PR #20, opened 2019-11-11
Fix memory leak in rav1e codec

AurelC2G:
The packet holding the decoded image was never freed.


joedrago:
https://github.com/AOMediaCodec/libavif/commit/8a6023cca8cfd985af79b2451aa916dba4626920

PR #24, opened 2019-11-19
Some fixes to cmake and support for find_package(libavif)

cryptomilk:
Here are some fixes for cmake and config files so other projects can find it using:

`find_package(libavif 0.4.7)`


cryptomilk:
I just checked for sanity as the libaom on my system is 1.0.0 and it is not compatible with libavif.

Also for compiler flag checking you should do it in a way that all compilers are supported. See e.g.

https://gitlab.com/cmocka/cmocka/blob/master/CompilerChecks.cmake

cryptomilk:
Took some iterations to fix appveyor but everything is green now.

joedrago:
There are some fantastic changes here. The only one I'm unsure on is enforcing BUILD_SHARED_LIBS. As I mentioned in my recent commit removing STATIC from add_library, I'd prefer to stay un-opinionated and leave CMake's defaults here.  Are there distro packaging concerns you have if that line is removed?

cryptomilk:
Normally you want shared libs by default. It makes sense resource wise and security wise.

joedrago:
Hmm, alright. I'm on board with shared-by-default, but I don't want to have libavif change a cmake option default, so I'll remove that line and make `AVIF_BUILD_STATIC` do more work. Starting the merge now.

darix:
distro people like cryptomilk and me prefer shared libraries :)

joedrago:
Alright, I merged in your changes and then reworked some of the static-vs-shared stuff a bit. The important takeaway is that it is still shared by default, and all of the find_package stuff survived my tweaking without any changes. I'm going to bump the libavif version soon, so you'll be able to use that new version number for whatever packaging scripts you've got cooking.

Keep those distro-packaging-related PRs coming! I don't claim to be an expert in that specific kind of voodoo, and I'm all for wider adoption.

joedrago:
`v0.4.8` is pushed and has the goods.

cryptomilk:
Normally projects set BUILD_SHARED_LIBS as an option. This is meant that way, see also the documentation for it https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

joedrago:
Oh, I didn't realize it was recommended / common practice to include your own default for it. I figured it would virally pollute parent CMake projects if a library project was added as a child project. 

I'll do some local testing and possibly reintroduce that to my CMakeLists, but seeing as functionally we've arrived at the same default behavior, this should no longer be a blocker, right?

Issue #9, opened 2019-09-04
Why not use the release version of libaom?


patrick22414:
I'm on Arch Linux, and `pacman -S aom` gives me `libaom v1.0.0-errata1`, which is the newest release of libaom.

However, libavif seems to be requiring a specific commit of libaom.
https://github.com/AOMediaCodec/libavif/blob/a78fa98a3ddbb45041e3f5eab3a2f15d8d374d75/ext/aom.cmd#L6

Building with `libaom v1.0.0-errata1` will result in compile error in `src/codec_aom.c`. So I have to run `ext/aom.cmd` and build libaom as well.

Is there a reason for not using the release version of libaom? At least in the release version of libavif?

joedrago:
I don't work directly on the libaom team and simply hadn't noticed that they made that tag 2 weeks ago. The checked-in SHA there was only from a month prior. As I note in aom.cmd, the SHA in there is just the most recent commit I've personally tested and known to work, and I'll update it periodically.

I've updated it just now to point at v1.0.0-errata1, fixed the compile issues related to the SHA update, and bumped the version to v0.3.7. It is going through the builders now, but assuming no new issues crop up, you should be able to use v0.3.7 along with your baked-in libaom on your distro now.

joedrago:
I think they've internally switched to using 16bpc all the time, which has broken libavif's decode path. I'm investigating now.

joedrago:
Oh god, I just realized what happened. The v1.0.0-errata1 version is ancient, and I missed the timestamp on that tag (2018!). I'm undoing all of this nonsense now.

joedrago:
Okay, I've successfully reverted these changes, but had to push a new build. 0.3.8 is effectively 0.3.6 in functionality, and I've flagged 0.3.7 as do-not-use. This will teach me to read GitHub issues right when I wake up in the mornings.

Let's start over here:

The v1.0.0-errata1 tag is very old, and libaom has had many improvements/fixes in the >12 months since the tag was made. I think the best bet here would be to update the PKGBUILD file for Arch to point at a fresher SHA, or build libaom locally with libavif (using aom.cmd and `-DAVIF_BUILD_AOM=1` on the CMake cmdline).

PR #25, opened 2019-11-19
codec: Support building with libaom 1.0.0

cryptomilk:
Support building with libaom 1.0.0

I haven't tested if it works yet ...


joedrago:
I recognize and appreciate the intent here, but I've [been down this road already](https://github.com/AOMediaCodec/libavif/issues/9), and there are a lot of issues with the generated files when using various configurations (I can't remember which). I'll pin libavif to support the next versioned release and let libaom support stay there, but I don't want to support random bugs from not having 18 months of good bugfixes, sorry.

Let's see what happens with this tracker first:

https://bugs.chromium.org/p/aomedia/issues/detail?id=2544

PR #28, opened 2019-11-21
cmake: Don't use the project version as library version

cryptomilk:
You should never use the project version as the SO version. This will sooner or later lead to trouble :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
       - cd ..
       - mkdir build
       - cd build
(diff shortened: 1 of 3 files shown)


cryptomilk:
Appveyor doesn't find ninja.

cryptomilk:
Looks good now:
```
lrwxrwxrwx 1 asn users     12 Nov 21 08:58 libavif.so -> libavif.so.0
lrwxrwxrwx 1 asn users     16 Nov 21 08:58 libavif.so.0 -> libavif.so.0.4.0
-rwxr-xr-x 1 asn users 190272 Nov 21 08:58 libavif.so.0.4.0
```

PR #32, opened 2019-11-26
Enable tiling with rav1e which allows multi threaded encoding!

cryptomilk:
Enable tiling with rav1e which allows multi threaded encoding.

Yay, more speed!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,8 @@
 /build*
+/obj*
 /ext/aom
 /ext/rav1e
 /ext/dav1d
+.clangd/
+cscope.*
+tags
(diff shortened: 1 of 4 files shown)


joedrago:
libavif most certainly already supports tiling with rav1e already. What does this behavior do instead of the (adjacent settings) `tileRowsLog2` / `tileColsLog2`?

joedrago:
I read `encoder.rs` to be sure this is a redunant PR, and `.tiles` is an alternate path to the same settings, with less control. Please switch your usage of this new code to using `tileRowsLog2` and `tileColsLog2` instead.

PR #17, opened 2019-10-06
AppVeyor: Use Visual Studio 2019 image

EwoutH:

PR #34, opened 2019-12-09
ext: Clone shallow, single and stable branch in aom.cmd, dav1d.cmd and rav1e.cmd

EwoutH:
Speedup `aom.cmd`, `dav1d.cmd` and `rav1e.cmd` in `/ext` massively by online cloning a single and shallow branch. dav1d and rav1e use the latest tagged (pre-)release, aom the latest build from master.

In Travis, the builds are approximately a minute faster.

### DIFF
--- a/ext/aom.cmd
+++ b/ext/aom.cmd
@@ -8,7 +8,7 @@
 : # If you're running this on Windows, be sure you've already run this (from your VC2017 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Auxiliary\Build\vcvars64.bat"
 
-git clone -n --shallow-since=2019-12-01 --single-branch https://aomedia.googlesource.com/aom && cd aom && git checkout aadb0b4b23b0aaa031c8aca15c9f86a42010acf6 && cd ..
+git clone -b libaom-v0.9.0 --depth 1 https://aomedia.googlesource.com/aom
 
 cd aom
 mkdir build.libavif
--- a/ext/aom.cmd
+++ b/ext/aom.cmd
@@ -8,7 +8,7 @@
 : # If you're running this on Windows, be sure you've already run this (from your VC2017 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Auxiliary\Build\vcvars64.bat"
 
-git clone -n https://aomedia.googlesource.com/aom && cd aom && git checkout 60d06c8721605ecd40c3b21bdd3685115b82ebf2 && cd ..
+git clone -n --shallow-since=2019-12-01 --single-branch https://aomedia.googlesource.com/aom && cd aom && git checkout aadb0b4b23b0aaa031c8aca15c9f86a42010acf6 && cd ..
 
 cd aom
 mkdir build.libavif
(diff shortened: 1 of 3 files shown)


joedrago:
Your shallow-since setup for libaom isn't going to work:

```
git clone -n --shallow-since=2019-12-01 --single-branch https://aomedia.googlesource.com/aom   && cd aom   && git checkout aadb0b4b23b0aaa031c8aca15c9f86a42010acf6   && cd ..
Cloning into 'aom'...
fatal: Server does not support --shallow-since
fatal: the remote end hung up unexpectedly
```

EwoutH:
Strange that it works on Travis but doesn't on AppVeyor. Could it be the Git version installed?

Maybe we could try with the Visual Studio 2019 image on AppVeyor.

joedrago:
I'm on board with shallow clones here, it should just be using a different method than `--shallow-since`.

Can you not do it the way you did dav1d and rav1e?

EwoutH:
The problem is libaom doesn't have that many releases, you need a recent tag to use the dav1d/rav1e method. The latest aom tag is `v1.0.0-errata1`, which is now 11 months old.

EwoutH:
I suspect it's actually the older Git version AppVeyor uses with the VS 2017 image (2.17). Travis uses Git 2.21 currently, and the VS 2019 image on AppVeyor also. So #17 should solve this issue.

joedrago:
I don't want to solve this with a newer image, and enforce 2019 requirements on people. I'm currently trying to get libaom to publish an interim tag so I can point people at that instead of a SHA.

Is there any reason you don't want to just use `--depth 1`?

EwoutH:
@joedrago For stability you want to build from a fixed commit, and if you clone with a depth=1 you can only checkout one commit earlier. With depth=50 you can checkout 50 commits earlier, etc. But eventually there are so many new commits that you can't checkout the commit you originally wanted anymore.

By cloning to a certain date you avoid this, because you can always make sure the date is before the commit you want to checkout.

joedrago:
They added a tag!

https://aomedia.googlesource.com/aom/+/refs/tags/libaom-v0.9.0

EwoutH:
Updated aom.cmd!

> `libaom-v0.9.0`

What a very weird tag btw (it should have been `v0.9.0` or `0.9.0`). Also, how the heck do you go from `v1.0.0` to `v0.9.0`.

joedrago:
I'm not completely sure, but from my understanding, they planned to keep the tags in sync with the AV1 spec version, so once that hit v1.0.0, the tags stopped coming. This is the "library version", which isn't quite 1.0 yet (I suppose). 

The important part for sure is that it is an official tag and it is from the last 24 hours, so I can stop getting requests to roll back to 18-months-old code. :-)

EwoutH:
It looks like they removed the tag again. Some people on Reddit where also getting compile errors.

joedrago:
Yes, the tag format had issues with their build system, so they are switching to something like v1.0.0-avif. I suspect the tag will land today at some point.

Where on Reddit?

EwoutH:
@joedrago Is there any chance we could get a bimonthly avif tag? So like `v1.0.0-errata1-avif2` this month, `v1.0.0-errata1-avif3` in April, etc.? libaom is still developing quite fast so it would be nice to have a fixed reference point we could test periodically.

joedrago:
I would also like that. I have no authority to make tags in libaom though (libavif is a separate project). I'll ask them about it.

EwoutH:
> I would also like that. I have no authority to make tags in libaom though (libavif is a separate project). I'll ask them about it.

Have you got an answer?

joedrago:
Yes, the last reply I got from them was that they are trying to finish up the next full release of the library, and once they have that out they will be releasing tags with a regular cadence. 

This appears to be their bug tracker for that release:

https://bugs.chromium.org/p/aomedia/issues/detail?id=2545&q=&can=2

PR #38, opened 2020-01-19
Unbreak build on Unix systems without /usr/lib/libdl.so

jbeich:
NetBSD and OpenBSD cannot use `-ldl` while [DragonFly](https://github.com/DragonFlyBSD/DragonFlyBSD/commit/2f60b541a1a5) and [FreeBSD](https://github.com/freebsd/freebsd/commit/84be924362c6) don't need it.


jbeich:
Why `-ldl` is required on Linux given libavif doesn't use `dlopen`?

joedrago:
I merged this, but not via the GitHub UI (as I did with the other), so it didn't catch my rebase. One of the underlying codecs requires dl (rav1e?). I could probably rearrange this dependency a bit in the future. Thanks for the fix though, I didn't realize CMake had that variable!

PR #44, opened 2020-02-07
Ext: Update to rav1e v0.3.0

EwoutH:


joedrago:
I'm on board with taking these kinds of PRs going forward, but can you add a line to the CHANGELOG as well when you do this?

cryptomilk:
@joedrago could you also do a release after merging this, so we have picture mode support?

EwoutH:
dav1d 0.6.0 will very likely be released this week, I will suggest tagging a release after that.

joedrago:
I like the idea of pushing a release after dav1d 0.6, we can wait for that.

My responses this week have been slow as my family and I all have the flu. Things should go back to normal response times later this week.

PR #47, opened 2020-02-10
Build local libaom with PIC enabled.

wantehchang:
Building libavif with a local libaom source tree in libavif/ext fails
with a linker error. Here are the steps to reproduce the bug.

  # In the top-level libavif directory, run the following commands.
  $ cd ext
  $ git clone -b v1.0.0-errata1-avif --depth 1 https://aomedia.googlesource.com/aom
  $ cd aom
  $ mkdir build.libavif
  $ cd build.libavif
  $ cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DENABLE_DOCS=0 -DENABLE_EXAMPLES=0 -DENABLE_TESTDATA=0 -DENABLE_TESTS=0 -DENABLE_TOOLS=0 ..
  $ ninja
  $ cd ../..
  $ cd ..
  $ mkdir build
  $ cd build
  $ cmake -G Ninja -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON ..
  $ ninja

The build fails while linking libavif.so.0.1.4:

[11/12] Linking C shared library libavif.so.0.1.4
FAILED: libavif.so.0.1.4
: && /usr/bin/cc -fPIC    -shared -Wl,-soname,libavif.so.0 -o libavif.so.0.1.4 CMakeFiles/avif.dir/src/avif.c.o CMakeFiles/avif.dir/src/colr.c.o CMakeFiles/avif.dir/src/mem.c.o CMakeFiles/avif.dir/src/rawdata.c.o CMakeFiles/avif.dir/src/read.c.o CMakeFiles/avif.dir/src/reformat.c.o CMakeFiles/avif.dir/src/stream.c.o CMakeFiles/avif.dir/src/utils.c.o CMakeFiles/avif.dir/src/write.c.o CMakeFiles/avif.dir/src/codec_aom.c.o  ../ext/aom/build.libavif/libaom.a -lm -pthread && :
/usr/bin/ld: ../ext/aom/build.libavif/libaom.a(decodeframe.c.o): relocation R_X86_64_PC32 against symbol `av1_sgr_params' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status

Fix this bug by adding -DCONFIG_PIC=1 to libaom's cmake command line to
compile libaom with position-independent code (PIC) enabled.


joedrago:
This isn't necessarily a bug as much as a failure to document, and perhaps a missing feature.

There are two supported build paths for libavif right now:

The default is to build with BUILD_SHARED_LIBS on and only enabling the AVIF_CODEC_* that you need, which will use the installed versions of that on your machine. This path is the default as it is the most likely to be used for *nix distributions, and should all build nicely. In this case, the ext/ directory is mostly just there as a nod to which codec versions are known-good, and it is expected that you have your own copies of them resident on your machine.

The alternative is to build with BUILD_SHARED_LIBS off, typically for embedded usage or (perhaps) Windows static builds, or simple testing in CI. In this mode, You can additionally enable AVIF_LOCAL_* and use static libs built in there, which plays nice with embedded/Win32 development as there might not be any notion of "system libs".

You appear to be building with BUILD_SHARED_LIBS on, but are trying to use AVIF_LOCAL_*, which I think isn't a configuration we should be supporting. If your intent is to statically link everything together, disable BUILD_SHARED_LIBS (as I mentioned in my previous paragraph) and things will link. If your intent is to use shared libs for everything, do a proper `make install` of the codec libraries so that they exist as shared libs on your machine, then don't enable AVIF_LOCAL_* in your build process. 

The path I believe you're going down right now would be trying to statically link codec implementations into libavif.so, which I think is bad. Distros are going to want to be able to independently update dav1d, rav1e, libaom, libgav1, etc, independent of libavif. Statically linking into libavif.so isn't the right answer.

I'll wait for your reply to this before closing this bug, in case I've misunderstood the issue.

wantehchang:
Hi Joe,

Thank you for the reply. I read the "Build Notes" and "Local Builds" sections in libavif/README.md and learned about the AVIF_CODEC_* and AVIF_LOCAL_* CMake options there. That file does not mention BUILD_SHARED_LIBS.

I just tried adding -DBUILD_SHARED_LIBS=OFF to the cmake command line:

  $ cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON ..

That solved the problem. Thank you!

I think the solution is to document BUILD_SHARED_LIBS, e.g., in the "Local Builds" section. I will close this pull request.

joedrago:
I agree with you. I've adjusted the README to make it clearer what BUILD_SHARED_LIBS role is and more tightly couple the AVIF_LOCAL_* variables to static builds.

PR #48, opened 2020-02-11
CI: Create initial GitHub Actions workflow

EwoutH:
A simple initial GitHub Actions setup that builds libavif with aom, dav1d and rav1e using gcc 8 on Ubuntu 18.04

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,4 +34,4 @@ jobs:
         cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON -DAVIF_CODEC_DAV1D=ON -DAVIF_LOCAL_DAV1D=ON -DAVIF_CODEC_RAV1E=ON -DAVIF_LOCAL_RAV1E=ON
     - name: Build libavif (make)
       working-directory: ./build
-      run: make -j ($(nproc)+1)
+      run: make -j $(($(nproc) + 1))
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,36 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+    env:
+      CC: gcc-8
+      CXX: g++-8
+    steps:
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.8'
+    - name: Install dependencies
+      run: |
+        sudo apt-get install -y ninja-build nasm gcc-8 g++-8
+        pip install --upgrade pip
+        pip install setuptools
+        pip install meson
+    - uses: actions/checkout@v2
+    - name: Setup aom
+      working-directory: ./ext
+      run: bash aom.cmd
+    - name: Setup dav1d
+      working-directory: ./ext
+      run: bash dav1d.cmd
+    - name: Setup rav1e
+      working-directory: ./ext
+      run: bash rav1e.cmd
+    - name: Prepare libavif (cmake)
+      run: |
+        mkdir build && cd build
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+      CC: gcc-8
+      CXX: g++-8
+    steps:
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.8'
+    - name: Install dependencies
+      run: |
+        sudo apt-get install -y ninja-build nasm gcc-8 g++-8
```
1480c1:
```suggestion
        DEBIAN_FRONTEND=noninteractive sudo apt-get update
        DEBIAN_FRONTEND=noninteractive sudo apt-get install -y ninja-build nasm gcc-8 g++-8
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Setup rav1e
+      working-directory: ./ext
+      run: bash rav1e.cmd
+    - name: Prepare libavif (cmake)
+      run: |
+        mkdir build && cd build
+        cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON -DAVIF_CODEC_DAV1D=ON -DAVIF_LOCAL_DAV1D=ON -DAVIF_CODEC_RAV1E=ON -DAVIF_LOCAL_RAV1E=ON
+    - name: Build libavif (make)
+      working-directory: ./build
+      run: make
```
1480c1:
```suggestion
      run: make -j4
```
or `-j2`

EwoutH: Isn't make multi-threaded by default nowadays?

https://www.gnu.org/software/make/manual/html_node/Parallel.html

> By default, there is no load limit.

.github/workflows/ci.yml
EwoutH: Thansk for the review! What does DEBIAN_FRONTEND=noninteractive exactly do?

1480c1: https://askubuntu.com/questions/556385/how-can-i-install-apt-packages-non-interactively
```none
noninteractive
      This  is  the anti-frontend. It never interacts with you at all,
      and makes the default answers be  used  for  all  questions.  It
      might  mail  error messages to root, but that's it; otherwise it
      is completely silent and unobtrusive,  a  perfect  frontend  for
      automatic installs. If you are using this front-end, and require
      non-default answers to questions, you will need to  preseed  the
      debconf  database;  see  the section below on Unattended Package
      Installation for more details.
```

Makes it so that it will never ask you anything

.github/workflows/ci.yml
1480c1: No, that only applies to `-l` to limit the total cpu usage of the  jobs.
Without `-j`, make should run sequentially. I will test it out regardless though

joedrago:
This seems pretty cool. It looks like you and 1480c1 are still figuring out some nuances with the script, but once you're satisfied with its state, I'll accept this PR.

.github/workflows/ci.yml
EwoutH: Got it. What do you think using `-j $(nproc)+1`?

1480c1: I don't think `$(nproc)+1` is valid bash

`$(($(nproc) + 1))` works though

1480c1: Note the double parentheses to tell bash to do arithmetic and the inner command substitution to supplement the output of nproc

EwoutH: Thanks!

EwoutH:
Good to go from my side!

We now have 3 CI systems up and running which build libavif, I think the next step is to add some tests.

joedrago:
Yeah, I'm considering harvesting my AVIF roundtrip IO testing from another project of mine (colorist). It takes a test image (some color bars) and encodes it in a bunch of different QPs, bpcs, etc, and then expects specific tolerances from each. I can recreate the important subset of that in libavif and we can run it in CI for regression tests.

PR #54, opened 2020-02-16
Use cmake find modules to look for library dependencies

### LINKED ISSUE #52: Build is broken with external dependencies
The CMake infrastructure currently picks up the header paths when building with external copies of the  dependent libraries but not the library paths thus the build fails.

cryptomilk:
This should fix #52


cmake/Modules/Findaom.cmake
```diff
@@ -0,0 +1,50 @@
...
+#  Distributed under the OSI-approved BSD License (the "License");
+#  see accompanying file Copyright.txt for details.
+#
+#  This software is distributed WITHOUT ANY WARRANTY; without even the
+#  implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#  See the License for more information.
+#=============================================================================
+#
+
+if (UNIX)
```
1480c1: PkgConfig is available on msys2 and other non-unix systems

cmake/Modules/Findaom.cmake
```diff
@@ -0,0 +1,50 @@
...
+  find_package(PkgConfig)
```
1480c1:
```suggestion
  find_package(PkgConfig QUIET)
```

cmake/Modules/Findaom.cmake
cryptomilk: That's why it has `if (UNIX)` and not `if (UNIX AND NOT WIN32)`!

1480c1: msys2 cmake and windows cmake does not set UNIX to true

1480c1: On msys2's mingw-w64 cmake
```cmake
message("Unix: ${UNIX}")
message("mingw: ${MINGW}")
message("WIN32: ${WIN32}")
if (UNIX)
    message("Unix")
else()
    message("Not Unix")
endif()
```
```bash
cmake -P temp.cmake
Unix:
mingw:
WIN32: 1
Not Unix
```

cmake/Modules/Findaom.cmake
```diff
@@ -0,0 +1,50 @@
...
+#  Distributed under the OSI-approved BSD License (the "License");
+#  see accompanying file Copyright.txt for details.
+#
+#  This software is distributed WITHOUT ANY WARRANTY; without even the
+#  implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#  See the License for more information.
+#=============================================================================
+#
+
+if (UNIX OR MINGW)
```
1480c1: Maybe it's better to word it this way, just remove the `if (UNIX OR MINGW)`, let the `if (PKG_CONFIG_FOUND)` handle if PkgConfig was found

PR #55, opened 2020-02-19
AppVeyor: Compile with dav1d, refactor

EwoutH:
Changes the AppVeyor configuration to compile with dav1d. Also cleans up unused code and configuration options.

The compile time increased from about 7 to 10 minutes, and avifenc and avifdec increased from 8,4 to 9,3 MB.

I removed the debug build since I don't think it provides much value over the release builds. If we want to test a MSVC debug build I would suggest doing that in GitHub Actions.

Also, would you be open for a deploy configuration that pushes the libavif zip, avifenc and avifdec to GitHub, so that they can be downloaded from the release page (like [rav1e](https://github.com/AOMediaCodec/libavif/releases))?


joedrago:
Once the CI checks pass, I'll merge this PR.

Let me think about the deploy thing. I'm not against it in general, I just need to think about what that might sign me up for, support-wise.

EwoutH:
@joedrago Have you given the deploy functionality thought?

joedrago:
The end goal here is to copy the AppVeyor artifacts to the GitHub releases page, right? I'm fine with that idea if it is easy to setup. What would that entail?

EwoutH:
Yes, I opened #111 so that you could see what it looks like.

PR #57, opened 2020-02-20
Add libgav1 support (decoder only).

wantehchang:
Only the local, static library build (cmake -DAVIF_CODEC_LIBGAV1=ON
-DAVIF_LOCAL_LIBGAV1=ON) has been tested. Since libgav1 is written in
C++ internally, the executable programs need to be linked with the C++
compiler.

src/codec_libgav1.c is a port of src/codec_aom.c (the decoder portions)
and src/codec_dav1d.c.

Known issues:
* Discovery of the libgav1 package on the system is not implemented,
  i.e., cmake/Modules/Findlibgav1.cmake is yet to be written.
* libgav1 does not support the "output all layers" option yet.


joedrago:
This looks pretty solid. I do have a few notes though:

* Did you want to take a shot at writing the Find cmake module for libgav1? I can probably give it a shot if not.

* It doesn't appear libgav1 can build on Windows right now as `threadpool.cc` doesn't `#include <algorithm>`, but uses `std::min()`. Any chance and tweaking that and updating the SHA in libgav1.cmd? Also, are there no regular tagged releases we can use for libgav1? It'd be nice to follow the shallow clone pattern the other .cmd files use.

Also for Windows local static compilation support, CMakeLists.txt requires this additional tweak:

```
index 84a52d6..2e4846e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,7 +147,10 @@ if(AVIF_CODEC_LIBGAV1)
     if(AVIF_LOCAL_LIBGAV1)
         set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/libgav1/build/libgav1.a")
-            message(FATAL_ERROR "libavif: ${LIB_FILENAME} is missing, bailing out")
+            set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/libgav1/build/libgav1.lib")
+            if(NOT EXISTS "${LIB_FILENAME}")
+                message(FATAL_ERROR "libavif: ${LIB_FILENAME} is missing, bailing out")
+            endif()
         endif()
 
         include_directories(
```

wantehchang:
Joe,

Thanks for the suggestions! In the new version of this pull request, I added cmake/Modules/Findlibgav1.cmake, but I don't know how to test it.

I also added Windows local static compilation support to CMakeLists.txt. I used the approach from the AVIF_LOCAL_AOM block because it has a better error message. The change you suggested would report "libavif: .../libgav1.lib is missing, bailing out" all all platforms.

Libgav1 should build on Windows. I will fix the compilation error of threadpool.cc that you reported. I suspect it may depend on the version of Visual C++.

I will find out about creating a tagged release for libgav1.

By the way, how do I edit the commit message for this pull request? It needs to be updated regarding Findlibgav1.cmake.

jzern:
> Joe,
> 
> Thanks for the suggestions! In the new version of this pull request, I added cmake/Modules/Findlibgav1.cmake, but I don't know how to test it.
> 

As part of libgav1__compile [1] a quick test is used [2].

[1] https://build.webmproject.org/jenkins/job/libgav1__compile/
[2]
cat > CMakeLists.txt <<EOF
cmake_minimum_required(VERSION 3.7.1)

project(libgav1 CXX)

find_package(libgav1)
if (NOT libgav1_FOUND)
  message(FATAL_ERROR "libgav1 package not found")
endif ()
message("libgav1_FOUND: \${libgav1_FOUND}")
message("libgav1_INCLUDE_DIRS: \${libgav1_INCLUDE_DIRS}")
message("libgav1_LIBRARIES: \${libgav1_LIBRARIES}")
message("LIBGAV1_INCLUDE_DIRS: \${LIBGAV1_INCLUDE_DIRS}")
message("LIBGAV1_LIBRARIES: \${LIBGAV1_LIBRARIES}")
EOF
cmake . -DCMAKE_PREFIX_PATH="${BUILD_DIR}/libgav1-install/usr/local/share/cmake"

wantehchang:
Please take another look. I've made the following changes to the pull request.

1. Use CMAKE_STATIC_LIBRARY_SUFFIX in CMakeLists.txt. Benbuck, thank you for the suggestion!

2. Added Ninja as a build requirement for libgav1 in README.md. Libgav1 can also be built with make and Visual Studio, but the AVIF_LOCAL_LIBGAV1 code in CMakeLists.txt assumes the build output directory structure used by Ninja.

3. Set the number of threads to 2 in codec_libgav1.c. The default value is 1, which always performs worse than 2 on a multi-core system. If we have to use a hardcoded value, 2 is the safest value.

I also wanted to add another known issue (which I believe also applies to codec_aom.c). Unlike codec_dav1d.c, codec_libgav1.c cannot add a reference to the output frame buffer. The output frame buffer will be invalid after the next Libgav1DecoderDequeueFrame(). So I believe the code in the following "else" branch won't work:

    if (nextFrame) {
        codec->internal->gav1Image = nextFrame;
        codec->internal->colorRange = (nextFrame->color_range == kLibgav1ColorRangeStudio) ? AVIF_RANGE_LIMITED : AVIF_RANGE_FULL;
    } else {
        if (codec->decodeInput->alpha && codec->internal->gav1Image) {
            // Special case: reuse last alpha frame
        } else {
            return AVIF_FALSE;
        }
    }

To fix this will require providing frame buffer callbacks for libgav1. I will do that in a follow-up pull request. This pull request defends against this scenario by setting codec->internal->gav1Image to NULL before each Libgav1DecoderEnqueueFrame() call.

wantehchang:
I analyzed the code in codec_libgav1.c and codec_aom.c again. I am no longer sure that the "known issue" I described in my last comment is correct. If I understand the code correctly, the "else" branch in the code snippet can only be reached if:

1) codec->internal->inputSampleIndex >= codec->decodeInput->samples.count, or
2) Libgav1DecoderDequeueFrame() succeeded but returned a null nextFrame.

In case #1, we do not call Libgav1DecoderDequeueFrame(), so codec->internal->gav1Image remains valid.

Case #2 is problematic. It can happen if the frame being decoded is not in the operating point and is dropped (see AV1 spec, Section 5.3.1, page 27, the drop_obu() call in the open_bitstream_unit( sz ) function. I don't know if this case can happen when the bitstream is an AVIF image.

joedrago:
Why are you hardcoding a thread count? In other codec implementions, they just honor `encoder->maxThreads`.

joedrago:
Oh whoops, as soon as I hit Comment I realized the issue. I guess I don't expose maxThreads in avifDecoder, hm? I should probably add that.

joedrago:
That else case is intended to be when nextFrame doesn't exist, meaning the previous for loop ran out of samples, but never encountered an error. The special check in the else case is trying to handle the situation where an image might have a color AV1 payload of 10 images (say) but an alpha AV1 payload of less frames. In that case, this implementation will simply continue to use the last decoded alpha frame as its alpha channel for all subsequent decoded color frames.

I'm not seeing any referencing issues in the AOM implementation, but I'm unfamiliar with how gav1 does its refcounts.

wantehchang:
On Android devices, libgav1 gets the best performance when the number of threads in Libgav1DecoderSettings is equal to the number of "performance" processor cores. It requires some code to extract that info from the /sysfs filesystem. (The get_nprocs() and get_nprocs_online() functions return a count that also includes the "efficiency" processor cores.) So in this initial version I decided to use a hardcoded value.

joedrago:
I think it is fine to leave it at 2 until I plumb through a maxThreads in avifDecoder.

joedrago:
Can you change the copyright on your Find cmake file to Google? Unless @cryptomilk helped with it and I missed that.

joedrago:
Everything looks good, after the copyright thing. Once that is resolved, I'll merge the PR.

wantehchang:
Joe: Thank you for explaining that else case. I think codec_libgav1.c will be able to handle that case correctly.

I have changed the copyright on Findlibgav1.cmake to Google. If you can modify the commit message before you merge the PR, please change the first "known issue" from:

* Discovery of the libgav1 package on the system is not implemented,
  i.e., cmake/Modules/Findlibgav1.cmake is yet to be written.

to:

* cmake/Modules/Findlibgav1.cmake is not tested.

or just delete that known issue.

Thank you for all your help with the PR!

PR #60, opened 2020-02-21
Have avif_example1.c return 1 from main on failure

wantehchang:
In Unix, a program exits with status 0 on success and with a nonzero
status on failure. If any function call fails or any decoded pixel does
match the original, have avif_example1.c return 1 from the main()
function.

Note: Following current practice in libavif, I assume avifAlloc() always
succeeds and do not check for memory allocation failures.

### DIFF
--- a/examples/avif_example1.c
+++ b/examples/avif_example1.c
@@ -10,6 +10,7 @@ int main(int argc, char * argv[])
 {
     (void)argc;
     (void)argv;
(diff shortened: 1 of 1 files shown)


wantehchang:
Note that I use goto statements on error to avoid further indenting some code. Please let me know if you do not want goto statements in avif_example.c.

PR #62, opened 2020-02-21
Minor edit of ext/aom.cmd.

wantehchang:
If CMake's AVIF_CODEC_AOM and AVIF_LOCAL_AOM options are both ON,
libavif does not build libaom. Therefore the first sentence in aom.cmd:

    If you want libavif to build libaom for you, ...

is not accurate. Change it to:

    If you want to use a local build of libaom, ...

which is the form used in the other <codec>.cmd files.


wantehchang:
I hope I didn't misunderstand something. I have to build ext/libaom manually before I run libgav1's cmake command.

PR #66, opened 2020-02-22
Handle monochrome images correctly.

ledyba-z:
I found that `avifImageYUVToRGB` will always returns `AVIF_RESULT_REFORMAT_FAILED` for monochrome images.

And also, `avifImageYUVToRGB8Mono` and `avifImageYUVToRGB16Mono` will returns green images, because they convert monochrome images with `Cr=Cb=-0.5`, which means green.
 
Please take a look!


joedrago:
Ah. this is a nice and simple fix. Does it work for you? If so, I'll merge the PR.

ledyba-z:
Thank you to merge!
Yeah, It works in our android (proprietary) apps.

PR #68, opened 2020-02-24
More cmake fixes related to commit 382345b.

### LINKED ISSUE #58: Findaom.cmake and Findrav1e.cmake mention dav1d
Release tag: v0.5.6

Findaom.cmake and Findrav1e.cmake mention dav1d in the following code snippet:

find_package(PkgConfig QUIET)
if (PKG_CONFIG_FOUND)
    pkg_check_modules(_DAV1D dav1d)
endif (PKG_CONFIG_FOUND)

It looks like copy-and-paste errors. If it is correct, it would be good to add a comment to explain why we need to call pkg_check_modules(_DAV1D dav1d) when trying to discover the aom or rav1e package on the system.

### LINKED ISSUE #59: CMakeLists.txt uses DAV1D_INCLUDEDIR but Finddav1d.cmake defines DAV1D_INCLUDE_DIR
Release tag: v0.5.6

Note: I wrote this bug report using DAV1D as an example, but the bug also applies to AOM and RAV1E.

I am not very familiar with CMake, so I may have misunderstood how Finddav1d.cmake is used. While reading CMakeLists.txt and Finddav1d.cmake I spotted three things that seem wrong.

1. CMakeLists.txt uses DAV1D_INCLUDEDIR but Finddav1d.cmake defines DAV1D_INCLUDE_DIR. Should CMakeLists.txt use DAV1D_INCLUDE_DIR instead?

2. Finddav1d.cmake defines both DAV1D_LIBRARY and DAV1D_LIBRARIES, but the comment at the top of the file does not mention DAV1D_LIBRARY:

```
  # - Try to find dav1d
  # Once done this will define
  #
  #  DAV1D_FOUND - system has dav1d
  #  DAV1D_INCLUDE_DIR - the dav1d include directory
  #  DAV1D_LIBRARIES - Link these to use dav1d
```

3. The comment I quoted above suggests DAV1D_LIBRARIES should be used, but CMakeLists.txt uses DAV1D_LIBRARY:

        set(AVIF_CODEC_LIBRARIES ${AVIF_CODEC_LIBRARIES} ${DAV1D_LIBRARY})

Should CMakeLists.txt use DAV1D_LIBRARIES instead?

wantehchang:
* cmake: Fix typo in Findlibgav1.cmake

Fixes #58

* cmake: Change "_RAV1e" to "_RAV1E" in Findrav1e.cmake

Fixes #58

* cmake: Fix typo for aom, libgav1, and rav1e include dirs

Fixes #59


wantehchang:
I have now tested Findlibgav1.cmake.

I also figured out why the typos did not seem to break the build. GCC searches for headers in /usr/local/include by default. If the package is installed in /usr/local, then its headers can still be found by GCC without an explicit -I/usr/local/include flag.

PR #71, opened 2020-02-26
Simplify image->depth checks in avifEncoderWrite()

wantehchang:
avifEncoderWrite() requires that image->depth must be 8, 10, or 12 on
entry. So inside the function, !image->depth cannot possibly be true,
and (image->depth == 8) || (image->depth == 10) can be simplified to
image->depth <= 10 or image->depth != 12.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -116,7 +116,7 @@ avifResult avifEncoderWrite(avifEncoder * encoder, avifImage * image, avifRWData
     // -----------------------------------------------------------------------
     // Reformat pixels, if need be
 
-    if (!image->width || !image->height || !image->depth) {
+    if (!image->width || !image->height) {
         result = AVIF_RESULT_NO_CONTENT;
         goto writeCleanup;
     }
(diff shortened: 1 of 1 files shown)


src/write.c
```diff
@@ -165,7 +165,7 @@ avifResult avifEncoderWrite(avifEncoder * encoder, avifImage * image, avifRWData
     avifRWStreamWriteChars(&s, "avif", 4);                         // unsigned int(32) compatible_brands[];
     avifRWStreamWriteChars(&s, "mif1", 4);                         // ... compatible_brands[]
     avifRWStreamWriteChars(&s, "miaf", 4);                         // ... compatible_brands[]
-    if ((image->depth == 8) || (image->depth == 10)) {             //
+    if (image->depth <= 10) {                                      //
```
bnason-nf: This isn't equivalent because the new code allows a bit depth of 9, right?

wantehchang: Hi Benbuck,

This pull request modifies only the avifEncoderWrite() function. The first few lines of the function look like this:

avifResult avifEncoderWrite(avifEncoder * encoder, avifImage * image, avifRWData * output)
{
    if ((image->depth != 8) && (image->depth != 10) && (image->depth != 12)) {
    }

Therefore, the rest of this function can assume image->depth is 8, 10, or 12.

joedrago:
I'm on board with killing that extraneous depth check in the beginning, but I really do prefer the explicit checks for 8 and 10 in the second case.

wantehchang:
> I'm on board with killing that extraneous depth check in the beginning,
> but I really do prefer the explicit checks for 8 and 10 in the second case.

Done. Could you edit the commit message for me? Please delete the following part:

  and (image->depth == 8) || (image->depth == 10) can be simplified to
  image->depth <= 10 or image->depth != 12.

Thanks!

PR #75, opened 2020-02-28
Create a static library for files in apps/shared.

wantehchang:
The files in apps/shared, y4m.c and avifutil.c, are compiled repeatedly
for the avifdec, avifenc, and aviftest targets. Collect them into a
static library so that they are compiled only once.

I named the static library "avif_apps". Suggestions for a better name
are welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,25 +268,33 @@ endif()
 option(AVIF_BUILD_APPS "Build avif apps." OFF)
 option(AVIF_BUILD_TESTS "Build avif tests." OFF)
 if(AVIF_BUILD_APPS OR AVIF_BUILD_TESTS)
-    add_library(avif_apps STATIC
-        apps/shared/y4m.c
-        apps/shared/avifutil.c
-    )
-    target_link_libraries(avif_apps PRIVATE avif)
-    target_include_directories(avif_apps PUBLIC apps/shared)
+    add_library(y4m OBJECT apps/shared/y4m.c)
+endif()
+if(AVIF_BUILD_APPS)
+    add_library(avifutil OBJECT apps/shared/avifutil.c)
 endif()
 
 if(AVIF_BUILD_APPS)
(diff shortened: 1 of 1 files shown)


1480c1:
I recommend using an object library instead https://cmake.org/cmake/help/v3.5/command/add_library.html#object-libraries

wantehchang:
Christopher Degawa wrote:
> I recommend using an object library instead https://cmake.org/cmake/help/v3.5/command/add_library.html#object-libraries

Christopher: Thank you for the suggestion. I am not familiar with CMake's object libraries. When is an object library preferred over a static library?

Joe: If you prefer an object library, I can upload a new version of the pull request that uses an object library. (I already got it working.)

1480c1:
An object library is prefered if there's no specific reason to have a library file
In this case, since y4m.c and avifutil.c are not necessarily part of the libavif.(ext) but are both needed by separate targets ( `avifdec` and `aviftest` and `avifenc`) the object library will provide the .o files to all of them, without producing the intermediary .a file

joedrago:
I agree with @1480c1 that if we're going to bother doing this kind of cleanup at all, that an object library is correct. That said, compiling those two files a few times seems like a minor concern for me either way.

I'm fine with just leaving this one alone (no changes), honestly. Do you feel strongly about it, @wantehchang ?

wantehchang:
Christopher: Thank you for the explanation.

Joe: Thanks for the answer. I've updated the pull request to build y4m.c and avifutil.c as CMake object libraries. I leave the decision of whether this needs fixing to you. I mainly wanted to report the issue and contribute a solution.

wantehchang:
The change of include_directories(apps/shared) to target_include_directories(target PRIVATE apps/shared) can be turned into its own pull request if we don't want to fix the redundant compilation of y4m.c and avifutil.c.

wantehchang:
I messed up my git branch while trying to rebase it. I will close this pull request.

PR #76, opened 2020-02-28
Use just -std=gnu99, not -std=c99 -std=gnu99.

wantehchang:
If multiple -std flags are specified, I believe only the last one takes
effect. Therefore -std=c99 -std=gnu99 is equivalent to just -std=gnu99.
I found a Stack Overflow article on this issue:
https://stackoverflow.com/questions/40563269/passing-multiple-std-switches-to-g

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,10 +19,6 @@ project(libavif LANGUAGES C CXX VERSION 0.5.6)
 set(LIBRARY_VERSION "0.1.6")
 set(LIBRARY_SOVERSION "0")
 
-if(CMAKE_C_COMPILER_ID MATCHES "GNU")
-  add_definitions(-std=c99) # Enforce C99 for gcc
-endif()
-
 option(BUILD_SHARED_LIBS "Build shared avif library" ON)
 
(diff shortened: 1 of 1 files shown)


wantehchang:
Here is a snippet of the relevant code in CMakeLists.txt, which shows how we end up using -std=c99 -std=gnu99 for GCC:

if(CMAKE_C_COMPILER_ID MATCHES "GNU")
  add_definitions(-std=c99) # Enforce C99 for gcc
endif()
...
# Enable all warnings
if(CMAKE_C_COMPILER_ID MATCHES "Clang")
  MESSAGE(STATUS "libavif: Enabling warnings for Clang")
  add_definitions(
    -Weverything
    -Werror
    ...
    -Wno-undef
  )
elseif(CMAKE_C_COMPILER_ID MATCHES "GNU")
  add_definitions(-std=gnu99) # Enforce C99 for gcc
  MESSAGE(STATUS "libavif: Enabling warnings for GCC")
  add_definitions(-Werror -Wall -Wextra)
elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")

wantehchang:
The code snippet also shows we are not using any -std flag for Clang. I am not sure if that is intentional.

PR #78, opened 2020-02-29
Miscellaneous header file cleanup.

wantehchang:
1. Add #ifndef header guard to apps/shared/avifutil.h and
apps/shared/y4m.h.

2. Do not install the internal header file include/avif/internal.h.
It is only used by src/*.c. (It would be good to move it to the src/
directory.)


wantehchang:
I named the #ifndef header guard macros based on the full pathnames, e.g., LIBAVIF_APPS_SHARED_AVIFUTIL_H for libavif/apps/shared/avifutil.h. This naming convention is recommended by Google's C++ Style Guide. Please let me know if you prefer alternative header guard macro names.

PR #83, opened 2020-03-03
cmake: Provide a pkgconfig file

### LINKED ISSUE #80: Pkg-config file for libavif
libavif should provide a pkg-config file.

cryptomilk:
Fixes #80

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -313,6 +313,8 @@ if(AVIF_BUILD_TESTS)
     )
 endif()
 
+configure_file(libavif.pc.cmake ${CMAKE_CURRENT_BINARY_DIR}/libavif.pc @ONLY)
+
 if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
     install(TARGETS avif
         EXPORT ${PROJECT_NAME}-config
@@ -331,6 +333,9 @@ if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
         install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config-version.cmake
                 DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME})
     endif()
+
+    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/libavif.pc
+            DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
 endif()
 if(NOT SKIP_INSTALL_HEADERS AND NOT SKIP_INSTALL_ALL)
     install(FILES include/avif/avif.h include/avif/internal.h
--- a/libavif.pc.cmake
+++ b/libavif.pc.cmake
@@ -1,0 +1,10 @@
+prefix=@CMAKE_INSTALL_FULL_PREFIX@
+exec_prefix=${prefix}/bin
+libdir=${prefix}/lib
+includedir=${prefix}/include
+
+Name: @PROJECT_NAME@
+Description: Library for encoding and decoding .avif files
+Version: @PROJECT_VERSION@
+Libs: -L${libdir} -lavif
+Cflags: -I{includedir}


CMakeLists.txt
```diff
@@ -313,6 +313,8 @@ if(AVIF_BUILD_TESTS)
     )
 endif()
 
+configure_file(libavif.pc.cmake ${CMAKE_CURRENT_BINARY_DIR}/libavif.pc)
```
1480c1: I would recommend to use `@ONLY` and make the PC file using the variables in the `@PROJECT_NAME@` style to make sure we can use `${var}` in the shell script sense

libavif.pc.cmake
```diff
@@ -0,0 +1,5 @@
+Name: ${PROJECT_NAME}
```
1480c1: It's customary to start with
```bash
prefix=@CMAKE_INSTALL_PREFIX@
exec_prefix=${prefix}/bin
libdir=${prefix}/lib
includedir=${prefix}/include
```
in case someone installs it to a place other than what's expected.

This is also where it would be helpful to use `@ONLY` so the end distributor or user can sed the `'s#prefix=.*#prefix=/MyPrefix#'` in case they want to install the end libraries in some place other than the one specified to cmake without needing to sed the rest of the lines

libavif.pc.cmake
```diff
@@ -0,0 +1,5 @@
...
+Description: Library for encoding and decoding .avif files
+Version: ${PROJECT_VERSION}
+Libs: -L${CMAKE_INSTALL_FULL_LIBDIR} -lavif
+Cflags: -I${CMAKE_INSTALL_FULL_INCLUDEDIR}
```
1480c1:
```suggestion
Version: @PROJECT_VERSION@
Libs: -L@CMAKE_INSTALL_FULL_LIBDIR@ -lavif
Cflags: -I@CMAKE_INSTALL_FULL_INCLUDEDIR@
```

libavif.pc.cmake
```diff
@@ -0,0 +1,5 @@
+Name: ${PROJECT_NAME}
+Description: Library for encoding and decoding .avif files
```
1480c1: Perhaps you can try to add a `Requires:` and `Libs.private:` and other sections for when linking and compiling libavif etc so we can tell which encoder and decoder libavif was linked with

CMakeLists.txt
cryptomilk: Done, thanks.

libavif.pc.cmake
cryptomilk: I think this is a task for a package manager to make sure dependencies are correctly installed :-)

1480c1: I think this would be a task for the cmake list to put the line for
```bash
Requires: aom
```
if compiled with `AVIF_CODEC_AOM` instead of the packager to append that line every time to be able to use the pkg-config file for other projects since without it, if the user wants to compile something with libaif, they have to guess which one libaif was compiled with and pass `$(pkg-config --libs libavif [aom rav1e or dav1d])` to the libraries every time they want to compile instead of letting autoconf or others to find the libraries

1480c1: I'm assuming it's something like
```cmake
if(AVIF_CODEC_AOM)
    set(pkg_config_requires "aom ${pkg_config_requires}")
endif()
if(AVIF_CODEC_DAV1D)
    set(pkg_config_requires "dav1d ${pkg_config_requires}")
endif()
if(AVIF_CODEC_RAV1E)
    set(pkg_config_requires "rav1e ${pkg_config_requires}")
endif()
# Idk what it is for libgav1
```
and adding to the pc file
```pkg-config
Requires: @pkg_config_requires@
```

1480c1: I'm not entirely sure if it belongs in `Requires:` or `Requires.private:` since I mainly build using static libs and almost never shared and am mainly in the position of needing to sed almost all of the pc files since they never declare their dependencies properly.

joedrago:
Thanks (in advance) to @cryptomilk for jumping on this one. I almost explicitly asked you to help with this, but I appreciate you jumping on it. 

Is the discussion on this finished, or is there still something to work out? If @cryptomilk is satisfied with its current state, I'll merge it.

libavif.pc.cmake
cryptomilk: What if aom is statically linked into libavif? Then the Require would show up ...

cryptomilk: Also I don't see why we need a require for building against libavif. We do not need any header from aom, rav1e or dav1d.

1480c1: If libavif is statically linked to libaom, will libavif not require libaom when linking an executable?

1480c1: Compiling libavif.a myself and running nm shows that the aom, dav1d, and rav1e symbols as undefined

1480c1: attempting to compiling `avif_examples.c` manually `$CC ../examples/avif_example1.c -I../include -L. -lavif` (similar to what would be produced by `$(pkg-config --libs libavif)` when using the pkg-config file here)
produces
```none
ld.exe: ./libavif.a(codec_dav1d.c.o):codec_dav1d.c:(.text+0x22): undefined reference to `dav1d_close'
...
ld.exe: ./libavif.a(codec_rav1e.c.o):codec_rav1e.c:(.text+0x99): undefined reference to `rav1e_config_default'
...
ld.exe: ./libavif.a(codec_aom.c.o):codec_aom.c:(.text+0x25): undefined reference to `aom_codec_destroy'
...
```
(stripped down to reduce clutter)

joedrago:
I'll just merge this. @cryptomilk can throw another PR at me if he needs further changes.

PR #93, opened 2020-03-06
avifdec/avifenc: Check for file write error.

wantehchang:


apps/shared/y4m.c
```diff
@@ -375,9 +376,13 @@ avifBool y4mWrite(avifImage * avif, const char * outputFilename)
...
 
-    fprintf(f, "YUV4MPEG2 W%d H%d F25:1 Ip A0:0 %s %s\n", avif->width, avif->height, y4mHeaderFormat, rangeString);
+    avifBool success = AVIF_TRUE;
+    if (fprintf(f, "YUV4MPEG2 W%d H%d F25:1 Ip A0:0 %s %s\nFRAME\n", avif->width, avif->height, y4mHeaderFormat, rangeString) < 0) {
+        fprintf(stderr, "Cannot write to file: %s\n", outputFilename);
+        success = AVIF_FALSE;
+        goto cleanup;
+    }
 
-    fprintf(f, "FRAME\n");
```
wantehchang: I merged this fprintf() statement into the previous fprintf() statement, so that we need to check for the failure of just one fprintf() call.

PR #98, opened 2020-03-08
Copy NCLX color profile.

ledyba-z:
It is forgot to copy NCLX color profiles (from AV1 Sequence header)  from the first tile to `decoder->image`.

Please take a look!

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -563,6 +563,7 @@ static avifBool avifDataFillImageGrid(avifData * data,
     unsigned int tileHeight = firstTile->image->height;
     unsigned int tileDepth = firstTile->image->depth;
     avifPixelFormat tileFormat = firstTile->image->yuvFormat;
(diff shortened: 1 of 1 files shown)


joedrago:
This is a good catch, I am missing this, but I think there is a minor issue here. The intent here is to allow the nclx from the tile to come through *as a fallback* to the AVIF container's `colr` box. The call to `avifImageSetProfileNCLX()` should only happen if the current `decoder->image->profile` is set to NONE. If it is already set, whatever is in there wins, I think.

Come to think of it, I should probably set it back to none at the top of `avifDecoderReset()` too.

ledyba-z:
OK, I added the code that checks `decoder->image->profileFormat == AVIF_PROFILE_FORMAT_NONE`.

Please take another look!

ledyba-z:
Thanks!

PR #102, opened 2020-03-09
avifenc: Exit with status 1 on syntax error.

wantehchang:
On syntax error, the main() function should return 1 after printing the
syntax help message. The only time main() should return 0 after printing
the syntax help message is when -h or --help is specified.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -17,7 +17,7 @@
     }                                                                 \
     arg = argv[++argIndex]
 
-static int syntax(void)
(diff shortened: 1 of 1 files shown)


wantehchang:
Joe: I should have taken care of this as part of pull request #97. I didn't expect you would merge pull requests last night.

joedrago:
No problem; I haven't pushed a new version yet, and I've been sufficiently vague in the changelog about these tweaks. If it survives the CI, I'll just merge it in.

PR #105, opened 2020-03-09
Miscellaneous cleanup related to UV width, height.

wantehchang:
Clean up the calculations of UV width and height. If the chroma shift is
1, we need to add 1 to width or height before right shifting by 1. This
can be written in a branch-free form.

Also clean up some code in src/codec_aom.c:
1. The type of the 'flags' parameter of aom_codec_enc_init() is
aom_codec_flags_t.
2. The memset() calls in a for loop can be merged because the row
strides are contiguous.

### DIFF
--- a/src/codec_rav1e.c
+++ b/src/codec_rav1e.c
@@ -134,10 +134,7 @@ static avifBool rav1eCodecEncodeImage(avifCodec * codec, avifImage * image, avif
     if (alpha) {
         rav1e_frame_fill_plane(rav1eFrame, 0, image->alphaPlane, image->alphaRowBytes * image->height, image->alphaRowBytes, byteWidth);
     } else {
-        uint32_t uvHeight = image->height >> yShift;
-        if (uvHeight < 1) {
-            uvHeight = 1;
-        }
+        uint32_t uvHeight = (image->height + yShift) >> yShift;
         rav1e_frame_fill_plane(rav1eFrame, 0, image->yuvPlanes[0], image->yuvRowBytes[0] * image->height, image->yuvRowBytes[0], byteWidth);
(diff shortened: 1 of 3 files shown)


src/codec_rav1e.c
```diff
@@ -134,10 +134,7 @@ static avifBool rav1eCodecEncodeImage(avifCodec * codec, avifImage * image, avif
     if (alpha) {
         rav1e_frame_fill_plane(rav1eFrame, 0, image->alphaPlane, image->alphaRowBytes * image->height, image->alphaRowBytes, byteWidth);
     } else {
-        uint32_t uvHeight = image->height >> yShift;
-        if (uvHeight < 1) {
-            uvHeight = 1;
-        }
```
wantehchang: I think the new code does not need this for loop, assuming image->height is nonzero. In the new code, if image->height is nonzero, then uvHeight is nonzero.

wantehchang:
Joe: It would be good if you could review these >> expressions in the avifDataFillImageGrid() function in src/read.c:

                // UV
                widthToCopy >>= formatInfo.chromaShiftX;
                heightToCopy >>= formatInfo.chromaShiftY;
                size_t uvColOffset = yaColOffset >> formatInfo.chromaShiftX;
                size_t uvRowOffset = yaRowOffset >> formatInfo.chromaShiftY;
                size_t uvRowBytes = yaRowBytes >> formatInfo.chromaShiftX;

I don't know if any of them needs to be changed. I suspect "heightToCopy >>= formatInfo.chromaShiftY;" may needs to be changed.

joedrago:
You might be right on the avifDataFillImageGrid() math, I'll have to think about it. There just aren't a lot of grid images in the wild to test with, especially not ones with odd height dimensions.

PR #107, opened 2020-03-10
Use comparison results as avifBool values directly

wantehchang:
Omit the explicit uses of AVIF_TRUE and AVIF_FALSE to convert comparison
results to avifBool.


wantehchang:
Please feel free to reject any of these changes. Thanks.

PR #108, opened 2020-03-11
cmake: Fix the SOVERSION which needs to be 2

cryptomilk:
Fix the SOVERSION which needs to be 2.


joedrago:
I can bump the library version if you want a new tag to use for packaging. A new version would also bring PNG support to avifenc/avifdec (if you include those in the RedHat package). Yesno?

cryptomilk:
That would be great, else I would add that patch ...

PR #114, opened 2020-03-17
Set dav1dSettings.frame_size_limit to avoid OOM.

wantehchang:
Set dav1dSettings.frame_size_limit to avoid the out-of-memory errors in
avif_decode_fuzzer, which is run with the -rss_limit_mb=2560
command-line option.

BUG=oss-fuzz:20645

### DIFF
--- a/src/codec_dav1d.c
+++ b/src/codec_dav1d.c
@@ -201,5 +201,7 @@ avifCodec * avifCodecCreateDav1d(void)
     codec->internal = (struct avifCodecInternal *)avifAlloc(sizeof(struct avifCodecInternal));
     memset(codec->internal, 0, sizeof(struct avifCodecInternal));
     dav1d_default_settings(&codec->internal->dav1dSettings);
+    // Set a maximum frame size limit to avoid OOM'ing fuzzers.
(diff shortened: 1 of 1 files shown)


wantehchang:
Joe:

I first propose that we set a hardcoded value of frame_size_limit for dav1d for simplicity. I use the frame_size_limit used in Chrome, 16384 * 16384:

https://chromium-review.googlesource.com/c/chromium/src/+/1614449

This value is greater than the largest MaxPicSize value of 35651584 in the AV1 spec, Annex A.3. It is also greater than the largest product of MaxHSize and MaxVSize (16384 * 8704). Note, however, that page 644 of the AV1 spec says:

    If seq_level_idx is equal to 31 (indicating the maximum parameters level),
    then there are no level-based constraints on the bitstream.

    Note: The maximum parameters level should only be set for bitstreams that
    do not conform to any other level. Typically this would be used for large
    resolution still images.

If you do not like a hardcoded value of frame_size_limit, I can rework this pull request to use a build option to enable frame_size_limit. That will require a corresponding change to oss-fuzz/projects/libavif/build.sh to build libavif with frame_size_limit enabled.

joedrago:
I think this is a reasonable cap. Anything bigger than this could use a grid image instead.

PR #115, opened 2020-03-18
Add CIFuzz

Leo-Neat:
OSS-Fuzz was wondering if libavif would use CIFuzz, a new feature being rolled out
in OSS-Fuzz. It allows for pull requests to be fuzzed in CI using fuzz targets enabled
in OSS-Fuzz. Documentation for this feature can be found at https://google.github.io/oss-fuzz/getting-started/continuous-integration/
Let us know if you have any questions!

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,23 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'libavif'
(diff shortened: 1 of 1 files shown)


joedrago:
I'm definitely on board with this change/feature, but it looks like it is failing to build dav1d because it can't find nasm, despite it being built and available in the previous step. 

I'm not familiar with this fuzz.yml mechanism -- is this building exactly the same way oss-fuzz builds, or is there some oddity here?

Leo-Neat:
@joedrago Ill look into the build failure.

Leo-Neat:
@joedrago  What happens is we mount the checked out PR over the previous state and delete contents not in the checked out PR. This removes some of your files necessary for building. So the issue is on our end and I am working to patch it. Should have one up soon!

joedrago:
@Leo-Neat Any movement on this one? If your interest in adding this isn't there anymore, can I close this PR?

PR #117, opened 2020-03-22
Ext: Only clone single branch of libgav1

EwoutH:
Speeds up the build process and lowers memory requirements while compiling a little.


cryptomilk:
Would be `git clone --depth 1` the better option?

EwoutH:
Normally, yes, but in this case a specific commit is checked out so we need the ability to look back. And you can't use a higher specific depth, like 50 or something, because commits will be piled on and the depth will prove insufficient at some point.

`--shallow-since=<date>` is an option in some newer repos, unfortunately Google's repos don't suport it.

EwoutH:
@joedrago Very small patch this, can it be merged?

PR #120, opened 2020-04-03
Ext: Add cmake files to build external projects automatically

1480c1:
The Build*.cmake files replicate the *.cmd as much as possible

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ cmake_minimum_required(VERSION 3.5)
 
 # Specify search path for CMake modules to be loaded by include()
 # and find_package()
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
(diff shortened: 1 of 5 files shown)


1480c1:
I guess github ci is not working properly right now

Here's a build on the same commit https://github.com/1480c1/libavif/actions/runs/69565768

EwoutH:
@1480c1 Thanks for the effort! Could you rebase and check if AppVeyor completes this time?

@joedrago I think it's important to get this in, it helps for building from external projects.

joedrago:
I recognize what you are doing here, but I really do not like this approach, sorry. I have taken this all-inclusive approach in other projects (like colorist), and it really makes for a nightmare of maintenance of CMake dependencies. I am more likely to go the other direction and completely remove all references to `AVIF_LOCAL_*` before going in this direction.

I appreciate the effort, but seeing as I'm on the hook for keeping this machinery running, I am not likely to accept this PR.

PR #122, opened 2020-04-05
Remove an unnecessary null check for colorOBUItem

wantehchang:
The following check ensures colorOBUItem is not null:

2033         if (!colorOBUItem) {
2034             return AVIF_RESULT_NO_AV1_ITEMS_FOUND;
2035         }

After this check, colorOBUItem is dereferenced without further checking
in multiple places.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2098,7 +2098,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
             }
         }
 
-        if ((data->colorGrid.rows > 0) && (data->colorGrid.columns > 0) && colorOBUItem) {
+        if ((data->colorGrid.rows > 0) && (data->colorGrid.columns > 0)) {
             if (!avifDecoderDataGenerateImageGridTiles(data, &data->colorGrid, colorOBUItem, AVIF_FALSE)) {
                 return AVIF_RESULT_INVALID_IMAGE_GRID;
             }


bnason-nf:
Hi @wantehchang,

Yes certainly the colorOBUItem is unnecessary, I just added it for symmetry, but I'm fine with removing it.

The alphaOBUItem check is also probably unnecessary as you point out, but it a simple way to make the analyzer happy. Adding an assert instead would keep the debug build flavor happy, but since assert() is a no-op in release builds, wouldn't help with that.

Both of these checks have almost no impact since this code is rarely used as far as I understand and the extra conditional will not make any measurable difference in performance.

Thanks,
Benbuck

wantehchang:
Hi Benbuck,

I thought you might have added the colorOBUItem check for symmetry. Thank you for the confirmation.

Since colorOBUItem is mandatory while alphaOBUItem is optional, the code that handles them does not need to be symmetric. For example, the for loops that find them have different comments:

        // Find the colorOBU (primary) item
vs.
        // Find the alphaOBU item, if any

Also, the reason I wrote this pull request was to make the code easier to understand. (I would not be surprised if some static analyzer would warn that the colorOBUItem check is always true.) I was not worried about the performance overhead of the checks.

Thanks,
Wan-Teh

bnason-nf:
Hi Wan-Teh,

Makes sense, I have no arguments with your logic.

Thanks,
Benbuck

PR #124, opened 2020-04-14
decoder->image can't be NULL after decoder has run

wantehchang:
decoder->image is set to NULL only in avifDecoderCleanup(), which is
only called in avifDecoderDestroy() and at the beginning of
avifDecoderParse(). Therefore, after decoder->image is allocated in
avifDecoderParse(), decoder->image cannot be NULL.

In avif.h, instead of saying when decoder->image "can be NULL", say when
decoder->image "is invalid".

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -544,7 +544,7 @@ typedef struct avifDecoder
     // Set this via avifDecoderSetSource().
     avifDecoderSource requestedSource;
 
-    // The current decoded image, owned by the decoder. Can be NULL if the decoder hasn't run or has run
+    // The current decoded image, owned by the decoder. Is invalid if the decoder hasn't run or has run
     // out of images. The YUV and A contents of this image are likely owned by the decoder, so be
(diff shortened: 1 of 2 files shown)


include/avif/avif.h
```diff
@@ -544,7 +544,7 @@ typedef struct avifDecoder
     // Set this via avifDecoderSetSource().
     avifDecoderSource requestedSource;
 
-    // The current decoded image, owned by the decoder. Can be NULL if the decoder hasn't run or has run
+    // The current decoded image, owned by the decoder. Is invalid if the decoder hasn't run or has run
     // out of images. The YUV and A contents of this image are likely owned by the decoder, so be
```
wantehchang, line 548: Alternatively, I can change the second sentence to "Can be NULL if the decoder hasn't run." Please let me know which one you prefer.

In fact, decoder->image is invalid after any error, not just AVIF_RESULT_NO_IMAGES_REMAINING. So the proposed comment can still be improved.

PR #125, opened 2020-04-14
Use outTiming param in avifDecoderNthImageTiming

wantehchang:
Actually use the outTiming paramter in avifDecoderNthImageTiming(). It
was still using decoder->imageTiming in the copied code.


joedrago:
Hah, yikes! Excellent catch.

If it wasn't for that `memcpy()` early in the function, that wouldn't have even compiled (unused argument).

Issue #129, opened 2020-04-22
avifCodecVersions: maxChars needs to be reduced after each strncat call


wantehchang:
The current version of avifCodecVersions() makes several calls to strncat() but passes a fixed 'maxChars' (255) as the third argument.

maxChars needs to be reduced after each strncat() call.

I don't see a simple way to fix this. I probably would call strlen() on each string we want to concatendate, take the minimum of that length and maxChars, use memcpy() to copy that many bytes, and then reduce maxChars by that length.

Two other thoughts:

1. If I remember it correctly, the current version was a result of fixing a static analysis error. This seems to imply that static analyzer is not powerful enough to analyze a sequence of strncat() calls.

2. If it's okay to change the prototype of the avifCodecVersions(), I suggest adding a "size_t size" as the second parameter. The compiler treats char outBuffer[256] and char *outBuffer the same way, so the 256-element array notation is merely documentation.

joedrago:
`avifCodecVersions()` is just the worst, and so is `strncat()`. I had `strlcat()` on the brain.

I've thrown out all `str*cat()` calls and just made it explicit, here:

https://github.com/AOMediaCodec/libavif/commit/8bfe2fc44a7f9abe7ee3b5840a45b763c3236f54

Good catch!

wantehchang:
Thanks for the fix!

I looked into the strncat() issue last night. Among the solutions not available in C99 (strlcat, strcat_s, and strncat_s), I also like strlcat the best.

I also found the following solution in Robert C. Seacord's Secure Coding in C and C++, Second Edition, page 94. Change a call like:

    strncat(outBuffer, availableCodecs[i].name, maxChars);

to:

    strncat(outBuffer, availableCodecs[i].name, maxChars - strlen(outBuffer));

Although not efficient, this solution is compact and could be used where performance is not critical.

PR #130, opened 2020-04-22
Fix potential buffer overflow in avifCodecVersions

wantehchang:
This is not the most elegant fix. Note that it does not bother to break
out of the for loop early if we reach the end of outBuffer, because that
is an unlikely event.

Fix issue https://github.com/AOMediaCodec/libavif/issues/129.


wantehchang:
Joe: Please feel free to decline this PR and write a better fix. I wrote it quickly.

joedrago:
Fixed here:

https://github.com/AOMediaCodec/libavif/commit/8bfe2fc44a7f9abe7ee3b5840a45b763c3236f54

PR #132, opened 2020-04-23
avifCodecVersions: null-terminate string just once

wantehchang:
Since append() does not require outBuffer be null-terminated, it
suffices to null-terminate outBuffer just once, after all the append()
calls are done.

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -485,14 +485,12 @@ static void append(char ** writePos, size_t * remainingLen, const char * appendS
     memcpy(*writePos, appendStr, appendLen);
     *remainingLen -= appendLen;
     *writePos += appendLen;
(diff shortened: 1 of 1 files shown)


wantehchang:
Joe: This is probably a matter of personal preference. Feel free to decline this.

joedrago:
Heh, this definitely cross the threshold of over-optimization/bikeshedding for me. I will concede that this is closer to the "optimal" way to concatenate 3 strings one time in a version function that is called occasionally on startup/helpSyntax, but I'd prefer that the `append()` function always leaves the buffer in a valid state, and I'll suffer with the 3 additional 0-byte sets that occur.

Unless this function has a crash/security issue, I'm hoping to never touch it again. :-)

PR #141, opened 2020-04-23
Misc cleanup for avifenc and avifdec

wantehchang:
1. avifenc and avifdec: Document .jpeg as a recognized file extension.

2. avifdec: Output quality range is 0-100, not 1-100. Note: libjpeg
internally still converts 0 to 1 to avoid division by zero. See
jpeg_set_quality and jpeg_quality_scaling in jcparam.c.

3. avifenc and avifdec: Continue to clean up resources after an error
with file extension, rather than exit immediately.

4. avifdec: Fail with exit status 1 on unrecognized file extension.
(avifenc already does this.)

5. avifenc: Have the syntax help message show "l" and "f" as supported
abbreviations of "limited" and "full" and the default range is full.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -21,7 +21,7 @@
 
 static void syntax(void)
 {
-    printf("Syntax: avifenc [options] input.[jpg|png|y4m] output.avif\n");
+    printf("Syntax: avifenc [options] input.[jpg|jpeg|png|y4m] output.avif\n");
     printf("Options:\n");
     printf("    -h,--help                         : Show syntax help\n");
     printf("    -j,--jobs J                       : Number of jobs (worker threads, default: 1)\n");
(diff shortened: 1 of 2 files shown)


apps/avifenc.c
```diff
@@ -31,7 +31,7 @@ static void syntax(void)
     printf("                                        P = enum avifNclxColourPrimaries\n");
     printf("                                        T = enum avifNclxTransferCharacteristics\n");
     printf("                                        M = enum avifNclxMatrixCoefficients\n");
-    printf("    -r,--range RANGE                  : YUV range [limited, full]. (JPEG/PNG only; For y4m, range is retained)\n");
+    printf("    -r,--range RANGE                  : YUV range [limited or l, full or f]. (JPEG/PNG only, default: full; For y4m, range is retained)\n");
     printf("    --min Q                           : Set min quantizer for color (%d-%d, where %d is lossless)\n",
```
wantehchang, line 35: Joe: I am not sure if "[limited or l, full or f]" is the best way to express the supported abbreviations. Please feel free to edit this directly.

joedrago: How you have it is fine.

wantehchang: Thanks!

PR #142, opened 2020-04-23
nclx: change uint16_t casts to enum type casts

wantehchang:
The first three members of struct avifNclxColorProfile were of the
uint16_t type before. In v0.7.0 they were changed to enum types. So
change the uint16_t casts for those struct members to enum type casts.

Note: v0.7.0 added -Wno-conversion to CMakeLists.txt, so we can also
just delete the uint16_t casts for those struct members.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -139,9 +139,9 @@ static avifBool aomCodecGetNextImage(avifCodec * codec, avifImage * image)
         if (image->profileFormat == AVIF_PROFILE_FORMAT_NONE) {
             // If the AVIF container doesn't provide a color profile, allow the AV1 OBU to provide one as a fallback
             avifNclxColorProfile nclx;
-            nclx.colourPrimaries = (uint16_t)codec->internal->image->cp;
-            nclx.transferCharacteristics = (uint16_t)codec->internal->image->tc;
-            nclx.matrixCoefficients = (uint16_t)codec->internal->image->mc;
+            nclx.colourPrimaries = (avifNclxColourPrimaries)codec->internal->image->cp;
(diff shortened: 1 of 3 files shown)


src/codec_aom.c
```diff
@@ -139,9 +139,9 @@ static avifBool aomCodecGetNextImage(avifCodec * codec, avifImage * image)
         if (image->profileFormat == AVIF_PROFILE_FORMAT_NONE) {
             // If the AVIF container doesn't provide a color profile, allow the AV1 OBU to provide one as a fallback
             avifNclxColorProfile nclx;
-            nclx.colourPrimaries = (uint16_t)codec->internal->image->cp;
-            nclx.transferCharacteristics = (uint16_t)codec->internal->image->tc;
-            nclx.matrixCoefficients = (uint16_t)codec->internal->image->mc;
+            nclx.colourPrimaries = (avifNclxColourPrimaries)codec->internal->image->cp;
+            nclx.transferCharacteristics = (avifNclxTransferCharacteristics)codec->internal->image->tc;
+            nclx.matrixCoefficients = (avifNclxMatrixCoefficients)codec->internal->image->mc;
```
wantehchang, line 144: Joe: Because of the -Wno-conversion flag added in v0.7.0, we do not need any casts for these assignments. Please let me know if you would like me to simply remove the uint16_t casts. Thanks!

joedrago: I like the explicit casts (the way you changed it).

wantehchang: Thanks. That's good to know. I will submit another pull request that allows us to build libavif with -Wconversion enabled, with the exception of the warning about isnan and isinf in cJSON.c due to a clang bug:

  https://bugs.llvm.org/show_bug.cgi?id=35268

PR #143, opened 2020-04-24
Ext: Update libaom to 93c4d56

EwoutH:
Main reason for the update is to get the last months memory optimizations in, memory usage should be [down by about 50%](https://www.reddit.com/r/AV1/comments/fx1249/more_memory_use_reductions_in_libaom/) for libaom.


EwoutH:
Did a quick test encoding an image with the following specs:
```
 * Resolution     : 5472x3648
 * Bit Depth      : 8
 * Format         : YUV444
```
With the new libaom version, the memory usage dropped from 6.2 GB to 5.6 GB. Solid improvement, but both still quite a lot!

joedrago:
I love the idea of having a more recent libaom, believe me, but we really need them to start making new tags. The `--depth 2000` here is a moving target, and going back to not using `--depth` would be sad.

I'll ping them again, again. :-)

joedrago:
They're trying to finish up 2.0.0 now, and this is the public issue tracking it:

https://bugs.chromium.org/p/aomedia/issues/detail?id=2545

EwoutH:
Thanks for sorting this out. We could also replace the `--depth` flag with `--single-branch`. That will take a bit longer since the whole master branch is cloned but should still be acceptable.

But let's wait a bit on the 2.0 tag fist.

PR #148, opened 2020-04-28
Improve output file extension error handling.

wantehchang:
AVIF_APP_FILE_FORMAT_UNKNOWN should result in an error message and error
exit status.

Print outputFilename, not inputFilename, in error messages.

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -142,6 +142,8 @@ int main(int argc, char * argv[])
 
         avifAppFileFormat outputFormat = avifGuessFileFormat(outputFilename);
         if (outputFormat == AVIF_APP_FILE_FORMAT_UNKNOWN) {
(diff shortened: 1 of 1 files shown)


apps/avifdec.c
```diff
@@ -155,7 +157,7 @@ int main(int argc, char * argv[])
                 returnCode = 1;
             }
         } else {
-            fprintf(stderr, "Unrecognized file extension: %s\n", inputFilename);
+            fprintf(stderr, "Unrecognized file extension: %s\n", outputFilename);
             returnCode = 1;
```
wantehchang, line 161: If we get here, the output file extension is ".avif". I am not sure if ""Unrecognized file extension" is the most appropriate error message. But at least this makes avifdec match the error handling in avifenc.

PR #152, opened 2020-04-28
Do not ignore GCC's -Wclobbered warning

wantehchang:
The C Standard requires certain local variables in the function calling
setjmp be declared as volatile. I went through the code and declare
variables as volatile according to the C Standard, except for the 'rgb'
struct variable. Document my rationale.

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -12,11 +12,6 @@
 
 #include "iccjpeg.h"
 
-// This warning triggers false postives way too often in here.
-#if defined(__GNUC__) && !defined(__clang__)
-#pragma GCC diagnostic ignored "-Wclobbered"
-#endif
-
 struct my_error_mgr
 {
     struct jpeg_error_mgr pub;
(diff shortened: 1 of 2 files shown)


apps/shared/avifjpeg.c
```diff
@@ -30,31 +25,44 @@ static void my_error_exit(j_common_ptr cinfo)
...
+        goto cleanup;
+    }
+
     struct my_error_mgr jerr;
     struct jpeg_decompress_struct cinfo;
     cinfo.err = jpeg_std_error(&jerr.pub);
     jerr.pub.error_exit = my_error_exit;
     if (setjmp(jerr.setjmp_buffer)) {
-        return AVIF_FALSE;
+        goto cleanup;
```
wantehchang, line 61: This "goto cleanup" fixes memory and resource leaks after a longjmp call.

NOTE: all the potential problems fixed by this CL can only be observed if libjpeg or libpng calls longjmp. I did not test that. I studied the C Standard and applied its rules to the code.

apps/shared/avifjpeg.c
```diff
@@ -63,10 +71,12 @@ avifBool avifJPEGRead(avifImage * avif, const char * inputFilename, avifPixelFor
     int row_stride = cinfo.output_width * cinfo.output_components;
     JSAMPARRAY buffer = (*cinfo.mem->alloc_sarray)((j_common_ptr)&cinfo, JPOOL_IMAGE, row_stride, 1);
 
+    uint8_t * iccDataTmp;
```
wantehchang, line 74: By using this non-volatile local variable 'iccDataTmp', we avoided changing the function prototype of read_icc_profile().

PR #160, opened 2020-05-08
Ext: Update libaom to 2.0.0 release branch

EwoutH:
I would suggest merging this now so we can start testing the libaom 2.0.0 library, and then updating back to -b v2.0.0 when v2.0.0 is tagged.

### DIFF
--- a/ext/aom.cmd
+++ b/ext/aom.cmd
@@ -8,7 +8,7 @@
 : # If you're running this on Windows, be sure you've already run this (from your VC2017 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Auxiliary\Build\vcvars64.bat"
 
-git clone -b v1.0.0-errata1-avif --depth 1 https://aomedia.googlesource.com/aom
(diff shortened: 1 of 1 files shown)


ext/aom.cmd
```diff
@@ -8,7 +8,7 @@
 : # If you're running this on Windows, be sure you've already run this (from your VC2017 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Auxiliary\Build\vcvars64.bat"
 
-git clone -b v1.0.0-errata1-avif --depth 1 https://aomedia.googlesource.com/aom
+git clone -b applejack --depth 1 https://aomedia.googlesource.com/aom
```
wantehchang, line 11: The v2.0.0-rc1 tag has just been created. We can use that instead.

joedrago:
I switched us over to `v2.0.0-rc1` and reenabled the cpu-used=7+ code when I detect that libaom is new enough.

https://github.com/AOMediaCodec/libavif/commit/f864bd23111b59c552d1524786245d0172cbba79

EwoutH:
Way better, when I opened this PR the v2.0.0-rc1 tag wasn't made yet.

PR #163, opened 2020-05-11
Add 'const' to some pointer parameters

wantehchang:
I made a pass through the public header avif.h and added 'const' to the
pointer parameters that could be declared as 'const'. I then made the
necessary changes to make the code compile.


joedrago:
This looks nice; thank you.

PR #169, opened 2020-05-18
Ext: update to rav1e 0.4.0

EwoutH:
Updates rav1e.cmd to use rav1e v0.4.0, which includes monochrome support.


EwoutH:
A bit preliminary, since v0.4.0 has not been released yet. Will reopen and rebase in time.

PR #175, opened 2020-05-22
Pass avifProperty to avifParse*Property functions

wantehchang:
Pass avifProperty * instead of avifDecoderData * to the
avifParse*Property() functions. This gets rid of the repeated
appearances of the expression data->properties.prop[propertyIndex].

Fix a typo in the comment for the hasUnsupportedEssentialProperty member
in the avifDecoderItem struct.

Move the members of avifProperty into a union because only one of them
is used, depending on the 'type' member.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -117,7 +117,7 @@ typedef struct avifDecoderItem
     uint32_t auxForID;       // if non-zero, this item is an auxC plane for Item #{auxForID}
     uint32_t descForID;      // if non-zero, this item is a content description for Item #{descForID}
     uint32_t dimgForID;      // if non-zero, this item is a derived image for Item #{dimgForID}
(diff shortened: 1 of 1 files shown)


wantehchang:
Joe: I can undo the union change if you don't like to use unions.

src/read.c
```diff
@@ -117,23 +117,26 @@ typedef struct avifDecoderItem
     uint32_t auxForID;       // if non-zero, this item is an auxC plane for Item #{auxForID}
     uint32_t descForID;      // if non-zero, this item is a content description for Item #{descForID}
     uint32_t dimgForID;      // if non-zero, this item is a derived image for Item #{dimgForID}
-    avifBool hasUnsupportedEssentialProperty; // If true, this file cites a property flagged as 'essential' that libavif doesn't support (yet). Ignore the item, if so.
```
wantehchang, line 120: I assume "this file" is a typo. I changed it to "this item".

joedrago:
This is a nice change, thanks!

PR #177, opened 2020-05-24
Add FreeBSD to CI

jbeich:
TravisCI [supports](https://github.com/travis-ci/travis-build/pulls?q=is%3Apr+freebsd+is%3Aclosed) FreeBSD nowadays. Let's see how well it works.

Coverage compared to Linux:
- Recent Clang/GCC
- System libaom/libdav1d/librav1e
- Static analysis


.travis.yml
```diff
@@ -45,3 +45,23 @@ matrix:
   - name: "Clang Release"
     compiler: clang
     config: Release
+  - &freebsd_common
+    name: "Clang 10 + system libraries"
+    os: freebsd
+    compiler: clang
+    before_install:
+      - | # FreeBSD 12.2 has Clang 10 but TravisCI uses FreeBSD 12.1
+        if [[ "$($CC --version)" =~ 8.0 ]]; then
```
1480c1, line 54:
```suggestion
        if [[ $(${CC:-clang} --version) =~ 8.0 ]]; then
```

bash `[[` does not require you to quote the lhs for a lot of comparisons, also it would be a good idea to add an alternative replacement in the case when someone is copy-pasting commands to test or reproduce

jbeich: Thanks. I mainly write in Ash (/bin/sh on FreeBSD), so rarely use `[[`.

.travis.yml
```diff
@@ -45,3 +45,23 @@ matrix:
...
+    os: freebsd
+    compiler: clang
+    before_install:
+      - | # FreeBSD 12.2 has Clang 10 but TravisCI uses FreeBSD 12.1
+        if [[ $(${CC:-clang} --version) =~ 8.0 ]]; then
+          export CC=clang10 CXX=clang++10
+          sudo pkg install -y llvm10
+        fi
+    before_script:
+      - sudo pkg install -y aom dav1d librav1e ninja
```
1480c1, line 59: Could we try using
```yml
      addons:
        pkg:
          packages:
            - aom
            - dav1d
            - librav1e
            - ninja
```
?

I haven't tested it out yet so I will quickly make sure that it's the right syntax
https://github.com/travis-ci/travis-yml/pull/154/files

1480c1:
Would it be better to change all of the other builds to using ninja to match all of them to use the same generator?

.travis.yml
jbeich: A few days ago it didn't work, except in recently created forks. Besides, `before_script` here overrides topmost definition (for Linux).

1480c1: I see, if pkg does work, we could just do `before_script: true` or `before_script: skip`

ps: seems to work

```diff
diff --git a/.travis.yml b/.travis.yml
index e44b654..d10e7b5 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,14 +49,20 @@ matrix:
     name: "Clang 10 + system libraries"
     os: freebsd
     compiler: clang
+    addons:
+      pkg:
+        packages:
+          - aom
+          - dav1d
+          - librav1e
+          - ninja
     before_install:
       - | # FreeBSD 12.2 has Clang 10 but TravisCI uses FreeBSD 12.1
         if [[ $(${CC:-clang} --version) =~ 8.0 ]]; then
           export CC=clang10 CXX=clang++10
           sudo pkg install -y llvm10
         fi
-    before_script:
-      - sudo pkg install -y aom dav1d librav1e ninja
+    before_script: skip
     script:
       - cmake -Bbuild -GNinja -DAVIF_CODEC_AOM=ON -DAVIF_CODEC_DAV1D=ON -DAVIF_CODEC_RAV1E=ON -DAVIF_BUILD_APPS=ON -DAVIF_BUILD_TESTS=ON
       - cmake --build build
```
image

jbeich: > ps: seems to work

Needs to be tested in a PR, not a fork. See https://github.com/Alexays/Waybar/commit/8d8c0489243fe2731161820e3013da7bf1ef434d.

1480c1: Hmm, interesting, wonder why it doesn't work. Maybe that should be reported upstream to travis?

jbeich: Found it. [This build](https://travis-ci.org/github/Alexays/Waybar/jobs/688252503) used `pkg` but failed, despite working in a fork.

jbeich:
> change all of the other builds to using ninja

Done. I also wanted to replace `mkdir build` + `cd build` + `cmake ..` with `cmake -B build` (like in `os: freebsd` case) but cmake package on `dist: bionic` is too old.

1480c1:
> > change all of the other builds to using ninja
> 
> Done. I also wanted to replace `mkdir build` + `cd build` + `cmake ..` with `cmake -B build` (like in `os: freebsd` case) but cmake package on `dist: bionic` is too old.

Yeah, I had that issue in some other places as well, one thing you could try is using kitware's apt repository with the sourceline option for the apt addon and add `cmake` to the packages


https://docs.travis-ci.com/user/installing-dependencies#adding-apt-sources
https://apt.kitware.com/

```yaml
addons:
  apt:
    sources:
    - sourceline: deb https://apt.kitware.com/ubuntu/ bionic main
      key_url: https://apt.kitware.com/keys/kitware-archive-latest.asc
```

1480c1:
nvm on the sources thing, forgot that travis-ci adds their cmake in front of everything.

ext/rav1e.cmd
```diff
@@ -11,7 +11,7 @@
 : # Also, the error that "The target windows-msvc is not supported yet" can safely be ignored provided that rav1e/target/release
 : # contains rav1e.h and rav1e.lib.
 
-git clone -b v0.3.1 --depth 1 https://github.com/xiph/rav1e.git
+git clone -b 0.3 --depth 1 https://github.com/xiph/rav1e.git
```
joedrago, line 14: Why is this change in this commit? It seems unrelated to the CI update.

1480c1: I think this is to fix an issue with rav1e compiling

https://github.com/xiph/rav1e/pull/2324

https://github.com/xiph/rav1e/pull/2325

joedrago: Ah, interesting. I imagine these fixes will land in rav1e v0.4 soon. I will probably wait until rav1e v0.4 hits before accepting this PR, but I will accept it shortly after that.

joedrago: Actually, I've changed my mind. I'll pull this now in order to get the CIs back up and running. Thanks for the information.

PR #182, opened 2020-06-20
Add a gdk-pixbuf module

linkmauve:
[gdk-pixbuf](https://developer.gnome.org/gdk-pixbuf/stable/) is a widely-used image loader on Linux, this module gives most Linux software instant support for AVIF.

gdk-pixbuf upstream [rejects any new format](https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/merge_requests/76#note_845043), so in order to make Linux distributions aware that this AVIF loader exists it makes sense to ship it alongside libavif.

For now it only supports loading still images (no sequences) and doesn’t support saving, but this can come in a later pull request.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -450,3 +450,30 @@ if(WIN32)
 endif()
 
 # ---------------------------------------------------------------------------------------
+# Contrib modules, enhancing other software with AVIF support
+
+option(AVIF_BUILD_GDK_PIXBUF "Build a gdk-pixbuf loader" OFF)
+if(AVIF_BUILD_GDK_PIXBUF)
+    find_package(PkgConfig)
+    if(PKG_CONFIG_FOUND)
+        pkg_search_module(GDK_PIXBUF gdk-pixbuf-2.0)
+        if(GDK_PIXBUF_FOUND)
+            set(GDK_PIXBUF_SRCS
+                contrib/gdk-pixbuf/loader.c
+            )
+            add_library(pixbufloader-avif ${GDK_PIXBUF_SRCS})
+
+            # This is required because glib stupidly uses invalid #define names, such as __G_LIB_H__…
+            add_definitions(-Wno-reserved-id-macro)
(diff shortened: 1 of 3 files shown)


joedrago:
Assuming you can survive the Travis (and other) CI checks, I'm generally okay with this kind of thing. That said, this is the kind of thing that should be in a `contrib/` directory, as I don't plan on actively maintaining it, and I'd want a new section added to LICENSE citing the `contrib/gdk-pixbuf/*` with your license info in it.

I'll probably want all contrib build options to be in their own section in CMakeLists.txt, but I can clean that up afterwards.

joedrago:
I'll make a stub `contrib/` directory now in the `master` branch, so you can move this in there (if you want to continue to tweak / submit this PR).

linkmauve:
I can commit to becoming the maintainer of this loader, if that’s what you’re worried about. :)

As for the license, I went with the same one as libavif, as I don’t have much of a preference.  With gdk-pixbuf itself being under LGPL 2.1, that shouldn’t be an issue.

I’ll rearrange the cmake stuff and move it there, fix the const issues from clang, and then this MVP should be good to go, thanks!

joedrago:
I recognize that you might simply have it on right now to verify the build with CI, but `AVIF_BUILD_GDK_PIXBUF` certainly needs to be `OFF` by default in the final PR.

linkmauve:
There, CI passed!

joedrago:
I've merged this, but I'm going to move the CMake block into the contrib subdir. Let me know if it breaks for you after I've done this, please.

joedrago:
Alright, I've moved the block here (with one tiny change to the sources line):

https://github.com/AOMediaCodec/libavif/commit/98d9f4b13c1c8a3e6030c490e98edcb7c86d8923

I've also put your own license header at the top, as you're the author of that block. This layout is much easier to delineate licensing and scales better others offer future contributions.

linkmauve:
I confirm it still works when adding `-DAVIF_BUILD_GDK_PIXBUF=ON` on the `cmake` invocation!

PR #190, opened 2020-06-24
Abort the process in case of memory allocation failure

### LINKED ISSUE #103: Memory allocation failures are not checked and handled.
Libavif does not check and handle memory allocation failures. Here is an example in src/avif.c:

avifImage * avifImageCreate(int width, int height, int depth, avifPixelFormat yuvFormat)
{
    avifImage * image = (avifImage *)avifAlloc(sizeof(avifImage));
    avifImageSetDefaults(image);
    image->width = width;
    image->height = height;
    image->depth = depth;
    image->yuvFormat = yuvFormat;
    return image;
}

Note that 'image' is dereferenced without a null pointer check.

This policy is acceptable to some clients. For example, in Chromium any memory allocation failure aborts the process. On the other hand, this policy would be a dealbreaker to clients that try to handle memory allocation failures.

If this policy is deliberate, it should be documented. Otherwise, we should add code to check for and  report out-of-memory errors. Some functions such as avifImageAllocatePlanes() that allocate memory and return void will need to return avifBool instead.

AurelC2G:
We currently do not check the results of our memory allocations, and simply dereference the result without any checks. This leads to undefined behavior whenever `malloc` returns null.

With this change, we replace this undefined behavior by deterministically aborting the process.

Fixes #103

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Update libaom.cmd to point at the v2.0.0 tag
 * Update dav1d.cmd to point at the 0.7.1 tag
 * Re-enable cpu-used=7+ in codec_aom when libaom major version > 1
+* Memory allocation failures now cause libavif to abort the process (rather than undefined behavior)
 
 ## [0.7.3] - 2020-05-04
 ### Added
--- a/src/mem.c
+++ b/src/mem.c
@@ -7,7 +7,11 @@
 
 void * avifAlloc(size_t size)
 {
-    return malloc(size);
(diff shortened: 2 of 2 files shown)


joedrago:
I do have an interest in trying to gracefully catch memory allocation failures and return a new avifError, but I'll accept this patch for now, and I'll remove this when I get around to changing the code to be the other way. Thanks for the PR.

AurelC2G:
Sounds good! I also tried to catch these failures and return a new error code, but that required quite a bit of changes in the internals of the library (a bunch of functions call `avifAlloc` but don't have an easy way to return a potential failure), so the changes started to build up and I gave up after a while. Would love it if you could get there though!

PR #193, opened 2020-06-25
Use libaom cpu_used 6 in "good quality" usage mode

wantehchang:
libaom 2.0.0 added cpu_used 6 in "good quality" usage mode. Although in
libaom 2.0.0 cpu_used 6 in "good quality" usage mode is just a place
holder and is identical to cpu_used 5, we can start using it now and it
will have an effect when a new version of libaom is released.


joedrago:
Do we need to protect against this in case someone is using libaom v1.0.0-errata1 still, or will it harmlessly clamp?

wantehchang:
This is harmless for libaom 2.0.0 and older. This has an effect if libavif is used with the current libaom master branch.

PR #195, opened 2020-06-30
Remove checks for the "av01" brand

wantehchang:
https://aomediacodec.github.io/av1-isobmff/ says a file conformant to
that spec shall have the 'av01' brand among the compatible brands array
of the FileTypeBox. The 'av01' brand (as opposed to the 'av01' image
item type) is not mentioned in the AVIF spec.

Therefore, 'av01' may be present in the compatible brands array of an
AVIF file, but it should not be the major brand of an AVIF file. In
addition, the presence of 'av01' in the compatible brands array does not
imply the file is AVIF.

Also use the || operator to connect the checks for the "avif" and "avis"
brands.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1872,25 +1872,11 @@ static avifBool avifParse(avifDecoderData * data, const uint8_t * raw, size_t ra
 
 static avifBool avifFileTypeIsCompatible(avifFileType * ftyp)
 {
-    avifBool avifCompatible = (memcmp(ftyp->majorBrand, "avif", 4) == 0);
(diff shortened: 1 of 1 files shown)


wantehchang:
Joe: It seems that the 'av01' brand requirement in AV1-ISOBMFF is irrelevant to AVIF, because the AVIF spec does not depend on that part of AV1-ISOBMFF. Correct?

joedrago:
I believe you are right. @cconcolato is this correct?

cconcolato:
I agree.

PR #196, opened 2020-06-30
Check a subtraction's result is in range of size_t

wantehchang:
In avifROStreamReadBoxHeader(), make sure that the result of
size - (stream->offset - startOffset) is within the range of size_t,
i.e., nonnegative and less than or equal to SIZE_MAX. Note that the
result of stream->offset - startOffset is positive.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -3,6 +3,7 @@
 
 #include "avif/internal.h"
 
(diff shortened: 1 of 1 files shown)


joedrago:
I merged this and added a few parens for subjective styling.

PR #198, opened 2020-07-01
Remove MAX_COMPATIBLE_BRANDS

wantehchang:
Change the compatibleBrands field of the avifFileType struct to a
pointer that points to compatible_brands in the read-only input. It is
not necessary to copy the compatible_brands to a buffer in the
avifFileType struct as long as we can keep the read-only input valid
for the lifetime of the avifFileType struct.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -7,7 +7,6 @@
 
 #define AUXTYPE_SIZE 64
 #define CONTENTTYPE_SIZE 64
-#define MAX_COMPATIBLE_BRANDS 32
 
 // class VisualSampleEntry(codingname) extends SampleEntry(codingname) {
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -1837,11 +1837,8 @@ static avifBool avifParseFileTypeBox(avifFileType * ftyp, const uint8_t * raw, s
     if ((compatibleBrandsBytes % 4) != 0) {
```
wantehchang, line 1837: We could also check compatibleBrandsBytes > (4 * MAX_COMPATIBLE_BRANDS) here to disallow a very-long compatible_brands list. Not sure if that is a good idea.

joedrago:
This is great!

PR #203, opened 2020-07-06
Use sampleIndex in the for loop's body

wantehchang:
Update the fix for oss-fuzz issue: 23940


joedrago:
Oh, I didn't see this PR before I fixed it (exactly the same way):

https://github.com/AOMediaCodec/libavif/commit/043311bee1c932d2a412b478a015254d91ac4d1a

Thanks!

PR #204, opened 2020-07-07
Miscellaneous cleanup of src/codec_rav1e.c

wantehchang:
In rav1eSupports400(), return the boolean expression minorVersion >= 4
directly at the end of the function.

In rav1eCodecEncodeImage(), move the declaration and initialization of
supports400 into the block that uses the variable, and replace the
boolean expression
  (codec->internal->chromaSampling != RA_CHROMA_SAMPLING_CS400) &&
  (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400)
with the equivalent expression
  image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400.

Note: The two boolean expressions are equivalent because
  image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400
implies
  codec->internal->chromaSampling != RA_CHROMA_SAMPLING_CS400
if 'alpha' is false.

### DIFF
--- a/src/codec_rav1e.c
+++ b/src/codec_rav1e.c
@@ -52,11 +52,7 @@ static avifBool rav1eSupports400(void)
         return AVIF_FALSE;
     }
     int minorVersion = atoi(minorVersionString);
-    if (minorVersion >= 4) {
-        return AVIF_TRUE;
-    }
-
-    return AVIF_FALSE;
+    return minorVersion >= 4;
 }
 
 static avifBool rav1eCodecEncodeImage(avifCodec * codec,
@@ -70,9 +66,9 @@ static avifBool rav1eCodecEncodeImage(avifCodec * codec,
 
     RaConfig * rav1eConfig = NULL;
     RaFrame * rav1eFrame = NULL;
-    const avifBool supports400 = rav1eSupports400();
 
     if (!codec->internal->rav1eContext) {
+        const avifBool supports400 = rav1eSupports400();
         RaPixelRange rav1eRange;
         if (alpha) {
             rav1eRange = (image->alphaRange == AVIF_RANGE_FULL) ? RA_PIXEL_RANGE_FULL : RA_PIXEL_RANGE_LIMITED;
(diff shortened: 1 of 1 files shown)


src/codec_rav1e.c
```diff
@@ -174,7 +170,7 @@ static avifBool rav1eCodecEncodeImage(avifCodec * codec,
         rav1e_frame_fill_plane(rav1eFrame, 0, image->alphaPlane, image->alphaRowBytes * image->height, image->alphaRowBytes, byteWidth);
     } else {
         rav1e_frame_fill_plane(rav1eFrame, 0, image->yuvPlanes[0], image->yuvRowBytes[0] * image->height, image->yuvRowBytes[0], byteWidth);
-        if ((codec->internal->chromaSampling != RA_CHROMA_SAMPLING_CS400) && (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400)) {
+        if (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400) {
```
wantehchang, line 173: Joe: Here I assume the original boolean expression is correct and merely simplified it.

It is possible that the correct boolean expression should be codec->internal->chromaSampling != RA_CHROMA_SAMPLING_CS400. Since I am not familiar with this code, I don't know whether that is the case.

joedrago: Both of these checks are necessary. One of them checks the requested output (CS400), and one of them checks the available input (YUV400). I haven't checked to see if fill_plane(1/2) are no-ops in CS400 mode, but it feels better to not rely on that and just not supply chroma planes if we know we're encoding monochrome.

wantehchang: I did not try to figure out the meaning of this code. I merely simplified the boolean expressions by the laws of logic.

Here is the detailed derivation:

Earlier in the function, we have:

        if (alpha) {
            rav1eRange = (image->alphaRange == AVIF_RANGE_FULL) ? RA_PIXEL_RANGE_FULL : RA_PIXEL_RANGE_LIMITED;
            codec->internal->chromaSampling = supports400 ? RA_CHROMA_SAMPLING_CS400 : RA_CHROMA_SAMPLING_CS420;
            codec->internal->yShift = 1;
        } else {
            rav1eRange = (image->yuvRange == AVIF_RANGE_FULL) ? RA_PIXEL_RANGE_FULL : RA_PIXEL_RANGE_LIMITED;
            codec->internal->yShift = 0;
            switch (image->yuvFormat) {
                case AVIF_PIXEL_FORMAT_YUV444:
                    codec->internal->chromaSampling = RA_CHROMA_SAMPLING_CS444;
                    break;
                case AVIF_PIXEL_FORMAT_YUV422:
                    codec->internal->chromaSampling = RA_CHROMA_SAMPLING_CS422;
                    break;
                case AVIF_PIXEL_FORMAT_YUV420:
                    codec->internal->chromaSampling = RA_CHROMA_SAMPLING_CS420;
                    codec->internal->yShift = 1;
                    break;
                case AVIF_PIXEL_FORMAT_YUV400:
                    codec->internal->chromaSampling = supports400 ? RA_CHROMA_SAMPLING_CS400 : RA_CHROMA_SAMPLING_CS420;
                    codec->internal->yShift = 1;
                    break;
                case AVIF_PIXEL_FORMAT_NONE:
                default:
                    return AVIF_FALSE;
            }
        }

Therefore, if 'alpha' is false, then image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400 implies codec->internal->chromaSampling != RA_CHROMA_SAMPLING_CS400.

If P imples Q, then Q && P is equivalent to P.

Therefore, if 'alpha' is false, then
    
    (codec->internal->chromaSampling != RA_CHROMA_SAMPLING_CS400) &&
   (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400)

is equivalent to
    
    image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400

joedrago: I see; you're right.

PR #210, opened 2020-07-09
Miscellaneous cleanup for src/codec_aom.c

wantehchang:
Clarify the comment for how aom_codec_version() packs the
major.minor.patch version into an int.

The local variable encoder_interface is renamed encoderInterface to
conform to libavif's naming convention.

Set codec->internal->encoderInitialized to true after the
aom_codec_enc_init() call.

The chroma_check() bug affects libaom's "real-time" usage mode, which
libavif uses for the faster libavif speeds, not slower.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -248,7 +248,7 @@ static avifBool aomCodecEncodeImage(avifCodec * codec,
             }
         }
 
-        // aom_codec.h says: aom_codec_version() == (1<<16 | 2<<8 | 3)
+        // aom_codec.h says: aom_codec_version() == (major<<16 | minor<<8 | patch)
         static const int aomVersion_2_0_0 = (2 << 16);
         const int aomVersion = aom_codec_version();
         if ((aomVersion < aomVersion_2_0_0) && (image->depth > 8)) {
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -278,10 +278,9 @@ static avifBool aomCodecEncodeImage(avifCodec * codec,
 
         avifGetPixelFormatInfo(image->yuvFormat, &codec->internal->formatInfo);
 
-        aom_codec_iface_t * encoder_interface = aom_codec_av1_cx();
+        aom_codec_iface_t * encoderInterface = aom_codec_av1_cx();
         struct aom_codec_enc_cfg cfg;
-        aom_codec_enc_config_default(encoder_interface, &cfg, aomUsage);
-        codec->internal->encoderInitialized = AVIF_TRUE;
+        aom_codec_enc_config_default(encoderInterface, &cfg, aomUsage);
```
wantehchang, line 283: The aom_codec_enc_config_default() call does not initialize the encoder, so I moved the "codec->internal->encoderInitialized = AVIF_TRUE;" after the aom_codec_enc_init() call.

PR #212, opened 2020-07-09
Miscellaneous cleanup for src/write.c

wantehchang:
It suffices to just check imageMetadata->icc.size > 0. It is not
necessary to also check imageMetadata->icc.data. See how we check
image->exif.size > 0 and image->xmp.size > 0.

Remove a redundant check for image->exif.size > 0.

Replace encoder->data->items.count < 1 with
encoder->data->items.count == 0.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -189,7 +189,7 @@ void avifEncoderDestroy(avifEncoder * encoder)
 
 static void avifEncoderWriteColorProperties(avifRWStream * s, const avifImage * imageMetadata, struct ipmaArray * ipma, uint8_t * itemPropertyIndex)
 {
-    if (imageMetadata->icc.data && (imageMetadata->icc.size > 0)) {
+    if (imageMetadata->icc.size > 0) {
         avifBoxMarker colr = avifRWStreamWriteBox(s, "colr", AVIF_BOX_SIZE_TBD);
         avifRWStreamWriteChars(s, "prof", 4); // unsigned int(32) colour_type;
         avifRWStreamWrite(s, imageMetadata->icc.data, imageMetadata->icc.size);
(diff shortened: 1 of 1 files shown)


src/write.c
```diff
@@ -300,12 +300,12 @@ static void avifEncoderWriteTrackMetaBox(avifEncoder * encoder, avifRWStream * s
...
-        avifRWStreamWriteU16(s, 1);                     // unsigned int(16) extent_count;
-        avifEncoderItemAddMdatFixup(item, s);           //
-        avifRWStreamWriteU32(s, 0 /* set later */);     // unsigned int(offset_size*8) extent_offset;
-        avifRWStreamWriteU32(s, (uint32_t)contentSize); // unsigned int(length_size*8) extent_length;
+        avifRWStreamWriteU16(s, item->id);          // unsigned int(16) item_ID;
+        avifRWStreamWriteU16(s, 0);                 // unsigned int(16) data_reference_index;
+        avifRWStreamWriteU16(s, 1);                 // unsigned int(16) extent_count;
+        avifEncoderItemAddMdatFixup(item, s);       //
+        avifRWStreamWriteU32(s, 0 /* set later */); // unsigned int(offset_size*8) extent_offset;
+        avifRWStreamWriteU32(s, contentSize);       // unsigned int(length_size*8) extent_length;
```
wantehchang, line 308: I removed the (uint32_t) cast for contentSize because contentSize is of the uint32_t type.

src/write.c
```diff
@@ -387,29 +387,27 @@ avifResult avifEncoderAddImage(avifEncoder * encoder, const avifImage * image, u
         if (image->exif.size > 0) {
             // Validate Exif payload (if any) and find TIFF header offset
             uint32_t exifTiffHeaderOffset = 0;
-            if (image->exif.size > 0) {
```
wantehchang, line 390: I removed this redundant image->exif.size > 0 check. That's the only change in this area.

PR #214, opened 2020-07-09
Use correct epoch for creation/modification times

wantehchang:


src/write.c
```diff
@@ -499,8 +499,11 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
     // -----------------------------------------------------------------------
     // Begin write stream
 
-    avifImage * imageMetadata = encoder->data->imageMetadata;
-    uint64_t now = (uint64_t)time(NULL);
+    const avifImage * imageMetadata = encoder->data->imageMetadata;
+    // The epoch for creation_time and modification_time is midnight, Jan. 1,
+    // 1904, in UTC time. Add the number of seconds between that epoch and the
+    // Unix epoch.
+    uint64_t now = (uint64_t)time(NULL) + 2082844800;
```
wantehchang, line 506: Joe: Would you like me to document how I calculated 2082844800?

Also, the declaration of the variable 'now' could be moved into the if (encoder->data->frames.count > 1) block because this variable is only used there. Would you like me to do that? I am not sure if you plan to set creation_time and modification_time for image items also.

joedrago:
Ah, I had missed that newer epoch in the BMFF spec. Thanks for this one.

PR #215, opened 2020-07-09
Declare data param of avifRWStreamWrite as void *

wantehchang:
Declare the 'data' parameter of avifRWStreamWrite() as const void *.
This avoids the need to cast that argument to const uint8_t * at some
call sites.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -239,7 +239,7 @@ void avifRWStreamFinishWrite(avifRWStream * stream)
     }
 }
 
-void avifRWStreamWrite(avifRWStream * stream, const uint8_t * data, size_t size)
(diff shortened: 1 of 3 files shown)


src/stream.c
```diff
@@ -239,7 +239,7 @@ void avifRWStreamFinishWrite(avifRWStream * stream)
     }
 }
 
-void avifRWStreamWrite(avifRWStream * stream, const uint8_t * data, size_t size)
+void avifRWStreamWrite(avifRWStream * stream, const void * data, size_t size)
```
wantehchang, line 242: We do not do pointer arithmetic with the 'data' parameter in this function, so we can just declare it as const void *, like the memcpy() and memcmp() functions.

PR #220, opened 2020-07-09
Fix regression preventing avifenc from working

linkmauve:
Since 436ab195de82b705cf0064591bd5c9593d33d415, `avifenc` would try to read its input file twice, in two consecutive calls to `avifInputReadImage()`, the latter would fail.

A test run during CI would catch this kind of issue.


joedrago:
I'm surprised that one got past me, thanks for the tweak. I'm sure I would have ran into it today.

I do have tests running in CI for the library, but you're right, I should add flow for the tooling as well.

PR #235, opened 2020-07-17
Optimise gbrp encoding

linkmauve:
In the identity reformat mode (as opposed to YUV coefficients
reformatting), the only operation needed is to deinterleave components
into multiple planes.

This reduces CPU usage of this function from 1.84% + 0.43% of the total
profiling time of avifenc, to 0.13% on my Thinkpad x280, and from 0.90%
to 0.02% on my Nintendo Switch.

I also included a non-SIMD codepath, in case someone is running on a
different architecture.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -6,6 +6,16 @@
 #include <math.h>
 #include <string.h>
 
+#ifdef __SSE4_1__
+#include <smmintrin.h>
+#elif defined(__ARM_NEON)
+#include <arm_neon.h>
+#endif
+
+#ifdef _MSC_VER
+#define __builtin_unreachable() __assume(0)
+#endif
+
 struct YUVBlock
 {
     float y;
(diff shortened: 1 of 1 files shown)


joedrago:
This is a great optimization, and assuming the builders eventually are happy with it, I'd love to take it. My only comment (from a brief skim) is that we'll have to add some more restrictions on this optimization path than just matrixCoefficients and bit depth. As unlikely (and/or useless) it'd be to have limited range with identity or YUV420 (say), we still have to handle those cases. Luckily, if someone makes a "bad" file we can fall back to the Slow path just fine, and most people will get your optimized path.

Top-of-head, this path assumes:

* Identity
* 8bpc
* YUV444
* Full range

We need to check for those second two in order to get your new fastness. I'll inspect this more on Monday (I'm out of the office today).

linkmauve:
I’m now passing CI, except for all of the Linux builders but it doesn’t seem caused by my PRs (see #236 which fails the same way).

joedrago:
Upon further inspection, it appears that these functions are super-optimized replacements specifically for `avifImageIdentity8ToRGB8ColorFullRange()`. Can you move the implementations of these adjacent to that function and `#ifdef` accordingly? Perhaps name the one that has alpha `avifImageIdentity8ToRGBA8ColorFullRange()` and further restrict that one with checks on `image->alphaRange == AVIF_RANGE_FULL`.

src/reformat.c
```diff
@@ -138,6 +414,13 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
         avifImageAllocatePlanes(image, AVIF_PLANES_A);
     }
 
+    if (state.mode == AVIF_REFORMAT_MODE_IDENTITY && image->depth == 8 && image->yuvFormat == AVIF_PIXEL_FORMAT_YUV444 && image->yuvRange == AVIF_RANGE_FULL) {
```
joedrago, line 417: Minor style change: Can you please parenthesize the inner conditionals the way the other checks do in this function?

linkmauve:
This family of functions actually do the opposite of `avifImageIdentity8ToRGB8ColorFullRange()`, it splits an interleaved RGB24 buffer into a planar GBR buffer for encoding, so I named them `avifRGB[A]8ToIdentity8ColorFullRange()`.

I’m also interested in optimising the decoding variants, but it seems there are already quite a bunch of fast paths for it (while this PR is the first one for encoding).

joedrago:
Perhaps it might be time to break these optimized routines into appropriate files (`reformat_neon.c`, etc). This file could get pretty unwieldy in if we don't organize it a little bit more soon.

linkmauve:
Are you fine with my usage of intrinsics btw?  dav1d for instance uses manual assembly files in order to avoid compilers generating potentially less than optimal assembly, as Lynne pointed during review at #dav1d (but that’d add a dependency on nasm (on x86 only IIRC)).  I’d be fine rewriting my functions into pure asm if you prefer so, but I’m also fine keeping them as intrinsics, your call. :)

1480c1:
an interjection, cmake does have an ASM_NASM language support, but it can be finicky with macOS's Xcode generator due to a limitation in cmake's Xcode generator

joedrago:
You know, even though you said it a handful of times (including the title), it didn't register over the weekend that this is an _encoding_ optimization. This makes it much less interesting to me, as this phase of the encode isn't really "slow" (in comparison). I'm not saying I don't want the commit; I just had my expectations calibrated wrong for this one.

linkmauve:
It’s still shoving off up to 2% out of encoding, which is (so far) my main usecase.

PR #238, opened 2020-07-17
Add an AArch64 run in CI

linkmauve:
Travis now has ARMv8 builders, so make use of them to discover any regression on AArch64.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,10 @@ matrix:
   - name: "GCC Release"
     compiler: gcc
     config: Release
+  - name: "AArch64 GCC Release"
+    arch: arm64
+    compiler: gcc
+    config: Release
   - name: "Clang Debug"
     compiler: clang
     config: Debug


linkmauve:
Since rav1e behaves weirdly when using clang on AArch64, let’s disable the clang pass for now.

EwoutH:
Could you rebase this PR? Curious how it performs now.

linkmauve:
Here you go. But now that travis is paid-for, do you want to keep using it?

.travis.yml
```diff
@@ -59,6 +59,10 @@ matrix:
   - name: "Clang Release"
     compiler: clang
     config: Release
+  - name: "AArch64 GCC Release"
+    arch: arm64
+    compiler: gcc
+    config: Release
```
wantehchang, line 65: I suggest moving this to line 44, so that it follows the similar "GCC Debug" and "GCC Release" immediately.

linkmauve: Done.

EwoutH:
I think other projects in the AOMediaCodec repository, mainly [SVT-AV1](https://github.com/AOMediaCodec/SVT-AV1), have used up all free OSS credits.

linkmauve:
This one too it seems, I don’t see Travis running in this PR.

1480c1:
it may be better to install an aarch64 cross compiler in github actions to at least test cross compilation

PR #241, opened 2020-07-19
Cmake: Don't use CXX unless needed and set std using cmake convention

1480c1:
planning another pr for changing other parts the cmakelist in a bit, particularly around cflags handling

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ cmake_minimum_required(VERSION 3.5)
 # and find_package()
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
 
-project(libavif LANGUAGES C CXX VERSION 0.8.0)
+project(libavif LANGUAGES C VERSION 0.8.0)
 
 # SOVERSION scheme: MAJOR.MINOR.PATCH
 #   If there was an incompatible interface change:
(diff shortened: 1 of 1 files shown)


cryptomilk:
Looks good to me.

1480c1:
Seems that perhaps dav1d might need c11 during compilation since it seems https://code.videolan.org/videolan/dav1d/-/blob/master/include/dav1d/headers.h#L95 is an anonymous struct, and it was not caught by travis-ci previously since the gnu99 standard flag was only applied to gcc, which I guess has automatic support for it.

I can't tell if this would be a clang issue since that means their gnu99 != gcc's gnu99 at least in terms of warning produced, but that could be just because of Weverything.

I mentioned it in #dav1d irc, but in the meantime, it might be necessary to bump the c standard to c11 or we might need to add `-Wno-error=c11-extensions` to the flags

joedrago:
Can you rebase your branch and force push it again to see if Travis likes it? I have disabled libgav1 in Travis on the master branch, so we shouldn't get any false failures anymore.

joedrago:
I'd prefer to disable the C11 warning over bumping to full C11 for now, while you're in there.

1480c1:
Done, now we wait for travis

1480c1:
https://code.videolan.org/videolan/dav1d/-/merge_requests/1056

dav1d's wip mr for the c99 "incompatible" headers issue

1480c1:
seems it was successful

PR #242, opened 2020-07-19
CMake: Set flags based if the compiler supports it rather than by name

1480c1:
This should help with setting generic cflags for those that do support them

https://cmake.org/cmake/help/v3.5/module/CheckCCompilerFlag.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,46 +79,72 @@ endif()
 # ---------------------------------------------------------------------------------------
 
 # Enable all warnings
-if(CMAKE_C_COMPILER_ID MATCHES "Clang")
-  MESSAGE(STATUS "libavif: Enabling warnings for Clang")
-  add_definitions(
-    -Weverything
-    -Werror
-    -Wno-bad-function-cast
-    -Wno-cast-align
-    -Wno-conversion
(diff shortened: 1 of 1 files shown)


cryptomilk:
The right thing would be to test if they are supported by the compiler, see e.g.

https://gitlab.com/cmocka/cmocka/-/blob/master/CompilerChecks.cmake

1480c1:
I'm not sure if you are expressing disapproval, approval, or just stating something? Your message seems a bit... cryptic (lol)

But either way, this pr does change it to use `check_c_compiler_flag` https://github.com/AOMediaCodec/libavif/pull/242/files#diff-af3b638bc2a3e6c650974192a53c7291R126

joedrago:
I'd prefer to not take this one, honestly. I think this will slow down CMake generation time and create a whole new set of compile combinations I'm going to have to support. I'd rather occasionally hit an issue with a specific build env and directly target a fix for it.

wantehchang:
I think we can close this pull request. Most of those compiler warning flags are known to be supported by the versions of Clang, GCC, and MSVC that we need to support, so we don't need to detect their presence at build time. We should take advantage of the knowledge and save time.

Issue #237, opened 2020-07-17
Segmentation fault in contrib/gdk-pixbuf/loader.c


novomesk:
```
Thread 6 "pool-gthumb" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffe6718700 (LWP 28518)]
load_increment () at /var/tmp/portage/media-libs/libavif-0.8.0/work/libavif-0.8.0/contrib/gdk-pixbuf/loader.c:185
185     /var/tmp/portage/media-libs/libavif-0.8.0/work/libavif-0.8.0/contrib/gdk-pixbuf/loader.c: Adresár alebo súbor neexistuje.
(gdb) bt
#0  load_increment () at /var/tmp/portage/media-libs/libavif-0.8.0/work/libavif-0.8.0/contrib/gdk-pixbuf/loader.c:185
#1  0x00007ffff7298921 in gdk_pixbuf_loader_load_module () at /usr/lib64/libgdk_pixbuf-2.0.so.0
#2  0x00007ffff72993ea in gdk_pixbuf_loader_write () at /usr/lib64/libgdk_pixbuf-2.0.so.0
#3  0x00007ffff7296d58 in gdk_pixbuf_get_file_info () at /usr/lib64/libgdk_pixbuf-2.0.so.0
#4  0x000055555560bbe5 in gth_pixbuf_new_from_file () at pixbuf-io.c:186
#5  0x000055555560be0b in gth_pixbuf_animation_new_from_file () at pixbuf-io.c:264
#6  0x00005555555dba89 in load_image_thread () at gth-image-loader.c:212
#7  0x00007ffff73481b9 in run_in_thread () at /usr/lib64/libgio-2.0.so.0
#8  0x00007ffff732bb56 in io_job_thread () at /usr/lib64/libgio-2.0.so.0
#9  0x00007ffff735c2be in g_task_thread_pool_thread () at /usr/lib64/libgio-2.0.so.0
#10 0x00007ffff717cf94 in g_thread_pool_thread_proxy () at /usr/lib64/libglib-2.0.so.0
#11 0x00007ffff717c72d in g_thread_proxy () at /usr/lib64/libglib-2.0.so.0
#12 0x00007ffff6c22fa7 in start_thread () at /lib64/libpthread.so.0
#13 0x00007ffff6b5adbf in clone () at /lib64/libc.so.6
```

novomesk:
[...]

PR #244, opened 2020-07-22
Fix a crash in the gdk-pixbuf loader, removed unnecessary asserts

novomesk:
These are the changes I suggested in https://github.com/AOMediaCodec/libavif/issues/237


novomesk:
@linkmauve I had to make these changes to make the loader to work for me.
I would like to discuss with with you further changes/improvements of this loader.
How can I contact you?

linkmauve:
Thanks for that, and sorry for not having answered/fixed it earlier.

You can either open issues here, or contact me by XMPP or email, both should be available in my profile.

The main feature missing AFAIK is encoding support, what else are you planning on supporting?

PR #245, opened 2020-07-24
cmake: Use correct include dir order

cryptomilk:


cryptomilk:
https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#transitive-usage-requirements

PR #246, opened 2020-07-28
Save support in gdk-pixbuf component

novomesk:
Upgraded gdk-pixbuf plugin with save support.

### DIFF
--- a/contrib/gdk-pixbuf/loader.c
+++ b/contrib/gdk-pixbuf/loader.c
@@ -1,5 +1,6 @@
-// Copyright 2020 Emmanuel Gil Peyrot. All rights reserved.
-// SPDX-License-Identifier: BSD-2-Clause
+/* Copyright 2020 Emmanuel Gil Peyrot. All rights reserved.
+   SPDX-License-Identifier: BSD-2-Clause
+*/
 
 #include <avif/avif.h>
 
@@ -76,7 +77,7 @@ static gboolean avif_context_try_load(struct avif_context * context, GError ** e
 
     ret = avifDecoderNextImage(decoder);
     if (ret == AVIF_RESULT_NO_IMAGES_REMAINING) {
-        // No more images, bail out. Verify that you got the expected amount of images decoded.
+        /* No more images, bail out. Verify that you got the expected amount of images decoded. */
         return TRUE;
     } else if (ret != AVIF_RESULT_OK) {
         g_set_error(error, GDK_PIXBUF_ERROR, GDK_PIXBUF_ERROR_FAILED,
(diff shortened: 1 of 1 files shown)


joedrago:
Provided you and linkmauve are on board with this, I'm likely to just accept the PR, but I do have some quick questions.

* Why not support other depths? libavif does most of the heavy lifting on the depth rescaling. Is there just not a good way to say "I know this data is 16bpc but I'd like to pack it into 10bpc"? If you just want 8bpc for now, that's fine; just know that libavif's RGBToYUV can have disagreeing bpc (between RGB/YUV) and it'll do the right thing.

* Why fall back to 422? I feel like if you're not going to for for 444, you might as well go 420.

novomesk:
According gdk-pixbuf manual I understand they support only 8 bit. I'd like to support higher depth if possible.

For highest quality (quantizer range 0-10) I selected 444 format.
For relatively high quality (quantizer range 10-20) I selected 422 format.
For the rest (quantizer 20-63) I use traditional 420 format.
Shall I avoid 422 format?

joedrago:
I can't really say how well various subsamplings are going to be supported, outside of libavif. Of course, any bugs that might arise with any of them via libavif are on me to fix, but (for example) Microsoft doesn't even support 444 yet in their decoder. I haven't tried 422 with Microsoft, but I suspect they're assuming 420. 

I'm planning to push for everyone to support 444 in addition to 420 (which is likely to work everywhere), but I honestly can't say about 422. If the gains aren't visually significant with your current strategy over 420, I'd probably skip it.

novomesk:
I reported one YUV422 file to libheif project and there are some fresh changes in develop branch.

As you suggest I will remove the 422 tomorrow and I will let you know when ready.

novomesk:
I removed 422.
Now it uses YUV444 for quantizers 0-14 and YUV420 for quantizers 14-63.

linkmauve:
Woops, I hadn’t seen this PR before sending you my comments on https://github.com/novomesk/libavif/commit/9324352309a756bc06072ef6d18665bb2ed5367a (as per the GitHub notification I received).

If you don’t feel like adding all of the encoding parameters for now that’s fine too, we can add them over time.

The issue with >8-bit depth is that gdk-pixbuf annoyingly only supports 8-bit RGB(A) internally. Encoding at a higher bit depth will still improve the quality somewhat, but I’d really like to add support for 10-, 12- and 16-bit internal images to gdk-pixbuf too.

novomesk:
I will add the CPU detection via `g_get_num_processors` and will use only YUV420 for compatibility reasons.

Parameter *quality* (range 0-100) is quite common parameter for image formats, easiest to use/understand for everybody.

We can add AVIF specific parameters too but almost nobody will know that they can be used or how they should be used.

BTW, parameter parsing code is highly inspired by gdk-pixbuf internal sources.

novomesk:
@linkmauve The reason I chose `avif->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601;` is compatibility with released libheif 1.7.0 . It performed YUV->RGB via these coefficients only. And we are making changes here for the sake of maximum compatibility.
Future libheif will support other matrixCoefficients too.

Now the code use only the most compatible `AVIF_PIXEL_FORMAT_YUV420` format only.

Now I detect number of CPU threads and limit the maximum to 64 (I saw that limit in libaom code in the past).

I changed some C++ comments to C-style comments.

novomesk:
@joedrago or @linkmauve Do you have further comments/requirements? Could it be accepted in this form?

contrib/gdk-pixbuf/loader.c
```diff
@@ -187,11 +188,184 @@ static gboolean load_increment(gpointer data, const guchar * buf, guint size, GE
...
+        if ( gdk_pixbuf_get_n_channels(pixbuf) != 3) {
+            g_set_error(error,
+                        GDK_PIXBUF_ERROR,
+                        GDK_PIXBUF_ERROR_UNKNOWN_TYPE,
+                        "Unsupported number of channels");
+            return FALSE;
+        }
+    }
+
+    max_quantizer = AVIF_QUANTIZER_WORST_QUALITY * ( 100 - CLAMP(quality, 0, 100)) / 100;
```
joedrago, line 298: I'm sure you're aware, but this strategy probably doesn't really produce a smooth quality dial (and neither does my semi-arbitrary mapping in colorist). The "right answer" is going to be a more nuanced combination of min+max, only I don't know it yet.

Someday I hope to come up with a smoother mapping to a single quality value, and I hope this doesn't set a weird precedent for people.

novomesk: Of course, when you will have better calculations, change it.
I wanted to avoid using complicated curves and chose linear functions only. And for very low quality I start to increase maxQuantizerAlpha, otherwise HEIC achieve smaller file.

I observed that quality depends on image size too. For small images, the minQuantizer is more important, for large images, maxQuantizer has higher impact. When there is big distance between minQ and maxQ, we have little control over compression. My distance between minQ and maxQ is hot higher than 20 (still grant some freedom to encoder but not too much).
Based on my experiments I think it is quite good, but I am open to changes.

Only some apps have write support and only some of them will set the quality parameter. Most of the time, the default values will be used anyway.

PR #248, opened 2020-08-02
Add support for XRGB formats in RGB->YUV helpers

linkmauve:
These are similar to the ARGB formats, except the bits where there would
otherwise be the alpha channel are unused.

These formats are quite common, especially in real-time image generation
where proper alignment is more important than memory usage.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -399,6 +399,8 @@ typedef struct avifRGBImage
     avifRGBFormat format;                  // all channels are always full range
     avifChromaUpsampling chromaUpsampling; // How to upsample non-4:4:4 UV (ignored for 444) when converting to RGB.
                                            // Unused when converting to YUV. avifRGBImageSetDefaults() prefers quality over speed.
+    avifBool ignoreAlpha;                  // Used for XRGB formats, treats formats containing alpha (such as ARGB) as if they were
+                                           // RGB, treating the alpha bits as if they were all 1.
 
     uint8_t * pixels;
     uint32_t rowBytes;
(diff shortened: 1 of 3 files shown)


joedrago:
Hmm. I think this explosion of enum values for what is effectively a bool seems pretty heavy, and doesn't fully solve what I think you want it to solve. For example, `avifImageRGBToYUV()` is still going to fill alpha in this implementation, even if you don't want it to.

If your goal is to ignore alpha during RGB functions, I'd be happier with an explicit bool in `avifRGBImage` named something like `ignoreAlpha`, which explicitly avoids any alpha operations in any RGB functions.

(Rereading the code, I could probably move the call to `avifImageAllocatePlanes(image, AVIF_PLANES_A)` down into the alpha block later as well.)

linkmauve:
This is indeed a lot more elegant, thanks for the suggestion!

novomesk:
I'm using QImage::Format_RGBX64 format in Qt,  but X must be always 65535.
So alpha is ignored but X must be filled.

include/avif/avif.h
```diff
@@ -402,6 +402,8 @@ typedef struct avifRGBImage
 
     uint8_t * pixels;
     uint32_t rowBytes;
+    avifBool ignoreAlpha;                  // Used for XRGB formats, treats formats containing alpha (such as ARGB) as if they were
```
joedrago, line 405: Can you move these two new lines just below `chromaUpsampling`, and move the associated `ignoreAlpha` initialization in `avifRGBImageSetDefaults()`? I'll accept the PR after that.

linkmauve: Done, thanks!

Issue #262, opened 2020-08-10
avifDecoderDataFillImageGrid() should check if the tiles fail to cover the image grid


wantehchang:
avifDecoderDataFillImageGrid() does not check if the tiles fully cover the image grid when they are assembled. It should check that.

wantehchang:
I found that this check is specified in HEIF, Section 6.6.2.3.1:

> All input images shall have exactly the same width and height; call those `tile_width` and `tile_height`. The tiled input images shall completely “cover” the reconstructed image grid canvas, where `tile_width*columns` is greater than or equal to `output_width` and `tile_height*rows` is greater than or equal to `output_height`.

Note: The next paragraph in the spec is already implemented by `avifDecoderDataFillImageGrid()`:

> The reconstructed image is formed by tiling the input images into a grid with a column width (potentially excluding the right-most column) equal to `tile_width` and a row height (potentially excluding the bottom-most row) equal to `tile_height`, without gap or overlap, and then trimming on the right and the bottom to the indicated `output_width` and `output_height`.

Issue #264, opened 2020-08-11
Enforce grid tile size restrictions when chroma is subsampled


wantehchang:
MIAF Section 7.3.11.4.2 (Grid tile size) specifies some restrictions on grid tile size when chroma is subsampled:

- when the images are in the 4:2:2 chroma sampling format the horizontal tile offsets and widths, and the output width, shall be even numbers;
- when the images are in the 4:2:0 chroma sampling format, both the horizontal and vertical tile offsets and widths, and the output width annd height, shall be even numbers.

It also specifies the following constraints.

- the `tile_width` shall be greater than or equal to 64, and should be a multiple of 64
- the `tile_height` shall be greater than or equal to 64, and should be a multiple of 64

wantehchang:
The variables tile_width and tile_height are defined in HEIF, Section 6.6.2.3.1:

> All input images shall have exactly the same width and height; call those `tile_width` and `tile_height`.

PR #272, opened 2020-08-12
Validate grid image size and tile size

wantehchang:
Check the restrictions on grid image size and tile size in HEIF and
MIAF.

Fix https://github.com/AOMediaCodec/libavif/issues/262 and
https://github.com/AOMediaCodec/libavif/issues/264.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -699,6 +699,39 @@ static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
         }
     }
 
+    // Validate grid image size and tile size.
+    //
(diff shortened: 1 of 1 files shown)


joedrago:
As a style nit, I typically parenthesize chained binary ops to make it easier to skim and show intent, but I can tweak that later.

wantehchang:
> As a style nit, I typically parenthesize chained binary ops to make it easier to skim and show intent, but I can tweak that later.

Got it. I will write a pull request to fix this. (I also wanted to improve a comment in this pull request.)

PR #282, opened 2020-08-18
Transformations support (crop,rotate,mirror) in gdk-pixbuf loader

novomesk:
I tested with gcc 9.3.0 and with clang 10.0.1

@linkmauve you may take a look too.


joedrago:
If the two of you like this, I'll accept the PR. I just skimmed it to verify the transformations were done in the correct order (which they are):

> MIAF 7.3.6.7
> These properties, if used, shall be indicated to be applied in the following order: clean aperture first, then rotation, then mirror.

novomesk:
@joedrago I have no feedback from @linkmauve . Maybe he is enjoying holiday...

Also I think it is good to add this component to the automatic builds (other PR from jbeich). There are various combinations of compiler versions and platforms.

PR #285, opened 2020-08-21
Option to build with compilers that have more warnings

### LINKED ISSUE #284: find_package(rav1e) causes build error due to -Wpoison-system-directories
libavif fails to build with clang 12, because libavif sets `-Werror`, and clang 12 has added new warnings.

```cmake
        if(NOT TARGET rav1e)
            find_package(rav1e REQUIRED)
            set(AVIF_CODEC_INCLUDES ${AVIF_CODEC_INCLUDES} ${RAV1E_INCLUDE_DIR})
        endif()
```

I have `rav1e` in `/usr/local/{include,lib}` and `find_package` correctly finds it there. The problem is, that the result ends up in `AVIF_CODEC_INCLUDES`, which is then used as an include path.

Clang is given `-Werror`, and defaults to having `-Wpoison-system-directories`, which forbids adding `-I/usr/local/include`

> error: include location '/usr/local/include' is unsafe for cross-compilation [-Werror,-Wpoison-system-directories]

(for the record: I'm not cross-compiling)

----

I suggest using `-Werror` only in environment you control, such as your CI or your own development machines. It's asking for failures when the library is used with compilers you don't control.

kornelski:
Fixes #284

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,8 @@ set(LIBRARY_SOVERSION ${LIBRARY_VERSION_MAJOR})
 
 option(BUILD_SHARED_LIBS "Build shared avif library" ON)
 
+option(AVIF_ENABLE_WERROR "Enable all C compiler warnings and fail the build if any warning is emitted" ON)
+
 option(AVIF_CODEC_AOM "Use the AOM codec for encoding (and decoding if no other decoder is present)" OFF)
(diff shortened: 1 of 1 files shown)


joedrago:
Merged.

https://github.com/AOMediaCodec/libavif/commit/b177cc1e2e8145a2b51c59f33bfb15c25ecc915c

PR #286, opened 2020-08-23
Disable -Wpoison-system-directories if supported

wantehchang:
Disable the clang -Wpoison-system-directories warning if it is
supported. Its detection of cross compilation has false positives on
macOS because --sysroot is implicitly added.

This patch is one way to address issue 284.


wantehchang:
I asked Ben Nason to review this pull request. He said the change looks completely reasonable. He also suggested using target_compile_definitions() instead of add_definitions().

PR #293, opened 2020-08-28
Validate adds, casts in avifParseItemLocationBox

wantehchang:
Validate additions and uint32_t casts in avifParseItemLocationBox().


src/read.c
```diff
@@ -672,6 +672,7 @@ static const uint8_t * avifDecoderDataCalcItemPtr(avifDecoderData * data, avifDe
             return NULL;
         }
         if ((size_t)extent->size > remainingBytes) {
+            // This should be impossible?
```
wantehchang, line 675: I copied this comment from line 684:

```
    if (remainingBytes != 0) {
        // This should be impossible?
        return NULL;
    }
```

PR #295, opened 2020-08-29
gdk-pixbuf/loader.c: Include stdlib.h for strtol

Biswa96:
Add stdlib.h for strtol(3) used in .contrib/gdk-pixbuf/loader.c file.

### DIFF
--- a/contrib/gdk-pixbuf/loader.c
+++ b/contrib/gdk-pixbuf/loader.c
@@ -3,6 +3,7 @@
 */
 
 #include <avif/avif.h>
+#include <stdlib.h>
 
 #define GDK_PIXBUF_ENABLE_BACKEND
 #include <gdk-pixbuf/gdk-pixbuf-io.h>


include/avif/avif.h
```diff
@@ -6,6 +6,7 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include <stdlib.h> // for strtol(3)
```
wantehchang, line 9: Thank you for the pull request. The only file that uses strtol() is contrib/gdk-pixbuf/loader.c. Please move this #include <stdlib.h> statement to that file. I also suggest omitting the "for strtol(3)" comment. That kind of comment is not commonly used in libavif. Thanks!

novomesk:
@Biswa96 I'm curious, did you have problems to build the loader before this patch?

On which platform you experienced problems? You are known from MSYS2 project, so was it MSYS2?

Biswa96:
Yes it is in MSYS2/MinGW-Packages repository. Here is the error output:

```c
../contrib/gdk-pixbuf/loader.c: In function 'avif_image_saver':
../contrib/gdk-pixbuf/loader.c:227:27: error: implicit declaration of function 'strtol'; did you mean 'strtok'? [-Werror=implicit-function-declaration]
  227 |                 quality = strtol(*viter, &endptr, 10);
      |                           ^~~~~~
      |                           strtok
```

PR #296, opened 2020-08-29
Test gdk-pixbuf on TravisCI

jbeich:
gdk-pixbuf isn't tested on CI yet but Arch, Fedora (separate), FreeBSD, Gentoo (USE), MSYS2 already expose it downstream. This PR only covers FreeBSD as I'm not familar with other systems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,8 +56,8 @@ matrix:
           sudo pkg install -y llvm10
         fi
     before_script:
-      - sudo pkg install -y aom dav1d librav1e ninja
-      - $WRAPPER cmake -B build -G Ninja -DAVIF_{CODEC_{AOM,DAV1D,RAV1E},BUILD_{APPS,TESTS}}=ON
+      - sudo pkg install -y aom dav1d gdk-pixbuf2 librav1e ninja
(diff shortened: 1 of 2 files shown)


joedrago:
I'm sorry, I won't be accepting this one.

The whole point of the `contrib/` directory is to be an officially _unsupported_ list of contributions. I was already hesitant to accept this file in the first place, but I figured it'd be a one-and-done type of simple file people could use as a reference or point to. This kind of change would cause the official libavif CI builds to fail on this file, which completely betrays the notion of the contrib directory. 

This file has already seen way too much traffic. If it is becoming an important daily driver for people, perhaps @novomesk, @linkmauve , or @jbeich should make an official repo for `gdk-pixbuf-avif` and make `libavif` a git submodule in it, and I can remove it from the libavif repo entirely.

novomesk:
If we move the gdk-pixbuf loader to its own project, it will have very few users.

People would end up using [gdk-pixbuf loader from libheif project](https://github.com/strukturag/libheif/blob/master/gdk-pixbuf/pixbufloader-heif.c) instead which is officially supported there.

Being under shelter of libavif is a big advantage. The component is tested by more platforms and the bugs are discovered more quickly. libavif is becoming available on more and more platforms.

Right now (when we have support of transformations already), I consider the `libpixbufloader-avif.so` better for AVIF than gdk-pixbuf loader from libheif because we have save support too.

We can easily add animations support to the loader but can't do it right now with libheif one. I'd like to indicate there could be more commits in future.

PR #299, opened 2020-09-01
Make avifCodecSpecificOptionsGet a const function

wantehchang:
Originally the avifCodecSpecificValue struct had a boolean "used" field
and a codec needed to mark the codec-specific values it supported as
used. Therefore avifCodecSpecificValueArrayGet() had to return a
non-const avifCodecSpecificValue pointer. The "unsed" field has been
removed from the avifCodecSpecificOption struct. Therefore
avifCodecSpecificOptionsGet can return a const avifCodecSpecificOption
pointer.


wantehchang:
Joe: Depending on your future plans with avifCodecSpecificOptions, please feel free to reject this pull request. I only know the current version of libavif does not need avifCodecSpecificOptionsGet() to return a non-const pointer. Thanks!

joedrago:
This is a good change, but the existence of this function might not really be what we want anyway. If it is an error to feed an encoder and unknown/unsupported `csOption`, we're going to want to encourage walking the array, in which this function is meaningless. I'll accept this PR, but perhaps we should just remove the function.

wantehchang:
Joe: You are right. I didn't realize this function is unused. I will write a pull request to remove this function.

Issue #217, opened 2020-07-09
Add -tune option


Uch1haOb1to:
Visually, `-tune ssim` in libaom gives better quality than the default. I compared the frame received in libaom with the frame from
libavif, they are approximately the same with equal size, but if you use `-tune ssim` in libaom, then the quality is noticeably better.

joedrago:
@wantehchang and I were recently discussing codec specific options, and how to offer them in my API. I will bring it up with him again tomorrow and see if I can find a good way to do this.

Uch1haOb1to:
I’ll also add that for noisy photos it would be nice to have a noise filter like in libaom, which removes noise and then generates it.
Noisy photos, for example, from a Google camera with noise reduction turned off.
I compressed the photo 4 times, also compressed the same in Photoshop and I would not say that ssim is a big win in favor of avif.
orig [https://drive.google.com/file/d/1ffA7KeEt_1NPPE99pJIX-BcgqVD0rL5a/view?usp=sharing](url)
jpeg [https://drive.google.com/file/d/1Xh34LIwIT_6_XBE4E5flNZ5mkrkKAIj-/view?usp=sharing](url)
avif [https://drive.google.com/file/d/1D0GzwVvLTqKOrU69iLjAZfOIsMCiSNTM/view?usp=sharing](url)
avif with `-j 4 -y 420 -r full -s 1`, I don’t remember the quality anymore, I selected it so that the size was 4 times smaller than the original
ffmpeg ssim
jpeg `SSIM Y:0.938718 (12.126704) U:0.959499 (13.925375) V:0.959101 (13.882865) All:0.945579 (12.642336)`
avif `SSIM Y:0.946353 (12.704542) U:0.962295 (14.236041) V:0.961743 (14.172845) All:0.951575 (13.149299)`
According to my tests, with the settings that avif now has, it gives a tangible win in pictures without noise and with a small amount of detail, such as anime art.

joedrago:
Noise synthesis is being discussed at the codec level in AOM mailing lists, but is out of scope specifically for the AVIF container layer (this repo is about the container format). It is best to join the `av1-discuss` mailing list and bring this up there:

https://groups.google.com/a/aomedia.org/forum/#!forum/av1-discuss

PR #301, opened 2020-09-01
Add the libaom-specific option tune=METRIC

wantehchang:
METRIC can be psnr or ssim.

Fix https://github.com/AOMediaCodec/libavif/issues/217.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -92,6 +92,11 @@ static void syntax(void)
     printf("    --irot ANGLE                      : Add irot property (rotation). [0-3], makes (90 * ANGLE) degree rotation anti-clockwise\n");
     printf("    --imir AXIS                       : Add imir property (mirroring). 0=vertical, 1=horizontal\n");
     printf("\n");
+    if (avifCodecName(AVIF_CODEC_CHOICE_AOM, 0)) {
+        printf("aom-specific advanced options:\n");
+        printf("    tune=METRIC                       : Tune the encoder for distortion metric (psnr or ssim, default: psnr)\n");
+        printf("\n");
+    }
     avifPrintVersions();
 }
 
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -271,11 +271,6 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             }
         }
 
-        if (codec->csOptions->count > 0) {
-            // None are currently supported!
-            return AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION;
-        }
-
         codec->internal->aomFormat = avifImageCalcAOMFmt(image, alpha);
         if (codec->internal->aomFormat == AOM_IMG_FMT_NONE) {
             return AVIF_RESULT_UNKNOWN_ERROR;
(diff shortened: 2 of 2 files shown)


wantehchang:
Here is the syntax() message of avifenc:

$ ./avifenc 
Syntax: avifenc [options] input.[jpg|jpeg|png|y4m] output.avif
Options:
    -h,--help                         : Show syntax help
    -j,--jobs J                       : Number of jobs (worker threads, default: 1)
    -o,--output FILENAME              : Instead of using the last filename given as output, use this filename
    -l,--lossless                     : Set all defaults to encode losslessly, and emit warnings when settings/input don't allow for it
    -d,--depth D                      : Output depth [8,10,12]. (JPEG/PNG only; For y4m or stdin, depth is retained)
    -y,--yuv FORMAT                   : Output format [default=444, 422, 420, 400]. (JPEG/PNG only; For y4m or stdin, format is retained)
    --stdin                           : Read y4m frames from stdin instead of files; no input filenames allowed, must set before offering output filename
    --cicp,--nclx P/T/M               : Set CICP values (nclx colr box) (3 raw numbers, use -r to set range flag)
                                        P = enum avifColorPrimaries
                                        T = enum avifTransferCharacteristics
                                        M = enum avifMatrixCoefficients
                                        (use 2 for any you wish to leave unspecified)
    -r,--range RANGE                  : YUV range [limited or l, full or f]. (JPEG/PNG only, default: full; For y4m or stdin, range is retained)
    --min Q                           : Set min quantizer for color (0-63, where 0 is lossless)
    --max Q                           : Set max quantizer for color (0-63, where 0 is lossless)
    --minalpha Q                      : Set min quantizer for alpha (0-63, where 0 is lossless)
    --maxalpha Q                      : Set max quantizer for alpha (0-63, where 0 is lossless)
    --tilerowslog2 R                  : Set log2 of number of tile rows (0-6, default: 0)
    --tilecolslog2 C                  : Set log2 of number of tile columns (0-6, default: 0)
    -s,--speed S                      : Encoder speed (0-10, slowest-fastest, 'default' or 'd' for codec internal defaults. default speed: 8)
    -c,--codec C                      : AV1 codec to use (choose from versions list below)
    -a,--advanced KEY[=VALUE]         : Pass an advanced, codec-specific key/value string pair directly to the codec. avifenc will warn on any not used by the codec.
    --duration D                      : Set all following frame durations (in timescales) to D; default 1. Can be set multiple times (before supplying each filename)
    --timescale,--fps V               : Set the timescale to V. If all frames are 1 timescale in length, this is equivalent to frames per second
    -k,--keyframe INTERVAL            : Set the forced keyframe interval (maximum frames between keyframes). Set to 0 to disable (default).
    --ignore-icc                      : If the input file contains an embedded ICC profile, ignore it (no-op if absent)
    --pasp H,V                        : Add pasp property (aspect ratio). H=horizontal spacing, V=vertical spacing
    --clap WN,WD,HN,HD,HON,HOD,VON,VOD: Add clap property (clean aperture). Width, Height, HOffset, VOffset (in num/denom pairs)
    --irot ANGLE                      : Add irot property (rotation). [0-3], makes (90 * ANGLE) degree rotation anti-clockwise
    --imir AXIS                       : Add imir property (mirroring). 0=vertical, 1=horizontal

    tune=METRIC                       : Tune the encoder for distortion metric (psnr or ssim, default: psnr)

Version: 0.8.1 (dav1d:0.7.1-0-ge9df70c, aom:2.0.0, rav1e:0.3.3 (v0.3.3))

Note that I call it "aom-specific" rather than "libaom-specific" because the version string uses the name "aom".

joedrago:
Awesome! I'll accept this when the builders finish chewing on it.

wantehchang:
Thank you, Joe. This pull request does not use the avifCodecSpecificOptionsGet() function that we discussed earlier today. So we should be able to safely remove that function.

PR #308, opened 2020-09-09
Add man pages for command-line tools

sorairolake:
Add man pages for command-line tools (`avifenc` and `avifdec`). It is written in AsciiDoc, so it is required [Asciidoctor](https://asciidoctor.org/) or [AsciiDoc](https://asciidoc.org/) as dependencies at build-time on Unix-like systems.

### DIFF
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -1,0 +1,6 @@
+# Copyright 2020 Shun Sakai. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+if(UNIX OR NOT MSVC)
+    add_subdirectory(man/man1)
+endif()
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,23 @@ dist: bionic
 addons:
   apt:
     packages:
+      - docbook-xml
+      - docbook-xsl
+      - libxml2-utils
       - ninja-build
       - python3
       - python3-pip
+      - xsltproc
 
 before_script:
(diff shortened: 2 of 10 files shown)


joedrago:
This is cool, but I have a few thoughts:

* The licensing headers in the new files aren't exactly correct, as I didn't write them. You'll probably want to put your name there.

* I'm unfamiliar with AsciiDoctor. Is this a common manpages tool? Is there no reason to use something more GitHub friendly, like GitHub-flavored Markdown, and convert *that* to manpages? I'd prefer to not add heavy dependencies, even if they're optional. What is the common practice for manpage generation?

docs/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# Copyright 2020 Joe Drago. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+if(APPLE OR UNIX)
+    add_subdirectory(man)
+endif()
```
1480c1, line 6: Maybe better to use `NOT MSVC` as the conditional since `MINGW` still uses man pages, but they do not set `UNIX` to true

1480c1: Also, the `APPLE` part is redundant since `UNIX` is set when `APPLE` is also set

sorairolake:
@joedrago Fix to support the original implementation too. It is written in Python and [Python 3 port exist](https://github.com/asciidoc/asciidoc-py3), so I think it is improve the dependencies issues.

AsciiDoc is used in many projects such as Git, and support the description list. So, I think it is better than Markdown for the purpose of writing the man page.

PR #314, opened 2020-09-16
Document how to build dav1d with asan or ubsan

wantehchang:
Also change "might required" to "might require".

### DIFF
--- a/ext/dav1d.cmd
+++ b/ext/dav1d.cmd
@@ -14,7 +14,9 @@ cd dav1d
 mkdir build
 cd build
 
-: # macOS might required: -Dc_args=-fno-stack-check
+: # macOS might require: -Dc_args=-fno-stack-check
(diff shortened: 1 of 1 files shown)


joedrago:
I'll immediately accept any comments-only PRs you have for this file, but it looks like you're still tweaking this. Let me know when you're done and I'll merge it.

wantehchang:
Joe: I am done. I suggest merging only the first commit if you can. I found that we do not turn off the assembly code in dav1d when we build avif_decode_fuzzer:
https://github.com/google/oss-fuzz/blob/master/projects/libavif/build.sh

Thank you!

joedrago:
I `cherry-pick`'d the first commit into `master`. This should be all done.

https://github.com/AOMediaCodec/libavif/commit/16d429f49003f44c901d19c4e4956627d4d13b8c

PR #319, opened 2020-09-18
No need to set AV1D_SET_OUTPUT_ALL_LAYERS to 0

wantehchang:
The default value of AV1D_SET_OUTPUT_ALL_LAYERS is 0, so it is not
necessary to set AV1D_SET_OUTPUT_ALL_LAYERS to 0.


wantehchang:
It is unusual that dav1d's default value of all_layers is 1. In libaom and libgav1, the default is 0.

If you'd like to keep this code for emphasis, please feel free to reject this pull request.

joedrago:
I think I want to keep it there for emphasis. I'll close this.

PR #320, opened 2020-09-18
Cleanup in avifPrepareReformatState()

wantehchang:
Return AVIF_FALSE if image->matrixCoefficients is
AVIF_MATRIX_COEFFICIENTS_SMPTE2085. Calculate state->uvBias using the
formula 1 << (image->depth - 1).

Also update a comment in matrixCoefficientsTables to use the correct
enum name AVIF_MATRIX_COEFFICIENTS_SMPTE2085.


src/reformat.c
```diff
@@ -101,15 +102,7 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
-    if (image->depth == 8) {
-        state->uvBias = 128;
-    } else if (image->depth == 10) {
-        state->uvBias = 512;
-    } else if (image->depth == 12) {
-        state->uvBias = 2048;
-    } else {
-        return AVIF_FALSE;
-    }
+    state->uvBias = 1 << (image->depth - 1);
```
wantehchang, line 105: This formula is in Equations 30 and 31 in Sec. 8.3 of H.273.

PR #321, opened 2020-09-18
No need to negate alphaPass and item->alpha

wantehchang:
alphaPass and item->alpha are of the avifBool type, so we can safely
assume they hold boolean values (0 or 1). It is not necessary to ensure
a boolean value by using the ! operator.


wantehchang:
Please feel free to reject this pull request if you'd like to keep the defensive programming. I was confused by this code. (I have seen code like !!ptr to convert a pointer to a boolean value.)

maximos08 (approved):

joedrago:
This change has already been completed in another commit.

wantehchang:
Thank you, Joe!

PR #327, opened 2020-09-23
avifIO: Remove API functions that take FILE* arguments

dwbuiten:
```
It is not safe to have to FILE* arguments in a public API, since
the library and caller are not necessarily built with the same
libc. This is particularily common on Windows, where things
are commonly linked against different MSVCRT versions, where
the FILE internals may differ.
```

I think this should be OK to do, since no release version has been tagged since this API was introduced.

I am unsure whether a SONAME bump is required here, given the above.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -6,7 +6,6 @@
 
 #include <stddef.h>
 #include <stdint.h>
-#include <stdio.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -571,7 +570,6 @@ typedef struct avifIO
 
 avifIO * avifIOCreateMemoryReader(const uint8_t * data, size_t size);
 avifIO * avifIOCreateFileReader(const char * filename);
-avifIO * avifIOCreateFilePtrReader(FILE * f, avifBool closeOnDestroy);
 void avifIODestroy(avifIO * io);
 
 // ---------------------------------------------------------------------------
@@ -701,7 +699,6 @@ avifResult avifDecoderReadFile(avifDecoder * decoder, avifImage * image, const c
 avifResult avifDecoderSetSource(avifDecoder * decoder, avifDecoderSource source);
 avifResult avifDecoderSetIO(avifDecoder * decoder, avifIO * io);
 avifResult avifDecoderSetIOMemory(avifDecoder * decoder, const avifROData * rawInput);
-avifResult avifDecoderSetIOFilePtr(avifDecoder * decoder, FILE * f, avifBool closeOnDestroy);
 avifResult avifDecoderSetIOFile(avifDecoder * decoder, const char * filename);
 avifResult avifDecoderParse(avifDecoder * decoder);
 avifResult avifDecoderNextImage(avifDecoder * decoder);
--- a/src/io.c
+++ b/src/io.c
@@ -70,7 +70,6 @@ typedef struct avifIOFileReader
     avifIO io; // this must be the first member for easy casting to avifIO*
     avifRWData buffer;
     FILE * f;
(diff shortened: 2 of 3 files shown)


include/avif/avif.h
```diff
@@ -571,7 +571,6 @@ typedef struct avifIO
 
 avifIO * avifIOCreateMemoryReader(const uint8_t * data, size_t size);
 avifIO * avifIOCreateFileReader(const char * filename);
-avifIO * avifIOCreateFilePtrReader(FILE * f, avifBool closeOnDestroy);
```
wantehchang, line 574: Could you also remove #include <stdio.h> from avif.h? That #include statement was added for this function.

src/io.c
```diff
@@ -122,7 +122,7 @@ static void avifIOFileReaderDestroy(struct avifIO * io)
     avifFree(io);
 }
 
-avifIO * avifIOCreateFilePtrReader(FILE * f, avifBool closeOnDestroy)
+static avifIO * avifIOCreateFilePtrReader(FILE * f, avifBool closeOnDestroy)
```
wantehchang, line 125: As an internal function, we don't need the closeOnDestroy parameter, and we don't need the reader->closeOnDestroy field, because they are always set to true.

wantehchang (changes requested):
Derek: Thanks for writing the pull request. I also noticed the "File" and "FilePtr" functions are merely convenience functions and don't think they need to be in the public API. They can be moved to the apps/ directory to support the apps, examples, and tests.

I suggest two small changes to your pull request. Please let me know if you would like me to take care of these. Thanks,.

src/io.c
dwbuiten: Done.

include/avif/avif.h
dwbuiten: Done.

dwbuiten:
Both request changes applied.

src/io.c
```diff
@@ -135,7 +135,7 @@ avifIO * avifIOCreateFilePtrReader(FILE * f, avifBool closeOnDestroy)
     avifIOFileReader * reader = avifAlloc(sizeof(avifIOFileReader));
     memset(reader, 0, sizeof(avifIOFileReader));
     reader->f = f;
-    reader->closeOnDestroy = closeOnDestroy;
+    reader->closeOnDestroy = AVIF_TRUE;
```
wantehchang, line 138: Could you also delete the closeOnDestroy field of struct avifIOFileReader at line 73? Thanks.

src/io.c
```diff
@@ -122,7 +122,7 @@ static void avifIOFileReaderDestroy(struct avifIO * io)
     avifFree(io);
 }
 
-avifIO * avifIOCreateFilePtrReader(FILE * f, avifBool closeOnDestroy)
+static avifIO * avifIOCreateFilePtrReader(FILE * f)
```
wantehchang, line 125: We don't need to have avifIOCreateFilePtrReader() as a separate function. Please manually inline the only call to avifIOCreateFilePtrReader(). Thanks.

wantehchang (changes requested):
I suggest two changes (both of them have been posted as separate comments. I'm still learning to use the github code review UI).

src/io.c
dwbuiten: Good catch. Done.

src/io.c
dwbuiten: Done.

dwbuiten:
Thanks. I have addressed both comments.

joedrago:
This looks good. I can accept it when it survives CI.

wantehchang (approved):
LGTM. Thank you!

Issue #326, opened 2020-09-23
-Werror=clobbered


novomesk:
The favorite warning strikes back:
```
/var/tmp/portage/media-libs/libavif-9999/work/libavif-9999/apps/shared/avifpng.c: In function ‘avifPNGWrite’:
/var/tmp/portage/media-libs/libavif-9999/work/libavif-9999/apps/shared/avifpng.c:161:12: error: variable ‘f’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Werror=clobbered]
  161 |     FILE * f = NULL;
      |            ^
/var/tmp/portage/media-libs/libavif-9999/work/libavif-9999/apps/shared/avifpng.c:166:9: error: variable ‘rgbDepth’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Werror=clobbered]
  166 |     int rgbDepth = requestedDepth;
      |         ^~~~~~~~
cc1: all warnings being treated as errors
```

wantehchang:
Daniel: Thank you for the bug report. I will take a look at this.

Does this mean none of the CI builds use clang?

novomesk:
I don't know, my compiler is `gcc (Gentoo 9.3.0-r1 p3) 9.3.0`
cmake prints `-- The C compiler identification is GNU 9.3.0`

wantehchang:
Yes, I was mistaken. It's gcc that has this warning.

PR #328, opened 2020-09-23
Fix GCC -Wclobbered warnings

wantehchang:
GCC 9.3.0 warns that the variables 'f' and 'rgbDepth' in avifPNGWrite()
might be clobbered by 'longjmp' or 'vfork'. I am not familiar with
'vfork', but I believe this is a false positive for 'longjmp' because
those two variables are not modified between the 'setjmp' and 'longjmp'
calls.

Fix the warnings by moving the relevant code back to where it was before
commit 6ce2d81653878e4bef5304495a2876232b1c515c.

I tested with the command:

  avifdec Chimera_10bit_cropped_to_1920x1008_with_HDR_metadata.avif output.png

and got the expected error message:

  Conversion to RGB failed: output.png

Fix https://github.com/AOMediaCodec/libavif/issues/326.

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -158,30 +158,11 @@ avifBool avifPNGWrite(avifImage * avif, const char * outputFilename, uint32_t re
     png_structp png = NULL;
     png_infop info = NULL;
     png_bytep * volatile rowPointers = NULL;
-    FILE * f = NULL;
 
     avifRGBImage rgb;
     memset(&rgb, 0, sizeof(avifRGBImage));
 
-    int rgbDepth = requestedDepth;
-    if (rgbDepth == 0) {
-        if (avif->depth > 8) {
-            rgbDepth = 16;
-        } else {
-            rgbDepth = 8;
-        }
-    }
-
-    avifRGBImageSetDefaults(&rgb, avif);
-    rgb.depth = rgbDepth;
-    rgb.chromaUpsampling = chromaUpsampling;
-    avifRGBImageAllocatePixels(&rgb);
-    if (avifImageYUVToRGB(avif, &rgb) != AVIF_RESULT_OK) {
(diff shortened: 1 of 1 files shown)


wantehchang:
Joe: There are other ways to fix the GCC -Wclobbered warnings. With this patch, the net change to avifpng.c is:

$ git diff 2fadd1d19815317fd444d177623b9c1458f7c637 apps/shared/avifpng.c
diff --git a/apps/shared/avifpng.c b/apps/shared/avifpng.c
index 02cec10..7b57961 100644
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -210,7 +210,10 @@ avifBool avifPNGWrite(avifImage * avif, const char * outputFilename, uint32_t re
     rgb.depth = rgbDepth;
     rgb.chromaUpsampling = chromaUpsampling;
     avifRGBImageAllocatePixels(&rgb);
-    avifImageYUVToRGB(avif, &rgb);
+    if (avifImageYUVToRGB(avif, &rgb) != AVIF_RESULT_OK) {
+        fprintf(stderr, "Conversion to RGB failed: %s\n", outputFilename);
+        goto cleanup;
+    }
     rowPointers = (png_bytep *)malloc(sizeof(png_bytep) * rgb.height);
     for (uint32_t y = 0; y < rgb.height; ++y) {

Is it okay to do "goto cleanup" in the middle of the PNG call sequence? My testing shows it seems to work, but I am not familiar with libpng.

wantehchang:
By the net change to avifpng.c, I mean the net change of your commit 6ce2d81653878e4bef5304495a2876232b1c515c and this pull request.

joedrago:
My original change looked a lot like your proposed change, but the reason I reordered it was because I was trying to avoid opening the file for write if I knew I wasn't going to be able to write it at all. Perhaps we can find an approach where we do this without hitting the clobbered warning?

wantehchang:
Yes, I can do that.

PR #332, opened 2020-09-24
Miscellaneous minor comment fixes

wantehchang:
include/avif/internal.h: Change "full box" to "whole box" to avoid being
misinterpreted as the FullBox type in ISO BMFF.

src/read.c: Describe how the ftypData field of avifDecoderData is used.
Move the Note for avifParse() to the outside of the function to make it
more prominent. Remove an extra "This".

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -524,7 +524,7 @@ static avifDecoderItem * avifMetaFindItem(avifMeta * meta, uint32_t itemID)
 typedef struct avifDecoderData
 {
     avifFileType ftyp;
-    avifRWData ftypData; // cloned from avifIO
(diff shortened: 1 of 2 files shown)


src/read.c
```diff
@@ -524,7 +524,7 @@ static avifDecoderItem * avifMetaFindItem(avifMeta * meta, uint32_t itemID)
 typedef struct avifDecoderData
 {
     avifFileType ftyp;
-    avifRWData ftypData; // cloned from avifIO
+    avifRWData ftypData; // Contents of the ftyp box. Cloned from avifIO. ftyp.compatibleBrands points into ftypData.
```
wantehchang, line 527: A more robust design might be to have ftyp.compatibleBrands own the memory that it points to. In this pull request I only document the current design.

wantehchang:
This pull request has become out of date.

PR #333, opened 2020-09-24
Remove ftyp and ftypData from avifDecoderData

wantehchang:
We can remove the ftyp and ftypData fields from avifDecoderData if we
are willing to call avifFileTypeIsCompatible() in avifParse() rather
than after the avifParse() call.


wantehchang:
Joe: Please feel free to decline this pull request. I am not sure if you would prefer to keep avifParse() a parsing-only function and process the parsed data outside avifParse().

joedrago:
I like this one. We can always reintroduce this in a different way if we want to expose ftyp to the enduser.

PR #335, opened 2020-09-25
avifROStreamHasBytesLeft: avoid adding byteCount

wantehchang:
It is not clear whether the byteCount parameter of
avifROStreamHasBytesLeft always comes from a trusted source, so adding
byteCount to the stream offset could result in an integer overflow.
Avoid this potential risk by subtracting the stream offset from the
stream size, which is safe to do when the stream is valid.


wantehchang:
Joe: This is similar to https://github.com/AOMediaCodec/libavif/commit/a6d78740159dab3e85b441f19902d1cd3d344d15.

PR #336, opened 2020-09-27
Add version option to avifenc and avifdec

sorairolake:
This is an option that only print the version number.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -48,6 +48,7 @@ static void syntax(void)
     printf("Syntax: avifenc [options] input.[jpg|jpeg|png|y4m] output.avif\n");
     printf("Options:\n");
     printf("    -h,--help                         : Show syntax help\n");
+    printf("    -V,--version                      : Show the version number\n");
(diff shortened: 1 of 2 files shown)


apps/avifenc.c
```diff
@@ -102,9 +103,7 @@ static void syntax(void)
         printf("    end-usage=MODE                    : Rate control mode (vbr, cbr, cq, or q)\n");
         printf("    sharpness=S                       : Loop filter sharpness (0-7, default: 0)\n");
         printf("    tune=METRIC                       : Tune the encoder for distortion metric (psnr or ssim, default: psnr)\n");
-        printf("\n");
```
joedrago, line 105: I'd like this newline in the AOM block to stay, for future expansions.

joedrago:
Also, why would you stop printing the version in the full syntax? (I'm curious as to the motivation for this PR in general.)

sorairolake:
@joedrago Because I thought it is redundant to print the version even in `syntax()` if the version option exist.

The purpose of this PR is to add an option to print only the version. It is not the purpose to stop printing the version in `syntax()`, so I will restore it.

PR #338, opened 2020-09-28
Read fails if offset is past end of buffer/fil

wantehchang:
Change avifIOMemoryReaderRead() and avifIOFileReaderRead() to return the
AVIF_RESULT_BMFF_PARSE_FAILED error if offset is past the end of the
buffer or file, because this can only happen if the size field of some
box (which we encountered but did not read the contents of) is too big.

In avifIOMemoryReaderRead(), if offset is exactly at the end of the
buffer, set out->data equal to the end of the buffer rather than the
beginning of the buffer. This is just for convenience and should not
matter because out->size is set equal to 0.

Also restructure avifIOMemoryReaderRead() and avifIOFileReaderRead() to
make the two functions as similar as possible. This minimizes the
behavior differences between the two functions.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -31,12 +31,13 @@ static avifResult avifIOMemoryReaderRead(struct avifIO * io, uint32_t readFlags,
     avifIOMemoryReader * reader = (avifIOMemoryReader *)io;
 
     // Sanitize/clamp incoming request
-    if (offset >= reader->rodata.size) {
-        offset = 0;
-        size = 0;
+    if (offset > reader->rodata.size) {
+        // The offset is past the end of the buffer.
+        return AVIF_RESULT_IO_ERROR;
     }
-    if (size > (reader->rodata.size - offset)) {
-        size = reader->rodata.size - offset;
+    uint64_t availableSize = reader->rodata.size - offset;
+    if (size > availableSize) {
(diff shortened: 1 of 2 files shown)


joedrago:
I like the changes and protections offered in this PR in a vacuum, but I don't like where they are or how they signal an error. 

The core issue here is that these functions are effectively across an _API boundary_, and can/will be replaced by endusers implementing their own readers. I'd prefer to distribute out any kinds of overflow checks to the calling code to avoid every implementer of `avifIO` having to do `uint64_t` overflow checks. Also, `AVIF_RESULT_BMFF_PARSE_FAILED` is overreaching here as the `avifIO` implementation cannot possibly know that we're parsing BMFF, as we might be reading samples or metadata out of the file after the parse.

I could be persuaded to leave the overflow checks here if they don't make a lot of sense in `avifParse()` (say), but at a minimum, we should be returning something like `AVIF_RESULT_IO_ERROR`, not anything with BMFF in it.

wantehchang:
Joe: Thank you very much for the review and suggestions! Sorry about the late reply. I had to prepare for an important meeting this morning.

I just pushed a new commit for review. I hope I implemented your suggestions correctly. Please take a look.

PR #343, opened 2020-10-06
Cleanup related to avifCodecGetNextImageFunc

wantehchang:
Declare the 'sample' input parameter as const.

The avifCodecGetNextImageFunc implementations can assume the 'sample'
input parameter is not NULL.

The for (;;) loop in aomCodecGetNextImage() needs to set 'sample' to
NULL after 'sample' has been passed to a successful aom_codec_decode()
call.

Delete comments like "Check if there are more samples to feed" and
"Feed another sample" because they describe the following check in the
original code:
 codec->internal->inputSampleIndex < codec->decodeInput->samples.count
which has been deleted.

### DIFF
--- a/src/codec_libgav1.c
+++ b/src/codec_libgav1.c
@@ -33,28 +33,24 @@ static avifBool gav1CodecOpen(avifCodec * codec)
     return AVIF_TRUE;
 }
 
-static avifBool gav1CodecGetNextImage(struct avifCodec * codec, avifDecodeSample * sample, avifBool alpha, avifImage * image)
+static avifBool gav1CodecGetNextImage(struct avifCodec * codec, const avifDecodeSample * sample, avifBool alpha, avifImage * image)
 {
-    const Libgav1DecoderBuffer * nextFrame = NULL;
-    // Check if there are more samples to feed
-    if (sample) {
-        // Feed another sample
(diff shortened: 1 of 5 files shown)


src/codec_libgav1.c
```diff
@@ -33,28 +33,24 @@ static avifBool gav1CodecOpen(avifCodec * codec)
...
+                                   /*buffer_private_data=*/NULL) != kLibgav1StatusOk) {
+        return AVIF_FALSE;
+    }
+    // Each Libgav1DecoderDequeueFrame() call invalidates the output frame
+    // returned by the previous Libgav1DecoderDequeueFrame() call. Clear
+    // our pointer to the previous output frame.
+    codec->internal->gav1Image = NULL;
     const Libgav1DecoderBuffer * nextFrame = NULL;
-    // Check if there are more samples to feed
-    if (sample) {
```
wantehchang, line 40: It is not easy to see the changes in this diff view. There are only two changes here:

1. I removed the if (sample) check because we can assume 'sample' is not NULL.
2. I moved the declaration of nextFrame (at line 38) to be right before the Libgav1DecoderDequeueFrame() call.

wantehchang: Argh, there is a third change: I removed the "Check if there are more samples to feed" and "Feed another sample" comments.

PR #345, opened 2020-10-06
Fix two bugs in avifDecoderReadItem()

wantehchang:
The two bug fixes are:

1. Change the test item->idatID == 1 to item->idatID != 0.

2. In the idatBuffer case, consider extent->offset when setting up
offsetBuffer.

Also make three cleanup changes:

1. In the !singlePersistentBuffer case, the size of item->mergedExtents
just needs to be totalBytesToRead, which may be less than item->size if
partialByteCount is specified. This requires moving forward the code
that sets totalBytesToRead.

2. In a memcpy() call, the second argument could be &item->mergedExtents
(the address of the struct) rather than &item->mergedExtents.data (the
address of the first member of the struct).

3. Remove a bytesToRead != offsetBuffer.size check, which is redundant
with the bytesToRead > offsetBuffer.size check later. At that point, we
know bytesToRead >= offsetBuffer.size, so checking
bytesToRead != offsetBuffer.size is equivalent to checking
bytesToRead > offsetBuffer.size.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -627,7 +627,7 @@ static avifResult avifDecoderReadItem(avifDecoder * decoder, avifDecoderItem * i
 {
     if (item->mergedExtents.data && !item->partialMergedExtents) {
         // Multiple extents have already been concatenated for this item, just return it
-        memcpy(outData, &item->mergedExtents.data, sizeof(avifROData));
+        memcpy(outData, &item->mergedExtents, sizeof(avifROData));
         return AVIF_RESULT_OK;
     }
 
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -702,9 +709,6 @@ static avifResult avifDecoderReadItem(avifDecoder * decoder, avifDecoderItem * i
             if (readResult != AVIF_RESULT_OK) {
                 return readResult;
             }
```
wantehchang, line 711: At this point, we have just called decoder->io->read(decoder->io, 0, extent->offset, bytesToRead, &offsetBuffer)  successfully. Therefore we know bytesToRead >= offsetBuffer.size. That implies the check at line 705 is equivalent to the check at line 710.

PR #348, opened 2020-10-08
CI: test system libgav1 and Clang 11

jbeich:
- I've [packaged](https://www.freshports.org/multimedia/libgav1) libgav1 0.16.0 which recenetly became available in (default) `/quarterly` set
- Clang 11 is approaching release. It's already used on FreeBSD 13.0 ([in development](https://www.freebsd.org/releases/13.0R/schedule.html))


wantehchang (approved):
LGTM. Thanks!

PR #349, opened 2020-10-08
Miscellaneous cleanup changes for src/io.c

wantehchang:
In the debug printf statements, use the PRIu64 format to print the
uint64_t parameters 'offset' and 'size'.

In avifIOFileReaderRead() and avifIOFileReaderRead(), assume reader->f
is not NULL. This is guaranteed by avifIOCreateFileReader().

In avifIOFileReaderRead(), make sure 'offset' is in the range of 'long'
before casting it to 'long'. 'offset' is of the uint64_t type. On a
32-bit platform and also 64-bit Windows, 'long' is 32 bits. So casting
'offset' to 'long' could be problematic.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -3,6 +3,7 @@
 
 #include "avif/internal.h"
 
+#include <limits.h>
 #include <stdio.h>
 #include <string.h>
 
@@ -24,7 +25,7 @@ typedef struct avifIOMemoryReader
 
 static avifResult avifIOMemoryReaderRead(struct avifIO * io, uint32_t readFlags, uint64_t offset, uint64_t size, avifROData * out)
 {
-    // printf("avifIOMemoryReaderRead offset %zu size %zu\n", offset, size);
+    // printf("avifIOMemoryReaderRead offset %" PRIu64 " size %" PRIu64 "\n", offset, size);
 
     (void)readFlags;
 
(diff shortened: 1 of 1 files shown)


src/io.c
```diff
@@ -24,7 +25,7 @@ typedef struct avifIOMemoryReader
 
 static avifResult avifIOMemoryReaderRead(struct avifIO * io, uint32_t readFlags, uint64_t offset, uint64_t size, avifROData * out)
 {
-    // printf("avifIOMemoryReaderRead offset %zu size %zu\n", offset, size);
+    // printf("avifIOMemoryReaderRead offset %" PRIu64 " size %" PRIu64 "\n", offset, size);
```
wantehchang, line 28: Joe: Would you like me to just delete these two printf statements?

src/io.c
```diff
@@ -95,6 +93,9 @@ static avifResult avifIOFileReaderRead(struct avifIO * io, uint32_t readFlags, u
     }
 
     if (size > 0) {
+        if (offset > LONG_MAX) {
+            return AVIF_RESULT_IO_ERROR;
+        }
         if (reader->buffer.size < size) {
             avifRWDataRealloc(&reader->buffer, size);
         }
```
wantehchang, line 101: The next line is:
        if (fseek(reader->f, (long)offset, SEEK_SET) != 0) {

So we need to make sure the (long)offset cast is safe.

PR #350, opened 2020-10-08
Miscellaneous comment fixes

wantehchang:
include/avif/avif.h:
* In the comment for the avifIODestroyFunc type, change "the reader" to
  "the avifIO object".
* In the comment for the avifIOReadFunc type, change "another call" to
  "another read call", and change the return value from AVIF_RESULT_OK
  to AVIF_RESULT_IO_ERROR when offset exceeds the size of the content
  (past EOF).
* In the comment for the 'persistent' field of the avifIO struct, change
  "the avifDecoder object" to "the avifIO object".
* In the comment for the 'io' field of the avifIO struct, change "the IO
  implementation" to "the avifIO implementation".
* In the comment for the avifDecoderRead() function, change "deestroyed"
  "destroyed".

src/read.c:
* In avifParse(), remove an extraneous "This".
* In avifDecoderPrepareSample(), change "make reuse the buffer" to
  "reuse the buffer".

src/write.c:
* In avifEncoderFinish(), change "alpha coming first" to "alpha coming
  before color" because now it is Exif and XMP that are packed first.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -523,20 +523,20 @@ typedef struct avifCodecConfigurationBox
 struct avifIO;
 
 // Destroy must completely destroy all child structures *and* free the avifIO object itself.
-// This function pointer is optional, however, if the reader isn't intended to be owned by
+// This function pointer is optional, however, if the avifIO object isn't intended to be owned by
 // a libavif encoder/decoder.
 typedef void (*avifIODestroyFunc)(struct avifIO * io);
 
-// This function should return a block of memory that *must* remain valid until another call to this
-// avifIO struct is used (reusing a read buffer is acceptable/expected).
+// This function should return a block of memory that *must* remain valid until another read call to
+// this avifIO struct is made (reusing a read buffer is acceptable/expected).
 //
+// * If offset exceeds the size of the content (past EOF), return AVIF_RESULT_IO_ERROR.
 // * If (offset+size) does not exceed the contents' size but the *entire range* is unavailable yet
 //   (due to network conditions or any other reason), return AVIF_RESULT_WAITING_ON_IO.
 // * If (offset+size) does not exceed the contents' size, it must provide the *entire range* and
 //   return AVIF_RESULT_OK.
 // * If (offset+size) exceeds the contents' size, it must provide a truncated buffer that provides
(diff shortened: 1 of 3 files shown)


include/avif/avif.h
```diff
@@ -523,20 +523,20 @@ typedef struct avifCodecConfigurationBox
 struct avifIO;
 
 // Destroy must completely destroy all child structures *and* free the avifIO object itself.
-// This function pointer is optional, however, if the reader isn't intended to be owned by
```
wantehchang, line 526: At this point, it is not clear what "reader" refers to. Also, avifIO is also designed for writing (not implemented yet). So I changed "reader" to "avifIO object".

include/avif/avif.h
```diff
@@ -523,20 +523,20 @@ typedef struct avifCodecConfigurationBox
...
+// This function pointer is optional, however, if the avifIO object isn't intended to be owned by
 // a libavif encoder/decoder.
 typedef void (*avifIODestroyFunc)(struct avifIO * io);
 
-// This function should return a block of memory that *must* remain valid until another call to this
-// avifIO struct is used (reusing a read buffer is acceptable/expected).
+// This function should return a block of memory that *must* remain valid until another read call to
```
wantehchang, line 530: This change assumes a write call (not implemented yet) won't modify the internal read buffer.

include/avif/avif.h
```diff
@@ -523,20 +523,20 @@ typedef struct avifCodecConfigurationBox
...
+// this avifIO struct is made (reusing a read buffer is acceptable/expected).
 //
+// * If offset exceeds the size of the content (past EOF), return AVIF_RESULT_IO_ERROR.
```
wantehchang, line 533: This reflects the change in https://github.com/AOMediaCodec/libavif/commit/b856dc2750bf54cc90d67074b92650c9a9ad7db3

include/avif/avif.h
```diff
@@ -557,13 +557,13 @@ typedef struct avifIO
     size_t sizeHint;
 
     // If true, *all* memory regions returned from *all* calls to read are guaranteed to be
-    // persistent and exist for the lifetime of the avifDecoder object. If false, libavif will make
```
wantehchang, line 560: I assume "avifDecoder" is a typo.

PR #351, opened 2020-10-08
Rewrite a loop as a do-while loop

wantehchang:
A for (;;) loop with an "if (cond) break;" statement at the end of the
loop body is equivalent to a "do {} while (!done);" loop.


wantehchang:
Joe: This pull request is optional.

PR #355, opened 2020-10-08
avifDecoderReadItem callers can rely on contract

wantehchang:
The callers of avifDecoderReadItem() with partialByteCount=0 can rely on
the API contract and assume that outData is valid (i.e., outData->data
is not NULL) and full (i.e., outData->size == item->size) when
avifDecoderReadItem() returns AVIF_RESULT_OK.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2485,10 +2485,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                 if (readResult != AVIF_RESULT_OK) {
                     return readResult;
                 }
-                if ((readData.data == NULL) || (readData.size != item->size)) {
-                    return AVIF_RESULT_BMFF_PARSE_FAILED;
-                }
-                if (!avifParseImageGridBox(&data->colorGrid, readData.data, item->size)) {
+                if (!avifParseImageGridBox(&data->colorGrid, readData.data, readData.size)) {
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -2485,10 +2485,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                 if (readResult != AVIF_RESULT_OK) {
                     return readResult;
                 }
-                if ((readData.data == NULL) || (readData.size != item->size)) {
-                    return AVIF_RESULT_BMFF_PARSE_FAILED;
-                }
-                if (!avifParseImageGridBox(&data->colorGrid, readData.data, item->size)) {
```
wantehchang, line 2491: At this point, we know readData.size == item->size, so we can replace item->size with readData.size, which looks nicer with readData.data.

src/read.c
```diff
@@ -2485,10 +2485,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                 if (readResult != AVIF_RESULT_OK) {
                     return readResult;
                 }
-                if ((readData.data == NULL) || (readData.size != item->size)) {
```
wantehchang, line 2488: I analyzed the avifDecoderReadItem() code and concluded that this cannot be true if partialByteCount is 0 and avifDecoderReadItem() returns AVIF_RESULT_OK. So we can delete this check.

PR #356, opened 2020-10-08
Declare the sizeHint field of avifIO as uint64_t

wantehchang:
Since the sizeHint field of the avifIO struct may represent the size of
a file, it should be declared as uint64_t in order to support large
files (files larger than 2 GB). The current size_t type is 32 bits on
32-bit platforms.

Since the 'offset' parameter of the avifIO read() function is uint64_t,
sizeHint should be of the same type. In other words, file offset and
file size should have the same type. (I assume 'offset' is declared as
uint64_t in order to support large files.)

In avifIOCreateFileReader(), check for ftell() failure to avoid casting
-1 to an unsigned type (whether it's size_t or uint64_t).

Also declared the 'data' parameter of the avifIO write() function as a
const pointer.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -129,7 +129,11 @@ avifIO * avifIOCreateFileReader(const char * filename)
     }
 
     fseek(f, 0, SEEK_END);
-    size_t fileSize = (size_t)ftell(f);
+    long fileSize = ftell(f);
+    if (fileSize < 0) {
+        fclose(f);
+        return NULL;
+    }
     fseek(f, 0, SEEK_SET);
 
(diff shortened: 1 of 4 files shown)


src/io.c
```diff
@@ -129,15 +129,19 @@ avifIO * avifIOCreateFileReader(const char * filename)
     }
 
     fseek(f, 0, SEEK_END);
-    size_t fileSize = (size_t)ftell(f);
```
wantehchang, line 132: ftell() returns -1 on failure. It would be bad to cast -1 to size_t or uint64_t. So I added error checking.

Strictly speaking, we should also check for fseek() failures, but I was too lazy to do that...

Issue #357, opened 2020-10-08
The 'size' parameter of avifIOReadFunc and avifIOWriteFunc just needs to be size_t


wantehchang:
The 'size' parameter of avifIOReadFunc and avifIOWriteFunc is declared as uint64_t. Since the 'size' parameter represents the size of a memory buffer, it just needs to be size_t.

wantehchang:
Joe: I'm planning to write a patch to make this change this afternoon. I think this should be a straightforward change.

PR #360, opened 2020-10-12
Change 'size' param of avifIOReadFunc to size_t

wantehchang:
Change the 'size' parameter of avifIOReadFunc and avifIOWriteFunc to
size_t, because it represents the size of a memory buffer. Only the size
and offset of a file need to be a 64-bit type (in order to support large
files > 2 GB).

Fix https://github.com/AOMediaCodec/libavif/issues/357

### DIFF
--- a/tests/aviftest.c
+++ b/tests/aviftest.c
@@ -261,9 +261,9 @@ typedef struct avifIOTestReader
     size_t availableBytes;
 } avifIOTestReader;
 
-static avifResult avifIOTestReaderRead(struct avifIO * io, uint32_t readFlags, uint64_t offset, uint64_t size, avifROData * out)
+static avifResult avifIOTestReaderRead(struct avifIO * io, uint32_t readFlags, uint64_t offset, size_t size, avifROData * out)
 {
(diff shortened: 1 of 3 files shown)


tests/aviftest.c
```diff
@@ -261,9 +261,9 @@ typedef struct avifIOTestReader
     size_t availableBytes;
 } avifIOTestReader;
 
-static avifResult avifIOTestReaderRead(struct avifIO * io, uint32_t readFlags, uint64_t offset, uint64_t size, avifROData * out)
+static avifResult avifIOTestReaderRead(struct avifIO * io, uint32_t readFlags, uint64_t offset, size_t size, avifROData * out)
 {
-    // printf("avifIOTestReaderRead offset %zu size %zu\n", offset, size);
+    // printf("avifIOTestReaderRead offset %" PRIu64 " size %zu\n", offset, size);
```
joedrago, line 266: This looks backwards?

joedrago: Oh, nevermind, this is for offset.

PR #363, opened 2020-10-13
Miscellaneous cleanup changes to apps/ and tests/

wantehchang:
apps/avifdec.c:
  * Handle avifDecoderSetIOFile() failures.

apps/avifenc.c:
  * Check for not only a zero return value (empty file) but also a
    negative return value (failure) of ftell().
  * In error messages, change "content" to "metadata" when referring to
    Exif and XMP.

tests/aviftest.c:
  * Carry over my recent changes from avifIOMemoryReaderRead() to
    avifIOTestReaderRead().
  * Declare local variables holding strlen() return values as size_t
    rather than int to avoid the casts to int.
  * Add the new --io-only option to the syntax help message.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -242,11 +242,12 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
     }
 
     fseek(f, 0, SEEK_END);
-    size_t fileSize = (size_t)ftell(f);
-    if (!fileSize) {
+    long pos = ftell(f);
+    if (pos <= 0) {
         fclose(f);
         return AVIF_FALSE;
     }
+    size_t fileSize = (size_t)pos;
     fseek(f, 0, SEEK_SET);
 
     avifRWDataRealloc(raw, fileSize);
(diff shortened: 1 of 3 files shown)


apps/avifenc.c
```diff
@@ -242,11 +242,12 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
     }
 
     fseek(f, 0, SEEK_END);
-    size_t fileSize = (size_t)ftell(f);
-    if (!fileSize) {
+    long pos = ftell(f);
+    if (pos <= 0) {
```
wantehchang, line 246: Joe: I kept the test for equality to 0. I assume you want to treat an empty file as error.

tests/aviftest.c
```diff
@@ -270,18 +270,19 @@ static avifResult avifIOTestReaderRead(struct avifIO * io, uint32_t readFlags, u
...
     }
-    if ((offset + size) > reader->rodata.size) {
-        size = reader->rodata.size - offset;
+    uint64_t availableSize = reader->rodata.size - offset;
+    if (size > availableSize) {
+        size = availableSize;
     }
 
-    if (offset >= reader->availableBytes) {
+    if (offset > reader->availableBytes) {
```
wantehchang, line 282: Note that if offset == reader->availableBytes && size == 0, the new code will return AVIF_RESULT_OK.  I made this change so that this test looks like the test at line 273. I believe the function's behavior in this corner case doesn't really matter.

tests/aviftest.c
```diff
@@ -317,16 +318,16 @@ static int runIOTests(const char * dataDir)
     static const char * ioSuffix = "/io/";
 
     char ioDir[FILENAME_MAX_LENGTH + 1];
-    int dataDirLen = (int)strlen(dataDir);
-    int ioSuffixLen = (int)strlen(ioSuffix);
+    size_t dataDirLen = strlen(dataDir);
+    size_t ioSuffixLen = strlen(ioSuffix);
 
     if ((dataDirLen + ioSuffixLen) > FILENAME_MAX_LENGTH) {
```
wantehchang, line 324: Joe: If the reason you declared these variables as int was to avoid compiler warnings about signed-unsigned comparison here, I can change this test to:

`    if ((dataDirLen + ioSuffixLen) >= sizeof(ioDir)) {`

PR #365, opened 2020-10-13
avifIO read() should require readFlags == 0

wantehchang:
Since no read flags are defined yet, current implementations of
avifIOReadFunc should require the readFlags parameter be 0.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -27,7 +27,10 @@ static avifResult avifIOMemoryReaderRead(struct avifIO * io, uint32_t readFlags,
 {
     // printf("avifIOMemoryReaderRead offset %" PRIu64 " size %zu\n", offset, size);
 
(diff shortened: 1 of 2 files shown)


src/io.c
```diff
@@ -27,7 +27,10 @@ static avifResult avifIOMemoryReaderRead(struct avifIO * io, uint32_t readFlags,
 {
     // printf("avifIOMemoryReaderRead offset %" PRIu64 " size %zu\n", offset, size);
 
-    (void)readFlags;
+    if (readFlags != 0) {
```
wantehchang, line 30: Joe: An alternative is to remove the readFlags parameter from avifIOReadFunc. Would you prefer that?

joedrago:
I think this is a good approach.

Issue #364, opened 2020-10-13
The meaning of the error code "AVIF_RESULT_NO_IO" is not clear


wantehchang:
When I first saw the new error code "AVIF_RESULT_NO_IO", its meaning was not clear to me. The corresponding error message is simply "No IO" and therefore does not help to explain this error condition.

After I read the implementation, I learned that this error code means an avifDecoder (or in the futre, an avifEncoder) object does not have an avifIO object associated with it, or the associated avifIO object does not have a read() or write() function defined.

Perhaps we can rename this error code or make its error message more informative?

For example, "AVIF_RESULT_IO_NOT_SET", which suggests avifDecoderSetIO*() has not been called. Or "AVIF_RESULT_NO_IO_OBJECT" or "AVIF_RESULT_NO_IO_READER", which clarifies what "IO" refers to.

joedrago:
I'm on board with changing it to `AVIF_RESULT_IO_NOT_SET`.

PR #367, opened 2020-10-13
AVIF_RESULT_NO_IO renamed AVIF_RESULT_IO_NOT_SET

wantehchang:
AVIF_RESULT_NO_IO is renamed AVIF_RESULT_IO_NOT_SET.

Change the error checking in avifDecoderSetIOMemory() to an assertion
because avifIOCreateMemoryReader() does not fail (libavif does not
report memory allocation errors).

Change avifDecoderSetIOFile() to return AVIF_RESULT_IO_ERROR on failure.
Ideally it should return a "file not found" error, but there is no such
avifResult error code.

Fix https://github.com/AOMediaCodec/libavif/issues/364.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2186,9 +2186,7 @@ void avifDecoderSetIO(avifDecoder * decoder, avifIO * io)
 avifResult avifDecoderSetIOMemory(avifDecoder * decoder, const uint8_t * data, size_t size)
 {
     avifIO * io = avifIOCreateMemoryReader(data, size);
-    if (!io) {
-        return AVIF_RESULT_NO_IO;
-    }
+    assert(io);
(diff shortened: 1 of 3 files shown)


src/read.c
```diff
@@ -2197,7 +2195,7 @@ avifResult avifDecoderSetIOFile(avifDecoder * decoder, const char * filename)
 {
     avifIO * io = avifIOCreateFileReader(filename);
     if (!io) {
-        return AVIF_RESULT_NO_IO;
+        return AVIF_RESULT_IO_ERROR;
```
wantehchang, line 2198: I changed the error code because "IO error" seems like the closest approximation to "file not found".

PR #369, opened 2020-10-15
add SVT-AV1 encoder

jonsneyers:
This pull request adds support for the SVT encoder.


joedrago:
Reviewing now.

joedrago:
Okay, first off, thank you for the contribution. I really appreciate it.

There are a handful of issues with it that we should address, but nothing too big. First, I've already done most of the tweaks here in this commit: https://github.com/jonsneyers/libavif/commit/46088d926fa64a661872ddc1708b564e6d4711ce

Let me know if you object to any of them.

Also, I've opened a PR (mentioned above) for `SVT-AV1` where some functions are present in both libaom and SVT-AV1, which causes static link time woes. Luckily it is a minor/benign tweak in SVT-AV1, and I hope it can be resolved before I accept this PR, simply to ensure that static libavif builds continue to work, as some of the CI builds depend on it.

Finally, it doesn't seem like SVT-AV1 plays nice in OSX at all. Is that true? I wouldn't block this PR due to it simply because enabling SVT is optional anyway, but are there any plans to add OSX support? I'd be surprised if it wasn't just some minor massaging to the Linux build, but I obviously have no clue what Linux-specific things are going on in SVT-AV1, so I could be way off there.

Long story short: If you can help get that SVT PR accepted (or otherwise solved) and you don't mind my sugar-on-top commit above, I'll accept this soon.

jonsneyers:
No objection to any of those tweaks.

I am not involved in the SVT-AV1 project, we just have an interest in relatively fast (on the fly) AVIF encoding so SVT-AV1 looked attractive to us as a payload encoder and libavif looked attractive to us to handle the container stuff :)
So I also don't know about SVT-AV1 on OSX.

I hope the SVT PR will be accepted quickly, it does not seem controversial to me.

Thanks for your quick review! Always happy to contribute back a little to FOSS projects :)

Issue #379, opened 2020-10-21
Error message "Could NOT find ZLIB (missing: ZLIB_INCLUDE_DIR) (found version "1.2.11")" on a Linux computer without system zlib and libpng dev packages


wantehchang:
To reproduce this bug, uninstall the zlib and libpng dev packages on a Linux computer. For example, on a Debian or Ubuntu Linux computer, do the following:

```
sudo apt-get remove zlib1g-dev
sudo apt-get remove libpng-dev
```

Then run the following commands to checkout and build libavif:

```
git clone https://github.com/AOMediaCodec/libavif.git
cd libavif/ext
sh aom.cmd
sh libjpeg.cmd
sh zlibpng.cmd
cd ..
mkdir build
cd build
cmake -G Ninja .. -DBUILD_SHARED_LIBS=OFF \
    -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON \
    -DAVIF_LOCAL_JPEG=ON -DAVIF_LOCAL_ZLIBPNG=ON \
    -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON -DAVIF_BUILD_TESTS=ON \
    -DCMAKE_BUILD_TYPE=Release
```

The output of the cmake command contains the error message "Could NOT find ZLIB (missing: ZLIB_INCLUDE_DIR) (found version "1.2.11")":

```
...
-- Found ZLIB: zlibstatic (found version "1.2.11") 
-- Could NOT find ZLIB (missing: ZLIB_INCLUDE_DIR) (found version "1.2.11")
-- Found PNG: png_static  
-- Found JPEG: jpeg (found version "90") 
...
```

The corresponding code in libavif/CMakeLists.txt is the following three `find_package()` calls:

```
if(AVIF_BUILD_APPS)
    find_package(ZLIB REQUIRED)
    find_package(PNG REQUIRED)
    find_package(JPEG REQUIRED)
```

It turns out that `find_package(PNG REQUIRED)` calls `find_package(ZLIB)` internally, so we end up calling `find_package(ZLIB)` twice. The error message is emitted by the second `find_package(ZLIB)` call. In addition, the second `find_package(ZLIB)` call erases the value of the `ZLIB_INCLUDE_DIR` variable, so it becomes an empty string.

My experiments showed that `find_package(JPEG)` also fails similarly if called twice:

[...]

PR #380, opened 2020-10-21
Remove the find_package(ZLIB REQUIRED) call.

wantehchang:
If find_package(ZLIB) is called twice, the second call fails. Since
find_package(PNG) calls find_package(ZLIB) internally, we should not
call find_package(ZLIB), otherwise the internal find_package(ZLIB) call
will fail.

Note: This is probably a bug of find_package(ZLIB) because one would
expect find_package(ZLIB) to be idempotent.

Fix https://github.com/AOMediaCodec/libavif/issues/379.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,7 +371,8 @@ endif()
 
 option(AVIF_BUILD_APPS "Build avif apps." OFF)
 if(AVIF_BUILD_APPS)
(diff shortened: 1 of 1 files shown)


wantehchang:
I am abandoning this pull request. I got an official workaround from https://gitlab.kitware.com/cmake/cmake/-/issues/21343.

PR #383, opened 2020-10-22
Fix a crash in avifJPEGRead() on fopen() failure

wantehchang:
If the fopen() call in avifJPEGRead() fails, the local variable cinfo is
not initialized yet, so we can't goto cleanup.

Note: In fact, the local variable cinfo is not even declared at that
point. I don't know why the compiler does not warn about that.


apps/shared/avifjpeg.c
```diff
@@ -50,7 +50,7 @@ avifBool avifJPEGRead(avifImage * avif, const char * inputFilename, avifPixelFor
     FILE * f = fopen(inputFilename, "rb");
     if (!f) {
         fprintf(stderr, "Can't open JPEG file for read: %s\n", inputFilename);
-        goto cleanup;
+        return ret;
```
wantehchang, line 53: We may be able to shuffle the code so that cinfo is declared and initialized before we call fopen(), but I was too lazy to look into that.

PR #389, opened 2020-10-31
Set max image size to 32768 * 32768

cryptomilk:
I've just checked and a milky way panorama (single row) shot with my 24MP camera. The image is 15286x5476 pixels. However we have cameras with 42MP (Sony a7R3) for a single photo in the meantime. So if I create the same image with that camera it would easily exceed 16384 pixels ...

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -282,7 +282,7 @@ avifBool avifSequenceHeaderParse(avifSequenceHeader * header, const avifROData *
 
 // A maximum image size to avoid out-of-memory errors or integer overflow in
 // (32-bit) int or unsigned int arithmetic operations.
-#define AVIF_MAX_IMAGE_SIZE (16384 * 16384)
+#define AVIF_MAX_IMAGE_SIZE (32768 * 32768)
 
 #ifdef __cplusplus
 } // extern "C"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+* Set max image size to 32768 * 32768
 
 ## [0.8.2] - 2020-10-14
 ### Added


joedrago:
@wantehchang was the one that chose the current limits based on oss-fuzz ceilings and 32bit arithmetic. I think it should be up to him what to do with this PR.

joedrago:
My gut reaction to this, however, is that the right answer is probably not a constant at all. Whatever value we choose for `AVIF_MAX_IMAGE_SIZE` can simply be the default limit, and we should probably have a member variable on `avifDecoder` (which is initialized to this constant) and allow `libavif` library users to override or completely disable it (set it to zero).

include/avif/internal.h
```diff
@@ -282,7 +282,7 @@ avifBool avifSequenceHeaderParse(avifSequenceHeader * header, const avifROData *
 
 // A maximum image size to avoid out-of-memory errors or integer overflow in
 // (32-bit) int or unsigned int arithmetic operations.
-#define AVIF_MAX_IMAGE_SIZE (16384 * 16384)
```
wantehchang, line 285: The current max image size of 16384 * 16384 (= 2^28) can be justified as follows:

> INT_MAX is 2^31 - 1. We want to allow our clients to compute the buffer size of four planes (RGB and alpha) of bit depth 10 or 12 using the 'int' type. So 4 * 2 * AVIF_MAX_IMAGE_SIZE < 2^31, which implies AVIF_MAX_IMAGE_SIZE < 2^28.

The current AVIF_MAX_IMAGE_SIZE value is already lenient because it allows 4 * 2 * width * height to be 2^31, which is greater than INT_MAX by 1. Therefore, we cannot increase AVIF_MAX_IMAGE_SIZE by four times.

The use of the 'int' type for width and height is common, so this integer overflow is a real security concern. If the size_t type is used instead, the current AVIF_MAX_IMAGE_SIZE value also helps avoid size_t overflow on 32-bit platforms.

Now, for libavif clients that know how to avoid integer overflows, we can make this an option in avifDecoder to allow a larger max image size on 64-bit platforms.

Note: Another consideration when choosing a max image size is to avoid memory allocation failures in avifAlloc(). avifAlloc() calls abort() rather than return NULL when memory allocation fails. It is important to avoid passing a huge 'size' argument to avifAlloc().

joedrago:
I'd like to note that this limit implementation is area-based as well, not dimension-based. A camera emitting a 20000x13000 image would not be subject to this limitation, FWIW.

joedrago:
This conversation seems to have ended. I'm going to close this PR.

PR #391, opened 2020-11-06
Don't set image->alphaRange to yuvRange

wantehchang:
Set image->alphaRange to AVIF_RANGE_FULL instead of yuvRange.


joedrago:
I'll accept this one, but the actual value here is completely benign (`avifyuv` isn't actually testing alpha).

wantehchang:
Thanks, Joe. I was making sure that libavif does not set alphaRange to limited unless specifically told to. I didn't check whether avifyuv actualy uses this.

PR #393, opened 2020-11-06
Change avifDecoderReadItem to take avifIO as input

wantehchang:
Change avifDecoderReadItem to take avifIO (instead of avifDecoder) as
input. This makes it easier to analyze the side effects of
avifDecoderReadItem.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -625,7 +625,7 @@ static void avifDecoderDataDestroy(avifDecoderData * data)
     avifFree(data);
 }
 
-static avifResult avifDecoderReadItem(avifDecoder * decoder, avifDecoderItem * item, avifROData * outData, size_t partialByteCount)
+static avifResult avifDecoderItemRead(avifDecoderItem * item, avifIO * io, avifROData * outData, size_t partialByteCount)
 {
     if (item->mergedExtents.data && !item->partialMergedExtents) {
         // Multiple extents have already been concatenated for this item, just return it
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -987,7 +987,7 @@ static avifResult avifDecoderFindMetadata(avifDecoder * decoder, avifMeta * meta
 
         if (!decoder->ignoreExif && !memcmp(item->type, "Exif", 4)) {
             avifROData exifContents;
-            avifResult readResult = avifDecoderReadItem(decoder, item, &exifContents, 0);
+            avifResult readResult = avifDecoderReadItem(decoder->io, item, &exifContents, 0);
```
wantehchang, line 990: This change makes it obvious at a glance that avifDecoderReadItem doesn't modify any fields in 'decoder'.

joedrago:
I'm on board with this change, but the function's name probably needs a change too. If an `avifDecoder*` isn't the first argument, the function shouldn't be prefixed with `avifDecoder`.

src/read.c
```diff
@@ -625,7 +625,7 @@ static void avifDecoderDataDestroy(avifDecoderData * data)
     avifFree(data);
 }
 
-static avifResult avifDecoderReadItem(avifDecoder * decoder, avifDecoderItem * item, avifROData * outData, size_t partialByteCount)
+static avifResult avifDecoderItemRead(avifDecoderItem * item, avifIO * io, avifROData * outData, size_t partialByteCount)
```
wantehchang, line 628: I renamed the function "avifDecoderItemRead" and made avifDecoderItem * item the first argument. Please take a look. Thanks!

Alternatively, I can rename the function "avifIOReadItem" or "avifIOReadDecoderItem". Would you prefer that?

joedrago: No, this is a nice choice. I'll let it go through the builders, and then merge it.

Issue #394, opened 2020-11-06
avifDecoderNthImage may call avifDecoderFlush unnecessarily


wantehchang:
Joe: I am filing an issue for discussion and will propose a patch next.

I noticed this problem while investigating why a Chrome AVIF unit test became very slow after I changed Chrome to use avifIO. The test function is `TestByteByByteDecode()` in third_party/blink/renderer/platform/image-decoders/image_decoder_test_helpers.cc. I reproduce the relevant code snippet here:

```
void TestByteByByteDecode(DecoderCreator create_decoder,
                          SharedBuffer* shared_data,
                          size_t expected_frame_count,
                          int expected_repetition_count) {
  const Vector<char> data = shared_data->CopyAs<Vector<char>>();
  ...
  std::unique_ptr<ImageDecoder> decoder = create_decoder();

  size_t frame_count = 0;
  size_t frames_decoded = 0;

  // Pass data to decoder byte by byte.
  scoped_refptr<SharedBuffer> source_data[2] = {SharedBuffer::Create(),
                                                SharedBuffer::Create()};
  const char* source = data.data();

  for (size_t length = 1; length <= data.size() && !decoder->Failed();
       ++length) {
    source_data[0]->Append(source, 1u);
    source_data[1]->Append(source++, 1u);
    // Alternate the buffers to cover the JPEGImageDecoder::OnSetData restart
    // code.
    decoder->SetData(source_data[length & 1].get(), length == data.size());

    EXPECT_LE(frame_count, decoder->FrameCount());
    frame_count = decoder->FrameCount();

    if (!decoder->IsSizeAvailable())
      continue;

    for (size_t i = frames_decoded; i < frame_count; ++i) {
      // In ICOImageDecoder memory layout could differ from frame order.
      // E.g. memory layout could be |<frame1><frame0>| and frame_count
      // would return 1 until receiving full file.
      // When file is completely received frame_count would return 2 and
      // only then both frames could be completely decoded.
      ImageFrame* frame = decoder->DecodeFrameBufferAtIndex(i);
      if (frame && frame->GetStatus() == ImageFrame::kFrameComplete)
        ++frames_decoded;
    }
  }

  EXPECT_FALSE(decoder->Failed());
  EXPECT_EQ(expected_frame_count, decoder->FrameCount());
  EXPECT_EQ(expected_frame_count, frames_decoded);
  EXPECT_EQ(expected_repetition_count, decoder->RepetitionCount());
  ...
}
```

[...]

PR #395, opened 2020-11-06
avifDecoderNthImage: tighten decoder flush

wantehchang:
Tighten the condition for a decoder flush in avifDecoderNthImage.

Fix https://github.com/AOMediaCodec/libavif/issues/394.


joedrago:
This is a nice change.

PR #397, opened 2020-11-07
Fix miscellaneous comment nits

wantehchang:


include/avif/avif.h
```diff
@@ -406,7 +406,7 @@ avifBool avifRGBFormatHasAlpha(avifRGBFormat format);
 
 typedef enum avifChromaUpsampling
 {
-    AVIF_CHROMA_UPSAMPLING_AUTOMATIC = 0,    // Chooses best trade off of speed/quality (prefers libyuv, else uses BEST_QUALITY)
+    AVIF_CHROMA_UPSAMPLING_AUTOMATIC = 0,    // Chooses best trade off of speed/quality (prefers libyuv, else uses BEST_QUALITY) (default)
```
wantehchang, line 409: It may be better to document the default chromaUpsampling at line 422 instead.

joedrago: Yes, I'd prefer it there.

wantehchang: Done.

PR #398, opened 2020-11-07
Use existing checks for AVIF_CODEC_FLAG_CAN_ENCODE

wantehchang:
We just need to move the code that dereferences 'codec' after the null
pointer checks.


joedrago:
I don't like this change, as it happens after a few items are allocated. I'd prefer to have explicit, immediate early-out.

wantehchang:
I see. I will abandon this change and write a new one with only the cleanup.

PR #404, opened 2020-11-07
Reorder the offset checks in avifIOTestReaderRead

wantehchang:
Perform the offset checks in the same order as the comments for
avifIOReadFunc in avif.h.

### DIFF
--- a/tests/aviftest.c
+++ b/tests/aviftest.c
@@ -273,14 +273,14 @@ static avifResult avifIOTestReaderRead(struct avifIO * io, uint32_t readFlags, u
     avifIOTestReader * reader = (avifIOTestReader *)io;
 
     // Sanitize/clamp incoming request
+    if (offset > reader->rodata.size) {
+        // The offset is past the end of the buffer.
+        return AVIF_RESULT_IO_ERROR;
+    }
     if (offset == reader->rodata.size) {
         // The parser is *exactly* at EOF: return a 0-size pointer to any valid buffer
         offset = 0;
         size = 0;
-    }
-    if (offset > reader->rodata.size) {
-        // The offset is past the end of the buffer.
-        return AVIF_RESULT_IO_ERROR;
     }
     uint64_t availableSize = reader->rodata.size - offset;
     if (size > availableSize) {


tests/aviftest.c
```diff
@@ -273,15 +273,15 @@ static avifResult avifIOTestReaderRead(struct avifIO * io, uint32_t readFlags, u
     avifIOTestReader * reader = (avifIOTestReader *)io;
 
     // Sanitize/clamp incoming request
+    if (offset > reader->rodata.size) {
+        // The offset is past the end of the buffer.
+        return AVIF_RESULT_IO_ERROR;
+    }
```
wantehchang, line 279: This matches the order of the following comments in avif.h:

```
// This function should return a block of memory that *must* remain valid until another read call to
// this avifIO struct is made (reusing a read buffer is acceptable/expected).
//
// * If offset exceeds the size of the content (past EOF), return AVIF_RESULT_IO_ERROR.
// * If offset is *exactly* at EOF, provide a 0-byte buffer and return AVIF_RESULT_OK.
// * If (offset+size) does not exceed the contents' size but the *entire range* is unavailable yet
//   (due to network conditions or any other reason), return AVIF_RESULT_WAITING_ON_IO.
// * If (offset+size) does not exceed the contents' size, it must provide the *entire range* and
//   return AVIF_RESULT_OK.
// * If (offset+size) exceeds the contents' size, it must provide a truncated buffer that provides
//   all bytes from the offset to EOF, and return AVIF_RESULT_OK.
typedef avifResult (*avifIOReadFunc)(struct avifIO * io, uint32_t readFlags, uint64_t offset, size_t size, avifROData * out);
```

Note: The last case in the comments needs work because we may also return AVIF_RESULT_WAITING_ON_IO. Specifically, we need to return AVIF_RESULT_WAITING_ON_IO if the truncated range is unavailable.

joedrago: I believe that case falls under an earlier check:

    // * If (offset+size) does not exceed the contents' size but the *entire range* is unavailable yet
    //   (due to network conditions or any other reason), return AVIF_RESULT_WAITING_ON_IO.

The emphasis on "entire range" here is key. If the entire range isn't available (due to temporary reasons), we're supposed to return WAITING_ON_IO. The truncated range at the end is for when we have "all" of the data and it is less than what is being requested.

joedrago: Oh, right after hit hit Comment I realized what you meant. If a file has 5000 bytes, 4900 have been downloaded, and libavif asks for 300 bytes at offset 4800, it is supposed to return WAITING_ON_IO, as it cannot return the truncated buffer *yet*. I agree that better wording can be found here.

wantehchang: Thank you for taking a look. Yes, that is the problem I reported. I should have given a concrete example like you did.

One way to fix this is to refine the last step as follows:

```
// * If (offset+size) exceeds the contents' size, it must provide a truncated buffer that provides
//   all bytes from the offset to EOF. If the the *truncated range* is unavailable yet (due to
//   network conditions or any other reason), return AVIF_RESULT_WAITING_ON_IO. Otherwise provide
//   the *truncated range* and return AVIF_RESULT_OK.
```

Another way to fix this is to match the code and insert a step after the second step and remove the last step, so that the comments look like this:

```
// * If offset exceeds the size of the content (past EOF), return AVIF_RESULT_IO_ERROR.
// * If offset is *exactly* at EOF, provide a 0-byte buffer and return AVIF_RESULT_OK.
// * If (offset+size) exceeds the contents' size, it must truncate the range to provide all
//   bytes from the offset to EOF.
// * If the range is unavailable yet (due to network conditions or any other reason),
//   return AVIF_RESULT_WAITING_ON_IO.
// * Otherwise, provide the range and return AVIF_RESULT_OK.
```

joedrago: I think the second approach is the right choice. It is easier to follow, I think.

PR #410, opened 2020-11-13
CI: Update to Ubuntu 20.04 and Nasm 2.15.05, 

EwoutH:
- Updates GitHub Actions, Travis CI and AppVeyor to Nasm 2.15.05
- Updates GitHub Actions and Travis CI to Ubuntu 20.04
- Simplify dependency setup in GitHub Actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,7 @@ jobs:
       CC: gcc-10
       CXX: g++-10
     steps:
+    - uses: actions/setup-python@v2
     - name: Install dependencies
       run: |
         DEBIAN_FRONTEND=noninteractive sudo apt-get update || true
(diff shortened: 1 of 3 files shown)


EwoutH:
For some reason GitHub Actions didn't use the newest version of ilammy/setup-nasm (it was just released earlier today), but I verified in [this run](https://github.com/EwoutH/libavif/runs/1396200405) the newest version with Nasm 2.15.05 as default also works on GitHub Actions.

joedrago:
I'm unfamiliar with `ilammy`. How guaranteed is the `@v1` to never change? I'm hesitant to depend on yet another repository for building, if I can avoid it.

EwoutH:
It's an GitHub Action, so just a small script that automates installing Nasm. An advantage is that future bugs in the script get fixed upstream, or if the Nasm repo changes or something. The developer is also quite responsive. See https://github.com/ilammy/setup-nasm

It's probably good to fixate the Nasm version however. I added that just now.

PR #412, opened 2020-11-17
Fix bugs in MaxExtent and ExtentMerge

wantehchang:
In avifDecoderItemMaxExtent, check for size_t overflow.

In avifExtentMerge, calculate the 'dst' extent size correctly and check
for size_t overflow.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -673,7 +673,11 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifExt
     }
 
     outExtent->offset = minOffset;
-    outExtent->size = maxOffset - minOffset;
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -673,7 +673,11 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifExt
     }
 
     outExtent->offset = minOffset;
-    outExtent->size = maxOffset - minOffset;
+    const uint64_t extengLength = maxOffset - minOffset;
```
joedrago, line 676: Typo: `extengLength`

wantehchang: Done. Sorry about the typo!

PR #413, opened 2020-11-17
Miscellaneous cleanup in read.c

wantehchang:
In avifDecoderReset, use "else if" to check prop->u.colr.hasNCLX after
checking prop->u.colr.hasICC.

In avifDecoderNextImage, avifDecoderIsKeyframe, and
avifDecoderNearestKeyframe, resurrect good changes that were reverted
along with the avifDecoderNthImageReady function.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2875,11 +2875,16 @@ avifResult avifDecoderReset(avifDecoder * decoder)
 
 avifResult avifDecoderNextImage(avifDecoder * decoder)
 {
+    if (!decoder->data) {
+        // Nothing has been parsed yet
+        return AVIF_RESULT_NO_CONTENT;
+    }
+
     if (!decoder->io || !decoder->io->read) {
         return AVIF_RESULT_IO_NOT_SET;
     }
 
-    uint32_t nextImageIndex = (uint32_t)(decoder->imageIndex + 1);
+    const uint32_t nextImageIndex = (uint32_t)(decoder->imageIndex + 1);
 
     // Acquire all sample data for the current image first, allowing for any read call to bail out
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -2779,8 +2779,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                 }
                 colrICCSeen = AVIF_TRUE;
                 avifImageSetProfileICC(decoder->image, prop->u.colr.icc, prop->u.colr.iccSize);
-            }
-            if (prop->u.colr.hasNCLX) {
+            } else if (prop->u.colr.hasNCLX) {
```
joedrago, line 2782: All of the other changes are good, but I prefer this to not be else-if, honestly. I recognize it is logically sound right now, but I wonder if in the future it'll be better to have a *single* `colr` property in the array that can have one ICC profile *and* CICP, in which this change would break it. I'd rather this code stay the way it was.

wantehchang: Done. The reason I suggested "else if" was to match the avifParseColourInformationBox() function and the syntax of ColourInformationBox in ISO BMFF:

```
class ColourInformationBox extends Box(‘colr’){
    unsigned int(32) colour_type;
    if (colour_type == ‘nclx’) /* on-screen colours */
    {
        unsigned int(16) colour_primaries;
        unsigned int(16) transfer_characteristics;
        unsigned int(16) matrix_coefficients;
        unsigned int(1) full_range_flag;
        unsigned int(7) reserved = 0;
    }
    else if (colour_type == ‘rICC’)
    {
        ICC_profile; // restricted ICC profile
    }
    else if (colour_type == ‘prof’)
    {
        ICC_profile; // unrestricted ICC profile
    }
}
```

PR #414, opened 2020-11-17
Clean up comments related to idat fox for an item

wantehchang:
Clean up the comments for the search for the idat box associated with an
item.

Note: I find "this meta box" a little confusing in those two functions
because the subject of the two functions is an item, not a meta box. So
I changed "this meta box" to "the meta box".


wantehchang:
Joe: Please feel free to decline this PR if you think the current comments are just fine. Or you can rephrase. Thanks!

PR #416, opened 2020-11-17
Replace firstExtent in avifDecoderItemMaxExtent

wantehchang:
The firstExtent boolean variable in avifDecoderItemMaxExtent can be
replaced by initializing minOffset to the maximum value (UINT64_MAX) and
maxOffset to the minimum value (0). This method works because the
extentIter for loop has at least one iteration
(item->extents.count != 0).


wantehchang:
Joe: You can decline this pull request if you prefer the original code.

Issue #418, opened 2020-11-18
Add AVIF_MATRIX_COEFFICIENTS_YCGCO=8 (YCgCo) support to src/reformat.c


wantehchang:
I am going to add AVIF_MATRIX_COEFFICIENTS_YCGCO=8 (YCgCo) support to src/reformat.c. Here is what Joe Drago outlined:

> Note that MC=8 (YCgCo) is not YCgCo-R.  You'll just have to implement AVIF_REFORMAT_MODE_YCGCO in avifPrepareReformatState() and avifImageYUVAnyToRGBAnySlow() and avifImageRGBToYUV() and make sure avifImageYUVToRGB() always routes to avifImageYUVAnyToRGBAnySlow() for it.

PR #419, opened 2020-11-18
Add AVIF_MATRIX_COEFFICIENTS_YCGCO (YCgCo) support

wantehchang:
Add AVIF_MATRIX_COEFFICIENTS_YCGCO=8 (YCgCo) support to src/reformat.c.

The YCgCo color space has the following two differences:

1. For the limited-full range conversion, the U,V planes use the same
equations as the Y plane. In this aspect, it is the same as Identity.

2. For the RGB-to-YUV conversion, it is similar to normal YUV
conversion but uses different equations (its equations do not use the Kb
and Kr coefficients).

In H.273 the equations for YCgCo also differ from the equations for
normal YUV conversion in when the Round() and Clip1Y() functions are
called. (In this aspect YCgCo is the same as Identity.) But libavif
seems to ignore these differences.

Fix https://github.com/AOMediaCodec/libavif/issues/418

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -24,7 +24,11 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * 
     }
 
     // These matrix coefficients values are currently unsupported. Revise this list as more support is added.
-    if ((image->matrixCoefficients == 3 /* CICP reserved */) || (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO) ||
+    //
+    // YCgCo performs limited-full range adjustment on R,G,B but the current implementation performs range adjustment
+    // on Y,U,V. So YCgCo with limited range is unsupported.
+    if ((image->matrixCoefficients == 3 /* CICP reserved */) ||
+        ((image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO) && (image->yuvRange == AVIF_RANGE_LIMITED)) ||
         (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_BT2020_CL) ||
         (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_SMPTE2085) ||
         (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_CHROMA_DERIVED_CL) ||
@@ -41,6 +45,8 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * 
 
     if (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
         state->mode = AVIF_REFORMAT_MODE_IDENTITY;
+    } else if (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO) {
+        state->mode = AVIF_REFORMAT_MODE_YCGCO;
     }
 
     if (state->mode != AVIF_REFORMAT_MODE_YUV_COEFFICIENTS) {
@@ -214,6 +220,11 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                         yuvBlock[bI][bJ].y = rgbPixel[1]; // G
                         yuvBlock[bI][bJ].u = rgbPixel[2]; // B
                         yuvBlock[bI][bJ].v = rgbPixel[0]; // R
+                    } else if (state.mode == AVIF_REFORMAT_MODE_YCGCO) {
+                        // Formulas 44,45,46 from https://www.itu.int/rec/T-REC-H.273-201612-I/en
+                        yuvBlock[bI][bJ].y = 0.5f * rgbPixel[1] + 0.25f * (rgbPixel[0] + rgbPixel[2]);
+                        yuvBlock[bI][bJ].u = 0.5f * rgbPixel[1] - 0.25f * (rgbPixel[0] + rgbPixel[2]);
+                        yuvBlock[bI][bJ].v = 0.5f * (rgbPixel[0] - rgbPixel[2]);
                     } else {
                         float Y = (kr * rgbPixel[0]) + (kg * rgbPixel[1]) + (kb * rgbPixel[2]);
                         yuvBlock[bI][bJ].y = Y;
@@ -523,6 +534,12 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
                     G = Y;
                     B = Cb;
                     R = Cr;
+                } else if (state->mode == AVIF_REFORMAT_MODE_YCGCO) {
+                    // Identity (GBR): Formulas 47,48,49,50 from https://www.itu.int/rec/T-REC-H.273-201612-I/en
+                    const float t = Y - Cb;
+                    G = Y + Cb;
+                    B = t - Cr;
+                    R = t + Cr;
                 } else {
                     // Normal YUV
                     R = Y + (2 * (1 - kr)) * Cr;
@@ -1028,7 +1045,7 @@ avifResult avifImageYUVToRGB(const avifImage * image, avifRGBImage * rgb)
             }
 
             // TODO: Add more fast paths for identity
-        } else {
+        } else if (state.mode == AVIF_REFORMAT_MODE_YUV_COEFFICIENTS) {
             if (image->depth > 8) {
                 // yuv:u16
 
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -456,7 +456,8 @@ int avifLimitedToFullUV(int depth, int v);
 typedef enum avifReformatMode
 {
     AVIF_REFORMAT_MODE_YUV_COEFFICIENTS = 0, // Normal YUV conversion using coefficients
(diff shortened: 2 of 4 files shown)


joedrago:
Wow, this is a surprisingly terse PR! I'll review it now.

joedrago:
> But libavif seems to ignore these differences.

Can you explain this? If we're missing something, I'm all for fixing it.

src/reformat.c
```diff
@@ -116,6 +117,8 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
         state->unormFloatTableY[cp] = (float)unormY / state->yuvMaxChannelF;
         if (state->mode == AVIF_REFORMAT_MODE_IDENTITY) {
             state->unormFloatTableUV[cp] = (float)unormY / state->yuvMaxChannelF;
+        } else if (state->mode == AVIF_REFORMAT_MODE_YCGCO) {
+            state->unormFloatTableUV[cp] = ((float)unormY - state->uvBias) / state->yuvMaxChannelF;
```
joedrago, line 121: This line feels wrong. I feel like it should either be the math from the above block or the below block, and somehow it is something in between. Is this correct?

wantehchang:
> > But libavif seems to ignore these differences.
> 
> Can you explain this? If we're missing something, I'm all for fixing it.

It is possible that the way libavif does it is equivalent to the equations in H.273.

In H.273, the equations for 0 and 8 (Identity and YCgCo) and the equations for others differ in when and the order in which Clip1Y() and Round() are called.

Suppose VideoFullRangeFlag == 1 and BitDepthY == BitDepthC.

0 and 8 use equations 26-28 and 41-46. Note that we first call Clip1Y() in equations 26-28 and then call Round() in equations 41-46. Clip1Y() is called in the middle of the conversion, and Round() is called at the end of the conversion.

Other value of MC use equations 29-31 and 38-40. Note that we first call Round() and then call Clip1Y() (both happen in equations 29-31). Also, equations 38-40 are executed before equations 29-31, so Round() and Clip1Y() are both called at the end of the conversion.

src/reformat.c
wantehchang: You're right. For YCgCo, unormFloatTableY and unormFloatTableUV should be applied at the end of the YUV-to-RGB conversion. Let me look into this.

joedrago: I think you should just avoid `uvBias` the way IDENTITY does, in both `avifPrepareReformatState()` and `avifReformatStateUVToUNorm()`.

joedrago: Something like this, perhaps?

    index 33c0f7b..de550c2 100644
    --- a/src/reformat.c
    +++ b/src/reformat.c
    @@ -115,10 +115,8 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
                 unormUV = avifLimitedToFullUV(image->depth, unormUV);
             }
             state->unormFloatTableY[cp] = (float)unormY / state->yuvMaxChannelF;
    -        if (state->mode == AVIF_REFORMAT_MODE_IDENTITY) {
    +        if ((state->mode == AVIF_REFORMAT_MODE_IDENTITY) || (state->mode == AVIF_REFORMAT_MODE_YCGCO)) {
                 state->unormFloatTableUV[cp] = (float)unormY / state->yuvMaxChannelF;
    -        } else if (state->mode == AVIF_REFORMAT_MODE_YCGCO) {
    -            state->unormFloatTableUV[cp] = ((float)unormY - state->uvBias) / state->yuvMaxChannelF;
             } else {
                 state->unormFloatTableUV[cp] = ((float)unormUV - state->uvBias) / state->yuvMaxChannelF;
             }
    @@ -139,7 +137,7 @@ static int avifReformatStateYToUNorm(avifReformatState * state, float v)
     static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
     {
         int unorm = (int)avifRoundf(v * state->yuvMaxChannelF);
    -    if (state->mode != AVIF_REFORMAT_MODE_IDENTITY) { // Don't adjust U or V when data is raw RGB
    +    if (state->mode == AVIF_REFORMAT_MODE_YUV_COEFFICIENTS) { // Don't adjust U or V when data is raw RGB or YCgCo
             unorm += state->uvBias;
         }
         unorm = AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);

joedrago: I could probably even flip the conditional in the `avifPrepareReformatState()` to match the one in `avifReformatStateUVToUNorm()` and simplify it.

joedrago: Ignore that diff; I've done something wrong.

joedrago: Okay, upon further reading of the CICP formulas, I think your usage of `uvBias` is _correct_. The only open question is whether or not to use the limited Y or limited UV range. Your original code might have been correct this whole time. Sorry for the swirl, if so.

joedrago: Yeah, the line above equation (20) makes it quite clear. Your current implementation is correct. Sorry about that.

wantehchang:
Joe: Thank you very much for taking a look. I will need to study H.273 carefully tomorrow. Please don't merge this pull request.

I think this pull request is probably correct for full range, ignoring the clip and round issue. However, I am not sure if this pull request is correct for limited range. I have only tested the code with "avifyuv -m 1", which uses full range only.

For 0 and 8, the limited-full range adjustment and matrix multiplication are applied in the opposite order. Since 0 uses the identity matrix, using the normal order doesn't matter. However, the order matters to YCgCo.

joedrago:
This PR seems to work and behave reasonably with Chrome's YCgCo rendering, based on this test:

    avifenc --min 0 --max 0 --cicp 1/13/1 src.png 1.avif
    avifenc --min 0 --max 0 --cicp 1/13/8 src.png 8.avif
    avifdec 1.avif 1.png
    avifdec 8.avif 8.png

I can verify that `src.png` is at most 1 codepoint off in a channel from both `1.png` and `8.png` (good/expected), and `8.avif` appears to render the same as `src.png` and `8.png` in Chrome. I'll send you a zip privately so you can see it.

wantehchang:
Joe: I studied H.273 Section 8.3 carefully last night. After considering the options, I propose that we first support YCgCo with full range only. This allows us to reuse the current implementation for normal YUV conversion with minimal changes and get the most bang for the buck. I have pushed a new update to this PR. Please review.

Ignoring the float-unorm conversion, the equations for Identity and YCgCo in H.273 can be summarized as follows:

> R,G,B -> range adjustment -> matrix transformation -> Y,Cb,Cr

The equations for other values of MC (i.e., normal YUV conversion) can be summarized as follows:

> R,G,B -> matrix transformation -> range adjustment -> Y,Cb,Cr

Therefore, range adjustment differs in not only the equations but also which samples are adjusted. The current code in src/reformat.c implements the second conversion process. Since the matrix transofmration for Identity is a no-op, the second conversion process can be customized for Identity by using different equations for range adjustment. However, it does not seem correct to use the second conversion process for YCgCo (even if we use different equations for range adjustment) because the order is incorrect.

An option I considered is to write separate functions for Identity and YCgCo; these functions would only need integer arithmetic. I found two difficulties.

1. The inputs to the Round() function in Equations 45 and 46 may be negative. Since Round() is defined to round midpoints away from 0 (e.g., 0.5 => 1 and -0.5 => -1), it is more complicated to implement that using integer arithmetic.

2. Equations 45 and 46 have a bug: for bit depth n, the maximum value of Cb and Cr is 2^n, which overflows an n-bit unsigned integer. For example, if G=255 and R=B=0, then Equation 25 says Cb = Round(0.5 * 255) + 128 = Round(127.5) + 128 = 256.

There are two possible fixes:

1. Add a Clip1C() call to Equations 45 and 46. Note: This clip operation is present in the second conversion process I described above. So this PR is essentially using this fix.

2. Replace the Round() function in Equations 45 and 46 with a function that rounds midpoints up for both positive and negative values (e.g., 0.5 => 1 and -0.5 => 0). This solution is easier to implement using integer arithmetic.

Because of this bug in H.273, I did not pursue this option.

joedrago:
I can accept this PR as it is now, unless you want to investigate other YCgCo implementations to see where the disagreement between H.273 and common implementations lies. This is a strict enough PR (disallowing YCgCo limited) that a supplemental PR could enable that once it has been figured out.

wantehchang:
Joe: Please accept the current version of this PR. It gets the most bang for the buck.

Summary of my findings:

1. The specification of YCgCo in H.273 has a bug: If BitDepthY=BitDepthC=n, the range of Cb and Cr based on Equations 45 and 46 is 0..2^n (inclusive). The maximum value 2^n cannot be represented as an unsigned n-bit integer.

This PR addresses this issue by the AVIF_CLAMP() call in avifReformatStateUVToUNorm().

2. The different limited-full range adjustment required by YCgCo is hard to fit into the existing implementation in src/reformat.c but can be dealt with by adding separate functions for YCgCo. I did not do that in this PR because I don't need to use YCgCo with limited range.

For the same reason, I probably won't spend the time to look into whether common implementations get the limited-full range adjustment right for YCgCo. (I am familiar with Chrome's implementation. Chrome doesn't do it correctly.) If I have spare time, I will just implement it correctly in libavif.

PR #420, opened 2020-11-19
Print MC value when warning resetting to defaults

wantehchang:
Print the default value when we warn we're restting matrix coefficients
to defaults.


joedrago:
Won't this always print 6? I'm not against the change, either way.

wantehchang:
Are you suggesting we print AVIF_MATRIX_COEFFICIENTS_BT601 instead?

The most important thing is to print the default value that we reset MC to. Whether we print image->matrixCoefficients, AVIF_MATRIX_COEFFICIENTS_BT601, or just hardcode "(6)" in the printf format string doesn't really matter.

PR #421, opened 2020-11-19
Allow override of HAVE_POISON_SYSTEM_DIRECTORIES_WARNING

bnason-nf:
This change allows overriding of the check for HAVE_POISON_SYSTEM_DIRECTORIES_WARNING in case that check is undesired for some reason.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,9 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang")
     )
     # The detection of cross compilation by -Wpoison-system-directories has false positives on macOS because
     # --sysroot is implicitly added. Turn the warning off.
-    check_c_compiler_flag(-Wpoison-system-directories HAVE_POISON_SYSTEM_DIRECTORIES_WARNING)
+    if(NOT DEFINED HAVE_POISON_SYSTEM_DIRECTORIES_WARNING)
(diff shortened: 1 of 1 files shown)


wantehchang:
Ben: I have a question about this change. If I understand it correctly, to enable the -Wpoison-system-directories warning, I need to pass -DHAVE_POISON_SYSTEM_DIRECTORIES_WARNING=OFF to cmake? That seems confusing.

joedrago:
I can explain this. There exist some embedded builds with CMake that are somewhat hostile to configuration flag detection, and this allows for a benign way to pre-set this value in a parent `CMakeLists.txt`. It doesn't impact anything that doesn't directly set this, so I accepted it.

bnason-nf:
You're both correct. @wantehchang, maybe it's a little less confusing if I explain that it's also possible to set HAVE_POISON_SYSTEM_DIRECTORIES_WARNING to ON? This change basically allows hardcoding that HAVE instead of detecting it if you want to avoid the check_c_compiler_flag() for the reason @joedrago said.

wantehchang:
Thank you both for the replies. Does this mean we should not add more check_c_compiler_flag() calls to CMakeLists.txt?

bnason-nf:
Not at all, though it would be appreciated (by me at least) if any new checks like that were wrapped with a similar if(NOT DEFINED ...).

PR #422, opened 2020-11-20
Add the avifContainerDump() function

wantehchang:
Use the avifContainerDump() function in the info() function in avifdec.c
to dump the container info instead of the first frame's info.

I noticed this issue while trying to debug monochrome images encoded
with libaom before commit 2172ed05169f08b20245221a7e95619a65884fd7.
The container yuvFormat was YUV400 but the first frame's yuvFormat was
YUV420. avifdec -i printed YUV420, so I could not see the container
yuvFormat was wrong. Obviously it would be best to dump both the
container info and every frame's info, but that could be confusing to
average users, or at least too verbose.

### DIFF
--- a/apps/shared/avifutil.h
+++ b/apps/shared/avifutil.h
@@ -21,6 +21,7 @@
 #endif
 
 void avifImageDump(avifImage * avif);
+void avifContainerDump(avifDecoder * decoder);
 void avifPrintVersions(void);
 
 typedef enum avifAppFileFormat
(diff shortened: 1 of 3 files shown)


apps/shared/avifutil.h
```diff
@@ -21,6 +21,7 @@
 #endif
 
 void avifImageDump(avifImage * avif);
+void avifContainerDump(avifDecoder * decoder);
```
wantehchang, line 24: Joe: Please let me know if you would prefer a different name for the new function.

PR #423, opened 2020-11-20
Document the avifInputHasRemainingData() function

wantehchang:
I did not know what the function was for until I saw how it was used.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -188,6 +188,8 @@ static avifInputFile * avifInputGetNextFile(avifInput * input)
     }
     return &input->files[input->fileIndex];
 }
+
+// Returns whether the next avifInputGetNextFile() call will succeed.
 static avifBool avifInputHasRemainingData(avifInput * input)
 {
     if (input->useStdin) {


apps/avifenc.c
```diff
@@ -188,6 +188,8 @@ static avifInputFile * avifInputGetNextFile(avifInput * input)
     }
     return &input->files[input->fileIndex];
 }
+
+// Returns whether the next avifInputGetNextFile() call will succeed.
```
wantehchang, line 192: Renaming this function something like "avifInputHasMoreFiles" would also be helpful by making the name more self-explanatory.

joedrago: "HasMoreFiles" is slightly incorrect, as this function will return true while you are still reading out of the middle of a multiframe y4m.

joedrago: I see why you chose the name you chose. Perhaps if anything, `avifInputGetNextFile()` needs a new name.

wantehchang: Thanks for taking a look! Note that this pull request does not rename the function. It only adds a comment.

PR #425, opened 2020-11-23
mark scripts as executable

skal65535:
some +x were missing...


yaowuxu (approved):

joedrago:
I'm not against the general idea of this commit, but I'm not sure how this helps you at all? These scripts don't have a shebang at the top, so you are always going to execute them with the name of your shell in front instead of directly, which means they don't need the executable bit set.

skal65535:
On Mon, Nov 23, 2020 at 6:46 PM Joe Drago <notifications@github.com> wrote:

> I'm not against the general idea of this commit, but I'm not sure how this
> helps you at all? These scripts don't have a shebang at the top, so you are
> always going to execute them with the name of your shell in front instead
> of directly, which means they don't need the executable bit set.
>

well i needed them on my system. And i found it strange that only aom.cmd
and libgav1.cmd had the +x bit and not the others...

./rav1e.cmd
-bash: ./rav1e.cmd: Permission denied

skal/

PR #426, opened 2020-11-25
Upgrade to libaom v2.0.1

wantehchang:


joedrago:
Neat! Did you check to see if this actually encodes in monochrome (400) properly now?

wantehchang:
Yes, I verified this encodes in monochrome (YUV 4:0:0) properly now.

Issue #372, opened 2020-10-15
Can avifJPEGRead() set cinfo.out_color_space to JCS_YCbCr to avoid the conversion to/from RGB?


wantehchang:
Joe,

When I run avifenc in the debugger and dump the contents of the 'cinfo' local variable in avifJPEGRead(), I see the following values:

    jpeg_color_space = JCS_YCbCr, out_color_space = JCS_RGB

In jpeglib.h, these two fields of struct jpeg_decompress_struct are documented as follows:

```
  J_COLOR_SPACE jpeg_color_space; /* colorspace of JPEG image */

  /* Decompression processing parameters --- these fields must be set before
   * calling jpeg_start_decompress().  Note that jpeg_read_header() initializes
   * them to default values.
   */

  J_COLOR_SPACE out_color_space; /* colorspace for output */
```

That comment seems to suggest we could set cinfo.out_color_space to JCS_YCbCr between our jpeg_read_header() and jpeg_start_decompress() calls to avoid the conversion to/from RGB. Do you know if this will work?

I found this code in the third_party/blink/renderer/platform/image-decoders/jpeg/jpeg_image_decoder.cc file in the Chrome source tree:

https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/image-decoders/jpeg/jpeg_image_decoder.cc;l=763?q=%22out_color_space%20%3D%20JCS_YCbCr%22&ss=chromium

```
      case JPEG_START_DECOMPRESS:
        if (decoding_mode == JPEGImageDecoder::DecodingMode::kDecodeToYuv) {
          DCHECK(decoder_->CanDecodeToYUV());
          DCHECK(decoder_->HasImagePlanes());
          info_.out_color_space = JCS_YCbCr;
          info_.raw_data_out = TRUE;
          uv_size_ = ComputeYUVSize(&info_, 1);
          // U size and V size have to be the same if we got here
          DCHECK_EQ(uv_size_, ComputeYUVSize(&info_, 2));
        }
```

Thank you.

joedrago:
I have never tried to receive YUV from libjpeg before, but assuming it can emit YUV420 (or YUV444?), it should play nicely with the current image loader architecture in `avifenc`. The conversion from RGB to YUV already happens inside of each image decoder. 

[...]

PR #430, opened 2020-12-02
Allow read from JPEG without colorspace conversion

tongyuantongyu:
Implementation for #372. Require a explicit flag(-p, --preserve) to use this path. This patch also adds support to read and write 12bit JPEG files (example [here](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/master/testimages/testorig12.jpg)).

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -12,6 +12,9 @@
 
 #include "iccjpeg.h"
 
+#define AVIF_MIN(a, b) (((a) < (b)) ? (a) : (b))
+#define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
+
 struct my_error_mgr
 {
     struct jpeg_error_mgr pub;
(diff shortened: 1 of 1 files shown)


EwoutH:
This looks quite amazing!

@joedrago or @wantehchang, does one of you have a chance to review?

joedrago:
This seems to be 3 semi-unrelated things all mixed together in a big PR. The general idea of preserving YUV (avoiding YUV -> RGB -> YUV) from JPEG seems quite separable from the larger changes (supporting arbitrary compiled-in-one-at-a-time JPEG depths, additional enums, all of the preserve checking in avifdec, etc.)

Which part looks amazing to you? It might be easier to separate it out and make a PR of just that part.

EwoutH:
I didn't dive too deep in the implementation, just the feature of avoiding colorspace conversion would be very useful, especially for screen content.

I agree with splitting in multiple PRs.

ValZapod:
> The time needed on figuring this out

I just use mediainfo. 🙄 Not all JPEG sampling modes are supported even by ffmpeg. https://trac.ffmpeg.org/ticket/8930 and others. Also JPEG supports YCCK and Kodak's PhotoYCC (almost like xvYCC), yet nobody implemented the latter, including reference libraries, and mozjpeg/turbojpeg and YCCK is buggy in ffmpeg...

tongyuantongyu:
YUV->RGB->YUV should be lossless, if all YUV value are valid in RGB range, so it's mostly just about performance - but your effort on checking this info and changing flags also count.

tongyuantongyu:
I updated this PR, and now it only contains 'preserving YUV when reading JPEG if possible' part.

I suggest we add an 'automatic' value for `requestedFormat` (or let `AVIF_PIXEL_FORMAT_NONE` be that), like 0 for `requestedDepth`, which allows selecting best format according to input file.

tongyuantongyu:
@joedrago, now this PR is only about preserving YUV when reading from JPEG, could you have a review if you think this is a good feature, or otherwise inform me so that I'll close this PR?

joedrago:
I haven't considered this a high priority feature as I don't believe there is any demand for it. Is there any reason you don't want to just let the PR stay open? It is okay with me if you want to close it and we can make a fresh PR in the future if need be.

ValZapod:
> this a high priority feature

It most certainly is. 

For example, full range R'G'B' values 238, 77, 45 are encoded into limited YCbCr with BT.601 matrix: you will get limited range 120, 90, 201 after school rounding, but if you will round it you will get 238, 77, 44 back in R'G'B'. And 238, 77, 44 value will go to the same value. 

Now in the case of JPEG Full range YCBCR with BT.601 matrix is used, it is not that bad but also not perfect.

joedrago:
> Franky speacking, without it libavif is too dangerous to use.

This seems a bit hyperbolic. 

The feature @tongyuantongyu is providing in this PR is for avifenc, not libavif. libavif is designed to accept unaltered YUV already, so I'm not sure what you're intending to say here.

I'm quite familiar with the precision loss when converting between RGB and YUV, but in this specific case (reading in a JPEG for packing), you're going to take the image given from JPEG and compress it down even further, creating more loss. I suppose you could create a QP0 AVIF which could keep the YUV in the JPEG perfectly intact, but not only would the resultant AVIF be larger than the JPEG (making compressing it into AVIF moot), but eventually someone is going to convert those precise YUV values back into RGB using the same mechanism before displaying it in any typical viewer (such as a browser).

So _in this specific case_ (avifenc, not libavif), this actually buys you very little. At this point I've spent more time replying to this PR than it probably would have taken to review it, so I'll review it now.

joedrago:
@tongyuantongyu I think the way that `avifJPEGReadCopy()` ensures that the requested destination format matches exactly is very likely to cause this code to not be used often, without the end user being aware at all. I think `avifenc.c` and `avifjpeg.c` might need to signal a bit more information to each other on the enduser's expectations, or perhaps we just need to add some careful printfs in `avifjpeg.c` indicating to the user that they are or are not receiving this benefit. As an example, I took a JPEG that is `JCS_YCbCr` internally, and simply ran `avifenc abstract.jpg out.avif` on it, and it silently avoided this path because I didn't also specify `--yuv 420`. We need to think about how the enduser can at least be made aware of whether this path occurred, or perhaps ensure it, if they require it for some reason. This will only get worse as the commandline gets more "interesting". For example, someone simply doing `avifenc --lossless input.jpg output.avif` will _never_ get this path (ironically).

I have a few comments I can add inline with the diff and we can discuss each there, but generally I think this current design is almost never going to end up invoking `avifJPEGCopyPixels()` in a way that the enduser can rely on or even know about.

apps/shared/avifjpeg.c
```diff
@@ -25,6 +28,176 @@ static void my_error_exit(j_common_ptr cinfo)
...
+{
+    if (avif->depth != 8) {
+        return AVIF_FALSE;
+    }
+
+    if (cinfo->jpeg_color_space == JCS_YCbCr) {
+        // Import from YUV: must using compatible matrixCoefficients.
+        if ((avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_BT601) ||
+            (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_CHROMA_DERIVED_NCL &&
+             avif->colorPrimaries == AVIF_COLOR_PRIMARIES_BT470M)) {
```
joedrago, line 113: I think checking against `MC=12` here might not be worth it, and I don't believe `AVIF_COLOR_PRIMARIES_BT470M` is correct either way? The derived matrix coefficients for BT601 either come from `AVIF_COLOR_PRIMARIES_BT470BG` or `AVIF_COLOR_PRIMARIES_BT601` (I'd have to do the math to remember precisely), but I'm almost positive it isn't `AVIF_COLOR_PRIMARIES_BT470M`.

apps/shared/avifjpeg.c
```diff
@@ -79,6 +247,25 @@ avifBool avifJPEGRead(const char * inputFilename, avifImage * avif, avifPixelFor
...
     }
 
+    avif->yuvFormat = requestedFormat;
+    avif->depth = requestedDepth ? requestedDepth : 8;
+    // JPEG doesn't have alpha. Prevent confusion.
+    avif->alphaPremultiplied = AVIF_FALSE;
+
+    if (avifJPEGReadCopy(avif, &cinfo)) {
+        ret = AVIF_TRUE;
+        goto cleanup;
```
joedrago, line 257: For another PR of yours, I undid the pattern of "call a function that is only called once, otherwise `goto`". I think there are good usages for `goto` and single-use functions (certainly both exist in libavif), but single-use `goto` feels sloppy to me. Is there another way this can be done?

This won't block the PR (as I'll change it myself before accepting it), but as you are a regular contributor to libavif (thank you!), I wanted to let you know what I'm thinking as I review these.

apps/shared/avifjpeg.c
joedrago: Alright, I did the math, and you're right:

    calcYUVCoefficients = (primaries) ->
      rX = primaries[0]
      rY = primaries[1]
      gX = primaries[2]
      gY = primaries[3]
      bX = primaries[4]
      bY = primaries[5]
      wX = primaries[6]
      wY = primaries[7]
      rZ = 1.0 - (rX + rY) # (Eq. 34)
      gZ = 1.0 - (gX + gY) # (Eq. 35)
      bZ = 1.0 - (bX + bY) # (Eq. 36)
      wZ = 1.0 - (wX + wY) # (Eq. 37)
      kr = (rY * (wX * (gY * bZ - bY * gZ) + wY * (bX * gZ - gX * bZ) + wZ * (gX * bY - bX * gY))) /
           (wY * (rX * (gY * bZ - bY * gZ) + gX * (bY * rZ - rY * bZ) + bX * (rY * gZ - gY * rZ)))

      # (Eq. 32)
      kb = (bY * (wX * (rY * gZ - gY * rZ) + wY * (gX * rZ - rX * gZ) + wZ * (rX * gY - gX * rY))) /
           (wY * (rX * (gY * bZ - bY * gZ) + gX * (bY * rZ - rY * bZ) + bX * (rY * gZ - gY * rZ)))

      # (Eq. 33)
      coeffs = []
      coeffs[0] = kr
      coeffs[2] = kb
      coeffs[1] = 1.0 - coeffs[0] - coeffs[2]

      return coeffs

    main = ->
      primaries =
        "BT470M":  [ 0.67, 0.33, 0.21, 0.71, 0.14, 0.08, 0.310, 0.316 ]
        "BT470BG": [ 0.64, 0.33, 0.29, 0.60, 0.15, 0.06, 0.3127, 0.3290 ]
        "BT601":   [ 0.630, 0.340, 0.310, 0.595, 0.155, 0.070, 0.3127, 0.3290 ]

      console.log "Looking for 0.299, 0.114:\n"

      for name, p of primaries
        c = calcYUVCoefficients(p)
        console.log "#{name}: Kr: #{c[0].toFixed(4)} Kb: #{c[2].toFixed(4)}"

    main()

... outputs:

    Looking for 0.299, 0.114:

    BT470M: Kr: 0.2990 Kb: 0.1146
    BT470BG: Kr: 0.2220 Kb: 0.0713
    BT601: Kr: 0.2124 Kb: 0.0866

So you're right that `AVIF_COLOR_PRIMARIES_BT470M` happens to match. That said, I don't think CICP `4/x/12` is an interesting combination to check. Perhaps if we come up with a solution for the "enduser expectations" issue cited above, this might end up being fine.

joedrago:
As an aside, I have general concerns that people might expect this to be in some what similar/equivalent to JPEG-XL's ability to repackage a JPEG to JPEG-XL without any data loss, which will never be the case. As I alluded to earlier, the best case we can offer in terms of preserving these values would be a `--min 0 --max 0`,  subsampling-matched, CICP-matched AVIF which would certainly be larger than the original JPEG.

I don't think this is the case with anyone commenting currently on this thread, but I wanted to make it clear for anyone that reads this in the future. Assuming we can figure out the enduser ergonomics around how to request/demand this path and be signaled whether or not it used this path, this will be a minor precision bump (at best, most JPEGs come from RGB anyway) on decode just before AVIF compresses the data down even further. That resultant AVIF will then be converted back to RGB in the future before being displayed.

apps/shared/avifjpeg.c
ValZapod: > The derived matrix coefficients for BT601 either come from

You know how the matrix coeff. are derived from primaries? Can you paste on Wikipedia? I set two citation needed in YCbCr article for that. https://en.wikipedia.org/wiki/YCbCr#ITU-R_BT.709_conversion

Now, from T.871 you can read that JPEG uses full range JPEG that uses BT.601 matrix (coeff. of 0.299, 0.587, 0.114) and after that without ICC profile sRGB is assumed. That means not 2.2, but almost 2.2 piesewise with linear segment function and BT.709 primaries.

It is NOT 470M since 470M uses (from ITU-T H.273) perfect 2.2 gamma and aboslutely different primaries, including DIFFERENT WHITE POINT, white C and not even series D of illuminats.

joedrago: I've supplied the math I did for my check afterwards. I think we should avoid offering this functionality when `MC=12`, to be safe.

ValZapod:
> repackage a JPEG to JPEG-XL without any data loss, which will never be the case

No, it will not. **I** already know that, for example. But really, this in not really the point. This one issue about different JPEG algorithm really destroys the whole idea of this stuff. It does not really matter. https://github.com/mozilla/mozjpeg/issues/182

apps/shared/avifjpeg.c
ValZapod: Oh, that is page 10 of ITU-T H.273! Sigh.

joedrago:
> I updated this PR, and now it only contains 'preserving YUV when reading JPEG if possible' part.
> 
> I suggest we add an 'automatic' value for `requestedFormat` (or let `AVIF_PIXEL_FORMAT_NONE` be that), like 0 for `requestedDepth`, which allows selecting best format according to input file.

Using `AVIF_PIXEL_FORMAT_NONE` for this might be fine. We'd have to audit all of the decoders to ensure that they behave correctly when the requestedFormat is `AVIF_PIXEL_FORMAT_NONE`. It might be smarter to be explicit about whether or not the format (or any other value) was explicitly set, though.

joedrago:
@tongyuantongyu I've looked over the decoders and `avifdec.c`, and I think we can get away with defaulting `input.requestedFormat` to `AVIF_PIXEL_FORMAT_NONE`, and then adding these two lines to `avifutil.h`:

    // Used by image decoders when the user doesn't explicitly choose a format with --yuv
    #define AVIF_APP_DEFAULT_PIXEL_FORMAT AVIF_PIXEL_FORMAT_YUV444

The Y4M decoder won't have to change, as it doesn't leverage `requestedFormat` already. In both the JPEG and PNG decoders though, they'll have to detect when `requestedFormat == AVIF_PIXEL_FORMAT_NONE` and will have to (at least) fall back to `AVIF_APP_DEFAULT_PIXEL_FORMAT`. This would allow for the "automatic" feature you've suggested.

apps/shared/avifjpeg.c
ValZapod: Wait a second, is not those equestions only for "Otherwise (MatrixCoefficients is equal to 12 or 13)," which are for very special "Chromaticity-derived non-constant luminance system"? Or are those others also derived like that?

joedrago:
If having an automatic format makes the most common paths use this new feature, then the only main issue left would be signaling when the JPEG reader uses this. Perhaps we can start simple and just printf() when it _does_ use it, as the case up until now has been to print nothing. We can make this more explicit in the future if we want, but as long as the enduser knows that they will at least see a line printed when this happens, we can add more restrictions to this later.

apps/shared/avifjpeg.c
joedrago: (I'm not sure if this was a question for me or not, but I will supply some info and hope I answer something for you.)

For any `AVIF_MATRIX_COEFFICIENTS_*` enum that ends up being a typical matrix multiply, we have a table of coefficients that simply cite the values in Table 4 of H.273:

https://github.com/AOMediaCodec/libavif/blob/master/src/colr.c#L81

libavif won't do any of that math on MC=1, for example, it'll just use whatever is in that table. That said, you can certainly start with any set of color primaries and (using the math above or in libavif's `calcYUVInfoFromCICP()`), you can derive the coefficients. They correspond exactly for CP/MC=1 (BT709 and CP/MC=9 (BT2020), for example.

ValZapod: Yeah, I already found the originated source with proper equestions. The derivation of the matrix from primaries is called normalized primary matrix and then luminance equation for Y' is derived. Yeah. It is SMPTE RP 177.

Just a little bit of linaer algebra.

apps/shared/avifjpeg.c
tongyuantongyu: Done

tongyuantongyu:
Command line tools need some changes or this feature, probably a `--yuv auto` and make it default is good. @joedrago, do you want these changes in this PR or have another PR for it?

apps/shared/avifjpeg.c
tongyuantongyu: > I've supplied the math I did for my check afterwards. I think we should avoid offering this functionality when `MC=12`, to be safe.

My consideration is, if YUV->RGB->YUV gives the same result, we just skip it and use YUV directly.

ValZapod: "BT601":   [ 0.630, 0.340, 0.310, 0.595, 0.155, 0.070, 0.3127, 0.3290 ]

>these are actually BT.709 primaries...

joedrago: No, they aren't.

From H.273:

image

I appreciate your interest in this particular PR and curiosities surrounding color math, but this is way out of scope, and incorrect. Please open a fresh GitHub issue if you discover incorrect constants in libavif.

ValZapod: Yeah, you right. It is for SMPTE 270M matrix. Wow.

ValZapod: Now, I do not understand how DCI-D65 primaries that are used in Netflix are used with BT.2020-ncl matrix. Should not a different matrix be used? Uuurgh.

ValZapod: Wow, try to google: 

BT470BG: Kr: 0.2220 Kb: 0.0713
Apparently it was defined in 1987, even though is not known now! Wow. Haha.

Also BT.709-1 (that is version 1) used slighly different coeff. from version 2, interesting why. (See wiki again.)

joedrago:
Alright, I've done a full review, and have made minor adjustments to the code, starting with the previous push from @tongyuantongyu (the repeated `goto success` was heading the opposite direction than I was intending with my comment). I then implemented `--yuv auto`, and took a shot at interpreting it correctly in `avifjpeg.c`. All commits are here:

https://github.com/AOMediaCodec/libavif/compare/raw_input_from_jpeg

It should be 2 commits from @tongyuantongyu (`9f87e58` and `5295c1a`), followed by two commits from me (`cdf3765` and `844da84`). 

[cdf3765](https://github.com/AOMediaCodec/libavif/commit/cdf3765dc2f5fa14f9785293e591c2c0619cc145) contains all the changes to the PR. It includes some clarifying comments, fixing the detection of "compatible YUV coefficients", and a minor flow change where I eliminated one goto and distributed out one call to `jpeg_finish_decompress()`. 

844da84[](https://github.com/AOMediaCodec/libavif/commit/844da84e8b1b0c5553b6a9b6f625e0c0bc650e81) is my implementation of `--yuv auto`, using `AVIF_PIXEL_FORMAT_NONE` as "auto" during image reading, as we discussed.

@tongyuantongyu and @wantehchang - Please review these two new commits (and for @wantehchang, please review all 4 commits if you haven't read @tongyuantongyu's commits yet), and let me know if they are acceptable. If so, these 4 commits are what I'll end up merging into `master`.

tongyuantongyu:
In [avifenc.c#746](https://github.com/AOMediaCodec/libavif/compare/raw_input_from_jpeg#diff-fd2436a9cdc2b36bc8bbde03350f5bd6e510eaceb44da44510e1425eb7d5b5ddR746) we may instead set `input.requestedFormat` to `AVIF_PIXEL_FORMAT_YUV444` if it's `AVIF_PIXEL_FORMAT_NONE`.

And we may also set yuvFormat to  `AVIF_PIXEL_FORMAT_YUV400` if we got a grayscale PNG input.

joedrago:
This is now in. We can add minor changes/features to it in separate PRs.

PR #434, opened 2020-12-07
Replace non-ASCII character in avif.h to solve C4819 on MSVC.

tongyuantongyu:
MSVC warns C4819 (The file contains a character that cannot be represented in the current code page) and wouldn't compile, due to a non-ASCII character (`\u2010`, HYPHEN) in `avif.h`.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -280,11 +280,11 @@ typedef struct avifCleanApertureBox
     uint32_t heightN;
     uint32_t heightD;
 
-    // a fractional number which defines the horizontal offset of clean aperture centre minus (width‐1)/2. Typically 0.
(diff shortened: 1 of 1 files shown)


wantehchang:
LGTM. Thanks!

I seem to remember I compiled libavif with MSVC before. I don't know why I didn't run into this compiler warning.

wantehchang (approved):

tongyuantongyu:
> LGTM. Thanks!
> 
> I seem to remember I compiled libavif with MSVC before. I don't know why I didn't run into this compiler warning.

It depends on the current code page being used by Windows (which usually related to system language).

wantehchang:
Thank you for the explanation. I confirm that I did compile libavif with MSVC successfully before. I am using US English version of Windows. That must be why I didn't run into this MSVC warning.

PR #438, opened 2020-12-08
Solve all possible MSVC C4819

tongyuantongyu:
Supplement for #434. Build using VS project files generates warns on more files, so I just went through the whole project, replaced all occurrences of non-ASCII characters in every file, and add macro around every third-party header to ensure it will never happen again.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,9 +153,14 @@ elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")
         /wd4820 # Disable: bytes padding added after data member
         /wd4996 # Disable: potentially unsafe stdlib methods
         /wd5045 # Disable: Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified
-        # See https://github.com/AOMediaCodec/libavif/pull/438
-        # Error when attempting to print non-ASCII character to console
-        /source-charset:.65001 /execution-charset:.20127
+        # Note: This tells MS CL to read source code as UTF-8,
+        # and assume console can only use ASCII (minimal safe).
+        # libavif uses ANSI API to print to console, which is not portable between systems using different
+        # languages and result in mojibake unless we only use codes that shared by every code page: ASCII.
+        # A C4556 warning will be generated on violation.
+        # Commonly used /utf-8 flag assume UTF-8 for both source and console, which is usually not the case.
+        # Warnings can be suppressed but there will still be random characters printed to the console.
+        /source-charset:utf-8 /execution-charset:us-ascii
     )
 else()
     MESSAGE(FATAL_ERROR "libavif: Unknown compiler, bailing out")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,6 +153,9 @@ elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")
         /wd4820 # Disable: bytes padding added after data member
         /wd4996 # Disable: potentially unsafe stdlib methods
         /wd5045 # Disable: Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified
+        # See https://github.com/AOMediaCodec/libavif/pull/438
+        # Error when attempting to print non-ASCII character to console
+        /source-charset:.65001 /execution-charset:.20127
     )
 else()
     MESSAGE(FATAL_ERROR "libavif: Unknown compiler, bailing out")
--- a/src/obu.c
+++ b/src/obu.c
@@ -1,6 +1,6 @@
 /*
- * Copyright © 2018, VideoLAN and dav1d authors
- * Copyright © 2018, Two Orioles, LLC
+ * Copyright (C) 2018, VideoLAN and dav1d authors
+ * Copyright (C) 2018, Two Orioles, LLC
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
--- a/src/obu.c
+++ b/src/obu.c
@@ -1,6 +1,6 @@
 /*
- * Copyright (C) 2018, VideoLAN and dav1d authors
- * Copyright (C) 2018, Two Orioles, LLC
+ * Copyright © 2018, VideoLAN and dav1d authors
+ * Copyright © 2018, Two Orioles, LLC
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -8,7 +8,14 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable : 4819) // The file contains a character that cannot be represented in the current code page
+#endif
 #include "jpeglib.h"
+#if defined(_MSC_VER)
+#pragma warning(pop)
+#endif
 
 #include "iccjpeg.h"
 
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -8,14 +8,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#if defined(_MSC_VER)
-#pragma warning(push)
-#pragma warning(disable : 4819) // The file contains a character that cannot be represented in the current code page
-#endif
 #include "jpeglib.h"
-#if defined(_MSC_VER)
-#pragma warning(pop)
-#endif
 
 #include "iccjpeg.h"
 
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -3,14 +3,7 @@
 
 #include "avifpng.h"
 
-#if defined(_MSC_VER)
-#pragma warning(push)
-#pragma warning(disable : 4819) // The file contains a character that cannot be represented in the current code page
-#endif
 #include "png.h"
-#if defined(_MSC_VER)
-#pragma warning(pop)
-#endif
 
 #include <stdio.h>
 #include <stdlib.h>
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -3,7 +3,14 @@
 
 #include "avifpng.h"
 
+#if defined(_MSC_VER)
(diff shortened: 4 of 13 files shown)


wantehchang:
Hi Yuan Tong,

Thank you very much for writing this pull request! After reading it through, I think something is wrong with this compiler warning if the public header files of all the external dependencies trigger this compiler warning. So I suggest two things:

1. Move the changes to libavif's own files to a separate pull request. This part is not controversial and can be committed immediately.

2. For the public headers of external dependencies, we have several options:

- I suggest we disable the warning globally by adding /wd4819 to CMakeLists.txt.
  * https://bugzilla.mozilla.org/show_bug.cgi?id=826611
- It seems that /utf-8 can also fix this warning. See
  * https://stackoverflow.com/questions/47690822/possible-to-force-cmake-msvc-to-use-utf-8-encoding-for-source-files-without-a-bo
  * https://github.com/ebassi/graphene/pull/173/
- If we decide to disable the warning locally as done in this pull request, the comments should ideally note the offending characters.

Ideally, we should also send patches to the maintainers of the external dependencies so that they can fix the offending characters, if the characters can be easily avoided. I can take care of libaom and libgav1 if you tell me what the offending characters are.

tongyuantongyu:
Here's my opinion:

- I'm against adding /wd4819 as this warning actually helps sometimes, if the characeter is in a string. For example, this is some possible output from `contrib/gdk-pixbuf/loader.c`:

> AVIF quality must be a value between 0 and 100; value “101” is not allowed.

And on my system (Chinese Simplified, Code page 936 (GBK)), when printed to the console, it becomes:

> AVIF quality must be a value between 0 and 100; value 鈥?01鈥?is not allowed.

Which is quite confusing.

- /utf-8 doesn't solve the issue above. It does the same as /wd4819.

(There was a lot of words, but while writing I found a good solution)

I found a good way to solve this: add `/source-charset:.65001 /execution-charset:.20127`, it means read the source as UTF-8, but assume a console that can only print ASCII. When encountering a non-ASCII character in string, a C4556 (character represented by universal-character-name 'char' cannot be represented in the current code page (20127)) will be generated, but comment usage will be ignored silently.

I'll update this pr.

CMakeLists.txt
```diff
@@ -153,6 +153,9 @@ elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")
         /wd4820 # Disable: bytes padding added after data member
         /wd4996 # Disable: potentially unsafe stdlib methods
         /wd5045 # Disable: Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified
+        # See https://github.com/AOMediaCodec/libavif/pull/438
```
wantehchang, line 156: Instead of referring to this pull request, could you replace this comment with an explanation? You can just edit and convert the following explanation of yours to a comment:

> I found a good way to solve this: add /source-charset:.65001 /execution-charset:.20127, it means read the source as UTF-8, but assume a console that can only print ASCII. When encountering a non-ASCII character in string, a C4556 (character represented by universal-character-name 'char' cannot be represented in the current code page (20127)) will be generated, but comment usage will be ignored silently.

src/obu.c
```diff
@@ -1,6 +1,6 @@
 /*
- * Copyright © 2018, VideoLAN and dav1d authors
- * Copyright © 2018, Two Orioles, LLC
+ * Copyright (C) 2018, VideoLAN and dav1d authors
+ * Copyright (C) 2018, Two Orioles, LLC
```
wantehchang, line 3: Is it possible to change these back to the copyright symbol? This is dav1d's license header. It would be better if we can have it in the original form.

CMakeLists.txt
```diff
@@ -153,6 +153,9 @@ elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")
...
+        # Error when attempting to print non-ASCII character to console
+        /source-charset:.65001 /execution-charset:.20127
```
wantehchang, line 158: 1. Could you change this to
         /source-charset:utf-8 /execution-charset:us-ascii
?

If this form works, it is easier to understand than the code page identifiers as decimal numbers.

2. Could you also add a comment to explain why /source-charset:.65001 /execution-charset:.20127 is better than /utf-8?

Details:
I found this Chromium issue on using the /utf-8 compiler flag on Windows:
https://bugs.chromium.org/p/chromium/issues/detail?id=637203

and the changelist:
https://codereview.chromium.org/2543743002

I verified that the /utf-8 flag is still in the current Chromium source tree:
https://source.chromium.org/chromium/chromium/src/+/master:build/config/win/BUILD.gn;l=64?q=%22%2Futf-8%22&ss=chromium&start=51

I also found the /utf-8 flag used by two third-party libraries used by Chromium: re2 and harfbuzz-ng. So the /utf-8 flag seems like a well-known solution to this problem. It would be good to explain why /source-charset:.65001 /execution-charset:.20127 is a superior solution.

tongyuantongyu:
As for the offending characters, it's a '√' from `rav1e.h`, introduced by this [commit](https://github.com/xiph/rav1e/commit/3bf5b60265e3abcc59ed3607802a7f0a93e01213#diff-d127913419285b5fb83bb79e44d3556b25f71ba47154bc4dbe4551e8008b0346R142) 16 months ago. It was there since libavif supports rav1e, and no one even noticed that. I'm worrying if another dependency also introduced something like this without being noticed, and constantly block users of a specific language not being able to compile it, so I did that.

Anyway, there's actually offending characters in libaom and libgav1, but not in the header libavif includes. I still got warning, but they compiled anyway.

libaom:
```diff
diff --git a/aom_ports/x86.h b/aom_ports/x86.h
index 8c18448..cf1c3e6 100644
--- a/aom_ports/x86.h
+++ b/aom_ports/x86.h
@@ -236,7 +236,7 @@ static INLINE int x86_simd_caps(void) {
 // x86_readtsc directly, but prevent the CPU's out-of-order execution from
 // affecting the measurement (by having earlier/later instructions be evaluated
 // in the time interval). See the white paper, "How to Benchmark Code
-// Execution Times on Intel® IA-32 and IA-64 Instruction Set Architectures" by
+// Execution Times on Intel(R) IA-32 and IA-64 Instruction Set Architectures" by
 // Gabriele Paoloni for more information.
 //
 // If you are timing a large function (CPU time > a couple of seconds), use
diff --git a/av1/av1_dx_iface.c b/av1/av1_dx_iface.c
index d821a52..3ba4dfe 100644
--- a/av1/av1_dx_iface.c
+++ b/av1/av1_dx_iface.c
@@ -153,7 +153,7 @@ static aom_codec_err_t parse_timing_info(struct aom_read_bit_buffer *rb) {
   if (equal_picture_interval) {
     const uint32_t num_ticks_per_picture_minus_1 = aom_rb_read_uvlc(rb);
     if (num_ticks_per_picture_minus_1 == UINT32_MAX) {
-      // num_ticks_per_picture_minus_1 cannot be (1 << 32) − 1.
+      // num_ticks_per_picture_minus_1 cannot be (1 << 32) - 1.
       return AOM_CODEC_UNSUP_BITSTREAM;
     }
   }
diff --git a/av1/decoder/decodeframe.c b/av1/decoder/decodeframe.c
index 7abfac4..ea4febf 100644
--- a/av1/decoder/decodeframe.c
+++ b/av1/decoder/decodeframe.c
@@ -4127,7 +4127,7 @@ void av1_read_timing_info_header(aom_timing_info_t *timing_info,
     if (num_ticks_per_picture_minus_1 == UINT32_MAX) {
       aom_internal_error(
           error, AOM_CODEC_UNSUP_BITSTREAM,
-          "num_ticks_per_picture_minus_1 cannot be (1 << 32) − 1.");
+          "num_ticks_per_picture_minus_1 cannot be (1 << 32) - 1.");
     }
     timing_info->num_ticks_per_picture = num_ticks_per_picture_minus_1 + 1;
   }
```

libgav1:
```diff
diff --git a/src/utils/raw_bit_reader.h b/src/utils/raw_bit_reader.h
index 76e7bfa..7d8ce8f 100644
--- a/src/utils/raw_bit_reader.h
+++ b/src/utils/raw_bit_reader.h
@@ -38,7 +38,7 @@ class RawBitReader : public BitReader, public Allocable {
                         size_t* value);    // le(n) in the spec.
   bool ReadUnsignedLeb128(size_t* value);  // leb128() in the spec.
   // Reads a variable length unsigned number and stores it in |*value|. On a
-  // successful return, |*value| is in the range of 0 to UINT32_MAX − 1,
+  // successful return, |*value| is in the range of 0 to UINT32_MAX - 1,
   // inclusive.
   bool ReadUvlc(uint32_t* value);  // uvlc() in the spec.
   bool Finished() const;
```

The third from libaom in `av1/decoder/decodeframe.c` is a non-comment usage, which might result in some problem.

wantehchang (changes requested):
Hi Tong Yuan,

Thank you again! I suggest adding more detail to the comment for your fix. I will create changelists for libaom and libgav1 now.

CMakeLists.txt
tongyuantongyu: If those strings that contains unicode is not going to be printed to console using `printf` then use /utf-8 flag is fine. Both re2 and harfbuzz are text-processing related libraries, so /utf-8 precisely fit their need (note that /execution-charset also specify the encoding used to store strings in executable). But for command-line tools' messages, it's better to use ascii only - even if you use `SetConsoleOutputCP` WinAPI to change current code page to UTF-8, there's still a chance that the user's font can't display it.

CMakeLists.txt
```diff
@@ -153,6 +153,14 @@ elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")
         /wd4820 # Disable: bytes padding added after data member
         /wd4996 # Disable: potentially unsafe stdlib methods
         /wd5045 # Disable: Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified
+        # Note: This tells MS CL to read source code as UTF-8,
```
wantehchang, line 156: Nit: Change "MS CL" to "MSVC" or "Visual C++".

wantehchang (commented):
LGTM. Thanks!

PR #444, opened 2020-12-12
Add Sharp YUV conversion

tongyuantongyu:
Sharp YUV is a YUV->RGB conversion algorithm from libwebp, which pays extra effort to reduce the negative effects caused by chroma subsampling.

I rewrote the algorithm to use float point to provide enough precision on high depth image, add support for all TC and MC libavif has (MC only includes those can be represented as Kr and Kb because the algorithm only make improvement on these MCs), and extend to also support YUV422.

The performance should be acceptable, though not being turned. It takes extra ~5s on my computer to process a 3840x2160 resolution image.


joedrago:
This PR will require a much more detailed audit than I can give it right now, but a couple of first impressions:

* No additional members should be on the avifImage struct. YUV options go on the avifRGBImage struct.

* There is already an opaque checking function in libavif. Either use it or replace it, but I don't want two of them.

I will offer more comments on Monday when I get a chance to read the whole thing.

wantehchang:
Hi Yuan Tong (@tongyuantongyu),

I am investigating an AVIF YUV 4:2:0 image quality issue. The AVIF YUV 4:4:4 and WebP versions of the image do not have the quality issue.

I did an experiment, which showed that using bilinear interpolation in the chrome upsampling process does not seem to correct this quality issue. So I suspect this quality issue could be corrected by using the sharp YUV conversion.

Would you have time to rebase this pull request on the current tip of the trunk? If not, I can manually apply the changes and experiment with it. Thanks.

tongyuantongyu:
> Hi Yuan Tong (@tongyuantongyu),
> 
> I am investigating an AVIF YUV 4:2:0 image quality issue. The AVIF YUV 4:4:4 and WebP versions of the image do not have the quality issue.
> 
> I did an experiment, which showed that using bilinear interpolation in the chrome upsampling process does not seem to correct this quality issue. So I suspect this quality issue could be corrected by using the sharp YUV conversion.
> 
> Would you have time to rebase this pull request on the current tip of the trunk? If not, I can manually apply the changes and experiment with it. Thanks.

Done. Please check out https://github.com/tongyuantongyu/libavif/tree/sharp_yuv.
If the quality issue is something about saturated sharp border appeared dimmer then it's probably related to sharp YUV.

I'm not very satisfied with the current code and is planning a rewrite, but the functionality should be fine. The result may be slightly different from libwebp's version, possibly due to libwebp uses fixed point arithmetic but this PR uses float point.

Note that `cwebp` only uses sharp YUV conversion if `-sharp_yuv` option is used. You can check if the webp version produced without `-sharp_yuv` option also has this issue. You may also consider let libwebp do the conversion and steal YUV data from `WebPPicture` into `avifImage` and check the result.

wantehchang:
Hi Yuan Tong,

Thank you very much for your help! I will test your patch. I may also try to use the WebPPictureSharpARGBToYUVA() function directly.

wantehchang:
Hi Yuan Tong (@tongyuantongyu),

Your https://github.com/tongyuantongyu/libavif/tree/sharp_yuv seems to have a merge error. I need to apply the following patch to it:

```
$ diff -pu libavif/src/write.c.orig libavif/src/write.c
--- libavif/src/write.c.orig        2021-06-13 11:52:24.000000000 -0700
+++ libavif/src/write.c     2021-06-13 11:54:14.000000000 -0700
@@ -500,6 +500,7 @@ static avifResult avifEncoderAddImageInt
             }
         }
 
+        encoder->data->alphaPresent = (firstCell->alphaPlane != NULL);
         if (encoder->data->alphaPresent && (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE)) {
             // If encoding a single image in which the alpha plane exists but is entirely opaque,
             // simply skip writing an alpha AV1 payload entirely, as it'll be interpreted as opaque
```

Also, the change from `avifImageIsOpaque()` to `avifCheckAlphaOpaque()` is not necessary. The only behavioral change is to handle limited-range alpha, which has been deprecated in the draft of the next revision of the AVIF spec. More importantly, the use of `avifAlphaParams` as the input parameter is confusing because none of the `dst*` fields of `avifAlphaParams` are initialized.

tongyuantongyu:
> Your https://github.com/tongyuantongyu/libavif/tree/sharp_yuv seems to have a merge error. I need to apply the following patch to it:

Updated. I also updated the code to handle alpha premultiply correctly.

 > Also, the change from `avifImageIsOpaque()` to `avifCheckAlphaOpaque()` is not necessary.

The main purpose is to reuse code. In `write.c`, it checks if the alpha plane of an `avifImage` is opaque, and skip encoding alpha plane if so. In `reformat_sharp_yuv.c`, it checks if the alpha channel of an `avifRGBImage` is opaque, and skip doing some work related to alpha. I've updated the code to drop the support for deprecated limited range alpha plane.

> More importantly, the use of avifAlphaParams as the input parameter is confusing because none of the dst* fields of avifAlphaParams are initialized.

There's also `avifFillAlpha()` that uses `avifAlphaParams` but none of `src*` fields are initialized. I think the better solution is to pass alpha info of src and dst in different paramters.

maryla-uc:
Hey, thank you for your work on this. I was wondering why you closed the PR? you said you were planning on rewriting it?
I haven't looked at it in great detail, but here are a few thoughts:
- I tried out your PR on a few images and the visual quality was great, nice job!
- I see that this is not just a reimplementation of webp's sharpyuv. Looking at the code, there's a bunch of extra things around edge preservation (in avifSharpYUVDRGBWeightFilter and avifSharpYUVRGB2DRGB422Weighted). What did you base this on?
- I compared the result with libwebp's sharp yuv and it's actually pretty different. Your version gives a smoother border. Do you know why that is? I tried disabling the weights mentioned above but it didn't change much. Do you think it's just the floating point arithmetic? I wouldn't expect it to make a huge difference but maybe it does.

tongyuantongyu:
@maryla-uc 
Hi Maryla: glad to hear your interest on this!

Plan on rewriting: I used interleaved RGB layout for temporary buffers because it was easier to understand for me. Later I found it'd be harder to add SIMD support than using planar RGB layout, so I plan to rewrite to use planar layout. (and maybe write a dedicated library)

About extra things around edge preservation: These are my own modifications on top of the original SharpYUV conversion, based on some observations of the original algorithm's output:
- Original SharpYUV algorithm still dims out sharp lines quite a lot, because it treats all pixels the same importance. Foreground line pixels affect perception more, so I'm giving it more weight during the decision of result chroma value.
- After the adjustment above, around sharp lines appear some heavy halos. This is due to pixels around lines also assigned bigger weights, so I use the stddev around some pixel to adaptively switch between line prefered and uniform weight mode.

About differences with libwebp's SharpYUV: I'm also a bit confused about the difference here. I wrote this version while studying libwebp's sharp yuv code, and about using float instead of fixed point, it's because I want to support 10/12bit and HDR input.
Note that changing matrix coefficient makes big difference to the result, and using MC=7 gives a result looks quite like the original sharp yuv, but I can't explain why it is the case, and it looks quite coincidental.
Maybe you can try converting my implementation to fixed point, to see if it is caused by fixed/float arithmetic.

maryla-uc:
Hey, thank you for the details.

Your changes to handle sharp lines are intesresting. I tried with and without it on [this image](https://www.ctrl.blog/media/image/image-subsampling-test.png) and it does make a noticeable difference. On other real life images I couldn't really tell the difference however. I guess that's the good thing if it improves fine lines without changing anything else.

I also tried using different matrix coeffcients and different transfer characteristics.
You are right that MC=7 is a bit closer to webp but not that much. Other MC values didn't seem to make a big difference except for YCgCo (8) which seems a bit broken (gives black borders worse than without sharpyuv).

Changing the transfer characteristics (TC) makes a bigger difference, with a lot of the images ending up with a black border. By the way it's a minor detail but I think that in avifTransferCharacteristicsGetConverter the default should be to/from SRGB instead of 709, so that it matches the default when no icc and co cicp info is provided (see https://github.com/AOMediaCodec/libavif/wiki/CICP#unspecified )

By the way, as you saw we're also looking at adding SharpYuv to libheif https://github.com/strukturag/libheif/issues/586 so it would make sense to have this code in a separate library.

tongyuantongyu:
> On other real life images

I'm not sure what "real life images" means, but if it means photographs, I suggest also test with illustrations and synthetic images. People will use avif for all kinds of images.

> except for YCgCo (8) which seems a bit broken

SharpYUV only works on MCs defined through Kr and Kb, which doesn't include YCgCo. The reformat part in libavif didn't have a way to emit warning, so it silently fallback to traditional YUV conversion.

 > Changing the transfer characteristics (TC) makes a bigger difference

Chaging TC changes the interpretation of colors. You should view the result with tools that handles it correctly. Chrome handles it but has chroma upsampling issue. Joe wrote a tool that handles it: https://github.com/joedrago/vantage. 
But generally, you shouldn't change TC as it changes the meaning of  **_input_** RGB data.

> make sense to have this code in a separate library

Yes I also planned that. It may also worth consider to add it to libyuv. Anyway I'd like to help.



As you've already seen, SharpYUV expect decode end to use bilinear chroma upsampling, and using nearest can instead decrease the result quality, which is libyuv's case (and in consequence Chrome and Firefox). Chrome can follow the change after libyuv is updated, but I'm a bit worry about Firefox. Its libyuv version is now lagged far behind, due to some build infra issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1698699.

baumanj:
Hey @tongyuantongyu, I just wanted to let you know that we haven't forgotten about this issue. We're still the the process of investigating whether we want to pursue continuing to use libyuv in Firefox (and getting our version up to date, and keeping it that way) or use different code for the YUV -> RGB conversion in the AVIF case.

PR #445, opened 2020-12-13
Fix few compile warnings

skal65535:
Fixes:

* missing 'static' in function signatures
* use (void*) case to work around obscure C rule about 'const ptr**' versus 'ptr**':
    
    avifenc.c:997:86: error: cast from 'avifImage **' (aka 'struct avifImage **') to 'const avifImage **' (aka 'const struct avifImage **')
          must have all intermediate pointers const qualified to be safe [-Werror,-Wcast-qual]
                encoder, (uint8_t)gridDims[0], (uint8_t)gridDims[1], (const avifImage **)gridCells, AVIF_ADD_IMAGE_FLAG_SINGLE);
                                                                                         ^


tongyuantongyu:
I think the right way to solve it should be changing the type of `cellImages` parameter to `const avifImage * const *`, and then cast `gridCells` from `avifImage **` to `const avifImage * const *`.

There's some potential problems by allowing `T **` implicitly casts to `const T **`: http://c-faq.com/ansi/constmismatch.html. Implicit castfrom `T **` to `const T * const *` should be fine, and is allowed in C++ but C chose to use simpler rule: always forbid nested pointer being implicitly cast to its const variant and require a explicit cast. (See how complicated C++'s rule is: https://en.cppreference.com/w/cpp/language/implicit_conversion#Qualification_conversions)

The compiler employs C++'s rule on the casting, and warns about possible wrongs here, so we shouldn't ignore that.

apps/avifenc.c
```diff
@@ -994,7 +994,7 @@ int main(int argc, char * argv[])
 
     if (gridDimsCount > 0) {
         avifResult addImageResult = avifEncoderAddImageGrid(
-            encoder, (uint8_t)gridDims[0], (uint8_t)gridDims[1], (const avifImage **)gridCells, AVIF_ADD_IMAGE_FLAG_SINGLE);
+            encoder, (uint8_t)gridDims[0], (uint8_t)gridDims[1], (void*)gridCells, AVIF_ADD_IMAGE_FLAG_SINGLE);
```
wantehchang, line 997: Hi Pascal: Thank you for writing a fix! Casting to the intermediate void* is not ideal. I like Yuan Tong's suggestion of adding one more const to the declaration of the gridCells parameter of avifEncoderAddImageGrid(). I verified it works.

wantehchang:
Hi Yuan Tong,

This is an annoying problem in C. (I previously ran into a similar problem with the execve() system call.) Thank you for the research. Your suggestion works.

```
diff --git a/apps/avifenc.c b/apps/avifenc.c
index 56169df..daabbde 100644
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -272,7 +272,7 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
     return AVIF_TRUE;
 }
 
-avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gridCols, uint32_t gridRows, avifImage ** gridCells)
+static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gridCols, uint32_t gridRows, avifImage ** gridCells)
 {
     if ((gridSplitImage->width % gridCols) != 0) {
         fprintf(stderr, "ERROR: Can't split image width (%u) evenly into %u columns.\n", gridSplitImage->width, gridCols);
@@ -994,7 +994,7 @@ int main(int argc, char * argv[])
 
     if (gridDimsCount > 0) {
         avifResult addImageResult = avifEncoderAddImageGrid(
-            encoder, (uint8_t)gridDims[0], (uint8_t)gridDims[1], (const avifImage **)gridCells, AVIF_ADD_IMAGE_FLAG_SINGLE);
+            encoder, (uint8_t)gridDims[0], (uint8_t)gridDims[1], (const avifImage * const *)gridCells, AVIF_ADD_IMAGE_FLAG_SINGLE);
         if (addImageResult != AVIF_RESULT_OK) {
             fprintf(stderr, "ERROR: Failed to encode image: %s\n", avifResultToString(addImageResult));
             returnCode = 1;
diff --git a/include/avif/avif.h b/include/avif/avif.h
index c65ba9c..96fdaec 100644
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -865,7 +865,11 @@ enum avifAddImageFlags
 //
 
 AVIF_API avifResult avifEncoderAddImage(avifEncoder * encoder, const avifImage * image, uint64_t durationInTimescales, uint32_t addImageFlags);
-AVIF_API avifResult avifEncoderAddImageGrid(avifEncoder * encoder, uint8_t gridCols, uint8_t gridRows, const avifImage ** cellImages, uint32_t addImageFlags);
+AVIF_API avifResult avifEncoderAddImageGrid(avifEncoder * encoder,
+                                            uint8_t gridCols,
+                                            uint8_t gridRows,
+                                            const avifImage * const * cellImages,
+                                            uint32_t addImageFlags);
 AVIF_API avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output);
 
 // Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs. These
index a6c3671..f45afcd 100644
--- a/src/write.c
+++ b/src/write.c
@@ -382,7 +382,7 @@ static void avifWriteGridPayload(avifRWData * data, uint8_t gridCols, uint8_t gr
 static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
                                               uint8_t gridCols,
                                               uint8_t gridRows,
-                                              const avifImage ** cellImages,
+                                              const avifImage * const * cellImages,
                                               uint64_t durationInTimescales,
                                               uint32_t addImageFlags)
 {
@@ -626,13 +626,13 @@ avifResult avifEncoderAddImage(avifEncoder * encoder, const avifImage * image, u
     return avifEncoderAddImageInternal(encoder, 1, 1, &image, durationInTimescales, addImageFlags);
 }
 
-avifResult avifEncoderAddImageGrid(avifEncoder * encoder, uint8_t gridCols, uint8_t gridRows, const avifImage ** cellImages, uint32_t addImageFlags)
+avifResult avifEncoderAddImageGrid(avifEncoder * encoder, uint8_t gridCols, uint8_t gridRows, const avifImage * const * cellImages, uint32_t addImageFlags)
 {
     return avifEncoderAddImageInternal(
         encoder, gridCols, gridRows, cellImages, 1, addImageFlags | AVIF_ADD_IMAGE_FLAG_SINGLE); // only single image grids are supported
 }
 
-size_t avifEncoderFindExistingChunk(avifRWStream * s, size_t mdatStartOffset, const uint8_t * data, size_t size)
+static size_t avifEncoderFindExistingChunk(avifRWStream * s, size_t mdatStartOffset, const uint8_t * data, size_t size)
 {
     const size_t mdatCurrentOffset = avifRWStreamOffset(s);
     const size_t mdatSearchSize = mdatCurrentOffset - mdatStartOffset;
```

apps/avifenc.c
skal65535: Wan-Teh, Yuan:  yes, the c-faq description of the problem was the one i was referring to as "obscure C rule".
I thought using "void*" was an easy way for users to not ask questions they probably don't really want to know the answer of ;)

I'm totally fine with doing it the explicit way! ("const ptr* const * ...")

wantehchang:
Pascal: I have replaced this pull request with https://github.com/AOMediaCodec/libavif/pull/446. Thank you again for your help.

PR #446, opened 2020-12-14
Fix -Wmissing-prototypes and -Wcast-qual warnings

wantehchang:
Add 'static' to function signatures of avifImageSplitGrid() and
avifEncoderFindExistingChunk(). This fix was suggested by Pascal
Massimino.

Change the type of the cellImages parameter of avifEncoderAddImageGrid()
from:
    const avifImage ** cellImages
to
    const avifImage * const * cellImages
so that we can cast an avifImage ** pointer to the new type without the
-Wcast-qual warning. This fix was suggested by Yuan Tong.

Fix https://crbug.com/oss-fuzz/28622.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -272,7 +272,7 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
     return AVIF_TRUE;
 }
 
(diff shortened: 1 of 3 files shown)


skal65535:
Thanks for the fix!

PR #447, opened 2020-12-15
Better handling for export headers

tongyuantongyu:
This solve some remaining problems in #439.

- Add case for `__declspec(dllimport)`, and macro `AVIF_DLL` to enable it.
- Add comments to explain this.
- Automatically defines `AVIF_DLL` for user if we can.

Here's what `AVIF_API` will be defined to:
| (Shared) | Windows | Unix |
| --- | --- | --- |
| Build | `__declspec(dllexport)` | `__attribute__((visibility("default")))` |
| Use | `__declspec(dllimport)`| |

For static build, `AVIF_API` always defines to nothing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -406,13 +406,11 @@ target_include_directories(avif
                            PUBLIC $<BUILD_INTERFACE:${libavif_SOURCE_DIR}/include>
                                   $<INSTALL_INTERFACE:include>
                            PRIVATE ${AVIF_PLATFORM_INCLUDES} ${AVIF_CODEC_INCLUDES})
-set(AVIF_PC_EXTRA_DEFINITION "")
+set(AVIF_PKG_CONFIG_EXTRA_CFLAGS "")
 if(BUILD_SHARED_LIBS)
-    target_compile_definitions(avif PRIVATE -DAVIF_BUILDING_SHARED_LIBS)
-    if(WIN32)
-        target_compile_definitions(avif INTERFACE -DAVIF_DLL)
-        set(AVIF_PC_EXTRA_DEFINITION " -DAVIF_DLL")
-    endif()
+    target_compile_definitions(avif PUBLIC AVIF_DLL)
+    target_compile_definitions(avif PRIVATE AVIF_BUILDING_SHARED_LIBS)
+    set(AVIF_PKG_CONFIG_EXTRA_CFLAGS " -DAVIF_DLL")
     if(AVIF_LOCAL_LIBGAV1)
         set_target_properties(avif PROPERTIES LINKER_LANGUAGE "CXX")
     endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -406,13 +406,17 @@ target_include_directories(avif
                            PUBLIC $<BUILD_INTERFACE:${libavif_SOURCE_DIR}/include>
                                   $<INSTALL_INTERFACE:include>
                            PRIVATE ${AVIF_PLATFORM_INCLUDES} ${AVIF_CODEC_INCLUDES})
+set(AVIF_PC_EXTRA_DEFINITION "")
 if(BUILD_SHARED_LIBS)
     target_compile_definitions(avif PRIVATE -DAVIF_BUILDING_SHARED_LIBS)
+    if(WIN32)
+        target_compile_definitions(avif INTERFACE -DAVIF_DLL)
+        set(AVIF_PC_EXTRA_DEFINITION " -DAVIF_DLL")
+    endif()
     if(AVIF_LOCAL_LIBGAV1)
         set_target_properties(avif PROPERTIES LINKER_LANGUAGE "CXX")
     endif()
 endif()
-
 
 option(AVIF_BUILD_EXAMPLES "Build avif Examples." OFF)
 if(AVIF_BUILD_EXAMPLES)
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -7,7 +7,16 @@
 #include <stddef.h>
 #include <stdint.h>
 
-#ifndef AVIF_API
+// ---------------------------------------------------------------------------
+// Export macros
+
+// AVIF_BUILDING_SHARED_LIBS should only be defined when building shared
+// version of libavif library.
+// AVIF_DLL may only be defined when your project dynamically linking to libavif
+// on Windows, to allow compiler to generate more efficient function call
+// assembly. If you are using libavif as CMake dependency, through CMake
+// package config file or through pkg-config this is defined automatically.
+
 #if defined(AVIF_BUILDING_SHARED_LIBS)
 #if defined(_WIN32)
 #define AVIF_API __declspec(dllexport)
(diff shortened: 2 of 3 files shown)


wantehchang (changes requested):
Thank you for writing this pull request! I suggest some changes.

CMakeLists.txt
```diff
@@ -406,14 +406,18 @@ target_include_directories(avif
                            PUBLIC $<BUILD_INTERFACE:${libavif_SOURCE_DIR}/include>
                                   $<INSTALL_INTERFACE:include>
                            PRIVATE ${AVIF_PLATFORM_INCLUDES} ${AVIF_CODEC_INCLUDES})
+set(AVIF_PC_EXTRA_DEFINITION "")
 if(BUILD_SHARED_LIBS)
     target_compile_definitions(avif PRIVATE -DAVIF_BUILDING_SHARED_LIBS)
+    if(WIN32)
```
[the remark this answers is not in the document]
tongyuantongyu: Then maybe we should also define the `AVIF_API` to `__attribute__((visibility("default")))` on Unix when using libavif shared library as "Step-by-step guide" suggest? In my test it doesn't make any difference, and I don't want to bother user with nothing, so I only ask Windows user to define it.

include/avif/avif.h
```diff
@@ -7,7 +7,16 @@
...
-#ifndef AVIF_API
+// ---------------------------------------------------------------------------
+// Export macros
+
+// AVIF_BUILDING_SHARED_LIBS should only be defined when building shared
+// version of libavif library.
+// AVIF_DLL may only be defined when your project dynamically linking to libavif
+// on Windows, to allow compiler to generate more efficient function call
+// assembly. If you are using libavif as CMake dependency, through CMake
+// package config file or through pkg-config this is defined automatically.
```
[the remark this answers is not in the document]
tongyuantongyu: Yes because in CMakeLists.txt#413 I added it as an interface definition: `target_compile_definitions(avif INTERFACE -DAVIF_DLL)`

If you check the installed file (`${install_prefix}/lib/cmake/libavif/libavif-config.cmake`) you can see something like this:
```cmake
set_target_properties(avif PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "AVIF_DLL"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
)
```

CMakeLists.txt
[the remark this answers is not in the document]
wantehchang: When compiling against the header files of a shared library, the header files do not need __attribute__((visibility("default"))) on its public API functions. For example, you can inspect the /usr/include/pthread.h and /usr/include/stdlib.h headers on a Unix system (including Linux). You will see that the function declarations do not have __attribute__((visibility("default"))). I understand that the "Step-by-step guide" has __attribute__((visibility("default"))). But the BoringSSL and protobuf examples I gave in another pull request do not have __attribute__((visibility("default"))).

It is true that AVIF_DLL only matters on Windows. However, if we document that AVIF_DLL should be defined on all platforms when libavif is a shared library, then our clients' build system can be slightly simpler. This is a design trade-off.

CMakeLists.txt
```diff
@@ -406,14 +406,18 @@ target_include_directories(avif
...
+        target_compile_definitions(avif INTERFACE -DAVIF_DLL)
```
wantehchang, line 413: We can omit "-D" in target_compile_definitions:

        target_compile_definitions(avif INTERFACE AVIF_DLL)

Note: The "Step-by-step guide" suggests making AVIF_DLL a PUBLIC definition rather than an INTERFACE definition. Either way will work. We can start with INTERFACE and only change it to PUBLIC when libavif itself needs to know whether it is a shared library or not.

include/avif/avif.h
```diff
@@ -7,7 +7,16 @@
 #include <stddef.h>
 #include <stdint.h>
 
-#ifndef AVIF_API
+// ---------------------------------------------------------------------------
+// Export macros
+
+// AVIF_BUILDING_SHARED_LIBS should only be defined when building shared
+// version of libavif library.
+// AVIF_DLL may only be defined when your project dynamically linking to libavif
```
wantehchang, line 15: Nit: dynamically linking => is dynamically linked

include/avif/avif.h
wantehchang, line 18: Add a comma (,) after "or through pkg-config".

include/avif/avif.h
```diff
@@ -7,7 +7,16 @@
 #include <stddef.h>
 #include <stdint.h>
 
-#ifndef AVIF_API
+// ---------------------------------------------------------------------------
+// Export macros
+
+// AVIF_BUILDING_SHARED_LIBS should only be defined when building shared
+// version of libavif library.
```
wantehchang, line 14: Nit: building shared version of libavif library => libavif is being built as a shared library ?

include/avif/avif.h
```diff
@@ -7,23 +7,39 @@
...
+#endif
+
+// ---------------------------------------------------------------------------
+// Export macros
+
+// AVIF_BUILDING_SHARED_LIBS should only be defined when libavif is being built
+// as a shared library.
+// AVIF_DLL should be defined if libavif is a shared library. If you are using
+// libavif as CMake dependency, through CMake package config file or
+// through pkg-config, this is defined automatically.
```
wantehchang, line 21: Optional: I like your summary in the commit message very much:

> Here's what `AVIF_API` will be defined to:
> | (Shared) | Windows | Unix |
> | --- | --- | --- |
> | Build | `__declspec(dllexport)` | `__attribute__((visibility("default")))` |
> | Use | `__declspec(dllimport)`| |
> 
> For static build, `AVIF_API` always defines to nothing. 

Do you think a plaintext versiopn of that (without the table markup) could become a comment here? Or do you think that comment would merely repeat the code?

wantehchang (approved):
LGTM. Nice work! If you don't have time to make the minor changes I suggested, I can merge this pull request first and then make those changes myself.

include/avif/avif.h
tongyuantongyu: I think it can help the future maintenance of the project by explaining what's going on here cleanly. I would like to add it.

include/avif/avif.h
```diff
@@ -7,23 +7,46 @@
...
+//
+// Here's what `AVIF_API` will be defined to in shared build:
```
wantehchang, line 23: Nit: Change "defined to" to "defined as" (or perhaps "defined to be") here and in line 28.

Also, I think AVIF_API does not need to be quoted in ``. Is `` part of the markdown language?

tongyuantongyu: Updated. Thanks for your advices on my expressions.

wantehchang: You're welcome!

I just used GitHub's "Squash and merge" UI to merge this pull request. It did not seem to give me an opportunity to edit the commit message, so the commit message is just the summary lines of the four commits. For your other pull request I will try the command-line instructions.

wantehchang: Actually, the "Squash and merge" UI has a dialog box but I missed it. I will try it now with the other pull request.

Issue #435, opened 2020-12-07
Failed to compile on MINGW using Clang due to -Wused-but-marked-unused


tongyuantongyu:
Clang warns every use of `printf` and `sprintf`, due to their declaration in MINGW headers is marked unused. (definition of `printf`: https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/crt/stdio.h#L354 , definition of macro `__mingw_ovr`: https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/crt/_mingw_mac.h#L290 ) 
This simple code can reproduce the problem:
```bash
clang -Weverything -Werror -x c - <<EOF
#include <stdio.h>

int main() {
    printf("");
}
EOF
```

Here's GCC's (https://gcc.gnu.org/onlinedocs/gcc-10.2.0/gcc/Common-Function-Attributes.html) and Clang's (https://clang.llvm.org/docs/AttributeReference.html#maybe-unused-unused) document of the `unused` attribute, and Clang's (https://releases.llvm.org/11.0.0/tools/clang/docs/DiagnosticsReference.html#wused-but-marked-unused) document of the diagnostic.

For solution, we either disable this warning by adding `-Wno-used-but-marked-unused`, or by adding definition `-D__USE_MINGW_ANSI_STDIO=0` (on this case MINGW headers use another definition of `*printf` which doesn't contain the `unused` attribute: https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/crt/stdio.h#L561 ), but I'm feeling doing something wrong in both cases, so I created this issue.

wantehchang:
Hi Yuan Tong,

Ideally we should ask the MINGW maintainers how they suggest we fix this compiler warning.

If you don't have time to do that, I suggest we disable the warning with -Wno-used-but-marked-unused. libavif itself does not use the 'unused' attribute, so this compiler warning is not useful to libavif right now. (In fact I think this compiler warning is only appropriate in some cases, and it can be manually enabled in those cases.)

PR #448, opened 2020-12-17
Solve some clang warnings

tongyuantongyu:
- Disable `-Wused-but-marked-unused` on Clang on MINGW. This solves #435.
  I created an issue but didn't get response from MINGW maintainers for a week, so for now just disable the warning in this case.

- Remove extra semicolon in `codec_svt.c`. This fixes `-Wextra-semi-stmt` warning.


CMakeLists.txt
```diff
@@ -137,6 +137,10 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang")
     if(HAVE_POISON_SYSTEM_DIRECTORIES_WARNING)
         add_definitions(-Wno-poison-system-directories)
     endif()
+    # MINGW defines printf with __attribute__ ((__unused__)). It not our fault so we just ignore it.
```
wantehchang, line 140: Nit: change "defines" to "declares" and add "in stdio.h".

Change "It not our fault" to "It is out of our control".

Please add the URL of the MINGW issue you filed. Thanks!

wantehchang (changes requested):
Thank you very much for reporting the problem to the MINGW maintainers. I suggest some small changes.

CMakeLists.txt
tongyuantongyu: Updated. Again thanks for the advice!

wantehchang: You're welcome! I found that GitHub UI allows me to edit the files directly. This is convenient because I won't need to trouble you for minor issues. I also got to edit the commit message when I did "Squash and commit" from the GitHub UI, although I still repeated the summary line by mistake :-)

PR #451, opened 2020-12-18
Rename the gridX and gridY params of avifImageDump

wantehchang:
The gridX and gridY parameters of avifImageDump() are renamed gridCols
and gridRows, respectively.

### DIFF
--- a/apps/shared/avifutil.h
+++ b/apps/shared/avifutil.h
@@ -20,7 +20,7 @@
 #define AVIF_FMT_ZU "%zu"
 #endif
 
-void avifImageDump(avifImage * avif, uint32_t gridX, uint32_t gridY);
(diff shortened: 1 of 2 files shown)


apps/shared/avifutil.h
```diff
@@ -20,7 +20,7 @@
 #define AVIF_FMT_ZU "%zu"
 #endif
 
-void avifImageDump(avifImage * avif, uint32_t gridX, uint32_t gridY);
+void avifImageDump(avifImage * avif, uint32_t gridCols, uint32_t gridRows);
```
wantehchang, line 23: The caller (avifenc.c) calls avifImageDump() as follows:

`        avifImageDump(avif, 0, 0);`

and

`    avifImageDump(gridCells ? gridCells[0] : image, gridDims[0], gridDims[1]);`

The second call makes it clear that these two parameters are gridCols and gridRows.

PR #453, opened 2020-12-19
Simplify avifEncoderAddImageInternal conditionals

wantehchang:
Simplify some conditionals in avifEncoderAddImageInternal().

Since we require all cellImages have the same width, height, and depth,
move the checks for invalid values of width, height, and depth outside
the for loop and perform those checks on just firstCell.

Check gridCount > 1 instead of whether gridColorID or gridAlphaID is
nonzero. gridCount > 1 is an equivalent condition and is used elsewhere
in the function.

In the single-image case, we only need to check if the alpha planes of
cellImages have fully opaque values if firstCell has an alpha plane. (If
firstCell does not have an alpha plane, then no cellImages have an alpha
plane.)

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -402,22 +402,26 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
     }
 
     const avifImage * firstCell = cellImages[0];
+    if ((firstCell->depth != 8) && (firstCell->depth != 10) && (firstCell->depth != 12)) {
+        return AVIF_RESULT_UNSUPPORTED_DEPTH;
+    }
+
+    if (!firstCell->width || !firstCell->height) {
+        return AVIF_RESULT_NO_CONTENT;
+    }
+
     if ((cellCount > 1) && ((firstCell->width < 64) || (firstCell->height < 64))) {
         return AVIF_RESULT_INVALID_IMAGE_GRID;
     }
 
     for (uint32_t cellIndex = 0; cellIndex < cellCount; ++cellIndex) {
         const avifImage * cellImage = cellImages[cellIndex];
-        if ((cellImage->depth != 8) && (cellImage->depth != 10) && (cellImage->depth != 12)) {
-            return AVIF_RESULT_UNSUPPORTED_DEPTH;
-        }
-
         if ((cellImage->depth != firstCell->depth) || (cellImage->width != firstCell->width) ||
             (cellImage->height != firstCell->height) || (!!cellImage->alphaPlane != !!firstCell->alphaPlane)) {
             return AVIF_RESULT_INVALID_IMAGE_GRID;
         }
 
-        if (!cellImage->width || !cellImage->height || !cellImage->yuvPlanes[AVIF_CHAN_Y]) {
+        if (!cellImage->yuvPlanes[AVIF_CHAN_Y]) {
             return AVIF_RESULT_NO_CONTENT;
         }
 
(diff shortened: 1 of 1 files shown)


src/write.c
```diff
@@ -402,22 +402,26 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
     }
 
     for (uint32_t cellIndex = 0; cellIndex < cellCount; ++cellIndex) {
         const avifImage * cellImage = cellImages[cellIndex];
-        if ((cellImage->depth != 8) && (cellImage->depth != 10) && (cellImage->depth != 12)) {
-            return AVIF_RESULT_UNSUPPORTED_DEPTH;
-        }
-
         if ((cellImage->depth != firstCell->depth) || (cellImage->width != firstCell->width) ||
             (cellImage->height != firstCell->height) || (!!cellImage->alphaPlane != !!firstCell->alphaPlane)) {
```
wantehchang, line 420: The (!!cellImage->alphaPlane != !!firstCell->alphaPlane) check here ensures that either all cellImages have an alpha plane or no cellImages have an alpha plane. This fact will be used below.

src/write.c
```diff
@@ -476,15 +480,15 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
             }
             item->codec->csOptions = encoder->csOptions;
 
-            if (gridColorID) {
+            if (cellCount > 1) {
                 item->dimgFromID = gridColorID;
-            } else if (!encoder->data->primaryItemID) {
+            } else {
                 encoder->data->primaryItemID = item->id;
             }
```
wantehchang, line 487: Please see line 460 (line 464 in new code). cellCount > 1 if and only if gridColorID is nonzero.

It is not necessary to check !encoder->data->primaryItemID. If cellCount == 1 (or equivalently gridColorID == 0), then encoder->data->primaryItemID must be 0, and this loop has only one iteration.

src/write.c
```diff
@@ -476,15 +480,15 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
         }
 
         encoder->data->alphaPresent = (firstCell->alphaPlane != NULL);
-        if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) {
+        if (encoder->data->alphaPresent && (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE)) {
```
wantehchang, line 491: If encoder->data->alphaPresent is false, no cellImages have an alpha plane, so we can skip the for loop at line 502.

src/write.c
```diff
@@ -584,15 +588,10 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
-                if (!cellImage->alphaPlane) {
-                    // If the first image in the sequence had an alpha plane (even if fully opaque), all
-                    // subsequence images must have alpha as well.
-                    return AVIF_RESULT_ENCODE_ALPHA_FAILED;
-                }
-            }
+        if (encoder->data->alphaPresent && !firstCell->alphaPlane) {
+            // If the first image in the sequence had an alpha plane (even if fully opaque), all
+            // subsequence images must have alpha as well.
+            return AVIF_RESULT_ENCODE_ALPHA_FAILED;
```
wantehchang, line 594: It suffices to just test if firstCell has an alpha plane.

Issue #331, opened 2020-09-24
Wish list: Large image encoding (via image grid?)


novomesk:
I observed that some people like very large images (even larger than 100 mega pixels) but encoding big images to AVIF is very memory hungry process.
For example I was able to encode 10000x10000 image on my computer, but with higher dimensions there is lot of swap use or the encoding process runs out memory.

Is the image grid solution for encoding large images? The smaller regions would be not so memory-consuming as entire image.

Maybe the image grid encoding would be another breaking change and maybe not.
Maybe the encoder API can remain mostly unchanged - user pass entire image and libavif create grid automatically when image is above specified size.
Or there will be a new API - user splits image and pass image fragments to libavif one after another.

This is not a top urgent feature now for me but I am writing so it could be considered/planned for future.

joedrago:
Image grid support for avifEncoder is a very reasonable request, but I'm not likely to personally add it for a while. avifDecoder can already decode grid images.

Out of curiosity, how much does enabling the encoder's tiling mechanism help? That functionality is already exposed and available to try.

novomesk:
Encoders create tiles automatically because maximum tile width is 4k and maximum size 9 mega pixel.

Even if you don't request tiles for 21 MP photo from digital camera, it will have 4 tiles at least.

joedrago:
Sure, I'm saying if you turned up the tiling, do you see an improvement in encoding time?

novomesk:
I tried few months ago. No difference when I encoded 21Mp picture using 4 threads because tiles were always 4.
I saw speedup only on smaller images, something like 10-20% relatively, but small images are generally fast.

Uch1haOb1to:
If I understand the work of tiles correctly, then they at least do not reduce memory consumption.
Earlier, before optimizing libaom, I also saw a lack of memory and use of paging, but now the memory is consumed significantly less. But still, the pictures can be such that there is not enough memory.
You may add a parameter that is responsible for the amount of memory available for encoding, and when there is not enough memory for normal encoding, then divide the picture. Also, you may make the encoder set this parameter by default depending on the available memory or depending on the total amount of memory.

0xC0000054:
> If I understand the work of tiles correctly, then they at least do not reduce memory consumption.

Tiles can refer to both the encoder's tiling mechanism and the individual images that make up an image grid.
The encoder's tiling mechanism does not reduce memory usage, creating an image grid can dramatically reduce memory usage.

[...]

PR #454, opened 2020-12-22
avifImageSplitGrid need not allocate cell planes

wantehchang:
avifImageSplitGrid() does not need to call avifImageAllocatePlanes() on
the cell images. It can just point the cell images' planes to the pixel
data of gridSplitImage. This issue was reported by Nicholas Hayes
(0xC0000054) in https://github.com/AOMediaCodec/libavif/issues/331.

Fix a bug in aomCodecEncodeImage() when copying pixel data from
avifImage to aom_image_t. The number of bytes per row should be the
width times bytesPerPixel, not image->yuvRowBytes[yuvPlane].


wantehchang:
I don't know how to resolve the merge conflicts. I will create a new pull request for this change.

PR #455, opened 2020-12-22
Correct grid split place on high-depth image

tongyuantongyu:
`avifImageSplitGrid` split image into grid at wrong place on x-axis on high-depth images.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -310,7 +310,7 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
             const uint32_t bytesPerRowY = bytesPerPixel * cellWidth;
             const uint32_t srcRowBytesY = gridSplitImage->yuvRowBytes[AVIF_CHAN_Y];
             const uint8_t * srcPlaneY =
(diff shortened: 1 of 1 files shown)


wantehchang:
Hi Yuan Tong: Good catch! I will edit your pull request to use existing local variables.

apps/avifenc.c
```diff
@@ -355,7 +355,7 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
 
                 const uint32_t bytesPerRowA = bytesPerPixel * cellWidth;
                 const uint32_t srcRowBytesA = gridSplitImage->alphaRowBytes;
-                const uint8_t * srcPlaneA = &gridSplitImage->alphaPlane[(gridX * cellWidth) + (gridY * cellHeight) * srcRowBytesA];
+                const uint8_t * srcPlaneA = &gridSplitImage->alphaPlane[(gridX * bytesPerRowA) + (gridY * cellHeight) * srcRowBytesA];
```
wantehchang, line 358: While editing this pull request, I found that bytesPerRowA could be replaced by bytesPerRowY because the two variables have the same value. But I did not delete bytesPerRowA.

PR #456, opened 2020-12-22
Expose AOM specific film grain advanced options

tongyuantongyu:
Film grain is a post-decode step in AV1 which adds synthetic noise on the image. 

When encoding an image using 8 bit, limited range (this is the best choice If performance and compatibility are both the case), color banding artifact can usually appear on color gradient areas. Adding noise (dithering) can reduce such artifact, but doing such will significantly increase the file size (random data is hard to compress). Film grain uses several parameters to model the noise, so it solves banding without inflating file size.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -107,8 +107,8 @@ static void syntax(void)
         printf("    end-usage=MODE                    : Rate control mode (vbr, cbr, cq, or q)\n");
         printf("    sharpness=S                       : Loop filter sharpness (0-7, default: 0)\n");
         printf("    tune=METRIC                       : Tune the encoder for distortion metric (psnr or ssim, default: psnr)\n");
-        printf("    film-grain-test=FV                : Film grain test vectors (0: none (default), 1: test-1  2: test-2, ... 16: test-16)\n");
-        printf("    film-grain-table=FT               : Path to file containing film grain parameters\n");
+        printf("    film-grain-test=TEST              : Film grain test vectors (0: none (default), 1: test-1  2: test-2, ... 16: test-16)\n");
+        printf("    film-grain-table=FILENAME         : Path to file containing film grain parameters\n");
         printf("\n");
     }
     avifPrintVersions();
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -107,6 +107,8 @@ static void syntax(void)
         printf("    end-usage=MODE                    : Rate control mode (vbr, cbr, cq, or q)\n");
         printf("    sharpness=S                       : Loop filter sharpness (0-7, default: 0)\n");
         printf("    tune=METRIC                       : Tune the encoder for distortion metric (psnr or ssim, default: psnr)\n");
+        printf("    film-grain-test=FV                : Film grain test vectors (0: none (default), 1: test-1  2: test-2, ... 16: test-16)\n");
+        printf("    film-grain-table=FT               : Path to file containing film grain parameters\n");
         printf("\n");
     }
     avifPrintVersions();
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -259,7 +259,7 @@ static avifBool aomOptionParseUInt(const char * str, unsigned int * val)
     const unsigned long rawval = strtoul(str, &endptr, 10);
 
     if (str[0] != '\0' && endptr[0] == '\0' && rawval <= UINT_MAX) {
-        *val = (int)rawval;
+        *val = (unsigned int)rawval;
         return AVIF_TRUE;
     }
 
@@ -326,18 +326,18 @@ static avifBool avifProcessAOMOptionsPreInit(avifCodec * codec, struct aom_codec
 typedef enum
 {
     AVIF_AOM_NUL_OPTION = 0,
-
     AVIF_AOM_STR_OPTION,
     AVIF_AOM_INT_OPTION,
     AVIF_AOM_UINT_OPTION,
     AVIF_AOM_ENUM_OPTION,
-} avifAOMOptionType;
+} aomOptionType;
 
 struct aomOptionDef
 {
     const char * name;
     int controlId;
-    avifAOMOptionType type;
+    aomOptionType type;
+    // if type is AVIF_AOM_ENUM_OPTION, this must be set. Otherwise should be NULL
     const struct aomOptionEnumList * enums;
 };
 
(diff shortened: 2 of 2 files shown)


apps/avifenc.c
```diff
@@ -107,6 +107,8 @@ static void syntax(void)
         printf("    end-usage=MODE                    : Rate control mode (vbr, cbr, cq, or q)\n");
         printf("    sharpness=S                       : Loop filter sharpness (0-7, default: 0)\n");
         printf("    tune=METRIC                       : Tune the encoder for distortion metric (psnr or ssim, default: psnr)\n");
+        printf("    film-grain-test=FV                : Film grain test vectors (0: none (default), 1: test-1  2: test-2, ... 16: test-16)\n");
+        printf("    film-grain-table=FT               : Path to file containing film grain parameters\n");
```
wantehchang, line 111: Let's name this argument "FILENAME", which is already used for several other (non-codec-specific) options.

apps/avifenc.c
```diff
@@ -107,6 +107,8 @@ static void syntax(void)
         printf("    end-usage=MODE                    : Rate control mode (vbr, cbr, cq, or q)\n");
         printf("    sharpness=S                       : Loop filter sharpness (0-7, default: 0)\n");
         printf("    tune=METRIC                       : Tune the encoder for distortion metric (psnr or ssim, default: psnr)\n");
+        printf("    film-grain-test=FV                : Film grain test vectors (0: none (default), 1: test-1  2: test-2, ... 16: test-16)\n");
```
wantehchang, line 110: Nit: Maybe name this argument "TEST"?

src/codec_aom.c
```diff
@@ -253,6 +253,19 @@ static avifBool aomOptionParseInt(const char * str, int * val)
     return AVIF_FALSE;
 }
 
+static avifBool aomOptionParseUInt(const char * str, unsigned int * val)
+{
+    char * endptr;
+    const unsigned long rawval = strtoul(str, &endptr, 10);
+
+    if (str[0] != '\0' && endptr[0] == '\0' && rawval <= UINT_MAX) {
+        *val = (int)rawval;
```
wantehchang, line 262: Change the cast to (unsigned int)

src/codec_aom.c
```diff
@@ -310,10 +323,21 @@ static avifBool avifProcessAOMOptionsPreInit(avifCodec * codec, struct aom_codec
...
+    AVIF_AOM_INT_OPTION,
+    AVIF_AOM_UINT_OPTION,
+    AVIF_AOM_ENUM_OPTION,
+} avifAOMOptionType;
+
 struct aomOptionDef
 {
     const char * name;
     int controlId;
+    avifAOMOptionType type;
```
wantehchang, line 340: Please document that if 'type' is AOM_OPTION_ENUM, then 'enums' must be specified.

src/codec_aom.c
```diff
@@ -345,12 +380,26 @@ static avifBool avifProcessAOMOptionsPostInit(avifCodec * codec)
...
             if (!strcmp(entry->key, aomOptionDefs[j].name)) {
                 match = AVIF_TRUE;
-                int val;
-                avifBool parsed;
-                if (aomOptionDefs[j].enums) {
-                    parsed = aomOptionParseEnum(entry->value, aomOptionDefs[j].enums, &val);
-                } else {
-                    parsed = aomOptionParseInt(entry->value, &val);
+                // store either an int, an unsigned int or a char *.
+                void * val = NULL;
```
wantehchang, line 384: This is not type-safe, because 'val' will be passed to aom_codec_control() at line 407 below, which expects its third argument to be of a specific type depending on the second argument.

I suggest declaring 'val' to be a local variable of the correct type in each of the cases of the switch statement, and duplicate the aom_codec_control() call in all the cases.

src/codec_aom.c
```diff
@@ -554,6 +603,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     if (alpha) {
         aomImage->range = (image->alphaRange == AVIF_RANGE_FULL) ? AOM_CR_FULL_RANGE : AOM_CR_STUDIO_RANGE;
         aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_RANGE, aomImage->range);
+        // film grain should not be applied on alpha plane
+        aom_codec_control(&codec->internal->encoder, AV1E_SET_FILM_GRAIN_TABLE, NULL);
+        aom_codec_control(&codec->internal->encoder, AV1E_SET_FILM_GRAIN_TEST_VECTOR, 0);
```
wantehchang, line 608: I noticed this issue before. The codec-specific options may need to be separated into a set for "color" and a set for "alpha".

src/codec_aom.c
```diff
@@ -554,6 +603,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     if (alpha) {
         aomImage->range = (image->alphaRange == AVIF_RANGE_FULL) ? AOM_CR_FULL_RANGE : AOM_CR_STUDIO_RANGE;
         aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_RANGE, aomImage->range);
+        // film grain should not be applied on alpha plane
```
wantehchang, line 606: Nit: Change "applied on" to "applied to".  Add "the" before "alpha plane".

wantehchang (changes requested):
Hi Yuan Tong,

I suggest some changes. Note that I may not be able to review new versions of this pull request until next year.

Happy New Year!

wantehchang:
In the commit message, you wrote:

> When encoding an image using 8 bit, limited range (this is the best choice If performance and compatibility are both the case), ...

I think it is better to use full range in image encoding. If "limited range" is not a typo, could you explain why it is better to use limited range in image encoding?

andrey-norkin:
> In the commit message, you wrote:
> 
> > When encoding an image using 8 bit, limited range (this is the best choice If performance and compatibility are both the case), ...
> 
> I think it is better to use full range in image encoding. If "limited range" is not a typo, could you explain why it is better to use limited range in image encoding?

Both limited range and full range would work for film grain.

tongyuantongyu:
> In the commit message, you wrote:
> 
> > When encoding an image using 8 bit, limited range (this is the best choice If performance and compatibility are both the case), ...
> 
> I think it is better to use full range in image encoding. If "limited range" is not a typo, could you explain why it is better to use limited range in image encoding?

That's for compatibility. Windows supports preview AVIF (and open in Paint) by install its AV1 Video extension in Microsoft Store from late 2018, but it treats full range images as limited range. Same as Firefox, the bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1654461) has been there for 5 months. And there's likely to be more. If everyone uses Chrome and libavif then I'd rather never bother with limited range, but if we want the image display correctly everywhere then we have to use limited range.

wantehchang (approved):
Hi Yuan Tong,

Happy New Year! I edited the pull request slightly. I will commit the pull request when it passes the CI tests.

wantehchang:
> > I think it is better to use full range in image encoding. If "limited range" is not a typo, could you explain why it is better to use limited range in image encoding?
> 
> That's for compatibility. Windows supports preview AVIF (and open in Paint) by install its AV1 Video extension in Microsoft Store from late 2018, but it treats full range images as limited range. Same as Firefox, the bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1654461) has been there for 5 months.

Thank you. I didn't know that. This is unfortunate. We should try to contact the right people and have these bugs fixed. (I don't know how to file a bug report on Windows 10's AV1 Video extension.)

Firefox seems to use libyuv for the YUV-to-RGB conversion. If that is the case, it should be straightforward to fix the Firefox bug for the BT.601 full range case because libyuv has functions (the "J" functions) for that case.

wantehchang:
The latest version of Microsoft Edge supports AVIF if the AV1 Video extension is installed. Have you tested that? Does Microsoft Edge also have the bug with full range?

tongyuantongyu:
> Thank you. I didn't know that. This is unfortunate. We should try to contact the right people and have these bugs fixed. (I don't know how to file a bug report on Windows 10's AV1 Video extension.)

I doubt if there is even one...

> Firefox seems to use libyuv for the YUV-to-RGB conversion. If that is the case, it should be straightforward to fix the Firefox bug for the BT.601 full range case because libyuv has functions (the "J" functions) for that case.

I created this issue in libyuv: https://bugs.chromium.org/p/libyuv/issues/detail?id=877 and after it implemented firefox should be able to handle limited and full range in MC=0, 1, 4, 5, 6, 7, 8, 9, 11, 12.
By the way, libavif's support for limited range YCgCo(8) and SMPTE2085(11) is also ready (https://github.com/tongyuantongyu/libavif/tree/more_matrix). I'll open a pr after #459 is merged. Chrome seems to have some issues on limited range YCgCo (https://bugs.chromium.org/p/chromium/issues/detail?id=1162784) though.

> The latest version of Microsoft Edge supports AVIF if the AV1 Video extension is installed. Have you tested that? Does Microsoft Edge also have the bug with full range?

Not sure why, but Microsoft Edge doesn't even support avif now: https://caniuse.com/avif
Edge will just download avif on my computer even with AV1 Video extension installed, and this is the request header for image:
```
accept: image/webp,image/apng,image/*,*/*;q=0.8
accept-encoding: gzip, deflate, br
sec-fetch-dest: image
sec-fetch-mode: no-cors
sec-fetch-site: cross-site
user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.88 Safari/537.36 Edg/87.0.664.66
```

wantehchang:
Hi Yuan Tong,

I believe Chrome's bug with limited range YCgCo has the same underlying reason as libavif's bug with limited range YCgCo: YCgCo uses different equations from the other color spaces. (I chose to make libavif fail with an error, rather than produce incorrect output, when limited range YCgCo is used.)

Thank you for testing Microsoft Edge. Your testing result made me remember that Microsoft Edge supports AV1 but not AVIF when the AV1 Video extension is installed. Sorry to have wasted your time.

By the way, I just realized that I have gone too far when I renamed AVIF_AOM_xxx_OPTION to AOM_OPTION_xxx -- those enum constants now look like enum constants defined by libaom. I will add the AVIF_ prefix back (but keep the _xxx part at the end).

PR #458, opened 2020-12-22
Declare gridCols and gridRows as uint32_t

wantehchang:
Declare gridCols and gridRows as uint32_t (instead of uint8_t) so that
256 (the maximum legal value) can be represented.

Also make some cleanup changes.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -939,7 +939,7 @@ int main(int argc, char * argv[])
         }
 
         if (gridCellIndex == 0) {
-            printf("Single image input for a grid image. Attempting split into %u cells...\n", gridCellCount);
+            printf("Single image input for a grid image. Attempting to split into %u cells...\n", gridCellCount);
             gridSplitImage = image;
             gridCells[0] = NULL;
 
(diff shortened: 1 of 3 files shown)


apps/avifenc.c
```diff
@@ -950,12 +950,10 @@ int main(int argc, char * argv[])
             gridCellIndex = gridCellCount - 1;
         }
 
-        for (uint32_t i = 0; i < gridCellCount; ++i) {
-            if (!gridCells[i]) {
-                fprintf(stderr, "ERROR: Not enough input files for grid image! (expecting %u, or a single image to be split)\n", gridCellCount);
-                returnCode = 1;
-                goto cleanup;
-            }
+        if (gridCellIndex != gridCellCount - 1) {
```
wantehchang, line 953: By construction, gridCells[i] != NULL for i = 0 .. gridCellIndex. So if gridCellIndex != gridCellCount - 1, at least one gridCells[i] is NULL.

apps/avifenc.c
```diff
@@ -1016,9 +1014,7 @@ int main(int argc, char * argv[])
             returnCode = 1;
             goto cleanup;
         }
-    }
 
-    if (!gridDimsCount) {
```
wantehchang, line 1021: This if block can be merged with the else block above because they have the same condition.

PR #459, opened 2020-12-23
Use procedure specified in H.273 to quantize YUV

tongyuantongyu:
Old `avifReformatStateYToUNorm` and `avifReformatStateUVToUNorm` functions quantize YUV value at very first without extra precision, and can introduce severer than normal banding. This PR change to follow the procedure specified in H.273, and can give better result on limited range images.

Difference example:
| Original | Old | New |
| --- | --- | --- |
| original | old | new |

Images are created by following command:
```
avifenc -r l --min 0 --max 0 --cicp 1/13/1 original.png output.avif
avifdec output.avif test.png
```

Updated formulas passed avifyuv test with maxDrift=1 (only tested 8bit and 10bit. 12bit is too slow).

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -520,16 +520,14 @@ typedef struct avifReformatState
     uint32_t rgbOffsetBytesA;
 
     uint32_t yuvDepth;
-    uint32_t rgbDepth;
     avifRange yuvRange;
     int yuvMaxChannel;
     int rgbMaxChannel;
-    float yuvMaxChannelF;
     float rgbMaxChannelF;
-    float yBias; // minimum y value
-    float uvBias; // the value of 0.5 for the appropriate bit depth [128, 512, 2048]
-    float yRange; // difference between max and min y
-    float uvRange; // difference between max and min uv
+    float biasY; // minimum y value
+    float biasUV; // the value of 0.5 for the appropriate bit depth [128, 512, 2048]
+    float rangeY; // difference between max and min y
+    float rangeUV; // difference between max and min uv
 
     avifPixelFormatInfo formatInfo;
 
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -526,7 +526,10 @@ typedef struct avifReformatState
     int rgbMaxChannel;
     float yuvMaxChannelF;
     float rgbMaxChannelF;
-    int uvBias; // the integer value of 0.5 for the appropriate bit depth [128, 512, 2048]
+    float yBias; // minimal y value
+    float uvBias; // the value of 0.5 for the appropriate bit depth [128, 512, 2048]
+    float yRange; // difference between max and min y
+    float uvRange; // difference between max and min uv
 
     avifPixelFormatInfo formatInfo;
 
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -526,7 +526,7 @@ typedef struct avifReformatState
     int rgbMaxChannel;
     float yuvMaxChannelF;
     float rgbMaxChannelF;
-    float yBias; // minimal y value
+    float yBias; // minimum y value
     float uvBias; // the value of 0.5 for the appropriate bit depth [128, 512, 2048]
     float yRange; // difference between max and min y
     float uvRange; // difference between max and min uv
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -109,7 +109,10 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * 
     state->rgbMaxChannel = (1 << rgb->depth) - 1;
     state->yuvMaxChannelF = (float)state->yuvMaxChannel;
     state->rgbMaxChannelF = (float)state->rgbMaxChannel;
-    state->uvBias = 1 << (image->depth - 1);
+    state->yBias = (float)(state->yuvRange == AVIF_RANGE_LIMITED ? (16 << (state->yuvDepth - 8)) : 0);
+    state->uvBias = (float)(1 << (state->yuvDepth - 1));
+    state->yRange = state->yuvRange == AVIF_RANGE_LIMITED ? ((float)(235 << (state->yuvDepth - 8)) - state->yBias) : state->yuvMaxChannelF;
+    state->uvRange = state->yuvRange == AVIF_RANGE_LIMITED ? ((float)(240 << (state->yuvDepth - 8)) - state->yBias) : state->yuvMaxChannelF;
 
     uint32_t cpCount = 1 << image->depth;
     for (uint32_t cp = 0; cp < cpCount; ++cp) {
@@ -129,30 +132,35 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * 
     return AVIF_TRUE;
 }
 
+// Formulas 20-31 from https://www.itu.int/rec/T-REC-H.273-201612-I/en
 static int avifReformatStateYToUNorm(avifReformatState * state, float v)
 {
-    int unorm = (int)avifRoundf(v * state->yuvMaxChannelF);
+    int unorm = (int)avifRoundf(v * state->yRange + state->yBias);
     unorm = AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
-    if (state->yuvRange == AVIF_RANGE_LIMITED) {
-        unorm = avifFullToLimitedY(state->yuvDepth, unorm);
-    }
     return unorm;
 }
 
 static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
 {
-    int unorm = (int)avifRoundf(v * state->yuvMaxChannelF);
-    if (state->mode != AVIF_REFORMAT_MODE_IDENTITY) { // Don't adjust U or V when data is raw RGB
-        unorm += state->uvBias;
+    int unorm = 0;
+
+    // For YCgCo, this is only correct for full range
+    if (state->mode == AVIF_REFORMAT_MODE_YCGCO) {
+        assert(state->yuvRange == AVIF_RANGE_FULL);
     }
-    unorm = AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
-    if (state->yuvRange == AVIF_RANGE_LIMITED) {
-        if (state->mode == AVIF_REFORMAT_MODE_IDENTITY) { // use Y range for all channels when data is raw RGB
-            unorm = avifFullToLimitedY(state->yuvDepth, unorm);
-        } else {
-            unorm = avifFullToLimitedUV(state->yuvDepth, unorm);
-        }
+
+    switch (state->mode) {
+        case AVIF_REFORMAT_MODE_YCGCO:
+        case AVIF_REFORMAT_MODE_YUV_COEFFICIENTS:
+            unorm = (int)avifRoundf(v * state->uvRange + state->uvBias);
+            unorm = AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
+            break;
+        case AVIF_REFORMAT_MODE_IDENTITY:
+            unorm = (int)avifRoundf(v * state->yRange + state->yBias);
+            unorm = AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
+            break;
     }
+
     return unorm;
 }
 
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -133,13 +133,12 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * 
 static int avifReformatStateYToUNorm(avifReformatState * state, float v)
 {
     int unorm = (int)avifRoundf(v * state->rangeY + state->biasY);
-    unorm = AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
-    return unorm;
+    return AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
 }
 
 static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
 {
-    int unorm = 0;
+    int unorm;
 
     // YCgCo performs limited-full range adjustment on R,G,B but the current implementation performs range adjustment
     // on Y,U,V. So YCgCo with limited range is unsupported.
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -118,6 +118,7 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * 
     for (uint32_t cp = 0; cp < cpCount; ++cp) {
         int unorm = cp;
         switch (state->mode) {
+            // review this when implementing YCgCo limited range support.
             case AVIF_REFORMAT_MODE_YCGCO:
             case AVIF_REFORMAT_MODE_YUV_COEFFICIENTS:
                 state->unormFloatTableY[cp] = ((float)unorm - state->yBias) / state->yRange;
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -109,24 +109,24 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * 
     state->rgbMaxChannel = (1 << rgb->depth) - 1;
     state->yuvMaxChannelF = (float)state->yuvMaxChannel;
     state->rgbMaxChannelF = (float)state->rgbMaxChannel;
-    state->yBias = (float)(state->yuvRange == AVIF_RANGE_LIMITED ? (16 << (state->yuvDepth - 8)) : 0);
+    state->yBias = state->yuvRange == AVIF_RANGE_LIMITED ? (float)(16 << (state->yuvDepth - 8)) : 0.f;
     state->uvBias = (float)(1 << (state->yuvDepth - 1));
-    state->yRange = state->yuvRange == AVIF_RANGE_LIMITED ? ((float)(235 << (state->yuvDepth - 8)) - state->yBias) : state->yuvMaxChannelF;
-    state->uvRange = state->yuvRange == AVIF_RANGE_LIMITED ? ((float)(240 << (state->yuvDepth - 8)) - state->yBias) : state->yuvMaxChannelF;
+    state->yRange = state->yuvRange == AVIF_RANGE_LIMITED ? (float)(219 << (state->yuvDepth - 8)) : state->yuvMaxChannelF;
+    state->uvRange = state->yuvRange == AVIF_RANGE_LIMITED ? (float)(224 << (state->yuvDepth - 8)) : state->yuvMaxChannelF;
 
     uint32_t cpCount = 1 << image->depth;
     for (uint32_t cp = 0; cp < cpCount; ++cp) {
-        int unormY = cp;
-        int unormUV = cp;
-        if (image->yuvRange == AVIF_RANGE_LIMITED) {
-            unormY = avifLimitedToFullY(image->depth, unormY);
-            unormUV = avifLimitedToFullUV(image->depth, unormUV);
-        }
-        state->unormFloatTableY[cp] = (float)unormY / state->yuvMaxChannelF;
-        if (state->mode == AVIF_REFORMAT_MODE_IDENTITY) {
-            state->unormFloatTableUV[cp] = (float)unormY / state->yuvMaxChannelF;
-        } else {
-            state->unormFloatTableUV[cp] = ((float)unormUV - state->uvBias) / state->yuvMaxChannelF;
+        int unorm = cp;
+        switch (state->mode) {
+            case AVIF_REFORMAT_MODE_YCGCO:
+            case AVIF_REFORMAT_MODE_YUV_COEFFICIENTS:
+                state->unormFloatTableY[cp] = ((float)unorm - state->yBias) / state->yRange;
+                state->unormFloatTableUV[cp] = ((float)unorm - state->uvBias) / state->uvRange;
+                break;
+            case AVIF_REFORMAT_MODE_IDENTITY:
+                state->unormFloatTableY[cp] = ((float)unorm - state->yBias) / state->yRange;
+                state->unormFloatTableUV[cp] = ((float)unorm - state->yBias) / state->yRange;
+                break;
         }
     }
     return AVIF_TRUE;
@@ -144,7 +144,8 @@ static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
 {
     int unorm = 0;
 
-    // For YCgCo, this is only correct for full range
+    // YCgCo performs limited-full range adjustment on R,G,B but the current implementation performs range adjustment
+    // on Y,U,V. So YCgCo with limited range is unsupported.
     if (state->mode == AVIF_REFORMAT_MODE_YCGCO) {
         assert(state->yuvRange == AVIF_RANGE_FULL);
     }
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -103,40 +103,36 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * 
     }
 
     state->yuvDepth = image->depth;
-    state->rgbDepth = rgb->depth;
     state->yuvRange = image->yuvRange;
     state->yuvMaxChannel = (1 << image->depth) - 1;
     state->rgbMaxChannel = (1 << rgb->depth) - 1;
-    state->yuvMaxChannelF = (float)state->yuvMaxChannel;
     state->rgbMaxChannelF = (float)state->rgbMaxChannel;
-    state->yBias = state->yuvRange == AVIF_RANGE_LIMITED ? (float)(16 << (state->yuvDepth - 8)) : 0.f;
-    state->uvBias = (float)(1 << (state->yuvDepth - 1));
-    state->yRange = state->yuvRange == AVIF_RANGE_LIMITED ? (float)(219 << (state->yuvDepth - 8)) : state->yuvMaxChannelF;
-    state->uvRange = state->yuvRange == AVIF_RANGE_LIMITED ? (float)(224 << (state->yuvDepth - 8)) : state->yuvMaxChannelF;
+    state->biasY = (state->yuvRange == AVIF_RANGE_LIMITED) ? (float)(16 << (state->yuvDepth - 8)) : 0.0f;
+    state->biasUV = (float)(1 << (state->yuvDepth - 1));
+    state->rangeY = (float)((state->yuvRange == AVIF_RANGE_LIMITED) ? (219 << (state->yuvDepth - 8)) : state->yuvMaxChannel);
+    state->rangeUV = (float)((state->yuvRange == AVIF_RANGE_LIMITED) ? (224 << (state->yuvDepth - 8)) : state->yuvMaxChannel);
 
     uint32_t cpCount = 1 << image->depth;
-    for (uint32_t cp = 0; cp < cpCount; ++cp) {
-        int unorm = cp;
-        switch (state->mode) {
-            // review this when implementing YCgCo limited range support.
-            case AVIF_REFORMAT_MODE_YCGCO:
-            case AVIF_REFORMAT_MODE_YUV_COEFFICIENTS:
-                state->unormFloatTableY[cp] = ((float)unorm - state->yBias) / state->yRange;
-                state->unormFloatTableUV[cp] = ((float)unorm - state->uvBias) / state->uvRange;
-                break;
-            case AVIF_REFORMAT_MODE_IDENTITY:
-                state->unormFloatTableY[cp] = ((float)unorm - state->yBias) / state->yRange;
-                state->unormFloatTableUV[cp] = ((float)unorm - state->yBias) / state->yRange;
-                break;
+    if (state->mode == AVIF_REFORMAT_MODE_IDENTITY) {
+        for (uint32_t cp = 0; cp < cpCount; ++cp) {
+            state->unormFloatTableY[cp] = ((float)cp - state->biasY) / state->rangeY;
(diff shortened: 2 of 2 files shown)


src/reformat.c
```diff
@@ -109,7 +109,10 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
     state->rgbMaxChannel = (1 << rgb->depth) - 1;
     state->yuvMaxChannelF = (float)state->yuvMaxChannel;
     state->rgbMaxChannelF = (float)state->rgbMaxChannel;
-    state->uvBias = 1 << (image->depth - 1);
+    state->yBias = (float)(state->yuvRange == AVIF_RANGE_LIMITED ? (16 << (state->yuvDepth - 8)) : 0);
+    state->uvBias = (float)(1 << (state->yuvDepth - 1));
+    state->yRange = state->yuvRange == AVIF_RANGE_LIMITED ? ((float)(235 << (state->yuvDepth - 8)) - state->yBias) : state->yuvMaxChannelF;
+    state->uvRange = state->yuvRange == AVIF_RANGE_LIMITED ? ((float)(240 << (state->yuvDepth - 8)) - state->yBias) : state->yuvMaxChannelF;
```
wantehchang, line 115: Lines 114-115 are equivalent to:

```
    state->yRange = state->yuvRange == AVIF_RANGE_LIMITED ? (float)(219 << (state->yuvDepth - 8)) : state->yuvMaxChannelF;
    state->uvRange = state->yuvRange == AVIF_RANGE_LIMITED ? (float)(224 << (state->yuvDepth - 8)) : state->yuvMaxChannelF;
```

This form is more similar to Equations 23-25 in H.273.

src/reformat.c
```diff
@@ -129,30 +132,35 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
+    // For YCgCo, this is only correct for full range
+    if (state->mode == AVIF_REFORMAT_MODE_YCGCO) {
+        assert(state->yuvRange == AVIF_RANGE_FULL);
     }
-    unorm = AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
-    if (state->yuvRange == AVIF_RANGE_LIMITED) {
-        if (state->mode == AVIF_REFORMAT_MODE_IDENTITY) { // use Y range for all channels when data is raw RGB
-            unorm = avifFullToLimitedY(state->yuvDepth, unorm);
-        } else {
-            unorm = avifFullToLimitedUV(state->yuvDepth, unorm);
```
wantehchang, line 153: We should also fix avifFullToLimitedY() and avifFullToLimitedUV(). These two functions are implemented as follows:

```
#define FULL_TO_LIMITED(MINLIMITEDY, MAXLIMITEDY, FULLY)                           \
    v = (((v * (MAXLIMITEDY - MINLIMITEDY)) + (FULLY / 2)) / FULLY) + MINLIMITEDY; \
    v = AVIF_CLAMP(v, MINLIMITEDY, MAXLIMITEDY)

int avifFullToLimitedY(int depth, int v)
{
    switch (depth) {
        case 8:
            FULL_TO_LIMITED(16, 235, 255);
            break;
        case 10:
            FULL_TO_LIMITED(64, 940, 1023);
            break;
        case 12:
            FULL_TO_LIMITED(256, 3760, 4095);
            break;
    }
    return v;
}

int avifFullToLimitedUV(int depth, int v)
{
    switch (depth) {
        case 8:
            FULL_TO_LIMITED(16, 240, 255);
            break;
        case 10:
            FULL_TO_LIMITED(64, 960, 1023);
            break;
        case 12:
            FULL_TO_LIMITED(256, 3840, 4095);
            break;
    }
    return v;
}
```

Note that the FULLY arguments are all odd numbers. Some numerical inaccuracy may come from the integer division (FULLY / 2) in the definition of the FULL_TO_LIMITED macro. Could you try this definition?

```
#define FULL_TO_LIMITED(MINLIMITEDY, MAXLIMITEDY, FULLY)                           \
    v = (((v * (MAXLIMITEDY - MINLIMITEDY) * 2) + FULLY) / (FULLY * 2)) + MINLIMITEDY; \
    v = AVIF_CLAMP(v, MINLIMITEDY, MAXLIMITEDY)
```

src/reformat.c
```diff
@@ -129,30 +132,35 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
-        }
+
+    switch (state->mode) {
+        case AVIF_REFORMAT_MODE_YCGCO:
```
wantehchang, line 153: AVIF_REFORMAT_MODE_YCGCO and AVIF_REFORMAT_MODE_IDENTITY should be handled in the same way. Move this after line 158.

```
        case AVIF_REFORMAT_MODE_IDENTITY:
        case AVIF_REFORMAT_MODE_YCGCO:
            unorm = (int)avifRoundf(v * state->yRange + state->yBias);
            unorm = AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
            break;
```

include/avif/avif.h
```diff
@@ -526,7 +526,10 @@ typedef struct avifReformatState
     int rgbMaxChannel;
     float yuvMaxChannelF;
     float rgbMaxChannelF;
-    int uvBias; // the integer value of 0.5 for the appropriate bit depth [128, 512, 2048]
+    float yBias; // minimal y value
```
wantehchang, line 529: Nit: minimal => minimum

src/reformat.c
```diff
@@ -129,30 +132,35 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
 }
 
 static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
 {
-    int unorm = (int)avifRoundf(v * state->yuvMaxChannelF);
-    if (state->mode != AVIF_REFORMAT_MODE_IDENTITY) { // Don't adjust U or V when data is raw RGB
-        unorm += state->uvBias;
+    int unorm = 0;
+
+    // For YCgCo, this is only correct for full range
```
wantehchang, line 147: I suggest copying the comment at lines 28-29.

include/avif/avif.h
```diff
@@ -526,7 +526,10 @@ typedef struct avifReformatState
     int rgbMaxChannel;
     float yuvMaxChannelF;
```
wantehchang, line 527: Could you find out if we still need the yuvMaxChannelF field in this struct?

wantehchang (changes requested):
Hi Yuan Tong,

I think I understand the issue. I suggest some changes. Also, we should let Joe Drago take a look, so the review of this pull request may not be finished until next year.

src/reformat.c
wantehchang: Do you know why the avifyuv test still passes with the AVIF_REFORMAT_MODE_YCGCO case in the wrong place? I wonder if I misunderstood something. Thanks!

tongyuantongyu: AVIF_REFORMAT_MODE_YCGCO is not following H.273's procedure, it's just coincidentally we can reuse AVIF_REFORMAT_MODE_YUV_COEFFICIENTS's route to produce mostly same result. Note that current implementation of `avifReformatStateUVToUNorm` also treats AVIF_REFORMAT_MODE_YCGCO the same as AVIF_REFORMAT_MODE_YUV_COEFFICIENTS.

Suppose range=AVIF_RANGE_FULL and mc=AVIF_MATRIX_COEFFICIENTS_YCGCO and depth=8, if strictly follows H.273 we should use Equations 26-28 and 44-46, that is:
| Representation / Operation |  Range |
| --- | --- |
| {E′R,E′G,E′B} | [0., 1.] |
| * 255 | |
| (Not denoted as variable in H.273) | [0., 255.] |
| Clip1Y | |
| {R, G, B} | [0., 255.] |
| Kr-Kb like transformation with Kr = 0.25, Kb = 0.25 | |
| {Y'1, Cg'1, Co'1} (Not denoted as variable in H.273) | Y'1: [0., 255.], Cg'1: [-127.5, 127.5], Co'1: [-127.5, 127.5] |
| Round (rounding halfway cases away from zero) | |
| {Y'2, Cg'2, Co'2} (Not denoted as variable in H.273) | Y'1: [0, 255], Cg'1: [-128, 128], Co'1: [-128, 128] |
| Add bias 128 to Cg'2 and Co'2 | |
| {Y, Cb, Cr} | Y: [0, 255], Cb: [0, 256], Cr: [0, 256] |

libavif's current implementation uses the following procedure:
| Representation / Operation |  Range |
| --- | --- |
| {E′R,E′G,E′B} | [0., 1.] |
| Kr-Kb like transformation with Kr = 0.25, Kb = 0.25 | |
| {Y'3, Cg'3, Co'3} | Y'3: [0., 1.], Cg'3: [-0.5, 0.5], Co'3: [-0.5, 0.5] |
| * 255 (operations from here done in `avifReformatStateUVToUNorm`) | |
| {Y'1, Cb'1, Cr'1} | Y'1: [0., 255.], Cb'1: [-127.5, 127.5], Cr'1: [-127.5, 127.5] |
| Round (rounding halfway cases up, see `avifRoundf`) | |
| {Y'2, Cb'2, Cr'2} | Y'2: [0, 255], Cb'2: [-127, 128], Cr'2: [-127, 128] |
| Add bias 128 to Cb'2 and Cr'2 | |
| {Y'3, Cb'3, Cr'3} | Y'3: [0, 255], Cb'3: [1, 256], Cr'3: [1, 256] |
| Clip1Y | |
| {Y, Cb, Cr} | Y: [0, 255], Cb: [1, 255], Cr: [1, 255] |

This commit changes to this procedure:
| Representation / Operation |  Range |
| --- | --- |
| {E′R,E′G,E′B} | [0., 1.] |
| Kr-Kb like transformation with Kr = 0.25, Kb = 0.25 | |
| {Y'3, Cg'3, Co'3} | Y'3: [0., 1.], Cg'3: [-0.5, 0.5], Co'3: [-0.5, 0.5] |
| * 255 (operations from here done in `avifReformatStateUVToUNorm`) | |
| {Y'4, Cg'4, Co'4} | Y'4: [0., 255.], Cg'4: [-127.5, 127.5], Co'4: [-127.5, 127.5] |
| Add bias 128 to Cg'4 and Co'4 | |
| {Y'4, Cb'4, Cr'4} | Y'4: [0., 255.], Cb'4: [0.5, 255.5], Cr'4: [0.5, 255.5] |
| Round (rounding halfway cases up) | |
| {Y'5, Cb'5, Cr'5} | Y'5: [0, 255], Cb'5: [1, 256], Cr'5: [1, 256] |
| Clip1Y | |
| {Y, Cb, Cr} | Y: [0, 255], Cb: [1, 255], Cr: [1, 255] |

You can see that these all do the same things but in different orders.

include/avif/avif.h
tongyuantongyu: It is used as an intermediate value to initialize `yRange`, `uvRange`, `unormFloatTableY` and `unormFloatTableUV`. Maybe we can make it a local variable of `avifPrepareReformatState`.

tongyuantongyu: After update only initialization of `yRange` and `uvRange` uses `yuvMaxChannelF`. But maybe I'll left it here for completeness. By the way, `rgbDepth` is not used anywhere.

tongyuantongyu:
avifyuv test result on 8bit limited range (apply following patch to make intermediate yuv image use limited range):
```diff
diff --git a/tests/avifyuv.c b/tests/avifyuv.c
index 580a955..a046934 100644
--- a/tests/avifyuv.c
+++ b/tests/avifyuv.c
@@ -123,7 +123,7 @@ int main(int argc, char * argv[])
                     image->colorPrimaries = AVIF_COLOR_PRIMARIES_BT709;
                     image->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_SRGB;
                     image->matrixCoefficients = matrixCoeffs;
-                    image->yuvRange = AVIF_RANGE_FULL;
+                    image->yuvRange = AVIF_RANGE_LIMITED;
                     avifImageAllocatePlanes(image, AVIF_PLANES_YUV);

                     avifRGBImage srcRGB;
```

Current master:
```
$ avifyuv -m drift -v
avif version: 0.8.4
[ 256/ 256] RGB depth: 8, YUV depth: 8, matrixCoeffs: 1
 * RGB depth: 8, YUV depth: 8, matrixCoeffs: 1, maxDrift:  4
   * drift:  0 ->      2748713 /     16777216 pixels (16.38 %)
   * drift:  1 ->     10947363 /     16777216 pixels (65.25 %)
   * drift:  2 ->      2663782 /     16777216 pixels (15.88 %)
   * drift:  3 ->       396523 /     16777216 pixels (2.36 %)
   * drift:  4 ->        20835 /     16777216 pixels (0.12 %)
[ 256/ 256] RGB depth: 8, YUV depth: 8, matrixCoeffs: 6
 * RGB depth: 8, YUV depth: 8, matrixCoeffs: 6, maxDrift:  4
   * drift:  0 ->      2654731 /     16777216 pixels (15.82 %)
   * drift:  1 ->     11085666 /     16777216 pixels (66.08 %)
   * drift:  2 ->      2667021 /     16777216 pixels (15.90 %)
   * drift:  3 ->       352453 /     16777216 pixels (2.10 %)
   * drift:  4 ->        17345 /     16777216 pixels (0.10 %)
[ 256/ 256] RGB depth: 8, YUV depth: 8, matrixCoeffs: 9
 * RGB depth: 8, YUV depth: 8, matrixCoeffs: 9, maxDrift:  4
   * drift:  0 ->      2744794 /     16777216 pixels (16.36 %)
   * drift:  1 ->     10772666 /     16777216 pixels (64.21 %)
   * drift:  2 ->      2679580 /     16777216 pixels (15.97 %)
   * drift:  3 ->       548441 /     16777216 pixels (3.27 %)
   * drift:  4 ->        31735 /     16777216 pixels (0.19 %)
[ 256/ 256] RGB depth: 8, YUV depth: 8, matrixCoeffs: 0
 * RGB depth: 8, YUV depth: 8, matrixCoeffs: 0, maxDrift:  1
   * drift:  0 ->     10648000 /     16777216 pixels (63.47 %)
   * drift:  1 ->      6129216 /     16777216 pixels (36.53 %)
```

This PR:
```
$ avifyuv -m drift -v
avif version: 0.8.4
[ 256/ 256] RGB depth: 8, YUV depth: 8, matrixCoeffs: 1
 * RGB depth: 8, YUV depth: 8, matrixCoeffs: 1, maxDrift:  2
   * drift:  0 ->      2753771 /     16777216 pixels (16.41 %)
   * drift:  1 ->     13893861 /     16777216 pixels (82.81 %)
   * drift:  2 ->       129584 /     16777216 pixels (0.77 %)
[ 256/ 256] RGB depth: 8, YUV depth: 8, matrixCoeffs: 6
 * RGB depth: 8, YUV depth: 8, matrixCoeffs: 6, maxDrift:  2
   * drift:  0 ->      2660527 /     16777216 pixels (15.86 %)
   * drift:  1 ->     14058291 /     16777216 pixels (83.79 %)
   * drift:  2 ->        58398 /     16777216 pixels (0.35 %)
[ 256/ 256] RGB depth: 8, YUV depth: 8, matrixCoeffs: 9
 * RGB depth: 8, YUV depth: 8, matrixCoeffs: 9, maxDrift:  2
   * drift:  0 ->      2749680 /     16777216 pixels (16.39 %)
   * drift:  1 ->     13871113 /     16777216 pixels (82.68 %)
   * drift:  2 ->       156423 /     16777216 pixels (0.93 %)
[ 256/ 256] RGB depth: 8, YUV depth: 8, matrixCoeffs: 0
 * RGB depth: 8, YUV depth: 8, matrixCoeffs: 0, maxDrift:  1
   * drift:  0 ->     10648000 /     16777216 pixels (63.47 %)
   * drift:  1 ->      6129216 /     16777216 pixels (36.53 %)
```

src/reformat.c
tongyuantongyu: I tested and both gives identical (good enough) result. avifFullToLimitedY() and avifFullToLimitedUV() are doing their best, it's just the result is not suitable as an intermediate value.

joedrago:
I'm generally on board with this PR. I'd like to see @wantehchang 's recommendations addressed, but otherwise, if the numbers look better in all cases coming from avifyuv's drift mode, I think it is a good change.

It looks like I need to add a range argument to avifyuv for easier testing?

tongyuantongyu:
> It looks like I need to add a range argument to avifyuv for easier testing?

Yes we should always test drift of limited range as well. I also think maxDrfit=10 is too loose, I would advise it to be 2 or 3.

wantehchang:
Hi Yuan Tong,

I will review this pull request tomorrow (Wednesday) or Thursday.

tongyuantongyu:
@wantehchang  Any update?

wantehchang:
Hi Yuan Tong. I am very sorry for the delay. I am going to review this pull request today.

src/reformat.c
```diff
@@ -109,50 +109,60 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
 static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
 {
-    int unorm = (int)avifRoundf(v * state->yuvMaxChannelF);
-    if (state->mode != AVIF_REFORMAT_MODE_IDENTITY) { // Don't adjust U or V when data is raw RGB
-        unorm += state->uvBias;
+    int unorm = 0;
+
+    // YCgCo performs limited-full range adjustment on R,G,B but the current implementation performs range adjustment
+    // on Y,U,V. So YCgCo with limited range is unsupported.
+    if (state->mode == AVIF_REFORMAT_MODE_YCGCO) {
```
wantehchang, line 150: Nit: Change this if statement to the following:

  assert((state->mode != AVIF_REFORMAT_MODE_YCGCO) || (state->yuvRange == AVIF_RANGE_FULL));

src/reformat.c
```diff
@@ -109,50 +109,60 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
+            case AVIF_REFORMAT_MODE_IDENTITY:
+                state->unormFloatTableY[cp] = ((float)unorm - state->yBias) / state->yRange;
+                state->unormFloatTableUV[cp] = ((float)unorm - state->yBias) / state->yRange;
+                break;
         }
     }
     return AVIF_TRUE;
 }
 
+// Formulas 20-31 from https://www.itu.int/rec/T-REC-H.273-201612-I/en
```
wantehchang, line 136: Nit: Formulas => Equations

src/reformat.c
```diff
@@ -109,50 +109,60 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
-            unorm = avifFullToLimitedY(state->yuvDepth, unorm);
-        } else {
-            unorm = avifFullToLimitedUV(state->yuvDepth, unorm);
-        }
+
+    switch (state->mode) {
+        case AVIF_REFORMAT_MODE_YCGCO:
+        case AVIF_REFORMAT_MODE_YUV_COEFFICIENTS:
+            unorm = (int)avifRoundf(v * state->uvRange + state->uvBias);
+            unorm = AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
```
wantehchang, line 158: Move this line and line 162 outside the switch statement. They are the same.

src/reformat.c
```diff
@@ -109,50 +109,60 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
     }
-    unorm = AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
-    if (state->yuvRange == AVIF_RANGE_LIMITED) {
-        if (state->mode == AVIF_REFORMAT_MODE_IDENTITY) { // use Y range for all channels when data is raw RGB
-            unorm = avifFullToLimitedY(state->yuvDepth, unorm);
-        } else {
-            unorm = avifFullToLimitedUV(state->yuvDepth, unorm);
-        }
+
+    switch (state->mode) {
```
wantehchang, line 154: Nit: Please change the switch statement back to an if statement (as in the original code). I believe the if statement will be compiled to more efficient code in this case.

src/reformat.c
```diff
@@ -109,50 +109,60 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
-            unormY = avifLimitedToFullY(image->depth, unormY);
-            unormUV = avifLimitedToFullUV(image->depth, unormUV);
-        }
-        state->unormFloatTableY[cp] = (float)unormY / state->yuvMaxChannelF;
-        if (state->mode == AVIF_REFORMAT_MODE_IDENTITY) {
-            state->unormFloatTableUV[cp] = (float)unormY / state->yuvMaxChannelF;
-        } else {
-            state->unormFloatTableUV[cp] = ((float)unormUV - state->uvBias) / state->yuvMaxChannelF;
+        int unorm = cp;
+        switch (state->mode) {
```
wantehchang, line 120: Nit: Please change the switch statement back to an if statement (as in the original code). I believe the if statement will be compiled to more efficient code in this case.

In fact, we should put the if statement outside and put the for loop inside the "if" and "else" branches, so that the for loop bodies have no conditionals.

Note: It is instructive to look at the assembly code that the compiler generates for this kind of statement. Although the switch statement exhaustively lists all possible values of the enum type, the compiler will still generate code equivalent to the following:

    if ((state->mode == AVIF_REFORMAT_MODE_YCGCO) || (state->mode == AVIF_REFORMAT_MODE_YUV_COEFFICIENTS)) {
        state->unormFloatTableY[cp] = ((float)unorm - state->yBias) / state->yRange;
        state->unormFloatTableUV[cp] = ((float)unorm - state->uvBias) / state->uvRange;
    } else if (state->mode == AVIF_REFORMAT_MODE_IDENTITY) {
        state->unormFloatTableY[cp] = ((float)unorm - state->yBias) / state->yRange;
        state->unormFloatTableUV[cp] = ((float)unorm - state->yBias) / state->yRange;
    }

I.e., the compiler will transform the last case in the switch statement to an "else if" rather than an "else".

include/avif/avif.h
```diff
@@ -526,7 +526,10 @@ typedef struct avifReformatState
...
     float rgbMaxChannelF;
-    int uvBias; // the integer value of 0.5 for the appropriate bit depth [128, 512, 2048]
+    float yBias; // minimum Y value
+    float uvBias; // the value of 0.5 for the appropriate bit depth [128, 512, 2048]
+    float yRange; // difference between max and min Y
+    float uvRange; // difference between max and min UV
```
wantehchang, line 532: Nit/optional: yRange and uvRange look like the yuvRange member at line 524 above, but they have different meanings. This could be confusing, but it may not be a real problem. We can address this by  changing yRange and uvRange to rangeY and rangeUV. If we do that, it would be good to also change yBias and uvBias to biasY and biasUV for consistency.

include/avif/avif.h
```diff
@@ -526,7 +526,10 @@ typedef struct avifReformatState
     int rgbMaxChannel;
     float yuvMaxChannelF;
     float rgbMaxChannelF;
-    int uvBias; // the integer value of 0.5 for the appropriate bit depth [128, 512, 2048]
+    float yBias; // minimum Y value
```
wantehchang, line 529: I suggest we remove the unused member rgbDepth and change yuvMaxChannelF to a local variable.

wantehchang (changes requested):
Hi Yuan Tong, I suggest some changes.

src/reformat.c
wantehchang: Thank you for the detailed reply. I know our AVIF_REFORMAT_MODE_YCGCO code doe not follow H.273's procedure. In fact, our AVIF_REFORMAT_MODE_IDENTITY code does not follow H.273's procedure, either. Ignoring rounding errors, for AVIF_REFORMAT_MODE_IDENTITY, we will still get the correct results for both full and limited ranges. For AVIF_REFORMAT_MODE_YCGCO, we will only get the correct results for full range. That's why I disallow limited range with AVIF_REFORMAT_MODE_YCGCO.

wantehchang:
> > It looks like I need to add a range argument to avifyuv for easier testing?
> 
> Yes we should always test drift of limited range as well. I also think maxDrfit=10 is too loose, I would advise it to be 2 or 3.

Joe, Yuan Tong: Could one of you file an issue about this proposed change to avifyuv? Thanks.

joedrago:
I believe you misread -- The change is for `avifyuv` (our tool in libavif), not libyuv.

wantehchang:
> I believe you misread -- The change is for `avifyuv` (our tool in libavif), not libyuv.

Hmm... I wrote "Could one of you file an issue about this proposed change to avifyuv?" I don't think libyuv was ever mentioned in the discussions on this pull request...

joedrago:
Ah, I misread what I thought you misread. I'll file a bug.

src/reformat.c
tongyuantongyu: Word "Formulas" was used at line 531 and 536 and I'm following that. Do we want to change these usages as well?

src/reformat.c
tongyuantongyu: I see. By the way I've done an experimental implementation to support all matrix and that needs 6 reformat modes. In that case, should we switch to a switch statement, or keep using if-else chain and sort the most common case first?

tongyuantongyu: I did some tests on this: I set up 6 branches with exponential distributed probability, in this case jumptable based switch and if chain ordered by probability resulted in similar speed. Specially Clang is generating jumptable even for if chain.
Based on the result, I may use switch statement when there's more modes added. Using switch also enables compiler to check if the reformat route is correctly implemented when new matrix coefficients are added in the future.

src/reformat.c
```diff
@@ -535,7 +533,7 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
                     B = Cb;
                     R = Cr;
                 } else if (state->mode == AVIF_REFORMAT_MODE_YCGCO) {
-                    // Identity (GBR): Formulas 47,48,49,50 from https://www.itu.int/rec/T-REC-H.273-201612-I/en
+                    // YCgCo: Formulas 47,48,49,50 from https://www.itu.int/rec/T-REC-H.273-201612-I/en
```
wantehchang, line 536: Thank you for fixing my copy-and-paste mistake!

src/reformat.c
wantehchang: Ah, I didn't know that. No need to change this then. (I suggested "Equations" only because that's the term in H.273.)

src/reformat.c
```diff
@@ -103,57 +103,55 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
     }
+
     return AVIF_TRUE;
 }
 
+// Formulas 20-31 from https://www.itu.int/rec/T-REC-H.273-201612-I/en
 static int avifReformatStateYToUNorm(avifReformatState * state, float v)
 {
-    int unorm = (int)avifRoundf(v * state->yuvMaxChannelF);
+    int unorm = (int)avifRoundf(v * state->rangeY + state->biasY);
```
wantehchang, line 135: I will change the next two lines to
    return AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
to match the return statement in avifReformatStateUVToUNorm().

wantehchang (approved):
LGTM. Thanks a lot!

src/reformat.c
wantehchang: Thank you for the report of your experiments!

I suggested if-else statements in this PR because they are simpler than the equivalent switch statements to handle two options. As soon as we need to handle three or more options, then switch statements are often clearer.

I should correct/clarify my previous comment "I believe the if statement will be compiled to more efficient code in this case."

For a switch statement that lists all values of an enum type exhaustively but does not have a default case, the compiler still treats it as if the switch statement had a default case that does nothing. This is why gcc warns that `unorm` may be used uninitialized in the following version of `avifReformatStateUVToUNorm()`, even though the switch statement assigns a value to `unorm` in all possible cases:

```
static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
{
    int unorm;

    // ...

    switch (state->mode) {
        case AVIF_REFORMAT_MODE_IDENTITY:
            unorm = (int)avifRoundf(v * state->rangeY + state->biasY);
            break;
        case AVIF_REFORMAT_MODE_YUV_COEFFICIENTS:
        case AVIF_REFORMAT_MODE_YCGCO:
            unorm = (int)avifRoundf(v * state->rangeUV + state->biasUV);
            break;
    }

    return AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
}
```

Similarly, gcc warns that control reaches end of non-void function in the following version of version of `avifReformatStateUVToUNorm()`:

```
static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
{
    int unorm;

    // ...

    switch (state->mode) {
        case AVIF_REFORMAT_MODE_IDENTITY:
            unorm = (int)avifRoundf(v * state->rangeY + state->biasY);
            return AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
        case AVIF_REFORMAT_MODE_YUV_COEFFICIENTS:
        case AVIF_REFORMAT_MODE_YCGCO:
            unorm = (int)avifRoundf(v * state->rangeUV + state->biasUV);
            return AVIF_CLAMP(unorm, 0, state->yuvMaxChannel);
    }
}
```

Note: In my experiments, clang does not warn about these two versions of avifReformatStateUVToUNorm(). Only gcc does. (I have gcc 10.2.0.) That may be a sign that clang and gcc handle this kind of switch statement differently.

By using a default case in the switch statement (or equivalently, ending an if-else-if chain with an "else"), we avoid this issue. But the benefits of this kind of switch statement may be more important than the minor annoyance of dealing with these compiler warnings. (I believe any performance difference of the generated code cannot be measured.)

PR #460, opened 2020-12-23
Improve message for avifInputReadImage failure

wantehchang:
Improve the error message for avifInputReadImage failure. The current
message is "Cannot determine input file format", but that is not the
only reason for avifInputReadImage failure. For example,
avifInputReadImage() fails if the CICP in 'image' is not supported.

Apparently we try to address this issue by having avifJPEGRead() print
an error message if its avifImageRGBToYUV() call fails. But I was still
confused by the "Cannot determine input file format" message when I
passed an unsupported --cicp option to avifenc.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -615,7 +615,7 @@ int main(int argc, char * argv[])
     uint32_t sourceDepth = 0;
     avifAppFileFormat inputFormat = avifInputReadImage(&input, image, &sourceDepth);
     if (inputFormat == AVIF_APP_FILE_FORMAT_UNKNOWN) {
-        fprintf(stderr, "Cannot determine input file format: %s\n", firstFile->filename);
+        fprintf(stderr, "Cannot determine input file format or read input file into avifImage: %s\n", firstFile->filename);
         returnCode = 1;
         goto cleanup;
     }


apps/avifenc.c
```diff
@@ -615,7 +615,7 @@ int main(int argc, char * argv[])
     uint32_t sourceDepth = 0;
     avifAppFileFormat inputFormat = avifInputReadImage(&input, image, &sourceDepth);
     if (inputFormat == AVIF_APP_FILE_FORMAT_UNKNOWN) {
-        fprintf(stderr, "Cannot determine input file format: %s\n", firstFile->filename);
+        fprintf(stderr, "Cannot determine input file format or read input file into avifImage: %s\n", firstFile->filename);
```
joedrago, line 618: Nit: I don't like `into avifImage` in the added text. Is `or read input file` sufficient here?

wantehchang: Thanks for the review. The reason I wrote this pull request was that the avifImageRGBToYUV() call in avifJPEGRead() failed. So "read input file" won't describe that error. Maybe we can enumerate all possible errors?

`        fprintf(stderr, "Cannot determine input file format, read input file, or convert to YUV: %s\n", firstFile->filename);`

vrabaud:
Closing as the message has been changed at HEAD: https://github.com/AOMediaCodec/libavif/blob/e107e095f1ada7c9de872ccdbbf556816e2e37a5/apps/avifenc.c#L624

Issue #440, opened 2020-12-09
(minor) Adjust min/max/speed defaults in avifenc?


joedrago:
As I think any actual avifenc commandline "recipe" will be quite explicit in its settings so this is semi-moot, my goal with the defaults for `avifenc` are to create reasonably-nice looking AVIFs at a decent speed, with opinionated-but-sensible defaults for other settings. With that in mind, I think the `min/max/speed` defaults might be missing the mark a bit.

`speed` is currently set to `8`, which is a nice and snappy experience for new `avifenc` users, but internally it will engage libaom's `AOM_USAGE_REALTIME`, which is probably not what we want. I think `4` might be a better default, but it will make the default experience slower.

 `min` and `max` are currently set to `0` and `10` respectively, which I believe is a bit too "high quality". The resultant files look very nice, but should probably be larger QP to show off AVIF for the first time for new people. 

My suggestion is to switch to something like `speed=4`, `min=24`, `max=26`. I'm hoping to hear feedback on `min/max/speed` defaults from some folks I know have experimented with recipes or are very familiar with AV1's guts. 

@wantehchang @AurelC2G @kornelski @jonsneyers @dwbuiten @cconcolato (I'm sure I'm missing folks)

kornelski:
Yes, I think for command-line tools it's better to do their best, even if that's slow. Image conversions done off-line or as part of build scripts can be slow. 

There's another use-case of realtime on-the-fly conversion, but implementation of such setup is much more involved, so more advanced developers can be expected to customize speed settings for their need. Also, to be honest, AV1 encoding is too slow for such use-cases, so it's better to focus on off-line conversion.

adityamavlankar:
The proposed defaults look good to me. From my experiments on compression efficiency vs speed I've observed that speed 4 incurs a single digit BD-rate loss (4-9%) compared to the best case (speed 0). Speed 8 incurs a bigger loss (20-35%). So speed 4 seems like a good tradeoff. Also agree on min/max 24/26. What I've seen with min=0 (regardless of max setting) is the encoder can employ very low QPs for several blocks -- thereby spending lot of bits for not-really-noticeable quality improvement. Perhaps the min/max could be tuned to output resolution. For smaller images (boxshot like with plenty visual info packed in there) a good default might be lower -- like min/max 20/22 or 22/24.

wantehchang:
Joe: The current default speed of 8 is not ideal for the libaom encoder. The new default speed can be one of the following:
[...]

PR #465, opened 2021-01-06
Change avifenc default speed from 4 to 6

wantehchang:
Speed 6 is the fastest speed in libaom's good-quality usage mode. In one
testing Frank Galligan performed on 20 images, the average encoding
times for speeds 3, 4, 6, 8 are roughly:
  - Speed 3: 125 seconds
  - Speed 4:  63 seconds
  - Speed 6:  42 seconds
  - Speed 8:  11 seconds

So speed 6 is likely to save 1/3 of the encoding time compared with
speed 4.

https://github.com/AOMediaCodec/libavif/issues/440

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -369,7 +369,7 @@ int main(int argc, char * argv[])
     int maxQuantizerAlpha = AVIF_QUANTIZER_LOSSLESS;
     int tileRowsLog2 = 0;
     int tileColsLog2 = 0;
-    int speed = 4;
+    int speed = 6;
     int paspCount = 0;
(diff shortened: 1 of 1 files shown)


FrankGalligan:
I think speed 6 is probably a better choice right now. At least going from 8 to 6 is not a big a jump as going to 8 to 4 encode time wise. We could always change the default later to 4 if we need too. Especially given that most complaints are not about quality, but encode time.

Avg PSNR for speeds above:
- Speed 3: 50.69
- Speed 4: 50.59
- Speed 6: 50.41
- Speed 8: 45.44

PR #467, opened 2021-01-11
Sync libgav1, dav1d and abseil-cpp versions between ext and docker

EwoutH:
Use the same versions in ext and docker:

- dav1d 0.8.1
- libgav1 4a89dc3
- abseil-cpp lts_2020_09_23 maintenance branch


EwoutH:
@joedrago Could you review this PR?

PR #469, opened 2021-01-13
Update rav1e.cmd and docker/build.sh to use rav1e 0.4

EwoutH:
Switches both rav1e.cmd and docker/build.sh to update from the rav1e 0.3 maintenance branch to the [0.4 maintenance branch](https://github.com/xiph/rav1e/tree/0.4).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     - name: Install dependencies
       run: |
         DEBIAN_FRONTEND=noninteractive sudo apt-get update || true
-        DEBIAN_FRONTEND=noninteractive sudo apt-get install -y ninja-build gcc-10 g++-10
+        DEBIAN_FRONTEND=noninteractive sudo apt-get install -y ninja-build gcc-10 g++-10 build-essential
         pip install --upgrade pip
         pip install setuptools
         pip install meson
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 * Update dav1d.cmd to point at the 0.8.1 tag
+* Update rav1e.cmd to point at the 0.4 maintenance branch
 
 ## [0.8.4] - 2020-11-23
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,7 +309,7 @@ if(AVIF_CODEC_RAV1E)
         endif()
 
         set(AVIF_CODEC_INCLUDES ${AVIF_CODEC_INCLUDES}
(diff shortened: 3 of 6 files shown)


EwoutH:
Currently the CI is stuck on [the following error](https://github.com/AOMediaCodec/libavif/runs/1695889933#step:12:27):
```
/home/runner/work/libavif/libavif/src/codec_rav1e.c:6:10: fatal error: rav1e/rav1e.h: No such file or directory
    6 | #include "rav1e/rav1e.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/avif.dir/build.make:277: CMakeFiles/avif.dir/src/codec_rav1e.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:138: CMakeFiles/avif.dir/all] Error 2
make: *** [Makefile:149: all] Error 2
Error: Process completed with exit code 2.
```
I'm going to need a bit of help here, @joedrago can you make anything of it?

@lu-zero Could you confirm if [rav1e.cmd](https://github.com/EwoutH/libavif/blob/rav1e-0.4/ext/rav1e.cmd#L17) now installs the c-compatible rav1e library correctly?

lu-zero:
which version of cargo-c are you using ?

EwoutH:
`cargo-c v0.7.0+cargo-0.49` [currently](https://github.com/AOMediaCodec/libavif/runs/1695889933#step:8:12)

lu-zero:
you may include `rav1e.h` directly if `pkg-config --cflags` provides `-I{includedir}/rav1e`

joedrago:
Your change to `rav1e.cmd` breaks the intent of the script. The goal of that script is to create a static, _local_ build of the library, not install it into the system, and then require that the system has it available.

joedrago:
If you want rav1e 0.4, please just change the version branches and leave the call to cbindgen alone.

EwoutH:
Thanks for the reply, I tried that in the [first commit](https://github.com/AOMediaCodec/libavif/pull/469/commits/3f71899ef009e274351d4b94f5e6f686d4f4298d) of this PR, but unfortunately that [run](url) also fails.
`  libavif: compiled rav1e library is missing (in ext/rav1e/target/release),`

lu-zero:
The current cargo-c introduced {pkg}-uninstalled.pc to address this, but looks like it need a [fix](https://github.com/lu-zero/cargo-c/issues/157), I'll look at it during the weekend.

EwoutH:
Something I missed was that rav1e got stuck on this error: `Error: CliError { error: Some(Permission denied (os error 13)), exit_code: 101 }`

Which is probably a GitHub Actions problem. I will see if I can modify it to run with sudo privileges.

EwoutH:
Here is a similar issue: https://github.com/rust-lang/cargo/issues/6757

EwoutH:
Succeeded by #476.

PR #476, opened 2021-01-18
Upgrade to rav1e 0.4.0

tongyuantongyu:
rav1e 0.4.0 changed its way to build C library. Following the new way.

### DIFF
--- a/ext/rav1e.cmd
+++ b/ext/rav1e.cmd
@@ -11,11 +11,10 @@
 : # Also, the error that "The target windows-msvc is not supported yet" can safely be ignored provided that rav1e/target/release
 : # contains rav1e.h and rav1e.lib.
 
-git clone -b 0.3 --depth 1 https://github.com/xiph/rav1e.git
+git clone -b 0.4 --depth 1 https://github.com/xiph/rav1e.git
 
 cd rav1e
-cargo install cbindgen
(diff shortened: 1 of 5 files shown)


EwoutH:
Great work, nice you found the path changes!

Could you include a line in the [changelog](https://github.com/AOMediaCodec/libavif/blob/master/CHANGELOG.md) and update the [docker build](https://github.com/AOMediaCodec/libavif/blob/master/tests/docker/build.sh)? (see #476)

ext/rav1e.cmd
```diff
@@ -11,11 +11,10 @@
...
-git clone -b 0.3 --depth 1 https://github.com/xiph/rav1e.git
+git clone -b 0.4 --depth 1 https://github.com/xiph/rav1e.git
 
 cd rav1e
-cargo install cbindgen
-cbindgen -c cbindgen.toml -l C -o target/release/include/rav1e/rav1e.h --crate rav1e .
+cargo install cargo-c
 
-cargo build --lib --release --features capi
+cargo cbuild --release
```
lu-zero, line 19: If you want to use cargo-c you should pass `--library-type=static` to not build an unnecessary dynamic library

tongyuantongyu: It seems to be `--library-type=staticlib` instead of `--library-type=static`. Anyway, thanks for pointing that out!

EwoutH (approved):

PR #477, opened 2021-01-22
Support premultiply alpha

tongyuantongyu:
Add support to signaling image is premultiplied, and converting between premultiply and non-premultiply RGB values.

Produced file and premultiply flag can be recognized correctly by mp4parse-rust, and can be opened by Firefox, though currently Firefox is not handling alpha correctly. (always premultiply once more than should)

Also changed JPEG writer to always output premultiplied RGB value, as it's the real appearance on black background.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -54,6 +54,7 @@ static void syntax(void)
     printf("    -l,--lossless                     : Set all defaults to encode losslessly, and emit warnings when settings/input don't allow for it\n");
     printf("    -d,--depth D                      : Output depth [8,10,12]. (JPEG/PNG only; For y4m or stdin, depth is retained)\n");
     printf("    -y,--yuv FORMAT                   : Output format [default=444, 422, 420, 400]. (JPEG/PNG only; For y4m or stdin, format is retained)\n");
+    printf("    -p,--premultiply                  : Premultiply color with alpha channel\n");
     printf("    --stdin                           : Read y4m frames from stdin instead of files; no input filenames allowed, must set before offering output filename\n");
     printf("    --cicp,--nclx P/T/M               : Set CICP values (nclx colr box) (3 raw numbers, use -r to set range flag)\n");
     printf("                                        P = enum avifColorPrimaries\n");
@@ -391,6 +392,7 @@ int main(int argc, char * argv[])
     int timescale = 1; // 1 fps by default
     int keyframeInterval = 0;
     avifBool cicpExplicitlySet = AVIF_FALSE;
+    avifBool premultiplyAlpha = AVIF_FALSE;
 
     int gridDimsCount = 0;
     uint32_t gridDims[8]; // only the first two are used
@@ -669,6 +671,8 @@ int main(int argc, char * argv[])
                                                               // https://github.com/xiph/rav1e/issues/151
             requestedRange = AVIF_RANGE_FULL;                 // avoid limited range
             matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_IDENTITY; // this is key for lossless
+        } else if (!strcmp(arg, "-p") || !strcmp(arg, "--premultiply")) {
+            premultiplyAlpha = AVIF_TRUE;
         } else {
             // Positional argument
             input.files[input.filesCount].filename = arg;
@@ -708,6 +712,7 @@ int main(int argc, char * argv[])
     image->transferCharacteristics = transferCharacteristics;
     image->matrixCoefficients = matrixCoefficients;
     image->yuvRange = requestedRange;
+    image->alphaPremultiplied = premultiplyAlpha;
 
     avifInputFile * firstFile = avifInputGetNextFile(&input);
     uint32_t sourceDepth = 0;
@@ -880,6 +885,7 @@ int main(int argc, char * argv[])
             cellImage->transferCharacteristics = image->transferCharacteristics;
             cellImage->matrixCoefficients = image->matrixCoefficients;
             cellImage->yuvRange = image->yuvRange;
+            cellImage->alphaPremultiplied = image->alphaPremultiplied;
             gridCells[gridCellIndex] = cellImage;
 
             avifAppFileFormat nextInputFormat = avifInputReadImage(&input, cellImage, NULL);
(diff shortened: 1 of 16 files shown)


joedrago:
Thanks for this. I will be out of the office until Tuesday, but I will review it then.

joedrago:
Alright, I'm reading this now, and there are a few minor style changes I'll be proposing, but I'm generally really pleased with this approach. 

I'm curious about the commandline usage of this though. It looks like your design (`-p`) assumes that the incoming data from a PNG (say) is _never_ premultiplied, and avifdec never un-premultiplies the data, thus _always_ creating a premultiplied PNG. This seems inconsistent. 

Perhaps we should distinguish on both avifenc and avifdec commandlines between the following use cases:

    * avifenc
      * User doesn't provide any settings (assumes not premultiplied)
      * User wants to give us an un-premultiplied PNG and create a premultiplied AVIF (`-p`)
      * User wants to give us a premultiplied PNG and create a premultiplied AVIF (?)
      * User wants to give us a premultiplied PNG and create a not-premultiplied AVIF (perhaps not a useful case)
    * avifdec
      * User doesn't specify what kind of PNG they want, but gives us an AVIF (always output _not_ premultiplied?)
      * User specifies they want a premultiplied PNG, and hands us a premultiplied AVIF (current default behavior, but specified)
      * User specifies that they want a premultiplied PNG, but hands us a not premultiplied AVIF (perhaps not useful?)

I haven't done a deepdive yet, but I think the high-level implementation at the `avifImage` and `avifRGBImage` is nice (especially leveraging libyuv when possible). We need to iron out the tooling, though.

tongyuantongyu:
PNG is assumed not premultiplied everywhere, so I believe we don't need to produce premultiplied PNG. If one really want this, then just use the API itself, but don't expect anything display the Image correctly.
 I remember I both read / write PNG data as not premultiplied. I made alphaPremultiplied in `avifImage` and `avifRGBImage` always reflect the real status between color and alpha, and let `avifRGBImagePremultiplyAlpha` change this field if conversation is successful.

joedrago:
Ah, I had missed that you forced it off in the PNG Writer.

Why aren't the pre-multiply steps a part of YUV<->RGB conversion (unless I missed it again)? Why make the end-user manually do it, and error in these calls? avifRGBImage exists to be params into these functions, so I would prefer that instead of erroring out if they don't match, it just automatically performed the appropriate work.

tongyuantongyu:
`avifRGBImagePremultiplyAlpha` and `avifRGBImageUnpremultiplyAlpha` can and is implemented to work in-place. If the pixel data of the source `avifRGBImage` can be modified freely then it's the best, otherwise we need to make a copy and work on it.
The source `avifRGBImage` parameter in `avifImageRGBToYUV` function is const-qualified, so we have to 
 - either make a copy of the source image, which can be insufficient;
 - or modify the source image, which can sometimes surprise the user / break something.
Considering this, I asked user to do this manually. For consistency, `avifImageYUVToRGB` doesn't do it either.

If we can assume modifying source `avifRGBImage` in `avifImageRGBToYUV` is fine, I would like to make this a part of YUV<->RGB conversion too.

joedrago:
Is there any reason it can't behave similarly to the rest of `alpha.c`, where it simply operates on an alpha pointer with a stride/offset, so it could work on either a separated alpha plane or an interleaved one? Or do we just need a few more functions? I don't see why it couldn't read from the interleaved alpha in avifRGBImage and write to the separate alpha in avifImage, or vice versa.

tongyuantongyu:
The libyuv function only works on interleaved RGBA(BGRA) plane. I planned to utilize libyuv's functions when writing our own version, so I adopted its design.

joedrago:
Your strategy makes sense, and it is unfortunate that the pre-existing API doesn't make this easier.

That said, they don't want to introduce a new error here where there wasn't one before, and I don't want someone to have to call another function to avoid it. Perhaps the compromise can simply be to leave your architecture in place exactly as it is with the additional function, but replace the error-out in the YUV conversion functions with a slower path? 

I'd rather be in a situation where somebody can opt into additional speed by adding a call, and we can optimize that slower path in the future, then to introduce an error case and forever require that people call two separate functions. 

If you're not interested in writing that slower path, I can take a shot at it. It is up to you. I appreciate your contribution so far.

joedrago:
Very nice! I'm going to ask Wan-Teh Chang if he can review the code itself as well, but if he doesn't get around to it by Friday evening, I'll give it another pass and accept it. Again, thanks for the effort here; this feature was a long time coming.

joedrago:
Alright, I just merged this, along with [this set of changes](ceb2fa0535cf5f0fde7a45d9b5a652c8b1af9659):

    * clang-format all files
    * Make it clear in avifenc syntax that -p signals prem in the resultant AVIF
    * Fix tref prem parsing to match tref auxl parsing (same syntax/box)
    * parenthesize binops in assignments (style)
    * Continue new convertResult pattern in avifImageYUVToRGB instead of early-out
    * Replace pColorIrefType/pColorIrefToID with new avifEncoderDataFindItemByID() call, as
      avifEncoderItem* mem storage may shift during new calls to avifEncoderDataCreateItem() so
      deep pointers into them during this is bad, plus this happens to remove accidental
      Hungarian notation, which is good

There is still a bit to be done here, but this covers most of it. This'll need a bunch of testing and an audit from @wantehchang before I'm ready to bump the next version of libavif with this in it.

@cconcolato Can you run this new `avifenc -p` on some transparent PNGs and verify that the prem references in the container are valid, please? I find it odd that the `prem` reference is on the primary item, but if it comes out correctly in the AVIF, I'm happy.

joedrago:
@baumanj @kornelski for visibility (upcoming premultiplied alpha support)

novomesk:
We should force unpremultiplied in https://github.com/AOMediaCodec/libavif/blob/master/contrib/gdk-pixbuf/loader.c because gdk pixbuf do not handle premultiplied.

joedrago:
If you throw that line in as a PR, I'll immediately accept it.

novomesk:
I will do it tomorrow.

joedrago:
Hmm, actually, based on this [principle](https://en.m.wikipedia.org/wiki/Principle_of_least_astonishment), perhaps the default for avifRGBImage should just be false?

novomesk:
It would be indeed easier to have it deactivated by default. Otherwise whole ecosystem need to make a tiny code update.

joedrago:
Done.

Those upgrading libavif and wanting to plumb through the unaltered (premultiplied) data can trivially opt _into_ this new behavior, and those that continue to expect not-premultiplied data will do so as well, even if the AVIF has `prem` set.

PR #478, opened 2021-01-22
Helptexts: note difference between min and max Q

rubo77:
Maybe you have an even better idea how to phraze that.

There could an alike explanation for alpha, though it is not clear to me:  I still wasn't able to use anything else than 0 there without errors


novomesk:
Why do you think that distance between minQ and maxQ cannot be larger than 10? I often use difference up to 20 without problem.

rubo77:
I got an error When using a larger difference By the time writing

vrabaud:
That should not be the case. If that is still happening at head, please open up a bug with the command your are using. Thx!

PR #483, opened 2021-01-30
Fix memcpy/sizeof typo in avifImageCopy

fdintino:
I stumbled across this typo while trying to debug i686 CI failures for python-pillow/Pillow#5201.

`sizeof(dstImage->pasp)` is 8, but `sizeof(dstImage->imir)` is 1. On 32-bit architectures that meant that the `avifImage->exif` pointer was fully copied to `dstImage`. This resulted in a double free abort, because both `avifEncoderDataDestroy` and `avifImageDestroy` were doing a `free` of the same location in memory (of `encoder->data->metadataImage->exif` and `image->exif`, respectively).


wantehchang (approved):
LGTM. Nice catch. Thanks a lot for the fix!

Joe: We should switch to struct assignment to avoid this class of bug.

PR #489, opened 2021-02-08
Impose a maximum of 4096 bytes on searchSampleSize

wantehchang:
Fix https://crbug.com/oss-fuzz/30478.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2838,6 +2838,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
             // Harvest CICP from the AV1's sequence header, which should be very close to the front
             // of the first sample. Read in successively larger chunks until we successfully parse the sequence.
             static const size_t searchSampleChunkIncrement = 64;
+            static const size_t searchSampleSizeMax = 4096;
             size_t searchSampleSize = 0;
             do {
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -2838,6 +2838,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
             // Harvest CICP from the AV1's sequence header, which should be very close to the front
             // of the first sample. Read in successively larger chunks until we successfully parse the sequence.
             static const size_t searchSampleChunkIncrement = 64;
+            static const size_t searchSampleSizeMax = 4096;
```
wantehchang, line 2841: We could also fix this bug by doubling searchSampleSize instead of adding a fixed increment.

joedrago: I think if you're confident that the sequence header needs to be in the front of the AV1 payload that this should be sufficient.

wantehchang: The first sequence header OBU must come before any frame. The only OBUs that may come before the first sequence header OBU are:
- the temporal delimiter OBU (very small because it has no payload data)
- padding OBUs
- unlikely but seems to be allowed by the spec: maybe certain metadata OBUs

PR #491, opened 2021-02-09
Refine Premultiply alpha handling during conversion, update libyuv usage

tongyuantongyu:
It turns out that handling premultiply alpha during YUV<->RGB conversion is more complicated than I thought it was, and code in #477 need some further changes.

The absent part is: if we are converting some image with alpha into a format without alpha, we should do 'premultiply alpha' before discarding alpha plane. This has the same effect of rendering this image on a black background, which makes sense.

For RGB->YUV, the logic is simpler because YUV and RGB image always both have or don't have alpha:

| ↓YUV, RGB→ | premultiplied | not-premultiplied | absent |
| --- | --- | --- | --- |
| premultiplied | no-op | premultiply | - |
| not-premultiplied | unpremultiply | no-op | - |
| absent | - | - | no-op |

And this is the logic for YUV->RGB:
| ↓YUV, RGB→ | premultiplied | not-premultiplied | absent |
| --- | --- | --- | --- |
| premultiplied | no-op | unpremultiply | no-op |
| not-premultiplied | premultiply | no-op | premultiply |
| absent | fill opaque | fill opaque | no-op |

There are two cases if RGB is absent:
1. RGB format has alpha, but ignoreAlpha is set. (XRGB)
2. RGB format doesn't have alpha. (RGB)

2 is tricky, because we were doing premultiply alpha as a post step of conversion but A channel is already discarded during conversion. I added premultiply alpha handling in `avifImageYUVAnyToRGBAnySlow`, routed any YUVA(Not premultiplied)->RGB conversion to it, and keep using the fast paths for other cases.

This PR also utilized more functions in libyuv.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -140,7 +140,8 @@ typedef enum avifResult
     AVIF_RESULT_IO_ERROR,
     AVIF_RESULT_WAITING_ON_IO, // similar to EAGAIN/EWOULDBLOCK, this means the avifIO doesn't have necessary data available yet
     AVIF_RESULT_INVALID_ARGUMENT, // an argument passed into this function is invalid
-    AVIF_RESULT_NOT_IMPLEMENTED   // a requested code path is not (yet) implemented
+    AVIF_RESULT_NOT_IMPLEMENTED,  // a requested code path is not (yet) implemented
+    AVIF_RESULT_NEED_POST_PROCESS // a requested code path need post process
 } avifResult;
 
 AVIF_API const char * avifResultToString(avifResult result);
@@ -476,7 +477,8 @@ typedef struct avifRGBImage
                                            // Unused when converting to YUV. avifRGBImageSetDefaults() prefers quality over speed.
     avifBool ignoreAlpha;        // Used for XRGB formats, treats formats containing alpha (such as ARGB) as if they were
                                  // RGB, treating the alpha bits as if they were all 1.
-    avifBool alphaPremultiplied; // indicates if RGB value is pre-multiplied by alpha. Default: false
+    avifBool alphaPremultiplied; // Indicates if RGB value is pre-multiplied by alpha. Default: false
+                                 // This field has no effect for RGB formats without alpha.
 
     uint8_t * pixels;
     uint32_t rowBytes;
@@ -537,8 +539,10 @@ typedef struct avifReformatState
     float rgbMaxChannelF;
     float biasY;   // minimum Y value
     float biasUV;  // the value of 0.5 for the appropriate bit depth [128, 512, 2048]
+    float biasA;   // minimum A value
     float rangeY;  // difference between max and min Y
     float rangeUV; // difference between max and min UV
+    float rangeA;  // difference between max and min A
 
     avifPixelFormatInfo formatInfo;
 
@@ -547,6 +551,11 @@ typedef struct avifReformatState
     float unormFloatTableUV[1 << 12];
 
     avifReformatMode mode;
+
+    avifBool toRGBPremultiply;
+    avifBool toRGBUnpremultiply;
+    avifBool toYUVPremultiply;
+    avifBool toYUVUnpremultiply;
 } avifReformatState;
 AVIF_API avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * rgb, avifReformatState * state);
 
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -93,17 +93,19 @@ avifBool avifFillAlpha(const avifAlphaParams * const params);
 avifBool avifReformatAlpha(const avifAlphaParams * const params);
 
 // Returns:
-// * AVIF_RESULT_OK              - Converted successfully with libyuv
-// * AVIF_RESULT_NOT_IMPLEMENTED - The fast path for this combination is not implemented with libyuv, use built-in YUV conversion
-// * [any other error]           - Return error to caller
+// * AVIF_RESULT_OK                - Converted successfully with libyuv
+// * AVIF_RESULT_NOT_IMPLEMENTED   - The fast path for this combination is not implemented with libyuv, use built-in YUV conversion
+// * AVIF_RESULT_NEED_POST_PROCESS - Only color conversion part for this combination is implemented with libyuv, use built-in alpha conversion and premultiply
+// * [any other error]             - Return error to caller
 avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb);
 
 // Returns:
-// * AVIF_RESULT_OK              - (Un)Premultiply successfully with libyuv
-// * AVIF_RESULT_NOT_IMPLEMENTED - The fast path for this combination is not implemented with libyuv, use built-in (Un)Premultiply
-// * [any other error]           - Return error to caller
+// * AVIF_RESULT_OK                - (Un)Premultiply successfully with libyuv
+// * AVIF_RESULT_NOT_IMPLEMENTED   - The fast path for this combination is not implemented with libyuv, use built-in (Un)Premultiply
+// * [any other error]             - Return error to caller
 avifResult avifRGBImagePremultiplyAlphaLibYUV(avifRGBImage * rgb);
 avifResult avifRGBImageUnpremultiplyAlphaLibYUV(avifRGBImage * rgb);
+avifResult avifImageIdentity8ToRGB8ColorFullRangeLibYUV(const avifImage * image, avifRGBImage * rgb, avifReformatState * state);
 
 // ---------------------------------------------------------------------------
 // avifCodecDecodeInput
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -44,6 +44,10 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * 
     state->mode = AVIF_REFORMAT_MODE_YUV_COEFFICIENTS;
 
     if (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
+        if (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV444) {
+            return AVIF_FALSE;
+        }
+
         state->mode = AVIF_REFORMAT_MODE_IDENTITY;
     } else if (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO) {
         state->mode = AVIF_REFORMAT_MODE_YCGCO;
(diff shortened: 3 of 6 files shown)


tongyuantongyu:
Some notes on the changed libyuv YuvConstants selection logic:
BT.601 MC is not derived from BT.601 CP but BT.470M(FCC) CP, so use BT.601 YuvConstants on MC=CHROMA_NCL and CP=BT470M, instead of CP=BT601/BT470BG. However FCC MC is very inaccurate that it won't match the MC computed from FCC CP (under the precision libyuv YuvConstants have), so BT.601 YuvConstants is not used on MC=FCC.

wantehchang (changes requested):
Hi Yuan Tong,

Could you move the avifjpeg.c changes to a separate pull request? Also please remove lines 86-87:

```
    // JPEG doesn't have alpha. Prevent confusion.
    avif->alphaPremultiplied = AVIF_FALSE;
```
and remove all the code related to the `rgbPremultiplied` variable in `avifJPEGWrite()`.

Reverting the premultiplied alpha code in avifjpeg.c should fix issues 502-504. Thanks.

include/avif/avif.h
```diff
@@ -140,7 +140,8 @@ typedef enum avifResult
     AVIF_RESULT_IO_ERROR,
     AVIF_RESULT_WAITING_ON_IO, // similar to EAGAIN/EWOULDBLOCK, this means the avifIO doesn't have necessary data available yet
     AVIF_RESULT_INVALID_ARGUMENT, // an argument passed into this function is invalid
-    AVIF_RESULT_NOT_IMPLEMENTED   // a requested code path is not (yet) implemented
+    AVIF_RESULT_NOT_IMPLEMENTED,  // a requested code path is not (yet) implemented
+    AVIF_RESULT_NEED_POST_PROCESS // a requested code path need post process
```
wantehchang, line 144: 1. Nit: In the comment, change "need" to "needs". (The enum name can say "NEED".)

Also, it is more common to say "post-processing". "Post-process" sounds like a verb.

2. This avifResult value is only used by libavif internally. Ideally it should not be in the public API. I don't have a good solution right now, other than adding an internal enum for the return value of avifImageYUVToRGBLibYUV() and similar functions.

include/avif/internal.h
```diff
@@ -93,17 +93,19 @@ avifBool avifFillAlpha(const avifAlphaParams * const params);
 avifBool avifReformatAlpha(const avifAlphaParams * const params);
 
 // Returns:
-// * AVIF_RESULT_OK              - Converted successfully with libyuv
-// * AVIF_RESULT_NOT_IMPLEMENTED - The fast path for this combination is not implemented with libyuv, use built-in YUV conversion
-// * [any other error]           - Return error to caller
+// * AVIF_RESULT_OK                - Converted successfully with libyuv
+// * AVIF_RESULT_NOT_IMPLEMENTED   - The fast path for this combination is not implemented with libyuv, use built-in YUV conversion
+// * AVIF_RESULT_NEED_POST_PROCESS - Only color conversion part for this combination is implemented with libyuv, use built-in alpha conversion and premultiply
```
wantehchang, line 98: Nit: Add "the" before "color conversion".

src/reformat.c
```diff
@@ -44,6 +44,10 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
     state->mode = AVIF_REFORMAT_MODE_YUV_COEFFICIENTS;
 
     if (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
+        if (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV444) {
+            return AVIF_FALSE;
+        }
```
wantehchang, line 49: This restriction should be documented in the commit message. (I agree with this change, but in my experience this restriction is somewhat controversial.) I suggest we move this change to its own pull request so that it can be reviewed separately.

include/avif/avif.h
```diff
@@ -547,6 +551,11 @@ typedef struct avifReformatState
     float unormFloatTableUV[1 << 12];
 
     avifReformatMode mode;
+
+    avifBool toRGBPremultiply;
+    avifBool toRGBUnpremultiply;
+    avifBool toYUVPremultiply;
+    avifBool toYUVUnpremultiply;
```
wantehchang, line 558: Nit: These four probably can be local variables in the respective functions.

src/reformat.c
```diff
@@ -126,6 +132,24 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
+    if (image->alphaPlane) {
+        avifBool rgbShouldPremultiply = rgb->alphaPremultiplied || !avifRGBFormatHasAlpha(rgb->format) || rgb->ignoreAlpha;
+        state->toRGBPremultiply = !image->alphaPremultiplied && rgbShouldPremultiply;
+        state->toRGBUnpremultiply = image->alphaPremultiplied && !rgbShouldPremultiply;
+    } else {
+        state->toRGBPremultiply = AVIF_FALSE;
+        state->toRGBUnpremultiply = AVIF_FALSE;
+    }
+
+
```
wantehchang, line 152: Nit: Delete one blank line.

src/reformat.c
```diff
@@ -126,6 +132,24 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
...
+    if (avifRGBFormatHasAlpha(rgb->format) && !rgb->ignoreAlpha) {
+        state->toYUVPremultiply = image->alphaPremultiplied && !rgb->alphaPremultiplied;
+        state->toYUVUnpremultiply = !image->alphaPremultiplied && rgb->alphaPremultiplied;
+    } else {
+        state->toYUVPremultiply = AVIF_FALSE;
+        state->toYUVUnpremultiply = AVIF_FALSE;
+    }
+
+    if (image->alphaPlane) {
+        avifBool rgbShouldPremultiply = rgb->alphaPremultiplied || !avifRGBFormatHasAlpha(rgb->format) || rgb->ignoreAlpha;
```
wantehchang, line 144: IMPORTANT: We need some comments for this condition. You can adapt the following from the commit message:

> if we are converting some image with alpha into a format without alpha, we should do 'premultiply alpha' before discarding alpha plane. This has the same effect of rendering this image on a black background, which makes sense.

I also suggest we file an issue and discuss this policy there. I don't know what is the right thing to do in this case.

src/reformat.c
```diff
@@ -1008,17 +1077,21 @@ avifResult avifImageYUVToRGB(const avifImage * image, avifRGBImage * rgb)
...
     }
 
-    if (avifRGBFormatHasAlpha(rgb->format) && !rgb->ignoreAlpha) {
+    if (convertResult == AVIF_RESULT_OK) {
+        return AVIF_RESULT_OK;
+    } else if (convertResult != AVIF_RESULT_NOT_IMPLEMENTED && convertResult != AVIF_RESULT_NEED_POST_PROCESS) {
+        return convertResult;
+    }
+
+    // XRGB is considered as RGB which need premultiplied pixel, so we still need alpha if we need to do premultiply.
```
wantehchang, line 1093: Nit: Remove "do" before "premultiply" because we can use "premultiply" as a verb.

src/reformat.c
```diff
@@ -1040,15 +1113,13 @@ avifResult avifImageYUVToRGB(const avifImage * image, avifRGBImage * rgb)
...
-                avifFillAlpha(&params);
-            }
+            avifFillAlpha(&params);
         }
     }
 
-    // Do this after alpha conversion
-    if (convertedWithLibYUV) {
-        return AVIF_RESULT_OK;
+    // Done copy alpha. Do alpha (un)premultiply.
```
wantehchang, line 1120: Nit: copy => copying

include/avif/avif.h
```diff
@@ -537,8 +539,10 @@ typedef struct avifReformatState
     float rgbMaxChannelF;
     float biasY;   // minimum Y value
     float biasUV;  // the value of 0.5 for the appropriate bit depth [128, 512, 2048]
+    float biasA;   // minimum A value
```
wantehchang, line 542: Hi Yuan Tong: In https://github.com/AOMediaCodec/libavif/pull/515 I propose to remove avifReformatMode, avifReformatState, and avifPrepareReformatState() from the public API. If that pull request is accepted, the definition of struct avifReformatState will be moved to src/reformat.c or include/avif/internal.h. Please follow that pull request. Thanks!

Note: Based on your changes to reformat_libyuv.c in this pull request, I think I should move the definition of struct avifReformatState to include/avif/internal.h.

src/reformat_libyuv.c
```diff
@@ -174,262 +200,404 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
...
+
+    return AVIF_RESULT_OK;
+}
+
+avifResult avifImageIdentity8ToRGB8ColorFullRangeLibYUV(const avifImage * image, avifRGBImage * rgb, avifReformatState * state)
+{
+    const uint8_t * planePtr[4];
+    int planeRowBytes[4];
+
+    // state->rgbOffsetBytesA is 0 is RGB format doesn't have alpha.
```
wantehchang, line 552: Nit: Change the second "is" to "if"

src/reformat.c
```diff
@@ -220,7 +244,7 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                             a = rgb->pixels[state.rgbOffsetBytesA + (i * state.rgbPixelBytes) + (j * rgb->rowBytes)] / rgbMaxChannelF;
                         }
 
-                        if (!rgb->alphaPremultiplied && image->alphaPremultiplied) {
+                        if (state.toYUVPremultiply) {
```
wantehchang, line 247: I have a pending pull request that changes the conditional structures of the code here. Please take a look: https://github.com/AOMediaCodec/libavif/pull/510

An observation in that pull request is that we should not enter the outer if statement at line 238 unless we need to premultiply or unpremultiply.

wantehchang (changes requested):
Hi Yuan Tong: Here are some preliminary review comments. I didn't review reformat.c and reformat_libyuv.c.

tongyuantongyu:
Most changes of this PR had been separated into several different PRs, and for updating libyuv usage part, I would wait until libyuv implements more functions we can utilize, and create another PR for it then. So close this PR now.

Jamaika1:
I add tune butteraugli with AVIF_LIBYUV_ENABLED.
Unfortunately there are bugs.
```
ERROR: Failed to finish encoding: Encoding of color planes failed
```
https://chromium.googlesource.com/libyuv/libyuv/+log

tongyuantongyu:
> I add tune butteraugli with AVIF_LIBYUV_ENABLED.
> Unfortunately there are bugs.

I believe libyuv is unrelated as it is only used during decoding?

PR #494, opened 2021-02-11
adds support for PNG and JPEG reading from stdin when --stdin is specified

dbquarrel:
Adding this patch will allow <code>avifenc</code> to read png and jpg from stdin and so allow avifenc to participate in pipelines with imagemagick or similar image manipulation programs rather than having to write to a temp file, and then read from it.

Patch does the following:

1. sets a user defined buffer to use for stdin so we can do lookaheads on it
2. adds magic numbers for JPG and PNG headers (maybe better to use library code for doing this but I don't know those libraries) ... code is extensible to other file types simply by adding additional lines with magic numbers in the appropriate data structure.
3. when reading from stdin will check for magic numbers, and if it can recognize PNG or JPG will pass control back to the functions for reading those instead of the default 
4. if not recognizing PNG or JPG, program flow is restored to default handling
5. a bug is fixed on recognizing EOF on stdin (with the swizzled input buffer the current code couldn't recognize EOF)

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -94,22 +94,6 @@ static void avifSwizzleSTDIN(void) {
     }
 }
 
-#if 0
-#define TOUPPER(x) (((x) <= 'z' && (x) >= 'a') ? ((x) - 'a'+ 'A') : (x))
-static void testme(char *buf){
-#define mymax  8
-    char outbuf[mymax] = {0};
-    for(int i=0; i<mymax;i++) {
-	sprintf(outbuf,"%02x",(unsigned char)buf[i]);
-	for (unsigned long j=0;outbuf[j];j++){
-	    outbuf[j] = TOUPPER(outbuf[j]);
-	}
-	fprintf(stderr,"%s ",outbuf);
-    }
-    fprintf(stderr,"\n");
-}
-#endif
-
 static inline avifBool avifMagicEntryTest(const avifMagicEntry *entry,
 					  char *data)
 {
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -18,6 +18,114 @@
 #include <io.h>
 #endif
 
+/* Describes a single magic number to identify a file. */
+typedef struct {
+    int64_t length; /* Length of sequence */
+    int64_t offset; /* Offset from start of file in bytes */
+    const char * const sequence; /* the magic sequence */
+} avifMagicNumber;
+
+/* Describes the magic number sequence of a file type. */
+#define avifMagicNumberMax 2
+typedef struct {
+    /* two magic numbers possible in one file */
+    avifMagicNumber magic[avifMagicNumberMax];
+    /* The format associated with these magics */
+    avifAppFileFormat format; 
+} avifMagicEntry;
+
+/* A map of magic numbers to test to come up with a file identification. */
+static avifMagicEntry avifMagicMap[] = {
+    { /* entry */
+	/* magic */ { /*[0]*/  {4,0,"\xFF\xD8\xFF\xDB"}, /*[1]*/ {0}},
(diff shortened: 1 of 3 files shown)


joedrago:
There are many issues with this PR before it can be accepted:

1. libavif already has file format magic detection in `avifutil.c`: 

        avifAppFileFormat avifGuessFileFormat(const char * filename);

Instead of having two blocks of code in libavif's `app/` directory that does almost the exact same thing, I'd prefer it if a portion of `avifGuessFileFormat()` was split into into a second function: `avifGuessHeaderFormat()` which takes a `uint8_t*` and a `size_t` and returned the enum `avifGuessFileFormat()` returns. `avifGuessFileFormat()` can simply use this function internally, as can this new stdin detection code.

2. Merge the functionality of `avifSwizzleSTDIN()` and `avifInputDetermineFileType()` into a single function which uses (the new) `avifGuessHeaderFormat()` function? `avifInputDetermineFileType()` is redundant (see 1 above), but has a better name. `avifGuessStdinFormat()` might be a good choice for this kind of functionality, which would return whatever `avifGuessHeaderFormat()` came up with after your call to `setvbuf()`. 
 
3. Remove the #if0/else. If your new implementation is more correct, we can just always use that. Don't leave cruft (we can always find the old code in git history).

4. Please use end-of-line "C++" (C99) style comments everywhere (as all of libavif does).

5. Run `clang-format` on any file you touch, in order to leverage the checked in .clang-format provided by libavif and keep the syntax consistent.

6. Replace usages of `exit()` with error handling which results in returning from main(). There should be no need for `exit()`. If `avifGuessStdinFormat()` returns `AVIF_APP_FILE_FORMAT_UNKNOWN`, the rest of the code will gracefully clean up and exit.

7. It must survive all continuous builds (appveyor's Windows build failed here).

joedrago:
8. It does not update `avifenc`'s `syntax()` function to indicate that `--stdin` will tolerate a single JPEG or PNG payload as an alternative to multiple y4m frames.

dbquarrel:
Basically this is just an example of how to fix the design flaw in your program. It's one step better than making a report of an issue.

I don't have the time to go through and fine tune everything to harmonize it with your project, I understand your goals and they are correct.

It is a design flaw to not properly read and write in a pipeline, especially for an image processing program. If you want this to function and be a nice neighbor in a UNIX like environment you need to be able to pipe data to it and from it. Forcing people to write and clean up temp files is a major design issue. 

I just wanted to point it out and solve the problem with example code rather than to learn your entire project and fine tune the submission, as I don't have time for that and need to move on with my own stuff. 

So just look at it as example code and pick and choose whatever you want from it and throw away the rest. But it is necessary functionality for image processing at the commandline.

joedrago:
Understood. Thank you for the PR.

dbquarrel:
Also, your code for <code>avifGuessFileFormat</code> is problematic in several ways:

1. the magic number used for JPEG is a prefix but is not the full magic number, it's not necessarily guaranteed to be correct even though it works on a practical basis 

See:

https://en.wikipedia.org/wiki/List_of_file_signatures

This is pedantic, but do note you turn the pedantic flag on for compilation so it's in line with the goals to do it right.

2. it requires a file to exist in the filesystem, that can be opened. This does not work at all for standard input or an already open file, unless you want to reopen it for reading and close again, which is wasteful.

You cannot fseek() on stdin so reading and then moving the file pointer back is no solution, hence the only way possible to get this information out of input piped to your program (as far as I know) is via the method I implemented. Or, you can force the user to state what the file is, but this is kind of bad when it can be detected. 

This approach of mine was made to try to drop into your code without having to modify the rest of it, rather than swizzling the input buffer, a better way of doing it would be to read on your own into memory *before* making any decisions about what to do with the data. As is the design requires this approach which is not the best design or approach, but is fast to implement as it's opaque to the rest of the program.

PR #495, opened 2021-02-16
ext: Update SVT-AV1 repo to GitLab

EwoutH:
Updates the SVT-AV1 repository from GitHub to GitLab.

See: https://github.com/AOMediaCodec/SVT-AV1/issues/1679

I also noticed that `ext/svt.cmd` is cloning the `v0.8.6` tag, while `ext/svt.sh` is checking out the default (master) branch. Should that be the case?


joedrago:
The tags should match, that is an oversight. If you update this PR to make them match, I'll accept it.

EwoutH:
Done, they now both checkout the `v0.8.6` tag

PR #498, opened 2021-02-18
Recognize the Y4M format string "C420mpeg2"

wantehchang:
The aom/common/y4menc.c file in libaom says:

      if (csp == AOM_CSP_VERTICAL) {
        return "C420mpeg2 XYSCSS=420MPEG2";
      } else if (csp == AOM_CSP_COLOCATED) {
        // Note that Y4M does not have a dedicated header for colocated chroma,
        // and that FFMPEG interprets C420 as C420jpeg.
        return "C420";
      } else {
        return "C420jpeg";
      }

Extend y4mColorSpaceParse() to also set the chroma sample position.

### DIFF
--- a/apps/shared/y4m.c
+++ b/apps/shared/y4m.c
@@ -20,94 +20,103 @@ struct y4mFrameIterator
     avifBool hasAlpha;
     avifPixelFormat format;
     avifRange range;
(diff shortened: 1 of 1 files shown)


FrankGalligan (commented):
Looks good. Thanks

PR #499, opened 2021-02-18
Check out libaom v2.0.2

wantehchang:


EwoutH:
At this point I would suggest checking out the `applejack` [branch](https://aomedia.googlesource.com/aom/+/refs/heads/applejack), which is aom's v2.0 maintenance branch (just like we do with rav1e). It's very stable but essential fixes will be committed in the future.

joedrago:
My intent is to always point at actual releases, as that is what is likely to be the most stable and vetted. I believe at some point we had to move rav1e due to either an issue with monochrome or some other weird build issue (?), but if we're not pointing at a proper tagged release for rav1e, I'm more likely to move that than I am to switch libaom to `applejack`.

EwoutH:
Personally I think for a master/development branch it's fine to checkout maintenance branches, especially ones that have proven very stable. The advantage is that quite often they provide fixes for minor corner cases that only are discovered months after being stable, for example: https://github.com/xiph/rav1e/commit/16f88cd720ad9b31ee5a2114d2c7b5b2a02ead85. There isn't currently a tag that includes it, so you would have to checkout that specific commit (which is also an option though).

wantehchang:
The applejack branch of libaom does not have continuous or nightly tests. Tests are only run on that branch before a release. It is safer to check out a release tag.

joedrago:
I think you put way too much of a premium on the values inside of the .cmd files. They are merely meant as suggestions for known-good versions of our dependencies, and what we use when doing our automated builds. The values in those cmd files in no way restricts what anyone can build libavif against. Feel free to build your own avifenc / avifdec against any version of libaom, including with local changes.

Someday, those brand new changes will be vetted and considered to be known-good and we will start recommending them. If you want to take a chance with those changes for that additional speed, by all means use a different check out when you build.

wantehchang:
Ewout: Thank you for testing the tip of libaom. We are always improving the libaom encoder. Also, since June or July last year we have been improving libaom for still image encoding, focusing on speed 6. The speedup you observed comes from both efforts.

My coworker Jingning Han recently added an "all intra" mode to the libaom encoder. The commit of his that you linked to is only active in the all intra mode. You can apply the following patch and experiment with the new all intra mode:

diff --git a/src/codec_aom.c b/src/codec_aom.c
index 8990184..8acdd0a 100644
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -437,10 +437,14 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         // Speed  9: RealTime    CpuUsed 7
         // Speed 10: RealTime    CpuUsed 8
         unsigned int aomUsage = AOM_USAGE_GOOD_QUALITY;
+#if defined(AOM_USAGE_ALL_INTRA)
+        if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) {
+            aomUsage = AOM_USAGE_ALL_INTRA;
+        }
+#endif
         int aomCpuUsed = -1;
         if (encoder->speed != AVIF_SPEED_DEFAULT) {
             if (encoder->speed < 8) {
-                aomUsage = AOM_USAGE_GOOD_QUALITY;
                 aomCpuUsed = AVIF_CLAMP(encoder->speed, 0, 6);
             } else {

Incidentally, the libaom 2.1.0 release candidate 1 just become available. I will write a pull request to check out the v2.1.0-rc1 tag. The new all intra mode missed the 2.1.0 release cutoff.

Issue #497, opened 2021-02-18
Harvest framerate from y4m headers as the "default", if present


joedrago:
The default fps for `avifenc` is 1 fps (arbitrary) as the enduser is expected to supply a framerate when encoding an image sequence. However, if the enduser doesn't supply a framerate (with `--fps` or `--timescale`) and the source file contains a framerate, use that as the default instead of 1 fps.

(Currently this would just be for y4m source data.)

PR #500, opened 2021-02-18
Harvest framerate from y4m headers as the "default", if present

joedrago:
Change avifenc's internal defaults for duration/timescale to 0 (as a sentinel), and allow the first
source file's framerate (if detected) to override the original defaults if so.

Fixes: #497

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -399,13 +399,18 @@ int main(int argc, char * argv[])
     int keyframeInterval = 0;
     avifBool cicpExplicitlySet = AVIF_FALSE;
     avifBool premultiplyAlpha = AVIF_FALSE;
-    avifAppSourceTiming sourceTiming = { 0, 0 }; // See declaration of avifAppSourceTiming for documentation
     int gridDimsCount = 0;
     uint32_t gridDims[8]; // only the first two are used
     uint32_t gridCellCount = 0;
     avifImage ** gridCells = NULL;
     avifImage * gridSplitImage = NULL; // used for cleanup tracking
     memset(gridDims, 0, sizeof(gridDims));
+
+    // This holds the output timing for image sequences. The timescale member in this struct will
+    // become the timescale set on avifEncoder, and the duration member will be the default duration
+    // for any frame that doesn't have a specific duration set on the commandline. See the
+    // declaration of avifAppSourceTiming for more documentation.
+    avifAppSourceTiming outputTiming = { 0, 0 };
 
     // By default, the color profile itself is unspecified, so CP/TC are set (to 2) accordingly.
     // However, if the end-user doesn't specify any CICP, we will convert to YUV using BT601
@@ -593,7 +598,7 @@ int main(int argc, char * argv[])
                 returnCode = 1;
                 goto cleanup;
             }
-            sourceTiming.duration = (uint64_t)durationInt;
+            outputTiming.duration = (uint64_t)durationInt;
         } else if (!strcmp(arg, "--timescale") || !strcmp(arg, "--fps")) {
             NEXTARG();
             int timescaleInt = atoi(arg);
@@ -602,7 +607,7 @@ int main(int argc, char * argv[])
                 returnCode = 1;
                 goto cleanup;
             }
-            sourceTiming.timescale = (uint64_t)timescaleInt;
+            outputTiming.timescale = (uint64_t)timescaleInt;
         } else if (!strcmp(arg, "-c") || !strcmp(arg, "--codec")) {
             NEXTARG();
             codecChoice = avifCodecChoiceFromName(arg);
@@ -684,7 +689,7 @@ int main(int argc, char * argv[])
         } else {
             // Positional argument
             input.files[input.filesCount].filename = arg;
-            input.files[input.filesCount].duration = sourceTiming.duration;
+            input.files[input.filesCount].duration = outputTiming.duration;
             ++input.filesCount;
         }
 
@@ -692,7 +697,7 @@ int main(int argc, char * argv[])
     }
 
     stdinFile.filename = "(stdin)";
-    stdinFile.duration = sourceTiming.duration;
+    stdinFile.duration = outputTiming.duration;
 
     if (!outputFilename) {
         if (((input.useStdin && (input.filesCount == 1)) || (!input.useStdin && (input.filesCount > 1)))) {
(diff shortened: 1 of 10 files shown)


apps/avifenc.c
```diff
@@ -388,12 +393,21 @@ int main(int argc, char * argv[])
...
     avifRWData xmpOverride = AVIF_DATA_EMPTY;
     avifRWData iccOverride = AVIF_DATA_EMPTY;
-    int duration = 1;  // in timescales, stored per-inputFile (see avifInputFile)
-    int timescale = 1; // 1 fps by default
     int keyframeInterval = 0;
     avifBool cicpExplicitlySet = AVIF_FALSE;
     avifBool premultiplyAlpha = AVIF_FALSE;
 
+    // The timescale specifies how many units of time are in a second, and the duration is how many
+    // of those timescales this frame should be displayed for. For example, if they match, it signals
```
wantehchang, line 401: I suggest changing "timescales" to "time units" or "units of time". The wording used in ISO BMFF is "in the indicated timescale".

Alternatively, we can make a pass through the source tree later to correct this misuse of terminology.

wantehchang (approved):
Thank you for writing the patch! I suggest some changes.

apps/shared/y4m.c
```diff
@@ -173,7 +200,7 @@ static int y4mReadLine(FILE * inputFile, avifRWData * raw, const char * displayF
             goto cleanup; \
     } while (0)
 
-avifBool y4mRead(avifImage * avif, const char * inputFilename, struct y4mFrameIterator ** iter)
+avifBool y4mRead(avifImage * avif, const char * inputFilename, struct y4mFrameIterator ** iter, avifImageTiming * imageTiming)
```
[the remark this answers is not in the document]
joedrago: I've pushed a commit that reorders these arguments. I've standardized now on the filename being the first argument for all Read/Write functions in the `apps` layer, and I've moved the timing argument here adjacent to the image argument.

apps/shared/y4m.c
```diff
@@ -185,6 +212,7 @@ avifBool y4mRead(avifImage * avif, const char * inputFilename, struct y4mFrameIt
     frame.format = AVIF_PIXEL_FORMAT_NONE;
     frame.range = AVIF_RANGE_LIMITED;
     frame.chromaSamplePosition = AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN;
+    memset(&frame.timing, 0, sizeof(avifImageTiming));
```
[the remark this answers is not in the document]
joedrago: I'll admit to abusing `avifImageTiming` here, and I can see that the ambiguity isn't worth it. I've created a proper struct specifically for this role named `avifAppSourceTiming` which is much clearer in its usage and goals, and doesn't have these extra members to add confusion.

apps/avifenc.c
```diff
@@ -28,7 +29,7 @@
 typedef struct avifInputFile
 {
     const char * filename;
-    int duration;
+    uint64_t duration; // If 0, use the default duration
```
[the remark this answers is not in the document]
joedrago: This should be fixed in my second commit.

apps/avifenc.c
```diff
@@ -202,13 +203,17 @@ static avifBool avifInputHasRemainingData(avifInput * input)
     return (input->fileIndex < input->filesCount);
 }
 
-static avifAppFileFormat avifInputReadImage(avifInput * input, avifImage * image, uint32_t * outDepth)
+static avifAppFileFormat avifInputReadImage(avifInput * input, avifImage * image, uint32_t * outDepth, avifImageTiming * imageTiming)
 {
+    if (imageTiming) {
+        memset(imageTiming, 0, sizeof(avifImageTiming));
+    }
```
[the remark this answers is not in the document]
joedrago: I added a comment in my second commit here.

apps/shared/y4m.c
```diff
@@ -122,6 +123,32 @@ static avifBool y4mColorSpaceParse(const char * formatString, struct y4mFrameIte
...
+    if (framerateString[0] != 'F') {
+        return AVIF_FALSE;
+    }
+    ++framerateString; // skip past 'F'
+
+    char * colonLocation = strchr(framerateString, ':');
+    if (!colonLocation) {
+        return AVIF_FALSE;
+    }
+    *colonLocation = 0;
```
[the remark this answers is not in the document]
joedrago: No, I don't typically use it; note the zeroing of my end-of-strings after usages of `strncpy()` in a few places.

I'm not against it though. We could do a pass where we sprinkle in `'\0'` in all of the right places after this.

apps/shared/y4m.c
```diff
@@ -122,6 +123,32 @@ static avifBool y4mColorSpaceParse(const char * formatString, struct y4mFrameIte
...
+    ++colonLocation;
+
+    int numerator = atoi(framerateString);
+    int denominator = atoi(colonLocation);
+    if ((numerator < 1) || (denominator < 1)) {
+        return AVIF_FALSE;
+    }
+
+    imageTiming->timescale = (uint64_t)numerator;
+    imageTiming->durationInTimescales = (uint64_t)denominator;
```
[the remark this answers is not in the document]
joedrago: This should be resolved in my second commit, as I don't use `avifImageTiming` for this anymore.

apps/avifenc.c
joedrago: I think this is resolved as I rewrote the comment entirely when I moved it to above the new `avifAppSourceTiming` declaration. Let me know if it needs more tweaks.

apps/avifenc.c
```diff
@@ -396,18 +399,7 @@ int main(int argc, char * argv[])
...
-    // 1 fps. If duration is 1 and timescale is 30, it signals 30 fps.
-    //
-    // These are set to 0 as sentinel values on initialization, and (if unchanged) are set to actual
-    // defaults later (search for "Set timing defaults"). This is done so that if the commandline
-    // doesn't set either of these values and the first source image provides a framerate, avifenc
-    // can use it instead of a 1 fps default.
-    uint64_t duration = 0;
-    uint64_t timescale = 0;
-
+    avifAppSourceTiming sourceTiming = { 0, 0 }; // See declaration of avifAppSourceTiming for documentation
```
wantehchang, line 402: Nit: This variable name is a little confusing because it's not clear what "source" means. Note that this variable stores the values from the command line.

wantehchang (approved):
LGTM. Thanks! My only nit is that I don't really know what "source" means in "sourceTiming". I guess "source" means the input files (source materials), but sourceTiming may receive its values from the command line.

apps/avifenc.c
joedrago: I only named it that to make it clear it was an `avifAppSourceTiming` instead of an `avifImageTiming`, but you're right that this specific variable could have a better name. Suggestions? I could see it being `outputTiming`, perhaps?

joedrago: `timingDefaults` seems even more appropriate. Thoughts?

joedrago: I went with `outputTiming`. I'll push this now.

wantehchang: "outputTiming" or "animationTiming" sounds good.

PR #501, opened 2021-02-19
Remove extra parentheses around (x) < (low)

wantehchang:
There are two pairs of parentheses around (x) < (low). Remove one pair.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -11,7 +11,7 @@ extern "C" {
 #endif
 
 // Yes, clamp macros are nasty. Do not use them.
-#define AVIF_CLAMP(x, low, high) ((((x) < (low))) ? (low) : (((high) < (x)) ? (high) : (x)))
+#define AVIF_CLAMP(x, low, high) (((x) < (low)) ? (low) : (((high) < (x)) ? (high) : (x)))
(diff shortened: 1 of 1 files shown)


include/avif/internal.h
```diff
@@ -11,7 +11,7 @@ extern "C" {
 #endif
 
 // Yes, clamp macros are nasty. Do not use them.
-#define AVIF_CLAMP(x, low, high) ((((x) < (low))) ? (low) : (((high) < (x)) ? (high) : (x)))
+#define AVIF_CLAMP(x, low, high) (((x) < (low)) ? (low) : (((high) < (x)) ? (high) : (x)))
```
wantehchang, line 14: It's not easy to see, but the original code has:

`(((x) < (low)))`

joedrago: @tongyuantongyu Added these extra parens 29 days ago, perhaps they can provide reasoning? I'm happy to accept this PR, but I don't want to keep changing it. One of these is better and I'm ambivalent, but both of you appear to have a stronger opinion.

https://github.com/AOMediaCodec/libavif/commit/3cf850bbc89ba178c44dcbd56e4427ac7c559ef7

joedrago: Nevermind, I guess I'm mistaken on which parens are being discussed. I'll accept this PR.

Issue #504, opened 2021-02-20
avifenc.c: the premultiplyAlpha value is dropped by avifJPEGRead()


wantehchang:
In apps/avifenc.c, the `premultiplyAlpha` variable can be set by the command-line option `-p,--premultiply`. Although we set `image->alphaPremultiplied` to `premultiplyAlpha`:

`    image->alphaPremultiplied = premultiplyAlpha;`

`image` is passed to `avifInputReadImage()`. If the input file is a JPEG file, `avifInputReadImage()` calls `avifJPEGRead()`, which sets `avif->alphaPremultiplied` to false:

```
    avif->width = cinfo.output_width;
    avif->height = cinfo.output_height;
    avif->yuvFormat = requestedFormat;
    avif->depth = requestedDepth ? requestedDepth : 8;
    // JPEG doesn't have alpha. Prevent confusion.
    avif->alphaPremultiplied = AVIF_FALSE;
    avifRGBImageSetDefaults(&rgb, avif);
```

So by the time we call `avifEncoderAddImage()`, `avif->alphaPremultiplied` is false and the `premultiplyAlpha` value is not used.

I think we can fix this by removing these two lines from `avifJPEGRead()`:

```
    // JPEG doesn't have alpha. Prevent confusion.
    avif->alphaPremultiplied = AVIF_FALSE;
```

and optionally moving the statement

`    image->alphaPremultiplied = premultiplyAlpha;`

after the  `avifInputReadImage()` call to avoid confusion.

wantehchang:
Fixed by https://github.com/AOMediaCodec/libavif/pull/505.

Issue #503, opened 2021-02-20
avifJPEGWrite() is suboptimal if the avifImage does not have an alpha channel.


wantehchang:
`avifJPEGWrite()` is suboptimal if the `avifImage avif` does not have an alpha channel. In this case, `avif->alphaPremultiplied` is false, so we set `rgb.format` to `AVIF_RGB_FORMAT_RGBA`:

`    rgb.format = avif->alphaPremultiplied ? AVIF_RGB_FORMAT_RGB : AVIF_RGB_FORMAT_RGBA;`

convert the `avifImage` to the RGBA format, and then call `avifRGBAToRGB()` to convert it to the RGB format. We should instead convert an `avifImage` without an alpha channel to the RGB format directly.

wantehchang:
Fixed by https://github.com/AOMediaCodec/libavif/pull/505.

Issue #502, opened 2021-02-20
Should we set cinfo.input_components to 4 when we set cinfo.in_color_space to JCS_EXT_RGBX?


wantehchang:
The `avifJPEGWrite()` function in apps/shared/avifjpeg.c has the following code:

```
#if defined(JCS_ALPHA_EXTENSIONS)
    cinfo.input_components = 3;
    cinfo.in_color_space = JCS_EXT_RGBX;
#else
    cinfo.input_components = 3;
    cinfo.in_color_space = JCS_RGB;
#endif
```

Should we set cinfo.input_components to 4 when we set cinfo.in_color_space to JCS_EXT_RGBX?

joedrago:
@tongyuantongyu - Can you field this question, please?

wantehchang:
Does cinfo.input_components = 3 tell the JPEG library to use only the first three components?

If this code is correct, maybe we can rewrite it as:

```
    cinfo.input_components = 3;
#if defined(JCS_ALPHA_EXTENSIONS)
    cinfo.in_color_space = JCS_EXT_RGBX;
#else
    cinfo.in_color_space = JCS_RGB;
#endif
```

tongyuantongyu:
`avifJPEGWrite` has some wrongs I didn't caught (including the above one). I found them later, and while trying to solve them I'm feeling having these things in `avifjpeg.c` is wrong, so I moved them to conversion code, which resulted in #491. Could you take a look on that?

wantehchang:
Fixed by https://github.com/AOMediaCodec/libavif/pull/505.

PR #505, opened 2021-02-20
Remove alpha premultiply code from avifjpeg.c

tongyuantongyu:
Alpha premultiply related logic should be done in conversion code, remove them from `avifjpeg.c`.

This removes the problematic code related to #502, #503, #504.


wantehchang (approved):
LGTM. Thank you, Yuan Tong!

PR #506, opened 2021-02-20
Fix off-by-one errors in err msg/code in avifyuv.c

wantehchang:
The condition in which it is unsafe to increment driftPixelCounts[drift]
should be drift >= MAX_DRIFT, not drift > MAX_DRIFT, and the error
message should say "drift greater than or equal to MAX_DRIFT".

Declare the maxDrift variable in the scope it is used in.

Zero-initialize the avifRGBImage rgb with memset() to ensure all struct
members are initialized.

Set rgb.width to the actual width (a + 1) before the
avifRGBImageUnpremultiplyAlpha() and avifRGBImagePremultiplyAlpha()
calls, otherwise the two functions will access uninitialized pixels.

Use "return 1" to exit the test on failure in test mode 3. This is
equivalent to exit(1) inside the main function, and is used for
consistency with test mode 1.

### DIFF
--- a/tests/avifyuv.c
+++ b/tests/avifyuv.c
@@ -247,7 +247,7 @@ int main(int argc, char * argv[])
                                             maxDrift = drift;
                                         }
                                     } else {
-                                        printf("ERROR: Encountered a drift greater than MAX_DRIFT(%d): %d\n", MAX_DRIFT, drift);
+                                        printf("ERROR: Encountered a drift greater than or equal to MAX_DRIFT(%d): %d\n", MAX_DRIFT, drift);
                                         return 1;
                                     }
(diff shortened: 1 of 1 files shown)


tests/avifyuv.c
```diff
@@ -468,21 +468,22 @@ int main(int argc, char * argv[])
                     }
                 }
 
+                rgb.width = a + 1;
```
wantehchang, line 471: In line 443 above, rgb.width is set to 'size' so that the avifRGBImageAllocatePixels() call at line 451 allocates an rgb.pixels buffer large enough for all the tests. Here, we set rgb.width to the actual width of the current test.

Issue #512, opened 2021-02-20
Add a variant of avifRGBImageSetDefaults() that does not need an 'avifImage * image' input


wantehchang:
The `avifRGBImageSetDefaults()` function has the following prototype:

`AVIF_API void avifRGBImageSetDefaults(avifRGBImage * rgb, const avifImage * image);`

The `avifImage * image` input parameter has two problems.

1. The `avifImage` struct has a lot of fields. It is not clear which fields are used to initialize the `avifRGBImage` struct.

This issue came up recently when we added an `alphaPremultiplied` field to both structs. It is not clear if `avifRGBImageSetDefaults()` sets `rgb->alphaPremultiplied` to `avif->alphaPremultiplied`.

This problem can be addressed by adding comments for `avifRGBImageSetDefaults()` in avif.h.

2. There may not be an avifImage struct available. In this case libavif clients can either initialize avifRGBImage fields manually (which is brittle because new avifRGBImage fields may be added) or resort to memset() the struct (which requires all new members to default to zero or AVIF_FALSE).

I suggest we add a variant of avifRGBImageSetDefaults() that takes width, height, and depth as input.

PR #513, opened 2021-02-20
Document avifRGBImageSetDefaults()

wantehchang:
Address the first problem reported in
https://github.com/AOMediaCodec/libavif/issues/512.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -482,6 +482,9 @@ typedef struct avifRGBImage
     uint32_t rowBytes;
 } avifRGBImage;
 
+// Sets rgb->width, rgb->height, and rgb->depth to image->width, image->height, and image->depth.
+// Sets rgb->pixels to NULL and rgb->rowBytes to 0. Sets the other fields of 'rgb' to default
+// values.
 AVIF_API void avifRGBImageSetDefaults(avifRGBImage * rgb, const avifImage * image);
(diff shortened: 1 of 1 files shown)


include/avif/avif.h
```diff
@@ -482,6 +482,7 @@ typedef struct avifRGBImage
     uint32_t rowBytes;
 } avifRGBImage;
 
+// Sets rgb->width, rgb->height, and rgb->depth to image->width, image->height, and image->depth. Sets rgb->pixels to NULL and rgb->rowBytes to 0. Sets the other fields of 'rgb' to default values.
```
wantehchang, line 485: Joe: I forgot your preference for how long a comment line can be before it is folded.

joedrago: My editor breaks those comment lines like this:

```
// Sets rgb->width, rgb->height, and rgb->depth to image->width, image->height, and image->depth.
// Sets rgb->pixels to NULL and rgb->rowBytes to 0. Sets the other fields of 'rgb' to default
// values.
```

I believe it is 100 lines, although I can't legislate that in clang-format independent of other things, I don't believe.

wantehchang: Thanks! Yes, it looks like a 100-char line length. clang-format did not fold that very long comment line. Maybe we can document the 100-char comment line length in README.md?

Issue #514, opened 2021-02-20
Remove avifReformatMode, avifReformatState, and avifPrepareReformatState() from the public API


wantehchang:
While reviewing Yuan Tong's pull request https://github.com/AOMediaCodec/libavif/pull/459, I suggested Yuan Tong remove the unused fields rgbDepth and yuvMaxChannelF of the avifReformatState struct. I just discovered that the avifReformatState struct is part of the public API, so the removal of those two fields from avifReformatState breaks backward compatibility.

I suggest that we remove avifReformatMode, avifReformatState, and avifPrepareReformatState() from the public API. They are only used in src/reformat.c.

PR #515, opened 2021-02-20
Remove avifPrepareReformatState() from public API

wantehchang:
Remove the avifReformatMode enum, the avifReformatState struct, and the
avifPrepareReformatState() function from the public API. They are only
used in src/reformat.c. (To my surprise, they are not even used in
src/reformat_libyuv.c.)

Fix https://github.com/AOMediaCodec/libavif/issues/514.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -507,49 +507,6 @@ AVIF_API int avifFullToLimitedUV(int depth, int v);
 AVIF_API int avifLimitedToFullY(int depth, int v);
 AVIF_API int avifLimitedToFullUV(int depth, int v);
 
-typedef enum avifReformatMode
-{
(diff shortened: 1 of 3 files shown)


joedrago:
Why is this PR changing libavif's version?

wantehchang:
Because this PR removes two public types and a public function. Should I just bump the AVIF_VERSION_DEVEL in this PR and let you change the version when you create the next release?

joedrago:
Just leave the versions alone (including DEVEL, which is intended to be boolean in nature), and I'll bump them on the next release, yes.

DEVEL helps people distinguish upcoming builds (against `master`), but `master` is allowed to be in flux in these kinds of ways. I should probably rename the branch something more meaningful like `next`, `upcoming`, or `future`, honestly.

wantehchang:
I changed the version back. Thanks!

I didn't realize DEVEL is intended to be boolean. Would be good to document that.

PR #517, opened 2021-02-21
Fail on reformat Identity with subsampling, make avifenc correct this before reformat

tongyuantongyu:
According to [AV1 Spec](https://aomediacodec.github.io/av1-spec/#color-config-semantics), Identity is only valid with YUV444:

But `avifImageRGBToYUV` still tries to do the reformat. avifenc tries to correct this, but after reformat, which will produce an valid image but with wrong color.

This PR let `avifImageRGBToYUV` failed on `image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY && image->yuvFormat != AVIF_PIXEL_FORMAT_YUV444` so that libavif won't produce corrupted file. This PR also make avifenc correct MC before reformat so that the output image has correct color.

I'm still thinking should we change MC or yuvFormat in this case? If user wants lossless, then maybe we should change yuvFormat instead, which makes output lossless as requested.

By the way, if we try to make an Identity monochrome image (forbid by spec), libavif, Chrome and Firefox all wouldn't reject decoding it. However libavif and Firefox will copy the Y(G) value to U(B) and V(R), but Chrome will fill U(B) and V(R) with 128.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -728,6 +728,18 @@ int main(int argc, char * argv[])
     image->yuvRange = requestedRange;
     image->alphaPremultiplied = premultiplyAlpha;
 
+    if ((image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) && (input.requestedFormat != AVIF_PIXEL_FORMAT_YUV444)) {
+        // matrixCoefficients was likely set to AVIF_MATRIX_COEFFICIENTS_IDENTITY as a side effect
+        // of --lossless, and Identity is only valid with YUV444. Set this back to the default.
+        image->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601;
+
+        if (cicpExplicitlySet) {
+            // Only warn if someone explicitly asked for identity.
+            printf("WARNING: matrixCoefficients may not be set to identity (0) when subsampling. Resetting MC to defaults (%d).\n",
+                   image->matrixCoefficients);
+        }
+    }
+
     avifInputFile * firstFile = avifInputGetNextFile(&input);
     uint32_t sourceDepth = 0;
     avifAppSourceTiming firstSourceTiming;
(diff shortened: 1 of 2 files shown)


src/reformat.c
```diff
@@ -44,6 +44,9 @@ avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage *
     state->mode = AVIF_REFORMAT_MODE_YUV_COEFFICIENTS;
 
     if (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
+        if (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV444) {
+            return AVIF_FALSE;
+        }
```
wantehchang, line 49: Nit/optional: I suggest we check for this invalid combination earlier, before doing any work, i.e., before line 42.

apps/avifenc.c
```diff
@@ -727,6 +727,18 @@ int main(int argc, char * argv[])
...
+        // of --lossless, and Identity is only valid with YUV444. Set this back to the default.
+        image->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601;
+
+        if (cicpExplicitlySet) {
+            // Only warn if someone explicitly asked for identity.
+            printf("WARNING: matrixCoefficients may not be set to identity (0) when subsampling. Resetting MC to defaults (%d).\n",
+                   image->matrixCoefficients);
+        }
+    }
+
```
wantehchang, line 741: IMPORTANT: This change is incorrect. 'image' is just created by an avifImageCreateEmpty() call at line 721. So at line 730, image->yuvFormat is still the default value set by the avifImageCreateEmpty() call.

wantehchang (changes requested):
Hi Yuan Tong,

The change to avifenc.c is incorrect. I understand the problem you were trying to fix. I don't have a good suggestion.

apps/avifenc.c
tongyuantongyu: Done
I changed to check `input.requestedFormat`, which holds the value specified by user. However y4m input doesn't use this setting so we have to check again. There's still the case that a yuv444 y4m input with `--cicp 1/13/0 -y 420` will make avifenc wrongly change the MC, but as the user is doing wrong things I think it's OK.

apps/avifenc.c
```diff
@@ -739,6 +751,16 @@ int main(int argc, char * argv[])
     }
     avifBool sourceWasRGB = (inputFormat != AVIF_APP_FILE_FORMAT_Y4M);
 
+    // check again for y4m input (y4m input decide yuv format by file but not command arguments)
```
wantehchang, line 754: I don't understand this change. By the time we get here, the avifInputReadImage() call at line 746 has already called avifImageRGBToYUV() using image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY, so the damage is already done. Isn't it too late to change image->matrixCoefficients to AVIF_MATRIX_COEFFICIENTS_BT601?

wantehchang: I see you already described this change:

`I changed to check input.requestedFormat, which holds the value specified by user. However y4m input doesn't use this setting so we have to check again. There's still the case that a yuv444 y4m input with --cicp 1/13/0 -y 420 will make avifenc wrongly change the MC, but as the user is doing wrong things I think it's OK.`

I think the right thing to do in this case is to exit the program with an error message, rather than encoding the image with an incorrect MC.

Also, this seems like a pre-existing problem, not introduced by your PR. You merely moved this block of code forward a few lines.

tongyuantongyu: Done

wantehchang (approved):
LGTM. I edited a comment slightly. Thanks!

PR #518, opened 2021-02-21
Ext: Update to dav1d 0.8.2

Jamaika1:


EwoutH:
You were a bit early, before the 0.8.2 tag was actually placed, which is why the CI fails. Please also update the [changelog](https://github.com/AOMediaCodec/libavif/blob/master/CHANGELOG.md), looks good otherwise.

wantehchang (approved):
LGTM. Thanks!

PR #519, opened 2021-02-21
Added definition of AOM

Jamaika1:


joedrago:
Can you provide reasoning for this change, please?

This seems redundant as the build system won't even build this file if you don't try to enable libaom. If you do enable libaom, you always want this function. This feels like you are trying to accommodate a used case in which somebody wants to build this file but not use either the encoder or decoder from libaom?

Jamaika1:
It is a matter of taste.
When I'm going to use rav1e only.
```
codec_aom.c: In function 'avifCodecVersionAOM':
codec_aom.c:726:12: warning: implicit declaration of function 'aom_codec_version_str' [-Wimplicit-function-declaration]
  726 |     return aom_codec_version_str();
      |            ^~~~~~~~~~~~~~~~~~~~~
codec_aom.c:726:12: warning: returning 'int' from a function with return type 'const char *' makes pointer from integer without a cast [-Wint-conversion]
  726 |     return aom_codec_version_str();
 ```

joedrago:
If you are only going to use rav1e, why build this file at all? Are you not using cmake?

1480c1:
> If you are only going to use rav1e, why build this file at all? Are you not using cmake?

From what I remember, his preferred way of building is globbing then just passing them to gcc or g++

Jamaika1:
Hi 1480c1
I've been compiling libwebp2 with libavif in gcc 10.2.0 lately. Not everything works fine for me, but that's another thing. Libavif doesn't need to include all functions with libwepb2, hence my whining.
I use manual functions in gcc as an amateur.

wantehchang:
Jamaika1: Thank you for the pull request. The build system should not compile libavif/src/codec_aom.c if libaom is not used. codec_aom.c assumes libaom is used, just like codec_rav1e.c assumes rav1e is used and codec_dav1d.c assumes dav1d is used.

Joe: I suggest we decline this pull request.

PR #523, opened 2021-02-23
Check out libaom v2.1.0-rc1

wantehchang:


joedrago:
What is the typical lead time from an `rc1` to the actual release? Is it long enough that we wouldn't just wait?

wantehchang:
The lead time is usually a week or two. The question is: do we want the users of the libavif master branch (which include ourselves) to help the libaom team test their release candidate?

Please feel free to decline this, for example, if there is a risk for this to make it into a libavif tagged release accidentally.

wantehchang:
I abandoned this pull request.

PR #525, opened 2021-02-25
Add MA1B brand for 8 or 10-bit monochrome images

wantehchang:
See the following sections in the AV1 Specification, Version 1.0.0 with
Errata 1:

Section 6.4.1. General sequence header OBU semantics, pages 112-113. The
table for seq_profile, bit depth, monochrome support, and chroma
subsampling.

A.2. Profiles, page 638. The Note says:
  Note: The Main profile supports YUV 4:2:0 or monochrome bitstreams
  with bit depth equal to 8 or 10. The High profile further adds support
  for 4:4:4 bitstreams with the same bit depth constraints. ...


joedrago:
The code change is straightforward and reasonable. If @cconcolato approves, I'll merge it in.

cconcolato (commented):
The old code or the new code might generate false positives. There are several constraints that make an encode out of profile. For example, even if 420 but using level 6.0, it cannot be tagged with the Baseline profile. Also why do we switch based on image metadata and not on bitstream information (seq_profile, seq_level_idx, ...)?

wantehchang:
Cyril is right. I abandoned this pull request.

PR #526, opened 2021-02-26
Library libavif with tune butteraugli

Jamaika1:
Library libavif with new aom 2.0.2-8de6aba + tune butteraugli for jpeg + integrated jpegxl

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -344,6 +344,9 @@ struct aomOptionDef
 static const struct aomOptionEnumList tuningEnum[] = { //
     { "psnr", AOM_TUNE_PSNR },                         //
     { "ssim", AOM_TUNE_SSIM },                         //
+#if CONFIG_TUNE_BUTTERAUGLI
+    { "butteraugli", AOM_TUNE_BUTTERAUGLI },           //
+#endif
     { NULL, 0 }
 };
 
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -109,7 +109,7 @@ static void syntax(void)
         printf("    enable-chroma-deltaq=B            : Enable delta quantization in chroma planes (0: disable (default), 1: enable)\n");
         printf("    end-usage=MODE                    : Rate control mode (vbr, cbr, cq, or q)\n");
         printf("    sharpness=S                       : Loop filter sharpness (0-7, default: 0)\n");
-        printf("    tune=METRIC                       : Tune the encoder for distortion metric (psnr or ssim, default: psnr)\n");
+        printf("    tune=METRIC                       : Tune the encoder for distortion metric (butteraugli, psnr or ssim, default: psnr)\n");
         printf("    film-grain-test=TEST              : Film grain test vectors (0: none (default), 1: test-1  2: test-2, ... 16: test-16)\n");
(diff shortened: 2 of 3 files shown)


src/codec_aom.c
```diff
@@ -344,6 +344,9 @@ struct aomOptionDef
 static const struct aomOptionEnumList tuningEnum[] = { //
     { "psnr", AOM_TUNE_PSNR },                         //
     { "ssim", AOM_TUNE_SSIM },                         //
+#if CONFIG_TUNE_BUTTERAUGLI
```
wantehchang, line 347: Jamaika1: Thank you for the pull request. This test assumes the build system passes -DCONFIG_TUNE_BUTTERAUGLI=1 on the compiler command line. The official libavif build system doesn't do this.

Here is the libaom enum type in question:

```
typedef enum {
  AOM_TUNE_PSNR = 0,
  AOM_TUNE_SSIM = 1,
  /* NOTE: enums 2 and 3 unused */
  AOM_TUNE_VMAF_WITH_PREPROCESSING = 4,
  AOM_TUNE_VMAF_WITHOUT_PREPROCESSING = 5,
  AOM_TUNE_VMAF_MAX_GAIN = 6,
  AOM_TUNE_VMAF_NEG_MAX_GAIN = 7,
  AOM_TUNE_BUTTERAUGLI = 8,
} aom_tune_metric;
```

Ideally we want libaom's public headers to provide a macro that we can test for the presence of AOM_TUNE_BUTTERAUGLI in the aom_tune_metric enum type. I don't think this was done when AOM_TUNE_BUTTERAUGLI was added to the aom_tune_metric enum type. I will look into this.

An alternative is to use the numeric value of AOM_TUNE_BUTTERAUGLI instead, like so:

    { "butteraugli",  8},           // AOM_TUNE_BUTTERAUGLI = 8

wantehchang (changes requested):

src/codec_aom.c
Jamaika1: av1enc.exe -q 100 -420 -tune BUTTERAUGLI -size 1563x1558 -effort 5 -thread 4 image_21447_24bit.png -d image_21447_24bitx.av1
OK
test libwebp2 -> libavif -> aom
av1enc.exe -avif -q 100 -420 -tune BUTTERAUGLI -size 1563x1558 -effort 5 -thread 4 image_21447_24bit.png -d image_21447_24bitw.avif
```
ERROR: Failed to encode: Encoding of color planes failed
ERROR: Failed to encode: No AV1 items found
Error while computing the distortion.
```

joedrago: I'm unfamiliar with `av1enc` (it appears to be gstreamer). Does that use libavif internally if `-avif` is supplied on the commandline?

Jamaika1: This is the libwebp 2 program that uses libavif for google products.
I was unable to compile the avif file from libavif
https://www.sendspace.com/file/dzkvri
I was able to compile an avif file from libheif / avif 
https://github.com/strukturag/libheif/pull/448

wantehchang:
This pull request has been replaced by https://github.com/AOMediaCodec/libavif/pull/567.

PR #529, opened 2021-02-27
Support reading image using WIC on Windows

tongyuantongyu:
Add support to use WIC API to support reading from any image formats Windows supported, including JPEG, PNG (these two are not used by us), TIFF, BMP, GIF and some others. It's also extensible so if with addon Windows is able to open it we are also able to read from it. (which means we also supports reading from WEBP, HEIF and even AVIF ourselves)

This PR only supports decoding, and only reads the first frame from an input file. In the future we may explore encoding using WIC to get rid of complicated dependency configuration of libjpeg and libpng on Windows, and also multiple-frame input support.

### DIFF
--- a/apps/shared/avifwic.c
+++ b/apps/shared/avifwic.c
@@ -1,0 +1,357 @@
+// Copyright 2021 Yuan Tong. All rights reserved.
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "avifwic.h"
+
+#ifdef AVIF_WINCODEC_ENABLED
+#include <stdio.h>
+#define COBJMACROS
+#include <wincodec.h>
+#include <windows.h>
+
+#define AVIF_MAKE_GUID(name, l, w1, w2, b1, b2, b3, b4, b5, b6, b7, b8) \
+    {                                                                   \
+        l, w1, w2, { b1, b2, b3, b4, b5, b6, b7, b8 }                   \
+    }
+#define AVIF_DEFINE_GUID(name, l, w1, w2, b1, b2, b3, b4, b5, b6, b7, b8) \
+    static const GUID AVIF_##name = AVIF_MAKE_GUID(name, l, w1, w2, b1, b2, b3, b4, b5, b6, b7, b8)
+
+// Pixel formats that has 10 bit precision. Make our own copy for compatibility with older SDK.
+static const GUID avifWIC10BitDepthGUIDsTable[] = {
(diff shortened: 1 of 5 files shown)


apps/shared/avifwic.c
```diff
@@ -0,0 +1,361 @@
...
+    if (pFactory != NULL) {
+        IUnknown_Release(pFactory);
+    }
+    if (f != INVALID_HANDLE_VALUE) {
+        CloseHandle(f);
+    }
+    return ret;
+}
+#else
+avifBool avifWICRead(const char * inputFilename, avifImage * avif, avifPixelFormat requestedFormat, uint32_t requestedDepth)
```
Jamaika1, line 353: avifBool avifWICRead(const char * inputFilename, avifImage * avif, avifPixelFormat requestedFormat, uint32_t requestedDepth, uint32_t * outDepth)

tongyuantongyu: Done

suphamster:
Any plans to finally add this in avifenc since for now it's not possible directly convert webp to avif with it?

juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **2 points that might be worth attention** (could be false positives, please use your judgment):

1. <strong>The ICC profile data buffer could be managed more efficiently to ensure resources are released</strong>
   - [Location](https://github.com/AOMediaCodec/libavif/blob/c1faa8fd1869d7b87abac0848ddf67833921edc2/apps/shared/avifwic.c#L248-L255): `apps/shared/avifwic.c:L248-L255`
   - Impact: Allocated memory for ICC data might persist after it is copied or when errors occur, potentially increasing memory usage during batch tasks.
   - Suggestion: Consider adding `avifFree(iccData)` after `avifImageSetProfileICC` and within the error handling block to ensure the buffer is cleared.

2. <strong>The intermediate RGB pixel buffer appears to remain in memory after the image conversion process</strong>
   - [Location](https://github.com/AOMediaCodec/libavif/blob/c1faa8fd1869d7b87abac0848ddf67833921edc2/apps/shared/avifwic.c#L326-L350): `apps/shared/avifwic.c:L326-L350`
   - Impact: Unreleased pixel buffers could accumulate over time, which might lead to higher memory pressure when processing image sequences.
   - Suggestion: You might want to call `avifRGBImageFreePixels(&rgb)` in the cleanup section to ensure the allocated pixels are properly reclaimed.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

Issue #520, opened 2021-02-21
Behavior on reformating image with alpha to format without alpha


tongyuantongyu:
Make this an issue to discuss as @wantehchang suggested.

Assuming the color is not premultiplied by alpha, when doing such conversion, there are two possible choices: discard alpha directly and multiply color by alpha first. Here is the difference between the two choices (notice the boundary):
| Original | Discard | Multiply |
| --- | --- | --- |
| ![chrome-l](https://user-images.githubusercontent.com/13075180/108618192-f0bf2400-7456-11eb-9b6b-2bde11313fda.png) | ![chrome-lnp](https://user-images.githubusercontent.com/13075180/108618194-f288e780-7456-11eb-9213-ed1b4cdf5960.jpg) | ![chrome-lp](https://user-images.githubusercontent.com/13075180/108618195-f3217e00-7456-11eb-935f-3930df3121a5.jpg) |

Currently libavif takes the 'Discard' choice.
Now we supported premultiply alpha, which is effectively the same as 'Multiply', so there are 3 choices now:

1. Naive: keep as is
 This is the current state. Results of images with premultiplied and not premultiplied alpha will be different, even if the images looks the same.

2. Natural: always multiply
 In this choice, we ensure color is multiplied by alpha. The result is the same as displaying the source image on black background.

3. Consistent: always not multiply
 In this choice, we ensure color is not multiplied by alpha. This keeps the old behavior even on images with premultiplied alpha, but the result looks unnatural and is probably not very useful.

I personally prefer 2.

joedrago:
This only impacts the JPEG writer, correct? I'm on board with 2 if that is the only place we need to handle this.

joedrago:
Actually, if it only affects the JPEG writer, then it could easily be a setting in `avifdec`. How about we make the default 2, but offer a way to avoid it on the commandline?

tongyuantongyu:
I think let user make the decision is a good idea. Is 'avoid' meaning 1 or 3?

PR #530, opened 2021-03-01
Multiply color with alpha for opaque RGB format during conversion

tongyuantongyu:
Implements #520.


joedrago:
This commit has been accepted with changes, but GitHub isn't catching it:

https://github.com/AOMediaCodec/libavif/commit/d4f317bbaf104ca00b546d9b35d7e73bafdc129b

and my additional fixes here:

https://github.com/AOMediaCodec/libavif/commit/dfa2f0a31959f6e66ba0825d9e1418f4d805968e

apps/avifdec.c
```diff
@@ -225,6 +229,10 @@ int main(int argc, char * argv[])
                 returnCode = 1;
             }
         } else if (outputFormat == AVIF_APP_FILE_FORMAT_JPEG) {
+            // Bypass alpha multiply step during conversion
```
wantehchang, line 232: Hi Tong Yuan,

The help message says raw color is always true for y4m. If so, shouldn't we always set avif->alphaPremultiplied to AVIF_TRUE between lines 227 and 228?

tongyuantongyu: `avif->alphaPremultiplied` only affects the conversion between RGB and YUV, and since y4m stores YUV data, this value is unrelated.

PR #531, opened 2021-03-02
Disable key frame filtering when encoding images

wantehchang:
Disable key frame filtering when encoding still images.


wantehchang:
Joe: Please do not merge this pull request until I have tested it.

wantehchang:
Joe: This pull request is ready for review. This is recommended by my coworker Jingning Han. Strictly speaking this is not necessary; this is for extra assurance that temporal filtering won't be performed on key frames.

PR #532, opened 2021-03-05
libavif.pc: respect libdir setting

vapier:
Do not hardcode "lib" as that is often the wrong path with multilib.
On an x86_64 system for example, it should actually be "lib64".


wantehchang (approved):
LGTM. Thanks!

PR #537, opened 2021-03-11
Bail out of avifDecoderParse() if an essential-required item property is not flagged as essential

joedrago:
This should ensure that all `av1C` property associations are flagged as `essential`. 

@wantehchang @cconcolato are there other properties we should enforce this way, while we're at it?

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1481,6 +1481,23 @@ static avifBool avifParseItemPropertyAssociation(avifMeta * meta, const uint8_t 
                 }
             }
             if (supportedType) {
+                if (!essential) {
+                    // Verify that it is legal for this property to not be flagged as essential. Any
+                    // types in this list are *required* in the spec to be flagged as essential when
+                    // associated with an item.
+                    static const char * essentialTypes[] = {
+                        "av1C" // AVIF: Section 2.2.1: "This property shall be marked as essential."
+                    };
+                    size_t essentialTypesCount = sizeof(essentialTypes) / sizeof(essentialTypes[0]);
+                    for (size_t i = 0; i < essentialTypesCount; ++i) {
+                        if (!memcmp(srcProp->type, essentialTypes[i], 4)) {
+                            // An essential-required property is not flagged as essential, bail out
+                            return AVIF_FALSE;
+                        }
+                    }
+                }
+
+                // Supported and valid; associate it with this item.
                 avifProperty * dstProp = (avifProperty *)avifArrayPushPtr(&item->properties);
                 memcpy(dstProp, srcProp, sizeof(avifProperty));
             } else {


src/read.c
```diff
@@ -1481,6 +1481,23 @@ static avifBool avifParseItemPropertyAssociation(avifMeta * meta, const uint8_t
                 }
             }
             if (supportedType) {
+                if (!essential) {
+                    // Verify that it is legal for this property to not be flagged as essential. Any
+                    // types in this list are *required* in the spec to be flagged as essential when
+                    // associated with an item.
+                    static const char * essentialTypes[] = {
```
wantehchang, line 1488: This should have another const:
`                    static const char * const essentialTypes[] = {`

We can make the same change to line 1474:
            static const char * const supportedTypes[] = { "ispe", "auxC", "colr", "av1C", "pasp", "clap", "irot", "imir", "pixi" };

src/read.c
```diff
@@ -1481,6 +1481,23 @@ static avifBool avifParseItemPropertyAssociation(avifMeta * meta, const uint8_t
...
+                        "av1C" // AVIF: Section 2.2.1: "This property shall be marked as essential."
```
baumanj, line 1489:
```suggestion
                        "a1op" // AVIF: Section 2.3.2.1: "If associated, it shall be marked as essential."
```
Not sure if `a1op` needs to be handled differently here.

Also, should there be an opposite check for `a1lx` since the spec says "[If associated, it shall not be marked as essential](https://aomediacodec.github.io/av1-avif/#layered-image-indexing-property-description)"

wantehchang, line 1489: We can add the following:

`"lsel",  // HEIF: Section 6.5.11.1: "essential shall be equal to 1 for an 'lsel' item property."`

Also note the following from MIAF, which I don't know how to interpret:

> 7.3.9 Transformations and derived items
> The image grid, overlay, and identity derivations shall be supported by the MIAF reader.
> All transformative properties associated with coded and derived images required or conditionally required by this document shall be marked as essential, and shall be from the set that are permitted by this document or the applicable profile. No other essential transformative property shall be associated with such images.

wantehchang (changes requested):

joedrago:
One issue with this is any file created with a sufficiently old `avifenc` (prior to `v0.7.2`) will break with this change, as av1C weren't being properly flagged as essential until this commit:

https://github.com/AOMediaCodec/libavif/commit/0da2997ed287014f543efdc6c42f53a5fb7dc22a

This means I have to fix the IO data checked into `tests` and oss-fuzz's seed corpus.

wantehchang:
I see. Let's also check the AVIF encoder in libheif.

wantehchang:
The libheif function to inspect is `Box_ipma::add_property_for_item_ID()`, declared in libheif/box.h:

```
  class Box_ipma : public Box
  {
  public:
    ...

    struct PropertyAssociation
    {
      bool essential;
      uint16_t property_index;
    };

    ...

    void add_property_for_item_ID(heif_item_id itemID,
                                  PropertyAssociation assoc);
```

The relevant libheif code is in libheif/heif_file.cc:

```
void HeifFile::add_av1C_property(heif_item_id id)
{
  auto av1C = std::make_shared<Box_av1C>();
  int index = m_ipco_box->append_child_box(av1C);

  m_ipma_box->add_property_for_item_ID(id, Box_ipma::PropertyAssociation{true, uint16_t(index + 1)});
}
```

The code was originally added in this commit, with `essential` set to true:
https://github.com/strukturag/libheif/commit/105928035e46eb7bca8100c508d7e457f572c5c7

joedrago:
I've submitted a PR to `oss-fuzz` to correct the seed corpus. It is the same content, but with updated shasums due to the updated essential bit being set for av1C properties.

joedrago:
I think this is a good change in the longrun, but I admit to being a bit concerned that there will be some "old" AVIFs (generated with avifenc prior to `v0.7.2`) that will "break". 

Perhaps before rolling this out (or alongside it), we should offer some kind of cheap fixup tool that toggles the correct essential bit in "old" files?

joedrago:
I've created a standalone script to fix old AVIFs with this issue. Should I check this into libavif somewhere?

[fix_essential.zip](https://github.com/AOMediaCodec/libavif/files/6126335/fix_essential.zip)

src/read.c
joedrago: Hmm. I'm thinking about removing `lsel` from the list for now and just adding boxes to this list as I add support for new boxes. `lsel` isn't written or read at all right now, so it seems odd to preemptively block a read on it.

Issue #538, opened 2021-03-11
Use libaom's new all intra mode


wantehchang:
I am going to upload a pull request that uses libaom's new all intra mode (added in https://crbug.com/aomedia/2959) if it is available at compile time.

The all intra mode is intended for encoding still images and all intra videos. It is only available on the master branch of libaom right now. I provide sample encoding times and compressed image sizes using three versions of libaom.
- v2.0.2 + the bug fix https://aomedia-review.googlesource.com/c/aom/+/132881: 45.3 seconds. 1130337 bytes.
- v2.1.0-rc1: 23.1 seconds. 1134011 bytes.
- the tip of the master branch (commit 9cd916ae1a704e56d60c11e7fcf66862e46545a1) + the all intra mode patch: 10.3 seconds. 1057324 bytes.

The avifenc command line is:
`avifenc --speed 6 happy_dog.jpg happy_dog.avif --jobs 1 -y 420 --min 0 --max 63 -a end-usage=q -a cq-level=18`

I used libavif speed 6, single-threaded encoding. The happy_dog.avif image has dimensions 4032x3024.

PR #539, opened 2021-03-11
Use libaom's new all intra mode

wantehchang:
Use libaom's new all intra mode (added in
https://crbug.com/aomedia/2959) if it is available at compile time. The
all intra mode is intended for encoding still images and all intra
videos.

Fix https://github.com/AOMediaCodec/libavif/issues/538.


wantehchang:
Yes, it it acceptable to set `cfg.g_lag_in_frames` to 0 in builds that do not have `AOM_USAGE_ALL_INTRA`.

The encoder config settings `cfg.g_lag_in_frames`. `cfg.kf_mode` , and `cfg.kf_max_dist` are not new. They are present in all versions of libaom.

PR #543, opened 2021-03-16
Warn about using end-usage=q with libaom <= 2.0.2

wantehchang:
Warn about using the aom-specific option end-usage=q with libaom 2.0.2
or older because of https://crbug.com/aomedia/2725. The compressed image
size may be twice as big!


Jamaika1:
Maybe use
```
#if defined(AVIF_CODEC_AOM_ENCODE) || defined(AVIF_CODEC_AOM_DECODE)
#include "config/aom_config.h"
#include "config/aom_version.h"
#endif
```
```
#if defined(VERSION_PACKED) && (VERSION_PACKED >= 020200)
            if (tempBuffer = "end-usage") {
                fprintf(stderr, "Do not use end-usage=q with libaom 2.0.2 or older.\n", arg);
                returnCode = 1;
                goto cleanup;
            }
#endif
```

wantehchang:
Hi Jamaika1,

Thank you for the suggestion. We will need to do a run-time version check. (Also the header files your code snippet includes are private headers.) I will abandon this pull request and look into either a run-time version check or a workaround for https://crbug.com/aomedia/2725.

PR #547, opened 2021-03-18
Merge two if statements with the same body

wantehchang:


joedrago:
This one is borderline for me (and definitely subjective), as I typically don't like chaining multiple expressions in a conditional when some of them are function calls. It sometimes can hurt debuggability (stepping). I do like the removal of the duplicated block though. Perhaps a simple fix would be something like:

    const avifBool finishSuccess = item->codec->encodeFinish(item->codec, item->encodeOutput);
    if(!finishSuccess || ...) { ... }
    
Again, this is purely subjective / style, and I'm willing to be convinced otherwise. Thoughts?

wantehchang:
Hi Joe,

I didn't know you prefer this style. I will abandon this pull request.

PR #549, opened 2021-03-19
Clean up the code that checks imageCountLimit

wantehchang:
Add the avifGetSampleCountOfChunk() function for the code that was
duplicated.

Declare the imageCountLimit parameter of
avifCodecDecodeInputGetSamples() to have the uint32_t type (because
decoder->imageCountLimit has the uint32_t type) and replace the uint64_t
variable imageCount with the uint32_t variable imageCountLeft. It is not
necessary to use the uint64_t type in this code.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -329,35 +329,42 @@ void avifCodecDecodeInputDestroy(avifCodecDecodeInput * decodeInput)
     avifFree(decodeInput);
 }
 
+// Returns how many samples are in the chunk.
+static uint32_t avifGetSampleCountOfChunk(const avifSampleTableSampleToChunkArray * sampleToChunks, uint32_t chunkIndex)
+{
+    uint32_t sampleCount = 0;
+    for (int sampleToChunkIndex = sampleToChunks->count - 1; sampleToChunkIndex >= 0; --sampleToChunkIndex) {
+        const avifSampleTableSampleToChunk * sampleToChunk = &sampleToChunks->sampleToChunk[sampleToChunkIndex];
+        if (sampleToChunk->firstChunk <= (chunkIndex + 1)) {
+            sampleCount = sampleToChunk->samplesPerChunk;
+            break;
+        }
+    }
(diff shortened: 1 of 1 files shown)


wantehchang:
Joe: Please see if the function name "avifGetSampleCountOfChunk" conforms to libavif's naming convention. Thanks!

joedrago:
I like the reorganization here, but you switched `imageCountLimit` and `imageCount` back to `uint32_t`, which I had recently promoted to `uint64_t` to avoid overflow issues on this line:

    imageCount += sampleCount;

I should have added a clarifying comment here. Which direction should we go here? Explicitly check for overflow before this arithmetic, or put those two variables back to `uint64_t` with a comment?

wantehchang:
There is more than one way to avoid the overflow. The approach I used in this pull request is to subtract after checking is safe to subtract:

```
            if (sampleCount > imageCountLeft) {
                // This file exceeds the imageCountLimit, bail out
                return AVIF_FALSE;
            }
            imageCountLeft -= sampleCount;
```

This requires using a new variable `imageCountLeft` that is initialized to `imageCountLimit`.

Alternatively I can use the following approach, which maintains the invariant `imageCount <= imageCountLimit` and is more similar to the original code:

```
        uint32_t imageCount = 0;
        for (uint32_t chunkIndex = 0; chunkIndex < sampleTable->chunks.count; ++chunkIndex) {
            // First, figure out how many samples are in this chunk
            uint32_t sampleCount = avifGetSampleCountOfChunk(&sampleTable->sampleToChunks, chunkIndex);
            ...

            if (sampleCount > imageCountLimit - imageCount) {
                // This file exceeds the imageCountLimit, bail out
                return AVIF_FALSE;
            }
            imageCount += sampleCount;
        }
```

Would you prefer this instead?

joedrago:
Oh, I missed the new technique (`imageCountLeft`), sorry. This is perfectly acceptable.

PR #551, opened 2021-03-20
In libaom all intra mode, set cq-level for user

wantehchang:
If libaom's new all intra mode is used, the rc_end_usage encoder setting
defaults to AOM_Q, which requires cq-level to function. Since a typical
user is unlikely to know this esoteric internal detail, if the user does
not set cq-level, set it to the average value of the min and max
quantizers for the user.

NOTE: An alternative is to replace the min and max quantizers with a
single quantizer in the avifEncoder struct. This will be an incompatible
change but is worth considering.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -30,6 +30,7 @@
 #pragma clang diagnostic ignored "-Wassign-enum"
 #endif
 
+#include <assert.h>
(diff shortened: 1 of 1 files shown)


jingninghan (commented):
lgtm

Issue #548, opened 2021-03-19
Add more aom-specific advanced options?


linkboyx:
Hi, thanks for your great job! 
I'm now developing on Android with libavif, and hope to speed up the aom-based avif encoder on the phone. I found that some aom-specific options(like enable-angle-delta , ...) could reduce the encoding time (may loss efficiency or decrease qulity, I guess so but not tested yet).  
Since I lack of codec knowledge and have no ideas about which options are critical, so could you please pick  some key options for encoding speed in AOM and add them into avif settings?. I will test them on the tradeoff between efficiency and speed on the phone.
Thanks a lot!

joedrago:
Perhaps @wantehchang can offer some insight on this, but I imagine any option that is useful for images (AVIF) he's already made available.

linkboyx:
Thanks for you reply!
On my desktop, if some aom options were closed, the encoding speed can increase from 1.7 fps to 2.9 fps.   For example, if i set --enable-angle-delta=0 only, time decrease 80ms, but file size increases 6KB. 
I also care about the RAM, but have not found any options could reduce RAM.  But there is a huge RAM reduction from ~500MB(aom 2.0.2) to ~110MB(aom 2.1.0/3.0.0) on a 1080P image.

wantehchang:
Hi linkboyx,

Thank you for your interest in using the libaom-based AVIF encoder on Android. Our team uses the following `avifenc` options to encode AVIF images without alpha channels (please replace `T` and `QP` with suitable values):

`avifenc --speed 6 --jobs T -y 420 --min 0 --max 63 -a end-usage=q -a cq-level=QP happy_dog.jpg happy_dog.avif `

Replace `T` with the number of worker threads you want the encoder to use.

Replace `QP` with a value in the range from 0 (best quality) to 63 (worst quality).

You can also try `-a tune=ssim` with a high quality (i.e., a low `QP` value).

Please use a recent stable snapshot from the master branch of libaom. (The `v3.0.0-rc1` tag does not have the new all-intra encoding mode.) Note that the master branch of libaom is not fully "continuously releasable" because the full test suites are run once a day. So you need to run `test_libaom` on a snapshot that you are considering to use. You can try https://aomedia.googlesource.com/aom/+/41fd15c9c999e03247ba88da2c2f54c54a4c4488.

[...]

PR #567, opened 2021-03-30
Use aom_codec_set_option() if available

wantehchang:
Use the new aom_codec_set_option() function, if available, to set
libaom-specific advanced options.

Fix https://github.com/AOMediaCodec/libavif/issues/548.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -111,6 +111,9 @@ static void syntax(void)
         printf("       enable-chroma-deltaq=B, should not be used with the alpha sub-image. In addition, the film grain options\n");
         printf("       are unlikely to make sense for the alpha sub-image.\n");
         printf("\n");
+        printf("    When used with libaom 3.0.0 or later, any key-value pairs supported by the aom_codec_set_option() function\n");
+        printf("    can be used. When used with libaom 2.0.x or older, the following key-value pairs can be used:\n");
+        printf("\n");
         printf("    aq-mode=M                         : Adaptive quantization mode (0: off (default), 1: variance, 2: complexity, 3: cyclic refresh)\n");
         printf("    cq-level=Q                        : Constant/Constrained Quality level (0-63, end-usage must be set to cq or q)\n");
         printf("    enable-chroma-deltaq=B            : Enable delta quantization in chroma planes (0: disable (default), 1: enable)\n");
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -35,6 +35,12 @@
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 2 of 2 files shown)


wantehchang:
This PR adds the second paragraph to the section on aom-specific advanced options in the syntax help message of avifenc. That section now reads:

```
    1. <key>=<value> applies to both the color sub-image and the alpha sub-image (if present).
    2. color:<key>=<value> or c:<key>=<value> applies only to the color sub-image.
    3. alpha:<key>=<value> or a:<key>=<value> applies only to the alpha sub-image (if present).
       Since the alpha sub-image is a monochrome image, the options that refer to the chrome planes, such as
       enable-chroma-deltaq=B, should not be used with the alpha sub-image. In addition, the film grain options
       are unlikely to make sense for the alpha sub-image.

    When used with libaom 3.0.0 or later, any key-value pairs supported by the aom_codec_set_option() function
    can be used. When used with libaom 2.0.x or older, the following key-value pairs can be used.

    aq-mode=M                         : Adaptive quantization mode (0: off (default), 1: variance, 2: complexity, 3: cyclic refresh)
    cq-level=Q                        : Constant/Constrained Quality level (0-63, end-usage must be set to cq or q)
    enable-chroma-deltaq=B            : Enable delta quantization in chroma planes (0: disable (default), 1: enable)
    end-usage=MODE                    : Rate control mode (vbr, cbr, cq, or q)
    sharpness=S                       : Loop filter sharpness (0-7, default: 0)
    tune=METRIC                       : Tune the encoder for distortion metric (psnr or ssim, default: psnr)
    film-grain-test=TEST              : Film grain test vectors (0: none (default), 1: test-1  2: test-2, ... 16: test-16)
    film-grain-table=FILENAME         : Path to file containing film grain parameters
```

apps/avifenc.c
```diff
@@ -111,6 +111,9 @@ static void syntax(void)
         printf("       enable-chroma-deltaq=B, should not be used with the alpha sub-image. In addition, the film grain options\n");
         printf("       are unlikely to make sense for the alpha sub-image.\n");
         printf("\n");
+        printf("    When used with libaom 3.0.0 or later, any key-value pairs supported by the aom_codec_set_option() function\n");
+        printf("    can be used. When used with libaom 2.0.x or older, the following key-value pairs can be used.\n");
```
joedrago, line 115: Nit: should it be `used:` ?

wantehchang: Done. Please review the new commit. Thanks!

joedrago:
It all appears to build locally with both libaom 2.x and 3.x, so LGTM.

I see that SVT has some additional duplicated symbols with libaom, but that is unrelated to these changes. I'll file an issue about it.

PR #575, opened 2021-04-06
Describe color and alpha as planes, not sub-images

wantehchang:
Also clarify that color planes refer to the YUV planes, not just the
chroma planes.


wantehchang:
The modified part of the avifenc help message now reads:

```
    1. <key>=<value> applies to both the color (YUV) planes and the alpha plane (if present).
    2. color:<key>=<value> or c:<key>=<value> applies only to the color (YUV) planes.
    3. alpha:<key>=<value> or a:<key>=<value> applies only to the alpha plane (if present).
       Since the alpha plane is encoded as a monochrome image, the options that refer to the chroma planes,
       such as enable-chroma-deltaq=B, should not be used with the alpha plane. In addition, the film grain
       options are unlikely to make sense for the alpha plane.
```

joedrago:
This is all reasonable to me. I'll merge it in.

Issue #563, opened 2021-03-27
A HandlerBox (hdlr) of type 'pict' must be the first box within the MetaBox (meta)


baumanj:
Per the [AVIF specification § 6](https://aomediacodec.github.io/av1-avif/#file-constraints):
> The file shall be compliant with the [MIAF] specification

The MIAF ([ISO 23000-22:2019](https://www.iso.org/standard/74417.html)) specification § 7.2.1.5 states:
> The file-level `MetaBox` shall be identified as carrying images by carrying in the associated `HandlerBox` the value required by ISO/IEC 23008-12 (which is `'pict'`).

Additionally § 7.2.1.5 states:
> The `HandlerBox` shall be the first contained box within the `MetaBox`.

The `MetaBox` and `HandlerBox` are defined in ISOBMFF ([ISO 14496-12:2015](https://www.iso.org/standard/68960.html)) § 8.11.1 and 8.4.3 respectively

This can be checked with the [ComplianceWarden AVIF](https://gpac.github.io/ComplianceWarden-wasm/avif.html) validator. Three examples of invalid input are attached:
1. [`HandlerBox` box not present](https://github.com/AOMediaCodec/libavif/files/6216540/no-hdlr.avif.zip)
2. [`HandlerBox` not the first contained box within the `MetaBox`](https://github.com/AOMediaCodec/libavif/files/6216542/hdlr-not-first.avif.zip)
3. [`HandlerBox` value not `pict`](https://github.com/AOMediaCodec/libavif/files/6216541/hdlr-not-pict.avif.zip)

See also https://github.com/mozilla/mp4parse-rust/issues/286

PR #578, opened 2021-04-06
A HandlerBox (hdlr) of type 'pict' must be the first box within the M…

joedrago:
…etaBox (meta)

Fixes: #563

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1108,6 +1108,37 @@ static avifBool isAlphaURN(const char * urn)
 // ---------------------------------------------------------------------------
 // BMFF Parsing
 
+static avifBool avifParseHandlerBox(const uint8_t * raw, size_t rawLen)
+{
+    BEGIN_STREAM(s, raw, rawLen);
+
+    CHECK(avifROStreamReadAndEnforceVersion(&s, 0));
+
+    uint32_t predefined;
+    CHECK(avifROStreamReadU32(&s, &predefined)); // unsigned int(32) pre_defined = 0;
+    if (predefined != 0) {
+        return AVIF_FALSE;
+    }
+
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -1108,6 +1108,37 @@ static avifBool isAlphaURN(const char * urn)
...
+    if (memcmp(handlerType, "pict", 4) != 0) {
+        return AVIF_FALSE;
+    }
+
+    for (int i = 0; i < 3; ++i) {
+        uint32_t reserved;
+        CHECK(avifROStreamReadU32(&s, &reserved)); // const unsigned int(32)[3] reserved = 0;
+        if (reserved != 0) {
+            return AVIF_FALSE;
+        }
```
wantehchang, line 1134: In many systems, encoders must set reserved fields to 0 but decoders are supposed to ignore the reserved fields. If decoders require reserved fields to be 0 and some decoders cannot be upgraded, then we will never be able to use the reserved fields in the future.

wantehchang (commented):
Please see my comment about requiring the reserved fields to be 0. Other than that, this pull request is good.

src/read.c
joedrago: Ah, this is good information. I'm perfectly willing to ignore the reserved value here; I just saw that it was `= 0` in the spec and took that to mean that it was required to be such.

joedrago: I've removed the check.

wantehchang (approved):
LGTM. Thanks!

I don't know if the reserved fields in ISO BMFF are supposed to be handled this way. The reserved fields are not described in the "Semantics" sections in ISO BMFF. (Neither are the pre_defined fields described.)

Issue #564, opened 2021-03-28
Require the pixel information property (pixi) for all displayable images


baumanj:
Per the [AVIF specification § 6](https://aomediacodec.github.io/av1-avif/#file-constraints):
> The file shall be compliant with the [MIAF] specification

The MIAF ([ISO 23000-22:2019](https://www.iso.org/standard/74417.html)) specification § 7.3.6.6 states:
> The pixel information property shall be associated with every image that is displayable (not hidden)

The `PixelInformationProperty` descriptive item property is defined in HEIF [ISO/IEC 23008-12:2017](https://www.iso.org/standard/66067.html) ([free download](https://standards.iso.org/ittf/PubliclyAvailableStandards/c066067_ISO_IEC_23008-12_2017.zip)) § 6.5.6

This can be checked with the [ComplianceWarden AVIF](https://gpac.github.io/ComplianceWarden-wasm/avif.html) validator. Two examples of invalid input are attached:
1. [A file with 1 image and no pixel information property](https://github.com/AOMediaCodec/libavif/files/6216676/no-pixi.avif.zip)
2. [A file with 2 images (YUV and alpha) and only one pixel information property (for YUV)](https://github.com/AOMediaCodec/libavif/files/6216677/no-pixi-for-alpha.avif.zip)

See also https://github.com/mozilla/mp4parse-rust/issues/287

PR #579, opened 2021-04-06
Require the pixel information property (pixi) for all displayable images

joedrago:
Fixes: #564

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -708,6 +708,30 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifExt
     return AVIF_RESULT_OK;
 }
 
+static avifResult avifDecoderItemValidateAV1(const avifDecoderItem * item)
+{
+    const avifProperty * av1CProp = avifPropertyArrayFind(&item->properties, "av1C");
+    if (!av1CProp) {
+        // An av1C box is mandatory in all valid AVIF configurations. Bail out.
+        return AVIF_RESULT_BMFF_PARSE_FAILED;
+    }
+    const uint32_t av1CDepth = avifCodecConfigurationBoxGetDepth(&av1CProp->u.av1C);
+
+    const avifProperty * pixiProp = avifPropertyArrayFind(&item->properties, "pixi");
+    if (!pixiProp) {
+        // A pixi box is mandatory in all valid AVIF configurations. Bail out.
+        return AVIF_RESULT_BMFF_PARSE_FAILED;
+    }
+
+    for (uint8_t i = 0; i < pixiProp->u.pixi.planeCount; ++i) {
+        if (pixiProp->u.pixi.planeDepths[i] != av1CDepth) {
+            // pixi depth must match av1C depth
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -708,6 +708,30 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifExt
...
+    }
+    const uint32_t av1CDepth = avifCodecConfigurationBoxGetDepth(&av1CProp->u.av1C);
+
+    const avifProperty * pixiProp = avifPropertyArrayFind(&item->properties, "pixi");
+    if (!pixiProp) {
+        // A pixi box is mandatory in all valid AVIF configurations. Bail out.
+        return AVIF_RESULT_BMFF_PARSE_FAILED;
+    }
+
+    for (uint8_t i = 0; i < pixiProp->u.pixi.planeCount; ++i) {
```
wantehchang, line 726: We should first validate pixiProp->u.pixi.planeCount.

pixiProp->u.pixi.planeCount should be equal to av1CProp->u.av1C.monochrome ? 1 : 3.

src/read.c
```diff
@@ -708,6 +708,30 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifExt
     return AVIF_RESULT_OK;
 }
 
+static avifResult avifDecoderItemValidateAV1(const avifDecoderItem * item)
```
wantehchang, line 711: Nit/optional: Change "AV1" to "AV1C"?

wantehchang (changes requested):

src/read.c
joedrago: The goal of the function is to validate an item that is backed by an AV1 payload. This function checks more than just the `av1C` block (such as `pixi`).

src/read.c
joedrago: The wording on pixi (HEIF 6.5.6.3) is:

> num_channels: This field signals the number of channels by each pixel of the reconstructed image of
the associated image item.

For a long time, the primary encoders out there (libaom & rav1e) were incapable of encoding proper YUV400 for various reasons, so it had to fall back to encoding YUV420, even if it was only going to use the Y plane (such as an alpha plane).

For `av1C`, the bits of this structure are clear: they should match the actual encoded data as the intent of this box is to initialize hardware encoders. However, if the "reconstructed image" is an alpha plane (say), should the two meaningless chroma planes that come out of the AV1 decoder be mentioned in the `pixi` block?

Maybe it is more of a philosophical question - Does a decoded ("reconstructed") alpha plane that was encoded as YUV420 now have 3 channels? Or is it still just an alpha plane?

I'm willing to be convinced otherwise (and I haven't tested it), but I suspect this would be a pretty heavy breaking change for alpha planes when they were encoded with semi-incomplete AV1 encoders.

wantehchang: Ah, I forgot that problem. Maybe just check the color image? Or just require pixiProp->u.pixi.planeCount is 1 or 3?

wantehchang (approved):

PR #581, opened 2021-04-07
Clarify the help message for -r,--raw-color

wantehchang:
Explicitly say the raw values are RGB, which helps explain why this
option is not applicable to Y4M.


wantehchang:
Here are the help messages for -r,--raw-color.

Before:

```
    -r,--raw-color    : Output raw values instead of multiplying alpha when saving to opaque formats (JPEG only; always true for y4m)
```

After:

```
    -r,--raw-color    : Output raw RGB values instead of multiplying by alpha when saving to opaque formats
                        (JPEG only; not applicable to y4m)
```

PR #591, opened 2021-04-08
Warn if alpha is limited range (deprecated)

wantehchang:
NOTE: The alpha range is not available when the --info option is used.
So this warning is emitted only when the image is decoded.


wantehchang:
When invoked on an AVIF image with limited-range alpha, the warning looks like this:

```
$ ./avifdec red-with-limited-alpha-8bpc.avif output.png
    idat->id = meta->idatID;
Image decoded: red-with-limited-alpha-8bpc.avif
Image details:
 * Resolution     : 3x3
 * Bit Depth      : 8
 * Format         : YUV420
 * Alpha          : Not premultiplied
                    Limited range
                    WARNING: Limited-range alpha is deprecated. Use full-range alpha instead.
 * Range          : Full
 * Color Primaries: 1
 * Transfer Char. : 13
 * Matrix Coeffs. : 1
 * ICC Profile    : Absent (0 bytes)
 * XMP Metadata   : Absent (0 bytes)
 * EXIF Metadata  : Absent (0 bytes)
 * Transformations:
    * pasp (Aspect Ratio)  : 1/1
Wrote PNG: output.png
```

Issue #587, opened 2021-04-07
Guard the avifImageYUVToRGBLibYUV() call with alphaMultiplyMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP || avifRGBFormatHasAlpha(rgb->format)


wantehchang:
In src/reformat.c, the following code in `avifImageYUVToRGB()` is guarded with the conditional `alphaMultiplyMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP || avifRGBFormatHasAlpha(rgb->format)`:

```
        if ((!hasColor || (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV444) || (chromaUpsampling == AVIF_CHROMA_UPSAMPLING_NEAREST)) &&
            (alphaMultiplyMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP || avifRGBFormatHasAlpha(rgb->format))) {
            // Explanations on the above conditional:
            // * None of these fast paths currently support bilinear upsampling, so avoid all of them
            //   unless the YUV data isn't subsampled or they explicitly requested AVIF_CHROMA_UPSAMPLING_NEAREST.
            // * None of these fast paths currently handle alpha (un)multiply, so avoid all of them
            //   if we can't do alpha (un)multiply as a separated post step (destination format doesn't have alpha).

            ...
        }
```

Before this block of code, `avifImageYUVToRGB()` has already tried avifImageYUVToRGBLibYUV():

```
    avifBool convertedWithLibYUV = AVIF_FALSE;
    avifResult libyuvResult = avifImageYUVToRGBLibYUV(image, rgb);
    if (libyuvResult == AVIF_RESULT_OK) {
        convertedWithLibYUV = AVIF_TRUE;
    } else {
        if (libyuvResult != AVIF_RESULT_NOT_IMPLEMENTED) {
            return libyuvResult;
        }
    }
```

Since `avifImageYUVToRGBLibYUV()` does not handle alpha (un)multiply, we should also avoid it if we can't do alpha (un)multiply as a separated post step (destination format doesn't have alpha). Therefore, I think we need to guard the `avifImageYUVToRGBLibYUV()` call with `alphaMultiplyMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP || avifRGBFormatHasAlpha(rgb->format)`.

tongyuantongyu:
[...]

PR #592, opened 2021-04-09
Check if avifImageYUVToRGBLibYUV can do the desired conversion

tongyuantongyu:
As suggested in #587, before calling `avifImageYUVToRGBLibYUV()` we should check if libyuv can do the desired conversion for us.
Utilize `I420ToRGB24Matrix()` from libyuv to convert to RGB/BGR formats.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -1081,18 +1081,22 @@ avifResult avifImageYUVToRGB(const avifImage * image, avifRGBImage * rgb)
         return AVIF_RESULT_REFORMAT_FAILED;
     }
 
+    avifAlphaMultiplyMode alphaMultiplyMode = state.toRGBAlphaMode;
     avifBool convertedWithLibYUV = AVIF_FALSE;
-    avifResult libyuvResult = avifImageYUVToRGBLibYUV(image, rgb);
-    if (libyuvResult == AVIF_RESULT_OK) {
-        convertedWithLibYUV = AVIF_TRUE;
-    } else {
-        if (libyuvResult != AVIF_RESULT_NOT_IMPLEMENTED) {
-            return libyuvResult;
+    avifResult libyuvResult = AVIF_RESULT_NOT_IMPLEMENTED;
+    if (alphaMultiplyMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP || avifRGBFormatHasAlpha(rgb->format)) {
+        libyuvResult = avifImageYUVToRGBLibYUV(image, rgb);
+
+        if (libyuvResult == AVIF_RESULT_OK) {
+            convertedWithLibYUV = AVIF_TRUE;
+        } else {
+            if (libyuvResult != AVIF_RESULT_NOT_IMPLEMENTED) {
+                return libyuvResult;
+            }
         }
     }
 
     // Reformat alpha, if user asks for it, or (un)multiply processing needs it.
-    avifAlphaMultiplyMode alphaMultiplyMode = state.toRGBAlphaMode;
     if (avifRGBFormatHasAlpha(rgb->format) && (!rgb->ignoreAlpha || (alphaMultiplyMode != AVIF_ALPHA_MULTIPLY_MODE_NO_OP))) {
         avifAlphaParams params;
 
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -1083,10 +1083,8 @@ avifResult avifImageYUVToRGB(const avifImage * image, avifRGBImage * rgb)
 
     avifAlphaMultiplyMode alphaMultiplyMode = state.toRGBAlphaMode;
     avifBool convertedWithLibYUV = AVIF_FALSE;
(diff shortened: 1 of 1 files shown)


src/reformat.c
```diff
@@ -1082,7 +1082,10 @@ avifResult avifImageYUVToRGB(const avifImage * image, avifRGBImage * rgb)
     }
 
     avifBool convertedWithLibYUV = AVIF_FALSE;
-    avifResult libyuvResult = avifImageYUVToRGBLibYUV(image, rgb);
+    avifResult libyuvResult = AVIF_RESULT_NOT_IMPLEMENTED;
```
wantehchang, line 1085: I suggest two changes.

1. Move the if-else statement for `libyuvResult == AVIF_RESULT_OK` into the new if statement you added.
2. Move the declaration of the `alphaMultiplyMode` local variable earlier and use it in the new if statement.


Here is the patch:

```
index 68b25ea..acbbfda 100644
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -1078,18 +1078,20 @@ avifResult avifImageYUVToRGB(const avifImage * image, avifRGBImage * rgb)
     }
 
+    avifAlphaMultiplyMode alphaMultiplyMode = state.toRGBAlphaMode;
     avifBool convertedWithLibYUV = AVIF_FALSE;
-    avifResult libyuvResult = avifImageYUVToRGBLibYUV(image, rgb);
-    if (libyuvResult == AVIF_RESULT_OK) {
-        convertedWithLibYUV = AVIF_TRUE;
-    } else {
-        if (libyuvResult != AVIF_RESULT_NOT_IMPLEMENTED) {
-            return libyuvResult;
+    if (alphaMultiplyMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP || avifRGBFormatHasAlpha(rgb->format)) {
+        avifResult libyuvResult = avifImageYUVToRGBLibYUV(image, rgb);
+        if (libyuvResult == AVIF_RESULT_OK) {
+            convertedWithLibYUV = AVIF_TRUE;
+        } else {
+            if (libyuvResult != AVIF_RESULT_NOT_IMPLEMENTED) {
+                return libyuvResult;
+            }
         }
     }
 
     // Reformat alpha, if user asks for it, or (un)multiply processing needs it.
-    avifAlphaMultiplyMode alphaMultiplyMode = state.toRGBAlphaMode;
     if (avifRGBFormatHasAlpha(rgb->format) && (!rgb->ignoreAlpha || (alphaMultiplyMode != AVIF_ALPHA_MULTIPLY_MODE_NO_OP))) {
         avifAlphaParams params;
 
```

wantehchang (changes requested):
Hi Yuan Tong,

Thank you for writing the pull request! I suggest the following.

1. Split the pull request into two pull requests. The first one contains the change for src/reformat.c only, which fixes https://github.com/AOMediaCodec/libavif/issues/587. The second one contains the changes to src/reformat_libyuv.c only.

2. Remove all the `#if 0` code in src/reformat_libyuv.c.

src/reformat.c
tongyuantongyu: Done.
Changes to `src/reformat_libyuv.c` are now removed from this PR. I'll include it in the planned PR to update libyuv usage,  when libyuv's 16 bit support is finished.

src/reformat.c
```diff
@@ -1081,18 +1081,22 @@ avifResult avifImageYUVToRGB(const avifImage * image, avifRGBImage * rgb)
...
 
+    avifAlphaMultiplyMode alphaMultiplyMode = state.toRGBAlphaMode;
     avifBool convertedWithLibYUV = AVIF_FALSE;
-    avifResult libyuvResult = avifImageYUVToRGBLibYUV(image, rgb);
-    if (libyuvResult == AVIF_RESULT_OK) {
-        convertedWithLibYUV = AVIF_TRUE;
-    } else {
-        if (libyuvResult != AVIF_RESULT_NOT_IMPLEMENTED) {
-            return libyuvResult;
+    avifResult libyuvResult = AVIF_RESULT_NOT_IMPLEMENTED;
```
wantehchang, line 1086: Please delete line 1086:

```
    avifResult libyuvResult = AVIF_RESULT_NOT_IMPLEMENTED;
````

and declare libyuvResult directly at line 1088:

```
        avifResult libyuvResult = avifImageYUVToRGBLibYUV(image, rgb);
````

The blank line at 1089 could be deleted.

wantehchang (changes requested):

src/reformat.c
tongyuantongyu: Done.
Tip: You can also push commits to the branch in my fork directly, if you prefer.

wantehchang: Thank you. I will modify your pull requests directly to make this kind of small change in the future. (I recently worked with someone who preferred to make all the changes I suggested, big or small.)

By the way, I think it is good to write a pull request now to use only the I420ToRGB24() function.

wantehchang (approved):

PR #594, opened 2021-04-14
Disable incompatible libyuv. Added AVIF_ENABLE_LIBYUV option [closes #593]

### LINKED ISSUE #593: Ignore incompatible libyuv
Hi!

avif enables libyuv support even if the installed libyuv version is not compatible or outdated which leads to compilation errors:

```
reformat_libyuv.c(113): error C2065: 'kYuv2020Constants': undeclared identifier
```

It would be great to add more tests to check if the installed libyuv is compatible and enable/disable its support appropriately.
It would be also great to add an option to disable libyuv explicitly like `cmake -DAVIF_ENABLE_LIBYUV=OFF <...> ..`

smoked-herring:


joedrago:
I see what you're trying to fix here and appreciate the PR, but seeing as `libyuv` is basically forever unversioned (out of my control) and CMake's "see if this builds" can be odd on certain embedded cross compilation setups, I'd prefer to solve this with some `#ifdefs` in `reformat_libyuv.c`, if possible. (To be clear, I am 100% on board with solving this issue with you.)

As a simple example, if `kYuv2020Constants` is something that doesn't exist in common libyuv releases in the wild, it appears that the declaration of that also defines end-swapped variations such as `kYvu2020Constants` using the preprocessor. Can we instead solve this by wrapping the setting of `matrixYUV` and `matrixYVU` with something like `#if defined(kYuv2020ConstantsVU)` (etc)? Ideally, we'd be able to switch in the code on version boundaries, but seeing as we don't have that, I think a few careful ifdefs in `reformat_libyuv.c` would go a long way. 

If you don't want to try this yourself, I'm happy to take a shot at it myself if you can provide the SHA of libyuv that your system provides. I can then verify that I'm protecting against these kinds of things.

smoked-herring:
Thanks for the quick response! Sure, if you have a better approach in mind, that would be cool to implement it. Any solution to fix #593 works. I don't think I can fix it by myself with `ifdef-s` as I'm not really aware of the libyuv API, so I would ask you then to implement the fix you proposed :)

The libyuv version I'm talking about is the version available in `vcpkg`. It's commit hash is `fec9121b676eccd9acea2460aec7d6ae219701b9`. Yes, it's almost a year old! The error log is attached in #593 . Please decline this PR then.

joedrago:
Alright, I will decline this PR and commit something today that makes libavif happily build alongside libyuv `fec9121b676eccd9acea2460aec7d6ae219701b9`.

PR #595, opened 2021-04-14
CI: Add caching and macOS job

EwoutH:
This PR updates the CI configuration to implement caching of all external dependencies (currently aom, dav1d, rav1e, libgav1 and libyuv), speeding up the build from 15 to 20 minutes to under 1 minute with a cache hit.

It also adds a macOS build using Xcode with Apple Clang and includes a bit of cleanup. It updates libyuv to resolve an error on macOS.

I have split the PR into 3 commits to separate each feature. You can see a run without caching [here](https://github.com/EwoutH/libavif/actions/runs/749698421) and a build with caching [here](https://github.com/EwoutH/libavif/actions/runs/749829968).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,24 +15,38 @@ jobs:
       run: echo "CC=gcc-10" >> $GITHUB_ENV && echo "CXX=g++-10" >> $GITHUB_ENV
     - uses: actions/setup-python@v2
 
+    - name: Cache external dependencies
+      id: cache-ext
+      uses: actions/cache@v2
+      with:
+        path: ext
+        key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}
     - uses: ilammy/setup-nasm@v1
+      if: steps.cache-ext.outputs.cache-hit != 'true'
       with:
         version: 2.15.05
     - uses: seanmiddleditch/gha-setup-ninja@v3
+      if: steps.cache-ext.outputs.cache-hit != 'true'
     - run: pip install meson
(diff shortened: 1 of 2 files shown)


joedrago:
I certainly love the idea of this PR, but I am quite ignorant on the magic going on in these changes, and would love to know how to interpret these changes before accepting them. Who is storing this compile cache? Is this just something GitHub securely offers for free? Should I be concerned about depending on `seanmiddleditch/gha-setup-ninja@v3`? Any additional information would be great here.

wantehchang:
Joe: Perhaps we can merge https://github.com/AOMediaCodec/libavif/pull/595/commits/98fb2a197a0a0159b9877a580d9800340a7a5b14 first? It is straightforward.

EwoutH:
Right, this was a long an _very_ iterative process so let me expand a bit:

- seanmiddleditch/gha-setup-ninja@v3 is used to install Ninja. Using apt-get wasn't ideal since macOS doesn't support it. We can fix this action to a specific commit if prefered.
- Compile cache is stored on GitHub servers. There is a compile cache for each branch, and all branches have access to the master-branch cache (for PRs etc.). It's indeed offered free of charge, and we don't store any sensitive things in there.

I think that's most of it. The cache will be updated every time `*.cmd` file in `ext/` is updated. And I reordered some steps to use them in the cache or not.

joedrago:
Considering the source (@EwoutH has proven trustworthy over the years), I'm willing to give it a shot. I might have to lean on you a bit if this starts to fail though. 😄

EwoutH:
Happy with the new CI config so far? Any feedback?

PR #600, opened 2021-04-16
Avoid unneeded clamping of R,G,B

wantehchang:
In avifImageYUVAnyToRGBAnySlow(), avoid unneeded AVIF_MIN(., 1.0f) calls
on R,G,B.

Use Rc, Gc, Bc variables similar to other functions and the original
version of avifImageYUVAnyToRGBAnySlow(), and add an 'Ac' variable for
the clamped version of 'A'.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -621,9 +621,9 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
                 B = Y;
             }
 
-            R = AVIF_CLAMP(R, 0.0f, 1.0f);
-            G = AVIF_CLAMP(G, 0.0f, 1.0f);
-            B = AVIF_CLAMP(B, 0.0f, 1.0f);
+            float Rc = AVIF_CLAMP(R, 0.0f, 1.0f);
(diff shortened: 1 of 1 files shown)


wantehchang:
@tongyuantongyu: Could you please review this pull request? Thanks.

src/reformat.c
```diff
@@ -621,9 +621,9 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
                 B = Y;
             }
 
-            R = AVIF_CLAMP(R, 0.0f, 1.0f);
-            G = AVIF_CLAMP(G, 0.0f, 1.0f);
-            B = AVIF_CLAMP(B, 0.0f, 1.0f);
```
wantehchang, line 626: I considered simply moving these three lines after the if statement, to replace lines 664-666 below. Unfortunately, that may potentially allow a large value of R, G, or B to be multiplied by an A less than but very close to 1.0f at lines 645-647 and then be clamped to exactly 1.0f. That seems incorrect. So I didn't do that.

tongyuantongyu:
This looks good to me.

PR #601, opened 2021-04-16
Check colorPrimaries is BT.2020 for HDR10

wantehchang:
HDR10's CICP is 9/16/9.


wantehchang:
@tongyuantongyu: Please review this PR. Thanks!

PR #603, opened 2021-04-17
Check yuvFormat is 444 or NONE if MC is IDENTITY

wantehchang:
In avifJPEGReadCopy(), if avif->matrixCoefficients is
AVIF_MATRIX_COEFFICIENTS_IDENTITY, also check if avif->yuvFormat is
AVIF_PIXEL_FORMAT_YUV444 or AVIF_PIXEL_FORMAT_NONE. Change
AVIF_PIXEL_FORMAT_NONE to AVIF_PIXEL_FORMAT_YUV444 in that case.

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -193,8 +193,10 @@ static avifBool avifJPEGReadCopy(avifImage * avif, struct jpeg_decompress_struct
                 }
             }
 
-            // Grayscale->RGB: copy Y to G, duplicate to R and B.
-            if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
+            // Grayscale->RGB: copy Y to G, duplicate to B and R.
+            if ((avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) &&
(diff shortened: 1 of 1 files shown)


apps/shared/avifjpeg.c
```diff
@@ -193,8 +193,10 @@ static avifBool avifJPEGReadCopy(avifImage * avif, struct jpeg_decompress_struct
                 }
             }
 
-            // Grayscale->RGB: copy Y to G, duplicate to R and B.
-            if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
+            // Grayscale->RGB: copy Y to G, duplicate to B and R.
```
wantehchang, line 196: I changed "R and B" to "B and R" because the order of the channels is G, B, R.

wantehchang:
@tongyuantongyu: Please review this PR. Thanks!

tongyuantongyu:
Maybe it's better to check and set yuvFormat at [avifenc.c#L746](https://github.com/AOMediaCodec/libavif/blob/master/apps/avifenc.c#L746)?
For PNG input, `--cicp 1/13/0 --yuv auto` result in 1/13/6 YUV444 but not 1/13/0 YUV444.

joedrago:
I believe Wan-Teh has a second PR planned to address the block at avifenc.c#L746.

PR #606, opened 2021-04-20
Generalize the "no alpha, no premultiply" handling

wantehchang:
Move the "JPEG doesn't have alpha. Prevent confusion." code in
avifJPEGRead() to avifInputReadImage() so that it applies to all input
image formats.

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -276,8 +276,6 @@ avifBool avifJPEGRead(const char * inputFilename, avifImage * avif, avifPixelFor
 
     avif->yuvFormat = requestedFormat; // This may be AVIF_PIXEL_FORMAT_NONE, which is "auto" to avifJPEGReadCopy()
     avif->depth = requestedDepth ? requestedDepth : 8;
-    // JPEG doesn't have alpha. Prevent confusion.
-    avif->alphaPremultiplied = AVIF_FALSE;
 
     if (avifJPEGReadCopy(avif, &cinfo)) {
         // JPEG pixels were successfully copied without conversion. Notify the enduser.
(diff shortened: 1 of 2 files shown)


wantehchang:
@joedrago, @tongyuantongyu: Please review. Does this PR make sense to you?

joedrago:
LGTM.

tongyuantongyu:
LGTM.

apps/shared/avifjpeg.c
```diff
@@ -276,8 +276,6 @@ avifBool avifJPEGRead(const char * inputFilename, avifImage * avif, avifPixelFor
 
     avif->yuvFormat = requestedFormat; // This may be AVIF_PIXEL_FORMAT_NONE, which is "auto" to avifJPEGReadCopy()
     avif->depth = requestedDepth ? requestedDepth : 8;
-    // JPEG doesn't have alpha. Prevent confusion.
-    avif->alphaPremultiplied = AVIF_FALSE;
```
wantehchang, line 280: I just realized I neglected an issue: this is before the `avifImageRGBToYUV()` call (at line 316). So this may be intended to prevent confusion in `avifImageRGBToYUV()`.

Yuan, Joe, is this intended to prevent confusion in `avifImageRGBToYUV()` or in `avifEncoderAddImage()`?

If it is the former (or both), then we cannot move this to a later point.

joedrago: Ah, yes. This value is going to be utilized by the YUV->RGB conversion (if invoked). It might be smarter to abandon this PR.

wantehchang: Joe: Thank you for the reply. I will abandon this PR.

Note: `avifImageRGBToYUV()` won't be confused by this because it does not use `avif->alphaPremultiplied` unless the RGB format has alpha. But I can see that programmers may be confused when they debug inside the `avifImageRGBToYUV()` function and dump the contents of the `avifImage` struct.

PR #608, opened 2021-04-21
Remove an unneeded test and duplicate code

wantehchang:
Remove an unneeded test for YUV 444, 422, and 420 in avifJPEGReadCopy().
That test always evaluates to true.

Remove two lines of duplicate code in avifJPEGRead().

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -291,9 +291,6 @@ avifBool avifJPEGRead(const char * inputFilename, avifImage * avif, avifPixelFor
         cinfo.out_color_space = JCS_RGB;
         jpeg_start_decompress(&cinfo);
 
-        avif->width = cinfo.output_width;
-        avif->height = cinfo.output_height;
-
         int row_stride = cinfo.output_width * cinfo.output_components;
         JSAMPARRAY buffer = (*cinfo.mem->alloc_sarray)((j_common_ptr)&cinfo, JPOOL_IMAGE, row_stride, 1);
 


apps/shared/avifjpeg.c
```diff
@@ -291,9 +288,6 @@ avifBool avifJPEGRead(const char * inputFilename, avifImage * avif, avifPixelFor
         cinfo.out_color_space = JCS_RGB;
         jpeg_start_decompress(&cinfo);
 
-        avif->width = cinfo.output_width;
-        avif->height = cinfo.output_height;
```
wantehchang, line 295: These two lines are the same as lines 300-301 below.

wantehchang:
@tongyuantongyu: Please review. Thanks!

tongyuantongyu:
I was using this to explicitly tell the current situation, but I'm fine removing it.

joedrago:
Perhaps as a good middle ground on "unnecessary checks" vs "self-documenting" we could switch this check to an assert?

wantehchang:
Since YUV 4:0:0 is the only monochrome pixel format in AVIF, if image->yuvFormat is not YUV 4:0:0 (including 'auto'), then avif->yuvFormat has U, V planes. This is obvious to me. As an analogy, for me this code is similar to the following code:

```
  if (dayOfWeek == Saturday || dayOfWeek == Sunday) {
    ... code for the weekend ...
    return;
  }

  if (dayOfWeek == Monday || dayOfWeek == Tuesday || ... || dayOfWeek == Friday) {
    ... code for a weekday ...
    return;
  }
```

But I didn't realize that this explicit check was intentional (for clarity). I will resurrect it in this PR.

wantehchang:
@joedrago, @tongyuantongyu: Please review the updated commit in this PR. Thanks!

joedrago:
> Since YUV 4:0:0 is the only monochrome pixel format in AVIF, if image->yuvFormat is not YUV 4:0:0 (including 'auto'), then avif->yuvFormat has U, V planes. This is obvious to me. As an analogy, for me this code is similar to the following code:
> 
> ```
>   if (dayOfWeek == Saturday || dayOfWeek == Sunday) {
>     ... code for the weekend ...
>     return;
>   }
> 
>   if (dayOfWeek == Monday || dayOfWeek == Tuesday || ... || dayOfWeek == Friday) {
>     ... code for a weekday ...
>     return;
>   }
> ```

I'll fully admit that keeping the original conditional or converting it to an assert is defensive programming, but I think it is warranted. Your approach only makes sense in an environment where the enum's values are set in stone, and nobody will ever add another, as your example with days-of-the-week suggests. However, using your example, what if that environment had a strong need for enum values `Yesterday`, `Today`, and `Tomorrow`? Now your code that saved one conditional needs to be fully audited (maybe this is the case, but if you always wrote defensively, you wouldn't have to be as rigorous).

Even in the context of this new code, this is the case. There isn't a reason to expect a new subsampling other than 400/420/422/444 in this code, but we _also_ have `NONE`, and instead of reusing it here, we _could_ have added `AUTO`. This kind of change in the name of optimization (saving a conditional) or simplicity (no complicate this code with another check) actually might serve to make the code less obvious to a new reader. If we remove the conditional _and_ don't convert it to an assert, then the only other viable choice would be to make it a clarifying comment, so when we must audit the code months from now if we ever have to add another format, we'll have that comment hinting to us that this codepath is assuming that the value must be 420/422/444.

wantehchang:
Joe: Thank you for your comment. Could you review the latest version of this PR? Thanks.

joedrago:
The new PR looks quite friendly. LGTM

PR #609, opened 2021-04-21
Miscellaneous cleanup for src/codec_svt.c

wantehchang:
If svt_av1_enc_init_handle() succeeds, we don't need to check if
codec->internal->svt_encoder is null.

The value of a conditional expression is 1 (AVIF_TRUE) or 0
(AVIF_FALSE), so we don't need to follow a conditional expression
with ? AVIF_TRUE : AVIF_FALSE.

Check the return value of allocate_svt_buffers().

No need to set the local variable input_buffer to null after freeing it
because the function is about to return.


wantehchang:
@tongyuantongyu: Please review. Thanks!

tongyuantongyu:
LGTM. Thanks!

PR #613, opened 2021-04-26
Set dav1dSettings.n_postfilter_threads if present

wantehchang:
Set dav1dSettings.n_postfilter_threads to decoder->maxThreads (after
clamping) in dav1d 0.8.2 or later.

The n_postfilter_threads member was added to the Dav1dSettings struct in
https://code.videolan.org/videolan/dav1d/-/commit/549086e4d3024410c6145c89266b37c916808b04.

### DIFF
--- a/src/codec_dav1d.c
+++ b/src/codec_dav1d.c
@@ -54,7 +54,26 @@ static avifBool dav1dCodecOpen(avifCodec * codec, avifDecoder * decoder)
     if (codec->internal->dav1dContext == NULL) {
         // Give all available threads to decode a single frame as fast as possible
         codec->internal->dav1dSettings.n_frame_threads = 1;
-        codec->internal->dav1dSettings.n_tile_threads = AVIF_CLAMP(decoder->maxThreads, 1, DAV1D_MAX_TILE_THREADS);
+        int maxThreads = AVIF_MAX(decoder->maxThreads, 1);
+        int tileThreads = AVIF_MIN(maxThreads, DAV1D_MAX_TILE_THREADS);
+#ifdef DAV1D_MAX_POSTFILTER_THREADS
+        // In dav1d 0.8.2 or later (dav1d API version 5.0.1 or later), n_postfilter_threads is
+        // available. The tile threads are way more useful than the postfilter threads, so allocate
+        // a quarter or so of the threads to postfilters and the rest to tiles.
+        int postfilterThreads;
+        if (maxThreads <= 8) {
+            postfilterThreads = (maxThreads + 3) / 4; // ceil(maxThreads / 4.0)
+        } else {
+            postfilterThreads = AVIF_MIN(maxThreads / 4, DAV1D_MAX_POSTFILTER_THREADS);
+        }
+        // If n_postfilter_threads is equal to 1, dav1d won't create any postfilter threads.
+        if ((postfilterThreads > 1) && ((tileThreads + postfilterThreads) > maxThreads)) {
+            int excess = tileThreads + postfilterThreads - maxThreads;
+            tileThreads -= excess;
+        }
+        codec->internal->dav1dSettings.n_postfilter_threads = postfilterThreads;
+#endif
+        codec->internal->dav1dSettings.n_tile_threads = tileThreads;
 
         if (dav1d_open(&codec->internal->dav1dContext, &codec->internal->dav1dSettings) != 0) {
             return AVIF_FALSE;


src/codec_dav1d.c
```diff
@@ -55,6 +55,11 @@ static avifBool dav1dCodecOpen(avifCodec * codec, avifDecoder * decoder)
         // Give all available threads to decode a single frame as fast as possible
         codec->internal->dav1dSettings.n_frame_threads = 1;
         codec->internal->dav1dSettings.n_tile_threads = AVIF_CLAMP(decoder->maxThreads, 1, DAV1D_MAX_TILE_THREADS);
+#ifdef DAV1D_MAX_POSTFILTER_THREADS
+        // In dav1d 0.8.2 or later (dav1d API version 5.0.1 or later), n_postfilter_threads is
+        // available.
+        codec->internal->dav1dSettings.n_postfilter_threads = AVIF_CLAMP(decoder->maxThreads, 1, DAV1D_MAX_POSTFILTER_THREADS);
+#endif
```
wantehchang, line 62: We can alternatively check the dav1d version macros, defined in `dav1d.h`:

```
#define DAV1D_API_VERSION_MAJOR 5
#define DAV1D_API_VERSION_MINOR 0
#define DAV1D_API_VERSION_PATCH 1
```

But it will be much more verbose.

joedrago:
I think this looks good. Would you like me to merge this in now, or do you want to wait for confirmation of the usage of this variable (via email)?

wantehchang:
Joe: Please don't merge this PR. I am still discussing with Victorien Le Couviour about how to set n_postfilter_threads properly.

wantehchang:
Joe: Please review the new commit. I implemented the simple heuristic that Victorien Le Couviour suggested in email. Thanks!

src/codec_dav1d.c
```diff
@@ -54,7 +54,26 @@ static avifBool dav1dCodecOpen(avifCodec * codec, avifDecoder * decoder)
...
+        // In dav1d 0.8.2 or later (dav1d API version 5.0.1 or later), n_postfilter_threads is
+        // available. The tile threads are way more useful than the postfilter threads, so allocate
+        // a quarter or so of the threads to postfilters and the rest to tiles.
+        int postfilterThreads;
+        if (maxThreads <= 8) {
+            postfilterThreads = (maxThreads + 3) / 4; // ceil(maxThreads / 4.0)
+        } else {
+            postfilterThreads = AVIF_MIN(maxThreads / 4, DAV1D_MAX_POSTFILTER_THREADS);
+        }
+        // If n_postfilter_threads is equal to 1, dav1d won't create any postfilter threads.
```
psilokos-twoorioles, line 69: well in fact even if the postfilters aren't parallelized they're still running in their own thread (the frame thread, or in your case the application thread), since all the entropy decoding / block recon is happening in the tile threads, that means they run linearly, but still in parallel of the tile threading stuff. So you effectively have n_tile_threads + n_pf_threads, in any case, even if the latter is one. (basically if any of these 2 values is 1, your application thread does actual work, so imo you should consider that as part of your maxThreads. Obviously tho if both are 1 you just have your app thread active, nothing is spawned).

src/codec_dav1d.c
```diff
@@ -54,7 +54,26 @@ static avifBool dav1dCodecOpen(avifCodec * codec, avifDecoder * decoder)
...
-        codec->internal->dav1dSettings.n_tile_threads = AVIF_CLAMP(decoder->maxThreads, 1, DAV1D_MAX_TILE_THREADS);
+        int maxThreads = AVIF_MAX(decoder->maxThreads, 1);
+        int tileThreads = AVIF_MIN(maxThreads, DAV1D_MAX_TILE_THREADS);
+#ifdef DAV1D_MAX_POSTFILTER_THREADS
+        // In dav1d 0.8.2 or later (dav1d API version 5.0.1 or later), n_postfilter_threads is
+        // available. The tile threads are way more useful than the postfilter threads, so allocate
+        // a quarter or so of the threads to postfilters and the rest to tiles.
+        int postfilterThreads;
+        if (maxThreads <= 8) {
+            postfilterThreads = (maxThreads + 3) / 4; // ceil(maxThreads / 4.0)
```
psilokos-twoorioles, line 65: actually for 5 threads you'd probably want 4 tt / 1 pft as well, that would result in 3 tt / 2 pft, for the same reason your comment states above

wantehchang:
I abandoned this pull request because dav1d's new threading model has been checked in:
https://code.videolan.org/videolan/dav1d/-/commit/753eef833b

This pull request is replaced by https://github.com/AOMediaCodec/libavif/pull/751.

PR #614, opened 2021-04-28
Update the documentation for libaom sharpness param

jayasanker-j:
The documentation for libaom sharpness parameter is updated to comply with the current implementation in libaom. These changes were introduced in the following CL:

https://aomedia-review.googlesource.com/c/aom/+/60901

Loop filtering does not make use of the sharpness parameter now. It is currently used to control sharpness in the block through transform coeff RD.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -120,7 +120,7 @@ static void syntax(void)
         printf("    cq-level=Q                        : Constant/Constrained Quality level (0-63, end-usage must be set to cq or q)\n");
         printf("    enable-chroma-deltaq=B            : Enable delta quantization in chroma planes (0: disable (default), 1: enable)\n");
         printf("    end-usage=MODE                    : Rate control mode (vbr, cbr, cq, or q)\n");
-        printf("    sharpness=S                       : Loop filter sharpness (0-7, default: 0)\n");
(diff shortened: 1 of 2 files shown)


apps/avifenc.c
```diff
@@ -120,7 +120,7 @@ static void syntax(void)
         printf("    cq-level=Q                        : Constant/Constrained Quality level (0-63, end-usage must be set to cq or q)\n");
         printf("    enable-chroma-deltaq=B            : Enable delta quantization in chroma planes (0: disable (default), 1: enable)\n");
         printf("    end-usage=MODE                    : Rate control mode (vbr, cbr, cq, or q)\n");
-        printf("    sharpness=S                       : Loop filter sharpness (0-7, default: 0)\n");
+        printf("    sharpness=S                       : Bias towards block sharpness in transform coeff RD (0-7, default: 0)\n");
```
wantehchang, line 123: Since this is documentation for end users, we should avoid technical abbreviations. Could you please change "transform coeff RD" to "transform coefficient rate-distortion optimization" or something like that? Thanks.

wantehchang (changes requested):

apps/avifenc.c
jayasanker-j: Hi @wantehchang
We have updated the documentation as per your comments.
Please review.

wantehchang (approved):
LGTM. Thanks!

PR #616, opened 2021-04-29
Use libyuv BT.709 & 2020 full range YuvConstants

wantehchang:
Use libyuv BT.709 and BT.2020 full range YuvConstants if they are
available.


joedrago:
Ah, this is awesome, and exactly how I hoped these switch statements would evolve.

PR #617, opened 2021-05-02
Add GitHub Actions workflow for Windows builds

aosoft:
* Build applications (avifdec.exe, avifenc.exe) and libs (avif_static.lib)
* Build shared libraries (avif.dll)
* Support feature switches (aom, dav1d, rav1e, libgav1, svt, libyuv)

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -45,6 +45,8 @@ jobs:
     - uses: ilammy/msvc-dev-cmd@v1
     - uses: ilammy/setup-nasm@v1
     - uses: seanmiddleditch/gha-setup-ninja@v3
+    - name: Setup meson
+      run: pip install meson
     - name: Short SHA
       run: |
         $ssha=$env:GITHUB_SHA.Substring(0, 8)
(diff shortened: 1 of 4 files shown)


tongyuantongyu:
We probably don't want to hardcode dependency versions in CI config file.
meson can be installed through pip.
libyuv has some trouble building with Ninja generator and MSVC compiler. We may fix it in `libyuv.cmd` directly.

aosoft:
Thanks for reply.

>meson can be installed through pip.

On Windows, when meson was installed by pip, it could not be executed by meson command, so I got it from GitHub.
https://stackoverflow.com/questions/42712896/how-to-run-meson-build-system-on-windows

tongyuantongyu:
> On Windows, when meson was installed by pip, it could not be executed by meson command, so I got it from GitHub.

It should work if the `Scripts` folder of your active python environment is in your `PATH` environment variable. 
Github Action's environment has this set up, so you can install it using pip.

I made some changes on your commits here: https://github.com/tongyuantongyu/libavif/commits/build-windows-action . If you think these changes are good, feel free to cherry pick them into your branch.

aosoft:
> > On Windows, when meson was installed by pip, it could not be executed by meson command, so I got it from GitHub.
> 
> It should work if the `Scripts` folder of your active python environment is in your `PATH` environment variable.
> Github Action's environment has this set up, so you can install it using pip.
> 
> I made some changes on your commits here: https://github.com/tongyuantongyu/libavif/commits/build-windows-action . If you think these changes are good, feel free to cherry pick them into your branch.

Thank you for the fix. I've confirmed that I can build it.
When I was trying, installing meson with pip failed to build even on GitHub Actions, so I had that implementation.

vrabaud:
There is now Windows CI, so closing.

PR #619, opened 2021-05-05
Add the allowMissingPixi decoder setting

wantehchang:
The allowMissingPixi decoder setting allows AVIF images to not have the
PixelInformationProperty (pixi) in AV1 image items. Images encoded with
libheif v1.11.0 or older do not have the 'pixi' item property. (This bug
has been fixed in libheif v1.12.0, just released on 2021-05-05.)

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -734,6 +734,14 @@ typedef struct avifDecoder
     // (see comment above), and setting this to 0 disables the limit.
     uint32_t imageCountLimit;
 
+    // Allow the PixelInformationProperty ('pixi') to be missing in AV1 image items. Default to
+    // false.
+    //
+    // libheif v1.11.0 or older does not add the 'pixi' item property to AV1 image items. If you
+    // need to decode AVIF images encoded by libheif v1.11.0 or older, set allowMissingPixi to
+    // true. (This issue has been corrected in libheif v1.12.0.)
+    avifBool allowMissingPixi;
+
     // stats from the most recent read, possibly 0s if reading an image sequence
     avifIOStats ioStats;
 
--- a/src/read.c
+++ b/src/read.c
@@ -712,7 +712,7 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifExt
     return AVIF_RESULT_OK;
 }
 
-static avifResult avifDecoderItemValidateAV1(const avifDecoderItem * item)
+static avifResult avifDecoderItemValidateAV1(const avifDecoderItem * item, avifBool allowMissingPixi)
 {
     const avifProperty * av1CProp = avifPropertyArrayFind(&item->properties, "av1C");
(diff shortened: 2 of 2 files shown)


joedrago:
The code itself here is simple to understand and reasonable, but I'm curious about its usage. I imagine it exists to give some time for libheif users to update their files without breaking them in Chrome, but what is the timeframe that this will stay enabled in Chrome?

wantehchang:
Joe: I need the allowMissingPixi decoder setting available in case I need to disable the 'pixi' check when future versions of Chrome advance to the Beta channel. (I am planning to leave the 'pixi' check enabled in the Canary and Dev channels of Chrome.)

While working on https://crbug.com/1199242, I realized that there are potentially a large number of websites that encode AVIF images using libheif. I fixed https://crbug.com/1199242 by reverting to an older version of libavif, but I won't be able to do that next time.

In https://crbug.com/1199242, two websites in India, https://www.thequint.com/ and https://swarajyamag.com/, are known to use AVIF images encoded with libheif. Imagine the amount of work it would take to contact the webmasters of those two websites and ask them to update libheif to v1.12.0 or later. (I made an attempt to contact them. See https://crbug.com/1199242#c9.) I am afraid that "we have bigger fish to fry."

joedrago:
I have a solution for this, which I'm working on now. We can discuss it in a few hours.

joedrago:
@wantehchang Please review my additional changes in the `diagnostics` branch. If this alternative direction fixes the issue you're trying to solve here, I'll merge in `diagnostics` completely and we can close this PR.

farindk:
In my opinion, it is good to have a strict decoding mode for checking encoder compliance, but in normal operation, a decoder should be as forgiving as possible as you will always come across non-compliant files.

The `pixi` box in particular seems to be for informal purposes only. It doesn't seem to be required at all for decoding as the same information is in `av1C` and even that is not needed for decoding. In HEIF (23008-12), it is optional.

EwoutH:
Personally I think the reference decoder, which libavif is for AVIF, should follow the specification very strict, at least by default. If even the reference decoders starts accepting non-standard things by default, less effort will be made to follow the spec anymore, since even the reference decoder allows it.

joedrago:
I haven't made any permanent decisions for the new strict mode's defaults. I could see inverting/replacing the new `avifdec` commandline switch `--strict` to a new switch that was the opposite (`--lenient` or `--no-strict`, etc, I'll think about it), and make the default have all of the strict flags set. Implementations that use libavif can decide for themselves when they instantiate an `avifDecoder` which strict flags they'd like to set by default.

joedrago:
https://github.com/AOMediaCodec/libavif/commit/4d102edd78dc0c433093f2ed6d9d3c3fe364aa73

I just changed `avifDecoder`'s default to have all strict flags set. Now to decide how on `avifdec`'s defaults...

PR #620, opened 2021-05-08
Docker: switch to ubuntu 20.04, fix tzdata install

paskal:
- Ubuntu 19.10 is not supported and I couldn't even install packages on it when I tried to run docker build
- without `DEBIAN_FRONTEND="noninteractive"` apt is asking the user for timezone choice when `tzdata` package is installed
- docker-compose v3 manifest is for swarm features that are clearly not in use there, so it's safe to decrease the version, which unblocks more docker versions on the client to run this manifest without problems

### DIFF
--- a/tests/docker/Dockerfile
+++ b/tests/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:19.10
+FROM ubuntu:20.04
 
 ADD build.sh /root/build.sh
 RUN apt update && apt install -y dos2unix
--- a/tests/docker/build.sh
+++ b/tests/docker/build.sh
@@ -17,7 +17,7 @@
 
 # build env
 apt update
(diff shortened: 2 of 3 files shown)


paskal:
Gentle ping @wantehchang @EwoutH

joedrago:
This looks reasonable to me, but I'm not using our Docker script often. I think if @EwoutH thinks these changes are benign, I'll accept it.

EwoutH:
We actually don't test the Docker build in CI yet, that should be setup if we want to actively maintain a Docker build.

`DEBIAN_FRONTEND="noninteractive"` is frequently used in CI when installing apt packages, and using a supported LTS version of Ubuntu is logical.

Thanks for the PR, LGTM.

EwoutH (approved):

paskal:
After this one will be merged I'll try to build a GitHub workflow for docker build for this repo, which would be a good smoke test.
If then someone will go a step further and publish resulting images on DockerHub, I'll be absolutely happy: building this images takes too much time and resources when you want to just play around with it.

PR #624, opened 2021-05-17
Check the return values of some aom_codec_ calls

wantehchang:
Check the return values of the following functions:
  aom_codec_enc_config_default
  aom_codec_enc_init
  aom_codec_encode

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -547,7 +547,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
 
         aom_codec_iface_t * encoderInterface = aom_codec_av1_cx();
         struct aom_codec_enc_cfg cfg;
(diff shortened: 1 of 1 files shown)


joedrago:
Do these aom errors offer more detailed information? This might be a good time to employ `avifDiagnosticsPrintf()` to help people learn exactly what broke.

wantehchang:
Thanks for the suggestion. We should pass `aom_codec_error(&codec->internal->encoder)` and `aom_codec_error_detail(&codec->internal->encoder)` to `avifDiagnosticsPrintf()`.

I took a quick look. It requires some plumbing because `avifEncoder` does not have an `avifDiagnostics diag` member yet. So I'd like to check in this pull request first.

joedrago:
Ah, right. This looks good to me.

PR #628, opened 2021-05-18
CI: Update GitHub Actions workflow to use GCC 11

EwoutH:
Updates the GitHub Actions workflow to use GCC 11 (instead of GCC 10) on the Ubuntu and macOS runs.

GCC 11 [was added](https://github.com/actions/virtual-environments/pull/3291) to the Ubuntu 20.04 environment earlier this month.


EwoutH:
Rebased

EwoutH:
@y-guyon @wantehchang I'm doing some cleanup of old, open PRs. If you would like to get these changes in, please leave a reply or review.

In two weeks I will close this PR if no reaction received.

.github/workflows/ci.yml
```diff
@@ -10,13 +10,12 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Set GCC & G++ 10 compiler (on Linux)
+    - name: Set GCC & G++ 11 compiler (on Linux)
       if: runner.os == 'Linux'
-      run: echo "CC=gcc-10" >> $GITHUB_ENV && echo "CXX=g++-10" >> $GITHUB_ENV
+      run: echo "CC=gcc-11" >> $GITHUB_ENV && echo "CXX=g++-11" >> $GITHUB_ENV
```
jzern, line 15: unrelated, but could this simply be an `env:` entry? Or is there something special about the use of $GITHUB_ENV here?

EwoutH: Looking at the git blame, I modified it in PR https://github.com/AOMediaCodec/libavif/pull/595 about one and a half years ago. I think this has to do with this issue:

- [GitHub Actions: Deprecating set-env and add-path commands](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/)

Or it has to do with retention between steps.

But I don't remember exactly anymore. We could try if it works with a regular `env:`, if you want.

jzern: For this change we can leave it as is. I was just curious when comparing other workflow files.

y-guyon (approved):
Should we also migrate other GitHub workflows such as `ci-unix-shared.yml`?

vrabaud:
GCC 13 is out. Should we update all the CI? Or maybe use that action? https://github.com/marketplace/actions/install-gcc

vrabaud:
Closing in favor of https://github.com/AOMediaCodec/libavif/pull/1886

Issue #636, opened 2021-05-24
avifDecoderParse() should avoid calling tile->codec->open()


wantehchang:
Right now `avifDecoderParse()` calls `tile->codec->open()` at the very end via this call chain:
    `avifDecoderParse()` > `avifDecoderReset()` > `avifDecoderFlush()` > `tile->codec->open()`

```
static avifResult avifDecoderFlush(avifDecoder * decoder)
{
    avifDecoderDataResetCodec(decoder->data);

    for (unsigned int i = 0; i < decoder->data->tiles.count; ++i) {
        avifTile * tile = &decoder->data->tiles.tile[i];
        tile->codec = avifCodecCreateInternal(decoder->codecChoice);
        if (!tile->codec) {
            return AVIF_RESULT_NO_CODEC_AVAILABLE;
        }
        tile->codec->diag = &decoder->diag;
        if (!tile->codec->open(tile->codec, decoder)) {
            return AVIF_RESULT_DECODE_COLOR_FAILED;
        }
    }
    return AVIF_RESULT_OK;
}
```

Since an application may call `avifDecoderParse()` just to parse the container without decoding the image, it would be good to delay the `avifCodecCreateInternal()` and `tile->codec->open()` calls until we decode the image.

Joe, if you agree with this proposed change and have some spare time, could you please implement this change because you know this code the best? Even outlining how this should be done would be very helpful. Thank you!

joedrago:
Similarly to how the encoding pipeline works with lazy initialization, I tried doing that with the decoding pipeline and was quite surprised at how friendly it all worked out. Even the diffs are nice and pretty!

Check it out and let me know if I missed something.

https://github.com/AOMediaCodec/libavif/pull/637

wantehchang:
Joe: Thank you so much for working on this right away!

I forgot to mention that the reason I looked into this today was to see if it was possible to set decoder->maxThreads based on the image's width and height. But I had been wondering about this for a long time, so today I traced the code in a debugger and verified that avifDecoderParse() opens the AV1 decoder.

joedrago:
Ah, yes. This change would allow for that quite nicely.

PR #637, opened 2021-05-24
Merge avifCodec's open call with its getNextImage call to avoid codec…

joedrago:
… init during parse, and simplify the codec API

Fixes: #636

This attempts to mimic the encoder path's symmetry of lazy initialization for the decoding path. This should avoid costly codec initialization during avifDecoderParse().

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3231,7 +3231,7 @@ avifResult avifDecoderNextImage(avifDecoder * decoder)
 
         const avifDecodeSample * sample = &tile->input->samples.sample[nextImageIndex];
 
-        if (!tile->codec->getNextImage(decoder, tile->codec, sample, tile->input->alpha, tile->image)) {
+        if (!tile->codec->getNextImage(tile->codec, decoder, sample, tile->input->alpha, tile->image)) {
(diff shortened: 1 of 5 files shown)


src/read.c
```diff
@@ -2745,9 +2745,6 @@ static avifResult avifDecoderFlush(avifDecoder * decoder)
             return AVIF_RESULT_NO_CODEC_AVAILABLE;
         }
         tile->codec->diag = &decoder->diag;
-        if (!tile->codec->open(tile->codec, decoder)) {
```
wantehchang, line 2748: Is it possible to also postpone the avifCodecCreateInternal() call to avifDecoderNextImage()?

wantehchang (approved):
LGTM. Thank you very much!

It is strange that dav1dCodecOpen() and gav1CodecOpen() already have the initialize-once protection, which is not needed in the original code.

src/read.c
joedrago: That would be more expensive for almost no gain. All calls to `avifCodecCreateInternal()` are little more than a `malloc()`/`memset()` pair, but they set up the existence of the codec in `read.c`. I think this change will give you the flexibility and lazy-init that you're looking for, unless you wanted to pick a _codec_ based on the image's size (which I don't think is a worthwhile use case).

PR #640, opened 2021-05-26
Support for progressive AVIFs and operating point selection

joedrago:
* Support for parsing boxes: a1op, a1lx, lsel
* Added operating point and "skip" support to avifCodec impls for layer selection support
* Reworked avifdec decoding to support nth image decoding and minor optimization
* New avifProgressiveState enum to expose info about progressive AVIFs to users of libavif
* New avifdec args: --progressive, --index

@wantehchang Do not merge this yet -- There are still outstanding issues with a few of the Apple example files. It is unclear if they are issues with the file itself (`animals_00_multilayer_grid_a1lx.avif`), or an issue with our usage of dav1d (`animals_00_multilayer.avif`).

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -36,6 +36,9 @@ static void syntax(void)
     printf("    -u,--upsampling U : Chroma upsampling (for 420/422). automatic (default), fastest, best, nearest, or bilinear\n");
     printf("    -r,--raw-color    : Output raw RGB values instead of multiplying by alpha when saving to opaque formats\n");
     printf("                        (JPEG only; not applicable to y4m)\n");
+    printf("    --index           : When decoding an image sequence or progressive image, specify which frame index to decode (Default: 0)\n");
+    printf("    --progressive     : Enable progressive AVIF processing. If a progressive image is encountered, avifdec will use\n");
+    printf("                        --index to choose which variant to decode (in progressive order).\n");
     printf("    --no-strict       : Disable strict decoding, which disables strict validation checks and errors\n");
     printf("    -i,--info         : Decode all frames and display all image information instead of saving to disk\n");
     printf("    --ignore-icc      : If the input file contains an embedded ICC profile, ignore it (no-op if absent)\n");
@@ -55,7 +58,9 @@ int main(int argc, char * argv[])
     avifChromaUpsampling chromaUpsampling = AVIF_CHROMA_UPSAMPLING_AUTOMATIC;
     avifBool ignoreICC = AVIF_FALSE;
     avifBool rawColor = AVIF_FALSE;
+    avifBool allowProgressive = AVIF_FALSE;
     avifStrictFlags strictFlags = AVIF_STRICT_ENABLED;
+    uint32_t frameIndex = 0;
 
     if (argc < 2) {
         syntax();
@@ -124,6 +129,11 @@ int main(int argc, char * argv[])
             }
         } else if (!strcmp(arg, "-r") || !strcmp(arg, "--raw-color")) {
             rawColor = AVIF_TRUE;
+        } else if (!strcmp(arg, "--progressive")) {
+            allowProgressive = AVIF_TRUE;
+        } else if (!strcmp(arg, "--index")) {
+            NEXTARG();
+            frameIndex = (uint32_t)atoi(arg);
         } else if (!strcmp(arg, "--no-strict")) {
             strictFlags = AVIF_STRICT_DISABLED;
         } else if (!strcmp(arg, "-i") || !strcmp(arg, "--info")) {
@@ -161,6 +171,7 @@ int main(int argc, char * argv[])
         decoder->maxThreads = jobs;
         decoder->codecChoice = codecChoice;
         decoder->strictFlags = strictFlags;
+        decoder->allowProgressive = allowProgressive;
         avifResult result = avifDecoderSetIOFile(decoder, inputFilename);
         if (result != AVIF_RESULT_OK) {
             fprintf(stderr, "Cannot open file for read: %s\n", inputFilename);
@@ -178,17 +189,31 @@ int main(int argc, char * argv[])
                    decoder->durationInTimescales,
                    decoder->imageCount,
                    (decoder->imageCount == 1) ? "" : "s");
-            printf(" * Frames:\n");
+            if (decoder->imageCount > 1) {
+                printf(" * %s Frames: (%u expected frames)\n",
+                       (decoder->progressiveState != AVIF_PROGRESSIVE_STATE_UNAVAILABLE) ? "Progressive Image" : "Image Sequence",
+                       decoder->imageCount);
+            } else {
+                printf(" * Frame:\n");
+            }
 
-            int frameIndex = 0;
-            while (avifDecoderNextImage(decoder) == AVIF_RESULT_OK) {
-                printf("   * Decoded frame [%d] [pts %2.2f (%" PRIu64 " timescales)] [duration %2.2f (%" PRIu64 " timescales)]\n",
-                       frameIndex,
+            int currIndex = 0;
+            avifResult nextImageResult = avifDecoderNextImage(decoder);
+            while (nextImageResult == AVIF_RESULT_OK) {
+                printf("   * Decoded frame [%d] [pts %2.2f (%" PRIu64 " timescales)] [duration %2.2f (%" PRIu64 " timescales)] [%ux%u]\n",
+                       currIndex,
                        decoder->imageTiming.pts,
                        decoder->imageTiming.ptsInTimescales,
                        decoder->imageTiming.duration,
-                       decoder->imageTiming.durationInTimescales);
-                ++frameIndex;
+                       decoder->imageTiming.durationInTimescales,
+                       decoder->image->width,
+                       decoder->image->height);
+                ++currIndex;
+                nextImageResult = avifDecoderNextImage(decoder);
+            }
+            if (nextImageResult != AVIF_RESULT_NO_IMAGES_REMAINING) {
+                printf("ERROR: Failed to decode frame: %s\n", avifResultToString(nextImageResult));
+                avifDumpDiagnostics(&decoder->diag);
             }
         } else {
             printf("ERROR: Failed to decode image: %s\n", avifResultToString(result));
@@ -210,52 +235,73 @@ int main(int argc, char * argv[])
            (jobs == 1) ? "" : "s");
 
     int returnCode = 0;
-    avifImage * avif = avifImageCreateEmpty();
     avifDecoder * decoder = avifDecoderCreate();
     decoder->maxThreads = jobs;
     decoder->codecChoice = codecChoice;
     decoder->strictFlags = strictFlags;
-    avifResult decodeResult = avifDecoderReadFile(decoder, avif, inputFilename);
-    if (decodeResult == AVIF_RESULT_OK) {
-        printf("Image decoded: %s\n", inputFilename);
-        printf("Image details:\n");
-        avifImageDump(avif, 0, 0);
+    decoder->allowProgressive = allowProgressive;
 
-        if (ignoreICC && (avif->icc.size > 0)) {
-            printf("[--ignore-icc] Discarding ICC profile.\n");
-            avifImageSetProfileICC(avif, NULL, 0);
-        }
+    avifResult result = avifDecoderSetIOFile(decoder, inputFilename);
+    if (result != AVIF_RESULT_OK) {
+        fprintf(stderr, "Cannot open file for read: %s\n", inputFilename);
+        returnCode = 1;
+        goto cleanup;
+    }
 
-        avifAppFileFormat outputFormat = avifGuessFileFormat(outputFilename);
-        if (outputFormat == AVIF_APP_FILE_FORMAT_UNKNOWN) {
-            fprintf(stderr, "Cannot determine output file extension: %s\n", outputFilename);
+    result = avifDecoderParse(decoder);
+    if (result != AVIF_RESULT_OK) {
+        fprintf(stderr, "ERROR: Failed to parse image: %s\n", avifResultToString(result));
+        returnCode = 1;
+        goto cleanup;
+    }
+
+    result = avifDecoderNthImage(decoder, frameIndex);
+    if (result != AVIF_RESULT_OK) {
+        fprintf(stderr, "ERROR: Failed to decode image: %s\n", avifResultToString(result));
+        returnCode = 1;
+        goto cleanup;
+    }
+
+    printf("Image decoded: %s\n", inputFilename);
+    printf("Image details:\n");
+    avifImageDump(decoder->image, 0, 0, decoder->progressiveState);
+
+    if (ignoreICC && (decoder->image->icc.size > 0)) {
+        printf("[--ignore-icc] Discarding ICC profile.\n");
+        avifImageSetProfileICC(decoder->image, NULL, 0);
+    }
+
+    avifAppFileFormat outputFormat = avifGuessFileFormat(outputFilename);
+    if (outputFormat == AVIF_APP_FILE_FORMAT_UNKNOWN) {
+        fprintf(stderr, "Cannot determine output file extension: %s\n", outputFilename);
+        returnCode = 1;
+    } else if (outputFormat == AVIF_APP_FILE_FORMAT_Y4M) {
+        if (!y4mWrite(outputFilename, decoder->image)) {
             returnCode = 1;
-        } else if (outputFormat == AVIF_APP_FILE_FORMAT_Y4M) {
-            if (!y4mWrite(outputFilename, avif)) {
-                returnCode = 1;
-            }
-        } else if (outputFormat == AVIF_APP_FILE_FORMAT_JPEG) {
-            // Bypass alpha multiply step during conversion
-            if (rawColor) {
-                avif->alphaPremultiplied = AVIF_TRUE;
-            }
-            if (!avifJPEGWrite(outputFilename, avif, jpegQuality, chromaUpsampling)) {
-                returnCode = 1;
-            }
-        } else if (outputFormat == AVIF_APP_FILE_FORMAT_PNG) {
-            if (!avifPNGWrite(outputFilename, avif, requestedDepth, chromaUpsampling)) {
-                returnCode = 1;
-            }
-        } else {
-            fprintf(stderr, "Unrecognized file extension: %s\n", outputFilename);
+        }
+    } else if (outputFormat == AVIF_APP_FILE_FORMAT_JPEG) {
+        // Bypass alpha multiply step during conversion
+        if (rawColor) {
+            decoder->image->alphaPremultiplied = AVIF_TRUE;
+        }
+        if (!avifJPEGWrite(outputFilename, decoder->image, jpegQuality, chromaUpsampling)) {
+            returnCode = 1;
+        }
+    } else if (outputFormat == AVIF_APP_FILE_FORMAT_PNG) {
+        if (!avifPNGWrite(outputFilename, decoder->image, requestedDepth, chromaUpsampling)) {
             returnCode = 1;
         }
     } else {
-        printf("ERROR: Failed to decode image: %s\n", avifResultToString(decodeResult));
-        avifDumpDiagnostics(&decoder->diag);
+        fprintf(stderr, "Unrecognized file extension: %s\n", outputFilename);
         returnCode = 1;
     }
-    avifDecoderDestroy(decoder);
-    avifImageDestroy(avif);
+
+cleanup:
+    if (decoder) {
+        if (returnCode != 0) {
+            avifDumpDiagnostics(&decoder->diag);
+        }
+        avifDecoderDestroy(decoder);
+    }
     return returnCode;
 }
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -37,8 +37,8 @@ static void syntax(void)
     printf("    -r,--raw-color    : Output raw RGB values instead of multiplying by alpha when saving to opaque formats\n");
     printf("                        (JPEG only; not applicable to y4m)\n");
     printf("    --index           : When decoding an image sequence or progressive image, specify which frame index to decode (Default: 0)\n");
-    printf("    --progressive     : Enable progressive AVIF processing. If a progressive image is encountered, avifdec will use\n");
-    printf("                        --index to choose which layer to decode (in progressive order).\n");
+    printf("    --progressive     : Enable progressive AVIF processing. If a progressive image is encountered and --progressive is passed,\n");
+    printf("                        avifdec will use --index to choose which layer to decode (in progressive order).\n");
     printf("    --no-strict       : Disable strict decoding, which disables strict validation checks and errors\n");
     printf("    -i,--info         : Decode all frames and display all image information instead of saving to disk\n");
     printf("    --ignore-icc      : If the input file contains an embedded ICC profile, ignore it (no-op if absent)\n");
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -38,7 +38,7 @@ static void syntax(void)
     printf("                        (JPEG only; not applicable to y4m)\n");
     printf("    --index           : When decoding an image sequence or progressive image, specify which frame index to decode (Default: 0)\n");
     printf("    --progressive     : Enable progressive AVIF processing. If a progressive image is encountered, avifdec will use\n");
-    printf("                        --index to choose which variant to decode (in progressive order).\n");
+    printf("                        --index to choose which layer to decode (in progressive order).\n");
     printf("    --no-strict       : Disable strict decoding, which disables strict validation checks and errors\n");
     printf("    -i,--info         : Decode all frames and display all image information instead of saving to disk\n");
     printf("    --ignore-icc      : If the input file contains an embedded ICC profile, ignore it (no-op if absent)\n");
--- a/apps/shared/avifutil.c
+++ b/apps/shared/avifutil.c
@@ -105,7 +105,6 @@ static void avifImageDumpInternal(const avifImage * avif, uint32_t gridCols, uin
             printf("    * imir (Mirror)        : Mode %u (%s)\n", avif->imir.mode, (avif->imir.mode == 0) ? "top-to-bottom" : "left-to-right");
         }
     }
-    printf(" * Alpha          : %s\n", alphaPresent ? (avif->alphaPremultiplied ? "Premultiplied" : "Not premultiplied") : "Absent");
     printf(" * Progressive    : %s\n", avifProgressiveStateToString(progressiveState));
 }
 
--- a/apps/shared/avifutil.c
+++ b/apps/shared/avifutil.c
@@ -38,7 +38,7 @@ static void printClapFraction(const char * name, int32_t n, int32_t d)
     printf(", ");
 }
 
-static void avifImageDumpInternal(const avifImage * avif, uint32_t gridCols, uint32_t gridRows, avifBool alphaPresent)
+static void avifImageDumpInternal(const avifImage * avif, uint32_t gridCols, uint32_t gridRows, avifBool alphaPresent, avifProgressiveState progressiveState)
 {
     uint32_t width = avif->width;
     uint32_t height = avif->height;
@@ -105,17 +105,19 @@ static void avifImageDumpInternal(const avifImage * avif, uint32_t gridCols, uin
             printf("    * imir (Mirror)        : Mode %u (%s)\n", avif->imir.mode, (avif->imir.mode == 0) ? "top-to-bottom" : "left-to-right");
         }
     }
+    printf(" * Alpha          : %s\n", alphaPresent ? (avif->alphaPremultiplied ? "Premultiplied" : "Not premultiplied") : "Absent");
+    printf(" * Progressive    : %s\n", avifProgressiveStateToString(progressiveState));
 }
 
-void avifImageDump(avifImage * avif, uint32_t gridCols, uint32_t gridRows)
+void avifImageDump(avifImage * avif, uint32_t gridCols, uint32_t gridRows, avifProgressiveState progressiveState)
 {
     const avifBool alphaPresent = avif->alphaPlane && (avif->alphaRowBytes > 0);
-    avifImageDumpInternal(avif, gridCols, gridRows, alphaPresent);
+    avifImageDumpInternal(avif, gridCols, gridRows, alphaPresent, progressiveState);
 }
 
 void avifContainerDump(avifDecoder * decoder)
 {
-    avifImageDumpInternal(decoder->image, 0, 0, decoder->alphaPresent);
+    avifImageDumpInternal(decoder->image, 0, 0, decoder->alphaPresent, decoder->progressiveState);
 }
 
 void avifPrintVersions(void)
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -760,8 +760,31 @@ typedef struct avifImageTiming
     uint64_t durationInTimescales; // duration in "timescales"
 } avifImageTiming;
 
+typedef enum avifProgressiveState
+{
+    // The current AVIF/Source does not offer a progressive image. This will always be the state
+    // for an image sequence.
+    AVIF_PROGRESSIVE_STATE_UNAVAILABLE = 0,
+
+    // The current AVIF/Source offers a progressive image, but avifDecoder.allowProgressive is not
+    // enabled, so it will behave as if the image was not progressive and will simply decode the
+    // best version of this item.
+    AVIF_PROGRESSIVE_STATE_AVAILABLE,
+
+    // The current AVIF/Source offers a progressive image, and avifDecoder.allowProgressive is true.
+    // In this state, avifDecoder.imageCount will be the count of all of the available progressive
+    // layers, and any specific layer can be decoded using avifDecoderNthImage() as if it was an
+    // image sequence, or simply using repeated calls to avifDecoderNextImage() to decode better and
+    // better versions of this image.
+    AVIF_PROGRESSIVE_STATE_ACTIVE
+} avifProgressiveState;
+AVIF_API const char * avifProgressiveStateToString(avifProgressiveState progressiveState);
+
 typedef struct avifDecoder
 {
+    // --------------------------------------------------------------------------------------------
+    // Inputs
+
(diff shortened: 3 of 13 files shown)


joedrago:
Some open questions, pinging folks directly (@leo-barnes, @cconcolato):

For some reason, the example file `animals_00_multilayer.avif` refuses to decode the primary item's AV1 payload when I use dav1d, but decodes perfectly fine when I use libaom or libgav1. I'd consider it to be a bug in libavif if it wasn't the only file (out of all AVIFs I've ever seen) with this specific issue. It doesn't give any specific error, I just don't think it is "enough" data to output a frame, at least for dav1d. I could use some help here on this one.

`animals_00_multilayer_grid_a1lx.avif` is an entirely different question. This is offering a _progressive grid_ image, but the earlier layers are not large enough to "fill the grid", as they are half resolution to the "real" grid cells. I fear that the only way this could possibly work is if these smaller cells are expected to be scaled prior to being stitched into a grid, which seems like it is taking these transformations too far. We already put in limitations on AVIF's derived image transformations for grid. Perhaps this can be another (e.g. "progressive grid cell variants must use fixed dimensions").

joedrago:
I've diagnosed the issue somewhat with `animals_00_multilayer.avif`, but I don't know what to do with the discrepancy between `dav1d` and `libaom`'s behavior. dav1d has this block:

```
    // skip lower spatial layers
    if (c->operating_point_idc && !c->all_layers) {
        const int max_spatial_id = ulog2(c->operating_point_idc >> 8);
        if (max_spatial_id > c->out.frame_hdr->spatial_id) {
            dav1d_picture_unref_internal(&c->out);
            return 0;
        }
    }
```
debug

This file's primary item does not signal any hint that it is multiple layers. In fact, it is one of the simplest types of primary items you can craft for AVIF (one extent, typical av1C/ispe/pixi/colr props, no alpha). Therefore, I don't configure any of the AV1 codec implementations (dav1d, aom, gav1) to have "all layers" turn on, as I just want this one AV1 payload's image.  As you can see above from the inspect variables and the logic here, dav1d is aware that there is another layer coming "later", so it discards the layer it has already encoded.

I feel like dav1d is doing the correct thing here and the file is invalid, but it happens to work just fine with libaom and libgav1, which I can't explain. I'm hoping either @leo-barnes or @wantehchang has some insight on this one.

leo-barnes:
@joedrago 
> As you can see above from the inspect variables and the logic here, dav1d is aware that there is another layer coming "later", so it discards the layer it has already decoded.

That can't be correct. It's fine if `dav1d` doesn't output the layer, but it can't discard it since future layers are predicted from it.

Either it's a bug in `dav1d`, or I somehow didn't correctly configure `libaom` when creating the two layers so it's not correctly signalling that the second layer is predicted from the first. @wantehchang, would it be possible for you to check if the second layer is correctly marked as using the first layer for prediction?

leo-barnes:
> animals_00_multilayer_grid_a1lx.avif is an entirely different question. This is offering a progressive grid image, but the earlier layers are not large enough to "fill the grid", as they are half resolution to the "real" grid cells. I fear that the only way this could possibly work is if these smaller cells are expected to be scaled prior to being stitched into a grid, which seems like it is taking these transformations too far. We already put in limitations on AVIF's derived image transformations for grid. Perhaps this can be another (e.g. "progressive grid cell variants must use fixed dimensions").

Spec-wise, this is perfectly valid:
```
NOTE: The dimensions of possible intermediate output images might not match the ones given in the ispe property. If they display these intermediate images, renderers are expected to scale the output image to match the ispe property.
```

If you want to support progressive refinement of grids, you will need to be prepared for intermediate layers having a different size than the tile size and will need to upscale them. If you don't want to do progressive refinement, you're guaranteed that the final layer (or the layer selected by `lsel`) will have the correct size.

joedrago:
I guess we should decide if we want AVIF implementations to be on the hook for YUV plane scaling.

wantehchang:
[Note: In this comment I use the variable names used in the AV1 spec to minimize the risk of confusion.]

I ran avifdec on `animals_00_multilayer.avif` in the debugger. The AV1 sequence header OBU has operating_points_cnt_minus_1=1 (which means there are two operating points) and operating_point_idc[0]=0x301 and operating_point_idc[1]=0x101.

The image item decoded by avifdec has only one AV1 frame OBU after the AV1 sequence header OBU. That AV1 frame OBU has obu_extension_flag=0, so it does not have an obu_extension_header() and its temporal_id and spatial_id are both 0 by default. Note: Under the draft revision of the AV1 spec, this AV1 frame OBU should set obu_extension_flag to 1 and have an obu_extension_header() with temporal_id=0 and spatial_id=0. It would be good to follow the new requirement.

The image item decoded by avifdec has two ispe properties, which violates the requirement in HEIF Section 6.5.3.1 that the quantity per item is one. The first ispe property has image_width=2048 and image_width=1536. The second ispe property has image_width=1024 and image_width=768.

Joe said `animals_00_multilayer.avif` has two image items. I think it should be changed to have only one image item. That image item should correspond to operatingPoint=0 and OperatingPointIdc=0x301 and have two AV1 Frame OBUs. And it should have only one ispe property.

Re: how an AV1 decoder should handle the current truncated AV1 bitstream in `animals_00_multilayer.avif`: I think the way dav1d handles this error condition is better.

joedrago:
@wantehchang It seems like the two outstanding issues I had (progressive grids and Leo's undefined-behavior file) are resolved sufficiently that you can review this at your leisure.

Remember that part of the most recent commit's design here (which uses `libyuv`'s `ScalePlane*()` functions) is that we would adopt/recontribute that small subset of functions into libavif, hopefully. Consider that as you review this, please.

apps/avifdec.c
```diff
@@ -36,6 +36,9 @@ static void syntax(void)
     printf("    -u,--upsampling U : Chroma upsampling (for 420/422). automatic (default), fastest, best, nearest, or bilinear\n");
     printf("    -r,--raw-color    : Output raw RGB values instead of multiplying by alpha when saving to opaque formats\n");
     printf("                        (JPEG only; not applicable to y4m)\n");
+    printf("    --index           : When decoding an image sequence or progressive image, specify which frame index to decode (Default: 0)\n");
+    printf("    --progressive     : Enable progressive AVIF processing. If a progressive image is encountered, avifdec will use\n");
+    printf("                        --index to choose which variant to decode (in progressive order).\n");
```
wantehchang, line 41: Should we change "variant" to "layer" or "frame"?

apps/shared/avifutil.c
```diff
@@ -105,17 +105,19 @@ static void avifImageDumpInternal(const avifImage * avif, uint32_t gridCols, uin
             printf("    * imir (Mirror)        : Mode %u (%s)\n", avif->imir.mode, (avif->imir.mode == 0) ? "top-to-bottom" : "left-to-right");
         }
     }
+    printf(" * Alpha          : %s\n", alphaPresent ? (avif->alphaPremultiplied ? "Premultiplied" : "Not premultiplied") : "Absent");
```
wantehchang, line 108: This line is the same as line 55 above. Is this line added by mistake?

include/avif/avif.h
```diff
@@ -772,6 +795,16 @@ typedef struct avifDecoder
     // Set this via avifDecoderSetSource().
     avifDecoderSource requestedSource;
 
+    // If this is true and a progressive AVIF is decoded, avifDecoder will behave as if the AVIF
+    // is an image sequence, in that it will set imageCount to the available number of progressive
```
wantehchang, line 799: Delete one of the two "available"s in "the available number of progressive frames available"

include/avif/avif.h
```diff
@@ -772,6 +795,16 @@ typedef struct avifDecoder
...
+    // frames available, and avifDecoderNextImage()/avifDecoderNthImage() will all for specific
+    // variants of a progressive image to be decoded. To distinguish between a progressive AVIF
```
wantehchang, line 801: "will all for specific variants of a progressive image to be decoded" sounds strange. The word "all" seems incorrect, but I don't know what word it should be changed to.

include/avif/avif.h
```diff
@@ -772,6 +795,16 @@ typedef struct avifDecoder
...
+    // and an AVIF image sequence, inspect avifDecoder.progressiveState.
+    avifBool allowProgressive;
+
+    // --------------------------------------------------------------------------------------------
+    // Outputs
```
wantehchang, line 806: I like splitting the avifDecoder struct members into three sections: Inputs, Outputs, and Internals. But the current division is not quite right.

For example, in the Output sections, `ignoreExif,` `ignoreXMP`, `imageCountLimit`, and `strictFlags` are inputs, not outputs. Perhaps they should be moved to the Inputs section?

Also, in the Internals section, `diag` is not really internal. Should we move it to the Output section?

include/avif/internal.h
```diff
@@ -244,6 +254,9 @@ typedef struct avifCodec
     struct avifCodecInternal * internal;  // up to each codec to use how it wants
                                           //
     avifDiagnostics * diag;               // Shallow copy; owned by avifEncoder or avifDecoder
+                                          //
+    uint8_t operatingPointIndex;          // Operating point Index, defaults to 0.
```
wantehchang, line 258: I suggest renaming this member `operatingPoint`. , because this corresponds to the `operatingPoint` variable in the AV1 spec.

Note: AV1 spec also has an `operating_point_idc[]` array and an `OperatingPointIdc` variable. The current name of this member, `operatingPointIndex`,  may potentially be misinterpreted as the `OperatingPointIdc` variable in the AV1 spec.

src/avif.c
```diff
@@ -101,6 +101,20 @@ const char * avifResultToString(avifResult result)
...
+    // clang-format off
+    switch (progressiveState) {
+        case AVIF_PROGRESSIVE_STATE_UNAVAILABLE: return "Unavailable";
+        case AVIF_PROGRESSIVE_STATE_AVAILABLE:   return "Available";
+        case AVIF_PROGRESSIVE_STATE_ACTIVE:      return "Active";
+        default:
+            break;
+    }
+    // clang-format on
+    return "Unknown";
```
wantehchang, line 115: Move this return statement to the default case in the switch statement.

wantehchang (changes requested):
Joe: I haven't reviewed the main changes in this PR. Here is the first round of review comments.

apps/avifdec.c
joedrago: Layer is probably better here.

apps/shared/avifutil.c
joedrago: Yes, fixing.

include/avif/avif.h
joedrago: It should be `will allow for specific layers`. Fixing.

src/avif.c
joedrago: This is the same layout as `avifPixelFormatToString()`. If we want to change this pattern in all functions, let's do it in a separate PR.

joedrago:
I've updated with most of the changes you requested. I left the default case return statement as it doesn't match other functions and we can adjust that style separately.

apps/avifdec.c
```diff
@@ -36,6 +36,9 @@ static void syntax(void)
     printf("    -u,--upsampling U : Chroma upsampling (for 420/422). automatic (default), fastest, best, nearest, or bilinear\n");
     printf("    -r,--raw-color    : Output raw RGB values instead of multiplying by alpha when saving to opaque formats\n");
     printf("                        (JPEG only; not applicable to y4m)\n");
+    printf("    --index           : When decoding an image sequence or progressive image, specify which frame index to decode (Default: 0)\n");
```
wantehchang, line 39: For a progressive image, it seems that the default should be the best layer, which has the largest frame index. Is it easy to implement that?

apps/avifdec.c
```diff
@@ -178,17 +189,31 @@ int main(int argc, char * argv[])
...
                        decoder->imageTiming.ptsInTimescales,
                        decoder->imageTiming.duration,
-                       decoder->imageTiming.durationInTimescales);
-                ++frameIndex;
+                       decoder->imageTiming.durationInTimescales,
+                       decoder->image->width,
+                       decoder->image->height);
+                ++currIndex;
+                nextImageResult = avifDecoderNextImage(decoder);
+            }
```
wantehchang, line 213: Lines 201-213 can be simplified slightly (only one avifDecoderNextImage call) if you are willing to do an assignment in the while loop's conditional expression:

```
            avifResult nextImageResult;
            while ((nextImageResult = avifDecoderNextImage(decoder)) == AVIF_RESULT_OK) {
                printf("   * Decoded frame [%d] [pts %2.2f (%" PRIu64 " timescales)] [duration %2.2f (%" PRIu64 " timescales)] [%ux%u]\n",
                       currIndex,
                       decoder->imageTiming.pts,
                       decoder->imageTiming.ptsInTimescales,
                       decoder->imageTiming.duration,
                       decoder->imageTiming.durationInTimescales,
                       decoder->image->width,
                       decoder->image->height);
                ++currIndex;
            }
```

apps/avifdec.c
```diff
@@ -210,52 +235,73 @@ int main(int argc, char * argv[])
...
+            returnCode = 1;
+        }
+    } else if (outputFormat == AVIF_APP_FILE_FORMAT_PNG) {
+        if (!avifPNGWrite(outputFilename, decoder->image, requestedDepth, chromaUpsampling)) {
             returnCode = 1;
         }
     } else {
-        printf("ERROR: Failed to decode image: %s\n", avifResultToString(decodeResult));
-        avifDumpDiagnostics(&decoder->diag);
+        fprintf(stderr, "Unrecognized file extension: %s\n", outputFilename);
```
wantehchang, line 295: Nit: If we reach here, we recognize the file extension (currently only .avif), but the file cannot be the input to avifenc. So this error message is a little confusing.

apps/avifdec.c
```diff
@@ -210,52 +235,73 @@ int main(int argc, char * argv[])
...
         returnCode = 1;
     }
-    avifDecoderDestroy(decoder);
-    avifImageDestroy(avif);
+
+cleanup:
+    if (decoder) {
```
wantehchang, line 300: This null pointer check for `decoder` is not necessary because `decoder` cannot be null. See line 238:

`avifDecoder * decoder = avifDecoderCreate();`

src/read.c
```diff
@@ -37,6 +37,8 @@ static const size_t xmpContentTypeSize = sizeof(xmpContentType);
 // can't be more than 4 unique tuples right now.
 #define MAX_IPMA_VERSION_AND_FLAGS_SEEN 4
 
+#define MAX_AV1_LAYER_COUNT 4
```
wantehchang, line 40: Optional: Would be good to comment that 4 is the maximum number of spatial layers in AV1.

src/read.c
```diff
@@ -361,11 +384,11 @@ static uint32_t avifGetSampleCountOfChunk(const avifSampleTableSampleToChunkArra
...
 
-static avifBool avifCodecDecodeInputGetSamples(avifCodecDecodeInput * decodeInput,
-                                               avifSampleTable * sampleTable,
-                                               const uint32_t imageCountLimit,
-                                               const uint64_t sizeHint,
-                                               avifDiagnostics * diag)
+static avifBool avifCodecDecodeInputFillFromSampleTable(avifCodecDecodeInput * decodeInput,
+                                                        avifSampleTable * sampleTable,
+                                                        const uint32_t imageCountLimit,
+                                                        const uint64_t sizeHint,
```
wantehchang, line 390: Nit/optional: Remove "const" from these two parameters.

src/read.c
```diff
@@ -451,6 +475,119 @@ static avifBool avifCodecDecodeInputGetSamples(avifCodecDecodeInput * decodeInpu
     return AVIF_TRUE;
 }
 
+static avifBool avifCodecDecodeInputFillFromDecoderItem(avifCodecDecodeInput * decodeInput,
+                                                        avifDecoderItem * item,
+                                                        avifBool allowProgressive,
+                                                        const uint32_t imageCountLimit,
+                                                        const uint64_t sizeHint,
```
wantehchang, line 482: Nit/optional: Remove "const" from these two parameters.

src/read.c
```diff
@@ -451,6 +475,119 @@ static avifBool avifCodecDecodeInputGetSamples(avifCodecDecodeInput * decodeInpu
...
+                                                        avifDiagnostics * diag)
+{
+    if (sizeHint && (item->size > sizeHint)) {
+        avifDiagnosticsPrintf(diag, "Exceeded avifIO's sizeHint, possibly truncated data");
+        return AVIF_FALSE;
+    }
+
+    uint8_t layerCount = 0;
```
wantehchang, line 490: Nit/optional: This local variable can be declared as `int`.

src/read.c
```diff
@@ -451,6 +475,119 @@ static avifBool avifCodecDecodeInputGetSamples(avifCodecDecodeInput * decodeInpu
...
+                layerSizes[i] = remainingSize;
+                remainingSize = 0;
+                break;
+            }
+        }
+        if (remainingSize > 0) {
+            assert(layerCount == 3);
+            ++layerCount;
+            layerSizes[3] = remainingSize;
+        }
```
wantehchang, line 518: Nit/optional: Lines 496 and 518 can be simplified if we declared the layerSize array in avifAV1LayeredImageIndexingProperty to have 4 elements and set the last element layerSize[3] to 0. Then the for loop at line 497 can have 4 iterations (` i < 4` ) and we can delete the if statement at line 514.

src/read.c
```diff
@@ -451,6 +475,119 @@ static avifBool avifCodecDecodeInputGetSamples(avifCodecDecodeInput * decodeInpu
...
+        size_t sampleSize = 0;
+        if (layerCount > 0) {
+            // Optimization: If we're selecting a layer that doesn't require the entire image's payload (hinted via the a1lx box)
+
+            if (lselProp->u.lsel.layerID >= layerCount) {
+                avifDiagnosticsPrintf(diag,
+                                      "lsel property requests layer index [%u] which isn't present in a1lx property ([%u] layers)",
+                                      lselProp->u.lsel.layerID,
+                                      layerCount);
+                return AVIF_FALSE;
```
wantehchang, line 539: IMPORTANT: This condition is not an error. It just means we cannot use the optimization. I think lines 531-548 should read:

```
        if (layerCount > 0 && lselProp->u.lsel.layerID < layerCount) {
            // Optimization: If we're selecting a layer that doesn't require the entire image's payload (hinted via the a1lx box)
            for (uint8_t i = 0; i <= lselProp->u.lsel.layerID; ++i) {
                sampleSize += layerSizes[i];
            }
        } else {
            // This layer's payload subsection is unknown, just use the whole payload
            sampleSize = item->size;
        }
```

src/read.c
```diff
@@ -451,6 +475,119 @@ static avifBool avifCodecDecodeInputGetSamples(avifCodecDecodeInput * decodeInpu
...
+            }
+
+            for (uint8_t i = 0; i <= lselProp->u.lsel.layerID; ++i) {
+                sampleSize += layerSizes[i];
```
wantehchang, line 543: [Just a comment, no changes requested] Note that `sampleSize` may be larger than the actual value if the `spatial_id` values are not consecutive, but at least `sampleSize` may be smaller than `item->size`.

src/read.c
```diff
@@ -451,6 +475,119 @@ static avifBool avifCodecDecodeInputGetSamples(avifCodecDecodeInput * decodeInpu
...
+            // This layer's payload subsection is unknown, just use the whole payload
+            sampleSize = item->size;
+        }
+
+        avifDecodeSample * sample = (avifDecodeSample *)avifArrayPushPtr(&decodeInput->samples);
+        sample->itemID = item->id;
+        assert(lselProp->u.lsel.layerID < MAX_AV1_LAYER_COUNT);
+        sample->offset = 0;
+        sample->size = sampleSize;
+        sample->skip = (uint8_t)lselProp->u.lsel.layerID;
```
wantehchang, line 555: IMPORTANT: This is incorrect if the `spatial_id` values are not consecutive.

src/read.c
```diff
@@ -451,6 +475,119 @@ static avifBool avifCodecDecodeInputGetSamples(avifCodecDecodeInput * decodeInpu
...
+        decodeInput->allLayers = AVIF_TRUE;
+
+        size_t offset = 0;
+        for (int i = 0; i < layerCount; ++i) {
+            avifDecodeSample * sample = (avifDecodeSample *)avifArrayPushPtr(&decodeInput->samples);
+            sample->itemID = item->id;
+            sample->offset = offset;
+            sample->size = layerSizes[i];
+            sample->skip = 0;
+            sample->sync = (i == 0); // Assume all layers depend on the first layer
```
wantehchang, line 574: Just wanted to confirm that things will still work if we set `sample->sync` to false incorrectly.

src/read.c
```diff
@@ -833,9 +990,17 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item, avifIO * io, avifR
         return AVIF_RESULT_TRUNCATED_DATA;
     }
 
-    size_t totalBytesToRead = item->size;
-    if (partialByteCount && (totalBytesToRead > partialByteCount)) {
-        totalBytesToRead = partialByteCount;
+    const size_t readOutputSize = (partialByteCount && (partialByteCount < item->size)) ? partialByteCount : item->size;
+
+    // Read in everything from the item's offset 0 all the way to the end of the read
```
wantehchang, line 995: Delete the "0" after "the item's offset"?

src/read.c
```diff
@@ -833,9 +990,17 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item, avifIO * io, avifR
         return AVIF_RESULT_TRUNCATED_DATA;
     }
 
-    size_t totalBytesToRead = item->size;
-    if (partialByteCount && (totalBytesToRead > partialByteCount)) {
-        totalBytesToRead = partialByteCount;
+    const size_t readOutputSize = (partialByteCount && (partialByteCount < item->size)) ? partialByteCount : item->size;
```
wantehchang, line 993: Should we replace `item->size` with `item->size - offset` (after making sure that `offset < item->size`)?

src/read.c
```diff
@@ -981,11 +1149,21 @@ static avifBool avifDecoderDataGenerateImageGridTiles(avifDecoderData * data, av
...
+
+                if (!alpha) {
+                    if (item->progressive) {
+                        decoder->progressiveState = AVIF_PROGRESSIVE_STATE_AVAILABLE;
+                        if (tile->input->samples.count > 1) {
+                            decoder->progressiveState = AVIF_PROGRESSIVE_STATE_ACTIVE;
+                            decoder->imageCount = tile->input->samples.count;
+                        }
+                    }
+                }
```
wantehchang, line 1166: Merge the nested ifs:

```
                if (!alpha && item->progressive) {
                    decoder->progressiveState = AVIF_PROGRESSIVE_STATE_AVAILABLE;
                    if (tile->input->samples.count > 1) {
                        decoder->progressiveState = AVIF_PROGRESSIVE_STATE_ACTIVE;
                        decoder->imageCount = tile->input->samples.count;
                    }
                }
```

src/read.c
```diff
@@ -1704,6 +1942,32 @@ static avifBool avifParseItemPropertyAssociation(avifMeta * meta, const uint8_t
...
+                    for (size_t i = 0; i < essentialTypesCount; ++i) {
+                        if (!memcmp(srcProp->type, essentialTypes[i], 4)) {
+                            avifDiagnosticsPrintf(diag,
+                                                  "Item ID [%u] has a %s property association which must be marked essential, but is not",
+                                                  itemID,
+                                                  essentialTypes[i]);
+                            return AVIF_FALSE;
+                        }
+                    }
+                }
```
wantehchang, line 1968: Would you like to check that a1lx is not marked as essential?

>   AVIF: Section 2.3.2.3.2: "If associated, it shall not be marked as essential."

src/read.c
```diff
@@ -2729,6 +2997,37 @@ avifResult avifDecoderParse(avifDecoder * decoder)
...
+            item->height = ispeProp->u.ispe.height;
+
+            if ((item->width == 0) || (item->height == 0) || (item->width > (AVIF_MAX_IMAGE_SIZE / item->height))) {
+                avifDiagnosticsPrintf(data->diag, "Item ID [%u] has an invalid size [%ux%u]", item->id, item->width, item->height);
+                return AVIF_RESULT_BMFF_PARSE_FAILED;
+            }
+        } else {
+            avifDiagnosticsPrintf(data->diag, "Item ID [%u] is missing a mandatory ispe property", item->id);
+            return AVIF_RESULT_BMFF_PARSE_FAILED;
+        }
```
wantehchang, line 3029: [Just a comment] I like this change. So far I have had to check if the container width and height is 0 after avifDecoderParse() returns successfully. I will be able to remove those checks in the future.

src/scale.c
```diff
@@ -0,0 +1,122 @@
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wstrict-prototypes" // "this function declaration is not a prototype"
+#endif
+#include <libyuv.h>
+#if defined(__clang__)
+#pragma clang diagnostic pop
+#endif
+
+// This should be configurable and/or smarter
+#define AVIF_LIBYUV_FILTER_MODE kFilterBox
```
wantehchang, line 29: Add a comment to note that `kFilterBox` has the highest quality (but is also the slowest).

But actually, since the low-resolution frame will be displayed only temporarily, it seems better to use the fastest filter mode.

src/scale.c
```diff
@@ -0,0 +1,122 @@
...
+        // Nothing to do
+        return AVIF_TRUE;
+    }
+
+    if ((dstWidth == 0) || (dstHeight == 0) || (dstWidth > (AVIF_MAX_IMAGE_SIZE / dstHeight))) {
+        avifDiagnosticsPrintf(diag, "avifImageScale requested invalid dst dimensions [%ux%u]", dstWidth, dstHeight);
+        return AVIF_FALSE;
+    }
+
+    uint8_t * srcYUVPlanes[AVIF_PLANE_COUNT_YUV];
```
wantehchang, line 43: Nit: const pointer:

`const uint8_t * srcYUVPlanes[AVIF_PLANE_COUNT_YUV];`

src/scale.c
```diff
@@ -0,0 +1,122 @@
...
+    for (int i = 0; i < AVIF_PLANE_COUNT_YUV; ++i) {
+        srcYUVPlanes[i] = image->yuvPlanes[i];
+        image->yuvPlanes[i] = NULL;
+        srcYUVRowBytes[i] = image->yuvRowBytes[i];
+        image->yuvRowBytes[i] = 0;
+    }
+    const avifBool srcImageOwnsYUVPlanes = image->imageOwnsYUVPlanes;
+    image->imageOwnsYUVPlanes = AVIF_FALSE;
+
+    uint8_t * srcAlphaPlane = image->alphaPlane;
```
wantehchang, line 54: Nit: const pointer:

const uint8_t * srcAlphaPlane = image->alphaPlane;

src/scale.c
```diff
@@ -0,0 +1,122 @@
...
+
+        avifPixelFormatInfo formatInfo;
+        avifGetPixelFormatInfo(image->yuvFormat, &formatInfo);
+        const uint32_t srcUVWidth = (srcWidth + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX;
+        const uint32_t srcUVHeight = (srcHeight + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY;
+        const uint32_t dstUVWidth = (dstWidth + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX;
+        const uint32_t dstUVHeight = (dstHeight + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY;
+
+        for (int i = 0; i < AVIF_PLANE_COUNT_YUV; ++i) {
+            if (!srcYUVPlanes[i]) {
```
wantehchang, line 77: Nit/optional: Instead of always doing three iterations and checking if srcYUVPlanes[i] is a null pointer, it would be better to check if `image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400` to determine the number of iterations (1 or 3) of this for loop. See the code in `avifImageAllocatePlanes()`.

src/scale.c
```diff
@@ -0,0 +1,122 @@
...
+                continue;
+            }
+
+            const uint32_t srcW = (i == AVIF_CHAN_Y) ? srcWidth : srcUVWidth;
+            const uint32_t srcH = (i == AVIF_CHAN_Y) ? srcHeight : srcUVHeight;
+            const uint32_t dstW = (i == AVIF_CHAN_Y) ? dstWidth : dstUVWidth;
+            const uint32_t dstH = (i == AVIF_CHAN_Y) ? dstHeight : dstUVHeight;
+            if (image->depth > 8) {
+                uint16_t * srcPlane = (uint16_t *)srcYUVPlanes[i];
```
wantehchang, line 86: Nit: const pointer:

const uint16_t * srcPlane = (const uint16_t *)srcYUVPlanes[i];

src/scale.c
```diff
@@ -0,0 +1,122 @@
...
+                avifFree(srcYUVPlanes[i]);
+            }
+        }
+    }
+
+    if (srcAlphaPlane) {
+        avifImageAllocatePlanes(image, AVIF_PLANES_A);
+
+        if (image->depth > 8) {
+            uint16_t * srcPlane = (uint16_t *)srcAlphaPlane;
```
wantehchang, line 105: Nit: const pointer:

const uint16_t * srcPlane = (const uint16_t *)srcAlphaPlane;

src/scale.c
```diff
@@ -0,0 +1,122 @@
...
+            const uint32_t srcW = (i == AVIF_CHAN_Y) ? srcWidth : srcUVWidth;
+            const uint32_t srcH = (i == AVIF_CHAN_Y) ? srcHeight : srcUVHeight;
+            const uint32_t dstW = (i == AVIF_CHAN_Y) ? dstWidth : dstUVWidth;
+            const uint32_t dstH = (i == AVIF_CHAN_Y) ? dstHeight : dstUVHeight;
+            if (image->depth > 8) {
+                uint16_t * srcPlane = (uint16_t *)srcYUVPlanes[i];
+                uint16_t * dstPlane = (uint16_t *)image->yuvPlanes[i];
+                ScalePlane_12(srcPlane, srcYUVRowBytes[i], srcW, srcH, dstPlane, image->yuvRowBytes[i], dstW, dstH, AVIF_LIBYUV_FILTER_MODE);
+            } else {
+                uint8_t * srcPlane = srcYUVPlanes[i];
```
wantehchang, line 90: Nit: const pointer:

const uint8_t * srcPlane = srcYUVPlanes[i];

src/scale.c
```diff
@@ -0,0 +1,122 @@
...
+
+    if (srcAlphaPlane) {
+        avifImageAllocatePlanes(image, AVIF_PLANES_A);
+
+        if (image->depth > 8) {
+            uint16_t * srcPlane = (uint16_t *)srcAlphaPlane;
+            uint16_t * dstPlane = (uint16_t *)image->alphaPlane;
+            ScalePlane_12(srcPlane, srcAlphaRowBytes, srcWidth, srcHeight, dstPlane, image->alphaRowBytes, dstWidth, dstHeight, AVIF_LIBYUV_FILTER_MODE);
+        } else {
+            uint8_t * srcPlane = srcAlphaPlane;
```
wantehchang, line 109: Nit: const pointer:

`const uint8_t * srcPlane = srcAlphaPlane;`

wantehchang (changes requested):
I have reviewed every file. I don't understand the new offset parameter to `avifDecoderItemRead()`. I will try to study that code.

src/scale.c
tongyuantongyu: I believe smooth version is preferred here. libjpeg-turbo even applies extra blur to smooth the decoded image: https://github.com/libjpeg-turbo/libjpeg-turbo/issues/459.

apps/avifdec.c
joedrago: If you don't pass `--progressive` at all, you will get the non-progressive behavior from the progressive AVIF, which will be the topmost/best layer.

apps/avifdec.c
joedrago: Perhaps "Unsupported input file extension" might be better wording here.

apps/avifdec.c
wantehchang: Thanks for the explanation. Perhaps in line 40 we can change

> If a progressive image is encountered,
to

> If a progressive image is encountered and --progressive is passed,

src/read.c
wantehchang: Since sample->skip is the spatial_id value, we can have the AV1 decoder output all layers and look for a matching spatial_id in the output frames.

- dav1d: check the `frame_hdr->spatial_id` field of the `Dav1dPicture` struct.
- libaom: check the `spatial_id` field of the `aom_image_t` struct. Unfortunately libaom has a bug in this area and I forgot to check in my bug fix: https://aomedia-review.googlesource.com/c/aom/+/133003. Sorry about that.
- libgav1: check the `spatial_id` field of the `Libgav1DecoderBuffer` struct.

wantehchang (changes requested):

src/read.c
```diff
@@ -451,6 +475,119 @@ static avifBool avifCodecDecodeInputGetSamples(avifCodecDecodeInput * decodeInpu
...
+
+        avifDecodeSample * sample = (avifDecodeSample *)avifArrayPushPtr(&decodeInput->samples);
+        sample->itemID = item->id;
+        assert(lselProp->u.lsel.layerID < MAX_AV1_LAYER_COUNT);
+        sample->offset = 0;
+        sample->size = sampleSize;
+        sample->skip = (uint8_t)lselProp->u.lsel.layerID;
+        sample->sync = AVIF_TRUE;
+    } else if (allowProgressive && a1lxProp) {
```
wantehchang, line 557: Nit: Replace a`1lxProp` with `item->progressive`:

`} else if (allowProgressive && item->progressive) {`

wantehchang (changes requested):

joedrago:
@wantehchang I have updated the `lsel` implementation to check `spatial_id` instead of an index-based skip. You said that there was some issue with libaom where we'll need to do a version check, but I'll defer writing the proper #ifdef to you.  I've also added in your changes above.

We also still need to address the scaling code and potential libyuv scaling recontribution.

wantehchang (approved):
Joe: Thank you for the new commits. I think this PR is ready to be merged. Please add a warning comment that the allowProgressive decoder setting should not be enabled except for experiments, because the implementation still has some issues and there is a small chance that the spec may change based on your implementation experience.

PR #642, opened 2021-05-27
Validate the num_channels field of the 'pixi' box

wantehchang:
Validate the num_channels field of the 'pixi' box against the av1C
property.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -738,7 +738,7 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifExt
     return AVIF_RESULT_OK;
 }
 
-static avifResult avifDecoderItemValidateAV1(const avifDecoderItem * item, avifDiagnostics * diag, const avifStrictFlags strictFlags)
+static avifResult avifDecoderItemValidateAV1(const avifDecoderItem * item, avifDiagnostics * diag, avifStrictFlags strictFlags, avifBool alpha)
 {
     const avifProperty * av1CProp = avifPropertyArrayFind(&item->properties, "av1C");
     if (!av1CProp) {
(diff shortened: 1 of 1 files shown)


joedrago:
We talked about this elsewhere, but just as a reminder for myself/others: I'm waiting on this one until I can ensure that the monochrome check is safe for really old AVIFs. (I can't remember how pixi was written in early versions of libavif before AV1 codecs could properly write monochrome.)

wantehchang:
Joe: I checked the relevant code in src/write.c.

In v0.3.5, the code is:

```
            if (hasAlpha) {
                avifBoxMarker pixiA = avifStreamWriteBox(&s, "pixi", 0, 0);
                avifStreamWriteU8(&s, 1);                     // unsigned int (8) num_channels;
```

In the current trunk, the code is:

```
        uint8_t channelCount = (item->alpha || (imageMetadata->yuvFormat == AVIF_PIXEL_FORMAT_YUV400)) ? 1 : 3;
        avifBoxMarker pixi = avifRWStreamWriteFullBox(&s, "pixi", AVIF_BOX_SIZE_TBD, 0, 0);
        avifRWStreamWriteU8(&s, channelCount); // unsigned int (8) num_channels;
```

So libavif has been setting num_channels to 1 in the 'pixi' box for the alpha auxiliary image for a long time.

I am requesting a sample image from Dirk Farin to see how libheif handles this.

farindk:
libheif behaves in the same way. Even if the encoder needs to save the alpha image in 4:2:0, libheif writes a one channel `pixi` box because there actually is only 1 channel of intended information. The other 2 channels are just dummy data without any use.

This is with libheif v1.12.0. Earlier versions of libheif did not write `pixi` boxes for AVIF.

PR #643, opened 2021-05-27
Require the 'ispe' property to be unique

wantehchang:
HEIF (ISO/IEC 23008-12:2017) Section 6.5.3.1 says:
  Box type: 'ispe'
  ...
  Mandatory (per item): Yes
  Quantity (per item): One

Also the diagnostic messages in avifParseItemPropertyContainerBox()
should say "Box[ipco]", not "Box[iprp]".

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1578,8 +1578,9 @@ static avifBool avifParsePixelInformationProperty(avifProperty * prop, const uin
 
 static avifBool avifParseItemPropertyContainerBox(avifPropertyArray * properties, const uint8_t * raw, size_t rawLen, avifDiagnostics * diag)
 {
-    BEGIN_STREAM(s, raw, rawLen, diag, "Box[iprp]");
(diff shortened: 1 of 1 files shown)


joedrago:
This is not valid for multiple reasons, unfortunately.

* The image which prompted you to write this review (`animals_00_multilayer.avif`) does have multiple `ispe` properties, but only one is actually associated to each of the two items in the file, which is correct.
* It is perfectly legal to have multiple properties of the same type in a single `ipco` box, as long as they're not multiply associated to the same item.

I think there might be a useful check to be added to `avifParseItemPropertyAssociation()` which avoids multiply _associating_ the same property type to a given item, but this check is incorrect and the original file is not in error. @cconcolato to confirm.

wantehchang:
Joe: Thank you for correcting me. We can let the AVIF Validator (ComplianceWarden) detect that error :-)

PR #653, opened 2021-05-29
Some straightforward changes to clapFraction code

wantehchang:
In calcCenter(), set the numerator to dim directly if dim is odd, rather
than calculate it from dim >> 1.

In clapFractionCD(), use *= whenever possible.

In avifCropRectConvertCleanApertureBox(), reduce halfW and halfH
immediately because the fractions widthN / widthD and heightN / heightD
have been verified to be integers.

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -400,8 +400,8 @@ static clapFraction calcCenter(int32_t dim)
     clapFraction f;
     f.n = dim >> 1;
     f.d = 1;
-    if ((dim % 2) == 1) {
-        f.n = (f.n * 2) + 1;
+    if ((dim % 2) != 0) {
+        f.n = dim;
         f.d = 2;
     }
     return f;
@@ -443,9 +443,9 @@ static void clapFractionCD(clapFraction * a, clapFraction * b)
     if ((a->d != b->d)) {
         const int32_t ad = a->d;
         const int32_t bd = b->d;
-        a->n = a->n * bd;
+        a->n *= bd;
         a->d *= bd;
-        b->n = b->n * ad;
+        b->n *= ad;
         b->d *= ad;
     }
 }


src/avif.c
```diff
@@ -400,8 +400,8 @@ static clapFraction calcCenter(int32_t dim)
     clapFraction f;
     f.n = dim >> 1;
     f.d = 1;
-    if ((dim % 2) == 1) {
-        f.n = (f.n * 2) + 1;
+    if ((dim % 2) != 0) {
```
wantehchang, line 403: The change from `(dim % 2) == 1` to `(dim % 2) != 0` is to handle the unlikely case that we pass a negative `dim` argument to this function.

src/avif.c
```diff
@@ -560,17 +560,17 @@ avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
     clapFraction croppedCenterY = clapFractionAdd(uncroppedCenterY, vertOff);
 
     clapFraction halfW;
-    halfW.n = widthN;
-    halfW.d = widthD * 2;
```
wantehchang, line 564: If the original code was written this way to avoid divisions, I can revert these two changes.

joedrago: Yes, my intent here was to avoid division. All of the other changes are good though.

wantehchang: Thanks for the reply. I will remove these two changes from this PR. Note that a good compiler should be able to combine these divisions with the earlier modulo operations when generating code for x86:

```
    if ((widthN % widthD) != 0) {
        avifDiagnosticsPrintf(diag, "[Strict] clap width %d/%d is not an integer", widthN, widthD);
        return AVIF_FALSE;
    }
    if ((heightN % heightD) != 0) {
        avifDiagnosticsPrintf(diag, "[Strict] clap height %d/%d is not an integer", heightN, heightD);
        return AVIF_FALSE;
    }
```

PR #659, opened 2021-06-02
Enable CMake configs in VCPKG mode [closes #658]

### LINKED ISSUE #658: Any particular reason not to install CMake configs in static mode?
Currently, avif installs CMake configs only in `SHARED` mode. But do we really have any particular reason not to install CMake configs in static mode?

If I re-enable installing CMake configs in static mode (by simply commenting out `if (BUILD_SHARED_LIBS)` at line 567), they are generated, installed, and used OK by a client application.

This prevents, for example, `vcpkg` integration from working correctly.

HappySeaFox:


joedrago:
Excellent! This is exactly the kind of targeted change I was hoping to see. I'll wait for this to survive CI, and then I'll merge it in. Thanks for the contribution.

wantehchang:
Joe: I just kicked off a CI build of this pull request.

wantehchang (approved):

PR #661, opened 2021-06-02
Check for int32_t overflows in 'clap' code

wantehchang:
Check if a uint32_t variable is <= INT32_MAX before casting it to
int32_t.

Check for overflows of int32_t arithmetic operations. Two methods are
used. 1) Perform the arithmetic operations in int64_t and then check for
int32_t overflows. 2) Check for potential overflows before performing
the arithmetic operations in int32_t.

Change clapFractionCD(), clapFractionAdd(), and clapFractionSub() to
return avifBool. They return AVIF_FALSE if any arithmetic operation
would overflow int32_t.

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -435,43 +435,66 @@ static void clapFractionSimplify(clapFraction * f)
     }
 }
 
+static avifBool overflowsInt32(int64_t x)
+{
+    return (x < INT32_MIN) || (x > INT32_MAX);
+}
+
 // Make the fractions have a common denominator
-static void clapFractionCD(clapFraction * a, clapFraction * b)
+static avifBool clapFractionCD(clapFraction * a, clapFraction * b)
 {
     clapFractionSimplify(a);
     clapFractionSimplify(b);
     if (a->d != b->d) {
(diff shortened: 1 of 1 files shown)


wantehchang:
Joe: Please take a look. This pull request is just one of several possible solutions. This pull request does not change the definition of the clapFraction struct, so the 'n' and 'd' fields are still int32_t. It detects and reports int32_t arithmetic overflows.

Other options include:
1. Detect int32_t arithmetic overflows and handle them with saturation, i.e., clamp the results to INT32_MIN or INT32_MAX.
2. Declare the 'n' and 'd' fields of the clapFraction struct as int64_t. (It also suffices to just declare 'n' as int64_t if we make sure 'd' is either 1 or 2.)
3. Port my ValidateClapProperty() function to libavif. The ValidateClapProperty() function can be found in the avif_image_decoder.cc file in https://chromium-review.googlesource.com/c/chromium/src/+/2919936.

wantehchang:
I just found a good way to avoid the `widthD * 2` and `heightD * 2` calculations. I will address that first and then resubmit this PR.

PR #665, opened 2021-06-03
Refactor imir implementation to match HEIF Draft Amendment 2

joedrago:
... , which replaces 'axis' with 'mode' and inverts the behavior

Due to misinterpretations of the original HEIF standard's imir description, most of the common implementations of HEIF interpret the `axis` bit incorrectly. The HEIF standard is in the process of being amended to match the prevailing implementations, which means AVIF must adjust as well. 

For those that do not query/leverage `axis` in their code, this is a benign change. For those that do, this will fail to compile in all of the places where they might need to invert their logic, which is expected and good.


wantehchang (approved):
LGTM. Thanks!

PR #668, opened 2021-06-04
Adjust gdk-pixbuf loader for new API change

novomesk:


contrib/gdk-pixbuf/loader.c
```diff
@@ -208,7 +208,7 @@ static gboolean avif_context_try_load(struct avif_context * context, GError ** e
     if (image->transformFlags & AVIF_TRANSFORM_IMIR) {
         GdkPixbuf *output_mirrored = NULL;
 
-        switch (image->imir.axis) {
+        switch (image->imir.mode) {
```
wantehchang, line 211: Daniel: Thank you for the pull request. Could you please confirm that we don't need to invert the FALSE/TRUE argument passed to the `gdk_pixbuf_flip()` calls at lines 213 and 216?

The meaning of `image->imir.mode` is as follows:
- 0: The top and bottom parts of the image are exchanged.
- 1: The left and right parts of the image are exchanged.

novomesk: It should stay unchanged. By coincidence, I used the current interpretation in the past too.

wantehchang: Thank you for the confirmation, Daniel.

wantehchang (approved):

PR #669, opened 2021-06-04
Q option

skal65535:
I find it practical to be able to set both min and max Quantizer value with one single option instead of using --min and --max.

Up to you if you think that's a good idea!


joedrago:
Hmm. I'm hesitant, but I think mainly for the choice of `-q`, as I don't want people to misinterpret it as "quality" and try to assume it is JPEG-ish. I'll think about it.

joedrago:
We need to be considerate of other AV1 encoders (currently rav1e and SVT). Do they offer a similar mechanism?

1480c1:
svt does have a qmin and qmax option, haven't tested with it much though

wantehchang:
Christopher: Thank you for your reply. Joe's question is the other way around: do rav1e and SVT-AV1 have an encoder option that sets the quantizer **value** as opposed to the quantizer **range** (min and max). (In libaom, the quantizer range is used to achieve a target bitrate.)

Joe: Both rav1e and SVT-AV1 have a similar mechanism.

rav1e: The `quantizer` field, which libavif already uses. Note that the `min_quantizer` field is intended for the bitrate mode:

```
/// - "quantizer": 0-255, default 100
...
/// - "min_quantizer": minimum allowed base quantizer to use in bitrate mode, default 0
/// - "bitrate": target bitrate for the bitrate mode (required for two pass mode), default 0
```

SVT-AV1: [The `-q` command-line option](https://github.com/AOMediaCodec/SVT-AV1/blob/master/Docs/svt-av1_encoder_user_guide.md) sets the "Quantization parameter used when RateControl is set to 0". libavif/src/codec_svt.c currently uses these options as follows:

```
        if (alpha) {
        } else {
            svt_config->qp = AVIF_CLAMP(encoder->maxQuantizer, 0, 63);
        }
```

I don't understand why we map `encoder->maxQuantizerAlpha` and `encoder->maxQuantize`r to different members of the svt_config struct.

vrabaud:
This was implemented in https://github.com/AOMediaCodec/libavif/pull/1138, with tests. So closing this one. Thx!

Issue #645, opened 2021-05-28
[Feature Request] How about -mt as well as -j N?


bat999:
Hi
**avifenc** and **avifdec** have an option...
`-j,--jobs J   : Number of jobs (worker threads, default: 1)`

That's OK, but **cwebp** and **dwebp** have an option ...
`-mt .... use multi-threading if available`
I suppose this means "all of them".

It would be neat if **avifenc** and **avifdec** had this option too.
So it would be...
`-j N`
or
`-mt`

:-)

bat999:
Or maybe don't bother with the "-mt" but change it to...
`-j,--jobs J : Number of jobs (worker threads, default: all)
`

bat999:
Until then...
`$ avifenc -j $(nproc) -o output.avif input.png`

joedrago:
I could see making a special `-j` option such as `all` or `max` which would effectively set it to what `nproc` emits (but in a cross-platform way).

bat999:
@joedrago 
Yes "**-j all**" will be OK if you're not interested in "**-mt**" like webp.

Do you think "-j,--jobs J : Number of jobs (worker threads, default: **all**)" is dangerous?

joedrago:
It's a good question; I'm not sure. It would be a pretty large default change to go from single-threaded to "use all cores" if `-j` isn't specified at all, but it might be a good change.

1480c1:
Would something like GNU make's rules like no `-j` == single, `-j` no argument passed == `$(nproc)`, and `-j arg` == `arg  jobs` be okay?

joedrago:
I don't support a single argument that can either have an argument or no argument, as the tool uses positional arguments for filenames. `-j` with no arguments is definitely out. I think `all` or `max` is a reasonable argument for `-j` and I'm still undecided on the default, but no-arg isn't a choice here.

1480c1:
If no-arg is out, then my personal leaning would be towards having the default being 1, since that way there's no unexpected changes for people casually updating and testing, and having `all` meaning `all of my cores` and potentially `max` meaning `maximum the library can support before breaking down`

1480c1:
Either that, or to preserve the idea of using numbers, maybe `-1` to mean `all`

bat999:
@joedrago 
Hi
They have done something different at the cavif program. 
Here ---> https://github.com/kornelski/cavif-rs

[...]

PR #670, opened 2021-06-07
avifenc, avifdec: Allow "-j all" to automatically use all of the cores on the machine

joedrago:
This will need to survive the builders first, but this is a pretty simple QoL feature for the `-j` argument. 

Fixes: #645

### DIFF
--- a/apps/shared/avifutil.c
+++ b/apps/shared/avifutil.c
@@ -220,3 +220,74 @@ void avifDumpDiagnostics(const avifDiagnostics * diag)
     printf("Diagnostics:\n");
     printf(" * %s\n", diag->error);
 }
+
+// ---------------------------------------------------------------------------
+// avifQueryMaxThreads (separated into OS implementations)
+
+#if defined(_WIN32)
+
+// Windows
+
+#pragma warning(disable : 5031)
+#pragma warning(disable : 5032)
+#include <windows.h>
+
+int avifQueryMaxThreads(void)
+{
+    int numCPU;
+    SYSTEM_INFO sysinfo;
+    GetSystemInfo(&sysinfo);
+    numCPU = sysinfo.dwNumberOfProcessors;
+    return numCPU;
+}
+
+#elif defined(__APPLE__)
+
+// Apple
+
+#include <sys/sysctl.h>
+
+int avifQueryMaxThreads()
+{
+    int mib[4];
+    int numCPU;
+    size_t len = sizeof(numCPU);
+
+    /* set the mib for hw.ncpu */
+    mib[0] = CTL_HW;
+    mib[1] = HW_AVAILCPU; // alternatively, try HW_NCPU;
+
+    /* get the number of CPUs from the system */
+    sysctl(mib, 2, &numCPU, &len, NULL, 0);
+
+    if (numCPU < 1) {
+        mib[1] = HW_NCPU;
+        sysctl(mib, 2, &numCPU, &len, NULL, 0);
+        if (numCPU < 1)
+            numCPU = 1;
+    }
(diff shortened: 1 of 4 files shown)


apps/shared/avifutil.c
```diff
@@ -220,3 +220,74 @@ void avifDumpDiagnostics(const avifDiagnostics * diag)
...
+
+#else
+
+// POSIX
+
+#include <unistd.h>
+
+int avifQueryMaxThreads()
+{
+    int numCPU = (int)sysconf(_SC_NPROCESSORS_ONLN);
```
1480c1, line 289: I vaguely remember something about BSD not having `_SC_NPROCESSORS_ONLN`, but don't remember what it used instead

joedrago: This is basically just a recontribution of my original block of code from `colorist`. I'm perfectly willing to adjust it to support more platforms as we encounter issues.

1480c1: Wait, nvm, it seems I had it backwards, it was that `_SC_NPROCESSORS_ONLN` was available on BSD, but some other options weren't available

joedrago:
@wantehchang Can you review this? I'd like to resolve this before the next version. Also, do you have an opinion on changing (or not changing) defaults for --jobs in the tooling?

apps/shared/avifutil.c
wantehchang: This kind of code is available in a lot of open-source libraries, so it is easy to find sample code. Some of the platform-specific code may no longer be necessary for the current versions of those Unix variants because they generally try to conform to POSIX or the Single Unix Specification or be compatible with Linux. For example, the code for __APPLE__ above may no longer be necessary for current versions of macOS and iOS.

A good strategy is to start with the POSIX code here and add platform-specific code only in response to bug report.

Re: BSD: The Chromium source code only has [a special case for OpenBSD](https://source.chromium.org/chromium/chromium/src/+/main:base/system/sys_info_openbsd.cc) in its POSIX code, which is similar to your __APPLE__ code above:

```
// static
int SysInfo::NumberOfProcessors() {
  int mib[] = {CTL_HW, HW_NCPU};
  int ncpu;
  size_t size = sizeof(ncpu);
  if (sysctl(mib, base::size(mib), &ncpu, &size, NULL, 0) < 0) {
    NOTREACHED();
    return 1;
  }
  return ncpu;
}
```

Chromium calls `sysconf(_SC_NPROCESSORS_CONF)` for the [general POSIX case](https://source.chromium.org/chromium/chromium/src/+/main:base/system/sys_info_posix.cc). Whether `_SC_NPROCESSORS_CONF` or `_SC_NPROCESSORS_ONLN` is more appropriate doesn't really matter in practice.

apps/shared/avifutil.h
```diff
@@ -24,6 +24,7 @@ void avifImageDump(avifImage * avif, uint32_t gridCols, uint32_t gridRows);
 void avifContainerDump(avifDecoder * decoder);
 void avifPrintVersions(void);
 void avifDumpDiagnostics(const avifDiagnostics * diag);
+int avifQueryMaxThreads(void);
```
wantehchang, line 27: This function should be named `avifQueryNumProcessors()` or `avifQueryNumCpus()` because that's what it returns ("to use all available cores").

We may want to document that this returns 1 if we cannot get the number of processors.

wantehchang (approved):
I suggest leaving the defaults at 1.

apps/shared/avifutil.h
joedrago: I've updated both of these.

PR #671, opened 2021-06-07
Cleanup avifutil.h comment to match libavif style

joedrago:
Minor: I noticed this C89-style block comment when adding `-j all` and wanted to have it match the other block comments in the codebase.


wantehchang (approved):
LGTM.

PR #672, opened 2021-06-08
Fix the clang -Wunused-macros warning

wantehchang:
Fix the clang -Wunused-macros (macro is not used) warning about the
HAVE_AOM_CODEC_SET_OPTION macro when libavif is configured to use libaom
for decode only (-DAVIF_CODEC_AOM_ENCODE=OFF
-DAVIF_CODEC_AOM_DECODE=ON).

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -35,10 +35,12 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if defined(AVIF_CODEC_AOM_ENCODE)
 // Detect whether the aom_codec_set_option() function is available. See aom/aom_codec.h
 // in https://aomedia-review.googlesource.com/c/aom/+/126302.
 #if AOM_CODEC_ABI_VERSION >= (6 + AOM_IMAGE_ABI_VERSION)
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -35,11 +35,13 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if defined(AVIF_CODEC_AOM_ENCODE)
 // Detect whether the aom_codec_set_option() function is available. See aom/aom_codec.h
 // in https://aomedia-review.googlesource.com/c/aom/+/126302.
 #if AOM_CODEC_ABI_VERSION >= (6 + AOM_IMAGE_ABI_VERSION)
 #define HAVE_AOM_CODEC_SET_OPTION 1
 #endif
+#endif
```
wantehchang, line 44: The `HAVE_AOM_CODEC_SET_OPTION` macro is only used in the encode part of this file, so put the definition of this macro inside a `#if defined(AVIF_CODEC_AOM_ENCODE)` block.

PR #680, opened 2021-06-17
Pass TestCase's minQuantizer, maxQuantizer, speed to encoder.

vrabaud:
The .json got updated accordingly.

### DIFF
--- a/tests/data/tests.json
+++ b/tests/data/tests.json
@@ -32,8 +32,8 @@
     "maxQP": 40,
     "speed": -1,
     "active": true,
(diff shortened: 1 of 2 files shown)


joedrago:
This is a nice change, and quite the oversight! Thanks.

wantehchang:
Joe: Should we open an issue to investigate the failure of "./aviftest ../tests/data" after this change?

joedrago:
I have a tracking issue for this here: https://github.com/AOMediaCodec/libavif/issues/684

Issue #679, opened 2021-06-17
MSYS2 build errors: unknown pragma ignored


eddiezato:
Commands to build:
```Bash
export CC=clang CXX=clang++
cd libavif
git pull
cd ext
./aom.cmd
./zlibpng.cmd
./libjpeg.cmd
cd ..
cmake -B build -G Ninja -S ./ \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=OFF \
    -DAVIF_CODEC_AOM=ON \
    -DAVIF_LOCAL_AOM=ON \
    -DAVIF_LOCAL_JPEG=ON \
    -DAVIF_LOCAL_ZLIBPNG=ON \
    -DAVIF_BUILD_APPS=ON \
    -DCMAKE_EXE_LINKER_FLAGS='-static' ..
ninja -C build
```
Received errors:
```
ninja: Entering directory `build'
[15/134] Building C object CMakeFiles/avifenc.dir/apps/shared/avifutil.c.obj
FAILED: CMakeFiles/avifenc.dir/apps/shared/avifutil.c.obj
D:\msys64\mingw64\bin\clang.exe  -Iext/zlib -Iext/libpng -I../include -I../apps/shared -I../ext/libpng -I../ext/libjpeg -O3 -DNDEBUG   -Weverything -Wno-bad-function-cast -Wno-cast-align -Wno-conversion -Wno-covered-switch-default -Wno-disabled-macro-expansion -Wno-documentation -Wno-documentation-unknown-command -Wno-double-promotion -Wno-float-equal -Wno-missing-noreturn -Wno-padded -Wno-sign-conversion -Wno-error=c11-extensions -Wno-poison-system-directories -Wno-used-but-marked-unused -Werror -std=gnu99 -MD -MT CMakeFiles/avifenc.dir/apps/shared/avifutil.c.obj -MF CMakeFiles\avifenc.dir\apps\shared\avifutil.c.obj.d -o CMakeFiles/avifenc.dir/apps/shared/avifutil.c.obj -c ../apps/shared/avifutil.c
../apps/shared/avifutil.c:231:9: error: unknown pragma ignored [-Werror,-Wunknown-pragmas]
#pragma warning(disable : 5031)
        ^
../apps/shared/avifutil.c:232:9: error: unknown pragma ignored [-Werror,-Wunknown-pragmas]
#pragma warning(disable : 5032)
        ^
2 errors generated.
[...]

PR #681, opened 2021-06-18
Don't need to disable MSVC warnings 5031 and 5032

wantehchang:
Since #pragma warning(push) and #pragma warning(pop) are not used in
this file, we cannot possibly get MSVC warnings 5031 and 5032. Therefore
it is not necessary to disable these two warnings.

Fix https://github.com/AOMediaCodec/libavif/issues/679.

### DIFF
--- a/apps/shared/avifutil.c
+++ b/apps/shared/avifutil.c
@@ -228,8 +228,6 @@ void avifDumpDiagnostics(const avifDiagnostics * diag)
 
 // Windows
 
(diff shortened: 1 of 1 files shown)


wantehchang:
@1480c1 @tongyuantongyu: Hi Christopher and Yuan Tong, please review this PR. Thanks!

1480c1 (approved):

PR #683, opened 2021-06-23
Make avifDecoderItemMaxExtent progressive-aware

joedrago:
This should make `avifDecoderNthImageMaxExtent()` behave correctly when in progressive mode.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -863,7 +863,7 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifDec
     uint64_t maxOffset = 0;
     for (uint32_t extentIter = 0; extentIter < item->extents.count; ++extentIter) {
         avifExtent * extent = &item->extents.extent[extentIter];
-        const size_t usedExtentSize = (extent->size > remainingBytes) ? extent->size : remainingBytes;
+        const size_t usedExtentSize = (extent->size < remainingBytes) ? extent->size : remainingBytes;
 
         if (usedExtentSize > UINT64_MAX - extent->offset) {
             return AVIF_RESULT_BMFF_PARSE_FAILED;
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -828,7 +828,7 @@ static void avifDecoderDataDestroy(avifDecoderData * data)
 // This returns the max extent that has to be read in order to decode this item. If
 // the item is stored in an idat, the data has already been read during Parse() and
 // this function will return AVIF_RESULT_OK with a 0-byte extent.
-static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifExtent * outExtent)
+static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifDecodeSample * sample, avifExtent * outExtent)
```
wantehchang, line 831: Please declare the `sample` parameter as const: `const avifDecodeSample * sample`

src/read.c
```diff
@@ -852,24 +852,40 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, avifExt
...
+    }
+    size_t remainingBytes = sample->size; // This may be smaller than item->size if the item is progressive
+
     // Assert that the for loop below will execute at least one iteration.
     assert(item->extents.count != 0);
     uint64_t minOffset = UINT64_MAX;
     uint64_t maxOffset = 0;
     for (uint32_t extentIter = 0; extentIter < item->extents.count; ++extentIter) {
         avifExtent * extent = &item->extents.extent[extentIter];
+        const size_t usedExtentSize = (extent->size > remainingBytes) ? extent->size : remainingBytes;
```
wantehchang, line 866: IMPORTANT: The ? : expression seems wrong. I think you intend to take the minimum, i.e.,

`const size_t usedExtentSize = (extent->size > remainingBytes) ? remainingBytes : extent->size;`

wantehchang (changes requested):

src/read.c
joedrago: Oh, that does look backwards. Fixing.

src/read.c
wantehchang: Please remember to fix this before merging.

wantehchang (approved):

wantehchang (approved):

PR #685, opened 2021-06-23
Set diagnostic message for aom_codec_set_option()

wantehchang:


wantehchang:
For example, if I pass `-a deltaq-mode=3` to avifenc,

Before:

`ERROR: Failed to encode image: Invalid codec-specific option`

After:

```
ERROR: Failed to encode image: Invalid codec-specific option
Diagnostics:
 * aom_codec_set_option("deltaq-mode", "3") failed: Invalid parameter: deltaq_mode out of range [..DELTA_Q_MODE_COUNT - 1]
```

PR #686, opened 2021-06-23
Add diagnostic messages for AV1 decode failures

wantehchang:
Add diagnostic messages for all AVIF_RESULT_DECODE_ALPHA_FAILED or
AVIF_RESULT_DECODE_COLOR_FAILED errors.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3556,12 +3556,14 @@ avifResult avifDecoderNextImage(avifDecoder * decoder)
         const avifDecodeSample * sample = &tile->input->samples.sample[nextImageIndex];
 
         if (!tile->codec->getNextImage(tile->codec, decoder, sample, tile->input->alpha, tile->image)) {
+            avifDiagnosticsPrintf(&decoder->diag, "tile->codec->getNextImage() failed");
             return tile->input->alpha ? AVIF_RESULT_DECODE_ALPHA_FAILED : AVIF_RESULT_DECODE_COLOR_FAILED;
         }
 
         // Scale the decoded image so that it corresponds to this tile's output dimensions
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -3623,12 +3626,14 @@ avifResult avifDecoderNextImage(avifDecoder * decoder)
...
         } else {
             if (decoder->data->alphaTileCount != 1) {
+                avifDiagnosticsPrintf(&decoder->diag, "decoder->data->alphaTileCount should be 1 but is %u", decoder->data->alphaTileCount);
                 return AVIF_RESULT_DECODE_ALPHA_FAILED;
             }
 
             avifImage * srcAlpha = decoder->data->tiles.tile[decoder->data->colorTileCount].image;
             if ((decoder->image->width != srcAlpha->width) || (decoder->image->height != srcAlpha->height) ||
                 (decoder->image->depth != srcAlpha->depth)) {
+                avifDiagnosticsPrintf(&decoder->diag, "decoder->image does not match srcAlpha in width, height, or bit depth");
```
joedrago, line 3636: This one is interesting. If we plan to scale to `ispe`, this entire check might be spurious now, right?

wantehchang: Hmm... Are you saying this error may not be a fatal error? One solution would be for the code that handles this error to clear the diagnostic message first.

Alternatively, we can document that decoder->diag.error should only be used after a public API function failure.

joedrago: I think it is safe to add this error while it continues to be a fatal error, but we should audit alpha size checks now that we are in this new world of ispe + ScalePlane.

PR #687, opened 2021-06-24
Fix avifDecoderItemMaxExtent() - compensate for sample->offset correctly

joedrago:
This should fix max extent calculations on progressive images.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -855,6 +855,7 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, const a
     if (sample->size == 0) {
         return AVIF_RESULT_TRUNCATED_DATA;
     }
+    size_t remainingOffset = sample->offset;
     size_t remainingBytes = sample->size; // This may be smaller than item->size if the item is progressive
 
     // Assert that the for loop below will execute at least one iteration.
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -855,6 +855,7 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, const a
     if (sample->size == 0) {
         return AVIF_RESULT_TRUNCATED_DATA;
     }
+    size_t remainingOffset = sample->offset;
```
wantehchang, line 858: Declare this as uint64_t, because `sample->offset` is uint64_t:

`    uint64_t remainingOffset = sample->offset;`

src/read.c
```diff
@@ -863,15 +864,31 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, const a
     uint64_t maxOffset = 0;
     for (uint32_t extentIter = 0; extentIter < item->extents.count; ++extentIter) {
         avifExtent * extent = &item->extents.extent[extentIter];
-        const size_t usedExtentSize = (extent->size < remainingBytes) ? extent->size : remainingBytes;
+
+        // Make local copies of extent->offset and extent->size as they might need to be adjusted
+        // due to the sample's offset.
+        uint64_t startOffset = extent->offset;
+        uint64_t extentSize = extent->size;
```
wantehchang, line 871: Declare this as size_t because `extent->size` is size_t:

`size_t extentSize = extent->size;`

src/read.c
```diff
@@ -863,15 +864,31 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, const a
...
+            }
+        }
+
+        const size_t usedExtentSize = (extentSize < remainingBytes) ? extentSize : remainingBytes;
 
         if (usedExtentSize > UINT64_MAX - extent->offset) {
             return AVIF_RESULT_BMFF_PARSE_FAILED;
         }
-        const uint64_t endOffset = extent->offset + usedExtentSize;
+        const uint64_t endOffset = startOffset + usedExtentSize;
```
wantehchang, line 888: The overflow check at line 885 needs to be updated:

        if (usedExtentSize > UINT64_MAX - startOffset) {
        }

src/read.c
```diff
@@ -863,15 +864,31 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, const a
...
+        // Make local copies of extent->offset and extent->size as they might need to be adjusted
+        // due to the sample's offset.
+        uint64_t startOffset = extent->offset;
+        uint64_t extentSize = extent->size;
+        if (remainingOffset) {
+            if (remainingOffset >= extentSize) {
+                remainingOffset -= extentSize;
+                continue;
+            } else {
+                startOffset += remainingOffset;
```
wantehchang, line 877: We may need to check for overflow before the addition:

```
        if (remainingOffset> UINT64_MAX - startOffset) {
        }
```

This check can be omitted if we know extent->offset + extent->size won't overflow.

wantehchang (changes requested):

joedrago:
I believe I've resolved all 4 review issues with https://github.com/AOMediaCodec/libavif/pull/687/commits/9732392b23a9a307aa3400b6a16db73d423ca39c.

wantehchang (approved):
LGTM. Thanks!

PR #688, opened 2021-06-25
Assert the presence of the appropriate alpha plane pointer in avifImageYUVAnyToRGBAnySlow()

joedrago:
This makes intentions clear to clang's analyzer and protect against any future changes to avifPrepareReformatState().

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -442,6 +442,11 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
     // These are the only supported built-ins
     assert((chromaUpsampling == AVIF_CHROMA_UPSAMPLING_BILINEAR) || (chromaUpsampling == AVIF_CHROMA_UPSAMPLING_NEAREST));
 
+    // If toRGBAlphaMode is active (not no-op), assert that the alpha plane is present. The end of
+    // the avifPrepareReformatState() function should ensure this, but this assert makes it clear
+    // to clang's analyzer.
+    assert((state->toRGBAlphaMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP) || aPlane);
+
     for (uint32_t j = 0; j < image->height; ++j) {
         const uint32_t uvJ = j >> state->formatInfo.chromaShiftY;
         const uint8_t * ptrY8 = &yPlane[j * yRowBytes];


src/reformat.c
```diff
@@ -457,6 +457,12 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
         uint8_t * ptrG = &rgb->pixels[state->rgbOffsetBytesG + (j * rgb->rowBytes)];
         uint8_t * ptrB = &rgb->pixels[state->rgbOffsetBytesB + (j * rgb->rowBytes)];
 
+        // If toAlphaMode is active (not no-op), assert that the appropriate alpha plane pointer is
```
wantehchang, line 460: Typo: toAlphaMode => toRGBAlphaMode

src/reformat.c
```diff
@@ -457,6 +457,12 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
...
+        // present (for this image's depth). The end of the avifPrepareReformatState() function should
+        // ensure this, but this assert makes it clear to clang's analyzer.
+        assert((state->toRGBAlphaMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP) ||
+               (((image->depth == 8) && ptrA8) || ((image->depth > 8) && ptrA16)));
```
wantehchang, line 464: Optional: The property ensured by the end of the avifPrepareReformatState() function is:

`assert((state->toRGBAlphaMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP) || aPlane);`

Would this simpler assertion work? This assertion can be moved outside the for loop, to line 436.

src/reformat.c
```diff
@@ -457,6 +457,12 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
...
+
```
wantehchang, line 465: An alternative is to move this assertion to the code that uses `ptrA8` and `ptrA16`, inside the if statement that starts at line 634:

```
            if (state->toRGBAlphaMode != AVIF_ALPHA_MULTIPLY_MODE_NO_OP) {
                // Calculate A
                uint16_t unormA;
                if (image->depth == 8) {
                    unormA = ptrA8[i];
                } else {
                    unormA = AVIF_MIN(ptrA16[i], yuvMaxChannel);
                }
                ...
```

Do you prefer the current location of the assertion because we assert immediately after initializing `ptrA8` and `ptrA16`?

wantehchang (changes requested):
Joe: I am sorry for the very late review. (I saw this PR at the beginning of my vacation and then forgot to review it.) Please fix a typo in the comment. I also suggest two possible changes.

joedrago:
Your simpler assert (outside of the loop) satisfies clang's analyzer, so I switched it.

wantehchang (approved):

PR #692, opened 2021-06-30
Update libavif.pc.cmake

kmilos:
See https://www.gnu.org/prep/standards/html_node/Directory-Variables.html and https://people.freedesktop.org/~dbn/pkg-config-guide.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -555,6 +555,12 @@ if(AVIF_BUILD_TESTS)
     endif()
 endif()
 
+if(NOT DEFINED CMAKE_INSTALL_LIBDIR)
+    set(CMAKE_INSTALL_LIBDIR "lib")
+endif()
+if(NOT DEFINED CMAKE_INSTALL_INCLUDEDIR)
+    set(CMAKE_INSTALL_INCLUDEDIR "include")
+endif()
 configure_file(libavif.pc.cmake ${CMAKE_CURRENT_BINARY_DIR}/libavif.pc @ONLY)
 
 if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -555,14 +555,6 @@ if(AVIF_BUILD_TESTS)
     endif()
 endif()
 
-if(NOT DEFINED CMAKE_INSTALL_LIBDIR)
-    set(CMAKE_INSTALL_LIBDIR "lib")
-endif()
-if(NOT DEFINED CMAKE_INSTALL_INCLUDEDIR)
-    set(CMAKE_INSTALL_INCLUDEDIR "include")
-endif()
-configure_file(libavif.pc.cmake ${CMAKE_CURRENT_BINARY_DIR}/libavif.pc @ONLY)
-
 if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
     install(TARGETS avif
         EXPORT ${PROJECT_NAME}-config
@@ -584,6 +576,7 @@ if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
                 DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME})
     endif()
 
+    configure_file(libavif.pc.cmake ${CMAKE_CURRENT_BINARY_DIR}/libavif.pc @ONLY)
     install(FILES ${CMAKE_CURRENT_BINARY_DIR}/libavif.pc
             DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
 endif()
--- a/libavif.pc.cmake
+++ b/libavif.pc.cmake
@@ -7,4 +7,4 @@ Name: @PROJECT_NAME@
 Description: Library for encoding and decoding .avif files
 Version: @PROJECT_VERSION@
 Libs: -L${libdir} -lavif
-Cflags: -I${includedir}/avif@AVIF_PKG_CONFIG_EXTRA_CFLAGS@
+Cflags: -I${includedir}@AVIF_PKG_CONFIG_EXTRA_CFLAGS@
--- a/libavif.pc.cmake
+++ b/libavif.pc.cmake
@@ -1,7 +1,7 @@
 prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=${prefix}
-libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+exec_prefix=@CMAKE_INSTALL_PREFIX@
+libdir=@CMAKE_INSTALL_FULL_LIBDIR@
+includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
 
 Name: @PROJECT_NAME@
 Description: Library for encoding and decoding .avif files
--- a/libavif.pc.cmake
+++ b/libavif.pc.cmake
@@ -1,10 +1,10 @@
 prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=${prefix}/bin
-libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/include
+exec_prefix=${prefix}
+libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
+includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
 
 Name: @PROJECT_NAME@
 Description: Library for encoding and decoding .avif files
 Version: @PROJECT_VERSION@
 Libs: -L${libdir} -lavif
-Cflags: -I${includedir}@AVIF_PKG_CONFIG_EXTRA_CFLAGS@
+Cflags: -I${includedir}/avif@AVIF_PKG_CONFIG_EXTRA_CFLAGS@


wantehchang:
@1480c1 Christopher: Could you please review this pull request? Thanks.

Miloš: Thank you for the pull request. I am going to take a look now and post some comments soon.

libavif.pc.cmake
```diff
@@ -1,10 +1,10 @@
...
+exec_prefix=${prefix}
+libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
+includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
 
 Name: @PROJECT_NAME@
 Description: Library for encoding and decoding .avif files
 Version: @PROJECT_VERSION@
 Libs: -L${libdir} -lavif
-Cflags: -I${includedir}@AVIF_PKG_CONFIG_EXTRA_CFLAGS@
+Cflags: -I${includedir}/avif@AVIF_PKG_CONFIG_EXTRA_CFLAGS@
```
1480c1, line 10: Just to make sure, is it recommended for users of libavif to use `#include <avif/avif.h>` or `#include <avif.h>`?

I see in https://github.com/AOMediaCodec/libavif/blob/master/apps/avifenc.c#L4 that @joedrago uses the former, so in that case, this change would be unnecessary or incorrect

CMakeLists.txt
```diff
@@ -555,6 +555,12 @@ if(AVIF_BUILD_TESTS)
     endif()
 endif()
 
+if(NOT DEFINED CMAKE_INSTALL_LIBDIR)
+    set(CMAKE_INSTALL_LIBDIR "lib")
+endif()
+if(NOT DEFINED CMAKE_INSTALL_INCLUDEDIR)
+    set(CMAKE_INSTALL_INCLUDEDIR "include")
+endif()
```
wantehchang, line 563: (Please note my question at the end of this comment.)

Earlier in this file at lines 456-458, we have:

```
if(NOT SKIP_INSTALL_ALL)
    include(GNUInstallDirs)
endif()
```

The [CMake documentation for GNUInstallDirs](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) says the GNUInstallDirs module defines the `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` variables. Therefore, it seems that it would be better to provide the default values of these two variables by rewriting lines 456-458 as follows:

```
if(NOT SKIP_INSTALL_ALL)
    include(GNUInstallDirs)
else()
    set(CMAKE_INSTALL_LIBDIR "lib")
    set(CMAKE_INSTALL_INCLUDEDIR "include")
endif()
```

**IMPORTANT:** I verified that this file uses `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` only if `NOT SKIP_INSTALL_ALL`. Does this mean the GNUInstallDirs module sometimes does not define these two variables?

libavif.pc.cmake
```diff
@@ -1,10 +1,10 @@
 prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=${prefix}/bin
-libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/include
+exec_prefix=${prefix}
+libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
+includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
```
1480c1, line 4: Not much to comment on if these are correct or not (they seem to be, but my experience of using pkg-config is more often using the provided Cflags and Libs rather than writing them), but I do know that nixOS people prefer just using the `_FULL_` version https://github.com/OpenVisualCloud/SVT-HEVC/pull/517. Not sure if it's something that should be considered, as it's somewhat niche.

wantehchang (commented):
The changes to libavif.pc.cmake are correct. I have a question about CMakeLists.txt.

libavif.pc.cmake
wantehchang: Good point. The apps and tests in libavif use `#include "avif/avif.h"`.

kmilos: Obviously extraneous then, I'll update the PR.

libavif.pc.cmake
kmilos: I haven't affected ${prefix} with this change (and the others were already relative, in an incorrect way though), so this is a separate issue.

CMakeLists.txt
kmilos: From my understanding the `configure_file(libavif.pc.cmake ...` is run regardless of `SKIP_INSTALL_ALL` so that's why a fallback was added for when `GNUInstallDirs` was not included. I'll change to above if you feel it is more readable...

Also, any reason not to include(GNUInstallDirs) always so we don't have to do this?

Alternatively, any reason not to run `configure_file(libavif.pc.cmake ...` only if `if(NOT SKIP_INSTALL_ALL)` as well?

libavif.pc.cmake
joedrago: `avif/avif.h` is the correct and intended include.

libavif.pc.cmake
kmilos: Btw, autotools can automagically replace @exec_prefix@, @libdir@, and @includedir@ without this relative scheme, so there is an option mimic those variables inside CMakeLists.txt, [like libheif does for example](https://github.com/strukturag/libheif/blob/bc5db10bd95c4fa3e66a2028e3c0da0745aff7a2/CMakeLists.txt#L119-L122).

CMakeLists.txt
wantehchang: Hi Miloš,

Thank you for the reply. Now that I understand the bug a little better, I probably would suggest running `configure_file(libavif.pc.cmake ...` only if `NOT SKIP_INSTALL_ALL`.

Since I am not familiar with this part of the build system, could you explain how to reproduce this bug so that I can take a closer look?

kmilos: Yes, this is what I did in the last commit, now `configure` of the .pc file is right next to the `install` of the .pc file.

Without this PR, you should still see the placeholder `libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@` inside the produced .pc file when `-DCMAKE_SKIP_INSTALL_RULES` is used for cmake configuration.

libavif.pc.cmake
```diff
@@ -1,7 +1,7 @@
 prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=${prefix}/bin
-libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/include
+exec_prefix=@CMAKE_INSTALL_PREFIX@
```
wantehchang, line 2: This is equivalent to `exec_prefix=${prefix}`, which you used earlier. I assume you think this is better.

wantehchang (approved):
@1480c1 Christopher: PTAL. Thanks!

1480c1 (approved):

CMakeLists.txt
wantehchang: > Without this PR, you should still see the placeholder `libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@` inside the produced .pc file when `-DCMAKE_SKIP_INSTALL_RULES` is used for cmake configuration.

If I pass `-DCMAKE_SKIP_INSTALL_RULES=ON` to cmake, cmake emits a warning:

```
CMake Warning in CMakeLists.txt:
  CMAKE_SKIP_INSTALL_RULES was enabled even though installation rules have
  been specified
```

And the produced libavif.pc file has the line

`libdir=${prefix}/`

rather than the line you described:

`libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@`

I guess you meant @CMAKE_INSTALL_LIBDIR@ has been substituted with an empty string?

kmilos: Seems to be the case indeed... incorrect/undesired anyway.

libavif.pc.cmake
kmilos: Yes, equivalent, and it's a toss-up I guess. We could either go w/ relative way based on ${prefix} for all the entries, or placeholders for all the entries. I don't really have a preference, but as you already commented, this way maybe lets some corner use cases override `CMAKE_INSTALL_FULL_LIBDIR` and `CMAKE_INSTALL_FULL_INCLUDEDIR`...

kmilos: The third option is to use (Autoconf default) placeholders `@prefix@`, `@exec_prefix@`, `@libdir@` and `@includedir@` and associated variables inside the CMakeLists.txt with these defaults (just like libheif and libtiff do as mentioned earlier), which maybe makes tweaking these a smidgeon easier (only one file to be patched).

wantehchang: I don't have a preference either. I just wanted to confirm you prefer the new form.

It seems that the variables `prefix` and `exec_prefix` are now defined but not used. Should we delete lines 1-2?

kmilos: Yeah, let's go w/ the PR as is. I think all the lines are potentially used by the pkg-config tool, even if they're not used in the .pc file directly, e.g. there are options to tweak the prefix and I presume the one in the file is the used to remove it from `includedir` for any replacements...

```
     --define-prefix
             Attempts to determine the prefix variable to use for CFLAGS and
             LIBS entry relocations.  This is mainly useful for platforms
             where framework SDKs are relocatable, such as Windows.

     --dont-define-prefix
             Disables the ‘define-prefix’ feature.

     --prefix-variable=VARIABLE
             Sets the ‘prefix’ variable used by the ‘define-prefix’ feature.
```

Issue #694, opened 2021-07-07
Avoid the heap allocation of aomImage plane buffers and copying of input image in aomCodecEncodeImage()


wantehchang:
`aomCodecEncodeImage()` calls `aom_img_alloc()` to allocate the plane buffers of `aomImage` from the heap and then copies the input image into the plane buffers in `aomImage`. This is not necessary. We can point the plane buffers in `aomImage` to the existing plane buffers in the input image.

Note: Since libaom is a C library, it is not clear which members of the `aom_image_t` struct are considered public and which members are considered private. I will also take a look at libaom documentation to see if this can be clarified.

PR #695, opened 2021-07-09
Do not copy input image when encoding with libaom

wantehchang:
Avoid the heap allocation of aomImage plane buffers and copying of input
image in aomCodecEncodeImage(). This requires manually setting up the
aom_image_t struct 'aomImage'.

Fix https://github.com/AOMediaCodec/libavif/issues/694.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -727,55 +727,71 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
 #endif
     }
 
-    int yShift = codec->internal->formatInfo.chromaShiftY;
-    uint32_t uvHeight = (image->height + yShift) >> yShift;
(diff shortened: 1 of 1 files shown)


joedrago:
This all looks good to me, although I can't really comment on the validity of the stack usage of `aom_image_t` or whether or not you're missing various members. Considering you're an expert on libaom, I trust that the usage is accurate, and the logic itself is sound.

wantehchang:
Joe: Thank you for taking a look. I found that there is no suitable public libaom function for this purpose. The closest function is `aom_img_wrap()`, but it assumes the buffers of the three planes are allocated together. This is why I resorted to setting the `aom_image_t` struct members directly. I imitated how `aom_img_alloc()` would set the struct members.

One issue I ran into is the `-Wswitch-enum` clang warning. That is why I rewrote the switch statement that set `bps` as an if-else-if-else statement.

PR #700, opened 2021-07-13
Call avifDecoderNextImage in while loop condition

wantehchang:
Call avifDecoderNextImage() in an assignment statement in the while
loop's conditional expression. This allows us to have only one
avifDecoderNextImage() call.


wantehchang:
Joe: I suggested this change in a code review before but am not sure if you missed it or didn't like it. Please feel free to decline this pull request.

PR #702, opened 2021-07-13
Remove unneeded null check for decoder

wantehchang:
Remove the unneeded null check for 'decoder' under the 'cleanup' label.
The 'decoder' variable is initialized as follows:

    avifDecoder * decoder = avifDecoderCreate();
    decoder->maxThreads = jobs;
    decoder->codecChoice = codecChoice;
    ...

so it cannot possibly be a null pointer.

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -301,11 +301,9 @@ int main(int argc, char * argv[])
     }
 
 cleanup:
-    if (decoder) {
-        if (returnCode != 0) {
-            avifDumpDiagnostics(&decoder->diag);
-        }
-        avifDecoderDestroy(decoder);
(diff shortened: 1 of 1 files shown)


apps/avifdec.c
```diff
@@ -301,11 +301,9 @@ int main(int argc, char * argv[])
     }
 
 cleanup:
-    if (decoder) {
-        if (returnCode != 0) {
-            avifDumpDiagnostics(&decoder->diag);
-        }
-        avifDecoderDestroy(decoder);
```
wantehchang, line 308: Joe: The current code does not need this null check for `decoder`. If you prefer to keep it, you can decline this pull request.

PR #703, opened 2021-07-14
Compare partialByteCount with item->size - offset

wantehchang:
Compare partialByteCount with item->size - offset rather than
item->size, because the read should start from 'offset'.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1031,18 +1031,13 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item,
         return AVIF_RESULT_TRUNCATED_DATA;
     }
 
-    const size_t readOutputSize = (partialByteCount && (partialByteCount < item->size)) ? partialByteCount : item->size;
-
-    // Read in everything from the item's offset 0 all the way to the end of the read
-    if (offset > SIZE_MAX - readOutputSize) {
+    if (offset >= item->size) {
         avifDiagnosticsPrintf(diag, "Item ID %u read has overflowing offset", item->id);
         return AVIF_RESULT_TRUNCATED_DATA;
     }
+    const size_t maxOutputSize = item->size - offset;
+    const size_t readOutputSize = (partialByteCount && (partialByteCount < maxOutputSize)) ? partialByteCount : maxOutputSize;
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -1031,18 +1031,13 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item,
...
-    const size_t readOutputSize = (partialByteCount && (partialByteCount < item->size)) ? partialByteCount : item->size;
-
-    // Read in everything from the item's offset 0 all the way to the end of the read
-    if (offset > SIZE_MAX - readOutputSize) {
+    if (offset >= item->size) {
         avifDiagnosticsPrintf(diag, "Item ID %u read has overflowing offset", item->id);
         return AVIF_RESULT_TRUNCATED_DATA;
     }
+    const size_t maxOutputSize = item->size - offset;
+    const size_t readOutputSize = (partialByteCount && (partialByteCount < maxOutputSize)) ? partialByteCount : maxOutputSize;
```
wantehchang, line 1039: Joe: Please review this PR carefully because it involves sizes and offsets.

The key point of this PR is that we should read at most `item->size - offset` bytes rather than `item->size` bytes. Therefore, if `partialByteCount` is specified, it should be compared with `item->size - offset` rather than `item->size`.

Also, since `readOutputSize <= maxOutputSize = item->size - offset`, it follows that `offset + readOutputSize <= offset + item->size - offset = item->size`, so `totalBytesRead` (calculated in the next line) cannot possibly overflow.

joedrago: I read it right when it hit my inbox. I can go over my thought process though as I read it:

* Switching the offset check to just be based on `item->size` is a good change.
* `maxOutputSize` will always be legal as the previous check ensures offset <= item->size
* readOutputSize always takes min, so it is also safe.

I did assume that you ran it and it still decodes normal and progressive files, but I think CI should catch that if it isn't true.

wantehchang: Joe: Thanks for the review. You understand the PR correctly. (The inequality in your second bullet point should use <, not <=.)

I tested this change on normal and progressive files.

Issue #706, opened 2021-07-14
avifdec: PNG compression level argument?


tannerhelland:
Hello.  Thank you for this great library.

Using avifdec to convert AVIFs to PNG is currently much slower than JPEG (10x or worse on large images).  A bulk of the time difference is spent in DEFLATE compression, which isn't necessary for many use-cases.

libpng provides a built-in `png_set_compression_level` function [for setting the underlying zlib compression level to 0-9](https://refspecs.linuxbase.org/LSB_3.1.0/LSB-Desktop-generic/LSB-Desktop-generic/libpng12.png.set.compression.level.1.html).  Level 0 specifies "no compression" and is extremely useful when speed is preferable to file size.  

Exposing this setting to users of avifdec would allow for much faster PNG conversion.

(Overloading the existing `-q --quality` command-line argument to specify compression for PNGs is one option... but it's also confusing because it has nothing to do with quality.  Maybe `-z --zlib` or similar?)

Thank you!

PR #707, opened 2021-07-14
avifdec: Add PNG compression level arg

joedrago:
Fixes: #706

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -33,6 +33,7 @@ static void syntax(void)
     printf("    -c,--codec C      : AV1 codec to use (choose from versions list below)\n");
     printf("    -d,--depth D      : Output depth [8,16]. (PNG only; For y4m, depth is retained, and JPEG is always 8bpc)\n");
     printf("    -q,--quality Q    : Output quality [0-100]. (JPEG only, default: %d)\n", DEFAULT_JPEG_QUALITY);
+    printf("    --png-compress L  : Set PNG compression level (PNG only; 0-9, 0=none, 9=max). Defaults to libpng's builtin default.\n");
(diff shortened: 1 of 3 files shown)


apps/avifdec.c
```diff
@@ -115,6 +117,14 @@ int main(int argc, char * argv[])
             } else if (jpegQuality > 100) {
                 jpegQuality = 100;
             }
+        } else if (!strcmp(arg, "--png-compress")) {
+            NEXTARG();
+            pngCompressionLevel = atoi(arg);
+            if (pngCompressionLevel < 0) {
+                pngCompressionLevel = 0;
```
bnason-nf, line 124: Would it make sense to print something here if you clamp the value?

apps/shared/avifpng.h
```diff
@@ -8,6 +8,10 @@
 
 // if (requestedDepth == 0), do best-fit
 avifBool avifPNGRead(const char * inputFilename, avifImage * avif, avifPixelFormat requestedFormat, uint32_t requestedDepth, uint32_t * outPNGDepth);
-avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint32_t requestedDepth, avifChromaUpsampling chromaUpsampling);
+avifBool avifPNGWrite(const char * outputFilename,
+                      const avifImage * avif,
+                      uint32_t requestedDepth,
+                      avifChromaUpsampling chromaUpsampling,
+                      int compressionLevel);
```
wantehchang, line 15: Document that `compressionLevel` is ignored if negative.

apps/avifdec.c
wantehchang: Or fail with an "invalid PNG compression level" message.

wantehchang (approved):

PR #708, opened 2021-07-14
Check that a1lx is not marked as essential

wantehchang:


joedrago:
LGTM (waiting for CI)

PR #682, opened 2021-06-23
Re-map libavif-libaom speed settings

fkyslov:
Make libavif speed setting to map 1-to-1 to libaom speed setting.
Adjust usage for speed >= 7 according to AOM_USAGE_ALL_INTRA
availability and sequence coding.

PR #710, opened 2021-07-15
Changes to comments and CHANGELOG.md

wantehchang:


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
+* Mandate ispe and disallow zero width or height (#640).
```
wantehchang, line 10: Joe: This change is part of your "progressive" branch but it is independent of progressive decoding, so I wanted to make sure we describe it separately in the CHANGELOG.

Could you also check the "Re-map libavif speed 7-10 to libaom speed 7-9 (#682)}" item at lines 11-16? This item has multiple lines. I am not sure if we wrote it correctly.

src/codec_aom.c
```diff
@@ -127,6 +127,7 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
         nextFrame = aom_codec_get_frame(&codec->internal->decoder, &codec->internal->iter);
         if (nextFrame) {
             if (spatialID != AVIF_SPATIAL_ID_UNSET) {
+                // This requires a version of libaom with the fix for https://crbug.com/aomedia/2993.
```
wantehchang, line 130: I decided to not do a version check here, because I expect the use of the 'lsel' property to be rare.

The version check will need to be a runtime version check because it is not possible to detect this bug fix at compile time. (There are no suitable libaom macros.)

PR #713, opened 2021-07-16
Add the --no-strict-pixi option to avifdec

wantehchang:
Add an easy way to verify that the missing pixi property is the only
problem with an AVIF image.

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -40,6 +40,7 @@ static void syntax(void)
     printf("    --progressive     : Enable progressive AVIF processing. If a progressive image is encountered and --progressive is passed,\n");
     printf("                        avifdec will use --index to choose which layer to decode (in progressive order).\n");
     printf("    --no-strict       : Disable strict decoding, which disables strict validation checks and errors\n");
+    printf("    --no-strict-pixi  : Allow the PixelInformationProperty ('pixi') to be missing in AV1 image items\n");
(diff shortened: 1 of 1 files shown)


wantehchang:
Joe: If you think this option is not generally useful, please feel free to reject this PR. Thanks!

joedrago:
Hmm. I guess this change is fine, but I'm curious when it is used? We already print out a very clear diagnostic error when the reason an image fails is because of pixi (or any other `[Strict]` check).

wantehchang:
Joe: Since avifdec stops parsing after the pixi check fails, one cannot tell if the image has other problems. This option allows us to check if the missing pixi property is the only problem.

joedrago:
My only remaining comment would be that this feels very specific / incomplete. Should we have `--no-strict-*` for all strict flags? It seems weird to single out this specific flag as a commandline switch.

wantehchang:
Because not all strict checks are equal. Because libheif is used in two popular image tools (libvips, used by other libraries, and ImageMagick), the missing pixi property problem is encountered much more often.

I don't mind making this change manually when I need to debug this kind of problem though. If you think this option could be confusing, I can close this PR.

joedrago:
I think if it benefits your workflow, we can add it. Based on other conversations though, I could see this specific flag being deprecated at some point and this new switch being moot.

wantehchang:
I am going to close this pull request. I can make the `strictFlags &= ~AVIF_STRICT_PIXI_REQUIRED` change when I need it.

Issue #715, opened 2021-07-16
avifImageScale() passes incorrect src_stride and dst_stride arguments to libyuv's ScalePlane_12() function


wantehchang:
This bug was discovered by Google's internal fuzzing (b/193792566). I guess the reason oss-fuzz has not detected this bug is that oss-fuzz does not build libavif with libyuv enabled.

The libyuv-based version of `avifImageScale()` calls libyuv's `ScalePlane_12()` function as follows:

```
            if (image->depth > 8) {
                uint16_t * srcPlane = (uint16_t *)srcYUVPlanes[i];
                uint16_t * dstPlane = (uint16_t *)image->yuvPlanes[i];
                ScalePlane_12(srcPlane, srcYUVRowBytes[i], srcW, srcH, dstPlane, image->yuvRowBytes[i], dstW, dstH, AVIF_LIBYUV_FILTER_MODE);
            } else {
```

The `src_stride` (second) and `dst_stride` (sixth) arguments should be in uint16_t samples, so we need to divide row bytes by 2.

PR #716, opened 2021-07-16
Pass correct src and dst strides to ScalePlane_12

wantehchang:
Fix https://github.com/AOMediaCodec/libavif/issues/715.


joedrago:
Should those local stride variables be const?

wantehchang:
I can declare the local stride variables as const. (I wasn't sure what your preference was.) Would you like me to declare the local planes variables as const for consistency? For example,

`uint8_t * const dstPlane = image->yuvPlanes[i];`

joedrago:
Yes, I think so.

wantehchang:
Done. Please review the new commit. Thanks.

PR #723, opened 2021-07-23
In 32-bit builds set frame_size_limit to 8192*8192

wantehchang:
This avoids the following dav1d_log() message in 32-bit builds:
    Frame size limit reduced from 268435456 to 67108864.


joedrago:
Do we want to alter `AVIF_MAX_IMAGE_SIZE` based on the build type as well, or is this just to fix dav1d?

wantehchang:
This is just to fix dav1d. In our own code we can use the larger AVIF_MAX_IMAGE_SIZE because our goal is different. We aim to prevent int32_t arithmetic overflows in our own code and our clients' code, not out-of-memory errors.

joedrago (approved):

PR #725, opened 2021-07-28
Copy input image data if width or height is 1

wantehchang:
Commit ea3e97e9af6623de16d60f336c9d96ef79ba06ce eliminated the
unnecessary copying of input image data into the aom_image_t struct, but
it exposed a libaom bug (https://crbug.com/aomedia/3113). Work around
the libaom bug by resurrecting the original code that copies input image
data and using it when the image width or height is equal to 1.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -730,47 +730,70 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     }
 
     aom_image_t aomImage;
-    memset(&aomImage, 0, sizeof(aomImage));
-    aomImage.fmt = codec->internal->aomFormat;
-    aomImage.bit_depth = (image->depth > 8) ? 16 : 8;
-    aomImage.w = image->width;
-    aomImage.h = image->height;
-    aomImage.d_w = image->width;
-    aomImage.d_h = image->height;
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -830,6 +879,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     }
     aom_codec_err_t encodeErr = aom_codec_encode(&codec->internal->encoder, &aomImage, 0, 1, encodeFlags);
     avifFree(monoUVPlane);
+    aom_img_free(&aomImage);
```
joedrago, line 882: Is this safe to call here in all cases? This seems like it'll perform frees on memory it doesn't own?

wantehchang: This turns out to be safe, because `aom_img_free` is defined as follows:

```
void aom_img_free(aom_image_t *img) {
  if (img) {
    aom_img_remove_metadata(img);
    if (img->img_data && img->img_data_owner) aom_free(img->img_data);

    if (img->self_allocd) free(img);
  }
}
```

I added a test for `aomImageAllocated` to make this clearer.

joedrago:
LGTM.

PR #727, opened 2021-08-03
Skip YUV/RGB conversion tests except first image

wantehchang:
Skip the YUV<->RGB conversion tests except for the first loop iteration
(loop == 0) and the first image of an image sequence. The YUV<->RGB
conversion tests are time-consuming for large images, so we should try
to avoid duplicate tests.


wantehchang:
Hi Joe,

Google's internal fuzzer tests reported timeout errors recently. In one case, the fuzzer test input is an image sequence of three images of size 12600x13072. So I am looking into various ways to reduce the time avif_decode_fuzzer spends on the YUV<->RGB conversion tests. This is my first PR related to this.

For the YUV<->RGB conversion tests in avif_decode_fuzzer, the pixel values do not really matter to detection of memory errors. Therefore it suffices to run these tests only for the first image of an image sequence, and only in the first loop iteration (`loop == 0`).

PR #729, opened 2021-08-04
Make image size limit configurable

wantehchang:
Adapted from Joe Drago's pull request
https://github.com/AOMediaCodec/libavif/pull/527, with the limitation
that decoder->imageSizeLimit must be less than or equal to the default
value and must not be set to 0 (reserved for future use). This way we
don't need to audit our code for integer overflows due to a large image
width or height.

Set decoder->imageSizeLimit to 11 * 1024 * 10 * 1024 in
avif_decode_fuzzer.cc to keep its memory consumption under 2560 MB.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -68,6 +68,10 @@ typedef int avifBool;
 
 #define AVIF_DIAGNOSTICS_ERROR_BUFFER_SIZE 256
 
+// A reasonable default for maximum image size to avoid out-of-memory errors or integer overflow in
+// (32-bit) int or unsigned int arithmetic operations.
+#define AVIF_DEFAULT_IMAGE_SIZE_LIMIT (16384 * 16384)
+
 // a 12 hour AVIF image sequence, running at 60 fps (a basic sanity check as this is quite ridiculous)
 #define AVIF_DEFAULT_IMAGE_COUNT_LIMIT (12 * 3600 * 60)
 
@@ -809,6 +813,12 @@ typedef struct avifDecoder
     // If you don't actually leverage this data, it is best to ignore it here.
     avifBool ignoreExif;
     avifBool ignoreXMP;
+
+    // This represents the maximum size of a image (in pixel count) that libavif and the underlying
+    // AV1 decoder should attempt to decode. It defaults to AVIF_DEFAULT_IMAGE_SIZE_LIMIT, and can be
+    // set to a smaller value. The value 0 is reserved.
+    // Note: Only some underlying AV1 codecs support a configurable size limit (such as dav1d).
+    uint32_t imageSizeLimit;
 
     // This provides an upper bound on how many images the decoder is willing to attempt to decode,
     // to provide a bit of protection from malicious or malformed AVIFs citing millions upon
--- a/tests/oss-fuzz/avif_decode_fuzzer.cc
+++ b/tests/oss-fuzz/avif_decode_fuzzer.cc
@@ -18,6 +18,11 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t * Data, size_t Size)
     static size_t yuvDepthsCount = sizeof(yuvDepths) / sizeof(yuvDepths[0]);
 
     avifDecoder * decoder = avifDecoderCreate();
+    // ClusterFuzz passes -rss_limit_mb=2560 to avif_decode_fuzzer. Empirically setting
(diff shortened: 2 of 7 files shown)


include/avif/avif.h
```diff
@@ -68,6 +68,10 @@ typedef int avifBool;
 
 #define AVIF_DIAGNOSTICS_ERROR_BUFFER_SIZE 256
 
+// A reasonable default for maximum image size to avoid out-of-memory errors or integer overflow in
+// (32-bit) int or unsigned int arithmetic operations.
+#define AVIF_DEFAULT_MAX_IMAGE_SIZE (16384 * 16384)
```
wantehchang, line 73: Joe: Would you like me to rename this macro `AVIF_DEFAULT_IMAGE_SIZE_LIMIT` to match the naming convention of `AVIF_DEFAULT_IMAGE_COUNT_LIMIT`? (See line 76 below.)

include/avif/avif.h
```diff
@@ -810,6 +814,12 @@ typedef struct avifDecoder
     avifBool ignoreExif;
     avifBool ignoreXMP;
 
+    // This represents the maximum size of a image (in pixel count) that libavif and the underlying
+    // AV1 decoder should attempt to decode. It defaults to AVIF_DEFAULT_MAX_IMAGE_SIZE, and can be
+    // set to a smaller value. The value 0 is reserved.
+    // Note: Only some underlying AV1 codecs support a configurable size limit (such as dav1d).
+    uint32_t imageSizeLimit;
```
wantehchang, line 821: Joe: Any preference for whether the `imageSizeLimit` member should be declared before or after the `imageCountLimit` member?

tests/oss-fuzz/avif_decode_fuzzer.cc
```diff
@@ -18,6 +18,11 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t * Data, size_t Size)
     static size_t yuvDepthsCount = sizeof(yuvDepths) / sizeof(yuvDepths[0]);
 
     avifDecoder * decoder = avifDecoderCreate();
+    // ClusterFuzz passes -rss_limit_mb=2560 to avif_decode_fuzzer. Empirically setting
+    // decoder->imageSizeLimit to this value allows avif_decode_fuzzer to consume no more than
+    // 2560 MB of memory.
+    static_assert(11 * 1024 * 10 * 1024 <= AVIF_DEFAULT_MAX_IMAGE_SIZE, "");
+    decoder->imageSizeLimit = 11 * 1024 * 10 * 1024;
```
wantehchang, line 25: We will need to set decoder->imageSizeLimit to an even smaller value to prevent avif_decode_fuzzer from timing out in the YUV<->RGB conversion tests. But I haven't done the experiments to figure out what that value should be.

Alternatively, we can add a dedicated test to call the YUV<->RGB conversion functions with all kinds of sizes and options, and remove the YUV<->RGB conversion tests from avif_decode_fuzzer.

wantehchang (commented):
Joe: Your PR also contains changes to avifdec.c. Would you like me to include those changes in this PR?

I didn't include your avifdec.c changes because my goal is to set `decoder->imageSizeLimit` to a smaller value in avif_decode_fuzzer.cc.

wantehchang:
> Joe: Your PR also contains changes to avifdec.c. Would you like me to include those changes in this PR?

You can ignore this question now because I have included your changes to avifdec.c in the PR locally. I will push a new commit after you answer my other questions. Thanks.

include/avif/avif.h
joedrago: Either order here is acceptable.

include/avif/avif.h
joedrago: I don't have a strong preference here, but having them match might be nice.

joedrago (approved):
Other than possibly renaming the macro to the LIMIT style, LGTM.

include/avif/avif.h
wantehchang: Done. `AVIF_DEFAULT_MAX_IMAGE_SIZE` has been renamed `AVIF_DEFAULT_IMAGE_SIZE_LIMIT`.

wantehchang:
Joe: I just pushed a new commit. Please take a look. Thanks.

You just need to review avifdec.c, which is new in this commit. The only other change in the new commit is the renaming of the AVIF_DEFAULT_MAX_IMAGE_SIZE macro.

joedrago (approved):

Issue #263, opened 2020-08-10
Add a maxImageSize decoder setting to avifDecoder.


wantehchang:
In the following two commits I added a hardcoded AVIF_MAX_IMAGE_SIZE:

https://github.com/AOMediaCodec/libavif/commit/0a8e7244d494ae98e9756355dfbfb6697ded2ff9
https://github.com/AOMediaCodec/libavif/commit/2fb636141296abdf608f381aa006ddad21a9507a

Ideally the maximum image size should be a configurable decoder setting. We can add a maxImageSize decoder setting to the avifDecoder struct.

We will need to validate the maxImageSize value. Also, to support large maxImageSize values, some arithmetic operations involving the image width, height, or row bytes that currently use the int or unsigned int type will need to use the size_t type instead. We also need to handle a 32-bit size_t correctly.

wantehchang:
Fixed by commit 980d585b6f43da0a599fd0e1674ac2ba85e9a880 and commit bd1492e1812fff842c31c0f4da84d48f1349265c.

PR #732, opened 2021-08-06
Add diagnostic messages if imageSizeLimit exceeded

wantehchang:
Add custom diagnostic messages if item size, track size, or grid
dimensions exceed imageSizeLimit.

Related to https://github.com/AOMediaCodec/libavif/issues/263.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1631,8 +1631,12 @@ static avifBool avifParseImageGridBox(avifImageGrid * grid, const uint8_t * raw,
         CHECK(avifROStreamReadU32(&s, &grid->outputWidth));  // unsigned int(FieldLength) output_width;
         CHECK(avifROStreamReadU32(&s, &grid->outputHeight)); // unsigned int(FieldLength) output_height;
     }
-    if ((grid->outputWidth == 0) || (grid->outputHeight == 0) || (grid->outputWidth > (imageSizeLimit / grid->outputHeight))) {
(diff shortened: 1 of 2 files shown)


src/read.c
```diff
@@ -2365,10 +2369,14 @@ static avifBool avifParseTrackHeaderBox(avifTrack * track, const uint8_t * raw,
...
 
-    if ((track->width == 0) || (track->height == 0) || (track->width > (imageSizeLimit / track->height))) {
+    if ((track->width == 0) || (track->height == 0)) {
         avifDiagnosticsPrintf(diag, "Track ID [%u] has an invalid size [%ux%u]", track->id, track->width, track->height);
         return AVIF_FALSE;
     }
+    if (track->width > (imageSizeLimit / track->height)) {
+        avifDiagnosticsPrintf(diag, "Track ID [%u] size is too large [%ux%u]", track->id, track->width, track->height);
+        return AVIF_FALSE;
+    }
```
wantehchang, line 2379: Joe: I was planning to add the AVIF_RESULT_IMAGE_TOO_LARGE error code as we previously discussed. But I found that it would take a lot of plumbing (changing three or four functions) to surface the new error code from here. (The grid and item sizes are easy to handle.) In the end I decided to just reply on diagnostic messages and not add a new error code.

joedrago (approved):
I like the new messages, LGTM.

Issue #734, opened 2021-08-11
Certain scale values can cause overflow in libyuv


FrankGalligan:
This bug was discovered by Google's internal fuzzing.

In parts of libyuv they convert values to fixed point (16.16) integers. Certain scale values can cause integer overflow. 

In one example the src width is 36681 and the dst width is 49. In the scaling code this will generate a max value of 2367705984, which is > 2^31 -1 and causes an integer overflow.

joedrago:
Thanks for the report. I'll open a PR to try to mitigate this.

FrankGalligan:
Hi @joedrago I put up a PR.

Kind of filed the bug and out up the PR, but you were too fast on the interrupt :)

FrankGalligan:
But if you have something better in mind, that is fine.

FrankGalligan:
I'm also going to see if more scale factors could be handled in libyuv.

joedrago:
Heh, I replied quickly, but I hadn't started writing any code just yet (nobody is _that_ fast). I've left a few quick comments on your PR.

PR #735, opened 2021-08-11
Check for scale values that cause overflow

FrankGalligan:
This addresses #734

### DIFF
--- a/src/scale.c
+++ b/src/scale.c
@@ -64,6 +64,12 @@ avifBool avifImageScale(avifImage * image, uint32_t dstWidth, uint32_t dstHeight
     image->height = dstHeight;
 
     if (srcYUVPlanes[0]) {
+        const int fixed_point_div = (int)(((int64_t)(srcWidth) << 16) / dstWidth);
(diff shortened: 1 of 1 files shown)


joedrago:
A few observations:

* Don't we need to do similar checks for height? Or is this issue with libyuv fixed point math only width-related?
* Style: Can you change your `snake_case` variable names to `camelCase` for consistency? (`fixedPointDiv`, `maxFixedValue`)

FrankGalligan:
I didn't add a check for height because the code is different and looks to have a check that might address it. I'm actually hoping this can be fixed better in libyuv, I just don't know the code that well to make a "fix" there.  If you want I can add the same check for height, I don't think it would hurt as it will perform the same scaling.

As for the `snake_case`, my bad I should have looked more in the file first.

joedrago:
I'm also not familiar enough with libyuv's internals, but if it works entirely in fixed point for dimensions, I think putting a height check here would be good. We can always relax these restrictions by libyuv version if it gains support for larger sizes.

FrankGalligan:
ok sounds good

FrankGalligan:
I also asked that question of libyuv author, will see what they say.

PR #736, opened 2021-08-19
When creating the read buffer in avifDecoderItemRead(), always make the buffer the item's full size

joedrago:
This avoids progressive image partial item reads causing reallocs of this buffer, which can cause
dangling pointers in the underlying AV1 decoders (use-after-free). As the natural pattern of this
type of usage is to ultimately decode the final image, those additional reallocations are also
typically unnecessary overhead, as we'll eventually be using this sized buffer anyway.

This should address Chromium issue 1239472.


wantehchang (approved):
LGTM. Thanks!

PR #737, opened 2021-08-19
New define in mem.c DEBUG_USE_AFTER_FREE_ON_WINDOWS which helps catch use-after-free issues in a debugger on Windows

joedrago:
This is a low priority PR, but this was how I successfully reproduced PR #736 in a debugger, so I figured it would be good to have it checked-in, in case we need to repro use-after-free issues in the future.

### DIFF
--- a/src/mem.c
+++ b/src/mem.c
@@ -5,6 +5,13 @@
 
 #include <stdlib.h>
 
+// Enable this to catch use-after-free issues when debugging on Windows.
+// #define DEBUG_USE_AFTER_FREE_ON_WINDOWS
+
+#if !defined(DEBUG_USE_AFTER_FREE_ON_WINDOWS)
+
+// The standard implementation, using malloc() and free().
+
 void * avifAlloc(size_t size)
 {
     void * out = malloc(size);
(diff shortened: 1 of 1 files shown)


src/mem.c
```diff
@@ -18,3 +25,72 @@ void avifFree(void * p)
...
+// should immediately catch it in a debugger right when it happens.
+
+#include <stdio.h>
+#include <windows.h>
+
+typedef struct Allocation
+{
+    void * ptr;
+    size_t originalSize;
+    size_t size;
```
wantehchang, line 44: Optional: If I understand the MSDN pages for VirtualAlloc() and VirtualProtect() correctly, we can just pass the original size to these two functions and they will round the size up to the next page boundary for us.

wantehchang (approved):

wantehchang:
Here is a screen shot of Visual Studio installer showing "C++ AddressSanitizer" under "Desktop development with C++":

Visual Studio Installer 9_7_2021 3_23_32 PM

joedrago:
This PR can sit for a while. I'd want to make sure I could catch #736 in the debugger when using this VS feature first. 

One advantage of this PR is it doesn't require any special VS magic; simply building against the Win32 API and it busts in the right spot.

PR #740, opened 2021-08-22
Allocate alpha alongside YUV (if necessary) during y4m decode to avoid incorrect alphaRowBytes math

joedrago:
There is code later in the function that relies on `alphaRowBytes` being correct, but it isn't set until the alpha plane  is allocated. This simply moves the alpha plane allocation a bit earlier, and fixes `yuva444p` decoding.


wantehchang (approved):
LGTM.

PR #742, opened 2021-08-26
Add docker build CI pipeline

paskal:
As discussed in #620, this PR adds docker build CI pipeline. Here is the pipeline run in my repo: https://github.com/paskal/libavif/runs/3429529148

The next step I propose to take after this change will be merged is to start publishing resulting images so that users could benefit from the pre-build library.

### DIFF
--- a/.github/workflows/docker-ci.yml
+++ b/.github/workflows/docker-ci.yml
@@ -1,0 +1,20 @@
+name: Docker Build
+on:
+  push:
+    branches:
+    tags:
+    paths:
+      - "!**.md"
+  pull_request:
+    paths:
+      - "!**.md"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: build docker image
+        run: docker build  ./tests/docker


paskal:
Gentle ping  @wantehchang @EwoutH @joedrago.

wantehchang (approved):

wantehchang:
@EwoutH Could you also review this PR? Thanks.

paskal:
Another gentle ping for @EwoutH.

EwoutH:
Hi, thanks for the ping, sorry I missed the first one.

In general, it looks good. I like that documentation changes are excluded. 18 minutes runtime is quite long however, I think we could do something with caching. All the dependencies are fixed versions so we should be able to build libaom, dav1d and rav1e once and cache them somewhere. Maybe setup something ([like this](https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching)), but that adds a lot of complexity so is fine for another PR.

Anyway, as long as we have a Dockerfile it’s good to test it in CI, so lets get this in and then discuss possible improvements.

EwoutH (approved):

PR #744, opened 2021-08-28
Compute maxFixedValueWidth in 64-bit operation

wantehchang:
Improve the check added in
https://github.com/AOMediaCodec/libavif/pull/735.

Compute maxFixedValueWidth and maxFixedValueHeight in 64-bit
multiplication by casting the first operand to int64_t. Otherwise, the
first operand is an int and the second operand is a uint32_t, so the
multiplication is 32-bit.

Replace 0x7fffffff with the symbolic constant INT_MAX because the
product will be stored in an int variable.

### DIFF
--- a/src/scale.c
+++ b/src/scale.c
@@ -17,6 +17,8 @@ avifBool avifImageScale(avifImage * image, uint32_t dstWidth, uint32_t dstHeight
 
 #else
 
+#include <limits.h>
+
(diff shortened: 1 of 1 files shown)


wantehchang:
@FrankGalligan Please review. Thanks!

FrankGalligan (commented):
LGTM Thanks!

PR #745, opened 2021-09-01
NON-STANDARD: Allow alpha subimage items to not have an ispe property, infer ispe from color item

joedrago:
See: https://crbug.com/1245673

Despite it being mandatory for `ispe` to be associated with all image items, libavif did not explicitly block missing `ispe` on subimages until recently, which is seen as a regression for websites using encoders that produce nonstandard files that Chrome used to tolerate/decode. This is a (hopefully temporary) solution to this perceived regression.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3095,7 +3095,7 @@ avifResult avifDecoderParse(avifDecoder * decoder)
                 avifDiagnosticsPrintf(data->diag, "Item ID [%u] size is too large [%ux%u]", item->id, item->width, item->height);
                 return AVIF_RESULT_BMFF_PARSE_FAILED;
             }
-        } else {
+        } else if (!item->auxForID) { // NON-STANDARD: Allow auxiliary images to not have an ispe property. See: https://crbug.com/1245673
             avifDiagnosticsPrintf(data->diag, "Item ID [%u] is missing a mandatory ispe property", item->id);
             return AVIF_RESULT_BMFF_PARSE_FAILED;
         }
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -3081,6 +3081,11 @@ avifResult avifDecoderParse(avifDecoder * decoder)
             // probably exif or some other data
             continue;
         }
+        if (item->auxForID) {
+            // NON-STANDARD: Allow subimages to not have an ispe property
```
wantehchang, line 3085: Nit: Change "subimages" to "auxiliary images"?

The reason is that we also refer to the primary image item as the "color subimage" in some comments and messages.

src/read.c
```diff
@@ -3081,6 +3081,11 @@ avifResult avifDecoderParse(avifDecoder * decoder)
...
+            // See: https://crbug.com/1245673
+            continue;
+        }
```
wantehchang, line 3088: At the end of avifDecoderItemValidateAV1(), we have this code:

```
    if (strictFlags & AVIF_STRICT_CLAP_VALID) {
        const avifProperty * clapProp = avifPropertyArrayFind(&item->properties, "clap");
        if (clapProp) {
            const avifProperty * ispeProp = avifPropertyArrayFind(&item->properties, "ispe");
            if (!ispeProp) {
                avifDiagnosticsPrintf(diag,
                                      "[Strict] Item ID %u is missing an ispe property, so its clap property cannot be validated",
                                      item->id);
            }

            ...
        }
    }
```

Should we also allow the auxiliary images to not have an ispe property there?

wantehchang (changes requested):
Joe: Thank you very much for writing this PR! I suggest a comment change and have a question.

src/read.c
joedrago: I've updated this comment to use `auxiliary images`.

src/read.c
joedrago: I don't believe auxiliary images will have a `clap` property (it will typically only be on the primary item), so this should be benign. I'd rather keep this strict.

joedrago:
@wantehchang I've updated `avifDecoderParse()` to properly adopt the alpha `ispe`, if present.

wantehchang (approved):
LGTM. Thanks!

PR #750, opened 2021-09-04
Fix: change name for david 0.9.2+

Jamaika1:


joedrago:
I believe @wantehchang was considering a PR for this name change, but is there a way you can detect when to use the new API, and when to use the old API? The change is definitely _after_ v0.9.2, so we need some sort of ifdef/test to support old and new versions of dav1d.

This might mean that we need to wait for `v0.9.3` and switch on that version, if there is no other signal.

src/codec_dav1d.c
```diff
@@ -57,8 +57,8 @@ static avifBool dav1dCodecGetNextImage(struct avifCodec * codec,
 {
     if (codec->internal->dav1dContext == NULL) {
         // Give all available threads to decode a single frame as fast as possible
-        codec->internal->dav1dSettings.n_frame_threads = 1;
-        codec->internal->dav1dSettings.n_tile_threads = AVIF_CLAMP(decoder->maxThreads, 1, DAV1D_MAX_TILE_THREADS);
+        codec->internal->dav1dSettings.n_threads = 1;
+        codec->internal->dav1dSettings.max_frame_delay = AVIF_CLAMP(decoder->maxThreads, 1, 256);
```
1480c1, line 61: Perhaps it's better to use `DAV1D_MAX_THREADS`  instead of a bare number

1480c1, line 61: You might want to use
```c
#if defined(DAV1D_API_VERSION_MAJOR) && DAV1D_API_VERSION_MAJOR >= 6
// new fields and code here
#else
// old code here
#endif
```
in order to keep compile compatibility with older dav1d versions and installs

Jamaika1: `DAV1D_MAX_THREADS` is delete in david 0.9.2+. I guess that changed the value from 64 to 256.

1480c1: I see it at https://code.videolan.org/videolan/dav1d/-/merge_requests/1252/diffs#95378459063068eb66ef004414047378fc28a963_49_46 and at https://code.videolan.org/videolan/dav1d/-/blob/master/include/dav1d/dav1d.h#L46

Jamaika1: OK. I added DAV1D_API_VERSION_MAJOR

1480c1: This can be marked as resolved

wantehchang:
Hi Jamaika1,

Thank you for writing the pull request. My coworker Michael Bradshaw wrote a patch for internal use at Google yesterday. I just converted his patch to https://github.com/AOMediaCodec/libavif/pull/751.

src/codec_dav1d.c
1480c1: Wrong macro

PR #751, opened 2021-09-04
Update codec_dav1d.c for the new threading model

wantehchang:
See https://code.videolan.org/videolan/dav1d/-/commit/753eef833b:
Merge the 3 threading models into a single one
by Victorien Le Couviour--Tuffet <victorien@videolan.org>

### DIFF
--- a/src/codec_dav1d.c
+++ b/src/codec_dav1d.c
@@ -57,8 +57,13 @@ static avifBool dav1dCodecGetNextImage(struct avifCodec * codec,
 {
     if (codec->internal->dav1dContext == NULL) {
         // Give all available threads to decode a single frame as fast as possible
+#if DAV1D_API_VERSION_MAJOR >= 6
+        codec->internal->dav1dSettings.max_frame_delay = 1;
+        codec->internal->dav1dSettings.n_threads = AVIF_CLAMP(decoder->maxThreads, 1, DAV1D_MAX_THREADS);
+#else
         codec->internal->dav1dSettings.n_frame_threads = 1;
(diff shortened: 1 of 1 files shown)


wantehchang:
@psilokos-twoorioles Hi Victorien, I would appreciate it if you could review this pull request. Thank you!

@mjbshaw Michael: I converted your patch to this pull request.

src/codec_dav1d.c
```diff
@@ -57,8 +57,13 @@ static avifBool dav1dCodecGetNextImage(struct avifCodec * codec,
 {
     if (codec->internal->dav1dContext == NULL) {
         // Give all available threads to decode a single frame as fast as possible
+#if DAV1D_API_VERSION_MAJOR >= 6
```
wantehchang, line 60: Victorien: I have a question for you. If the max_frame_delay and n_threads are renamed before the next dav1d release, will DAV1D_API_VERSION_MAJOR be incremented?

psilokos-twoorioles: Before the next release? hmmm I don't think so, this patch isn't part of a release yet, this will be in for 1.0.0, so I don't think we need to bump major several times between releases... maybe ask on the irc channel to be sure
Tho I don't see why we would rename those as of now.

psilokos-twoorioles:
> @psilokos-twoorioles Hi Victorien, I would appreciate it if you could review this pull request. Thank you!

This looks fine.

src/codec_dav1d.c
mjbshaw: In case it helps, I'm happy to make another patch if Dav1d's API changes again.

PR #752, opened 2021-09-08
Generalized ipco property deduplication

joedrago:
This contains a new helper struct called `avifItemPropertyDedup`, which keeps records of properties that have been written to an `ipco` box so far, and ensures that a duplicate property is never written. 

In my local testing of a fairly common case (single image with alpha), this eliminates a duplicate `ispe` property, which ends up saving 20 bytes. 

[ipco_dedup.zip](https://github.com/AOMediaCodec/libavif/files/7131013/ipco_dedup.zip)

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -181,6 +181,79 @@ static void avifEncoderItemAddMdatFixup(avifEncoderItem * item, const avifRWStre
 }
 
 // ---------------------------------------------------------------------------
+// avifItemPropertyDedup - Provides ipco deduplication
+
+typedef struct avifItemProperty
+{
+    uint8_t index;
+    size_t offset;
+    size_t size;
+} avifItemProperty;
+AVIF_ARRAY_DECLARE(avifItemPropertyArray, avifItemProperty, property);
+
+typedef struct avifItemPropertyDedup
+{
+    avifItemPropertyArray properties;
+    avifRWStream s;    // Temporary stream for each new property, checked against already-written boxes for deduplications
+    avifRWData buffer; // Temporary storage for 's'
+    uint8_t nextIndex; // 1-indexed, incremented every time another unique property is finished
+} avifItemPropertyDedup;
+
+static avifItemPropertyDedup * avifItemPropertyDedupCreate(void)
+{
+    avifItemPropertyDedup * dedup = (avifItemPropertyDedup *)avifAlloc(sizeof(avifItemPropertyDedup));
+    memset(dedup, 0, sizeof(avifItemPropertyDedup));
+    avifArrayCreate(&dedup->properties, sizeof(avifItemProperty), 8);
+    avifRWDataRealloc(&dedup->buffer, 2048); // This will resize automatically (if necessary)
+    return dedup;
+}
+
+static void avifItemPropertyDedupDestroy(avifItemPropertyDedup * dedup)
+{
+    avifArrayDestroy(&dedup->properties);
+    avifRWDataFree(&dedup->buffer);
+    avifFree(dedup);
+}
+
+// Resets the dedup's temporary write stream in preparation for a single item property's worth of writing
+static void avifItemPropertyDedupStart(avifItemPropertyDedup * dedup)
+{
+    avifRWStreamStart(&dedup->s, &dedup->buffer);
+}
+
+// This compares the newly written item property (in the dedup's temporary storage buffer) to
+// already-written properties (whose offsets/sizes in outputStream are recorded in the dedup). If a
+// match is found, the previous item's index is used. If this new property is unique, it is
+// assigned the next available property index, written to the output stream, and its offset/size in
+// the output stream is recorded in the dedup for future comparisons.
+//
+// This function always returns a valid 1-indexed property index for usage in a property association
+// (ipma) box later. If the most recent property was a duplicate of a previous property, the return
+// value will be the index of the original property, otherwise it will be the index of the newly
+// created property.
+static uint8_t avifItemPropertyDedupFinish(avifItemPropertyDedup * dedup, avifRWStream * outputStream)
+{
+    const size_t newPropertySize = avifRWStreamOffset(&dedup->s);
+
+    for (size_t i = 0; i < dedup->properties.count; ++i) {
+        avifItemProperty * property = &dedup->properties.property[i];
+        if ((property->size == newPropertySize) &&
+            !memcmp(&outputStream->raw->data[property->offset], dedup->buffer.data, newPropertySize)) {
+            // We've already written this exact property, reuse it
+            return property->index;
+        }
+    }
+
+    // Write a new property, and remember its location in the output stream for future deduplication
+    avifItemProperty * property = (avifItemProperty *)avifArrayPushPtr(&dedup->properties);
+    property->index = ++dedup->nextIndex; // preincrement so the first new index is 1 (as ipma is 1-indexed)
+    property->size = newPropertySize;
+    property->offset = avifRWStreamOffset(outputStream);
+    avifRWStreamWrite(outputStream, dedup->buffer.data, newPropertySize);
+    return property->index;
+}
+
+// ---------------------------------------------------------------------------
 
 avifEncoder * avifEncoderCreate(void)
 {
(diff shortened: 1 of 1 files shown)


src/write.c
```diff
@@ -180,6 +180,74 @@ static void avifEncoderItemAddMdatFixup(avifEncoderItem * item, const avifRWStre
...
+        if ((property->size == newPropertySize) &&
+            !memcmp(&outputStream->raw->data[property->offset], dedup->buffer.data, newPropertySize)) {
+            // We've already written this exact property, reuse it
+            return property->index;
+        }
+    }
+
+    // Write a new property, and remember its location in the output stream for future deduplication
+    avifItemProperty * property = (avifItemProperty *)avifArrayPushPtr(&dedup->properties);
+    property->index = ++dedup->nextIndex; // ipma is 1-indexed, doing this afterwards is correct
```
wantehchang, line 244: Nit: It is not clear what "doing this afterwards" refers to here. Doing this after what?

Perhaps this is referring to the pre-increment `++dedup->nextIndex` as opposed to the post-increment `dedup->nextIndex++`?

src/write.c
```diff
@@ -180,6 +180,74 @@ static void avifEncoderItemAddMdatFixup(avifEncoderItem * item, const avifRWStre
...
+    size_t offset;
+    size_t size;
+} avifItemProperty;
+AVIF_ARRAY_DECLARE(avifItemPropertyArray, avifItemProperty, property);
+
+typedef struct avifItemPropertyDedup
+{
+    avifItemPropertyArray properties;
+    avifRWStream s;    // Temporary stream for each new property, checked against already-written boxes for deduplications
+    avifRWData buffer; // Temporary storage for 's'
```
wantehchang, line 198: Optional: It seems that we should be able to write each new property to the output stream tentatively, without the need for the temporary stream.

src/write.c
```diff
@@ -213,20 +281,37 @@ void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const char * key,
...
 }
 
-static void avifEncoderWriteColorProperties(avifRWStream * s, const avifImage * imageMetadata, struct ipmaArray * ipma, uint8_t * itemPropertyIndex)
+static void avifEncoderWriteColorProperties(avifRWStream * outputStream,
+                                            const avifImage * imageMetadata,
+                                            struct ipmaArray * ipma,
+                                            avifItemPropertyDedup * dedup)
 {
+    avifRWStream * s = outputStream;
+    if (dedup) {
```
wantehchang, line 290: These `if (dedup)` checks should be omitted because we never call this function with `dedup = NULL`.

If the  `if (dedup)` checks were added because of static analysis warnings, we can deal with that by adding `assert(dedup)` just like the assert(ipma) at line 291.

src/write.c
```diff
@@ -899,81 +996,87 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
...
-                avifEncoderItem * dedupItem = &encoder->data->items.item[dedupIndex];
-                if (item->dimgFromID == dedupItem->dimgFromID) {
-                    // We've already written dedup's items out. Steal their ipma indices and move on!
-                    memcpy(&item->ipma, &dedupItem->ipma, sizeof(struct ipmaArray));
-                    foundPreviousCell = AVIF_TRUE;
-                    break;
-                }
-            }
-            if (foundPreviousCell) {
+    {
```
wantehchang, line 999: It does not seem necessary to put this block of code inside `{ }`. It seems that the purpose is to clearly limit the scope of the variable `dedup`? If so, an alternative is to set `dedup` to `NULL` after the `avifItemPropertyDedupDestroy(dedup)` call.

src/write.c
```diff
@@ -180,6 +180,74 @@ static void avifEncoderItemAddMdatFixup(avifEncoderItem * item, const avifRWStre
...
+        }
+    }
+
+    // Write a new property, and remember its location in the output stream for future deduplication
+    avifItemProperty * property = (avifItemProperty *)avifArrayPushPtr(&dedup->properties);
+    property->index = ++dedup->nextIndex; // ipma is 1-indexed, doing this afterwards is correct
+    property->size = newPropertySize;
+    property->offset = outputStream->offset;
+    avifRWStreamWrite(outputStream, dedup->buffer.data, newPropertySize);
+    return property->index;
```
wantehchang, line 248: We should add a check like this:

    if (property->index & 0x80) {
        //fprintf(stderr, "Some error message);
        abort();
    }

because our `property_index` is 7 bits.

wantehchang (changes requested):
Joe: Thank you for writing this pull request. This is a general solution. The linear search in avifItemPropertyDedupFinish() is brute force. It is fine when the number of properties is small.

I suggest some changes.

src/write.c
```diff
@@ -180,6 +180,74 @@ static void avifEncoderItemAddMdatFixup(avifEncoderItem * item, const avifRWStre
...
+static void avifItemPropertyDedupStart(avifItemPropertyDedup * dedup)
+{
+    avifRWStreamStart(&dedup->s, &dedup->buffer);
+}
+
+// This compares the newly written item property (in the dedup's temporary storage buffer) to
+// already-written properties (whose offsets/sizes in outputStream are recorded in the dedup). If a
+// match is found, the previous item's index is used. If this new property is unique, it is
+// assigned the next available property index, written to the output stream, and its offset/size in
+// the output stream is recorded in the dedup for future comparisons.
```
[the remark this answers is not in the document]
joedrago: Adding now.

src/write.c
joedrago: This is the same wording as the original usage of the pre-increment operator for this index, but I agree that it is awkward. I'll fix it.

src/write.c
joedrago: I agree that this would be a viable approach, but I prefer keeping the buffer separate.

src/write.c
joedrago: `avifEncoderWriteColorProperties()` is usage when writing tracks, which doesn't leverage `ipma`. Therefore, it is legal to call this function with `NULL` for the last two arguments, and this function must honor that. This is why `if (dedup)` is sprinkled throughout the function. We could make `avifItemPropertyDedupStart()` internally `NULL` check and early-out to save on some code paperwork in this function, but I think it would cost the function a bit of readability of intent. I'll add a comment above this function making it clear that it is legal to pass `NULL` for these arguments.

src/write.c
joedrago: I'll remove the block and set `dedup` to `NULL` instead.

src/write.c
joedrago: I'm not against this, but I think this is out-of-scope for this PR as this issue already existed prior to this PR, and if anything, this PR will lessen the chance of this happening (which I already believe is currently impossible with the available set of properties libavif is willing to write). We can address this one in a separate PR.

joedrago:
I've updated the commit with the latest changes from your review.

src/write.c
wantehchang: I thought there was only one call to avifEncoderWriteColorProperties(). Sorry about my confusion.

src/write.c
```diff
@@ -180,6 +180,79 @@ static void avifEncoderItemAddMdatFixup(avifEncoderItem * item, const avifRWStre
...
+//
+// This function always returns a valid 1-indexed property index for usage in a property association
+// (ipma) box later. If the most recent property was a duplicate of a previous property, the return
+// value will be the index of the original property, otherwise it will be the index of the newly
+// created property.
+static uint8_t avifItemPropertyDedupFinish(avifItemPropertyDedup * dedup, avifRWStream * outputStream)
+{
+    const size_t newPropertySize = dedup->s.offset;
```
wantehchang, line 236: Nit: Change `dedup->s.offset` to `avifRWStreamOffset(&dedup->s)`.

We have avoided accessing the `offset` field of `avifRWStream` directly in this file.

src/write.c
```diff
@@ -180,6 +180,79 @@ static void avifEncoderItemAddMdatFixup(avifEncoderItem * item, const avifRWStre
...
+            // We've already written this exact property, reuse it
+            return property->index;
+        }
+    }
+
+    // Write a new property, and remember its location in the output stream for future deduplication
+    avifItemProperty * property = (avifItemProperty *)avifArrayPushPtr(&dedup->properties);
+    property->index = ++dedup->nextIndex; // preincrement so the first new index is 1 (as ipma is 1-indexed)
+    property->size = newPropertySize;
+    property->offset = outputStream->offset;
```
wantehchang, line 251: Nit: Similarly, change `outputStream->offset` to `avifRWStreamOffset(outputStream)`,

wantehchang (approved):
LGTM. I suggest two small changes.

src/write.c
joedrago: Done!

src/write.c
joedrago: Done!

wantehchang (approved):
LGTM. Thanks!

PR #754, opened 2021-09-14
Reland allow alpha to not have ispe

wantehchang:


wantehchang:
I will redo this pull request to show Joe Drago as the author of the first commit.

PR #755, opened 2021-09-14
Reland allow alpha to not have ispe

wantehchang:


src/read.c
```diff
@@ -3101,7 +3101,7 @@ avifResult avifDecoderParse(avifDecoder * decoder)
                 avifDiagnosticsPrintf(data->diag, "Item ID [%u] size is too large [%ux%u]", item->id, item->width, item->height);
                 return AVIF_RESULT_BMFF_PARSE_FAILED;
             }
-        } else {
+        } else if (!item->auxForID || (decoder->strictFlags & AVIF_STRICT_ALPHA_ISPE_REQUIRED)) {
             avifDiagnosticsPrintf(data->diag, "Item ID [%u] is missing a mandatory ispe property", item->id);
```
joedrago, line 3105: All of our current calls to `avifDiagnosticsPrintf()` that can be circumvented by disabling strict flags have the prefix `"[Strict] "` in front of them. Do you think it is worth prefixing this string if the reason we're entering this block is because the strictFlags is set (`item->auxForID` is nonzero)? The prefix exists in other places as a hint to use `--no-strict` if you ever encounter a failure with `[Strict]` in it.

Other than this idea, LGTM.

wantehchang: Done. I also tightened the test for an alpha auxiliary image.

joedrago (approved):
LGTM. I think this is much easier to intuitively understand.

PR #756, opened 2021-09-14
Simplify idat storage for avifMeta structure

joedrago:
The "idats" array inside of an avifMeta structure actually pre-dates the
avifMeta structure itself, and existed to distinguish multiple MetaBox's idat
chunks from one another. This array hasn't been necessary ever since the
avifMeta structure was created and each item gained a back-pointer to its
associated avifMeta. Each avifMeta simply needs to keep track of its own idat
buffer (each meta box must have 0 or 1 idat boxes), and each item is tightly
associated to the meta box it was found in.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1500,12 +1500,22 @@ static avifBool avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw, s
     }
     for (uint32_t i = 0; i < itemCount; ++i) {
         uint32_t itemID;
-        avifBool idatStored = AVIF_FALSE;
         if (version < 2) {
             CHECK(avifROStreamReadU16(&s, &tmp16)); // unsigned int(16) item_ID;
             itemID = tmp16;
         } else {
             CHECK(avifROStreamReadU32(&s, &itemID)); // unsigned int(32) item_ID;
+        }
+
+        avifDecoderItem * item = avifMetaFindItem(meta, itemID);
+        if (!item) {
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -1521,7 +1500,7 @@ static avifBool avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw, s
     }
     for (uint32_t i = 0; i < itemCount; ++i) {
         uint32_t itemID;
-        uint32_t idatID = 0;
+        avifBool idatStored = AVIF_FALSE;
```
wantehchang, line 1503: This local variable does not seem necessary. Why don't we just set `item->idatStored` directly?

wantehchang (approved):
I have a question.

src/read.c
joedrago: We have not yet called `avifMetaFindItem()`, which is what lazily creates the item structure. We could create it before the `constructionMethod` block to avoid the temporary variable, I suppose. I believe my original intent was to avoid calling `avifMetaFindItem()` until I was sure the data is as valid as possible, but we have plenty of potential errors after this call, so it isn't a big deal to move it earlier.

joedrago: I've added a second commit to the PR (so you can see the specific change for this). Do you prefer this?

wantehchang: I didn't know `item` was created later. This is fine then.

wantehchang: No need to move the `avifMetaFindItem()` call unless you prefer that.

PR #759, opened 2021-09-16
fix build on apple m1 arm64

junmer:


ext/aom.cmd
```diff
@@ -14,6 +14,11 @@ cd aom
 mkdir build.libavif
 cd build.libavif
 
-cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DENABLE_DOCS=0 -DENABLE_EXAMPLES=0 -DENABLE_TESTDATA=0 -DENABLE_TESTS=0 -DENABLE_TOOLS=0 ..
+if [[ `uname` == 'Darwin' && `uname -m` == 'arm64' ]]; then
+  cmake -G Ninja -DCONFIG_RUNTIME_CPU_DETECT=0 -DCMAKE_BUILD_TYPE=Release -DENABLE_DOCS=0 -DENABLE_EXAMPLES=0 -DENABLE_TESTDATA=0 -DENABLE_TESTS=0 -DENABLE_TOOLS=0 ..
+else
+  cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DENABLE_DOCS=0 -DENABLE_EXAMPLES=0 -DENABLE_TESTDATA=0 -DENABLE_TESTS=0 -DENABLE_TOOLS=0 ..
+fi
+
```
1480c1, line 22: This would break batch file compatibility, it might be better to mention this in a BUILDING,md file or similar instead or set  `CONFIG_RUNTIME_CPU_DETECT` to false inside the cmakelist instead

joedrago:
There has to be a good reason this runtime CPU detection is on by default. I am very unlikely to accept this patch as-is, and opt for a build note as @1480c1 suggests.

@wantehchang what do we lose if we always had this option enabled for libaom, even on CPUs other than M1?

junmer:
I found the method to fix runtime CPU detection on https://aomedia.googlesource.com/aom/+/refs/heads/master/aom_ports/arm_cpudetect.c#146
```
#elif defined(__APPLE__) /* end __linux__ */
int aom_arm_cpu_caps(void) {
  return HAS_EDSP|HAS_MEDIA|HAS_NEON;
}
#else /* end __APPLE__ */
```

wantehchang (changes requested):
Hi Lingjun,

Thank you very much for the patch. Sorry about the late reply. I was on vacation last Wednesday - Friday.

The best way to fix this build issue is to fix it in libaom. Could you please file a bug report in the [libaom issue tracker](https://bugs.chromium.org/p/aomedia/issues/entry)? Please include the compiler or linker error messages in the bug report.

Then, please convert your suggested change to a libaom changelist by following the [instructions for submitting patches](https://aomedia.googlesource.com/aom/+/master/README.md#submitting-patches).

Once your patch is accepted and checked in, I will create a new libaom patch release (v3.1.3) and we can then change libavif/ext/aom.cmd to get the new version.

wantehchang:
Hi Lingjun,

I didn't know you already uploaded a libaom patch for review. I saw the original version of your comment this morning, which had the link to https://aomedia.googlesource.com/aom/+/refs/heads/master/aom_ports/arm_cpudetect.c#146.

I filed a bug report: https://crbug.com/aomedia/3147.

ext/aom.cmd
```diff
@@ -8,12 +8,11 @@
 : # If you're running this on Windows, be sure you've already run this (from your VC2019 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
 
-git clone -b v3.1.2 --depth 1 https://aomedia.googlesource.com/aom
+git clone -b v3.1.3 --depth 1 https://aomedia.googlesource.com/aom
```
wantehchang, line 11: @junmer Could you please test the v3.1.3-rc1 tag on macOS with the Apple M1 chip? That tag is libaom v3.1.3 release candidate 1. Thank you for your help!

junmer: the v3.1.3-rc1 on MacBook Pro (13-inch, M1, 2020) Apple M1 test passed.   thank you.

wantehchang: Thank you for testing!

There is another file in libavif that contains the libaom version number. I don't remember it off the top of my head, but you can find it by searching for "v3.1.2" in the libavif source tree. Please also update that file in this pull request.

wantehchang:
I took care of this in https://github.com/AOMediaCodec/libavif/pull/769.

PR #760, opened 2021-09-20
Fix oss-fuzz coverage build failure of dav1d

wantehchang:
The cc.symbols_have_underscore_prefix() test in dav1d/meson.build
returns the incorrect value (true) if -fprofile-instr-generate is
specified in the CFLAGS environment variable. This is apparently a bug
in Meson's symbols_have_underscore_prefix() function. See
https://github.com/mesonbuild/meson/issues/5482. Work around this bug by
adding a special case for Linux in dav1d/meson.build.

Part 1 of the fix for https://crbug.com/oss-fuzz/38512. Part 2 of the
fix is to change oss-fuzz/projects/libavif/build.sh to run
"bash dav1d_oss_fuzz.sh" instead of "bash dav1d.cmd".

### DIFF
--- a/ext/dav1d_oss_fuzz.patch
+++ b/ext/dav1d_oss_fuzz.patch
@@ -1,0 +1,19 @@
+diff --git a/meson.build b/meson.build
+index a9cfa04..085653f 100644
+--- a/meson.build
++++ b/meson.build
+@@ -382,7 +382,13 @@ endif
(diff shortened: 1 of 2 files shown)


wantehchang (commented):
@jannau just FYI.

joedrago (approved):

PR #765, opened 2021-09-22
Redesign AVIF_DECODER_SOURCE_AUTO to honor the FileTypeBox's major brand

joedrago:
When using AVIF_DECODER_SOURCE_AUTO, use the major_brand field in the FileTypeBox ('ftyp') in the
beginning of the file as a hint to whether or not to treat this file as an AVIF sequence or an
items-based single image.

In practice (and currently in the field), this won't change pre-existing behavior as the vast
majority of AVIFs in existence right now are single images and signal the major brand 'avif'
anyway, and sequences generated by avifenc have always used the major brand 'avis'.

This design is more in the spirit of ISO-BMFF Section 4.3's explanation of major_brand:

    Files would normally be externally identified (e.g. with a file extension or mime type) that
    identifies the ‘best use’ (major brand), or the brand that the author believes will provide the
    greatest compatibility.

This also would allow for currently unexercised use-cases such as "live photos", where an AVIF can
internally hold an image sequence recorded by a camera, but suggest (via the major brand) to only
display it as a single image, specifying one of the frame samples as an item.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -747,7 +747,9 @@ struct avifDecoderData;
 
 typedef enum avifDecoderSource
 {
-    // If a moov box is present in the .avif(s), use the tracks in it, otherwise decode the primary item.
+    // Honor the major brand signaled in the beginning of the file to pick between an AVIF sequence
+    // ('avis', tracks-based) or a single image ('avif', item-based). If the major brand is neither
+    // of these, prefer the AVIF sequence ('avis', tracks-based), if present.
     AVIF_DECODER_SOURCE_AUTO = 0,
 
     // Use the primary item and the aux (alpha) item in the avif(s).
--- a/src/read.c
+++ b/src/read.c
@@ -721,6 +721,7 @@ typedef struct avifDecoderData
     avifImageGrid colorGrid;
     avifImageGrid alphaGrid;
     avifDecoderSource source;
+    uint8_t majorBrand[4];                     // From the file's ftyp, used by AVIF_DECODER_SOURCE_AUTO
     avifDiagnostics * diag;                    // Shallow copy; owned by avifDecoder
     const avifSampleTable * sourceSampleTable; // NULL unless (source == AVIF_DECODER_SOURCE_TRACKS), owned by an avifTrack
     avifBool cicpSet;                          // True if avifDecoder's image has had its CICP set correctly yet.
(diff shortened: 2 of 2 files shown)


include/avif/avif.h
```diff
@@ -747,7 +747,9 @@ struct avifDecoderData;
 
 typedef enum avifDecoderSource
 {
-    // If a moov box is present in the .avif(s), use the tracks in it, otherwise decode the primary item.
+    // Honor the major brand signaled in the beginning of the file to pick between an AVIF sequence
+    // ('avis', tracks-based) or a single image ('avif', item-based). If the major brand is neither
+    // of these, prefer the AVIF sequence ('avis', tracks-based), if present.
```
wantehchang, line 752: Joe: Chrome's [avif_image_decoder.cc](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/avif/avif_image_decoder.cc) has the following code:

```
    // TODO(wtc): Currently libavif always prioritizes the animation, but that's
    // not correct. It should instead select animation or still image based on
    // the preferred and major brands listed in the file.
    if (animation_option_ != AnimationOption::kUnspecified &&
        avifDecoderSetSource(
            decoder_.get(),
            animation_option_ == AnimationOption::kPreferAnimation
                ? AVIF_DECODER_SOURCE_TRACKS
                : AVIF_DECODER_SOURCE_PRIMARY_ITEM) != AVIF_RESULT_OK) {
      return false;
    }
```

Once Chrome picks up this pull request, I can delete that TODO comment (but leave the code unchanged). Correct?

wantehchang (approved):
LGTM.

include/avif/avif.h
joedrago: Hmm. I'm not familiar with `AnimationOption`, but it seems like this code would still be useful, no?

When `animation_option_` is set to `kUnspecified`, you probably want the AUTO behavior (don't bother calling `avifDecoderSetSource()`, leave it defaulted to AUTO). If it is specified, perhaps you still want to force this?

Either way, the TODO comment can probably go away. AUTO will now leverage the major brand, which I believe satisfies the TODO.

wantehchang: Joe: Thanks for your reply. I agree with you fully.

I am sorry that my question was not clear. I was suggesting that *only* the TODO comment be deleted. I think the code should be left unchanged. What makes the TODO comment confusing is that it is actually a to-do for libavif, not for Chrome.

Issue #767, opened 2021-10-01
The detection code for integer overflows in libyuv's scale functions is off by one


wantehchang:
This bug was discovered by Google's internal fuzzing (b/201759760 and b/201783525).

In https://github.com/AOMediaCodec/libavif/pull/735 we added code to detect integer overflows in libyuv's scale functions. The detection code is off by one:

```
        const int fixedPointDivWidth = (int)(((int64_t)(srcWidth) << 16) / dstWidth);
        const int64_t maxFixedValueWidth = (int64_t)fixedPointDivWidth * (dstWidth - 1);
        if (maxFixedValueWidth > INT_MAX) {
            avifDiagnosticsPrintf(diag, "avifImageScale requested invalid width scale for libyuv [%u -> %u]", srcWidth, dstWidth);
            return AVIF_FALSE;
        }
        const int fixedPointDivHeight = (int)(((int64_t)(srcHeight) << 16) / dstHeight);
        const int64_t maxFixedValueHeight = (int64_t)fixedPointDivHeight * (dstHeight - 1);
        if (maxFixedValueHeight > INT_MAX) {
            avifDiagnosticsPrintf(diag, "avifImageScale requested invalid height scale for libyuv [%u -> %u]", srcHeight, dstHeight);
            return AVIF_FALSE;
        }
```

The multiplications by `(dstWidth - 1)` and `(dstHeight - 1)` should be changed to multiplications by `dstWidth` and `dstHeight`, respectively.

For example, consider the `x += dx` statement in the following code:

```
static void ScaleAddCols2_C(int dst_width,
                            int boxheight,
                            int x,
                            int dx,
                            const uint16_t* src_ptr,
                            uint8_t* dst_ptr) {
  int i;
  int scaletbl[2];
  int minboxwidth = dx >> 16;
  int boxwidth;
  scaletbl[0] = 65536 / (MIN1(minboxwidth) * boxheight);
  scaletbl[1] = 65536 / (MIN1(minboxwidth + 1) * boxheight);
  for (i = 0; i < dst_width; ++i) {
    int ix = x >> 16;
    x += dx;
    boxwidth = MIN1((x >> 16) - ix);
    *dst_ptr++ =
        SumPixels(boxwidth, src_ptr + ix) * scaletbl[boxwidth - minboxwidth] >>
        16;
  }
}
```

Note that the initial value of `x` is 0. So the maximum value of `x` is `dx * dst_width`, not `dx * (dst_width - 1)`.

PR #768, opened 2021-10-01
Improve detection of libyuv integer overflow bugs

wantehchang:
The detection code added in
https://github.com/AOMediaCodec/libavif/pull/735 is off by one.

Also add comments to explain the detection code.

Fix https://github.com/AOMediaCodec/libavif/issues/767.


wantehchang:
@jzern PTAL. Thanks!

joedrago (approved):

jzern (commented):
lgtm

PR #770, opened 2021-10-07
Require srcWidth,srcHeight <= 16384 for ScalePlane

wantehchang:
In most cases a srcWidth or srcHeight of 32767 pixels or less should be
safe for libyuv's ScalePlane() and ScalePlane_12() functions. However,
in some cases a srcHeight less than 32767 still results in integer
overflows. It is not practical to emulate the complicated logic in
libyuv's scale functions and detect integer overflows. Require srcWidth
and srcHeight be less than or equal to 16384 as a simple conservative
check to avoid integer overflows in libyuv.

### DIFF
--- a/src/scale.c
+++ b/src/scale.c
@@ -71,27 +71,13 @@ avifBool avifImageScale(avifImage * image, uint32_t dstWidth, uint32_t dstHeight
     image->height = dstHeight;
 
     if (srcYUVPlanes[0]) {
-        // Detect integer overflows in libyuv's ScalePlane() and ScalePlane_12() functions. For a
-        // representative example, see the ScalePlaneBox() function. It calls ScaleScape() to set
-        // x, y, dx, and dy. It then does "y += dy" in a loop with dst_height iterations. It may
-        // call ScaleAddCols2_C(), which does "x += dx" in a loop with dst_width iterations. We
-        // simulate the libyuv code and detect if the addition would overflow.
-        //
-        // Divide srcWidth by dstWidth and return as 16.16 fixed point result.
-        const int fixedPointDivWidth = (int)(((int64_t)(srcWidth) << 16) / dstWidth);
-        // The maximum value of the sum when adding fixedPointDivWidth in a loop with dstWidth
-        // iterations.
-        const int64_t maxFixedValueWidth = (int64_t)fixedPointDivWidth * dstWidth;
-        if (maxFixedValueWidth > INT_MAX) {
+        // A simple conservative check to avoid integer overflows in libyuv's ScalePlane() and
+        // ScalePlane_12() functions.
+        if (srcWidth > 16384) {
             avifDiagnosticsPrintf(diag, "avifImageScale requested invalid width scale for libyuv [%u -> %u]", srcWidth, dstWidth);
             return AVIF_FALSE;
         }
-        // Divide srcHeight by dstHeight and return as 16.16 fixed point result.
-        const int fixedPointDivHeight = (int)(((int64_t)(srcHeight) << 16) / dstHeight);
-        // The maximum value of the sum when adding fixedPointDivHeight in a loop with dstHeight
-        // iterations.
-        const int64_t maxFixedValueHeight = (int64_t)fixedPointDivHeight * dstHeight;
-        if (maxFixedValueHeight > INT_MAX) {
+        if (srcHeight > 16384) {
             avifDiagnosticsPrintf(diag, "avifImageScale requested invalid height scale for libyuv [%u -> %u]", srcHeight, dstHeight);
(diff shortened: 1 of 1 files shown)


src/scale.c
```diff
@@ -135,15 +121,13 @@ avifBool avifImageScale(avifImage * image, uint32_t dstWidth, uint32_t dstHeight
...
             avifDiagnosticsPrintf(diag, "avifImageScale requested invalid width scale for libyuv [%u -> %u]", srcWidth, dstWidth);
             return AVIF_FALSE;
         }
-        const int fixedPointDivHeight = (int)(((int64_t)(srcHeight) << 16) / dstHeight);
-        const int64_t maxFixedValueHeight = (int64_t)fixedPointDivHeight * dstHeight;
-        if (maxFixedValueHeight > INT_MAX) {
+        if (srcHeight > 16384) {
             avifDiagnosticsPrintf(diag, "avifImageScale requested invalid height scale for libyuv [%u -> %u]", srcHeight, dstHeight);
             return AVIF_FALSE;
         }
```
y-guyon, line 133: I believe these `(srcWidth > 16384)` and `(srcHeight > 16384)` tests could be merged and moved to line 72, to avoid code duplication and unnecessary work. `if (srcYUVPlanes[0] || srcAlphaPlane)` can be prepended to keep the same behavior.

src/scale.c
```diff
@@ -71,27 +71,13 @@ avifBool avifImageScale(avifImage * image, uint32_t dstWidth, uint32_t dstHeight
...
-        //
-        // Divide srcWidth by dstWidth and return as 16.16 fixed point result.
-        const int fixedPointDivWidth = (int)(((int64_t)(srcWidth) << 16) / dstWidth);
-        // The maximum value of the sum when adding fixedPointDivWidth in a loop with dstWidth
-        // iterations.
-        const int64_t maxFixedValueWidth = (int64_t)fixedPointDivWidth * dstWidth;
-        if (maxFixedValueWidth > INT_MAX) {
+        // A simple conservative check to avoid integer overflows in libyuv's ScalePlane() and
+        // ScalePlane_12() functions.
+        if (srcWidth > 16384) {
```
y-guyon, line 76: Should `dstWidth` be checked too? I did not check libyuv's implementation but it seems likely that the indexing uses the same variable types for dst as src.

y-guyon (approved):

src/scale.c
wantehchang: Done. Thanks for the suggestion!

Note: I believe srcYUVPlanes[0] cannot be possibly be NULL, because an AVIF image must have at least a Y plane. So the if (srcYUVPlanes[0]) test could be omitted. But I did not clean this up in this CL.

src/scale.c
wantehchang: libyuv's `ScalePlane()` and `ScalePlane_12()` functions have many code paths. I did not inspect all of them. But in the ones I inspected, it turns out that `dstWidth` and `dstHeight` don't need to be checked. I will explain this using the test in the original code.

We compute `maxFixedValueWidth` as follows:

```
    const int fixedPointDivWidth = (int)(((int64_t)(srcWidth) << 16) / dstWidth);
    const int64_t maxFixedValueWidth = (int64_t)fixedPointDivWidth * dstWidth;
```

And then we require `maxFixedValueWidth <= INT_MAX` to avoid integer overflows in libyuv.

It is easy to verify that if `srcWidth <= 32767`, then the requirement  `maxFixedValueWidth <= INT_MAX` will be satisfied for any value of `dstWidth`.

I hope this explanation makes it plausible that `dstWidth` does not need to be checked.

Note: Ideally I should add comments to justify the proposed checks, but I don't know how to explain the checks concisely...

jzern:
It might be worth filing an issue to have a fallback because this means valid AV1 bitstreams can't be downscaled if I'm understanding the check correctly. Is upscaling allowed by the library?

src/scale.c
y-guyon: Alright, thanks for the detailed explanation. Out of curiosity, did anyone try fixing libyuv?

wantehchang: Yes. Frank Barchard (the author of libavif) is looking into fixing libyuv.

wantehchang:
jzern wrote:
> It might be worth filing an issue to have a fallback because this means valid AV1 bitstreams can't be downscaled if I'm understanding the check correctly. Is upscaling allowed by the library?

The checks I proposed disallow large `srcWidth` and large `srcHeight`. As long as both `srcWidth` and `srcHeight` are <= 16384, `dstWidth` and `dstHeight` can be any values allowed by libavif, so the AV1 bitstreams can be downscaled or upscaled.

PR #778, opened 2021-10-20
Print the item type in two diagnostic messages

wantehchang:
Print the item type in the diagnostic messages for missing mandatory
av1C or pixi property.

Before:
 * [Strict] Item ID 8 is missing mandatory pixi property

After:
 * [Strict] Item ID 8 of type 'grid' is missing mandatory pixi property

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -917,14 +917,17 @@ static avifResult avifDecoderItemValidateAV1(const avifDecoderItem * item, avifD
     const avifProperty * av1CProp = avifPropertyArrayFind(&item->properties, "av1C");
     if (!av1CProp) {
         // An av1C box is mandatory in all valid AVIF configurations. Bail out.
-        avifDiagnosticsPrintf(diag, "Item ID %u is missing mandatory av1C property", item->id);
+        avifDiagnosticsPrintf(diag, "Item ID %u of type '%.4s' is missing mandatory av1C property", item->id, (const char *)item->type);
         return AVIF_RESULT_BMFF_PARSE_FAILED;
     }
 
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -917,14 +917,17 @@ static avifResult avifDecoderItemValidateAV1(const avifDecoderItem * item, avifD
     const avifProperty * av1CProp = avifPropertyArrayFind(&item->properties, "av1C");
     if (!av1CProp) {
         // An av1C box is mandatory in all valid AVIF configurations. Bail out.
-        avifDiagnosticsPrintf(diag, "Item ID %u is missing mandatory av1C property", item->id);
+        avifDiagnosticsPrintf(diag, "Item ID %u of type '%.4s' is missing mandatory av1C property", item->id, (const char *)item->type);
```
wantehchang, line 920: The declaration of `item->type` is:

```
typedef struct avifDecoderItem
{
    ...
    uint8_t type[4];
    ...
} avifDecoderItem;
```

The use of the `%s` format assumes that the `item->type` array consists of four printable characters, which I believe is the case.

There are no compiler warnings if I simply pass `item->type` without the typecast, but I added the typecast anyway.

wantehchang: James: Good point about a corrupted file. I did not consider that. I analyzed the callers of `avifDecoderItemValidateAV1` and found that this function is only called when `item->type` is equal to "av01" or "grid". (This fact is implied by the "ValidateAV1" in the function's name.) So it is safe for this function to print `item->type` using the `%.4s` format.

y-guyon:
LGTM

PR #785, opened 2021-10-28
AVIF parser

y-guyon:
The AVIF parser is a standalone library that can be used to extract
the width, height, bit depth and number of channels from an AVIF
payload. The code is kept short enough to be used by applications that
cannot link to the whole libavif library.


y-guyon:
Thanks. To avoid the confusion you mentioned, a new repository was opened elsewhere:
https://aomedia.googlesource.com/libavifinfo/

The commit is sent for review, if you would like to take a look:
https://aomedia-review.googlesource.com/c/libavifinfo/+/148321
There is a sentence in the README.md describing when to use libavifinfo over libavif.
It is also specified that libavifinfo is a standalone.

PR #786, opened 2021-11-01
Upgrade to rav1e 0.5.0

Jamaika1:


ext/rav1e.cmd
```diff
@@ -10,7 +10,7 @@
 : #
 : # On a successful local build, rav1e/build.libavif/ should contain the necessary headers and static library.
 
-git clone -b 0.4 --depth 1 https://github.com/xiph/rav1e.git
+git clone -b 0.5 --depth 1 https://github.com/xiph/rav1e.git
```
wantehchang, line 13: Please also update the rav1e version in tests/docker/build.sh.

Note that there are CI / build failures. I wonder if they were caused by not updating tests/docker/build.sh.

wantehchang (changes requested):

ext/rav1e.cmd
tongyuantongyu: Try just rerun the CI: rav1e 0.5 requires rust 1.56 for edition 2021 of rust language, but Github is a bit late updating the rust toolchain in its CI environment.

Issue #781, opened 2021-10-21
Build failure (-Werror: implicit declaration of function ‘ScalePlane_12’)


ArchangeGabriel:
When trying to build the latest release on Arch Linux:
```
/build/libavif/src/libavif-0.9.3/src/scale.c: In function ‘avifImageScale’:
/build/libavif/src/libavif-0.9.3/src/scale.c:110:17: error: implicit declaration of function ‘ScalePlane_12’; did you mean ‘ScalePlane_16’? [-Werror=implicit-function-declaration]
  110 |                 ScalePlane_12(srcPlane, srcStride, srcW, srcH, dstPlane, dstStride, dstW, dstH, AVIF_LIBYUV_FILTER_MODE);
      |                 ^~~~~~~~~~~~~
      |                 ScalePlane_16
cc1: all warnings being treated as errors
```
<details>
<summary>Full log</summary>

```
-- The C compiler identification is GNU 11.1.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- libavif: Enabling warnings for GCC
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - yes
-- Found Threads: TRUE  
-- Checking for module 'libyuv'
--   Package 'libyuv', required by 'virtual:world', not found
-- libavif: libyuv (1768) found; libyuv-based fast paths enabled.
-- libavif: Codec enabled: dav1d (decode)
-- Checking for module 'dav1d'
--   Found dav1d, version 0.9.2
-- Found dav1d: /usr/lib/libdav1d.so (found version "0.9.2") 
-- libavif: Codec enabled: rav1e (encode)
-- Checking for module 'rav1e'
--   Found rav1e, version 0.4.1
-- Found rav1e: /usr/lib/librav1e.so (found version "0.4.1") 
-- libavif: Codec enabled: svt (encode)
-- Checking for module 'SvtAv1Enc'
--   Found SvtAv1Enc, version 0.8.7
-- Found svt: /usr/lib/libSvtAv1Enc.so (found version "0.8.7") 
-- libavif: Codec enabled: aom (encode/decode)
[...]

PR #788, opened 2021-11-04
Update required libyuv version

ArchangeGabriel:
Fixes GH-781 (the function was introduced in libavif https://github.com/AOMediaCodec/libavif/commit/46104d6db42ec2a94a7da6fc0693f24fa90a1b98).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,12 +226,16 @@ endif()
 find_package(libyuv QUIET) # not required
 if(libyuv_FOUND)
     # libyuv 1755 exposed all of the I*Matrix() functions, which libavif relies on.
+    # libyuv 1774 exposed ScalePlane_12 function, which libavif can use for some additional optimizations.
     if(NOT LIBYUV_VERSION)
         message(STATUS "libavif: libyuv found, but version unknown; libyuv-based fast paths disabled.")
     elseif(LIBYUV_VERSION LESS 1755)
         message(STATUS "libavif: libyuv (${LIBYUV_VERSION}) found, but is too old; libyuv-based fast paths disabled.")
     else()
         message(STATUS "libavif: libyuv (${LIBYUV_VERSION}) found; libyuv-based fast paths enabled.")
+        if(LIBYUV_VERSION LESS 1774)
+            message(STATUS "libavif: some libyuv optimizations require at least version 1774 to work.")
+        endif()
         set(AVIF_PLATFORM_DEFINITIONS ${AVIF_PLATFORM_DEFINITIONS} -DAVIF_LIBYUV_ENABLED=1)
         set(AVIF_PLATFORM_INCLUDES ${AVIF_PLATFORM_INCLUDES} ${LIBYUV_INCLUDE_DIR})
         set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ${LIBYUV_LIBRARY})
--- a/src/scale.c
+++ b/src/scale.c
@@ -107,7 +107,11 @@ avifBool avifImageScale(avifImage * image, uint32_t dstWidth, uint32_t dstHeight
                 const uint32_t srcStride = srcYUVRowBytes[i] / 2;
                 uint16_t * const dstPlane = (uint16_t *)image->yuvPlanes[i];
                 const uint32_t dstStride = image->yuvRowBytes[i] / 2;
+#if LIBYUV_VERSION >= 1774
                 ScalePlane_12(srcPlane, srcStride, srcW, srcH, dstPlane, dstStride, dstW, dstH, AVIF_LIBYUV_FILTER_MODE);
+#else
+                ScalePlane_16(srcPlane, srcStride, srcW, srcH, dstPlane, dstStride, dstW, dstH, AVIF_LIBYUV_FILTER_MODE);
+#endif
             } else {
                 uint8_t * const srcPlane = srcYUVPlanes[i];
                 const uint32_t srcStride = srcYUVRowBytes[i];
@@ -130,7 +134,11 @@ avifBool avifImageScale(avifImage * image, uint32_t dstWidth, uint32_t dstHeight
             const uint32_t srcStride = srcAlphaRowBytes / 2;
             uint16_t * const dstPlane = (uint16_t *)image->alphaPlane;
             const uint32_t dstStride = image->alphaRowBytes / 2;
+#if LIBYUV_VERSION >= 1774
             ScalePlane_12(srcPlane, srcStride, srcWidth, srcHeight, dstPlane, dstStride, dstWidth, dstHeight, AVIF_LIBYUV_FILTER_MODE);
+#else
+            ScalePlane_16(srcPlane, srcStride, srcWidth, srcHeight, dstPlane, dstStride, dstWidth, dstHeight, AVIF_LIBYUV_FILTER_MODE);
+#endif
         } else {
             uint8_t * const srcPlane = srcAlphaPlane;
             const uint32_t srcStride = srcAlphaRowBytes;


CMakeLists.txt
```diff
@@ -225,10 +225,10 @@ endif()
 
 find_package(libyuv QUIET) # not required
 if(libyuv_FOUND)
-    # libyuv 1755 exposed all of the I*Matrix() functions, which libavif relies on.
+    # libyuv 1778 exposed ScalePlane_12 function, which libavif relies on.
```
wantehchang, line 228: libyuv 1778 exposed I420Scale_12 and I444Scale_12, not ScalePlane12:
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2706966

ScalePlane_12 was added in libyuv 1774:
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2658097

It is certainly fine to require a newer version such as libyuv 1778. I am just wondering if you know something wrong with the ScalePlane_12() function in libyuv 1774, such as it was an incomplete or buggy implementation.

wantehchang (commented):
Hi Bruno,

Thank you very much for the bug report and the pull request. I have a question.

CMakeLists.txt
tongyuantongyu: I recommended to use 1789:
There's some issues in the neon64 assembly when building with gcc:
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2922147

ArchangeGabriel: @wantehchang You are right, actually I only did quick commit msg search, and Scale 12 matched on https://chromium.googlesource.com/libyuv/libyuv/+/08815a29766a78398a8e2b9ed095280e9d0a73c2^!/. Grepping the commit diff for `ScalePlane_12` was further positive, and this commit allowed libavif to compile, so I thought this was good. But indeed looking more deeply at the diff I now see that `ScalePlane_12` is called but not defined in this commit (which is as you said about `I420/I444Scale_12`), you are evidently right that 1774 should be enough. But I can update the MR to 1789 instead to avoid the issue mentionned by @tongyuantongyu.

ArchangeGabriel: Apparently they are other neon fixes in 1796: https://chromium-review.googlesource.com/c/libyuv/libyuv/+/3216739

CMakeLists.txt
```diff
@@ -225,10 +225,10 @@ endif()
...
     if(NOT LIBYUV_VERSION)
         message(STATUS "libavif: libyuv found, but version unknown; libyuv-based fast paths disabled.")
-    elseif(LIBYUV_VERSION LESS 1755)
+    elseif(LIBYUV_VERSION LESS 1778)
```
wantehchang, line 231: Hi Bruno,

I just realized that this approach has a serious downside: it will cause libavif to not use libyuv for RGB-to-YUV and YUV-to-RGB conversions if libyuv is not version 1778 or newer.

libavif uses libyuv for two purposes: RGB<->YUV conversion and scaling. Between the two, RGB<->YUV conversion is much more important.

I suggest we use a different approach. We check for libyuv version at the place where we call ScalePlane_12(). Please search for "LIBYUV_VERSION" in libavif/src/reformat_libyuv.c for examples.

I think something like this patch should work. In this patch I use libyuv version 1774 for simplicity and leave the responsibility of using a libyuv version that is good enough to the user.

```
diff --git a/src/scale.c b/src/scale.c
index 6ca651b..89bf2fe 100644
--- a/src/scale.c
+++ b/src/scale.c
@@ -71,6 +71,13 @@ avifBool avifImageScale(avifImage * image, uint32_t dstWidth, uint32_t dstHeight
     image->height = dstHeight;
 
     if (srcYUVPlanes[0] || srcAlphaPlane) {
+#if LIBYUV_VERSION < 1774
+        if (image->depth > 8) {
+            avifDiagnosticsPrintf(diag, "ScalePlane_12() is not available in libyuv version %d", LIBYUV_VERSION);
+        }
+#endif
+
         // A simple conservative check to avoid integer overflows in libyuv's ScalePlane() and
         // ScalePlane_12() functions.
         if (srcWidth > 16384) {
@@ -103,11 +110,13 @@ avifBool avifImageScale(avifImage * image, uint32_t dstWidth, uint32_t dstHeight
             const uint32_t dstW = (i == AVIF_CHAN_Y) ? dstWidth : dstUVWidth;
             const uint32_t dstH = (i == AVIF_CHAN_Y) ? dstHeight : dstUVHeight;
             if (image->depth > 8) {
+#if LIBYUV_VERSION >= 1774
                 uint16_t * const srcPlane = (uint16_t *)srcYUVPlanes[i];
                 const uint32_t srcStride = srcYUVRowBytes[i] / 2;
                 uint16_t * const dstPlane = (uint16_t *)image->yuvPlanes[i];
                 const uint32_t dstStride = image->yuvRowBytes[i] / 2;
                 ScalePlane_12(srcPlane, srcStride, srcW, srcH, dstPlane, dstStride, dstW, dstH, AVIF_LIBYUV_FILTER_MODE);
+#endif
             } else {
                 uint8_t * const srcPlane = srcYUVPlanes[i];
                 const uint32_t srcStride = srcYUVRowBytes[i];
@@ -126,11 +135,13 @@ avifBool avifImageScale(avifImage * image, uint32_t dstWidth, uint32_t dstHeight
         avifImageAllocatePlanes(image, AVIF_PLANES_A);
 
         if (image->depth > 8) {
+#if LIBYUV_VERSION >= 1774
             uint16_t * const srcPlane = (uint16_t *)srcAlphaPlane;
             const uint32_t srcStride = srcAlphaRowBytes / 2;
             uint16_t * const dstPlane = (uint16_t *)image->alphaPlane;
             const uint32_t dstStride = image->alphaRowBytes / 2;
             ScalePlane_12(srcPlane, srcStride, srcWidth, srcHeight, dstPlane, dstStride, dstWidth, dstHeight, AVIF_LIBYUV_FILTER_MODE);
+#endif
         } else {
             uint8_t * const srcPlane = srcAlphaPlane;
             const uint32_t srcStride = srcAlphaRowBytes;
```

wantehchang (changes requested):

CMakeLists.txt
tongyuantongyu: Alternatively, we can use `ScalePlane_16` instead if `LIBYUV_VERSION < 1774`.
`ScalePlane_12` has special optimization for 2x scale up using linear/bilinear filter (when scaling up, box and bilinear are the same), and for the rest cases, it simply calls `ScalePlane_16`, which handles 2x scale up just fine (slower though).

ArchangeGabriel: Seems a good idea. However I’m not exactly sure whether the arguments for `ScalePlane_16` are the same as for the `_12` versions, and especially I don’t know (the) code enough to avoid issues such as https://github.com/AOMediaCodec/libavif/issues/715.

tongyuantongyu: The arguments are the same: https://chromium.googlesource.com/libyuv/libyuv/+/a04e4f87fbf40405707b1d0ae9fcba8fc93f7856/source/scale.cc#2073, so it should be fine to use `ScalePlane_16` when `ScalePlane_12` is unavailable.

ArchangeGabriel:
I’ve changed the commit to conditionally use ScalePlane_12 as suggested, and modified CMake to print additional information instead when in the partially supported version range. I can change the message (e.g. specify what can of optimizations, wording. etc.), remove it (but I think people compiling should be told about it so they can leverage it) or merge it with the enabling line (so making to enable message depending on version).

wantehchang (approved):
LGTM. Thanks.

Bruno: I am very sorry for the late review. I was out of the office.

ArchangeGabriel:
It’s fine, no need to apologize. ;)

PR #791, opened 2021-11-06
Upgrade to nasm-2.16rc0

Jamaika1:


wantehchang (commented):
Thank you for the pull request. As a policy we do not use release candidate tags. We can consider making an exception if the final release is imminent, but nasm-2.16rc0-20201104.tar.gz is not really a release candidate; it is a snapshot.

PR #803, opened 2021-12-15
findrav1e: add LDFLAGS to LIBRARIES

1480c1:
matches findaom's behavior

fixes static linking of librav1e.a on windows since it requires bcrypt

As requested at https://github.com/AOMediaCodec/libavif/pull/802#issuecomment-995200877

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -333,7 +333,7 @@ if(AVIF_CODEC_RAV1E)
 
     # Unfortunately, rav1e requires a few more libraries
     if(WIN32)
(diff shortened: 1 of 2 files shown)


wantehchang (approved):
LGTM. Thank you!

wantehchang:
@1480c1 Christopher: Do you know why appveyor still fails with the unresolved reference to BCryptGenRandom?

1480c1:
That's because msvc doesn't use pkg-config, I can manually add bcrypt to the list of extra libs in cmake to fix that

wantehchang (approved):

PR #804, opened 2021-12-17
Support local android builds for libgav1

vigneshvg:
When building libavif for android using the NDK, the underlying
AV1 codec can be built for multiple architectures with specific
optimizations. It is useful to build them as it is sometimes
necessary to generate a fat APK (an APK that contains the library
for multiple architectures).

Upate the CMake file to include an architecture sub-directory while
finding libgav1. Also add a shell script that will clone and build
libgav1 using the andoroid NDK for the architectures supported on
Android by default.


joedrago:
I haven't actually read the change here yet, but just as a note, there is no requirement to build any of the AV1 libraries locally to this library. ideally you would just build libgav1 for the Android system as a shared library using whatever optimizations you want, and then simply let libavif use the system shared library.

If @wantehchang thinks this is a good change, I will defer to him, but the "LOCAL" features are intended to be used in embedded/static builds, which I'm not sure Android qualifies. This feels like a little bit of a misfeature (at first glance).

vigneshvg:
The idea here is to let apps bundle libavif if they choose to. For example, an app could write a jni wrapper around libavif and choose to build the necessary native libraries using CMake. This example supports that use case.

There is also a plan to add a JNI wrapper for libavif. Whether or not the JNI wrapper belongs in this repository is another question. But this PR basically lets app build libavif (along with the underlying libraries) from source for multiple architectures if the apps choose to do that. :)

ext/libgav1_android.sh
```diff
@@ -0,0 +1,41 @@
...
+  for abi in ${ABI_LIST}; do
+    mkdir "${abi}"
+    cd "${abi}"
+    cmake ../.. \
+      -DCMAKE_TOOLCHAIN_FILE=../../cmake/toolchains/android.cmake \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DLIBGAV1_ANDROID_NDK_PATH=${1} \
+      -DLIBGAV1_THREADPOOL_USE_STD_MUTEX=1 \
+      -DANDROID_ABI=${abi}
+    make -j
```
wantehchang, line 36: Nit: Is it possible to use ninja instead of make?

ext/libgav1_android.sh
```diff
@@ -0,0 +1,41 @@
...
+# Android NDK: https://developer.android.com/ndk/downloads
+#
+# The git SHA below is known to work, and will occasionally be updated. Feel
+# free to use a more recent commit.
+#
+# cmake and make must be in your PATH.
+
+if [ $# -ne 1 ]; then
+  echo "Usage: ${0} <path_to_android_ndk>"
+else
```
wantehchang, line 18: Nit: After the 'echo' statement, should the script just do "exit 1"? Then we won't need the "else".

android_jni/.gitignore
```diff
@@ -0,0 +1,15 @@
+*.iml
```
wantehchang, line 1: Please remove the second commit (Add Android JNI Wrapper) from this pull request. It should become a separate pull request.

I am not very familiar with GitHub's pull requests, so I don't know what is the best way to do this.

wantehchang (changes requested):

android_jni/.gitignore
vigneshvg: Adding a commit to my fork seems to update this pull request as well. I was not expecting that. I have reverted that change now. This pull request only contains the one commit now.

ext/libgav1_android.sh
vigneshvg: Done.

ext/libgav1_android.sh
vigneshvg: Done.

CMakeLists.txt
```diff
@@ -286,7 +286,13 @@ if(AVIF_CODEC_LIBGAV1)
...
 
     if(AVIF_LOCAL_LIBGAV1)
-        set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/libgav1/build/libgav1${CMAKE_STATIC_LIBRARY_SUFFIX}")
+        # If ${ANDROID_ABI} is set, look for the library under that subdirectory.
+        if (DEFINED ANDROID_ABI)
+            set(AVIF_ANDROID_PREFIX "${ANDROID_ABI}/")
+        else()
+            set(AVIF_ANDROID_PREFIX "")
+        endif()
+        set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/libgav1/build/${AVIF_ANDROID_PREFIX}libgav1${CMAKE_STATIC_LIBRARY_SUFFIX}")
```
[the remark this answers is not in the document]
vigneshvg: Done.

vigneshvg:
I believe i have addressed your comments Wan-Teh.

My github skills are quite limited. So hopefully i didn't screw anything up. Please take another look and let me know if you have any more comments.

CMakeLists.txt
```diff
@@ -293,7 +293,12 @@ if(AVIF_CODEC_LIBGAV1)
     )
 
     if(AVIF_LOCAL_LIBGAV1)
-        set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/libgav1/build/libgav1${CMAKE_STATIC_LIBRARY_SUFFIX}")
+        set (AVIF_LIBGAV1_BUILD_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/libgav1/build")
+        # If ${ANDROID_ABI} is set, look for the library under that subdirectory.
```
wantehchang, line 297: Is it true that we must pass -DANDROID_ABI=xxx to cmake when we build libavif with the Android NDK? This webpage says ANDROID_ABI is a required argument:

https://developer.android.com/ndk/guides/cmake#android_abi

wantehchang (changes requested):
Vignesh: Thank you for updating the pull request. I have some more comments.

ext/libgav1_android.sh
```diff
@@ -0,0 +1,42 @@
+# If you want to use a local build of libgav1 to be used with the android NDK,
+# you must clone the libgav1 repo in this directory first, then enable CMake's
+# AVIF_CODEC_LIBGAV1 and AVIF_LOCAL_LIBGAV1 options.
```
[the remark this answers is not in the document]
vigneshvg: Removed.

ext/libgav1_android.sh
```diff
@@ -0,0 +1,42 @@
...
+# This script will build libgav1 for the default ABI targets supported by
+# android. You must pass the path to the android NDK as a parameter to this
+# script.
+#
+# Android NDK: https://developer.android.com/ndk/downloads
+#
+# The git SHA below is known to work, and will occasionally be updated. Feel
+# free to use a more recent commit.
+#
+# cmake and make must be in your PATH.
```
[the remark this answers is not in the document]
vigneshvg: Removed.

ext/libgav1_android.sh
```diff
@@ -0,0 +1,42 @@
...
+# you must clone the libgav1 repo in this directory first, then enable CMake's
+# AVIF_CODEC_LIBGAV1 and AVIF_LOCAL_LIBGAV1 options.
+#
+# This script will build libgav1 for the default ABI targets supported by
+# android. You must pass the path to the android NDK as a parameter to this
+# script.
+#
+# Android NDK: https://developer.android.com/ndk/downloads
+#
+# The git SHA below is known to work, and will occasionally be updated. Feel
```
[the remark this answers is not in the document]
vigneshvg: Changed SHA to tag. It refers to the tag of libgav1 and abseil that's being checked out.

CMakeLists.txt
vigneshvg: Yes that's correct. In the subsequent pull request, where we actually make use of this in libavif's android JNI, that argument will be set by the android gradle build system automatically.

ext/libgav1_android.sh
```diff
@@ -0,0 +1,36 @@
+# This script will build libgav1 for the default ABI targets supported by
```
jzern, line 1: This is consistent with the other ext/.sh files, but I think it would be better to have a #! on an executable file unless this is meant to be run from another script, then maybe it doesn't need to be directly executable.

ext/libgav1_android.sh
```diff
@@ -0,0 +1,36 @@
...
+# Android NDK: https://developer.android.com/ndk/downloads
+#
+# The git tag below is known to work, and will occasionally be updated. Feel
+# free to use a more recent commit.
+
+if [ $# -ne 1 ]; then
+  echo "Usage: ${0} <path_to_android_ndk>"
+  exit
+fi
+git clone -b v0.17.0 --depth 1 https://chromium.googlesource.com/codecs/libgav1
```
jzern, line 14: you might want to add the same comment here about updating libgav1.cmd

ext/libgav1_android.sh
vigneshvg: Done.

ext/libgav1_android.sh
vigneshvg: Done

ext/libgav1_android.sh
```diff
@@ -0,0 +1,39 @@
...
+# script.
+#
+# Android NDK: https://developer.android.com/ndk/downloads
+#
+# The git tag below is known to work, and will occasionally be updated. Feel
+# free to use a more recent commit.
+
+if [ $# -ne 1 ]; then
+  echo "Usage: ${0} <path_to_android_ndk>"
+  exit
```
wantehchang, line 14: Nit: exit 1

because this is an error.

wantehchang (approved):
LGTM. I would like to test this pull request on my Linux computer first.

ext/libgav1_android.sh
vigneshvg: Done.

vigneshvg:
> LGTM. I would like to test this pull request on my Linux computer first.

Thanks Wan-Teh.

I have updated the exit code to exit 1.

wantehchang (approved):
LGTM. I tested this pull request with Android NDK r22b. Android NDK r23b (the latest LTS version) requires a change to libgav1/cmake/toolchains/android.cmake.

After running ext/libgav1_android.sh, I created the build/arm64-v8a directory in libavif and then ran the following cmake command:

`cmake -G Ninja ../.. -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_LIBGAV1=ON -DAVIF_LOCAL_LIBGAV1=ON -DCMAKE_TOOLCHAIN_FILE=~/android-ndk-r22b/build/cmake/android.toolchain.cmake -DANDROID_ABI=arm64-v8a`

Then I ran the ninja command to build libavif.a.

PR #805, opened 2021-12-17
Make external library subdir configurable

tongyuantongyu:
I'm mainly developing on Windows and testing on multiple toolchains (MinGW (MINGW64), MinGW (CLANG64) and MSVC). Libraries built using one toolchain can't be used in another, but they all expect external libraries in `ext/`, which means only one toolchain is buildable at one time. This is quite inconvenient, so I'm making it configurable so that they can coexist.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,11 +39,26 @@ option(AVIF_CODEC_SVT "Use the SVT-AV1 codec for encoding" OFF)
 option(AVIF_CODEC_AOM_DECODE "if AVIF_CODEC_AOM is on, use/offer libaom's decoder" ON)
 option(AVIF_CODEC_AOM_ENCODE "if AVIF_CODEC_AOM is on, use/offer libaom's encoder" ON)
 
-option(AVIF_LOCAL_AOM "Build the AOM codec by providing your own copy of the repo in ext/aom (see Local Builds in README)" OFF)
-option(AVIF_LOCAL_DAV1D "Build the dav1d codec by providing your own copy of the repo in ext/dav1d (see Local Builds in README)" OFF)
-option(AVIF_LOCAL_LIBGAV1 "Build the libgav1 codec by providing your own copy of the repo in ext/libgav1 (see Local Builds in README)" OFF)
-option(AVIF_LOCAL_RAV1E "Build the rav1e codec by providing your own copy of the repo in ext/rav1e (see Local Builds in README)" OFF)
-option(AVIF_LOCAL_SVT "Build the SVT-AV1 codec by providing your own copy of the repo in ext/SVT-AV1 (see Local Builds in README)" OFF)
+option(AVIF_LOCAL_AOM "Build the AOM codec by providing your own copy of the repo (see Local Builds in README)" OFF)
+option(AVIF_LOCAL_DAV1D "Build the dav1d codec by providing your own copy of the repo (see Local Builds in README)" OFF)
+option(AVIF_LOCAL_LIBGAV1 "Build the libgav1 codec by providing your own copy of the repo (see Local Builds in README)" OFF)
+option(AVIF_LOCAL_RAV1E "Build the rav1e codec by providing your own copy of the repo (see Local Builds in README)" OFF)
+option(AVIF_LOCAL_SVT "Build the SVT-AV1 codec by providing your own copy of the repo (see Local Builds in README)" OFF)
+
+set(AVIF_LOCAL_AOM_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/aom" CACHE PATH "Source directory of your own copy of AOM codec")
+set(AVIF_LOCAL_AOM_LIBRARY_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/aom/build.libavif" CACHE PATH "Library directory of your own copy of AOM codec")
+set(AVIF_LOCAL_DAV1D_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/usr/include" CACHE PATH "Source directory of your own copy of dav1d codec")
+set(AVIF_LOCAL_DAV1D_LIBRARY_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/usr/lib" CACHE PATH "Library directory of your own copy of dav1d codec")
+set(AVIF_LOCAL_LIBGAV1_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/libgav1/src" CACHE PATH "Source directory of your own copy of libgav1 codec")
+set(AVIF_LOCAL_LIBGAV1_LIBRARY_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/libgav1/build" CACHE PATH "Library directory of your own copy of libgav1 codec")
+set(AVIF_LOCAL_RAV1E_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/rav1e/build.libavif/usr/include/rav1e" CACHE PATH "Source directory of your own copy of rav1e codec")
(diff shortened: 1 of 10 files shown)


wantehchang:
@tongyuantongyu Yuan Tong: Thank you for the pull request. I agree this problem is annoying and your solution works. But a drawback of your solution is that the source trees of the external dependencies are not shared. It seems better if we could instead make the build directory configurable. Right now the build directory is named "build", except for libaom. We can look into generalizing the approach in ext/libgav1_android.sh, allowing an optional subdirectory to be used under the build directory for a particular toolchain or target architecture.

tongyuantongyu:
@wantehchang : Thanks for your suggestion, and I agree with you. I updated my PR to accept custom `AVIF_LOCAL_*_{INCLUDE,LIBRARY}_DIR` options that point to each codec's include and library path.

CMakeLists.txt
```diff
@@ -39,11 +39,26 @@ option(AVIF_CODEC_SVT "Use the SVT-AV1 codec for encoding" OFF)
...
-option(AVIF_LOCAL_LIBGAV1 "Build the libgav1 codec by providing your own copy of the repo in ext/libgav1 (see Local Builds in README)" OFF)
-option(AVIF_LOCAL_RAV1E "Build the rav1e codec by providing your own copy of the repo in ext/rav1e (see Local Builds in README)" OFF)
-option(AVIF_LOCAL_SVT "Build the SVT-AV1 codec by providing your own copy of the repo in ext/SVT-AV1 (see Local Builds in README)" OFF)
+option(AVIF_LOCAL_AOM "Build the AOM codec by providing your own copy of the repo (see Local Builds in README)" OFF)
+option(AVIF_LOCAL_DAV1D "Build the dav1d codec by providing your own copy of the repo (see Local Builds in README)" OFF)
+option(AVIF_LOCAL_LIBGAV1 "Build the libgav1 codec by providing your own copy of the repo (see Local Builds in README)" OFF)
+option(AVIF_LOCAL_RAV1E "Build the rav1e codec by providing your own copy of the repo (see Local Builds in README)" OFF)
+option(AVIF_LOCAL_SVT "Build the SVT-AV1 codec by providing your own copy of the repo (see Local Builds in README)" OFF)
+
+set(AVIF_LOCAL_AOM_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/aom" CACHE PATH "Source directory of your own copy of AOM codec")
```
tongyuantongyu, line 48: These paths look quite random. If we instead 'install' these packages, they can be more consistent and maybe we can just set a prefix, and then search under `include` and `lib` dirs. But this means during development you'll need to 'install' instead of 'build' each time you make some changes to some codec.

CMakeLists.txt
```diff
@@ -230,8 +231,14 @@ if(UNIX)
     set(AVIF_PLATFORM_LIBRARIES m Threads::Threads)
 endif()
 
-find_package(libyuv QUIET) # not required
+find_package(libyuv) # not required
```
tongyuantongyu, line 234: The line
```-- Found libyuv: path/to/libyuv.a```
Is useful to indicate which library file is being linked against, so I removed the `QUIET`. Inside `Findlibyuv.cmake` the call `pkg_check_modules(_LIBYUV libyuv QUIET)` doesn't provide extra info, so I silented that one.

File: CMakeLists.txt [hunk 3]
```diff
@@ -252,167 +259,113 @@ set(AVIF_CODEC_INCLUDES)
 set(AVIF_CODEC_LIBRARIES)
 
 if(AVIF_CODEC_DAV1D)
-    message(STATUS "libavif: Codec enabled: dav1d (decode)")
```
tongyuantongyu, line 255: I moved this message to the bottom of this block, because it sounds like we've done all the checks and everything seems working.

cmake/Modules/Findaom.cmake
```diff
@@ -17,19 +17,30 @@
...
 
-find_package(PkgConfig QUIET)
-if (PKG_CONFIG_FOUND)
-    pkg_check_modules(_AOM aom)
-endif (PKG_CONFIG_FOUND)
+if (AVIF_LOCAL_AOM)
+    find_path(AOM_INCLUDE_DIR
+              NAMES aom/aom.h
+              PATHS ${AVIF_LOCAL_AOM_INCLUDE_DIR}
+              NO_DEFAULT_PATH)
```
tongyuantongyu, line 24: Is there anyway to pass `NO_DEFAULT_PATH` option to `find_path` only when `AVIF_LOCAL_AOM` is true?

cmake/Modules/Findlibyuv.cmake
```diff
@@ -17,17 +17,30 @@
 #=============================================================================
 #
 
-find_package(PkgConfig QUIET)
```
tongyuantongyu, line 20: `Findlibyuv.cmake` was a bit weird: it calls pkg-config to find libyuv installed in system even when `AVIF_LOCAL_LIBYUV` is set. I made it more similar to other `Find*.cmake`.

ext/dav1d.cmd
```diff
@@ -18,6 +18,8 @@ cd build
 : # macOS might require: -Dc_args=-fno-stack-check
 : # Build with asan: -Db_sanitize=address
 : # Build with ubsan: -Db_sanitize=undefined
-meson --default-library=static --buildtype release ..
-ninja
+: # Tell MSYS2 don't mess with local prefix, no-op for other platforms
+: ; export MSYS2_ARG_CONV_EXCL="--prefix="
+meson --default-library=static --buildtype release --prefix=/usr ..
```
tongyuantongyu, line 23: I'm changing dav1d to 'install', because otherwise header files are scattered everywhere and hard to specify using one include dir.

src/codec_rav1e.c
```diff
@@ -3,7 +3,7 @@
 
 #include "avif/internal.h"
 
-#include "rav1e.h"
+#include "rav1e/rav1e.h"
```
tongyuantongyu, line 6: rav1e header is installed to `<prefix>/include/rav1e/rav1e.h`, so this seems more standard.

wantehchang:
In CMakeLists.txt, you wrote:

> The line
> ```-- Found libyuv: path/to/libyuv.a```
> Is useful to indicate which library file is being linked against, so I removed the `QUIET`. Inside `Findlibyuv.cmake` the call `pkg_check_modules(_LIBYUV libyuv QUIET)` doesn't provide extra info, so I silented that one.

Could you move these two changes to a new PR so that they can be reviewed separately?

wantehchang (commented):
Hi Yuan Tong,

The new PR is going in a different direction and seems much more complex. I suggest moving some independent changes to new PRs to help simplify this PR.

But I am wondering whether we need this complexity. The approach I suggested is to create a subdirectory under the 'build' directory if an environment variable is specified. Did you consider this approach? I guess shell scripts and Windows batch files use different syntaxes to get the value of an environment variable, so it will be hard to make the ext/*.cmd files do that in a cross-platform way?

CMakeLists.txt
tongyuantongyu: My bad, I forgot to explain the 'install' I said.
I'm quoting 'install' to mean we are not going to really copy the files into system directory, but to use install ability to copy necessary headers and library files into a standard path structure under build directory. This way they can be found by simply provide a prefix path, and this is exactly what we've done for rav1e.

src/codec_rav1e.c
tongyuantongyu: I re-examined the situation.
rav1e's pkg-config file set additional include path to `${prefix}/include/rav1e`, instead of the more common `${prefix}/include`, so despite placed at `${prefix}/include/rav1e/rav1e.h`, it is included by using `#include "rav1e.h"`. `#include "rav1e/rav1e.h"` works probably just because other dependency added include path `${prefix}/include`. I'm reverting these changes.

tongyuantongyu:
@wantehchang: I see most time we only want one copy of source, but then I found sometimes I do need multiple source trees, like one at main branch to experiment stuffs, another at some release tag to verify if code builds. Using one variable can't provide such flexibility.

If you feel it's too much complexity for little benefit, I'll keep these changes locally.

CMakeLists.txt
tongyuantongyu: Moved into #816.

vrabaud:
Now that `ext/` is using FetchContent, you should be able to use one toolchain for `libavif` and its local deps.

PR #806, opened 2022-01-07
Split monolithic src/read.c into .h and .c files

y-guyon:
I find 4000-line-long read.c difficult to navigate. I would prefer to split it into several .h/.c for easier struct definition side-by-side lookup and per-theme function grouping.
Unfortunately git/Github is not file-splitting-friendly, but this change only adds header guards into the .h files. Everything else is branched from read.c, in order.


wantehchang:
Yannis: Thank you for writing this pull request. I am torn about this PR. On the one hand, I probably would also use smaller files if I wrote this from scratch. On the other hand, libavif is already fully functional, so it seems a little late to restructure the read.c file like this.

I admit I am biased because I am already familiar with the current file structure. Also, I use a source code browser such as Code Search when I analyze code, so long files are not a problem (unless they are too long to load quickly in a Web browser).

If we decide to split read.c like this, it would be better to do it gradually, one foo.{h,c} pair at a time. That will make code review easier.

joedrago:
I echo Wan-Teh's sentiment that smaller files probably would have been better if done over time, but at this point, splitting this would do more harm than good, IMO. I've locally pulled down your branch and tried to do some fairly tame "blames" to discern history on various lines, and it is significantly harder to pin down, and would require people knowing that `read.c` was the home for the first 3 years of all read code in order to find commits.

We can make an effort in future larger-scale features to split them out, but I think we should leave `read.c` alone for now.

PR #807, opened 2022-01-08
Change function in SVT-AV1 0.9.0

Jamaika1:


src/codec_svt.c
```diff
@@ -73,7 +73,9 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         }
         svt_config->encoder_color_format = color_format;
         svt_config->encoder_bit_depth = (uint8_t)image->depth;
+#if SVT_VERSION_PATCHLEVEL < 8
```
wantehchang, line 76: We also need to test SVT_VERSION_MAJOR and SVT_VERSION_MINOR. In pseudo code, the correct check is

`#if SVT_VERSION < 0.8.8`

I took a quick look. We should be able to use the SVT_CHECK_VERSION(major,minor,patch) macro. So I think this test should read:

`#if !SVT_CHECK_VERSION(0, 8, 8)`

src/codec_svt.c
```diff
@@ -85,8 +87,12 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         svt_config->logical_processors = encoder->maxThreads;
         svt_config->enable_adaptive_quantization = AVIF_FALSE;
         // disable 2-pass
+#if SVT_VERSION_PATCHLEVEL >= 8
```
wantehchang, line 90: I think this test should read

`#if SVT_CHECK_VERSION(0, 8, 8)`

wantehchang (changes requested):
Thank you for the pull request. I suggest some changes.

src/codec_svt.c
wantehchang: I just realized that I found the SVT_CHECK_VERSION macro in SVT-VP9, not SVT-AV1:

https://github.com/OpenVisualCloud/SVT-VP9/blob/master/Source/API/EbApiVersion.h.in

SVT-AV1 does not seem to have an equivalent macro. So we will need to define a similar macro ourselves or use some other approach.

src/codec_svt.c
Jamaika1: No problem but in my opinion it isn't future-proof. It will work badly for 0.8.9.
This is just unapproved information. I also don't know if it makes sense to put `#if` since cmd version is defined.
I will add cmd changes after approving version 0.8.8.

src/codec_svt.c
```diff
@@ -73,7 +73,7 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         }
         svt_config->encoder_color_format = color_format;
         svt_config->encoder_bit_depth = (uint8_t)image->depth;
-#if SVT_VERSION_PATCHLEVEL < 8
+#if SVT_VERSION < 0.8.8
```
wantehchang, line 76: I just wanted to clarify a point. In my previous comment, I wrote

> In pseudo code, the correct check is
> 
> `#if SVT_VERSION < 0.8.8`

Note that `#if SVT_VERSION < 0.8.8` is pseudo code. We cannot use `#if SVT_VERSION < 0.8.8` directly.

There are two common ways to do this check correctly. The first way is to write a complicated expression to check `SVT_VERSION_MAJOR`, `SVT_VERSION_MINOR,` and `SVT_VERSION_PATCHLEVEL`. For example

```
#if SVT_VERSION_MAJOR == 0 && \
  (SVT_VERSION_MINOR < 8 || \
   (SVT_VERSION_MINOR == 8 && \
    SVT_VERSION_PATCHLEVEL < 8))
```

The second way is to pack `SVT_VERSION_MAJOR`, `SVT_VERSION_MINOR`, and `SVT_VERSION_PATCHLEVEL` into a packed value, e.g.,

```
#define SVT_VERSION \
  (SVT_VERSION_MAJOR * 10000 + SVT_VERSION_MINOR * 100 + SVT_VERSION_PATCHLEVEL)
```

and then check the packed value, e.g.,

`#if SVT_VERSION < 808`

wantehchang (changes requested):

src/codec_svt.c
wantehchang: It would be best to add either the SVT_CHECK_VERSION macro (which is in [SVT-VP9](https://github.com/OpenVisualCloud/SVT-VP9/blob/master/Source/API/EbApiVersion.h.in)) or the packed-version SVT_VERSION macro to SVT-AV1 first, and then libavif can use the new macro in src/codec_svt.c, providing its own backup definition of the new macro for older versions of SVT-AV1.

src/codec_svt.c
1480c1: @wantehchang https://gitlab.com/AOMediaCodec/SVT-AV1/-/merge_requests/1783, I don't know if you have a gitlab account I can ping, so I will mention it here

wantehchang: @1480c1 Thanks a lot! I assume you prefer SVT_AV1_CHECK_VERSION over a packed-version SVT_AV1_VERSION macro.

src/codec_svt.c
```diff
@@ -73,7 +73,9 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         }
         svt_config->encoder_color_format = color_format;
         svt_config->encoder_bit_depth = (uint8_t)image->depth;
+#if !SVT_CHECK_VERSION(0, 8, 8)
```
wantehchang, line 76: Given the new SVT_AV1_CHECK_VERSION macro that @1480c1 added, I suggest the following.

1. After line 14 in this file, add our backup definition of that macro:

```
#ifndef SVT_AV1_CHECK_VERSION
#define SVT_AV1_CHECK_VERSION(major, minor, patch)                            \
    (SVT_AV1_VERSION_MAJOR > (major) ||                                       \
     (SVT_AV1_VERSION_MAJOR == (major) && SVT_AV1_VERSION_MINOR > (minor)) || \
     (SVT_AV1_VERSION_MAJOR == (major) && SVT_AV1_VERSION_MINOR == (minor) && \
      SVT_AV1_VERSION_PATCHLEVEL >= (patch)))
#endif
```

2. Change this line to
```
#if !SVT_AV1_CHECK_VERSION(0, 8, 8)
```

src/codec_svt.c
```diff
@@ -85,8 +87,12 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         svt_config->logical_processors = encoder->maxThreads;
         svt_config->enable_adaptive_quantization = AVIF_FALSE;
         // disable 2-pass
+#if SVT_CHECK_VERSION(0, 8, 8)
```
wantehchang, line 90: Change this line to
```
#if SVT_AV1_CHECK_VERSION(0, 8, 8)
```

Then run the clang-format command on this file:
```
clang-format -i --style=file src/codec_svt.c
```

wantehchang (changes requested):

src/codec_svt.c
1480c1: > @1480c1 Thanks a lot! I assume you prefer SVT_AV1_CHECK_VERSION over a packed-version SVT_AV1_VERSION macro.

Indifferent right now, since I'm not entirely sure on what the limits of the packed version check would be, but I assume if we manage to go to `0.9999999.1` in svt-av1, then it might break

src/codec_svt.c
wantehchang: Jamaika1: We need to add our backup definition of SVT_AV1_CHECK_VERSION as I suggested in my previous comment.

Please test this pull request with SVT-AV1 0.8.7.

src/codec_svt.c
```diff
@@ -73,7 +73,9 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         }
         svt_config->encoder_color_format = color_format;
         svt_config->encoder_bit_depth = (uint8_t)image->depth;
+#if !SVT_AV1_CHECK_VERSION(0, 8, 8)
```
wantehchang, line 76: Note that I suggest two changes in this comment.

1. Apply the following patch to this file. We need to provide a backup definition of the `SVT_AV1_CHECK_VERSION` macro so that this file can be compiled against a version of SVT-AV1 older than v0.9.0:

```
diff --git a/src/codec_svt.c b/src/codec_svt.c
index d7a7e88..32853fd 100644
--- a/src/codec_svt.c
+++ b/src/codec_svt.c
@@ -9,6 +9,18 @@
 
 #include <string.h>
 
+// The SVT_AV1_CHECK_VERSION macro was added in SVT-AV1 v0.9.0. Define this macro for older versions
+// of SVT-AV1.
+#ifndef SVT_AV1_CHECK_VERSION
+// clang-format off
+#define SVT_AV1_CHECK_VERSION(major, minor, patch)                            \
+    (SVT_AV1_VERSION_MAJOR > (major) ||                                       \
+     (SVT_AV1_VERSION_MAJOR == (major) && SVT_AV1_VERSION_MINOR > (minor)) || \
+     (SVT_AV1_VERSION_MAJOR == (major) && SVT_AV1_VERSION_MINOR == (minor) && \
+      SVT_AV1_VERSION_PATCHLEVEL >= (patch)))
+// clang-format on
+#endif
+
 #define STR_HELPER(x) #x
 #define STR(x) STR_HELPER(x)
 #define SVT_FULL_VERSION STR(SVT_VERSION_MAJOR) "." STR(SVT_VERSION_MINOR) "." STR(SVT_VERSION_PATCHLEVEL)

```

2. Change this line to

`#if !SVT_AV1_CHECK_VERSION(0, 9, 0)`

src/codec_svt.c
```diff
@@ -85,8 +87,12 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         svt_config->logical_processors = encoder->maxThreads;
         svt_config->enable_adaptive_quantization = AVIF_FALSE;
         // disable 2-pass
+#if SVT_AV1_CHECK_VERSION(0, 8, 8)
```
wantehchang, line 90: Change this line to

`#if SVT_AV1_CHECK_VERSION(0, 9, 0)`

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 *   - speed 8-9: re-test and re-adjust speed according to your app needs
 * Update aom.cmd: v3.2.0
 * Update dav1d.cmd: 0.9.2
+* Update svt-av1.cmd: 0.9.0
```
wantehchang, line 29: Nit: 0.9.0 => v0.9.0

The tag is named "v0.9.0".

wantehchang (changes requested):

wantehchang:
@1480c1 Christopher: Should we add SVT-AV1 to tests/docker/build.sh?

1480c1:
> @1480c1 Christopher: Should we add SVT-AV1 to tests/docker/build.sh?

That would probably be good as I don't think we have any integration tests for using shared library on SVT-AV1's side, mostly static libraries.

src/codec_svt.c
wantehchang: Jamaika1 : You didn't add the code that defines the SVT_AV1_CHECK_VERSION macro for older versions of SVT-AV1. We need that code to compile against an installed SVT-AV1, which may be older than v0.9.0. I took care of that in https://github.com/AOMediaCodec/libavif/pull/823.

wantehchang:
@hassount @1480c1 Please review the changes to src/codec_svt.c in this PR. I think those changes are the result of the following item in the SVT-AV1 v0.9 release notes:

> - API configuration structure cleanup removing all invalid or out of date parameters

I added a backup definition of `SVT_AV1_CHECK_VERSION` in https://github.com/AOMediaCodec/libavif/pull/823, so we can use that macro safely.

Thank you!

hassount:
@wantehchang LGTM

wantehchang (approved):

Issue #684, opened 2021-06-23
Refactor/Redesign aviftest


joedrago:
`aviftest` and its JSON-based tests are unreliable, and due to an oversight in `aviftest` (487e3425ee59715afe0656a01e93c084ddc6e36a), I didn't realize it. It is time to rethink our automated/smoke tests in favor of something more reliable and comprehensive.

vrabaud:
Fixed in https://github.com/AOMediaCodec/libavif/pull/811

PR #811, opened 2022-01-14
Remove JSON-based tests (as they are unreliable), along with associated helper code (cJSON, compare)

joedrago:
Related: #684

I don't have a strategy/design yet for a replacement set of tests that do pixel thresholds, but this design doesn't make sense with the spread of AV1 codecs / versions we have these days. We can resurrect pieces of this code in the future if we determine we need it.


tests/aviftest.c
```diff
@@ -467,14 +299,10 @@ int main(int argc, char * argv[])
     // Parse cmdline
     for (int i = 1; i < argc; ++i) {
         char * arg = argv[i];
-        if (!strcmp(arg, "-g")) {
-            generate = AVIF_TRUE;
-        } else if (!strcmp(arg, "--io-only")) {
-            ioOnly = AVIF_TRUE;
+        if (!strcmp(arg, "--io-only")) {
+            // Deprecated, do nothing
```
wantehchang, line 303: Should we print a warning message that says this option is deprecated and should be removed?

CMakeLists.txt
```diff
@@ -532,9 +532,6 @@ if(AVIF_BUILD_TESTS)
     add_executable(aviftest
         apps/shared/y4m.c
         tests/aviftest.c
-        tests/cJSON.c
```
wantehchang, line 535: Please remove the cJSON license from the LICENSE file.

wantehchang (changes requested):

tests/aviftest.c
joedrago: Done!

CMakeLists.txt
joedrago: Done!

wantehchang (approved):
LGTM. Thanks!

PR #814, opened 2022-01-15
Do not create AV1 codecs in avifDecoderParse()

wantehchang:
The creation of AV1 codecs can be completely postponed to
avifDecoderNextImage().

This builds on top of https://github.com/AOMediaCodec/libavif/pull/794.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3342,11 +3342,6 @@ static avifCodec * avifCodecCreateInternal(avifCodecChoice choice)
     return avifCodecCreate(choice, AVIF_CODEC_FLAG_CAN_DECODE);
 }
 
-static void avifDecoderFlush(avifDecoder * decoder)
-{
-    avifDecoderDataResetCodec(decoder->data);
-}
-
 // If alpha is AVIF_FALSE, searches for the primary color item (parentItemID is ignored in this case).
 // If alpha is AVIF_TRUE, searches for the auxiliary alpha item whose parent item ID is parentItemID.
 // Returns the target item if found, or NULL.
@@ -3800,7 +3795,6 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         return AVIF_RESULT_BMFF_PARSE_FAILED;
     }
 
-    avifDecoderFlush(decoder);
     return AVIF_RESULT_OK;
 }
 
@@ -3812,9 +3806,7 @@ static avifResult avifDecoderPrepareTiles(avifDecoder * decoder, uint32_t nextIm
         // Ensure there's an AV1 codec available before doing anything else
         if (!tile->codec) {
             tile->codec = avifCodecCreateInternal(decoder->codecChoice);
-            if (!tile->codec) {
-                return AVIF_RESULT_NO_CODEC_AVAILABLE;
-            }
+            AVIF_CHECKERR(tile->codec, AVIF_RESULT_NO_CODEC_AVAILABLE);
             tile->codec->diag = &decoder->diag;
             tile->codec->operatingPoint = tile->operatingPoint;
             tile->codec->allLayers = tile->input->allLayers;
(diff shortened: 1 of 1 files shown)


wantehchang:
Joe: Do you think this will work?

y-guyon (approved):

src/read.c
```diff
@@ -3800,7 +3795,6 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         return AVIF_RESULT_BMFF_PARSE_FAILED;
     }
 
-    avifDecoderFlush(decoder);
```
wantehchang: Can this `avifDecoderFlush()` call at the end of `avifDecoderReset()` be omitted?

src/read.c
```diff
@@ -3821,7 +3805,13 @@ static avifResult avifDecoderPrepareTiles(avifDecoder * decoder, uint32_t nextIm
 
         // Ensure there's an AV1 codec available before doing anything else
         if (!tile->codec) {
-            return AVIF_RESULT_NO_CODEC_AVAILABLE;
+            tile->codec = avifCodecCreateInternal(decoder->codecChoice);
+            if (!tile->codec) {
+                return AVIF_RESULT_NO_CODEC_AVAILABLE;
+            }
```
y-guyon, line 3811: Optional:

```c
AVIF_CHECKERR(tile->codec, AVIF_RESULT_NO_CODEC_AVAILABLE);
```

src/read.c
```diff
@@ -4121,7 +4111,7 @@ avifResult avifDecoderNthImage(avifDecoder * decoder, uint32_t frameIndex)
     if ((nearestKeyFrame > (decoder->imageIndex + 1)) || (requestedIndex <= decoder->imageIndex)) {
         // If we get here, a decoder flush is necessary
```
y-guyon, line 4112: `avifDecoderFlush()` does not longer exists. I suggest replacing the word "flush" by a more verbose explanation, such as

```
// We need to start from a fresh key frame so discard the unused decoder state and its previous frames. This will force the setup of a new decoder instance in avifDecoderNextImage().
```

y-guyon (approved):

src/read.c
vigneshvg: Done.

src/read.c
vigneshvg: Done.

src/read.c
wantehchang: I was asking why it is correct to remove the `avifDecoderFlush()` call at the end of `avifDecoderReset()`. I see this was suggested by Yannis and he explained why this `avifDecoderFlush()` call can be removed.

vigneshvg: Yes, essentially this is okay because we call `avifDecoderDataClearTiles` in line 3416 and that does more than what `avifDecoderFlush` used to do.

src/read.c
```diff
@@ -4114,14 +4102,18 @@ avifResult avifDecoderNthImage(avifDecoder * decoder, uint32_t frameIndex)
...
+        // Fall through to resetting the decoder data and start decoding from
+        // the nearest key frame.
     }
 
     int nearestKeyFrame = (int)avifDecoderNearestKeyframe(decoder, frameIndex);
     if ((nearestKeyFrame > (decoder->imageIndex + 1)) || (requestedIndex <= decoder->imageIndex)) {
-        // If we get here, a decoder flush is necessary
+        // If we get here, we need to start decoding from the nearest key frame.
+        // So discard the unused decoder state and its previous frames. This
+        // will force the setup of a new decoder instance in
```
wantehchang, line 4113: Nit: Please clarify "a new decoder instance". It is ambiguous because "decoder" could be misinterpreted as avifDecoder. But the comment is referring to avifCodec instances.

One way to clarify this is to change "a new decoder instance" to "new AV1 decoder instances". Note the plural, because we may create multiple AV1 decoder instances, e.g., to decode the color and alpha sub-images.

An alternative is to say "new avifCodec instances".

vigneshvg: Oops, i just merged the change. I will update the comment in a follow-up PR.

vigneshvg: done here: https://github.com/AOMediaCodec/libavif/pull/1317

src/read.c
wantehchang: One complication is that after we call `avifDecoderDataClearTiles` in line 3416, we call `avifDecoderDataCreateTile`, which creates `tile->image` as empty images. So we need to verify that `tile->image` is still empty at the end of `avifDecoderReset`. I validated that by adding some assertions and running our test suite, but this is also very plausible because we don't have any `tile->codec` to decode data into `tile->image`.

This analysis shows that more work could be moved out of `avifDecoderParse` to optimize for the "parse container metadata only" use case. For example, the empty `tile->image` that we create is not used in this use case.

PR #815, opened 2022-01-17
Return false when compareYUVA() empty images

y-guyon:


joedrago:
This file is being removed in its entirety in #811, so this is not necessary.

y-guyon:
Ah yes sorry, I missed that.

PR #816, opened 2022-01-19
Move codec enabled message after check passed

tongyuantongyu:
Currently these messages
`libavif: Codec enabled: libfoo (decode)`
are printed before any check about whether the codec works. This is a bit confusing.

Print these messages after we've done all the check and config seems to be more suitable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,7 +256,6 @@ set(AVIF_CODEC_INCLUDES)
 set(AVIF_CODEC_LIBRARIES)
 
 if(AVIF_CODEC_DAV1D)
-    message(STATUS "libavif: Codec enabled: dav1d (decode)")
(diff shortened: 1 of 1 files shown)


wantehchang (approved):
LGTM.

wantehchang:
I tested this pull request. I passed `-DAVIF_CODEC_SVT=ON -DAVIF_LOCAL_SVT=ON` to cmake but there was not a SVT-AV1 source tree in the ext/ directory.

Before this pull request, the cmake output was:

```
-- libavif: Codec enabled: svt (encode)
CMake Error at CMakeLists.txt:368 (message):
  libavif: compiled svt library is missing (in ext/SVT-AV1/Bin/Release),
  bailing out
```

After this pull request, the cmake output was:

```
CMake Error at CMakeLists.txt:370 (message):
  libavif: compiled svt library is missing (in ext/SVT-AV1/Bin/Release),
  bailing out
```

So the "`-- libavif: Codec enabled: svt (encode)`" line was not printed.

On the other hand, this pull request does not change the cmake output when cmake succeeds.

PR #817, opened 2022-01-20
reformat: Support F16 Half Float conversion in avifRGBImage

vigneshvg:
Some platforms (Android for example [1]) support only the F16 Half
Float format for images with bits-per-pixel == 16. This PR adds
support for converting the RGBA pixels into half float format.

It does so by adding a setting variable to avifRGBImage and then
doing the conversion at the end of avifImageYUVToRGB. It also
invokes libyuv if available to do the conversion with SIMD
optimizations.

While there is potential to make this even faster (instead of
doing this conversion in the end, we could potentially do it while
we do the YUV -> RGB conversion), that is a much larger change and
the performance gain from that may or may not be worth the trade
off in complicating the code paths.

[1] https://developer.android.com/reference/android/graphics/Bitmap.Config#RGBA_F16

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -22,6 +22,9 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
     if ((rgb->depth != 8) && (rgb->depth != 10) && (rgb->depth != 12) && (rgb->depth != 16)) {
         return AVIF_FALSE;
     }
+    if (rgb->isFloat && rgb->depth != 16) {
+        return AVIF_FALSE;
+    }
 
     // These matrix coefficients values are currently unsupported. Revise this list as more support is added.
     //
@@ -186,6 +189,10 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
     avifReformatState state;
     if (!avifPrepareReformatState(image, rgb, &state)) {
         return AVIF_RESULT_REFORMAT_FAILED;
+    }
+
+    if (rgb->isFloat) {
+        return AVIF_RESULT_NOT_IMPLEMENTED;
     }
 
     avifAlphaMultiplyMode alphaMode = AVIF_ALPHA_MULTIPLY_MODE_NO_OP;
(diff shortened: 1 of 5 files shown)


src/reformat.c
```diff
@@ -1073,6 +1081,32 @@ static avifResult avifImageYUV8ToRGB8Mono(const avifImage * image, avifRGBImage
...
+};
+
+static avifResult avifRGBImageToF16(avifRGBImage * rgb) {
+    avifResult libyuvResult = avifRGBImageToF16LibYUV(rgb);
+    if (libyuvResult != AVIF_RESULT_NOT_IMPLEMENTED) {
+        return libyuvResult;
+    }
+    const int channelCount = avifRGBFormatChannelCount(rgb->format);
+    const float scale = 1.0f / ((1 << rgb->depth) - 1);
+    const float multiplier = 1.9259299444e-34f * scale;
```
wantehchang, line 1096: It would be good to get Frank Barchard's permission to copy his comments before the `HalfFloatRow_C()` function.

Until then, let's add a link to the libyuv function as the source of this algorithm.

src/reformat.c
```diff
@@ -1073,6 +1081,32 @@ static avifResult avifImageYUV8ToRGB8Mono(const avifImage * image, avifRGBImage
...
+  float f;
+  uint32_t u32;
+};
+
+static avifResult avifRGBImageToF16(avifRGBImage * rgb) {
+    avifResult libyuvResult = avifRGBImageToF16LibYUV(rgb);
+    if (libyuvResult != AVIF_RESULT_NOT_IMPLEMENTED) {
+        return libyuvResult;
+    }
+    const int channelCount = avifRGBFormatChannelCount(rgb->format);
```
wantehchang, line 1094: Nit: int => uint32_t

wantehchang (changes requested):
Vignesh: I suggest a few simple changes.

src/reformat.c
```diff
@@ -1073,6 +1081,32 @@ static avifResult avifImageYUV8ToRGB8Mono(const avifImage * image, avifRGBImage
     return AVIF_RESULT_OK;
 }
 
+union F16 {
+  float f;
+  uint32_t u32;
+};
```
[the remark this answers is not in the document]
vigneshvg: Done.

src/reformat.c
```diff
@@ -1073,6 +1081,32 @@ static avifResult avifImageYUV8ToRGB8Mono(const avifImage * image, avifRGBImage
...
+    }
+    const int channelCount = avifRGBFormatChannelCount(rgb->format);
+    const float scale = 1.0f / ((1 << rgb->depth) - 1);
+    const float multiplier = 1.9259299444e-34f * scale;
+    uint16_t * pixelRowBase = (uint16_t*)rgb->pixels;
+    const uint32_t stride = rgb->rowBytes >> 1;
+    for (uint32_t j = 0; j < rgb->height; ++j) {
+        uint16_t * pixel = pixelRowBase;
+        for (uint32_t i = 0; i < rgb->width * channelCount; ++i, ++pixel) {
+            const union F16 f16 = {.f = *pixel * multiplier};
```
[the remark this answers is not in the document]
vigneshvg: Done.

src/reformat.c
```diff
@@ -1073,6 +1081,32 @@ static avifResult avifImageYUV8ToRGB8Mono(const avifImage * image, avifRGBImage
     return AVIF_RESULT_OK;
 }
 
+union F16 {
+  float f;
+  uint32_t u32;
+};
+
+static avifResult avifRGBImageToF16(avifRGBImage * rgb) {
```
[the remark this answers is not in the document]
vigneshvg: Done.

src/reformat.c
vigneshvg: Done.

src/reformat.c
vigneshvg: Done.

src/reformat_libyuv.c
```diff
@@ -499,6 +504,16 @@ avifResult avifRGBImageUnpremultiplyAlphaLibYUV(avifRGBImage * rgb)
     return AVIF_RESULT_NOT_IMPLEMENTED;
 }
 
+avifResult avifRGBImageToF16LibYUV(avifRGBImage * rgb)
+{
+    const float scale = 1.0f / ((1 << rgb->depth) - 1);
+    HalfFloatPlane((const uint16_t*)rgb->pixels, rgb->rowBytes,
```
wantehchang, line 510: I found that HalfFloatPlane() returns 0 on success, -1 on failure. So in theory we should check its return value. This means avifRGBImageToF16LibYUV() will also return `AVIF_RESULT_INVALID_ARGUMENT`, we will need to change avifRGBImageToF16() back to return avifResult.

However, in the current implementation of HalfFloatPlane(), it only returns -1 here:

```
  if (!src_y || !dst_y || width <= 0 || height == 0) {
    return -1;
  }
```

So we can verify we don't pass such invalid arguments to HalfFloatPlane(), and therefore this HalfFloatPlane() call cannot fail.

What do you think?

vigneshvg: Ah nice catch. I didn't notice that HalfFloatPlane() returned a return value (most libyuv functions don't).

I have updated the code to check the return value.

wantehchang (approved):
LGTM.

PR #818, opened 2022-01-21
Fix alpha copy in aomCodecEncodeImage()

y-guyon:
The stride in bytes was mistaken for the number of samples to copy.


wantehchang (approved):
LGTM. Thanks!

I don't know why I missed the "if (alpha)" branch. The "else" branch is correct.

PR #819, opened 2022-01-21
android_jni: Support RGBA_F16 Bitmaps

vigneshvg:
libavif now supports conversion of RGBA pixels into F16 formats
for AVIF images with depths 10 and 12. Make use of that feature to
support RGBA_F16 bitmaps on the android JNI wrapper.

Note that apps can still pass an ARGB_8888 Bitmap for 10/12 bit
images and the wrapper will merely downscale them to 8 bits in that
case (existing behavior).

### DIFF
--- a/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
+++ b/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
@@ -23,6 +23,7 @@ namespace {
 
 jfieldID global_info_width;
 jfieldID global_info_height;
+jfieldID global_info_depth;
 
 // RAII wrapper class that properly frees the decoder related objects on
 // destruction.
@@ -75,6 +76,7 @@ jint JNI_OnLoad(JavaVM* vm, void* /*reserved*/) {
       env->FindClass("org/aomedia/avif/android/AvifDecoder$Info");
   global_info_width = env->GetFieldID(info_class, "width", "I");
   global_info_height = env->GetFieldID(info_class, "height", "I");
+  global_info_depth = env->GetFieldID(info_class, "depth", "I");
   return JNI_VERSION_1_6;
 }
 
@@ -94,6 +96,7 @@ FUNC(jboolean, getInfo, jobject encoded, int length, jobject info) {
   }
   env->SetIntField(info, global_info_width, decoder.decoder->image->width);
   env->SetIntField(info, global_info_height, decoder.decoder->image->height);
+  env->SetIntField(info, global_info_depth, decoder.decoder->image->depth);
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


vigneshvg:
@wantehchang this PR is ready for review.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -124,9 +127,15 @@ FUNC(jboolean, decode, jobject encoded, int length, jobject bitmap) {
         decoder.decoder->image->height);
     return false;
   }
-  // Ensure that the bitmap format is RGBA_8888. For now, we only support that.
-  if (bitmap_info.format != ANDROID_BITMAP_FORMAT_RGBA_8888) {
-    LOGE("Bitmap format (%d) is not supported.", bitmap_info.format);
+  // Ensure that the bitmap format is either RGBA_8888 or RGBA_F16 based on the
+  // image's bitdepth.
+  if ((decoder.decoder->image->depth == 8 &&
+       bitmap_info.format != ANDROID_BITMAP_FORMAT_RGBA_8888) ||
```
wantehchang, line 133: libavif can convert 8-bit YUVA to RGBA_F16. Would you like to support that?

I.e, we can simply check:

```
  // Ensure that the bitmap format is either RGBA_8888 or RGBA_F16.
  if (bitmap_info.format != ANDROID_BITMAP_FORMAT_RGBA_8888 &&
      bitmap_info.format != ANDROID_BITMAP_FORMAT_RGBA_F16) {
```

vigneshvg: I considered that feature. But F16 bitmaps and displaying HDR content in android are generally considered expensive and allowing this would indirectly encourage bad practice. So i thought it was best to keep this out.

Please let me know if you feel otherwise and i can update this.

tongyuantongyu: Whether an image is WCG or HDR depends on the image's ColorPrimaries and TranferCharacteristic value, so if F16 is required to display WCG/HDR content then maybe we shouldn't limit it there. Storing HDR in 8bit is probably bad, but it's not the fault of the image viewer (or alike).

wantehchang: This issue came up in Chrome recently: https://crbug.com/1266456. As I understand it, Chrome renders HDR images in F16 image frame buffers, and 8-bit HDR images can't be rendered because the blink::AVIFImageDecoder class in Chrome does not have the code to convert 8-bit YUVA to F16 RGBA.

Your Android JNI bindings can easily support this if someone requests it.

vigneshvg: Sounds good. I have updated the PR to allow 8 -> 16 conversion as well.

wantehchang (approved):
LGTM.

Issue #820, opened 2022-01-21
Handle out-of-memory errors, such as avifAlloc() failures


wantehchang:
This issue tracks the work on handling out-of-memory errors, starting with avifAlloc() failures.

I will do this work in installments because this requires a lot of changes.

y-guyon:
All `avifAlloc()` calls should be checked now.

vrabaud:
I confirm all avifAlloc are checked.

wantehchang:
Yannis: Thank you very much for all your work on this issue! This is an enormous amount of work and requires patience and perseverance.

I changed the assignee to you and removed myself. (The initial work I did is insignificant compared with what you have done.)

Does this mean we can remove the abort() call in avifAlloc() now?

y-guyon:
Thanks.

> Does this mean we can remove the abort() call in avifAlloc() now?

Yes. I was wondering if adding a test or fuzz target to exercise `malloc()` failures would be worth it. Maybe it can be done after the change.

y-guyon:
Once the changes above are merged, the `apps` and `src` folders are handled.

There are a few allocations in `third_party/libyuv` (a small portion of code copied from upstream [libyuv](https://chromium.googlesource.com/libyuv/libyuv/)). These are not checked.  
Unfortunately it also means using the whole libyuv library as a dependency is a liability. I filed https://bugs.chromium.org/p/libyuv/issues/detail?id=968.

I can imagine other dependencies such as libaom bringing the same risk. Should this bug be closed before all dependencies are safe?

wantehchang:
I suggest closing this issue as completed and open a new issue about libyuv.

Unlike libyuv, libaom aims to handle memory allocation failures (see https://crbug.com/aomedia/3276), so it doesn't seem necessary to file a new issue about libaom.

wantehchang:
Sorry, I forgot about the code in third_party/libyuv. In my previous comment, I was referring to the upstream libyuv.

We can probably also open a new issue about the code in third_party/libyuv. Or we can leave this issue open until that code is fixed.

y-guyon:
> we can leave this issue open until that code is fixed.

I agree.

> Unlike libyuv, libaom aims to handle memory allocation failures (see https://crbug.com/aomedia/3276), so it doesn't seem necessary to file a new issue about libaom.

Great. What about other dependencies, especially dav1d? I think libsharpyuv is fine too, right @vrabaud?

vrabaud:
[...]

PR #821, opened 2022-01-21
Handle avifArrayCreate() failures

wantehchang:
Add the AVIF_RESULT_OUT_OF_MEMORY error code.

Add the avifArrayPop() function to remove the last element in the array.

Bug: https://github.com/AOMediaCodec/libavif/issues/820.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -237,17 +237,35 @@ typedef struct avifSampleTable
     uint32_t allSamplesSize; // If this is non-zero, sampleSizes will be empty and all samples will be this size
 } avifSampleTable;
 
+static void avifSampleTableDestroy(avifSampleTable * sampleTable);
+
 static avifSampleTable * avifSampleTableCreate()
 {
     avifSampleTable * sampleTable = (avifSampleTable *)avifAlloc(sizeof(avifSampleTable));
     memset(sampleTable, 0, sizeof(avifSampleTable));
-    avifArrayCreate(&sampleTable->chunks, sizeof(avifSampleTableChunk), 16);
-    avifArrayCreate(&sampleTable->sampleDescriptions, sizeof(avifSampleDescription), 2);
-    avifArrayCreate(&sampleTable->sampleToChunks, sizeof(avifSampleTableSampleToChunk), 16);
-    avifArrayCreate(&sampleTable->sampleSizes, sizeof(avifSampleTableSampleSize), 16);
-    avifArrayCreate(&sampleTable->timeToSamples, sizeof(avifSampleTableTimeToSample), 16);
-    avifArrayCreate(&sampleTable->syncSamples, sizeof(avifSyncSample), 16);
+    if (!avifArrayCreate(&sampleTable->chunks, sizeof(avifSampleTableChunk), 16)) {
+        goto error;
+    }
+    if (!avifArrayCreate(&sampleTable->sampleDescriptions, sizeof(avifSampleDescription), 2)) {
+        goto error;
+    }
+    if (!avifArrayCreate(&sampleTable->sampleToChunks, sizeof(avifSampleTableSampleToChunk), 16)) {
+        goto error;
+    }
+    if (!avifArrayCreate(&sampleTable->sampleSizes, sizeof(avifSampleTableSampleSize), 16)) {
+        goto error;
+    }
+    if (!avifArrayCreate(&sampleTable->timeToSamples, sizeof(avifSampleTableTimeToSample), 16)) {
+        goto error;
+    }
+    if (!avifArrayCreate(&sampleTable->syncSamples, sizeof(avifSyncSample), 16)) {
+        goto error;
+    }
     return sampleTable;
+
+error:
+    avifSampleTableDestroy(sampleTable);
+    return NULL;
 }
 
 static void avifSampleTableDestroy(avifSampleTable * sampleTable)
@@ -346,8 +364,14 @@ avifCodecDecodeInput * avifCodecDecodeInputCreate(void)
 {
     avifCodecDecodeInput * decodeInput = (avifCodecDecodeInput *)avifAlloc(sizeof(avifCodecDecodeInput));
     memset(decodeInput, 0, sizeof(avifCodecDecodeInput));
-    avifArrayCreate(&decodeInput->samples, sizeof(avifDecodeSample), 1);
+    if (!avifArrayCreate(&decodeInput->samples, sizeof(avifDecodeSample), 1)) {
+        goto error;
+    }
     return decodeInput;
+
+error:
+    avifFree(decodeInput);
+    return NULL;
 }
 
 void avifCodecDecodeInputDestroy(avifCodecDecodeInput * decodeInput)
@@ -666,13 +690,23 @@ typedef struct avifMeta
     uint32_t primaryItemID;
 } avifMeta;
 
+static void avifMetaDestroy(avifMeta * meta);
+
 static avifMeta * avifMetaCreate()
 {
     avifMeta * meta = (avifMeta *)avifAlloc(sizeof(avifMeta));
     memset(meta, 0, sizeof(avifMeta));
-    avifArrayCreate(&meta->items, sizeof(avifDecoderItem), 8);
-    avifArrayCreate(&meta->properties, sizeof(avifProperty), 16);
+    if (!avifArrayCreate(&meta->items, sizeof(avifDecoderItem), 8)) {
+        goto error;
+    }
+    if (!avifArrayCreate(&meta->properties, sizeof(avifProperty), 16)) {
+        goto error;
+    }
     return meta;
+
+error:
+    avifMetaDestroy(meta);
+    return NULL;
 }
 
 static void avifMetaDestroy(avifMeta * meta)
@@ -704,11 +738,21 @@ static avifDecoderItem * avifMetaFindItem(avifMeta * meta, uint32_t itemID)
     }
 
     avifDecoderItem * item = (avifDecoderItem *)avifArrayPushPtr(&meta->items);
-    avifArrayCreate(&item->properties, sizeof(avifProperty), 16);
-    avifArrayCreate(&item->extents, sizeof(avifExtent), 1);
+    if (!avifArrayCreate(&item->properties, sizeof(avifProperty), 16)) {
+        goto error;
+    }
+    if (!avifArrayCreate(&item->extents, sizeof(avifExtent), 1)) {
+        goto error;
+    }
     item->id = itemID;
     item->meta = meta;
     return item;
+
+error:
+    avifArrayDestroy(&item->extents);
+    avifArrayDestroy(&item->properties);
+    avifArrayPop(&meta->items);
+    return NULL;
 }
 
 typedef struct avifDecoderData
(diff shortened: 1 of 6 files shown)


wantehchang:
@y-guyon Please review. Thanks.

src/read.c
```diff
@@ -237,17 +237,35 @@ typedef struct avifSampleTable
...
 {
     avifSampleTable * sampleTable = (avifSampleTable *)avifAlloc(sizeof(avifSampleTable));
     memset(sampleTable, 0, sizeof(avifSampleTable));
-    avifArrayCreate(&sampleTable->chunks, sizeof(avifSampleTableChunk), 16);
-    avifArrayCreate(&sampleTable->sampleDescriptions, sizeof(avifSampleDescription), 2);
-    avifArrayCreate(&sampleTable->sampleToChunks, sizeof(avifSampleTableSampleToChunk), 16);
-    avifArrayCreate(&sampleTable->sampleSizes, sizeof(avifSampleTableSampleSize), 16);
-    avifArrayCreate(&sampleTable->timeToSamples, sizeof(avifSampleTableTimeToSample), 16);
-    avifArrayCreate(&sampleTable->syncSamples, sizeof(avifSyncSample), 16);
+    if (avifArrayCreate(&sampleTable->chunks, sizeof(avifSampleTableChunk), 16) != AVIF_RESULT_OK) {
```
vigneshvg, line 246: it seems like we can combine these if conditions with an || and then do the "error" path inside the if block ? This way we can avoid the goto.

For example:
```
if (avifArrayCreate(...) != AVIF_RESULT_OK ||
    avifArrayCreate(...) != AVIF_RESULT_OK ||
    avifArrayCreate(...) != AVIF_RESULT_OK) {
    avifSampleTableDestroy(sampleTable);
    return NULL;
}
return sampleTable;
```

This comment also applies for the rest of the code changes where we try to create several arrays and have to fail if one of them fails.

vigneshvg (changes requested):
I suggest a minor change to avoid the (seemingly unnecessary) use of goto.

src/utils.c
```diff
@@ -78,15 +78,21 @@ uint64_t avifNTOH64(uint64_t l)
 
 AVIF_ARRAY_DECLARE(avifArrayInternal, uint8_t, ptr);
 
-void avifArrayCreate(void * arrayStruct, uint32_t elementSize, uint32_t initialCapacity)
+// On error, must set arr->ptr to NULL and both arr->count and arr->capacity to 0.
+avifResult avifArrayCreate(void * arrayStruct, uint32_t elementSize, uint32_t initialCapacity)
```
y-guyon, line 82: Are you planning to automate (e.g. through a macro) the error code forwarding? Because currently the status is always compared to OK and the value itself never transmitted up the call stack. Otherwise a bool would be enough (which seems traditional to me for functions that "just" allocate memory).

src/write.c
```diff
@@ -46,7 +46,10 @@ avifCodecEncodeOutput * avifCodecEncodeOutputCreate(void)
 {
     avifCodecEncodeOutput * encodeOutput = (avifCodecEncodeOutput *)avifAlloc(sizeof(avifCodecEncodeOutput));
     memset(encodeOutput, 0, sizeof(avifCodecEncodeOutput));
-    avifArrayCreate(&encodeOutput->samples, sizeof(avifEncodeSample), 1);
+    if (avifArrayCreate(&encodeOutput->samples, sizeof(avifEncodeSample), 1) != AVIF_RESULT_OK) {
```
y-guyon, line 49: The pattern is different from avifEncoderDataCreate() but I do not see why. I suggest both or neither use a goto.

y-guyon (approved):

src/read.c
y-guyon: Although I prefer a more modern pattern, I believe keeping a goto everywhere for consistency is clearer. I suggest merging the condition with || for shortness but keeping the goto.

wantehchang: Thank you both for the suggestion of merging the conditions with `||`. I wrote this as multiple if statements to make it easier to debug this in a debugger. This style is more common in the libavif codebase. Here is an example, which uses a macro (`CHECK`) to make the code less verbose:

```
#define CHECK(A)               \
    do {                       \
        if (!(A))              \
            return AVIF_FALSE; \
    } while (0)

static avifBool avifParseCleanApertureBoxProperty(avifProperty * prop, const uint8_t * raw, size_t rawLen, avifDiagnostics * diag)
{
    BEGIN_STREAM(s, raw, rawLen, diag, "Box[clap]");

    avifCleanApertureBox * clap = &prop->u.clap;
    CHECK(avifROStreamReadU32(&s, &clap->widthN));    // unsigned int(32) cleanApertureWidthN;
    CHECK(avifROStreamReadU32(&s, &clap->widthD));    // unsigned int(32) cleanApertureWidthD;
    CHECK(avifROStreamReadU32(&s, &clap->heightN));   // unsigned int(32) cleanApertureHeightN;
    CHECK(avifROStreamReadU32(&s, &clap->heightD));   // unsigned int(32) cleanApertureHeightD;
    CHECK(avifROStreamReadU32(&s, &clap->horizOffN)); // unsigned int(32) horizOffN;
    CHECK(avifROStreamReadU32(&s, &clap->horizOffD)); // unsigned int(32) horizOffD;
    CHECK(avifROStreamReadU32(&s, &clap->vertOffN));  // unsigned int(32) vertOffN;
    CHECK(avifROStreamReadU32(&s, &clap->vertOffD));  // unsigned int(32) vertOffD;
    return AVIF_TRUE;
}
```

src/utils.c
wantehchang: Good question. This function is an internal function. We can make sure it never receives invalid arguments, so the only reason for failure is the out-of-memory error.

I am indeed debating whether this function should return avifBool instead. Is that also your preference?

src/write.c
wantehchang: This function has only one error path, so it is not necessary to use goto to share error handling code. But I will switch this function to goto for consistency.

src/utils.c
```diff
@@ -78,15 +78,21 @@ uint64_t avifNTOH64(uint64_t l)
 
 AVIF_ARRAY_DECLARE(avifArrayInternal, uint8_t, ptr);
 
-void avifArrayCreate(void * arrayStruct, uint32_t elementSize, uint32_t initialCapacity)
+// On error, must set arr->ptr to NULL and both arr->count and arr->capacity to 0.
```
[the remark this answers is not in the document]
wantehchang: This comment specifies a contract that this function must satisfy. I will add "this function" to the comment to clarify this.

src/utils.c
wantehchang: Good question. This function is an internal function. We can make sure it never receives invalid arguments, so it can only fail because it runs out of memory.

I am indeed debating whether this function should return `avifBool` instead. Is that also your preference?

src/write.c
wantehchang: The reason I use `goto` in this PR is to let error paths share error handling code. This function has only one error path, so it is not necessary to share error handling code. But I will switch this function to `goto` for consistency.

src/read.c
wantehchang: Thank you both for the suggestion of merging the conditions with `||`. I wrote this as multiple if statements to make it easier to debug this in a debugger. This style is more common in the libavif codebase. Here is an example, which uses a macro (`CHECK`) to make the code less verbose:

```
#define CHECK(A)               \
    do {                       \
        if (!(A))              \
            return AVIF_FALSE; \
    } while (0)

static avifBool avifParseCleanApertureBoxProperty(avifProperty * prop, const uint8_t * raw, size_t rawLen, avifDiagnostics * diag)
{
    BEGIN_STREAM(s, raw, rawLen, diag, "Box[clap]");

    avifCleanApertureBox * clap = &prop->u.clap;
    CHECK(avifROStreamReadU32(&s, &clap->widthN));    // unsigned int(32) cleanApertureWidthN;
    CHECK(avifROStreamReadU32(&s, &clap->widthD));    // unsigned int(32) cleanApertureWidthD;
    CHECK(avifROStreamReadU32(&s, &clap->heightN));   // unsigned int(32) cleanApertureHeightN;
    CHECK(avifROStreamReadU32(&s, &clap->heightD));   // unsigned int(32) cleanApertureHeightD;
    CHECK(avifROStreamReadU32(&s, &clap->horizOffN)); // unsigned int(32) horizOffN;
    CHECK(avifROStreamReadU32(&s, &clap->horizOffD)); // unsigned int(32) horizOffD;
    CHECK(avifROStreamReadU32(&s, &clap->vertOffN));  // unsigned int(32) vertOffN;
    CHECK(avifROStreamReadU32(&s, &clap->vertOffD));  // unsigned int(32) vertOffD;
    return AVIF_TRUE;
}
```

y-guyon: I would greatly appreciate a library-wide AVIF_CHECK macro for error handling. Unfortunately many but not all places use a goto, some with a "cleanup" label, others with an "error" one. This is past the scope of this PR but unifying all these with a single macro (handling a bool or even better an AVIF_RESULT) would be handy and less verbose, at the expense of having a goto label nearly everywhere.

src/utils.c
y-guyon: If there is a global AVIF_CHECK_STATUS macro as mentioned above, I would tend to return a status to leverage it. Otherwise I do not have a strong preference, maybe slightly towards an avifBool to indicate that the only possible error is a memory one indeed. Up to you.

src/read.c
wantehchang: I took a look at the "cleanup" label. It is used in functions that must free resources (clean up) before returning, whether success or failure. The functions modified in this pull request free resources only when returning on error. That's why I named the label "error" in this pull request.

src/utils.c
wantehchang: OK, I will change `avifArrayCreate()` to return `avifBool`.

wantehchang:
@y-guyon @vigneshvg Please review the new patch. Thanks.

src/read.c
wantehchang: We only need to use a `goto` label if we need to perform cleanup actions before returning and we want to share the cleanup actions between multiple code paths. If no cleanup is necessary, then it's simpler to just return from the if statements.

src/write.c
```diff
@@ -142,8 +158,16 @@ static avifEncoderItem * avifEncoderDataCreateItem(avifEncoderData * data, const
...
-    avifArrayCreate(&item->mdatFixups, sizeof(avifOffsetFixup), 4);
+    if (!avifArrayCreate(&item->mdatFixups, sizeof(avifOffsetFixup), 4)) {
+        goto error;
+    }
     return item;
+
+error:
+    avifCodecEncodeOutputDestroy(item->encodeOutput);
+    --data->lastItemID;
+    avifArrayPushPtr(&data->items);
```
vigneshvg, line 169: This is incorrect. It should be `avifArrayPop()` ?

src/write.c
```diff
@@ -142,8 +158,16 @@ static avifEncoderItem * avifEncoderDataCreateItem(avifEncoderData * data, const
     item->infeNameSize = infeNameSize;
     item->encodeOutput = avifCodecEncodeOutputCreate();
     item->cellIndex = cellIndex;
-    avifArrayCreate(&item->mdatFixups, sizeof(avifOffsetFixup), 4);
+    if (!avifArrayCreate(&item->mdatFixups, sizeof(avifOffsetFixup), 4)) {
```
vigneshvg, line 161: optional: if you move this to line 154 (immediately after *item is populated for the first time), the error handling code can be simplified to simply pop the array element instead of having to revert a bunch of other things.

vigneshvg (changes requested):

src/write.c
wantehchang: Yes! Thank you so much for catching this.

src/write.c
wantehchang: Thank you for the suggestion. We still need to handle the failure of `item->encodeOutput = avifCodecEncodeOutputCreate()` at line 159, so in this pull request I will just move the `++data->lastItemID` operation right before the successful return.

wantehchang:
@vigneshvg Please review the new patch. Thanks!

vigneshvg (approved):
lgtm.

y-guyon (approved):

PR #824, opened 2022-01-26
Change  SVT_FULL_VERSION

Jamaika1:


wantehchang (approved):
LGTM. Thanks!

The `SVT_AV1_CVS_VERSION` macro was added in SVT-AV1 v0.8.2, which was released on 2020-04-18. Since this file was added to libavif on 2020-10-05 (in commit 17f67ffba36780ce6564f1155afae3fd565b5b11), we can safely assume `SVT_AV1_CVS_VERSION` is available.

wantehchang:
Jamaika1: I merged this pull request, but had to revert it because it broke the compilation with a local SVT-AV1 tree in the ext/ directory. The compiler error message is:

```
libavif/src/codec_svt.c:8:10: fatal error: 'svt-av1/EbVersion.h' file not found
#include "svt-av1/EbVersion.h"
         ^~~~~~~~~~~~~~~~~~~~~
```

Did you test this pull request with SVT-AV1 installed on your computer?

1480c1:
As far as I remember, EbVersion.h isn't installed, `svt_av1_get_version()` should return the string that has the same content though

wantehchang:
Christopher: Thank you for the suggestion. `svt_av1_get_version()` was added in v0.9.0, so we will need to guard it with `#if SVT_AV1_CHECK_VERSION(0, 9, 0)`.

wantehchang:
I confirm EbVersion.h isn't installed.

Issue #638, opened 2021-05-25
Incremental decoding for grid image


tongyuantongyu:
Though currently no decoder can do partial decoding, for grid images every grid tile is an independent AV1 item, so we should be able to do incremental decoding at least for grid images.

PR #827, opened 2022-01-26
Incremental decoding of AVIF grid tiles

y-guyon:
Allow parts of an image encoded as a grid to be accessed by the user
before all bytes are available. It uses AVIF cells and not AV1 tiles.

sofa_grid1x5_420.avif has no copyright (created from a photo I took).

Code formatted with the .clang-format config file.


tongyuantongyu:
I also considered this idea at #638, but I found this can be very complicated.

Data of each grid may be reordered to put the most important one at first, so that it can be displayed first. There are also interactions with progressive ability.

I crafted a series of images to show the possible cases: [incremental_grid_examples.zip](https://github.com/AOMediaCodec/libavif/files/7963426/incremental_grid_examples.zip)

Note: In the filename, "CGL" (or alike) represents interleave order from outer loop to inner, where C means color&alpha, G means grid and L means layer, and "rev" means the grid loop goes in reverse order (last cell stored first).

The image with transparency is from here: https://commons.wikimedia.org/wiki/File:PNG_transparency_demonstration_1.png

wantehchang (approved):
LGTM. I didn't review avifincrtest.c, but reviewed everything else carefully.

PR #828, opened 2022-01-26
Do not ignore -Wunused-macros in src/codec_svt.c

wantehchang:
Remove the temporary GCC pragma that ignores -Wunused-macros in
src/codec_svt.c. It was needed only before
commit 74a21444397e323a088fa1f274c9fc942ee7a870.


wantehchang:
@1480c1 Christopher: PTAL. Thanks!

1480c1 (approved):

Issue #830, opened 2022-01-26
Add SVT-AV1 to the CI


wantehchang:
I just found that https://github.com/AOMediaCodec/libavif/blob/master/.github/workflows/ci.yml does not currently build SVT-AV1. We should add SVT-AV1 to our CI.

We should also add SVT-AV1 to tests/docker/build.sh.

@EwoutH I am wondering if you could take a stab at this. Thank you.

EwoutH:
I looked shorty into it, but I'm not that up to speed anymore with the libavif built setup, and unfortunately I'm quite busy the coming weeks.

PR #831, opened 2022-01-27
Add SVT-AV1 to CI and build scripts

wantehchang:
Fix https://github.com/AOMediaCodec/libavif/issues/830.


.github/workflows/ci.yml
```diff
@@ -41,6 +41,10 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: bash rav1e.cmd
+    - name: Build SVT-AV1
+      if: steps.cache-ext.outputs.cache-hit != 'true'
```
wantehchang, line 45: If I leave this line, then it seems that the "`bash svt.cmd`" command at line 47 is not executed.

If I delete this line, then the "`bash svt.cmd`" command is executed, but cmake fails to find NASM. Here is the output log:
https://github.com/AOMediaCodec/libavif/actions/runs/1758324173

The relevant excerpt is:

```
-- Looking for a ASM_NASM compiler
-- Looking for a ASM_NASM compiler - NOTFOUND
-- The ASM_NASM compiler identification is unknown
-- Didn't find assembler
CMake Error at CMakeLists.txt:72 (enable_language):
  No CMAKE_ASM_NASM_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "ASM_NASM" or the CMake cache entry CMAKE_ASM_NASM_COMPILER to the
  full path to the compiler, or to the compiler name if it is in the PATH.
```

wantehchang:
@joedrago @1480c1 I added SVT-AV1 to all the CI and build scripts in this pull request. But I am having a hard time with .github/workflows/ci.yml. Please see my comment in that file.

joedrago:
I believe @EwoutH wrote the cache-related lines. If they don't have time to investigate this issue, perhaps we can remove the cache checking lines in the script entirely and just have slower builds for a while?

1480c1:
If I remember correctly about github actions and caching, it may be better to instead use ccache to cache compilation output by setting CCACHE_DIR to somewhere inside the project build's dir and then cache that instead of caching the whole ext folder and unconditionally build the projects under ext. That could probably be done in another pr and the line for if hit can be removed for now

wantehchang:
@joedrago @1480c1 Please review. Thanks!

1480c1 (approved):

joedrago (approved):

PR #833, opened 2022-01-28
Fix y4mRead() and y4mWrite() for 4:0:0

y-guyon:
Add avify4mtest.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -542,7 +542,7 @@ if(AVIF_BUILD_TESTS)
     # Reference tests for CTest to work.
     enable_testing()
     # Using a CMake FIXTURES_SETUP/FIXTURES_CLEANUP to create and delete a
-    # subdirectory would be cleaner but this is way simpler this way.
+    # subdirectory would be cleaner but this is way simpler.
     set(AVIF_TEST_TMP_DIR ${PROJECT_BINARY_DIR})
 
     add_executable(aviftest
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -539,7 +539,12 @@ endif()
 
 option(AVIF_BUILD_TESTS "Build avif tests." OFF)
 if(AVIF_BUILD_TESTS)
+    # Reference tests for CTest to work.
     enable_testing()
+    # Using a CMake FIXTURES_SETUP/FIXTURES_CLEANUP to create and delete a
+    # subdirectory would be cleaner but this is way simpler this way.
+    set(AVIF_TEST_TMP_DIR ${PROJECT_BINARY_DIR})
+
     add_executable(aviftest
         apps/shared/y4m.c
         tests/aviftest.c
@@ -559,6 +564,18 @@ if(AVIF_BUILD_TESTS)
     target_link_libraries(avifgridapitest avif ${AVIF_PLATFORM_LIBRARIES})
     add_test(NAME avifgridapitest
         COMMAND ${CMAKE_BINARY_DIR}/avifgridapitest)
+
+    add_executable(avify4mtest
+        apps/shared/y4m.c
+        tests/avify4mtest.c
+    )
+    if(AVIF_LOCAL_LIBGAV1)
+        set_target_properties(avify4mtest PROPERTIES LINKER_LANGUAGE "CXX")
+    endif()
+    target_link_libraries(avify4mtest avif ${AVIF_PLATFORM_LIBRARIES})
+    target_include_directories(avify4mtest PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+    add_test(NAME avify4mtest
+        COMMAND ${CMAKE_BINARY_DIR}/avify4mtest ${AVIF_TEST_TMP_DIR})
 
     add_executable(avifyuv
         tests/avifyuv.c
--- a/tests/avify4mtest.c
+++ b/tests/avify4mtest.c
@@ -1,0 +1,198 @@
+// Copyright 2022 Google LLC. All rights reserved.
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "apps/shared/y4m.h"
+#include "avif/avif.h"
+
+#include <assert.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+//------------------------------------------------------------------------------
+
+// Returns true if image1 and image2 are identical.
+static avifBool compareYUVA(const avifImage * image1, const avifImage * image2)
+{
+    if (image1->width != image2->width || image1->height != image2->height || image1->depth != image2->depth ||
+        image1->yuvFormat != image2->yuvFormat || image1->yuvRange != image2->yuvRange) {
+        printf("ERROR: input mismatch\n");
+        return AVIF_FALSE;
+    }
+    assert(image1->width * image1->height > 0);
+
+    avifPixelFormatInfo formatInfo;
+    avifGetPixelFormatInfo(image1->yuvFormat, &formatInfo);
+    const uint32_t uvWidth = (image1->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX;
+    const uint32_t uvHeight = (image1->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY;
+
+    for (uint32_t plane = 0; plane < (formatInfo.monochrome ? 1 : AVIF_PLANE_COUNT_YUV); ++plane) {
+        const uint32_t widthByteCount =
+            ((plane == AVIF_CHAN_Y) ? image1->width : uvWidth) * ((image1->depth > 8) ? sizeof(uint16_t) : sizeof(uint8_t));
+        const uint32_t height = (plane == AVIF_CHAN_Y) ? image1->height : uvHeight;
+        const uint8_t * row1 = image1->yuvPlanes[plane];
+        const uint8_t * row2 = image2->yuvPlanes[plane];
+        for (uint32_t y = 0; y < height; ++y) {
+            if (memcmp(row1, row2, widthByteCount) != 0) {
+                printf("ERROR: different px at row %" PRIu32 ", channel %" PRIu32 "\n", y, plane);
+                return AVIF_FALSE;
+            }
+            row1 += image1->yuvRowBytes[plane];
+            row2 += image2->yuvRowBytes[plane];
+        }
+    }
+
+    if (image1->alphaPlane != NULL || image2->alphaPlane != NULL) {
+        if (image1->alphaPlane == NULL || image2->alphaPlane == NULL || image1->alphaRange != image2->alphaRange ||
+            image1->alphaPremultiplied != image2->alphaPremultiplied) {
+            printf("ERROR: input mismatch\n");
+            return AVIF_FALSE;
+        }
+        const uint32_t widthByteCount = image1->width * ((image1->depth > 8) ? sizeof(uint16_t) : sizeof(uint8_t));
+        const uint8_t * row1 = image1->alphaPlane;
+        const uint8_t * row2 = image2->alphaPlane;
+        for (uint32_t y = 0; y < image1->height; ++y) {
+            if (memcmp(row1, row2, widthByteCount) != 0) {
+                printf("ERROR: different px at row %" PRIu32 ", alpha\n", y);
+                return AVIF_FALSE;
+            }
+            row1 += image1->alphaRowBytes;
+            row2 += image2->alphaRowBytes;
+        }
+    }
+    return AVIF_TRUE;
+}
+
+//------------------------------------------------------------------------------
+
+// Fills each plane of the image with the maximum allowed value.
+static void fillPlanes(avifImage * image)
+{
+    const uint32_t yuvValue = (image->yuvRange == AVIF_RANGE_LIMITED) ? (235 << (image->depth - 8)) : ((1 << image->depth) - 1);
+    avifPixelFormatInfo info;
+    avifGetPixelFormatInfo(image->yuvFormat, &info);
+    for (int plane = 0; plane < (info.monochrome ? 1 : AVIF_PLANE_COUNT_YUV); ++plane) {
+        if (image->yuvPlanes[plane] != NULL) {
+            const uint32_t planeWidth = (plane == AVIF_CHAN_Y) ? image->width
+                                                               : ((image->width + info.chromaShiftX) >> info.chromaShiftX);
+            const uint32_t planeHeight = (plane == AVIF_CHAN_Y) ? image->height
+                                                                : ((image->height + info.chromaShiftY) >> info.chromaShiftY);
+            for (uint32_t y = 0; y < planeHeight; ++y) {
+                uint8_t * const row = image->yuvPlanes[plane] + y * image->yuvRowBytes[plane];
+                if (image->depth == 8) {
+                    memset(row, yuvValue, planeWidth);
+                } else {
+                    for (uint32_t x = 0; x < planeWidth; ++x) {
+                        ((uint16_t *)row)[x] = yuvValue;
+                    }
+                }
+            }
+        }
+    }
+    if (image->alphaPlane != NULL) {
+        assert(image->alphaRange == AVIF_RANGE_FULL);
+        const uint32_t alphaValue = (1 << image->depth) - 1;
+        for (uint32_t y = 0; y < image->height; ++y) {
+            uint8_t * const row = image->alphaPlane + y * image->alphaRowBytes;
+            if (image->depth == 8) {
+                memset(row, alphaValue, image->width);
+            } else {
+                for (uint32_t x = 0; x < image->width; ++x) {
+                    ((uint16_t *)row)[x] = alphaValue;
+                }
+            }
+        }
+    }
+}
+
+// Creates an image and encodes then decodes it as a y4m file.
+static avifBool encodeDecodeY4m(uint32_t width,
+                                uint32_t height,
+                                uint32_t depth,
+                                avifPixelFormat yuvFormat,
+                                avifRange yuvRange,
+                                avifBool createAlpha,
+                                const char filePath[])
+{
+    avifBool success = AVIF_FALSE;
+    avifImage * image = avifImageCreateEmpty();
+    avifImage * decoded = avifImageCreateEmpty();
+    if (image == NULL || decoded == NULL) {
+        printf("ERROR: avifImageCreate() failed\n");
+        goto cleanup;
+    }
+    image->width = width;
+    image->height = height;
+    image->depth = depth;
+    image->yuvFormat = yuvFormat;
+    image->yuvRange = yuvRange;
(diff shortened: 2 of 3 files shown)


tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+        image1->depth != image2->depth || image1->yuvFormat != image2->yuvFormat || image1->yuvRange != image2->yuvRange) {
+        printf("ERROR: input mismatch\n");
+        return AVIF_FALSE;
+    }
+
+    avifPixelFormatInfo formatInfo;
+    avifGetPixelFormatInfo(image1->yuvFormat, &formatInfo);
+    uint32_t uvWidth = (image1->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX;
+    uint32_t uvHeight = (image1->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY;
+    if (uvWidth < 1)
```
jzern, line 29: it seems this would only happen if width/height is 0 which probably should be invalid

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+                printf("ERROR: different px at row %" PRIu32 ", channel %" PRIu32 "\n", y, plane);
+                return AVIF_FALSE;
+            }
+        }
+    }
+
+    if (image1->alphaPlane != NULL) {
+        if (image2->alphaPlane == NULL || image1->alphaRange != image2->alphaRange ||
+            image1->alphaPremultiplied != image2->alphaPremultiplied) {
+            printf("ERROR: input mismatch\n");
```
jzern, line 51: do other tests prefer stdout to stderr?

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+
+//------------------------------------------------------------------------------
+
+// Fills each plane of the image with the maximum allowed value.
+static void fillPlanes(avifImage * image)
+{
+    const uint32_t yuvValue = ((image->yuvRange == AVIF_RANGE_LIMITED) ? 235 : 255) << (image->depth - 8);
+    avifPixelFormatInfo info;
+    avifGetPixelFormatInfo(image->yuvFormat, &info);
+    for (int plane = 0; plane < AVIF_PLANE_COUNT_YUV; ++plane) {
```
jzern, line 73: should this check for monochrome?

CMakeLists.txt
```diff
@@ -539,6 +539,7 @@ endif()
 
 option(AVIF_BUILD_TESTS "Build avif tests." OFF)
 if(AVIF_BUILD_TESTS)
+    enable_testing()
```
wantehchang, line 542: What does enable_testing() do? Why didn't we need this before?

CMakeLists.txt
```diff
@@ -549,6 +550,18 @@ if(AVIF_BUILD_TESTS)
...
+        apps/shared/y4m.c
+        tests/avify4mtest.c
+    )
+    if(AVIF_LOCAL_LIBGAV1)
+        set_target_properties(avify4mtest PROPERTIES LINKER_LANGUAGE "CXX")
+    endif()
+    target_link_libraries(avify4mtest avif ${AVIF_PLATFORM_LIBRARIES})
+    target_include_directories(avify4mtest PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+    add_test(NAME avify4mtest
+        COMMAND ${CMAKE_BINARY_DIR}/avify4mtest)
```
wantehchang, line 563: Why don't the other two tests (aviftest and avifyuv) need an add_test()?

Instead of this add_test(), can we add avify4mtest to the add_custom_target(avif_test_all ...) call at line 573?

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+        uvWidth = 1;
+    if (uvHeight < 1)
+        uvHeight = 1;
+
+    for (uint32_t plane = 0; plane < (formatInfo.monochrome ? 1 : AVIF_PLANE_COUNT_YUV); ++plane) {
+        const uint32_t pixelByteCount =
+            ((plane == AVIF_CHAN_Y) ? image1->width : uvWidth) * ((image1->depth > 8) ? sizeof(uint16_t) : sizeof(uint8_t));
+        const uint32_t height = (plane == AVIF_CHAN_Y) ? image1->height : uvHeight;
+        for (uint32_t y = 0; y < height; ++y) {
+            if (memcmp(image1->yuvPlanes[plane] + y * image1->yuvRowBytes[plane],
```
wantehchang, line 39: Since `image1->yuvRowBytes[plane]` is a uint32_t:

    uint32_t yuvRowBytes[AVIF_PLANE_COUNT_YUV];

the product `y * image1->yuvRowBytes[plane]` can potentially overflow a uint32_t **in general**. In this test, I am sure it won't overflow, but it would be better to eliminate this concern. A common solution is to cast one of the numbers to `size_t`. Alternatively you can declare and increment a row pointer, as is done in y4m.c.

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+
+#include "apps/shared/y4m.h"
+#include "avif/avif.h"
+
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static const char kPath[] = "/tmp/avify4mtest.y4m";
```
wantehchang, line 12: This path may not work on Windows.

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+
+//------------------------------------------------------------------------------
+
+// Returns true if image1 and image2 are identical.
+static avifBool compareYUVA(const avifImage * image1, const avifImage * image2)
+{
+    if (image1 == NULL || image2 == NULL || image1->width != image2->width || image1->height != image2->height ||
```
wantehchang, line 19: Since `compareYUVA()` is an internal function, we can safely assume the callers never pass null pointers to this function, so we don't need to `check image1 == NULL || image2 == NULL`.

tests/avify4mtest.c
wantehchang: Yes, if `uvWidth < 1`, we should return AVIF_FALSE immediately rather than set `uvWidth` to 1 and go on.

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+    avifGetPixelFormatInfo(image1->yuvFormat, &formatInfo);
+    uint32_t uvWidth = (image1->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX;
+    uint32_t uvHeight = (image1->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY;
+    if (uvWidth < 1)
+        uvWidth = 1;
+    if (uvHeight < 1)
+        uvHeight = 1;
+
+    for (uint32_t plane = 0; plane < (formatInfo.monochrome ? 1 : AVIF_PLANE_COUNT_YUV); ++plane) {
+        const uint32_t pixelByteCount =
```
wantehchang, line 35: Nit: I suggest naming this variable `widthByteCount`.

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+            if (memcmp(image1->yuvPlanes[plane] + y * image1->yuvRowBytes[plane],
+                       image2->yuvPlanes[plane] + y * image2->yuvRowBytes[plane],
+                       pixelByteCount) != 0) {
+                printf("ERROR: different px at row %" PRIu32 ", channel %" PRIu32 "\n", y, plane);
+                return AVIF_FALSE;
+            }
+        }
+    }
+
+    if (image1->alphaPlane != NULL) {
```
wantehchang, line 48: We should also check for the case where `image1->alphaPlane == NULL && image2->alphaPlane != NULL`.

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+                } else {
+                    for (uint32_t x = 0; x < planeWidth; ++x) {
+                        ((uint16_t *)row)[x] = yuvValue;
+                    }
+                }
+            }
+        }
+    }
+    if (image->alphaPlane != NULL) {
+        const uint32_t alphaValue = ((image->alphaRange == AVIF_RANGE_LIMITED) ? 235 : 255) << (image->depth - 8);
```
wantehchang, line 92: Limited-range alpha has been removed from the next version of AVIF. So we don't need to test that case.

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+                return AVIF_FALSE;
+            }
+        }
+    }
+    return AVIF_TRUE;
+}
+
+//------------------------------------------------------------------------------
+
+// Fills each plane of the image with the maximum allowed value.
```
wantehchang, line 67: For a full-range pixel, the maximum allowed value is `(1 << image->depth) - 1`, not `255 << (image->depth - 8)`.

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+    if (!compareYUVA(image, decoded)) {
+        goto cleanup;
+    }
+
+    success = AVIF_TRUE;
+cleanup:
+    if (image != NULL) {
+        avifImageDestroy(image);
+    }
+    return success;
```
wantehchang, line 143: We also need to destroy the `decoded` image before returning.

tests/avify4mtest.c
```diff
@@ -0,0 +1,177 @@
...
+}
+
+//------------------------------------------------------------------------------
+
+int main(int argc, char * argv[])
+{
+    (void)argc, (void)argv;
```
wantehchang, line 150: We can declare the main() function as follows:

`int main(void)`

Does the compiler warn if we don't use `argc` and `argv`? I wonder if the compiler makes an exception for the `argc` and `argv` parameters of `main()`.

wantehchang (changes requested):

CMakeLists.txt
y-guyon: See https://github.com/AOMediaCodec/libavif/pull/834#discussion_r800521989

tests/avify4mtest.c
y-guyon: I found both: [`stdout`](https://github.com/AOMediaCodec/libavif/blob/0c11cb19478149b1cc0853288944b1351b9174e6/tests/avifyuv.c#L250) and [`stderr`](https://github.com/AOMediaCodec/libavif/blob/0c11cb19478149b1cc0853288944b1351b9174e6/tests/aviftest.c#L448). The latter seems mostly related to main arguments. Test failures currently use stdout.

tests/avify4mtest.c
y-guyon: Note: I tried to use [`tmpnam()`](https://en.cppreference.com/w/cpp/io/c/tmpnam) for easier IDE automatic setup but got `error: 'tmpnam' is deprecated`. [GoogleTest's `TempDir()`](https://github.com/google/googletest/blob/504eb989534f26033db3edce41a8b3d883cc8e52/googletest/include/gtest/gtest.h#L2211) seems convenient. I fell back to passing the binary dir as an argument to the avify4mtest binary.

tests/avify4mtest.c
y-guyon: I got a compiler error about unused variables, yes. I tried anonymous arguments but this is not valid in C.

tests/avify4mtest.c
```diff
@@ -0,0 +1,196 @@
...
+//------------------------------------------------------------------------------
+
+int main(int argc, char * argv[])
+{
+    if (argc < 2 || !strlen(argv[1])) {
+        fprintf(stderr, "Missing temporary file path argument\n");
+        return EXIT_FAILURE;
+    }
+    char filePath[100];
+    sprintf(filePath, "%s/avify4mtest.y4m", argv[1]);
```
wantehchang, line 169: I looked into this. Among the alternatives suggested by the [tmpnam page](https://en.cppreference.com/w/c/io/tmpnam) on cppreference.com, I like the POSIX function [`tempnam()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/tempnam.html) the best. On Windows we can use [`_tempnam()`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/tempnam-wtempnam-tmpnam-wtmpnam?view=msvc-150).

tests/avify4mtest.c
wantehchang: We can choose a policy and follow it consistently in the tests.

tests/avify4mtest.c
```diff
@@ -0,0 +1,196 @@
...
+//------------------------------------------------------------------------------
+
+// Returns true if image1 and image2 are identical.
+static avifBool compareYUVA(const avifImage * image1, const avifImage * image2)
+{
+    if (image1->width != image2->width || image1->height != image2->height || image1->depth != image2->depth ||
+        image1->yuvFormat != image2->yuvFormat || image1->yuvRange != image2->yuvRange) {
+        printf("ERROR: input mismatch\n");
+        return AVIF_FALSE;
+    } else if (image1->width * image1->height < 1) {
```
wantehchang, line 22: Nit: Don't use `else` if the previous block ends with a return statement.

Also, I believe that `image1->width` and `image1->height` are both nonzero for a successfully decoded image. This property is true for non-grid images because in frame width and height are at least 1 in AV1. So we just need to check the code that decodes grid images to see if it disallows zero output image width or height.

tests/avify4mtest.c
```diff
@@ -0,0 +1,196 @@
...
+        avifImageDestroy(decoded);
+    }
+    return success;
+}
+
+//------------------------------------------------------------------------------
+
+int main(int argc, char * argv[])
+{
+    if (argc < 2 || !strlen(argv[1])) {
```
wantehchang, line 164: Nit: Use a stricter check `argc != 2`.

tests/avify4mtest.c
```diff
@@ -0,0 +1,196 @@
...
+        fprintf(stderr, "Missing temporary file path argument\n");
+        return EXIT_FAILURE;
+    }
+    char filePath[100];
```
wantehchang, line 168: 100 is may not be big enough if `AVIF_TEST_TMP_DIR` is a long pathname. If we use the `tempnam()` function I suggested below, then we don't need to provide the file pathname buffer.

wantehchang (changes requested):
We need to rebase this pull request. I also suggest some changes.

tests/avify4mtest.c
y-guyon: > Don't use `else` if the previous block ends with a return statement.

Sometimes I find it makes sense to put an `else` after a `return` from a code flow point of view (either a switch-like block series or very similar conditions).

> we just need to check the code that decodes grid images to see if it disallows zero output image width or height.

There are two checks in [`avifDecoderDataFillImageGrid()`](https://github.com/AOMediaCodec/libavif/blob/a8cd44119112b55f0433025d57a1ca245ec23d7c/src/read.c#L1285) that would fail if `outputWidth/Height` is `0` or if `(firstTile->image->width * grid->columns) < grid->outputWidth` so the test is not needed. An assertion is enough then.

tests/avify4mtest.c
y-guyon: 1. I would rather avoid non-portable solutions that would introduce platform-dependent defines.
2. Done.

y-guyon:
> We need to rebase this pull request. I also suggest some changes.

Squashed (suggested [here](https://github.com/openedx/edx-platform/wiki/How-to-Rebase-a-Pull-Request#squash-your-changes)) and rebased.

tests/avify4mtest.c
wantehchang: Omitting `else` after a `return` statement is a common recommendation. For example,

https://clang.llvm.org/extra/clang-tidy/checks/readability-else-after-return.html

I understand your argument, so I don't always pick this nit. In this particular case, we have a series of error checks. Therefore the early-exit recommendation is appropriate.

Thank you for the links to the two checks in `avifDecoderDataFillImageGrid()`. Just wanted to make sure I understand it correctly: the first check cannot fail if `outputWidth` or `outputHeight` is 0, right?

```
    if (((firstTile->image->width * grid->columns) < grid->outputWidth) ||
        ((firstTile->image->height * grid->rows) < grid->outputHeight)) {
        avifDiagnosticsPrintf(data->diag,
                              "Grid image tiles do not completely cover the image (HEIF (ISO/IEC 23008-12:2017), Section 6.6.2.3.1)");
        return AVIF_FALSE;
    }
```

On the other hand, I agree the second check will fail if `outputWidth` or `outputHeight` is 0:

```
    if (((firstTile->image->width * (grid->columns - 1)) >= grid->outputWidth) ||
        ((firstTile->image->height * (grid->rows - 1)) >= grid->outputHeight)) {
        avifDiagnosticsPrintf(data->diag,
                              "Grid image tiles in the rightmost column and bottommost row do not overlap the reconstructed image grid canvas. See MIAF (ISO/IEC 23000-22:2019), Section 7.3.11.4.2, Figure 2");
        return AVIF_FALSE;
    }
```

Note: I found the check I was referring to. It is the following code in the `avifParseImageGridBox()` function:

```
    if ((grid->outputWidth == 0) || (grid->outputHeight == 0)) {
        avifDiagnosticsPrintf(diag, "Grid box contains illegal dimensions: [%u x %u]", grid->outputWidth, grid->outputHeight);
        return AVIF_FALSE;
    }
```

tests/avify4mtest.c
wantehchang: Using `tempnam()` allows us to run the same avify4mtest binary from multiple terminal windows at the same time. This is admittedly an uncommon use case. This kind of simple portability issue (where `tempname()` and `_tempname()` differ only in the function name) can be handled by using a macro or adding a wrapper function.

CMakeLists.txt
```diff
@@ -539,7 +539,12 @@ endif()
 
 option(AVIF_BUILD_TESTS "Build avif tests." OFF)
 if(AVIF_BUILD_TESTS)
+    # Reference tests for CTest to work.
```
wantehchang, line 542: Nit: What does "Reference tests" mean?

Since CTest is not as well known as CMake, it may be helpful to be a little verbose and say this allows us to use add_test() to add a test to be run by ctest, and provide the ctest command line we should use.

wantehchang (approved):
LGTM. Thanks. I suggest a comment change, which you can address in a follow-up pull request.

tests/avify4mtest.c
```diff
@@ -0,0 +1,198 @@
+// Copyright 2022 Google LLC. All rights reserved.
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "apps/shared/y4m.h"
+#include "avif/avif.h"
```
wantehchang, line 5: Please change these two lines to:

```
#include "avif/avif.h"

#include "y4m.h"
```

This is how apps/avifdec.c and apps/avifenc.c include y4m.h.

CMakeLists.txt
y-guyon: Done in #846

tests/avify4mtest.c
y-guyon: Done in #846

PR #834, opened 2022-01-28
Move checks to avifAreGridDimensionsValid()

y-guyon:
It was possible to encode a 65x65 4:2:0 image with avifenc that could
not be decoded with avifdec. Calling avifAreGridDimensionsValid() from
both encoding and decoding pipelines removes this asymmetric behavior.


jzern (commented):
Is there any test coverage to ensure the encode/decode roundtrip in the commit message?

src/avif.c
```diff
@@ -721,6 +721,51 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+        avifDiagnosticsPrintf(diag,
+                              "[Strict] Grid image tiles are smaller than 64x64 (%ux%u). See MIAF (ISO/IEC 23000-22:2019), Section 7.3.11.4.2",
+                              tileW,
+                              tileH);
+        return AVIF_FALSE;
+    }
+
+    // ISO/IEC 23000-22:2019, Section 7.3.11.4.2:
+    //   - when the images are in the 4:2:2 chroma sampling format the horizontal tile offsets and widths, and the output width, shall be even numbers;
+    //   - when the images are in the 4:2:0 chroma sampling format both the horizontal and vertical tile offsets and widths, and the output width and height, shall be even numbers.
```
wantehchang, line 747: I seem to remember that I analyzed these requirements and concluded that the requirements on imageW and imageH (i.e., output width and height) were overly strict. Is this the problem you reported in the team meeting? If so, we can choose to not enforce the requirements on imageW and imageH and document our decision in the comment.

Note: In fact, it seems that even the requirements on tileW and tileH can be relaxed when the image grid has only one row or one column. For example, if the image grid has only one row, then it seems that tileH can be an odd number when chroma is subsampled vertically.

include/avif/avif.h
```diff
@@ -411,6 +411,16 @@ AVIF_API avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * cla
                                                       avifPixelFormat yuvFormat,
                                                       avifDiagnostics * diag);
 
+// ---------------------------------------------------------------------------
+
+// Returns false if tiles of tileW*tileH pixels each in an image of imageW*imageH pixels violate any standards.
```
wantehchang, line 416: Nit: I suggest adding parentheses to make this sentence easier to parse. Also add the words "the" and "grid".

`// Returns false if the tiles (of tileW*tileH pixels each) in a grid image of imageW*imageH pixels violate any standards.`

Or split the comment into three sentences:

`// Returns false if the tiles in a grid image violate any standards. The tiles are of tileW*tileH pixels each. The grid image is of imageW*imageH pixels.`

src/read.c
```diff
@@ -1295,38 +1295,15 @@ static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
...
-    // Check the restrictions in MIAF (ISO/IEC 23000-22:2019), Section 7.3.11.4.2.
-    //
-    // The tile_width shall be greater than or equal to 64, and the tile_height shall be greater than or equal to 64.
-    if ((firstTile->image->width < 64) || (firstTile->image->height < 64)) {
-        avifDiagnosticsPrintf(data->diag,
-                              "Grid image tiles are smaller than 64x64 (%ux%u). See MIAF (ISO/IEC 23000-22:2019), Section 7.3.11.4.2",
-                              firstTile->image->width,
-                              firstTile->image->height);
+
+    if (!avifAreGridDimensionsValid(alpha ? AVIF_PIXEL_FORMAT_NONE : firstTile->image->yuvFormat,
```
wantehchang, line 1299: Use AVIF_PIXEL_FORMAT_YUV400 for alpha. AVIF_PIXEL_FORMAT_NONE is often used as an invalid value.

wantehchang, line 1299: Use AVIF_PIXEL_FORMAT_YUV400 for alpha. AVIF_PIXEL_FORMAT_NONE is often used as an invalid value.

Note: I believe that if `alpha` is true, then `firstTile->image->yuvFormat` should be equal to AVIF_PIXEL_FORMAT_YUV400.

src/write.c
```diff
@@ -585,6 +586,12 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
         if (cellImage->yuvFormat == AVIF_PIXEL_FORMAT_NONE) {
             return AVIF_RESULT_NO_YUV_FORMAT_SELECTED;
         }
+
+        // MIAF (ISO 23000-22:2019), Section 7.3.11.4.1:
+        //   All input image of a grid image item shall use the same chroma sampling format.
```
wantehchang, line 591: Nit: "image" should be plural "images".

Since this comment is written as a quote, we should quote the exact text from MIAF:

`All input images of a grid image item shall use the same coding format, chroma sampling format, ...`

src/write.c
```diff
@@ -585,6 +586,12 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
+        if (cellImage->yuvFormat != firstCell->yuvFormat) {
+            return AVIF_RESULT_INVALID_IMAGE_GRID;
+        }
```
wantehchang, line 594: It seems better to merge this check into the existing if statement at lines 576-580.

Also, I wonder if we should match the check performed in src/read.c:

```
        if ((tile->image->width != firstTile->image->width) || (tile->image->height != firstTile->image->height) ||
            (tile->image->depth != firstTile->image->depth) || (tile->image->yuvFormat != firstTile->image->yuvFormat) ||
            (tile->image->yuvRange != firstTile->image->yuvRange) || (uvPresent != firstTileUVPresent) ||
            (tile->image->colorPrimaries != firstTile->image->colorPrimaries) ||
            (tile->image->transferCharacteristics != firstTile->image->transferCharacteristics) ||
            (tile->image->matrixCoefficients != firstTile->image->matrixCoefficients) ||
            (tile->image->alphaRange != firstTile->image->alphaRange)) {
            avifDiagnosticsPrintf(data->diag, "Grid image contains mismatched tiles");
            return AVIF_FALSE;
        }
```

src/write.c
```diff
@@ -562,11 +562,12 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
         return AVIF_RESULT_UNSUPPORTED_DEPTH;
     }
 
-    if (!firstCell->width || !firstCell->height) {
-        return AVIF_RESULT_NO_CONTENT;
-    }
-
-    if ((cellCount > 1) && ((firstCell->width < 64) || (firstCell->height < 64))) {
+    if ((cellCount > 1) && !avifAreGridDimensionsValid(firstCell->yuvFormat,
```
wantehchang, line 565: IMPORTANT: This change loses the `if (!firstCell->width || !firstCell->height)` check in the original code when `cellCount` is 1.

Perhaps we should move the `if ((imageW == 0) || (imageH == 0)` check out of `avifAreGridDimensionsValid()` and resurrect the `if (!firstCell->width || !firstCell->height)` check here.

wantehchang (changes requested):

src/write.c
```diff
@@ -567,15 +567,23 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
         return AVIF_RESULT_NO_CONTENT;
     }
 
-    if ((cellCount > 1) && ((firstCell->width < 64) || (firstCell->height < 64))) {
+    if (!avifAreGridDimensionsValid(firstCell->yuvFormat, gridCols, gridRows, firstCell->width, firstCell->height, &encoder->diag)) {
         return AVIF_RESULT_INVALID_IMAGE_GRID;
     }
 
+    avifBool firstTileUVPresent = (firstCell->yuvPlanes[AVIF_CHAN_U] && firstCell->yuvPlanes[AVIF_CHAN_V]);
```
wantehchang, line 574: It is not necessary to check `uvPresent != firstTileUVPresent`, because this check is implied by the `cellImage->yuvFormat != firstCell->yuvFormat` check.

Note that `uvPresent` is true if and only if `cellImage->yuvFormat != AVIF_PIXEL_FORMAT_YUV400`.

tests/avifgridapitest.c
```diff
@@ -0,0 +1,234 @@
...
+    avifPixelFormatInfo formatInfo;
+    avifGetPixelFormatInfo((*image)->yuvFormat, &formatInfo);
+    uint32_t uvWidth = ((*image)->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX;
+    uint32_t uvHeight = ((*image)->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY;
+    if (uvWidth < 1) {
+        uvWidth = 1;
+    }
+    if (uvHeight < 1) {
+        uvHeight = 1;
+    }
```
wantehchang, line 36: I think we should remove these two checks (lines 31-36).

Note that `uvWidth` can only be 0 if `width` is 0.

tests/avifgridapitest.c
```diff
@@ -0,0 +1,234 @@
...
+
+// Generates then encodes a grid image. Returns false in case of failure.
+static avifBool encodeGrid(int columns, int rows, int cellWidth, int cellHeight, int depth, avifPixelFormat yuvFormat, avifRWData * output)
+{
+    avifBool success = AVIF_FALSE;
+    avifEncoder * encoder = NULL;
+    avifImage ** cellImages = avifAlloc(sizeof(avifImage *) * columns * rows);
+    memset(cellImages, 0, sizeof(avifImage *) * columns * rows);
+    for (int row = 0, iCell = 0; row < rows; ++row) {
+        for (int col = 0; col < columns; ++col, ++iCell) {
```
wantehchang, line 58: Nit: We can collapse the two for loops into one:

```
    for (int iCell = 0; iCell < columns * rows; ++iCell) {
```

Note that you do it this way under the `cleanup` label, at line 87.

tests/avifgridapitest.c
```diff
@@ -0,0 +1,234 @@
...
+    success = AVIF_TRUE;
+cleanup:
+    avifRWDataFree(&encodedAvif);
+    return success;
+}
+
+//------------------------------------------------------------------------------
+
+// For each dimension, for each combination of cell count and size, generates, encodes then decodes a grid image.
+static avifBool encodeDecodeSizes(const int (*columnsCellWidths)[2],
```
wantehchang, line 146: Nit/optional: The following syntax may be easier to understand:

const int columnsCellWidths[][2]

wantehchang (changes requested):

src/write.c
y-guyon: I copy-pasted the line from Draft ISO/IEC 23000-22(E). The typo was likely fixed in a later version.

src/avif.c
y-guyon: The requirements on tileW and tileH are now relaxed when the image grid has only one row or one column. See the comments I added for details on why it cannot easily be relaxed further.

CMakeLists.txt
```diff
@@ -548,6 +549,20 @@ if(AVIF_BUILD_TESTS)
     endif()
     target_link_libraries(aviftest avif ${AVIF_PLATFORM_LIBRARIES})
     target_include_directories(aviftest PRIVATE apps/shared)
+    add_custom_target(avif_test_all
+        COMMAND $<TARGET_FILE:aviftest> ${CMAKE_CURRENT_SOURCE_DIR}/tests/data
+        DEPENDS aviftest
+    )
```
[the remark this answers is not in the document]
y-guyon: 1. I added `enable_testing()` so that the `ctest` command will run all targets referenced by `add_test()`. \
I moved the `add_custom_target(avif_test_all ...)` call closer to its only dependency `aviftest`. \
These two changes are not required but `ctest` is convenient and the move seemed clearer to me.
2. I do not mind adding `avifgridapitest` and `avifyuv` to the `avif_test_all` custom target but this seems beyond the scope of this change. Is `avif_test_all` used by the CI of this GitHub repo?

CMakeLists.txt
```diff
@@ -548,6 +549,20 @@ if(AVIF_BUILD_TESTS)
...
+
+    add_executable(avifgridapitest
+            tests/avifgridapitest.c
+    )
+    if(AVIF_LOCAL_LIBGAV1)
+        set_target_properties(avifgridapitest PROPERTIES LINKER_LANGUAGE "CXX")
+    endif()
+    target_link_libraries(avifgridapitest avif ${AVIF_PLATFORM_LIBRARIES})
+    add_test(NAME avifgridapitest
```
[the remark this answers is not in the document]
y-guyon: Yes but probably in another change. I would also suggest replacing `avif_test_all` by CMake tests, but it depends how and where `avif_test_all` is currently used.

include/avif/avif.h
```diff
@@ -411,6 +411,17 @@ AVIF_API avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * cla
                                                       avifPixelFormat yuvFormat,
                                                       avifDiagnostics * diag);
 
+// ---------------------------------------------------------------------------
+
+// Returns false if the tiles in a grid image violate any standards.
+// There are columns*rows tiles. The tiles are of tileW*tileH pixels each.
+AVIF_API avifBool avifAreGridDimensionsValid(avifPixelFormat yuvFormat,
```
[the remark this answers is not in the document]
y-guyon: `avifAreGridDimensionsValid()` is similar to `avifCropRectConvertCleanApertureBox()`, that is why I put it nearby. I do not mind moving it to internal.h.

CMakeLists.txt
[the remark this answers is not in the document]
y-guyon: I put `add_custom_target()` back to where it was.

y-guyon:
> Is there any test coverage to ensure the encode/decode roundtrip in the commit message?

Done.

src/avif.c
```diff
@@ -721,6 +721,54 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+    //   - when the images are in the 4:2:2 chroma sampling format the horizontal tile offsets and widths,
+    //     and the output width, shall be even numbers;
+    //   - when the images are in the 4:2:0 chroma sampling format both the horizontal and vertical tile
+    //     offsets and widths, and the output width and height, shall be even numbers.
+    // Only the tile dimensions are enforced since tile offsets and image dimensions depend on them.
+    // If the rules above were not respected, the following problematic situation may happen:
+    //   Some 4:2:0 image is Pi pixels wide and has N cell columns.
+    //   Each cell is Pc pixels wide, Pc being odd.
+    //   The chroma plane of the whole image is Pi/2 or (Pi+1)/2 pixels wide, depending on the rounding.
+    //   The chroma plane of the summed cells is (Pc/2)*N or ((Pc+1)/2)*N pixels wide, depending on the rounding.
```
wantehchang, line 757: Given the assumption that `Pc` is odd, the chroma plane of the summed cells is `((Pc+1)/2)*N` pixels wide. So this sentence should read:

`The chroma plane of the summed cells is ((Pc+1)/2)*N pixels wide.`

Assuming `Pi = Pc*N`, we have `((Pc+1)/2)*N = (Pc*N+N)/2 = (Pi+N)/2`.

src/avif.c
```diff
@@ -721,6 +721,54 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+
+    // ISO/IEC 23000-22:2019, Section 7.3.11.4.2:
+    //   - when the images are in the 4:2:2 chroma sampling format the horizontal tile offsets and widths,
+    //     and the output width, shall be even numbers;
+    //   - when the images are in the 4:2:0 chroma sampling format both the horizontal and vertical tile
+    //     offsets and widths, and the output width and height, shall be even numbers.
+    // Only the tile dimensions are enforced since tile offsets and image dimensions depend on them.
+    // If the rules above were not respected, the following problematic situation may happen:
+    //   Some 4:2:0 image is Pi pixels wide and has N cell columns.
+    //   Each cell is Pc pixels wide, Pc being odd.
```
wantehchang, line 755: I believe you also assume `Pi = Pc*N`, right?

Note that in general this assumption is not true. Please see Figure 2 on page 18 of MIAF. It shows that in general grid width doesn't need to be equal to output width, and grid height doesn't need to be equal to output height.

src/avif.c
```diff
@@ -721,6 +721,54 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+    //   The chroma plane of the whole image is Pi/2 or (Pi+1)/2 pixels wide, depending on the rounding.
+    //   The chroma plane of the summed cells is (Pc/2)*N or ((Pc+1)/2)*N pixels wide, depending on the rounding.
+    //   This leads to either N missing pixels or N extra pixels, compared to the whole image.
```
wantehchang, line 758: If I understand your example correctly, this sentence should read

`This leads to either N/2 or (N-1)/2 extra pixels, compared to the whole image.`

because we are comparing` (Pi+N)/2` with `Pi/2` or `(Pi+1)/2`.

src/avif.c
```diff
@@ -721,6 +721,54 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+    if ((((yuvFormat == AVIF_PIXEL_FORMAT_YUV420) || (yuvFormat == AVIF_PIXEL_FORMAT_YUV422)) && ((tileW % 2) != 0)) ||
+        ((yuvFormat == AVIF_PIXEL_FORMAT_YUV420) && ((tileH % 2) != 0))) {
+        avifDiagnosticsPrintf(diag,
+                              "[Strict] Grid image tile width (%u) or height (%u) shall be even "
```
wantehchang, line 762: Remove "[Strict]" from this error message.

src/avif.c
```diff
@@ -721,6 +721,54 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+                              "if there is more than one cell and if chroma is subsampled in that dimension. "
```
wantehchang, line 763: Remove "if there is more than one cell and" from this error message.

src/write.c
```diff
@@ -567,15 +567,21 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
         return AVIF_RESULT_NO_CONTENT;
     }
 
-    if ((cellCount > 1) && ((firstCell->width < 64) || (firstCell->height < 64))) {
+    if (!avifAreGridDimensionsValid(firstCell->yuvFormat, gridCols, gridRows, firstCell->width, firstCell->height, &encoder->diag)) {
```
wantehchang, line 570: Add `(cellCount > 1) &&` to the conditional expression.

src/avif.c
```diff
@@ -721,6 +721,54 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+        return AVIF_FALSE;
+    }
+
+    // ISO/IEC 23000-22:2019, Section 7.3.11.4.2:
+    //   - when the images are in the 4:2:2 chroma sampling format the horizontal tile offsets and widths,
+    //     and the output width, shall be even numbers;
+    //   - when the images are in the 4:2:0 chroma sampling format both the horizontal and vertical tile
+    //     offsets and widths, and the output width and height, shall be even numbers.
+    // Only the tile dimensions are enforced since tile offsets and image dimensions depend on them.
+    // If the rules above were not respected, the following problematic situation may happen:
```
wantehchang, line 753: In the next three comments I described the problems I found while working through your example here. This at least shows the example is a little complicated. I think we can make the example simpler by replacing the variables `Pi`, `N`, and `Pc` with concrete numbers, such as 130 (= Pi), 2 (= N), and 65 (= Pc).

wantehchang (changes requested):

src/avif.c
```diff
@@ -721,6 +721,54 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+        return AVIF_TRUE;
+    }
+
+    // ISO/IEC 23000-22:2019, Section 7.3.11.4.2:
+    //   - the tile_width shall be greater than or equal to 64, and should be a multiple of 64
+    //   - the tile_height shall be greater than or equal to 64, and should be a multiple of 64
+    // The "should" part is ignored here.
+    if ((tileW < 64) || (tileH < 64)) {
+        avifDiagnosticsPrintf(diag,
+                              "[Strict] Grid image tile width (%u) or height (%u) cannot be smaller "
```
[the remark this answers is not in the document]
y-guyon: Is this the case for the "[Strict]" in [avifCropRectIsValid()](https://github.com/AOMediaCodec/libavif/blob/d9cffc5f46b62aeff46eebf51449726386d6c485/src/avif.c#L531)? I do not see how this specific check can be turned off, except not using `avifenc --crop`.

src/avif.c
```diff
@@ -721,6 +721,54 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+                              "than 64 if there is more than one cell. "
```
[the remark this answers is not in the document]
y-guyon: I will remove it but why? It is more accurate.

src/avif.c
```diff
@@ -721,6 +721,54 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+                              tileH);
+        return AVIF_FALSE;
+    }
+
+    // ISO/IEC 23000-22:2019, Section 7.3.11.4.2:
+    //   - when the images are in the 4:2:2 chroma sampling format the horizontal tile offsets and widths,
+    //     and the output width, shall be even numbers;
+    //   - when the images are in the 4:2:0 chroma sampling format both the horizontal and vertical tile
+    //     offsets and widths, and the output width and height, shall be even numbers.
+    // Only the tile dimensions are enforced since tile offsets and image dimensions depend on them.
```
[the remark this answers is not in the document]
y-guyon: Currently in libavif, at encoding, the output width and height do depend on the tile dimensions. However you are right that it is not the case at decoding. Let's keep the same decoding constraints as before in this change and just sync encoding constraints.

src/avif.c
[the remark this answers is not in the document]
wantehchang: I looked into this. I found that the policy on "[Strict]" cannot be followed perfectly because right now only `avifDecoder` has a `strictFlags` field. So I will only show how the policy on "[Strict]" is followed by `avifCropRectIsValid()` on the decoder side.

`avifCropRectIsValid()` is called by `avifCropRectConvertCleanApertureBox()`. Note that all the diagnostic messages in `avifCropRectIsValid()` and `avifCropRectConvertCleanApertureBox()` have the "[Strict]" prefix.

The decoder calls `avifCropRectConvertCleanApertureBox()` as follows:

```
    if (strictFlags & AVIF_STRICT_CLAP_VALID) {
        const avifProperty * clapProp = avifPropertyArrayFind(&item->properties, "clap");
        if (clapProp) {
            const avifProperty * ispeProp = avifPropertyArrayFind(&item->properties, "ispe");
            if (!ispeProp) {
                avifDiagnosticsPrintf(diag,
                                      "[Strict] Item ID %u is missing an ispe property, so its clap property cannot be validated",
                                      item->id);
            }

            avifCropRect cropRect;
            const uint32_t imageW = ispeProp->u.ispe.width;
            const uint32_t imageH = ispeProp->u.ispe.height;
            const avifPixelFormat av1CFormat = avifCodecConfigurationBoxGetFormat(&av1CProp->u.av1C);
            avifBool validClap = avifCropRectConvertCleanApertureBox(&cropRect, &clapProp->u.clap, imageW, imageH, av1CFormat, diag);
            if (!validClap) {
            }
        }
    }
```

Note that it is inside an if statement that checks `strictFlags & AVIF_STRICT_CLAP_VALID`.

src/avif.c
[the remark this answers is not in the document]
wantehchang: The reason is that the libavif decoder calls avifAreGridDimensionsValid() on all image grids, even when the image grid has only one cell. The libavif encoder won't encode an image grid if there is only one cell, but other AVIF encoders may do that because it is allowed by the image format.

src/write.c
```diff
@@ -567,15 +567,26 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
+                                                       &encoder->diag)) {
         return AVIF_RESULT_INVALID_IMAGE_GRID;
     }
 
     for (uint32_t cellIndex = 0; cellIndex < cellCount; ++cellIndex) {
         const avifImage * cellImage = cellImages[cellIndex];
-        if ((cellImage->depth != firstCell->depth) || (cellImage->width != firstCell->width) ||
-            (cellImage->height != firstCell->height) || (!!cellImage->alphaPlane != !!firstCell->alphaPlane) ||
-            (cellImage->alphaPremultiplied != firstCell->alphaPremultiplied)) {
+        // MIAF (ISO 23000-22:2019), Section 7.3.11.4.1:
```
wantehchang, line 581: Please also cite the following requirement in HEIF Section 6.6.2.3.1:

> All input images shall have exactly the same width and height; call those `tile_width` and `tile_height`.

This requirement is enforced by the check `(cellImage->width != firstCell->width) || (cellImage->height != firstCell->height)`.

In our video call today, you mentioned that the right-most column or the bottom-most row can have a smaller tile width or tile height. Are you referring to this requirement in HEIF Section 6.6.2.3.1?

> The reconstructed image is formed by tiling the input images into a grid with a column width (potentially excluding the right-most column) equal to `tile_width` and a row height (potentially excluding the bottom-most row) equal to `tile_height`, without gap or overlap, and then trimming on the right and the bottom to the indicated `output_width` and `output_height`.

The "potentially excluding the right-most column" and "potentially excluding the bottom-most row" parts in this requirement seem to allow the right-most column to be narrower than `tile_width` and the bottom-most row to be shorter than `tile_height`. But it is not written clearly.

src/avif.c
```diff
@@ -721,6 +721,51 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+
+    // ISO/IEC 23000-22:2019, Section 7.3.11.4.2:
+    //   - when the images are in the 4:2:2 chroma sampling format the horizontal tile offsets and widths,
+    //     and the output width, shall be even numbers;
+    //   - when the images are in the 4:2:0 chroma sampling format both the horizontal and vertical tile
+    //     offsets and widths, and the output width and height, shall be even numbers.
+    // If the rules above were not respected, the following problematic situation may happen:
+    //   Some 4:2:0 image is 650 pixels wide and has 10 cell columns, each being 65 pixels wide.
+    //   The chroma plane of the whole image is 325 pixels wide. The chroma plane of each cell is 33 pixels wide.
+    //   33*10 - 325 gives 5 extra pixels from cells with no specified destination in the reconstructed image.
```
wantehchang, line 749: Nit: I am not sure if I understand what "from cells" means. Maybe rephrase it or omit it?

wantehchang (approved):
LGTM. Thanks!

y-guyon:
> LGTM. Thanks!

Thank you for the thorough review and for helping me navigate the specification maze.

src/avif.c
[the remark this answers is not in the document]
y-guyon: The MIAF issue was filed: [Consider relaxing forbidden odd dimensions of subsampled chroma grid images](https://github.com/MPEGGroup/MIAF/issues/11)

wantehchang (approved):
LGTM. I made a minor change to avifgridapitest.c: change 0/1 to `AVIF_FALSE`/`AVIF_TRUE` in the `encodeDecodeSizes()` function.

PR #835, opened 2022-01-29
Increment lastItemID first when creating an item

wantehchang:
Fix a bug introduced in commit f732a4d240d1562151685b71e822d0ba3a352ad9.

In avifEncoderDataCreateItem(), increment data->lastItemID before
setting item->id equal to data->lastItemID. This is what the original
code did.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -151,6 +151,7 @@ error:
 static avifEncoderItem * avifEncoderDataCreateItem(avifEncoderData * data, const char * type, const char * infeName, size_t infeNameSize, uint32_t cellIndex)
 {
     avifEncoderItem * item = (avifEncoderItem *)avifArrayPushPtr(&data->items);
+    ++data->lastItemID;
     item->id = data->lastItemID;
     memcpy(item->type, type, sizeof(item->type));
     item->infeName = infeName;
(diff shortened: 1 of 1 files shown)


wantehchang:
@vigneshvg @y-guyon @joedrago Please review. This is a critical bug; this broke the encoder. I am very sorry.

wantehchang:
This bug was just reported in https://github.com/AOMediaCodec/libavif/issues/836. Because of the criticality of this bug, I am going to merge this pull request now.

src/write.c
```diff
@@ -160,11 +161,11 @@ static avifEncoderItem * avifEncoderDataCreateItem(avifEncoderData * data, const
     if (!avifArrayCreate(&item->mdatFixups, sizeof(avifOffsetFixup), 4)) {
         goto error;
     }
-    ++data->lastItemID;
     return item;
 
 error:
     avifCodecEncodeOutputDestroy(item->encodeOutput);
+    --data->lastItemID;
```
y-guyon, line 168: This is fine. Another option to avoid the decrement in the error path would be `item->id = data->lastItemID + 1;` at line 155.

y-guyon (commented):
LGTM

y-guyon:
> This is a critical bug; this broke the encoder.

Any idea on how to include a regression test about that specific issue into the GitHub CI?

wantehchang:
> Any idea on how to include a regression test about that specific issue into the GitHub CI?

Yannis: I believe the GitHub CI does not call use the avifEncoder at all. The encode-decode tests in tests/aviftest.c were removed in https://github.com/AOMediaCodec/libavif/pull/811 because the way they checked the decoded images were close to the original images was too sensitive to changes to the AV1 encoders.

Rather than add a regression test for this bug, we can add the encode-decode tests back and just remove the compareYUVA() call from the testCaseRun() function, similar to what I did in Google internal cl/423410661. That will allow us to detect this bug.

PR #839, opened 2022-02-08
Document the "[Strict]" prefix in error strings

wantehchang:


wantehchang:
@joedrago @y-guyon Please review. Thanks.

joedrago (approved):

PR #840, opened 2022-02-09
Only consider a frame index to be a keyframe if all tiles/planes are sync frames

joedrago:
Related: https://crbug.com/1274200


wantehchang (approved):
LGTM. Thanks a lot!

PR #842, opened 2022-02-10
avifArrayPop() should zero the popped element

wantehchang:
avifArrayPop() should zero the memor of the popped element. If we ever
push an element to the array, the memory of the new element will be all
zeros.

Also make sure we cast arr->elementSize to size_t when we multiply it by
another uint32_t variable.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -3,6 +3,7 @@
 
 #include "avif/internal.h"
 
+#include <assert.h>
 #include <math.h>
 #include <string.h>
 
@@ -128,6 +129,7 @@ void avifArrayPush(void * arrayStruct, void * element)
 void avifArrayPop(void * arrayStruct)
 {
     avifArrayInternal * arr = (avifArrayInternal *)arrayStruct;
+    assert(arr->count > 0);
     --arr->count;
     memset(&arr->ptr[arr->count * (size_t)arr->elementSize], 0, arr->elementSize);
 }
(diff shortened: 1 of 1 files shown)


joedrago:
This change seems fine, but I have a question about this function in general (it looks like it was added when the array OOM changes were put in): Should it protect / assert / check or assert that `arr->count > 0`? The 3 usages of it seem to make it impossible otherwise, but it might be worth an assert() here.

src/utils.c
```diff
@@ -129,6 +129,7 @@ void avifArrayPop(void * arrayStruct)
 {
     avifArrayInternal * arr = (avifArrayInternal *)arrayStruct;
     --arr->count;
```
vigneshvg, line 133: should we bail if count == 0 when this function is called?

if we are sure that could will always be > 0, then maybe add an assert?

src/utils.c
```diff
@@ -115,7 +115,7 @@ void * avifArrayPushPtr(void * arrayStruct)
 {
     uint32_t index = avifArrayPushIndex(arrayStruct);
     avifArrayInternal * arr = (avifArrayInternal *)arrayStruct;
-    return &arr->ptr[index * arr->elementSize];
+    return &arr->ptr[index * (size_t)arr->elementSize];
```
vigneshvg, line 119: optional: make the index computation into a function or macro so that the cast doesn't have to be repeated.

wantehchang:
Yes, I added `avifArrayPop()` when the array OOM changes were put in. I added an assertion for `arr->count > 0`.

It may be good to call `abort()` in release builds, otherwise the `memset()` call will corrupt memory.

src/utils.c
wantehchang: It is a programming error to call `avifArrayPop()` on an empty array. So I don't want to silently bail out if arr->count == 0. We can call `abort()` or print an error message.

PR #843, opened 2022-02-10
avifDecoderReset() should not return AVIF_FALSE

wantehchang:
The return type of avifDecoderReset() is avifResult, so it should not
return AVIF_FALSE on failure, because AVIF_FALSE has the same value (0)
as AVIF_RESULT_OK.


vigneshvg:
lgtm.

PR #844, opened 2022-02-14
Handle avifDecoderDataCreateTile() failures

wantehchang:
Allow avifDecoderDataCreateTile() to fail on avifImageCreateEmpty() or
avifCodecDecodeInputCreate() failures. Note that we don't check for
avifArrayPushPtr() failure yet. That will be done later.

Change all avifDecoderDataCreateTile() callers to handle its failures.


vigneshvg:
lgtm.

y-guyon:
LGTM

PR #845, opened 2022-02-14
Document that SVT-AV1 doesn't support lossless yet

wantehchang:


wantehchang:
@hassount @1480c1 Just FYI.

vigneshvg:
lgtm.

PR #846, opened 2022-02-15
Match include style in avify4mtest and avifenc

y-guyon:
Also describe ctest usage in CMakeLists.txt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -539,7 +539,9 @@ endif()
 
 option(AVIF_BUILD_TESTS "Build avif tests." OFF)
 if(AVIF_BUILD_TESTS)
-    # Reference tests for CTest to work.
+    # With testing enabled, all targets referenced by add_test() can be run with
+    # CMake's ctest command line tool from the build folder: ctest [options]
+    # Options include filtering tests with a regexp -R, multithreading -j etc.
     enable_testing()
     # Using a CMake FIXTURES_SETUP/FIXTURES_CLEANUP to create and delete a
     # subdirectory would be cleaner but this is way simpler.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -539,7 +539,9 @@ endif()
 
 option(AVIF_BUILD_TESTS "Build avif tests." OFF)
 if(AVIF_BUILD_TESTS)
-    # Reference tests for CTest to work.
+    # With testing enabled, all targets referenced by add_test() can be run with
+    # CMake's ctest command line tool from the build folder: ctest [options]
+    # Options include filtering tests with a regexp -R, multithreading -j etc.
```
wantehchang, line 544: Thank you! I didn't realize that we can just say "ctest" without any options. Given that, we can omit everything after "from the build folder" if you prefer a concise comment.

CMakeLists.txt
```diff
@@ -573,7 +575,8 @@ if(AVIF_BUILD_TESTS)
         set_target_properties(avify4mtest PROPERTIES LINKER_LANGUAGE "CXX")
     endif()
     target_link_libraries(avify4mtest avif ${AVIF_PLATFORM_LIBRARIES})
-    target_include_directories(avify4mtest PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+    target_include_directories(avify4mtest
+        PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/apps/shared)
```
wantehchang, line 579: Nit: `${CMAKE_CURRENT_SOURCE_DIR}/` can be omitted. Please compare with line 559. We should use the same approach at both places.

I am fine with either approach, but slightly prefer the more concise approach unless it is broken in some build configurations.

wantehchang (changes requested):
LGTM. I have a suggested change.

CMakeLists.txt
y-guyon: Indeed, from [CMake doc](https://cmake.org/cmake/help/latest/command/target_include_directories.html):

> A relative path will be interpreted as relative to the current source directory (i.e. CMAKE_CURRENT_SOURCE_DIR)

wantehchang (approved):
LGTM. Thanks!

PR #848, opened 2022-02-16
Support parsing of version 3 of ItemInfoEntry

wantehchang:
Version 2 and version 3 of ItemInfoEntry differ only in the size of the
item_ID field:

    if (version >= 2) {
        if (version == 2) {
            unsigned int(16) item_ID;
        } else if (version == 3) {
            unsigned int(32) item_ID;
        }
	...
    }

In the rest of src/read.c, we already use the uint32_t type for item
IDs. So we only need to change avifParseItemInfoEntry().

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2175,10 +2175,31 @@ static avifBool avifParseItemInfoEntry(avifMeta * meta, const uint8_t * raw, siz
 {
     BEGIN_STREAM(s, raw, rawLen, diag, "Box[infe]");
 
-    CHECK(avifROStreamReadAndEnforceVersion(&s, 2)); // TODO: support version > 2? 2+ is required for item_type
+    uint8_t version;
+    uint32_t flags;
+    CHECK(avifROStreamReadVersionAndFlags(&s, &version, &flags));
+    // Version 2+ is required for item_type
+    if (version != 2 && version != 3) {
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -2175,10 +2175,31 @@ static avifBool avifParseItemInfoEntry(avifMeta * meta, const uint8_t * raw, siz
...
+    // Version 2+ is required for item_type
+    if (version != 2 && version != 3) {
+        avifDiagnosticsPrintf(s.diag, "%s: Expecting box version 2 or 3, got version %u", s.diagContext, version);
+        return AVIF_FALSE;
+    }
+    // TODO: check flags. ISO/IEC 14496-12:2022, Section 8.11.6.1 says:
+    // The flags field of ItemInfoEntry with version greater than or equal to 2 is specified as
+    // follows:
+    // -- (flags & 1) equal to 1 indicates that the item is not intended to be a part of the
+    //    presentation. .
```
y-guyon, line 2190: Nit: Extra dot

y-guyon (approved):

src/read.c
wantehchang: Done. I also changed the comment to quote HEIF, which gives an example of a hidden image item.

> For example, when (flags & 1) is equal to 1 for an image item, the image item should not be displayed.

Joe pointed out that "intended to be a part of the presentation" seems ambiguous. HEIF's example should help clarify it.

PR #849, opened 2022-02-16
Fix signed/unsigned mismatch and loss-of-data conversion warnings for CL in new avif tests

joedrago:
MS' CL emits signed/unsigned comparison warnings and `uint32_t` -> `uint16_t` conversion warnings in these places. This should fix them.

Nit: We don't really have any explicit code style conventions for libavif, but ternarys in a for loop conditional hurts readability a bit, and I'd prefer in the future if those were simply a const value on the previous line with a good self-documenting name.

### DIFF
--- a/apps/shared/y4m.c
+++ b/apps/shared/y4m.c
@@ -360,12 +360,13 @@ avifBool y4mRead(const char * inputFilename, avifImage * avif, avifAppSourceTimi
         avifImageAllocatePlanes(avif, AVIF_PLANES_A);
     }
 
-    avifPixelFormatInfo info;
-    avifGetPixelFormatInfo(avif->yuvFormat, &info);
+    avifPixelFormatInfo formatInfo;
+    avifGetPixelFormatInfo(avif->yuvFormat, &formatInfo);
 
-    for (int plane = 0; plane < (info.monochrome ? 1 : AVIF_PLANE_COUNT_YUV); ++plane) {
-        uint32_t planeWidth = (plane > 0) ? ((avif->width + info.chromaShiftX) >> info.chromaShiftX) : avif->width;
-        uint32_t planeHeight = (plane > 0) ? ((avif->height + info.chromaShiftY) >> info.chromaShiftY) : avif->height;
+    const int planeCount = formatInfo.monochrome ? 1 : AVIF_PLANE_COUNT_YUV;
+    for (int plane = 0; plane < planeCount; ++plane) {
+        uint32_t planeWidth = (plane > 0) ? ((avif->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX) : avif->width;
+        uint32_t planeHeight = (plane > 0) ? ((avif->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY) : avif->height;
         uint32_t rowBytes = planeWidth << (avif->depth > 8);
         uint8_t * row = avif->yuvPlanes[plane];
         for (uint32_t y = 0; y < planeHeight; ++y) {
@@ -512,8 +513,8 @@ avifBool y4mWrite(const char * outputFilename, const avifImage * avif)
         rangeString = "XCOLORRANGE=LIMITED";
     }
 
-    avifPixelFormatInfo info;
-    avifGetPixelFormatInfo(avif->yuvFormat, &info);
+    avifPixelFormatInfo formatInfo;
+    avifGetPixelFormatInfo(avif->yuvFormat, &formatInfo);
 
     FILE * f = fopen(outputFilename, "wb");
     if (!f) {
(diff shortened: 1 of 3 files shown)


tests/avifgridapitest.c
```diff
@@ -29,7 +29,7 @@ static avifBool createImage(int width, int height, int depth, avifPixelFormat yu
     uint32_t uvWidth = ((*image)->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX;
     uint32_t uvHeight = ((*image)->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY;
 
-    for (uint32_t plane = 0; plane < (formatInfo.monochrome ? 1 : AVIF_PLANE_COUNT_YUV); ++plane) {
+    for (uint32_t plane = 0; plane < (formatInfo.monochrome ? 1U : AVIF_PLANE_COUNT_YUV); ++plane) {
```
wantehchang, line 32: Since `AVIF_PLANE_COUNT_YUV` is defined as 3, not 3U:

`#define AVIF_PLANE_COUNT_YUV 3`

won't `AVIF_PLANE_COUNT_YUV` also cause the signed/unsigned mismatch warning?

It seems simpler to just declare `plane` as `int` (and also use a local variable as you suggested in Chat):

```
const int planeCount = formatInfo.monochrome ? 1 : AVIF_PLANE_COUNT_YUV;
for (int plane = 0; plane < planeCount; ++plane) {
```

wantehchang (changes requested):
Joe: Thank you for the pull request. I'd like to suggest a change.

tests/avifgridapitest.c
joedrago: I agree. I've updated the commit to unify the look of all y4m-related plane iterations. They all now use the same variable names and types.

joedrago:
After further discussions with @wantehchang, I switched the loop variable back to `int` (from `uint32_t`) as that is more consistent with the rest of the codebase (and the original y4m code).

wantehchang (approved):
LGTM.

apps/shared/y4m.c
```diff
@@ -360,12 +360,13 @@ avifBool y4mRead(const char * inputFilename, avifImage * avif, avifAppSourceTimi
...
-    avifGetPixelFormatInfo(avif->yuvFormat, &info);
+    avifPixelFormatInfo formatInfo;
+    avifGetPixelFormatInfo(avif->yuvFormat, &formatInfo);
 
-    for (int plane = 0; plane < (info.monochrome ? 1 : AVIF_PLANE_COUNT_YUV); ++plane) {
-        uint32_t planeWidth = (plane > 0) ? ((avif->width + info.chromaShiftX) >> info.chromaShiftX) : avif->width;
-        uint32_t planeHeight = (plane > 0) ? ((avif->height + info.chromaShiftY) >> info.chromaShiftY) : avif->height;
+    const int planeCount = formatInfo.monochrome ? 1 : AVIF_PLANE_COUNT_YUV;
+    for (int plane = 0; plane < planeCount; ++plane) {
+        uint32_t planeWidth = (plane > 0) ? ((avif->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX) : avif->width;
```
y-guyon, line 368: Nit: `planeWidth` could be const too

tests/avifgridapitest.c
```diff
@@ -29,7 +29,8 @@ static avifBool createImage(int width, int height, int depth, avifPixelFormat yu
     uint32_t uvWidth = ((*image)->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX;
     uint32_t uvHeight = ((*image)->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY;
 
-    for (uint32_t plane = 0; plane < (formatInfo.monochrome ? 1 : AVIF_PLANE_COUNT_YUV); ++plane) {
+    const int planeCount = formatInfo.monochrome ? 1 : AVIF_PLANE_COUNT_YUV;
```
y-guyon, line 32: Is there any guideline for picking `int` over `uint32_t` besides avoiding warnings? I found a mix of types in the code base. Example: `avifImageCreate(int width, int height)` but `uint32_t avifImage::width`.

wantehchang: In general I would suggest going for consistency.

For the type of the loop index variable `plane`, Joe and I looked at existing code in libavif that iterates over the planes (I searched for "< 3;" and "AVIF_PLANE_COUNT_YUV"). We found that the` int` type is commonly used for this index variable. That's why we went with `int` for consistency.

`avifImageCreate(int width, int height)` vs `uint32_t avifImage::width` does seem like unintended inconsistency, because the three `int` parameters are assigned to `uint32_t` struct members without negative value checks:

```
avifImage * avifImageCreate(int width, int height, int depth, avifPixelFormat yuvFormat)
{
    avifImage * image = (avifImage *)avifAlloc(sizeof(avifImage));
    avifImageSetDefaults(image);
    image->width = width;
    image->height = height;
    image->depth = depth;
    image->yuvFormat = yuvFormat;
    return image;
}
```

Issue #851, opened 2022-02-16
Mark the input images of an image grid as hidden image items


wantehchang:
Yannis Guyon found that when using GIMP to view an AVIF grid image encoded by avifenc, GIMP asked the user to select an image to display from a list of not only the image grid item but also its input images (i.e., tiles). Leo Barnes pointed out that this is because the input images were not marked as hidden image items.

The libavif encoder should mark the input images of an image grid as hidden image items.

y-guyon:
GIMP prompt example of an AVIF image encoded with `avifenc --grid1x5`:
![image8](https://user-images.githubusercontent.com/51164219/154445289-bc44633b-1abe-4923-9b2c-a8bc8d9d2419.png)

PR #852, opened 2022-02-16
Mark the input images of an image grid as hidden

wantehchang:
Add the hiddenImage boolean field to the avifEncoderItem struct. When we
write the ItemInfoEntry boxes, if hiddenImage is true, set (flags & 1)
to 1.

Fix https://github.com/AOMediaCodec/libavif/issues/851.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -86,6 +86,7 @@ typedef struct avifEncoderItem
     avifCodecConfigurationBox av1C;       // Harvested in avifEncoderFinish(), if encodeOutput has samples
     uint32_t cellIndex;                   // Which row-major cell index corresponds to this item. ignored on non-av01 types
     avifBool alpha;
+    avifBool hiddenImage; // A hidden image item has (flags & 1) equal to 1 in its ItemInfoEntry.
 
     const char * infeName;
     size_t infeNameSize;
@@ -654,6 +655,7 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
 
             if (cellCount > 1) {
                 item->dimgFromID = gridColorID;
+                item->hiddenImage = AVIF_TRUE;
             } else {
                 encoder->data->primaryItemID = item->id;
             }
@@ -712,6 +714,7 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
 
                 if (cellCount > 1) {
                     item->dimgFromID = gridAlphaID;
+                    item->hiddenImage = AVIF_TRUE;
                 } else {
                     item->irefToID = encoder->data->primaryItemID;
                     item->irefType = "auxl";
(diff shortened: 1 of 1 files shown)


src/write.c
```diff
@@ -993,7 +996,8 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
     for (uint32_t itemIndex = 0; itemIndex < encoder->data->items.count; ++itemIndex) {
         avifEncoderItem * item = &encoder->data->items.item[itemIndex];
 
-        avifBoxMarker infe = avifRWStreamWriteFullBox(&s, "infe", AVIF_BOX_SIZE_TBD, 2, 0);
+        uint32_t flags = item->hiddenImage ? 1 : 0;
+        avifBoxMarker infe = avifRWStreamWriteFullBox(&s, "infe", AVIF_BOX_SIZE_TBD, 2, flags);
```
wantehchang, line 1000: Joe: There is an `avifRWStreamWriteFullBox(s, "infe", AVIF_BOX_SIZE_TBD, 2, 0)` call in `avifEncoderWriteTrackMetaBox()`.  Should I also update that call to consider `item->hiddenImage`?

src/write.c
```diff
@@ -86,6 +86,7 @@ typedef struct avifEncoderItem
     avifCodecConfigurationBox av1C;       // Harvested in avifEncoderFinish(), if encodeOutput has samples
     uint32_t cellIndex;                   // Which row-major cell index corresponds to this item. ignored on non-av01 types
     avifBool alpha;
+    avifBool hiddenImage; // A hidden image item has (flags & 1) equal to 1 in its ItemInfoEntry.
```
wantehchang, line 89: Right now, `item->hiddenImage` is true iff `item->dimgFromID` is nonzero. So we could just check `item->dimgFromID != 0` and not add the `hiddenImage` field. But I think adding the `hiddenImage` field will be a more general solution.

wantehchang (commented):
Joe: I have a question for you.

Yannis: Could you please test this pull request by encoding a grid image and viewing it using GIMP? Thanks.

src/write.c
joedrago: Anything that is written for a `trak`'s `meta` box is going to be considered a part of the presentation as it will be a frame in the sequence, so I think not. If anything, it should always be false when writing out a track's meta, right?

y-guyon:
> Yannis: Could you please test this pull request by encoding a grid image and viewing it using GIMP? Thanks.

Yes, it works. Using `avifenc --grid 1x5 img.png -o img.avif` and opening `img.avif` in GIMP opens the reconstructed image without prompting for selection.

y-guyon (approved):

wantehchang:
> > Yannis: Could you please test this pull request by encoding a grid image and viewing it using GIMP? Thanks.
> 
> Yes, it works. Using `avifenc --grid 1x5 img.png -o img.avif` and opening `img.avif` in GIMP opens the reconstructed image without prompting for selection.

Thanks a lot for the review and testing.

PR #853, opened 2022-02-17
Have avify4mtest get temp dir path from env var

wantehchang:


tests/avify4mtest.c
```diff
@@ -158,14 +158,15 @@ static avifBool encodeDecodeY4m(uint32_t width,
 
 //------------------------------------------------------------------------------
 
-int main(int argc, char * argv[])
+int main(void)
 {
-    if (argc != 2 || !strlen(argv[1])) {
-        fprintf(stderr, "Missing temporary folder path argument\n");
+    const char * testTmpdir = getenv("TEST_TMPDIR");
```
wantehchang, line 163: Using the hardcoded environment variable name "TEST_TMPDIR" optimizes for the Bazel and Blaze build environments.

We probably should use our own environment variable name, such as "AVIF_TEST_TMP_DIR". That will require passing the environment variable name as a command-line argument to avify4mtest.c. What do you think?

y-guyon: Yes, passing the environment variable name as a command-line argument to avify4mtest is a good idea.

wantehchang: Done.

tests/avify4mtest.c
```diff
@@ -161,11 +161,16 @@ static avifBool encodeDecodeY4m(uint32_t width,
 int main(int argc, char * argv[])
 {
     if (argc != 2 || !strlen(argv[1])) {
-        fprintf(stderr, "Missing temporary folder path argument\n");
+        fprintf(stderr, "Missing temporary directory environment variable argument\n");
```
y-guyon, line 164: "Missing temporary directory path environment variable name argument"?

y-guyon (approved):

tests/avify4mtest.c
wantehchang: Done. The error message is now precise but a little hard to parse.

PR #854, opened 2022-02-17
Fix FillImageGrid() yuvFormat assertion for alpha

y-guyon:
Also fix AVIF_PIXEL_FORMAT_NONE comment.
Also adapt avifgridapitest to cover this use case and >8 bit depths.

An avifImage with NONE as yuvFormat means there is no YUV pixel, but
alpha samples can still be present, which is the way tiles are
decoded. See how yuvFormat is left to NONE for alpha images in
aomCodecGetNextImage() and other codec implementations.

### DIFF
--- a/tests/avifgridapitest.c
+++ b/tests/avifgridapitest.c
@@ -30,7 +30,7 @@ static void fillPlane(int width, int height, int depth, uint8_t * row, uint32_t 
 
 // Creates an image where the pixel values are defined but do not matter.
 // Returns false in case of memory failure.
-static avifBool createImage(int width, int height, int depth, avifPixelFormat yuvFormat, int createAlpha, avifImage ** image)
+static avifBool createImage(int width, int height, int depth, avifPixelFormat yuvFormat, avifBool createAlpha, avifImage ** image)
 {
     *image = avifImageCreate(width, height, depth, yuvFormat);
     if (*image == NULL) {
@@ -63,7 +63,7 @@ static avifBool createImage(int width, int height, int depth, avifPixelFormat yu
 }
 
 // Generates then encodes a grid image. Returns false in case of failure.
-static avifBool encodeGrid(int columns, int rows, int cellWidth, int cellHeight, int depth, avifPixelFormat yuvFormat, int createAlpha, avifRWData * output)
+static avifBool encodeGrid(int columns, int rows, int cellWidth, int cellHeight, int depth, avifPixelFormat yuvFormat, avifBool createAlpha, avifRWData * output)
 {
     avifBool success = AVIF_FALSE;
     avifEncoder * encoder = NULL;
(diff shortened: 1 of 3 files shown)


tests/avifgridapitest.c
```diff
@@ -3,23 +3,41 @@
...
+            }
+        }
+        row += rowBytes;
+    }
+}
+
 // Creates an image where the pixel values are defined but do not matter.
 // Returns false in case of memory failure.
-static avifBool createImage(int width, int height, int depth, avifPixelFormat yuvFormat, avifImage ** image)
+static avifBool createImage(int width, int height, int depth, avifPixelFormat yuvFormat, int createAlpha, avifImage ** image)
```
wantehchang, line 33: Please use the `avifBool` type for the `createAlpha` parameter. Search for "int createAlpha" in this file and change all occurrences.

wantehchang (changes requested):
Please run `clang-format -i --style=file` on the files in the pull request, if you haven't done so. (I won't remind you in the future.)

tests/avifgridapitest.c
```diff
@@ -117,11 +137,11 @@ static avifBool decode(const avifRWData * encodedAvif)
 //------------------------------------------------------------------------------
 
 // Generates, encodes then decodes a grid image.
-static avifBool encodeDecode(int columns, int rows, int cellWidth, int cellHeight, avifPixelFormat yuvFormat, int expected_success)
+static avifBool encodeDecode(int columns, int rows, int cellWidth, int cellHeight, int depth, avifPixelFormat yuvFormat, int createAlpha, int expected_success)
```
wantehchang, line 140: Please also declare `expected_success` as `avifBool` and rename it `expectedSuccess`. See https://github.com/AOMediaCodec/libavif/pull/857/commits/5d8075e258cf412facca823a728d5baaf79eb8d5.

y-guyon:
> Please run `clang-format -i --style=file` on the files in the pull request, if you haven't done so. (I won't remind you in the future.)

`clang-format -i --style=file tests/avifgridapitest.c --verbose` did not modify or output anything. For the record I use CLion's "Reformat Code" with "Code style: Defined in .clang-format file". I believe it has the same behavior as the command-line tool.

tests/avifgridapitest.c
y-guyon: Sorry about that, I mixed up libwebp and libavif code styles.

Thanks for the commit example. Next time feel free to submit trivial fixes.

wantehchang:
> `clang-format -i --style=file tests/avifgridapitest.c --verbose` did not modify or output anything. For the record I use CLion's "Reformat Code" with "Code style: Defined in .clang-format file". I believe it has the same behavior as the command-line tool.

I see my reminder was confusing. Sorry. You don't need to use the exact clang-format command line I suggested. Any equivalent method that uses the .clang-format file is fine.

Issue #31, opened 2019-11-24
crash when passing in a png as input to avifenc


darix:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007feb30f5155d in __GI_abort () at abort.c:79
#2  0x00007feb30fab45f in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7feb310b9f49 "%s\n") at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007feb30fb28fc in malloc_printerr (str=str@entry=0x7feb310bbb80 "free(): double free detected in tcache 2") at malloc.c:5332
#4  0x00007feb30fb472d in _int_free (av=0x7feb310ec9e0 <main_arena>, p=0x555794e21290, have_lock=0) at malloc.c:4201
#5  0x00005557934da7a8 in main (argc=<optimized out>, argv=<optimized out>) at /usr/src/debug/libavif-0.5.2-1.1.x86_64/apps/avifenc.c:214
```

joedrago:
Currently, avifenc only supports y4m as a valid input, but I'll do a bit better job protecting against malformed y4m files.

joedrago:
https://github.com/AOMediaCodec/libavif/commit/70e16bb58da03f45117dd82c980c4a1a7908065c

PR #855, opened 2022-02-17
Add iso8 to compatible_brands for animated images

vigneshvg:
Section 10.3.1.1. of the HEIF spec:
It is required that 'iso8' is present among the compatible brands array.

Fixes the Compliance Warden error:
[heif][Rule #31] Error: 'msf1' brand: 'iso8' shall be present among the compatible brands array


wantehchang (approved):
LGTM. Cyril, could you also take a look? Thanks.

cconcolato:
I'm not sure I like the explosion of brands but that ship has sailed ... The idea was that an `iso8` readers/writers that are not aware of `msf1` would still be able to process the file in a meaningful way. That is a nice idea but in practice, this means they would have to remove the `msf1` brand (since they do not understand it and can't guarantee that their edits haven't violated the requirements of `msf1`).

And yes we should add it to the example in the spec. I created an issue https://github.com/AOMediaCodec/av1-avif/issues/187

Issue #12, opened 2019-09-24
Provide key frame list.


dalecurtis:
To save memory, some clients will want to evict things that are not part of the currently displayed set of frames. So having a list of which frames are required to decode a frame at index i is helpful.

Keyframe status can be found in stss, trun, or the OBU header.

joedrago:
I'm adding stss parsing now (it works locally), but what kind of API are you looking for here? 

Possibilities:

* avifDecoderIsKeyframe(decoder, frameIndex) // 0-based
* avifDecoderGetNearestKeyframe(decoder, frameIndex) // 0-based
* exposing the internal syncSample table

With this new keyframe data, I'm experimenting now with an avifDecoderGetNthImage(decoder, frameIndex) which will flush the codec, find the nearest keyframe, and decode up to it. If you had that, would knowing which are keyframes be less interesting?

dalecurtis:
I think both are required for our use case. The first or last API would be best since the Blink decoded image cache tries to be smart about what it can evict and what it doesn't need to decode using knowledge of which frames depend on what.

Later when it tries to refill some of the cache, the GetNthImage you propose would be useful to avoid spinning all the way from the beginning in the worst case (or over several keyframes in the average case).

joedrago:
Alright, I'll add all 3 calls hinted at above and see that goes.

joedrago:
Alright, v0.3.10 should have all of the goodies.

dalecurtis:
Thanks!

PR #856, opened 2022-02-17
Write ccst box in Sample Entry for animated images

vigneshvg:
Section 7.2.3 of the HEIF spec:
The CodingConstraintsBox shall be present in the sample description
entry for tracks with handler_type equal to 'pict'.

The values of the fields are chosen to be permissive rather than
figuring out the exact values by inspecting the bitstream which can be
too cumbersome. The values are as follows:
all_ref_pics_intra - 0 - all samples can use any type of reference.
intra_pred_used - 1 - intra prediction may or may not be used.
max_ref_per_pic - 15 - reserved value to indicate that any number of
                       reference images can be used.

Fixes the following Compliance Warden error:
[heif][Rule #12] Error: CodingConstraintsBox ('ccst') shall be present
once

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -1383,6 +1383,15 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
             if (!item->alpha) {
                 avifEncoderWriteColorProperties(&s, imageMetadata, NULL, NULL);
             }
+
+            avifBoxMarker ccst = avifRWStreamWriteFullBox(&s, "ccst", AVIF_BOX_SIZE_TBD, 0, 0);
+            const uint8_t ccstValue = (0 << 7) | // unsigned int(1) all_ref_pics_intra;
+                                      (1 << 6) | // unsigned int(1) intra_pred_used;
+                                      (15 << 2); // unsigned int(4) max_ref_per_pic;
+            avifRWStreamWriteU8(&s, ccstValue);
+            avifRWStreamWriteZeros(&s, 3); // unsigned int(26) reserved; (two zero bits are written along with ccstValue).
+            avifRWStreamFinishBox(&s, ccst);
+
             avifRWStreamFinishBox(&s, av01);
             avifRWStreamFinishBox(&s, stsd);
 


src/write.c
```diff
@@ -1380,6 +1380,15 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
             avifRWStreamWriteU16(&s, 0x0018);                          // template unsigned int(16) depth = 0x0018;
             avifRWStreamWriteU16(&s, (uint16_t)0xffff);                // int(16) pre_defined = -1;
             writeConfigBox(&s, &item->av1C);
+
+            avifBoxMarker ccst = avifRWStreamWriteFullBox(&s, "ccst", AVIF_BOX_SIZE_TBD, 0, 0);
+            const uint8_t ccstValue = (0 << 7) | // unsigned int(1) all_ref_pics_intra;
+                                      (1 << 6) | // unsigned int(1) intra_pred_used;
+                                      (15 << 2); // unsigned int(4) max_ref_per_pic;
```
wantehchang, line 1387: I think it's fine to use the special value 15 here, but I am just wondering if this value could be set to 7 or 8 for AV1.

src/write.c
```diff
@@ -1380,6 +1380,15 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
...
+            avifRWStreamWriteU8(&s, ccstValue);
+            avifRWStreamWriteZeros(&s, 3); // unsigned int(26) reserved; (two zeros are written along with ccstValue).
+            avifRWStreamFinishBox(&s, ccst);
+
```
wantehchang, line 1391: Optional: Although the order of the boxes probably doesn't matter, since the ccst box is apparently not used by the decoder, let's put it last, after the colr boxes:

```
            if (!item->alpha) {
                avifEncoderWriteColorProperties(&s, imageMetadata, NULL, NULL);
            }
```

wantehchang (approved):

src/write.c
```diff
@@ -1380,6 +1380,15 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
             avifRWStreamWriteU16(&s, 0x0018);                          // template unsigned int(16) depth = 0x0018;
             avifRWStreamWriteU16(&s, (uint16_t)0xffff);                // int(16) pre_defined = -1;
             writeConfigBox(&s, &item->av1C);
+
+            avifBoxMarker ccst = avifRWStreamWriteFullBox(&s, "ccst", AVIF_BOX_SIZE_TBD, 0, 0);
+            const uint8_t ccstValue = (0 << 7) | // unsigned int(1) all_ref_pics_intra;
+                                      (1 << 6) | // unsigned int(1) intra_pred_used;
+                                      (15 << 2); // unsigned int(4) max_ref_per_pic;
+            avifRWStreamWriteU8(&s, ccstValue);
+            avifRWStreamWriteZeros(&s, 3); // unsigned int(26) reserved; (two zeros are written along with ccstValue).
```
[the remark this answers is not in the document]
vigneshvg: Done.

src/write.c
vigneshvg: Yeah it is not a big deal. We can use either. This box is just here for conformance purposes and it's very unlikely that any decoder will actually use it in a meaningful way.

wantehchang (approved):
LGTM. Thanks!

PR #863, opened 2022-03-01
Remove mention of avifImageClear() in a comment

wantehchang:
The avifImageClear() function morphed into avifImageFreePlanes().

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -116,7 +116,6 @@ const char * avifProgressiveStateToString(avifProgressiveState progressiveState)
     return "Unknown";
 }
 
-// This function assumes nothing in this struct needs to be freed; use avifImageClear() externally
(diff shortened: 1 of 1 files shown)


src/avif.c
```diff
@@ -116,7 +116,7 @@ const char * avifProgressiveStateToString(avifProgressiveState progressiveState)
     return "Unknown";
 }
 
-// This function assumes nothing in this struct needs to be freed; use avifImageClear() externally
+// This function assumes nothing in this struct needs to be freed.
```
wantehchang, line 119: Joe: I can also remove this comment completely. Would you prefer that?

joedrago: Sure, just remove the whole comment.

wantehchang: Done.

joedrago (approved):

PR #865, opened 2022-03-02
Create internal avif_apps library

y-guyon:
To be used by apps and tests.

Next steps:
1. Split `CMakeLists.txt` into `tests/CMakeLists.txt` (`avify4mtest` requires `avifappsshared` to avoid including ../apps/shared/y4m.c from a neighbor directory)
2. Migrate `avify4mtest` and `avifgridapitest` to C++ and googletest (requires them to be in a separate CMake directory so that -Weverything can be disabled on the tests/ directory because it triggers too many C++ warnings)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -491,7 +491,7 @@ if(AVIF_BUILD_APPS OR AVIF_BUILD_TESTS)
     find_package(PNG REQUIRED)
     find_package(JPEG REQUIRED)
 
-    add_library(avif_apps STATIC # For internal use only. Not installed.
+    add_library(avif_apps STATIC # Static so it does not have to be installed.
         apps/shared/avifjpeg.c
         apps/shared/iccjpeg.c
         apps/shared/avifpng.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -487,23 +487,18 @@ option(AVIF_BUILD_APPS "Build avif apps." OFF)
 option(AVIF_BUILD_TESTS "Build avif tests." OFF)
 
 if(AVIF_BUILD_APPS OR AVIF_BUILD_TESTS)
-    # Not all targets depending on avifappsshared, such as tests, need png and jpeg.
-    # However it is simpler this way and likely that a test environment provides these.
     find_package(ZLIB REQUIRED)
     find_package(PNG REQUIRED)
     find_package(JPEG REQUIRED)
 
-    add_library(avifappsshared STATIC # Static so it does not have to be installed.
+    add_library(avif_apps STATIC # For internal use only. Not installed.
         apps/shared/avifjpeg.c
         apps/shared/iccjpeg.c
         apps/shared/avifpng.c
         apps/shared/avifutil.c
         apps/shared/y4m.c)
-    if(AVIF_LOCAL_LIBGAV1 OR AVIF_CODEC_LIBRARIES MATCHES vmaf)
-        set_target_properties(avifappsshared PROPERTIES LINKER_LANGUAGE "CXX")
-    endif()
-    target_link_libraries(avifappsshared avif ${PNG_LIBRARY} ${ZLIB_LIBRARY} ${JPEG_LIBRARY})
-    target_include_directories(avifappsshared
+    target_link_libraries(avif_apps avif ${PNG_LIBRARY} ${ZLIB_LIBRARY} ${JPEG_LIBRARY})
+    target_include_directories(avif_apps
                                PRIVATE
                                    $<TARGET_PROPERTY:avif,INTERFACE_INCLUDE_DIRECTORIES>
                                    ${PNG_PNG_INCLUDE_DIR}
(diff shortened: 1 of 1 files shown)


jzern (commented):
This looks like it's missing tests/CMakeLists.txt

wantehchang:
I wrote a pull request to do this a long time ago: https://github.com/AOMediaCodec/libavif/pull/75. At that time, Christopher Degawa suggested the library should be a cmake OBJECT library. You might want to consider that. We can ask Christopher to review this PR.

y-guyon:
> This looks like it's missing tests/CMakeLists.txt

The [next steps](https://github.com/AOMediaCodec/libavif/pull/865#issue-1157343487) are suggested future changes and not part of this change.

> I wrote a pull request to do this a long time ago: https://github.com/AOMediaCodec/libavif/pull/75. At that time, Christopher Degawa suggested the library should be a cmake OBJECT library. You might want to consider that. We can ask Christopher to review this PR.

The point of this change was also to get rid of:
```cmake
target_include_directories(avifenc|avifdec|avify4mtest PRIVATE apps/shared)
```
This can be done with the following that propagates the included directories and linked libraries:
```cmake
add_library(avifappsshared OBJECT)
target_link_libraries(avifappsshared avif ${PNG_LIBRARY} ${ZLIB_LIBRARY} ${JPEG_LIBRARY})
target_include_directories(avifappsshared INTERFACE apps/shared)
...
target_link_libraries(avifenc|avifdec|avify4mtest avif avifappsshared)
```
Unfortunately only the non-propagating `$<TARGET_OBJECTS:>` is available (until [CMake 3.12](https://cmake.org/cmake/help/v3.12/manual/cmake-buildsystem.7.html#object-libraries)):
```cmake
add_library(avifappsshared OBJECT)
...
target_link_libraries(avifenc|avifdec|avify4mtest avif $<TARGET_OBJECTS:avifappsshared> ${PNG_LIBRARY} ${ZLIB_LIBRARY} ${JPEG_LIBRARY})
target_include_directories(avifenc|avifdec|avify4mtest PRIVATE apps/shared)
```
We can either bump the CMake version to a fairly recent one, keep it `STATIC` for now (it is just one extra non-installed .a, right?) or include and link each of `avifenc|avifdec|avify4mtest`. I suggest keeping it `STATIC` for now and bumping the CMake version in another change, as I remember discarding newer CMake features for other matters because of the 3.5 target. What do you think?

CMakeLists.txt
```diff
@@ -484,49 +484,47 @@ if(AVIF_CODEC_LIBRARIES MATCHES vmaf)
...
+    # Not all targets depending on avifappsshared, such as tests, need png and jpeg.
+    # However it is simpler this way and likely that a test environment provides these.
     find_package(ZLIB REQUIRED)
     find_package(PNG REQUIRED)
     find_package(JPEG REQUIRED)
 
-    add_executable(avifenc
-        apps/avifenc.c
-
+    add_library(avifappsshared STATIC # Static so it does not have to be installed.
```
wantehchang, line 496: Nit: Shorten the library name to "avifapps" or "avif_apps". "avif_apps" is the name I used in the BUILD file in Google's internal repository. It would be good to be consistent.

wantehchang, line 496: The comment "Static so it does not have to be installed." is confusing, because we can install a static library, and we don't need to install all shared libraries. I think we can omit this comment, or change it to say something like "For internal use only. Not installed."

CMakeLists.txt
```diff
@@ -484,49 +484,47 @@ if(AVIF_CODEC_LIBRARIES MATCHES vmaf)
 endif()
 
 option(AVIF_BUILD_APPS "Build avif apps." OFF)
-if(AVIF_BUILD_APPS)
+option(AVIF_BUILD_TESTS "Build avif tests." OFF)
+
+if(AVIF_BUILD_APPS OR AVIF_BUILD_TESTS)
+    # Not all targets depending on avifappsshared, such as tests, need png and jpeg.
+    # However it is simpler this way and likely that a test environment provides these.
```
wantehchang, line 491: Nit: I think we can omit this comment. The new library must depend on png and jpeg because we decided to put avifjpeg.c and avifpng.c in the new library. Also, if new tests use png or jpeg in the future, this comment will become stale.

CMakeLists.txt
```diff
@@ -484,49 +484,47 @@ if(AVIF_CODEC_LIBRARIES MATCHES vmaf)
...
         apps/shared/avifjpeg.c
         apps/shared/iccjpeg.c
         apps/shared/avifpng.c
         apps/shared/avifutil.c
-        apps/shared/y4m.c
-    )
+        apps/shared/y4m.c)
     if(AVIF_LOCAL_LIBGAV1 OR AVIF_CODEC_LIBRARIES MATCHES vmaf)
-        set_target_properties(avifenc PROPERTIES LINKER_LANGUAGE "CXX")
+        set_target_properties(avifappsshared PROPERTIES LINKER_LANGUAGE "CXX")
```
wantehchang, line 503: Delete these three lines (502-504). The C++ linker language is only necessary when linking shared libraries and executables.

wantehchang (approved):
I suggest some changes. Also two high-level comments.

1. It seems better to avoid adding tests/CMakeLists.txt. Is dropping -Weverything for C++ files the only reason for adding tests/CMakeLists.txt? I am fine with dropping -Weverything for C files.

2. I am fine with creating the new library as a static library. It is not important to create it as a cmake OBJECT library.

CMakeLists.txt
y-guyon: If `apps_apps` is shared and `avifenc` is installed, `apps_apps` must be installed too, right?

wantehchang: Thanks for the explanation. I understand the comment now and have changed it back. Sorry about my confusion. I only considered the case of installing a library (including its headers) for external use (often called a -dev package).

PR #866, opened 2022-03-02
Fix avifpng.c for libpng 1.4

y-guyon:
The current default macOS CI version of libpng is 1.4; Ubuntu is 1.6.
`apps/shared/avifpng.c` is not built on the current macOS CI setup but will be in #865.
See log at https://github.com/y-guyon/libavif/runs/5393667489?check_suite_focus=true

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -12,9 +12,9 @@
 
 // See libpng-manual.txt, section XI.
 #if PNG_LIBPNG_VER_MAJOR > 1 || (PNG_LIBPNG_VER_MAJOR == 1 && PNG_LIBPNG_VER_MINOR >= 5)
-typedef png_bytep avifIccpDataP;
+typedef png_bytep png_iccp_datap;
 #else
-typedef png_charp avifIccpDataP;
+typedef png_charp png_iccp_datap;
 #endif
 
 // Note on setjmp() and volatile variables:
@@ -82,7 +82,7 @@ avifBool avifPNGRead(const char * inputFilename, avifImage * avif, avifPixelForm
     int iccpCompression = 0;
     unsigned char * iccpData = NULL;
     png_uint_32 iccpDataLen = 0;
-    if (png_get_iCCP(png, info, &iccpProfileName, &iccpCompression, (avifIccpDataP *)&iccpData, &iccpDataLen) == PNG_INFO_iCCP) {
+    if (png_get_iCCP(png, info, &iccpProfileName, &iccpCompression, (png_iccp_datap *)&iccpData, &iccpDataLen) == PNG_INFO_iCCP) {
         avifImageSetProfileICC(avif, iccpData, iccpDataLen);
     }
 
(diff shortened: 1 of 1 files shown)


apps/shared/avifpng.c
```diff
@@ -10,6 +10,13 @@
 #include <stdlib.h>
 #include <string.h>
 
+// See libpng-manual.txt, section XI.
+#if PNG_LIBPNG_VER_MAJOR > 1 || (PNG_LIBPNG_VER_MAJOR == 1 && PNG_LIBPNG_VER_MINOR >= 5)
+#define png_iccp_datap png_bytep
```
wantehchang, line 15: Nit: Use typedef instead of #define.

Optional: It would be good to avoid the `png_` prefix to make it clear that this is not a libpng type. Since this type is only used within this file, this is not that important.

wantehchang (changes requested):
I suggest a small change.

Do you know what is the version of macOS on the CI machines?

An alternative is to use a local libpng in ext/ by running ext/zlibpng.cmd and passing -DAVIF_LOCAL_ZLIBPNG=ON to cmake. Did you consider that?

wantehchang:
> The current default macOS CI version of libpng is 1.4; Ubuntu is 1.6. `apps/shared/avifpng.c` is not built on the current macOS CI setup but will be in #865. See log at https://github.com/y-guyon/libavif/runs/5393667489?check_suite_focus=true

I looked into this. The CI build log file shows it's using the libpng header `/Library/Frameworks/Mono.framework/Headers/png.h`. But Mono is not a standard system framework on macOS. It needs to be installed as described in https://www.mono-project.com/docs/about-mono/supported-platforms/macos/.

I checked my Mac Pro. It runs macOS 12.2.1 and has Xcode 13.2.1. It does not have the png.h header, although it does have zlib.h in `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include/zlib.h`. In any case, this shows macOS does not come with libpng. So I suggest we run zlibpng.cmd in `ext/` and pass -DAVIF_LOCAL_ZLIBPNG=ON to cmake.

Note: On my Mac Pro, I also had to run libjpeg.cmd in `ext/` and pass -DAVIF_LOCAL_JPEG=ON to cmake. It would be good to also do that on the CI macoO machines.

y-guyon:
> Do you know what is the version of macOS on the CI machines?

macOS 11.6 20G417 according to the [log](https://github.com/y-guyon/libavif/runs/5393667489?check_suite_focus=true#step:1:4).

> An alternative is to use a local libpng in ext/ by running ext/zlibpng.cmd and passing -DAVIF_LOCAL_ZLIBPNG=ON to cmake. Did you consider that?

No. I believe this patch can help others. I suggest submitting both #866 and #868.

wantehchang (approved):
LGTM.

I first wanted to note that I compared this pull request carefully with the information in libpng-manual.txt and think it is a well-written solution to the problem of building libavif with libpng 1.4.12.

Libpng 1.4.12 was released on July 10, 2012, almost ten years ago. It seems unlikely that people building the latest version of libavif cannot use a more recent version of libpng. That's why I wondered if we needed to solve that problem.

wantehchang:
I forgot to mention that I changed avifIccpDataP back to png_iccp_datap. It is helpful to have PNG in the type's name. I think either avif_png_iccp_datap or png_iccp_datap would be better, so I changed it back to the name you used before.

PR #867, opened 2022-03-02
Remove extra comma warning in avif.h

y-guyon:
It is triggered by -Wc++98-compat-pedantic when compiling a C++ target
including include/avif/avif.h with -Weverything set for Clang.


wantehchang (approved):
LGTM.

PR #868, opened 2022-03-02
ci.yml: Build examples and apps

wantehchang:
macOS does not come with libpng out of the box.


jzern:
Was jpeg a problem or installed somewhere in the platform SDK?

wantehchang:
macOS does not come with libjpeg out of the box either. So the first version of this pull request also built libjpeg locally. I removed it because I thought the jpeg library installed on the CI machines is likely to be libjpeg-turbo.

wantehchang:
The original motivation of this pull request was to test an alternative solution to the libpng problem Yannis is trying to fix in https://github.com/AOMediaCodec/libavif/pull/866. That's why it's optional to build libjpeg locally.

jzern:
Yannis' fix is a more general one and seems like it would be better beyond just this CI if people wanted to build on older distributions. If we wanted to put a minimum version requirement on libpng then we might automate the check in CMake and force a local build if needed.

wantehchang:
I removed the local build of zlib and libpng from this pull request so that Yannis's new code is exercised on the "macos-latest" CI build machine.

PR #871, opened 2022-03-09
Remove extra comma warning in avif.h

y-guyon:
It is triggered by `-Wc++98-compat-pedantic` when compiling a C++ target including `include/avif/avif.h` with `-Weverything` set for Clang (the current default setup in [libavif's CMakeLists.txt](https://github.com/AOMediaCodec/libavif/blob/b7c428c840afd1cf22aeb7b04d90520de27f14f0/CMakeLists.txt#L117)).

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -10,6 +10,13 @@
 #include <stdlib.h>
 #include <string.h>
 
+// See libpng-manual.txt, section XI.
+#if PNG_LIBPNG_VER_MAJOR > 1 || (PNG_LIBPNG_VER_MAJOR == 1 && PNG_LIBPNG_VER_MINOR >= 5)
+typedef png_bytep png_iccp_datap;
+#else
+typedef png_charp png_iccp_datap;
+#endif
+
 // Note on setjmp() and volatile variables:
 //
 // K & R, The C Programming Language 2nd Ed, p. 254 says:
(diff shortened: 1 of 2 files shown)


jzern:
@wantehchang there seems to be a disconnect between the C/C++ settings since C is requesting C99 as the standard. Yannis,  what C++ target are you compiling?

wantehchang:
James: Yannis said (in another pull request or in private email) that he ran into the compiler warning when writing libavif tests using the gtest (GoogleTest) framework.

Yannis: Was this pull request created in error? I already merged the extra comma warning fix:

https://github.com/AOMediaCodec/libavif/commit/ccde367c896798bf7f4535eb563cb268be30ca83

Although libavif uses -Weverything with clang right now, we can consider dropping that. -Weverything has been a maintenance burden because new warning flags are added to -Weverything in new versions of clang.

jzern:
Googletest requires C++11 now, so with that standard enabled this warning shouldn't appear I'd think.

https://github.com/google/googletest#supported-platforms

joedrago:
I'm closing this PR as it looks like @wantehchang has already removed the comma in the `master` branch.

y-guyon:
> Yannis: Was this pull request created in error?

Ah yes sorry. I did not remember it being merged, could not find it in open pull requests and forgot to sync it with master. My bad.

PR #873, opened 2022-03-10
Utilize bilinear chroma upsample in libyuv when possible

tongyuantongyu:
libyuv recently introduced the `I420ToARGBMatrixFilter` series functions with an extra `filter` parameter to select chroma upsampling behavior: nearest or bilinear. This PR utilize this ability when user requested bilinear upsampling.

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -37,8 +37,10 @@ unsigned int avifLibYUVVersion(void)
 #if defined(__clang__)
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wstrict-prototypes" // "this function declaration is not a prototype"
-// The newline at the end of libyuv/version.h was accidentally deleted in version 1792:
+// The newline at the end of libyuv/version.h was accidentally deleted in version 1792 and restored
+// in version 1813:
 // https://chromium-review.googlesource.com/c/libyuv/libyuv/+/3183182
+// https://chromium-review.googlesource.com/c/libyuv/libyuv/+/3527834
 #pragma clang diagnostic ignored "-Wnewline-eof" // "no newline at end of file"
 #endif
 #include <libyuv.h>
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -37,6 +37,9 @@ unsigned int avifLibYUVVersion(void)
 #if defined(__clang__)
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wstrict-prototypes" // "this function declaration is not a prototype"
+// The newline at the end of libyuv/version.h was accidentally deleted in version 1792:
+// https://chromium-review.googlesource.com/c/libyuv/libyuv/+/3183182
+#pragma clang diagnostic ignored "-Wnewline-eof" // "no newline at end of file"
 #endif
 #include <libyuv.h>
 #if defined(__clang__)
@@ -52,7 +55,8 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
     }
 
     if ((rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_AUTOMATIC) && (rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_FASTEST)) {
-        // libyuv uses its own upsampling filter. If the enduser chose a specific one, avoid using libyuv.
+        // We do not ensure specific filter is used when calling libyuv, so if the end user chose a specific one, avoid using it.
+        // Also libyuv trade a bit of accuracy for speed, so if the end user requested best quality, avoid using it as well.
         return AVIF_RESULT_NOT_IMPLEMENTED;
     }
 
@@ -187,6 +191,14 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
         return AVIF_RESULT_NOT_IMPLEMENTED;
     }
 
+#if LIBYUV_VERSION >= 1813
+    enum FilterMode filter = kFilterBilinear;
+    if (rgb->chromaUpsampling == AVIF_CHROMA_UPSAMPLING_FASTEST) {
+        // 'None' (Nearest neighbor) filter is faster than bilinear.
+        filter = kFilterNone;
+    }
+#endif
+
     // This following section might be a bit complicated to audit without a bit of explanation:
     //
     // libavif uses byte-order when describing pixel formats, such that the R in RGBA is the lowest address,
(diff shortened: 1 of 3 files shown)


src/reformat_libyuv.c
```diff
@@ -51,7 +52,12 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
         return AVIF_RESULT_NOT_IMPLEMENTED;
     }
 
+#if LIBYUV_VERSION >= 1812
+    if ((rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_AUTOMATIC) &&
+        (rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_FASTEST) && (rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_BILINEAR)) {
+#else
     if ((rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_AUTOMATIC) && (rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_FASTEST)) {
+#endif
         // libyuv uses its own upsampling filter. If the enduser chose a specific one, avoid using libyuv.
```
maryla-uc, line 61: This comment might need to be updated.

maryla-uc:
I agree the comments on the current options are a bit too specific and I would say it's ok to change the behavior slightly.

But in the interest of getting this patch submitted, I would suggest keeping the current libavif behavior for now, that is: reformat_libyuv only deals with AUTOMATIC and FASTEST. If the user specifies BEST_QUALITY, NEAREST or BILINEAR, fallback to C.

Then in another PR we can think about changing this behavior and adding more options. In general, I think we should try to have reasonable defaults and make it easy to use the most common options. I think the default should be to use libyuv when possible, preferring bilinear if available. And if the user asks for nearest or bilinear, it should also default to using libyuv if available.

For more obscure use cases, like using the slow C bilinear upsampling to benefit from floating point precision, then yes we could have an extra option.

src/reformat_libyuv.c
```diff
@@ -51,7 +52,12 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
         return AVIF_RESULT_NOT_IMPLEMENTED;
     }
 
+#if LIBYUV_VERSION >= 1812
+    if ((rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_AUTOMATIC) &&
+        (rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_FASTEST) && (rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_BILINEAR)) {
+#else
     if ((rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_AUTOMATIC) && (rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_FASTEST)) {
```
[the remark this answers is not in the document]
wantehchang: This may be due to the current definition of `AVIF_CHROMA_UPSAMPLING_NEAREST`:

```
typedef enum avifChromaUpsampling
{
    AVIF_CHROMA_UPSAMPLING_AUTOMATIC = 0,    // Chooses best trade off of speed/quality (prefers libyuv, else uses BEST_QUALITY)
    AVIF_CHROMA_UPSAMPLING_FASTEST = 1,      // Chooses speed over quality (prefers libyuv, else uses NEAREST)
    AVIF_CHROMA_UPSAMPLING_BEST_QUALITY = 2, // Chooses the best quality upsampling, given settings (avoids libyuv)
    AVIF_CHROMA_UPSAMPLING_NEAREST = 3,      // Uses nearest-neighbor filter (built-in)
    AVIF_CHROMA_UPSAMPLING_BILINEAR = 4      // Uses bilinear filter (built-in)
} avifChromaUpsampling;
```

Note that it is documented to use the built-in nearest-neighbor filter. We should review the definitions of these settings.

src/reformat_libyuv.c
```diff
@@ -37,6 +37,7 @@ unsigned int avifLibYUVVersion(void)
 #if defined(__clang__)
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wstrict-prototypes" // "this function declaration is not a prototype"
+#pragma clang diagnostic ignored "-Wnewline-eof" // "no newline at end of file"
```
wantehchang, line 40: Yuan: Could you add a newline character at the end of libyuv.h?

tongyuantongyu: The newline was removed (probably by mistake) in libyuv commit 48d167108ff60850ef788d25ef328d223a3f25c4 (version 1792, commited at 2021-09-25). I can write a CL to add it back, but there's many affected versions. Should I remove this line as well?

src/reformat_libyuv.c
```diff
@@ -52,7 +53,7 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
     }
 
     if ((rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_AUTOMATIC) && (rgb->chromaUpsampling != AVIF_CHROMA_UPSAMPLING_FASTEST)) {
-        // libyuv uses its own upsampling filter. If the enduser chose a specific one, avoid using libyuv.
+        // If the enduser chose a specific one, avoid using libyuv.
```
wantehchang, line 56: If we delete the first sentence "libyuv uses its own upsampling filter.", we need to change "one" to "upsampling filter" in the second sentence.

It would still be good to provide a reason. I guess you deleted "libyuv uses its own upsampling filter." because it is incorrect?

Maybe we can say something like the following:

> If the enduser chose a built-in upsampling filter or the best quality upsampling, avoid using libyuv.

wantehchang (changes requested):
Hi Yuan Tong,

I suggest two simple changes.

I think this is a reasonable solution without changing the avifChromeUpsampling enum type. This means if a recent version of libyuv is used, it will be slower but have better quality.

wantehchang:
Maryla wrote:
> 
> [...] in another PR we can think about changing this behavior and adding more options. In general, I think we should try to have reasonable defaults and make it easy to use the most common options. I think the default should be to use libyuv when possible, preferring bilinear if available. And if the user asks for nearest or bilinear, it should also default to using libyuv if available.
> 
> For more obscure use cases, like using the slow C bilinear upsampling to benefit from floating point precision, then yes we could have an extra option.

I agree with this general approach.

src/reformat_libyuv.c
tongyuantongyu: Done.
I added reasons of why we avoided libyuv in these cases.

tongyuantongyu:
> and we should treat the large +-18 errors of libyuv as a bug and work with Frank Barchard to fix it.

This is already fixed at code level at least for x86 and arm, but not enabled by default. We should start to roll it out.

wantehchang:
Hi Yuan Tong,

I used GitHub's "Squash and merge" command to merge this pull request. But the trackpad of my laptop was overly sensitive, so I didn't get a chance to remove the following (which was automatically populated by the "Squash and merge" command) from the commit message:

```
* Add more info to comment about -Wnewline-eof

* Add a comment about -Wnewline-eof

* Edit comment

Co-authored-by: Wan-Teh Chang <wtc@google.com>
```

Sorry about that.

PR #874, opened 2022-03-10
Fix -Wformat / -Wformat-non-iso on MinGW UCRT

tongyuantongyu:
MinGW can be configured to use UCRT, which is the new C runtime library for Windows. I'm moving my MinGW dev environment to the UCRT variant.

UCRT supports `"%zu"` properly so MinGW didn't patch`*printf` functions and set `__USE_MINGW_ANSI_STDIO` to 1. This makes `AVIF_FMT_ZU` defined to `"%Iu"`, and triggers warning from GCC and Clang.

Detect UCRT by check `_UCRT` macro, which MSSVC and MinGW both defines, and use `"%zu"` when using UCRT.

### DIFF
--- a/apps/shared/avifutil.h
+++ b/apps/shared/avifutil.h
@@ -6,14 +6,18 @@
 
 #include "avif/avif.h"
 
-// The %z format specifier is not available with Visual Studios before 2013 and mingw-w64 toolchains
-// with `__USE_MINGW_ANSI_STDIO` not set to 1. Hence the %I format specifier must be used instead
-// to print out `size_t`. Newer Visual Studios and mingw-w64 toolchains built with the commit
-// mentioned with c99 set as the standard supports the %z specifier properly.
+// The %z format specifier is not available in the old Windows CRT msvcrt,
+// Hence the %I format specifier must be used instead to print out `size_t`.
+// The new Windows CRT UCRT, which is used by Visual Studio from 2015, supports
+// the %z specifier properly.
+// Additionally, with c99 set as the standard mingw-w64 toolchains built with
+// the commit mentioned can patch format functions to support the %z specifier,
+// even it's using the old msvcrt, and this can be detected by
+// the `__USE_MINGW_ANSI_STDIO` macro.
 //
 // Related mingw-w64 commit: bfd33f6c0ec5e652cc9911857dd1492ece8d8383
 
-#if (defined(_MSVC) && _MSVC < 1800) || (defined(__USE_MINGW_ANSI_STDIO) && __USE_MINGW_ANSI_STDIO == 0)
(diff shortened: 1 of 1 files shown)


apps/shared/avifutil.h
```diff
@@ -6,14 +6,18 @@
...
+// Visual Studio from 2015 uses UCRT, which supports the %z specifier properly;
+// mingw-w64 toolchains can also be configured to use UCRT.
+// Additionally, with c99 set as the standard mingw-w64 toolchains built with
+// the commit mentioned can patch format functions to support the %z specifier,
+// and this can be detected by `__USE_MINGW_ANSI_STDIO` macro.
 //
 // Related mingw-w64 commit: bfd33f6c0ec5e652cc9911857dd1492ece8d8383
 
-#if (defined(_MSVC) && _MSVC < 1800) || (defined(__USE_MINGW_ANSI_STDIO) && __USE_MINGW_ANSI_STDIO == 0)
+#if !defined(_UCRT) && ((defined(_MSVC) && _MSVC < 1800) || (defined(__USE_MINGW_ANSI_STDIO) && __USE_MINGW_ANSI_STDIO == 0))
```
y-guyon, line 20: Is it possible to use UCRT with Visual Studio before 2013? Otherwise it can be "simplified" to:

```c++
#if (defined(_MSVC) && _MSVC < 1800) || (defined(__USE_MINGW_ANSI_STDIO) && __USE_MINGW_ANSI_STDIO == 0 && !defined(_UCRT))
```

tongyuantongyu: VS <= 2013 never use UCRT and VS >= 2015 always use UCRT.
However, I find this expression easier to understand: 

we want a feature in the C runtime. 
- UCRT definitely provides it.
- msvcr120 (the version of msvcrt Visual Studio 2013 uses) provides it, but no macro can be used to test CRT version so test Visual Studio version instead.
- MinGW sometimes provides it by patch and we can test if the patch is applied.

wantehchang: Hi Yuan,

It is reasonable to require Visual Studio 2015 or later today. For example, libaom requires Visual Studio 2017 or later:

https://aomedia.googlesource.com/aom/+/master/README.md#microsoft-visual-studio-builds

Let's drop support for Visual Studio 2013 or older. Could you simplify this expression?

kmilos:
Actually, this is much better fixed by [using standard `PRI*` print formatters from `inttypes`](https://www.cplusplus.com/reference/cinttypes/) directly in libavif instead of trying to figure it out here.

`inttypes.h` under MinGW implementation already does the correct `#if defined(_UCRT) || __USE_MINGW_ANSI_STDIO` for you.

apps/shared/avifutil.h
tongyuantongyu: Done

wantehchang (approved):
LGTM. I edited the comment slightly.

PR #878, opened 2022-03-15
Replace some memcpy calls with struct assignments

wantehchang:
An advantage of struct assignments is that the compiler does type
checking. It will also avoid the potential mistake of passing an
incorrect third argument to memcpy().

To avoid conflict with a pending pull request, this pull request does
not change src/read.c.


jzern:
lgtm. This is part of the motivation of preferring sizeof(var) over sizeof(type) in other constructions. In the end these should become memcpy's. Was there a bug?

y-guyon (approved):

wantehchang:
@jzern James: There was a bug caused by a copy-and-paste error of the sizeof(var) argument to memcpy(): https://github.com/AOMediaCodec/libavif/pull/483.

PR #882, opened 2022-03-15
No need to use the & operator with an array

wantehchang:
The name of an array can be used as a pointer to the first element of
the array, so it is not necessary to use the & operator with an array.
In other words, if `data` is an array, then `data` can be used as
`&data[0]`.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -27,7 +27,7 @@ uint16_t avifHTONS(uint16_t s)
 uint16_t avifNTOHS(uint16_t s)
 {
     uint8_t data[2];
-    memcpy(&data, &s, sizeof(data));
+    memcpy(data, &s, sizeof(data));
 
     return (uint16_t)((data[1] << 0) | (data[0] << 8));
 }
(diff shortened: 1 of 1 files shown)


wantehchang:
@y-guyon @jzern Please review.

src/utils.c
```diff
@@ -27,7 +27,7 @@ uint16_t avifHTONS(uint16_t s)
 uint16_t avifNTOHS(uint16_t s)
 {
     uint8_t data[2];
-    memcpy(&data, &s, sizeof(data));
+    memcpy(data, &s, sizeof(data));
 
     return (uint16_t)((data[1] << 0) | (data[0] << 8));
```
y-guyon, line 32: Out of curiousity, why not just the following? Is it about speed? I do not see how it solves any endianness issue even after reading the blog article above.
```c++
const uint8_t* data = (const uint8_t*)&s;
return (uint16_t)((data[1] << 0) | (data[0] << 8));
```

y-guyon (approved):

src/utils.c
wantehchang: Yannis: I believe your code is equivalent and saves a copy.

tongyuantongyu: Usually when reinterpreting some memory (for example getting bit representation of float point value), `memcpy` is used to avoid violating strict aliasing rule (https://en.cppreference.com/w/c/language/object#Strict_aliasing). But as a special case `char`/`unsigned char` aliasing is allowed, so `memcpy` is not necessary here. Anyway, with proper optimization level set, the call to `memcpy` should be optimized away by most compilers.

y-guyon: Thank you for the explanation.

PR #886, opened 2022-03-17
Fix compilation with clang 14

DimitryAndric:
Clang >= 14 warns that mixing declarations and code is incompatible with standards before C99, even if you compile with `-std=c99` or `-std=gnu99`.

Add `-Wno-declaration-after-statement` if it is supported, to silence the warning.


wantehchang (approved):
LGTM. Joe Drago and I discussed our use of -Weverything with clang.  We will replace it with -Wall -Wextra plus additional compiler warnings.

PR #889, opened 2022-03-19
Ext: Update to dav1d 1.0.0

wantehchang:
The patch ext/dav1d_oss_fuzz.patch no longer needs to be applied and is
deleted. Will delete ext/dav1d_oss_fuzz.sh after updating
oss-fuzz/projects/libavif/build.sh to use ext/dav1d.cmd again.


wantehchang:
@jzern @vigneshvg Please review. Thanks.

jzern (approved):

PR #893, opened 2022-03-25
Remove an avifImageFreePlanes call for alpha plane

wantehchang:
decoder->image is initially created as an empty image. If
decoder->data->alphaTileCount is 0, we will never allocate the alpha
plane of decoder->image or steal the alpha plane from the only "tile".
So we don't need to free the alpha plane of decoder->image in this case.


y-guyon (approved):

vigneshvg (approved):
lgtm

PR #894, opened 2022-03-25
Remove ext/dav1d_oss_fuzz.sh

wantehchang:
Now that oss-fuzz/projects/libavif/build.sh is using ext/dav1d.cmd
again (https://github.com/google/oss-fuzz/pull/7454), it is safe to
remove ext/dav1d_oss_fuzz.sh.


wantehchang:
@jzern Please review. This is a follow-up to https://github.com/AOMediaCodec/libavif/pull/889. Thanks.

jzern (approved):

PR #895, opened 2022-03-25
Reorder the if statements in avifDecoderNthImage

wantehchang:
Reorder the if statements in avifDecoderNthImage so that the one that
may fall through is the last one, after the if statements that always
return.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3982,12 +3982,17 @@ avifResult avifDecoderNthImage(avifDecoder * decoder, uint32_t frameIndex)
         return AVIF_RESULT_NO_CONTENT;
     }
 
-    if (frameIndex > INT_MAX) {
+    if ((frameIndex > INT_MAX) || ((int)frameIndex >= decoder->imageCount)) {
         // Impossible index
         return AVIF_RESULT_NO_IMAGES_REMAINING;
     }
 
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -3982,12 +3982,17 @@ avifResult avifDecoderNthImage(avifDecoder * decoder, uint32_t frameIndex)
...
     }
 
     int requestedIndex = (int)frameIndex;
+    if (requestedIndex == (decoder->imageIndex + 1)) {
+        // It's just the next image (already partially decoded or not at all), nothing special here
+        return avifDecoderNextImage(decoder);
+    }
+
     if (requestedIndex == decoder->imageIndex) {
         if ((decoder->data->decodedColorTileCount == decoder->data->colorTileCount) &&
```
wantehchang, line 3997: We can collapse the nested if conditionals. Should we do that?

y-guyon: If it is easier to understand, sure.

y-guyon (approved):

src/read.c
wantehchang: If we collapse the nested if conditionals, then we will need to move the comment that describes the fall-through. I will let you look into that.

PR #896, opened 2022-03-25
Simpler check for fully decoded image in NextImage

wantehchang:
Simplify the check for a completely decoded image in
avifDecoderNextImage(). We only need a yes/no answer; we don't need to
know how many rows are available if the image is not completely decoded.


y-guyon (approved):

wantehchang:
Thanks. I switched to `!=` in the conditionals and removed those two assertions.

PR #897, opened 2022-03-28
Remove unnecessary memcpy() calls in src/utils.c

y-guyon:
See https://en.cppreference.com/w/c/language/object#Strict_aliasing

Following https://github.com/AOMediaCodec/libavif/pull/882#discussion_r827426932

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -17,7 +17,7 @@ float avifRoundf(float v)
 uint16_t avifHTONS(uint16_t s)
 {
     uint16_t result;
-    uint8_t* data = (uint8_t*)&result;
+    uint8_t * data = (uint8_t *)&result;
     data[0] = (s >> 8) & 0xff;
     data[1] = (s >> 0) & 0xff;
     return result;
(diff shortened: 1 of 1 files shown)


src/utils.c
```diff
@@ -16,45 +16,40 @@ float avifRoundf(float v)
...
-    memcpy(&result, data, sizeof(uint16_t));
     return result;
 }
 
 uint16_t avifNTOHS(uint16_t s)
 {
-    uint8_t data[2];
-    memcpy(data, &s, sizeof(data));
-
+    uint8_t* data = (uint8_t*)&s;
```
wantehchang, line 28: In the NTOH functions, `data` should be declared as a const pointer.

wantehchang (changes requested):
Thank you for the pull request. Please run clang-format (or an equivalent method) on this file.

y-guyon:
> Thank you for the pull request. Please run clang-format (or an equivalent method) on this file.

Sorry, I forgot again. Switching between code bases and code styles...

wantehchang (approved):
LGTM. Thanks.

PR #898, opened 2022-03-28
Replace some checks by assertions in read.c

y-guyon:
Moving these checks to `avifDecoderReset()` as suggested in some previous review makes little sense as these variables are defined there. Keeping assertions in `avifDecoderNextImage()` is reassuring.


wantehchang (approved):
LGTM. I added parentheses around subexpressions to conform to the current coding style in libavif.

PR #900, opened 2022-03-30
Split CMakeLists.txt into tests/CMakeLists.txt

y-guyon:
See rationale at https://github.com/AOMediaCodec/libavif/pull/865

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -535,65 +535,8 @@ if(AVIF_BUILD_APPS)
 endif()
 
 if(AVIF_BUILD_TESTS)
-    # With testing enabled, all targets referenced by add_test() can be run
-    # at once with CMake's ctest command line tool from the build folder.
-    enable_testing()
-    # Using a CMake FIXTURES_SETUP/FIXTURES_CLEANUP to create and delete a
-    # subdirectory would be cleaner but this is way simpler.
-    set(AVIF_TEST_TMP_DIR ${PROJECT_BINARY_DIR})
-
-    add_executable(aviftest tests/aviftest.c)
-    if(AVIF_LOCAL_LIBGAV1)
-        set_target_properties(aviftest PROPERTIES LINKER_LANGUAGE "CXX")
-    endif()
-    target_link_libraries(aviftest avif ${AVIF_PLATFORM_LIBRARIES})
-
-    add_executable(avifgridapitest tests/avifgridapitest.c)
-    if(AVIF_LOCAL_LIBGAV1)
-        set_target_properties(avifgridapitest PROPERTIES LINKER_LANGUAGE "CXX")
-    endif()
-    target_link_libraries(avifgridapitest avif ${AVIF_PLATFORM_LIBRARIES})
-    add_test(NAME avifgridapitest
-        COMMAND ${CMAKE_BINARY_DIR}/avifgridapitest)
-
-    add_executable(avifincrtest tests/avifincrtest.c)
-    if(AVIF_LOCAL_LIBGAV1)
-        set_target_properties(avifincrtest PROPERTIES LINKER_LANGUAGE "CXX")
-    endif()
-    target_link_libraries(avifincrtest avif ${AVIF_PLATFORM_LIBRARIES})
-    add_test(NAME avifincrtest
-        COMMAND ${CMAKE_BINARY_DIR}/avifincrtest ${CMAKE_CURRENT_SOURCE_DIR}/tests/data/sofa_grid1x5_420.avif)
-
-    add_executable(avify4mtest tests/avify4mtest.c)
-    if(AVIF_LOCAL_LIBGAV1)
-        set_target_properties(avify4mtest PROPERTIES LINKER_LANGUAGE "CXX")
-    endif()
-    target_link_libraries(avify4mtest avif avif_apps ${AVIF_PLATFORM_LIBRARIES})
-    add_test(NAME avify4mtest
-        COMMAND ${CMAKE_BINARY_DIR}/avify4mtest AVIF_TEST_TMP_DIR)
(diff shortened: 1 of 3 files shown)


vrabaud (approved):

tests/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+    set_target_properties(aviftest PROPERTIES LINKER_LANGUAGE "CXX")
+endif()
+target_link_libraries(aviftest avif ${AVIF_PLATFORM_LIBRARIES})
+
+add_executable(avifgridapitest avifgridapitest.c)
+if(AVIF_LOCAL_LIBGAV1)
+    set_target_properties(avifgridapitest PROPERTIES LINKER_LANGUAGE "CXX")
+endif()
+target_link_libraries(avifgridapitest avif ${AVIF_PLATFORM_LIBRARIES})
+add_test(NAME avifgridapitest COMMAND avifgridapitest)
```
wantehchang, line 22: In this original code, this `add_test()` call is:

```
    add_test(NAME avifgridapitest
        COMMAND ${CMAKE_BINARY_DIR}/avifgridapitest)
```

So you removed `${CMAKE_BINARY_DIR}/` from the COMMAND. Is it because the test executable is now `${CMAKE_BINARY_DIR}/tests/avifgridapitest`?

tests/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+target_link_libraries(avifyuv avif ${AVIF_PLATFORM_LIBRARIES})
+
+add_custom_target(avif_test_all
+    COMMAND $<TARGET_FILE:aviftest> ${CMAKE_CURRENT_SOURCE_DIR}/data
+    DEPENDS aviftest
+)
+
+if(AVIF_ENABLE_COVERAGE)
+    add_custom_target(avif_coverage
+        COMMAND LLVM_PROFILE_FILE=${CMAKE_BINARY_DIR}/aviftest.profraw $<TARGET_FILE:aviftest> ${CMAKE_CURRENT_SOURCE_DIR}/data
```
wantehchang, line 52: You did not remove `${CMAKE_BINARY_DIR}/` from these `COMMAND`s. Is this okay?

Should we add `/tests` after `${CMAKE_BINARY_DIR}`?

CMakeLists.txt
```diff
@@ -535,65 +535,8 @@ if(AVIF_BUILD_APPS)
...
-        add_custom_target(avif_coverage
-            COMMAND LLVM_PROFILE_FILE=${CMAKE_BINARY_DIR}/aviftest.profraw $<TARGET_FILE:aviftest> ${CMAKE_CURRENT_SOURCE_DIR}/tests/data
-            COMMAND ${XCRUN} llvm-profdata merge -sparse ${CMAKE_BINARY_DIR}/aviftest.profraw -o ${CMAKE_BINARY_DIR}/aviftest.profdata
-            COMMAND cmake -E make_directory ${CMAKE_BINARY_DIR}/coverage
-            COMMAND ${XCRUN} llvm-cov show $<TARGET_FILE:aviftest> -instr-profile=${CMAKE_BINARY_DIR}/aviftest.profdata -project-title=libavif --format html -output-dir=${CMAKE_BINARY_DIR}/coverage
-            COMMAND echo Coverage report here: ${CMAKE_BINARY_DIR}/coverage/index.html
-            DEPENDS aviftest
-        )
-    endif()
+    enable_testing() # Allow ctest to be called from root directory.
```
wantehchang, line 538: Nit: root => top-level

The root directory may be interpreted as `/`.

wantehchang (changes requested):
Hi Yannis,

This change is straightforward and looks good. I have some questions and two suggested changes.

tests/CMakeLists.txt
1480c1: I think it's because
https://cmake.org/cmake/help/latest/command/add_test.html
> COMMAND
    Specify the test command-line. If <command> specifies an executable target (created by [add_executable()](https://cmake.org/cmake/help/latest/command/add_executable.html#command:add_executable)) it will automatically be replaced by the location of the executable created at build time.

wantehchang: Thank you for the answer, Christopher.

tests/CMakeLists.txt
1480c1, line 52: It may be better to use cmake's `env` command for `LLVM_PROFILE_FILE` instead of what it is currently (relies on posix shell)

https://stackoverflow.com/a/35042808


```suggestion
        COMMAND ${CMAKE_COMMAND} -E env "LLVM_PROFILE_FILE=${CMAKE_BINARY_DIR}/aviftest.profraw" $<TARGET_FILE:aviftest> ${CMAKE_CURRENT_SOURCE_DIR}/data
```

wantehchang: Christopher,

Thank you for the suggestion. I committed your suggested change with a truncated commit message accidentally.

wantehchang: Do you think we should change `${CMAKE_BINARY_DIR}` to `${CMAKE_BINARY_DIR}/tests` in these `COMMAND`s?

1480c1: Personally, leaning towards yes since that would ensure all "tests" are in the "tests" directory, regardless of source or build directory, but in practice, it would probably have minimal impact assuming the builder properly uses a separate build tree outside the source tree

joedrago: If you're switching to a child subdir's `CMakeLists.txt`, instead of `${CMAKE_BINARY_DIR}/tests`, you should use `${CMAKE_CURRENT_BINARY_DIR}`, which doesn't make assumptions about the root of the project (in case someone includes libavif as a child CMake project). It'll resolve to the same path here, but it is cleaner.

PR #904, opened 2022-04-04
Format CMakeLists.txt

y-guyon:
Use https://github.com/cheshirekow/cmake_format/tree/v0.6.13
Add config file .cmake-format.yaml.
Set CMakeLists.txt line width to 130 characters, as per .clang-format.
Format and add formatting instructions to README.md.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,10 +6,7 @@ cmake_minimum_required(VERSION 3.5)
 # Specify search path for CMake modules to be loaded by include() and find_package()
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
 
-project(
-    libavif
-    LANGUAGES C
-    VERSION 0.9.3)
+project(libavif LANGUAGES C VERSION 0.9.3)
 
 # Set C99 as the default
 set(CMAKE_C_STANDARD 99)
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -64,16 +70,24 @@ if(AVIF_LOCAL_ZLIBPNG)
...
-    set(PNG_TESTS OFF CACHE BOOL "")
+    set(PNG_BUILD_ZLIB
+        "${CMAKE_CURRENT_SOURCE_DIR}/ext/zlib"
+        CACHE STRING "" FORCE)
+    set(PNG_SHARED
+        OFF
+        CACHE BOOL "")
+    set(PNG_TESTS
+        OFF
+        CACHE BOOL "")
```
wantehchang, line 90: I think it's a good idea to format the cmake files with a tool, but the style file results in many short lines such as the three `set()` calls here (lines 82-90) and the `project()` call at lines 9-12.

Is it possible to avoid this problem?

I believe libaom also formats its cmake files with a tool, but its cmake files contains `set()` calls like the following:

```
      set(${config} "${${config}} ${c_flag}" CACHE STRING "" FORCE)
```
and
```
    set(${name} ${value} CACHE STRING "${helpstring}")
```

This gives me hope that it is possible.

jzern: I believe that's max_subargs_per_line, it's set to 10 in libaom (the default is 3).

y-guyon: Indeed there is the config file [`.cmake-format.py`](https://aomedia.googlesource.com/aom/+/refs/heads/main/.cmake-format.py) in libaom. Should I reuse the entire file in libavif? Or just add `max_subargs_per_line`?

I used YAML for the config file to match `.travis.yml` and `appveyor.yml` but python is fine too.

wantehchang: I suggest we aim to minimize the reformatting changes, so we should figure out the config settings for the current prevalent style in libavif. This is likely to require some experimentation. We can't use libaom's config file directly because the tab_size is different (2 vs. 4).

It seems that the config file consists of name-value pairs, so it doesn't matter whether the config file is written in YAML of Python.

y-guyon: (`max_subargs_per_line` [does no longer exist](https://github.com/cheshirekow/cmake_format/blob/904c42da298f7fda932d9ac774517c20e2667fb8/cmakelang/doc/release_notes.rst#v060) in cmake-format 0.6.3. It was replaced by `max_subgroups_hwrap` and `max_pargs_hwrap`.)

README.md
```diff
@@ -44,7 +51,30 @@ codec.
...
 
-If you're building on Windows with Visual Studio 2019 and want to try out libavif without going through the build process, static library builds for both Debug and Release are available on [AppVeyor](https://ci.appveyor.com/project/louquillio/libavif).
+If you're building on Windows with Visual Studio 2019 and want to try out
+libavif without going through the build process, static library builds for both
+Debug and Release are available on
+[AppVeyor](https://ci.appveyor.com/project/louquillio/libavif).
+
+## Development Notes
+
+libavif is written mainly in C99.
```
wantehchang, line 61: Can we remove the word "mainly"? It's not clear whether it means there is not only C but also C++ code, or some C99 features are not used.

wantehchang (changes requested):

CMakeLists.txt
```diff
@@ -127,8 +131,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang")
         -Wno-missing-noreturn
         -Wno-padded
         -Wno-sign-conversion
-        -Wno-error=c11-extensions
-    )
+        -Wno-error=c11-extensions)
```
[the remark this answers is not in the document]
y-guyon: Not a big fan but some lines are closer to original indeed.

wantehchang (approved):
LGTM. Thanks. I will merge this pull request after I create the 0.10.1 release.

wantehchang (approved):
Yannis: I found that there are some other CMakeLists.txt and *.cmake files in libavif:

cmake/Modules/Find*.cmake
contrib/CMakeLists.txt
contrib/gdk-pixbuf/CMakeLists.txt
android_jni/avifandroidjni/src/main/jni/CMakeLists.txt

I am going to merge this pull request, and let you format those files in a new pull request.

CMakeLists.txt
```diff
@@ -3,8 +3,7 @@
 
 cmake_minimum_required(VERSION 3.5)
 
-# Specify search path for CMake modules to be loaded by include()
-# and find_package()
```
wantehchang, line 7: Yannis: It seems that these two lines were manually merged into one line. When I ran cmake-format on the original CMakeLists.txt, it did not change these two lines.

This doesn't really matter. I am just testing cmake-format.

PR #906, opened 2022-04-05
Prepare avif example for non-aborting avifAlloc()

y-guyon:
This is part of a series of changes to replace abort() by return null
upon memory allocation failure in avifAlloc().

Bug: #820


wantehchang (approved):
LGTM. I made a small edit. The local variable `io` can be declared at the original location.

PR #909, opened 2022-04-06
tile_rows/columns in EbSvtAv1EncConfiguration log2

wantehchang:
The tile_rows and tile_columns members of the EbSvtAv1EncConfiguration
struct are log2 of tile rows and columns.

### DIFF
--- a/src/codec_svt.c
+++ b/src/codec_svt.c
@@ -121,12 +121,10 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         }
 
         if (encoder->tileRowsLog2 != 0) {
-            int tileRowsLog2 = AVIF_CLAMP(encoder->tileRowsLog2, 0, 6);
-            svt_config->tile_rows = 1 << tileRowsLog2;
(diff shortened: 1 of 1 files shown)


wantehchang:
@hassount @1480c1 Please review. Thanks.

I tested libavif with SVT-AV1 v1.0.0-rc1. I encoded three images using 1 and 4 worker threads, with and without tiles. The only problem I found is a bug in libavif, which I fixed in this pull request.

hassount:
@wantehchang thank you very much this is much appreciated!

1480c1 (approved):

wantehchang:
@1480c1 While looking into this issue, I found a misspelling of "columns" in the SVT-AV1 v1.0.0-rc1. Could you please fix them? You can find them with this command:

`find . -name "*" -a -type f | xargs grep colums`

1480c1:
> colums

https://gitlab.com/AOMediaCodec/SVT-AV1/-/merge_requests/1889


as a note, `git grep` might be better since it will only grep for files that are tracked by git

Issue #825, opened 2022-01-26
libavif only reports one XMP blob, despite it being perfectly valid for an item to reference many


leo-barnes:
I haven't actually checked if this is the case, but @wantehchang indicated that it's probably how it's currently working in the last AOM Image Container meeting.

HEIF, and by extension AVIF, supports an image item having multiple EXIF and XMP blobs associated with it. There's probably not much benefit to having multiple EXIF blobs, but since XMP is the main way that users can store arbitrary metadata into HEIF files, it's perfectly reasonable for an image item to have multiple XMP blobs.

One prime example is if you want to transcode a JPEG that has an IPTC IIM segment and an XMP segment into AVIF. The recommended way of storing IPTC these days is in XMP, which is a superset of the old IPTC IIM segment. So I would expect the clearest way of storing this metadata in an AVIF file is with one XMP blob containing the IPTC, and one XMP blob containing what was in the JPEG XMP blob.

y-guyon:
I believe you are talking about the single [`avifRWData xmp` in `avifImage`](https://github.com/AOMediaCodec/libavif/blob/6ee4926eb4650bd9ab22df60bf4c6fa56d1f8bfa/include/avif/avif.h#L464).

I see two main approaches to solving this issue.

**Allow the libavif encoder and decoder to accept and output multiple XMP blobs**

Although the API would be slightly messier (`avifRwDataArray xmps`?), this seems fairly simple and straightforward. However it moves the burden of leveraging several chunks to the client.

**Merging XMP and IPTC metadata into a single XMP blob**

As you mentioned, it is possible to "convert" an IIM chunk into an XMP one according to [XMP Specification Part 3 - 3.2.2](https://www.adobe.com/content/dam/acom/en/devnet/xmp/pdfs/XMPSDKReleasecc-2020/XMPSpecificationPart3.pdf). I guess one just has to insert each IPTC DataSet as a new XMP property, among the already existing XMP properties.

Is it libavif's duty to merge this kind of metadata, or should it be done with another tool as a preprocessing step?

Note: When using `exiv2 -e X ex` on [this example](https://upload.wikimedia.org/wikipedia/commons/5/5a/Metadata_test_file_-_includes_data_in_IIM%2C_XMP%2C_and_Exif.jpg), I found some `Iptc4xmpCore` and `Iptc4xmpExt` properties which look relevant.

y-guyon:
[...]

PR #911, opened 2022-04-07
Add multi XMP/Exif metadata encode/decode API

y-guyon:
The single metadata chunk API is kept as deprecated for now.
It may be removed in a later API-breaking release.
Add test.
See bug #825.


y-guyon:
As discussed, there is no need to handle multiple XMP nor EXIF chunks in libavif. libheif can be used for this rather specific use case.

Issue #912, opened 2022-04-07
The "build.sh install" command of SVT-AV1 does not work in our Docker image


wantehchang:
This bug was introduced in https://github.com/AOMediaCodec/libavif/commit/1594b6c5c4a320178ec72ef76668be37ba3c719b in the libavif v0.10.0 release.

The "build.sh install" command of SVT-AV1 exits with status 127 if the sudo command is not available. This is the case for our Docker image.

There are three possible solutions.

1. Instal the sudo package in the Docker image.
2. Do not use the build.sh script of SVT-AV1. Run cmake and ninja install directly.
3. Modify the build.sh script of SVT-AV1 to allow the sudo command to be missing when installing.

wantehchang:
This patch for the build.sh script of SVT-AV1 fixes the bug:

```
diff --git a/Build/linux/build.sh b/Build/linux/build.sh
index ebdb2c0c..75fd4eca 100755
--- a/Build/linux/build.sh
+++ b/Build/linux/build.sh
@@ -180,7 +180,7 @@ check_executable() (
 
 install_build() (
     build_type=Release
-    sudo=$(check_executable -p sudo)
+    sudo=$(check_executable -p sudo) || :
     while [ -n "$*" ]; do
         case $(printf %s "$1" | tr '[:upper:]' '[:lower:]') in
         release) build_type="Release" && shift ;;
```

PR #913, opened 2022-04-07
Build SVT-AV1 using cmake and ninja directly

wantehchang:
The build.sh script of SVT-AV1 does not work when install is requested
and the sudo command is not instaled. Run the cmake and ninja commands
directly.

Also add "set -e" to make the script exit if any command fails.

Fix https://github.com/AOMediaCodec/libavif/issues/912

### DIFF
--- a/tests/docker/build.sh
+++ b/tests/docker/build.sh
@@ -15,6 +15,8 @@
 #
 #     bash docker_ubuntu_shared_libs.sh
 
+set -e
+
 # build env
 apt update
 DEBIAN_FRONTEND="noninteractive" apt install -y build-essential libjpeg-dev libpng-dev libssl-dev ninja-build cmake pkg-config git perl vim curl python3-pip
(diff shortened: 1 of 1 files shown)


tests/docker/build.sh
```diff
@@ -70,8 +72,9 @@ cargo cinstall --prefix=/usr --release
 cd
 git clone -b v0.9.1 --depth 1 https://gitlab.com/AOMediaCodec/SVT-AV1.git
 cd SVT-AV1
-cd Build/linux
-./build.sh --prefix=/usr release install
+cd Build
+cmake -G Ninja -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release ..
+ninja install
```
wantehchang, line 77: @1480c1 @hassount Please review. Thanks.

I modeled after the commands at https://gitlab.com/AOMediaCodec/SVT-AV1#1-build-and-install-svt-av1.

1480c1, line 77: I personally would recommend instead

```suggestion
cmake -B Build -S . -G Ninja -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release
cmake --build Build --target install
```
to remove the cd, but this seems good too

wantehchang: Thank you for the suggestion. I suggest we first check in this version because it is similar to the other cmake and ninja commands in this file.

We can write another pull request to switch to your method throughout the file. Does this plan sound good?

1480c1: Sounds good to me

Issue #914, opened 2022-04-08
Declaration of f16 is invalid under ANSI C


hmaarrfk:
The definition of `f16` in https://github.com/AOMediaCodec/libavif/blob/656caa65b3caa5b1ad8c244a97a277c1b65ffce1/src/reformat.c#L1103

is causing our compilations to fail at conda-forge https://github.com/conda-forge/libavif-feedstock/pull/5

It seems to be non-standard according to

https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4204?view=msvc-170

hmaarrfk:
I guess we are going to disable treating warnings as error, but I just wanted to bring it to your attention.

wantehchang:
Mark: Thank you for the bug report. MSVC did not support all C99 features. I don't know if it's still the case. In any case, we can easily avoid that kind of initializer. I will write a pull request.

wantehchang:
Mark: I reproduced the compiler warning with Visual Studio 2017 (but not with Visual Studio 2019). It is not warning about the use of `.f =` in the initializer; it is warning about the use of a non-constant expression in the initializer.

hmaarrfk:
Is the answer to remove `const`?

wantehchang:
No. That `const` means the variable is immutable after initialization. The fix is to use an assignment instead of initialization. Note that the assignment requires removing the `const` from the variable declaration.

```
diff --git a/src/reformat.c b/src/reformat.c
index 1e3d365..ea48196 100644
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -1096,11 +1096,12 @@ static avifResult avifRGBImageToF16(avifRGBImage * rgb)
     for (uint32_t j = 0; j < rgb->height; ++j) {
         uint16_t * pixel = pixelRowBase;
         for (uint32_t i = 0; i < rgb->width * channelCount; ++i, ++pixel) {
-            const union
+            union
             {
                 float f;
                 uint32_t u32;
-            } f16 = { .f = *pixel * multiplier };
+            } f16;
+            f16.f = *pixel * multiplier;
             *pixel = (uint16_t)(f16.u32 >> 13);
         }
         pixelRowBase += stride;
```

PR #915, opened 2022-04-08
Don't use a non-constant expr as union initializer

wantehchang:
Don't use a non-constant expression as a union initializer. Fix the
Visual Studio 2017 compiler warning:

..\src\reformat.c(1103): warning C4204:
nonstandard extension used: non-constant aggregate initializer

Fix https://github.com/AOMediaCodec/libavif/issues/914.


wantehchang:
@hmaarrfk This is the fix.

y-guyon (approved):

hmaarrfk:
Thanks this fixed things for us.

Our build logs for cross reference
https://github.com/conda-forge/libavif-feedstock/pull/6

vigneshvg:
oops sorry about this. thanks for fixing @wantehchang.

PR #917, opened 2022-04-11
Fix help message of `avifdec`

sorairolake:
`--index` takes a value, but this is missing. So fix this.


wantehchang (approved):
LGTM. Thanks.

PR #921, opened 2022-04-13
Format all CMake files

y-guyon:
See https://github.com/AOMediaCodec/libavif/pull/904#pullrequestreview-939983042


wantehchang (approved):
LGTM. Thanks.

PR #923, opened 2022-04-20
Split avifincrtest.c into avifincrtest_helpers.c

y-guyon:
To reduce the file size and to be able to reuse some test functions from a C++ fuzz target.
First commit is only to branch the file for easier diff view.


tests/avifincrtest_helpers.h
```diff
@@ -0,0 +1,34 @@
+#ifndef LIBAVIF_TESTS_AVIFINCRTEST_HELPERS_H_
```
jzern, line 1: no copyright in the header?

jzern (approved):

wantehchang (approved):
LGTM. I tweaked the C system header #include statements.

PR #924, opened 2022-04-20
Convert some tests to C++ and GoogleTest

y-guyon:
Convert feature-oriented, unit tests to C++ and to the GoogleTest testing framework.
Migrating remaining C behaviors will be done in another change (`goto`, casts, etc.). Migrating other test targets (avifincrtest) will be done in another change.
Keep aviftest in C with no GoogleTest dependency as a minimum test suite when GoogleTest is unavailable.

Add `extern "C"` in `apps/shared` headers.
Add CMake local googletest lookup in `ext/` with `find_package` fallback.

Build GoogleTest and run all tests in the GitHub workflow.

Set `CMAKE_CXX_STANDARD` to C++11 and disable `-Weverything` for C++.

First commit is for correctly branched diff view.


.github/workflows/ci.yml
```diff
@@ -51,15 +52,28 @@ jobs:
...
-        cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON -DAVIF_CODEC_DAV1D=ON -DAVIF_LOCAL_DAV1D=ON -DAVIF_CODEC_RAV1E=ON -DAVIF_LOCAL_RAV1E=ON -DAVIF_CODEC_SVT=ON -DAVIF_LOCAL_SVT=ON -DAVIF_CODEC_LIBGAV1=ON -DAVIF_LOCAL_LIBGAV1=ON -DAVIF_LOCAL_LIBYUV=ON -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON -DAVIF_BUILD_TESTS=ON
+        cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF \
+          -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON \
+          -DAVIF_CODEC_DAV1D=ON -DAVIF_LOCAL_DAV1D=ON \
+          -DAVIF_CODEC_RAV1E=ON -DAVIF_LOCAL_RAV1E=ON \
+          -DAVIF_CODEC_SVT=ON -DAVIF_LOCAL_SVT=ON \
+          -DAVIF_CODEC_LIBGAV1=ON -DAVIF_LOCAL_LIBGAV1=ON \
+          -DAVIF_LOCAL_LIBYUV=ON \
+          -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON \
+          -DAVIF_BUILD_TESTS=1 -DAVIF_BUILD_GTESTS=1
```
jzern, line 72: they have the same meaning, but I guess for consistency with the rest of the command line =ON might be more appropriate

README.md
```diff
@@ -49,6 +49,14 @@ If you want to build/install shared libraries for AV1 codecs, you can still
 peek inside of each script to see where the current known-good SHA is for each
 codec.
 
+### Tests
+
+A few tests can be built by enabling the `AVIF_BUILD_TESTS` CMake option.
+
+The remaining tests can be built by enabling the `AVIF_BUILD_TESTS` and
+`AVIF_BUILD_GTESTS` CMake options. They require GoogleTest to be built locally
```
jzern, line 57: should AVIF_BUILD_GTESTS automatically set AVIF_BUILD_TESTS?

tests/avifgridapitest.cc
```diff
@@ -0,0 +1,276 @@
...
+                             bitDepth,
+                             yuvFormat,
+                             createAlpha,
+                             expectedSuccess));
+}
+
+// A cell cannot be smaller than 64px in any dimension if there are several cells.
+// A cell cannot have an odd size in any dimension if there are several cells and chroma subsampling.
+// Image size must be a multiple of cell size.
+const Cell kValidCells[] = { { 1, 64 }, { 1, 66 }, { 2, 64 }, { 3, 68 } };
```
jzern, line 201: these could be constexpr given we're requring c++11

jzern (commented):
Just a quick pass, I didn't look at the content of the tests

README.md
y-guyon: It could be useful to still have a way to at least run `aviftest` in environments where GoogleTest is unavailable.

I would suggest to set `AVIF_BUILD_GTESTS` to `ON` by default instead. This way enabling `AVIF_BUILD_TESTS` would be enough to run all tests. Another option would be a `AVIF_BUILD_NO_GTEST` flag but it seems bad.

jzern: > It could be useful to still have a way to at least run `aviftest` in environments where GoogleTest is unavailable.
> 
> I would suggest to set `AVIF_BUILD_GTESTS` to `ON` by default instead. This way enabling `AVIF_BUILD_TESTS` would be enough to run all tests. Another option would be a `AVIF_BUILD_NO_GTEST` flag but it seems bad.

I agree NO isn't the best. I was only thinking that since AVIF_BUILD_GTESTS requires AVIF_BUILD_TESTS, cmake -DAVIF_BUILD_GTESTS=1 could be a shorthand for cmake -DAVIF_BUILD_GTESTS=1 -DAVIF_BUILD_TESTS=1

joedrago:
Aside: I'd just like to go on record that while I don't really mind what test tooling we use and the languages they use therein, I never want `libavif` _itself_ or its "apps" (`avifenc`, `avifdec`) to ever require C++. The base library will continue to stay C, and provided downstream projects only use C-based codecs, they should be able to stay entirely in C.

jzern:
Joe, that makes sense. We made similar decisions in libvpx, libaom and libwebp.

y-guyon:
> The base library will continue to stay C

I did not plan to change that, this pull request is only to make testing easier.

> I never want `libavif` to ever require C++

Out of curiosity, may I know why? I do not wish to start a debate, this is just to clarify. Are you talking about the public interface `avif.h` only or also the core library (the latter having no impact on downstream projects nor C-based codecs)?

joedrago:
I don't want to debate either; it is personal preference. If I wanted libavif to be written in C++, I would have chosen it instead of C when I first wrote it.

wantehchang (changes requested):
Yannis: Thank you for the pull request. Sorry about the late review. I am giving this PR high priority.

Overall this looks good. I suggest some changes.

.github/workflows/ci.yml
```diff
@@ -51,15 +52,28 @@ jobs:
...
+          -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON \
+          -DAVIF_BUILD_TESTS=ON -DAVIF_BUILD_GTESTS=ON
     - name: Build libavif (make)
       working-directory: ./build
       run: make -j $(($(nproc) + 1))
     - name: Run AVIF Tests (on Linux)
       if: runner.os == 'Linux'
       working-directory: ./build
-      run: ./tests/aviftest ../tests/data
+      run: ./tests/aviftest ../tests/data && ctest -j $(($(nproc) + 1))
```
wantehchang, line 79: Optional: I suggest we move this change and the other changes I suggested in this comment to a separate pull request, because they are independent of GoogleTest.

Nit: Change `&&` to a semicolon `;`. Alternatively, delete `&&` and move the `ctest -j $(($(nproc) + 1))` command to the next line. If `aviftest` fails, we should still run the other tests.

Note 1: We should also run the `avifyuv` test.

Note 2: We should move `aviftest` and `avifyuv` to the ctest framework (and remove the custom target `avif_test_all`). Then the `ctest -j $(($(nproc) + 1))` command will run all tests.

tests/CMakeLists.txt
```diff
@@ -44,6 +26,36 @@ target_link_libraries(avifyuv avif ${AVIF_PLATFORM_LIBRARIES})
...
+    else()
+        message(STATUS "Missing local ext/googletest, falling back to finding installed GoogleTest")
+        find_package(GTest REQUIRED)
+    endif()
+
+    if(CMAKE_C_COMPILER_ID MATCHES "Clang")
+        # -Weverything is mostly fine for C but triggers too many C++ warnings in GoogleTest.
+        add_compile_options(-Wno-everything) # Applied to current directory only.
+    endif()
+
```
wantehchang, line 47: The help message for the `AVIF_BUILD_GTESTS` option says "Requires GoogleTest to be installed and AVIF_BUILD_TESTS to be ON." Should we enforce these two requirements here?

CMakeLists.txt
```diff
@@ -479,8 +479,11 @@ endif()
 
 option(AVIF_BUILD_APPS "Build avif apps." OFF)
 option(AVIF_BUILD_TESTS "Build avif tests." OFF)
+option(AVIF_BUILD_GTESTS
+       "Build avif C++ tests depending on GoogleTest. Requires GoogleTest to be installed and AVIF_BUILD_TESTS to be ON." OFF
```
wantehchang, line 483: Nit: change "installed" to "present" or "available"?

If we use ext/googletest, then GoogleTest doesn't need to be installed on the machine.

wantehchang (changes requested):

.github/workflows/ci.yml
```diff
@@ -41,6 +41,7 @@ jobs:
       working-directory: ./ext
       run: bash rav1e.cmd
     - name: Build SVT-AV1
+      if: steps.cache-ext.outputs.cache-hit != 'true'
```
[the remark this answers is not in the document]
y-guyon: I think it needs to build and pass once before being cached.

tests/avifgridapitest.cc
```diff
@@ -0,0 +1,276 @@
+// Copyright 2022 Google LLC. All rights reserved.
```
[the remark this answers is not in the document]
y-guyon: As far as I know, there is no file branching/renaming concept in git. [`git mv` is equivalent to `mv`](https://stackoverflow.com/a/1094392). `git diff` may autodetect it as such and adapt the diff accordingly, but there is no guarantee. You can force it with `git diff HEAD:./oldfilename newfilename` apparently.

The PR description mentions a first commit that only renames .c to .cc. Unfortunately GitHub UI does not pick that as a hint; it squashes all commits and then computes the diff. Fortunately you can select all commits but the first one in the [GitHub UI](https://github.com/AOMediaCodec/libavif/pull/924/files/4a222e42d4baaf0c2108fbef7617424ea536a24f..9a5ea574be5f31d71255a8dfc0a2f25ec4fc9d7d), showing no new file.

tests/CMakeLists.txt
y-guyon: I added a message when `AVIF_BUILD_GTESTS` is `ON` but `AVIF_ENABLE_GTEST` is `OFF`.

I added a warning when `AVIF_BUILD_GTESTS` is `OFF` but `AVIF_ENABLE_GTEST` is `ON`.

.github/workflows/ci.yml
y-guyon: Done at PR https://github.com/AOMediaCodec/libavif/pull/943.

tests/CMakeLists.txt
```diff
@@ -44,6 +26,36 @@ target_link_libraries(avifyuv avif ${AVIF_PLATFORM_LIBRARIES})
...
+    if(EXISTS ${GTEST_INCLUDE_DIRS}/gtest/gtest.h AND EXISTS ${GTEST_LIBRARIES} AND EXISTS ${GTEST_MAIN_LIBRARIES})
+        message(STATUS "Found local ext/googletest")
+    else()
+        message(STATUS "Missing local ext/googletest, falling back to finding installed GoogleTest")
+        find_package(GTest REQUIRED)
+    endif()
+
+    if(CMAKE_C_COMPILER_ID MATCHES "Clang")
+        # -Weverything is mostly fine for C but triggers too many C++ warnings in GoogleTest.
+        add_compile_options(-Wno-everything) # Applied to current directory only.
```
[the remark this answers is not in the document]
y-guyon: Done in PR https://github.com/AOMediaCodec/libavif/pull/944.

.github/workflows/ci.yml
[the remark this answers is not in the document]
y-guyon: Yes. I found [this](https://news.ycombinator.com/item?id=25224469) and [that](https://unix.stackexchange.com/questions/519092/what-is-the-logic-of-using-nproc-1-in-make-command) on the subject, but no definitive proven guideline. Anyway I enabled `ctest` in the separate PR https://github.com/AOMediaCodec/libavif/pull/943 and removed the +1 (I guess there is no I/O strain for tests, only for compilation).

tests/avifgridapitest.cc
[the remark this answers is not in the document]
jzern: Locally you can also tune the rename detection with -M<similarity%> (and -C for copies). --follow helps with git-log.

tests/avifgridapitest.cc
```diff
@@ -0,0 +1,276 @@
...
+using testing::Values;
+using testing::ValuesIn;
+
+namespace
+{
+
+//------------------------------------------------------------------------------
+
+// Fills a plane with a repeating gradient.
+static void fillPlane(int width, int height, int depth, uint8_t * row, uint32_t rowBytes)
```
jzern, line 24: static is redundant in an anonymous namespace

tests/avifgridapitest.cc
```diff
@@ -0,0 +1,276 @@
...
+
+//------------------------------------------------------------------------------
+
+// Pair of cell count and cell size for a single dimension.
+struct Cell
+{
+    int count, size;
+};
+
+class GridApiTest : public testing::TestWithParam<std::tuple<Cell, Cell, int, avifPixelFormat, bool, bool>>
```
jzern, line 175: you should explicitly #include <tuple>

tests/avify4mtest.cc
```diff
@@ -5,21 +5,31 @@
...
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
+#include <cassert>
+#include <cinttypes>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <sstream>
+#include <string>
```
jzern, line 14: same here for <tuple>

CMakeLists.txt
```diff
@@ -464,8 +467,11 @@ endif()
 
 option(AVIF_BUILD_APPS "Build avif apps." OFF)
 option(AVIF_BUILD_TESTS "Build avif tests." OFF)
+option(AVIF_ENABLE_GTEST
+       "Build avif C++ tests, which depend on GoogleTest. Requires GoogleTest. Has no effect unless AVIF_BUILD_TESTS is ON." OFF
```
wantehchang, line 471: Would you like to change the default to ON as you suggested before? If you do that, remember to delete the warning message at lines 515-516.

wantehchang:
On Fri, May 13, 2022 at 9:08 AM Yannis Guyon ***@***.***> wrote:
>
> @y-guyon commented on this pull request.
>
> ________________________________
>
> In tests/CMakeLists.txt:
>
> > +    set(GTEST_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/ext/googletest/googletest/include)
> +    set(GTEST_LIBRARIES ${CMAKE_SOURCE_DIR}/ext/googletest/build/lib/libgtest${CMAKE_STATIC_LIBRARY_SUFFIX})
> +    set(GTEST_MAIN_LIBRARIES ${CMAKE_SOURCE_DIR}/ext/googletest/build/lib/libgtest_main${CMAKE_STATIC_LIBRARY_SUFFIX})
> +    set(GTEST_BOTH_LIBRARIES ${GTEST_LIBRARIES} ${GTEST_MAIN_LIBRARIES})
> +    if(EXISTS ${GTEST_INCLUDE_DIRS}/gtest/gtest.h AND EXISTS ${GTEST_LIBRARIES} AND EXISTS ${GTEST_MAIN_LIBRARIES})
> +        message(STATUS "Found local ext/googletest")
> +    else()
> +        message(STATUS "Missing local ext/googletest, falling back to finding installed GoogleTest")
> +        find_package(GTest REQUIRED)
> +    endif()
> +
> +    if(CMAKE_C_COMPILER_ID MATCHES "Clang")
> +        # -Weverything is mostly fine for C but triggers too many C++ warnings in GoogleTest.
> +        add_compile_options(-Wno-everything) # Applied to current directory only.
> +    endif()
> +
>
> I added a message when AVIF_BUILD_GTESTS is ON but AVIF_ENABLE_GTEST is OFF.
>
> I added a warning when AVIF_BUILD_GTESTS is OFF but AVIF_ENABLE_GTEST is ON.

Hi Yannis,

In this comment, "AVIF_BUILD_GTESTS" is a typo. Should be
"AVIF_BUILD_TESTS". I edited the comment and fixed the typo (two
occurrences).

Wan-Teh

wantehchang (commented):
Yannis: I will let James review the two new .cc files. I have reviewed everything else and they look good to me. I will wait until James has reviewed this PR to approve it.

.github/workflows/ci.yml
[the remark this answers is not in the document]
y-guyon: Done at https://github.com/AOMediaCodec/libavif/pull/949

tests/avifgridapitest.cc
```diff
@@ -0,0 +1,276 @@
...
+
+//------------------------------------------------------------------------------
+
+// Decodes the data. Returns false in case of failure.
+bool decode(const avifRWData * encodedAvif)
+{
+    bool success = false;
+    avifImage * const image = avifImageCreateEmpty();
+    avifDecoder * const decoder = avifDecoderCreate();
+    if (image == NULL || decoder == NULL) {
```
jzern, line 127: maybe in a follow up these checks could be normalized like in your other commit [1] to keep the move simple in this change

[1] https://github.com/AOMediaCodec/libavif/commit/14c948e7c0c5bc15c334cf1eaa5ad3384f58ee0a

tests/avifgridapitest.cc
```diff
@@ -0,0 +1,276 @@
...
+
+//------------------------------------------------------------------------------
+
+// Pair of cell count and cell size for a single dimension.
+struct Cell
+{
+    int count, size;
+};
+
+class GridApiTest : public testing::TestWithParam<std::tuple<Cell, Cell, int, avifPixelFormat, bool, bool>>
```
jzern, line 175: Can you document the parameters at this level too?

tests/avifgridapitest.cc
```diff
@@ -172,7 +172,7 @@ struct Cell
     int count, size;
 };
 
-class GridApiTest : public testing::TestWithParam<std::tuple<Cell, Cell, int, avifPixelFormat, bool, bool>>
+class GridApiTest : public testing::TestWithParam<std::tuple</*horizontal=*/Cell, /*vertical=*/Cell, /*bitDepth=*/int, /*yuvFormat=*/avifPixelFormat, /*createAlpha=*/bool, /*expectedSuccess=*/bool>>
```
jzern, line 175: this seems to break clang-format's ColumnLimit setting of 130

tests/avifgridapitest.cc
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    if (createAlpha) {
+        fillPlane((*image)->width, (*image)->height, (*image)->depth, (*image)->alphaPlane, (*image)->alphaRowBytes);
+    }
+    return true;
+}
+
+// Generates then encodes a grid image. Returns false in case of failure.
+bool encodeGrid(int columns, int rows, int cellWidth, int cellHeight, int depth, avifPixelFormat yuvFormat, bool createAlpha, avifRWData * output)
```
jzern, line 75: no change required, but it seems like something in .clang-format is allowing this to exceed the column limit of 130.

jzern (approved):

wantehchang (approved):
LGTM. You need to resolve the merge conflict in tests/CMakeLists.txt.

Please merge this PR using the "Squash and merge" method.

PR #925, opened 2022-04-26
Add comment in avif.h for avifRWData functions

y-guyon:


wantehchang (approved):
LGTM. Thanks!

I shortened "zero-initialized or defined" to "initialized". But it is also not accurate.

I think a complete description would be something like "zero-initialized and optionally only manipulated with these functions." Do you have any suggestions?

y-guyon:
> I shortened "zero-initialized or defined" to "initialized".
> 
> But it is also not accurate. I think a complete description would be something like "zero-initialized and optionally only manipulated with these functions." Do you have any suggestions?

- "The avifRWData input must be zero-initialized before being manipulated with these functions."
- "Initialize avifRWData to AVIF_DATA_EMPTY, then manipulate it with these functions."

wantehchang:
Thank you. I switched to the first comment you suggested.

PR #926, opened 2022-04-28
Remove alphaRange from avifImage

vigneshvg:
The AVIF draft specification disallows limited range alpha and it
will eventually be in the specification once AOM ratifies it [1] [2].

Even though the default value of alphaRange is set to
AVIF_RANGE_FULL, it is possible for applications to override it
and set it to AVIF_RANGE_LIMITED thereby producing AVIF images
that will eventually be invalid.

On the encoder side, This PR prevents such images from being
produced by libavif.

On the decoder side, if we encounter an image with limited range
alpha, we always convert it to full range alpha and the applications
can assume that alpha plane is always in full range (as it will be
required by the spec eventually).

This change breaks the ABI and a note has been added to CHANGELOG.md
in the "Unreleased" section to document that.

[1] https://github.com/AOMediaCodec/av1-avif/pull/101
[2] https://aomediacodec.github.io/av1-avif/#auxiliary-images

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+There is an incompatible ABI change in this release. Members were removed from
+avifImage struct. It is necessary to recompile your code.
+
+### Removed
+* alphaRange field was removed from the avifImage struct. It it presumed that
+  alpha plane is always full range.
+
 ## [0.10.1] - 2022-04-11
 
 ### Changed
--- a/apps/shared/avifutil.c
+++ b/apps/shared/avifutil.c
@@ -53,10 +53,6 @@ static void avifImageDumpInternal(const avifImage * avif, uint32_t gridCols, uin
         printf(" * Chroma Sam. Pos: %u\n", avif->yuvChromaSamplePosition);
     }
     printf(" * Alpha          : %s\n", alphaPresent ? (avif->alphaPremultiplied ? "Premultiplied" : "Not premultiplied") : "Absent");
-    if (avif->alphaRange == AVIF_RANGE_LIMITED) {
-        printf("                    Limited range\n");
-        printf("                    WARNING: Limited-range alpha is deprecated. Use full-range alpha instead.\n");
-    }
     printf(" * Range          : %s\n", (avif->yuvRange == AVIF_RANGE_FULL) ? "Full" : "Limited");
 
     printf(" * Color Primaries: %u\n", avif->colorPrimaries);
--- a/examples/avif_example_decode_file.c
+++ b/examples/avif_example_decode_file.c
@@ -49,7 +49,7 @@ int main(int argc, char * argv[])
     while (avifDecoderNextImage(decoder) == AVIF_RESULT_OK) {
         // Now available (for this frame):
         // * All decoder->image YUV pixel data (yuvFormat, yuvPlanes, yuvRange, yuvChromaSamplePosition, yuvRowBytes)
-        // * decoder->image alpha data (alphaRange, alphaPlane, alphaRowBytes)
+        // * decoder->image alpha data (alphaPlane, alphaRowBytes)
         // * this frame's sequence timing
 
         avifRGBImageSetDefaults(&rgb, decoder->image);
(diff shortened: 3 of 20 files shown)


jzern:
lgtm. Just one update for the commit message: Eventhough -> Even though...

wantehchang (approved):
LGTM. Thanks. I made some minor edits in src/read.c.

Jamaika1:
I cann't merge libavif in libwebp2.
c:/msys1201/bin/../lib/gcc/x86_64-w64-mingw32/12.0.1/../../../../x86_64-w64-mingw32/bin/ld.exe: avif\read.o:read.c:(.text+0x32df): undefined reference to `avifLimitedToFullY'
c:/msys1201/bin/../lib/gcc/x86_64-w64-mingw32/12.0.1/../../../../x86_64-w64-mingw32/bin/ld.exe: avif\read.o:read.c:(.text+0x33f8): undefined reference to `avifLimitedToFullY'
collect2.exe: error: ld returned 1 exit status

vigneshvg:
> I cann't merge libavif in libwebp2. c:/msys1201/bin/../lib/gcc/x86_64-w64-mingw32/12.0.1/../../../../x86_64-w64-mingw32/bin/ld.exe: avif\read.o:read.c:(.text+0x32df): undefined reference to `avifLimitedToFullY' c:/msys1201/bin/../lib/gcc/x86_64-w64-mingw32/12.0.1/../../../../x86_64-w64-mingw32/bin/ld.exe: avif\read.o:read.c:(.text+0x33f8): undefined reference to `avifLimitedToFullY' collect2.exe: error: ld returned 1 exit status

I didn't try the windows build after this patch. I will take a look at this issue.

jzern:
@Jamaika1 Thanks for the report. I'm not able to reproduce the issue under msys2 with a gcc 11.3.0 toolchain. Can you give the set of commands you're using?

Jamaika1:
I use gcc 12.0.1 20220429. http://msystem.waw.pl/x265/mingw-gcc1201-20220429.7z
I use plugins for gcc thread/mutex/future/condition_variable
https://github.com/Jamaika1/mingw_std_threads

opensource libwebp2
https://www.sendspace.com/file/z639jj

jzern:
Thanks, can you share the cmake commands you used?

> I use gcc 12.0.1 20220429. http://msystem.waw.pl/x265/mingw-gcc1201-20220429.7z I use plugins for gcc thread/mutex/future/condition_variable https://github.com/Jamaika1/mingw_std_threads
> 
> opensource libwebp2 https://www.sendspace.com/file/z639jj

I used https://chromium.googlesource.com/codecs/libwebp2, I'm assuming the rar is a copy of that.

Jamaika1:
RAR is a copy of that and plugins. You create codec via .bat file.

jzern:
The batch file has include path issues, which are easily fixed and then runs into a command line length limitation. I would suggest using cmake to build the intermediate libraries rather than listing all of the .o's manually. In this change use of avifLimitedToFullY was added to read.c, but you may be missing reformat.o in one or more of your command lines.

Jamaika1:
Nothing is missing. As you compiled what I put in rar, you can see that the avifenc/dec file is creating and rd_curve/libwebp2 is not.

jzern:
I didn't mean anything was missing, just that the compile failed with webp2.bat as it was. There's a pattern like 'cd dir; for ... gcc' this needed to add -I<top-level>/include. Later the link would fail in a command prompt due to overlong command lines. Either way, I suggest checking the .o's in the link line or switching to building the intermediate libraries with cmake, which works for me.

Jamaika1:
At first it will fail.
You need to copy the GCC to the msys1201 directory on the C drive and save it to a bat file
some files from rar include you need to copy to gcc msys1201\x86_64-w64-mingw32\include
the second files from rar include must be copied to gcc msys1201\include\c ++\12.0.1
It should work.

jzern:
Those aren't the includes I was referring to, I meant the include path to files like aom_config.h, etc. Since you cd into the directories the top-level paths aren't visible without adding -I<top-level>/include to the command lines in the batch file.

As for the object files I mentioned, if you look at the command lines that produce rd_curve.exe and av1enc.exe they have avif\read.o (which uses `avifLimitedToFullY` as of this change), but do not have avif\reformat.o (which defines the function). avifdec.exe does have this and I'm assuming links all right.

PR #928, opened 2022-05-02
Free metadata upon avifenc exit

y-guyon:


wantehchang (approved):
LGTM. Thanks for the fix!

PR #929, opened 2022-05-02
Add clang-format-check GitHub action

y-guyon:


.github/workflows/clang-format-check.yml
```diff
@@ -0,0 +1,12 @@
...
+on: [push]
+jobs:
+  clang-format-checking:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: RafikFarhad/clang-format-github-action@v2.1.0
+        with:
+          style: file
+          sources: "apps/**/*.h,apps/**/*.c,examples/*.c,includes/**/*.h,src/**/*.h,src/**/*.c,tests/**/*.h,tests/**/*.c"
```
wantehchang, line 11: Should we add `tests/**/*.cc` now? Or would you like to wait until libavif actually has C++ files?

wantehchang (approved):
LGTM. Thanks.

.github/workflows/clang-format-check.yml
y-guyon: We can wait, it does not matter.

PR #930, opened 2022-05-03
Avoid decoding partial samples in avifDecoderDecodeTiles()

y-guyon:
Testing !data in avifDecoderDecodeTiles() is not enough because
samples gathered from merged extents may be partially available.
This behavior happens with incremental decoding of progressive
images, for example.


wantehchang (approved):
LGTM.

PR #932, opened 2022-05-04
Increase avifincrtest API coverage

y-guyon:
Test sizeHint and persistent fields from avifIO struct.

### DIFF
--- a/tests/avifincrtest_helpers.c
+++ b/tests/avifincrtest_helpers.c
@@ -110,36 +110,41 @@ static uint32_t getMinDecodedRowCount(uint32_t height, uint32_t cellHeight, avif
 
 //------------------------------------------------------------------------------
 
+typedef struct
+{
+    avifROData available;
+    size_t fullSize;
+} avifROPartialData;
+
 // Implementation of avifIOReadFunc simulating a stream from an array.
-// The full array size must be in io->sizeHint and io->data must be the available avifROData.
+// io->data is expected to point to avifROPartialData.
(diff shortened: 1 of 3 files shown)


jzern (approved):

tests/avifincrtest_helpers.c
```diff
@@ -331,11 +343,11 @@ avifBool decodeIncrementally(const avifRWData * encodedAvif, const avifImage * r
     // Parsing is not incremental.
     avifResult parseResult = avifDecoderParse(decoder);
     while (parseResult == AVIF_RESULT_WAITING_ON_IO) {
-        if (availableEncodedAvif.size >= encodedAvif->size) {
+        if (data.available.size >= data.fullSize) {
             printf("ERROR: avifDecoderParse() returned WAITING_ON_IO instead of OK\n");
             goto cleanup;
         }
-        availableEncodedAvif.size = availableEncodedAvif.size + 1;
+        data.available.size = data.available.size + 1;
```
wantehchang, line 350: Nit: This doesn't really matter. This can be written as data.available.size++. I was just wondering if it was deliberately written this way (perhaps for style reasons).

wantehchang (approved):
LGTM.

tests/avifincrtest_helpers.c
y-guyon: You are right, I tend to avoid `++` for quantity-related vars to hint that it is not an iterator nor a counter. I would have used `+= 1` but I did not remember if it was valid in C I guess.

Issue #826, opened 2022-01-26
Update the handling of 'lsel' and progressive decoding


leo-barnes:
Once https://github.com/AOMediaCodec/av1-avif/issues/180 lands, libavif will need to update how it handles progressive decoding.

wantehchang:
I wrote pull request https://github.com/AOMediaCodec/libavif/pull/939 to fix this.

PR #939, opened 2022-05-11
Update handling of 'lsel' and progressive decoding

wantehchang:
Previously, progressive decoding of a layered image was allowed if
'lsel' was absent. Now, progressivg decoding is allowed if the layer_id
value in 'lsel' is set to 0xFFFF.

Fix https://github.com/AOMediaCodec/libavif/issues/826.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -170,7 +170,7 @@ typedef struct avifDecoderItem
     uint32_t premByID;             // if non-zero, this item is premultiplied by Item #{premByID}
     avifBool hasUnsupportedEssentialProperty; // If true, this item cites a property flagged as 'essential' that libavif doesn't support (yet). Ignore the item, if so.
     avifBool ipmaSeen;    // if true, this item already received a property association
-    avifBool progressive; // if true, this item has progressive layers (a1lx), but does not select a specific layer (lsel)
+    avifBool progressive; // if true, this item has progressive layers (a1lx), but does not select a specific layer (the layer_id value in lsel is set to 0xFFFF)
 } avifDecoderItem;
 AVIF_ARRAY_DECLARE(avifDecoderItemArray, avifDecoderItem, item);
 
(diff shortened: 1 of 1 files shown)


jzern:
> Now, progressivg decoding...

progressive

> Previously, progressive decoding of a layered image was allowed if
> 'lsel' was absent

From the updated text absent and layer_id == 0xffff are treated equivalently for backward compatibility, correct?

wantehchang:
> > Now, progressivg decoding...
> 
> progressive

Fixed. Thanks!

> > Previously, progressive decoding of a layered image was allowed if
> > 'lsel' was absent
> 
> From the updated text absent and layer_id == 0xffff are treated equivalently for backward compatibility, correct?

In the current commit, I do not treat absent and layer_id == 0xffff equivalently. I follow the final v1.1.0 spec and require layer_id == 0xffff for a layered image to be decoded progressively.

HEIF requires a layered image to have 'lsel'. So a layered image with 'lsel' absent is not compliant with HEIF. This is why the final v1.1.0 spec switched to layer_id == 0xffff.

I am tempted to treat absent and layer_id == 0xffff equivalently for backward compatibility, until the test images have been updated to conform to the final v1.1.0 spec. Do you think this is a good idea?

src/read.c
```diff
@@ -535,8 +535,9 @@ static avifBool avifCodecDecodeInputFillFromDecoderItem(avifCodecDecodeInput * d
     }
 
     const avifProperty * lselProp = avifPropertyArrayFind(&item->properties, "lsel");
-    item->progressive = (a1lxProp && !lselProp); // Progressive images offer layers via the a1lxProp, but don't specify a layer selection with lsel.
-    if (lselProp) {
+    // Progressive images offer layers via the a1lxProp, but don't specify a layer selection with lsel.
+    item->progressive = (a1lxProp && lselProp && (lselProp->u.lsel.layerID == 0xFFFF));
```
wantehchang, line 539: James: If we want to treat absent and layer_id == 0xffff equivalently for backward compatibility, we just need to change this line to:
```
    item->progressive = (a1lxProp && (!lselProp || (lselProp->u.lsel.layerID == 0xFFFF)));
```

jzern: I would lean towards compatibility. This change on this line made me wonder. I didn't read through the current spec, but was basing my comment on the linked diff.

jzern: In talking about this over chat, we're not mandating a lsel box in this change, just potentially changing the rendering so this is ok if we wanted to deal with the possibility of bug reports or questions about why particular files aren't being rendered consistently.

jzern (approved):

PR #941, opened 2022-05-13
Move code avifEncoderDataCreate*Item() in write.c

y-guyon:


wantehchang (approved):
LGTM. I verified that the code was copied verbatim to the two new functions. The only new code is the checks for avifEncoderDataCreateItem() failure.

PR #942, opened 2022-05-13
Add avifmetadatatest

y-guyon:


tests/avifmetadatatest.c
```diff
@@ -0,0 +1,221 @@
...
+int main(void)
+{
+    if (!encodeDecodeMetadataItems(/*useICC=*/AVIF_TRUE, /*useExif=*/AVIF_FALSE, /*useXMP=*/AVIF_FALSE) ||
+        !encodeDecodeMetadataItems(/*useICC=*/AVIF_FALSE, /*useExif=*/AVIF_TRUE, /*useXMP=*/AVIF_FALSE) ||
+        !encodeDecodeMetadataItems(/*useICC=*/AVIF_FALSE, /*useExif=*/AVIF_FALSE, /*useXMP=*/AVIF_TRUE)) {
+        return EXIT_FAILURE;
+    }
+    // TODO: Negative test
+    // TODO: Multi xmp test
+    return EXIT_SUCCESS;
```
wantehchang, line 220: Nit: Print a message to indicate the test passed. Otherwise one needs to check the exit status.

tests/avifmetadatatest.c
```diff
@@ -0,0 +1,221 @@
...
+
+//------------------------------------------------------------------------------
+
+// Encodes the image. Returns false in case of failure.
+static avifBool encode(const avifImage * image, avifRWData * output)
+{
+    avifBool success = AVIF_FALSE;
+    avifEncoder * encoder = NULL;
+
+    encoder = avifEncoderCreate();
```
wantehchang, line 95: Nit/optional: lines 93-95 can be merged:
```
    avifEncoder * encoder = avifEncoderCreate();
```

tests/avifmetadatatest.c
```diff
@@ -0,0 +1,221 @@
...
+        row += rowBytes;
+    }
+}
+
+// Creates an image where the pixel values are defined but do not matter.
+// Returns false in case of memory failure.
+static avifBool createImage(int width, int height, int depth, avifPixelFormat yuvFormat, avifBool createAlpha, avifImage ** image)
+{
+    *image = avifImageCreate(width, height, depth, yuvFormat);
+    if (*image == NULL) {
```
wantehchang, line 56: Nit: In libavif, the concise form of null pointer checks such as `!ptr` and `ptr` are more common than the explicit form `ptr == NULL` and `ptr != NULL`. Please try to use the concise form.

tests/avifmetadatatest.c
```diff
@@ -0,0 +1,221 @@
...
+                                     0x67, 0x69, 0x6e, 0x3d, 0x22, 0xef, 0xbb, 0xbf, 0x22, 0x20, 0x69, 0x64 };
+static const size_t sampleXMPSize = sizeof(sampleXMP) / sizeof(sampleXMP[0]);
+
+//------------------------------------------------------------------------------
+// TODO: Move these functions to a tests/common/helper.c shared with avifgridapitest.c
+
+// Fills a plane with a repeating gradient.
+static void fillPlane(int width, int height, int depth, uint8_t * row, uint32_t rowBytes)
+{
+    assert(depth == 8 || depth == 10 || depth == 12); // Values allowed by AV1.
```
wantehchang, line 37: Note: In the past I've asked people to parenthesize subexpressions when comparisons are involved, to imitate the style used by Joe Drago. For example, in that style this line would be written as follows:
```
    assert((depth == 8) || (depth == 10) || (depth == 12)); // Values allowed by AV1.
```


Since it is easy to forget to do this, Joe and I have decided that we should be more lenient about this.

wantehchang (changes requested):
Overall this looks good. I suggest a few simple changes.

PR #944, opened 2022-05-13
Replace -Weverything by -Wall -Wextra for Clang

y-guyon:
These warnings are enough and -Weverything triggers too many errors in C++ files (notably GoogleTest).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,22 +122,7 @@ endif()
 include(CheckCCompilerFlag)
 if(CMAKE_C_COMPILER_ID MATCHES "Clang")
     message(STATUS "libavif: Enabling warnings for Clang")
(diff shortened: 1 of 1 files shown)


vrabaud (approved):
ok as discussed with @wantehchang

wantehchang (commented):
LGTM. This pull request is correct but is missing the removal of the -Wno-xxx flags that are added conditionally. I guess you wanted to merge this PR as soon as possible so that you don't need to add -Wno-everything to the GoogleTest pull request.

I will take care of removing those conditonally-added -Wno-xxx flags.

Note: Another thing we need to do is to add back some warning flags that are enabled by -Weverything. In particular, it would be good to add the warning flags used by Chrome's build system and Google's internal build system.

PR #950, opened 2022-05-17
Set our own default cfg.rc_end_usage value

wantehchang:
Set our own default cfg.rc_end_usage value if it differs from libaom's
default.

This change allows users to remove "-a end-usage=q" from their avifenc
command line.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -591,6 +591,27 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             return AVIF_RESULT_UNKNOWN_ERROR;
         }
 
+        // Set our own default cfg.rc_end_usage value, which may differ from libaom's default.
+        switch (aomUsage) {
+            case AOM_USAGE_GOOD_QUALITY:
+                // libaom's default is AOM_VBR. Change the default to AOM_Q since we don't need to
+                // hit a certain target bit rate. It's easier to control the worst quality in Q
+                // mode.
+                assert(cfg.rc_end_usage == AOM_VBR);
+                cfg.rc_end_usage = AOM_Q;
+                break;
+            case AOM_USAGE_REALTIME:
+                // For real-time mode we need to use CBR rate control mode. AOM_Q doesn’t fit the
+                // rate control requirements for real-time mode. CBR does.
+                assert(cfg.rc_end_usage == AOM_CBR);
+                break;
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -591,6 +591,27 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+        switch (aomUsage) {
+            case AOM_USAGE_GOOD_QUALITY:
+                // libaom's default is AOM_VBR. Change the default to AOM_Q since we don't need to
+                // hit a certain target bit rate. It's easier to control the worst quality in Q
+                // mode.
+                assert(cfg.rc_end_usage == AOM_VBR);
+                cfg.rc_end_usage = AOM_Q;
+                break;
+            case AOM_USAGE_REALTIME:
+                // For real-time mode we need to use CBR rate control mode. AOM_Q doesn’t fit the
```
jzern, line 604: you have a non-ascii character in doesn't (’)

jzern (commented):
Though the defaults are pretty stable, the assert might be fragile if a libaom was released and paired with an older libavif.

wantehchang:
> Though the defaults are pretty stable, the assert might be fragile if a libaom was released and paired with an older libavif.

James, do you think we should not rely on libaom's default cfg.rc_end_usage value here?

Since cfg.rc_end_usage is a fundamental config setting, I don't think libaom can change its default cfg.rc_end_usage value except in a major release. But it is certainly cheap to just go ahead and set our own default value.

jzern:
> > Though the defaults are pretty stable, the assert might be fragile if a libaom was released and paired with an older libavif.
> 
> James, do you think we should not rely on libaom's default cfg.rc_end_usage value here?
> 
> Since cfg.rc_end_usage is a fundamental config setting, I don't think libaom can change its default cfg.rc_end_usage value except in a major release. But it is certainly cheap to just go ahead and set our own default value.

I agree it's unlikely to happen, but I think it's ok to be explicit here rather than take the default value for something we have a certain expectation for.

wantehchang:
> I agree it's unlikely to happen, but I think it's ok to be explicit here rather than take the default value for something we have a certain expectation for.

Done.

jzern (approved):

vigneshvg:
late for the review. but lgtm.

Issue #953, opened 2022-05-20
-a cq-level=QP is overridden if -a end-usage=q is not also specified


wantehchang:
src/codec_aom.c has some code that sets the aom-specific `cq-level` option for the user if the user does not set `-a cq-level=QP` and `-a end-usage=q`. This depends on the `endUsageSet` and `cqLevelSet `booleans being set correctly.

I found that we neglect to set `cqLevelSet ` if the aom library has the `aom_codec_set_option()` function. Since all recent versions of the aom library (v3.0.0 or later) have the `aom_codec_set_option()` function, this bug affects all recent versions of the aom library.

To reproduce this bug, run this command:

`avifenc --min 0 --max 63 -a cq-level=18 -a tune=ssim happy_dog.jpg happy_dog.avif`

src/codec_aom.c will set the `cq-level` option to (min + max) / 2 = (0 + 63) / 2 = 31, overriding the `-a cq-level=18` option from the user.

Note that if -q end-usage=q is passed, then things work correctly:

`avifenc --min 0 --max 63 -a end-usage=q -a cq-level=18 -a tune=ssim happy_dog.jpg happy_dog.avif`

Note: Since src/codec_aom.c uses the aom library's all-intra mode for still image encoding, and `-a end-usage=q` is the default for the all-intra mode, it is tempting to omit the `-a end-usage=q` option. Until this bug is fixed, the `-a end-usage=q` option cannot be omitted.

PR #954, opened 2022-05-20
Set the cqLevelSet boolean correctly

wantehchang:
Set the cqLevelSet boolean correctly when libaom has the
aom_codec_set_option() function.

Fix https://github.com/AOMediaCodec/libavif/issues/953.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -464,6 +464,9 @@ static avifBool avifProcessAOMOptionsPostInit(avifCodec * codec, avifBool alpha)
                                   aom_codec_error_detail(&codec->internal->encoder));
             return AVIF_FALSE;
         }
+        if (!strcmp(key, "cq-level")) {
+            codec->internal->cqLevelSet = AVIF_TRUE;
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -464,6 +464,9 @@ static avifBool avifProcessAOMOptionsPostInit(avifCodec * codec, avifBool alpha)
                                   aom_codec_error_detail(&codec->internal->encoder));
             return AVIF_FALSE;
         }
+        if (!strcmp(key, "cq-level")) {
+            codec->internal->cqLevelSet = AVIF_TRUE;
+        }
```
wantehchang, line 469: Right now we only set the `cqLevelSet` boolean correctly at lines 501-503 below, when libaom does not have the `aom_codec_set_option()` function. I don't know why I missed this spot.

Since libaom v3.0.0 or later has the `aom_codec_set_option()` function, this bug affects all recent versions of libaom.

jzern (approved):

PR #957, opened 2022-05-20
Add assertion for boolean parameter useNthImageApi

wantehchang:
Assert that the boolean parameter useNthImageApi of
decodeNonIncrementallyAndIncrementally() only takes the value of
AVIF_FALSE or AVIF_TRUE. This detects the mistake of passing the
useNthImageApi and cellHeight arguments in the wrong order to this
function.

### DIFF
--- a/tests/avifincrtest_helpers.c
+++ b/tests/avifincrtest_helpers.c
@@ -4,6 +4,7 @@
 #include "avifincrtest_helpers.h"
 #include "avif/avif.h"
 
+#include <assert.h>
 #include <inttypes.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


wantehchang:
Yannis: Please feel free to decline this pull request, especially if avifincrtest.c and avifincrtest_helpers.c will be converted to C++ imminently.

y-guyon:
> Yannis: Please feel free to decline this pull request, especially if avifincrtest.c and avifincrtest_helpers.c will be converted to C++ imminently.

This is planned but the assertion can be added in the meantime.

y-guyon (approved):

PR #964, opened 2022-06-15
C++ify and shorten avifgridapitest and avify4mtest

y-guyon:
Move code to aviftest_helpers for clarity and to be reused by other tests.


tests/CMakeLists.txt
```diff
@@ -56,12 +56,12 @@ if(AVIF_ENABLE_GTEST)
...
     endif()
 
-    add_executable(avifgridapitest avifgridapitest.cc)
+    add_executable(avifgridapitest avifgridapitest.cc aviftest_helpers.cc)
     target_link_libraries(avifgridapitest avif ${AVIF_PLATFORM_LIBRARIES} ${GTEST_BOTH_LIBRARIES})
     target_include_directories(avifgridapitest PRIVATE ${GTEST_INCLUDE_DIRS})
     add_test(NAME avifgridapitest COMMAND avifgridapitest)
 
-    add_executable(avify4mtest avify4mtest.cc)
+    add_executable(avify4mtest avify4mtest.cc aviftest_helpers.cc)
```
jzern, line 64: for a single file it might be overkill, but this could go into a convenience object lib

jzern (approved):
nice cleanup

y-guyon:
> Are you planning to update the C++ test files to conform to Google's C++ Style Guide?

It would require another call to `clang-format` with a different `.clang-format` config file but it can be done. See https://github.com/AOMediaCodec/libavif#formatting. Unfortunately IDEs such as CLion will not be able to pick the correct style guide to apply because they seek the closest `.clang-format` (in same directory, then in parent etc.). Maybe we can wait to do that once almost all tests are converted to C++, or move those to a separate directory from the C ones. What do you think?

wantehchang:
Yes, we should move the C++ test files to a separate directory so that they can use a different `.clang-format` config file. You can make these changes in the order most convenient for you.

PR #969, opened 2022-06-21
Move unit tests to tests/unit

y-guyon:
`aviftest.c` is a generic encode/decode test so it can stay in `tests/`.
`avifyuv.c` is more a conversion benchmarking tool than a unit test so it can stay in `tests/`.
Move other test files to `tests/unit/`. It will allow to use a separate `.clang-tidy` config file once all files in `tests/unit` are converted to C++. This was suggested in # 964.


maryla-uc (approved):

tests/CMakeLists.txt
```diff
@@ -5,36 +5,42 @@
 # at once with CMake's ctest command line tool from the build folder.
 enable_testing()
 
+################################################################################
+# General tests and tools
```
wantehchang, line 9: We can consider moving this CMakeLists.txt file to the unit tests directory and moving aviftest.c and avifyuv.c back to the top-level CMakeLists.txt file.

I seem to remember one reason we added this CMakeLists.txt file was to allow us to compile the C++ test files with different compiler flags.

tests/CMakeLists.txt
```diff
@@ -5,36 +5,42 @@
...
-add_executable(avifincrtest avifincrtest.c avifincrtest_helpers.c)
+add_executable(avifyuv avifyuv.c)
+if(AVIF_LOCAL_LIBGAV1)
+    set_target_properties(avifyuv PROPERTIES LINKER_LANGUAGE "CXX")
+endif()
+target_link_libraries(avifyuv avif ${AVIF_PLATFORM_LIBRARIES})
+foreach(AVIFYUV_MODE limited rgb) # Modes drift and premultiply take more than 2 minutes each so they are disabled.
+    add_test(NAME avifyuv_${AVIFYUV_MODE} COMMAND avifyuv -m ${AVIFYUV_MODE})
+endforeach()
+
```
wantehchang, line 26: Optional: The `if(AVIF_ENABLE_COVERAGE)` part at lines 78-91 probably could be moved to this section because it seems to depend on `aviftest` only.

tests/CMakeLists.txt
```diff
@@ -5,36 +5,42 @@
...
+################################################################################
+# Unit tests
```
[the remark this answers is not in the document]
y-guyon: Done at https://github.com/AOMediaCodec/libavif/pull/972

PR #970, opened 2022-06-21
Use CMake OBJECT libraries instead of STATIC

y-guyon:
Suggested at https://github.com/AOMediaCodec/libavif/pull/964#discussion_r898560739 and at https://github.com/AOMediaCodec/libavif/pull/865#issuecomment-1057447356.


jzern (approved):
In the description of the change you could mention that now that the minimum cmake version has been increased it's ok to use this feature. It may help in people reading the git log to understand the connection.

PR #971, opened 2022-06-21
Use same CMake version in CI and CMakeLists.txt

y-guyon:
Make sure it works at the specified cmake_minimum_required version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,12 @@ jobs:
       with:
         path: ext
         key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}
+    - name: Setup cmake
+      uses: jwlawson/actions-setup-cmake@v1.12
+      with:
+        cmake-version: '3.13.x'
+    - name: Print cmake version
+      run: cmake --version
     - uses: ilammy/setup-nasm@v1
       with:
         version: 2.15.05


.github/workflows/ci.yml
```diff
@@ -21,6 +21,12 @@ jobs:
       with:
         path: ext
         key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}
+    - name: Setup cmake
+      uses: jwlawson/actions-setup-cmake@v1.12
+      with:
+        cmake-version: '3.13.x'
+    - name: Print cmake version
+      run: cmake --version
```
wantehchang, line 29: I verified that the code you added follows the example in https://github.com/jwlawson/actions-setup-cmake.

I am curious as to how you discovered jwlawson/actions-setup-cmake or determined it to be stable enough for our use. Is it well known?

wantehchang (approved):
LGTM.

.github/workflows/ci.yml
y-guyon: > I am curious as to how you discovered jwlawson/actions-setup-cmake

I searched for "github action set cmake version" on Google.

> determined it to be stable enough for our use. Is it well known?

I did not assess the stability of the module but it only impacts the CI, it uses a fixed version of the module that should not change in the future, and printing `cmake --version` reassured me. It can also be easily removed if any issue occurs.

PR #975, opened 2022-06-27
Add avifrgbtoyuvtest to test RGB-YUV subsampling

y-guyon:
The current tests/avifyuv.c tool only displays 4:4:4 conversions.
Adding a convenient C++ test for chroma subsampling loss evaluation
will help when working on faster alternatives to the current slow
RGB-to-YUV conversion in avifenc.

Related: https://bugs.chromium.org/p/libyuv/issues/detail?id=936


y-guyon:
`continuous-integration/appveyor/pr` failed with

> Build execution time has reached the maximum allowed time for your plan (60 minutes).

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -0,0 +1,350 @@
...
+    const avifMatrixCoefficients matrixCoefficients = std::get<5>(GetParam());
+    // Whether to add noise to the input RGB samples. Should only impact subsampled chroma (4:2:2 and 4:2:0).
+    const bool plainColor = std::get<6>(GetParam());
+    // Testing each RGB combination would be more accurate but results are similar with faster settings.
+    const uint32_t rgbStep = std::get<7>(GetParam());
+    // Thesholds to pass.
+    const double maxAbsAverageDiff = std::get<8>(GetParam());
+    const double minPsnr = std::get<9>(GetParam());
+    // Deduced constants.
+    const bool isMonochrome = yuvFormat == AVIF_PIXEL_FORMAT_YUV400; // If true, only test greyish input.
```
jzern, line 116: Are the tests adopting the () style used in the rest of the codebase? Should this be '(yuvFormat == AVIF_PIXEL_FORMAT_YUV400)'?

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -0,0 +1,350 @@
...
+    // Stats and thresholds.
+    // Note: The thresholds defined in this test are calibrated for libyuv fast paths. See reformat_libyuv.c.
+    //       Slower non-libyuv conversions in libavif have a higher precision (using floating point operations for example).
+    const double averageDiff = static_cast<double>(diffSum) / static_cast<double>(numDiffs);
+    const double averageAbsDiff = static_cast<double>(absDiffSum) / static_cast<double>(numDiffs);
+    const double psnr = getPsnr(sqDiffSum, numDiffs, rgbMax);
+    EXPECT_LE(std::abs(averageDiff), maxAbsAverageDiff);
+    EXPECT_GE(psnr, minPsnr);
+
+    // Printing stats for convenience and easier threshold tuning.
```
jzern, line 192: Maybe, "Print stats..."

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -0,0 +1,350 @@
...
+    const int yuvDepth = std::get<1>(GetParam());
+    const avifRGBFormat rgbFormat = std::get<2>(GetParam());
+    const avifPixelFormat yuvFormat = std::get<3>(GetParam());
+    const avifRange yuvRange = std::get<4>(GetParam());
+    const avifMatrixCoefficients matrixCoefficients = std::get<5>(GetParam());
+    // Whether to add noise to the input RGB samples. Should only impact subsampled chroma (4:2:2 and 4:2:0).
+    const bool plainColor = std::get<6>(GetParam());
+    // Testing each RGB combination would be more accurate but results are similar with faster settings.
+    const uint32_t rgbStep = std::get<7>(GetParam());
+    // Thesholds to pass.
```
jzern, line 112: Thresholds

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -0,0 +1,350 @@
...
+    const double maxAbsAverageDiff = std::get<8>(GetParam());
+    const double minPsnr = std::get<9>(GetParam());
+    // Deduced constants.
+    const bool isMonochrome = (yuvFormat == AVIF_PIXEL_FORMAT_YUV400); // If true, only test greyish input.
+    const uint32_t rgbMax = (1 << rgbDepth) - 1;
+
+    // The YUV upsampling treats the first and last rows and columns differently than the remaining pairs of rows and columns.
+    // An image of 16 pixels is used to test all these possibilities.
+    const int width = 4;
+    const int height = 4;
```
jzern, line 122: these could be constexpr int kWidth (assuming that naming convention is common)

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -0,0 +1,350 @@
...
+    testutil::avifRGBImageCleaner dstRgb(yuv.get(), rgbDepth, rgbFormat);
+
+    // Alpha values are not tested here. Keep it opaque.
+    if (avifRGBFormatHasAlpha(srcRgb.format)) {
+        testutil::fillImageChannel(&srcRgb, AVIF_CHAN_A, rgbMax);
+    }
+
+    // To exercise the YUV subsampling loss, the input samples must differ in each of the RGB channels.
+    // YUV subsampling expects the input RGB channels to be correlated to minimize the quality loss.
+    const int32_t redModifier[] = { 7,  14, 11, 5,  // Random permutation of 16 values.
```
jzern, line 136: these could be static constexpr

jzern (approved):
You may want to hold this until Wan-Teh can comment.

tests/gtest/avifrgbtoyuvtest.cc
y-guyon: Naming convention for constants in libavif seems to be [lowercase](https://github.com/AOMediaCodec/libavif/blob/e6368422d891ca940ed9f77a6c5e80d26191f280/src/write.c#L33). It may change to kSomething in C++ files when the Google C++ style guide is applied to them.

jzern (approved):

wantehchang (changes requested):
Yannis: I only reviewed the changes to the public header avif.h and related changes.

include/avif/avif.h
```diff
@@ -94,7 +94,7 @@ typedef enum avifPlanesFlag
 } avifPlanesFlag;
 typedef uint32_t avifPlanesFlags;
 
-enum avifChannelIndex
+typedef enum avifChannelIndex
```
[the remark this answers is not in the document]
y-guyon: - I removed R,G,B,A from `avifChannelIndex`.
- I added the generic `avifChannel` enum in `aviftest_helpers.h`. For now it contains just R,G,B,A but it can be extended to Y,U,V when needed.
- I moved `avifChannelOffset()` to `aviftest_helpers.h`. Note that I find "offset" less confusing because "avifChannelIndex" already exists as en enum.

`include/avif/internal.h` is currently inaccessible from tests.

tests/gtest/aviftest_helpers.h
```diff
@@ -24,12 +24,20 @@ class avifRWDataCleaner : public avifRWData
     ~avifRWDataCleaner() { avifRWDataFree(this); }
 };
 
+class avifRGBImageCleaner : public avifRGBImage
```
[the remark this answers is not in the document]
y-guyon: For structs that have a destroying function such as `avifImage`, `unique_ptr` can be used directly. `avifRWData` and `avifRGBImage` do not though, so we could add one and use `unique_ptr` instead. However I do not see the point of dynamic allocation here besides consistency, but the public API is not consistent on this matter either.

If it is mostly the name and not the inheritance that is problematic, what do you suggest? `avifRGBImageWrapper`? `avifRGBImageAuto`? `AvifRGBImage`?

Replacing with a `unique_ptr` or renaming should be done in another PR.

include/avif/avif.h
```diff
@@ -546,6 +548,7 @@ typedef enum avifRGBFormat
     AVIF_RGB_FORMAT_ABGR
 } avifRGBFormat;
 AVIF_API uint32_t avifRGBFormatChannelCount(avifRGBFormat format);
+AVIF_API uint32_t avifRGBFormatChannelOffset(avifRGBFormat format, avifChannelIndex channel); // Maps from avifChannelIndex to byte position in avifRGBFormat.
```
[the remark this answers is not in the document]
y-guyon: Done.

> Note that this function returns an invalid value if `channel` is `AVIF_CHAN_A` but `format` does not have alpha.

There is an assertion for that.

tests/gtest/aviftest_helpers.h
```diff
@@ -13,6 +13,17 @@ namespace libavif
...
+enum avifChannel
+{
+    R = 0,
+    G = 1,
+    B = 2,
+    A = 3
+};
+
+// Maps from avifChannel to sample position in avifRGBFormat.
+uint32_t avifChannelOffset(avifRGBFormat format, avifChannel channel);
```
wantehchang, line 25: Please name this function `GetRGBChannelOffset`. The `avif` prefix is not necessary because this is a C++ function in the `libavif` namespace.

If I understand it correctly, the code in this pull request always get the R, G, B, A channel offsets at the same time. So we should consider a function that returns the four channel offsets in one shot, for example,

```
struct RGBChannelOffsets {
  uint8_t r;
  uint8_t g;
  uint8_t b;
  uint8_t a;
};

RGBChannelOffsets GetRGBChannelOffsets(avifRGBFormat format);
```

And then we pass `uint32_t channelOffset` instead of `avifChannel channel` to functions like `modifyImageChannel`, `setImageChannel`, and `fillImageChannel`. And we won't need the `avifChannel` enum type.

tests/gtest/aviftest_helpers.h
```diff
@@ -13,6 +13,17 @@ namespace libavif
 namespace testutil
 {
 
+enum avifChannel
```
wantehchang, line 16: I believe C++ allows us to use just `R`, `G`, `B`, `A`. We need to define this as an enum class to force the use of `avifChannel::R`, `avifChannel::G`, `avifChannel::B`, `avifChannel::A`.

tests/gtest/aviftest_helpers.cc
```diff
@@ -37,16 +67,16 @@ void fillImagePlain(avifImage * image, const uint32_t yuva[4])
     avifPixelFormatInfo info;
     avifGetPixelFormatInfo(image->yuvFormat, &info);
 
-    for (int c : AVIF_CHANS) {
-        uint8_t * row = (c == AVIF_CHAN_A) ? image->alphaPlane : image->yuvPlanes[c];
+    for (int c : AVIF_YUV_CHANS) {
```
wantehchang, line 70: Nit/optional: We can write this as a traditional for loop:

```
for (int c = 0; c < 4; c++) {
```

Since `c` is used to index the `image->yuvPlanes` and `image->yuvRowBytes` arrays, we must assume `c` is 0, 1, or 2 (when `c` is not the alpha channel).

src/avif.c
```diff
@@ -3,6 +3,7 @@
 
 #include "avif/internal.h"
 
+#include <assert.h>
```
wantehchang, line 6: Please undo the changes to this file.

wantehchang (changes requested):
Hi Yannis,

I analyzed how the `avifChannelOffset` function is used. I proposed an alternative design that returns the offsets of all channels in one shot.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -0,0 +1,387 @@
...
+                            int64_t* sq_diff_sum, int64_t* max_abs_diff) {
+  (a.depth <= 8) ? GetDiffSumAndSqDiffSum<uint8_t>(a, b, diff_sum, abs_diff_sum,
+                                                   sq_diff_sum, max_abs_diff)
+                 : GetDiffSumAndSqDiffSum<uint16_t>(
+                       a, b, diff_sum, abs_diff_sum, sq_diff_sum, max_abs_diff);
+}
+
+// Returns the Peak Signal-to-Noise Ratio from accumulated stats.
+double GetPsnr(double sq_diff_sum, double num_diffs, double max_abs_diff) {
+  if (sq_diff_sum == 0.) {
```
wantehchang, line 88: [Just a comment. No change requested.] I seem to remember there was a style cleanup effort in Chrome or Google's internal repository to change all floating-point constants of the form `ddd.` to `ddd.0`. Would be good to look into that.

tests/gtest/aviftest_helpers.cc
```diff
@@ -13,13 +13,39 @@ namespace testutil {
...
+                           avifRGBFormat rgbFormat) {
+  avifRGBImageSetDefaults(this, yuv);
+  depth = rgbDepth;
+  format = rgbFormat;
+  avifRGBImageAllocatePixels(this);
+}
+
+//------------------------------------------------------------------------------
+
+RgbChannelOffsets GetRgbChannelOffsets(avifRGBFormat format) {
```
wantehchang, line 31: Optional: This function could return a fixed value for `a`, such as 0, for formats that do not have an alpha channel.

tests/gtest/aviftest_helpers.h
```diff
@@ -23,6 +26,20 @@ class AvifRwData : public avifRWData {
...
+class AvifRgbImage : public avifRGBImage {
+ public:
+  AvifRgbImage(const avifImage* yuv, int rgbDepth, avifRGBFormat rgbFormat);
+  ~AvifRgbImage() { avifRGBImageFreePixels(this); }
+};
+
+//------------------------------------------------------------------------------
+// Samples and images
+
+struct RgbChannelOffsets {
```
wantehchang, line 38: Nit: Add a comment for this struct. We can just adapt the following comment from a previous version of this pull request:

```
// Maps from avifChannel to sample position in avifRGBFormat.
uint32_t avifChannelOffset(avifRGBFormat format, avifChannel channel);
```

wantehchang (approved):
LGTM. Thanks.

PR #980, opened 2022-06-30
Format C++ tests with Google style

y-guyon:
Add tests/gtest/.clang-format for that purpose.
Fix .github/workflows/clang-format-check.yml

### DIFF
--- a/tests/gtest/.clang-format
+++ b/tests/gtest/.clang-format
@@ -1,0 +1,4 @@
+---
+Language: Cpp
+BasedOnStyle: Google
+...
--- a/tests/gtest/.clang-format
+++ b/tests/gtest/.clang-format
@@ -1,4 +1,5 @@
 ---
 Language: Cpp
 BasedOnStyle: Google
+DerivePointerAlignment: false
 ...
--- a/tests/gtest/.clang-format
+++ b/tests/gtest/.clang-format
@@ -2,4 +2,5 @@
 Language: Cpp
 BasedOnStyle: Google
 DerivePointerAlignment: false
+PointerAlignment: Left
 ...
--- a/tests/gtest/avifincrtest.cc
+++ b/tests/gtest/avifincrtest.cc
@@ -21,17 +21,17 @@ namespace {
 //------------------------------------------------------------------------------
 
 // Used to pass the data folder path to the GoogleTest suites.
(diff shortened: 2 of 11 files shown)


tests/gtest/avifincrtest.cc
```diff
@@ -16,86 +15,97 @@ using testing::Bool;
...
-namespace
-{
+namespace libavif {
+namespace {
 
 //------------------------------------------------------------------------------
 
 // Used to pass the data folder path to the GoogleTest suites.
-const char * dataPath = nullptr;
+const char *dataPath = nullptr;
```
jzern, line 24: more common would be 'const char* dataPath'. The tool just took a guess here, but it's inconsistent with the formatting in tests/gtest/avifincrtest_helpers.h for example

jzern: As a one off you could force the option one way or another, then the tool should keep the files in sync

tests/gtest/.clang-format
```diff
@@ -0,0 +1,5 @@
+---
+Language: Cpp
+BasedOnStyle: Google
+DerivePointerAlignment: false
```
wantehchang, line 4: The example in https://clang.llvm.org/docs/ClangFormatStyleOptions.html also sets `PointerAlignment` to `Left`:

```
Language: Cpp
# Force pointers to the type for C++.
DerivePointerAlignment: false
PointerAlignment: Left
```

wantehchang (changes requested):
This looks good. We should also follow the naming conventions of the Google C++ Style Guide. I assume you are planning to rename the functions and variables in a separate pull request.

tests/gtest/.clang-format
y-guyon: Done but I believe `PointerAlignment: Left` is set by `BasedOnStyle: Google` already.

y-guyon:
> This looks good. We should also follow the naming conventions of the Google C++ Style Guide. I assume you are planning to rename the functions and variables in a separate pull request.

Yes.

vrabaud (approved):

tests/gtest/.clang-format
wantehchang: You are right. Would you like to remove `PointerAlignment: Left`?

PR #983, opened 2022-07-04
Use Google Style naming convention in tests/gtest

y-guyon:


maryla-uc (approved):

vrabaud (approved):

tests/gtest/aviftest_helpers.h
```diff
@@ -17,23 +17,23 @@ using avifEncoderPtr =
 using avifDecoderPtr =
     std::unique_ptr<avifDecoder, decltype(&avifDecoderDestroy)>;
 
-class avifRWDataCleaner : public avifRWData {
+class AvifRwData : public avifRWData {
```
wantehchang, line 20: `avifImagePtr`, `avifEncoderPtr`, and `avifDecoderPtr` at lines 14, 15, 17 above should be renamed `AvifImagePtr`, `AvifEncoderPtr`, and `AvifDecoderPtr`. Did you leave them unchanged as an exception to the naming convention?

y-guyon: I thought that they should match `avifImage`, and that `AvifImagePtr` would become `ImagePtr` because it is in the `libavif` namespace.

What do you prefer:
 - `libavif::testutil::AvifImagePtr`
 - `libavif::testutil::ImagePtr`
 - `libavif::testutil::AvifImageCleaner`
 - `libavif::testutil::ImageCleaner`
 - `libavif::testutil::Image`

y-guyon: > Is there an `AvifImage` class?

No.

> I probably would choose `AvifImagePtr`.

Done in https://github.com/AOMediaCodec/libavif/pull/985

> If you search for "std::unique_ptr<avifImage," in Google's internal Code Search, you can find how other Google teams deal with this.

I only found `AvifImage` and `AvifImageUniquePtr`.

Issue #787, opened 2021-11-03
`avifdec in.avif out.png` adds alpha channel even when the input doesn’t have one


da2x:
`avifdec any-image.avif output.png` always adds an alpha channel in the output PNG regardless of whether it’s needed or not.

```
Version: 0.9.2 (dav1d [dec]:0.9.1, aom [enc/dec]:v3.1.2, rav1e [enc]:0.4.1 (UNKNOWN), svt [enc]:0.8.6)
libyuv : unavailable
```

wantehchang:
@FrankGalligan I wrote a fix at https://github.com/AOMediaCodec/libavif/pull/1021.

@joedrago I found that PNG also has `PNG_COLOR_TYPE_GRAY` and `PNG_COLOR_TYPE_GRAY_ALPHA`, so AVIF monochrome (YUV 4:0:0) images could be encoded in those two PNG color types.

PR #984, opened 2022-07-06
Adding a basic test for lossless.

vrabaud:
This will be used when YCgCo-R is added next.


joedrago:
Please run clang-format on all files in your commit. Files outside of the gtest dir do not use Google's format style.

apps/shared/avifutil.c
```diff
@@ -220,6 +224,37 @@ avifAppFileFormat avifGuessFileFormat(const char * filename)
     return AVIF_APP_FILE_FORMAT_UNKNOWN;
 }
 
+avifAppFileFormat avifReadImage(const char * filename,
```
wantehchang, line 227: Is it possible to modify the `avifInputReadImage` function in avifenc.c to call this new function?

apps/shared/avifutil.c
```diff
@@ -220,6 +224,37 @@ avifAppFileFormat avifGuessFileFormat(const char * filename)
...
+                                avifPixelFormat requestedFormat, int requestedDepth,
+                                avifImage *image, uint32_t *outDepth) {
```
wantehchang, line 229: Please run clang-format. Use the clang-format command line in https://github.com/AOMediaCodec/libavif/blob/main/README.md.

tests/are_images_equal.cc
```diff
@@ -0,0 +1,34 @@
...
+    std::cerr
+        << "Wrong argument: ./are_image_equal file1 file2 ignore_alpha_flag"
+        << std::endl;
+    return 1;
+  }
+  avifImagePtr decoded0(avifImageCreateEmpty(), avifImageDestroy);
+  uint32_t depth0, depth1;
+  constexpr int kRequestedDepth = 8;
+  constexpr avifPixelFormat requestedFormat = AVIF_PIXEL_FORMAT_NONE;
+  avifReadImage(argv[1], requestedFormat, kRequestedDepth, decoded0.get(), &depth0);
```
wantehchang, line 23: Check the return value of `avifReadImage` (two occurrences)

tests/gtest/aviftest_helpers.h
```diff
@@ -33,7 +33,8 @@ void FillImagePlain(avifImage* image, const uint32_t yuva[4]);
 void FillImageGradient(avifImage* image);
 
 // Returns true if both images have the same features and pixel values.
-bool AreImagesEqual(const avifImage& image1, const avifImage& image2);
+bool AreImagesEqual(const avifImage &image1, const avifImage &image2,
+                    bool ignore_alpha = false);
```
wantehchang, line 37: Document the `ignore_alpha` parameter.

tests/are_images_equal.cc
```diff
@@ -0,0 +1,34 @@
+// Copyright 2022 Google LLC. All rights reserved.
```
wantehchang, line 1: Our current policy allows C++ code only in the tests/gtest directory. Since this file doesn't use GoogleTest, it may seem inappropriate to put this file in tests/gtest.

One solution is to move the `libavif::testutil::AreImagesEqual()` function to apps/shared/avifutil.h. We will need to convert it to a C function, and therefore won't be able to specify a default argument for the `ignore_alpha` parameter.

Another solution is to move this file to tests/gtest.

Note: We can also write the lossless test using GoogleTest. Then we don't need to add this file.

tests/are_images_equal.cc
```diff
@@ -0,0 +1,34 @@
...
+using libavif::testutil::avifImagePtr;
+
+int main(int argc, char **argv) {
+  if (argc != 4) {
+    std::cerr
+        << "Wrong argument: ./are_image_equal file1 file2 ignore_alpha_flag"
+        << std::endl;
+    return 1;
+  }
+  avifImagePtr decoded0(avifImageCreateEmpty(), avifImageDestroy);
```
wantehchang, line 19: Nit: Check if `decoded0` and `decoded1` are null pointers.

wantehchang (changes requested):

apps/shared/avifutil.c
y-guyon: [The clang-format-check GitHub action](https://github.com/vrabaud/libavif/runs/7217915080?check_suite_focus=true#step:4:10) did not catch the issue. There might be some configuration error, maybe because of the `**` patterns?

`/entrypoint.sh: line 44: local: 'apps/shared/iccjpeg.h': invalid variable name for name reference` also puzzles me.

tests/test_cmd.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# Basic calls.
+"${AVIFENC}" --version
+"${AVIFDEC}" --version
+
+# Lossless test.
+echo "Testing basic lossless"
+"${AVIFENC}" -s 10 -l "${PNG_FILE}" -o "${TMP_ENCODED_FILE}"
+"${AVIFDEC}" "${TMP_ENCODED_FILE}" "${DECODED_FILE}"
+"${ARE_IMAGES_EQUAL}" "${PNG_FILE}" "${DECODED_FILE}" 1
```
wantehchang, line 56: IMPORTANT: The `are_images_equal` program compares the two images after converting them to YUV. For lossless compression, it seems that we should compare the images in RGB form.

apps/shared/avifutil.c
vrabaud: Done

apps/shared/avifutil.c
vrabaud: dir/**.h does not take subdirectories.

vrabaud: Done

apps/shared/avifutil.h
```diff
@@ -46,6 +46,12 @@ typedef enum avifAppFileFormat
 
 avifAppFileFormat avifGuessFileFormat(const char * filename);
 
+// Reads an image from a file.
```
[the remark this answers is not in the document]
vrabaud: Done

tests/are_images_equal.cc
vrabaud: The test is about testing avifenc/avifdec, not the algorithms. Moving it to the gtest folder.

tests/are_images_equal.cc
```diff
@@ -0,0 +1,34 @@
...
+#include "gtest/aviftest_helpers.h"
+
+using libavif::testutil::avifImagePtr;
+
+int main(int argc, char **argv) {
+  if (argc != 4) {
+    std::cerr
+        << "Wrong argument: ./are_image_equal file1 file2 ignore_alpha_flag"
+        << std::endl;
+    return 1;
```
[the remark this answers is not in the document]
vrabaud: Done

tests/are_images_equal.cc
vrabaud: Done

tests/are_images_equal.cc
vrabaud: Done

tests/are_images_equal.cc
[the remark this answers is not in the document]
vrabaud: Indeed: I set the MC to identity so that it stays in RGB for PNG or YUV for JPEG.

tests/test_cmd.sh
vrabaud: Indeed, done.

tests/gtest/are_images_equal.cc
```diff
@@ -0,0 +1,54 @@
...
+                      NULL) == AVIF_APP_FILE_FORMAT_UNKNOWN) {
+      std::cerr << "Image " << argv[i + 1] << " cannot be read." << std::endl;
+      return 3;
+    }
+  }
+
+  if (depth[0] != depth[1]) {
+    std::cerr << "Images " << argv[1] << " and " << argv[2]
+              << " have different depth." << std::endl;
+    return 4;
```
wantehchang, line 43: It would be better to use the exit status 1 for "images are not equal", so that this exit status is independent of the number of errors we want to differentiate by exit status.

I think it is fine to use a single exit status for all errors. Your call.

tests/test_cmd.sh
wantehchang, line 56: It would be good to explain why alpha is ignored in this test.

tests/gtest/are_images_equal.cc
```diff
@@ -0,0 +1,54 @@
...
+
+#include "aviftest_helpers.h"
+#include "avifutil.h"
+
+using libavif::testutil::AvifImagePtr;
+
+int main(int argc, char** argv) {
+  if (argc != 4) {
+    std::cerr
+        << "Wrong argument: ./are_image_equal file1 file2 ignore_alpha_flag"
```
wantehchang, line 15: Nit: Change `./are_image_equal` to `argv[0]`:

```
        << "Wrong argument: " << argv[0] << " file1 file2 ignore_alpha_flag"
```

tests/gtest/are_images_equal.cc
```diff
@@ -0,0 +1,54 @@
...
+  }
+  AvifImagePtr decoded[2] = {
+      AvifImagePtr(avifImageCreateEmpty(), avifImageDestroy),
+      AvifImagePtr(avifImageCreateEmpty(), avifImageDestroy)};
+  if (!decoded[0] || !decoded[1]) {
+    std::cerr << "Cannot create AVIF images." << std::endl;
+    return 2;
+  }
+  uint32_t depth[2];
+  constexpr int kRequestedDepth = 8;
```
wantehchang, line 27: It seems that we should set kRequestedDepth to 0. 0 means "request the bit depth closest to the bit depth of the input file".

The `avifImage` cannot use the exact bit depth of the input fie when the input file is 16-bit PNG; we are forced to use 12-bit AVIF in that case.

For lossless test, if the bit depth of the `avifImage` is not equal to the bit depth of the input file, there is potentially loss from the `avifReadImage()` call.

tests/CMakeLists.txt
```diff
@@ -96,3 +98,15 @@ if(AVIF_ENABLE_GTEST)
 else()
     message(STATUS "Most tests are disabled because AVIF_ENABLE_GTEST is OFF.")
 endif()
+
+################################################################################
+# Bash tests
+
+if(AVIF_BUILD_APPS)
```
wantehchang, line 105: `AVIF_BUILD_APPS` means "build the executables in the apps/ directory".

Note: We can move are_images_equal.cc to the apps/ directory, but we will need to convert it to a C file, and we will also need to move `libavif::testutil::AreImagesEqual` to apps/shared/avifutil.{h,c} and convert it to a C function.

wantehchang (changes requested):

apps/avifenc.c
```diff
@@ -278,27 +278,14 @@ static avifAppFileFormat avifInputReadImage(avifInput * input, avifImage * image
...
-    } else {
-        fprintf(stderr, "Unrecognized file format: %s\n", input->files[input->fileIndex].filename);
+    const avifAppFileFormat nextInputFormat = avifReadImage(input->files[input->fileIndex].filename,
+                                                            input->requestedFormat,
+                                                            input->requestedDepth,
+                                                            image,
+                                                            outDepth,
+                                                            sourceTiming,
+                                                            &input->frameIter);
+    if (nextInputFormat == AVIF_APP_FILE_FORMAT_UNKNOWN) {
```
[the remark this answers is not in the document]
vrabaud: I'd rather keep the current format which is what is done above too: exit as soon as you can.

tests/CMakeLists.txt
```diff
@@ -42,9 +42,14 @@ endforeach()
 ################################################################################
 # GoogleTest
 
-if(AVIF_ENABLE_GTEST)
+if(AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
```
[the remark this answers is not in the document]
vrabaud: `AVIF_BUILD_TESTS` should not be used in this file as `AVIF_BUILD_TESTS` already controls whether we `add_subdirectory` this folder.
`are_images_equal` is, for now, only used to test avifenc/avifdec. Therefore, it is only used if the apps are built, and therefore is `AVIF_BUILD_APPS` is set.

If you want, we can decide that `AVIF_BUILD_TESTS` sets `AVIF_BUILD_APPS` to ON.

tests/CMakeLists.txt
vrabaud: I am adding a test to test the app so I indeed need to check for `AVIF_BUILD_APPS`

tests/gtest/are_images_equal.cc
```diff
@@ -0,0 +1,54 @@
...
+  constexpr avifPixelFormat requestedFormat = AVIF_PIXEL_FORMAT_NONE;
+  for (int i : {0, 1}) {
```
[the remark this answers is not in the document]
vrabaud: I wrote it this way because it is easier to understand to me, and shorter. With one glance, I know that it tests for 0 and 1, I do not have to think about what is the last value.

tests/test_cmd.sh
vrabaud: I am not ignoring it anymore. It will be ignored for a future test comparing with a JPEG: JPEG doe snot have alpha loaded, while PNG has it loaded (with full opacity though).

tests/gtest/are_images_equal.cc
```diff
@@ -0,0 +1,54 @@
...
+    // Make sure no color conversion happens.
+    decoded[i]->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_IDENTITY;
+    if (avifReadImage(argv[+1], requestedFormat, kRequestedDepth,
```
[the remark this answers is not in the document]
vrabaud: Done

tests/test_cmd.sh
wantehchang: I think I understand the issue now. We should enhance `libavif::testutil::AreImagesEqual` to treat an absent alpha plane and a fully-opaque alpha plane as equal. We can do that in a follow-up CL. The `avifImageIsOpaque()` function in libavif/src/write.c could be useful.

Also, generating a PNG with a fully-opaque alpha plane when the image has no alpha is a know issue: https://github.com/AOMediaCodec/libavif/issues/787. That issue should be straightforward to fix.

tests/gtest/are_images_equal.cc
```diff
@@ -0,0 +1,54 @@
...
+    std::cerr << "Wrong argument: " << argv[0]
+              << " file1 file2 ignore_alpha_flag" << std::endl;
+    return 2;
+  }
+  AvifImagePtr decoded[2] = {
+      AvifImagePtr(avifImageCreateEmpty(), avifImageDestroy),
+      AvifImagePtr(avifImageCreateEmpty(), avifImageDestroy)};
+  if (!decoded[0] || !decoded[1]) {
+    std::cerr << "Cannot create AVIF images." << std::endl;
+    return 1;
```
wantehchang, line 23: We should return an exit status >= 2 here because this is an error. (1 now means "images are not equal".)

tests/gtest/are_images_equal.cc
```diff
@@ -0,0 +1,54 @@
...
+  constexpr int kRequestedDepth = 0;
+  constexpr avifPixelFormat requestedFormat = AVIF_PIXEL_FORMAT_NONE;
+  for (int i : {0, 1}) {
+    // Make sure no color conversion happens.
+    decoded[i]->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_IDENTITY;
+    if (avifReadImage(argv[i + 1], requestedFormat, kRequestedDepth,
+                      decoded[i].get(), &depth[i], NULL,
+                      NULL) == AVIF_APP_FILE_FORMAT_UNKNOWN) {
+      std::cerr << "Image " << argv[i + 1] << " cannot be read." << std::endl;
+      return 1;
```
wantehchang, line 36: We should return an exit status >= 2 here because this is an error. (1 now means "images are not equal".)

tests/test_cmd.sh
```diff
@@ -0,0 +1,62 @@
...
+"${AVIFENC}" --version
+"${AVIFDEC}" --version
+
+# Lossless test.
+echo "Testing basic lossless"
+"${AVIFENC}" -s 10 -l "${PNG_FILE}" -o "${TMP_ENCODED_FILE}"
+"${AVIFDEC}" "${TMP_ENCODED_FILE}" "${DECODED_FILE}"
+"${ARE_IMAGES_EQUAL}" "${PNG_FILE}" "${DECODED_FILE}" 0
+set +e
+"${ARE_IMAGES_EQUAL}" "${TESTDATA_DIR}"/kodim23_yuv420_8bpc.y4m "${DECODED_FILE}" 0
```
wantehchang, line 58: It would be better to check if the exit status is equal to 1, right? We can check the `$?` variable.

wantehchang (changes requested):
Just some simple changes, otherwise this looks good.

tests/test_cmd.sh
vrabaud: Indeed, the flag should probably be something like: ignore_alpha_if_opaque

tests/gtest/are_images_equal.cc
```diff
@@ -0,0 +1,54 @@
...
+  uint32_t depth[2];
+  // Request the bit depth closest to the bit depth of the input file.
+  constexpr int kRequestedDepth = 0;
+  constexpr avifPixelFormat requestedFormat = AVIF_PIXEL_FORMAT_NONE;
+  for (int i : {0, 1}) {
+    // Make sure no color conversion happens.
+    decoded[i]->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_IDENTITY;
+    if (avifReadImage(argv[i + 1], requestedFormat, kRequestedDepth,
+                      decoded[i].get(), &depth[i], NULL,
+                      NULL) == AVIF_APP_FILE_FORMAT_UNKNOWN) {
```
jzern, line 34: NULL could be nullptr, though I don't know how consistent the c++ code is at the moment

tests/test_cmd.sh
```diff
@@ -0,0 +1,67 @@
...
+if [ $? -ne 1 ]
+then
```
jzern, line 60: join this with the line above and prefer [[ ]] there

tests/test_cmd.sh
```diff
@@ -0,0 +1,67 @@
...
+    echo "Image should be different"
+    exit 1
```
jzern, line 62: indent can be 2 to match the rest of the file

tests/test_cmd.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+set -e
+
+echo "TEST OK"
+exit 0
```
jzern, line 67: This isn't strictly necessary given the echo should succeed, but fine either way.

tests/gtest/are_images_equal.cc
wantehchang: Since this is a new file, let's use `nullptr` instead of `NULL`.

wantehchang (commented):
Vincent: LGTM. I will approve this pull request after you have addressed James's comments.

tests/test_cmd.sh
```diff
@@ -0,0 +1,67 @@
...
+# Basic calls.
+"${AVIFENC}" --version
+"${AVIFDEC}" --version
+
+# Lossless test.
+echo "Testing basic lossless"
+"${AVIFENC}" -s 10 -l "${PNG_FILE}" -o "${TMP_ENCODED_FILE}"
+"${AVIFDEC}" "${TMP_ENCODED_FILE}" "${DECODED_FILE}"
+"${ARE_IMAGES_EQUAL}" "${PNG_FILE}" "${DECODED_FILE}" 0
+set +e
```
wantehchang, line 57: James, Vincent: Is it possible to avoid temporarily setting `+e` for this test?

jzern: if "${ARE_IMAGES_EQUAL}" "${TESTDATA_DIR}"/kodim23_yuv420_8bpc.y4m "${DECODED_FILE}" 0; then
...
fi

wantehchang: How do we check if the exit status of that `${ARE_IMAGES_EQUAL}` command is equal to 1?

Exit status 1 means "images are not equal". Exit status 2 means "the command failed with an error". So this test expects the exit status to be equal to 1, rather than any nonzero value.

vrabaud: I set it as final: failing tests will be executed last.

wantehchang (approved):
LGTM.

PR #989, opened 2022-07-11
Add avifImageRGBToYUVLibYUV()

y-guyon:
Use libyuv for RGB to YUV conversion when possible. For now only 8-bit
BT.601 is handled and only for some RGB and YUV layouts.
The conversion itself can be several times faster than the current
built-in floating point conversion. For example, avifImageRGBToYUV()
took 1.6s (built-in) vs 0.2s (libyuv) during a local test.

libyuv needs to be compiled with LIBYUV_BIT_EXACT for limited range
conversion to be accurate enough, and without LIBYUV_BIT_EXACT for
full range conversion to be accurate enough. When respecting these
constraints, the loss is comparable to the precision loss between
built-in and libyuv YUV to RGB conversion, both being used by libavif
already. Full range is the default setting in avifImageSetDefaults()
and avifenc, so libavif does not recommend building libyuv with
LIBYUV_BIT_EXACT defined, in order to favor fast full range encoding.
A comment is added to ext/libyuv.cmd though.

See also https://bugs.chromium.org/p/libyuv/issues/detail?id=936.

This change does not impact the core library. Only the tooling dealing
with non-YUV color spaces should be faster in some cases. For example,
avifenc is expected to be up to 7% faster when encoding PNG images at
AVIF_SPEED_FASTEST for available libyuv conversions.

Summary using tests/gtest/avifrgbtoyuvtest results (8-bit BT.601),
libyuv compared to built-in conversion:
- YUV to RGB (LIBYUV_BIT_EXACT has no impact):
    Overall average drift increased by 0.03 on average, 0.08 max (change over absolute difference of average of samples)
    Quality loss is increased by 10% on average, 22% max (change over average of absolute sample value differences)
    Sample value difference is increased by 0.9 in average, up to 2 (change over max absolute sample value difference)
    PSNR drops by 0.6dB on average, 1.6dB max (on a test 4x4 noisy or plain pattern)
- RGB to YUV limited range (with LIBYUV_BIT_EXACT):
    Overall average drift increased by 0.03 on average, 0.09 max
    Quality loss is increased by 12% on average, 25% max
    Sample value difference is increased by 0.6 in average, up to 1
    PSNR drops by 0.8dB on average, 1.7dB max
- RGB to YUV full range (without LIBYUV_BIT_EXACT):
    Sample values are as centered (no bigger overall average drift)
    Quality loss is increased by 17% on average, 57% max
    Sample value difference is increased by 1.1 in average, up to 2
    PSNR drops by 1dB on average, 2.6dB max

Note: GitHub badly highlights the minor changes to `src/reformat.c`
because of the indent change.

### DIFF
--- a/ext/libyuv.cmd
+++ b/ext/libyuv.cmd
@@ -4,10 +4,6 @@
 
 : # cmake and ninja must be in your PATH.
 
-: # Add -DCMAKE_C_FLAGS="-DLIBYUV_BIT_EXACT" -DCMAKE_CXX_FLAGS="-DLIBYUV_BIT_EXACT" to the cmake command
-: # to enable faster RGB to YUV limited range BT.601 conversion and disable faster full range BT.601 conversion.
-: # See avifIsLibyuvBitExact() call sites in src/reformat_libyuv.c for further information.
-
 : # If you're running this on Windows, be sure you've already run this (from your VC2019 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
 
--- a/ext/libyuv.cmd
+++ b/ext/libyuv.cmd
@@ -4,6 +4,10 @@
 
 : # cmake and ninja must be in your PATH.
 
+: # Add -DCMAKE_C_FLAGS="-DLIBYUV_BIT_EXACT" -DCMAKE_CXX_FLAGS="-DLIBYUV_BIT_EXACT" to the cmake command
+: # to enable faster RGB to YUV limited range BT.601 conversion (disables faster full range BT.601 conversion).
+: # See avifIsLibyuvBitExact() call sites in src/reformat_libyuv.c for further information.
+
 : # If you're running this on Windows, be sure you've already run this (from your VC2019 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
 
--- a/ext/libyuv.cmd
+++ b/ext/libyuv.cmd
@@ -5,7 +5,7 @@
 : # cmake and ninja must be in your PATH.
 
 : # Add -DCMAKE_C_FLAGS="-DLIBYUV_BIT_EXACT" -DCMAKE_CXX_FLAGS="-DLIBYUV_BIT_EXACT" to the cmake command
-: # to enable faster RGB to YUV limited range BT.601 conversion (disables faster full range BT.601 conversion).
+: # to enable faster RGB to YUV limited range BT.601 conversion and disable faster full range BT.601 conversion.
 : # See avifIsLibyuvBitExact() call sites in src/reformat_libyuv.c for further information.
 
 : # If you're running this on Windows, be sure you've already run this (from your VC2019 install dir):
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -531,6 +531,18 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
 //   * x/x/[1|2|5|6|9]/Limited
 //   * [1|2|5|6|9]/x/12/Limited
 
+// If libavif is built with libyuv fast paths enabled, libavif will use libyuv for conversion from
+// RGB to YUV if the following requirements are met:
+//
+// * YUV depth: 8
+// * RGB depth: 8
+// * rgb.chromaDownsampling: AVIF_CHROMA_UPSAMPLING_AUTOMATIC, AVIF_CHROMA_UPSAMPLING_FASTEST
+// * One of the following combinations (avifRGBFormat to avifPixelFormat/Range):
+//   *  BGRA            to [YUV400|YUV420|YUV422]/[Full|Limited]
+//   *  BGRA            to  YUV444               / Limited
+//   *  BGR             to  YUV420               /[Full|Limited]
+//   * [RGBA|ARGB|ABGR] to  YUV420               / Limited
+
 typedef enum avifRGBFormat
 {
     AVIF_RGB_FORMAT_RGB = 0,
@@ -552,6 +564,14 @@ typedef enum avifChromaUpsampling
     AVIF_CHROMA_UPSAMPLING_BILINEAR = 4      // Uses bilinear filter (built-in)
 } avifChromaUpsampling;
 
+typedef enum avifChromaDownsampling
+{
+    AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC = 0,    // Chooses best trade off of speed/quality (prefers libyuv, else uses BEST_QUALITY)
+    AVIF_CHROMA_DOWNSAMPLING_FASTEST = 1,      // Chooses speed over quality (prefers libyuv, else uses AVERAGE)
+    AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY = 2, // Chooses the best quality upsampling (avoids libyuv, uses AVERAGE)
+    AVIF_CHROMA_DOWNSAMPLING_AVERAGE = 3       // Uses floating point RGB-to-YUV conversion then averaging (built-in)
+} avifChromaDownsampling;
+
 typedef struct avifRGBImage
 {
     uint32_t width;       // must match associated avifImage
@@ -560,6 +580,8 @@ typedef struct avifRGBImage
     avifRGBFormat format; // all channels are always full range
     avifChromaUpsampling chromaUpsampling; // Defaults to AVIF_CHROMA_UPSAMPLING_AUTOMATIC: How to upsample non-4:4:4 UV (ignored for 444) when converting to RGB.
                                            // Unused when converting to YUV. avifRGBImageSetDefaults() prefers quality over speed.
+    avifChromaDownsampling chromaDownsampling; // How to convert (and downsample to non-4:4:4 UV) when converting from RGB.
+                                               // Unused when converting to RGB. Defaults to AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC.
     avifBool ignoreAlpha;        // Used for XRGB formats, treats formats containing alpha (such as ARGB) as if they were
                                  // RGB, treating the alpha bits as if they were all 1.
     avifBool alphaPremultiplied; // indicates if RGB value is pre-multiplied by alpha. Default: false
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -466,7 +466,7 @@ typedef struct avifImage
 
 AVIF_API avifImage * avifImageCreate(int width, int height, int depth, avifPixelFormat yuvFormat);
 AVIF_API avifImage * avifImageCreateEmpty(void); // helper for making an image to decode into
-AVIF_API void avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes); // deep copy
+AVIF_API avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes); // deep copy
 AVIF_API avifResult avifImageSetViewRect(avifImage * dstImage, const avifImage * srcImage, const avifCropRect * rect); // shallow copy, no metadata
 AVIF_API void avifImageDestroy(avifImage * image);
 
@@ -476,8 +476,8 @@ AVIF_API void avifImageSetProfileICC(avifImage * image, const uint8_t * icc, siz
 AVIF_API void avifImageSetMetadataExif(avifImage * image, const uint8_t * exif, size_t exifSize);
 AVIF_API void avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_t xmpSize);
 
-AVIF_API void avifImageAllocatePlanes(avifImage * image, avifPlanesFlags planes); // Ignores any pre-existing planes
-AVIF_API void avifImageFreePlanes(avifImage * image, avifPlanesFlags planes);     // Ignores already-freed planes
+AVIF_API avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags planes); // Ignores any pre-existing planes
+AVIF_API void avifImageFreePlanes(avifImage * image, avifPlanesFlags planes);           // Ignores already-freed planes
 AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, avifPlanesFlags planes);
 
 // ---------------------------------------------------------------------------
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -569,7 +569,7 @@ typedef enum avifChromaDownsampling
     AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC = 0,    // Chooses best trade off of speed/quality (prefers libyuv, else uses BEST_QUALITY)
     AVIF_CHROMA_DOWNSAMPLING_FASTEST = 1,      // Chooses speed over quality (prefers libyuv, else uses AVERAGE)
     AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY = 2, // Chooses the best quality upsampling (avoids libyuv, uses AVERAGE)
-    AVIF_CHROMA_DOWNSAMPLING_AVERAGE = 3       // Uses floating point RGB-to-YUV conversion then averaging (built-in)
+    AVIF_CHROMA_DOWNSAMPLING_FLOAT_AVERAGE = 3 // Uses floating point RGB-to-YUV conversion then averaging (built-in)
 } avifChromaDownsampling;
 
 typedef struct avifRGBImage
@@ -580,7 +580,7 @@ typedef struct avifRGBImage
     avifRGBFormat format; // all channels are always full range
     avifChromaUpsampling chromaUpsampling; // Defaults to AVIF_CHROMA_UPSAMPLING_AUTOMATIC: How to upsample non-4:4:4 UV (ignored for 444) when converting to RGB.
                                            // Unused when converting to YUV. avifRGBImageSetDefaults() prefers quality over speed.
-    avifChromaDownsampling chromaDownsampling; // How to convert (and downsample to non-4:4:4 UV) when converting from RGB.
+    avifChromaDownsampling chromaDownsampling; // How to convert (and downsample to non-4:4:4 UV) when converting to YUV.
                                                // Unused when converting to RGB. Defaults to AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC.
     avifBool ignoreAlpha;        // Used for XRGB formats, treats formats containing alpha (such as ARGB) as if they were
                                  // RGB, treating the alpha bits as if they were all 1.
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -537,11 +537,12 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
 // * YUV depth: 8
 // * RGB depth: 8
 // * rgb.chromaDownsampling: AVIF_CHROMA_UPSAMPLING_AUTOMATIC, AVIF_CHROMA_UPSAMPLING_FASTEST
-// * One of the following combinations (avifRGBFormat to avifPixelFormat/Range):
-//   *  BGRA            to [YUV400|YUV420|YUV422]/[Full|Limited]
-//   *  BGRA            to  YUV444               / Limited
-//   *  BGR             to  YUV420               /[Full|Limited]
-//   * [RGBA|ARGB|ABGR] to  YUV420               / Limited
+// * One of the following combinations (avifRGBFormat to avifPixelFormat/MC/Range):
+//   *  BGRA            to  YUV400        /  x  /[Full|Limited]
+//   *  BGRA            to [YUV420|YUV422]/[5|6]/[Full|Limited]
+//   *  BGRA            to  YUV444        /[5|6]/ Limited
+//   *  BGR             to  YUV420        /[5|6]/[Full|Limited]
+//   * [RGBA|ARGB|ABGR] to  YUV420        /[5|6]/ Limited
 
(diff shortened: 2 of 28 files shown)


src/reformat_libyuv.c
```diff
@@ -50,6 +56,126 @@ unsigned int avifLibYUVVersion(void)
...
+                RGBtoYUV = (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) ? ABGRToI420 : NULL;
+            } else if (rgb->format == AVIF_RGB_FORMAT_ARGB) {
+                RGBtoYUV = (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) ? BGRAToI420 : NULL;
+            } else if (rgb->format == AVIF_RGB_FORMAT_BGR) {
+                RGBtoYUV = (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) ? RGB24ToI420 : NULL;
+            } else if (rgb->format == AVIF_RGB_FORMAT_BGRA) {
+                RGBtoYUV = (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV444)   ? ARGBToI444
+                           : (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV422) ? ARGBToI422
+                           : (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) ? ARGBToI420
+                                                                            : NULL;
```
wantehchang, line 142: This should definitely be rewritten as if-else-if-else or a switch statement:

```
            } else if (rgb->format == AVIF_RGB_FORMAT_BGRA) {
                if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV444)  {
                    RGBtoYUV = ARGBToI444;
                } else if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV422) {
                    RGBtoYUV = ARGBToI422;
                } else if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) {
                    RGBtoYUV = ARGBToI420;
                }
```

src/reformat_libyuv.c
```diff
@@ -50,6 +56,126 @@ unsigned int avifLibYUVVersion(void)
...
+    if ((image->depth == 8) && (rgb->depth == 8)) {
+        return avifImageRGBToYUVLibYUV8bpc(image, rgb);
+    }
+
+    // This function didn't do anything; use the built-in conversion.
+    return AVIF_RESULT_NOT_IMPLEMENTED;
+}
+
+// Returns whether libyuv was compiled with LIBYUV_BIT_EXACT defined or not.
+static avifBool avifIsLibyuvBitExact()
```
wantehchang, line 72: Please ask Frank Barchard (the author of libyuv) to review this function. He may be able to suggest a better way to detect if libyuv was compiled with LIBYUV_BIT_EXACT defined.

ext/libyuv.cmd
```diff
@@ -4,6 +4,10 @@
 
 : # cmake and ninja must be in your PATH.
 
+: # Add -DCMAKE_C_FLAGS="-DLIBYUV_BIT_EXACT" -DCMAKE_CXX_FLAGS="-DLIBYUV_BIT_EXACT" to the cmake command
+: # to enable faster RGB to YUV limited range BT.601 conversion (disables faster full range BT.601 conversion).
```
wantehchang, line 8: Nit: I suggest omitting the parentheses:

```
: # to enable faster RGB to YUV limited range BT.601 conversion and disable faster full range BT.601 conversion.
```

src/reformat_libyuv.c
```diff
@@ -50,6 +56,126 @@ unsigned int avifLibYUVVersion(void)
...
+
+        return !RGBtoY ? AVIF_RESULT_NOT_IMPLEMENTED
+               : RGBtoY(rgb->pixels,
+                        rgb->rowBytes,
+                        image->yuvPlanes[AVIF_CHAN_Y],
+                        image->yuvRowBytes[AVIF_CHAN_Y],
+                        image->width,
+                        image->height)
+                   ? AVIF_RESULT_REFORMAT_FAILED
+                   : AVIF_RESULT_OK;
```
wantehchang, line 117: Nit: I think it'd be clearer to rewrite this using if statements:
```
        if (!RGBtoY) {
        }
        if (RGBtoY(rgb->pixels, ...)) {
            AVIF_RESULT_REFORMAT_FAILED;
        }
        return AVIF_RESULT_OK;
```

Please make a similar change to the return statement at line 164.

wantehchang (changes requested):

src/reformat_libyuv.c
y-guyon: Done at http://crbug/libyuv/936#c9

include/avif/internal.h
```diff
@@ -143,6 +143,7 @@ typedef struct avifReformatState
 // * AVIF_RESULT_OK              - Converted successfully with libyuv
 // * AVIF_RESULT_NOT_IMPLEMENTED - The fast path for this combination is not implemented with libyuv, use built-in YUV conversion
 // * [any other error]           - Return error to caller
+avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb);
```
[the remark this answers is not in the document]
wantehchang: I didn't know you were following the parameter order of `avifImageRGBToYUV()` in avif.h. Since `avifImageRGBToYUV()` is a public function, and it may be commonly used, let's not change it in this pull request. Please undo this change. Sorry about the extra work!

src/reformat_libyuv.c
fbarchard: Shouldnt this compare ARGBToI420 results?  The difference with LIBYUV_BIT_EXACT relates to subsampling.

y-guyon: As mentioned at http://crbug/libyuv/936#c10, `avifIsLibyuvBitExact()` is not a very sustainable solution. As there is no equivalent alternative in libyuv, I suggest we drop any `LIBYUV_BIT_EXACT` detection.

There is nothing to change for full range because it gives more accurate results without `LIBYUV_BIT_EXACT`, `LIBYUV_BIT_EXACT` is disabled by default in libyuv, and it is unlikely to be enabled by accident by the user.

However for limited range, here are the stats without `LIBYUV_BIT_EXACT`:
- RGB to YUV limited range (without LIBYUV_BIT_EXACT):
    Overall average drift increased by 0.26 on average, 0.32 max
    Quality loss is increased by 46% on average, 103% max
    Sample value difference is increased by 1.44 in average, up to 2
    PSNR drops by 2.2dB on average, 4.4dB max

It is significantly worse than the ["RGB to YUV limited range (with LIBYUV_BIT_EXACT)" quality loss from my original post](https://github.com/AOMediaCodec/libavif/pull/989#issue-1301079230) (results obtained with an Intel CPU). On the other hand, there is no visually detectable loss, it only impacts the objective metrics at high quality, so it might be fine. On top of that, limited range is already bringing a significant quality drop, so users of this setting should not be that concerned by this RGB-to-YUV conversion loss.

Thus I suggest we keep the libyuv calls in this PR as is but we do not check for `avifIsLibyuvBitExact()`. This will make it easier to improve libyuv accuracy in the future, if necessary.

src/reformat.c
```diff
@@ -381,6 +369,35 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
...
+                            *pU = (uint16_t)avifReformatStateUVToUNorm(&state, avgU);
+                            uint16_t * pV = (uint16_t *)&yuvPlanes[AVIF_CHAN_V][(uvI * 2) + (uvJ * yuvRowBytes[AVIF_CHAN_V])];
+                            *pV = (uint16_t)avifReformatStateUVToUNorm(&state, avgV);
+                        } else {
+                            yuvPlanes[AVIF_CHAN_U][uvI + (uvJ * yuvRowBytes[AVIF_CHAN_U])] =
+                                (uint8_t)avifReformatStateUVToUNorm(&state, avgU);
+                            yuvPlanes[AVIF_CHAN_V][uvI + (uvJ * yuvRowBytes[AVIF_CHAN_V])] =
+                                (uint8_t)avifReformatStateUVToUNorm(&state, avgV);
+                        }
+                    }
```
wantehchang, line 400: IMPORTANT: Please take a look at lins 406-428:

```
    if (image->alphaPlane && image->alphaRowBytes) {
        avifAlphaParams params;
        params.width = image->width;
        params.height = image->height;
        params.dstDepth = image->depth;
        params.dstPlane = image->alphaPlane;
        params.dstRowBytes = image->alphaRowBytes;
        params.dstOffsetBytes = 0;
        params.dstPixelBytes = state.yuvChannelBytes;
        if (avifRGBFormatHasAlpha(rgb->format) && !rgb->ignoreAlpha) {
            params.srcDepth = rgb->depth;
            params.srcPlane = rgb->pixels;
            params.srcRowBytes = rgb->rowBytes;
            params.srcOffsetBytes = state.rgbOffsetBytesA;
            params.srcPixelBytes = state.rgbPixelBytes;
            avifReformatAlpha(&params);
        } else {
            avifFillAlpha(&params);
        }
    }
```

Compare with lines 1149-1174:

```
    // Reformat alpha, if user asks for it, or (un)multiply processing needs it.
    if (avifRGBFormatHasAlpha(rgb->format) && (!rgb->ignoreAlpha || (alphaMultiplyMode != AVIF_ALPHA_MULTIPLY_MODE_NO_OP))) {
        avifAlphaParams params;
        params.width = rgb->width;
        params.height = rgb->height;
        params.dstDepth = rgb->depth;
        params.dstPlane = rgb->pixels;
        params.dstRowBytes = rgb->rowBytes;
        params.dstOffsetBytes = state.rgbOffsetBytesA;
        params.dstPixelBytes = state.rgbPixelBytes;
        if (image->alphaPlane && image->alphaRowBytes) {
            params.srcDepth = image->depth;
            params.srcPlane = image->alphaPlane;
            params.srcRowBytes = image->alphaRowBytes;
            params.srcOffsetBytes = 0;
            params.srcPixelBytes = state.yuvChannelBytes;
            avifReformatAlpha(&params);
        } else {
            if (!convertedWithLibYUV) { // libyuv fills alpha for us
                avifFillAlpha(&params);
            }
        }
    }
```

Do you think we should guard the `avifFillAlpha(&params)` call at line 426 with `if (!convertedWithLibYUV)`?

I don't know the answer to this question. I was just doing pattern matching.

src/reformat_libyuv.c
```diff
@@ -50,6 +56,113 @@ unsigned int avifLibYUVVersion(void)
...
+            return AVIF_RESULT_REFORMAT_FAILED;
+        }
+        return AVIF_RESULT_OK;
+    }
+
+    // Generic mapping from any RGB layout (with or without alpha) to any YUV layout (subsampled or not).
+    int (*RGBtoYUV)(const uint8_t *, int, uint8_t *, int, uint8_t *, int, uint8_t *, int, int, int) = NULL;
+
+    // libyuv only handles BT.601 for RGB to YUV, and not all range/order/subsampling combinations.
+    if (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_BT601) {
```
wantehchang, line 105: Also check for `AVIF_MATRIX_COEFFICIENTS_BT470BG`, which has the same coefficients as `AVIF_MATRIX_COEFFICIENTS_BT601`.

wantehchang (changes requested):

wantehchang (changes requested):

src/reformat_libyuv.c
```diff
@@ -50,6 +56,113 @@ unsigned int avifLibYUVVersion(void)
 #pragma clang diagnostic pop
 #endif
 
+static avifResult avifImageRGBToYUVLibYUV8bpc(avifImage * image, const avifRGBImage * rgb);
+
+avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb)
```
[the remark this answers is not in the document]
y-guyon: Currently there are inconsistencies with `avifChromaUpsampling`:

- the name is misleading as it does not only impact upsampling but also YUV to RGB conversion results, with and without upsampling (because of different conversion results with libyuv).
- comment mentions `ignored for 444` but this is wrong: `avifImageYUVToRGBLibYUV()` will bail out also for 444 on `BEST_QUALITY` for example.
- comment mentions `avifRGBImageSetDefaults() prefers quality over speed` but `avifRGBImageSetDefaults()` sets `AVIF_CHROMA_UPSAMPLING_AUTOMATIC` whose comment says `prefers libyuv, else uses BEST_QUALITY` so it actually prefers libyuv-grade speed over bilinear quality.

---

I added `avifChromaDownsampling` to be consistent with `avifChromaUpsampling` but I suggest we fix the problems mentioned above in a future change, especially the naming, even if it is an API breaking change.

> We may need to have a new option, for example, to select sharp YUV.

Good idea. It may be of interest to @maryla-uc.

src/reformat.c
y-guyon: > Do you think we should guard the `avifFillAlpha(&params)` call at line 426 with `if (!convertedWithLibYUV)`?

No. I added a comment for clarity.

src/reformat_libyuv.c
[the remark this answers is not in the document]
wantehchang: Yannis: I will review this pull request on Monday. Sorry about the delay.

I agree with your analysis. `avifChromaUpsampling` is really a parameter for the `avifImageYUVToRGB()` function and does not need to be stored in the `avifRGBImage` struct. Its meaning also became somewhat inconsistent when chroma upsampling with bilinear filtering was added to libyuv so that there were two impmenetations (libavif built-in and libyuv) of bilinear interpolation. We should look into how to add a `flags` parameter (a bitmask) to the `avifImageYUVToRGB()` function and define the bit flags appropriately.

Similarly `avifChromaDownsampling` is really a parameter of the `avifImageRGBToYUV()` function. An advantage of making it a member of the `avifRGBImage` struct is that existing `avifImageRGBToYUV()` calls don't need to be modified and just need to be recompiled. We can consider adding a `flags` parameter to the `avifImageRGBToYUV()` function now, or doing that at the same time as adding a `flags` parameter to the `avifImageYUVToRGB()` function.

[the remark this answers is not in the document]
y-guyon: Unless there is still a lot of comments to make on this PR, I suggest we move `avifChromaUpsampling` and `avifChromaDownsampling` from `avifRGBImage` to `avifImageYUVToRGB()` and `avifImageRGBToYUV()` arguments in another change.

Alternatively, we can make the change for `avifChromaUpsampling` first, and then do the change for `avifChromaDownsampling` in this PR before merging it.

include/avif/avif.h
```diff
@@ -552,6 +564,14 @@ typedef enum avifChromaUpsampling
     AVIF_CHROMA_UPSAMPLING_BILINEAR = 4      // Uses bilinear filter (built-in)
 } avifChromaUpsampling;
 
+typedef enum avifChromaDownsampling
```
wantehchang, line 567: In the current implementation, `AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC` and `AVIF_CHROMA_DOWNSAMPLING_FASTEST` are the same, and `AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY` and `AVIF_CHROMA_DOWNSAMPLING_AVERAGE` are the same.

Please confirm this is the intention. Also, do you think it will stay this way?

wantehchang (changes requested):

src/reformat_libyuv.c
fbarchard: If there is an accuracy issue with libyuv RGB subsampling, I'd be interested to fix it.
libyuv takes the average of the 4 RGB pixels, producing a single RGB pixel.  And then proceeds to convert to UV.
Is this ok, or should the averaging be moved to after the conversion?
Do we know where 'drift' is coming from?


LIBYUV_BIT_EXACT makes intel use the arm method.  Instead of average of 4 pixels, it uses the sum of 4 pixels without the divide by 4.  Then the matrix has adjusted coefficients.  This hurts quality for some color spaces.  If this is the problem, I'd suggest making ARM / BIT EXACT use the intel method.
It just adds a shift, hurting performance on ARM by 2 instructions.

exact flag will likely go away... it makes sense for ARM and Intel to match.
But if flags need to remembered, a function could save all the compile flags into bits of an int so they can be queried or changed at runtime.

wantehchang: Frank: If we ignore the clipping and rounding operations, the RGB-to-UV conversion and the four-pixel averaging operation are both linear, so we should be able to interchange them and the results should only differ due to clipping and rounding. Of course, we would prefer if the averaging is moved to after the conversion to eliminate this source of inaccuracy.

src/reformat_libyuv.c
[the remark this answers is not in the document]
wantehchang: Yannis: Your suggestion is fine by me.

include/avif/avif.h
```diff
@@ -552,6 +564,14 @@ typedef enum avifChromaUpsampling
...
+{
+    AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC = 0,    // Chooses best trade off of speed/quality (prefers libyuv, else uses BEST_QUALITY)
+    AVIF_CHROMA_DOWNSAMPLING_FASTEST = 1,      // Chooses speed over quality (prefers libyuv, else uses AVERAGE)
+    AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY = 2, // Chooses the best quality upsampling (avoids libyuv, uses AVERAGE)
```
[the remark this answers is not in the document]
y-guyon: I can imagine so, or at least a dedicated AVIF_CHROMA_DOWNSAMPLING_SHARP_YUV as a start.

include/avif/avif.h
y-guyon: The intention was consistency by having flags as similar to `avifChromaUpsampling` as possible. The flags and/or behavior may change when SharpYUV and/or other libyuv functions are made available.

include/avif/avif.h
```diff
@@ -552,6 +564,14 @@ typedef enum avifChromaUpsampling
...
+    AVIF_CHROMA_DOWNSAMPLING_FLOAT_AVERAGE = 3 // Uses floating point RGB-to-YUV conversion then averaging (built-in)
```
wantehchang, line 572: Let's remove "FLOAT_" from this enum name.

I discussed the API problem with Joe Drago and he suggested that we add a `preferLibYUV` setting to control whether libyuv should be used if available. Then the `chromaUpsampling` and `chromaDownsampling` settings will only specify the upsampling and downsampling filters. To prepare for this future change, this enum name just needs to specify that chroma is downsampled by averaging.

wantehchang (commented):
I will review this PR on Wednesday.

include/avif/avif.h
y-guyon: So should we keep `chromaUpsampling` and `chromaDownsampling` in `avifRGBImage` or should we move them to `vifImageRGBToYUV()` and `avifImageYUVToRGB()` arguments as discussed? I guess the `preferLibYUV` setting will be at the same location.

wantehchang: In this pull request we will keep `chromaUpsampling` and `chromaDownsampling` in `avifRGBImage`, as you suggested earlier.

y-guyon: Yes. I meant in the future, not in this PR. Sorry, that was unclear.

include/avif/avif.h
```diff
@@ -531,6 +531,18 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
 //   * x/x/[1|2|5|6|9]/Limited
 //   * [1|2|5|6|9]/x/12/Limited
 
+// If libavif is built with libyuv fast paths enabled, libavif will use libyuv for conversion from
+// RGB to YUV if the following requirements are met:
+//
+// * YUV depth: 8
+// * RGB depth: 8
+// * rgb.chromaDownsampling: AVIF_CHROMA_UPSAMPLING_AUTOMATIC, AVIF_CHROMA_UPSAMPLING_FASTEST
```
wantehchang, line 539: Also image->matrixCoefficients must be `AVIF_MATRIX_COEFFICIENTS_BT470BG` or `AVIF_MATRIX_COEFFICIENTS_BT601`.  We probably should use `5` and `6`, following the convention used at lines 529-532.

include/avif/avif.h
```diff
@@ -531,6 +531,18 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
...
+// * One of the following combinations (avifRGBFormat to avifPixelFormat/Range):
+//   *  BGRA            to [YUV400|YUV420|YUV422]/[Full|Limited]
+//   *  BGRA            to  YUV444               / Limited
+//   *  BGR             to  YUV420               /[Full|Limited]
+//   * [RGBA|ARGB|ABGR] to  YUV420               / Limited
```
wantehchang, line 544: [Just a comment. No change requested.] It is difficult to verify this table matches the code in `avifImageRGBToYUVLibYUV8bpc()`.

If possible, we should reorder the code to make it more obvious that it matches this table in a follow-up pull request.

wantehchang (approved):
LGTM. Thanks!

include/avif/avif.h
wantehchang: No worries.

We will need to think about this API design issue. Ideally these settings should become function parameters for `avifImageRGBToYUV()` and `avifImageYUVToRGB()`. But we need to avoid changing the function signatures again if we ever need to add another new setting. So we may need to use a bitmask (and combine these settings using bitwise-OR) or a parameter struct.

include/avif/avif.h
y-guyon: Done at https://github.com/AOMediaCodec/libavif/pull/1009

include/avif/avif.h
y-guyon: I sent https://github.com/AOMediaCodec/libavif/pull/1026 with modifications to `avif.h` only, as a start. @wantehchang could you take a quick look and tell me if it matches what Joe and you discussed, please? I will refactor the remaining source files afterwards, to avoid unnecessary work in the wrong direction.

PR #991, opened 2022-07-12
Use I420ToRGB24Matrix() in reformat_libyuv.c

y-guyon:
As suggested in https://crbug.com/libyuv/936.

Summary using `tests/gtest/avifrgbtoyuvtest` results (8-bit BT.601 4:2:0),
libyuv compared to built-in conversion:
- YUV to RGB/BGR on noisy samples:
    Overall average drift increased by 0.03
    Quality loss is decreased by 1.7%
    Max sample value difference is increased by 1.5 in average, up to 2
    PSNR rises by 0.18dB
- YUV to RGB/BGR on single color:
    Overall average drift increased by 0.03
    Quality loss is increased by 20%
    Max sample value difference is increased by 1
    PSNR drops by 1.1dB on average, 1.6dB max

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -239,14 +239,52 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
     //
     // libavif format        libyuv Func     UV matrix (and UV argument ordering)
     // --------------------  -------------   ------------------------------------
-    // AVIF_RGB_FORMAT_RGB   n/a             n/a
-    // AVIF_RGB_FORMAT_BGR   n/a             n/a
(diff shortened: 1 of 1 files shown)


jzern (approved):

wantehchang (commented):
LGTM. Thanks!

wantehchang (commented):
I also looked into whether we need to guard the `I420ToRGB24Matrix()` calls with a libyuv version check. `I420ToRGB24Matrix()` was added in libyuv version 1755:

https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2202748

libyuv currently requires libyuv version 1755, so we can call `I420ToRGB24Matrix()` without a libyuv version check.

PR #994, opened 2022-07-25
svt: null metadata field on EbBufferHeaderType

1480c1:
fixes [svt-av1's issue 1957 ](https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1957)relating to uninitialized variables crashing the encoder

Signed-off-by: Christopher Degawa <ccom@randomderp.com>

ping @wantehchang

### DIFF
--- a/src/codec_svt.c
+++ b/src/codec_svt.c
@@ -207,6 +207,7 @@ static avifBool svtCodecEncodeFinish(avifCodec * codec, avifCodecEncodeOutput * 
     input_buffer.p_app_private = NULL;
     input_buffer.flags = EB_BUFFERFLAG_EOS;
     input_buffer.p_buffer = NULL;
+    input_buffer.metadata = NULL;
 
     // flush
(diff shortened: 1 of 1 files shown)


src/codec_svt.c
```diff
@@ -207,6 +207,7 @@ static avifBool svtCodecEncodeFinish(avifCodec * codec, avifCodecEncodeOutput *
     input_buffer.p_app_private = NULL;
     input_buffer.flags = EB_BUFFERFLAG_EOS;
     input_buffer.p_buffer = NULL;
+    input_buffer.metadata = NULL;
```
wantehchang, line 210: Do you think we should set input_buffer.size after line 203?

```
    input_buffer.size = sizeof(EbBufferHeaderType);
```

wantehchang (approved):
LGTM. Thanks.

src/codec_svt.c
1480c1: From what I understand based on the sample app, .size does not need to be set if sending an EOS

Issue #992, opened 2022-07-16
Encoding of big png silently fails


danielniccoli:
I have a png file with the size of 37120 x 73984 pixels (W x H). Encoding it with `.\avifenc.exe  .\output_world.png .\output_world.avif` (on Windows) silently fails and there is no output file.

Upon further inspection, I could find an exit code: 0xc0000409

Version: 0.10.1 (dav1d [dec]:1.0.0-0-g99172b1, aom [enc/dec]:3.3.0, rav1e [enc]:0.5.1 (v0.5.1), svt [enc]:v1.1.0)

danielniccoli:
How do I get a stack trace?

Link to image: https://1drv.ms/u/s!AmQg3eUYceo2pA7sKgtRKjsILT0U?e=XyeVrs

jzern:
Thanks for uploading the image.

For the stacktrace I was hoping a debug build might give some more information in any failure pop-up or you could try running the binary under Visual Studio, for instance.

How are you building libavif and the dependent libs?

joedrago:
On my build this is abort()ing in `malloc()` as this line of code in `avifImageAllocatePlanes()` is overflowing:

```
int fullSize = fullRowBytes * image->height;
```
`36864 * 73728 = 2717908992`, which overflows the signed i32 here, so it attempts to allocate `-1577058304` bytes, which comes through as `18,446,744,072,132,493,312` bytes in `avifAlloc()`. 

We need overflow checks here, and probably should honor `AVIF_DEFAULT_IMAGE_SIZE_LIMIT` too.

danielniccoli:
I was not building from code, but using a Windows build from https://ci.appveyor.com/project/louquillio/libavif/builds/43929886/artifacts

PR #995, opened 2022-07-26
Make avifImageAllocatePlanes() return avifResult

y-guyon:
To catch memory allocation failures and invalid parameters, it is
safer to return a status from avifImageAllocatePlanes() and from
avifImageCopy().

Fixes https://github.com/AOMediaCodec/libavif/issues/992.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-There are incompatible ABI changes in this release. Members were removed from
-avifImage struct. avifImageCopy() and avifImageAllocatePlanes() signatures
-changed. It is necessary to recompile your code. It is also recommended to check
-the return values of avifImageCopy() and avifImageAllocatePlanes().
+There are incompatible ABI changes in this release. The alphaRange member was
+removed from avifImage struct. avifImageCopy() and avifImageAllocatePlanes()
+signatures changed. It is necessary to recompile your code. Also check the
+return values of avifImageCopy() and avifImageAllocatePlanes().
 
 ### Changed
 * Update aom.cmd: v3.4.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,15 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-There is an incompatible ABI change in this release. Members were removed from
-avifImage struct and avifImage*() signatures changed. It is necessary to
-recompile your code.
+There are incompatible ABI changes in this release. Members were removed from
+avifImage struct. avifImageCopy() and avifImageAllocatePlanes() signatures
+changed. It is necessary to recompile your code. It is also recommended to check
+the return values of avifImageCopy() and avifImageAllocatePlanes().
 
 ### Changed
 * Update aom.cmd: v3.4.0
 * Update svt.cmd/svt.sh: v1.1.0
 * avifImageCopy() and avifImageAllocatePlanes() now return avifResult instead of
-  void to warn about invalid parameters or memory allocation failures
+  void to report invalid parameters or memory allocation failures
 
 ### Removed
 * alphaRange field was removed from the avifImage struct. It it presumed that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 There is an incompatible ABI change in this release. Members were removed from
-avifImage struct. It is necessary to recompile your code.
+avifImage struct and avifImage*() signatures changed. It is necessary to
+recompile your code.
 
 ### Changed
 * Update aom.cmd: v3.4.0
 * Update svt.cmd/svt.sh: v1.1.0
+* avifImageCopy() and avifImageAllocatePlanes() now return avifResult instead of
+  void to warn about invalid parameters or memory allocation failures
 
 ### Removed
 * alphaRange field was removed from the avifImage struct. It it presumed that
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -40,7 +40,7 @@ static void my_error_exit(j_common_ptr cinfo)
 
 // An internal function used by avifJPEGReadCopy(), this is the shared libjpeg decompression code
 // for all paths avifJPEGReadCopy() takes.
-static void avifJPEGCopyPixels(avifImage * avif, struct jpeg_decompress_struct * cinfo)
+static avifBool avifJPEGCopyPixels(avifImage * avif, struct jpeg_decompress_struct * cinfo)
 {
     cinfo->raw_data_out = TRUE;
     jpeg_start_decompress(cinfo);
@@ -68,7 +68,9 @@ static void avifJPEGCopyPixels(avifImage * avif, struct jpeg_decompress_struct *
         readLines = AVIF_MAX(readLines, linesPerCall[i]);
     }
 
-    avifImageAllocatePlanes(avif, AVIF_PLANES_YUV);
+    if (avifImageAllocatePlanes(avif, AVIF_PLANES_YUV) != AVIF_RESULT_OK) {
+        return AVIF_FALSE;
+    }
 
     // destination avif channel for each jpeg channel
     avifChannelIndex targetChannel[3] = { AVIF_CHAN_Y, AVIF_CHAN_Y, AVIF_CHAN_Y };
@@ -102,6 +104,7 @@ static void avifJPEGCopyPixels(avifImage * avif, struct jpeg_decompress_struct *
             alreadyRead[i] += linesPerCall[i];
         }
     }
+    return AVIF_TRUE;
 }
 
 static avifBool avifJPEGHasCompatibleMatrixCoefficients(avifMatrixCoefficients matrixCoefficients)
@@ -148,9 +151,7 @@ static avifBool avifJPEGReadCopy(avifImage * avif, struct jpeg_decompress_struct
                 }
                 if (avif->yuvFormat == jpegFormat) {
                     cinfo->out_color_space = JCS_YCbCr;
-                    avifJPEGCopyPixels(avif, cinfo);
-
-                    return AVIF_TRUE;
+                    return avifJPEGCopyPixels(avif, cinfo);
                 }
             }
 
@@ -158,9 +159,7 @@ static avifBool avifJPEGReadCopy(avifImage * avif, struct jpeg_decompress_struct
             if ((avif->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) && (cinfo->comp_info[0].h_samp_factor == cinfo->max_h_samp_factor &&
                                                                   cinfo->comp_info[0].v_samp_factor == cinfo->max_v_samp_factor)) {
                 cinfo->out_color_space = JCS_YCbCr;
-                avifJPEGCopyPixels(avif, cinfo);
-
-                return AVIF_TRUE;
+                return avifJPEGCopyPixels(avif, cinfo);
             }
         }
     } else if (cinfo->jpeg_color_space == JCS_GRAYSCALE) {
@@ -173,16 +172,16 @@ static avifBool avifJPEGReadCopy(avifImage * avif, struct jpeg_decompress_struct
                 if ((avif->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) || (avif->yuvFormat == AVIF_PIXEL_FORMAT_NONE)) {
                     avif->yuvFormat = AVIF_PIXEL_FORMAT_YUV400;
                     cinfo->out_color_space = JCS_GRAYSCALE;
-                    avifJPEGCopyPixels(avif, cinfo);
-
-                    return AVIF_TRUE;
+                    return avifJPEGCopyPixels(avif, cinfo);
                 }
 
                 // Grayscale->YUV: copy Y, fill UV with monochrome value.
                 if ((avif->yuvFormat == AVIF_PIXEL_FORMAT_YUV444) || (avif->yuvFormat == AVIF_PIXEL_FORMAT_YUV422) ||
                     (avif->yuvFormat == AVIF_PIXEL_FORMAT_YUV420)) {
                     cinfo->out_color_space = JCS_GRAYSCALE;
-                    avifJPEGCopyPixels(avif, cinfo);
+                    if (!avifJPEGCopyPixels(avif, cinfo)) {
+                        return AVIF_FALSE;
+                    }
 
                     avifPixelFormatInfo info;
                     avifGetPixelFormatInfo(avif->yuvFormat, &info);
@@ -199,7 +198,9 @@ static avifBool avifJPEGReadCopy(avifImage * avif, struct jpeg_decompress_struct
                 ((avif->yuvFormat == AVIF_PIXEL_FORMAT_YUV444) || (avif->yuvFormat == AVIF_PIXEL_FORMAT_NONE))) {
                 avif->yuvFormat = AVIF_PIXEL_FORMAT_YUV444;
                 cinfo->out_color_space = JCS_GRAYSCALE;
-                avifJPEGCopyPixels(avif, cinfo);
+                if (!avifJPEGCopyPixels(avif, cinfo)) {
+                    return AVIF_FALSE;
+                }
 
                 memcpy(avif->yuvPlanes[AVIF_CHAN_U], avif->yuvPlanes[AVIF_CHAN_Y], (size_t)avif->yuvRowBytes[AVIF_CHAN_U] * avif->height);
                 memcpy(avif->yuvPlanes[AVIF_CHAN_V], avif->yuvPlanes[AVIF_CHAN_Y], (size_t)avif->yuvRowBytes[AVIF_CHAN_V] * avif->height);
@@ -216,9 +217,7 @@ static avifBool avifJPEGReadCopy(avifImage * avif, struct jpeg_decompress_struct
              cinfo->comp_info[2].h_samp_factor == 1 && cinfo->comp_info[2].v_samp_factor == 1)) {
             avif->yuvFormat = AVIF_PIXEL_FORMAT_YUV444;
             cinfo->out_color_space = JCS_RGB;
-            avifJPEGCopyPixels(avif, cinfo);
-
-            return AVIF_TRUE;
+            return avifJPEGCopyPixels(avif, cinfo);
         }
     }
 
--- a/src/avif.c
+++ b/src/avif.c
@@ -293,8 +293,8 @@ avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags planes)
         avifGetPixelFormatInfo(image->yuvFormat, &info);
 
         // Intermediary computation as 64 bits in case width or height is exactly UINT32_MAX.
-        const uint32_t shiftedW = (uint32_t)((image->width + (uint64_t)info.chromaShiftX) >> info.chromaShiftX);
-        const uint32_t shiftedH = (uint32_t)((image->height + (uint64_t)info.chromaShiftY) >> info.chromaShiftY);
+        const uint32_t shiftedW = (uint32_t)(((uint64_t)image->width + info.chromaShiftX) >> info.chromaShiftX);
+        const uint32_t shiftedH = (uint32_t)(((uint64_t)image->height + info.chromaShiftY) >> info.chromaShiftY);
 
         // These are less than or equal to fullRowBytes/fullSize. No need to check overflows.
         const size_t uvRowBytes = channelSize * shiftedW;
--- a/src/avif.c
+++ b/src/avif.c
@@ -3,6 +3,7 @@
 
 #include "avif/internal.h"
 
+#include <limits.h>
 #include <string.h>
 
 #define STR_HELPER(x) #x
@@ -163,7 +164,7 @@ static void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImag
     dstImage->imir = srcImage->imir;
 }
 
-void avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
+avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
 {
     avifImageFreePlanes(dstImage, AVIF_PLANES_ALL);
     avifImageCopyNoAlloc(dstImage, srcImage);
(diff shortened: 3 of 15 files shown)


tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,167 @@
...
+  testutil::AvifImagePtr image(avifImageCreateEmpty(), avifImageDestroy);
+  ASSERT_NE(image, nullptr);
+  image->width = width;
+  image->height = height;
+  image->depth = depth;
+  image->yuvFormat = AVIF_PIXEL_FORMAT_YUV444;
+
+  // This is the maximum number of bytes that can safely be allocated in
+  // this test.
+  static constexpr uint64_t kMaxAlloc = std::numeric_limits<int32_t>::max();
```
jzern, line 103: in some environments you might get a signed to unsigned conversion warning

jzern (commented):
This should probably have an entry in CHANGELOG.md

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,169 @@
...
+  encoder->speed = AVIF_SPEED_FASTEST;
+  testutil::AvifRwData encoded_avif;
+  ASSERT_EQ(avifEncoderWrite(encoder.get(), image.get(), &encoded_avif),
+            expected_result);
+}
+
+TEST(EncodingTest, MinimumValid) { TestAllocation(1, 1, 8, AVIF_RESULT_OK); }
+
+TEST(EncodingTest, MaximumValid) {
+  TestEncoding(65535, 1, 12, AVIF_RESULT_OK);
```
jzern, line 143: should this be AVIF_DEFAULT_IMAGE_SIZE_LIMIT?

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,169 @@
...
+  TestEncoding(1, 65535, 12, AVIF_RESULT_OK);
+  // TestEncoding(16384, 8096, 12, AVIF_RESULT_OK);  // Too slow.
+}
+
+TEST(EncodingTest, MinimumInvalid) {
+  TestEncoding(0, 1, 8, AVIF_RESULT_NO_CONTENT);
+  TestEncoding(1, 0, 8, AVIF_RESULT_NO_CONTENT);
+  TestEncoding(1, 1, 0, AVIF_RESULT_UNSUPPORTED_DEPTH);
+  TestEncoding(65536, 1, 12, AVIF_RESULT_ENCODE_COLOR_FAILED);
```
jzern, line 152: and AVIF_DEFAULT_IMAGE_SIZE_LIMIT + 1 here?

src/avif.c
```diff
@@ -264,11 +272,27 @@ void avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_t xmpS
...
+    const size_t channelSize = avifImageUsesU16(image) ? 2 : 1;
+    if (image->width > SIZE_MAX / channelSize) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
+    const size_t fullRowBytes = channelSize * image->width;
+    if ((image->height > SIZE_MAX / fullRowBytes) || (fullRowBytes > UINT_MAX)) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
+    const size_t fullSize = fullRowBytes * image->height;
+    if (image->width * image->height > AVIF_DEFAULT_IMAGE_SIZE_LIMIT) {
```
wantehchang, line 292: We should not do this check here. One reason is that we have performed a stronger check above. Another reason is that the image size limit is a configurable decoder option, and `AVIF_DEFAULT_IMAGE_SIZE_LIMIT` is merely the default value of that option.

src/avif.c
```diff
@@ -264,11 +272,27 @@ void avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_t xmpS
...
+    }
+    const size_t channelSize = avifImageUsesU16(image) ? 2 : 1;
+    if (image->width > SIZE_MAX / channelSize) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
+    const size_t fullRowBytes = channelSize * image->width;
+    if ((image->height > SIZE_MAX / fullRowBytes) || (fullRowBytes > UINT_MAX)) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
+    const size_t fullSize = fullRowBytes * image->height;
```
wantehchang, line 291: IMPORTANT: Also declare the `uvRowBytes` and `uvSize` variables (at lines 303-304) as `size_t`, similar to the `fullRowBytes` and `fullSize` variables here.

src/avif.c
```diff
@@ -282,16 +306,25 @@ void avifImageAllocatePlanes(avifImage * image, avifPlanesFlags planes)
...
+                return AVIF_RESULT_OUT_OF_MEMORY;
+            }
         }
 
         if (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400) {
             if (!image->yuvPlanes[AVIF_CHAN_U]) {
                 image->yuvRowBytes[AVIF_CHAN_U] = uvRowBytes;
                 image->yuvPlanes[AVIF_CHAN_U] = avifAlloc(uvSize);
+                if (!image->yuvPlanes[AVIF_CHAN_U]) {
+                    return AVIF_RESULT_OUT_OF_MEMORY;
```
wantehchang, line 319: IMPORTANT: We need to make sure that `avifImageFreePlanes()` can free such a partially-allocated image.

I didn't look at this closely, but it seems that one way to make this work would be to set image->imageOwnsYUVPlanes to AVIF_TRUE early. Another way would be to free the earlier planes if the allocation of a later plane fails.

wantehchang (changes requested):
Yannis: Thanks for writing this pull request. I reviewed everything carefully except the new test file.

tests/gtest/avifallocationtest.cc
y-guyon: 65535 is the maximum width allowed by [libaom](https://aomedia.googlesource.com/aom/+/5f9433e1053ec56b72ac6760a33d4270b8cca562/av1/av1_cx_iface.c#615).

tests/gtest/avifallocationtest.cc
y-guyon: Changed to maximum AV1 dimension + 1.

CHANGELOG.md
```diff
@@ -7,11 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 There is an incompatible ABI change in this release. Members were removed from
-avifImage struct. It is necessary to recompile your code.
+avifImage struct and avifImage*() signatures changed. It is necessary to
```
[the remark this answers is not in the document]
y-guyon: 1. Done.
2. Done but why stay evasive about "Members removed from avifImage" and precise about which functions changed, while both pieces of information are written right below?

apps/shared/avifjpeg.c
```diff
@@ -40,7 +40,7 @@ static void my_error_exit(j_common_ptr cinfo)
 
 // An internal function used by avifJPEGReadCopy(), this is the shared libjpeg decompression code
 // for all paths avifJPEGReadCopy() takes.
-static void avifJPEGCopyPixels(avifImage * avif, struct jpeg_decompress_struct * cinfo)
+static avifBool avifJPEGCopyPixels(avifImage * avif, struct jpeg_decompress_struct * cinfo)
```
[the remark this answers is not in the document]
y-guyon: According to this [jpeglib.h](https://github.com/kornelski/libjpeg/blob/41f55c093a551ff5df0431155de978618862bbc9/jpeglib.h#L744), no.

src/avif.c
y-guyon: Done. I had to add an extra conversion to `uint64_t` for `shiftedW/H`. Also [`decoder->imageSizeLimit` cannot be greater than `AVIF_DEFAULT_IMAGE_SIZE_LIMIT` during decoding for now](https://github.com/AOMediaCodec/libavif/blob/1d2462a20d7140ccfb4edd659ad38eea05aab41a/src/read.c#L3132).

src/avif.c
y-guyon: Done.

This is brittle anyway: if you pass an `avifImage` with some plane set to external memory (with `imageOwnsYUVPlanes = AVIF_FALSE`), `avifImageAllocatePlanes()` will not replace those pointers but set `imageOwnsYUVPlanes = AVIF_TRUE`, leading to memory freed twice (unrelated to allocation failures).

src/avif.c
```diff
@@ -264,45 +273,70 @@ void avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_t xmpS
...
+    }
+    const size_t fullSize = fullRowBytes * image->height;
+
     if ((planes & AVIF_PLANES_YUV) && (image->yuvFormat != AVIF_PIXEL_FORMAT_NONE)) {
         avifPixelFormatInfo info;
         avifGetPixelFormatInfo(image->yuvFormat, &info);
 
-        int shiftedW = (image->width + info.chromaShiftX) >> info.chromaShiftX;
-        int shiftedH = (image->height + info.chromaShiftY) >> info.chromaShiftY;
+        // Intermediary computation as 64 bits in case width or height is exactly UINT32_MAX.
```
wantehchang, line 295: Optional: Good catch! I think the following should also work:

```
        const size_t shiftedW = (image->width >> info.chromaShiftX) + (image->width & info.chromaShiftX);
```

src/avif.c
```diff
@@ -264,45 +273,70 @@ void avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_t xmpS
...
+        const size_t shiftedW = (image->width + (uint64_t)info.chromaShiftX) >> info.chromaShiftX;
+        const size_t shiftedH = (image->height + (uint64_t)info.chromaShiftY) >> info.chromaShiftY;
 
-        int uvRowBytes = channelSize * shiftedW;
-        int uvSize = uvRowBytes * shiftedH;
+        // These are equal to fullRowBytes/fullSize or lower. No need to check overflows.
```
wantehchang, line 299: Nit/optional: change "These are equal to fullRowBytes/fullSize or lower" to "These are less than or equal to fullRowBytes/fullSize". Alternatively, change "lower" to "smaller".

src/avif.c
```diff
@@ -264,45 +273,70 @@ void avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_t xmpS
...
+
     if ((planes & AVIF_PLANES_YUV) && (image->yuvFormat != AVIF_PIXEL_FORMAT_NONE)) {
         avifPixelFormatInfo info;
         avifGetPixelFormatInfo(image->yuvFormat, &info);
 
-        int shiftedW = (image->width + info.chromaShiftX) >> info.chromaShiftX;
-        int shiftedH = (image->height + info.chromaShiftY) >> info.chromaShiftY;
+        // Intermediary computation as 64 bits in case width or height is exactly UINT32_MAX.
+        const size_t shiftedW = (image->width + (uint64_t)info.chromaShiftX) >> info.chromaShiftX;
+        const size_t shiftedH = (image->height + (uint64_t)info.chromaShiftY) >> info.chromaShiftY;
```
wantehchang, line 297: 1. Optional: These two variables can be declared as `uint32_t`.

2. A `size_t` or `uint32_t` cast may be necessary to avoid a compiler warning in 32-bit builds.

tests/gtest/avifgridapitest.cc
```diff
@@ -41,6 +41,9 @@ TEST_P(GridApiTest, EncodeDecode) {
     cell_images.emplace_back(testutil::CreateImage(
         horizontal.size, vertical.size, bit_depth, yuv_format,
         create_alpha ? AVIF_PLANES_ALL : AVIF_PLANES_YUV));
+    if (cell_images.back() == nullptr && !expected_success) {
+      return;  // Bad dimensions may be already caught.
```
wantehchang, line 45: Nit: be => have been

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,177 @@
...
+
+namespace libavif {
+namespace {
+
+void TestAllocation(uint32_t width, uint32_t height, uint32_t depth,
+                    avifResult expected_result) {
+  // The format of the image and which planes are allocated should not matter.
+  // Test all combinations.
+  for (avifPixelFormat format :
+       {AVIF_PIXEL_FORMAT_NONE, AVIF_PIXEL_FORMAT_YUV444,
```
wantehchang, line 24: [Just a comment, no change requested] I am surprised that `avifImageAllocatePlanes()` allows `image->yuvFormat` to be `AVIF_PIXEL_FORMAT_NONE`.

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,177 @@
...
+      image->width = width;
+      image->height = height;
+      image->depth = depth;
+      image->yuvFormat = format;
+      EXPECT_EQ(avifImageAllocatePlanes(image.get(), planes), expected_result);
+
+      // Make sure the actual plane pointers are consistent with the settings.
+      if (expected_result == AVIF_RESULT_OK &&
+          format != AVIF_PIXEL_FORMAT_NONE &&
+          (planes == AVIF_PLANES_YUV || planes == AVIF_PLANES_ALL)) {
```
wantehchang, line 40: Replace `(planes == AVIF_PLANES_YUV || planes == AVIF_PLANES_ALL)` with `(planes & AVIF_PLANES_YUV)`?

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,177 @@
...
+                 AVIF_RESULT_OK);
+  TestAllocation(1, std::numeric_limits<typeof(avifImage::height)>::max(), 8,
+                 AVIF_RESULT_OK);
+  // 12 bits (impacts the width because avifImage stride is stored as uint32_t)
+  TestAllocation(std::numeric_limits<typeof(avifImage::width)>::max() / 2, 1,
+                 12, AVIF_RESULT_OK);
+  TestAllocation(1, std::numeric_limits<typeof(avifImage::height)>::max(), 12,
+                 AVIF_RESULT_OK);
+  // Some high number of bytes that malloc() accepts to allocate.
+  TestAllocation(1024 * 16, 1024 * 16, 12, AVIF_RESULT_OK);  // Up to 2 GB total
```
wantehchang, line 77: IMPORTANT: Please test this in 32-bit builds, whose address space is 4 GB total.

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,177 @@
...
+}
+
+TEST(AllocationTest, MaximumInvalid) {
+  TestAllocation(std::numeric_limits<typeof(avifImage::width)>::max(),
+                 std::numeric_limits<typeof(avifImage::height)>::max(), 12,
```
wantehchang, line 82: We can pass 1 as the second argument (height) here.

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,177 @@
...
+  image->width = width;
+  image->height = height;
+  image->depth = depth;
+  image->yuvFormat = AVIF_PIXEL_FORMAT_YUV444;
+
+  // This is a fairly high number of bytes that can safely be allocated in this
+  // test. The goal is to have something to give to libavif but libavif should
+  // return an error before attempting to read all of it, so it does not matter
+  // if there are fewer bytes than the provided image dimensions.
+  static constexpr uint64_t kMaxAlloc = 2147483647;
```
wantehchang, line 107: Please test this in 32-bit builds.

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  // Initialize pixels as 16b values to make sure values are valid for 10
+  // and 12-bit depths. The array will be cast to uint8_t for 8-bit depth.
+  std::vector<uint16_t> pixels(num_allocated_bytes / sizeof(uint16_t), 400);
+  uint8_t* bytes = reinterpret_cast<uint8_t*>(pixels.data());
+  // Avoid avifImageAllocatePlanes() to exercise the checks at encoding.
+  image->imageOwnsYUVPlanes = AVIF_FALSE;
+  image->imageOwnsAlphaPlane = AVIF_FALSE;
+  image->yuvRowBytes[AVIF_CHAN_Y] = image->width;
```
wantehchang, line 124: IMPORTANT: this needs to be multiplied by 2 if image->depth > 8. right?

wantehchang (changes requested):

tests/gtest/avifallocationtest.cc
y-guyon: Done. It passed for these 1024 * 16, 1024 * 16 dimensions but not for `std::numeric_limits<typeof(avifImage::width)>::max()`.

For the record, how to build as 32 bits:
- `sudo apt-get install gcc-multilib g++-multilib`
- Add `-DCMAKE_CXX_FLAGS=-m32 -DCMAKE_C_FLAGS=-m32` to libavif and local dependencies

tests/gtest/avifallocationtest.cc
y-guyon: It is to store alpha only, from what I understood.

src/avif.c
y-guyon: I did that at first and thought it would be more confusing than just casting to uint64_t, a data type that is used in some parts of libavif already.

src/avif.c
```diff
@@ -264,45 +273,70 @@ void avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_t xmpS
...
+avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags planes)
 {
-    int channelSize = avifImageUsesU16(image) ? 2 : 1;
-    int fullRowBytes = channelSize * image->width;
-    int fullSize = fullRowBytes * image->height;
+    if (image->width == 0 || image->height == 0) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
+    const size_t channelSize = avifImageUsesU16(image) ? 2 : 1;
+    if (image->width > SIZE_MAX / channelSize) {
```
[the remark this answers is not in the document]
y-guyon: `assert(sizeof(size_t) >= sizeof(uint32_t));` seems odder than the current implementation to me.

tests/gtest/avifallocationtest.cc
wantehchang: Thank you. Could you add this info to the README.md file?

wantehchang: I meant adding the 32-bit build info to README.md **in a separate PR**.

wantehchang (approved):
LGTM. I suggest some changes to tests/gtest/avifallocationtest.cc.

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,185 @@
...
+            expected_result);
+}
+
+TEST(EncodingTest, MinimumValid) { TestAllocation(1, 1, 8, AVIF_RESULT_OK); }
+
+TEST(EncodingTest, MaximumValid) {
+  // 65536 is the maximum AV1 frame dimension allowed by the AV1 specification.
+  // See the section 5.5.1. General sequence header OBU syntax.
+  // TODO(yguyon): Replace 65535 by 65536 once ext/aom.cmd clones a version of
+  //               aom that includes the fix for https://crbug.com/aomedia/3304.
```
wantehchang, line 157: libavif can also be built with a libaom library installed on the system (`-DAVIF_LOCAL_AOM=OFF`), so this TODO comment does not consider that case. I suggest we just mention the libaom bug and explain why we use 65535 instead of 65536 here, and remove the TODO comment.

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,185 @@
...
+  TestAllocation(1, kMaxAllocatableDimension, 8, AVIF_RESULT_OK);
+  // 12 bits (impacts the width because avifImage stride is stored as uint32_t)
+  TestAllocation(kMaxAllocatableDimension / 2, 1, 12, AVIF_RESULT_OK);
+  TestAllocation(1, kMaxAllocatableDimension, 12, AVIF_RESULT_OK);
+  // Some high number of bytes that malloc() accepts to allocate.
+  TestAllocation(1024 * 16, 1024 * 16, 12, AVIF_RESULT_OK);  // Up to 2 GB total
+}
+
+TEST(AllocationTest, MinimumInvalid) {
+  TestAllocation(std::numeric_limits<typeof(avifImage::width)>::max(), 1, 12,
```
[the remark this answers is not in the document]
y-guyon: `4294967295, 1, 12` are the minimum invalid dimensions. I renamed the test cases.

src/avif.c
```diff
@@ -264,45 +273,70 @@ void avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_t xmpS
...
+
     if ((planes & AVIF_PLANES_YUV) && (image->yuvFormat != AVIF_PIXEL_FORMAT_NONE)) {
         avifPixelFormatInfo info;
         avifGetPixelFormatInfo(image->yuvFormat, &info);
 
-        int shiftedW = (image->width + info.chromaShiftX) >> info.chromaShiftX;
-        int shiftedH = (image->height + info.chromaShiftY) >> info.chromaShiftY;
+        // Intermediary computation as 64 bits in case width or height is exactly UINT32_MAX.
+        const uint32_t shiftedW = (uint32_t)((image->width + (uint64_t)info.chromaShiftX) >> info.chromaShiftX);
+        const uint32_t shiftedH = (uint32_t)((image->height + (uint64_t)info.chromaShiftY) >> info.chromaShiftY);
```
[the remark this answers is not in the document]
y-guyon: `info.chromaShiftX` is an `int`. I had to look [integer promotion rules](https://en.cppreference.com/w/c/language/conversion) up:

> 1) the ranks of all signed integer types are different and increase with their precision
> 2) the ranks of all signed integer types equal the ranks of the corresponding unsigned integer types

So `uint64_t + int` yields `uint64_t`, assuming `sizeof(int) < sizeof(uint64_t)`. It seems to be fine to move the cast but I suggest we leave it as is, for safety and to avoid having to remember the promotion rules.

If I understood correctly, the standard says `uint32_t + int32_t` is implementation defined.

tests/gtest/avifallocationtest.cc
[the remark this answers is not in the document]
wantehchang: `std::numeric_limits<typeof(avifImage::width)>::max() / 2 + 1` is less than `std::numeric_limits<typeof(avifImage::width)>::max()` but is also invalid. So `std::numeric_limits<typeof(avifImage::width)>::max(), 1, 12` don't seem to be the minimum invalid dimensions.

Note: One can argue that `0, 0, 12`are the minumum invalid dimensions, so the name "MinimumInvalidDimensions" is still a little confusing to me. It may be simpler to just have a test named "InvalidDimensions".

src/avif.c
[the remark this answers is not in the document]
wantehchang: Thank you for looking into this. A lot of code (inside and outslde libavif) assumes `int` is the same as `int32_t`. Although strictly speaking this assumption is not portable, it is valid for all the general-purpose CPUs and compilers we need to support today. 

Analyzing this further, we see that `int` is unlikely to be 64 bits, otherwise int16_t and int32_t cannot both be defined as some standard integer types (`short int` is the only one available). It follows that in practice `int` is either 16 bits or 32 bits, and the cast I suggested will result in either `uint64_t + int16_t` or `uint64_t + int32_t`.

wantehchang (approved):
LGTM.

tests/gtest/avifallocationtest.cc
```diff
@@ -0,0 +1,187 @@
...
+TEST(AllocationTest, MinimumValidDimensions) {
+  TestAllocation(1, 1, 8, AVIF_RESULT_OK);
+}
+
+TEST(AllocationTest, MaximumValidDimensions) {
+  // On 32-bit builds, malloc() will fail with fairly low sizes.
+  // Adapt the tests to take that into account.
+  constexpr bool kIsPlatform64b = sizeof(void*) > 4;
+  constexpr uint32_t kMaxAllocatableDimension =
+      kIsPlatform64b ? std::numeric_limits<typeof(avifImage::width)>::max()
```
wantehchang, line 70: Yannis: When I tried to compile libavif with MSVC, I found that `typeof` is a GCC extension. It is not supported by MSVC.

y-guyon: Fixed in https://github.com/AOMediaCodec/libavif/pull/1014

Issue #766, opened 2021-09-22
libavif should default to a strict mode that only uses published features in HEIF/MIAF


leo-barnes:
Per the last Image Container SG meeting, Jon noted that he had encountered issues with files containing multiple `colr` boxes which is not legal according to the currently published version of HEIF.

Support for multiple `colr` boxes is coming in the next version of HEIF, but it's technically not legal yet, and unless someone is participating in the MPEG discussions they wouldn't know its changing.

The conclusion of the meeting was that libavif should probably default to a strict mode that only uses features in currently published specs, with new features only being enabled by clients who are willing to accept that the created files may not open everywhere.

leo-barnes:
@wantehchang @baumanj @cconcolato
Let me know if I missed some detail from the discussion.

joedrago:
I think this is a good strategy for any upcoming changes we have not yet implemented or shipped in libavif, and I'm completely on board in that context. However, other than that acknowledgment, what is the action item here in the context of multiple colr boxes?

As we've learned with both the `ispe` and `pixi` properties, once files like these are in the wild and at scale, it becomes near impossible to walk back these implementations. If it was just a decode strictness, we could add a strict flag (as libavif already has a notion of decode strictness), but seeing as this particular colr box change was [implemented almost a year ago](https://github.com/AOMediaCodec/libavif/commit/bf58fe7c1f1e16c2e50570432b04b16cdd07fc5d) and exists in any file encoded with an ICC profile using `avifenc` since `v0.8.4`, I think its a little late for this particular change. Isn't the new amendment going live very, very soon?

Again, I'm completely on board for any future amendments being locked behind some strict/future flag. Anything we do for the colr boxes here is going to just be unncessary swirl, in my opinion.

baumanj:
Seconding what @joedrago said. Missing the support for multiple `colr` boxes in mp4parse-rust was an unfortunate situation, but it's too late to start rejecting those files. Going forward, defaulting to what's in the specs sounds nice, but I'm assuming we'd make exceptions for these other unpublished changes as well:
- `imir` interpretation
- permitting absent `pixi`
- permitting absent `mif1` brand

Are there any others?

joedrago:
I think AVIF sequences are still a bit of wild wild west, honestly. We probably need to do an audit for them (in terms of what is and isn't documented), and still make a few decisions.

[...]

PR #999, opened 2022-07-28
Add experimental YCgCo-R support

vrabaud:
It is all protected by a C++ define and disabled by default in CMake. For now, it is just a proposal as an update to H263 with a new matrix coefficient for YCgCo-R.
It is specified here: https://jvet-experts.org/doc_end_user/current_document.php?id=10608


jzern:
I didn't notice the paper title and was expecting 'YCoCg-R', which seems more common, though they might be used interchangeably.

vrabaud:
I know ... H263 actually mentions YCgCo (cf https://www.itu.int/rec/T-REC-H.273-202107-I/fr) and libavif is also following that convention for the non-reversible one: https://github.com/AOMediaCodec/libavif/blob/3c12bb64f1476e49c69e32d49f0623bf3bc31a67/include/avif/avif.h#L286

apps/avifenc.c
```diff
@@ -730,17 +730,7 @@ int main(int argc, char * argv[])
...
-            maxQuantizer = AVIF_QUANTIZER_LOSSLESS;           // lossless
-            minQuantizerAlpha = AVIF_QUANTIZER_LOSSLESS;      // lossless
-            maxQuantizerAlpha = AVIF_QUANTIZER_LOSSLESS;      // lossless
-            codecChoice = AVIF_CODEC_CHOICE_AOM;              // rav1e doesn't support lossless transform yet:
-                                                              // https://github.com/xiph/rav1e/issues/151
-                                                              // SVT-AV1 doesn't support lossless encoding yet:
-                                                              // https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1636
-            requestedRange = AVIF_RANGE_FULL;                 // avoid limited range
-            matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_IDENTITY; // this is key for lossless
+            requestedRange = AVIF_RANGE_FULL; // avoid limited range
```
jzern, line 733: this is duplicated in the if (lossless) block added below.

apps/shared/avifpng.c
```diff
@@ -181,10 +190,19 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+        const avifBool useYCgCoR = (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_R);
+#else
+        const avifBool useYCgCoR = AVIF_FALSE;
+#endif
+        if (useYCgCoR) {
+            rgbDepth = avif->depth - 2;
         } else {
-            rgbDepth = 8;
+            if (avif->depth > 8) {
```
jzern, line 201: ?:s seems common enough, this could be one line

src/reformat.c
```diff
@@ -293,6 +310,21 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                         yuvBlock[bI][bJ].y = 0.5f * rgbPixel[1] + 0.25f * (rgbPixel[0] + rgbPixel[2]);
                         yuvBlock[bI][bJ].u = 0.5f * rgbPixel[1] - 0.25f * (rgbPixel[0] + rgbPixel[2]);
                         yuvBlock[bI][bJ].v = 0.5f * (rgbPixel[0] - rgbPixel[2]);
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+                    } else if (state.mode == AVIF_REFORMAT_MODE_YCGCO_R) {
+#else
+                    } else if (AVIF_FALSE) {
```
jzern, line 316: why not wrap the whole block with the #if?

src/reformat.c
```diff
@@ -616,6 +648,22 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
                     G = Y + Cb;
                     B = t - Cr;
                     R = t + Cr;
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+                } else if (state->mode == AVIF_REFORMAT_MODE_YCGCO_R) {
+#else
+                } else if (AVIF_FALSE) {
```
jzern, line 654: same here, you can avoid the #else with a #if around the whole block

wantehchang:
Vincent: Thank you for the pull request. I will review it on Monday.

CMakeLists.txt
```diff
@@ -28,6 +28,8 @@ option(BUILD_SHARED_LIBS "Build shared avif library" ON)
 
 option(AVIF_ENABLE_WERROR "Treat all compiler warnings as errors" ON)
 
+option(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R "Enable experimental YCgCo-R matrix code" OFF)
```
wantehchang, line 31: We can consider a single option that enables all experimental features. See https://github.com/AOMediaCodec/libavif/issues/766.

src/reformat.c
```diff
@@ -30,12 +30,20 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
     //
     // YCgCo performs limited-full range adjustment on R,G,B but the current implementation performs range adjustment
     // on Y,U,V. So YCgCo with limited range is unsupported.
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+    const avifBool useYCgCoR = (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_R);
+    const avifMatrixCoefficients lastMC = AVIF_MATRIX_COEFFICIENTS_YCGCO_R;
```
wantehchang, line 35: Nit: It may be better to add an `AVIF_MATRIX_COEFFICIENTS_LAST` or `AVIF_MATRIX_COEFFICIENTS_MAX` value to the enum type in avif.h.

tests/test_cmd.sh
```diff
@@ -37,23 +37,37 @@ ARE_IMAGES_EQUAL="${BINARY_DIR}/tests/are_images_equal"
...
 DECODED_FILE=/tmp/decoded.png
 PNG_FILE=/tmp/kodim03.png
+JPG_FILE=/tmp/kodim03.jpg
 
 # Prepare some extra data.
 set +x
 echo "Generating a color PNG"
 "${AVIFENC}" -s 10 "${TESTDATA_DIR}"/kodim03_yuv420_8bpc.y4m -o "${TMP_ENCODED_FILE}" &> /dev/null
 "${AVIFDEC}" "${TMP_ENCODED_FILE}" "${PNG_FILE}"  &> /dev/null
+"${AVIFDEC}" "${TMP_ENCODED_FILE}" "${JPG_FILE}"  &> /dev/null
```
wantehchang, line 47: Should the message at line 44 also mention JPEG now?

tests/test_cmd.sh
```diff
@@ -37,23 +37,37 @@ ARE_IMAGES_EQUAL="${BINARY_DIR}/tests/are_images_equal"
...
-echo "Testing basic lossless"
-"${AVIFENC}" -s 10 -l "${PNG_FILE}" -o "${TMP_ENCODED_FILE}"
-"${AVIFDEC}" "${TMP_ENCODED_FILE}" "${DECODED_FILE}"
-"${ARE_IMAGES_EQUAL}" "${PNG_FILE}" "${DECODED_FILE}" 0
+echo "Testing basic lossless on PNG and JPEG"
+for f in ${PNG_FILE} ${JPG_FILE}; do
+  "${AVIFENC}" -s 10 -l "${f}" -o "${TMP_ENCODED_FILE}"
+  "${AVIFDEC}" "${TMP_ENCODED_FILE}" "${DECODED_FILE}"
+  "${ARE_IMAGES_EQUAL}" "${f}" "${DECODED_FILE}" 1
+  echo "YCgCo works on ${f}"
```
wantehchang, line 60: Should this message say "Identity" or "RGB" instead of "YCgCo"?

include/avif/avif.h
```diff
@@ -289,7 +289,10 @@ enum
     AVIF_MATRIX_COEFFICIENTS_SMPTE2085 = 11,
     AVIF_MATRIX_COEFFICIENTS_CHROMA_DERIVED_NCL = 12,
     AVIF_MATRIX_COEFFICIENTS_CHROMA_DERIVED_CL = 13,
-    AVIF_MATRIX_COEFFICIENTS_ICTCP = 14
+    AVIF_MATRIX_COEFFICIENTS_ICTCP = 14,
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+    AVIF_MATRIX_COEFFICIENTS_YCGCO_R = 15,
```
wantehchang, line 294: Should this be named `AVIF_MATRIX_COEFFICIENTS_YCGCO_R`, because it is called YCgCo-Ro in the proposed change to the standard?

tests/gtest/are_images_equal.cc
```diff
@@ -7,7 +7,8 @@
 #include "aviftest_helpers.h"
 #include "avifutil.h"
 
-using libavif::testutil::AvifImagePtr;
+using ::libavif::testutil::AreImagesEqual;
+using ::libavif::testutil::AvifImagePtr;
```
wantehchang, line 11: Nit/optional: Just wondering why the root namespace :: is used with the libavif namespace in these two using statements. We don't use the root namespace :: with the std namespace in this file, for example.

wantehchang (changes requested):
Vincent: I did a partial review today. I will continue to review this tomorrow.

tests/gtest/are_images_equal.cc
maryla-uc: It's recommended for "using" statements, e.g. in https://abseil.io/tips/119 "When declaring namespace aliases and using-declarations, use fully qualified names (with leading `::`) unless you are referring to a name inside the current namespace."

src/reformat.c
```diff
@@ -616,6 +648,22 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
...
+                    const int16_t YY = unormY;
+                    const int16_t Cg = Cb * yuvMaxChannel;
+                    const int16_t Co = Cr * yuvMaxChannel;
+                    const int16_t t = YY - (Cg >> 1);
+                    G = AVIF_CLAMP(t + Cg, 0, maxValue);
+                    B = AVIF_CLAMP(t - (Co >> 1), 0, maxValue);
+                    R = AVIF_CLAMP(B + Co, 0, maxValue);
+                    G /= rgbMaxChannelF;
+                    B /= rgbMaxChannelF;
+                    R /= rgbMaxChannelF;
```
wantehchang, line 666: General comment: Same here. This implementation is inefficient. It is fine for exporing the comression ratio of YCoCg-R, but we will need to redo the implementation when we start to care about speed.

wantehchang (changes requested):
Vincent: I read everything carefully. I focused on correctness in this initial review. The `#if` conditionals made the code messy, but I am ignoring that issue for now (partly because I haven't tried to come up with a good alternative).

apps/avifenc.c
vrabaud: Indeed, that was fixed in https://github.com/AOMediaCodec/libavif/pull/1179

apps/shared/avifjpeg.c
```diff
@@ -297,12 +297,25 @@ avifBool avifJPEGRead(const char * inputFilename, avifImage * avif, avifPixelFor
...
-                                                                                              : AVIF_APP_DEFAULT_PIXEL_FORMAT;
+            // Identity and YCgCo-R are only valid with YUV444.
+            avif->yuvFormat = (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY || useYCgCoR)
+                                  ? AVIF_PIXEL_FORMAT_YUV444
+                                  : AVIF_APP_DEFAULT_PIXEL_FORMAT;
         }
         avif->depth = requestedDepth ? requestedDepth : 8;
+        if (useYCgCoR) {
+            if (requestedDepth == 12) {
+                fprintf(stderr, "Cannot perform YCgCo-R on 12 bit input.\n");
```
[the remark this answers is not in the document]
vrabaud: Re-worded

include/avif/avif.h
vrabaud: Indeed: I added support for RE and RO

src/reformat.c
```diff
@@ -293,6 +310,21 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
...
+#else
+                    } else if (AVIF_FALSE) {
+#endif
+                        // Formulas from JVET-U0093.
+                        const int16_t R = avifRoundf(rgbPixel[0] * rgbMaxChannelF);
+                        const int16_t G = avifRoundf(rgbPixel[1] * rgbMaxChannelF);
+                        const int16_t B = avifRoundf(rgbPixel[2] * rgbMaxChannelF);
+                        const int16_t Co = R - B;
+                        const int16_t t = B + (Co >> 1);
+                        const int16_t Cg = G - t;
```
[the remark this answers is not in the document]
vrabaud: I usually only use what is needed so that it makes the SIMD optimization easier but your call.

src/reformat.c
```diff
@@ -293,6 +310,21 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
...
+                        yuvBlock[bI][bJ].y = (t + (Cg >> 1)) / state.rangeY;
+                        yuvBlock[bI][bJ].u = Cg / state.rangeUV;
+                        yuvBlock[bI][bJ].v = Co / state.rangeUV;
```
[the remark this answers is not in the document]
vrabaud: Indeed. All that code is using floats anyway which is inefficient.

tests/test_cmd.sh
vrabaud: This is now a C++ test.

wantehchang:
Vincent: I will try to review this pull request tomorrow.

wantehchang:
Vincent: I created a copy of this pull request in Gerrit for code review: https://aomedia-review.googlesource.com/c/libavif/+/174864

I will read the code there and post my review comments here.

src/reformat.c
```diff
@@ -701,6 +734,22 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
                     G = Y + Cb;
                     B = t - Cr;
                     R = t + Cr;
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+                } else if (state->mode == AVIF_REFORMAT_MODE_YCGCO_RE || state->mode == AVIF_REFORMAT_MODE_YCGCO_RO) {
+                    const int bit_offset = (state->mode == AVIF_REFORMAT_MODE_YCGCO_RE) ? 2 : 1;
```
wantehchang, line 739: Nit: bit_offset => bitOffset

tests/gtest/aviflosslesstest.cc
```diff
@@ -21,8 +21,12 @@ TEST(BasicTest, EncodeDecodeMatrixCoefficients) {
     const testutil::AvifImagePtr ground_truth_image =
         testutil::ReadImage(data_path, file_name);
 
-    for (auto matrix_coefficient :
-         {AVIF_MATRIX_COEFFICIENTS_IDENTITY, AVIF_MATRIX_COEFFICIENTS_YCGCO}) {
+    for (auto& matrix_coefficient : {
```
wantehchang, line 24: Nit: It doesn't seem necessary to add `&`, because it is cheap to copy the `AVIF_MATRIX_COEFFICIENTS_xxx` values.

tests/gtest/aviflosslesstest.cc
```diff
@@ -36,7 +40,18 @@ TEST(BasicTest, EncodeDecodeMatrixCoefficients) {
...
           /*outDepth=*/nullptr, /*sourceTiming=*/nullptr,
           /*frameIter=*/nullptr);
-      ASSERT_NE(file_format, AVIF_APP_FILE_FORMAT_UNKNOWN);
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+      if (matrix_coefficient == AVIF_MATRIX_COEFFICIENTS_YCGCO_RO) {
+        // AVIF_MATRIX_COEFFICIENTS_YCGCO_RO does not work because the input
+        // depth is not odd.
+        ASSERT_EQ(file_format, AVIF_APP_FILE_FORMAT_UNKNOWN);
+        continue;
+      } else {
```
wantehchang, line 49: Omit this `else` because it follows a `continue` statement. This will allow us t delete lines 52-54.

wantehchang (changes requested):
Vincent: Sorry about the late review. I reviewed everything, but I didn't check the equations in reformat.c carefully. I will check that in the next round of review.

Also, I read the new code directly without referring to my last round of review comments. So please pardon me if I repeated some review comments.

src/reformat.c
```diff
@@ -701,6 +734,22 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
...
+                    const int maxValue = (1 << (state->yuvDepth - bit_offset)) - 1;
+                    assert((float)maxValue == state->rgbMaxChannelF);
```
wantehchang, line 741: Nit: state->rgbMaxChannelF => rgbMaxChannelF

src/reformat.c
```diff
@@ -313,6 +333,19 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                             yuvBlock[bI][bJ].y = 0.5f * rgbPixel[1] + 0.25f * (rgbPixel[0] + rgbPixel[2]);
                             yuvBlock[bI][bJ].u = 0.5f * rgbPixel[1] - 0.25f * (rgbPixel[0] + rgbPixel[2]);
                             yuvBlock[bI][bJ].v = 0.5f * (rgbPixel[0] - rgbPixel[2]);
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+                        } else if (state.mode == AVIF_REFORMAT_MODE_YCGCO_RE || state.mode == AVIF_REFORMAT_MODE_YCGCO_RO) {
+                            // Formulas from JVET-U0093.
+                            const int R = avifRoundf(rgbPixel[0] * rgbMaxChannelF);
+                            const int G = avifRoundf(rgbPixel[1] * rgbMaxChannelF);
+                            const int B = avifRoundf(rgbPixel[2] * rgbMaxChannelF);
```
wantehchang, line 341: In lines 339-341, should we apply Clip3() before avifRoundf()?

See Formulae (31) to (33) in ISO/IEC DIS 23091-2:2022(E).

src/reformat.c
```diff
@@ -313,6 +333,19 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                             yuvBlock[bI][bJ].y = 0.5f * rgbPixel[1] + 0.25f * (rgbPixel[0] + rgbPixel[2]);
                             yuvBlock[bI][bJ].u = 0.5f * rgbPixel[1] - 0.25f * (rgbPixel[0] + rgbPixel[2]);
                             yuvBlock[bI][bJ].v = 0.5f * (rgbPixel[0] - rgbPixel[2]);
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+                        } else if (state.mode == AVIF_REFORMAT_MODE_YCGCO_RE || state.mode == AVIF_REFORMAT_MODE_YCGCO_RO) {
+                            // Formulas from JVET-U0093.
```
wantehchang, line 338: I don't have JVET-U0093. I am using ISO/IEC DIS 23091-2:2022(E) (https://www.iso.org/obp/ui/#iso:std:iso-iec:23091:-2:dis:ed-3:v1:en).

wantehchang (commented):
Vincent: I have checked the equations in reformat.c. Here are a few more comments.

src/reformat.c
wantehchang: Update: I found JVET-U0093. I will compare its equations with the equations in ISO/IEC DIS 23091-2:2022(E) tomorrow.

In any case, your code seems to match the equations in ISO/IEC DIS 23091-2:2022(E).

apps/avifenc.c
```diff
@@ -1479,11 +1479,19 @@ int main(int argc, char * argv[])
...
-            fprintf(stderr, "Matrix coefficients have to be identity in lossless mode.\n");
-            returnCode = 1;
+        if (cicpExplicitlySet) {
+            avifBool changeMC = (settings.matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_IDENTITY);
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+            changeMC &= (settings.matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_YCGCO_RE &&
+                         settings.matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_YCGCO_RO);
+#endif
+            if (changeMC) {
+                fprintf(stderr, "Matrix coefficients have to be identity in lossless mode.\n");
```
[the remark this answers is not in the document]
vrabaud: No, this is just an extension of the conditions to when that error message is displayed. Before, it was only sent when CICP was set and the matrix coefficients not identity. Now, we also allow MC to be YCGCO_RE and _RO.

apps/avifenc.c
```diff
@@ -1479,11 +1479,19 @@ int main(int argc, char * argv[])
...
+                returnCode = 1;
+            }
+        } else {
```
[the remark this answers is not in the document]
vrabaud: No because we want to keep the MC is set to YCGCO_RE or _RO. We set to identity only by default.

apps/avifenc.c
```diff
@@ -949,7 +957,7 @@ int main(int argc, char * argv[])
             lossless = AVIF_FALSE;
         }
 
-        if (!depthMatches) {
+        if (usingIdentityMatrix && (sourceDepth != image->depth)) {
```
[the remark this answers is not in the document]
vrabaud: Line 962 is empty, which line are you referring to ? Thx

apps/shared/avifjpeg.c
```diff
@@ -356,12 +356,28 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+        const avifBool useYCgCoR = (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE ||
+                                    avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RO);
+#else
+        const avifBool useYCgCoR = AVIF_FALSE;
+#endif
         if (avif->yuvFormat == AVIF_PIXEL_FORMAT_NONE) {
-            // Identity is only valid with YUV444.
-            avif->yuvFormat = (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) ? AVIF_PIXEL_FORMAT_YUV444
-                                                                                              : AVIF_APP_DEFAULT_PIXEL_FORMAT;
+            // Identity and YCgCo-R are only valid with YUV444.
```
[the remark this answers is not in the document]
vrabaud: YCgCo-R is reversible so we do not want to do chroma subsampling. If the input in 420 or 400, we could do YCgCO-R but we would need to check the input is indeed that way.

apps/shared/avifpng.c
```diff
@@ -313,9 +313,15 @@ avifBool avifPNGRead(const char * inputFilename,
...
+    const avifBool useYCgCoR = (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE ||
+                                avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RO);
+#else
+    const avifBool useYCgCoR = AVIF_FALSE;
+#endif
     if (avif->yuvFormat == AVIF_PIXEL_FORMAT_NONE) {
-        if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
-            // Identity is only valid with YUV444.
+        if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY || useYCgCoR) {
+            // Identity and YCgCo-R is only valid with YUV444.
```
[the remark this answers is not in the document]
vrabaud: Done

apps/shared/avifpng.c
```diff
@@ -325,7 +331,11 @@ avifBool avifPNGRead(const char * inputFilename,
     }
     avif->depth = requestedDepth;
     if (avif->depth == 0) {
-        if (imgBitDepth == 8) {
+        if (useYCgCoR) {
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+            avif->depth = imgBitDepth + ((avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE) ? 2 : 1);
```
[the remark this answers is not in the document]
vrabaud: Good catch, fixed.

apps/shared/avifpng.c
```diff
@@ -396,11 +406,17 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
+                                    avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RO);
+        if (useYCgCoR) {
+            rgbDepth = avif->depth - ((avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE) ? 2 : 1);
         } else {
-            rgbDepth = 8;
+#endif
+            rgbDepth = (avif->depth > 8) ? 16 : 8;
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
         }
+#endif
```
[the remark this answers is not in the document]
vrabaud: Done

include/avif/internal.h
```diff
@@ -120,7 +120,11 @@ typedef enum avifReformatMode
 {
     AVIF_REFORMAT_MODE_YUV_COEFFICIENTS = 0, // Normal YUV conversion using coefficients
     AVIF_REFORMAT_MODE_IDENTITY,             // Pack GBR directly into YUV planes (AVIF_MATRIX_COEFFICIENTS_IDENTITY)
-    AVIF_REFORMAT_MODE_YCGCO                 // YUV conversion using AVIF_MATRIX_COEFFICIENTS_YCGCO
+    AVIF_REFORMAT_MODE_YCGCO,                // YUV conversion using AVIF_MATRIX_COEFFICIENTS_YCGCO
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+    AVIF_REFORMAT_MODE_YCGCO_RE, // YUV conversion using AVIF_MATRIX_COEFFICIENTS_YCGCO_RE
+    AVIF_REFORMAT_MODE_YCGCO_RO  // YUV conversion using AVIF_MATRIX_COEFFICIENTS_YCGCO_RO
```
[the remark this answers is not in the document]
vrabaud: Done.

src/reformat.c
vrabaud: Done.

src/reformat.c
vrabaud: Done

src/reformat.c
vrabaud: Done

tests/gtest/aviflosslesstest.cc
vrabaud: Done

tests/gtest/aviflosslesstest.cc
vrabaud: Done

apps/shared/avifpng.c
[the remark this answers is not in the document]
wantehchang: The new code still needs work. I will resolve this conversation and add a new comment about the new code.

apps/shared/avifpng.c
```diff
@@ -396,11 +407,16 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
 
     volatile int rgbDepth = requestedDepth;
     if (rgbDepth == 0) {
-        if (avif->depth > 8) {
-            rgbDepth = 16;
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+        if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE ||
+                                    avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RO) {
+            rgbDepth = avif->depth - ((avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE) ? 2 : 1);
```
wantehchang, line 413: IMPORTANT: For PNG, `rgbDepth` can only be 8 or 16. Since `avif->depth` is 8, 10, or 12, the only legal combination here is `avif->depth == 10`, `avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE`, and `rgbDepth == 8`.

I will let you work out the details.

src/reformat.c
```diff
@@ -313,6 +333,21 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
...
+                            const int bitOffset = (state.mode == AVIF_REFORMAT_MODE_YCGCO_RE) ? 2 : 1;
+                            const int maxValue = (1 << (state.yuvDepth - bitOffset)) - 1;
+                            const int R = AVIF_CLAMP(avifRoundf(rgbPixel[0] * rgbMaxChannelF), 0, maxValue);
+                            const int G = AVIF_CLAMP(avifRoundf(rgbPixel[1] * rgbMaxChannelF), 0, maxValue);
+                            const int B = AVIF_CLAMP(avifRoundf(rgbPixel[2] * rgbMaxChannelF), 0, maxValue);
```
wantehchang, line 343: Optional: In ISO/IEC DIS 23091-2:2022(E), the Clip3() operation is applied before the Round() operation, so the nesting of `AVIF_CLAMP` and `avifRoundf` should be reversed. For example,

```
                            assert((float)maxValue == rgbMaxChannelF);
                            const int R = avifRoundf(AVIF_CLAMP(rgbPixel[0] * rgbMaxChannelF, 0.0f, rgbMaxChannelF));
                            const int G = avifRoundf(AVIF_CLAMP(rgbPixel[1] * rgbMaxChannelF, 0.0f, rgbMaxChannelF));
                            const int B = avifRoundf(AVIF_CLAMP(rgbPixel[2] * rgbMaxChannelF, 0.0f, rgbMaxChannelF));
```

Note 1: I believe the order of Clip3() and Round() doesn't matter here. If you agree, then we don't really need to change this code. But it would be better to follow the spec more closely if it is convenient.

Note 2: JVET-U0093 applies the Clip1Y() and Round() operations in the same order. Clip1Y() is first applied in the three equations at the very end of Section 2. Round() is then applied in the first three equations at the beginning of Section 2.

wantehchang (commented):
Vincent: In this review, I realized that YCGCO_RO can't be used with AVIF if the RGB bit depth is an even number such as 8. Also, since the RGB output of libjpeg always has bit depth 8, and the RGB bit depth of PNG can only be 8 or 16, the corresponding AVIF bit depth can only be 10 if YCGCO_RE is used.

The code snippets I suggested in avifjpeg.c and avifpng.c are only meant to be sample code. They may have errors. Please review them carefully and feel free to rewrite them.

apps/avifenc.c
```diff
@@ -1479,11 +1479,19 @@ int main(int argc, char * argv[])
             returnCode = 1;
         }
         // Matrix coefficients.
-        if (cicpExplicitlySet && settings.matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
-            fprintf(stderr, "Matrix coefficients have to be identity in lossless mode.\n");
-            returnCode = 1;
+        if (cicpExplicitlySet) {
+            avifBool changeMC = (settings.matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_IDENTITY);
```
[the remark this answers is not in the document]
vrabaud: Done

apps/avifenc.c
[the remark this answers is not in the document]
vrabaud: Done

apps/avifenc.c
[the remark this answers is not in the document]
vrabaud: Yes, exactly

apps/avifenc.c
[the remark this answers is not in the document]
vrabaud: Done

apps/shared/avifjpeg.c
```diff
@@ -356,12 +356,28 @@ avifBool avifJPEGRead(const char * inputFilename,
...
         avif->depth = requestedDepth ? requestedDepth : 8;
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+        if (useYCgCoR) {
+            if (requestedDepth == 12) {
+                fprintf(stderr, "Cannot request 12 bits for YCgCo-R as it uses 1 or 2 extra bits.\n");
+                goto cleanup;
+            }
+            avif->depth += (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE) ? 2 : 1;
+        }
+#endif
```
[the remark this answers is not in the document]
vrabaud: I wanted to keep the code flexible but indeed, JPEG and PNG are not going to change any time soon ...

apps/shared/avifpng.c
```diff
@@ -330,6 +336,11 @@ avifBool avifPNGRead(const char * inputFilename,
         } else {
             avif->depth = 12;
         }
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+        if (useYCgCoR) {
+            avif->depth += (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE) ? 2 : 1;
+        }
+#endif
```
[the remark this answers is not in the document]
vrabaud: Done

apps/shared/avifpng.c
vrabaud: Done

src/reformat.c
vrabaud: Done

vrabaud:
The latest version of the spec is here BTW: https://jvet-experts.org/doc_end_user/documents/29_Teleconference/wg11/JVET-AC1008-v2.zip

apps/shared/avifpng.c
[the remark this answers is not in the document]
wantehchang: I think lines 339-355 should be moved after line 356 so that they are outside the `if (avif->depth == 0) { ... }` block.

I.e., the code should look like the following:

```
    avif->depth = requestedDepth;
    if (avif->depth == 0) {
        if (imgBitDepth == 8) {
            avif->depth = 8;
        } else {
            avif->depth = 12;
        }
    }
#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
    if (useYCgCoR) {
        ...
    }
#endif
```

apps/shared/avifpng.c
```diff
@@ -396,11 +419,15 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
-        } else {
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+        if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE) {
             rgbDepth = 8;
+        } else {
+            rgbDepth = (avif->depth > 8) ? 16 : 8;
         }
+#else
+        rgbDepth = (avif->depth > 8) ? 16 : 8;
+#endif
```
wantehchang, line 430: This is missing error checking. I think lines 421-431 should read something like the following:

```
    if (rgbDepth == 0) {
        rgbDepth = (avif->depth > 8) ? 16 : 8;
    }
#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
    if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RO) {
        fprintf(stderr, "AVIF_MATRIX_COEFFICIENTS_YCGCO_RO cannot be used with PNG because it has an even bit depth.\n");
        goto cleanup;
    }
    if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE) {
        if (avif->depth != 10 || (requestedDepth && requestedDepth != 8)) {
            fprintf(stderr, "...error message...\n");
            goto cleanup;
        }
        rgbDepth = 8;
    }   
#endif
```

src/reformat.c
```diff
@@ -313,6 +333,21 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
...
                             yuvBlock[bI][bJ].u = 0.5f * rgbPixel[1] - 0.25f * (rgbPixel[0] + rgbPixel[2]);
                             yuvBlock[bI][bJ].v = 0.5f * (rgbPixel[0] - rgbPixel[2]);
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+                        } else if (state.mode == AVIF_REFORMAT_MODE_YCGCO_RE || state.mode == AVIF_REFORMAT_MODE_YCGCO_RO) {
+                            // Formulas from JVET-U0093.
+                            const int bitOffset = (state.mode == AVIF_REFORMAT_MODE_YCGCO_RE) ? 2 : 1;
+                            const int maxValue = (1 << (state.yuvDepth - bitOffset)) - 1;
+                            const int R = avifRoundf(AVIF_CLAMP(rgbPixel[0] * rgbMaxChannelF, 0, maxValue));
+                            const int G = avifRoundf(AVIF_CLAMP(rgbPixel[1] * rgbMaxChannelF, 0, maxValue));
+                            const int B = avifRoundf(AVIF_CLAMP(rgbPixel[2] * rgbMaxChannelF, 0, maxValue));
```
wantehchang, line 343: Nit: In lines 341-343, change `0, maxValue` to `0.0f, rgbMaxChannelF` so that the constants have the correct type (float).

wantehchang (approved):
LGTM with three suggested changes.

apps/shared/avifpng.c
vrabaud: Done

src/reformat.c
vrabaud: Done

apps/shared/avifpng.c
```diff
@@ -396,12 +418,22 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
+        rgbDepth = (avif->depth > 8) ? 16 : 8;
+    }
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+    if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RO) {
+        fprintf(stderr, "AVIF_MATRIX_COEFFICIENTS_YCGCO_RO cannot be used with PNG because it has an even bit depth.\n");
+        goto cleanup;
+    }
+    if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE) {
+        if (avif->depth != 10 || (requestedDepth && requestedDepth != 8)) {
+            fprintf(stderr, "Cannot request %u bits for YCgCo-Re as it only works for 8 bits.\n", requestedDepth);
```
wantehchang, line 430: Nit: This error message is for the `requestedDepth && requestedDepth != 8` condition only. I guess we should check the `avif->depth != 10` condition separately so that we can print a different error message. Alternatively, we can modify this error message to also say avif->depth must be 10.

src/reformat.c
```diff
@@ -336,11 +336,9 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
...
                         } else if (state.mode == AVIF_REFORMAT_MODE_YCGCO_RE || state.mode == AVIF_REFORMAT_MODE_YCGCO_RO) {
                             // Formulas from JVET-U0093.
-                            const int bitOffset = (state.mode == AVIF_REFORMAT_MODE_YCGCO_RE) ? 2 : 1;
-                            const int maxValue = (1 << (state.yuvDepth - bitOffset)) - 1;
-                            const int R = avifRoundf(AVIF_CLAMP(rgbPixel[0] * rgbMaxChannelF, 0, maxValue));
-                            const int G = avifRoundf(AVIF_CLAMP(rgbPixel[1] * rgbMaxChannelF, 0, maxValue));
-                            const int B = avifRoundf(AVIF_CLAMP(rgbPixel[2] * rgbMaxChannelF, 0, maxValue));
+                            const int R = (int)avifRoundf(AVIF_CLAMP(rgbPixel[0] * rgbMaxChannelF, 0.f, rgbMaxChannelF));
+                            const int G = (int)avifRoundf(AVIF_CLAMP(rgbPixel[1] * rgbMaxChannelF, 0.f, rgbMaxChannelF));
+                            const int B = (int)avifRoundf(AVIF_CLAMP(rgbPixel[2] * rgbMaxChannelF, 0.f, rgbMaxChannelF));
```
wantehchang, line 341: Nit: 0.f => 0.0f

See https://google.github.io/styleguide/cppguide.html#Floating_Literals

vrabaud:
Fixed in https://github.com/AOMediaCodec/libavif/pull/1392

PR #1003, opened 2022-08-01
Build libyuv in the AppVeyor build

wantehchang:


ext/libyuv.cmd
```diff
@@ -16,5 +16,5 @@ mkdir build
 cd build
 
 cmake -G Ninja -DBUILD_SHARED_LIBS=0 -DCMAKE_BUILD_TYPE=Release ..
-ninja yuv
+ninja -w dupbuild=warn yuv
```
jzern, line 19: something to be reported for libyuv?

jzern (approved):

ext/libyuv.cmd
wantehchang: Yes, I just fixed this problem in libyuv:
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/3806073

I will try to fix the MSVC compiler warnings in libyuv and then update our libyuv.cmd script to check out the new revision.

kmilos: Btw, please don't use WIN32 in CMake to assume MSVC - WIN32 covers both MINGW and MSVC. So either use MSVC if you really mean that, or perhaps "WIN32 AND NOT MINGW". For example, MINGW builds libXYZ.dll shared lib, libXYZ.dll.a import lib, and libXYZ.a static lib by default.

wantehchang: Hi Miloš,

Thank you for your comment. Although I don't use MINGW, I am somewhat familiar with such porting issues. I wrote https://chromium-review.googlesource.com/c/libyuv/libyuv/+/3806073 for any toolchain on Windows. But we cannot be sure unless we test it. I haven't tested it with MINGW.

In another CL (https://chromium-review.googlesource.com/c/libyuv/libyuv/+/3807104) about the MSVC C runtime library (CRT), I test `MSVC` instead of `WIN32`. Is that correct?

Thanks.

kmilos: Yes, that second one is correct.

Btw, one could also add mingw builds easily (at least to libavif CI) via the [setup-msys2 action](https://github.com/msys2/setup-msys2).

PR #1007, opened 2022-08-04
Fix avifallocationtest for all environments

y-guyon:


tests/gtest/avifallocationtest.cc
```diff
@@ -58,22 +58,16 @@ TEST(AllocationTest, MinimumValidDimensions) {
...
-  TestAllocation(kMaxAllocatableDimension, 1, 8, AVIF_RESULT_OK);
-  TestAllocation(1, kMaxAllocatableDimension, 8, AVIF_RESULT_OK);
-  // 12 bits (impacts the width because avifImage stride is stored as uint32_t)
-  TestAllocation(kMaxAllocatableDimension / 2, 1, 12, AVIF_RESULT_OK);
-  TestAllocation(1, kMaxAllocatableDimension, 12, AVIF_RESULT_OK);
-  // Some high number of bytes that malloc() accepts to allocate.
-  TestAllocation(1024 * 16, 1024 * 8, 12, AVIF_RESULT_OK);  // Up to 1 GB total
+TEST(AllocationTest, MaximumAllocatableValidDimensions) {
+  // Limit the total allocation to 1 GB to prevent any failure no matter the
+  // platform or environment.
```
maryla-uc, line 63: In this test you're only checking that you *can* allocate 1GB, but not that you *can't* allocate more. Shouldn't you test e.g. (1 << 30) / 4 + 1 if (1 << 30) / 4 really is the max valid dimension?
I see other tests in this file testing bad dimensions but they all use numeric_limits::max() instead.

y-guyon: It depends on the platform and the environment (build/test farms have lower RAM limits, 32-bit have 1 or 2GB max etc.) so to avoid the pain of having to fine tune this test for each running context, I believe it is simpler to have low passing thresholds and high failing thresholds that work everywhere.

tests/gtest/avifallocationtest.cc
```diff
@@ -58,22 +58,17 @@ TEST(AllocationTest, MinimumValidDimensions) {
...
-  // 8 bits
-  TestAllocation(kMaxAllocatableDimension, 1, 8, AVIF_RESULT_OK);
-  TestAllocation(1, kMaxAllocatableDimension, 8, AVIF_RESULT_OK);
-  // 12 bits (impacts the width because avifImage stride is stored as uint32_t)
-  TestAllocation(kMaxAllocatableDimension / 2, 1, 12, AVIF_RESULT_OK);
-  TestAllocation(1, kMaxAllocatableDimension, 12, AVIF_RESULT_OK);
-  // Some high number of bytes that malloc() accepts to allocate.
-  TestAllocation(1024 * 16, 1024 * 8, 12, AVIF_RESULT_OK);  // Up to 1 GB total
+// The libavif API accepts allocating up to SIZE_MAX bytes. Testing this
+// threshold is unrealistic so allocate 1 GB: that should pass on all platforms
```
wantehchang, line 62: Nit: The comment "The libavif API accepts allocating up to SIZE_MAX bytes." is a little confusing because it seems to imply that up to `SIZE_MAX` bytes can be allocated. I suggest rephrasing the comment, for example, "Up to SIZE_MAX can be passed as the input argument to avifAlloc()."

It is also confusing to refer to `SIZE_MAX` bytes as a threshold. `SIZE_MAX` is the maximum value of the `size_t` type (the type of the input parameter of `avifAlloc()`). `SIZE_MAX` is not a threshold for memory allocation (i.e., the maximum number of bytes that can be allocated).

y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1015

PR #1008, opened 2022-08-04
Add support RGB565 format

vigneshvg:
RGB565 [1] is a pixel format used by Android. It stores each pixel
in 2 bytes (with 5, 6 and 5 bits respectively for R, G and B
channels).

This is used in some older android devices which have low RAM to
minimize the size of bitmaps on those devices.

[1] https://developer.android.com/reference/android/graphics/Bitmap.Config#RGB_565

### DIFF
--- a/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
+++ b/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
@@ -127,8 +127,9 @@ FUNC(jboolean, decode, jobject encoded, int length, jobject bitmap) {
         decoder.decoder->image->height);
     return false;
   }
-  // Ensure that the bitmap format is either RGBA_8888 or RGBA_F16.
+  // Ensure that the bitmap format is RGBA_8888, RGB_565 or RGBA_F16.
   if (bitmap_info.format != ANDROID_BITMAP_FORMAT_RGBA_8888 &&
+      bitmap_info.format != ANDROID_BITMAP_FORMAT_RGB_565 &&
       bitmap_info.format != ANDROID_BITMAP_FORMAT_RGBA_F16) {
     LOGE("Bitmap format (%d) is not supported.", bitmap_info.format);
     return false;
@@ -144,6 +145,9 @@ FUNC(jboolean, decode, jobject encoded, int length, jobject bitmap) {
   if (bitmap_info.format == ANDROID_BITMAP_FORMAT_RGBA_F16) {
     rgb_image.depth = 16;
     rgb_image.isFloat = AVIF_TRUE;
+  } else if (bitmap_info.format == ANDROID_BITMAP_FORMAT_RGB_565) {
+    rgb_image.format = AVIF_RGB_FORMAT_RGB_565;
+    rgb_image.depth = 8;
   } else {
     rgb_image.depth = 8;
   }
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -551,7 +551,17 @@ typedef enum avifRGBFormat
     AVIF_RGB_FORMAT_ARGB,
     AVIF_RGB_FORMAT_BGR,
     AVIF_RGB_FORMAT_BGRA,
-    AVIF_RGB_FORMAT_ABGR
+    AVIF_RGB_FORMAT_ABGR,
+    // RGB_565 format uses five bits for the red and blue components and six
+    // bits for the green component. Each RGB pixel is 16 bits (2 bytes), which
+    // is packed as follows:
+    //   uint16_t: [r4 r3 r2 r1 r0 g5 g4 g3 g2 g1 g0 b4 b3 b2 b1 b0]
+    //   r4 and r0 are the MSB and LSB of the red component respectively.
+    //   g5 and g0 are the MSB and LSB of the green component respectively.
+    //   b4 and b0 are the MSB and LSB of the blue component respectively.
+    // This format is only supported for YUV -> RGB conversion and when
+    // avifRGBImage.depth is set to 8.
+    AVIF_RGB_FORMAT_RGB_565
 } avifRGBFormat;
 AVIF_API uint32_t avifRGBFormatChannelCount(avifRGBFormat format);
 AVIF_API avifBool avifRGBFormatHasAlpha(avifRGBFormat format);
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -25,6 +25,9 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
     if (rgb->isFloat && rgb->depth != 16) {
         return AVIF_FALSE;
     }
+    if (rgb->format == AVIF_RGB_FORMAT_RGB_565 && rgb->depth != 8) {
+        return AVIF_FALSE;
+    }
 
     // These matrix coefficients values are currently unsupported. Revise this list as more support is added.
     //
@@ -105,6 +108,16 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
             state->rgbOffsetBytesG = state->rgbChannelBytes * 2;
             state->rgbOffsetBytesR = state->rgbChannelBytes * 3;
             break;
+        case AVIF_RGB_FORMAT_RGB_565:
+            // Since RGB_565 consists of two bytes per RGB pixel, we simply use
+            // the pointer to the red channel to populate the entire pixel value
+            // as a uint16_t. As a result only rgbOffsetBytesR is used and the
+            // other offsets are unused.
+            state->rgbOffsetBytesR = 0;
+            state->rgbOffsetBytesG = 0;
+            state->rgbOffsetBytesB = 0;
+            state->rgbOffsetBytesA = 0;
+            break;
 
         default:
             return AVIF_FALSE;
@@ -180,7 +193,7 @@ static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
 
 avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
 {
-    if (!rgb->pixels) {
+    if (!rgb->pixels || rgb->format == AVIF_RGB_FORMAT_RGB_565) {
         return AVIF_RESULT_REFORMAT_FAILED;
     }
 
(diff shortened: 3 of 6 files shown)


src/reformat_libyuv.c
```diff
@@ -406,6 +407,25 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
             }
             return AVIF_RESULT_OK;
         }
+    } else if (rgb->format == AVIF_RGB_FORMAT_RGB_565) {
+        // AVIF_RGB_FORMAT_BGR   *ToRGB565Matrix  matrixYUV
+
+        if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) {
```
vigneshvg, line 413: as of now only 420to565matrix is supported in libyuv.

it is trivial to add 422 and probably 444 as well. i will try to add those to libyuv first and then update the code here.

src/reformat.c
```diff
@@ -180,7 +190,7 @@ static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
 
 avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
 {
-    if (!rgb->pixels) {
+    if (!rgb->pixels || rgb->format == AVIF_RGB_FORMAT_RGB_565) {
```
y-guyon, line 193: Should this return `AVIF_RESULT_NOT_IMPLEMENTED`?

src/reformat.c
```diff
@@ -435,6 +445,17 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
     return AVIF_RESULT_OK;
 }
 
+static void avifStoreRGB8Pixel(avifRGBFormat format, uint8_t R, uint8_t G, uint8_t B, uint8_t * ptrR, uint8_t * ptrG, uint8_t * ptrB)
+{
+    if (format == AVIF_RGB_FORMAT_RGB_565) {
+        *(uint16_t *)ptrR = (uint16_t)((B >> 3) | ((G >> 2) << 5) | ((R >> 3) << 11));
```
y-guyon, line 451: No rounding for faster computation?

y-guyon (approved):

src/reformat.c
vigneshvg: it is very unlikely that we will ever support this conversion since there are no image formats where the source is rgb565. so i decided not to return `AVIF_RESULT_NOT_IMPLEMENTED` here.

This is also documented int he AVIF_RGB_FORMAT_RGB_565 enum definition.

src/reformat.c
vigneshvg: Sorry i am not sure what you mean by no rounding here. Can you please explain?

jzern: I think he means dithering here:
https://chromium.googlesource.com/webm/libwebp/+/34b1dc336d04a1b3bb54f38dce2a74e2a586427f/extras/extras.c#50

y-guyon: There is dithering yes, but also sometimes there is what I (mistakenly?) call rounding:

`(value >> (1 << shift >> 1)) >> shift` instead of `value >> shift` (flooring) to have lower loss when scaling back to `shifted_value << shift`.
Example: `value=15`, `shift=3` gives `15>>3=1` (flooring), `(15+(1<<2))>>3=2` (rounding), `1<<3=8` (flooring diff `15-8=7`), `2<<3=16` (rounding diff `15-16=-1`).
But maybe the rounding is done elsewhere (such as during the `<<shift`), or maybe it is faster not to take rounding into account. I was just wondering what was the reason.

vigneshvg: just to be clear, what's happening here is that we are converting R to 5 bits (taking the 5 MSB), G to 6 bits and B to 5 bits. We are then concatenating those bits to be a 16bit word.

i don't think rounding and scaling back to the original value is of much relevance here since RGB565 itself is quite a lossy.

i was trying to match libyuv's C implementation so that we have consistent results between libyuv and libavif.

https://chromium.googlesource.com/libyuv/libyuv/+/9892d70c965678381d2a70a1c9002d1cf136ee78/source/row_common.cc#2362

vigneshvg: > I think he means dithering here: https://chromium.googlesource.com/webm/libwebp/+/34b1dc336d04a1b3bb54f38dce2a74e2a586427f/extras/extras.c#50

Isn't this for going from RGB565 to something else?

We are going the other way here (i.e.) YUV -> RGB565. So i am not sure how dithering can be applied here.

wantehchang: Vignesh is right. (I saw the same dithering code in libyuv's `RGB565ToARGBRow_C` function.)

The rounding that Yannis suggested needs to be done carefully to avoid converting, e.g., a red value of 0xFF to 0x20 (which is a 6-bit value) instead of 0x1F (the maximum 5-bit value).

wantehchang (changes requested):

src/reformat.c
jzern: The example I gave was 565->888, but you can dither (add noise to) u & v before converting to rgb:
https://chromium.googlesource.com/webm/libwebp/+/refs/tags/v1.2.4/src/dec/frame_dec.c#371

include/avif/avif.h
```diff
@@ -551,7 +551,8 @@ typedef enum avifRGBFormat
     AVIF_RGB_FORMAT_ARGB,
     AVIF_RGB_FORMAT_BGR,
     AVIF_RGB_FORMAT_BGRA,
-    AVIF_RGB_FORMAT_ABGR
+    AVIF_RGB_FORMAT_ABGR,
+    AVIF_RGB_FORMAT_RGB_565 // This format is supported only for YUV -> RGB conversion.
```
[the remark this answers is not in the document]
jzern: In webp we describe the byte order like this:
https://chromium.googlesource.com/webm/libwebp/+/refs/tags/v1.2.4/src/webp/decode.h#137

wantehchang:
Vignesh: If the RGB 565 format is only used in some older Android devices, can we just support this format in the Glide library? Do you think this should be supported by the core libavif library?

src/reformat.c
vigneshvg: Yeah, so in order to keep things simple i am just going to match libyuv's C conversion.

Marking this conversation as resolved. Please feel free to re-open if anybody things otherwise. :)

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -127,8 +127,9 @@ FUNC(jboolean, decode, jobject encoded, int length, jobject bitmap) {
         decoder.decoder->image->height);
     return false;
   }
-  // Ensure that the bitmap format is either RGBA_8888 or RGBA_F16.
+  // Ensure that the bitmap format is one of RGB_565, RGBA_8888 or RGBA_F16.
```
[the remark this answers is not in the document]
vigneshvg: Updated the comment.

tests/avifyuv.c
```diff
@@ -39,6 +39,8 @@ static const char * rgbFormatToString(avifRGBFormat format)
             return "BGRA";
         case AVIF_RGB_FORMAT_ABGR:
             return "ABGR";
+        case AVIF_RGB_FORMAT_RGB_565:
+            return "RGB565";
```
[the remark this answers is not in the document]
vigneshvg: Done.

src/reformat.c
```diff
@@ -105,6 +108,13 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
             state->rgbOffsetBytesG = state->rgbChannelBytes * 2;
             state->rgbOffsetBytesR = state->rgbChannelBytes * 3;
             break;
+        case AVIF_RGB_FORMAT_RGB_565:
+            // In this case only rgbOffsetBytesR is used. The rest are unused.
```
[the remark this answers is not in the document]
vigneshvg: Done.

include/avif/avif.h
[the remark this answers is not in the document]
vigneshvg: Done.

src/reformat.c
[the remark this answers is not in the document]
vigneshvg: Done.

vigneshvg:
> Vignesh: If the RGB 565 format is only used in some older Android devices, can we just support this format in the Glide library? Do you think this should be supported by the core libavif library?

Glide library depends on libavif and it's JNI wrappers. It does not have any native code. Also, it is much easier to support this in the core library as there are several cases that need to be covered which is already being done in reformat.c and reformat_libyuv.c.

include/avif/avif.h
```diff
@@ -551,7 +551,17 @@ typedef enum avifRGBFormat
     AVIF_RGB_FORMAT_ARGB,
     AVIF_RGB_FORMAT_BGR,
     AVIF_RGB_FORMAT_BGRA,
-    AVIF_RGB_FORMAT_ABGR
+    AVIF_RGB_FORMAT_ABGR,
+    // RGB_565 format uses five bits for the red and blue components and six
+    // bits for the green component. Each RGB pixel is 16 bits (2 bytes) are
```
wantehchang, line 556: Nit: should "are" be changed to "and"?

Or perhaps this sentence should read:

> Each RGB pixel is 16 bits (2 bytes), which are packed as follows:

include/avif/avif.h
```diff
@@ -551,7 +551,17 @@ typedef enum avifRGBFormat
...
+    // packed as follows:
+    //   Byte 1: [r4 r3 r2 r1 r0 g5 g4 g3]
+    //   Byte 2: [g2 g1 g0 b4 b3 b2 b1 b0]
```
wantehchang, line 559: I suggest we describe this as a uint16_t value in memory instead of Byte 1 and Byte 2:

```
    //   uint16_t: [r4 r3 r2 r1 r0 g5 g4 g3 g2 g1 g0 b4 b3 b2 b1 b0]
```

This matches the code:

```
        *(uint16_t *)ptrR = (uint16_t)((B >> 3) | ((G >> 2) << 5) | ((R >> 3) << 11));
```

The reason "Byte 1" and "Byte 2" are confusing is that they seem to depend on endianness. On little-endian computers (all computers we actually have access to today), Byte 1 and Byte 2 are in the following order in memory:

```
    //   Byte 1 (lower address):  [g2 g1 g0 b4 b3 b2 b1 b0]
    //   Byte 2 (higher address): [r4 r3 r2 r1 r0 g5 g4 g3]
```

By describing RGB_565 pixels as uint16_t values, we avoid the byte order issue. (This is why I said that we could consider RGB_565 as having one channel of bit depth 16.)

include/avif/avif.h
```diff
@@ -551,7 +551,17 @@ typedef enum avifRGBFormat
...
+    //   r4 and r0 are the MSB and LSB of the red component respectively.
+    //   g5 and g0 are the MSB and LSB of the green component respectively.
+    //   b4 and b0 are the MSB and LSB of the blue component respectively.
```
wantehchang, line 562: Please also document that rgb.depth must be set to 8 (by convention).

wantehchang (changes requested):
Vignesh: Thank you for measuring the performance. Your plan sounds good.

I suggest some changes to the comments.

include/avif/avif.h
vigneshvg: Done.

include/avif/avif.h
vigneshvg: Done.

include/avif/avif.h
vigneshvg: Done.

vigneshvg:
> Vignesh: Thank you for measuring the performance. Your plan sounds good.
> 
> I suggest some changes to the comments.

Thanks for the review. I have addressed your comments.

wantehchang (approved):
LGTM.

PR #1010, opened 2022-08-05
Build libavif with ninja instead of make

wantehchang:
Prepare for adding windows-latest to the `os` list, because I am not
sure if make is available on Windows by default.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
       with:
         version: 2.15.05
     - uses: seanmiddleditch/gha-setup-ninja@v3
-      if: steps.cache-ext.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


jzern (approved):

wantehchang:
I edited the commit message to reflect my new understanding of the windows-latest os image. Please review.

I did a lot of experiments with the windows-latest os image in https://github.com/AOMediaCodec/libavif/pull/1011. I found that the windows-latest os image actually has the `make` program, but the current `cmake` command uses the default CMake generator, which is Visual Studio on Windows, so the `make` command still fails on Windows because no (Unix) makefiles are generated.

Passing `-G Ninja` to `cmake` and building with `ninja` is a good solution to this problem.

y-guyon (approved):

PR #1017, opened 2022-08-08
Use the latest versions of GitHub Actions

wantehchang:
Use the latest versions of actions/checkout, actions/setup-python, and
actions/cache.

Do not check for cache hit in the seanmiddleditch/gha-setup-ninja@v3
and "pip install meson" steps, because their results are not in the
"ext" directory. Only the "ext" directory is cached.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,15 +9,17 @@ jobs:
         os: [ubuntu-latest, macos-latest]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set GCC & G++ 10 compiler (on Linux)
       if: runner.os == 'Linux'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       with:
         version: 2.15.05
     - uses: seanmiddleditch/gha-setup-ninja@v3
-      #if: steps.cache-ext.outputs.cache-hit != 'true'
+      if: steps.cache-ext.outputs.cache-hit != 'true'
```
wantehchang, line 34: @EwoutH Caching does not seem to work for the seanmiddleditch/gha-setup-ninja@v3 step. Do you know what is wrong?

wantehchang: I think this is because the results of the seanmiddleditch/gha-setup-ninja@v3 are not in the `ext` directory and therefore are not cached.

jzern (approved):

PR #1020, opened 2022-08-08
Write cmake command as YAML Folded Block Scalar

wantehchang:
Write the multi-line cmake command as a YAML Folded Block Scalar. See
https://yaml-multiline.info/.

This change is needed because the default shell on Windows (PowerShell)
does not support line continuation with a backslash (\) character.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,18 +65,19 @@ jobs:
       run: bash googletest.cmd
 
     - name: Prepare libavif (cmake)
-      run: |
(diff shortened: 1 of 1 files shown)


jzern (approved):

kmilos:
> This change is needed because the default shell on Windows (PowerShell)
> does not support line continuation with a backslash () character.

Just for future reference: on PS you can use the backtick (\`) for line continuation.

y-guyon (approved):

PR #1021, opened 2022-08-09
avifPNGWrite: Output RGB if input has no alpha

wantehchang:
avifPNGWrite() should output RGB instead of RGBA if the input AVIF image
has no alpha plane.

Remove an unnecessary memset() call, because avifRGBImageSetDefaults()
assumes its avifRGBImage input is uninitialized.

Fix https://github.com/AOMediaCodec/libavif/issues/787.


wantehchang:
An alternative is to change `avifRGBImageSetDefaults()` to set `rgb->format` to a default value that depends on whether the `avifImage` input has an alpha plane or not. I didn't do that, because it may break code that depends on the current default value of `rgb->format`

wantehchang:
There are some bugs with the pull request. I will request a review after I sort them out.

wantehchang:
The pull request is ready for review now.

jzern (approved):

joedrago (approved):

FrankGalligan:
Thanks LGTM

PR #1022, opened 2022-08-09
Update .gitignore

vigneshvg:
Update .gitignore with cmake related files, binaries and
intermediate libraries.

This makes the git workspace saner and easier to track after
building the library locally.

Also sort the file alphabetically.


.gitignore
```diff
@@ -1,15 +1,27 @@
+avifdec
+avifenc
+avif_example_decode_file
+avif_example_decode_memory
+avif_example_decode_streaming
+avif_example_encode
 /build*
-/obj*
+.clangd/
```
wantehchang, line 8: The current .gitignore file assumes that you run cmake in the libavif/build directory (see line 7) or a directory outside the source tree. Can you use that setup?

vigneshvg: ah i see. sorry for the noise, will drop this one.

PR #1023, opened 2022-08-09
Guard I400 libyuv conversions with correct version

vigneshvg:
I400To* functions were only available on libyuv since version 1756.
The minimum version of libyuv required by libavif is 1755. So we
need to guard these function calls with the correct version.


wantehchang (approved):
LGTM. Thanks. I verified that I400ToARGBMatrix() was added in version 1756:
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2209718

PR #1024, opened 2022-08-10
Use libyuv for default PNG RGB-to-YUV in avifenc

y-guyon:
Speed up at the expense of a small conversion loss.

According to avifrgbtoyuvtest, for 8-bit RGBA to 8-bit full range BT.601:
```
C float conversion: avg 0.0728848, abs avg 2.75841, max 12, PSNR 37.1393dB
libyuv conversion:  avg 0.0442412, abs avg 2.81094, max 13, PSNR 37.0215dB
```
According to a profiling of a local encoding of 30 ~2k images with the following flags:
```
--speed 9 --jobs 1 -y 420 --min 0 --max 63 --tilecolslog2 2 --tilerowslog2 2
-a end-usage=q -a cq-level=28 -a tune=ssim -a deltaq-mode=3 -a sharpness=3 --range full
```
```
avifenc 7.74s (including 1.46s of avifImageRGBToYUV() )
avifenc 6.47s (including 0.22s of avifImageRGBToYUV() )
```
representing an 85% speedup of avifImageRGBToYUV() and a 16% speedup of avifenc with these settings.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -16,7 +16,7 @@ extern "C" {
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
 // Returns the equivalent of ceil(a/b) but for positive integers.
-#define AVIF_CEIL(a, b) (((a) + (b)-1) / (b))
+#define AVIF_DIV_CEIL(a, b) (((a) + (b)-1) / (b))
 
 // Used by stream related things.
 #define CHECK(A)               \
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -15,7 +15,7 @@ extern "C" {
 #define AVIF_MIN(a, b) (((a) < (b)) ? (a) : (b))
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
-// Returns the equivalent of ceil(a/b) but for positive integers.
+// Returns the equivalent of ceil(a/b) for positive integers a and b.
 #define AVIF_DIV_CEIL(a, b) (((a) + (b)-1) / (b))
 
 // Used by stream related things.
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -15,9 +15,6 @@ extern "C" {
 #define AVIF_MIN(a, b) (((a) < (b)) ? (a) : (b))
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
-// Returns the equivalent of ceil(a/b) for positive integers a and b.
-#define AVIF_DIV_CEIL(a, b) (((a) + (b)-1) / (b))
-
 // Used by stream related things.
 #define CHECK(A)               \
     do {                       \
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -15,6 +15,9 @@ extern "C" {
 #define AVIF_MIN(a, b) (((a) < (b)) ? (a) : (b))
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
+// Returns the equivalent of ceil(a/b) but for positive integers.
+#define AVIF_CEIL(a, b) (((a) + (b)-1) / (b))
+
 // Used by stream related things.
 #define CHECK(A)               \
     do {                       \
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -111,15 +111,15 @@ static int avifABGRToJ420(const uint8_t * src_abgr,
 
     for (int y = 0; y < height; y += num_allocated_rows) {
         const int num_rows = AVIF_MIN(num_allocated_rows, height - y);
-        if (ARGBToABGR(src_abgr, src_stride_abgr, src_argb, src_stride_argb, width, num_rows) ||
+        if (ABGRToARGB(src_abgr, src_stride_abgr, src_argb, src_stride_argb, width, num_rows) ||
             ARGBToJ420(src_argb, src_stride_argb, dst_y, dst_stride_y, dst_u, dst_stride_u, dst_v, dst_stride_v, width, num_rows)) {
             avifFree(src_argb);
             return -1;
         }
         src_abgr += (size_t)num_rows * src_stride_abgr;
         dst_y += (size_t)num_rows * dst_stride_y;
-        dst_u += (size_t)num_rows * dst_stride_u / 2; // 4:2:0
-        dst_v += (size_t)num_rows * dst_stride_v / 2; // (either num_rows is even or this is the last loop turn)
+        dst_u += (size_t)num_rows / 2 * dst_stride_u; // 4:2:0
+        dst_v += (size_t)num_rows / 2 * dst_stride_v; // (either num_rows is even or this is the last iteration)
     }
     avifFree(src_argb);
     return 0;
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -41,6 +41,7 @@ unsigned int avifLibYUVVersion(void)
 #else
 
 #include <assert.h>
+#include <limits.h>
 
 #if defined(__clang__)
 #pragma clang diagnostic push
@@ -98,13 +99,16 @@ static int avifABGRToJ420(const uint8_t * src_abgr,
         // Process the whole buffer in one go.
         num_allocated_rows = height;
     } else {
+        if ((uint64_t)src_stride_argb * 2 > INT_MAX) {
+            return -1;
+        }
         // The last row of an odd number of RGB rows to be converted to subsampled YUV is treated differently
         // by libyuv, so make sure all steps but the last one process an even number of rows.
         // Try to process as many row pairs as possible in a single step without allocating more than
         // soft_allocation_limit, unless two rows need more than that.
-        num_allocated_rows = AVIF_MAX(1, soft_allocation_limit / ((uint64_t)src_stride_argb * 2)) * 2;
+        num_allocated_rows = AVIF_MAX(1, soft_allocation_limit / (src_stride_argb * 2)) * 2;
     }
-    src_argb = avifAlloc((uint64_t)num_allocated_rows * src_stride_argb);
+    src_argb = avifAlloc(num_allocated_rows * src_stride_argb);
     if (!src_argb) {
         return -1;
     }
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -94,7 +94,7 @@ static int avifABGRToJ420(const uint8_t * src_abgr,
     const int src_stride_argb = width * 4;
     const uint64_t soft_allocation_limit = 16384; // Arbitrarily chosen trade-off between CPU and memory footprints.
     int num_allocated_rows;
-    if ((uint64_t)src_stride_argb * height <= soft_allocation_limit) {
+    if (height == 1 || (uint64_t)src_stride_argb * height <= soft_allocation_limit) {
         // Process the whole buffer in one go.
         num_allocated_rows = height;
     } else {
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -102,9 +102,9 @@ static int avifABGRToJ420(const uint8_t * src_abgr,
         // by libyuv, so make sure all steps but the last one process an even number of rows.
         // Try to process as many row pairs as possible in a single step without allocating more than
         // soft_allocation_limit, unless two rows need more than that.
-        num_allocated_rows = AVIF_MAX(1, (int)(soft_allocation_limit / (uint64_t)src_stride_argb * 2)) * 2;
+        num_allocated_rows = AVIF_MAX(1, soft_allocation_limit / ((uint64_t)src_stride_argb * 2)) * 2;
     }
-    src_argb = avifAlloc(num_allocated_rows * src_stride_argb);
+    src_argb = avifAlloc((uint64_t)num_allocated_rows * src_stride_argb);
     if (!src_argb) {
         return -1;
     }
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -102,11 +102,11 @@ static int avifABGRToJ420(const uint8_t * src_abgr,
         // by libyuv, so make sure all steps but the last one process an even number of rows.
         // Try to process as many row pairs as possible in a single step without allocating more than
         // soft_allocation_limit, unless two rows need more than that.
-        num_allocated_rows = (int)AVIF_CEIL(soft_allocation_limit, (uint64_t)src_stride_argb * 2) * 2;
+        num_allocated_rows = (int)AVIF_DIV_CEIL(soft_allocation_limit, (uint64_t)src_stride_argb * 2) * 2;
     }
     src_argb = avifAlloc(num_allocated_rows * src_stride_argb);
     if (!src_argb) {
-        return 1;
+        return -1;
     }
 
     for (int y = 0; y < height; y += num_allocated_rows) {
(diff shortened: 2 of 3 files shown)


src/reformat_libyuv.c
```diff
@@ -77,6 +77,54 @@ avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb)
...
+                          uint8_t * dst_y,
+                          int dst_stride_y,
+                          uint8_t * dst_u,
+                          int dst_stride_u,
+                          uint8_t * dst_v,
+                          int dst_stride_v,
+                          int width,
+                          int height)
+{
+    // A temporary buffer is needed to swap the R and B channels before calling ARGBToJ420().
```
jzern, line 92: The API docs for libyuv are a little sparse, but I'd expect this conversion to work in place for most optimizations. The C code seems to handle that case. That might be something to clarify with @fbarchard.

include/avif/internal.h
```diff
@@ -15,6 +15,9 @@ extern "C" {
 #define AVIF_MIN(a, b) (((a) < (b)) ? (a) : (b))
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
+// Returns the equivalent of ceil(a/b) but for positive integers.
+#define AVIF_CEIL(a, b) (((a) + (b)-1) / (b))
```
wantehchang, line 19: Name this macro `AVIF_CEIL_DIV` or `AVIF_DIV_CEIL` because it also does a division.

Add spaces around the `-` operator: `(((a) + (b) - 1) / (b))`

src/reformat_libyuv.c
wantehchang: James: The `src_abgr` parameter points to a const buffer, so we can't modify the `src_abgr` buffer.

Yannis: I believe we just need to call `ARGBToJ420()` with the `dst_u` and `dst_v` buffers swapped. We use this trick elsewhere, as described by the following comment in rc/reformat_libyuv.c:

```
    // AVIF_RGB_FORMAT_RGB       *ToRGB24Matrix   matrixYVU
    // AVIF_RGB_FORMAT_BGR       *ToRGB24Matrix   matrixYUV
    // AVIF_RGB_FORMAT_RGB_565   *ToRGB565Matrix  matrixYUV
    // AVIF_RGB_FORMAT_BGRA      *ToARGBMatrix    matrixYUV
    // AVIF_RGB_FORMAT_RGBA      *ToARGBMatrix    matrixYVU
    // AVIF_RGB_FORMAT_ABGR      *ToRGBAMatrix    matrixYUV
    // AVIF_RGB_FORMAT_ARGB      *ToRGBAMatrix    matrixYVU
```

src/reformat_libyuv.c
```diff
@@ -77,6 +77,54 @@ avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb)
...
+        const int num_rows = AVIF_MIN(num_allocated_rows, height - y);
+        if (ARGBToABGR(src_abgr, src_stride_abgr, src_argb, src_stride_argb, width, num_rows) ||
+            ARGBToJ420(src_argb, src_stride_argb, dst_y, dst_stride_y, dst_u, dst_stride_u, dst_v, dst_stride_v, width, num_rows)) {
+            avifFree(src_argb);
+            return 1;
+        }
+        src_abgr += (uint64_t)num_rows * src_stride_abgr;
+        dst_y += (uint64_t)num_rows * dst_stride_y;
+        dst_u += (uint64_t)num_rows * dst_stride_u / 2; // 4:2:0
+        dst_v += (uint64_t)num_rows * dst_stride_v / 2; // (either num_rows is even or this is the last loop turn)
```
wantehchang, line 122: In these four lines, cast to `size_t` or `ptrdiff_t` instead of `uint64_t`. The `uint64_t` type is too big for memory buffer offsets in 32-bit builds.

src/reformat_libyuv.c
```diff
@@ -77,6 +77,54 @@ avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb)
...
+    } else {
+        // The last row of an odd number of RGB rows to be converted to subsampled YUV is treated differently
+        // by libyuv, so make sure all steps but the last one process an even number of rows.
+        // Try to process as many row pairs as possible in a single step without allocating more than
+        // soft_allocation_limit, unless two rows need more than that.
+        num_allocated_rows = (int)AVIF_CEIL(soft_allocation_limit, (uint64_t)src_stride_argb * 2) * 2;
+    }
+    src_argb = avifAlloc(num_allocated_rows * src_stride_argb);
+    if (!src_argb) {
+        return 1;
```
wantehchang, line 109: libyuv functions return -1 on failure. Also at line 117.

wantehchang (changes requested):

src/reformat_libyuv.c
jzern: > James: The `src_abgr` parameter points to a const buffer, so we can't modify the `src_abgr` buffer.

Sorry about that, I didn't look at the signature closely enough, I was just being hopeful I guess.

jzern: Though the question still sta

> > James: The `src_abgr` parameter points to a const buffer, so we can't modify the `src_abgr` buffer.
> 
> Sorry about that, I didn't look at the signature closely enough, I was just being hopeful I guess.

Though the question still would be if it was ok to have the src and dst pointers be the same in this function and the calls to libyuv.

y-guyon: > Yannis: I believe we just need to call `ARGBToJ420()` with the `dst_u` and `dst_v` buffers swapped. We use this trick elsewhere, as described by the following comment in rc/reformat_libyuv.c:

Yes, but the comment also contains:

```
    // In addition, swapping U and V in any of these calls, along with using the Yvu matrix instead of Yuv matrix,
    // swaps B and R in these orderings as well.
```

It is possible with `I420ToARGBMatrixFilter()` because it has a `const struct YuvConstants* yuvconstants` argument. However the YUV matrix coefficients are hard-coded in `ARGBToJ420()` and swapping `dst_u` and `dst_v` is not enough.
See Frank's comment: https://bugs.chromium.org/p/libyuv/issues/detail?id=936#c5

- - -

> Though the question still would be if it was ok to have the src and dst pointers be the same in this function and the calls to libyuv.

I do not follow.
Are you suggesting to store the temporary swapped RGB samples into `dst_y, dst_u, dst_v` to avoid an allocation, and then to call `ARGBToJ420()` in-place?
Or are you suggesting to modify the `src_abgr` samples in-place (they do not have to be const in the use case where `avifenc` reads a PNG and owns the samples)?

Anyway I hope that `ABGRToJ420()` will be available in libyuv at some point, and we will only need `avifABGRToJ420()` for earlier libyuv versions.

include/avif/internal.h
y-guyon: > Name this macro `AVIF_CEIL_DIV` or `AVIF_DIV_CEIL` because it also does a division.

Done.

> Add spaces around the `-` operator: `(((a) + (b) - 1) / (b))`

This is a bug of clang-format. See https://github.com/llvm/llvm-project/issues/41502.

src/reformat_libyuv.c
y-guyon: `ARGBToABGR()` calls `ARGBShuffle()` which may call `ARGBShuffleRow_C()` which has the comment [`// To support in-place conversion.`](https://chromium.googlesource.com/libyuv/libyuv/+/9b17af9bef41aec80373a7d4689bb77bac33eab5/source/row_common.cc#3525). I did not check the intrinsic implementations but I would guess the whole function is in-place.

wantehchang (changes requested):
Yannis: Nice speedup! Overall the pull request looks good. I suggest some simple changes. I leave the changes marked as "optional" to your discretion.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -237,6 +246,69 @@ TEST_P(RGBToYUVTest, Convert) {
...
+      dst_rgb.chromaUpsampling = AVIF_CHROMA_UPSAMPLING_AUTOMATIC;
+      const testutil::RgbChannelOffsets offsets =
+          testutil::GetRgbChannelOffsets(rgb_format);
+
+      // Fill the input buffer with whatever content.
+      SetImageChannel(&src_rgb, offsets.r, /*value=*/0,
+                      add_noise ? kRedNoise : kPlainColor);
+      SetImageChannel(
+          &src_rgb, offsets.g, /*value=*/0,
+          add_noise ? is_monochrome ? kRedNoise : kGreenNoise : kPlainColor);
```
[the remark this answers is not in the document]
y-guyon: It cannot mean anything else, otherwise the syntax would be invalid.

src/reformat_libyuv.c
```diff
@@ -77,6 +77,54 @@ avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb)
...
+                          int dst_stride_u,
+                          uint8_t * dst_v,
+                          int dst_stride_v,
+                          int width,
+                          int height)
+{
+    // A temporary buffer is needed to swap the R and B channels before calling ARGBToJ420().
+    uint8_t * src_argb;
+    const int src_stride_argb = width * 4;
+    const uint64_t soft_allocation_limit = 16384; // Arbitrarily chosen trade-off between CPU and memory footprints.
```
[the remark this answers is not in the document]
y-guyon: It is a soft limit because we may allocate more than `soft_allocation_limit` if `src_stride_argb` is bigger, as `num_allocated_rows` is at least 1.

src/reformat_libyuv.c
```diff
@@ -77,6 +77,54 @@ avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb)
...
+    int num_allocated_rows;
+    if ((uint64_t)src_stride_argb * height <= soft_allocation_limit) {
+        // Process the whole buffer in one go.
+        num_allocated_rows = height;
+    } else {
+        // The last row of an odd number of RGB rows to be converted to subsampled YUV is treated differently
+        // by libyuv, so make sure all steps but the last one process an even number of rows.
+        // Try to process as many row pairs as possible in a single step without allocating more than
+        // soft_allocation_limit, unless two rows need more than that.
```
wantehchang, line 104: Nit: "two rows need more than that", i.e., `(uint64_t)src_stride_argb * 2 > soft_allocation_limit`, is not the only reason we allocate more than `soft_allocation_limit`. We allocate more than `soft_allocation_limit` when the division has a remainder, i.e., when `soft_allocation_limit` is not a multiple of `(uint64_t)src_stride_argb * 2`.

tests/gtest/avifrgbtoyuvtest.cc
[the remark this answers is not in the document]
wantehchang: You are right, but this expression is not easy to understand. Parenthesizing the inner `? :` expression would make the nesting structure clearer.

src/reformat_libyuv.c
wantehchang: Frank Barchard confirmed that `ARGBToABGR()` supports in-place conversion. In response to our question, he added a unit test for libyuv functions that should support in-place conversion:
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/3828485

wantehchang (approved):
LGTM. I suggest a small change to a comment. You can do that in a follow-up pull request.

src/reformat_libyuv.c
y-guyon: Ah yes, my bad. It was not my intent so I changed the formula and removed AVIF_DIV_CEIL().

fbarchard:
The function ABGRToJ420 does not exist in libyuv?
But ABGRToI420 does, and ARGBToJ420 does, so I suggest adding the ABGRToJ420.
Internally the RGB to YUV is starting to take matrix values, so the low levels share code with a different parameter.

src/reformat_libyuv.c
```diff
@@ -77,6 +77,54 @@ avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb)
...
+    int num_allocated_rows;
+    if ((height == 1) || ((uint64_t)src_stride_argb * height <= soft_allocation_limit)) {
+        // Process the whole buffer in one go.
+        num_allocated_rows = height;
+    } else {
+        // The last row of an odd number of RGB rows to be converted to subsampled YUV is treated differently
+        // by libyuv, so make sure all steps but the last one process an even number of rows.
+        // Try to process as many row pairs as possible in a single step without allocating more than
+        // soft_allocation_limit, unless two rows need more than that.
+        num_allocated_rows = AVIF_MAX(1, (int)(soft_allocation_limit / (uint64_t)src_stride_argb * 2)) * 2;
```
wantehchang, line 105: IMPORTANT: We need to divide `soft_allocation_limit` by `src_stride_argb * 2`. This bug was introduced when `AVIF_DIV_CEIL` was removed: https://github.com/AOMediaCodec/libavif/commit/4911be8ae90deba3334b14ffcf9b1a965b4afe98

Optional: We may be able to omit the `uint64_t` cast on `src_stride_argb`.

wantehchang (changes requested):

src/reformat_libyuv.c
```diff
@@ -77,6 +77,54 @@ avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb)
...
+        // Process the whole buffer in one go.
+        num_allocated_rows = height;
+    } else {
+        // The last row of an odd number of RGB rows to be converted to subsampled YUV is treated differently
+        // by libyuv, so make sure all steps but the last one process an even number of rows.
+        // Try to process as many row pairs as possible in a single step without allocating more than
+        // soft_allocation_limit, unless two rows need more than that.
+        num_allocated_rows = AVIF_MAX(1, soft_allocation_limit / ((uint64_t)src_stride_argb * 2)) * 2;
+    }
+    src_argb = avifAlloc((uint64_t)num_allocated_rows * src_stride_argb);
```
wantehchang, line 107: Cast to `size_t` instead (see lines 119-122), because `uint64_t` is too big for `avifAlloc()` in 32-bit builds.

Note: This cast seems unnecessary, because this product is at most `soft_allocation_limit` or `2 * src_stride_argb`. We already know `src_stride_argb <= src_stride_abgr <= INT_MAX`. We just need to make sure `2 * src_stride_argb <= INT_MAX`.

y-guyon:
> The function ABGRToJ420 does not exist in libyuv?

Not at the time of https://bugs.chromium.org/p/libyuv/issues/detail?id=936#c5.

> But ABGRToI420 does, and ARGBToJ420 does, so I suggest adding the ABGRToJ420. Internally the RGB to YUV is starting to take matrix values, so the low levels share code with a different parameter.

We can do that, but we would like to speed avifenc up even for older versions of libyuv, so I started with this PR. Thanks for the implementation hints.

wantehchang (changes requested):

wantehchang (approved):
LGTM!

src/reformat_libyuv.c
```diff
@@ -77,6 +78,57 @@ avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb)
...
+        num_allocated_rows = height;
+    } else {
+        if ((uint64_t)src_stride_argb * 2 > INT_MAX) {
+            return -1;
+        }
+        // The last row of an odd number of RGB rows to be converted to subsampled YUV is treated differently
+        // by libyuv, so make sure all steps but the last one process an even number of rows.
+        // Try to process as many row pairs as possible in a single step without allocating more than
+        // soft_allocation_limit, unless two rows need more than that.
+        num_allocated_rows = AVIF_MAX(1, soft_allocation_limit / (src_stride_argb * 2)) * 2;
```
wantehchang, line 109: Yannis: MSVC warns about this line:

```
src\reformat_libyuv.c(109,89): warning C4242: '=': conversion from 'const uint64_t' to 'int', possible loss of data
```

Please add back the `(int)` cast.

I will hurry up my work on adding a GitHub CI build for Windows. Also, we can look into if there is a suitable GCC or Clang warning for this.

y-guyon: Fixed in https://github.com/AOMediaCodec/libavif/pull/1036

PR #1025, opened 2022-08-10
Run ext/libjpeg.cmd in GitHub CI

y-guyon:
find_package(JPEG REQUIRED) does not seem to work anymore on Mac.

`# if: steps.cache-ext.outputs.cache-hit != 'true'` to be uncommented in a following change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,10 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: bash libyuv.cmd
+    - name: Build libjpeg
+      # if: steps.cache-ext.outputs.cache-hit != 'true'
+      working-directory: ./ext
+      run: bash libjpeg.cmd
     - name: Build GoogleTest
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -58,6 +58,10 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: bash libyuv.cmd
+    - name: Build libjpeg
+      # if: steps.cache-ext.outputs.cache-hit != 'true'
+      working-directory: ./ext
+      run: bash libjpeg.cmd
```
wantehchang, line 64: This is a good change. I may have suggested this when you modified avifjpeg.c and avifpng.c to work with very old versions of libjpeg and libpng, because (if I remember correctly) we were using libjpeg (or libpng?) in the Mono framework (an open-source implementation of .NET).

One thing we can consider is to do this on non-Linux platforms, if we can assume the ubuntu-latest os image has the libjpeg-dev package installed. This will speed up the ubuntu-latest builds.

y-guyon: If cached, I would imagine it is as fast.

wantehchang: Also, I just realized that making this non-Linux only will complicate the addition of `-DAVIF_LOCAL_JPEG=ON` to the cmake command line. So please disregard this suggestion.

wantehchang (approved):
LGTM.

.github/workflows/ci.yml
jzern: Is this the cause of the breakage in the mac CI builds?
https://github.com/AOMediaCodec/libavif/runs/7776218330?check_suite_focus=true

wantehchang: Yes, it is. Note that we find zlib and libpng in the macos-latest os image at these locations:

```
-- Found ZLIB: /Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/lib/libz.tbd (found version "1.2.11") 
-- Found PNG: /usr/local/lib/libpng.dylib (found version "1.4.12") 
```

It is not clear whether these were installed in the macos-latest os image deliberately or as a byproduct of installing something else. Previously we found libjpeg in the Mono framework in a similar manner. That libjpeg installation is gone now.

wantehchang:
I believe I figured out the underlying problem while trying to run tests/test_cmd.sh in GitHub CI's macos-latest os image. Please see https://github.com/AOMediaCodec/libavif/pull/1072.

PR #1026, opened 2022-08-10
Refactor AVIF_CHROMA_ into avifConversionFlags

y-guyon:


y-guyon:
Unrelated note: amending the commit message with [`skip-checks:true`](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs) does not seem to skip the GitHub CI.

include/avif/avif.h
```diff
@@ -602,9 +578,26 @@ AVIF_API uint32_t avifRGBImagePixelSize(const avifRGBImage * rgb);
...
+typedef enum avifConversionFlag
+{
+    // Conversion between RGB and YUV samples:
+    AVIF_CONVERSION_AUTO = 0,                // prefers libyuv if available (default)
+    AVIF_CONVERSION_AVOID_LIBYUV = (1 << 0), // only picks built-in paths
+
+    // Conversion from RGB to YUV (ignored when converting from RGB to YUV), set at most one:
+    AVIF_CHROMA_DOWNSAMPLING_AVERAGE = (1 << 10), // only use the averaging filter (libyuv or built-in) (default)
+
+    // Conversion from YUV to RGB (ignored when converting from YUV to RGB), set at most one:
```
wantehchang, line 591: In the parentheses, change "from YUV to RGB" to "from RGB to YUV".

These three flags specify the chroma upsampling filter, so the comment should start a new subsection for them, e.g.,

```
    // Conversion from YUV to RGB (ignored when converting from RGB to YUV)

    // Chroma upsampling filter. Set at most one:
```

We should say that they are ignored if the conversion does not need to upsample chroma (when input is YUV 4:4:4 or YUV 4:0:0). We shoudl also say what is the default chroma upsampling filter when none is specified.

include/avif/avif.h
```diff
@@ -602,9 +578,26 @@ AVIF_API uint32_t avifRGBImagePixelSize(const avifRGBImage * rgb);
...
-AVIF_API avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb);
-AVIF_API avifResult avifImageYUVToRGB(const avifImage * image, avifRGBImage * rgb);
+// The main conversion options.
+typedef enum avifConversionFlag
+{
+    // Conversion between RGB and YUV samples:
+    AVIF_CONVERSION_AUTO = 0,                // prefers libyuv if available (default)
+    AVIF_CONVERSION_AVOID_LIBYUV = (1 << 0), // only picks built-in paths
+
+    // Conversion from RGB to YUV (ignored when converting from RGB to YUV), set at most one:
```
wantehchang, line 588: In the parentheses, change "from RGB to YUV" to "from YUV to RGB".

These flags (currently only one) specify the chroma downsampling filter, so the comment should start a new subsection for them, e.g.,

```
    // Conversion from RGB to YUV (ignored when converting from YUV to RGB)

    // Chroma downsampling filter. Set at most one:
```

We should say that they are ignored if the conversion does not need to downsample chroma (when output is YUV 4:4:4 or YUV 4:0:0). We shoudl also say what is the default chroma downsampling filter when none is specified.

include/avif/avif.h
```diff
@@ -602,9 +578,26 @@ AVIF_API uint32_t avifRGBImagePixelSize(const avifRGBImage * rgb);
...
 AVIF_API void avifRGBImageFreePixels(avifRGBImage * rgb);
 
-// The main conversion functions
-AVIF_API avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb);
-AVIF_API avifResult avifImageYUVToRGB(const avifImage * image, avifRGBImage * rgb);
+// The main conversion options.
+typedef enum avifConversionFlag
+{
+    // Conversion between RGB and YUV samples:
+    AVIF_CONVERSION_AUTO = 0,                // prefers libyuv if available (default)
```
wantehchang, line 585: `AVIF_CONVERSION_AUTO` should be separated from the next line. It should form its own section, separate from the following three sections (libyuv preference, RGB-to-YUV conversion, YUV-to-RGB conversion).

`AVIF_CONVERSION_AUTO` means more than "prefers libyuv if available". It means everything (libyuv preference, chroma downsampling or upsampling filter)  is the default.

include/avif/avif.h
```diff
@@ -602,9 +578,26 @@ AVIF_API uint32_t avifRGBImagePixelSize(const avifRGBImage * rgb);
...
+    AVIF_CONVERSION_AVOID_LIBYUV = (1 << 0), // only picks built-in paths
```
wantehchang, line 586: Change the comment to "Only picks built-in paths. The default is to prefer libyuv if available."

wantehchang (commented):
Yannis: Overall this looks good. All of my changes are for the comments.

@joedrago Could you take a look at the proposed API changes to the `avifRGBImage` struct and the two conversion functions?

include/avif/avif.h
y-guyon: > In the parentheses, change "from YUV to RGB" to "from RGB to YUV".

Whoops.

> These three flags specify the chroma upsampling filter, so the comment should start a new subsection for them

Will there ever be another subsection? I find the `avifConversionFlag` enum difficult to read now.

> We should say that they are ignored if the conversion does not need to upsample chroma (when input is YUV 4:4:4 or YUV 4:0:0).

Done.

> We shoudl also say what is the default chroma upsampling filter when none is specified.

I did that through the `(default)` comments but I guess it was not visible or clear enough.

y-guyon:
> @joedrago Could you take a look at the proposed API changes to the `avifRGBImage` struct and the two conversion functions?

I will wait a bit.

joedrago:
Sorry for the delay. I've read the API/header changes and in general, I think this is more honest about how the YUV conversion thinks, which is good. 

However, the original design's intent around `avifRGBImage` was to effectively be an ephemeral struct representing the arguments to the YUV conversion functions, where `avifRGBImageSetDefaults()` would buy us backwards compatibility (especially for those not interested in tuning sampling). By changing the function signature and moving settings that could exist in this struct into the argument list, it breaks that notion and sets a precedent for adding future additional arguments (and more breaking of compat), which I don't like. This takes the notion of `avifRGBImage` being an object with state more seriously though, so I think I'm alright with that.

It is also piling quite a bit of settings into that flags member, but seeing as these functions have mostly stabilized in the actual functionality they offer, that might not be too big of a deal. I'm a big fan of `AUTO` being `0` (no flags), though. 

I think I'm on board with the overall API changes. I prefer the wording of these flags over the original enums that existed.

wantehchang (changes requested):

include/avif/avif.h
```diff
@@ -540,12 +540,9 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
 //   * x/x/[1|2|5|6|9]/Limited
 //   * [1|2|5|6|9]/x/12/Limited
 
-// If libavif is built with libyuv fast paths enabled, libavif will use libyuv for conversion from
-// RGB to YUV if the following requirements are met:
-//
+// Conversion from RGB to YUV:
 // * YUV depth: 8
 // * RGB depth: 8
-// * rgb.chromaDownsampling: AVIF_CHROMA_UPSAMPLING_AUTOMATIC, AVIF_CHROMA_UPSAMPLING_FASTEST
```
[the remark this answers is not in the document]
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1060

wantehchang (approved):
LGTM with suggested changes.

wantehchang:
Joe Drago wrote:
> ..., the original design's intent around `avifRGBImage` was to effectively be an ephemeral struct representing the arguments to the YUV conversion functions, where `avifRGBImageSetDefaults()` would buy us backwards compatibility (especially for those not interested in tuning sampling).

Thank you for the explanation. Now I see `avifRGBImage` can indeed to viewed as a parameter struct for the YUV conversion functions. We can preserve the original design, and improve it by adding an `avifBool avoidLibYUV` field to `avifRGBImage` and reworking the `chromaUpsampling` and `chromaDownsampling` fields.

But if we do the new design right, the new `flags` parameter will essentially be the bitwise-OR of either the `chromaUpsampling` enum and the `avoidLibYUV` boolean, or the `chromaDownsampling` enum and the `avoidLibYUV` boolean. I.e., we are piling only two settings into the `flags` parameter. If we ever need to pile another setting into the `flags` parameter, we will be able to do that without changing the prototypes of the YUV conversion functions.

include/avif/avif.h
```diff
@@ -628,9 +604,35 @@ AVIF_API uint32_t avifRGBImagePixelSize(const avifRGBImage * rgb);
...
+    // Conversion from RGB to YUV 4:2:2 or YUV 4:2:0 (ignored in all other cases)
+
+    // Chroma downsampling filter. Set at most one:
+    AVIF_CHROMA_DOWNSAMPLING_AVERAGE = (1 << 10), // only use the averaging filter (libyuv or built-in)
+
+    // Conversion from YUV 4:2:2 or YUV 4:2:0 to RGB (ignored in all other cases)
+
+    // Chroma upsampling filter. Set at most one:
+    AVIF_CHROMA_UPSAMPLING_NEAREST = (1 << 20), // only use the fast nearest-neighbor filter (libyuv or built-in)
+    AVIF_CHROMA_UPSAMPLING_BILINEAR = (1 << 21) // only use the good-quality bilinear filter (libyuv or built-in)
```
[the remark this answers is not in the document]
y-guyon: > 1. Since chroma downsampling and chroma upsampling flags are never used at the same time, they do not need to be assigned to different bit positions.

Based on this and the [other](https://github.com/AOMediaCodec/libavif/pull/1026#discussion_r954314330) comments, I decided to split `avifConversionFlags` into `avifRGBToYUVFlags` and `avifYUVToRGBFlags`.

> 2. One way to ensure at most one chroma upsampling filter is set is to number them `1 << 20`, `2 << 20`, `3 << 20`, etc., instead of `1 << 20`, `1 << 21`, `1 << 22`.

I would rather not mix bit flags and enum constants in the same enum. `(2 << 20) == (1 << 21)` so the values are the same for two possibilities anyway.

wantehchang (approved):
LGTM.

PR #1028, opened 2022-08-10
Support YUV422->RGB565 conversion with libyuv

vigneshvg:
This was added in libyuv r1838 (commit 9b17af9b).

Also update ext/libyuv.cmd to that version.

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -416,7 +416,24 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
     } else if (rgb->format == AVIF_RGB_FORMAT_RGB_565) {
         // AVIF_RGB_FORMAT_BGR   *ToRGB565Matrix  matrixYUV
 
-        if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) {
+        if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV422) {
+#if LIBYUV_VERSION >= 1838
+            if (I422ToRGB565Matrix(image->yuvPlanes[AVIF_CHAN_Y],
+                                   image->yuvRowBytes[AVIF_CHAN_Y],
+                                   image->yuvPlanes[AVIF_CHAN_U],
+                                   image->yuvRowBytes[AVIF_CHAN_U],
+                                   image->yuvPlanes[AVIF_CHAN_V],
+                                   image->yuvRowBytes[AVIF_CHAN_V],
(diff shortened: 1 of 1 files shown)


ext/libyuv.cmd
```diff
@@ -10,7 +10,7 @@
 git clone --single-branch https://chromium.googlesource.com/libyuv/libyuv
 
 cd libyuv
-git checkout d53f1be
+git checkout 9b17af9b
```
wantehchang, line 13: We should wait until this revision is used in Chrome. I can write a Chrome CL to update its third_party/libyuv to this revision.

The src/reformat_libyuv.c change in this pull request is fine. Would you like to remove the ext/libyuv.cmd change from this pull request?

wantehchang: See https://chromium-review.googlesource.com/c/chromium/src/+/3822634 for the libyuv update in Chrome.

Also please update the following item in CHANGELOG.md:
```
* Update libyuv.cmd: d53f1be (version 1837)
```

wantehchang (changes requested):

ext/libyuv.cmd
vigneshvg: I have removed the libyuv.cmd change from this PR. I will do it separately once it lands in chromium.

src/reformat_libyuv.c
```diff
@@ -431,6 +431,24 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
                 return AVIF_RESULT_REFORMAT_FAILED;
             }
             return AVIF_RESULT_OK;
+        } else if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV422) {
+#if LIBYUV_VERSION >= 1838
+            if (I422ToARGBMatrixFilter(image->yuvPlanes[AVIF_CHAN_Y],
```
wantehchang, line 436: We should call `I422ToRGB565Matrix()`, right? I am sorry I didn't notice this error earlier.

Also test `AVIF_PIXEL_FORMAT_YUV422` before `AVIF_PIXEL_FORMAT_YUV420`, i.e., in the order of the enum defintion.

wantehchang (changes requested):

src/reformat_libyuv.c
vigneshvg: Sorry it seems like i made a mess with the copy pasting on this CL.

I have updated it now and verified that it works. Please take another look.

wantehchang (approved):
LGTM.

PR #1029, opened 2022-08-10
Use per-row conditional in avifImageIdentity8ToRGB8Color

vigneshvg:
Performance benchmarking showed that having the branch "per-row"
instead of "per-pixel" makes the code much faster (on
microbenchmarks it was close to ~15% faster).

Note that the other functions don't have this problem because the
branch is not the bottleneck on the other functions.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -448,13 +448,15 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
     return AVIF_RESULT_OK;
 }
 
+#define RGB565(R, G, B) ((uint16_t)(((B) >> 3) | (((G) >> 2) << 5) | (((R) >> 3) << 11)))
+
 static void avifStoreRGB8Pixel(avifRGBFormat format, uint8_t R, uint8_t G, uint8_t B, uint8_t * ptrR, uint8_t * ptrG, uint8_t * ptrB)
 {
     if (format == AVIF_RGB_FORMAT_RGB_565) {
         // References for RGB565 color conversion:
         // * https://docs.microsoft.com/en-us/windows/win32/directshow/working-with-16-bit-rgb
         // * https://chromium.googlesource.com/libyuv/libyuv/+/9892d70c965678381d2a70a1c9002d1cf136ee78/source/row_common.cc#2362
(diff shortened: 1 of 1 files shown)


src/reformat.c
```diff
@@ -448,13 +448,15 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
     return AVIF_RESULT_OK;
 }
 
+#define RGB565(R, G, B) (uint16_t)(((B) >> 3) | (((G) >> 2) << 5) | (((R) >> 3) << 11))
```
wantehchang, line 451: Nit: Parenthesize the whole expression:

```
((uint16_t)(((B) >> 3) | (((G) >> 2) << 5) | (((R) >> 3) << 11)))
```

src/reformat.c
```diff
@@ -1039,12 +1041,23 @@ static avifResult avifImageIdentity8ToRGB8ColorFullRange(const avifImage * image
...
-        for (uint32_t i = 0; i < image->width; ++i) {
-            avifStoreRGB8Pixel(rgb->format, ptrV[i], ptrY[i], ptrU[i], ptrR, ptrG, ptrB);
-
-            ptrR += rgbPixelBytes;
-            ptrG += rgbPixelBytes;
-            ptrB += rgbPixelBytes;
+        // This is intentionally a per-row conditional instead of a per-pixel
+        // conditional. This makes the "else" path (much more common than the
+        // "if" path) much faster than having a per-pixel branch.
+        if (rgb->format == AVIF_RGB_FORMAT_RGB_565) {
```
wantehchang, line 1047: Optional: Move this conditional outside the outer for loop.

wantehchang (approved):
LGTM with two suggested changes (one optional). Thanks.

src/reformat.c
vigneshvg: this would mean duplicating the outer for loop as well twice for no good reason. so i will leave this as-is.

src/reformat.c
vigneshvg: Done.

src/reformat.c
wantehchang: The `AVIF_RGB_FORMAT_RGB_565` version of the outer loop can omit the `ptrG` and `ptrB` variables. Since `rgb->format == AVIF_RGB_FORMAT_RGB_565` is a loop invariant, the compiler may make this this change as an optimization.

PR #1031, opened 2022-08-11
Update libyuv.cmd to 9b17af9b

vigneshvg:
This landed in chromium in
https://chromium-review.googlesource.com/c/chromium/src/+/3822634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ return values of avifImageCopy() and avifImageAllocatePlanes().
 * Update aom.cmd: v3.4.0
 * Update svt.cmd/svt.sh: v1.2.0
 * Update libgav1.cmd: v0.18.0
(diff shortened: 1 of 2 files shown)


wantehchang (approved):
LGTM. Not sure if it is more common to use seven or eight hex digits in an abbreviated git revision.

vigneshvg:
> LGTM. Not sure if it is more common to use seven or eight hex digits in an abbreviated git revision.

I usually get short hashes from: `git rev-parse --short <revision>`

wantehchang:
That's a good way to get short hashes. The reason I asked is that seven-digit short hashes also seem common. For example, they are used in the commit listings in repositories:

https://github.com/AOMediaCodec/libavif/commits/main
https://chromium.googlesource.com/libyuv/libyuv/

(I was just curious about this topic. No need to change anything.)

PR #1032, opened 2022-08-11
Refactor libyuv yuv->rgb conversion with function pointers

vigneshvg:
Use function pointers and look up tables based on RGB and
YUV image formats to call the appropriate libyuv function.

This eliminates a lot of repeated code (which can be error prone
when copy-pasting, refactoring, etc). It also makes adding new
libyuv functions simpler.

Tested with several AVIF images and the existing yuv and rgb tests
continue to pass.

### DIFF
--- a/apps/shared/y4m.c
+++ b/apps/shared/y4m.c
@@ -457,7 +457,8 @@ avifBool y4mWrite(const char * outputFilename, const avifImage * avif)
                     y4mHeaderFormat = "Cmono XYSCSS=400";
                     break;
                 case AVIF_PIXEL_FORMAT_NONE:
-                    // will error later; this case is here for warning's sake
+                case AVIF_PIXEL_FORMAT_COUNT:
+                    // will error later; these cases are here for warning's sake
                     break;
             }
             break;
@@ -476,7 +477,8 @@ avifBool y4mWrite(const char * outputFilename, const avifImage * avif)
                     y4mHeaderFormat = "Cmono10 XYSCSS=400";
                     break;
                 case AVIF_PIXEL_FORMAT_NONE:
-                    // will error later; this case is here for warning's sake
+                case AVIF_PIXEL_FORMAT_COUNT:
+                    // will error later; these cases are here for warning's sake
                     break;
             }
             break;
@@ -495,7 +497,8 @@ avifBool y4mWrite(const char * outputFilename, const avifImage * avif)
                     y4mHeaderFormat = "Cmono12 XYSCSS=400";
                     break;
                 case AVIF_PIXEL_FORMAT_NONE:
-                    // will error later; this case is here for warning's sake
+                case AVIF_PIXEL_FORMAT_COUNT:
+                    // will error later; these cases are here for warning's sake
                     break;
             }
             break;
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -28,6 +28,10 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
     if (rgb->format == AVIF_RGB_FORMAT_RGB_565 && rgb->depth != 8) {
         return AVIF_FALSE;
     }
+    if (image->yuvFormat <= AVIF_PIXEL_FORMAT_NONE || image->yuvFormat >= AVIF_PIXEL_FORMAT_COUNT ||
+        rgb->format < AVIF_RGB_FORMAT_RGB || rgb->format >= AVIF_RGB_FORMAT_COUNT) {
+        return AVIF_FALSE;
+    }
 
     // These matrix coefficients values are currently unsupported. Revise this list as more support is added.
     //
@@ -43,10 +47,6 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
     }
 
     if ((image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) && (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV444)) {
-        return AVIF_FALSE;
-    }
-
-    if (image->yuvFormat == AVIF_PIXEL_FORMAT_NONE) {
         return AVIF_FALSE;
     }
 
@@ -119,7 +119,7 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
             state->rgbOffsetBytesA = 0;
             break;
 
-        default:
+        case AVIF_RGB_FORMAT_COUNT:
             return AVIF_FALSE;
     }
 
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -341,6 +341,37 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
     return AVIF_RESULT_NOT_IMPLEMENTED;
 }
 
+// These defines are used to create a NULL reference to libyuv functions that
+// did not exist prior to a particular version of libyuv.
+#if LIBYUV_VERSION < 1838
+#define I422ToRGB565Matrix NULL
+#endif
+#if LIBYUV_VERSION < 1813
+#define I422ToARGBMatrixFilter NULL
+#define I420ToARGBMatrixFilter NULL
+#define I210ToARGBMatrixFilter NULL
+#define I010ToARGBMatrixFilter NULL
+#endif
+#if LIBYUV_VERSION < 1780
+#define I410ToARGBMatrix NULL
+#endif
+#if LIBYUV_VERSION < 1756
+#define I400ToARGBMatrix NULL
+#endif
+
+// Lookup table for isYVU. If the entry in this table is AVIF_TRUE, then it
+// means that we are using a libyuv function with R and B channels swapped,
+// which requires U and V planes also be swapped.
+static const avifBool lutIsYVU[AVIF_RGB_FORMAT_COUNT] = {
+    AVIF_TRUE,  // RGB
+    AVIF_TRUE,  // RGBA
+    AVIF_TRUE,  // ARGB
+    AVIF_FALSE, // BGR
+    AVIF_FALSE, // BGRA
+    AVIF_FALSE, // ABGR
+    AVIF_FALSE, // RGB_565
+};
+
 avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
                                        avifRGBImage * rgb,
                                        const struct YuvConstants * matrixYUV,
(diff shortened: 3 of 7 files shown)


src/reformat_libyuv.c
```diff
@@ -350,429 +368,136 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
...
-            }
-            return AVIF_RESULT_OK;
-        }
-    } else if (rgb->format == AVIF_RGB_FORMAT_BGR) {
-        // AVIF_RGB_FORMAT_BGR   *ToRGB24Matrix  matrixYUV
+    // AVIF_RGB_FORMAT_RGB_565   *ToRGB565Matrix  matrixYUV
 
-        if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) {
-            if (I420ToRGB24Matrix(image->yuvPlanes[AVIF_CHAN_Y],
+    // Look up table for YUV To RGB Matrix (without filter).
```
y-guyon, line 386: Would be slightly faster to understand if the `[7][5]` coordinates were described.

src/reformat_libyuv.c
```diff
@@ -341,6 +341,24 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
     return AVIF_RESULT_NOT_IMPLEMENTED;
 }
 
+// These defines are used to to create a NULL reference to libyuv functions that
+// did not exist prior to a particular version of libyuv.
+#if LIBYUV_VERSION < 1838
+#define I422ToRGB565Matrix NULL
```
wantehchang, line 347: Nit: This code may potentially break if libavif is compiled against a modified version of libyuv. For example, a copy of libyuv may have local modifications that cherry-pick these functions from future versions.

It would be better to define a macro like the following:

```
#if LIBYUV_VERSION >= 1838
#define I422_TO_RGB565_MATRIX I422ToRGB565Matrix
#else
#define I422_TO_RGB565_MATRIX NULL
#endif
```

And then use `I422_TO_RGB565_MATRIX` in the code below.

y-guyon (approved):

src/reformat_libyuv.c
vigneshvg: Good point. It's basically the number of RGB formats and number of YUV formats.

Some projects have a delimiter enum like AVIF_RGB_FORMAT_COUNT or AVIF_RGB_FORMAT_MAX. We don't have that in libavif, so i just used the values here.

I have replaced them with a local #define.

src/reformat_libyuv.c
```diff
@@ -341,6 +341,24 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
     return AVIF_RESULT_NOT_IMPLEMENTED;
 }
 
+// These defines are used to to create a NULL reference to libyuv functions that
```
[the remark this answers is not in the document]
vigneshvg: Done.

src/reformat_libyuv.c
```diff
@@ -341,6 +341,24 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
...
+#define I210ToARGBMatrixFilter NULL
+#define I010ToARGBMatrixFilter NULL
+#endif
+#if LIBYUV_VERSION < 1780
+#define I410ToARGBMatrix NULL
+#endif
+#if LIBYUV_VERSION < 1756
+#define I400ToARGBMatrix NULL
+#endif
+
```
vigneshvg, line 361: these defines are not strictly needed and we can include them inline in the loop up table definitions. but i thought it was more readable to keep all the version specific quirks together this way.

src/reformat_libyuv.c
```diff
@@ -350,429 +368,136 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
...
-                                   image->yuvPlanes[AVIF_CHAN_V],
-                                   image->yuvRowBytes[AVIF_CHAN_V],
-                                   rgb->pixels,
-                                   rgb->rowBytes,
-                                   matrixYUV,
-                                   image->width,
-                                   image->height) != 0) {
+    } else if (yuvToRgbMatrixFilter != NULL || yuvToRgbMatrix != NULL) {
+        int uPlaneIndex = isYVU ? AVIF_CHAN_V : AVIF_CHAN_U;
+        int vPlaneIndex = isYVU ? AVIF_CHAN_U : AVIF_CHAN_V;
```
wantehchang, line 468: Nit/optional: Might be more accurate to name tese variables `plane1Index` and `plane2Index` or `firstChromaIndex` and `secondChromaIndex`. But I understand what they mean.

src/reformat_libyuv.c
```diff
@@ -341,6 +341,27 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
...
+#endif
+#if LIBYUV_VERSION < 1780
+#define I410ToARGBMatrix NULL
+#endif
+#if LIBYUV_VERSION < 1756
+#define I400ToARGBMatrix NULL
+#endif
+
+#define RGB_PIXEL_FORMATS 7
+#define YUV_PIXEL_FORMATS 5
```
wantehchang, line 363: Optional: The lookup tables depend on the current `avifRGBFormat` and `avifPixelFormat` enum values, not just how many enum values there are.

Since these two enum types are not based on some standard, the enum values may change until we freeze the API. It would be good to figure out a way to assert the asseumptions,

src/reformat_libyuv.c
wantehchang: Nit: Add a comment to note that these are the numbers of `avifRGBFormat` and `avifPixelFormat` enum values.

src/reformat_libyuv.c
```diff
@@ -788,208 +516,113 @@ avifResult avifImageYUVToRGBLibYUV10bpc(const avifImage * image,
...
+        { NULL, NULL, I210ToARGBMatrixFilter, I010ToARGBMatrixFilter, NULL }, // RGBA
+        {},                                                                   // ARGB
+        {},                                                                   // BGR
+        { NULL, NULL, I210ToARGBMatrixFilter, I010ToARGBMatrixFilter, NULL }, // BGRA
+        {},                                                                   // ABGR
+        {},                                                                   // RGB_565
+    };
+
+    // Look up table for isYVU.
+    avifBool lutIsYVU[RGB_PIXEL_FORMATS] = {
```
wantehchang, line 571: I am surprised that this `lutIsYVU` table is different from the other `lutIsYVU` table. Why?

Can we make these two `lutIsYVU` tables the same so that the table can be declared as const and shared by the two functions?

wantehchang (changes requested):

src/reformat_libyuv.c
vigneshvg: actually these two tables are the same. in the 10bit case we only use two values (RGBA and BGRA) which are the same as the table above and the other values are unused (and being set to false in this table since it doesn't matter).

But yes, sharing this table makes sense. Done.

src/reformat_libyuv.c
```diff
@@ -341,6 +341,38 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
...
+#define I410ToARGBMatrix NULL
+#endif
+#if LIBYUV_VERSION < 1756
+#define I400ToARGBMatrix NULL
+#endif
+
+#define RGB_PIXEL_FORMATS 7 // Number of enum values in avifRGBFormat.
+#define YUV_PIXEL_FORMATS 5 // Number of enum values in avifPixelFormat.
+
+// Look up table for isYVU.
```
wantehchang, line 365: Nit: It would be good to define `isYVU`. It means we are using a libyuv function with R and B swaped, which requires U and V also be swapped.

src/reformat_libyuv.c
wantehchang: Yes, let's add `AVIF_RGB_FORMAT_COUNT` (note the name does not have `PIXEL`) and `AVIF_PIXEL_FORMAT_COUNT`.

wantehchang (changes requested):
Vignesh: Overall this looks good. I suggest a few changes.

I do feel the lookup tables are less readable than switch statements or if-else-if statements. An alternative is to set up function pointers using switch statements or if-else-if statements, as Yannis did in `avifImageRGBToYUVLibYUV8bpc()`. Among these alternatives, there is a trade-off between code compactness and resistance to enum changes. In any case, let's go ahead with the lookup table approach.

src/reformat_libyuv.c
```diff
@@ -350,429 +382,125 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
...
+                                        const uint8_t *,
+                                        int,
+                                        uint8_t *,
+                                        int,
+                                        const struct YuvConstants *,
+                                        int,
+                                        int,
+                                        enum FilterMode);
+    YUVToRGBMatrixFilter lutYuvToRgbMatrixFilter[RGB_PIXEL_FORMATS][YUV_PIXEL_FORMATS] = {
+        {},                                                                   // RGB
```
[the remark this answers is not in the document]
wantehchang: UPDATE: I found that the AppVeyor build failed because of this:

```
C:\projects\libavif-93od3\src\reformat_libyuv.c(427,10): error C2059: syntax error: '}' [C:\projects\libavif-93od3\build\avif.vcxproj]
C:\projects\libavif-93od3\src\reformat_libyuv.c(538,10): error C2059: syntax error: '}' [C:\projects\libavif-93od3\build\avif.vcxproj]
C:\projects\libavif-93od3\src\reformat_libyuv.c(561,10): error C2059: syntax error: '}' [C:\projects\libavif-93od3\build\avif.vcxproj]
```

vigneshvg:
> Vignesh: Overall this looks good. I suggest a few changes.
> 
> I do feel the lookup tables are less readable than switch statements or if-else-if statements.

I think this is very subjective/perceptive. While we have a nested else-if ladder in our current code, we also have a "table-like" comment explaining what the nested else-if ladder is doing. So this CL is essentially trying to match the "table" in the comment with a simila table in the code.

> An alternative is to set up function pointers using switch statements or if-else-if statements, as Yannis did in `avifImageRGBToYUVLibYUV8bpc()`. Among these alternatives, there is a trade-off between code compactness and resistance to enum changes. In any case, let's go ahead with the lookup table approach.

apps/shared/y4m.c
```diff
@@ -457,6 +457,7 @@ avifBool y4mWrite(const char * outputFilename, const avifImage * avif)
                     y4mHeaderFormat = "Cmono XYSCSS=400";
                     break;
                 case AVIF_PIXEL_FORMAT_NONE:
+                case AVIF_PIXEL_FORMAT_COUNT:
```
wantehchang, line 460: Nit: The three comments at lines 461, 481, 501 should be updated to use plural nouns:

```
                    // will error later; these cases are here for warning's sake
```

src/reformat.c
```diff
@@ -28,6 +28,10 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
     if (rgb->format == AVIF_RGB_FORMAT_RGB_565 && rgb->depth != 8) {
         return AVIF_FALSE;
     }
+    if (image->yuvFormat <= AVIF_PIXEL_FORMAT_NONE || image->yuvFormat >= AVIF_PIXEL_FORMAT_COUNT ||
+        rgb->format < AVIF_RGB_FORMAT_RGB || rgb->format >= AVIF_RGB_FORMAT_COUNT) {
+        return AVIF_FALSE;
+    }
```
wantehchang, line 34: Optional: We should be able to omit the default case at lines 122-123 now:

```
        default:
            return AVIF_FALSE;
```

tests/avifyuv.c
```diff
@@ -41,6 +41,8 @@ static const char * rgbFormatToString(avifRGBFormat format)
             return "ABGR";
         case AVIF_RGB_FORMAT_RGB_565:
             return "RGB_565";
+        case AVIF_RGB_FORMAT_COUNT:
+            return "Unknown";
```
wantehchang, line 45: Nit: Use a `break` statement instead.

src/reformat_libyuv.c
```diff
@@ -350,428 +381,122 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
...
+    // AVIF_RGB_FORMAT_RGB_565   *ToRGB565Matrix  matrixYUV
 
-        if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) {
-            if (I420ToRGB24Matrix(image->yuvPlanes[AVIF_CHAN_Y],
+    avifBool isYVU = lutIsYVU[rgb->format];
+    const struct YuvConstants * matrix = isYVU ? matrixYVU : matrixYUV;
+    if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) {
+        // Lookup table for YUV400 to RGB Matrix.
+        typedef int (*YUV400ToRGBMatrix)(const uint8_t *, int, uint8_t *, int, const struct YuvConstants *, int, int);
+        YUV400ToRGBMatrix lutYuv400ToRgbMatrix[AVIF_RGB_FORMAT_COUNT] = {
```
wantehchang, line 404: We can deal with this in a follow-up pull request. Conceptually these lookup tables are all constants and can be declared as static const arrays outside the functions. (If they are declared as static inside a function, then they may not be initialized until the first time the function is called.)

What I am not sure about is whether the addresses of these libyuv functions are true constants, because obviously they depend on address space layout.

wantehchang (approved):
LGTM. Just three simple changes (one optional) and a comment.

wantehchang (approved):
LGTM.

PR #1035, opened 2022-08-16
Add SharpYuv RGB->YUV conversion function using libsharpyuv. Not used yet.

maryla-uc:


ext/libsharpyuv.cmd
```diff
@@ -0,0 +1,21 @@
+: # If you want to use a local build of libyuv, you must clone the libwebp repo in this directory first,
```
y-guyon, line 1: Update comment

y-guyon:
> Maryla: The AppVeyor build reported the following MSVC warning in src/reformat_libyuv.c. Since your pull request doesn't modify that file, it is a pre-existing problem that we need to fix first.
> 
> ```
> src\reformat_libyuv.c(109,89): warning C4242: '=': conversion from 'const uint64_t' to 'int', possible loss of data
> ```

Fixed in https://github.com/AOMediaCodec/libavif/pull/1036

y-guyon (approved):

CMakeLists.txt
```diff
@@ -228,6 +246,15 @@ else()
     message(STATUS "libavif: libyuv not found; libyuv-based fast paths disabled.")
 endif()
 
+find_package(WebP) # not required
+if(TARGET WebP::sharpyuv)
```
kmilos, line 250: `FIndWebP.cmake` is not shipped w/ CMake AFAICT (nor does the standard libwebp package supply the CMake config files), so please supply one.

kmilos: Sorry, [the missing CMake config file could be just a MSYS2 packaging issue](https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-libwebp?repo=ucrt64), will investigate (it seems it is in the source tree for the 1.2.4 tag...)

kmilos: Ah, MSYS2 (and I suspect many other distros) still build and package libwebp using autotools, so it still might not be advisable to rely on this? Maybe also include a fallback searching via pkg-config?

src/reformat_libsharpyuv.c
```diff
@@ -0,0 +1,51 @@
+// Copyright 2022 Google LLC. All rights reserved.
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "avif/internal.h"
+#include <stdio.h>
```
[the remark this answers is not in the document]
maryla-uc: Done.

ext/libsharpyuv.cmd
maryla-uc: Done.

CMakeLists.txt
maryla-uc: Thank you for pointing this out and investigating! I'll look into it.

kmilos: Thanks. Can you also update WebP's CMakeLists so the configure_file() commands for all the .pc files include "@ONLY" please?

jzern: > Thanks. Can you also update WebP's CMakeLists so the [configure_file() commands for all the .pc files](https://github.com/webmproject/libwebp/blob/9b902cbacedd0fb19de5ce4475d7a8345d7fe9cf/CMakeLists.txt#L158) include `@ONLY` please (i.e. behaviour is on par w/ autotools)?

Thanks for pointing that out. I made the change in:
https://chromium-review.googlesource.com/c/webm/libwebp/+/3836718

.github/workflows/ci.yml
```diff
@@ -62,6 +62,10 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: bash libjpeg.cmd
+    - name: Build libsharpyuv
+      if: steps.cache-ext.outputs.cache-hit != 'true'
+      working-directory: ./ext
+      run: bash libsharpyuv.cmd
```
wantehchang, line 68: Also add this step to the new file .github/workflows/ci-windows.yml. Change the last line to:

```
      run: ./libsharpyuv.cmd
```

.github/workflows/ci.yml
```diff
@@ -62,6 +62,10 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: bash libjpeg.cmd
+    - name: Build libsharpyuv
```
wantehchang, line 65: Add `-DAVIF_LOCAL_LIBSHARPYUV=ON` to the cmake command at lines 79-88?

ext/libsharpyuv.cmd
```diff
@@ -0,0 +1,21 @@
...
+
+: # libsharpyuv is part of the libwebp repo.
+git clone --single-branch https://chromium.googlesource.com/webm/libwebp
+
+cd libwebp
+git checkout c170df3
+
+mkdir build
+cd build
+cmake -G Ninja -DBUILD_SHARED_LIBS=0 -DCMAKE_BUILD_TYPE=Release ..
```
wantehchang, line 19: Nit: Change `0` to `OFF` in `-DBUILD_SHARED_LIBS=0`.

include/avif/internal.h
```diff
@@ -151,6 +151,12 @@ avifResult avifImageRGBToYUVLibYUV(avifImage * image, const avifRGBImage * rgb);
 // * [any other error]           - Return error to caller
 avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb);
 
+// Returns:
+// * AVIF_RESULT_OK              - Converted successfully with libsharpyuv
+// * AVIF_RESULT_NOT_IMPLEMENTED - libsharpyuv is not compiled in, or doesn't support this type of input
+// * [any other error]           - Return error to caller
+avifResult avifImageRGBToYUVLibSharpYUV(avifImage * image, const avifRGBImage * rgb, const avifReformatState * state);
```
wantehchang, line 158: The commit message should mention that this function is not used.

wantehchang (changes requested):

CMakeLists.txt
maryla-uc: Done, I've added a Findlibsharpyuv.cmake that relies on pkg-config.

.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     - name: Setup cmake
       uses: jwlawson/actions-setup-cmake@v1.12
       with:
-        cmake-version: '3.13.x'
+        cmake-version: '3.17.x'
```
wantehchang, line 29: Should we change `cmake_minimum_required(VERSION 3.13)` in CMakeLists.txt to `cmake_minimum_required(VERSION 3.17)`?

If not, we should add a comment here to explain why the cmake version here differs from the minimum required cmake version in CMakeLists.txt.

CMakeLists.txt
```diff
@@ -187,6 +204,7 @@ set(AVIF_SRCS
     src/read.c
     src/reformat.c
     src/reformat_libyuv.c
+    src/reformat_libsharpyuv.c
```
wantehchang, line 207: Nit: The files here are listed in alphabetical order, so this should be listed before `src/reformat_libyuv.c`.

PR #1036, opened 2022-08-16
Fix MSVC cast warning in avifABGRToJ420()

y-guyon:
src\reformat_libyuv.c(109,89): warning C4242: '=': conversion from
'const uint64_t' to 'int', possible loss of data

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -95,11 +95,11 @@ static int avifABGRToJ420(const uint8_t * src_abgr,
     const int src_stride_argb = width * 4;
     const int soft_allocation_limit = 16384; // Arbitrarily chosen trade-off between CPU and memory footprints.
     int num_allocated_rows;
-    if ((height == 1) || ((uint64_t)src_stride_argb * height <= (uint64_t)soft_allocation_limit)) {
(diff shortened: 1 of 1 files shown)


wantehchang (approved):
LGTM.

src/reformat_libyuv.c
```diff
@@ -93,9 +93,9 @@ static int avifABGRToJ420(const uint8_t * src_abgr,
     // A temporary buffer is needed to swap the R and B channels before calling ARGBToJ420().
     uint8_t * src_argb;
     const int src_stride_argb = width * 4;
-    const uint64_t soft_allocation_limit = 16384; // Arbitrarily chosen trade-off between CPU and memory footprints.
+    const int soft_allocation_limit = 16384; // Arbitrarily chosen trade-off between CPU and memory footprints.
     int num_allocated_rows;
-    if ((height == 1) || ((uint64_t)src_stride_argb * height <= soft_allocation_limit)) {
+    if ((height == 1) || ((uint64_t)src_stride_argb * height <= (uint64_t)soft_allocation_limit)) {
```
wantehchang, line 98: Optional: We could also cast `src_stride_argb` to `int64_t` instead. Then we should not need to cast `soft_allocation_limit` because there won't be a signed-unsigned comparison.

If you make this change, make a similar change to the cast in line 102.

PR #1037, opened 2022-08-16
Compile with Clang -Wshorten-64-to-32 warning flag

wantehchang:
Unfortunately, this does not warn about the conversion from
'const uint64_t' to 'int' that MSVC warns about, which is fixed in
https://github.com/AOMediaCodec/libavif/pull/1036.


apps/avifdec.c
```diff
@@ -159,11 +159,12 @@ int main(int argc, char * argv[])
             ignoreICC = AVIF_TRUE;
         } else if (!strcmp(arg, "--size-limit")) {
             NEXTARG();
-            imageSizeLimit = strtoul(arg, NULL, 10);
-            if ((imageSizeLimit > AVIF_DEFAULT_IMAGE_SIZE_LIMIT) || (imageSizeLimit == 0)) {
+            unsigned long value = strtoul(arg, NULL, 10);
+            if ((value > AVIF_DEFAULT_IMAGE_SIZE_LIMIT) || (value == 0)) {
```
wantehchang, line 163: If we pass the test here, we know `value <= AVIF_DEFAULT_IMAGE_SIZE_LIMIT <= UINT32_MAX`, so the `uint32_t` cast at line 167 is safe.

y-guyon (approved):

jzern (approved):

jzern:
I don't see CI runs for this patch. There's a warning that was missed:

```
src/read.c:3092:99: error: implicit conversion loses integer precision: 'uint64_t' (aka 'unsigned long long') to 'size_t' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
            avifResult readResult = avifDecoderItemRead(item, decoder->io, &itemContents, sample->offset, bytesToRead, &decoder->diag);
                                    ~~~~~~~~~~~~~~~~~~~                                   ~~~~~~~~^~~~~~
```

wantehchang:
James: Thanks for the report. I only tested 64-bit builds.

The [macos-latest CI](https://github.com/AOMediaCodec/libavif/runs/7865274310?check_suite_focus=true) should have tested this patch because it uses AppleClang 13.0.0.13000029 , but it is a 64-bit build.

PR #1039, opened 2022-08-16
Ext: Update to SVT-AV1 v1.2.1

wantehchang:
Do not build the SVT-AV1 decoder and apps. src/codec_svt.c only supports
using SVT-AV1 as an AV1 encoder, and it only needs the SVT-AV1 library.


wantehchang:
@1480c1 Please review. Thanks

hassount (approved):

PR #1042, opened 2022-08-17
Delete a stale (and malformed) comment

wantehchang:
The comment should have been deleted in
https://github.com/AOMediaCodec/libavif/pull/894. Also, the comment is
malformed for Windows batch files.

### DIFF
--- a/ext/dav1d.cmd
+++ b/ext/dav1d.cmd
@@ -8,7 +8,6 @@
 : # If you're running this on Windows, be sure you've already run this (from your VC2019 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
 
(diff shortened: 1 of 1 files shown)


ext/dav1d.cmd
```diff
@@ -8,7 +8,6 @@
 : # If you're running this on Windows, be sure you've already run this (from your VC2019 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
 
-# When updating the dav1d version, make the same change to dav1d_oss_fuzz.sh.
```
wantehchang, line 11: This comment generates the following message on Windows:

```
dav1d.cmd
'#' is not recognized as an internal or external command,
operable program or batch file.
```

The comment should start with `: #` instead of `#`.

jzern (approved):

PR #1043, opened 2022-08-17
Disable some strict mode flags in the JNI wrapper

vigneshvg:
It is ok for the JNI wrapper to be more permissive.

Also update the cmake version required in build.gradle. Some
sdk installations use 3.10 by default which is not enough to build
libavif.

### DIFF
--- a/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
+++ b/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
@@ -52,6 +52,15 @@ bool CreateDecoderAndParse(AvifDecoderWrapper* const decoder,
   }
   decoder->decoder->ignoreXMP = AVIF_TRUE;
   decoder->decoder->ignoreExif = AVIF_TRUE;
+
+  // Turn off 'clap' (clean aperture) property validation. The JNI wrapper
+  // ignores the 'clap' property.
+  decoder->decoder->strictFlags &= ~AVIF_STRICT_CLAP_VALID;
+  // Allow 'pixi' (pixel information) property to be missing. Older versions of
+  // libheif did not add the 'pixi' item property to AV1 image items (See
+  // crbug.com/1198455).
+  decoder->decoder->strictFlags &= ~AVIF_STRICT_PIXI_REQUIRED;
+
   avifResult res = avifDecoderSetIOMemory(decoder->decoder, buffer, length);
   if (res != AVIF_RESULT_OK) {
     LOGE("Failed to set AVIF IO to a memory reader.");
--- a/android_jni/avifandroidjni/src/main/jni/CMakeLists.txt
+++ b/android_jni/avifandroidjni/src/main/jni/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Copyright 2022 Google LLC. All rights reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
-cmake_minimum_required(VERSION 3.7)
+cmake_minimum_required(VERSION 3.13)
 
 project(avif_android_jni)
 


android_jni/avifandroidjni/build.gradle
```diff
@@ -24,6 +24,7 @@ android {
     }
     externalNativeBuild {
         cmake {
+            version '3.23.3'
```
wantehchang, line 27: In android_jni/avifandroidjni/src/main/jni/CMakeLists.txt, we have:

```
cmake_minimum_required(VERSION 3.7)
```

Should we change that to the actual minimum required version?

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -52,6 +52,7 @@ bool CreateDecoderAndParse(AvifDecoderWrapper* const decoder,
   }
   decoder->decoder->ignoreXMP = AVIF_TRUE;
   decoder->decoder->ignoreExif = AVIF_TRUE;
+  decoder->decoder->strictFlags = AVIF_STRICT_DISABLED;
```
wantehchang, line 55: It is better to disable specific flags. Chrome's third_party/blink/renderer/platform/image-decoders/avif/avif_image_decoder.cc has the following:

```
    // Turn off libavif's 'clap' (clean aperture) property validation. (We
    // ignore the 'clap' property.)
    decoder_->strictFlags &= ~AVIF_STRICT_CLAP_VALID;
    // Allow the PixelInformationProperty ('pixi') to be missing in AV1 image
    // items. libheif v1.11.0 or older does not add the 'pixi' item property to
    // AV1 image items. (This issue has been corrected in libheif v1.12.0.) See
    // crbug.com/1198455.
    decoder_->strictFlags &= ~AVIF_STRICT_PIXI_REQUIRED;
```

Pleae see if you can get by with only `decoder->decoder->strictFlags &= ~AVIF_STRICT_PIXI_REQUIRED;`

wantehchang (changes requested):

android_jni/avifandroidjni/build.gradle
wantehchang: libavif's top-level CMakeLists.txt has the following:

```
cmake_minimum_required(VERSION 3.13)
```

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
vigneshvg: The current android platform's AVIF parser is quite permissive and does not care about most errors (it only cares about those boxes that are actually necessary to decode the image).

So it is better for the android JNI wrapper to try and match that as much as possible. So turning off strict mode altogether is probably better.

android_jni/avifandroidjni/build.gradle
vigneshvg: Done.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
wantehchang: The libavif decoder performs a lot of standard compliance checks unconditionally. It is impossible for the Android JNI wrapper to match the permissiveness of the Android platform's AVIF parser. It's better to match the strictness of Chrome and help improve standard compliance of AVIF files.

vigneshvg: > The libavif decoder performs a lot of standard compliance checks unconditionally. It is impossible for the Android JNI wrapper to match the permissiveness of the Android platform's AVIF parser. 

Hmm, the whole point of bundling libavif along with an Android app is to make sure that avif decoding is available irrespective of the android platform version. it is also generally considered to be a super-set of what the android platform supports (like 10bit, 12bit, etc). So it would be weird if an avif file can be parsed and rendered by the platform but the libavif wrapper fails to display it because it failed some compliance checks.

> It's better to match the strictness of Chrome and help improve standard compliance of AVIF files.

Also in general, i believe it is not fair to enforce conformance checks on decoders in order to improve compliance. Because once such files are in the wild, they are in the wild and we should not go out of our way to fail on them especially if we can decode them.

wantehchang: The decision to add standard compliance checks to libavif was supported by several members of the AOM Still Image Working Group. The rationale is that it will help prevent non-standard-compliant AVIF files from being created, and we can only do this before too many non-compliant AVIF files have been created.

Based on Chrome's AVIF bug reports, I recommend only disabling `AVIF_STRICT_PIXI_REQUIRED`, but ultimately this is your call. My main request is to disable specific strict flags instead of `AVIF_STRICT_DISABLED`, with comments that justify why a strict flag is disabled. For example, this is how Chrome disables two strict flags:

```
    // Turn off libavif's 'clap' (clean aperture) property validation. (We
    // ignore the 'clap' property.)
    decoder_->strictFlags &= ~AVIF_STRICT_CLAP_VALID;
    // Allow the PixelInformationProperty ('pixi') to be missing in AV1 image
    // items. libheif v1.11.0 or older does not add the 'pixi' item property to
    // AV1 image items. (This issue has been corrected in libheif v1.12.0.) See
    // crbug.com/1198455.
    decoder_->strictFlags &= ~AVIF_STRICT_PIXI_REQUIRED;
```

Note: We may turn on `AVIF_STRICT_CLAP_VALID` in Chrome when we support 'clap' (likely with restrictions).

vigneshvg: > The decision to add standard compliance checks to libavif was supported by several members of the AOM Still Image Working Group.

Yes i understand that from an encoder perspective. We should strive as much as possible to not create invalid AVIF files. But it is impossible for us to control the entire ecosystem (there will always be that one popular tool that is stuck on an older version of ffmpeg or something that will produce slightly invalid AVIF files). Ultimately, i still feel that if we can decode a file, we should not go out of our way to bail out.

> The rationale is that it will help prevent non-standard-compliant AVIF files from being created, and we can only do this before too many non-compliant AVIF files have been created.
> 
Some of the files in the "official" AV1 github samples fail the CLAP check: https://github.com/AOMediaCodec/av1-avif/tree/master/testFiles/Link-U (kimono.crop.avif for example). So such files are already in the wild. :)

> Based on Chrome's AVIF bug reports, I recommend only disabling `AVIF_STRICT_PIXI_REQUIRED`, but ultimately this is your call. My main request is to disable specific strict flags instead of `AVIF_STRICT_DISABLED`, with comments that justify why a strict flag is disabled. For example, this is how Chrome disables two strict flags:
> 
> ```
>     // Turn off libavif's 'clap' (clean aperture) property validation. (We
>     // ignore the 'clap' property.)
>     decoder_->strictFlags &= ~AVIF_STRICT_CLAP_VALID;
>     // Allow the PixelInformationProperty ('pixi') to be missing in AV1 image
>     // items. libheif v1.11.0 or older does not add the 'pixi' item property to
>     // AV1 image items. (This issue has been corrected in libheif v1.12.0.) See
>     // crbug.com/1198455.
>     decoder_->strictFlags &= ~AVIF_STRICT_PIXI_REQUIRED;
> ```
> 
> Note: We may turn on `AVIF_STRICT_CLAP_VALID` in Chrome when we support 'clap' (likely with restrictions).

I have copied the specific flags from chromium in this CL. PTAL.

wantehchang (approved):
LGTM.

wantehchang (commented):
Please see if the commit message should be updated. Thanks.

wantehchang (commented):
Also, no need to wait for AppVeyor build to complete now, because the new windows-latest CI build provides MSVC build coverage.

vigneshvg:
> Please see if the commit message should be updated. Thanks.

Done.

wantehchang (approved):
Still LGTM.

PR #1044, opened 2022-08-17
Fix Clang and GCC -Wswitch-enum warnings

wantehchang:
-Wswitch-enum warns about unhandled enum values in switch statements
even when the switch statements have a default case.

MSVC also warns about this when /Wall is specified. Since we compile
with MSVC's /Wall option on Windows, this pull request is a prerequisite
for https://github.com/AOMediaCodec/libavif/pull/1011.

### DIFF
--- a/src/codec_rav1e.c
+++ b/src/codec_rav1e.c
@@ -90,6 +90,7 @@ static avifResult rav1eCodecEncodeImage(avifCodec * codec,
                     codec->internal->yShift = 1;
                     break;
                 case AVIF_PIXEL_FORMAT_NONE:
+                case AVIF_PIXEL_FORMAT_COUNT:
                 default:
                     return AVIF_RESULT_UNKNOWN_ERROR;
             }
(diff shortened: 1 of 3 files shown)


src/codec_rav1e.c
```diff
@@ -90,6 +90,7 @@ static avifResult rav1eCodecEncodeImage(avifCodec * codec,
                     codec->internal->yShift = 1;
                     break;
                 case AVIF_PIXEL_FORMAT_NONE:
+                case AVIF_PIXEL_FORMAT_COUNT:
```
vigneshvg, line 93: i updated this enum on codec_libaom.c because appveyor was failing without this.

i should've checked the other files as well while i was at it. sorry about that.

i think these are not being built on the appveyor presubmit builds.

vigneshvg (approved):

src/codec_rav1e.c
wantehchang: Vignesh: No worries, and thank you for the review.

I had to remove these AV1 codecs (libgav1, rav1e, and SVT-AV1) from the AppVeyor build because it was taking longer than 60 minutes and timing out. I am trying to add a GitHub CI for Windows MSVC build in https://github.com/AOMediaCodec/libavif/pull/1011. Once that pull request is merged, we will only use the AppVeyor build for Windows binary releases.

vigneshvg: Sounds good!

PR #1046, opened 2022-08-17
avifenc: Set YUV format to 400 for grayscale PNGs

vigneshvg:
If the input is a grayscale PNG file, then AVIF output format can
be set to monochrome (YUV400) if the user does not set the --yuv
option explicitly.

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -127,9 +127,14 @@ avifBool avifPNGRead(const char * inputFilename, avifImage * avif, avifPixelForm
     avif->height = rawHeight;
     avif->yuvFormat = requestedFormat;
     if (avif->yuvFormat == AVIF_PIXEL_FORMAT_NONE) {
-        // Identity is only valid with YUV444.
-        avif->yuvFormat = (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) ? AVIF_PIXEL_FORMAT_YUV444
-                                                                                          : AVIF_APP_DEFAULT_PIXEL_FORMAT;
+        if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
+            // Identity is only valid with YUV444.
+            avif->yuvFormat = AVIF_PIXEL_FORMAT_YUV444;
+        } else if ((rawColorType == PNG_COLOR_TYPE_GRAY) || (rawColorType == PNG_COLOR_TYPE_GRAY_ALPHA)) {
+            avif->yuvFormat = AVIF_PIXEL_FORMAT_YUV400;
+        } else {
+            avif->yuvFormat = AVIF_APP_DEFAULT_PIXEL_FORMAT;
+        }
     }
     avif->depth = requestedDepth;
     if (avif->depth == 0) {


apps/shared/avifpng.c
```diff
@@ -128,8 +128,13 @@ avifBool avifPNGRead(const char * inputFilename, avifImage * avif, avifPixelForm
     avif->yuvFormat = requestedFormat;
     if (avif->yuvFormat == AVIF_PIXEL_FORMAT_NONE) {
         // Identity is only valid with YUV444.
-        avif->yuvFormat = (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) ? AVIF_PIXEL_FORMAT_YUV444
-                                                                                          : AVIF_APP_DEFAULT_PIXEL_FORMAT;
+        if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
+            avif->yuvFormat = AVIF_PIXEL_FORMAT_YUV444;
```
wantehchang, line 132: Please move the comment `// Identity is only valid with YUV444.` into this block.

apps/shared/avifpng.c
```diff
@@ -128,8 +128,13 @@ avifBool avifPNGRead(const char * inputFilename, avifImage * avif, avifPixelForm
...
+        } else if (rawColorType == PNG_COLOR_TYPE_GRAY) {
```
wantehchang, line 133: Should we also allow `PNG_COLOR_TYPE_GRAY_ALPHA` here?

```
        } else if ((rawColorType == PNG_COLOR_TYPE_GRAY) || (rawColorType == PNG_COLOR_TYPE_GRAY_ALPHA)) {
```


See line 110.

wantehchang (changes requested):
Thank you for writing this pull request.

apps/shared/avifpng.c
vigneshvg: Yes, Done.

apps/shared/avifpng.c
```diff
@@ -127,9 +127,14 @@ avifBool avifPNGRead(const char * inputFilename, avifImage * avif, avifPixelForm
     avif->height = rawHeight;
     avif->yuvFormat = requestedFormat;
     if (avif->yuvFormat == AVIF_PIXEL_FORMAT_NONE) {
-        // Identity is only valid with YUV444.
-        avif->yuvFormat = (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) ? AVIF_PIXEL_FORMAT_YUV444
-                                                                                          : AVIF_APP_DEFAULT_PIXEL_FORMAT;
+        if (avif->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
+            // Identity is only valid with YUV444.
+            avif->yuvFormat = AVIF_PIXEL_FORMAT_YUV444;
+        } else if (rawColorType == PNG_COLOR_TYPE_GRAY || rawColorType == PNG_COLOR_TYPE_GRAY_ALPHA) {
```
wantehchang, line 133: Please quote the conditional expressions. This is libavif's local style. See the code above for examples.

We no longer require it, but follow the style whenever you remember:

```
        } else if ((rawColorType == PNG_COLOR_TYPE_GRAY) || (rawColorType == PNG_COLOR_TYPE_GRAY_ALPHA)) {
```

wantehchang (approved):
LGTM with a suggested change.

apps/shared/avifpng.c
vigneshvg: Done.

PR #1050, opened 2022-08-18
googletest.cmd: Pass -Dgtest_force_shared_crt=ON

wantehchang:
Pass the -Dgtest_force_shared_crt=ON option to cmake so that GoogleTest
is compiled with the DLL version of the MSVC runtime library (MSVCRT) on
Windows. By default GoogleTest uses the static library version of the
MSVCRT on Windows. All the other libraries (libavif, libaom, dav1d,
etc.) use the the MSVCRT DLL by default, so we will get errors about
using both the MSVCRT DLL and static library when linking
GoogleTest-based tests if GoogleTest is compiled with the MSVCRT static
library.

### DIFF
--- a/ext/googletest.cmd
+++ b/ext/googletest.cmd
@@ -11,6 +11,9 @@ git clone -b release-1.11.0 --depth 1 https://github.com/google/googletest.git
 cd googletest
 mkdir build
 cd build
-cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_GMOCK=OFF ..
+: # The gtest_force_shared_crt option makes gtest link the Microsoft C runtime library (CRT) dynamically
+: # on Windows:
+: # https://github.com/google/googletest/blob/main/googletest/README.md#visual-studio-dynamic-vs-static-runtimes
+cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_GMOCK=OFF -Dgtest_force_shared_crt=ON ..
 cmake --build .
 cd ../..


ext/googletest.cmd
```diff
@@ -11,6 +11,6 @@ git clone -b release-1.11.0 --depth 1 https://github.com/google/googletest.git
 cd googletest
 mkdir build
 cd build
-cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_GMOCK=OFF ..
+cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_GMOCK=OFF -Dgtest_force_shared_crt=ON ..
```
wantehchang, line 14: The `gtest_force_shared_crt` cmake option is documented in GoogleTest's README :
https://github.com/google/googletest/blob/main/googletest/README.md#visual-studio-dynamic-vs-static-runtimes

See also https://stackoverflow.com/questions/12540970/how-to-make-gtest-build-mdd-instead-of-mtd-by-default-using-cmake.

jzern: maybe add the first link above the command?

jzern (approved):

ext/googletest.cmd
wantehchang: Done.

jzern (approved):

jzern:
The opposite issue will now occur if you use -DBUILD_SHARED_LIBS=0. This is the default for cmake and may be where the discrepancy came from since libavif is forcing this on by default.

jzern:
> The opposite issue will now occur if you use -DBUILD_SHARED_LIBS=0. This is the default for cmake and may be where the discrepancy came from since libavif is forcing this on by default.

That or I may have mixed the debug target for libavif with the release build of googletest.

wantehchang:
James: This pull request fixed the /MT and /MD mismatch. I believe you ran into the /MD and /MDd mismatch when mixing a Debug build of libavif with Release builds of external dependencies. I don't have a good solution to that problem.

PR #1052, opened 2022-08-18
avifenc: Add the --autotiling option

wantehchang:
The --autotiling option sets --tilerowslog2 and --tilecolslog2
automatically.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -84,6 +84,7 @@ static void syntax(void)
            AVIF_QUANTIZER_LOSSLESS);
     printf("    --tilerowslog2 R                  : Set log2 of number of tile rows (0-6, default: 0)\n");
     printf("    --tilecolslog2 C                  : Set log2 of number of tile columns (0-6, default: 0)\n");
+    printf("    --autotiling                      : Set --tilerowslog2 and --tilecolslog2 automatically\n");
     printf("    -g,--grid MxN                     : Encode a single-image grid AVIF with M cols & N rows. Either supply MxN identical W/H/D images, or a single\n");
     printf("                                        image that can be evenly split into the MxN grid and follow AVIF grid image restrictions. The grid will adopt\n");
     printf("                                        the color profile of the first image supplied.\n");
@@ -436,8 +437,9 @@ int main(int argc, char * argv[])
     int maxQuantizer = 26;
     int minQuantizerAlpha = AVIF_QUANTIZER_LOSSLESS;
     int maxQuantizerAlpha = AVIF_QUANTIZER_LOSSLESS;
-    int tileRowsLog2 = 0;
-    int tileColsLog2 = 0;
+    int tileRowsLog2 = -1;
+    int tileColsLog2 = -1;
+    avifBool autoTiling = AVIF_FALSE;
     int speed = 6;
     int paspCount = 0;
     uint32_t paspValues[8]; // only the first two are used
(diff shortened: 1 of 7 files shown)


apps/avifenc.c
```diff
@@ -398,6 +399,68 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
     return AVIF_TRUE;
 }
 
+static int countLeadingZeros(uint32_t n)
```
y-guyon, line 402: What about
```cpp
static int countLeadingZeros(uint32_t n)
{
    int count = 32;
    while (n != 0) {
        --count;
        n >>= 1;
    }
    return count;
}
```

apps/avifenc.c
```diff
@@ -421,8 +484,9 @@ int main(int argc, char * argv[])
     int maxQuantizer = 26;
     int minQuantizerAlpha = AVIF_QUANTIZER_LOSSLESS;
     int maxQuantizerAlpha = AVIF_QUANTIZER_LOSSLESS;
-    int tileRowsLog2 = 0;
-    int tileColsLog2 = 0;
+    int tileRowsLog2 = -1;
+    int tileColsLog2 = -1;
```
y-guyon, line 488: Optional: `avifBool manualTiling = AVIF_FALSE;` might be simpler and no need for the `tileRowsLog2 < 0` checks below.

apps/avifenc.c
```diff
@@ -398,6 +399,68 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
...
+        }
+        if (tiles > (uint32_t)threads) {
+            tiles = threads;
+        }
+        int tilesLog2 = floorLog2(tiles);
+        // If the image's width is greater than the height, use more tile columns
+        // than tile rows.
+        if (width >= height) {
+            *tileRowsLog2 = tilesLog2 / 2;
+            *tileColsLog2 = tilesLog2 - *tileRowsLog2;
```
y-guyon, line 456: Consider this extreme example:
- 16384x64px so 4 tiles
- tilesLog2 = 2, tileRowsLog2 = 1, tileColsLog2 = 1

I would imagine splitting 16384x64px into 2x2 tiles is worse than 4x1 tiles.
The following or a similar ratio might help:
```cpp
*tileRowsLog2 = (tilesLog2 * height + (((width + height) >> 1)) / (width + height);
```

Unit testing this feature would be reassuring, unfortunately there is no way to get the number of tiles at decoding currently.

apps/avifenc.c
wantehchang: Done. Thanks.

The original code was copied from libgav1, which also has optimized code (based on compiler intrinsics) for common CPU architectures. In this use case, `n` is likely to be small, so your code is faster.

apps/avifenc.c
wantehchang: Thanks for the suggestion. This approach has the awkward property that `manualTiling` is not always the opposite of `autoTiling`, because `manualTiling` actually means whether `--tilerowslog2` or `--tilecolslog2` is specified. We can address this issue by renaming this boolean flag, such as `tilingExplicitlySet` or `tileConfigurationExplicitSet`, similar to the `cicpExplicitlySet` boolean flag.

apps/avifenc.c
wantehchang: Done. Thanks for the suggestion.

It should suffice to unit test the `setTileConfiguration()` function. That will require moving `setTileConfiguration()` to some other file, such as apps/shared/avifutil.{h,c}.

wantehchang:
I abandoned this pull request. It was replaced by https://github.com/AOMediaCodec/libavif/pull/1089.

PR #1056, opened 2022-08-19
Build the tests in the windows-latest os image

wantehchang:
In CMakeLists.txt, compile with MSVC warning /W4 instead of /Wall.
Remove the /wd (disable warning) options that are no longer needed.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -8,8 +8,7 @@
 #   * Runs ext/*.cmd as Windows batch files rather than using bash. This is
 #     important for ext/svt.cmd, to prevent it from running ext/svt.sh.
 #   * Builds with local zlib and libpng (-DAVIF_LOCAL_ZLIBPNG=ON).
-#   * Doesn't build GoogleTest and the tests, because MSVC's /Wall option
-#     reports too many warnings in the C++ test code.
+#   * Doesn't run ctest, because the test 'test_cmd' is a shell script.
 
 name: CI
 on: [push, pull_request]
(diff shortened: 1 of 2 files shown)


y-guyon (approved):

.github/workflows/ci-windows.yml
```diff
@@ -75,6 +74,10 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: ./zlibpng.cmd
+    - name: Build GoogleTest
+      #if: steps.cache-ext.outputs.cache-hit != 'true'
```
wantehchang, line 78: I think I understand how the cache works now. The cache step is specified as follows:

```
    - name: Cache external dependencies
      id: cache-ext
      uses: actions/cache@v3
      with:
        path: ext
        key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}
```

Note that the key is the hash of all the `ext/*.cmd` files. Since this pull request doesn't modify any `ext/*.cmd` file, we will get a cache hit, but the current cache for Windows doesn't contain a GoogleTest build. This is why we have to comment out the cache hit check in this pull request.

We can uncomment this line when we modify any `ext/*.cmd` file, or afterwards.

CMakeLists.txt
```diff
@@ -132,16 +137,9 @@ elseif(CMAKE_C_COMPILER_ID MATCHES "GNU")
 elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")
     message(STATUS "libavif: Enabling warnings for MSVC")
     add_definitions(
-        /Wall # All warnings
-        /wd4255 # Disable: no function prototype given
+        /W4
```
wantehchang, line 140: @joedrago FYI

`/Wall` generates a lot of warnings in GoogleTest headers and some warnings in our C++ tests. We could suppress the warnings in GoogleTest headers, and try to fix as many warnings in our C++ tests as we can and suppress the rest. But I decided to use `/W4` instead, which is recommended by this Visual Studio doc page:

https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-170

> **/W4** displays level 1, level 2, and level 3 warnings, and all level 4 (informational) warnings that aren't off by default. We recommend that you use this option to provide lint-like warnings. For a new project, it may be best to use **/W4** in all compilations. This option helps ensure the fewest possible hard-to-find code defects.

PR #1057, opened 2022-08-19
List avifchangesettingtest in alphabetical order

wantehchang:
Also it doesn't need to link with the avif_apps library.


wantehchang:
@tongyuantongyu Please review. Thanks.

tongyuantongyu (approved):
LGTM, thanks.

Should we add an comment above the block, to document the alphabetical order requirement here?

y-guyon (approved):

wantehchang:
> Should we add an comment above the block, to document the alphabetical order requirement here?

Alphabetical order is not a strict requirement, but it is a common convention. Adding new items at the end is also a common convention, so what you did was reasoable. It is best to just see if the items are listed in some apparent order, and if so, try to follow that order.

PR #1058, opened 2022-08-21
Correctly update encoder settings, clear csOptions after avifEncoderAddImage()

tongyuantongyu:
This is a follow up of #1033.

The init logic was skipping some calls if the option is the default value. Now it no longer skip calls so that encoder settings can be reverted to default.

Switch to discard csOptions after being applied to encoder.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1058,8 +1058,8 @@ struct avifCodecSpecificOptions;
 //   image in less bytes. AVIF_SPEED_DEFAULT means "Leave the AV1 codec to its default speed settings"./
 //   If avifEncoder uses rav1e, the speed value is directly passed through (0-10). If libaom is used,
 //   a combination of settings are tweaked to simulate this speed range.
-// * AV1 encoder settings and codec specific options set by avifEncoderSetCodecSpecificOption()
-//   will be applied / updated to AV1 encoder before each call to avifEncoderAddImage().
+// * Some encoder settings can be changed. Changes will take effect from next call to
+//   avifEncoderAddImage().
 typedef struct avifEncoder
 {
     // Defaults to AVIF_CODEC_CHOICE_AUTO: Preference determined by order in availableCodecs table (avif.c)
@@ -1068,15 +1068,15 @@ typedef struct avifEncoder
     // settings (see Notes above)
     int keyframeInterval; // How many frames between automatic forced keyframes; 0 to disable (default).
     uint64_t timescale;   // timescale of the media (Hz)
-    // AV1 encoder settings.
     int maxThreads;
+    int speed;
+    // changeable encoder settings.
     int minQuantizer;
     int maxQuantizer;
     int minQuantizerAlpha;
     int maxQuantizerAlpha;
     int tileRowsLog2;
     int tileColsLog2;
-    int speed;
 
     // stats from the most recent write
     avifIOStats ioStats;
@@ -1129,9 +1129,9 @@ AVIF_API avifResult avifEncoderAddImageGrid(avifEncoder * encoder,
 AVIF_API avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output);
 
 // Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs,
-// to be sent to codec at the next avifEncoderAddImage() call.
-// See the codec documentation to know if a setting is permanent or applied only to the next frame.
-// key must be non-NULL, but passing a NULL value will delete that pending key, if it exists.
+// to be consumed by the codec at the next avifEncoderAddImage() call.
+// See the codec documentation to know if a setting is persistent or applied only to the next frame.
+// key must be non-NULL, but passing a NULL value will delete the pending key, if it exists.
 // Setting an incorrect or unknown option for the current codec will cause errors of type
 // AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION from avifEncoderWrite() or avifEncoderAddImage().
 AVIF_API void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const char * key, const char * value);
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1129,7 +1129,7 @@ AVIF_API avifResult avifEncoderAddImageGrid(avifEncoder * encoder,
 AVIF_API avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output);
 
 // Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs,
-// to be sent to codec once at the next avifEncoderAddImage() call.
+// to be sent to codec at the next avifEncoderAddImage() call.
 // See the codec documentation to know if a setting is permanent or applied only to the next frame.
 // key must be non-NULL, but passing a NULL value will delete that pending key, if it exists.
 // Setting an incorrect or unknown option for the current codec will cause errors of type
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1128,9 +1128,10 @@ AVIF_API avifResult avifEncoderAddImageGrid(avifEncoder * encoder,
                                             avifAddImageFlags addImageFlags);
 AVIF_API avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output);
 
-// Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs.
-// key must be non-NULL, but passing a NULL value will delete that key, if it exists.
-// Settings will be sent to codec once at the next avifEncoderAddImage() call.
+// Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs,
+// to be sent to codec once at the next avifEncoderAddImage() call.
+// See the codec documentation to know if a setting is permanent or applied only to the next frame.
+// key must be non-NULL, but passing a NULL value will delete that pending key, if it exists.
 // Setting an incorrect or unknown option for the current codec will cause errors of type
 // AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION from avifEncoderWrite() or avifEncoderAddImage().
 AVIF_API void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const char * key, const char * value);
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1130,6 +1130,7 @@ AVIF_API avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output
 
 // Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs.
 // key must be non-NULL, but passing a NULL value will delete that key, if it exists.
+// Settings will be sent to codec once at the next avifEncoderAddImage() call.
 // Setting an incorrect or unknown option for the current codec will cause errors of type
 // AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION from avifEncoderWrite() or avifEncoderAddImage().
 AVIF_API void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const char * key, const char * value);
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1058,8 +1058,8 @@ struct avifCodecSpecificOptions;
 //   image in less bytes. AVIF_SPEED_DEFAULT means "Leave the AV1 codec to its default speed settings"./
 //   If avifEncoder uses rav1e, the speed value is directly passed through (0-10). If libaom is used,
 //   a combination of settings are tweaked to simulate this speed range.
-// * Some encoder settings can be changed. Changes will take effect from next call to
-//   avifEncoderAddImage().
+// * Some encoder settings can be changed after encoding starts. Changes will take effect in the next
+//   call to avifEncoderAddImage().
 typedef struct avifEncoder
 {
     // Defaults to AVIF_CODEC_CHOICE_AUTO: Preference determined by order in availableCodecs table (avif.c)
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -244,6 +244,7 @@ typedef struct avifCodecSpecificOption
 } avifCodecSpecificOption;
 AVIF_ARRAY_DECLARE(avifCodecSpecificOptions, avifCodecSpecificOption, entries);
 avifCodecSpecificOptions * avifCodecSpecificOptionsCreate(void);
+void avifCodecSpecificOptionsClear(avifCodecSpecificOptions * csOptions);
 void avifCodecSpecificOptionsDestroy(avifCodecSpecificOptions * csOptions);
 void avifCodecSpecificOptionsSet(avifCodecSpecificOptions * csOptions, const char * key, const char * value); // if(value==NULL), key is deleted
 
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -254,6 +254,18 @@ void avifCodecSpecificOptionsSet(avifCodecSpecificOptions * csOptions, const cha
 struct avifCodec;
 struct avifCodecInternal;
 
+typedef enum avifEncoderConfig
+{
+    AVIF_ENCODER_CONFIG_MIN_QUANTIZER = (1 << 0),
+    AVIF_ENCODER_CONFIG_MAX_QUANTIZER = (1 << 1),
+    AVIF_ENCODER_CONFIG_MIN_QUANTIZER_ALPHA = (1 << 2),
+    AVIF_ENCODER_CONFIG_MAX_QUANTIZER_ALPHA = (1 << 3),
+    AVIF_ENCODER_CONFIG_TILE_ROWS_LOG_2 = (1 << 4),
+    AVIF_ENCODER_CONFIG_TILE_COLS_LOG_2 = (1 << 5),
+
+    AVIF_ENCODER_CONFIG_CODEC_SPECIFIC = (1 << 31)
+} avifEncoderConfig;
+
 typedef avifBool (*avifCodecGetNextImageFunc)(struct avifCodec * codec,
                                               struct avifDecoder * decoder,
                                               const avifDecodeSample * sample,
@@ -268,7 +280,7 @@ typedef avifResult (*avifCodecEncodeImageFunc)(struct avifCodec * codec,
                                                avifEncoder * encoder,
                                                const avifImage * image,
                                                avifBool alpha,
-                                               avifBool updateConfig,
+                                               avifEncoderConfig updateConfig,
                                                avifAddImageFlags addImageFlags,
                                                avifCodecEncodeOutput * output);
 typedef avifBool (*avifCodecEncodeFinishFunc)(struct avifCodec * codec, avifCodecEncodeOutput * output);
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -254,17 +254,18 @@ void avifCodecSpecificOptionsSet(avifCodecSpecificOptions * csOptions, const cha
 struct avifCodec;
 struct avifCodecInternal;
 
-typedef enum avifEncoderConfig
+typedef enum avifEncoderChange
 {
-    AVIF_ENCODER_CONFIG_MIN_QUANTIZER = (1 << 0),
-    AVIF_ENCODER_CONFIG_MAX_QUANTIZER = (1 << 1),
-    AVIF_ENCODER_CONFIG_MIN_QUANTIZER_ALPHA = (1 << 2),
-    AVIF_ENCODER_CONFIG_MAX_QUANTIZER_ALPHA = (1 << 3),
-    AVIF_ENCODER_CONFIG_TILE_ROWS_LOG_2 = (1 << 4),
-    AVIF_ENCODER_CONFIG_TILE_COLS_LOG_2 = (1 << 5),
+    AVIF_ENCODER_CHANGE_MIN_QUANTIZER = (1 << 0),
+    AVIF_ENCODER_CHANGE_MAX_QUANTIZER = (1 << 1),
+    AVIF_ENCODER_CHANGE_MIN_QUANTIZER_ALPHA = (1 << 2),
+    AVIF_ENCODER_CHANGE_MAX_QUANTIZER_ALPHA = (1 << 3),
+    AVIF_ENCODER_CHANGE_TILE_ROWS_LOG2 = (1 << 4),
+    AVIF_ENCODER_CHANGE_TILE_COLS_LOG2 = (1 << 5),
 
-    AVIF_ENCODER_CONFIG_CODEC_SPECIFIC = (1 << 31)
-} avifEncoderConfig;
+    AVIF_ENCODER_CHANGE_CODEC_SPECIFIC = (1 << 31)
+} avifEncoderChange;
+typedef uint32_t avifEncoderChanges;
 
 typedef avifBool (*avifCodecGetNextImageFunc)(struct avifCodec * codec,
                                               struct avifDecoder * decoder,
@@ -280,7 +281,7 @@ typedef avifResult (*avifCodecEncodeImageFunc)(struct avifCodec * codec,
                                                avifEncoder * encoder,
                                                const avifImage * image,
                                                avifBool alpha,
-                                               avifEncoderConfig updateConfig,
+                                               avifEncoderChanges encoderChanges,
                                                avifAddImageFlags addImageFlags,
                                                avifCodecEncodeOutput * output);
 typedef avifBool (*avifCodecEncodeFinishFunc)(struct avifCodec * codec, avifCodecEncodeOutput * output);
--- a/src/avif.c
+++ b/src/avif.c
@@ -880,10 +880,6 @@ error:
 
 void avifCodecSpecificOptionsClear(avifCodecSpecificOptions * csOptions)
 {
-    if (!csOptions) {
-        return;
-    }
-
     for (uint32_t i = 0; i < csOptions->count; ++i) {
         avifCodecSpecificOption * entry = &csOptions->entries[i];
         avifFree(entry->key);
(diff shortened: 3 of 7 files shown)


tongyuantongyu:
@y-guyon , @wantehchang : Please review, thanks.

---

@wantehchang :
> the third option (quoted below) seems the best to me

After rethink I agree this is better.

> I think the original comment at line 863 "Another frame in an image sequence" should be updated to "Another frame in an image sequence, or another layer in a layered image" or simply "Another frame in an image sequence or layered image".

Maybe "the first image" and "subsequent images" is enough?

> the following new comment added at lines 866-870 should be removed

Done. Also removed one of the two conditions that effectively checking the same thing.

> I believe we are dealing with an image sequence or layered image here. So we should allow the frames to have different widths and heights

Done. I wonder if libavif decoder handles frames with different widths and heights correctly.

include/avif/avif.h
```diff
@@ -1130,6 +1130,7 @@ AVIF_API avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output
 
 // Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs.
 // key must be non-NULL, but passing a NULL value will delete that key, if it exists.
+// Settings will be sent to codec once at the next avifEncoderAddImage() call.
```
y-guyon, line 1133: > ```cpp
> // key must be non-NULL, but passing a NULL value will delete that key, if it exists.
> ```

This is now misleading:
- Consider setting option A to value `X` at frame 0; setting A to `NULL` at frame 1 will not discard the setting on the codec side (no-op).

I suggest either:
- Removing the possibility of passing `NULL` to `avifCodecSpecificOptionsSet()`.
- When `NULL` is passed to `avifCodecSpecificOptionsSet()`, instead of removing the setting from the `csOptions`, keep it as `NULL` so it can be "unset" from the codec.

I would lean toward the second solution but I do not know if it can be done.

_Note: By the way, when looking at the implementation of `avifCodecSpecificOptionsSet()` at head, if `NULL` is passed to `avifCodecSpecificOptionsSet()` for a never set option, it will call `avifStrdup(NULL)` and thus `strlen(NULL)`. This is a bug._

- - -

> ```cpp
> // Settings will be sent to codec once at the next avifEncoderAddImage() call.
> ```

It would not hurt to be slightly more precise, what about

```cpp
// See the codec documentation to know if a setting is permanent or applied only to the next frame.
```

y-guyon (commented):
It seems `tests/gtest/avifchangesettingtest.cc` passes as is, even with the `csOptions` behavior change in this PR. Ideally the coverage should be increased to catch future any `csOptions` behavior change. It can be done in another PR.

include/avif/avif.h
tongyuantongyu: > This is now misleading

How about now? I'm feeling it's clear enough, but would like you to check it.

> it can be "unset" from the codec

AOM is not providing unset ability, so probably not easy to implement.

The "Passing NULL to delete" behavior is already documented in comment, so maybe just keep it there?

> This is a bug

Fixed.

src/codec_aom.c
```diff
@@ -518,6 +518,7 @@ static avifBool avifProcessAOMOptionsPostInit(avifCodec * codec, avifBool alpha)
         }
 #endif // defined(HAVE_AOM_CODEC_SET_OPTION)
     }
+    codec->csOptions->count = 0;
```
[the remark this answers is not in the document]
tongyuantongyu: Done.
Moved into `avifEncoderAddImageInternal()` as it's more correct and suitable. Resetting it here will cause `csOptions` to only apply to the first cell in grid, and the resetting behavior is not encoder-specific.

src/codec_aom.c
```diff
@@ -606,26 +607,30 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 avifDiagnosticsPrintf(codec->diag, "aom_codec_enc_config_default() failed: %s", aom_codec_err_to_string(err));
                 return AVIF_RESULT_UNKNOWN_ERROR;
             }
+        } else {
+            cfg->g_usage = aomUsage;
```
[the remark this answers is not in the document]
tongyuantongyu: Done. `encoder->speed` decides `aomUsage` and `aomCpuUsed`, so updating `g_usage` here. In "init" code path this field is set in `aom_codec_enc_config_default()`.

src/codec_aom.c
```diff
@@ -606,26 +607,30 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
-                // hit a certain target bit rate. It's easier to control the worst quality in Q
-                // mode.
-                cfg->rc_end_usage = AOM_Q;
-                break;
-            case AOM_USAGE_REALTIME:
-                // For real-time mode we need to use CBR rate control mode. AOM_Q doesn't fit the
-                // rate control requirements for real-time mode. CBR does.
-                cfg->rc_end_usage = AOM_CBR;
-                break;
+        if (!codec->internal->endUsageSet) {
```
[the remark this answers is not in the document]
tongyuantongyu: This is before the call to `avifProcessAOMOptionsPreInit()`, so user-configured end-usage will override it. (and also set `codec->internal->endUsageSet` so it won't be overwritten)

tongyuantongyu (commented):
> Ideally the coverage should be increased to catch future any csOptions behavior change.

`csOptions` field is internal so we can't directly check it. And given that options that can be set in this way are probably all persistent, it's also hard to tell from the encoded file. Maybe we need a test encoder.

src/codec_aom.c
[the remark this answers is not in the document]
y-guyon: I see. Your message helped me more than the comment though, so maybe consider using the former as the latter. Something like `// aomUsage was taken into account by aom_codec_enc_config_default() but it can only be manually updated afterwards.`

include/avif/avif.h
y-guyon: > How about now? I'm feeling it's clear enough, but would like you to check it.

Good, thanks!

> AOM is not providing unset ability, so probably not easy to implement.

I agree.

> The "Passing NULL to delete" behavior is already documented in comment, so maybe just keep it there?

Adding the word "pending" is enough for me, yes.

src/avif.c
```diff
@@ -889,6 +889,13 @@ void avifCodecSpecificOptionsDestroy(avifCodecSpecificOptions * csOptions)
         avifFree(entry->key);
         avifFree(entry->value);
     }
+
+    csOptions->count = 0;
```
wantehchang, line 893: Move the null pointer protection (lines 883-885) to the `avifCodecSpecificOptionsDestroy()` function.

Otherwise, passing NULL to `avifCodecSpecificOptionsDestroy()` will result in null pointer dereferencing in the `avifArrayDestroy()` call.

`avifCodecSpecificOptionsClear()` can assume its input is not NULL.

src/codec_aom.c
```diff
@@ -744,20 +749,12 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
-        if (encoder->tileColsLog2 != 0) {
-            int tileColsLog2 = AVIF_CLAMP(encoder->tileColsLog2, 0, 6);
-            aom_codec_control(&codec->internal->encoder, AV1E_SET_TILE_COLUMNS, tileColsLog2);
-        }
+        aom_codec_control(&codec->internal->encoder, AV1E_SET_LOSSLESS, lossless);
+        aom_codec_control(&codec->internal->encoder, AV1E_SET_ROW_MT, encoder->maxThreads > 1);
+        int tileRowsLog2 = AVIF_CLAMP(encoder->tileRowsLog2, 0, 6);
+        aom_codec_control(&codec->internal->encoder, AV1E_SET_TILE_ROWS, tileRowsLog2);
+        int tileColsLog2 = AVIF_CLAMP(encoder->tileColsLog2, 0, 6);
+        aom_codec_control(&codec->internal->encoder, AV1E_SET_TILE_COLUMNS, tileColsLog2);
```
wantehchang, line 757: We should make these `aom_codec_control()` calls only when necessary. Is it possible to determine which calls are necessary when `updateConfig` is true?

The reason is that `aom_codec_control()` is not as cheap as it looks, if you trace its code. It will cause libaom to validate all of its configuration settings. See https://crbug.com/aomedia/2732.

src/codec_aom.c
```diff
@@ -783,6 +780,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 return AVIF_RESULT_UNKNOWN_ERROR;
             }
         }
+
+        // Sync our copy with updated config in AOM encoder.
+        *cfg = *codec->internal->encoder.config.enc;
```
wantehchang, line 785: Can this be avoided? We should treat `codec->internal->encoder` as an opaque structure. See these comments in aom/aom_codec.h:

```
/*!\brief Codec context structure
 *
 * All codecs \ref MUST support this context structure fully. In general,
 * this data should be considered private to the codec algorithm, and
 * not be manipulated or examined by the calling application. Applications
 * may reference the 'name' member to get a printable description of the
 * algorithm.
 */
typedef struct aom_codec_ctx {
  const char *name;             /**< Printable interface name */
  aom_codec_iface_t *iface;     /**< Interface pointers */
  aom_codec_err_t err;          /**< Last returned error */
  const char *err_detail;       /**< Detailed info, if available */
  aom_codec_flags_t init_flags; /**< Flags passed at init time */
  union {
    /**< Decoder Configuration Pointer */
    const struct aom_codec_dec_cfg *dec;
    /**< Encoder Configuration Pointer */
    const struct aom_codec_enc_cfg *enc;
    const void *raw;
  } config;               /**< Configuration pointer aliasing union */
  aom_codec_priv_t *priv; /**< Algorithm private storage */
} aom_codec_ctx_t;
```

src/codec_aom.c
tongyuantongyu: These options can also be set using `aom_codec_set_option()`.
If we only sync when these value changes, we can compare with previous value to find out which value needs to be updated; If these values should always be reflected in encoder, then we have to update them each time, as we can't know the real current value.

src/codec_aom.c
tongyuantongyu: `aom_codec_set_option()` can change some options in `cfg`. I didn't find API functions that can get current `cfg` from codec. If we don't update our copy here, these options will only apply to next frame.

include/avif/avif.h
```diff
@@ -1128,8 +1128,10 @@ AVIF_API avifResult avifEncoderAddImageGrid(avifEncoder * encoder,
                                             avifAddImageFlags addImageFlags);
 AVIF_API avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output);
 
-// Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs.
-// key must be non-NULL, but passing a NULL value will delete that key, if it exists.
+// Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs,
+// to be sent to codec at the next avifEncoderAddImage() call.
+// See the codec documentation to know if a setting is permanent or applied only to the next frame.
```
wantehchang, line 1133: Do you know of any codec-specific option that applies only to the next frame? This kind of option seems inconvenient to use.

I think the `flags` parameter of `aom_codec_encode()` is intended for options that apply only to the next frame.

include/avif/avif.h
```diff
@@ -1128,8 +1128,10 @@ AVIF_API avifResult avifEncoderAddImageGrid(avifEncoder * encoder,
                                             avifAddImageFlags addImageFlags);
 AVIF_API avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output);
 
-// Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs.
-// key must be non-NULL, but passing a NULL value will delete that key, if it exists.
+// Codec-specific, optional "advanced" tuning settings, in the form of string key/value pairs,
+// to be sent to codec at the next avifEncoderAddImage() call.
```
wantehchang, line 1132: Nit: add "the" before "codec"

Optional: Perhaps we can say "to be consumed by the codec ..." The word "consume" more strongly implies that the key/value pairs will be cleared after they are sent to the codec.

include/avif/avif.h
```diff
@@ -1128,8 +1128,10 @@ AVIF_API avifResult avifEncoderAddImageGrid(avifEncoder * encoder,
...
+// See the codec documentation to know if a setting is permanent or applied only to the next frame.
+// key must be non-NULL, but passing a NULL value will delete that pending key, if it exists.
```
wantehchang, line 1134: Nit: change "that" to "the"

src/codec_aom.c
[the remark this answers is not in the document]
wantehchang: Changing `cfg->g_usage` like this is risky, because some other fields in `cfg` may be tightly coupled to `cfg->g_usage` and need to be changed accordinly. It is best to avoid this potential issue, by not allowing `encoder->speed` to change.

Similarly, it is best to not allow `encoder->maxThreads` to change.

src/codec_aom.c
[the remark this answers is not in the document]
wantehchang: This block of code should follow the `aom_codec_enc_config_default()` call. The idea of this block of code is to change `cfg->rc_end_usage` from libaom's default to libavif's default.

wantehchang (changes requested):
Hi Yuan,

I suggest some changes. I will need to carefully review the current code in src/codec_aom.c after this pull request is merged.

src/codec_aom.c
```diff
@@ -715,57 +682,95 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 cfg->monochrome = 1;
             }
         }
+    }
+
+    avifBool dimensionChanged = AVIF_FALSE;
+    if ((cfg->g_w != image->width) || (cfg->g_h != image->height)) {
+        cfg->g_w = image->width;
+        cfg->g_h = image->height;
```
tongyuantongyu, line 690: Seems just changing `g_w` and `g_h` here only works when sending smaller frames, due to aom writing first frame's dimension to stream header.

To support larger frames, seems we need user to provide the expect dimension before starting encoding, and set `g_forced_max_frame_width` and `g_forced_max_frame_height`. I'm going to implement this in another PR.

src/codec_aom.c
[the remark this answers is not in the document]
wantehchang: I looked into this. I found that libaom's test/datarate_test.h changes `AOME_SET_CPUUSED` dynamically in the middle of a video sequence (search for `speed_change_test_` in the file), and I also saw this done in webrtc's libaom_av1_encoder.cc file. So I believe libaom supports this. However, libavif may need to change `cfg->g_usage` at the same time it changes `AOME_SET_CPUUSED`, so it is best to avoid this until it's really needed.

src/codec_aom.c
wantehchang: That's correct. An AV1 bitstream begins with a sequence header OBU, which advertises the maximum frame width and maximum frame height. Then each frame has a frame header OBU that may optionally specify a frame width and height smaller than the maximum. So if we are encoding multiple frames with increasing frame widths and heights, we will need a way to specify the maximum frame width and maximum frame height before encoding the first frame.

src/codec_aom.c
```diff
@@ -778,11 +783,39 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
                 return AVIF_RESULT_UNKNOWN_ERROR;
             }
         }
+        codec->internal->encoderInitialized = AVIF_TRUE;
+    } else if (updatedConfig) {
+        if (((!alpha) && ((updatedConfig & AVIF_ENCODER_CONFIG_MIN_QUANTIZER) || (updatedConfig & AVIF_ENCODER_CONFIG_MAX_QUANTIZER))) ||
+            (alpha && ((updatedConfig & AVIF_ENCODER_CONFIG_MIN_QUANTIZER_ALPHA) ||
+                       (updatedConfig & AVIF_ENCODER_CONFIG_MAX_QUANTIZER_ALPHA)))) {
+            aom_codec_control(&codec->internal->encoder, AV1E_SET_LOSSLESS, lossless);
+        }
```
wantehchang, line 812: This conditional expression is very complicated. I suggest we aim for clarify and rewrite it as follows:

```
        if (!alpha) {
            if ((updatedConfig & AVIF_ENCODER_CONFIG_MIN_QUANTIZER) || (updatedConfig & AVIF_ENCODER_CONFIG_MAX_QUANTIZER)) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_LOSSLESS, lossless);
            }
        } else {
            if ((updatedConfig & AVIF_ENCODER_CONFIG_MIN_QUANTIZER_ALPHA) || (updatedConfig & AVIF_ENCODER_CONFIG_MAX_QUANTIZER_ALPHA)) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_LOSSLESS, lossless);
            }
        }
```

Optional: We can simplify the conditionals even further:

```
        if (!alpha) {
            if (updatedConfig & (AVIF_ENCODER_CONFIG_MIN_QUANTIZER | AVIF_ENCODER_CONFIG_MAX_QUANTIZER)) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_LOSSLESS, lossless);
            }
        } else {
            if (updatedConfig & (AVIF_ENCODER_CONFIG_MIN_QUANTIZER_ALPHA | AVIF_ENCODER_CONFIG_MAX_QUANTIZER_ALPHA)) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_LOSSLESS, lossless);
            }
        }
```

src/write.c
```diff
@@ -339,13 +337,12 @@ static void avifBackupSettings(avifEncoder * encoder)
...
     lastEncoder->tileColsLog2 = encoder->tileColsLog2;
     lastEncoder->speed = encoder->speed;
-    encoder->data->csOptionsUpdated = AVIF_FALSE;
 }
 
 // This function detect changes made on avifEncoder.
 // It reports if the change is valid, i.e. if any setting that can't change was changed.
-// It also sets needUpdate to true if valid changes are detected.
-static avifBool avifEncoderSettingsChanged(const avifEncoder * encoder, avifBool * needUpdate)
+// It also reports detected changes in updatedConfig.
```
wantehchang, line 344: This function should sets `*updatedConfig` to 0 on entry. Otherwise, we need to document that it expects the caller to initialize `*updatedConfig` to 0.

src/codec_aom.c
```diff
@@ -715,57 +682,95 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 cfg->monochrome = 1;
             }
         }
+    }
+
+    avifBool dimensionChanged = AVIF_FALSE;
```
wantehchang, line 687: Nit: dimensionChanged => dimensionsChanged

"Dimensions" should be plural.

src/codec_aom.c
```diff
@@ -715,57 +682,95 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+            maxQuantizer = AVIF_CLAMP(encoder->maxQuantizerAlpha, 0, 63);
+        }
+        lossless = ((minQuantizer == AVIF_QUANTIZER_LOSSLESS) && (maxQuantizer == AVIF_QUANTIZER_LOSSLESS));
+        cfg->rc_min_quantizer = minQuantizer;
+        cfg->rc_max_quantizer = maxQuantizer;
 
         if (!avifProcessAOMOptionsPreInit(codec, alpha, cfg)) {
             return AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION;
         }
+    }
```
wantehchang, line 711: Since we do not allow "end-usage" to change, we only need to call `avifProcessAOMOptionsPreInit()` before the `aom_codec_enc_init()` call. Please move lines 708-710 right before line 736, inside the `if (!codec->internal->encoderInitialized)` block.

src/codec_aom.c
```diff
@@ -778,11 +783,39 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             aom_codec_control(&codec->internal->encoder, AOME_SET_CQ_LEVEL, cqLevel);
         }
 #endif
+    }
```
wantehchang, line 786: Lines 774-785 should only be done when initializing the encoder or when `cfg->rc_min_quantizer` or `cfg->rc_max_quantizer` changed. If it takes more than five lines to implement this, just add a TODO comment.

wantehchang (changes requested):
Yuan: Please be careful when making this round of suggested changes. For some reason I find it difficult to review the changes to src/codec_aom.c. So we should refrain from making more big changes and try to stabilize this pull request. Thank you!

include/avif/internal.h
```diff
@@ -253,6 +254,18 @@ void avifCodecSpecificOptionsSet(avifCodecSpecificOptions * csOptions, const cha
...
 struct avifCodecInternal;
 
+typedef enum avifEncoderConfig
+{
+    AVIF_ENCODER_CONFIG_MIN_QUANTIZER = (1 << 0),
+    AVIF_ENCODER_CONFIG_MAX_QUANTIZER = (1 << 1),
+    AVIF_ENCODER_CONFIG_MIN_QUANTIZER_ALPHA = (1 << 2),
+    AVIF_ENCODER_CONFIG_MAX_QUANTIZER_ALPHA = (1 << 3),
+    AVIF_ENCODER_CONFIG_TILE_ROWS_LOG_2 = (1 << 4),
+    AVIF_ENCODER_CONFIG_TILE_COLS_LOG_2 = (1 << 5),
```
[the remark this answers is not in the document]
tongyuantongyu: Done.

include/avif/internal.h
```diff
@@ -253,6 +254,18 @@ void avifCodecSpecificOptionsSet(avifCodecSpecificOptions * csOptions, const cha
...
+
+    AVIF_ENCODER_CONFIG_CODEC_SPECIFIC = (1 << 31)
+} avifEncoderConfig;
```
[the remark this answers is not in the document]
tongyuantongyu: Done

src/codec_aom.c
```diff
@@ -589,45 +593,19 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
-#endif
-        }
+        avifGetPixelFormatInfo(image->yuvFormat, &codec->internal->formatInfo);
 
+        encoderInterface = aom_codec_av1_cx();
+        aom_codec_err_t err = aom_codec_enc_config_default(encoderInterface, cfg, aomUsage);
+        if (err != AOM_CODEC_OK) {
+            avifDiagnosticsPrintf(codec->diag, "aom_codec_enc_config_default() failed: %s", aom_codec_err_to_string(err));
+            return AVIF_RESULT_UNKNOWN_ERROR;
+        }
```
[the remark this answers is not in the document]
tongyuantongyu: Done

src/codec_aom.c
tongyuantongyu: Done

src/codec_aom.c
```diff
@@ -715,57 +682,95 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 cfg->monochrome = 1;
             }
         }
+    }
+
+    avifBool dimensionChanged = AVIF_FALSE;
+    if ((cfg->g_w != image->width) || (cfg->g_h != image->height)) {
+        cfg->g_w = image->width;
+        cfg->g_h = image->height;
+        if (codec->internal->encoderInitialized) {
```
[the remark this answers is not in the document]
tongyuantongyu: Done

src/codec_aom.c
tongyuantongyu: Done

src/codec_aom.c
```diff
@@ -715,57 +682,95 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+                    // rate control requirements for real-time mode. CBR does.
+                    cfg->rc_end_usage = AOM_CBR;
+                    break;
+#if defined(AOM_USAGE_ALL_INTRA)
+                case AOM_USAGE_ALL_INTRA:
+                    cfg->rc_end_usage = AOM_Q;
+                    break;
+#endif
             }
+        }
```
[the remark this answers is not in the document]
tongyuantongyu: Done

src/codec_aom.c
tongyuantongyu: Done

include/avif/avif.h
```diff
@@ -1058,8 +1058,8 @@ struct avifCodecSpecificOptions;
 //   image in less bytes. AVIF_SPEED_DEFAULT means "Leave the AV1 codec to its default speed settings"./
 //   If avifEncoder uses rav1e, the speed value is directly passed through (0-10). If libaom is used,
 //   a combination of settings are tweaked to simulate this speed range.
-// * AV1 encoder settings and codec specific options set by avifEncoderSetCodecSpecificOption()
-//   will be applied / updated to AV1 encoder before each call to avifEncoderAddImage().
+// * Some encoder settings can be changed. Changes will take effect from next call to
```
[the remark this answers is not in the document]
tongyuantongyu: Done

src/codec_aom.c
tongyuantongyu: Done

src/codec_rav1e.c
```diff
@@ -51,11 +53,20 @@ static avifResult rav1eCodecEncodeImage(avifCodec * codec,
                                         avifEncoder * encoder,
                                         const avifImage * image,
                                         avifBool alpha,
-                                        avifBool updateConfig,
+                                        avifEncoderConfig updatedConfig,
                                         uint32_t addImageFlags,
                                         avifCodecEncodeOutput * output)
 {
-    if (updateConfig) {
+    // rav1e does not support changing config.
```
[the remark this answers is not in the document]
tongyuantongyu: It's rav1e does not support it.

rav1e's API design is:
1. Init `RaConfig *` with configs using `rav1e_config_parse_int()`.
2. Get `RaContext *` from `RaConfig *` using `rav1e_context_new()`.
3. Send frame to `RaContext *` using `rav1e_send_frame()`.

So there's no chance to change config during encoding.

---

For SVT, from comment its `svt_av1_enc_set_parameter()` is only expected to call before encoding start, and try to call it after encoding started causes encoder hang at next call to `svt_av1_enc_get_packet()`, so very likely it does not support either.

src/write.c
tongyuantongyu: Done

wantehchang (approved):
LGTM.

PR #1059, opened 2022-08-23
Add avifImageYUVToRGBLibYUV12bpc()

wantehchang:
It was created as a copy of avifImageYUVToRGBLibYUV10bpc(). It currently
uses I012ToARGBMatrix() only.

I012ToARGBMatrix() was added in
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2727775. That
commit is between version 1780 and version 1781. See
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2707003 and
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2746780.

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -240,6 +240,10 @@ static avifResult avifImageYUVToRGBLibYUV10bpc(const avifImage * image,
                                                avifRGBImage * rgb,
                                                const struct YuvConstants * matrixYUV,
                                                const struct YuvConstants * matrixYVU);
+static avifResult avifImageYUVToRGBLibYUV12bpc(const avifImage * image,
+                                               avifRGBImage * rgb,
+                                               const struct YuvConstants * matrixYUV,
+                                               const struct YuvConstants * matrixYVU);
 
 avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
 {
@@ -391,6 +395,10 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
         return avifImageYUVToRGBLibYUV10bpc(image, rgb, matrixYUV, matrixYVU);
     }
 
+    if ((image->depth == 12) && (rgb->depth == 8)) {
+        return avifImageYUVToRGBLibYUV12bpc(image, rgb, matrixYUV, matrixYVU);
+    }
+
     // This function didn't do anything; use the built-in YUV conversion
     return AVIF_RESULT_NOT_IMPLEMENTED;
 }
@@ -411,6 +419,9 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
 #endif
 #if LIBYUV_VERSION < 1756
 #define I400ToARGBMatrix NULL
+#endif
+#if LIBYUV_VERSION < 1781
+#define I012ToARGBMatrix NULL
 #endif
 
 // Lookup table for isYVU. If the entry in this table is AVIF_TRUE, then it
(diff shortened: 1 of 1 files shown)


y-guyon (approved):
Did you witness any significant quality drop (for example in avifrgbtoyuvtest)? I would expect it as it replaces the floating-point, filtered built-in conversion with the libyuv nearest-neighbor one.

src/reformat_libyuv.c
```diff
@@ -663,6 +674,111 @@ avifResult avifImageYUVToRGBLibYUV10bpc(const avifImage * image,
...
+                                        int,
+                                        const uint16_t *,
+                                        int,
+                                        uint8_t *,
+                                        int,
+                                        const struct YuvConstants *,
+                                        int,
+                                        int,
+                                        enum FilterMode);
+    YUVToRGBMatrixFilter lutYuvToRgbMatrixFilter[AVIF_RGB_FORMAT_COUNT][AVIF_PIXEL_FORMAT_COUNT] = {
```
vigneshvg, line 719: given that this is all NULLs. we can omit this for now and add it later when we actually have a non-NULL entry here.

vigneshvg (changes requested):

src/reformat_libyuv.c
```diff
@@ -663,6 +674,111 @@ avifResult avifImageYUVToRGBLibYUV10bpc(const avifImage * image,
     return AVIF_RESULT_NOT_IMPLEMENTED;
 }
 
+avifResult avifImageYUVToRGBLibYUV12bpc(const avifImage * image,
+                                        avifRGBImage * rgb,
+                                        const struct YuvConstants * matrixYUV,
+                                        const struct YuvConstants * matrixYVU)
```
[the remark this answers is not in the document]
wantehchang: Done.

src/reformat_libyuv.c
wantehchang: It was more convenient to also add a bit depth dimension to the lutYuvToRgbMatrixFilter table, which required a two-dimensional subtable of all NULLs.

wantehchang (commented):
Please review. Thanks.

vigneshvg (approved):
lgtm.

couple of things:
1) you may want to update the PR title to something more generic like "Use libyuv for 12bit YUV->RGB when available" or some such since the `avifImageYUVToRGBLibYUV12bpc` function is no longer being added.
2) you added an additional commit with title "implement vignesh's suggestions" or some such. just want to make sure that, that particular commit message will not be part of the actual merge.

wantehchang:
Vignesh,

Thank you for the comment. I am experimenting a new way of responding to review comments, by pushing a new commit. This kind of pull request should be merged with the "Squash and merge" method. That method allows us to edit the commit message.

A downside of this approach is that the code reviewer does not review the final commit message. Here is my proposed commit message when I squash this pull request:

```
Use libyuv for 12-bit YUV->RGB when available

Generalize avifImageYUVToRGBLibYUV10bpc() to
avifImageYUVToRGBLibYUVHighBitDepth() and use it for both 10bpc and
12bpc. The 12bpc part currently uses I012ToARGBMatrix() only.

I012ToARGBMatrix() was added in
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2727775. That
commit is between version 1780 and version 1781. See
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2707003 and
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/2746780.
```

Does this look good?

vigneshvg:
yep the commit message looks good.

wantehchang:
> Did you witness any significant quality drop (for example in avifrgbtoyuvtest)? I would expect it as it replaces the floating-point, filtered built-in conversion with the libyuv nearest-neighbor one.

I am not familiar with the avifrgbtoyuvtest. I just ran it. Here are the "before" and "after" numbers for one particular test. Is this what you expected?

Before:
```
[ RUN      ] All8b/RGBToYUVTest.ConvertWholeRange/164
 RGB 8 bits, YUV 12 bits, RGBA, YUV420, lmtd, MC 6, plain, avg 0, abs avg 0, max 0, PSNR 99dB
[       OK ] All8b/RGBToYUVTest.ConvertWholeRange/164 (0 ms)
```
After:
```
[ RUN      ] All8b/RGBToYUVTest.ConvertWholeRange/164
 RGB 8 bits, YUV 12 bits, RGBA, YUV420, lmtd, MC 6, plain, avg -0.919753, abs avg 0.969136, max 3, PSNR 45.809dB
[       OK ] All8b/RGBToYUVTest.ConvertWholeRange/164 (0 ms)
```

y-guyon:
> I am not familiar with the avifrgbtoyuvtest. I just ran it. Here are the "before" and "after" numbers for one particular test. Is this what you expected?

Yes, thank you.

> ```
> [ RUN      ] All8b/RGBToYUVTest.ConvertWholeRange/164
>  RGB 8 bits, YUV 12 bits, RGBA, YUV420, lmtd, MC 6, plain, avg -0.919753, abs avg 0.969136, max 3, PSNR 45.809dB
> [       OK ] All8b/RGBToYUVTest.ConvertWholeRange/164 (0 ms)
> ```

On my machine I get:

```
[ RUN      ] All8b/RGBToYUVTest.ConvertWholeRange/164
 RGB 8 bits, YUV 12 bits, RGBA, YUV420, lmtd, MC 6, plain, avg -0.966049, abs avg 1.0216, max 4, PSNR 45.1812dB
[       OK ] All8b/RGBToYUVTest.ConvertWholeRange/164 (0 ms)
```

I built libavif 1c33cd7403b9445b2f3692e99fc81a9b967fade9 with libyuv [9b17af9](https://chromium.googlesource.com/libyuv/libyuv/+/9b17af9bef41aec80373a7d4689bb77bac33eab5) and Clang 14.0.6. I guess my x86 architecture does not match yours and libyuv has different behaviors depending on the platform.

I think a loss of ~1 on average is a bit high for 8-bit RGB stored as 12-bit YUV, but I guess it does not matter for limited range (plain color so the subsampling is not taken into account). I am more concerned by full range not being lossless:

```
[ RUN      ] All8b/RGBToYUVTest.ConvertWholeRange/166
 RGB 8 bits, YUV 12 bits, RGBA, YUV420, full, MC 6, plain, avg -0.29321, abs avg 0.524691, max 2, PSNR 50.0348dB
[       OK ] All8b/RGBToYUVTest.ConvertWholeRange/166 (0 ms)
```

wantehchang:
Hi Yannis,

I built the same version of libyuv, but I built it with -DLIBYUV_UNLIMITED_DATA, which produces more accurate results:

```
// This macro controls YUV to RGB using unsigned math to extend range of
// YUV to RGB coefficients to 0 to 4 instead of 0 to 2 for more accuracy on B:
// LIBYUV_UNLIMITED_DATA

// Macros to enable unlimited data for each colorspace
// LIBYUV_UNLIMITED_BT601
// LIBYUV_UNLIMITED_BT709
// LIBYUV_UNLIMITED_BT2020
```

I am sorry I forgot to mention that.

I believe in general libyuv is supposed to produce the same results across CPU architectures.

Once I rebuilt libyuv without -DLIBYUV_UNLIMITED_DATA, I got the same results as yours for both `All8b/RGBToYUVTest.ConvertWholeRange/164` and `All8b/RGBToYUVTest.ConvertWholeRange/166`.

If you think the conversion loss is too high, such as your expectation that 8-bit RGB -> 12-bit YUV -> 8-bit RGB should be a lossless roundtrip, we should ask Frank Barchard if conversion accuracy can be further improved.

PR #1063, opened 2022-08-24
Refactor avifImageRGBToYUVLibYUV8bpc() and avif.h

y-guyon:
Describe RGB-to-YUV before YUV-to-RGB in avif.h
Order description and code by avifRGBFormat, avifPixelFormat, and
avifRange for both RGB-to-YUV and YUV-to-RGB.

Fix libyuv RGB-to-YUV 4:0:0 being used for any avifMatrixCoefficients
and not only for BT.601.

Note: Previous description in avif.h was inaccurate for 10-bit YUV-to-RGB and missing for 12-bit
(10-bit has fewer paths than 8-bit and 12-bit got added in https://github.com/AOMediaCodec/libavif/pull/1059).

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -529,8 +529,7 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
 // routines will fail if the width and height don't match the associated avifImage.
 
 // If libavif is built with a version of libyuv offering a fast conversion between RGB and YUV for
-// the given settings (avifMatrixCoefficients, avifColorPrimaries, avifPixelFormat, avifRGBFormat,
-// avifRange, bit depth), libavif will use it. See reformat_libyuv.c for the details.
+// the given inputs, libavif will use it. See reformat_libyuv.c for the details.
 
 // Note to libavif maintainers: The lookup tables in avifImageYUVToRGBLibYUV
 // rely on the ordering of this enum values for their correctness. So changing
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -528,38 +528,9 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
 // conversion, if necessary. Pixels in an avifRGBImage buffer are always full range, and conversion
 // routines will fail if the width and height don't match the associated avifImage.
 
-// If libavif is built with libyuv fast paths enabled, libavif will use libyuv for conversion from
-// RGB to YUV if the following requirements are met:
-//
-// * YUV depth: 8
-// * RGB depth: 8
-// * rgb.chromaDownsampling: AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC, AVIF_CHROMA_DOWNSAMPLING_FASTEST
-// * CICP is one of the following combinations (CP/TC/MC):
-//   *  x/x/[5|6]
-// * One of the following combinations (avifRGBFormat to avifPixelFormat/Range):
-//   * RGBA     to  [      |      |YUV420|      ]/[Limited|Full]
-//   * ARGB     to  [      |      |YUV420|      ]/[Limited|    ]
-//   * BGR      to  [      |      |YUV420|      ]/[Limited|Full]
-//   * BGRA     to  [YUV444|YUV422|YUV420|YUV400]/[Limited|    ] or
-//                  [      |YUV422|YUV420|YUV400]/[       |Full]
-//   * ABGR     to  [      |      |YUV420|      ]/[Limited|    ]
-
-// If libavif is built with libyuv fast paths enabled, libavif will use libyuv for conversion from
-// YUV to RGB if the following requirements are met:
-//
-// * RGB depth: 8
-// * rgb.chromaUpsampling: AVIF_CHROMA_UPSAMPLING_AUTOMATIC, AVIF_CHROMA_UPSAMPLING_FASTEST
-// * CICP is one of the following combinations (CP/TC/MC/Range):
-//   *      x     /x/[1|2|5|6|9]/[Limited|Full]
-//   * [1|2|5|6|9]/x/    12     /[Limited|Full]
-// * One of the following combinations (avifRGBFormat from avifPixelFormat/depth):
-//   * RGB      from  [      |      |YUV420|      ]/8-bit
-//   * RGBA     from  [YUV444|YUV422|YUV420|YUV400]/8-bit or [YUV444|YUV422|YUV420]/10-bit or [YUV420]/12-bit
-//   * ARGB     from  [      |YUV422|YUV420|      ]/8-bit
-//   * BGR      from  [      |      |YUV420|      ]/8-bit
-//   * BGRA     from  [YUV444|YUV422|YUV420|YUV400]/8-bit or [YUV444|YUV422|YUV420]/10-bit or [YUV420]/12-bit
-//   * ABGR     from  [      |YUV422|YUV420|      ]/8-bit
-//   * RGB_565  from  [      |YUV422|YUV420|      ]/8-bit
+// If libavif is built with a version of libyuv offering a fast conversion between RGB and YUV for
+// the given settings (avifMatrixCoefficients, avifColorPrimaries, avifPixelFormat, avifRGBFormat,
+// avifRange, bit depth), libavif will use it. See reformat_libyuv.c for the details.
 
 // Note to libavif maintainers: The lookup tables in avifImageYUVToRGBLibYUV
 // rely on the ordering of this enum values for their correctness. So changing
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -529,31 +529,37 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
 // routines will fail if the width and height don't match the associated avifImage.
 
 // If libavif is built with libyuv fast paths enabled, libavif will use libyuv for conversion from
-// YUV to RGB if the following requirements are met:
-//
-// * YUV depth: 8 or 10
-// * RGB depth: 8
-// * rgb.chromaUpsampling: AVIF_CHROMA_UPSAMPLING_AUTOMATIC, AVIF_CHROMA_UPSAMPLING_FASTEST
-// * rgb.format: AVIF_RGB_FORMAT_RGBA, AVIF_RGB_FORMAT_BGRA (420/422 support for
-//               AVIF_RGB_FORMAT_ABGR, AVIF_RGB_FORMAT_ARGB, AVIF_RGB_FORMAT_RGB_565)
-// * CICP is one of the following combinations (CP/TC/MC/Range):
-//   * x/x/[2|5|6]/Full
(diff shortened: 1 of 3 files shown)


include/avif/avif.h
```diff
@@ -529,31 +529,37 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
...
+//   *      x     /x/[1|2|5|6|9]/[Limited|Full]
+//   * [1|2|5|6|9]/x/    12     /[Limited|Full]
+// * One of the following combinations (avifRGBFormat from avifPixelFormat/depth):
+//   * RGB      from  [      |      |YUV420|      ]/8-bit
+//   * RGBA     from  [YUV444|YUV422|YUV420|YUV400]/8-bit or [YUV444|YUV422|YUV420]/10-bit or [YUV420]/12-bit
+//   * ARGB     from  [      |YUV422|YUV420|      ]/8-bit
+//   * BGR      from  [      |      |YUV420|      ]/8-bit
+//   * BGRA     from  [YUV444|YUV422|YUV420|YUV400]/8-bit or [YUV444|YUV422|YUV420]/10-bit or [YUV420]/12-bit
+//   * ABGR     from  [      |YUV422|YUV420|      ]/8-bit
+//   * RGB_565  from  [      |YUV422|YUV420|      ]/8-bit
```
wantehchang, line 562: I suggest we omit the detailed documentation and just say "if there is a libyuv function for the conversion".

I really appreciate the work you did to synthesize these tables, but the tables are starting to become too complicated, and it's hard to keep them up to date.

src/reformat_libyuv.c
```diff
@@ -135,101 +135,68 @@ avifResult avifImageRGBToYUVLibYUV8bpc(avifImage * image, const avifRGBImage * r
...
-            }
-        } else if (rgb->format == AVIF_RGB_FORMAT_ARGB) {
-            if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) {
-                if (image->yuvRange == AVIF_RANGE_LIMITED) {
-                    RGBtoYUV = BGRAToI420;
+        if ((image->yuvFormat == AVIF_PIXEL_FORMAT_YUV444) || (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV422) ||
+            (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420)) {
+            // Lookup table for RGB To YUV Matrix (average filter).
+            typedef int (*RGBtoYUV)(const uint8_t *, int, uint8_t *, int, uint8_t *, int, uint8_t *, int, int, int);
+            // Third dimension is for avifRange: 0 is AVIF_RANGE_LIMITED, 1 is AVIF_RANGE_FULL.
```
wantehchang, line 145: Optional: The "0 is AVIF_RANGE_LIMITED, 1 is AVIF_RANGE_FULL" part could be omitted, because they are the numeric values of `AVIF_RANGE_LIMITED` and `AVIF_RANGE_FULL`.

Also at line 174.

src/reformat_libyuv.c
```diff
@@ -135,101 +135,68 @@ avifResult avifImageRGBToYUVLibYUV8bpc(avifImage * image, const avifRGBImage * r
...
-                 image->yuvPlanes[AVIF_CHAN_U],
-                 image->yuvRowBytes[AVIF_CHAN_U],
-                 image->yuvPlanes[AVIF_CHAN_V],
-                 image->yuvRowBytes[AVIF_CHAN_V],
-                 image->width,
-                 image->height)) {
-        return AVIF_RESULT_REFORMAT_FAILED;
-    }
-    return AVIF_RESULT_OK;
+    return AVIF_RESULT_NOT_IMPLEMENTED;
```
wantehchang, line 199: The TODO comment at line 198 probably should be moved to a different place. It looks out of place now.

src/reformat_libyuv.c
```diff
@@ -555,6 +498,31 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
             }
             return AVIF_RESULT_OK;
         }
+    } else if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) {
```
wantehchang, line 501: Since we are checking for specific values of `image->yuvFormat` now, we should add an `else` branch and return `AVIF_RESULT_INVALID_ARGUMENT` or `AVIF_RESULT_REFORMAT_FAILED`.

On the other hand, if we reach this function, we already know `image->yuvFormat` is one of the four YUV formats in AV1 (it's been validated by the `avifPrepareReformatState()` call in `avifImageRGBToYUV()`, so the original `if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) { ... } else { ...}` structure is correct and simpler. We don't need to validate `image->yuvFormat` redundantly here.

y-guyon: I believe following enum order is simpler to maintain, so I just removed `if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400)`.

src/reformat_libyuv.c
y-guyon: It is right after the `if ((image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_BT470BG) || (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_BT601)) {...}` condition. There could be `else if (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {...}` here, so it looks correctly placed to me. Where would you suggest to move it?

src/reformat_libyuv.c
```diff
@@ -255,49 +222,37 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
     // Find the correct libyuv YuvConstants, based on range and CP/MC
     const struct YuvConstants * matrixYUV = NULL;
     const struct YuvConstants * matrixYVU = NULL;
-    if (image->yuvRange == AVIF_RANGE_FULL) {
+    if (image->yuvRange == AVIF_RANGE_LIMITED) {
```
wantehchang, line 225: If I understand it correctly, all the remaining changes are just to test `image->yuvRange` or `image->yuvFormat` in the order of the enum values. I think it's not necessary to make this kind of change.

Also, to separate monochrome from non-monochrome YUV formats, it is cheaper to test `image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400` (the only monochrome format) than testing three formats.

include/avif/avif.h
```diff
@@ -528,32 +528,9 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
...
-// * rgb.chromaDownsampling: AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC, AVIF_CHROMA_DOWNSAMPLING_FASTEST
-// * One of the following combinations (avifRGBFormat to avifPixelFormat/MC/Range):
-//   *  BGRA            to  YUV400        /  x  /[Full|Limited]
-//   *  BGRA            to [YUV420|YUV422]/[5|6]/[Full|Limited]
-//   *  BGRA            to  YUV444        /[5|6]/ Limited
-//   *  BGR             to  YUV420        /[5|6]/[Full|Limited]
-//   * [RGBA|ARGB|ABGR] to  YUV420        /[5|6]/ Limited
+// If libavif is built with a version of libyuv offering a fast conversion between RGB and YUV for
+// the given settings (avifMatrixCoefficients, avifColorPrimaries, avifPixelFormat, avifRGBFormat,
+// avifRange, bit depth), libavif will use it. See reformat_libyuv.c for the details.
```
wantehchang, line 533: I like the new comment. Thanks.

Nit: It doesn't seem necessary to enumerate all the settings we depend on. (It is already quite long but still missing chromeUpsampling/chromaDownsampling/conversion flags.)  It should suffice to just say "for the given inputs".

wantehchang (approved):
LGTM with comments and optional suggested changes. Also please wait for Vignesh's review.

src/reformat_libyuv.c
```diff
@@ -255,49 +221,37 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb)
...
```
vigneshvg, line 224: i think the reason behind the changes in this function is to order the if checks in the enum order?

if so, it could be a separate PR.

src/reformat_libyuv.c
```diff
@@ -135,101 +135,67 @@ avifResult avifImageRGBToYUVLibYUV8bpc(avifImage * image, const avifRGBImage * r
...
+            RGBtoYUV lutRgbToYuv[AVIF_RGB_FORMAT_COUNT][AVIF_PIXEL_FORMAT_COUNT][2] = {
+                { { NULL, NULL }, { NULL, NULL }, { NULL, NULL }, { NULL, NULL }, { NULL, NULL } },                 // RGB
+                { { NULL, NULL }, { NULL, NULL }, { NULL, NULL }, { ABGRToI420, avifABGRToJ420 }, { NULL, NULL } }, // RGBA
+                { { NULL, NULL }, { NULL, NULL }, { NULL, NULL }, { BGRAToI420, NULL }, { NULL, NULL } },           // ARGB
+                { { NULL, NULL }, { NULL, NULL }, { NULL, NULL }, { RGB24ToI420, RGB24ToJ420 }, { NULL, NULL } },   // BGR
+                { { NULL, NULL }, { ARGBToI444, NULL }, { ARGBToI422, ARGBToJ422 }, { ARGBToI420, ARGBToJ420 }, { NULL, NULL } }, // BGRA
+                { { NULL, NULL }, { NULL, NULL }, { NULL, NULL }, { RGBAToI420, NULL }, { NULL, NULL } }, // ABGR
+                { { NULL, NULL }, { NULL, NULL }, { NULL, NULL }, { NULL, NULL }, { NULL, NULL } }        // RGB_565
+            };
+            RGBtoYUV rgbToYuv = lutRgbToYuv[rgb->format][image->yuvFormat][image->yuvRange];
```
vigneshvg, line 179: since we are now using image->yuvRange in the lookup tables (as opposed to if/switch statements), we would not have to validate it in avifPrepareReformatState similar to how we validate yuvFormat and rgb->format [1].

[1] https://github.com/AOMediaCodec/libavif/blob/main/src/reformat.c#L31

vigneshvg (changes requested):

src/reformat_libyuv.c
vigneshvg: *we would now

src/reformat_libyuv.c
wantehchang: In general, handling enum values in the enum order makes it easier to verify all enum values are handled. In this particular case, the order doesn't matter because it's clear that range is either limited or full. This is why I think it's not necessary to reorder this if-else statement.

vigneshvg: yeah i am okay with not re-ordering this as well.

vigneshvg (approved):

vigneshvg:
lgtm. thanks!

wantehchang (approved):
Still LGTM!

PR #1064, opened 2022-08-24
avifdec: Return proper exit code in "info" mode

vigneshvg:
When running avifdec with --info, the exit code must not be 0 when
there is an error in decoding the frames.

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -236,8 +236,7 @@ int main(int argc, char * argv[])
             }
 
             int currIndex = 0;
-            avifResult nextImageResult;
-            while ((nextImageResult = avifDecoderNextImage(decoder)) == AVIF_RESULT_OK) {
+            while ((result = avifDecoderNextImage(decoder)) == AVIF_RESULT_OK) {
                 printf("   * Decoded frame [%d] [pts %2.2f (%" PRIu64 " timescales)] [duration %2.2f (%" PRIu64 " timescales)] [%ux%u]\n",
                        currIndex,
                        decoder->imageTiming.pts,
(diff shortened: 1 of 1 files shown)


apps/avifdec.c
```diff
@@ -248,17 +247,19 @@ int main(int argc, char * argv[])
                        decoder->image->height);
                 ++currIndex;
             }
-            if (nextImageResult != AVIF_RESULT_NO_IMAGES_REMAINING) {
-                printf("ERROR: Failed to decode frame: %s\n", avifResultToString(nextImageResult));
+            if (result != AVIF_RESULT_NO_IMAGES_REMAINING) {
```
y-guyon, line 250: I suggest swapping the condition, I think it is easier to read:

```cpp
            if (result == AVIF_RESULT_NO_IMAGES_REMAINING) {
                result = AVIF_RESULT_OK;
            } else {
                assert(result != AVIF_RESULT_OK);
                printf("ERROR: Failed to decode frame: %s\n", avifResultToString(result));
                avifDumpDiagnostics(&decoder->diag);
            }
```

y-guyon (approved):

apps/avifdec.c
vigneshvg: Done.

apps/avifdec.c
```diff
@@ -248,8 +247,10 @@ int main(int argc, char * argv[])
                        decoder->image->height);
                 ++currIndex;
             }
-            if (nextImageResult != AVIF_RESULT_NO_IMAGES_REMAINING) {
-                printf("ERROR: Failed to decode frame: %s\n", avifResultToString(nextImageResult));
+            if (result == AVIF_RESULT_NO_IMAGES_REMAINING) {
+                result = AVIF_RESULT_OK;
+            } else {
+                printf("ERROR: Failed to decode frame: %s\n", avifResultToString(result));
```
wantehchang, line 253: Optional: The error message at line 257 should say "parse", not "decode", because it reports the error of the `avifDecoderParse(decoder)` call at line 219.

Compare with lines 291-293.

Note: It seems that the error messages for --info should be printed to stderr instead of stdout. We can deal with this issue later.

wantehchang (approved):
LGTM.

apps/avifdec.c
vigneshvg: > Optional: The error message at line 257 should say "parse", not "decode", because it reports the error of the `avifDecoderParse(decoder)` call at line 219.
> 
> Compare with lines 291-293.
> 

Done.

> Note: It seems that the error messages for --info should be printed to stderr instead of stdout. We can deal with this issue later.

Ack. Will do.

vigneshvg: > Note: It seems that the error messages for --info should be printed to stderr instead of stdout. We can deal with this issue later.

There were only 2 occurrences of this. So i have fixed this as well in this PR.

wantehchang (approved):
LGTM.

PR #1065, opened 2022-08-24
Extract Exif to avifImage in avifPNGRead()

y-guyon:
Test it in avifmetadatatest.cc.
Add paris_exif_xmp_icc.jpg and paris_exif_xmp_icc.png. I took and
strongly compressed these pictures. I redacted the GPS coordinates
from the Exif entries.

Code is heavily inspired from https://chromium.googlesource.com/webm/libwebp/+/e8f83de2867456a75b13aa1e876b22e2d1e1c077/imageio/pngdec.c#60.
The license is also BSD.


jzern (commented):
This looks good to me. The only downside I see is that it's always extracted. I don't think there's a way for the user to control that, right?

wantehchang:
1. If the `--exif FILENAME` option is passed to avifenc, we don't need to extract the Exif metadata from the JPEG or PNG source image.

2. The reason we (Google employees) can copy code freely from https://chromium.googlesource.com/webm/libwebp/+/e8f83de2867456a75b13aa1e876b22e2d1e1c077/imageio/pngdec.c#60 is not because its license is also BSD, but rather because Google owns the copyright. The BSD license of that code requires a copy of the code to "retain the above copyright notice, this list of conditions and the following disclaimer."

joedrago:
I like this idea, but we should probably match the expectations that we do with ICC, where it would behave like this PR does unless they explicitly opt out, such as with `--ignore-exif` (matching `--ignore-icc`).

apps/shared/avifpng.c
```diff
@@ -17,6 +18,106 @@ typedef png_bytep png_iccp_datap;
...
+    uint8_t * dst = bytes->data;
+    size_t actualLength = 0;
+    for (; (actualLength < expectedLength) && (*src != '\0'); ++actualLength, ++dst) {
+        while (*src == '\n') {
+            ++src;
+        }
+        const char hexValue[3] = { src[0], src[1], '\0' };
+        src += 2;
+        char * end;
+        bytes->data[actualLength] = (uint8_t)strtol(hexValue, &end, 16);
```
wantehchang, line 38: If `hexValue` is `{ '+', '1', '\0' }` or `{ ' ', '1', '\0' }`, the code will convert it to `1` successfully. But this kind of input is malformed, right?

apps/shared/avifpng.c
```diff
@@ -17,6 +18,106 @@ typedef png_bytep png_iccp_datap;
...
 
+// Converts a hexadecimal string which contains 2-byte character representations of hexadecimal values to raw data.
+// hexString may contain values consisting of [A-F][a-f][0-9] in pairs, e.g., 7af2..., separated by any number of newlines.
+// On success the bytes are filled and AVIF_TRUE is returned.
+// AVIF_FALSE is returned if fewer than expectedLength hexadecimal pairs are converted.
+static avifBool avifHexStringToBytes(const char * hexString, size_t expectedLength, avifRWData * bytes)
+{
+    avifRWDataRealloc(bytes, expectedLength);
+    const char * src = hexString;
+    uint8_t * dst = bytes->data;
```
wantehchang, line 29: Delete `dst` because it is not used. Alternatively replace `bytes->data[actualLength]` with `*dst` at line 38.

apps/shared/avifpng.c
```diff
@@ -17,6 +18,106 @@ typedef png_bytep png_iccp_datap;
...
+    // Skip the profile name and extract the length.
+    for (; (*src != '\0') && (*src != '\n'); ++src, ++numReadCharacters) {
+        if (numReadCharacters == profileLen) {
+            fprintf(stderr, "Exif extraction failed: truncated raw profile of size " AVIF_FMT_ZU "\n", profileLen);
+            return AVIF_FALSE;
+        }
+    }
+    char * end;
+    const long expectedLength = strtol(src, &end, 10);
+    if ((expectedLength <= 0) || ((unsigned long)expectedLength > SIZE_MAX)) {
```
wantehchang, line 79: Should we check `expectedLength` against `profileLen`?

If we are relying on the NUL termination of the `profile` string, then we don't need the `if (numReadCharacters == profileLen)` check at line 72.

apps/shared/avifpng.c
```diff
@@ -17,6 +18,106 @@ typedef png_bytep png_iccp_datap;
...
+    size_t numReadCharacters = 1;
+    // Skip the profile name and extract the length.
+    for (; (*src != '\0') && (*src != '\n'); ++src, ++numReadCharacters) {
+        if (numReadCharacters == profileLen) {
+            fprintf(stderr, "Exif extraction failed: truncated raw profile of size " AVIF_FMT_ZU "\n", profileLen);
+            return AVIF_FALSE;
+        }
+    }
+    char * end;
+    const long expectedLength = strtol(src, &end, 10);
```
wantehchang, line 78: It seems that we should check if `errno` is equal to `ERANGE` after this `strtol()` call.

apps/shared/avifpng.c
```diff
@@ -17,6 +18,106 @@ typedef png_bytep png_iccp_datap;
...
+    }
+    ++end;
+
+    // 'end' now points to the profile payload.
+    return avifHexStringToBytes(end, (size_t)expectedLength, payload);
+}
+
+// Returns AVIF_TRUE if there was no Exif metadata located at info or if the Exif metadata located at info
+// was correctly parsed and imported to avif->exif. Returns AVIF_FALSE in case of error.
+static avifBool avifExtractExif(png_structp png, png_infop const info, avifImage * avif)
```
wantehchang, line 95: Does this support https://ftp-osl.osuosl.org/pub/libpng/documents/pngext-1.5.0.html#C.eXIf ?

wantehchang (changes requested):

apps/shared/avifpng.c
```diff
@@ -17,6 +18,120 @@ typedef png_bytep png_iccp_datap;
 typedef png_charp png_iccp_datap;
 #endif
 
+static avifBool avifIsHexDigit(char hexDigit, uint8_t * decimalValue)
```
jzern, line 21: isxdigit() from ctype.h? though that would mean converting back to strtol

apps/shared/avifpng.c
kmilos: Doesn't look like it indeed.

There are macros to check if the libpng used is recent enough and has a method to support the eXIf chunk.

Or implement from scratch regardless of libpng used.

apps/shared/avifpng.c
```diff
@@ -17,6 +18,120 @@ typedef png_bytep png_iccp_datap;
...
+{
+    if ((hexDigit >= 'A') && (hexDigit <= 'F')) {
+        *decimalValue = (uint8_t)(hexDigit - 'A') + 10;
```
wantehchang, line 24: Nit: The `(uint8_t)` cast needs to be applied to the entire expression, otherwise the result of the cast will be automatically converted to `int` before the addition with 10.

```
        *decimalValue = (uint8_t)(hexDigit - 'A' + 10);
```

Alternatively, cast only if the compiler warns.

y-guyon:
> This looks good to me. The only downside I see is that it's always extracted. I don't think there's a way for the user to control that, right?

Done through the addition of the `ignoreExif` argument.

> 1. If the `--exif FILENAME` option is passed to avifenc, we don't need to extract the Exif metadata from the JPEG or PNG source image.

Done.

> 2. The reason we (Google employees) can copy code freely from https://chromium.googlesource.com/webm/libwebp/+/e8f83de2867456a75b13aa1e876b22e2d1e1c077/imageio/pngdec.c#60 is not because its license is also BSD, but rather because Google owns the copyright. The BSD license of that code requires a copy of the code to "retain the above copyright notice, this list of conditions and the following disclaimer."

Yes. I meant that we do not need to modify the LICENSE file because it is the same boilerplate.

> I like this idea, but we should probably match the expectations that we do with ICC, where it would behave like this PR does unless they explicitly opt out, such as with `--ignore-exif` (matching `--ignore-icc`).

Done.

apps/shared/avifpng.c
y-guyon: Thanks for the links. I did not grasp everything on the eXIf topic. It is another kind of Exif chunk stored in a PNG, right? I assume libwebp does not handle those. I guess libavif should handle both Exif and eXIf chunks. Do you have an example image with an eXIf chunk?

I added a call to `png_get_eXIf_1()` but I did not find any code snippet using it on the web, so I have no idea how correct this is. "eXIf" being equivalent to "Exif" besides the case makes it hard to find documentation on the topic.

> Or implement own method from scratch regardless of libpng used. It is just a literal bytestream dump without any extra header nor coding like it is the case with this custom legacy scheme.

Could you precise what you meant:
- Implement a new way of parsing
  1. eXIf chunks in PNG files
  2. Exif or eXIf chunks in PNG files
- without relying on
  1. a version of libpng recent enough to support `png_get_eXIf_1()`.
  2. libpng at all.
- Any
  1. eXIf chunk
  2. Exif or eXIf chunk
- is just a literal bytestream dump without any extra header nor coding like it is the case with
  1. how it is handled in libwebp currently (`png_get_text()` + comparing `text->key` + raw profile parsing).
  2. how custom legacy Exif metadata chunks are laid out in PNG (raw profiles of the form `"\n<name>\n<length>(%8lu)\n<hex payload>\n"`).

wantehchang:
Note that we can safely use this code in libavif. I am just clarifying which mechanism we are using.

> Yes. I meant that we do not need to modify the LICENSE file because it is the same boilerplate.

The BSD license for https://chromium.googlesource.com/webm/libwebp/+/e8f83de2867456a75b13aa1e876b22e2d1e1c077/imageio/pngdec.c#60 is https://chromium.googlesource.com/webm/libwebp/+/e8f83de2867456a75b13aa1e876b22e2d1e1c077/COPYING. Compared with the BSD license used by libavif, that BSD license has an extra clause:

```
  * Neither the name of Google nor the names of its contributors may
    be used to endorse or promote products derived from this software
    without specific prior written permission.
```

Since you wrote this code as a Google employee, Google owns the copyright. Therefore, a Google employee (e.g., you or me) can contribute this code to libavif as the copyright owner. This is the mechanism we are using. In this mechanism, the license of this code doesn't matter as long as Google owns the copyright.

On the other hand, if someone who is not a Google employee (e.g., Joe Drago or Yuan Tong) wants to copy this code to libavif, they must do so under the terms of that BSD license. I will suggest they put the copied code, with that BSD license, in a separate file. We will also need to add that BSD license to the libavif/LICENSE file.

apps/shared/avifpng.c
```diff
@@ -17,6 +20,119 @@ typedef png_bytep png_iccp_datap;
...
+        if (profile[i] == '\0') {
+            fprintf(stderr, "Exif extraction failed: malformed raw profile, unexpected null-terminating character at " AVIF_FMT_ZU "\n", i);
+            return AVIF_FALSE;
+        } else if (profile[i] == '\n') {
+            if (!lengthStart) {
+                // Skip the name and store the beginning of the string containing the length of the payload.
+                lengthStart = &profile[i + 1];
+            } else {
+                const char * hexPayloadStart = &profile[i + 1];
+                const size_t hexPayloadMaxLength = profileLength - (i + 1);
```
wantehchang, line 74: `hexPayloadMaxLength` is the maximum length of the hexadecimal string. In line 83, we compare it with `expectedLength`, which is the expected number of converted bytes. These two values differ by a factor of 2, because two hexadecimal digits are converted to one byte.

It seems that in line 83 we should compare `(unsigned long)expectedLength > (hexPayloadMaxLength / 2)` instead.

apps/shared/avifpng.c
```diff
@@ -17,6 +20,119 @@ typedef png_bytep png_iccp_datap;
...
+                if (lengthEnd != &profile[i]) {
+                    fprintf(stderr, "Exif extraction failed: malformed raw profile, expected '\\n' but got '\\x%.2X'\n", *lengthEnd);
+                    return AVIF_FALSE;
+                }
+                // No need to check for errno. Just make sure expectedLength is not LONG_MIN and not LONG_MAX.
+                if ((expectedLength <= 0) || (expectedLength == LONG_MAX) || ((unsigned long)expectedLength > hexPayloadMaxLength)) {
+                    fprintf(stderr, "Exif extraction failed: invalid length %ld\n", expectedLength);
+                    return AVIF_FALSE;
+                }
+                return avifHexStringToBytes(hexPayloadStart, (size_t)expectedLength, payload);
```
wantehchang, line 87: Nit: This is just a comment.

`expectedLength` may be much smaller than `profileLength`, so that there are extra hexadecimal digits in the payload after we have `expectedLength` converted bytes. It seems that this kind of payload is also malformed.

I am just pointing this out for your reference, because I don't know if you aim to reject all malformed payloads.

apps/shared/avifpng.c
```diff
@@ -17,6 +20,119 @@ typedef png_bytep png_iccp_datap;
...
+static avifBool avifHexStringToBytes(const char * hexString, size_t expectedLength, avifRWData * bytes)
+{
+    avifRWDataRealloc(bytes, expectedLength);
+    const char * src = hexString;
+    size_t actualLength = 0;
+    for (; (actualLength < expectedLength) && (*src != '\0'); ++actualLength) {
+        while (*src == '\n') {
+            ++src;
+        }
+        if (!isxdigit(src[0]) || !isxdigit(src[1])) {
```
wantehchang, line 36: Nit/optional: An earlier version of this pull request does the conversion manually. We can use a variant of that:

```
// Returns the value of the hexadecimal digit. Returns -1 hexDigit is not a hexadecimal digit.
static int avifHexDigitValue(char hexDigit)
{
    if ((hexDigit >= 'A') && (hexDigit <= 'F')) {
        return hexDigit - 'A' + 10;
    } else if ((hexDigit >= 'a') && (hexDigit <= 'f')) {
        return hexDigit - 'a' + 10;
    } else if ((hexDigit >= '0') && (hexDigit <= '9')) {
        return hexDigit - '0';
    } else {
        return -1;
    }
}
```

wantehchang (approved):
LGTM with suggested changes.

It'd be good to ask Vincent to review the changes to are_images_equal.cc and test_cmd.sh.

apps/shared/avifpng.c
```diff
@@ -17,6 +20,119 @@ typedef png_bytep png_iccp_datap;
...
+        fprintf(stderr, "Exif extraction failed: expected " AVIF_FMT_ZU " tokens but got " AVIF_FMT_ZU "\n", expectedLength, actualLength);
+        return AVIF_FALSE;
+    }
+    return AVIF_TRUE;
+}
+
+// Parses the raw profile string of profileLen characters and extracts the payload.
+static avifBool avifCopyRawProfile(const char * profile, size_t profileLength, avifRWData * payload)
+{
+    // ImageMagick formats 'raw profiles' as "\n<name>\n<length>(%8lu)\n<hex payload>\n".
```
kmilos, line 57: I think ImageMagick also prepends the `Exif\0\0` 6 byte sequence that needs to be discarded after decoding. See again the note for the eXIf chunk specification that does not include this header.

kmilos (changes requested):

apps/shared/avifpng.c
kmilos: Yes, I meant in the case one builds libavif with a version of libpng that does not have support for it. OTOH, it might be valid to say "just use a more recent version of libpng" in order to have this chunk supported....

For eXIf chunk examples you can perhaps take a look at some Exiv2 test vectors.

kmilos: FWIW, Exiv2 will replace any eXIf chunk with a legacy raw one, while exiftool will always do the opposite, i.e. write the more modern eXIf chunk...

apps/shared/avifpng.c
y-guyon: Not sure how this is better than `isxdigit()` + `strtol()`.

vrabaud:
LGTM for are_images_equal.cc and test_cmd.sh. METADATA should indeed be compared.

apps/shared/avifpng.c
kmilos: The best I can offer is the Exif spec itself - the Exif00 header is described as required for the JPEG APP1 payload, but not for the TIFF one. Other image formats were not discussed at the time. Why ImageMagick devs decided to keep it for their custom raw PNG chunk that became so widespread is unknown (perhaps it was just more convenient to convert to/from JPEG). GIMP parser e.g. links to IM code for reference.

The PNG 1.5 extension spec linked earlier also specifically mentions the omission of this header for the more modern, standard eXIf chunk.

y-guyon: I guess this Exif00 header should not be imported into the `avifImage::exif` field then. We just have to be careful when generating Jpeg and PNG images from it, if we implement that one day.

Among the AVIF images with Exif metadata that I crawled, most start with Exif00 but not all.

As a reference, [`heif_enc.cc`](https://github.com/strukturag/libheif/blob/master/examples/heif_enc.cc) does not seem to import any kind of metadata besides the color profile. The HEIF specification ISO-23008 says in section A.2.1 that the Exif blob may or may not start with the TIFF header, leaving the choice to the user of including the Exif00 tag or not.

kmilos: Yes, I think it should be removed too.

Don't know what libheif implemented, but I have recently discussed this over at libjxl as well - they also decided to follow the official HEIF and JPEG XS spec where the Exif payload always starts with just II/MM directly (with the possible 4-byte offset at the start of the box).

apps/shared/avifpng.c
```diff
@@ -17,6 +20,134 @@ typedef png_bytep png_iccp_datap;
...
+{
+    // Preprocess the input hexString by removing a tag commonly added at encoding, if present.
+    // HEIF specification ISO-23008 section A.2.1 allows including and excluding it from AVIF files.
+    // The PNG 1.5 extension mentions the omission of this header for the modern standard eXIf chunk.
+    const char tagExif00[] = "457869660000"; // "Exif\0\0" tag encoded as a hexadecimal string.
+    const size_t tagExif00Len = 6 * 2;
+    if ((expectedLength >= tagExif00Len) && !memcmp(hexString, tagExif00, tagExif00Len)) {
+        hexString += tagExif00Len;
+        expectedLength -= tagExif00Len;
+    }
```
wantehchang, line 37: IMPORTANT: The variables `expectedLength` and `tagExif00Len` have different units. `expectedLength` is in converted bytes, whereas `tagExif00Len` is in hex digits. It seems that this if statement should read:

```
    if ((expectedLength >= tagExif00Len / 2) && !memcmp(hexString, tagExif00, tagExif00Len)) {
        hexString += tagExif00Len;
        expectedLength -= tagExif00Len / 2;
    }
```

Since you have apparently tested this code with Exif metadata containing this tag, I may have misunderstood this code.

Note: The use of `memcmp` (instead of `strcmp`) here requires `hexString` to have at least `tagExif00Len` characters. I verified this condition is ensured by the check `(unsigned long)expectedLength > (hexPayloadMaxLength / 2)` at line 96.

Nit: Should we allow initial `'\n'` characters before "457869660000"?

tests/gtest/aviftest_helpers.h
```diff
@@ -54,13 +54,20 @@ void FillImageGradient(avifImage* image);
 void FillImageChannel(avifRGBImage* image, uint32_t channel_offset,
                       uint32_t value);
 
-// Returns true if both images have the same features and pixel values.
-// If ignore_alpha is true, the alpha channel is not taken into account in the
-// comparison.
+// Returns true if both arrays are empty or have the same length and bytes.
+// data1 or data2 may be null only when data1_length is 0.
```
wantehchang, line 58: I originally suggested this comment right before line 164 in aviftest_helpers.cc, where we know `data1_length` is equal to `data2_length`. If we add the comment to this file, we cannot assume they are equal. So the comment should read something like this:

```
// data1 may be null only when data1_length is 0. data2 may be null only when data2_length is 0.
```

tests/CMakeLists.txt
```diff
@@ -105,6 +100,13 @@ if(AVIF_ENABLE_GTEST)
     target_link_libraries(avify4mtest aviftest_helpers avif_apps ${GTEST_BOTH_LIBRARIES})
     target_include_directories(avify4mtest PRIVATE ${GTEST_INCLUDE_DIRS})
     add_test(NAME avify4mtest COMMAND avify4mtest)
+
+    if(AVIF_BUILD_APPS)
```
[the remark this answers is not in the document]
wantehchang: Could you explain why we need `if(AVIF_BUILD_APPS)`?

In the top-level CMakeLists.txt file, the `avif_apps` library is built under this condition:

```
if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND AVIF_ENABLE_GTEST))
    ...
    add_library(
        avif_apps OBJECT ...
```

`(AVIF_BUILD_TESTS AND AVIF_ENABLE_GTEST)` is true here, so we should not need to add `if(AVIF_BUILD_APPS)`.

wantehchang (changes requested):

apps/shared/avifpng.c
wantehchang: It is more efficient than `isxdigit()` + `strtol()`, especially if we extend the function to process two hex digits. In any case, it's more important to make sure the parsing code is correct and safe.

wantehchang (changes requested):

apps/shared/avifutil.c
```diff
@@ -248,7 +251,8 @@ avifAppFileFormat avifReadImage(const char * filename,
             *outDepth = 8;
         }
     } else if (format == AVIF_APP_FILE_FORMAT_PNG) {
-        if (!avifPNGRead(filename, image, requestedFormat, requestedDepth, outDepth)) {
+        (void)ignoreICC, (void)ignoreXMP; // TODO(yguyon): Implement
```
kmilos, line 254: FYI, recent PNG spec extension 1.6 also added a cICp chunk, so for now you just might want to add it to the feature TODO list here (and also the force/ignore options). There is no libpng API for it yet, so a "user chunk" API can be used. So far, I think only libjxl started outputting PNGs like these when transcoding...

https://w3c.github.io/PNG-spec/

y-guyon: Good to know.

apps/shared/avifpng.c
```diff
@@ -17,6 +20,142 @@ typedef png_bytep png_iccp_datap;
...
+        return AVIF_FALSE;
+    }
+    return AVIF_TRUE;
+}
+
+// Parses the raw profile string of profileLen characters and extracts the payload.
+static avifBool avifCopyRawProfile(const char * profile, size_t profileLength, avifRWData * payload)
+{
+    // ImageMagick formats 'raw profiles' as "\n<name>\n<length>(%8lu)\n<hex payload>\n".
+    if (!profile || (profileLength < 3) || (profile[0] != '\n')) {
```
wantehchang, line 76: Nit: It would be good to document why `profileLength` must be at least 3. I think it is because the string must contain at least three `\n` characters.

Alternatively, change the `profileLength < 3` check to `profileLength == 0`.

apps/shared/avifpng.c
```diff
@@ -17,6 +20,142 @@ typedef png_bytep png_iccp_datap;
...
+        fprintf(stderr, "Exif extraction failed: truncated or malformed raw profile\n");
+        return AVIF_FALSE;
+    }
+
+    const char * lengthStart = NULL;
+    for (size_t i = 1; i < profileLength; ++i) {
```
wantehchang, line 82: Nit/optional: It's easy to miss that `i` starts from 1 (to skip the first `\n`). We could add a comment to point that out

apps/shared/avifpng.c
```diff
@@ -31,11 +170,17 @@ typedef png_charp png_iccp_datap;
...
+                     avifImage * avif,
+                     avifPixelFormat requestedFormat,
+                     uint32_t requestedDepth,
+                     avifBool ignoreExif,
+                     uint32_t * outPNGDepth)
 {
     volatile avifBool readResult = AVIF_FALSE;
     png_structp png = NULL;
     png_infop info = NULL;
+    png_infop info_end = NULL;
```
wantehchang, line 183: Nit: info_end => infoEnd

tests/gtest/avifmetadatatest.cc
```diff
@@ -81,32 +87,148 @@ TEST_P(MetadataTest, EncodeDecode) {
...
+}
+
+//------------------------------------------------------------------------------
 
 }  // namespace
 }  // namespace libavif
+
+int main(int argc, char** argv) {
+  ::testing::InitGoogleTest(&argc, argv);
+  if (argc < 2) {
```
wantehchang, line 226: Nit/Optional: Should we test `argc != 2` here to disallow extraneous arguments?

tests/test_cmd.sh
```diff
@@ -34,16 +34,17 @@ fi
...
+DECODED_FILE=/tmp/avif_test_cmd_decoded.png
+PNG_FILE=/tmp/avif_test_cmd_kodim03.png
 
 # Prepare some extra data.
 set +x
 echo "Generating a color PNG"
-"${AVIFENC}" -s 10 "${TESTDATA_DIR}"/kodim03_yuv420_8bpc.y4m -o "${TMP_ENCODED_FILE}" > /dev/null
-"${AVIFDEC}" "${TMP_ENCODED_FILE}" "${PNG_FILE}" > /dev/null
+"${AVIFENC}" -s 10 "${TESTDATA_DIR}/kodim03_yuv420_8bpc.y4m" -o "${ENCODED_FILE}" > /dev/null
+"${AVIFDEC}" "${ENCODED_FILE}" "${PNG_FILE}"  > /dev/null
```
wantehchang, line 47: Nit: There are two spaces before the `>`. Remove one space.

apps/shared/avifpng.c
```diff
@@ -17,6 +20,142 @@ typedef png_bytep png_iccp_datap;
...
 #endif
 
+// Converts a hexadecimal string which contains 2-byte character representations of hexadecimal values to raw data (bytes).
+// hexString may contain values consisting of [A-F][a-f][0-9] in pairs, e.g., 7af2..., separated by any number of newlines.
+// On success the bytes are filled and AVIF_TRUE is returned.
+// AVIF_FALSE is returned if fewer than expectedLength hexadecimal pairs are converted.
+static avifBool avifHexStringToBytes(const char * hexString, size_t hexStringLength, size_t numExpectedBytes, avifRWData * bytes)
+{
+    // Remove any trailing new line. They are not part of the numExpectedBytes.
+    while ((hexStringLength >= 1) && (hexString[0] == '\n')) {
```
wantehchang, line 30: Nit/optional: Change `hexStringLength >= 1` to `hexStringLength > 0`.

wantehchang (approved):
LGTM with suggested changes.

apps/shared/avifpng.c
```diff
@@ -17,6 +20,142 @@ typedef png_bytep png_iccp_datap;
...
+        ++hexString;
+        --hexStringLength;
+    }
+
+    // Preprocess the input hexString by removing a tag commonly added at encoding, if present.
+    // HEIF specification ISO-23008 section A.2.1 allows including and excluding it from AVIF files.
+    // The PNG 1.5 extension mentions the omission of this header for the modern standard eXIf chunk.
+    const char tagExif00[] = "457869660000"; // "Exif\0\0" tag encoded as a hexadecimal string.
+    const size_t tagExif00Len = 6 * 2;
+    if ((hexStringLength >= tagExif00Len) && !memcmp(hexString, tagExif00, tagExif00Len)) {
```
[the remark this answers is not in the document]
y-guyon: I added a `(hexStringLength < (numExpectedBytes * 2))` check above and replaced the condition by `(numExpectedBytes >= (tagExif00Len / 2))` instead.

I also added a `(numExpectedBytes == 0)` check to discard empty payloads as invalid. It could be part of the condition above because with the current code it can happen only there, but it looks safer and simpler not to nest them.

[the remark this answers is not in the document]
wantehchang: Your solution also works. `hexStringLength >= (numExpectedBytes * 2)` and `numExpectedBytes >= (tagExif00Len / 2)` implies `hexStringLength >= tagExif00Len`, so `hexStringLength -= tagExif00Len;` can be executed safely.

wantehchang (approved):
LGTM.

I may have a couple of minor comments, but I need to look at the code closer and we should merge this pull request first.

apps/shared/avifpng.c
```diff
@@ -17,6 +20,150 @@ typedef png_bytep png_iccp_datap;
...
+                const size_t hexPayloadMaxLength = profileLength - (i + 1);
+                // Parse the length, now that we are sure that it is surrounded by '\n' within the profileLength characters.
+                char * lengthEnd;
+                const long expectedLength = strtol(lengthStart, &lengthEnd, 10);
+                if (lengthEnd != &profile[i]) {
+                    fprintf(stderr, "Exif extraction failed: malformed raw profile, expected '\\n' but got '\\x%.2X'\n", *lengthEnd);
+                    return AVIF_FALSE;
+                }
+                // No need to check for errno. Just make sure expectedLength is not LONG_MIN and not LONG_MAX.
+                if ((expectedLength <= 0) || (expectedLength == LONG_MAX) ||
```
wantehchang, line 111: This comment is just for your consideration. If the `<length>(%8lu)` in the comment at line 83 means the length is at most 8 decimal digits, then we can replace the `expectedLength == LONG_MAX` check here with `expectedLength > 99999999` (eight 9s).

y-guyon: If I had an external source to cite than libwebp's implementation, sure, but it seems simpler to keep the code as is than to explain this `99999999`. In practice ~100MB of Exif metadata is too much anyway, accepting it as input in `avifenc` or not does not really matter.

wantehchang: [This comment is just a discussion, not a request for change.]

Let's not make the change I suggested.

Regarding an external source to cite, the comment at line 83 mentions ImageMagick. I found the ImageMagick code the comment refers to in ImageMagick-7.1.0/coders/png.c. It matches the comment:

```
   sp=profile_data;
   dp=text[0].text;
   *dp++='\n';
   (void) CopyMagickString(dp,(const char *) profile_description,
     allocated_length);
   dp+=description_length;
   *dp++='\n';
   (void) FormatLocaleString(dp,allocated_length-
     (png_size_t) (dp-text[0].text),"%8lu ",(unsigned long) length);
   dp+=8;

   for (i=0; i < (ssize_t) length; i++)
   {
     if (i%36 == 0)
       *dp++='\n';
     *(dp++)=(char) hex[((*sp >> 4) & 0x0f)];
     *(dp++)=(char) hex[((*sp++ ) & 0x0f)];
   }

   *dp++='\n';
   *dp='\0';
```

We can consider adding "(coders/png.c)" to that comment.

Note: In the ImageMagick code, there is a space in the format string `"%8lu "` for the length. That space is immediately overwritten with `'\n'` in the for loop. It should be deleted.

PR #1066, opened 2022-08-26
Add support for sharp yuv in avifImageRGBToYUV.

maryla-uc:
Add --sharpyuv flag to avifenc.

### DIFF
--- a/apps/shared/avifpng.h
+++ b/apps/shared/avifpng.h
@@ -11,7 +11,12 @@ extern "C" {
 #endif
 
 // if (requestedDepth == 0), do best-fit
-avifBool avifPNGRead(const char * inputFilename, avifImage * avif, avifPixelFormat requestedFormat, uint32_t requestedDepth, uint32_t * outPNGDepth);
+avifBool avifPNGRead(const char * inputFilename,
+                     avifImage * avif,
+                     avifPixelFormat requestedFormat,
+                     uint32_t requestedDepth,
+                     uint32_t * outPNGDepth,
+                     avifRGBToYUVFlags flags);
 avifBool avifPNGWrite(const char * outputFilename,
                       const avifImage * avif,
                       uint32_t requestedDepth,
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -586,7 +586,7 @@ AVIF_API void avifRGBImageFreePixels(avifRGBImage * rgb);
 // Conversion options.
 typedef enum avifRGBToYUVFlag
 {
-    AVIF_RGB_TO_YUV_DEFAULT = 0, // Uses the first available upsampling filter among:
+    AVIF_RGB_TO_YUV_DEFAULT = 0, // Uses the first available downsampling filter among:
                                  //   libyuv average, built-in average
 
     // libyuv preference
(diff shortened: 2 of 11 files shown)


include/avif/avif.h
```diff
@@ -586,7 +586,7 @@ AVIF_API void avifRGBImageFreePixels(avifRGBImage * rgb);
 // Conversion options.
 typedef enum avifRGBToYUVFlag
 {
-    AVIF_RGB_TO_YUV_DEFAULT = 0, // Uses the first available upsampling filter among:
+    AVIF_RGB_TO_YUV_DEFAULT = 0, // Uses the first available downsampling filter among:
```
wantehchang, line 589: IMPORTANT: Should we add the sharp yuv filter to the beginning of the list in the next line? I.e., should we prefer sharp yuv by default?

wantehchang (commented):
Maryla: I only reviewed the interface changes (in the header files and avifenc.c).

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -128,15 +129,16 @@ TEST_P(RGBToYUVTest, ConvertWholeRange) {
   const avifPixelFormat yuv_format = std::get<3>(GetParam());
   const avifRange yuv_range = std::get<4>(GetParam());
   const avifMatrixCoefficients matrix_coefficients = std::get<5>(GetParam());
+  const bool sharp_yuv = std::get<6>(GetParam());
```
y-guyon, line 132: Optional: This could be directly `avifRGBToYUVFlags rgb_to_yuv_flags`

y-guyon (approved):

apps/shared/avifpng.h
```diff
@@ -11,7 +11,12 @@ extern "C" {
 #endif
 
 // if (requestedDepth == 0), do best-fit
-avifBool avifPNGRead(const char * inputFilename, avifImage * avif, avifPixelFormat requestedFormat, uint32_t requestedDepth, uint32_t * outPNGDepth);
+avifBool avifPNGRead(const char * inputFilename,
+                     avifImage * avif,
+                     avifPixelFormat requestedFormat,
+                     uint32_t requestedDepth,
+                     uint32_t * outPNGDepth,
+                     avifRGBToYUVFlags flags);
```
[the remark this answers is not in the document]
maryla-uc: Done for avifPNGRead and avifReadImage. I left avifInputReadImage as is since it was less clear where to put the flags and it's a function internal to avifenc.c anyway.

wantehchang (approved):
LGTM with a suggested change.

include/avif/avif.h
maryla-uc: I spoke with Frank and he suggested getting it in first and then we can debate whether it should be on by default.
We have to keep in mind that sharpyuv can make images worse if the decoder uses nearest neighbour. However we can also argue that the main consumer of avif files is Chrome and it does support smooth upsampling.
Performance wise I haven't measured the difference, but on paper it should be a bit slower than normal libyuv conversion but I expect the difference to be negligible compared to actual encoding.

wantehchang (approved):
Still LGTM.

include/avif/avif.h
wantehchang: Thank you for the reply. I agree with your decision.

PR #1067, opened 2022-08-26
In avifenc/dec, treat all arguments that start with '-' as flags.

maryla-uc:
To pass filenames that actually start with '-', users can either
prefix the path with './' or use '--' to signal the end of flags, e.g.
./avifenc --min 0 --max 64 -- --my_input.png output.avif

This creates clearer error messages, e.g.
avifenc input.png output.avif --badflag
would print 'Unrecognized file format: output.avif' but now prints
'ERROR: unrecognized option --badflag' instead.

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -85,7 +85,7 @@ int main(int argc, char * argv[])
             ++argIndex;
             // Parse additional positional arguments if any.
             while (argIndex < argc) {
-                const char * arg = argv[argIndex];
+                arg = argv[argIndex];
                 if (!inputFilename) {
                     inputFilename = arg;
                 } else if (!outputFilename) {
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -79,7 +79,11 @@ int main(int argc, char * argv[])
     while (argIndex < argc) {
         const char * arg = argv[argIndex];
 
-        if (!strcmp(arg, "-h") || !strcmp(arg, "--help")) {
+        if (!strcmp(arg, "--")) {
+            // Strop parsing flags, everything after this is positional arguments
+            ++argIndex;
+            break;
+        } else if (!strcmp(arg, "-h") || !strcmp(arg, "--help")) {
             syntax();
             return 0;
         } else if (!strcmp(arg, "-V") || !strcmp(arg, "--version")) {
@@ -176,6 +180,10 @@ int main(int argc, char * argv[])
                 return 1;
             }
             imageDimensionLimit = (uint32_t)value;
+        } else if (arg[0] == '-') {
+            fprintf(stderr, "ERROR: unrecognized flag %s\n\n", arg);
+            syntax();
+            return 1;
         } else {
             // Positional argument
             if (!inputFilename) {
@@ -183,13 +191,26 @@ int main(int argc, char * argv[])
             } else if (!outputFilename) {
                 outputFilename = arg;
             } else {
-                fprintf(stderr, "Too many positional arguments: %s\n", arg);
+                fprintf(stderr, "Too many positional arguments: %s\n\n", arg);
                 syntax();
                 return 1;
             }
         }
 
         ++argIndex;
+    }
+    // Parse additional positional arguments if any
+    while (argIndex < argc) {
+        const char * arg = argv[argIndex];
+        if (!inputFilename) {
+            inputFilename = arg;
+        } else if (!outputFilename) {
+            outputFilename = arg;
+        } else {
+            fprintf(stderr, "Too many positional arguments: %s\n\n", arg);
+            syntax();
+            return 1;
+        }
     }
 
     if (!inputFilename) {
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -47,6 +47,7 @@ static void syntax(void)
     printf("                        Default: %u, set to a smaller value to further restrict.\n", AVIF_DEFAULT_IMAGE_SIZE_LIMIT);
     printf("  --dimension-limit C : Specifies the image dimension limit (width or height) that should be tolerated.\n");
     printf("                        Default: %u, set to 0 to ignore.\n", AVIF_DEFAULT_IMAGE_DIMENSION_LIMIT);
+    printf("    --                : Signals the end of options. Everything after is interpreted as file names.\n");
     printf("\n");
     avifPrintVersions();
 }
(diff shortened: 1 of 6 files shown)


apps/avifdec.c
```diff
@@ -79,7 +79,11 @@ int main(int argc, char * argv[])
     while (argIndex < argc) {
         const char * arg = argv[argIndex];
 
-        if (!strcmp(arg, "-h") || !strcmp(arg, "--help")) {
+        if (!strcmp(arg, "--")) {
+            // Strop parsing flags, everything after this is positional arguments
```
wantehchang, line 83: Strop => Stop

Since the `syntax()` function refers to these as "options" rather "flags", perhaps this comment should also say "options"?

Also in avifenc.c.

apps/shared/avifutil.c
```diff
@@ -252,7 +252,7 @@ avifAppFileFormat avifReadImage(const char * filename,
             return AVIF_APP_FILE_FORMAT_UNKNOWN;
         }
     } else {
-        fprintf(stderr, "Unrecognized file format: %s\n", filename);
+        fprintf(stderr, "Unrecognized file format for input file: %s\n", filename);
```
wantehchang, line 255: I believe `avifReadImage()` is only used with input files. If we are not sure, we can omit "input" in the error message.

apps/avifdec.c
```diff
@@ -176,21 +180,38 @@ int main(int argc, char * argv[])
...
+        const char * arg = argv[argIndex];
+        if (!inputFilename) {
+            inputFilename = arg;
+        } else if (!outputFilename) {
+            outputFilename = arg;
+        } else {
+            fprintf(stderr, "Too many positional arguments: %s\n\n", arg);
+            syntax();
+            return 1;
+        }
```
wantehchang, line 213: IMPORTANT: Add `++argIndex`.

apps/avifdec.c
```diff
@@ -79,7 +79,11 @@ int main(int argc, char * argv[])
     while (argIndex < argc) {
         const char * arg = argv[argIndex];
 
-        if (!strcmp(arg, "-h") || !strcmp(arg, "--help")) {
+        if (!strcmp(arg, "--")) {
```
wantehchang, line 82: The `syntax()` function should document `--`.

Also in avifenc.c.

Note that we also have man pages for avifdec and avifenc. doc/avifdec.1.md and doc/ avifenc.1.md should document `--`.

apps/avifdec.c
```diff
@@ -176,21 +180,38 @@ int main(int argc, char * argv[])
                 return 1;
             }
             imageDimensionLimit = (uint32_t)value;
+        } else if (arg[0] == '-') {
+            fprintf(stderr, "ERROR: unrecognized flag %s\n\n", arg);
```
wantehchang, line 184: Nit: Say "option" instead of "flag" to match the `syntax()` function?

Also in avifenc.c.

wantehchang (changes requested):
Maryla: Thank you for the pull request. I am not familiar with `--`. If it is a common convention, then we can use it.

jzern:
> Maryla: Thank you for the pull request. I am not familiar with `--`. If it is a common convention, then we can use it.

It's a common gnu convention, we use it in libwebp's tools too.
https://man7.org/linux/man-pages/man3/getopt.3.html
_The special argument "--" forces an end of option-scanning regardless of the scanning mode._

apps/avifdec.c
```diff
@@ -176,21 +180,38 @@ int main(int argc, char * argv[])
...
+        if (!inputFilename) {
+            inputFilename = arg;
+        } else if (!outputFilename) {
+            outputFilename = arg;
+        } else {
+            fprintf(stderr, "Too many positional arguments: %s\n\n", arg);
+            syntax();
+            return 1;
+        }
+    }
```
y-guyon, line 214: Consider moving this loop to line 85, because it is the only place where it is used.

y-guyon (approved):
@vrabaud added [`tests/test_cmd.sh`](https://github.com/AOMediaCodec/libavif/blob/main/tests/test_cmd.sh). Consider testing the apps with `--` (just add it to existing lines or create new calls).

apps/avifdec.c
maryla-uc: Done but note that this is a standard convention and not always documented. E.g. if you look at the man page of "cp" or "mv" they don't talk about -- but do support it.

apps/shared/avifutil.c
maryla-uc: I think it's always an input file in the sense that it's a file we're reading. I thought the error message was clearer this way.

apps/avifdec.c
maryla-uc: Woops, indeed! I thought I'd tested it but not properly, sorry.

apps/avifdec.c
wantehchang: When I wrote this suggestion, I didn't know `--` is a well-known convention. If you think it is not necessary to document this well-known convention, please remove it from the `syntax()` functions and the man pages.

maryla-uc: I think it's fine either way. cwebp does document it. It doesn't hurt to have it in the doc.

apps/avifdec.c
```diff
@@ -79,7 +80,24 @@ int main(int argc, char * argv[])
...
+                    inputFilename = arg;
+                } else if (!outputFilename) {
+                    outputFilename = arg;
+                } else {
+                    fprintf(stderr, "Too many positional arguments: %s\n\n", arg);
+                    syntax();
+                    return 1;
+                }
+                ++argIndex;
+            }
```
wantehchang, line 99: Nit: Add a `break` statement here.

Note: I like the earlier version better because it was easier to understand how `argIndex` was incremented.

apps/avifdec.c
```diff
@@ -47,6 +47,7 @@ static void syntax(void)
     printf("                        Default: %u, set to a smaller value to further restrict.\n", AVIF_DEFAULT_IMAGE_SIZE_LIMIT);
     printf("  --dimension-limit C : Specifies the image dimension limit (width or height) that should be tolerated.\n");
     printf("                        Default: %u, set to 0 to ignore.\n", AVIF_DEFAULT_IMAGE_DIMENSION_LIMIT);
+    printf("    --                : Signals the end of options. Everything after is interpreted as file names.\n");
```
wantehchang, line 50: Optional: add "this" before "is"

If you make this change, please make the same change to the other occurrences. (Search for "Everything after is".)

wantehchang (approved):
LGTM with suggested changes.

wantehchang (approved):
LGTM.

PR #1070, opened 2022-08-27
Clean up encoder settings update during encoding

wantehchang:
Clean up the changes made in the following two pull requests to support
updating encoder settings during encoding:
https://github.com/AOMediaCodec/libavif/pull/1033
https://github.com/AOMediaCodec/libavif/pull/1058

In particular, restore the aomCodecEncodeImage() function in
src/codec_aom.c to its original structure, plus a new block of code to
handle encoder changes.

Rename some functions and data members. Edit some comments and messages.

In the avifEncoderChange enum, left-shift the unsigned int constant 1u
because if we left-shift the signed int constant 1 by 31 bits, it will
be shifted into the sign bit.

Other miscellaneous cosmetic changes.

https://github.com/AOMediaCodec/libavif/pull/761


wantehchang:
@tongyuantongyu Yuan, please review. Thanks.

I reviewed your recent changes carefully and made these adjustments.

include/avif/avif.h
```diff
@@ -1054,11 +1054,11 @@ typedef struct avifEncoder
     avifCodecChoice codecChoice;
 
     // settings (see Notes above)
-    int keyframeInterval; // How many frames between automatic forced keyframes; 0 to disable (default).
-    uint64_t timescale;   // timescale of the media (Hz)
     int maxThreads;
     int speed;
-    // changeable encoder settings.
+    int keyframeInterval; // How many frames between automatic forced keyframes; 0 to disable (default).
+    uint64_t timescale;   // timescale of the media (Hz)
```
wantehchang, line 1060: Commonly used members are listed before rarely used members.

include/avif/internal.h
```diff
@@ -262,14 +262,14 @@ struct avifCodecInternal;
...
 typedef enum avifEncoderChange
 {
-    AVIF_ENCODER_CHANGE_MIN_QUANTIZER = (1 << 0),
-    AVIF_ENCODER_CHANGE_MAX_QUANTIZER = (1 << 1),
-    AVIF_ENCODER_CHANGE_MIN_QUANTIZER_ALPHA = (1 << 2),
-    AVIF_ENCODER_CHANGE_MAX_QUANTIZER_ALPHA = (1 << 3),
-    AVIF_ENCODER_CHANGE_TILE_ROWS_LOG2 = (1 << 4),
-    AVIF_ENCODER_CHANGE_TILE_COLS_LOG2 = (1 << 5),
-
-    AVIF_ENCODER_CHANGE_CODEC_SPECIFIC = (1 << 31)
```
wantehchang, line 272: Google's internal ClangTidy check warns that `1 << 31` shifts the 1 to the sign bit of a signed int.

src/codec_aom.c
```diff
@@ -532,10 +532,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                                       avifCodecEncodeOutput * output)
 {
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
-    aom_codec_iface_t * encoderInterface = NULL;
-    unsigned int aomUsage = AOM_USAGE_GOOD_QUALITY;
-    int aomCpuUsed = -1;
-    avifBool lossless = AVIF_FALSE;
+    avifBool quantizerUpdated = AVIF_FALSE;
```
wantehchang, line 535: I restored this function to its original structure. There is only one `if (!codec->internal->encoderInitialized)` block now (at line 537).

src/codec_aom.c
```diff
@@ -703,103 +712,84 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
-            aom_codec_control(&codec->internal->encoder, AV1E_SET_TILE_COLUMNS, tileColsLog2);
-        }
         if (!codec->internal->tuningSet) {
             if (aom_codec_control(&codec->internal->encoder, AOME_SET_TUNING, AOM_TUNE_SSIM) != AOM_CODEC_OK) {
                 return AVIF_RESULT_UNKNOWN_ERROR;
             }
         }
-        codec->internal->encoderInitialized = AVIF_TRUE;
-    } else if (encoderChanges) {
+    } else {
```
wantehchang, line 761: This `else` block handles the "update config" case after the encoder has been initialized.

Please review this `else` block carefully.

src/codec_aom.c
```diff
@@ -703,103 +712,84 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
-        }
-
         if (aom_codec_enc_init(&codec->internal->encoder, encoderInterface, cfg, encoderFlags) != AOM_CODEC_OK) {
             avifDiagnosticsPrintf(codec->diag,
                                   "aom_codec_enc_init() failed: %s: %s",
                                   aom_codec_error(&codec->internal->encoder),
                                   aom_codec_error_detail(&codec->internal->encoder));
             return AVIF_RESULT_UNKNOWN_ERROR;
         }
+        codec->internal->encoderInitialized = AVIF_TRUE;
```
wantehchang, line 731: Note that we must set `codec->internal->encoderInitialized` to true immediately after a successful `aom_codec_enc_init()` call, because we also check `codec->internal->encoderInitialized` to determine if `aom_codec_destroy(codec->internal->encoder)` should be called:

```
#if defined(AVIF_CODEC_AOM_ENCODE)
    if (codec->internal->encoderInitialized) {
        aom_codec_destroy(&codec->internal->encoder);
    }
#endif
```

Perhaps this should be documented better.

src/write.c
```diff
@@ -294,15 +294,15 @@ avifEncoder * avifEncoderCreate(void)
     avifEncoder * encoder = (avifEncoder *)avifAlloc(sizeof(avifEncoder));
     memset(encoder, 0, sizeof(avifEncoder));
     encoder->maxThreads = 1;
+    encoder->speed = AVIF_SPEED_DEFAULT;
+    encoder->keyframeInterval = 0;
+    encoder->timescale = 1;
```
wantehchang, line 299: Data members are initialized in the order they are declared.

src/write.c
```diff
@@ -320,59 +320,61 @@ void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const char * key,
     avifCodecSpecificOptionsSet(encoder->csOptions, key, value);
 }
 
-static void avifBackupSettings(avifEncoder * encoder)
+static void avifEncoderBackupSettings(avifEncoder * encoder)
```
wantehchang, line 323: This function copies many members, one by one. Should we just copy the entire `avifEncoder` struct?

```
    *lastEncoder = *encoder;
```

src/write.c
```diff
@@ -320,59 +320,61 @@ void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const char * key,
...
-    lastEncoder->speed = encoder->speed;
 }
 
-// This function detect changes made on avifEncoder.
-// It reports if the change is valid, i.e. if any setting that can't change was changed.
-// It also reports detected changes in updatedConfig.
-static avifBool avifEncoderSettingsChanged(const avifEncoder * encoder, avifEncoderChanges * encoderChanges)
+// This function detects changes made on avifEncoder. It returns true on success (i.e., if every
+// change is valid), or false on failure (i.e., if any setting that can't change was changed). It
+// reports detected changes in encoderChanges.
```
wantehchang, line 345: I edited this comment. Please check if the comment is accurate.

src/write.c
```diff
@@ -669,11 +671,11 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
         return AVIF_RESULT_NO_CODEC_AVAILABLE;
     }
 
-    avifEncoderChanges encoderChanges = 0;
-    if (!avifEncoderSettingsChanged(encoder, &encoderChanges)) {
+    avifEncoderChanges encoderChanges;
```
wantehchang, line 674: Note that I removed the initialization of this variable, to test that the `avifEncoderDetectChanges()` call below sets its output parameter properly.

src/write.c
```diff
@@ -876,7 +878,8 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
         // Another frame in an image sequence
 
         const avifImage * imageMetadata = encoder->data->imageMetadata;
-        // If the first image had an alpha plane (even if fully opaque), all subsequent images must have alpha as well.
+        // If the first image in the sequence had an alpha plane (even if fully opaque), all
+        // subsequent images must have alpha as well.
```
wantehchang, line 882: I added back "in the sequence" to this comment. Please let me know if you intended to delete it.

tests/gtest/aviftest_helpers.h
```diff
@@ -66,13 +66,15 @@ struct AvifIOLimitedReader {
   static constexpr uint64_t kNoClamp = std::numeric_limits<uint64_t>::max();
 
   avifIO io;
-  avifIO* underlayIO;
+  avifIO* underlyingIO;
```
wantehchang, line 69: I renamed this member "underlyingIO". I think "underlying" is more correct than "underlay". Please let me know if I misunderstood the meaning.

Other alternatives include "lowerIO", "lowerLayer", "wrappedIO".

tests/gtest/aviftest_helpers.cc
```diff
@@ -214,36 +215,39 @@ bool AreImagesEqual(const avifImage& image1, const avifImage& image2,
   return true;
 }
 
-static avifResult avifIOLimitedReaderRead(struct avifIO* io, uint32_t readFlags,
+static avifResult avifIOLimitedReaderRead(avifIO* io, uint32_t readFlags,
                                           uint64_t offset, size_t size,
                                           avifROData* out) {
   auto reader = reinterpret_cast<AvifIOLimitedReader*>(io);
 
+  if (offset > UINT64_MAX - size) {
```
wantehchang, line 223: This check detects overflow of the addition `offset + size` at line 226.

wantehchang (commented):
Yuan: I annotated some changes in this pull request.

src/write.c
tongyuantongyu: Let's add it back for now. I deleted it to cover layer image case, but now I think we should describe it clearly instead.

tests/gtest/aviftest_helpers.h
tongyuantongyu: Yes this is exactly what I mean. I couldn't remember this very word when writing the code, so used "underlay" instead which feels similar.

src/codec_aom.c
```diff
@@ -703,103 +712,84 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
                 quantizerUpdated = AVIF_TRUE;
-                aom_codec_control(&codec->internal->encoder, AV1E_SET_LOSSLESS, lossless);
+            }
+        }
+        if (quantizerUpdated) {
+            avifBool lossless =
+                ((cfg->rc_min_quantizer == AVIF_QUANTIZER_LOSSLESS) && (cfg->rc_max_quantizer == AVIF_QUANTIZER_LOSSLESS));
+            aom_codec_control(&codec->internal->encoder, AV1E_SET_LOSSLESS, lossless);
+        }
+        if (quantizerUpdated || dimensionsChanged) {
```
tongyuantongyu, line 785: For now using `quantizerUpdated` here is fine, but if more options are added, we may need a dedicated `cfgUpdated` here.

tongyuantongyu:
LGTM. Thanks for the update.

src/write.c
wantehchang: Thank you for the reply. Let's update this comment and the comment at line 878 above when you add support for layered images.

wantehchang (commented):
Yuan: Thank you for the review. I am wondering if you can approve this pull request using the GitHub UI.

If you select the "Files changed" tab when viewing this PR, do you see a green "Review changes" or "Finish your review" drop-down menu on the right?

In that menu, you can check the "Approve" radio button and hit the "Submit review" button.

tongyuantongyu (approved):
> I am wondering if you can approve this pull request using the GitHub UI.

Got it.

y-guyon (approved):

PR #1071, opened 2022-08-27
Run AVIF tests on Windows in GitHub CI

wantehchang:
Invoke test_cmd.sh using bash. This works on Windows if bash is installed.

Pass 2 to the -j option of the ctest command. The nproc command is not
available on Windows, but there is a NUMBER_OF_PROCESSORS environment
variable. We can consider replacing 2 with %NUMBER_OF_PROCESSORS%.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -2,13 +2,13 @@
 #
 #   * The os matrix consists of windows-latest only.
 #   * Installs Visual Studio in the os image.
-#   * Doesn't install cmake 3.13.x, because only recent versions of cmake with
-#     the following bug fix detect Visual Studio compilers correctly:
+#   * Doesn't install cmake 3.17.x, because only cmake 3.18 or later, with the
+#     following bug fix, detects Visual Studio compilers correctly:
 #     https://gitlab.kitware.com/cmake/cmake/-/issues/20585
 #   * Runs ext/*.cmd as Windows batch files rather than using bash. This is
 #     important for ext/svt.cmd, to prevent it from running ext/svt.sh.
+#   * Builds with local libjpeg (-DAVIF_LOCAL_JPEG=ON).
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-windows.yml
```diff
@@ -102,3 +100,6 @@ jobs:
     - name: Build libavif (ninja)
       working-directory: ./build
       run: ninja
+    - name: Run AVIF Tests
+      working-directory: ./build
+      run: ctest -j 2
```
wantehchang, line 105: Please let me know if you prefer a larger value for the `-j` option.

.github/workflows/ci-windows.yml
```diff
@@ -2,13 +2,11 @@
 #
 #   * The os matrix consists of windows-latest only.
 #   * Installs Visual Studio in the os image.
-#   * Doesn't install cmake 3.13.x, because only recent versions of cmake with
-#     the following bug fix detect Visual Studio compilers correctly:
+#   * Doesn't install cmake 3.17.x, because only cmake 3.18 or later, with the
+#     following bug fix, detects Visual Studio compilers correctly:
```
wantehchang, line 6: I updated this comment with the cmake version that `ci.yml` installs now. I also added the first cmake version that has the bug fix we need on Windows.

wantehchang:
I decided to move the macOS part of this pull request to a separate PR: https://github.com/AOMediaCodec/libavif/pull/1072.

Don't review this PR yet. I will request review when this is ready.

wantehchang (commented):
This pull request is ready for review.

.github/workflows/ci-windows.yml
jzern: This is ok, though it's unfortunate we can't pull that from the environment. It would need to be different for all 3 os images in that case.

jzern (approved):

PR #1072, opened 2022-08-28
Use libpng,libjpeg in /usr/local/ in macos-latest

wantehchang:
Without these changes, the avifdec command in tests/test_cmd.sh fails
with the following error in GitHub CI's macos-latest os image:

  libpng warning: Application built with libpng-1.4.12 but running with 1.6.37
  Cannot init libpng (png): /tmp/kodim03.png

The reason is that we compile apps/shared/avifpng.c with the old libpng
1.4.12 headers in /Library/Frameworks/Mono.framework/Headers, and then
run with /usr/local/lib/libpng.dylib, which is the latest libpng 1.6.37.

These changes allow us to run the AVIF tests on macOS in
.github/workflows/ci.yml. Pass 2 instead of $(nproc) to the -j option of
the ctest command because the nproc command is not available on macOS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,10 +59,6 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: bash libyuv.cmd
-    - name: Build libjpeg
-      if: steps.cache-ext.outputs.cache-hit != 'true'
-      working-directory: ./ext
-      run: bash libjpeg.cmd
     - name: Build libsharpyuv
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
@@ -84,14 +80,13 @@ jobs:
         -DAVIF_CODEC_RAV1E=ON -DAVIF_LOCAL_RAV1E=ON
         -DAVIF_CODEC_SVT=ON -DAVIF_LOCAL_SVT=ON
         -DAVIF_CODEC_LIBGAV1=ON -DAVIF_LOCAL_LIBGAV1=ON
-        -DAVIF_LOCAL_LIBYUV=ON -DAVIF_LOCAL_JPEG=ON
+        -DAVIF_LOCAL_LIBYUV=ON
         -DAVIF_LOCAL_LIBSHARPYUV=ON
         -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON
         -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON -DAVIF_LOCAL_GTEST=ON
     - name: Build libavif (ninja)
       working-directory: ./build
       run: ninja
-    - name: Run AVIF Tests (on Linux)
-      if: runner.os == 'Linux'
+    - name: Run AVIF Tests
       working-directory: ./build
-      run: ctest -j $(nproc)
+      run: ctest -j 2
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -480,7 +485,14 @@ if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND AVIF_ENABLE_GTEST))
...
     target_link_libraries(avif_apps avif ${AVIF_PLATFORM_LIBRARIES} ${PNG_LIBRARY} ${ZLIB_LIBRARY} ${JPEG_LIBRARY})
-    target_include_directories(avif_apps PRIVATE ${PNG_PNG_INCLUDE_DIR} ${JPEG_INCLUDE_DIR} INTERFACE apps/shared)
+    # In GitHub CI's macos-latest os image, /usr/local/include has not only the headers of libpng
+    # libjpeg but also the headers of an older version of libavif. Put the avif include directory
+    # before ${PNG_PNG_INCLUDE_DIR} ${JPEG_INCLUDE_DIR} to prevent picking up old libavif headers
+    # from /usr/local/include.
+    target_include_directories(
+        avif_apps PRIVATE $<TARGET_PROPERTY:avif,INTERFACE_INCLUDE_DIRECTORIES> ${PNG_PNG_INCLUDE_DIR} ${JPEG_INCLUDE_DIR}
+        INTERFACE apps/shared
+    )
```
wantehchang, line 495: I found this solution in https://stackoverflow.com/questions/47175683/cmake-target-link-libraries-propagation-of-include-directories.

.github/workflows/ci.yml
```diff
@@ -84,14 +80,13 @@ jobs:
...
         -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON -DAVIF_LOCAL_GTEST=ON
     - name: Build libavif (ninja)
       working-directory: ./build
       run: ninja
-    - name: Run AVIF Tests (on Linux)
-      if: runner.os == 'Linux'
+    - name: Run AVIF Tests
       working-directory: ./build
-      run: ctest -j $(nproc)
+      run: ctest -j 2
```
y-guyon, line 92: Why not using all available cores?

y-guyon (approved):

.github/workflows/ci.yml
wantehchang: I mentioned this in the commit message, but neglected to highlight it here. We can't pass `$(nproc)` to the `-j` option of the `ctest` command because the `nproc` command is not available on macOS.

My experiments showed that `nproc` returns 2 in the virtual machine that runs the ubuntu-latest os image, and the `NUMBER_OF_PROCESSORS` environment variable has the value 2 in the virtual machine that runs the windows-latest os image. This is why I pass 2.

It is possible to work harder to get the number of processor cores on macOS, but I didn't look into that in this pull request because it's more important to start running tests.

y-guyon: Thank you for the explanation.

wantehchang: If we need to use `nproc` on Linux and a different command on macOS, what is the best way to handle this? One idea I have is to have a shell script that invokes a platform-specific command to return the number of processor cores, and then invoke that shell script in place of `$(nproc)`. Do you think this is a good solution?

y-guyon: I do not think this is important enough to spend too much time on it, especially if it is just for going from 2 to 3 threads on MacOS only. Anyway having platform-specific commands seems to be the way to go, either in a shell script as you mentioned, or [directly as a GitHub action](https://github.com/orgs/community/discussions/25057) (maybe this is what you meant?).

PR #1074, opened 2022-08-29
test_cmd.sh: Don't redirect stderr to /dev/null

wantehchang:
In the "Generating a color PNG" step, don't redirect stderr to
/dev/null. This allows us to see the error messages if the avifenc or
avifdec command fails.

### DIFF
--- a/tests/test_cmd.sh
+++ b/tests/test_cmd.sh
@@ -41,8 +41,8 @@ PNG_FILE=/tmp/kodim03.png
 # Prepare some extra data.
 set +x
 echo "Generating a color PNG"
-"${AVIFENC}" -s 10 "${TESTDATA_DIR}"/kodim03_yuv420_8bpc.y4m -o "${TMP_ENCODED_FILE}" &> /dev/null
-"${AVIFDEC}" "${TMP_ENCODED_FILE}" "${PNG_FILE}"  &> /dev/null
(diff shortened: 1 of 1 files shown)


wantehchang:
@vrabaud @y-guyon Please review. Thanks.

I also noticed that tests/test_cmd.sh uses fixed file names in the /tmp directory:

```
TMP_ENCODED_FILE=/tmp/encoded.avif
DECODED_FILE=/tmp/decoded.png
PNG_FILE=/tmp/kodim03.png
```

Is there a common way to avoid this in shell tests?

y-guyon (approved):

y-guyon:
> I also noticed that tests/test_cmd.sh uses fixed file names in the /tmp directory:
> 
> ```
> TMP_ENCODED_FILE=/tmp/encoded.avif
> DECODED_FILE=/tmp/decoded.png
> PNG_FILE=/tmp/kodim03.png
> ```
> 
> Is there a common way to avoid this in shell tests?

Usually I try to prefix the names of temporary files with the test name to avoid any collision between concurrent test runs. This does not prevent collisions between runs of the same test though, so cleaning up afterwards is good practice. We could generate unique file names with a timestamp if the issue occurs.

See the changes I made to this file in https://github.com/AOMediaCodec/libavif/pull/1065/files#diff-e5f614bb84ef790fd29351c81b455c2458008f5b6d0ebe61274b25db773e3da9.

PR #1075, opened 2022-08-29
Use clang-cl to build libyuv on Windows on CI

tongyuantongyu:
libyuv has far more optimized routines written for GCC/Clang than MSVC. Use clang-cl to build libyuv so that these can be used on Windows.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -66,6 +66,9 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: ./libyuv.cmd
+      env:
+        CC: clang-cl
+        CXX: clang-cl
     - name: Build libjpeg
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -66,9 +66,6 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: ./libyuv.cmd
-      env:
-        CC: clang-cl
-        CXX: clang-cl
     - name: Build libjpeg
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,7 +24,9 @@ install:
  - dav1d.cmd
  - libjpeg.cmd
  - zlibpng.cmd
+ - set CC=clang-cl && set CXX=clang-cl
  - libyuv.cmd
+ - set "CC=" && set "CXX="
  - cd ..
  # Configure with CMake
  - mkdir build
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,7 +24,7 @@ install:
  - dav1d.cmd
  - libjpeg.cmd
  - zlibpng.cmd
- - set CC=clang-cl && set CXX=clang-cl
+ - set "CC=clang-cl" && set "CXX=clang-cl"
  - libyuv.cmd
  - set "CC=" && set "CXX="
  - cd ..
--- a/ext/libyuv.cmd
+++ b/ext/libyuv.cmd
@@ -6,6 +6,8 @@
 
 : # If you're running this on Windows, be sure you've already run this (from your VC2019 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
+: # Clang-cl is recommended to build libyuv on Windows. Run this if you have clang-cl installed:
+: #     set CC=clang-cl && set CXX=clang-cl
 
 git clone --single-branch https://chromium.googlesource.com/libyuv/libyuv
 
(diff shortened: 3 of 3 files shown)


.github/workflows/ci-windows.yml
```diff
@@ -66,6 +66,9 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: ./libyuv.cmd
+      env:
+        CC: clang-cl
+        CXX: clang-cl
```
y-guyon, line 71: I believe `.github/workflows/ci-windows.yml` was made to test MSVC compilation success rather than test performance. I recommend reverting this file.

This is a good idea for AppVeyor though, because its purpose is binary distribution for users.

y-guyon (approved):

.github/workflows/ci-windows.yml
tongyuantongyu: Done.

y-guyon:
Let's wait for one successful AppVeyor run then Squash and merge this PR.

ext/libyuv.cmd
```diff
@@ -6,6 +6,8 @@
 
 : # If you're running this on Windows, be sure you've already run this (from your VC2019 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
+: # Clang-cl is recommended to build libyuv on Windows. Run this if you have clang-cl installed:
```
wantehchang, line 9: Nit: Change the first sentence to "It is recommended to build libyuv with clang-cl on Windows." We can also explain the reason is that MSVC doesn't support the assembly code in libyuv, which is written in the GCC inline assembly syntax.

wantehchang (changes requested):
Yuan: This is a good change. I posted some comments in the files.

Ideally we should have a CI in .github/workflows/ that builds everything with clang-cl. But I didn't have time to look into that.

.github/workflows/ci-windows.yml
wantehchang: I just saw this discussion. I think this change is fine because libyuv is less valuable without its assembly code. I wanted to make this kind of change but I didn't have time to look into it.

Yuan: Because of the use of the cache files for the `ext/` directory, you will need to make a dummy change to one of the `ext/*cmd` files in this pull request, or temporarily comment out line 66 `if: steps.cache-ext.outputs.cache-hit != 'true'`.

wantehchang: Just to clarify -- it is okay to deter this change to a separate pull request.

jzern: > Yuan: Because of the use of the cache files for the `ext/` directory, you will need to make a dummy change to one of the `ext/*cmd` files in this pull request, or temporarily comment out line 66 `if: steps.cache-ext.outputs.cache-hit != 'true'`.

I wonder if an alternative in changes like this would be to add a version (v1,v2,etc) to the `key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}` and increase it to force a rerun.

wantehchang: Yuan: I just noticed that this pull request contains a change to `ext/libyuv.cmd`, so no need for a dummy change to one of the `ext/*cmd` files. Sorry about my confusion.

James: I am considering adding a dummy file such as `ext/touch.cmd` or `ext/update-cache.cmd`

ext/libyuv.cmd
tongyuantongyu: Done. The appveyor build has been pending for 10 hours, but since I only updated comment this should success.

appveyor.yml
```diff
@@ -24,7 +24,9 @@ install:
  - dav1d.cmd
  - libjpeg.cmd
  - zlibpng.cmd
+ - set CC=clang-cl && set CXX=clang-cl
  - libyuv.cmd
+ - set "CC=" && set "CXX="
```
wantehchang, line 29: Yuan: Do we need to quote `CC=` and `CXX=` in this command?

You did not quote `CC=clang-cl` and `CXX=clang-cl` in line 27.

ext/libyuv.cmd
```diff
@@ -6,6 +6,9 @@
 
 : # If you're running this on Windows, be sure you've already run this (from your VC2019 install dir):
 : #     "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
+: # It is recommended to build libyuv with clang-cl on Windows, because most libyuv's assembly code is written in GCC
+: # inline assembly syntax which MSVC doesn't support. Run this if you have clang-cl installed:
```
wantehchang, line 10: Nit: add "of" after "most".

Add a comma `,` between "syntax" and "which".

Keep the line length under 100 characters.

```
: # It is recommended to build libyuv with clang-cl on Windows, because most of libyuv's assembly
: # code is written in GCC inline assembly syntax, which MSVC doesn't support. Run this if you have
: # clang-cl installed:
```

wantehchang (changes requested):
Yuan: I have a question (which I somehow neglected to post last time) and a suggested change.

appveyor.yml
wantehchang: Thanks for the explanation. Then please quote the first one so that the value of `CC` doesn't have a trailing space:

```
- set "CC=clang-cl" && set "CXX=clang-cl"
```

Make the same change to the comment in ext/libyuv.cmd. Thanks.

PR #1076, opened 2022-08-30
Pass $(getconf _NPROCESSORS_ONLN) to ctest -j

wantehchang:
Use getconf _NPROCESSORS_ONLN as the portable replacement of nproc on
Linux and macOS.

On Windows use %NUMBER_OF_PROCESSORS%.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -104,4 +104,4 @@ jobs:
       run: ninja
     - name: Run AVIF Tests
       working-directory: ./build
-      run: ctest -j 2
+      run: ctest -j $Env:NUMBER_OF_PROCESSORS
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,4 +89,4 @@ jobs:
       run: ninja
     - name: Run AVIF Tests
       working-directory: ./build
-      run: ctest -j 2
+      run: ctest -j $(getconf _NPROCESSORS_ONLN)


.github/workflows/ci.yml
```diff
@@ -89,4 +89,4 @@ jobs:
       run: ninja
     - name: Run AVIF Tests
       working-directory: ./build
-      run: ctest -j 2
+      run: ctest -j $(getconf _NPROCESSORS_ONLN)
```
wantehchang, line 92: My experiments showed `getconf _NPROCESSORS_ONLN` returns 2 and 3 in the virtual machines running the ubuntu-latest and macos-latest os images, respectively.

`nproc` and `getconf _NPROCESSORS_ONLN` return the same values on my Linux workstation (72) and the virtual machine running ubuntu-latest (2).

On my Mac Pro desktop, `getconf _NPROCESSORS_ONLN` and `sysctl -n hw.logicalcpu` both return 12, whereas `sysctl -n hw.physicalcpu`  returns 6.

.github/workflows/ci-windows.yml
```diff
@@ -104,4 +104,4 @@ jobs:
       run: ninja
     - name: Run AVIF Tests
       working-directory: ./build
-      run: ctest -j 2
+      run: ctest -j $Env:NUMBER_OF_PROCESSORS
```
wantehchang, line 107: My experiments showed that this command is executed in PowerShell, not Command Prompt. So we need to use the PowerShell syntax `$Env:NUMBER_OF_PROCESSORS`, not the Command Prompt syntax `%NUMBER_OF_PROCESSORS%`.

The value of the environment variable `NUMBER_OF_PROCESSORS` is 2 in the virtual machine running the windows-latest os image.

jzern (approved):

PR #1088, opened 2022-09-05
Read XMP from PNG and Exif,XMP from Jpeg

y-guyon:


apps/shared/avifpng.c
```diff
@@ -144,21 +160,20 @@ static avifBool avifExtractExif(png_structp png, png_infop const info, avifImage
...
+    return AVIF_TRUE;
+}
+
+// Same for XMP.
+static avifBool avifExtractXMP(png_structp png, png_infop info, avifImage * avif)
+{
+    png_textp text = NULL;
+    const png_uint_32 numTextChunks = png_get_text(png, info, &text, NULL);
+    for (png_uint_32 i = 0; i < numTextChunks; ++i, ++text) {
+        if (!strcmp(text->key, "Raw profile type xmp")) {
```
kmilos, line 175: There are a couple of other possible locations for XMP data (the most other important one being "XML:com.adobe.xmp"), see (especially the comments):

https://exiftool.org/TagNames/PNG.html

Also interesting there is/was this "Raw profile type APP1" chunk in the wild that could be either Exif or XMP.

Finally, there is also a possible alternative ICC chunk, haven't checked if libavif supports that one id addition to iCCP, or perhaps libpng checks for it as well with their ICC set/get methods?

y-guyon: > There are a couple of other possible locations for XMP data (the most other important one being "XML:com.adobe.xmp"

Done.

> the obsoleted tXMP

How can I read that with libpng? Can `png_get_text()` return a `png_text_struct::key` being "tXMP"? Do you have an example image? I did not find any `tXMP` among ~50k crawled PNG images (using `exiv2 -pS | grep "tXMP"`) so it is probably not worth the hassle.

> Also interesting there is/was this "Raw profile type APP1" chunk in the wild that could be either Exif or XMP.

Done, based on whether it begins with "II" or "MM". Everything else is mapped to XMP as there seems to be many possible XMP first bytes.

> Finally, there is also a possible alternative ICC chunk as text, haven't checked if libavif supports that one in addition to the dedicated iCCP, or perhaps libpng already checks for it as well with their ICC set/get methods?

Do you have an example image?

kmilos: > How can I read that with libpng? Can `png_get_text()` return a `png_text_struct::key` being "tXMP"? Do you have an example image? I did not find any `tXMP` among ~50k crawled PNG images (using `exiv2 -pS | grep "tXMP"`) so it is probably not worth the hassle.

[I guess only w/ the "unknown" chunk API](https://refspecs.linuxbase.org/LSB_4.1.0/LSB-Desktop-generic/LSB-Desktop-generic/libpng-png-get-unknown-chunks-1.html). But I agree, probably not worth the hassle if it really was pulled from the spec 20 years ago, and indeed, there are no examples around.

Same for the ICC alternative - as there are probably no test images in the exiv2 database and the iCCP chunk has been around for a while, probably not worth the hassle either until someone reports it.

apps/shared/avifpng.c
```diff
@@ -104,61 +108,132 @@ static avifBool avifCopyRawProfile(const char * profile, size_t profileLength, a
...
+                return AVIF_FALSE;
+            }
+            avifRWDataFree(&avif->exif);
+            avif->exif = metadata;
+            *ignoreExif = AVIF_TRUE; // Ignore any other Exif chunk.
+        } else if (!*ignoreXMP && !strcmp(text->key, "Raw profile type xmp")) {
+            if (!avifCopyRawProfile(text->text, avifGetPngTextLength(text), &metadata, &tagExif00WasRemoved)) {
+                return AVIF_FALSE;
+            }
+            if (tagExif00WasRemoved) {
```
kmilos, line 184: Don't think this'll ever happen, but why not. Would've been better to check if there is a valid XML packet somehow (like checking there is valid Exif above), but probably an overkill...

apps/shared/avifpng.c
```diff
@@ -104,61 +108,132 @@ static avifBool avifCopyRawProfile(const char * profile, size_t profileLength, a
...
+                    // Ignore invalid Exif payloads too.
+                    avifRWDataFree(&metadata);
+                } else {
+                    avifRWDataFree(&metadata);
+                    fprintf(stderr, "Exif extraction failed: missing TIFF header\n");
+                    return AVIF_FALSE;
+                }
+            } else {
+                // For simplicity, consider everything else as XMP.
+                // In theory, it could be an ICC stored as text but iCCP has been around for a while,
```
kmilos, line 219: Don't think it applies here - we're inside the "Raw profile type APP1" case which was only ever used for Exif or XMP it would seem. I'd move this ICC comment to the end of this big decision tree and mention no support for "Raw profile type icc" or "Raw profile type icm" keys.

apps/shared/avifpng.c
```diff
@@ -104,61 +108,132 @@ static avifBool avifCopyRawProfile(const char * profile, size_t profileLength, a
...
+            if (metadata.size > tiffHeaderSize &&
+                (avifExtractExifTiffHeaderOffset(metadata.data, tiffHeaderSize, &exifTiffHeaderOffset) == AVIF_RESULT_OK)) {
+                if (*ignoreExif) {
+                    avifRWDataFree(&metadata);
+                } else {
+                    avifRWDataFree(&avif->exif);
+                    avif->exif = metadata;
+                    *ignoreExif = AVIF_TRUE; // Ignore any other Exif chunk.
+                }
+            } else if (tagExif00WasRemoved) {
```
kmilos, line 208: Again, not quite sure why the extra check if a valid Exif is already found above...

apps/shared/avifpng.c
y-guyon: Is there an official list of valid XMP headers? From crawled JPEGs I saw mostly `<?xpacket` but I found a few `<x:xmpmeta`, `?xpacket`, `http://ns.adobe.com/xap/1.0/`, `-<rdf:RDF`, etc.

Also in libwebp the spirit was: do not care about metadata content, just pass it around. I do not know to what extend we want to apply that in libavif.

apps/shared/avifpng.c
y-guyon: Sorry I mixed them up. I moved the comment near `png_get_iCCP()`.

apps/shared/avifpng.c
y-guyon: We check `tagExif00WasRemoved` when it is *not* a valid Exif. I wanted to:
- avoid outputting an XMP chunk with the heading bytes `Exif\0\0` silently removed,
- call `avifCopyRawProfile()` only once,
- avoid reallocating the `avifRWData metadata` just to add/remove the heading bytes `Exif\0\0`,
- keep the code somewhat short and easy to read.

Feel free to suggest alternatives or to discuss whether these constraints are less important than simplifying the code.

apps/shared/avifpng.c
kmilos: IMHO anything that starts w/ a "<" should be a bare (valid) XML load with no header that should be kept (i.e. as coming from TIFFs, Jpeg2000, compliant PNGs etc.). The only JPEG APP1 header I'm aware of is the `http://ns.adobe.com/xap/1.0/\0` - see the [XMP spec for JPEG](https://github.com/adobe/xmp-docs/blob/master/XMPSpecifications/XMPSpecificationPart3.pdf); [ImageMagick also implements this for JPEG](https://github.com/ImageMagick/ImageMagick/blob/80bd592dd20d4a1287842459ca3801c3ae3866cd/coders/jpeg.c#L185), and I think it is safe to assume this in the PNG case for "Raw profile type APP1" (and most probably also for "Raw profile type xml"), as, like "Raw profile type exif", these seem to be verbatim copies of the corresponding JPEG APP1 payloads before the eXIF chunk and "XML:com.adobe.xmp" keys were specified for the headerless payloads.

Caveat: note that there is also the "ExtendedXMP" JPEG APP1 header `http://ns.adobe.com/xmp/extension/\0` that is used for concatenated packets to work around the 64kB JPEG APP1 limit, so supporting fusion of multiple APP1 segments when reading JPEGs might be required as well (already done by e.g. exiv2 and XMPSdk). Be aware that there is also a possibility of concatenated Exif APP1 segments in JPEG as well.

apps/shared/avifpng.c
kmilos: I don't feel strongly about this, it was just a bit difficult to follow - as discussed above there really should be little difference in Exif or XMP treatment (you either have some APP1 header or you don't in either case) so this special flag just in the Exif case was throwing me off...

apps/shared/avifpng.c
```diff
@@ -162,75 +138,63 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
         } else if (!strcmp(text->key, "Raw profile type APP1")) {
-            if (!avifCopyRawProfile(text->text, avifGetPngTextLength(text), &metadata, &tagExif00WasRemoved)) {
+            // This can be either Exif, XMP or something else.
+            avifRWData metadata = { NULL, 0 };
+            if (!avifCopyRawProfile(text->text, textLength, &metadata)) {
                 return AVIF_FALSE;
             }
-            // Differentiate Exif and XMP by checking whether the chunk begins with a valid TIFF header ("IIx0x42" or "MMx42x0").
-            const uint32_t tiffHeaderSize = 4;
+            const uint32_t exifTiffHeaderSize = 10; // "MM\0\42" or "II\42\0", maybe prefixed by "Exif\0\0"
```
kmilos, line 165: In the raw APP1 case, this MUST be prefixed by `Exif\0\0`, so `exifTiffHeaderSize` is no longer appropriate here (and there's no point in checking for II or MM, you can assume it's a valid payload if the header is matched). Maybe just call this `const exifApp1HeaderSize = 6`?

apps/shared/avifpng.c
```diff
@@ -162,75 +138,63 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
             uint32_t exifTiffHeaderOffset;
-            if (metadata.size > tiffHeaderSize &&
-                (avifExtractExifTiffHeaderOffset(metadata.data, tiffHeaderSize, &exifTiffHeaderOffset) == AVIF_RESULT_OK)) {
+            const avifROData xmpJpegHeader = { (const uint8_t *)"http://ns.adobe.com/xap/1.0/\0", 29 };
```
kmilos, line 167: Similarly rename to `xmpApp1Header` to match?

apps/shared/avifpng.c
```diff
@@ -162,75 +138,63 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
-                    // Ignore invalid Exif payloads too.
-                    avifRWDataFree(&metadata);
-                } else {
-                    avifRWDataFree(&metadata);
-                    fprintf(stderr, "Exif extraction failed: missing TIFF header\n");
-                    return AVIF_FALSE;
-                }
-            } else {
-                // For simplicity, consider everything else as XMP.
+            } else if ((metadata.data[0] == '<') ||
```
kmilos, line 179: Don't think there is point in checking for "<" - in the APP1 case the payload must start w/ the header above.

apps/shared/avifpng.c
```diff
@@ -162,75 +138,63 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
+                       ((metadata.size > xmpJpegHeader.size) && !memcmp(metadata.data, xmpJpegHeader.data, xmpJpegHeader.size))) {
+                // metadata has an XMP header.
                 if (*ignoreXMP) {
                     avifRWDataFree(&metadata);
                 } else {
                     avifRWDataFree(&avif->xmp);
                     avif->xmp = metadata;
-                    *ignoreXMP = AVIF_TRUE; // Ignore any other XMP chunk.
+                    *ignoreXMP = AVIF_TRUE; // Ignore any other Exif chunk.
```
kmilos, line 187: Keep the old comment?

apps/shared/avifpng.c
```diff
@@ -119,46 +95,107 @@ static avifBool avifCopyRawProfile(const char * profile, size_t profileLength, a
...
+                    *ignoreExif = AVIF_TRUE; // Ignore any other Exif chunk.
+                }
+            } else if ((metadata.data[0] == '<') ||
+                       ((metadata.size > xmpJpegHeader.size) && !memcmp(metadata.data, xmpJpegHeader.data, xmpJpegHeader.size))) {
+                // metadata has an XMP header.
+                if (*ignoreXMP) {
+                    avifRWDataFree(&metadata);
+                } else {
+                    avifRWDataFree(&avif->xmp);
+                    avif->xmp = metadata;
```
kmilos, line 186: Need to skip the XMP APP1 header here.

apps/shared/avifpng.c
```diff
@@ -119,46 +95,107 @@ static avifBool avifCopyRawProfile(const char * profile, size_t profileLength, a
...
-                    break;
+
+        if (!*ignoreExif && !strcmp(text->key, "Raw profile type exif")) {
+            if (!avifCopyRawProfile(text->text, textLength, &avif->exif)) {
+                return AVIF_FALSE;
+            }
+            avifRemoveHeadingExif00TagIfAny(&avif->exif);
+            *ignoreExif = AVIF_TRUE; // Ignore any other Exif chunk.
+        } else if (!*ignoreXMP && !strcmp(text->key, "Raw profile type xmp")) {
+            if (!avifCopyRawProfile(text->text, textLength, &avif->xmp)) {
```
kmilos, line 155: Don't see where the potential `http://ns.adobe.com/xap/1.0/\0` header is handled in this case?

apps/shared/avifpng.c
kmilos: > I now realize that it is the same behavior between PNG raw profiles and Jpeg APP1, for the header part.

Great! I think for "Raw profile type APP1" it must be the same (what other choice is there?); for "Raw profile type exif" and "Raw profile type xmp" it should be the same (as e.g. implemented by ImageMagick), but again, as there is no official spec, there are no guarantees so better check for headers just in case, but one should also allow the case when they're not there as the key kinda identifies the data...

kmilos (commented):
Apart from the minor last nit, looking good now!

apps/shared/avifpng.c
```diff
@@ -119,46 +95,95 @@ static avifBool avifCopyRawProfile(const char * profile, size_t profileLength, a
...
-                case PNG_TEXT_COMPRESSION_zTXt:
-                default:
-                    textLength = text->text_length;
-                    break;
+
+        if (!*ignoreExif && !strcmp(text->key, "Raw profile type exif")) {
+            if (!avifCopyRawProfile(text->text, textLength, &avif->exif)) {
+                return AVIF_FALSE;
+            }
+            avifRemoveHeader(&exifApp1Header, &avif->exif); // Optional.
```
wantehchang, line 155: Nit: It's not clear what the comment "Optional" means. I think it means we don't need to check the return value of `avifRemoveHeader()` because the header is optional. If so, it would be good to say that, e.g., "Ignore the return value because the header is optional."

apps/shared/avifpng.c
```diff
@@ -119,46 +95,95 @@ static avifBool avifCopyRawProfile(const char * profile, size_t profileLength, a
...
+            *ignoreExif = AVIF_TRUE;                        // Ignore any other Exif chunk.
+        } else if (!*ignoreXMP && !strcmp(text->key, "Raw profile type xmp")) {
+            if (!avifCopyRawProfile(text->text, textLength, &avif->xmp)) {
+                return AVIF_FALSE;
+            }
+            avifRemoveHeader(&xmpApp1Header, &avif->xmp); // Optional.
```
wantehchang, line 161: Nit: Same here. It would be good to clarify ths "Optional" comment.

tests/gtest/avifmetadatatest.cc
```diff
@@ -155,40 +155,36 @@ TEST_P(MetadataTest, Read) {
 
 INSTANTIATE_TEST_SUITE_P(
     PngNone, MetadataTest,
-    Combine(Values("paris_exif_xmp_icc.png"),  // zTXt iCCP iTXt IDAT
+    Combine(Values("paris_icc_exif_xmp.png"),  // iCCP zTXt zTXt IDAT
```
wantehchang, line 158: Optional: It would be good to add a README file to tests/data/ to describe the properties of the test files that we depend on. For example, we can document the metadata that paris_icc_exif_xmp.png has and the formats/chunks the metadata are stored in.

apps/shared/avifjpeg.c
```diff
@@ -320,13 +334,49 @@ avifBool avifJPEGRead(const char * inputFilename, avifImage * avif, avifPixelFor
...
+        }
+    }
+    if (!ignoreXMP) {
+        const avifROData tagStandardXmp = { (const uint8_t *)"http://ns.adobe.com/xap/1.0/\0", 29 };
+        const avifROData tagExtendedXmp = { (const uint8_t *)"http://ns.adobe.com/xmp/extension/\0", 35 };
+        avifBool found = AVIF_FALSE;
+        for (jpeg_saved_marker_ptr marker = cinfo.marker_list; marker != NULL; marker = marker->next) {
+            if ((marker->marker == (JPEG_APP0 + 1)) && (marker->data_length > tagStandardXmp.size) &&
+                !memcmp(marker->data, tagStandardXmp.data, tagStandardXmp.size)) {
+                if (found) {
```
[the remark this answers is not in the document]
wantehchang: If there can be only one StandardXMP packet, then it is correct to return an error, and the error message at line 362 should not contain the word "unsupported".

wantehchang (changes requested):
Overall this looks good. Thanks. I suggest some changes.

tests/gtest/avifmetadatatest.cc
y-guyon: It can be done in another PR.

apps/shared/avifjpeg.c
```diff
@@ -359,7 +359,7 @@ avifBool avifJPEGRead(const char * inputFilename,
                 !memcmp(marker->data, tagStandardXmp.data, tagStandardXmp.size)) {
                 if (found) {
                     // TODO(yguyon): Implement instead of outputting an error.
-                    fprintf(stderr, "XMP extraction failed: unsupported XMP split into multiple chunks or invalid multiple XMP chunks\n");
+                    fprintf(stderr, "XMP extraction failed: invalid multiple XMP chunks\n");
```
wantehchang, line 362: Delete the TODO comment (line 361).

apps/shared/avifpng.c
```diff
@@ -60,7 +35,7 @@ static avifBool avifHexStringToBytes(const char * hexString, size_t hexStringLen
         }
         if (!isxdigit(hexString[i]) || !isxdigit(hexString[i + 1])) {
             avifRWDataFree(bytes);
-            fprintf(stderr, "Exif extraction failed: invalid character at " AVIF_FMT_ZU "\n", i);
+            fprintf(stderr, "Metadata extraction failed: invalid character at " AVIF_FMT_ZU "\n", i);
```
wantehchang, line 38: Note: This is just a comment, not a request for change.

The code in the apps/ directory should only use the public API of the libavif library. Strictly speaking, the way this function (and related code such as `avifRemoveHeader()`) directly sets `avif->exif`, `avif->xmp` violates API encapsulation. Since C doesn't have language features for encapsulation, we can only achieve encapsulation by convention. The intention of the libavif API is that `avif->exif` and `avif->xmp` should be set by using the setter functions `avifImageSetMetadataExif()` and `avifImageSetMetadataXMP()`:

```
    avifRWData exif;
    avifRWData xmp;
```

So ideally we should not bypass the setter functions. The reason this code could be potentially problematic is that this code knows how the setter functions currently do their job and sets `avif->exif` and `avif->xmp` in the same way. If we ever change the implementation of the setter functions and the corresponding cleanup code in `avifImageDestroy()`, then this code may break `avifImageDestroy()`.

I allowed this because this avoids copying and the implementation of the setter functions is unlikely to change. (With this kind of code outside the library, we won't be able to change the implementation of the setter functions.) But we should use the setter functions when possible.

wantehchang (approved):
LGTM with a change and a comment.

PR #1089, opened 2022-09-06
avifEncoder: Add the autoTiling option

wantehchang:
The autoTiling option, if set to AVIF_TRUE, causes avifEncoder to ignore
the tileRowsLog2 and tileColsLog2 options and automatically choose
suitable tiling values.

Also add the --autotiling option to avifenc, which sets --tilerowslog2
and --tilecolslog2 automatically.

The current implementation uses as many tiles as allowed by the minimum
tile area requirement and imposes a maximum of 8 tiles.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -172,6 +172,14 @@ avifResult avifRGBImageUnpremultiplyAlphaLibYUV(avifRGBImage * rgb);
 
 avifBool avifDimensionsTooLarge(uint32_t width, uint32_t height, uint32_t imageSizeLimit, uint32_t imageDimensionLimit);
 
+// Given the number of encoding threads or decoding threads available and the image dimensions,
+// chooses suitable values of *tileRowsLog2 and *tileColsLog2.
+//
+// Note: Although avifSetTileConfiguration() is only used in src/write.c and could be a static
+// function in that file, it is defined as an internal global function so that it can be tested by
+// unit tests.
+void avifSetTileConfiguration(int threads, uint32_t width, uint32_t height, int * tileRowsLog2, int * tileColsLog2);
+
 // ---------------------------------------------------------------------------
 // Scaling
 
@@ -283,10 +291,16 @@ typedef avifBool (*avifCodecGetNextImageFunc)(struct avifCodec * codec,
 // encoded and EncodeFinish is called, the number of samples emitted must match the number of submitted frames.
 // avifCodecEncodeImageFunc may return AVIF_RESULT_UNKNOWN_ERROR to automatically emit the appropriate
 // AVIF_RESULT_ENCODE_COLOR_FAILED or AVIF_RESULT_ENCODE_ALPHA_FAILED depending on the alpha argument.
+// avifCodecEncodeImageFunc should use tileRowsLog2 and tileColsLog2 instead of
+// encoder->tileRowsLog2, encoder->tileColsLog2, and encoder->autoTiling. The caller of
+// avifCodecEncodeImageFunc is responsible for automatic tiling if encoder->autoTiling is set to
+// AVIF_TRUE. The actual tiling values are passed to avifCodecEncodeImageFunc as parameters.
 typedef avifResult (*avifCodecEncodeImageFunc)(struct avifCodec * codec,
                                                avifEncoder * encoder,
                                                const avifImage * image,
                                                avifBool alpha,
+                                               int tileRowsLog2,
+                                               int tileColsLog2,
                                                avifEncoderChanges encoderChanges,
                                                avifAddImageFlags addImageFlags,
                                                avifCodecEncodeOutput * output);
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -295,6 +295,12 @@ typedef avifBool (*avifCodecGetNextImageFunc)(struct avifCodec * codec,
 // encoder->tileRowsLog2, encoder->tileColsLog2, and encoder->autoTiling. The caller of
 // avifCodecEncodeImageFunc is responsible for automatic tiling if encoder->autoTiling is set to
 // AVIF_TRUE. The actual tiling values are passed to avifCodecEncodeImageFunc as parameters.
+//
+// Note: The caller of avifCodecEncodeImageFunc always passes encoder->data->tileRowsLog2 and
+// encoder->data->tileColsLog2 as the tileRowsLog2 and tileColsLog2 arguments. Because
+// encoder->data is of a struct type defined in src/write.c, avifCodecEncodeImageFunc cannot
+// dereference encoder->data and has to receive encoder->data->tileRowsLog2 and
+// encoder->data->tileColsLog2 via function parameters.
 typedef avifResult (*avifCodecEncodeImageFunc)(struct avifCodec * codec,
                                                avifEncoder * encoder,
                                                const avifImage * image,
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -527,6 +527,8 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                                       avifEncoder * encoder,
                                       const avifImage * image,
                                       avifBool alpha,
+                                      int tileRowsLog2,
+                                      int tileColsLog2,
                                       avifEncoderChanges encoderChanges,
                                       avifAddImageFlags addImageFlags,
                                       avifCodecEncodeOutput * output)
@@ -737,12 +739,12 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         if (encoder->maxThreads > 1) {
             aom_codec_control(&codec->internal->encoder, AV1E_SET_ROW_MT, 1);
         }
-        if (encoder->tileRowsLog2 != 0) {
-            int tileRowsLog2 = AVIF_CLAMP(encoder->tileRowsLog2, 0, 6);
+        if (tileRowsLog2 != 0) {
+            tileRowsLog2 = AVIF_CLAMP(tileRowsLog2, 0, 6);
             aom_codec_control(&codec->internal->encoder, AV1E_SET_TILE_ROWS, tileRowsLog2);
         }
-        if (encoder->tileColsLog2 != 0) {
-            int tileColsLog2 = AVIF_CLAMP(encoder->tileColsLog2, 0, 6);
+        if (tileColsLog2 != 0) {
+            tileColsLog2 = AVIF_CLAMP(tileColsLog2, 0, 6);
             aom_codec_control(&codec->internal->encoder, AV1E_SET_TILE_COLUMNS, tileColsLog2);
         }
         if (aomCpuUsed != -1) {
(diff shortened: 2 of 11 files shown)


src/write.c
```diff
@@ -746,6 +825,27 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
 
+    // -----------------------------------------------------------------------
+    // Handle automatic tiling
+
+    encoder->data->tileRowsLog2 = encoder->tileRowsLog2;
+    encoder->data->tileColsLog2 = encoder->tileColsLog2;
+    if (encoder->autoTiling) {
+        // Use as many tiles as allowed by the minimum tile area requirement and impose a maximum
+        // of 8 tiles.
+        const int threads = 8;
```
wantehchang, line 836: The maximum number of tiles is a hardcoded value of 8. This could be an encoder option. It should not be based on `encoder->maxThreads` alone because it also depends on how many worker threads the decoder is expected to have.

One solution is to change the `autoTiling` option to be an `int` and specify the maximum number of tiles. But then the `autoTiling` option won't be fully automatic.

wantehchang (commented):
Vignesh, Yannis: Please review. Thanks.

@tongyuantongyu Yuan: Just FYI. This pull request adds a new `avifEncoder` option and as a result the existing `tileRowsLog2` and `tileColsLog2` options need special treatment in `avifCodecEncodeImageFunc`, `avifEncoderBackupSettings()`, and `avifEncoderDetectChanges()`.

wantehchang:
Note that https://github.com/AOMediaCodec/libavif/pull/1052 is a previous attempt, which requires libavif clients to call `avifSetTileConfiguration()` to set `encoder->tileRowsLog2` and `encoder->tilesColsLog2`. This pull request is easier to use but requires more changes to the code.

src/codec_aom.c
```diff
@@ -737,12 +739,12 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         if (encoder->maxThreads > 1) {
             aom_codec_control(&codec->internal->encoder, AV1E_SET_ROW_MT, 1);
         }
-        if (encoder->tileRowsLog2 != 0) {
-            int tileRowsLog2 = AVIF_CLAMP(encoder->tileRowsLog2, 0, 6);
+        if (tileRowsLog2 != 0) {
+            tileRowsLog2 = AVIF_CLAMP(tileRowsLog2, 0, 6);
```
y-guyon, line 743: Nit: It can never clamp to 0. Should it be `AVIF_CLAMP(tileRowsLog2, 1, 6)`?

tests/gtest/aviftilingtest.cc
```diff
@@ -0,0 +1,112 @@
...
+
+TEST(TilingTest, SetTileConfiguration) {
+  constexpr int kThreads = 8;
+  int tile_rows_log2;
+  int tile_cols_log2;
+  // 144p
+  avifSetTileConfiguration(kThreads, 256, 144, &tile_rows_log2,
+                           &tile_cols_log2);
+  EXPECT_EQ(tile_rows_log2, 0);
+  EXPECT_EQ(tile_cols_log2, 0);
```
y-guyon, line 18: Optional: To reduce this test's length, consider one of the following helper functions:
```cpp
void TestSetTileConfiguration(int threads, uint32_t width, uint32_t height, int expected_tile_rows_log2, int expected_tile_cols_log2) {
  int tile_rows_log2, tile_cols_log2;
  avifSetTileConfiguration(threads, width, height, &tile_rows_log2, &tile_cols_log2);
  EXPECT_EQ(tile_rows_log2, expected_tile_rows_log2);
  EXPECT_EQ(tile_cols_log2, expected_tile_cols_log2);
}
...
  TestSetTileConfiguration(kThreads, 256, 144, 0, 0);
```
```cpp
std::pair<int, int> GetTileConfig(int threads, uint32_t width, uint32_t height) {
  int tile_rows_log2, tile_cols_log2;
  avifSetTileConfiguration(threads, width, height, &tile_rows_log2, &tile_cols_log2);
  return {tile_rows_log2, tile_cols_log2};
}
...
  EXPECT_EQ(GetTileConfig(kThreads, 256, 144), {0, 0});
```

tests/gtest/aviftilingtest.cc
```diff
@@ -0,0 +1,112 @@
...
+  avifSetTileConfiguration(kThreads, 144, 256, &tile_rows_log2,
+                           &tile_cols_log2);
+  EXPECT_EQ(tile_rows_log2, 0);
+  EXPECT_EQ(tile_cols_log2, 0);
```
y-guyon, line 22: Optional: To reduce this test's length, consider having a separate symmetry test instead of doing everything twice:

```cpp
TEST(TilingTest, Symmetry) {
  for (uint32_t width : {1, 144, 256, ...}) {
    for (uint32_t height : {1, 144, 256, ...}) {
      EXPECT_EQ(GetTileConfig(kThreads, width, height), GetTileConfig(kThreads, height, width));
    }
  }
}
```
It could also be integrated directly into the helper function suggested above.

y-guyon (approved):

src/codec_aom.c
```diff
@@ -737,12 +739,12 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
             aom_codec_control(&codec->internal->encoder, AV1E_SET_TILE_ROWS, tileRowsLog2);
         }
-        if (encoder->tileColsLog2 != 0) {
-            int tileColsLog2 = AVIF_CLAMP(encoder->tileColsLog2, 0, 6);
+        if (tileColsLog2 != 0) {
+            tileColsLog2 = AVIF_CLAMP(tileColsLog2, 0, 6);
```
vigneshvg, line 747: nit/optional: seems like clamping is an AV1 spec thing. so we can move the clamping logic outside of the codec_* wrappers and just have it in one place.

src/codec_aom.c
wantehchang: Hmm... you are right. I found that `AVIF_CLAMP(tileRowsLog2, 1, 6)` will change the behavior for a negative `tileRowsLog2`. So the current code is correct. It will call `aom_codec_control(AV1E_SET_TILE_ROWS)` with the value 0 unnecessarily (0 is the default value). Since a negative `tileRowsLog2` is uncommon, I think this is fine.

The alternative is the following, but we will always do the `AVIF_CLAMP()`:

```
        tileRowsLog2 = AVIF_CLAMP(tileRowsLog2, 0, 6);
        if (tileRowsLog2 != 0) {
            aom_codec_control(&codec->internal->encoder, AV1E_SET_TILE_ROWS, tileRowsLog2);
        }
```

Would you prefer that? If so, I can make the change in a follow-up.

y-guyon: It does not matter much, it can be left as is.
Alternatively an error could be output when tileRowsLog2 < 0.

include/avif/internal.h
```diff
@@ -283,10 +291,16 @@ typedef avifBool (*avifCodecGetNextImageFunc)(struct avifCodec * codec,
 // encoded and EncodeFinish is called, the number of samples emitted must match the number of submitted frames.
 // avifCodecEncodeImageFunc may return AVIF_RESULT_UNKNOWN_ERROR to automatically emit the appropriate
 // AVIF_RESULT_ENCODE_COLOR_FAILED or AVIF_RESULT_ENCODE_ALPHA_FAILED depending on the alpha argument.
+// avifCodecEncodeImageFunc should use tileRowsLog2 and tileColsLog2 instead of
+// encoder->tileRowsLog2, encoder->tileColsLog2, and encoder->autoTiling. The caller of
```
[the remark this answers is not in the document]
wantehchang: I added a comment to explain this.

The reason is not obvious -- `encoder->data` is of a struct type defined in src/write.c, so src/codec_* cannot dereference `encoder->data`.

I did try to move the definition of that struct type (`struct avifEncoderData`) to internal.h to make it available to src/codec_*, but I found that I would also need to move the definitions of several other types (referenced by `encoder->data`) from src/write.c to internal.h, and eventually I wasn't sure if that's a good idea.

wantehchang (commented):
PTAL. Thanks!

include/avif/internal.h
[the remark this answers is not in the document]
vigneshvg: Ah okay. Makes sense.

vigneshvg (approved):

include/avif/internal.h
```diff
@@ -283,10 +291,22 @@ typedef avifBool (*avifCodecGetNextImageFunc)(struct avifCodec * codec,
...
+// Note: The caller of avifCodecEncodeImageFunc always passes encoder->data->tileRowsLog2 and
+// encoder->data->tileColsLog2 as the tileRowsLog2 and tileColsLog2 arguments. Because
+// encoder->data is of a struct type defined in src/write.c, avifCodecEncodeImageFunc cannot
+// dereference encoder->data and has to receive encoder->data->tileRowsLog2 and
+// encoder->data->tileColsLog2 via function parameters.
 typedef avifResult (*avifCodecEncodeImageFunc)(struct avifCodec * codec,
                                                avifEncoder * encoder,
                                                const avifImage * image,
                                                avifBool alpha,
+                                               int tileRowsLog2,
```
tongyuantongyu, line 308: I wonder if it's better to pass these values using `avifCodec` instead of as separated parameters? There's already `codec.allLayers` for decoder side whose value depends on `decoder.allowProgressive` field.

y-guyon (approved):

include/avif/internal.h
wantehchang: Yuan: Thank you for the suggestion. After finding that src/codec_*.c cannot dereference `encoder->data`, I looked into this option and several other options. They all have their pros and cons. My preference was to define a new struct to collect the existing `encoderChanges` parameter and the new `tileRowsLog2` and `tileColsLog2` parameters, but in the end I decided to imitate `encoderChanges` and just add `tileRowsLog2` and `tileColsLog2` as parameters.

I can also share my thought process behind this pull request. There were three design decisions.

1. Whether to handle automatic tiling in src/write.c or in the individual src/codec_*.c files.
2. Where to store the actual tiling values (we need to store them so we can detect changes)
3. If src/write.c handles automatic tiling, how to pass the actual tiling values to src/codec_*.c files.

I decided to handle automatic tiling in src/write.c to avoid duplicating code. But not much code would be duplicated if we were to handle automatic tiling in src/codec_*.c.

It's tempting to store the actual tiling values in `encoder->tileRowsLog2` and `encoder->tileColsLog2` directly. It would allow libavif clients to know the actual tiling values used by the encoder. But I think it'd be strange to let `encoder` become an in/out parameter. So I didn't do that.

tongyuantongyu: Wan-Teh: Thanks for your detail explanation. I also prefer using a new struct to collect them, especially if we are going to add more parameters like these. I think `avifCodec` may be used for this, but it's not a perfect fit either.

wantehchang: Yes. Some avifCodecs such as rav1e and SVT-AV1 are encoders only, so it isn't ideal to put decoder-specific fields such as `operatingPoint` and `allLayers` in the `avifCodec` struct.

PR #1090, opened 2022-09-06
Add CI for Android JNI Bindings

vigneshvg:
This will build the JNI bindings with each PR/push to ensure that
we don't break the Android JNI build.

It is quite fast (< 3 minutes) when run locally.

We could potentially add some tests as well and run them in an
emulator on the CI. But that is work for the future.


wantehchang (approved):
LGTM.

vigneshvg:
Example run: https://github.com/AOMediaCodec/libavif/runs/8213702213

PR #1093, opened 2022-09-07
android_jni: Set threads to number of available CPU cores

vigneshvg:
This will speed up the decoding process.

Also update the Android JNI CI to install the cmake from the SDK for building android specific code.

### DIFF
--- a/.github/workflows/ci-android-jni.yml
+++ b/.github/workflows/ci-android-jni.yml
@@ -20,8 +20,8 @@ jobs:
       - name: Setup cmake
         uses: jwlawson/actions-setup-cmake@v1.12
         with:
-          # We use the same version as the one in build.gradle.
-          cmake-version: "3.18.1"
+          # This is the minimum cmake version needed to build libgav1.
+          cmake-version: "3.7.x"
       - name: Build libgav1 with the Android NDK
         working-directory: ext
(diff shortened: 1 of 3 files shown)


jzern (approved):

.github/workflows/ci-android-jni.yml
```diff
@@ -20,8 +20,8 @@ jobs:
       - name: Setup cmake
         uses: jwlawson/actions-setup-cmake@v1.12
         with:
-          # We use the same version as the one in build.gradle.
-          cmake-version: "3.18.1"
+          # This is the minimum cmake version needed to build libgav1.
+          cmake-version: "3.7.x"
```
wantehchang, line 24: [This is just a comment.] That libgav1 can be built with cmake 3.7 should be checked by libgav1's CI builds. We don't need to check that here.

wantehchang (approved):
LGTM.

.github/workflows/ci-android-jni.yml
vigneshvg: yeah. i just need to specify some cmake version here so the cmake plugin can be run. instead of specifying an arbitrary version, i used the one that was required by libgav1 (since that is essentially why we use the cmake plugin here).

wantehchang: Thanks for the explanation. I am not familiar with the build process in this CI build, so my comment may have been ill-informed.

PR #1097, opened 2022-09-09
Adjust gdk-pixbuf loader for newest API changes

novomesk:
The plug-in doesn't build without these changes.


wantehchang (approved):
LGTM.

PR #1099, opened 2022-09-09
Parse Exif for orientation in avifEncoder

y-guyon:
Create avifExtractExifOrientation() in exif.c.
Move avifExtractExifTiffHeaderOffset() from write.c to exif.c.

Add tests in avifmetadatatest.


include/avif/internal.h
```diff
@@ -42,10 +42,16 @@ extern "C" {
 
 float avifRoundf(float v);
 
+// H is platform-dependent. Could be little- or big-endian.
+// N is big-endian: most- to least-significant bytes.
+// Z is little-endian: most- to least-significant bytes.
```
wantehchang, line 47: "H" and "N" come from network programming. "H" stands for "host", i.e., the computer. "N" stands for "network". The network byte order is big endian by definition. The `avifHTONS`, `avifNTOHS`, ... function names come from the Berkeley sockets API, which has the `htons`, `ntohs`, ... functions.

I have not heard of "Z". What does it stand for?

Is the comment for "Z" (little-endian) incorrect? It is the same as the comment for big-endian: "most- to least-significant bytes".

src/exif.c
```diff
@@ -0,0 +1,133 @@
...
+{
+    uint32_t exifTiffHeaderOffset;
+    const avifResult result = avifExtractExifTiffHeaderOffset(exif, &exifTiffHeaderOffset);
+    if (result != AVIF_RESULT_OK) {
+        // Couldn't find the TIFF header
+        return result;
+    }
+
+    avifROData raw = { exif->data + exifTiffHeaderOffset, exif->size - exifTiffHeaderOffset };
+    const avifBool isLittleEndian = (raw.data[0] == 'I');
```
wantehchang, line 45: Nit: name this variable `littleEndian` (the name used for the function parameters).

src/exif.c
```diff
@@ -0,0 +1,133 @@
...
+                        imir->mode = 0; // ignored
+                        return AVIF_RESULT_OK;
+                    case 4: // The 0th row is at the visual bottom of the image, and the 0th column is the visual left-hand side.
+                        *flags = AVIF_TRANSFORM_IMIR;
+                        irot->angle = 0; // ignored
+                        imir->mode = 0;
+                        return AVIF_RESULT_OK;
+                    case 5: // The 0th row is the visual left-hand side of the image, and the 0th column is the visual top.
+                        *flags = AVIF_TRANSFORM_IROT | AVIF_TRANSFORM_IMIR;
+                        irot->angle = 1; // applied before imir according to Chrome's rendering behavior
```
wantehchang, line 103: I believe the order in which `irot` and `imir` are applied is specified by HEIF. Please check HEIF and verify that. If that's the case, this comment should say "according to HEIF", not "according to Chrome's rendering behavior".

src/exif.c
```diff
@@ -0,0 +1,133 @@
...
+                return AVIF_RESULT_INVALID_EXIF_PAYLOAD;
+            }
+            // Orientation attribute according to JEITA CP-3451C section 4.6.4 (TIFF Rev. 6.0 Attribute Information):
+            if (tag == 0x0112 && type == /*SHORT=*/0x03 && count == 0x01) { // Exif IFD
+                // Mapping from Exif orientation as defined in JEITA CP-3451C section 4.6.4.A Orientation
+                // to irot and imir boxes as defined in HEIF ISO/IEC 28002-12:2021 sections 6.5.10 and 6.5.12.
+                switch (firstHalfOfValueOffset) {
+                    case 1: // The 0th row is at the visual top of the image, and the 0th column is the visual left-hand side.
+                        *flags = AVIF_TRANSFORM_NONE;
+                        irot->angle = 0; // ignored
```
wantehchang, line 83: I don't know what these "ignored" comments mean...

UPDATE: I understand it now. It means the caller will ignore the value because the corresponding bit in `*flags` is not set.

I think it may be better to just pass `avifImage *image` to this functon and let this function reads `image->exit` and modifies `image->transformFlags`, `image->irot`, and `image->imir` directly.

If you do that, then this function should be named `avifImageExtractExifOrientation` to follow libavif's naming convention for object-oriented, method-like functions.

src/write.c
```diff
@@ -961,15 +945,27 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
                 return result;
             }
+            // Import Exif orientation only if not already explicitly specified by the user.
+            if (!(encoder->data->imageMetadata->transformFlags & (AVIF_TRANSFORM_IROT | AVIF_TRANSFORM_IMIR))) {
+                avifTransformFlags transformFlags;
+                avifImageRotation irot;
+                avifImageMirror imir;
+                // Ignore any Exif parsing failure.
+                if (avifExtractExifOrientation(&encoder->data->imageMetadata->exif, &transformFlags, &irot, &imir) == AVIF_RESULT_OK) {
+                    encoder->data->imageMetadata->transformFlags = encoder->data->imageMetadata->transformFlags | transformFlags;
```
wantehchang, line 960: Nit: Use the `|=` operator?

src/exif.c
```diff
@@ -0,0 +1,133 @@
+// Copyright 2022 Google LLC. All rights reserved.
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "avif/internal.h"
+
+#include <stdint.h>
+#include <string.h>
+
+avifResult avifExtractExifTiffHeaderOffset(const avifRWData * exif, uint32_t * exifTiffHeaderOffset)
```
wantehchang, line 9: Nit/optional: Since there is only one offset in this function, the second parameter could be named `offset` with no risk of confusion. The function name `avifExtractExifTiffHeaderOffset` documents what kind of offset it is.

src/exif.c
```diff
@@ -0,0 +1,133 @@
...
+        return AVIF_RESULT_INVALID_EXIF_PAYLOAD;
+    }
+
+    const uint8_t tiffHeaderBE[4] = { 'M', 'M', 0, 42 };
+    const uint8_t tiffHeaderLE[4] = { 'I', 'I', 42, 0 };
+    for (; *exifTiffHeaderOffset < (exif->size - 4); ++*exifTiffHeaderOffset) {
+        if (!memcmp(&exif->data[*exifTiffHeaderOffset], tiffHeaderBE, sizeof(tiffHeaderBE))) {
+            break;
+        }
+        if (!memcmp(&exif->data[*exifTiffHeaderOffset], tiffHeaderLE, sizeof(tiffHeaderLE))) {
```
wantehchang, line 23: Nit: It's safe to use 4 instead of `sizeof(tiffHeaderBE)` and `sizeof(tiffHeaderLE)` because the size of those two arrays is fixed by the standard. Also this function already uses 4 in three places.

src/exif.c
```diff
@@ -0,0 +1,133 @@
...
+            break;
+        }
+    }
+
+    if (*exifTiffHeaderOffset >= exif->size - 4) {
+        // Couldn't find the TIFF header
+        return AVIF_RESULT_INVALID_EXIF_PAYLOAD;
+    }
+    return AVIF_RESULT_OK;
```
wantehchang, line 32: Nit: I believe lines 19-32 can be simplified as follows:

```
    for (; *exifTiffHeaderOffset < (exif->size - 4); ++*exifTiffHeaderOffset) {
        if (!memcmp(&exif->data[*exifTiffHeaderOffset], tiffHeaderBE, sizeof(tiffHeaderBE)) ||
            !memcmp(&exif->data[*exifTiffHeaderOffset], tiffHeaderLE, sizeof(tiffHeaderLE))) {
            return AVIF_RESULT_OK;
        }
    }

    // Couldn't find the TIFF header
    return AVIF_RESULT_INVALID_EXIF_PAYLOAD;
```

wantehchang (changes requested):
Yannis: The first round of comments is very incomplete/preliminary because I only took a quick look at a few functions.

src/exif.c
wantehchang: Another idea is to keep the current function prototype but use `flags`, `irot`, and `imir` as in/out parameters. Then case 1, case 2, etc. of this switch statement would be written as:

```
                    case 1: // The 0th row is at the visual top of the image, ...
                        return AVIF_RESULT_OK;
                    case 2: // The 0th row is at the visual top of the image, ...
                        *flags |= AVIF_TRANSFORM_IMIR;
                        imir->mode = 1;
                        return AVIF_RESULT_OK;
                    case 3: // The 0th row is at the visual bottom of the image, ...
                        *flags |= AVIF_TRANSFORM_IROT;
                        irot->angle = 2;
                        return AVIF_RESULT_OK;
                    ...
```

And we would call this function as follows:

```
            // Import Exif orientation only if not already explicitly specified by the user.
            if (!(encoder->data->imageMetadata->transformFlags & (AVIF_TRANSFORM_IROT | AVIF_TRANSFORM_IMIR))) {
                // Ignore any Exif parsing failure.
                (void)avifExtractExifOrientation(&encoder->data->imageMetadata->exif,
                                                 &encoder->data->imageMetadata->transformFlags,
                                                 &encoder->data->imageMetadata->irot,
                                                 &encoder->data->imageMetadata->imir);
            }
```

src/exif.c
```diff
@@ -0,0 +1,133 @@
...
+                        *flags = AVIF_TRANSFORM_IROT;
+                        irot->angle = 1;
+                        imir->mode = 0; // ignored
+                        return AVIF_RESULT_OK;
+                    default: // reserved
+                        break;
+                }
+            }
+            // Exif IFD according to JEITA CP-3451C is (tag == 0x8769 && type == /*LONG=*/0x04 && count == 0x01) with
+            // valueOffset being a "pointer" (meaning offset) to the Exif IFD but there is little to do with that information.
```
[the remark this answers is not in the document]
y-guyon: I removed the comment instead.

include/avif/internal.h
y-guyon: > "H" and "N" come from network programming. "H" stands for "host", i.e., the computer. "N" stands for "network". The network byte order is big endian by definition. The `avifHTONS`, `avifNTOHS`, ... function names come from the Berkeley sockets API, which has the `htons`, `ntohs`, ... functions.

Thanks for the context. I searched for that information but failed to obtain it on my own.

> I have not heard of "Z". What does it stand for?

I had no clue what H/N were so took another random letter. Let's take "C" for "custom".

> Is the comment for "Z" (little-endian) incorrect? It is the same as the comment for big-endian: "most- to least-significant bytes".

Fixed.

src/exif.c
y-guyon: I already searched for it and failed to find an explicit proof in HEIF. However I just checked MIAF again and it is specified there in section 7.3.6.7; I think I missed it because "irot" is not part of MIAF.

src/exif.c
y-guyon: > Another idea is to keep the current function prototype but use `flags`, `irot`, and `imir` as in/out parameters.

I did that at first, then thought that it was difficult to grasp what was the input and what was set. Also it does not make sense to set the angle without clearing the rotation, unless it is known that this function is only called when there is no `AVIF_TRANSFORM_IROT` and no `AVIF_TRANSFORM_IMIR` prior to calling it.

> I think it may be better to just pass `avifImage *image` to this functon and let this function reads `image->exit` and modifies `image->transformFlags`, `image->irot`, and `image->imir` directly.

I think the logic at the call site is slightly more obscure that way but done.

wantehchang: I didn't have time to review this pull review again today. I will review it tomorrow (Wednesday).

One thing I noticed last Friday was that this pull request always imports Exif orientation. It may be better to let the user control whether Exif orientation is imported. One reason is that it will allow the encoded output to match the input `avifImage` faithfully. In that design, we will need a function that the user can call to import Exif orientation, and that function can take `avifImage *image` as input.

y-guyon: > I didn't have time to review this pull review again today.

We can save time by agreeing on the general design before reviewing the implementation details.

> One thing I noticed last Friday was that this pull request always imports Exif orientation.

The default behavior should be to encode Exif orientation as irot/imir whenever possible. That is why I put directly within `avifEncoderAddImageInternal()`. It was possible to skip it but in a shady way: set `irot.angle` to `0` and use `AVIF_TRANSFORM_IROT`. It would keep the default unoriented behavior but would add an unnecessary `irot` box to the encoded bitstream.

> It may be better to let the user control whether Exif orientation is imported.

I agree. I added an `extractExifOrientationToIrotImir` argument to `avifImageSetMetadataExif()` instead of doing it in `avifEncoderAddImageInternal()`. If `avifImageSetMetadataExif()` is the only recommended way to deal with Exif metadata, the user must set that option. The downsides are: the user has to care about the irot/imir topic (no default), `avifImageSetMetadataExif(image, nullptr, 0, AVIF_TRUE)` will not clear irot/imir (easily overlooked).

The alternative would be going back to the `avifEncoderAddImageInternal()` but with an option such as `avifEncoder.extractExifOrientationToIrotImir` set to `AVIF_TRUE` by default. This alternative does not respect your constraint about matching input `avifImage` faithfully for default settings.

What do you think?

---

Somewhat related: I found the following in MIAF ISO 23000-22 section 7.3.10.

> Metadata of type Exif as specified by JEITA CP-3451 may be present in the file and shall be output by the MIAF reader; this means that a reader shall be able to transfer this data to the application. There is no need for a reader to understand the details of the metadata. **There should be no image transformations expressed by Exif (rotation, mirroring, etc.) indicated in the Exif metadata, in files encoded according to this document.** If present, they shall be ignored by a MIAF renderer, i.e. not cause a change in the presentation of associated images by a MIAF player.

Should we strip orientation data from the encoded Exif payload in libavif?

y-guyon:
Sorry for rebasing. I ran into this weird issue where _not_ syncing this PR to main produced the same [error](https://github.com/AOMediaCodec/libavif/actions/runs/3053683005/jobs/4924618517) as rebasing it without fixing the conflict, so I rebased and force pushed.

include/avif/internal.h
wantehchang: Here is the man page for the `htons`, `ntohs`, ... functions:

https://pubs.opengroup.org/onlinepubs/9699919799/functions/htonl.html

Their names are cryptic for people unfamiliar with socket programming. We can switch to more descriptive names if we are adding little-endian variants of these functions. But let's focus on correctness of this pull request first and deal with these function names later.

src/exif.c
wantehchang: I will comment on how to control extracting Exif orientation later.

If it is safe (i.e., not prone to software security issues) to strip orientation data from Exif, we can do that. I think keeping irot and imir in sync with the Exif orientation data is the next best thing and is good enough.

src/exif.c
```diff
@@ -0,0 +1,119 @@
...
+                !avifROStreamReadU16Endianness(&stream, &secondHalfOfValueOffset, littleEndian)) {
+                return AVIF_RESULT_INVALID_EXIF_PAYLOAD;
+            }
+            // Orientation attribute according to JEITA CP-3451C section 4.6.4 (TIFF Rev. 6.0 Attribute Information):
+            if (tag == 0x0112 && type == /*SHORT=*/0x03 && count == 0x01) { // Exif IFD
+                const avifTransformFlags otherFlags = image->transformFlags & ~(AVIF_TRANSFORM_IROT | AVIF_TRANSFORM_IMIR);
+                // Mapping from Exif orientation as defined in JEITA CP-3451C section 4.6.4.A Orientation
+                // to irot and imir boxes as defined in HEIF ISO/IEC 28002-12:2021 sections 6.5.10 and 6.5.12.
+                switch (firstHalfOfValueOffset) {
+                    case 1: // The 0th row is at the visual top of the image, and the 0th column is the visual left-hand side.
```
wantehchang, line 69: I checked the mapping using two methods. It is correct.

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ avifImageCopy() and avifImageAllocatePlanes().
 * avifImageRGBToYUV() now uses libyuv fast paths by default. It may slightly
   change conversion results. The old behavior can be restored by setting
   avifRGBImage::chromaDownsampling to AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY.
+* Exif and XMP metadata is imported from PNG and JPEG files.
+* irot/imir orientation is extracted from Exif metadata by the avifEncoder.
```
wantehchang, line 28: Nit: move "orientation" to a different place in the sentence, e.g.,

```
* irot/imir is extracted from the orientation in Exif metadata by the avifEncoder.
```

apps/avifenc.c
```diff
@@ -890,7 +890,7 @@ int main(int argc, char * argv[])
         avifImageSetProfileICC(image, iccOverride.data, iccOverride.size);
     }
     if (exifOverride.size) {
-        avifImageSetMetadataExif(image, exifOverride.data, exifOverride.size);
+        avifImageSetMetadataExif(image, exifOverride.data, exifOverride.size, /*extractExifOrientationToIrotImir=*/AVIF_TRUE);
```
wantehchang, line 893: Adding the extractExifOrientationToIrotImir parameter is an incompatible change. We can remain backward compatible by adding a new function to extract Exif orientation. We can just make `avifExtractExifOrientationToIrotImir()` a public API function.

wantehchang (changes requested):
I checked the `avifExtractExifOrientationToIrotImir()` function carefully.

src/exif.c
kmilos: Just a small subtle fact to think about: the Exif Orientation tag has a default value of 1 (first pixel supplied is top left) if not present, so you cannot really "strip" it. Rather than going through the acrobatics with reallocating the payload and copying chunks around, you might want to simply ensure it is set to 1 instead (i.e. instruct readers to do nothing further)?

apps/avifenc.c
y-guyon: Done. I dislike having to rely on the user to remember to call `avifImageExtractExifOrientationToIrotImir()` though.

src/exif.c
```diff
@@ -0,0 +1,119 @@
...
+                        return AVIF_RESULT_OK;
+                    case 8: // The 0th row is the visual left-hand side of the image, and the 0th column is the visual bottom.
+                        image->transformFlags = otherFlags | AVIF_TRANSFORM_IROT;
+                        image->irot.angle = 1;
+                        image->imir.mode = 0; // ignored
+                        return AVIF_RESULT_OK;
+                    default: // reserved
+                        break;
+                }
+            }
```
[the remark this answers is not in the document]
y-guyon: > IMPORTANT: I compared this code with the `ReadExifDirectory()` function in Chromium's third_party/blink/renderer/platform/image-decoders/jpeg/jpeg_image_decoder.cc. This code is missing the case of tag == ExifTags::kExifOffsetTag (0x8769), which indicates a subdirectory.
> 
> If the orientation data cannot be stored in a subdirectory, then this is fine.

From what I understand in JEITA CP-3451C, the `0x8769` tag contains a pointer (offset) to the `Exif IFD`. There is no mention of "directory" in this specification. The section 4.6.5 Exif IFD Attribute Information does not include the Orientation tag.

src/exif.c
```diff
@@ -0,0 +1,119 @@
...
+            uint16_t secondHalfOfValueOffset;
+            if (!avifROStreamReadU16Endianness(&stream, &tag, littleEndian) ||
+                !avifROStreamReadU16Endianness(&stream, &type, littleEndian) ||
+                !avifROStreamReadU32Endianness(&stream, &count, littleEndian) ||
+                !avifROStreamReadU16Endianness(&stream, &firstHalfOfValueOffset, littleEndian) ||
+                !avifROStreamReadU16Endianness(&stream, &secondHalfOfValueOffset, littleEndian)) {
+                return AVIF_RESULT_INVALID_EXIF_PAYLOAD;
+            }
+            // Orientation attribute according to JEITA CP-3451C section 4.6.4 (TIFF Rev. 6.0 Attribute Information):
+            if (tag == 0x0112 && type == /*SHORT=*/0x03 && count == 0x01) { // Exif IFD
```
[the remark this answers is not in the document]
y-guyon: > Is the comment "Exif IFD" correct? The conditional expression in this line indicates the orientation tag.

Removed. I think I am confused about what I call "Exif orientation" actually being "TIFF orientation" and not part of the "Exif IFD".

src/exif.c
```diff
@@ -0,0 +1,119 @@
...
+    avifROData raw = { image->exif.data + offset, image->exif.size - offset };
+    const avifBool littleEndian = (raw.data[0] == 'I');
+    avifROStream stream;
+    avifROStreamStart(&stream, &raw, NULL, NULL);
+
+    // TIFF Header
+    uint32_t offsetToNextIfd;
+    if (!avifROStreamSkip(&stream, 2 + 2) ||                                       // Byte Order ("II" or "MM" then "42")
+        !avifROStreamReadU32Endianness(&stream, &offsetToNextIfd, littleEndian) || // Offset to 0th IFD
+        offsetToNextIfd < 2 + 2 + 4) {
```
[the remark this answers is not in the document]
y-guyon: > Is the `offsetToNextIfd < 2 + 2 + 4` check necessary?

It was an easy bitstream-compliance check. I removed it.

> You don't do this check after the `avifROStreamReadU32(&stream, &offsetToNextIfd)` call at line 114.

No because I do not know if the chunks are always in order so the check to perform is not obvious.

> NOTE: It is important to check that `offsetToNextIfd` does not go past the end of Exif metadata. This is done implicitly by the `avifROStreamSetOffset(&stream, offsetToNextIfd)` call at line 45.

Acknowledged.

src/exif.c
```diff
@@ -0,0 +1,119 @@
...
+                    default: // reserved
+                        break;
+                }
+            }
+        }
+        if (!avifROStreamReadU32(&stream, &offsetToNextIfd)) {
+            return AVIF_RESULT_INVALID_EXIF_PAYLOAD;
+        }
+    }
+    return AVIF_RESULT_NO_CONTENT;
```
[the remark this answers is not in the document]
y-guyon: It does not matter much as the only use of `avifImageExtractExifOrientationToIrotImir()` does not check the returned value.
The orientation tag is not mandatory (only recommended) according to JEITA CP-3451C section 4.6.8.A so I changed it to `AVIF_RESULT_OK`.

src/exif.c
```diff
@@ -0,0 +1,119 @@
...
+                        image->irot.angle = 3; // applied before imir according to MIAF spec ISO/IEC 28002-12:2021 - section 7.3.6.7
+                        image->imir.mode = 0;
+                        return AVIF_RESULT_OK;
+                    case 8: // The 0th row is the visual left-hand side of the image, and the 0th column is the visual bottom.
+                        image->transformFlags = otherFlags | AVIF_TRANSFORM_IROT;
+                        image->irot.angle = 1;
+                        image->imir.mode = 0; // ignored
+                        return AVIF_RESULT_OK;
+                    default: // reserved
+                        break;
```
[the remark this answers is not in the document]
y-guyon: > IMPORTANT: It seems that we don't need to keep going if we get a reserved value. We can directly return a suitable error code.

Done but how do you justify that? What if there is another orientation attribute stored in the same file?
I briefly read JEITA CP-3451C again and could not find whether it is allowed, although I believe it is forbidden.

src/exif.c
```diff
@@ -0,0 +1,119 @@
...
+                }
+            }
+        }
+        if (!avifROStreamReadU32(&stream, &offsetToNextIfd)) {
```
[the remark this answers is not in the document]
y-guyon: > IMPORTANT: Should we use `avifROStreamReadU16Endianness` here? If not, add a comment to assure the reader that this is correct.

Good catch.

> UPDATE: I didn't find any code in Chromium's third_party/blink/renderer/platform/image-decoders/jpeg/jpeg_image_decoder.cc that matches this exactly. The closest match is the `CheckExifHeader()` call, but that call will also read the TIFF header.

I followed JEITA CP-3451C rather than another project. Most implementations I saw on the web when searching for parsers were just lazily looking for the first Exif IFD orientation tag.

src/write.c
```diff
@@ -961,15 +945,15 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
+            const avifResult result = avifEncoderDataCreateExifItem(encoder->data, &encoder->data->imageMetadata->exif);
             if (result != AVIF_RESULT_OK) {
                 return result;
             }
         }
 
-        if (firstCell->xmp.size > 0) {
-            const avifResult result = avifEncoderDataCreateXMPItem(encoder->data, &firstCell->xmp);
+        if (encoder->data->imageMetadata->xmp.size > 0) {
+            const avifResult result = avifEncoderDataCreateXMPItem(encoder->data, &encoder->data->imageMetadata->xmp);
```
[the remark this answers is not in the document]
y-guyon: > this is just a cleanup or clarification, not a bug fix, right?

It was a clarification when `avifExtractExifOrientationToIrotImir()` was applied to `encoder->data->imageMetadata` but not to `firstCell`.

> If so, then the original code seems simpler.

Done.

src/exif.c
```diff
@@ -0,0 +1,119 @@
+// Copyright 2022 Google LLC. All rights reserved.
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "avif/internal.h"
+
+#include <stdint.h>
+#include <string.h>
+
+avifResult avifExtractExifTiffHeaderOffset(const avifRWData * exif, uint32_t * offset)
```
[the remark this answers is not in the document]
y-guyon: > 1. Nit: Preserve the original comment:
> 
> `// Validate Exif payload (if any) and find TIFF header offset`

The `if any` part is no longer valid and the function is described in `internal.h` already.

> 3. Nit: Change "Extract" to "Find" or "Get" in the function's name.

Done.

src/exif.c
y-guyon: > I think keeping irot and imir in sync with the Exif orientation data is the next best thing and is good enough.

Let's do only that in this PR then, I agree. It was just a thought.

> the Exif Orientation tag has a default value of 1

Indeed, it is defined in JEITA CP-3451C section 4.6.4.A Orientation.

src/exif.c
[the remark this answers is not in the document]
wantehchang: My justification for not keep going when we get an unknown (reserved) value is that we also do not keep going when we get a known value (1-8). In other words, I suggest we stop at the first orientation attribute.

It is also reasonable to stop at the first **known** orientation attribute. Either policy is fine as long as we document it.

wantehchang (approved):
LGTM.

src/exif.c
```diff
@@ -0,0 +1,124 @@
...
+
+    // TIFF Header
+    uint32_t offsetToNextIfd;
+    if (!avifROStreamSkip(&stream, 4) ||                                           // Skip tiffHeaderBE or tiffHeaderLE.
+        !avifROStreamReadU32Endianness(&stream, &offsetToNextIfd, littleEndian) || // Offset to 0th IFD
+        offsetToNextIfd < 4 + 4) {
+        return AVIF_RESULT_INVALID_EXIF_PAYLOAD;
+    }
+
+    while (offsetToNextIfd) { // for each IFD
```
wantehchang, line 45: It seems that the while loop is not necessary, because orientation is in the 0th IFD.

wantehchang: BUG: I just realized that if the first `offsetToNextIfd` is 8 and the second `offsetToNextIfd` is also 8, this while loop will be an infinite loop. We need to either remove the while loop (i.e., process only the 0th IFD) or defend against an infinite loop.

It would be good to create a test JPEG file with this condition.

wantehchang (changes requested):

wantehchang (approved):

PR #1102, opened 2022-09-12
Reuse png_infop for png_read_end() in avifPNGRead()

y-guyon:
Otherwise there can be a memory leak.

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -214,7 +214,6 @@ avifBool avifPNGRead(const char * inputFilename,
     volatile avifBool readResult = AVIF_FALSE;
     png_structp png = NULL;
     png_infop info = NULL;
-    png_infop infoEnd = NULL;
     png_bytep * volatile rowPointers = NULL;
 
     avifRGBImage rgb;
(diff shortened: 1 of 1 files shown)


jzern (approved):

apps/shared/avifpng.c
```diff
@@ -365,7 +359,6 @@ avifBool avifPNGRead(const char * inputFilename,
     }
     if (png) {
         png_destroy_read_struct(&png, &info, NULL);
-        png_destroy_read_struct(&png, &infoEnd, NULL);
```
wantehchang, line 368: I am not familiar with libpng, so I could be wrong. But it seems that the original code was also correct except that it called `png_destroy_read_struct()` incorrectly. Would it work if we pass `&infoEnd` as the third argument?

```
        png_destroy_read_struct(&png, &info, &infoEnd);
```

The third parameter is named `end_info_ptr_ptr`:

https://refspecs.linuxbase.org/LSB_4.0.0/LSB-Desktop-generic/LSB-Desktop-generic/libpng12.png.destroy.read.struct.1.html

This will match how `end_info_ptr` is used in `test_one_file()` in https://source.chromium.org/chromium/chromium/src/+/main:third_party/libpng/pngtest.c.

See also https://source.chromium.org/chromium/chromium/src/+/main:third_party/libpng/contrib/oss-fuzz/libpng_read_fuzzer.cc and https://source.chromium.org/chromium/chromium/src/+/main:third_party/libwebp/src/imageio/pngdec.c. Search for "png_read_end" in these files.

wantehchang: I am not familiar with libpng. If the new code is correct, it is not necessary to go back to `infoEnd`. I was just sharing that the original code also seemed correct except for the `png_destroy_read_struct()` calls.

Issue #1086, opened 2022-09-04
avifdec writing PNG returns error


FrankGalligan:
Steps to reproduce:
1. ./avifenc input.png output.avif
2. ./avifdec output.avif new.png

I get error:
libpng error: profile 'libavif': 49002h: length does not match profile

If I comment out this line:
https://github.com/AOMediaCodec/libavif/blob/main/apps/shared/avifpng.c#L421

Then I do not see the error.

wantehchang:
Frank: I can't reproduce this bug. Could you provide a sample PNG file? You can email it to me and Yannis. Thanks.

wantehchang:
Yannis: Good job tracking this bug down. This kind of memory error is a serious problem. We should take the opportunity to review the relevant code carefully. Do you know why the Exif and XMP metadata are not affected by this bug? It seems that they avoid this bug by either reading the data early or storing the offset (in the `avifExtent` struct), the same approach you used to fix the bug.

y-guyon:
I sent https://github.com/AOMediaCodec/libavif/pull/1114 as an attempt to catch this kind of issue, when called by a fuzz target for example. A local run of an internal fuzz target did not report anything. Maybe the CI continuous fuzzing will.

wantehchang:
Fixed by https://github.com/AOMediaCodec/libavif/pull/1103.

PR #1103, opened 2022-09-12
Fix mem issue in avifParseColourInformationBox()

y-guyon:
See https://github.com/AOMediaCodec/libavif/issues/1086.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -75,7 +75,7 @@ typedef struct avifContentType
 typedef struct avifColourInformationBox
 {
     avifBool hasICC;
-    size_t iccOffset;
+    uint64_t iccOffset;
     size_t iccSize;
 
     avifBool hasNCLX;
@@ -1728,7 +1728,7 @@ static avifBool avifParseAuxiliaryTypeProperty(avifProperty * prop, const uint8_
     return AVIF_TRUE;
 }
 
-static avifBool avifParseColourInformationBox(avifProperty * prop, const uint8_t * raw, size_t rawOffset, size_t rawLen, avifDiagnostics * diag)
+static avifBool avifParseColourInformationBox(avifProperty * prop, const uint8_t * raw, uint64_t rawOffset, size_t rawLen, avifDiagnostics * diag)
 {
     BEGIN_STREAM(s, raw, rawLen, diag, "Box[colr]");
 
@@ -1919,7 +1919,7 @@ static avifBool avifParseAV1LayeredImageIndexingProperty(avifProperty * prop, co
 
 static avifBool avifParseItemPropertyContainerBox(avifPropertyArray * properties,
                                                   const uint8_t * raw,
-                                                  size_t rawOffset,
+                                                  uint64_t rawOffset,
                                                   size_t rawLen,
                                                   avifDiagnostics * diag)
 {
@@ -2152,7 +2152,7 @@ static avifBool avifParseItemDataBox(avifMeta * meta, const uint8_t * raw, size_
     return AVIF_TRUE;
 }
 
-static avifBool avifParseItemPropertiesBox(avifMeta * meta, const uint8_t * raw, size_t rawOffset, size_t rawLen, avifDiagnostics * diag)
+static avifBool avifParseItemPropertiesBox(avifMeta * meta, const uint8_t * raw, uint64_t rawOffset, size_t rawLen, avifDiagnostics * diag)
 {
     BEGIN_STREAM(s, raw, rawLen, diag, "Box[iprp]");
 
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -1739,7 +1739,10 @@ static avifBool avifParseColourInformationBox(avifProperty * prop, const uint8_t
     CHECK(avifROStreamRead(&s, colorType, 4));
     if (!memcmp(colorType, "rICC", 4) || !memcmp(colorType, "prof", 4)) {
         colr->hasICC = AVIF_TRUE;
-        colr->icc = avifROStreamCurrent(&s);
+        // Remember the offset of the ICC chunk relative to the beginning of the stream. A direct pointer cannot be stored because
+        // decoder->io->persistent could have been AVIF_FALSE when obtaining raw through decoder->io->read().
+        // The bytes could be copied now instead of remembering the offset, but it is as invasive as passing rawOffset everywhere.
+        colr->iccOffset = rawOffset + avifROStreamOffset(&s);
```
jzern, line 1745: at line 1939 rawOffset is passed in as `rawOffset + avifROStreamOffset(&s)` is this value correct?

y-guyon: Yes, `rawOffset` is the offset at which you can find `raw` within the input bitstream. Should it have a different name? I am not so sure I follow what is unclear here.

src/read.c
```diff
@@ -1740,7 +1740,10 @@ static avifBool avifParseColourInformationBox(avifProperty * prop, const uint8_t
     AVIF_CHECK(avifROStreamRead(&s, colorType, 4));
     if (!memcmp(colorType, "rICC", 4) || !memcmp(colorType, "prof", 4)) {
         colr->hasICC = AVIF_TRUE;
-        colr->icc = avifROStreamCurrent(&s);
+        // Remember the offset of the ICC chunk relative to the beginning of the stream. A direct pointer cannot be stored because
```
wantehchang, line 1743: Pedantic nit: ICC chunk => "ICC payload" or "ICC profile"

HEIF/MIAF/AVIF doesn't seem to use the term "chunk" in the same meaning as some other image formats such as PNG.

src/read.c
```diff
@@ -1728,7 +1728,7 @@ static avifBool avifParseAuxiliaryTypeProperty(avifProperty * prop, const uint8_
     return AVIF_TRUE;
 }
 
-static avifBool avifParseColourInformationBox(avifProperty * prop, const uint8_t * raw, size_t rawLen, avifDiagnostics * diag)
+static avifBool avifParseColourInformationBox(avifProperty * prop, const uint8_t * raw, uint64_t rawOffset, size_t rawLen, avifDiagnostics * diag)
```
wantehchang, line 1731: Nit: Since a {pointer, length} pair is a common construct, I suggest not inserting an offset in the middle. How about listing the three parameters in the order of raw, rawLen, rawOffset or rawOffset, raw, rawLen?

This change would need to be made to all the functions that take these three parameters.

src/read.c
```diff
@@ -3637,8 +3661,14 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                 if (colrICCSeen) {
                     return AVIF_RESULT_BMFF_PARSE_FAILED;
                 }
+                avifROData icc;
+                const avifResult readResult =
+                    decoder->io->read(decoder->io, /*readFlags=*/0, prop->u.colr.iccOffset, prop->u.colr.iccSize, &icc);
```
wantehchang, line 3666: Nit: For consistency, I suggest omitting the `/*readFlags=*/` comment here.

src/read.c
jzern: > Yes, `rawOffset` is the offset at which you can find `raw` within the input bitstream. Should it have a different name? I am not so sure I follow what is unclear here.

I was just wondering about the call setting `rawOffset = rawOffset + avifROStreamOffset(&s)` and then adding ` avifROStreamOffset(&s)` again here

jzern (approved):
This should be ok assuming the offsets are correct

src/read.c
wantehchang: James: The value `rawOffset + avifROStreamOffset(&s)` is correct. Note that each of these functions also passes `avifROStreamCurrent(&s)` as the `raw` argument to other functions, and `avifROStreamCurrent(&s)` is equal to the original value of `raw` + `avifROStreamOffset(&s)`.

wantehchang (approved):
LGTM with suggested changes. Good job tracking this memory error down.

src/read.c
jzern: Thanks I wasn't paying attention to the use of `avifROStreamCurrent`.

PR #1104, opened 2022-09-13
Rename CHECK and CHEKERR to AVIF_CHECK

y-guyon:
So that tests including include/avif/internal.h and depending on frameworks already defining CHECK do not trigger -Wmacro-redefined.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -16,17 +16,17 @@ extern "C" {
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
 // Used by stream related things.
-#define CHECK(A)               \
+#define AVIF_CHECK(A)          \
     do {                       \
         if (!(A))              \
             return AVIF_FALSE; \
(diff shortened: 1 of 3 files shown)


vrabaud (approved):

include/avif/internal.h
```diff
@@ -16,17 +16,17 @@ extern "C" {
...
 
 // Used instead of CHECK if needing to return a specific error on failure, instead of AVIF_FALSE
-#define CHECKERR(A, ERR) \
-    do {                 \
-        if (!(A))        \
-            return ERR;  \
+#define AVIF_CHECKERR(A, ERR) \
+    do {                      \
+        if (!(A))             \
+            return ERR;       \
```
wantehchang, line 29: This is an unexpected name collision. Perhaps everything in this header should have the `avif` or `AVIF_` prefix. I took a quick look and it seems that only the three macros at lines 35-38 are missing the prefix. I can take care of these.

PR #1105, opened 2022-09-14
Add AVIF_ prefix to three macros in internal.h

wantehchang:
With this change, everything in internal.h has the "avif" or "AVIF_" prefix. Since we have started to test internal functions declared in internal.h using GoogleTest, some unit test files may include both internal.h and external headers. Using a namespace prefix avoid name conflicts with external headers.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -32,10 +32,10 @@ extern "C" {
 // ---------------------------------------------------------------------------
 // URNs and Content-Types
 
-#define URN_ALPHA0 "urn:mpeg:mpegB:cicp:systems:auxiliary:alpha"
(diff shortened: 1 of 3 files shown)


wantehchang:
Yannis: Please review. I hope I understand the reason for https://github.com/AOMediaCodec/libavif/pull/1104 correctly (aviftilingtest.cc includes both avif/internal.h and gtest/gtest.h, and gtest/gtest.h may include Abseil headers that define `CHECK`).

y-guyon (approved):
> aviftilingtest.cc includes both avif/internal.h and gtest/gtest.h, and gtest/gtest.h may include Abseil headers that define `CHECK`.

Yes.

PR #1106, opened 2022-09-14
Use mktemp instead of /tmp in test_cmd.sh

y-guyon:
It should avoid test file conflicts for multiple testing environments started at the same time.

### DIFF
--- a/tests/test_cmd.sh
+++ b/tests/test_cmd.sh
@@ -34,7 +34,7 @@ fi
 if [[ "$#" -ge 3 ]]; then
   TMP_DIR="$(eval echo "$3")"
 else
-  TMP_DIR=/tmp
+  TMP_DIR="$(mktemp -d)"
 fi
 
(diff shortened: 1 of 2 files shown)


tests/test_cmd.sh
```diff
@@ -34,7 +34,7 @@ fi
 if [[ "$#" -ge 3 ]]; then
   TMP_DIR="$(eval echo "$3")"
```
maryla-uc, line 35: Is this still used in other places or can it be removed?

maryla-uc (approved):

tests/test_cmd.sh
y-guyon: I plan to use it internally.

jzern:
lgtm. Thanks for following up.

wantehchang (commented):
LGTM. Thanks.

PR #1107, opened 2022-09-15
Add a test for issue 1086

wantehchang:
Add a test for https://github.com/AOMediaCodec/libavif/issues/1086 to prevent regression.


y-guyon (approved):

wantehchang:
Here is how I created the test.

The tests/data/paris_icc_exif_xmp.avif file was generated using commit ed52c1b00c4d88169c3060f0a2256da45f39b14a (the commit before the bug fix), with the following command that Yannis mentioned in https://github.com/AOMediaCodec/libavif/issues/1086:

```
./avifenc -s 10 ../tests/data/paris_icc_exif_xmp.png ../tests/data/paris_icc_exif_xmp.avif
```

Without the bug fix (commit d5859e3c65ff1a8d3509b9f8ac38eca88c5a3c09), the test fails only under ASan:

```
tests/avifmetadatatest ../tests/data/
```

The following patch makes the test fail when it is not running under ASan:

```
diff --git a/src/io.c b/src/io.c
index 892726f..7c5be24 100644
--- a/src/io.c
+++ b/src/io.c
@@ -103,6 +103,9 @@ static avifResult avifIOFileReaderRead(struct avifIO * io, uint32_t readFlags, u
             return AVIF_RESULT_IO_ERROR;
         }
         if (reader->buffer.size < size) {
+            if (reader->buffer.size) {
+                memset(reader->buffer.data, 0xdd, reader->buffer.size);
+            }
             avifRWDataRealloc(&reader->buffer, size);
         }
         if (fseek(reader->f, (long)offset, SEEK_SET) != 0) {
```

PR #1108, opened 2022-09-15
Bring back deprecated avifImageRGBToYUV() API

y-guyon:
For backward compatibility and smoother update of projects depending on libavif.

The C API does not allow function overloading so there must be a new name. avifRGBImageToYUV() and avifRGBImageFromYUV() were picked because they are similar to avifRGBImageAllocatePixels(),
avifRGBImagePremultiplyAlpha() etc.

Use this opportunity to refactor argument order and naming.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-There are incompatible ABI changes in this release. The alphaRange member was
-removed from avifImage struct. The chromaDownsampling member was added to the
-avifRGBImage struct. The imageDimensionLimit member was added to the avifDecoder
-struct. avifImageCopy() and avifImageAllocatePlanes() signatures changed. It is
-necessary to recompile your code. Also check the return values of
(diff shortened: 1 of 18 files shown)


wantehchang:
Yannis: It is okay to go back to the old API. But we should delete the new API. We don't need both.

In the old API, the `avifRGBImage` struct plays two roles:

- represents an RGB image, and
- serves as the parameters struct for the two conversion functions.

The second role is less obvious and is the reason I suggested switching to the new API. But once we recognize `avifRGBImage` is also the parameters struct for the two conversion functions, the old API is just fine. The only change we should make is to add an `avifBool preferLibYUV` or `avifBool avoidLibYUV` field to the struct.

Also the `avifRGBImage` struct is a public struct that libavif clients allocate, so when libavif reaches version 1.0.0, we will not be able to change the `avifRGBImage` struct. So this refactoring should try to add all the necessary new fields. But a boolean field that controls whether libyuv should be used is the only new field I can think of now.

y-guyon:
> Yannis: It is okay to go back to the old API. But we should delete the new API. We don't need both.

Done in https://github.com/AOMediaCodec/libavif/pull/1113.

PR #1112, opened 2022-09-16
Youtube image

cabelo:


y-guyon:
Thank you for the suggestion. I do not think it is worth it to add a screenshot of the YouTube player in libavif. I do not even know what is the license of such an image containing YouTube UI elements.

PR #1114, opened 2022-09-16
Non-persistent input in incremental decoding test

y-guyon:


tests/gtest/avifincrtest_helpers.cc
```diff
@@ -118,13 +118,16 @@ uint32_t GetMinDecodedRowCount(uint32_t height, uint32_t cell_height,
 struct PartialData {
   avifROData available;
   size_t full_size;
+
+  // Only used as non persistent input.
+  std::vector<uint8_t> buffer, swapped_buffer;
```
wantehchang, line 123: Optional: This approach works, but it may not catch some errors when running under ASan because the buffer returned by the previous io read call is kept around as valid memory in `swapped_buffer`.

Since we can run this test under ASan, a simpler approach is to declare `std::unique_ptr<uint8_t[]> buffer`, allocate a new `buffer` in each io read call, and flip the bits in the old `buffer` (or memset the old `buffer` to 0 or a value like 0xdd) before freeing the old `buffer`.

wantehchang (approved):
LGTM.

tests/gtest/avifincrtest_helpers.cc
```diff
@@ -269,7 +287,7 @@ void DecodeIncrementally(const avifRWData& encoded_avif, bool is_persistent,
   ASSERT_NE(decoder, nullptr);
   avifDecoderSetIO(decoder.get(), &io);
   decoder->allowIncremental = AVIF_TRUE;
-  const size_t step = std::max(static_cast<size_t>(1), data.full_size / 10000);
```
wantehchang, line 272: Just a side note: I found that the Style Guide also recommends using `size_t{1}` instead of `static_cast<size_t>(1)`.

> Use brace initialization to convert arithmetic types (e.g., `int64_t{x}`). This is the safest approach because code will not compile if conversion can result in information loss. The syntax is also concise.

https://google.github.io/styleguide/cppguide.html#Casting

tests/gtest/avifincrtest_helpers.cc
```diff
@@ -259,7 +279,8 @@ void DecodeIncrementally(const avifRWData& encoded_avif, bool is_persistent,
 
   // Emulate a byte-by-byte stream.
   PartialData data = {/*available=*/{encoded_avif.data, 0},
-                      /*fullSize=*/encoded_avif.size};
+                      /*fullSize=*/encoded_avif.size,
+                      /*buffer=*/{}, /*swapped_buffer=*/{}};
```
wantehchang, line 283: IMPORTANT: The initializer needs to be updated.

wantehchang (changes requested):

tests/gtest/avifincrtest_helpers.cc
wantehchang: I am surprised that this doesn't break compilation. I guess I haven't mastered C++.

jzern: (2) applies here, I believe: https://en.cppreference.com/w/cpp/language/zero_initialization

clang-tidy should catch the incorrect comment though.

jzern: > clang-tidy should catch the incorrect comment though.

Actually maybe that only works on function parameters.

y-guyon: > I am surprised that this doesn't break compilation. I guess I haven't mastered C++.

It is only the comments `/*buffer=*/` and `/*swapped_buffer=*/` that were wrong. `nonpersistent_bytes` was correctly set to `{}` (meaning `nullptr`) and `num_nonpersistent_bytes` to `{}` (meaning `0`).

> clang-tidy should catch the incorrect comment though.

Maybe we could have a clang-tidy test as part of the GitHub CI.

PR #1118, opened 2022-09-19
Avoid setting the ICC profile when given an image in YUV400 format.

yochy4671:
I'm using apps/avifdec and got an error  when converting an avif file in YUV400 pixel format to a png file.
```
Image decoded: 00003.avif
Image details:
 * Resolution     : 1350x1920
 * Bit Depth      : 8
 * Format         : YUV400
 * Alpha          : Absent
 * Range          : Full
 * Color Primaries: 2
 * Transfer Char. : 2
 * Matrix Coeffs. : 6
 * ICC Profile    : Present (408 bytes)
 * XMP Metadata   : Absent
 * Exif Metadata  : Absent
 * Transformations: None
 * Progressive    : Unavailable
libpng error: profile 'libavif': 'GRAY': Gray color space not permitted on RGB PNG
Error writing PNG: 00003.png
```
Modified to avoid this.

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -439,7 +439,7 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
     }
 
     png_set_IHDR(png, info, avif->width, avif->height, rgb.depth, colorType, PNG_INTERLACE_NONE, PNG_COMPRESSION_TYPE_DEFAULT, PNG_FILTER_TYPE_DEFAULT);
-    if (avif->icc.data && (avif->icc.size > 0)) {
+    if (avif->icc.data && (avif->icc.size > 0) && (avif->yuvFormat != AVIF_PIXEL_FORMAT_YUV400)) {
         png_set_iCCP(png, info, "libavif", 0, (png_iccp_datap)avif->icc.data, (png_uint_32)avif->icc.size);
     }
     png_write_info(png, info);


y-guyon:
Thank you for the bug report and suggested change.

> Monochrome images can still have a TRC, so this can't just be discarded. Better create a true grayscale PNG in this case, or convert to RGB.

I agree that a wider fix would be to use `PNG_COLOR_TYPE_GRAY` instead of the current [`PNG_COLOR_TYPE_RGB`](https://github.com/AOMediaCodec/libavif/blob/f1a17c34e5e546328ed3ad9f1542efae5c707950/apps/shared/avifpng.c#L396) for monochrome input.

yochy4671:
I'm trying it like this: However, the correct image was not obtained.

Specifically, the input image and the output image have the same number of vertical and horizontal pixels, but are enlarged horizontally by a factor of 3, and only the left one-third of the image is output.

I guess it's not the input pixel data for a monochrome image that libpng expects.
Do you have any good ideas?

```
    int colorType = PNG_COLOR_TYPE_RGBA;
    if (!avif->alphaPlane) {
        colorType = PNG_COLOR_TYPE_RGB;
    }
    if (avif->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) {
	colorType = PNG_COLOR_TYPE_GLAY;
    }
```

y-guyon:
> I guess it's not the input pixel data for a monochrome image that libpng expects.

Indeed. The current [`avifImageYUVToRGB()`](https://github.com/AOMediaCodec/libavif/blob/f1a17c34e5e546328ed3ad9f1542efae5c707950/apps/shared/avifpng.c#L402) call converts YUV 4:0:0 to RGB which are three interleaved channels with the same values per pixel (gray levels). This is why your image's content is three times wider.

What is needed is an image with a single channel. I do not think the Y channel should be directly used for two reasons: it may not be equivalent to the RGB average of the output of `avifImageYUVToRGB()` (not sure about that) and it would only work for 8-bit, as `avifImage` only supports 8, 10 and 12-bit when generated by libavif. `avifRGBImage` was designed to also support 16-bit.

Unfortunately libavif does not implement single-channel `AVIF_RGB_FORMAT_GRAY` in [`avifRGBFormat`](https://github.com/AOMediaCodec/libavif/blob/1fa47f2149bdde77a3131deb89c061104c2dbcb2/include/avif/avif.h#L548) for `avifRGBImage`.

> Do you have any good ideas?

[My previous answer](https://github.com/AOMediaCodec/libavif/pull/1118#issuecomment-1250777075) was merely suggesting a lead. It does not contain the full needed implementation. I suggest implementing `AVIF_RGB_FORMAT_GRAY` for `avifRGBImage`, but this is a lot of work to make sure it works in all places where `avifRGBFormat` is used.

Whether you do the work yourself or not, I suggest filing the specific bug you have so that the issue can be easily referenced anyway. An image to reproduce the problem would be convenient.

> ```
>     int colorType = PNG_COLOR_TYPE_RGBA;
>     if (!avif->alphaPlane) {
>         colorType = PNG_COLOR_TYPE_RGB;
>         rgb.format = AVIF_RGB_FORMAT_RGB;
>     }
>     if (avif->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) {
> 	colorType = PNG_COLOR_TYPE_GLAY;
> 	rgb.format = AVIF_RGB_FORMAT_RGB;
>     }
> ```

This code does not support `PNG_COLOR_TYPE_GRAY_ALPHA`.

yochy4671:
Understood.
I'm withdrawing this pull request. I fixed it myself, because I wanted to fix this problem quickly.

I'm thinking about creating a new issue as you suggested.

This problem is solved, to the extent of my self-use (and I understand it's not the ideal solution).

I also don't think it's necessary to provide the image data that causes the problem, because I think you fully understand the problem. And this image data is personal, and I lack the know-how to make another new sample.

y-guyon:
> I'm thinking about creating a new issue as you suggested.

Thanks.

> I also don't think it's necessary to provide the image data that causes the problem, because I think you fully understand the problem. And this image data is private, and I lack the know-how to make another new sample.

No worries.

wantehchang:
@yochy4671 @y-guyon Thank you both for the bug report and investigation.

Yannis, we can handle 8-bit AVIF monochrome with no alpha as a special case now. I agree a general solution will require adding `AVIF_RGB_FORMAT_GRAY` and `AVIF_RGB_FORMAT_GRAY_ALPHA`. The special case can still serve as an optimization for the general solution.

wantehchang:
I prototyped my suggestion in https://github.com/AOMediaCodec/libavif/pull/1121. I am not familiar with libpng, so there may be a better way to do this.

@yochy4671: could you give https://github.com/AOMediaCodec/libavif/pull/1121 a try? Thanks.

yochy4671:
@wantehchang: I tried it. The result seems to be good, thanks.

The png file is now an 8bit depth grayscale image, and the file size is about half that of an RGB24bit image.

yochy4671:
This request is withdrawn, I created a new issue.

regards,

PR #1119, opened 2022-09-19
Write 'auxi' box for animated images with alpha channel

vigneshvg:
When alpha channel is present on animated images, the 'auxi' box is mandatory as it tells the decoder what kind of auxiliary track it is.

The current decoder gets the value from the still image item properties. But if a decoder ignores the still image fields and treats the animated AVIF as a video sequence, it has no way of knowing what kind of information the auxiliary track contains.

This patch resolves the following compliance warden error: [heif][Rule #1] Section 7.5.3.1
The nature of the auxiliary track is announced by the AuxiliaryTypeInfoBox that shall be included in the sample entry of the auxiliary track.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -1589,6 +1589,12 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
             avifRWStreamWriteZeros(&s, 3); // unsigned int(26) reserved; (two zero bits are written along with ccstValue).
             avifRWStreamFinishBox(&s, ccst);
 
+            if (item->alpha) {
+                avifBoxMarker auxi = avifRWStreamWriteFullBox(&s, "auxi", AVIF_BOX_SIZE_TBD, 0, 0);
+                avifRWStreamWriteChars(&s, alphaURN, alphaURNSize); //  string aux_track_type;
+                avifRWStreamFinishBox(&s, auxi);
+            }
+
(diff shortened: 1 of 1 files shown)


joedrago:
I believe the implementation of the box write is correct, and it sounds like it it belongs inside of the sample entry alongside `ccst`, so this seems correct. I believe for these additional boxes in a sample entry, there's a recommendation to keep the ordering consistent. Do we have any pre-existing files (or otherwise hints) on whether or not `auxi` should come before or after `ccst`?

vigneshvg:
> I believe for these additional boxes in a sample entry, there's a recommendation to keep the ordering consistent.

I scoured the ISO BMFF Base, MIAF and HEIF specifications for information on ordering. But i could not find anything concrete.

> Do we have any pre-existing files (or otherwise hints) on whether or not auxi should come before or after ccst?

I also could not find any other tools/files that have these boxes.

In the end it doesn't really matter since the reader has enough information to process these boxes in any order.

vigneshvg:
I have also written up a patch for ffmpeg for the same change [1]. I have maintained the same ordering there as well (i.e.) auxi comes after ccst. So we could keep those two in sync.

[1] http://ffmpeg.org/pipermail/ffmpeg-devel/2022-September/301706.html

joedrago (approved):

wantehchang (approved):
LGTM.

PR #1120, opened 2022-09-20
Write 'auxv' as handler_type for alpha channel track

vigneshvg:
For animated AVIF sequences, alpha channel track has to have the handler_type set to 'auxv'.

This is mentioned in Section 7.5.3.1 of the HEIF specification.

libavif's parser ignores the 'hdlr' box that occurs inside the 'mdia' box. So this change is backward compatible (i.e.) files produced after this change will still be parse'able by earlier libavif versions.


joedrago (approved):
Oh nice, I like how simple this change ended up being. I was anticipating a lot of painful plumbing!

wantehchang (approved):
LGTM.

Issue #1125, opened 2022-09-21
New libavif function avifImageExtractExifOrientationToIrotImir() breaks libavif building


BlueSwordM:
#### Describe the bug
Any master build at commit 1fa47f2149bdde77a3131deb89c061104c2dbcb2 and later fail to build

#### To Reproduce

Build libavif regularly.

#### Expected behavior
It should finish building the libavif libraries and binaries.

### Actual behavior
```ld.lld: error: undefined symbol: avifImageExtractExifOrientationToIrotImir
>>> referenced by avifjpeg.c
>>>               avifjpeg.c.o:(avifJPEGRead) in archive libavif_apps.a
>>> referenced by avifpng.c
>>>               avifpng.c.o:(avifExtractExifAndXMP) in archive libavif_apps.a
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [CMakeFiles/avifdec.dir/build.make:104: avifdec] Error 1
make[1]: *** [CMakeFiles/Makefile2:201: CMakeFiles/avifdec.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[100%] Linking CXX executable avifenc
ld.lld: error: undefined symbol: avifImageExtractExifOrientationToIrotImir
>>> referenced by avifenc.c
>>>               CMakeFiles/avifenc.dir/apps/avifenc.c.o:(main)
>>> referenced by avifjpeg.c
>>>               avifjpeg.c.o:(avifJPEGRead) in archive libavif_apps.a
>>> referenced by avifpng.c
>>>               avifpng.c.o:(avifExtractExifAndXMP) in archive libavif_apps.a
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [CMakeFiles/avifenc.dir/build.make:104: avifenc] Error 1
make[1]: *** [CMakeFiles/Makefile2:174: CMakeFiles/avifenc.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

#### Version:
git build from: https://github.com/AOMediaCodec/libavif/commit/1fa47f2149bdde77a3131deb89c061104c2dbcb2

#### Operating system and toolchain
Openmandriva 4.50 ROME Linux 5.19-3
Clang 15.0.0

wantehchang:
[...]

PR #1126, opened 2022-09-21
Export avifImageExtractExifOrientationToIrotImir

wantehchang:
Declare avifImageExtractExifOrientationToIrotImir() with AVIF_API to export the function from the shared library.

Fix https://github.com/AOMediaCodec/libavif/issues/1125.


y-guyon (approved):

y-guyon:
It would be convenient to detect that in the CI too, not sure how though.

y-guyon:
> 2. Add a CI build that builds libavif with `-DBUILD_SHARED_LIBS=ON`.

https://github.com/AOMediaCodec/libavif/pull/1170 is an attempt.

PR #1121, opened 2022-09-20
Use PNG_COLOR_TYPE_GRAY for 8-bit grayscale output

wantehchang:
When writing a PNG output file, if the AVIF image is 8-bit monochrome with no alpha and the PNG output bit depth is 8, use the Y plane of the AVIF image directly for the PNG_COLOR_TYPE_GRAY color type.

PR #1132, opened 2022-09-22
Exclude YUV 4:4:4 from rgb->chromaUpsampling

wantehchang:
Chroma upsampling does not apply to YUV 4:4:4, so the code that depends on rgb->chromaUpsampling can be skipped for YUV 4:4:4.

Also change avifImageYUVToRGBLibYUVHighBitDepth() to have the same structure as avifImageYUVToRGBLibYUV8bpc().

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -560,57 +560,60 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
     } else {
         int uPlaneIndex = isYVU ? AVIF_CHAN_V : AVIF_CHAN_U;
         int vPlaneIndex = isYVU ? AVIF_CHAN_U : AVIF_CHAN_V;
-        // Lookup table for YUV To RGB Matrix (with filter).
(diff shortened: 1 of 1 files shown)


wantehchang:
@jzern Just FYI. I am testing the libavif mirror you set up for us to make code review easier.

@vigneshvg @y-guyon Please review. A copy of this pull request is available at https://aomedia-review.googlesource.com/c/libavif/+/163401. Let's do most of the code review there.

y-guyon (approved):

vigneshvg (approved):

wantehchang:
Thanks for the review. It's not easy to check if libyuv is used in `avifImageYUVToRGB()`. I just tested it exhaustively and manually verified that if the AVIF image has no alpha, `avifdec` uses libyuv for 4:4:4, 4:2:2, and 4:2:0 for all `-u` options now. (For 4:0:0, `avifImageYUVToRGB()` is not called after https://github.com/AOMediaCodec/libavif/pull/1121.)

Issue #1122, opened 2022-09-20
avifenc fails with "XMP extraction failed: extended XMP is unsupported"


wantehchang:
When I ran the `avifenc` command on the attached happy_dog.jpg file (provided by @FrankGalligan), I got an error:

```
./avifenc happy_dog.jpg happy_dog.avif
Directly copied JPEG pixel data (no YUV conversion): happy_dog.jpg
XMP extraction failed: extended XMP is unsupported
Cannot determine input file format: happy_dog.jpg
```

![happy_dog](https://user-images.githubusercontent.com/16024799/191378209-4ac5d242-4665-498f-91ef-598b5d04af04.jpg)

PR #1133, opened 2022-09-23
Read extended XMP in avifjpeg.c

y-guyon:
Add test unit and test image.

Fixes https://github.com/AOMediaCodec/libavif/issues/1122.


apps/shared/avifjpeg.c
```diff
@@ -354,24 +383,134 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+        }
+
+        if (foundExtendedXMP) {
+            // Make sure there is no missing byte.
+            if (memchr(extendedXMPReadBytes.data, AVIF_FALSE, extendedXMPReadBytes.size)) {
+                fprintf(stderr, "XMP extraction failed: missing extended XMP segments\n");
                 goto cleanup;
             }
+
+            // According to Adobe XMP Specification Part 3 section 1.1.3.1:
```
jzern, line 491: Previously I had seen
https://archimedespalimpsest.net/Documents/External/XMP/XMPSpecificationPart3.pdf
where these are in chapter 3. What version are you using?

Adobe's links 404 currently:
  https://github.com/adobe/xmp-docs/blob/master/Specifications.md

apps/shared/avifjpeg.c
```diff
@@ -354,24 +383,134 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+                // According to Adobe XMP Specification Part 3 section 1.1.3.1:
+                //   "full length of the ExtendedXMP serialization as a 32-bit unsigned integer"
+                const uint32_t totalExtendedXMPSize =
+                    avifJPEGReadUint32BigEndian(&marker->data[tagExtendedXMP.size + extendedXMPGUIDLength]);
+                // Offset in totalXMP after standardXMP.
+                // According to Adobe XMP Specification Part 3 section 1.1.3.1:
+                //   "offset of this portion as a 32-bit unsigned integer"
+                const uint32_t extendedXMPOffset =
+                    avifJPEGReadUint32BigEndian(&marker->data[tagExtendedXMP.size + extendedXMPGUIDLength + 4]);
+                if ((uint64_t)standardXMP.size + totalExtendedXMPSize > SIZE_MAX) {
```
wantehchang, line 441: I usually check for potential overflow before adding two numbers. Casting the numbers to a larger integer type also works. You used that method here and at line 445 below.

When used here, it is more work to verify the correctness. The reason is that in 64-bit builds, `standardXMP.size` (a `size_t`) is already the same size as `uint64_t`, so the `(uint64_t)` cast is a no-op. The correctness comes from the fact that the value stored in `standardXMP.size` is less than `marker->data_length`, which is an `unsigned int`. This takes some work to discover.

```
struct jpeg_marker_struct {
  jpeg_saved_marker_ptr next;   /* next in list, or NULL */
  UINT8 marker;                 /* marker code: JPEG_COM, or JPEG_APP0+n */
  unsigned int original_length; /* # bytes of data in the file */
  unsigned int data_length;     /* # bytes of data saved at data[] */
  JOCTET *data;                 /* the data contained in the marker */
  /* the marker length word is not counted in data_length or original_length */
};
```

It would be good to add an assertion to indicate we are depending on this fact:

```
                    assert(standardXMP.size <= UINT32_MAX);
```

Or replace `standardXMP` with a pair of variables:

```
        const uint8_t *standardXMPData = NULL;
        uint32_t standardXMPSize = 0; // At most 64kB as defined by Adobe XMP Specification Part 3.
        for (jpeg_saved_marker_ptr marker = cinfo.marker_list; marker != NULL; marker = marker->next) {
            if ((marker->marker == (JPEG_APP0 + 1)) && (marker->data_length > tagStandardXMP.size) &&
                !memcmp(marker->data, tagStandardXMP.data, tagStandardXMP.size)) {
                if (foundStandardXMP) {
                    fprintf(stderr, "XMP extraction failed: invalid multiple XMP segments\n");
                    goto cleanup;
                }
                standardXMPData = marker->data + tagStandardXMP.size;
                standardXMPSize = marker->data_length - tagStandardXMP.size;
                foundStandardXMP = AVIF_TRUE;
            }
        }
```

wantehchang (changes requested):
Yannis: I only reviewed the pull request partially, so the first round of review comments is incomplete.

apps/shared/avifjpeg.c
y-guyon: > What version are you using?

I do not remember where I got it but it seems to be the "January 2020" version as written in the front page.

> Previously I had seen https://archimedespalimpsest.net/Documents/External/XMP/XMPSpecificationPart3.pdf where these are in chapter 3.

Yes, the section "Embedding XMP metadata in application files" was moved from index 3. to index 1.

> Adobe's links 404 currently

Could this be because it is now an ISO specification behind a paywall? [This one](https://www.iso.org/standard/79384.html) seems related but maybe it is just an extension or something.

apps/shared/avifjpeg.c
```diff
@@ -225,6 +226,26 @@ static avifBool avifJPEGReadCopy(avifImage * avif, struct jpeg_decompress_struct
...
+    for (size_t index = 0; index + substrLength <= strLength; ++index) {
+        if (!memcmp(&str[index], substr, substrLength)) {
+            return &str[index];
+        }
+    }
+    return NULL;
+}
+
+#define AVIF_EXTENDED_XMP_GUID_LENGTH 32
+#define AVIF_XMP_NOTE_TAG_LENGTH 24
```
wantehchang, line 247: Nit/optional: Would be good to note that this is the length of `"xmpNote:HasExtendedXMP=\""`

apps/shared/avifjpeg.c
```diff
@@ -254,6 +275,11 @@ avifBool avifJPEGRead(const char * inputFilename,
     avifRGBImage rgb;
     memset(&rgb, 0, sizeof(avifRGBImage));
 
+    // Standard XMP segment followed by all extended XMP segments.
+    avifRWData totalXMP = { NULL, 0 };
+    // Each byte set to AVIF_FALSE is a missing byte. Each byte set to AVIF_TRUE was read and copied to totalXMP.
```
wantehchang, line 280: Nit: 0 and 1 (as opposed to AVIF_FALSE and AVIF_TRUE) are used in line 463 and line 477.

apps/shared/avifjpeg.c
```diff
@@ -354,24 +380,133 @@ avifBool avifJPEGRead(const char * inputFilename,
         }
     }
     if (!ignoreXMP) {
-        const avifROData tagStandardXmp = { (const uint8_t *)"http://ns.adobe.com/xap/1.0/\0", 29 };
-        const avifROData tagExtendedXmp = { (const uint8_t *)"http://ns.adobe.com/xmp/extension/\0", 35 };
-        avifBool found = AVIF_FALSE;
+        const avifROData tagStandardXMP = { (const uint8_t *)"http://ns.adobe.com/xap/1.0/\0", 29 };
```
wantehchang, line 383: Nit/optional: `\0` can be omitted in this string and in the string in line 400. We can use the terminating NUL bytes in string literals.

apps/shared/avifjpeg.c
```diff
@@ -354,24 +380,133 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+        avifBool foundStandardXMP = AVIF_FALSE;
```
wantehchang, line 384: Nit/optional: `foundStandardXMP` can be omitted. We can just test if `standardXMPData` is non-null.

apps/shared/avifjpeg.c
```diff
@@ -354,24 +380,133 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+        uint32_t standardXMPSize = 0; // At most 64kB as defined by Adobe XMP Specification Part 3.
         for (jpeg_saved_marker_ptr marker = cinfo.marker_list; marker != NULL; marker = marker->next) {
-            if ((marker->marker == (JPEG_APP0 + 1)) && (marker->data_length > tagStandardXmp.size) &&
-                !memcmp(marker->data, tagStandardXmp.data, tagStandardXmp.size)) {
-                if (found) {
-                    fprintf(stderr, "XMP extraction failed: invalid multiple XMP chunks\n");
+            if ((marker->marker == (JPEG_APP0 + 1)) && (marker->data_length > tagStandardXMP.size) &&
+                !memcmp(marker->data, tagStandardXMP.data, tagStandardXMP.size)) {
+                if (foundStandardXMP) {
+                    fprintf(stderr, "XMP extraction failed: invalid multiple XMP segments\n");
```
wantehchang, line 391: Nit: Add "standard" to this error message? It is fine to have multiple XMP segments.

apps/shared/avifjpeg.c
```diff
@@ -354,24 +380,133 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+        for (jpeg_saved_marker_ptr marker = cinfo.marker_list; marker != NULL; marker = marker->next) {
+            if ((marker->marker == (JPEG_APP0 + 1)) && (marker->data_length > tagExtendedXMP.size) &&
+                !memcmp(marker->data, tagExtendedXMP.data, tagExtendedXMP.size)) {
+                if (!foundStandardXMP) {
+                    fprintf(stderr, "XMP extraction failed: extended XMP segment found, missing standard XMP segment\n");
+                    goto cleanup;
+                }
+
+                // Offset in marker->data segment (skip tag + guid + size + offset).
+                const size_t offsetTillExtendedXMP = tagExtendedXMP.size + AVIF_EXTENDED_XMP_GUID_LENGTH + 4 + 4;
```
wantehchang, line 412: Nit/optional: This will a compile-time constant if we replace `tagExtendedXMP.size` with 35.

apps/shared/avifjpeg.c
```diff
@@ -354,24 +380,133 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+                if ((extendedXMPSize == 0) || ((uint64_t)extendedXMPOffset + extendedXMPSize > totalExtendedXMPSize)) {
+                    fprintf(stderr, "XMP extraction failed: invalid extended XMP segment size or offset\n");
+                    goto cleanup;
+                }
+                if (foundExtendedXMP) {
+                    if (memcmp(guid, extendedXMPGUID, AVIF_EXTENDED_XMP_GUID_LENGTH)) {
+                        fprintf(stderr, "XMP extraction failed: extended XMP segment GUID mismatch\n");
+                        goto cleanup;
+                    }
+                    if (totalExtendedXMPSize != totalXMP.size) {
```
wantehchang, line 451: IMPORTANT: This should read

```
                   if (totalExtendedXMPSize != totalXMP.size - standardXMPSize) {
```

See line 458.

apps/shared/avifjpeg.c
```diff
@@ -354,24 +380,133 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+                        fprintf(stderr, "XMP extraction failed: extended XMP total size mismatch\n");
+                        goto cleanup;
+                    }
+                } else {
+                    memcpy(extendedXMPGUID, guid, AVIF_EXTENDED_XMP_GUID_LENGTH);
+
+                    avifRWDataRealloc(&totalXMP, standardXMPSize + totalExtendedXMPSize);
```
wantehchang, line 458: IMPORTANT: Cast `standardXMPSize` to `size_t`. The cast is necessary now that we declare `standardXMPSize` as `uint32_t`.

wantehchang (approved):
LGTM with changes.

The suggested changes marked with "IMPORTANT" are bug fixes.

The suggested changes marked with "optional" are left to your discretion.

Note: I created a copy of this pull request at https://aomedia-review.googlesource.com/c/libavif/+/164002 to help me review this, but I believe I could have reviewed this directly in GitHub.

apps/shared/avifjpeg.c
```diff
@@ -354,24 +380,133 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+                    avifJPEGReadUint32BigEndian(&marker->data[tagExtendedXMP.size + AVIF_EXTENDED_XMP_GUID_LENGTH + 4]);
+                if ((uint64_t)standardXMPSize + totalExtendedXMPSize > SIZE_MAX) {
+                    fprintf(stderr, "XMP extraction failed: total XMP size is too high\n");
+                    goto cleanup;
+                }
+                if ((extendedXMPSize == 0) || ((uint64_t)extendedXMPOffset + extendedXMPSize > totalExtendedXMPSize)) {
+                    fprintf(stderr, "XMP extraction failed: invalid extended XMP segment size or offset\n");
+                    goto cleanup;
+                }
+                if (foundExtendedXMP) {
```
wantehchang, line 446: Optional: To test this code path, we need to find a JPEG image with more than one extended XMP segment.

apps/shared/avifjpeg.c
```diff
@@ -354,24 +380,133 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+                //   "full length of the ExtendedXMP serialization as a 32-bit unsigned integer"
+                const uint32_t totalExtendedXMPSize =
+                    avifJPEGReadUint32BigEndian(&marker->data[tagExtendedXMP.size + AVIF_EXTENDED_XMP_GUID_LENGTH]);
+                // Offset in totalXMP after standardXMP.
+                // According to Adobe XMP Specification Part 3 section 1.1.3.1:
+                //   "offset of this portion as a 32-bit unsigned integer"
+                const uint32_t extendedXMPOffset =
+                    avifJPEGReadUint32BigEndian(&marker->data[tagExtendedXMP.size + AVIF_EXTENDED_XMP_GUID_LENGTH + 4]);
+                if ((uint64_t)standardXMPSize + totalExtendedXMPSize > SIZE_MAX) {
+                    fprintf(stderr, "XMP extraction failed: total XMP size is too high\n");
```
jzern, line 439: '...too large'?

apps/shared/avifjpeg.c
jzern: It's possible the standardization could be the cause. Thanks for pointing me to your version.

apps/shared/avifjpeg.c
```diff
@@ -354,24 +380,133 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+            memcpy(xmpNote + AVIF_XMP_NOTE_TAG_LENGTH, extendedXMPGUID, AVIF_EXTENDED_XMP_GUID_LENGTH);
+            if (!avifJPEGFindSubstr(standardXMPData, standardXMPSize, xmpNote, sizeof(xmpNote))) {
+                fprintf(stderr, "XMP extraction failed: standard and extended XMP GUID mismatch\n");
+                goto cleanup;
+            }
+
+            // According to Adobe XMP Specification Part 3 section 1.1.3.1:
+            //   "A JPEG reader must [...] remove the xmpNote:HasExtendedXMP property."
+            // This constraint is ignored here because leaving the xmpNote:HasExtendedXMP property is rather harmless
+            // and editing XMP metadata is quite involving.
```
jzern, line 502: s/involving/involved/

apps/shared/avifjpeg.c
y-guyon: I moved the string next to the size to make it clear. That was my original intent: keep the string, the size and where it is used close.

apps/shared/avifjpeg.c
y-guyon: I believe this is clearer this way, without a comment mentioning that the hard-coded length includes the implicit `\0`.

apps/shared/avifjpeg.c
y-guyon: I could not find one so I crafted one.

apps/shared/avifjpeg.c
wantehchang: I forgot to mention that this value is loop invariant. We can move it to line 405, outside the for loop.

Note: You should be able to define the constant as an enum in the middle of a function:

```
    enum {
        TAG_EXTENDED_XMP_SIZE = 35
    };
```

apps/shared/avifjpeg.c
```diff
@@ -354,24 +385,135 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+        if (foundExtendedXMP) {
+            // Make sure there is no missing byte.
+            if (memchr(extendedXMPReadBytes.data, 0, extendedXMPReadBytes.size)) {
+                fprintf(stderr, "XMP extraction failed: missing extended XMP segments\n");
                 goto cleanup;
             }
+
+            // According to Adobe XMP Specification Part 3 section 1.1.3.1:
+            //   "A reader must incorporate only ExtendedXMP blocks whose GUID matches the value of xmpNote:HasExtendedXMP."
+            uint8_t xmpNote[AVIF_XMP_NOTE_TAG_LENGTH + AVIF_EXTENDED_XMP_GUID_LENGTH];
```
wantehchang, line 493: IMPORTANT: This array is now also used for `AVIF_ALTERNATIVE_XMP_NOTE_TAG` (see lines 498-499). It happens to be big enough because both xmpNote tags have the same length (24).

We should add a comment, or define a new macro `AVIF_XMP_NOTE_TAG_LENGTH_MAX` and use it in this line. There may be a better solution.

wantehchang (approved):
Still LGTM. Thanks. The `xmpNote` array is now used for two xmpNote tags, which could potentially have different lengths.

apps/shared/avifjpeg.c
```diff
@@ -354,24 +385,135 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+                // Offset in totalXMP after standardXMP.
+                // According to Adobe XMP Specification Part 3 section 1.1.3.1:
+                //   "offset of this portion as a 32-bit unsigned integer"
+                const uint32_t extendedXMPOffset =
+                    avifJPEGReadUint32BigEndian(&marker->data[tagExtendedXMP.size + AVIF_EXTENDED_XMP_GUID_LENGTH + 4]);
+                if ((uint64_t)standardXMPSize + totalExtendedXMPSize > SIZE_MAX) {
+                    fprintf(stderr, "XMP extraction failed: total XMP size is too large\n");
+                    goto cleanup;
+                }
+                if ((extendedXMPSize == 0) || ((uint64_t)extendedXMPOffset + extendedXMPSize > totalExtendedXMPSize)) {
```
jzern, line 445: is it more common to group the subexpressions here? `(((uint64_t)extendedXMPOffset + extendedXMPSize) > totalExtendedXMPSize)`

apps/shared/avifjpeg.c
```diff
@@ -354,24 +385,135 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+                // According to Adobe XMP Specification Part 3 section 1.1.3.1:
+                //   "full length of the ExtendedXMP serialization as a 32-bit unsigned integer"
+                const uint32_t totalExtendedXMPSize =
+                    avifJPEGReadUint32BigEndian(&marker->data[tagExtendedXMP.size + AVIF_EXTENDED_XMP_GUID_LENGTH]);
+                // Offset in totalXMP after standardXMP.
+                // According to Adobe XMP Specification Part 3 section 1.1.3.1:
+                //   "offset of this portion as a 32-bit unsigned integer"
+                const uint32_t extendedXMPOffset =
+                    avifJPEGReadUint32BigEndian(&marker->data[tagExtendedXMP.size + AVIF_EXTENDED_XMP_GUID_LENGTH + 4]);
+                if ((uint64_t)standardXMPSize + totalExtendedXMPSize > SIZE_MAX) {
```
jzern, line 441: is it more common to group the subexpressions here? `((uint64_t)standardXMPSize + totalExtendedXMPSize)`

tests/data/README.md
```diff
@@ -50,6 +69,32 @@ Source: Personal photo edited with Gimp 2.10.
...
+|---------:|-------------|-------:|----------------------------------|
+|        0 | 0xffd8 SOI  |        |                                  |
+|        2 | 0xffe0 APP0 |     16 | `JFIF.....,.,.`                  |
+|       20 | 0xffe1 APP1 |   5531 | http://ns.adobe.com/xap/1.0/.<?x |
+|     5553 | 0xffe1 APP1 |  65535 | http://ns.adobe.com/xmp/extensio |
+|    71090 | 0xffe1 APP1 |  65535 | http://ns.adobe.com/xmp/extensio |
+|   136627 | 0xffe1 APP1 |   4791 | http://ns.adobe.com/xmp/extensio |
+|          |             |        | ...                              |
+
+https://en.wikipedia.org/wiki/De_finibus_bonorum_et_malorum
```
jzern, line 96: is this leftover since the link is given in the above paragraph?

jzern (approved):

tests/data/README.md
y-guyon: Good catch.

apps/shared/avifjpeg.c
y-guyon: A lot are missing in this file, see all the `cinfo->comp_info[0].h_samp_factor == 1` conditions at line 135 for example.

apps/shared/avifjpeg.c
y-guyon: I find it very verbose but the main issue is that I forget it indeed. Also it is hard to remember exactly when to put `()`, for example:

```c
int a = b + c; // int a = (b + c) ?
if (a == (b + c)) {}
```

apps/shared/avifjpeg.c
y-guyon: > I forgot to mention that this value is loop invariant. We can move it to line 405, outside the for loop.

I wanted to keep it close to where it is used.

> Note: You should be able to define the constant as an enum in the middle of a function:
> 
> ```
>     enum {
>         TAG_EXTENDED_XMP_SIZE = 35
>     };
> ```

I do not see any other use of this pattern in libavif.

---

Since it is moved out the loop anyway, a `#define` will do it.

PR #1135, opened 2022-09-24
Replace repeated subtraction by modulo in calcGCD

wantehchang:
calcGCD uses repeated subtraction as in the original Euclid's algorithm. But repeated subtraction takes a long time for inputs such as calcGCD(1937072755, 1) and calcGCD(1937073715, 1).

Bug: b/246649620

### DIFF
--- a/apps/shared/avifutil.c
+++ b/apps/shared/avifutil.c
@@ -3,7 +3,6 @@
 
 #include "avifutil.h"
 
-#include <assert.h>
 #include <ctype.h>
 #include <stdio.h>
 #include <string.h>
@@ -16,22 +15,18 @@
 // overflowing int32_t.
 static int64_t calcGCD(int64_t a, int64_t b)
 {
-    assert(b != 0);
     if (a < 0) {
         a *= -1;
     }
     if (b < 0) {
         b *= -1;
     }
-    for (;;) {
+    while (b != 0) {
         int64_t r = a % b;
-        if (r == 0) {
-            break;
-        }
         a = b;
         b = r;
     }
-    return b;
+    return a;
 }
 
 static void printClapFraction(const char * name, int32_t n, int32_t d)
--- a/apps/shared/avifutil.c
+++ b/apps/shared/avifutil.c
@@ -3,6 +3,7 @@
 
 #include "avifutil.h"
 
+#include <assert.h>
 #include <ctype.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


apps/shared/avifutil.c
```diff
@@ -11,32 +12,35 @@
 #include "avifpng.h"
 #include "y4m.h"
 
-static int32_t calcGCD(int32_t a, int32_t b)
+// |a| and |b| hold int32_t values. The int64_t type is used so that we can negate INT_MIN without
+// overflowing int32_t.
+static int64_t calcGCD(int64_t a, int64_t b)
 {
+    assert(b != 0);
```
joedrago, line 19: The actual math here is perfectly reasonable and solid.

About the assert though:

Based on the bug we discussed earlier, we know that the current implementation could allow for one of these values (from user input) to be zero. Instead of asserting here, perhaps just return `0` here and document returning `0` as a failure due to illegal inputs?

We really only call `calcGCD()` in two places in the whole codebase right now. I feel like it'd be better to protect against zeros at those call sites and fail properly instead of just hitting this assert (which would abort/crash).

joedrago: Per our discussion just now, I see that you've already verified that we protect against zeros in the denominators before this function is ever invoked. This change looks good.

joedrago:
LGTM.

apps/shared/avifutil.c
wantehchang: Done. I checked both files again and figured out a way to remove the assertion.

Now `calcGCD(0, 0)` returns 0, `calcGCD(a, 0)` returns `a`, and `calcGCD(0, b)` returns `b`.

jzern (approved):
libvpx does something similar. There are other options, but this is simple enough.

https://en.wikipedia.org/wiki/Binary_GCD_algorithm

wantehchang:
James: Thanks for the review and the pointer to the `gcd()` function in libvpx. I also found two `gcd()` functions in libaom.

apps/shared/avifutil.c
wantehchang: > Per our discussion just now, I see that you've already verified that we protect against zeros in the denominators before this function is ever invoked. This change looks good.

Just to clarify: I only verified that we protect against zeros in the denominators before this function is ever invoked in src/avif.c. This is not the case in apps/shared/avifutil.c, so we should not add the assertion in that file. In any case, the second commit in this pull request made this issue moot.

joedrago (approved):

PR #1138, opened 2022-09-28
Add quality and qualityAlpha to avifEncoder

wantehchang:
Change the default of minQuantizer and minQuantizerAlpha to
AVIF_QUANTIZER_BEST_QUALITY (0). Change the default of maxQuantizer and
maxQuantizerAlpha to AVIF_QUANTIZER_WORST_QUALITY (63).

The default of quality is based on the average of minQuantizer and
maxQuantizer. Similarly the default of qualityAlpha is based on the
average of minQuantizerAlpha and maxQuantizerAlpha.

Map quality and qualityAlpha in the range of 0..100 linearly to
quantizer values (QP) in the range of 63..0.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -147,7 +147,7 @@ static void syntax(void)
 // This is *very* arbitrary, I just want to set people's expectations a bit
 static const char * qualityString(int quality)
 {
-    if (quality == 100) {
+    if (quality == AVIF_QUALITY_LOSSLESS) {
         return "Lossless";
     }
     if (quality >= 80) {
@@ -436,6 +436,7 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
     return AVIF_TRUE;
 }
 
+#define INVALID_QUALITY -1
 #define DEFAULT_QUALITY 60 // Maps to a quantizer (QP) of 25.
 #define DEFAULT_QUALITY_ALPHA AVIF_QUALITY_LOSSLESS
 
@@ -458,8 +459,8 @@ int main(int argc, char * argv[])
 
     int returnCode = 0;
     int jobs = 1;
-    int quality = -1;
-    int qualityAlpha = -1;
+    int quality = INVALID_QUALITY;
+    int qualityAlpha = INVALID_QUALITY;
     int minQuantizer = -1;
     int maxQuantizer = -1;
     int minQuantizerAlpha = -1;
@@ -873,7 +874,8 @@ int main(int argc, char * argv[])
         // Don't subsample when using AVIF_MATRIX_COEFFICIENTS_IDENTITY.
         input.requestedFormat = AVIF_PIXEL_FORMAT_YUV444;
         // Quality.
-        if ((quality >= 0 && quality != AVIF_QUALITY_LOSSLESS) || (qualityAlpha >= 0 && qualityAlpha != AVIF_QUALITY_LOSSLESS)) {
+        if ((quality != INVALID_QUALITY && quality != AVIF_QUALITY_LOSSLESS) ||
+            (qualityAlpha != INVALID_QUALITY && qualityAlpha != AVIF_QUALITY_LOSSLESS)) {
             fprintf(stderr, "Quality cannot be set in lossless mode, except to %d.\n", AVIF_QUALITY_LOSSLESS);
             returnCode = 1;
         }
@@ -909,30 +911,30 @@ int main(int argc, char * argv[])
             goto cleanup;
     } else {
         // Set lossy defaults.
-        if (minQuantizer < 0) {
-            assert(maxQuantizer < 0);
-            if (quality < 0) {
+        if (minQuantizer == -1) {
+            assert(maxQuantizer == -1);
+            if (quality == INVALID_QUALITY) {
                 quality = DEFAULT_QUALITY;
             }
             minQuantizer = AVIF_QUANTIZER_BEST_QUALITY;
             maxQuantizer = AVIF_QUANTIZER_WORST_QUALITY;
         } else {
-            assert(maxQuantizer >= 0);
-            if (quality < 0) {
+            assert(maxQuantizer != -1);
+            if (quality == INVALID_QUALITY) {
                 const int quantizer = (minQuantizer + maxQuantizer) / 2;
                 quality = ((63 - quantizer) * 100 + 31) / 63;
             }
         }
-        if (minQuantizerAlpha < 0) {
-            assert(maxQuantizerAlpha < 0);
-            if (qualityAlpha < 0) {
+        if (minQuantizerAlpha == -1) {
+            assert(maxQuantizerAlpha == -1);
+            if (qualityAlpha == INVALID_QUALITY) {
                 qualityAlpha = DEFAULT_QUALITY_ALPHA;
             }
             minQuantizerAlpha = AVIF_QUANTIZER_BEST_QUALITY;
             maxQuantizerAlpha = AVIF_QUANTIZER_WORST_QUALITY;
         } else {
-            assert(maxQuantizerAlpha >= 0);
-            if (qualityAlpha < 0) {
+            assert(maxQuantizerAlpha != -1);
+            if (qualityAlpha == INVALID_QUALITY) {
                 const int quantizerAlpha = (minQuantizerAlpha + maxQuantizerAlpha) / 2;
                 qualityAlpha = ((63 - quantizerAlpha) * 100 + 31) / 63;
             }
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -66,22 +66,14 @@ static void syntax(void)
     printf("                                        M = matrix coefficients\n");
     printf("                                        (use 2 for any you wish to leave unspecified)\n");
     printf("    -r,--range RANGE                  : YUV range [limited or l, full or f]. (JPEG/PNG only, default: full; For y4m or stdin, range is retained)\n");
-    printf("    --min Q                           : Set min quantizer for color (%d-%d, where %d is lossless)\n",
-           AVIF_QUANTIZER_BEST_QUALITY,
-           AVIF_QUANTIZER_WORST_QUALITY,
-           AVIF_QUANTIZER_LOSSLESS);
-    printf("    --max Q                           : Set max quantizer for color (%d-%d, where %d is lossless)\n",
-           AVIF_QUANTIZER_BEST_QUALITY,
-           AVIF_QUANTIZER_WORST_QUALITY,
-           AVIF_QUANTIZER_LOSSLESS);
-    printf("    --minalpha Q                      : Set min quantizer for alpha (%d-%d, where %d is lossless)\n",
-           AVIF_QUANTIZER_BEST_QUALITY,
-           AVIF_QUANTIZER_WORST_QUALITY,
-           AVIF_QUANTIZER_LOSSLESS);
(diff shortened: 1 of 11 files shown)


jzern (commented):
Changelog entry?

apps/avifenc.c
```diff
@@ -66,22 +66,14 @@ static void syntax(void)
...
-           AVIF_QUANTIZER_LOSSLESS);
-    printf("    --minalpha Q                      : Set min quantizer for alpha (%d-%d, where %d is lossless)\n",
-           AVIF_QUANTIZER_BEST_QUALITY,
-           AVIF_QUANTIZER_WORST_QUALITY,
-           AVIF_QUANTIZER_LOSSLESS);
-    printf("    --maxalpha Q                      : Set max quantizer for alpha (%d-%d, where %d is lossless)\n",
-           AVIF_QUANTIZER_BEST_QUALITY,
-           AVIF_QUANTIZER_WORST_QUALITY,
-           AVIF_QUANTIZER_LOSSLESS);
+    printf("    --q Q                             : Set quality for color (%d-%d, where %d is lossless)\n",
```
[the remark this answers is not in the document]
wantehchang: `--q` does not override `--min` & `--max`, but rather it changes the default values of `--min` & `--max`. Since the help message printed by `syntax()` does not document the default values of `--min` & `--max`, I did not describe this behavior.

Do you think I should?

apps/avifenc.c
```diff
@@ -562,6 +575,24 @@ int main(int argc, char * argv[])
         } else if (!strcmp(arg, "-k") || !strcmp(arg, "--keyframe")) {
             NEXTARG();
             keyframeInterval = atoi(arg);
+        } else if (!strcmp(arg, "--q")) {
```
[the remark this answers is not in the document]
wantehchang: I also noticed this issue today. I've changed `--q` to `-q`.

Should we provide a long name for this option? If so, do you prefer `--qcolor` or `--quality`?

wantehchang (commented):
James: Thanks for the review (even though this pull request isn't ready for review yet).

I will add a changelog entry.

apps/avifenc.c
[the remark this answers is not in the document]
wantehchang: James: I just realized my previous comment is wrong. Please ignore it. Here is the correct reply.

`--q` does not override `--min` & `--max`. But the default value of `--q` depends on whether `--min` & `--max` are specified.

Since the help message (printed by `syntax()`) does not document the default values of `--min` & `--max`, it does not document the default value of `--q` either.

Do you think it should?

apps/avifenc.c
```diff
@@ -827,6 +858,12 @@ int main(int argc, char * argv[])
         }
         // Don't subsample when using AVIF_MATRIX_COEFFICIENTS_IDENTITY.
         input.requestedFormat = AVIF_PIXEL_FORMAT_YUV444;
+        // Quality.
+        if ((quality >= 0 && quality != AVIF_QUALITY_LOSSLESS) || (qualityAlpha >= 0 && qualityAlpha != AVIF_QUALITY_LOSSLESS)) {
+            fprintf(stderr, "Quality cannot be set in lossless mode, except to 100.\n");
```
vrabaud, line 863: replace 100 with %d and AVIF_QUALITY_LOSSLESS

apps/avifenc.c
```diff
@@ -1052,17 +1105,17 @@ int main(int argc, char * argv[])
             lossless = AVIF_FALSE;
         }
 
-        if (!losslessColorQP) {
+        if (!losslessColorQuality) {
             fprintf(stderr,
-                    "WARNING: [--lossless] Color quantizer range (--min, --max) not set to %d. Color output might not be lossless.\n",
-                    AVIF_QUANTIZER_LOSSLESS);
+                    "WARNING: [--lossless] Color quality (--q) not set to %d. Color output might not be lossless.\n",
```
vrabaud, line 1110: -q then ?

apps/avifenc.c
```diff
@@ -562,6 +575,24 @@ int main(int argc, char * argv[])
         } else if (!strcmp(arg, "-k") || !strcmp(arg, "--keyframe")) {
             NEXTARG();
             keyframeInterval = atoi(arg);
+        } else if (!strcmp(arg, "--q")) {
+            NEXTARG();
+            quality = atoi(arg);
+            if (quality < AVIF_QUALITY_WORST) {
```
vrabaud, line 581: maybe use AVIF_CLAMP for shorter code.

wantehchang: AVIF_CLAMP, AVIF_MIN, and AVIF_MAX are internal macros. This is why we can't use them in the apps/ directory, unfortunately.

We can certainly define equivalent macros in some header in apps/shared/ and use them here.

apps/avifenc.c
wantehchang: Done.

apps/avifenc.c
wantehchang: Done. I changed `--q` to `-q or --qcolor`.

Note: The long name for `-q` can be `--qcolor` or `--quality`. Which one do you prefer?

wantehchang (commented):
Vincent: Thanks a lot for the review. This pull request is not ready for review yet. apps/avifenc.c is ready, but the other files, in particular src/write.c and src/codec_xxx.c, are not.

apps/avifenc.c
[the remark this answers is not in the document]
jzern: `--qcolor` makes sense, I think there are other references to color & alpha.

wantehchang:
This pull request is ready for review. I created a copy of this pull request in the mirror:
https://aomedia-review.googlesource.com/c/libavif/+/166701

Let's do most of the code review in the mirror using Gerrit. Thanks.

wantehchang:
@FrankGalligan @jzern @y-guyon Please review. Thanks.

What you reviewed at https://aomedia-review.googlesource.com/c/libavif/+/166701 (patch set 2) corresponds to the second commit (titled "Address review comments") in this pull request.

I made some changes after your last review in the third commit (titled "Clamp quantizer in src/write.c") in this pull request.

jzern (approved):

joedrago:
I am quite late to this PR, but do we really just want a single generic quality value range of 100 to linearly map to the quantizer? Do we not want to fit some sort of curve so that the quality values feel distinct, or perhaps at least similar to other 100-range quality values in other formats? 

I'm only asking because "quantizer value" of 63-0 doesn't really have strong expectations for end-users in what it would mean necessarily, but a _quality_ value that goes 1-100 will. Do we not want to try to find some kind of curve to fit on this that makes each quality value distinct, and maybe linearizes for output payload size or PSNR or something?

I'm just expecting a sea of confusion from endusers when they use quality of 90, 80, 70, etc and it doesn't correspond in any way to some other metric.

y-guyon (approved):

apps/avifenc.c
```diff
@@ -1044,18 +1099,18 @@ int main(int argc, char * argv[])
         usingAOM = AVIF_TRUE;
     }
     avifBool hasAlpha = (image->alphaPlane && image->alphaRowBytes);
-    avifBool losslessColorQP = (minQuantizer == AVIF_QUANTIZER_LOSSLESS) && (maxQuantizer == AVIF_QUANTIZER_LOSSLESS);
-    avifBool losslessAlphaQP = (minQuantizerAlpha == AVIF_QUANTIZER_LOSSLESS) && (maxQuantizerAlpha == AVIF_QUANTIZER_LOSSLESS);
+    avifBool losslessColorQuality = (quality == AVIF_QUALITY_LOSSLESS);
```
vrabaud, line 1102: maybe rename to: usingLosslessColor or hasLosslessColor to be consistent with the naming around.

wantehchang: Done. Changed the variable names to `usingLosslessColor` and `usingLosslessAlpha`. Ideally the variable names should contain the word "quality", but the names would be too verbose.

src/write.c
```diff
@@ -811,6 +828,21 @@ static avifImage * avifImageCopyAndPad(const avifImage * srcImage, uint32_t dstW
...
 }
 
+static int avifQualityToQuantizer(int quality, int minQuantizer, int maxQuantizer)
+{
+    int quantizer;
+    if (quality == AVIF_QUALITY_DEFAULT) {
+        // In older libavif releases, avifEncoder didn't have the quality and qualityAlpha fields.
+        // Supply a default value for quantizer.
+        quantizer = (minQuantizer + maxQuantizer) / 2;
+        quantizer = AVIF_CLAMP(quantizer, 0, 63);
```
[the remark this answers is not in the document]
wantehchang: This is a tough call. The `AVIF_CLAMP` calls are easy to deal with. For example,

```
        quantizer = AVIF_CLAMP(quantizer, AVIF_QUANTIZER_BEST_QUALITY, AVIF_QUANTIZER_WORST_QUALITY);
```

The difficulty is in the conversion formula in line 841:

```
```

The 63 in the formula is actually `(AVIF_QUANTIZER_WORST_QUALITY - AVIF_QUANTIZER_BEST_QUALITY)`.

And we would need to add a `+ AVIF_QUANTIZER_BEST_QUALITY` term (which expands to + 0 and can be omitted).

So it would look like:

```
        quantizer = ((100 - quality) * (AVIF_QUANTIZER_WORST_QUALITY - AVIF_QUANTIZER_BEST_QUALITY) + 50) / 100
                    + AVIF_QUANTIZER_BEST_QUALITY;
```

And we would also need to deal with the 100 and 50 in the formula.

wantehchang (commented):
Vincent: Thanks for the review.

wantehchang:
Joe Drago wrote:
> I am quite late to this PR, but do we really just want a single generic quality value range of 100 to linearly map to the quantizer? Do we not want to fit some sort of curve so that the quality values feel distinct, or perhaps at least similar to other 100-range quality values in other formats?
> 
> I'm only asking because "quantizer value" of 63-0 doesn't really have strong expectations for end-users in what it would mean necessarily, but a _quality_ value that goes 1-100 will. Do we not want to try to find some kind of curve to fit on this that makes each quality value distinct, and maybe linearizes for output payload size or PSNR or something?
> 
> I'm just expecting a sea of confusion from endusers when they use quality of 90, 80, 70, etc and it doesn't correspond in any way to some other metric.

Joe, thank you for your comment (which you also gave to me in person before).

The quality value implemented in this pull request is indeed just the quantizer/QP value in disguise. It is more user friendly than the quantizer value because the range is the more familiar 100 and the two extreme values (0 and 100) correspond to worst and best qualities intuitively.

It does have some drawbacks:

- Its relation to image quality is indirect (through the quantizer value) and incomplete (because other encoder options, in particular the speed preset, also affect image quality somewhat).
- It is on a different scale from JPEG's quality value, even though the range is the same (0..100). I.e., an AVIF image encoded with quality=80 and a JPEG image encoded with quality=80 are likely to have different image qualities. I think end users will learn to expect the scales to be different.

I wonder if there is a way to be more honest about the "quantizer in disguise" nature of this quality value. Perhaps in documentation and syntax help message we can refer to it as the _quantizer quality_ or _quantization quality_?

wantehchang:
I looked at how ImageMagick handles its quality paremeter when it encodes AVIF images (using libheif).

libheif maps ImageMagick's quality value (1..100) to the quantizer/QP/CRF value of the HEIF (libx265) or AVIF (libaom, rav1e, and SVT-AV1) encoder linearly. So the approach of this pull request matches what libheif does.

PR #1143, opened 2022-10-03
Allow clamped cells in avifEncoderAddImageGrid()

y-guyon:
The right-most column and bottom-most row of cells given to avifEncoderAddImageGrid() can be smaller than the others. This will allow for arbitrarily sized images to be encoded as grids. avifEncoderAddImageInternal() will pad those truncated cells with border samples values because all AVIF cells must have the same dimensions according to HEIF.
(Not to be confused with AV1 tiles.)

Refactor avifgridapitest.cc.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -178,6 +178,9 @@ typedef struct avifEncoderItem
     uint32_t gridCols; // if non-zero (legal range [1-256]), this is a grid item
     uint32_t gridRows; // if non-zero (legal range [1-256]), this is a grid item
 
+    uint32_t outputWidth;  // displayed dimensions of the item
+    uint32_t outputHeight; // (only present on type==av01)
+
     uint16_t dimgFromID; // if non-zero, make an iref from dimgFromID -> this id
 
     struct ipmaArray ipma;
@@ -653,7 +656,7 @@ static void avifEncoderWriteTrackMetaBox(avifEncoder * encoder, avifRWStream * s
     avifRWStreamFinishBox(s, meta);
 }
 
-static void avifWriteGridPayload(avifRWData * data, uint32_t gridCols, uint32_t gridRows, const avifImage * firstCell)
+static void avifWriteGridPayload(avifRWData * data, uint32_t gridCols, uint32_t gridRows, uint32_t gridWidth, uint32_t gridHeight)
 {
     // ISO/IEC 23008-12 6.6.2.3.2
     // aligned(8) class ImageGrid {
@@ -666,8 +669,6 @@ static void avifWriteGridPayload(avifRWData * data, uint32_t gridCols, uint32_t 
     //     unsigned int(FieldLength) output_height;
     // }
 
-    uint32_t gridWidth = firstCell->width * gridCols;
-    uint32_t gridHeight = firstCell->height * gridRows;
     uint8_t gridFlags = ((gridWidth > 65535) || (gridHeight > 65535)) ? 1 : 0;
 
     avifRWStream s;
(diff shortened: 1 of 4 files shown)


src/write.c
```diff
@@ -1291,19 +1430,12 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
...
-        }
-
         // Properties all av01 items need
 
         avifItemPropertyDedupStart(dedup);
         avifBoxMarker ispe = avifRWStreamWriteFullBox(&dedup->s, "ispe", AVIF_BOX_SIZE_TBD, 0, 0);
-        avifRWStreamWriteU32(&dedup->s, imageWidth);  // unsigned int(32) image_width;
-        avifRWStreamWriteU32(&dedup->s, imageHeight); // unsigned int(32) image_height;
+        avifRWStreamWriteU32(&dedup->s, item->outputWidth);  // unsigned int(32) image_width;
+        avifRWStreamWriteU32(&dedup->s, item->outputHeight); // unsigned int(32) image_height;
```
y-guyon, line 1438: Tracks are a new topic to me and I am not so sure of using `item->outputWidth` here instead of `imageWidth` (same at lines 1587 and 1701). There is not enough test coverage for tracks with grids to know if this works either.

src/write.c
```diff
@@ -726,6 +727,102 @@ static avifResult avifEncoderDataCreateXMPItem(avifEncoderData * data, const avi
...
+                                uint32_t dstRowBytes,
+                                uint32_t dstWidth,
+                                uint32_t dstHeight,
+                                const uint8_t * srcPlane,
+                                uint32_t srcRowBytes,
+                                uint32_t srcWidth,
+                                uint32_t srcHeight,
+                                uint32_t sampleByteCount)
+{
+    for (uint32_t j = 0; j < srcHeight; ++j) {
```
jzern, line 742: are there checks anywhere to ensure dstWidth/Height are >= src? If so, then maybe an assert can be added here

wantehchang (changes requested):
Yannis: I only reviewed src/write.c in the first round of review.

Could you file an issue with either HEIF or MIAF about the contradiction in the spec?
HEIF (or rather, MPEF file formats) issue tracker: https://github.com/MPEGGroup/FileFormat/issues
MIAF issue tracker: https://github.com/MPEGGroup/MIAF/issues

y-guyon:
> Could you file an issue with either HEIF or MIAF about the contradiction in the spec? HEIF (or rather, MPEG file formats) issue tracker: https://github.com/MPEGGroup/FileFormat/issues MIAF issue tracker: https://github.com/MPEGGroup/MIAF/issues

Done in https://github.com/MPEGGroup/FileFormat/issues/66.

src/write.c
y-guyon: > are there checks anywhere to ensure dstWidth/Height are >= src?

Yes, at line 885:

```c
    if ((topRightCell->width > tileWidth) || (bottomLeftCell->height > tileHeight)) {
    }
```

> If so, then maybe an assert can be added here

Done.

src/write.c
y-guyon: > You can't have an image sequence of grids.

Can it be a grid of tracks of image sequences? Otherwise it means `isGrid` is always false here, right?

wantehchang: Here is Joe's answer:

> The only way to signal a grid in any way is using a "grid" item, which refers to other items. The only way to signal a sequence is with tracks. They don't mix at all. There's no such thing as image grid sequences.

See also https://github.com/AOMediaCodec/libavif/pull/1149.

As for your second question, I am running into a limitation of GitHub's code review tool. You asked "Otherwise it means `isGrid` is always false here, right?" I assume by "here" you're referring to the following two lines:

```
        avifRWStreamWriteU32(&dedup->s, imageWidth);  // unsigned int(32) image_width;
        avifRWStreamWriteU32(&dedup->s, imageHeight); // unsigned int(32) image_height;
```

These two lines are part of the `ispe` box, which is only used with image items, not with image sequences.

tests/gtest/avifgridapitest.cc
```diff
@@ -15,139 +14,233 @@ using testing::ValuesIn;
...
+    EXPECT_EQ(
+        EncodeDecodeGrid({{{100, 100}, {100, 100}, {66, 100}}}, pixel_format),
+        AVIF_RESULT_OK);
+    // Bottom-most cells are smaller.
+    EXPECT_EQ(EncodeDecodeGrid({{{100, 100}, {100, 100}},
+                                {{100, 100}, {100, 100}},
+                                {{100, 66}, {100, 66}}},
+                               pixel_format),
+              AVIF_RESULT_OK)
+        << pixel_format;
```
wantehchang, line 192: Nit: Delete `<< pixel_format`? Or add it to the other `EXPECT_EQ` statements for consistency?

wantehchang (changes requested):
Yannis: I read everything, although I didn't read the tests carefully. Overall this looks good and thank you for adding more tests for image grids. Just some final suggestions and then this is good to go.

src/write.c
```diff
@@ -749,32 +849,53 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
+    //   excluding the bottom-most row) equal to tile_height, without gap or overlap, and then
+    //   trimming on the right and the bottom to the indicated output_width and output_height.
+    // Consider the combined input cellImages as the user's final output intent.
+    // Right and bottom cells may be padded below to enforce all tiles to be tileWidth by tileHeight,
+    // and the output cropped to gridWidth by gridHeight.
+    const uint32_t tileWidth = firstCell->width;
+    const uint32_t tileHeight = firstCell->height;
+    const uint32_t gridWidth = (gridCols - 1) * tileWidth + topRightCell->width;
+    const uint32_t gridHeight = (gridRows - 1) * tileHeight + bottomLeftCell->height;
+    for (uint32_t cellIndex = 0; cellIndex < cellCount; ++cellIndex) {
```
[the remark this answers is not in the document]
y-guyon: I wanted to separate the dimension-related tests from the other tests. The specification citations take space and they should be next to `tileWidth`, `gridWidth` but also close to `expectedCellWidth` definitions. Duplicating the comment or referring to it seemed worse than keeping only one for loop.

I do not think it matters much performance-wise or which error is returned first.

src/write.c
```diff
@@ -726,6 +728,104 @@ static avifResult avifEncoderDataCreateXMPItem(avifEncoderData * data, const avi
...
+                                uint32_t srcRowBytes,
+                                uint32_t srcWidth,
+                                uint32_t srcHeight,
+                                uint32_t sampleByteCount)
+{
+    assert(dstWidth >= srcWidth);
+    assert(dstHeight >= srcHeight);
+    for (uint32_t j = 0; j < srcHeight; ++j) {
+        const uint8_t * srcRow = &srcPlane[j * srcRowBytes];
+        uint8_t * dstRow = &dstPlane[j * dstRowBytes];
```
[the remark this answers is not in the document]
y-guyon: Is there any convention for casting the first or the second operand? Why `j * (size_t)dstRowBytes` but `(size_t)dstWidth * sampleByteCount`?

src/write.c
```diff
@@ -726,6 +728,104 @@ static avifResult avifEncoderDataCreateXMPItem(avifEncoderData * data, const avi
...
+                    dstRow16[x] = dstRow16[srcWidth - 1];
+                }
+            }
+        }
+    }
+
+    // Pad rows.
+    for (uint32_t j = srcHeight; j < dstHeight; ++j) {
+        uint8_t * dstRow = &dstPlane[j * dstRowBytes];
+        memcpy(dstRow, dstRow - dstRowBytes, dstRowBytes);
```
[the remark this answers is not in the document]
y-guyon: `dstImage` was created in `avifImageCopyAndPad()` so it is known here that `dstRowBytes == (size_t)dstWidth * sampleByteCount`. I agree that your suggestion is cleaner though.

src/write.c
y-guyon: > Here is Joe's answer

I did not realize image sequences were a completely different format at the container level (we are still writing a `meta` box, though? what if there is no item?). Thank you for forwarding Joe's explanation.

> I assume by "here" you're referring to the following two lines

Sorry, I was confused. I was talking about tracks and I thought the `ispe` box also applied to them. So the grid stuff is irrelevant at the lines 1590 and 1704, good to know.

src/write.c
```diff
@@ -178,6 +178,10 @@ typedef struct avifEncoderItem
     uint32_t gridCols; // if non-zero (legal range [1-256]), this is a grid item
     uint32_t gridRows; // if non-zero (legal range [1-256]), this is a grid item
 
+    // the reconstructed grid item will be trimmed to these dimensions (only present on grid items)
```
wantehchang, line 181: Nit: Change "reconstructed grid item" to "reconstructed image of a grid item".

src/write.c
```diff
@@ -749,32 +849,52 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
     }
 
     const avifImage * firstCell = cellImages[0];
+    const avifImage * topRightCell = cellImages[gridCols - 1];
+    const avifImage * bottomLeftCell = cellImages[cellCount - gridCols];
```
wantehchang, line 853: Optional: I just realized that we can replace `topRightCell` and `bottomLeftCell` with `bottomRightCell`:

```
    const avifImage * bottomRightCell = cellImages[cellCount - 1];
```

The top-right cell is the first cell of the right-most column, and the bottom-left cell is the first cell of the bottom-most row. That's why we think of them first. But if the cells in the image grid satisfy the width and height requirements, we can also use the bottom-right cell because it's in both the right-most column and the bottom-left cell.

wantehchang (approved):
LGTM. In the interest of time, you can merge the current version of this pull request and address my final suggestions later. I will also answer your questions later today. (I wanted to approve this pull request as soon as possible.)

y-guyon:
> LGTM. In the interest of time, you can merge the current version of this pull request and address my final suggestions later. I will also answer your questions later today. (I wanted to approve this pull request as soon as possible.)

Thanks. There is no rush though.

tests/gtest/avifgridapitest.cc
```diff
@@ -15,139 +14,233 @@ using testing::ValuesIn;
...
-    // Decode the grid image.
-    testutil::AvifImagePtr image(avifImageCreateEmpty(), avifImageDestroy);
-    ASSERT_NE(image, nullptr);
-    testutil::AvifDecoderPtr decoder(avifDecoderCreate(), avifDecoderDestroy);
-    ASSERT_NE(decoder, nullptr);
-    ASSERT_EQ(avifDecoderReadMemory(decoder.get(), image.get(),
-                                    encoded_avif.data, encoded_avif.size),
+  // Reconstruct the input image by merging all cells into a single avifImage.
+  testutil::AvifImagePtr grid =
+      testutil::CreateImage((int)image->width, (int)image->height, /*depth=*/8,
```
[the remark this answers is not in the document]
wantehchang: I consulted the Style Guide. It says:

https://google.github.io/styleguide/cppguide.html#Casting

> - Use brace initialization to convert arithmetic types (e.g., `int64_t{x}`). This is the safest approach because code will not compile if conversion can result in information loss. The syntax is also concise.

Since the example uses a variable `x`, I think `int{}` is not limited to just literals and constants. But it is definitely fine to use `static_cast<int>`.

src/write.c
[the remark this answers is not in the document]
wantehchang: Thank you for confirming the separate for loop is intended.

src/write.c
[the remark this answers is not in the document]
wantehchang: Strictly speaking both operands should be cast to `size_t`. We are trying to be concise by casting only one operand and relying on the compiler's automatic conversion of the other operand to `size_t`.

It doesn't matter wich operand is cast. You can always cast the first operand. In my suggestions I cast the larger operand.

In this particular case there is another solution (which you used in the `CopyImageSamples()` function in tests/gtest/aviftest_helpers.cc) that doesn't need casting:

```
    const uint8_t * srcRow = srcPlane
    uint8_t * dstRow = dstPlane;
    for (uint32_t j = 0; j < srcHeight; ++j) {
        ...
        srcRow += srcRowBytes;
        dstRow += dstRowBytes;
    }
```

src/write.c
wantehchang: libavif also creates an image item when encoding an image sequence. That image item is intended for renderers that cannot playback an image sequence. So there is always an item. (For simplicity libavif just points at the first key frame as a standalone image.)

At Cyril's suggestion, libavif writes a meta box for each track unless there is no data for it. (See the beginning of the `avifEncoderWriteTrackMetaBox()` function.)

wantehchang (approved):
Still LGTM.

PR #1144, opened 2022-10-04
CMake: Use add_compile_options in place of add_definitions

jbms:
`add_definitions` has been superseded by other commands and can lead to surprising behavior.  This commit replaces the uses of it with `add_compile_options`, which is the appropriate replacement for compiler options that don't define macros or specify include directories.


wantehchang (approved):
LGTM.

PR #1145, opened 2022-10-04
Replace tagStandardXMP by define in avifjpeg.c

y-guyon:


apps/shared/avifjpeg.c
```diff
@@ -418,7 +423,7 @@ avifBool avifJPEGRead(const char * inputFilename,
                     fprintf(stderr, "XMP extraction failed: truncated extended XMP segment\n");
                     goto cleanup;
                 }
-                const uint8_t * guid = &marker->data[tagExtendedXMP.size];
+                const uint8_t * guid = &marker->data[AVIF_EXTENDED_XMP_TAG_LENGTH];
```
wantehchang, line 426: Optional: We can now define `offsetTillExtendedXMP` as a const variable near where it's used. Its value is a compile-time constant now.

```
                // Offset in marker->data segment (skip tag + guid + size + offset).
                const size_t offsetTillExtendedXMP = AVIF_EXTENDED_XMP_TAG_LENGTH + AVIF_EXTENDED_XMP_GUID_LENGTH + 4 + 4;
```

wantehchang (approved):
LGTM.

jzern (approved):

PR #1146, opened 2022-10-05
Change avifImageCreate to take uint32_t parameters

wantehchang:
Change the width, height, and depth parameters of avifImageCreate() to be uint32_t instead of int, because they are assigned to the corresponding members of avifImage, which are uint32_t.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -474,7 +474,7 @@ typedef struct avifImage
     avifRWData xmp;
 } avifImage;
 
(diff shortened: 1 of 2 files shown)


joedrago:
These changes look quite straightforward at first glance, but I'm surprised no associated changes were necessary in the tools. Does this change really not cause warnings (or spurious casts) elsewhere?

joedrago (approved):

y-guyon (approved):

PR #1148, opened 2022-10-05
Check if it's safe to cast width,height to uint16_t

wantehchang:


src/write.c
```diff
@@ -1558,6 +1558,12 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
             avifRWStreamSetOffset(&s, prevOffset);
             avifRWStreamFinishBox(&s, stts);
 
+            // Check whether it is safe to cast width and height to uint16_t. The maximum width and
+            // height of an AV1 frame are 65536, which just exceeds uint16_t.
+            if ((imageMetadata->width > 65535) || (imageMetadata->height > 65535)) {
+                return AVIF_RESULT_INVALID_ARGUMENT;
+            }
```
wantehchang, line 1565: Joe: We need to cast `imageMetadata->width` and `imageMetadata->height` to `uint16_t` when writing out the VisualSampleEntry, at lines 1575-1576 below:

```
            avifRWStreamWriteU16(&s, (uint16_t)imageMetadata->width);  // unsigned int(16) width;
            avifRWStreamWriteU16(&s, (uint16_t)imageMetadata->height); // unsigned int(16) height;
```

Do you know where is the best place to check this?

wantehchang: I discussed this with Joe. We will check this when we detect we will be writing an image sequence.

src/write.c
```diff
@@ -974,6 +974,25 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
+        //        unsigned int(16) width;
+        //        unsigned int(16) height;
+        //        ...
+        //     }
+        // Check whether it is safe to cast width and height to uint16_t. The maximum width and
+        // height of an AV1 frame are 65536, which just exceeds uint16_t.
+        assert(encoder->data->items.count > 0);
+        const avifImage * imageMetadata = encoder->data->imageMetadata;
+        if ((imageMetadata->width > 65535) || (imageMetadata->height > 65535)) {
+            return AVIF_RESULT_INVALID_ARGUMENT;
```
wantehchang, line 992: `AVIF_RESULT_INVALID_ARGUMENT` isn't the best error code, but I can't find a better error code.

src/write.c
```diff
@@ -974,6 +974,25 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
         }
     }
 
+    if (encoder->data->frames.count == 1) {
```
wantehchang, line 977: In `avifEncoderFinish()`, we check `if (encoder->data->frames.count > 1)` to determine if we will be writing tracks (an image sequence). This is why I check `if (encoder->data->frames.count == 1)` here. We will call `avifArrayPushPtr(&encoder->data->frames)` at the end of this function (at line 1026).

wantehchang:
@vigneshvg Vignesh: Please make sure your AVIF image sequence encoder in ffmpeg is taking the same precaution before writing the `width` and `height` fields of `VisualSampleEntry`.

joedrago (approved):

vigneshvg:
> @vigneshvg Vignesh: Please make sure your AVIF image sequence encoder in ffmpeg is taking the same precaution before writing the width and height fields of VisualSampleEntry.

Thanks for pointing this out. ffmpeg already has a similar check [1]. So we are good there.

[1] https://git.ffmpeg.org/gitweb/ffmpeg.git/blob/aaf4109a5f36f4a481c6d278fda6f180d31c5802:/libavformat/movenc.c#l7084

PR #1150, opened 2022-10-07
Add avifImagePlaneWidth/Height() in avif.h

y-guyon:
Add avifpng16btest.cc

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1107,6 +1107,9 @@ AVIF_API void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const cha
 
 // Helpers
 AVIF_API avifBool avifImageUsesU16(const avifImage * image);
+// channel can be an avifChannelIndex or 3 for alpha.
+AVIF_API uint32_t avifImagePlaneWidth(const avifImage * image, int channel);
+AVIF_API uint32_t avifImagePlaneHeight(const avifImage * image, int channel);
 
 // Returns AVIF_TRUE if input begins with a valid FileTypeBox (ftyp) that supports
 // either the brand 'avif' or 'avis' (or both), without performing any allocations.
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -102,7 +102,10 @@ typedef enum avifChannelIndex
     // These can be used as the index for the yuvPlanes and yuvRowBytes arrays in avifImage.
     AVIF_CHAN_Y = 0,
     AVIF_CHAN_U = 1,
-    AVIF_CHAN_V = 2
+    AVIF_CHAN_V = 2,
+
+    // This may not be used in yuvPlanes and yuvRowBytes, but is available for use with avifImagePlane().
+    AVIF_CHAN_A = 3
 } avifChannelIndex;
 
 // ---------------------------------------------------------------------------
@@ -1107,11 +1110,10 @@ AVIF_API void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const cha
 
 // Helpers
 AVIF_API avifBool avifImageUsesU16(const avifImage * image);
-// channel can be an avifChannelIndex or 3 for alpha.
-AVIF_API uint8_t * avifImagePlaneRow(const avifImage * image, int channel, uint32_t y);
-AVIF_API uint32_t avifImagePlaneRowBytes(const avifImage * image, int channel);
-AVIF_API uint32_t avifImagePlaneWidth(const avifImage * image, int channel);
-AVIF_API uint32_t avifImagePlaneHeight(const avifImage * image, int channel);
+AVIF_API uint8_t * avifImagePlane(const avifImage * image, avifChannelIndex channel);
+AVIF_API uint32_t avifImagePlaneRowBytes(const avifImage * image, avifChannelIndex channel);
+AVIF_API uint32_t avifImagePlaneWidth(const avifImage * image, avifChannelIndex channel);
+AVIF_API uint32_t avifImagePlaneHeight(const avifImage * image, avifChannelIndex channel);
 
 // Returns AVIF_TRUE if input begins with a valid FileTypeBox (ftyp) that supports
 // either the brand 'avif' or 'avis' (or both), without performing any allocations.
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1108,6 +1108,8 @@ AVIF_API void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const cha
 // Helpers
 AVIF_API avifBool avifImageUsesU16(const avifImage * image);
 // channel can be an avifChannelIndex or 3 for alpha.
+AVIF_API uint8_t * avifImagePlaneRow(const avifImage * image, int channel, uint32_t y);
+AVIF_API uint32_t avifImagePlaneRowBytes(const avifImage * image, int channel);
 AVIF_API uint32_t avifImagePlaneWidth(const avifImage * image, int channel);
 AVIF_API uint32_t avifImagePlaneHeight(const avifImage * image, int channel);
 
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1110,10 +1110,11 @@ AVIF_API void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const cha
 
 // Helpers
 AVIF_API avifBool avifImageUsesU16(const avifImage * image);
-AVIF_API uint8_t * avifImagePlane(const avifImage * image, avifChannelIndex channel);
-AVIF_API uint32_t avifImagePlaneRowBytes(const avifImage * image, avifChannelIndex channel);
-AVIF_API uint32_t avifImagePlaneWidth(const avifImage * image, avifChannelIndex channel);
-AVIF_API uint32_t avifImagePlaneHeight(const avifImage * image, avifChannelIndex channel);
+// channel can be an avifChannelIndex.
+AVIF_API uint8_t * avifImagePlane(const avifImage * image, int channel);
+AVIF_API uint32_t avifImagePlaneRowBytes(const avifImage * image, int channel);
+AVIF_API uint32_t avifImagePlaneWidth(const avifImage * image, int channel);
+AVIF_API uint32_t avifImagePlaneHeight(const avifImage * image, int channel);
 
 // Returns AVIF_TRUE if input begins with a valid FileTypeBox (ftyp) that supports
 // either the brand 'avif' or 'avis' (or both), without performing any allocations.
--- a/src/avif.c
+++ b/src/avif.c
@@ -422,7 +422,7 @@ uint32_t avifImagePlaneWidth(const avifImage * image, int channel)
         }
         return (image->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX;
     }
-    if ((channel == 3) || (image->alphaPlane)) {
+    if ((channel == 3) && (image->alphaPlane)) {
         return image->width;
     }
     return 0;
@@ -441,7 +441,7 @@ uint32_t avifImagePlaneHeight(const avifImage * image, int channel)
         }
         return (image->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY;
     }
-    if ((channel == 3) || (image->alphaPlane)) {
+    if ((channel == 3) && (image->alphaPlane)) {
         return image->height;
     }
     return 0;
--- a/src/avif.c
+++ b/src/avif.c
@@ -409,7 +409,7 @@ avifBool avifImageUsesU16(const avifImage * image)
     return (image->depth > 8);
 }
 
-uint8_t * avifImagePlane(const avifImage * image, avifChannelIndex channel)
+uint8_t * avifImagePlane(const avifImage * image, int channel)
 {
     if ((channel == AVIF_CHAN_Y) || (channel == AVIF_CHAN_U) || (channel == AVIF_CHAN_V)) {
         return image->yuvPlanes[channel];
@@ -420,7 +420,7 @@ uint8_t * avifImagePlane(const avifImage * image, avifChannelIndex channel)
     return NULL;
 }
 
-uint32_t avifImagePlaneRowBytes(const avifImage * image, avifChannelIndex channel)
+uint32_t avifImagePlaneRowBytes(const avifImage * image, int channel)
 {
     if ((channel == AVIF_CHAN_Y) || (channel == AVIF_CHAN_U) || (channel == AVIF_CHAN_V)) {
         return image->yuvRowBytes[channel];
@@ -431,7 +431,7 @@ uint32_t avifImagePlaneRowBytes(const avifImage * image, avifChannelIndex channe
     return 0;
 }
 
-uint32_t avifImagePlaneWidth(const avifImage * image, avifChannelIndex channel)
+uint32_t avifImagePlaneWidth(const avifImage * image, int channel)
 {
     if (channel == AVIF_CHAN_Y) {
         return image->width;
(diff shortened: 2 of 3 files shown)


include/avif/internal.h
```diff
@@ -137,8 +137,8 @@ typedef struct avifReformatState
     avifPixelFormatInfo formatInfo;
 
     // LUTs for going from YUV limited/full unorm -> full range RGB FP32
-    float unormFloatTableY[1 << 12];
-    float unormFloatTableUV[1 << 12];
+    float unormFloatTableY[1 << 16];
+    float unormFloatTableUV[1 << 16];
```
tongyuantongyu, line 141: Note that we are allocating `avifReformatState` on the stack. The two arrays will take 512KB of stack space, which can be a bit too big.

tongyuantongyu (commented):
I wonder why you want 16bit YUV support in libavif. AV1 supports only 8, 10 and 12 bit image if I recall correctly. I believe we only want some conversion utility in libavif if it's useful for handling AVIF, right?

include/avif/internal.h
joedrago: I agree that this is much too large for something that is going to be on the stack, and not useful to AV1 conversion.

y-guyon:
> I wonder why you want 16bit YUV support in libavif. AV1 supports only 8, 10 and 12 bit image if I recall correctly. I believe we only want some conversion utility in libavif if it's useful for handling AVIF, right?

It is not 16bit YUV support in libavif, it is just being able to read 16bit RGB samples from PNG. I used that for a local unrelated experiment and thought that it might be useful to someone else in the future, on top of not being a big change. The latter does no longer hold because of the increased stack size strain, so I abandoned this change.

I only kept the `avifImagePlaneWidth()` and `avifImagePlaneHeight()` additions, which I believe makes sample browsing simpler and shorter in most cases. The performance loss of calling `avifGetPixelFormatInfo()` multiple times is nothing compared to browsing pixels.

tests/gtest/aviftest_helpers.cc
```diff
@@ -75,24 +75,15 @@ AvifImagePtr CreateImage(int width, int height, int depth,
...
         (c == AVIF_CHAN_A) ? image->alphaRowBytes : image->yuvRowBytes[c];
-    const uint32_t plane_width =
-        (c == AVIF_CHAN_Y || c == AVIF_CHAN_A)
-            ? image->width
-            : ((image->width + info.chromaShiftX) >> info.chromaShiftX);
-    const uint32_t plane_height =
-        (c == AVIF_CHAN_Y || c == AVIF_CHAN_A)
-            ? image->height
-            : ((image->height + info.chromaShiftY) >> info.chromaShiftY);
+    const uint32_t plane_width = avifImagePlaneWidth(image, c);
```
wantehchang, line 85: For your use case (simplifying test code like this), we need two other functions that replace the code at lines 79 and 83-84.

src/avif.c
```diff
@@ -409,6 +409,44 @@ avifBool avifImageUsesU16(const avifImage * image)
...
+    }
+    if ((channel == AVIF_CHAN_U) || (channel == AVIF_CHAN_V)) {
+        avifPixelFormatInfo formatInfo;
+        avifGetPixelFormatInfo(image->yuvFormat, &formatInfo);
+        if (formatInfo.monochrome) {
+            return 0;
+        }
+        return (image->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX;
+    }
+    if ((channel == 3) || (image->alphaPlane)) {
```
wantehchang, line 425: This condition seems wrong. Should it be `&&`?

wantehchang (commented):
Yannis: Let's wait until after the v0.11.0 release to merge this pull request.

tests/gtest/aviftest_helpers.cc
```diff
@@ -75,57 +75,28 @@ AvifImagePtr CreateImage(int width, int height, int depth,
...
+    const uint32_t plane_height = avifImagePlaneHeight(image, c);
     for (uint32_t y = 0; y < plane_height; ++y) {
+      uint8_t* row = avifImagePlaneRow(image, c, y);
```
wantehchang, line 82: Unless the compiler inlines this `avifImagePlaneRow()` call, the compiler won't be able to perform the common optimization of converting this to the original code (i.e., replacing multiplcation with addition).

I think we just need to provide a `avifImagePlane()` function that returns a pointer to the first row of the plane.

wantehchang (changes requested):

include/avif/avif.h
```diff
@@ -1107,6 +1107,11 @@ AVIF_API void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const cha
 
 // Helpers
 AVIF_API avifBool avifImageUsesU16(const avifImage * image);
+// channel can be an avifChannelIndex or 3 for alpha.
```
wantehchang, line 1110: I discussed this pull request with Joe. Joe suggested that we add `AVIF_CHAN_A` to the `avifChannelIndex` enum. Then we can declare the `channel` parameter as `avifChannelIndex channel` and get rid of this comment.

I don't quite remember why I objected to this when you suggested this. It may have to do with the comment in the `avifChannelIndex` enum definition:

```
typedef enum avifChannelIndex
{
    // These can be used as the index for the yuvPlanes and yuvRowBytes arrays in avifImage.
    AVIF_CHAN_Y = 0,
    AVIF_CHAN_U = 1,
    AVIF_CHAN_V = 2
} avifChannelIndex;
```

Joe suggests adding a blank line and a comment, something like:

```
typedef enum avifChannelIndex
{
    // These can be used as the index for the yuvPlanes and yuvRowBytes arrays in avifImage.
    AVIF_CHAN_Y = 0,
    AVIF_CHAN_U = 1,
    AVIF_CHAN_V = 2,

    // This may not be used in yuvPlanes and yuvRowBytes, but is available for use with avifImagePlane().
    AVIF_CHAN_A = 3
} avifChannelIndex;
```

wantehchang (changes requested):

tests/gtest/aviftest_helpers.cc
wantehchang: I am not really a CS person. According to this Wikipedia page, the compiler optimization I described in the previous comment is called strength reduction:

https://en.wikipedia.org/wiki/Strength_reduction

include/avif/avif.h
y-guyon: Done. In C++ files it is fine because we can use `for (avifChannelIndex c : {AVIF_CHAN_Y, AVIF_CHAN_U, AVIF_CHAN_V, AVIF_CHAN_A})`, but C files will need `for (int i = 0; i < 4; ++i)` and a `(avifChannelIndex)c` cast to call `avifImagePlane()`.

tests/gtest/aviftest_helpers.cc
wantehchang: The code I have in mind would call `avifImagePlaneRowBytes()` outside the `y` for loop and save the return value in a local variable, similar to `plane_width` and `plane_height`.

For example, the `CopyImageSamples()` function in tests/gtest/aviftest_helpers.cc would be written as follows (I found that it's very close to the original code):

```
void CopyImageSamples(const avifImage& from, avifImage* to) {
  ...

  for (int c = 0; c < 4; c++) {
    const uint8_t* from_row = avifImagePlane(&from, c);
    uint8_t* to_row = avifImagePlane(to, c);
    assert(!from_row == !to_row);
    if (!from_row) {
      continue;
    }
    const uint32_t from_row_bytes = avifImagePlaneRowBytes(&from, c);
    const uint32_t to_row_bytes = avifImagePlaneRowBytes(to, c);
    const uint32_t plane_width = avifImagePlaneWidth(&from, c);
    const uint32_t plane_height = avifImagePlaneHeight(&from, c);
    for (uint32_t y = 0; y < plane_height; ++y) {
      if (avifImageUsesU16(&from)) {
        std::copy(reinterpret_cast<const uint16_t*>(from_row),
                  reinterpret_cast<const uint16_t*>(from_row) + plane_width,
                  reinterpret_cast<uint16_t*>(to_row));
      } else {
        std::copy(from_row, from_row + plane_width, to_row);
      }
      from_row += from_row_bytes;
      to_row += to_row_bytes;
    }
  }
}
```

Since you wrote the original code and I remember I didn't suggest this structure to you, I am wondering if you now see some problem with this structure. (This structure is commonly used.)

Or do you think `avifImagePlaneRow()` will be handy, for example, when we need to get a pointer to the last row:

```
    const uint32_t plane_height = avifImagePlaneHeight(image, channel);
    uint8_t * last_row = avifImagePlaneRow(image, channel, plane_height - 1);
```

include/avif/avif.h
y-guyon: > I would also change the for loops back to `for (int c = 0; c < 4; c++)`.

Why? I find `for (avifChannelIndex c : {AVIF_CHAN_Y, AVIF_CHAN_U, AVIF_CHAN_V, AVIF_CHAN_A})` longer (unfortunately) but type-safe and easier to understand.

tests/gtest/aviftest_helpers.cc
y-guyon: > Since you wrote the original code and I remember I didn't suggest this structure to you, I am wondering if you now see some problem with this structure.

I do not see any problem with this structure. It was the only structure available with libavif API at the time. In this PR, I was just trying to shorten it from

```c
  for (int c = 0; c < 4; c++) {
    const uint8_t* from_row = avifImagePlane(from, c);
    uint8_t* to_row = avifImagePlane(to, c);
    const uint32_t from_row_bytes = avifImagePlaneRowBytes(from, c);
    const uint32_t to_row_bytes = avifImagePlaneRowBytes(to, c);
    const uint32_t plane_height = avifImagePlaneHeight(from, c);
    for (uint32_t y = 0; y < plane_height; ++y) {
      // do something with from_row and to_row
      from_row += from_row_bytes;
      to_row += to_row_bytes;
    }
  }
```
to
```c
  for (int c = 0; c < 4; c++) {
    for (uint32_t y = 0; y < avifImagePlaneHeight(from, c); ++y) {
      const uint8_t* from_row = avifImagePlaneRow(from, c, y);
      uint8_t* to_row = avifImagePlaneRow(to, c, y);
      // do something with from_row and to_row
    }
  }
```

but let's keep `avifImagePlane()` for consistency with the commonly used patterns.

> Or do you think `avifImagePlaneRow()` will be handy, for example, when we need to get a pointer to the last row

This is not a very common case. This example was not part of my motivation to introduce `avifImagePlaneRow()`, although `avifImagePlaneRow()` being a superset of `avifImagePlane()`, it could be used for any pattern.

include/avif/avif.h
wantehchang: Using `for (int c = 0; c < 4; c++)` is optional. I didn't realize you prefer the new form. (I thought you may have used the new form to avoid the `avifChannelIndex` type cast.)

tests/gtest/aviftest_helpers.cc
wantehchang: Thank you for clarifying your previous comment. Calling `avifImagePlaneRow(from, c, y)` inside the `y` for loop means the code in `avifImagePlaneRow(from, c, y)` that depends only on `from` and `c` is executed repeatedly.

src/avif.c
```diff
@@ -409,6 +409,66 @@ avifBool avifImageUsesU16(const avifImage * image)
...
+    }
+    return NULL;
+}
+
+uint32_t avifImagePlaneRowBytes(const avifImage * image, int channel)
+{
+    if ((channel == AVIF_CHAN_Y) || (channel == AVIF_CHAN_U) || (channel == AVIF_CHAN_V)) {
+        return image->yuvRowBytes[channel];
+    }
+    if ((channel == AVIF_CHAN_A) && (image->alphaPlane)) {
```
wantehchang, line 428: Nit: omit `&& (image->alphaPlane)`.

Note that we don't check `image->yuvPlanes[channel]`, which can also be null (for monochrome images).

Also, in the `avifImage` structs that we construct, I believe we maintain the invariant that if `image->alphaPlane` is null, then `image->alphaRowBytes` is 0. We also maintain a similar invariant for `image->yuvPlanes[channel]` and `image->yuvRowBytes[channel]`.

tests/gtest/aviftest_helpers.cc
```diff
@@ -75,24 +70,11 @@ AvifImagePtr CreateImage(int width, int height, int depth,
...
-            : ((image->width + info.chromaShiftX) >> info.chromaShiftX);
-    const uint32_t plane_height =
-        (c == AVIF_CHAN_Y || c == AVIF_CHAN_A)
-            ? image->height
-            : ((image->height + info.chromaShiftY) >> info.chromaShiftY);
+  for (avifChannelIndex c :
+       {AVIF_CHAN_Y, AVIF_CHAN_U, AVIF_CHAN_V, AVIF_CHAN_A}) {
+    const uint32_t plane_width = avifImagePlaneWidth(image, c);
+    const uint32_t plane_height = avifImagePlaneHeight(image, c);
+    uint8_t* row = avifImagePlane(image, c);
```
wantehchang, line 77: Nit: save `avifImagePlaneRowBytes(image, c)` in a local variable `row_bytes` and use `row_bytes` in line 86.

Make similar changes to the three functions that were modified below.

tests/gtest/aviftest_helpers.cc
```diff
@@ -186,47 +155,33 @@ bool AreImagesEqual(const avifImage& image1, const avifImage& image2,
...
-        (c == AVIF_CHAN_A) ? image2.alphaRowBytes : image2.yuvRowBytes[c];
-    const uint32_t plane_width =
-        (c == AVIF_CHAN_Y || c == AVIF_CHAN_A)
-            ? image1.width
-            : ((image1.width + info.chromaShiftX) >> info.chromaShiftX);
-    const uint32_t plane_height =
-        (c == AVIF_CHAN_Y || c == AVIF_CHAN_A)
-            ? image1.height
-            : ((image1.height + info.chromaShiftY) >> info.chromaShiftY);
+    const uint32_t plane_width = avifImagePlaneWidth(&image1, c);
```
wantehchang, line 169: Optional: Lines 166-168 could be omitted:

```
    if (!row1) {
      continue;
    }
```

You omitted the corresponding code in the two functions that were modified above. It would be good to be consistent. Either form is fine by me.

tests/gtest/aviftest_helpers.cc
```diff
@@ -241,29 +196,16 @@ void CopyImageSamples(const avifImage& from, avifImage* to) {
...
-        (c == AVIF_CHAN_A) ? to->alphaRowBytes : to->yuvRowBytes[c];
-    const uint32_t plane_width =
-        (c == AVIF_CHAN_Y || c == AVIF_CHAN_A)
-            ? from.width
-            : ((from.width + info.chromaShiftX) >> info.chromaShiftX);
-    const uint32_t plane_height =
-        (c == AVIF_CHAN_Y || c == AVIF_CHAN_A)
-            ? from.height
-            : ((from.height + info.chromaShiftY) >> info.chromaShiftY);
+    const uint32_t plane_width = avifImagePlaneWidth(&from, c);
```
wantehchang, line 207: Nit: Lines 204-206 could be omitted. See my comment at line 169.

wantehchang (approved):
LGTM with suggested changes.

wantehchang (approved):
LGTM.

include/avif/avif.h
joedrago: For what it's worth, when I saw Yannis' comment about the loops, I had a very similar reaction to Wan-Teh (that we should leave the argument as `int`). This kind of function is almost always called in a loop from C, and it isn't worth the annoying casts. We should leave the function signature alone.

wantehchang:
@y-guyon Hi Yannis. Because of the need to create a v0.11.1 patch release next week, I'd appreciate it if you could wait a couple of days to merge this pull request. Thank you!

wantehchang:
Hi Yannis,

You can merge this pull request now. Sorry about the delay.

PR #1153, opened 2022-10-10
Two comment edits

wantehchang:
1. In "avoidLibYUV should be set to AVIF_TRUE ...", change "should" to "can" to make it clear this is optional, not a recommendation.

2. In "to be consumed by the codec at the next avifEncoderAddImage() call", change "at" to "in".

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -17,7 +17,7 @@ extern "C" {
 // AVIF_BUILDING_SHARED_LIBS should only be defined when libavif is being built
 // as a shared library.
 // AVIF_DLL should be defined if libavif is a shared library. If you are using
-// libavif as CMake dependency, through CMake package config file or through
+// libavif as a CMake dependency, through a CMake package config file or through
 // pkg-config, this is defined automatically.
 //
 // Here's what AVIF_API will be defined as in shared build:
@@ -567,7 +567,7 @@ AVIF_API avifBool avifRGBFormatHasAlpha(avifRGBFormat format);
 typedef enum avifChromaUpsampling
 {
     AVIF_CHROMA_UPSAMPLING_AUTOMATIC = 0,    // Chooses best trade off of speed/quality (uses BILINEAR libyuv if available,
-                                             // or fallbacks to NEAREST libyuv if available, or fallbacks to BILINEAR built-in)
+                                             // or falls back to NEAREST libyuv if available, or falls back to BILINEAR built-in)
     AVIF_CHROMA_UPSAMPLING_FASTEST = 1,      // Chooses speed over quality (same as NEAREST)
     AVIF_CHROMA_UPSAMPLING_BEST_QUALITY = 2, // Chooses the best quality upsampling, given settings (same as BILINEAR)
     AVIF_CHROMA_UPSAMPLING_NEAREST = 3,      // Uses nearest-neighbor filter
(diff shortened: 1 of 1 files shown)


include/avif/avif.h
```diff
@@ -538,7 +538,7 @@ AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, a
 // If libavif is built with a version of libyuv offering a fast conversion between RGB and YUV for
 // the given inputs, libavif will use it. See reformat_libyuv.c for the details.
 // libyuv is faster but may have slightly less precision than built-in conversion, so avoidLibYUV
-// should be set to AVIF_TRUE when AVIF_CHROMA_UPSAMPLING_BEST_QUALITY or
+// can be set to AVIF_TRUE when AVIF_CHROMA_UPSAMPLING_BEST_QUALITY or
```
wantehchang, line 541: The current wording seems to recommend that avoidLibYUV should be set to AVIF_TRUE when AVIF_CHROMA_UPSAMPLING_BEST_QUALITY or AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY is used. If what we meant is that avoidLibYUV can be set to AVIF_FALSE or AVIF_TRUE, and setting it to AVIF_TRUE when AVIF_CHROMA_UPSAMPLING_BEST_QUALITY or AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY is used produces the most precise but slowest results, then we can change "should" to "can" to clarify that. An alternative rewrite is the following:

> To get the most precise but slowest results, use AVIF_CHROMA_UPSAMPLING_BEST_QUALITY or AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY and set avoidLibYUV to AVIF_TRUE.

wantehchang (commented):
Yannis, please see my question below.

include/avif/avif.h
y-guyon: Your suggestion looks good. Should it be "and" instead of "or"?

y-guyon (approved):

jzern (commented):
I can't seem to add a comment except within the diff hunk.

from line 20:
...libavif as CMake dependency, through CMake package config file...

could be:
...libavif as a CMake dependency, through a CMake package config file...

from 570:
fallbacks -> falls back?

jzern (approved):

wantehchang:
James, thanks for reviewing avif.h. I made the changes you suggested.

> I can't seem to add a comment except within the diff hunk.

Yes, this is a limitation of GitHub's code review tool.

include/avif/avif.h
wantehchang: Thank you for confirming I understood the comment correctly. In the interest of time, I am going to merge this pull request as is, because the "should" to "can" change also clarifies the comment.

Regarding the alternative that I suggested:

> To get the most precise but slowest results, use AVIF_CHROMA_UPSAMPLING_BEST_QUALITY or AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY and set avoidLibYUV to AVIF_TRUE.

I used "AVIF_CHROMA_UPSAMPLING_BEST_QUALITY or AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY" because "or" was used in the original comment. I guess I thought "or" was okay because an `avifRGBImage` struct is typically only used for either YUV-to-RGB or RGB-to-YUV conversion, so we only need to set either `chromaUpsampling` or `chromaDownsampling`. It is harmless but not necessary to set both.

Perhaps we can clarify the alternative as follows:

> To get the most precise but slowest results, set chromaUpsampling to AVIF_CHROMA_UPSAMPLING_BEST_QUALITY (for avifImageYUVToRGB) or set chromaDownsampling to AVIF_CHROMA_DOWNSAMPLING_BEST_QUALITY (for avifImageRGBToYUV), and set avoidLibYUV to AVIF_TRUE.

But this is getting a little verbose.

PR #1154, opened 2022-10-10
Have avifImageSetMetadataExif extract orientation

wantehchang:
Have avifImageSetMetadataExif() parse the Exif metadata and extract the Exif orientation to irot and imir properties.

Remove avifImageExtractExifOrientationToIrotImir() from the public API but keep it as an internal function.

Call avifRWDataSet(&image->exif, exif, exifSize) directly when we just need to set image->exif.

Based on the reverse of the following commit of Yannis Guyon's: https://github.com/AOMediaCodec/libavif/pull/1099/commits/9b631db687834aeb8a52884d109f4529cc69a824

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ return values of avifImageCopy() and avifImageAllocatePlanes().
   upsampling chroma from 4:2:0 or 4:2:2 and has no impact on the use of libyuv.
   Set avifRGBImage::avoidLibYUV accordingly to control the use of libyuv.
 * Exif and XMP metadata is imported from PNG and JPEG files.
+* avifImageSetMetadataExif() parses the Exif metadata and converts any Exif
+  orientation found into transformFlags, irot and imir values.
 
 ### Removed
 * alphaRange field was removed from the avifImage struct. It it presumed that
@@ -40,9 +42,6 @@ return values of avifImageCopy() and avifImageAllocatePlanes().
 * Add imageDimensionLimit field to avifDecoder struct
 * Add autoTiling field to avifEncoder struct
 * Add AVIF_CHROMA_DOWNSAMPLING_SHARP_YUV value to avifChromaDownsampling enum
-* Add avifImageExtractExifOrientationToIrotImir() which parses the Exif metadata
-  of an image and converts any Exif orientation found into transformFlags, irot
-  and imir values.
 * avifdec: Add --dimension-limit, which specifies the image dimension limit
   (width or height) that should be tolerated
 * avifenc: Add --sharpyuv, which enables "sharp" RGB to YUV420 conversion, which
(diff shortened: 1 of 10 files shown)


wantehchang:
To help your code review, I created a copy of this pull request at https://aomedia-review.googlesource.com/c/libavif/+/164401.

joedrago (approved):

y-guyon (approved):

y-guyon:
Note: There is now no way for the user to know whether the Exif metadata was valid and did not contain any orientation information, or whether the Exif metadata was valid enough to pass `avifEncoderWrite()` but invalid enough to fail `avifImageExtractExifOrientationToIrotImir()`. This is a small issue, but `avifImageExtractExifOrientationToIrotImir()` could be called in `avifEncoderWrite()` just to check the correctness of the Exif metadata.

wantehchang:
Yannis: Thank you for the note. The change you suggested will make `avifEncoderWrite()` fail with `AVIF_RESULT_INVALID_EXIF_PAYLOAD` if the Exif metadata has the TIFF header but the data after the TIFF header is truncated. This differs from the behavior of avifenc.c, avifjpeg.c, and avifpng.c before this pull request, which ignored any Exif parsing failure. Did I understand your suggestion correctly?

The following comment for `avifImageSetMetadataExif()` that you mentioned was originally added in https://github.com/AOMediaCodec/libavif/commit/71b8805d7c9cf55a82044d4f0336c0ee6e4a6e6d

```
// Warning: If the Exif payload is set and invalid, avifEncoderWrite() may return AVIF_RESULT_INVALID_EXIF_PAYLOAD.
```

Since the comment says "may", `avifEncoderWrite()` doesn't need to return `AVIF_RESULT_INVALID_EXIF_PAYLOAD` if the Exif payload is invalid. Do you think `avifEncoderWrite()` should return `AVIF_RESULT_INVALID_EXIF_PAYLOAD` if it knows the Exif payload is invalid?

y-guyon:
> Did I understand your suggestion correctly?

Yes.

> Do you think `avifEncoderWrite()` should return `AVIF_RESULT_INVALID_EXIF_PAYLOAD` if it knows the Exif payload is invalid?

Yes, I believe it would be less obscure for the user. If the user provides truncated Exif metadata with some expected orientation that libavif fails to extract, getting an error or warning from libavif seems appropriate to me. The user could check the content of `avifImage`'s `irot`/`imir`after calling `avifImageSetMetadataExif()`, but there is no way to know if it failed to parse the Exif payload or if it successfully found orientation 1 (no transformation).

What would be even better in my opinion, is to return an error from `avifImageSetMetadataExif()`, with an option to bypass the Exif correctness check. The latter could be done by accessing `avifImage::exif` directly (but it is discouraged), by another function or by a new argument in `avifImageSetMetadataExif()` (as done in https://github.com/AOMediaCodec/libavif/pull/1099/commits/040eb30453dbc51b07633c1c78958e8d5c0ccf39). But these change the API.

With all that said, it is not an important matter, and the note was just a thought. The consequences of either implementations are insignificant.

PR #1155, opened 2022-10-11
Enable libaom-specific AV1E_SET_SKIP_POSTPROC_FILTERING by default

RemyaPrakasan:
This PR enables AV1E_SET_SKIP_POSTPROC_FILTERING by default as the application of post-processing filters on the reconstructed frame can be skipped in case of still-picture encoding as the frame is never referenced.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -750,6 +750,15 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 return AVIF_RESULT_UNKNOWN_ERROR;
             }
         }
+
+#if defined(AOM_USAGE_ALL_INTRA) && defined(AOM_CTRL_AV1E_SET_SKIP_POSTPROC_FILTERING)
+        if (cfg->g_usage == AOM_USAGE_ALL_INTRA) {
+            // Enable AV1E_SET_SKIP_POSTPROC_FILTERING for still-picture encoding, which is
+            // different from libaom's default.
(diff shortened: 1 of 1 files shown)


RemyaPrakasan:
@wantehchang , Please review.

src/codec_aom.c
```diff
@@ -816,6 +816,13 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         unsigned int cqLevel = (cfg->rc_min_quantizer + cfg->rc_max_quantizer) / 2;
         aom_codec_control(&codec->internal->encoder, AOME_SET_CQ_LEVEL, cqLevel);
     }
+#if defined(AOM_CTRL_AV1E_SET_SKIP_POSTPROC_FILTERING)
+    if (cfg->g_usage == AOM_USAGE_ALL_INTRA) {
+        // Enable AV1E_SET_SKIP_POSTPROC_FILTERING for still-picture encoding, which is
+        // different from libaom's default.
+        aom_codec_control(&codec->internal->encoder, AV1E_SET_SKIP_POSTPROC_FILTERING, 1);
+    }
+#endif
```
wantehchang, line 825: Please move this code earlier, either before line 735, or between line 752 and line 753. You will need to add `#if defined(AOM_USAGE_ALL_INTRA)` around this code.

wantehchang (changes requested):
Remya: Thank you for writing the pull request. I suggest a change.

src/codec_aom.c
RemyaPrakasan: Done.

PR #908, opened 2022-04-05
v0.10.0

wantehchang:

PR #1158, opened 2022-10-12
v0.11.0

wantehchang:


wantehchang:
I am requesting reviews from all of you so that more people will be familiar with the libavif release process.

To review this pull request, please refer to the following:
- [libavif release checklist](https://github.com/AOMediaCodec/libavif/wiki/Release-Checklist)
- The [pull request for v0.10.0](https://github.com/AOMediaCodec/libavif/pull/908)

It was a lot of work to update the changelog for this release. One reason is that it has been six months since the last release, so there are a lot of commits to audit. But we can start a new policy of updating the changelog as we write pull requests that should be logged. This will save us time in future releases.

CHANGELOG.md
```diff
@@ -6,14 +6,50 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  automatically.
+* avifenc: Input Exif orientation is converted to irot/imir by default.
+
 ### Changed
+* Fix memory leaks of metadata on avifenc exit
+* Update the handling of 'lsel' and progressive decoding to AVIF spec v1.1.0
+* Treat an absent lsel and layer_id == 0xFFFF equivalently for backward
+  compatibility with earlier drafts of AVIF spec v1.1.0
+* Set libavif's own default value of cfg.rc_end_usage for libaom
+* Fix https://github.com/AOMediaCodec/libavif/issues/953
```
y-guyon, line 46: The older entries in this CHANGELOG.md file do not link to issues. Maybe remove them or replace them by a sentence ending with ` (#[number])`.

CHANGELOG.md
```diff
@@ -6,14 +6,50 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.11.0] - 2022-10-12
+
 There are incompatible ABI changes in this release. The alphaRange member was
 removed from avifImage struct. The chromaDownsampling and avoidLibYUV members
```
jzern, line 12: ...from the avifImage struct.

jzern (approved):

CHANGELOG.md
y-guyon: One reason I prefer `Set the cqLevelSet boolean correctly (#953)` over `Fix https://github.com/AOMediaCodec/libavif/issues/953` is that if you happen to search for when `cqLevelSet` behavior changed, you can just search for the "cqLevelSet" string in `CHANGELOG.md`. The opposite holds: if you are looking for something in `CHANGELOG.md` and you know it has nothing to do with `cqLevelSet`, it avoids one click.

wantehchang: You are right and these "Fix [bug number]" changelog entries are shortcuts I took when I pored over a lot of commits. I think a good solution is to update the changelog as we write pull requests that should be described in the changelog. Another good solution is to shorten the gap between releases so that there are fewer commits to go through. Or give the release engineer more time to write the changelog.

Issue #1162, opened 2022-10-14
[v0.11.0] Tests fail to build


jbeich:
Regressed by b3e0f315fcf9

```
$ c++ --version
FreeBSD clang version 14.0.5 (https://github.com/llvm/llvm-project.git llvmorg-14.0.5-0-gc12386ae247c)
Target: x86_64-unknown-freebsd14.0
Thread model: posix
InstalledDir: /usr/bin
```
```
$ cmake -DAVIF_BUILD_TESTS=on -B /tmp/libavif_build -G Ninja
$ cmake --build /tmp/libavif_build
[...]
FAILED: tests/avifmetadatatest
: && /usr/bin/c++   tests/CMakeFiles/aviftest_helpers.dir/gtest/aviftest_helpers.cc.o tests/CMakeFiles/avifmetadatatest.dir/gtest/avifmetadatatest.cc.o -o tests/avifmetadatatest  -Wl,-rpath,/usr/local/lib:/tmp/libavif_build  /usr/local/lib/libgtest.so.1.12.1  libavif_apps.a  libavif.so.15.0.0  -lm  /usr/local/lib/libpng.so  /usr/lib/libz.so  /usr/local/lib/libjpeg.so  -pthread && :
ld: error: undefined symbol: avifImageExtractExifOrientationToIrotImir
>>> referenced by avifmetadatatest.cc
>>>               tests/CMakeFiles/avifmetadatatest.dir/gtest/avifmetadatatest.cc.o:(libavif::(anonymous namespace)::MetadataTest_ExifIfdOffsetLoopingTo8_Test::TestBody())
c++: error: linker command failed with exit code 1 (use -v to see invocation)
```

wantehchang:
Jan: Thank you for the bug report. I think I reproduced this bug locally. `tests/avifmetadatatest` must be linked with the static libavif library. I will try to figure out a solution.

A workaround is to pass `-DAVIF_BUILD_APPS=OFF -DAVIF_ENABLE_GTEST=OFF` to cmake. Since `-DAVIF_BUILD_APPS=OFF` is the default, it can be omitted.

Note: When creating the v0.11.0 release, I announced a release candidate (v0.11.0-rc1) two days ago (Wed Oct 12) to try to avoid this kind of build failure. Did you test the release candidate? If you only tested the final release, do you have any suggestions on how to publicize the release candidate better? Should I allow more days for people to test the release candidate?

Thanks.

jbeich:
[...]

PR #1163, opened 2022-10-14
avifmetadatatest: don't include avif/internal.h

wantehchang:
avifmetadatatest can only call public functions because it may be linked with the libavif shared library (when BUILD_SHARED_LIBS is set to ON). Replace the call to the internal function
avifImageExtractExifOrientationToIrotImir() with a call to the public function avifImageSetMetadataExif(), and include avif/avif.h instead of avif/internal.h.

Fix https://github.com/AOMediaCodec/libavif/issues/1162.

### DIFF
--- a/tests/gtest/avifmetadatatest.cc
+++ b/tests/gtest/avifmetadatatest.cc
@@ -4,7 +4,7 @@
 #include <array>
 #include <tuple>
 
-#include "avif/internal.h"
+#include "avif/avif.h"
 #include "aviftest_helpers.h"
 #include "gtest/gtest.h"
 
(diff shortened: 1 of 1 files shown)


jzern (approved):
I guess the build coverage needs to be extended.

y-guyon:
> I guess the build coverage needs to be extended.

https://github.com/AOMediaCodec/libavif/pull/1170 is an attempt.

y-guyon (approved):
LGTM.

I also sent an alternative https://github.com/AOMediaCodec/libavif/pull/1172. It avoids the temporary `TODO`. Feel free to submit #1163 or #1172.

wantehchang:
Thank you. To keep the test coverage of shared-library builds as high as possible, I will merge this pull request.

Issue #1164, opened 2022-10-14
SharpYuv*Bit/RGBToYUVTest.* in avifrgbtoyuvtest fail if libsharpyuv is not found


wantehchang:
Steps to reproduce:

1. Do not install libsharpyuv on the system.
2. In the `ext/` directory, run `./aom.cmd` and `./googletest.cmd`.
3. Create a `build/` directory and change into that directory.
4. Run `cmake -G Ninja .. -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON -DAVIF_BUILD_TESTS=ON -DAVIF_LOCAL_GTEST=ON`
5. Run `ninja`
6. Run `tests/avifrgbtoyuvtest`

The `SharpYuv*Bit/RGBToYUVTest.*` tests fail. Here are two examples:

```
[----------] 144 tests from SharpYuv8Bit/RGBToYUVTest
[ RUN      ] SharpYuv8Bit/RGBToYUVTest.ConvertWholeRange/0
/usr/local/google/home/wtc/libavif.1/libavif/tests/gtest/avifrgbtoyuvtest.cc:206: Failure
Expected equality of these values:
  avifImageRGBToYUV(yuv.get(), &src_rgb)
    Which is: 25
  AVIF_RESULT_OK
    Which is: 0
[  FAILED  ] SharpYuv8Bit/RGBToYUVTest.ConvertWholeRange/0, where GetParam() = (8, 8, 0, 3, 0, 6, 4, true, 17, 1.2, 34) (0 ms)
[ RUN      ] SharpYuv8Bit/RGBToYUVTest.ConvertWholeRange/1
/usr/local/google/home/wtc/libavif.1/libavif/tests/gtest/avifrgbtoyuvtest.cc:206: Failure
Expected equality of these values:
  avifImageRGBToYUV(yuv.get(), &src_rgb)
    Which is: 25
  AVIF_RESULT_OK
    Which is: 0
[  FAILED  ] SharpYuv8Bit/RGBToYUVTest.ConvertWholeRange/1, where GetParam() = (8, 8, 0, 3, 0, 1, 4, true, 17, 1.2, 34) (0 ms)
```

The reason is that libsharpyuv is not found, so we return AVIF_RESULT_NOT_IMPLEMENTED (25) from the following code in `avifImageRGBToYUV()`:

```
    avifBool converted = AVIF_FALSE;

[...]

PR #1166, opened 2022-10-14
avifrgbtoyuvtest: Test AVIF_LIBSHARPYUV_ENABLED

wantehchang:
The test suites (SharpYuv8Bit, SharpYuv10Bit, and SharpYuv12Bit) that use AVIF_CHROMA_DOWNSAMPLING_SHARP_YUV should only be run if the macro AVIF_LIBSHARPYUV_ENABLED is defined, otherwise the avifImageRGBToYUV() calls in those tests will fail with AVIF_RESULT_NOT_IMPLEMENTED (25).

Fix https://github.com/AOMediaCodec/libavif/issues/1164.


wantehchang:
I am abandoning this pull request. I found that the `AVIF_LIBSHARPYUV_ENABLED` macro is only defined when building the libavif library, not when building the tests.

PR #1168, opened 2022-10-15
avifrgbtoyuvtest: Test if libsharpyuv is found

wantehchang:
When compiling avifrgbtoyuvtest.cc, define the macro AVIF_LIBSHARPYUV_ENABLED if libsharpyuv is found.

The test suites (SharpYuv8Bit, SharpYuv10Bit, and SharpYuv12Bit) that use AVIF_CHROMA_DOWNSAMPLING_SHARP_YUV should only be run if the macro AVIF_LIBSHARPYUV_ENABLED is defined, otherwise the avifImageRGBToYUV() calls in those tests will fail with AVIF_RESULT_NOT_IMPLEMENTED (25).

Fix https://github.com/AOMediaCodec/libavif/issues/1164.


tests/CMakeLists.txt
```diff
@@ -99,6 +99,9 @@ if(AVIF_ENABLE_GTEST)
     add_test(NAME avifmetadatatest COMMAND avifmetadatatest ${CMAKE_CURRENT_SOURCE_DIR}/data/)
 
     add_executable(avifrgbtoyuvtest gtest/avifrgbtoyuvtest.cc)
+    if(libsharpyuv_FOUND)
+        target_compile_definitions(avifrgbtoyuvtest PRIVATE -DAVIF_LIBSHARPYUV_ENABLED=1)
+    endif()
```
y-guyon, line 104: What about the following, similar to line 123:

```cmake
else()
    message(STATUS "Some tests are disabled because libsharpyuv is unavailable.")
```

y-guyon (approved):
LGTM.

I also sent an alternative https://github.com/AOMediaCodec/libavif/pull/1171. It avoids modifying the compiler definitions for the test and calls `avifImageRGBToYUV()` when there is no libsharpyuv. The drawback would be a false positive if `avifImageRGBToYUV()` were to return `NOT_IMPLEMENTED` by mistake when there is libsharpyuv (it would just skip the test instead of failing). Feel free to submit #1168 or #1171.

wantehchang:
I abandoned this pull request. It is replaced by https://github.com/AOMediaCodec/libavif/pull/1171.

Yannis: I also considered the approach of https://github.com/AOMediaCodec/libavif/pull/1171 but did not seriously look into it. Since you are the author of tests/gtest/avifrgbtoyuvtest.cc, we should use the fix you suggested.

PR #1169, opened 2022-10-17
Add SHARED libavif CI with locally built dependencies

y-guyon:
Only done for `aom`, `libyuv`, `libsharpyuv`, `googletest` on `ubuntu-latest` as a first step.

It does not work because of libyuv:

> /usr/bin/ld: ../ext/libyuv/build/libyuv.so: undefined reference to `jpeg_read_raw_data'

I do not know the root cause.


jzern:
> Only done for `aom`, `libyuv`, `libsharpyuv`, `googletest` on `ubuntu-latest` as a first step.
> 
> It does not work because of libyuv:
> 
> > /usr/bin/ld: ../ext/libyuv/build/libyuv.so: undefined reference to `jpeg_read_raw_data'
> 
> I do not know the root cause.

The cmake file looks to glob all sources, so it's picking up mjpeg_decoder.cc, which uses libjpeg. One workaround might be to try -Wl,-z,undefs though that's a little risky.

jzern:
> > Only done for `aom`, `libyuv`, `libsharpyuv`, `googletest` on `ubuntu-latest` as a first step.
> > It does not work because of libyuv:
> > > /usr/bin/ld: ../ext/libyuv/build/libyuv.so: undefined reference to `jpeg_read_raw_data'
> > 
> > 
> > I do not know the root cause.
> 
> The cmake file looks to glob all sources, so it's picking up mjpeg_decoder.cc, which uses libjpeg. One workaround might be to try -Wl,-z,undefs though that's a little risky.

Or adding libjpeg.so to the link args should work.

y-guyon:
> adding libjpeg.so to the link args should work.

It did, thanks. Was https://github.com/AOMediaCodec/libavif/pull/1169/commits/3892b4231a282fbd702bb4ada11b7f2d1d0b6276 what you had in mind?

---

I did not manage to make the cached steps work. Maybe the PR must be submitted before enabling caching.

ext/libyuv.cmd
```diff
@@ -19,6 +19,6 @@ git checkout f9fda6e7
 mkdir build
 cd build
 
-cmake -G Ninja -DBUILD_SHARED_LIBS=0 -DCMAKE_BUILD_TYPE=Release ..
+cmake -G Ninja -DCMAKE_BUILD_TYPE=Release ..
```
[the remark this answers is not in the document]
y-guyon: [libyuv's CMakeLists.txt](https://chromium.googlesource.com/libyuv/libyuv/+/8084f0b0edd4b9710b74051cde4d87f360925024/CMakeLists.txt#29) ignores `BUILD_SHARED_LIBS` and generates a `STATIC` `yuv` target and a `SHARED` `yuv_shared` target. I wonder why it was set in the first place in the original commit a9cda518fafcdea37c0127bc10e93a4a005eb295. libyuv's behavior was the same [at that time](https://chromium.googlesource.com/libyuv/libyuv/+/19d71f6b351fe992ae34b114eebd872c383a6bdb/CMakeLists.txt).

jzern:
still lgtm.

.github/workflows/ci-unix-shared.yml
```diff
@@ -0,0 +1,81 @@
+# This is a copy of ci.yml for building shared libraries. It differs from ci.yml as follows:
+#
+#   * The os matrix consists of ubuntu-latest only.
+#   * Does not build dav1d, rav1e, SVT-AV1 nor libgav1.
+
+name: CI (shared libraries)
+on: [push, pull_request]
+jobs:
+  build:
```
wantehchang, line 9: There are two checks with the same name "build (ubuntu-latest)" now. Could you try changing this line to something like "build-shared"? Thanks.

y-guyon: I see the following  when clicking on the green check mark next to the last commit of this PR:

```
CI (shared libraries) / build (ubuntu-latest)
CI / build (ubuntu-latest)
```

Where are you seeing identical names?

wantehchang:
Yannis: I took a screen shot of what I am seeing (on Mac) in this pull request:

Screen Shot 2022-10-20 at 3 02 53 PM

PR #1170, opened 2022-10-17
Add SHARED libavif CI with installed dependencies

y-guyon:
Only done for `aom`, `libyuv`, `googletest` on `ubuntu-latest` as a first step.

It does not work:

> libgtest.a(gtest-all.cc.o): undefined reference to `std::__throw_bad_array_new_length()'

I do not know the root cause. The alternative to installing dependencies would be building them locally, see https://github.com/AOMediaCodec/libavif/pull/1169.

### DIFF
--- a/.github/workflows/ci-unix-shared-installed.yml
+++ b/.github/workflows/ci-unix-shared-installed.yml
@@ -1,5 +1,5 @@
-# This is a copy of ci-unix-shared-local.yml for building shared libraries, but
-# using installed dependencies instead of locally built ones.
+# This is a copy of ci-unix-shared-local.yml for building shared libraries
+# with an additional build configuration (using installed deps and dav1d).
 
 name: CI
 on: [push, pull_request]
@@ -39,7 +39,7 @@ jobs:
     - name: Update apt
       run: sudo apt-get update
     - name: Install dependencies
-      run: sudo apt-get install libaom-dev libyuv-dev
+      run: sudo apt-get install libaom-dev libdav1d-dev libyuv-dev
       # `sudo apt-get install googletest libgtest-dev` leads to the following:
       #   "libgtest.a(gtest-all.cc.o): undefined reference to `std::__throw_bad_array_new_length()'"
       # so build it locally instead.
@@ -55,6 +55,8 @@ jobs:
         cmake .. -G Ninja
         -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON
         -DAVIF_CODEC_AOM=ON
+        -DAVIF_CODEC_AOM_DECODE=OFF -DAVIF_CODEC_AOM_ENCODE=ON
+        -DAVIF_CODEC_DAV1D=ON
         -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON
         -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON -DAVIF_LOCAL_GTEST=ON
     - name: Build libavif (ninja)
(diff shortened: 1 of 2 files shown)


jzern:
> Only done for `aom`, `libyuv`, `googletest` on `ubuntu-latest` as a first step.
> 
> It does not work:
> 
> > libgtest.a(gtest-all.cc.o): undefined reference to `std::__throw_bad_array_new_length()'
> 
> I do not know the root cause. The alternative to installing dependencies would be building them locally, see #1169.

Looks like maybe a compiler version mismatch, e.g., some building with an older gcc and gtest built with a newer one, referencing a new function in libstdc++.

wantehchang:
We can try building GoogleTest locally.

y-guyon:
> We can try building GoogleTest locally.

Done.

---

Not sure that this cl is very useful, now that there is already a shared library CI with locally built dependencies. What do you think?

jzern:
> > We can try building GoogleTest locally.
> 
> Done.
> 
> Not sure that this cl is very useful, now that there is already a shared library CI with locally built dependencies. What do you think?

It is one way to check compatibility with different library versions, though it's of course limited to a single distro.

wantehchang:
James: Joe told me that tests/docker/build.sh is intended to check compatibility with dependencies installed as Linux packages. It seems that  .github/workflows/docker-ci.yml is supposed to run tests/docker/build.sh as a CI check, but it is not running.

.github/workflows/ci-unix-shared-installed.yml
```diff
@@ -0,0 +1,68 @@
...
+        cmake-version: '3.13.x'
+    - name: Print cmake version
+      run: cmake --version
+    - uses: seanmiddleditch/gha-setup-ninja@v3
+    - name: Set shared libs
+      if: steps.cache-ext.outputs.cache-hit != 'true'
+      working-directory: ./ext
+      run: sed -i'' -e 's/-DBUILD_SHARED_LIBS=OFF/-DBUILD_SHARED_LIBS=ON/' *.cmd
+    - name: Add recent repository for libyuv-dev
+      if: steps.cache-ext.outputs.cache-hit != 'true'
```
wantehchang, line 38: Remove this `if` statement, because the side effects (output files) of the run statement (line 39) are not in the cache files. Note that line 24 says the cache files only include the `ext` directory.

For the same reason, remove the `if` statements at line 41 and line 44.

Do keep the `if` statement at line 50 because GoogleTest build output files are in the `ext` directory.

y-guyon:
> James: Joe told me that tests/docker/build.sh is intended to check compatibility with dependencies installed as Linux packages. It seems that .github/workflows/docker-ci.yml is supposed to run tests/docker/build.sh as a CI check, but it is not running.

`tests/docker/build.sh` builds and installs the codecs. It uses the installed `libjpeg-dev libpng-dev` dependencies, but not `libyuv` or `libsharpyuv`.

jzern (approved):
This looks ok to me, but you may want to leave it open for Wan-Teh to comment.

.github/workflows/ci-unix-shared-installed.yml
```diff
@@ -0,0 +1,65 @@
+# This is a copy of ci-unix-shared-local.yml for building shared libraries, but
+# using installed dependencies instead of locally built ones.
```
wantehchang, line 2: The original purpose of ci-unix-shared-local.yml is to test whether we can build libavif itself as a shared library and export all the public API functions from the shared library correctly. Whether libavif also uses the dependencies as shared libraries doesn't really matter to the check of public API function exports. And whether the dependencies are installed on the system or built locally doesn't really matter either. So this new CI build is redundant for the original purpose, but it is useful for testing an additional build configuration, and for that purpose it would be good to test using dav1d as the AV1 decoder.

wantehchang (approved):
LGTM.

wantehchang:
Sorry about the late review. I didn't know James suggested you leave this pull request open for me to comment.

wantehchang (approved):
LGTM.

PR #1172, opened 2022-10-17
Build avifmetadatatest only statically

y-guyon:
Because it includes avif/internal.h which is not exported by the shared library.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -93,29 +93,31 @@ if(AVIF_ENABLE_GTEST)
     target_link_libraries(avifincrtest aviftest_helpers avifincrtest_helpers)
     add_test(NAME avifincrtest COMMAND avifincrtest ${CMAKE_CURRENT_SOURCE_DIR}/data/)
 
-    add_executable(avifmetadatatest gtest/avifmetadatatest.cc)
-    target_link_libraries(avifmetadatatest aviftest_helpers ${GTEST_LIBRARIES})
-    target_include_directories(avifmetadatatest PRIVATE ${GTEST_INCLUDE_DIRS})
-    add_test(NAME avifmetadatatest COMMAND avifmetadatatest ${CMAKE_CURRENT_SOURCE_DIR}/data/)
-
     add_executable(avifrgbtoyuvtest gtest/avifrgbtoyuvtest.cc)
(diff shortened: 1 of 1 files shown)


y-guyon:
See https://github.com/AOMediaCodec/libavif/pull/1163.

(Note: I think the GitHub auto cross-link feature does not work if the link is repeated, because my comment in https://github.com/AOMediaCodec/libavif/pull/1163 should have added a link here in https://github.com/AOMediaCodec/libavif/pull/1172.)

tests/CMakeLists.txt
```diff
@@ -93,29 +93,31 @@ if(AVIF_ENABLE_GTEST)
...
+        target_link_libraries(avifmetadatatest aviftest_helpers ${GTEST_LIBRARIES})
+        target_include_directories(avifmetadatatest PRIVATE ${GTEST_INCLUDE_DIRS})
+        add_test(NAME avifmetadatatest COMMAND avifmetadatatest ${CMAKE_CURRENT_SOURCE_DIR}/data/)
+
         add_executable(aviftilingtest gtest/aviftilingtest.cc)
         target_link_libraries(aviftilingtest avif ${GTEST_BOTH_LIBRARIES})
         target_include_directories(aviftilingtest PRIVATE ${GTEST_INCLUDE_DIRS})
         add_test(NAME aviftilingtest COMMAND aviftilingtest)
+    else()
+        message(STATUS "Some tests are disabled because BUILD_SHARED_LIBS is ON.")
```
wantehchang, line 119: Optional: We could keep this change (print a status message). The message could be made more concrete by changing "Some tests" to "The tests for internal functions".

A better solution is to figure out how to test internal functions in a shared-library build. One solution is to create a STATIC or cmake OBJECT library that tests for internal functons can use.

wantehchang (commented):
Yannis: Thank you for writing this pull request. I also considered this fix but concluded it was not the right trade-off. Unlike aviftilingtest.cc, which only tests the internal function `avifSetTileConfiguration()`, almost all of avifmetadatatest.cc tests public, exported functions. So this pull request will give up the entire avifmetadatatest.cc file for a small part of the file.

The bug reports filed by Jan Beich last Friday show that he does shared-library builds and runs our tests. So we should aim to keep the test coverage of shared-library builds as high as possible.

Issue #1175, opened 2022-10-17
testcmd (--lossless) fails with rav1e


jbeich:
```
      Start  1: aviftest
 1/11 Test  #1: aviftest .........................   Passed    0.10 sec
      Start  2: avifyuv_limited
 2/11 Test  #2: avifyuv_limited ..................   Passed    0.00 sec
      Start  3: avifyuv_rgb
 3/11 Test  #3: avifyuv_rgb ......................   Passed    0.02 sec
      Start  4: avifallocationtest
 4/11 Test  #4: avifallocationtest ...............   Passed    2.57 sec
      Start  5: avifchangesettingtest
 5/11 Test  #5: avifchangesettingtest ............   Passed    0.01 sec
      Start  6: avifgridapitest
 6/11 Test  #6: avifgridapitest ..................   Passed    3.18 sec
      Start  7: avifincrtest
 7/11 Test  #7: avifincrtest .....................   Passed    5.78 sec
      Start  8: avifmetadatatest
 8/11 Test  #8: avifmetadatatest .................   Passed    1.36 sec
      Start  9: avifrgbtoyuvtest
 9/11 Test  #9: avifrgbtoyuvtest .................   Passed    2.62 sec
      Start 10: avify4mtest
10/11 Test #10: avify4mtest ......................   Passed    0.01 sec
      Start 11: test_cmd
11/11 Test #11: test_cmd .........................***Failed    1.06 sec

91% tests passed, 1 tests failed out of 11

Total Test time (real) =  16.72 sec

The following tests FAILED:
         11 - test_cmd (Failed)
```
<details><summary>Verbose output (only failed)</summary>

```
test 11
      Start 11: test_cmd

11: Test command: /usr/local/bin/bash "tests/test_cmd.sh" "../.build" "tests/data"
11: Working Directory: ../.build/tests
11: Test timeout computed to be: 10000000
11: + [[ 2 -ge 1 ]]
11: ++ eval echo ../.build
11: +++ echo ../.build
11: + BINARY_DIR=../.build
11: + [[ 2 -ge 2 ]]
11: ++ eval echo tests/data
11: +++ echo tests/data
11: + TESTDATA_DIR=tests/data
11: + [[ 2 -ge 3 ]]
11: ++ mktemp -d
11: + TMP_DIR=/tmp/tmp.T0zvyqaQ
11: + AVIFENC=../.build/avifenc
11: + AVIFDEC=../.build/avifdec
[...]

PR #1176, opened 2022-10-18
Disable tests that may fail if the codec is not aom

y-guyon:
For example, the SVT-AV1 encoder requires 8-bit, 4:2:0 images with even dimensions of at least 64x64 px.

The first commit is just for convenience. Hide it for better diffs.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -117,6 +117,15 @@ if(AVIF_ENABLE_GTEST)
     target_include_directories(avify4mtest PRIVATE ${GTEST_INCLUDE_DIRS})
     add_test(NAME avify4mtest COMMAND avify4mtest)
 
+    if(NOT AVIF_CODEC_AOM OR NOT AVIF_CODEC_AOM_ENCODE OR NOT AVIF_CODEC_AOM_DECODE)
+        # These tests are supported with aom being the encoder and decoder. If aom is unavailable,
+        # these tests are disabled because other codecs may not implement all the necessary features.
+        # For example, SVT-AV1 requires 8-bit, 4:2:0 images with even dimensions of at least 64x64 px.
+        set_tests_properties(avifallocationtest avifgridapitest avifmetadatatest avifincrtest PROPERTIES DISABLED True)
+
+        message(STATUS "Some tests are disabled because aom is unavailable for encoding or decoding.")
+    endif()
+
     if(NOT libsharpyuv_FOUND)
         message(STATUS "Some tests are skipped because libsharpyuv is unavailable.")
     endif()
@@ -135,4 +144,20 @@ if(AVIF_BUILD_APPS)
     add_test(NAME test_cmd COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/test_cmd.sh ${CMAKE_BINARY_DIR}
                                    ${CMAKE_CURRENT_SOURCE_DIR}/data
     )
+    add_test(NAME test_cmd_lossless COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/test_cmd_lossless.sh ${CMAKE_BINARY_DIR}
+                                            ${CMAKE_CURRENT_SOURCE_DIR}/data
+    )
+    add_test(NAME test_cmd_metadata COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/test_cmd_metadata.sh ${CMAKE_BINARY_DIR}
+                                            ${CMAKE_CURRENT_SOURCE_DIR}/data
+    )
+
+    if(NOT AVIF_CODEC_AOM OR NOT AVIF_CODEC_AOM_ENCODE)
+        # Only aom encoder supports lossless encoding.
+        set_property(TEST test_cmd_lossless PROPERTY DISABLED True)
+
+        # SVT-AV1 does not support the images with odd dimensions that are used in this test.
+        set_property(TEST test_cmd_metadata PROPERTY DISABLED True)
+
+        message(STATUS "Some tests are disabled because aom is unavailable for encoding.")
+    endif()
 endif()
--- a/tests/test_cmd_lossless.sh
+++ b/tests/test_cmd_lossless.sh
@@ -1,0 +1,104 @@
+#!/bin/bash
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+#
+# tests for command lines
+
+# Very verbose but useful for debugging.
+set -ex
+
+if [[ "$#" -ge 1 ]]; then
+  # eval so that the passed in directory can contain variables.
+  BINARY_DIR="$(eval echo "$1")"
+else
+  # Assume "tests" is the current directory.
+  BINARY_DIR="$(pwd)/.."
+fi
+if [[ "$#" -ge 2 ]]; then
+  TESTDATA_DIR="$(eval echo "$2")"
+else
+  TESTDATA_DIR="$(pwd)/data"
+fi
+if [[ "$#" -ge 3 ]]; then
+  TMP_DIR="$(eval echo "$3")"
+else
+  TMP_DIR="$(mktemp -d)"
+fi
+
+AVIFENC="${BINARY_DIR}/avifenc"
+AVIFDEC="${BINARY_DIR}/avifdec"
+ARE_IMAGES_EQUAL="${BINARY_DIR}/tests/are_images_equal"
+
+# Basic calls.
+"${AVIFENC}" --version
+"${AVIFDEC}" --version
+
+# Input file paths.
+INPUT_Y4M="${TESTDATA_DIR}/kodim03_yuv420_8bpc.y4m"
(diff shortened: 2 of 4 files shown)


tests/test_cmd_metadata.sh
```diff
@@ -38,55 +38,23 @@ else
...
-  "${ARE_IMAGES_EQUAL}" "${DECODED_FILE}" "${DECODED_FILE_LOSSLESS}" 0
-
-  # Argument parsing test with filenames starting with a dash.
-  echo "Testing arguments"
-  "${AVIFENC}" -s 10 "${INPUT_PNG}" -- "${ENCODED_FILE_WITH_DASH}"
-  "${AVIFDEC}" --info  -- "${ENCODED_FILE_WITH_DASH}"
-  # Passing a filename starting with a dash without using -- should fail.
-  "${AVIFENC}" -s 10 "${INPUT_PNG}" "${ENCODED_FILE_WITH_DASH}" && exit 1
-  "${AVIFDEC}" --info "${ENCODED_FILE_WITH_DASH}" && exit 1
-
```
wantehchang: Nit: In line 91 `echo "Testing metadata enc/dec"`, should we remove `/dec` from the message? Or would you like to keep it for future tests?

tests/test_cmd_lossless.sh
```diff
@@ -40,65 +40,29 @@ fi
...
-    cmp "${ENCODED_FILE}" "${ENCODED_FILE_NO_METADATA}" && exit 1
-    "${AVIFENC}" "${INPUT}" -o "${ENCODED_FILE_NO_METADATA}" --ignore-exif
-    cmp "${ENCODED_FILE}" "${ENCODED_FILE_NO_METADATA}" && exit 1
-    "${AVIFENC}" "${INPUT}" -o "${ENCODED_FILE_NO_METADATA}" --ignore-xmp
-    cmp "${ENCODED_FILE}" "${ENCODED_FILE_NO_METADATA}" && exit 1
-  done
+  # TODO(yguyon): Make this test pass with IGNORE_ALPHA_FLAG set to 0.
+  "${AVIFENC}" -s 10 -l "${INPUT_PNG}" -o "${ENCODED_FILE}"
+  "${AVIFDEC}" "${ENCODED_FILE}" "${DECODED_FILE}"
+  "${ARE_IMAGES_EQUAL}" "${DECODED_FILE}" "${DECODED_FILE}" "${IGNORE_ALPHA_FLAG}"
```
wantehchang, line 65: IMPORTANT: Both arguments are the same (`${DECODED_FILE}`). One of the arguments should be changed to `${INPUT_PNG}`.

It would be good for this pull request to focus on fixing https://github.com/AOMediaCodec/libavif/issues/1175 and not attempt to fix the original TODO `TODO(yguyon): Make this test pass with INPUT_PNG instead of DECODED_FILE`.

tests/CMakeLists.txt
```diff
@@ -135,4 +144,20 @@ if(AVIF_BUILD_APPS)
...
+    add_test(NAME test_cmd_metadata COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/test_cmd_metadata.sh ${CMAKE_BINARY_DIR}
+                                            ${CMAKE_CURRENT_SOURCE_DIR}/data
+    )
+
+    if(NOT AVIF_CODEC_AOM OR NOT AVIF_CODEC_AOM_ENCODE)
+        # Only aom encoder supports lossless encoding.
+        set_property(TEST test_cmd_lossless PROPERTY DISABLED True)
+
+        # SVT-AV1 does not support the images with odd dimensions that are used in this test.
+        set_property(TEST test_cmd_metadata PROPERTY DISABLED True)
```
wantehchang, line 159: Just wondering if we can use `set_tests_properties(test_cmd_lossless test_cmd_metadata PROPERTIES DISABLED True)` here.

Did you disable the tests separately so that it's easier to give the reason why each test is disabled?

tests/CMakeLists.txt
```diff
@@ -135,4 +144,20 @@ if(AVIF_BUILD_APPS)
...
+    )
+    add_test(NAME test_cmd_metadata COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/test_cmd_metadata.sh ${CMAKE_BINARY_DIR}
+                                            ${CMAKE_CURRENT_SOURCE_DIR}/data
+    )
+
+    if(NOT AVIF_CODEC_AOM OR NOT AVIF_CODEC_AOM_ENCODE)
+        # Only aom encoder supports lossless encoding.
+        set_property(TEST test_cmd_lossless PROPERTY DISABLED True)
+
+        # SVT-AV1 does not support the images with odd dimensions that are used in this test.
```
wantehchang, line 158: Optional: Should we disable this test under the following additional condition?

```
        # SVT-AV1 does not support the images with odd dimensions that are used in this test.
        if(NOT AVIF_CODEC_RAV1E)
            set_property(TEST test_cmd_metadata PROPERTY DISABLED True)
        endif()
```

tests/CMakeLists.txt
```diff
@@ -117,6 +117,15 @@ if(AVIF_ENABLE_GTEST)
     target_include_directories(avify4mtest PRIVATE ${GTEST_INCLUDE_DIRS})
     add_test(NAME avify4mtest COMMAND avify4mtest)
 
+    if(NOT AVIF_CODEC_AOM OR NOT AVIF_CODEC_AOM_ENCODE OR NOT AVIF_CODEC_AOM_DECODE)
+        # These tests are supported with aom being the encoder and decoder. If aom is unavailable,
+        # these tests are disabled because other codecs may not implement all the necessary features.
+        # For example, SVT-AV1 requires 8-bit, 4:2:0 images with even dimensions of at least 64x64 px.
```
wantehchang, line 123: Nit/optional: This comment could be misinterpreted to mean SVT-AV1 requires all of those conditions to be true. After reading this comment, I was wondering if SVT-AV1 did not support 10-bit. But I think it does.

tests/CMakeLists.txt
```diff
@@ -117,6 +117,15 @@ if(AVIF_ENABLE_GTEST)
...
+        set_tests_properties(avifallocationtest avifgridapitest avifmetadatatest avifincrtest PROPERTIES DISABLED True)
```
wantehchang, line 124: [Just a comment, no change requested.] Disabling these tests when libaom is not available is a drastic measure. This is appropriate as a simple solution in the v0.11.1 patch release.

wantehchang (changes requested):
Yannis: Thank you very much for writing this pull request. There is a typo in tests/test_cmd_lossless.sh (marked with "IMPORTANT").

tests/CMakeLists.txt
y-guyon: > Just wondering if we can use `set_tests_properties(test_cmd_lossless test_cmd_metadata PROPERTIES DISABLED True)` here.

We can.

> Did you disable the tests separately so that it's easier to give the reason why each test is disabled?

Yes.

tests/CMakeLists.txt
y-guyon: I got a segfault when encoding a 10-bit 64x64 4:2:0 image with SVT-AV1 but I did not take the time to investigate. Maybe I missed something. I removed the 8-bit part.

tests/test_cmd_lossless.sh
```diff
@@ -0,0 +1,72 @@
...
+trap cleanup EXIT
+
+pushd ${TMP_DIR}
+  # Generate test data.
+  "${AVIFENC}" -s 8 "${INPUT_Y4M}" -o "${ENCODED_FILE}"
+  "${AVIFDEC}" "${ENCODED_FILE}" "${DECODED_FILE}"
+
+  # Lossless test. The decoded pixels should be the same as the original image.
+  echo "Testing basic lossless"
+  # TODO(yguyon): Make this test pass with INPUT_Y4M instead of DECODED_FILE.
```
wantehchang, line 66: I looked into this. We can't encode `INPUT_Y4M ` losslessly because `INPUT_Y4M ` is not YUV 4:4:4. The filename `kodim03_yuv420_8bpc.y4m` implies it is YUV 4:2:0. Since the standard AVIF lossless compression takes RGB input, it is better to use a PNG file as input.

I suggest we remove this TODO comment or change `INPUT_Y4M ` back to `INPUT_PNG` (which requires adding back the `INPUT_PNG="${TESTDATA_DIR}/paris_icc_exif_xmp.png"` line).

wantehchang (approved):
LGTM with a change.

tests/CMakeLists.txt
wantehchang: SVT-AV1 supports 10-bit:
https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/Docs/CommonQuestions.md#8-or-10-bit-encoding

The segfault may mean there is a bug in libavif/src/codec_svt.c.

wantehchang (approved):
Still LGTM.

tests/test_cmd_lossless.sh
```diff
@@ -42,11 +42,11 @@ AVIFDEC="${BINARY_DIR}/avifdec"
...
 # Input file paths.
-INPUT_Y4M="${TESTDATA_DIR}/kodim03_yuv420_8bpc.y4m"
+INPUT_PNG="${TESTDATA_DIR}/paris_icc_exif_xmp.png"
 # Output file names.
-ENCODED_FILE="avif_test_cmd_encoded.avif"
-DECODED_FILE="avif_test_cmd_decoded.png"
-DECODED_FILE_LOSSLESS="avif_test_cmd_decoded_lossless.png"
+ENCODED_FILE="avif_test_cmd_lossless_encoded.avif"
+DECODED_FILE="avif_test_cmd_lossless_decoded.png"
+DECODED_FILE_LOSSLESS="avif_test_cmd_lossless_decoded_lossless.png"
```
wantehchang, line 49: Yannis: These tmp files are all created in `${TMP_DIR}`, which is unique. So the tmp file names should not need to be unique, right?

y-guyon: Right, I confused the behavior with the .cc ones. I guess `/tmp` might still be given as an argument for `${TMP_DIR}` on some build systems so it does not hurt.

PR #1179, opened 2022-10-20
Do not allow parameter overrides in lossless.

vrabaud:
This patch fixes:
- the fact that the order of parameters matters for lossless
- the fact that we allow a mix of -l and parameters contradictory with lossless

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -827,11 +827,9 @@ int main(int argc, char * argv[])
             fprintf(stderr, "Codec can only be AOM in lossless mode.\n");
             returnCode = 1;
         }
-        // rav1e doesn't support lossless
-        // transform yet:
+        // rav1e doesn't support lossless transform yet:
         // https://github.com/xiph/rav1e/issues/151
-        // SVT-AV1 doesn't support lossless
-        // encoding yet:
+        // SVT-AV1 doesn't support lossless encoding yet:
         // https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1636
         codecChoice = AVIF_CODEC_CHOICE_AOM;
         // Range.
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -814,11 +814,11 @@ int main(int argc, char * argv[])
                     "mode.\n");
             returnCode = 1;
         }
-        input.requestedFormat = AVIF_PIXEL_FORMAT_YUV444; // don't subsample when using
-                                                          // AVIF_MATRIX_COEFFICIENTS_IDENTITY
+        // Don't subsample when using AVIF_MATRIX_COEFFICIENTS_IDENTITY.
+        input.requestedFormat = AVIF_PIXEL_FORMAT_YUV444;
         // Quantizers.
-        if (minQuantizer != -1 || maxQuantizer != -1 || minQuantizerAlpha != -1 || maxQuantizerAlpha != -1) {
-            fprintf(stderr, "Quantizers cannot be set in lossless mode.\n");
+        if (minQuantizer > 0 || maxQuantizer > 0 || minQuantizerAlpha > 0 || maxQuantizerAlpha > 0) {
+            fprintf(stderr, "Quantizers cannot be set in lossless mode, except to 0.\n");
             returnCode = 1;
         }
         minQuantizer = maxQuantizer = minQuantizerAlpha = maxQuantizerAlpha = AVIF_QUANTIZER_LOSSLESS;
@@ -827,12 +827,13 @@ int main(int argc, char * argv[])
             fprintf(stderr, "Codec can only be AOM in lossless mode.\n");
             returnCode = 1;
         }
-        codecChoice = AVIF_CODEC_CHOICE_AOM; // rav1e doesn't support lossless
-                                             // transform yet:
-                                             // https://github.com/xiph/rav1e/issues/151
-                                             // SVT-AV1 doesn't support lossless
-                                             // encoding yet:
-                                             // https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1636
+        // rav1e doesn't support lossless
+        // transform yet:
+        // https://github.com/xiph/rav1e/issues/151
+        // SVT-AV1 doesn't support lossless
+        // encoding yet:
+        // https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1636
+        codecChoice = AVIF_CODEC_CHOICE_AOM;
         // Range.
         if (requestedRange != AVIF_RANGE_FULL) {
             fprintf(stderr, "Range has to be full in lossless mode.\n");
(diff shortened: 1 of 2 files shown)


jzern:
I didn't review the patch that closely, but it looks like this could break existing workflows if the quantizer is set (but ignored) currently.

vrabaud:
Indeed, that is a behavioral change. Do we want/allow that ?

The next step is to also change the following behavior: if you set the quantizer to lossless (and also the mC to identiy), but do not ask for lossless, avifenc internally switches to lossless but then checks again if parameters are compatible with lossless, prints a warning if they are incompatible and continues. I propose to remove this magic guessing next.

jzern:
> Indeed, that is a behavioral change. Do we want/allow that ?
> 

Introducing it as a warning first might make sense. [aomenc](https://aomedia.googlesource.com/aom/+/refs/tags/v3.5.0/common/warnings.c#70) has some examples like this.

> The next step is to also change the following behavior: if you set the quantizer to lossless (and also the mC to identiy), but do not ask for lossless, avifenc internally switches to lossless but then checks again if parameters are compatible with lossless, prints a warning if they are incompatible and continues. I propose to remove this magic guessing next.

wantehchang:
@joedrago Joe: I seem to remember you wrote the code that's modified in this pull request. It'd be good if you could take a look. Thanks.

apps/avifenc.c
```diff
@@ -805,6 +792,63 @@ int main(int argc, char * argv[])
...
+        if (input.requestedFormat != AVIF_PIXEL_FORMAT_NONE && input.requestedFormat != AVIF_PIXEL_FORMAT_YUV444) {
+            fprintf(stderr,
+                    "When set, the pixel format can only be 444 in lossless "
+                    "mode.\n");
+            returnCode = 1;
+        }
+        input.requestedFormat = AVIF_PIXEL_FORMAT_YUV444; // don't subsample when using
+                                                          // AVIF_MATRIX_COEFFICIENTS_IDENTITY
+        // Quantizers.
+        if (minQuantizer != -1 || maxQuantizer != -1 || minQuantizerAlpha != -1 || maxQuantizerAlpha != -1) {
```
wantehchang, line 807: We should also allow the quantizers to be set to `AVIF_QUANTIZER_LOSSLESS`.

Note that you allow the other parameters (`input.requestedFormat`, `codecChoice`, and `matrixCoefficients`) to be set to the correct values.

apps/avifenc.c
```diff
@@ -805,6 +792,63 @@ int main(int argc, char * argv[])
...
+    if (lossless) {
+        // Pixel format.
+        if (input.requestedFormat != AVIF_PIXEL_FORMAT_NONE && input.requestedFormat != AVIF_PIXEL_FORMAT_YUV444) {
+            fprintf(stderr,
+                    "When set, the pixel format can only be 444 in lossless "
+                    "mode.\n");
+            returnCode = 1;
+        }
+        input.requestedFormat = AVIF_PIXEL_FORMAT_YUV444; // don't subsample when using
+                                                          // AVIF_MATRIX_COEFFICIENTS_IDENTITY
```
wantehchang, line 805: Nit: I suggest moving the comment before the assignment statement.

```
        // don't subsample when using AVIF_MATRIX_COEFFICIENTS_IDENTITY
        input.requestedFormat = AVIF_PIXEL_FORMAT_YUV444;
```

In the original code, there are multiple assignment statements and the comments are placed after the statements. We don't need to do that in the new code.

apps/avifenc.c
```diff
@@ -805,6 +792,63 @@ int main(int argc, char * argv[])
...
+        if (codecChoice != AVIF_CODEC_CHOICE_AUTO && codecChoice != AVIF_CODEC_CHOICE_AOM) {
+            fprintf(stderr, "Codec can only be AOM in lossless mode.\n");
+            returnCode = 1;
+        }
+        codecChoice = AVIF_CODEC_CHOICE_AOM; // rav1e doesn't support lossless
+                                             // transform yet:
+                                             // https://github.com/xiph/rav1e/issues/151
+                                             // SVT-AV1 doesn't support lossless
+                                             // encoding yet:
+                                             // https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1636
```
wantehchang, line 822: Nit: Similarly, I suggest moving the comment before the assignment statement:

```
        // rav1e doesn't support lossless transform yet:
        // https://github.com/xiph/rav1e/issues/151
        // SVT-AV1 doesn't support lossless encoding yet:
        // https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1636
        codecChoice = AVIF_CODEC_CHOICE_AOM;
```

apps/avifenc.c
```diff
@@ -773,19 +773,6 @@ int main(int argc, char * argv[])
             }
         } else if (!strcmp(arg, "-l") || !strcmp(arg, "--lossless")) {
             lossless = AVIF_TRUE;
-
-            // Set defaults, and warn later on if anything looks incorrect
```
wantehchang, line 777: The new code only allows correct parameters to be used with lossless. It seems that this pull request should also remove the code that the comment "warn later on if anything looks incorrect" refers to. I took a quick look and it seems more complicated than I thought because we sometimes set `lossless` to `AVIF_TRUE`.

wantehchang (changes requested):
Hi Vincent. All the changes look reaosnable except that this pull request doesn't allow `-l --min 0 --max 0`. Please see my comment at line 807.

apps/avifenc.c
vrabaud: Done

apps/avifenc.c
vrabaud: Done

apps/avifenc.c
vrabaud: Done

apps/avifenc.c
vrabaud: Indeed, I'd rather do it in a different pull request.

apps/avifenc.c
```diff
@@ -818,6 +805,62 @@ int main(int argc, char * argv[])
...
+        if (input.requestedFormat != AVIF_PIXEL_FORMAT_NONE && input.requestedFormat != AVIF_PIXEL_FORMAT_YUV444) {
+            fprintf(stderr,
+                    "When set, the pixel format can only be 444 in lossless "
+                    "mode.\n");
+            returnCode = 1;
+        }
+        // Don't subsample when using AVIF_MATRIX_COEFFICIENTS_IDENTITY.
+        input.requestedFormat = AVIF_PIXEL_FORMAT_YUV444;
+        // Quantizers.
+        if (minQuantizer > 0 || maxQuantizer > 0 || minQuantizerAlpha > 0 || maxQuantizerAlpha > 0) {
```
wantehchang, line 820: Nit: It may be better to use `AVIF_QUANTIZER_LOSSLESS` instead of 0 here to match line 824. I didn't make this change because it may require running clang-format on this file.

Note: The error message in line 821 should use 0 though because that's what's used in command lines.

wantehchang (approved):
LGTM. I am sorry for the late review. I didn't realize this hasn't been reviewed and merged.

I made two simple edits and will merge this pull request after it passes the CI checks.

PR #1180, opened 2022-10-20
Simplify test of AV1E_SET_SKIP_POSTPROC_FILTERING

wantehchang:
If AV1E_SET_SKIP_POSTPROC_FILTERING is available, the macro AOM_USAGE_ALL_INTRA must be defined. So the defined(AOM_USAGE_ALL_INTRA) test can be omitted.

Also edit the comment.

PR #1181, opened 2022-10-20
Use different cache file for shared library build

wantehchang:
Add "-shared" to the cache file name for the shared library build.

Make a simple change to ext/googletest.cmd to force the cache files to be recreated.

### DIFF
--- a/.github/workflows/ci-unix-shared.yml
+++ b/.github/workflows/ci-unix-shared.yml
@@ -27,7 +27,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ext
-        key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}
+        key: ${{ runner.os }}-shared-${{ hashFiles('ext/*.cmd') }}
     - name: Setup cmake
       uses: jwlawson/actions-setup-cmake@v1.12
       with:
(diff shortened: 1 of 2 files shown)


ext/googletest.cmd
```diff
@@ -15,5 +15,5 @@ cd build
 : # on Windows:
 : # https://github.com/google/googletest/blob/main/googletest/README.md#visual-studio-dynamic-vs-static-runtimes
 cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DBUILD_GMOCK=OFF -Dgtest_force_shared_crt=ON ..
-cmake --build .
+ninja
```
wantehchang, line 18: I am thinking about adding an `ext/touch.cmd` file that we can modify to force the cache files to be recreated.

.github/workflows/ci-unix-shared.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ext
-        key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}
+        key: ${{ runner.os }}-shared-${{ hashFiles('ext/*.cmd') }}
```
wantehchang, line 30: I hope that the `hashFiles('ext/*.cmd')` call uses the contents of the `ext/*.cmd` files before we modify them. It seems to this way.

wantehchang:
I wrote this pull request because I got a CI build failure in https://github.com/AOMediaCodec/libavif/pull/1180:
https://github.com/AOMediaCodec/libavif/actions/runs/3293552693/jobs/5430172707

y-guyon (approved):

ext/googletest.cmd
jzern: Would deleting the existing cache be enough?

https://docs.github.com/en/rest/actions/cache#delete-a-github-actions-cache-for-a-repository-using-a-cache-id

cmake --build . and ninja should be the same thing here, is this for consistency with other files?

wantehchang: I didn't know we can delete the existing cache. Thank you for the link to the doc.

I made this change to ext/googletest.cmd simply to force the cache files to be recreated. I didn't see any comment error that I could fix, so I just changed `cmake --build .` to an equivalent command. It wasn't really for consistency with other files.

PR #1183, opened 2022-10-21
Use more avifImagePlane*() functions

y-guyon:
To reduce and simplify the codebase.


apps/avifenc.c
```diff
@@ -375,38 +375,20 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
...
-                    &gridSplitImage->yuvPlanes[AVIF_CHAN_U][(gridX * bytesPerRowUV) + (gridY * uvHeight) * srcRowBytesU];
-                cellImage->yuvRowBytes[AVIF_CHAN_U] = srcRowBytesU;
-
-                const uint32_t srcRowBytesV = gridSplitImage->yuvRowBytes[AVIF_CHAN_V];
-                cellImage->yuvPlanes[AVIF_CHAN_V] =
-                    &gridSplitImage->yuvPlanes[AVIF_CHAN_V][(gridX * bytesPerRowUV) + (gridY * uvHeight) * srcRowBytesV];
-                cellImage->yuvRowBytes[AVIF_CHAN_V] = srcRowBytesV;
+            const size_t cellY = (size_t)gridY * cellHeight;
+
+            int lastYuvPlane = (gridSplitImage->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) ? AVIF_CHAN_Y : AVIF_CHAN_V;
```
jzern, line 380: this could be const

src/avif.c
```diff
@@ -179,44 +179,42 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
     avifRWDataSet(&dstImage->exif, srcImage->exif.data, srcImage->exif.size);
     avifImageSetMetadataXMP(dstImage, srcImage->xmp.data, srcImage->xmp.size);
 
+    int firstPlane = AVIF_CHAN_A; // No plane by default.
```
jzern, line 182: this should be stable, but an assert that A > V might clear up the meaning of the comment

src/write.c
```diff
@@ -728,100 +728,75 @@ static avifResult avifEncoderDataCreateXMPItem(avifEncoderData * data, const avi
...
-        memcpy(dstRow, dstRow - dstRowBytes, (size_t)dstWidth * sampleByteCount);
-    }
-}
-
 // Same as avifImageCopy() but pads the dstImage with border pixel values to reach dstWidth and dstHeight.
 static avifImage * avifImageCopyAndPad(const avifImage * srcImage, uint32_t dstWidth, uint32_t dstHeight)
 {
     avifImage * dstImage = avifImageCreate(dstWidth, dstHeight, srcImage->depth, srcImage->yuvFormat);
 
+    int firstPlane = AVIF_CHAN_A; // No plane by default.
```
jzern, line 736: same comment here for a potential assert addition

tests/gtest/avifincrtest_helpers.cc
```diff
@@ -33,37 +33,30 @@ void ComparePartialYuva(const avifImage& image1, const avifImage& image2,
...
-    const uint32_t width_byte_count = image1.width * pixel_byte_count;
-    const uint8_t* data1 = image1.alphaPlane;
-    const uint8_t* data2 = image2.alphaPlane;
-    for (uint32_t y = 0; y < row_count; ++y) {
-      ASSERT_EQ(std::memcmp(data1, data2, width_byte_count), 0);
-      data1 += image1.alphaRowBytes;
-      data2 += image2.alphaRowBytes;
+  }
+
+  int lastPlane = image1.alphaPlane ? AVIF_CHAN_A : AVIF_CHAN_V;
```
jzern, line 46: this could be const

apps/avifenc.c
```diff
@@ -375,38 +375,20 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
...
-                const uint32_t srcRowBytesU = gridSplitImage->yuvRowBytes[AVIF_CHAN_U];
-                cellImage->yuvPlanes[AVIF_CHAN_U] =
-                    &gridSplitImage->yuvPlanes[AVIF_CHAN_U][(gridX * bytesPerRowUV) + (gridY * uvHeight) * srcRowBytesU];
-                cellImage->yuvRowBytes[AVIF_CHAN_U] = srcRowBytesU;
-
-                const uint32_t srcRowBytesV = gridSplitImage->yuvRowBytes[AVIF_CHAN_V];
-                cellImage->yuvPlanes[AVIF_CHAN_V] =
-                    &gridSplitImage->yuvPlanes[AVIF_CHAN_V][(gridX * bytesPerRowUV) + (gridY * uvHeight) * srcRowBytesV];
-                cellImage->yuvRowBytes[AVIF_CHAN_V] = srcRowBytesV;
+            const size_t cellY = (size_t)gridY * cellHeight;
```
wantehchang, line 378: This code was written before `avifImageSetViewRect()` existed. I believe we can rewrite this code to use `avifImageSetViewRect()`.

apps/shared/y4m.c
```diff
@@ -361,42 +361,23 @@ avifBool y4mRead(const char * inputFilename, avifImage * avif, avifAppSourceTimi
...
-    const int planeCount = formatInfo.monochrome ? 1 : AVIF_PLANE_COUNT_YUV;
-    for (int plane = 0; plane < planeCount; ++plane) {
-        uint32_t planeWidth = (plane > 0) ? ((avif->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX) : avif->width;
-        uint32_t planeHeight = (plane > 0) ? ((avif->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY) : avif->height;
-        uint32_t rowBytes = planeWidth << (avif->depth > 8);
-        uint8_t * row = avif->yuvPlanes[plane];
+    const int lastPlane = frame.hasAlpha ? AVIF_CHAN_A : AVIF_CHAN_V;
+    for (int plane = AVIF_CHAN_Y; plane <= lastPlane; ++plane) {
+        uint32_t planeHeight = avifImagePlaneHeight(avif, plane);
+        uint32_t sampleBytes = avifImagePlaneWidth(avif, plane) << (avif->depth > 8);
```
wantehchang, line 367: Nit/optional: `sampleBytes` could be misinterpreted to mean the sample size in bytes, i.e., 1 or 2 bytes. We could name this variable `widthBytes` or `planeWidthBytes` (i.e., the width in bytes).

UPDATE: You named an equivalent variable `width_byte_count` in tests/gtest/avifincrtest_helpers.cc. That name is similar to my suggestion.

src/avif.c
```diff
@@ -179,44 +180,43 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
     avifRWDataSet(&dstImage->exif, srcImage->exif.data, srcImage->exif.size);
     avifImageSetMetadataXMP(dstImage, srcImage->xmp.data, srcImage->xmp.size);
 
+    assert(AVIF_CHAN_A > AVIF_CHAN_V);
```
wantehchang, line 183: Optional: This is probably a matter of preference. Although the original code is longer, it is more direct and easier to understand.

The new code splits the allocation of the plane buffers and the data copying, and the for loop that copies the data contains a special case for the YUV planes.

wantehchang (commented):
Hi Yannis,

Please consider my review as supplemental. I didn't review all the code in the pull request. You can merge this pull request with James' approval.

src/avif.c
y-guyon: > The new code splits the allocation of the plane buffers and the data copying

I find it more logical to allocate all the planes first, to avoid copying anything if the allocation failed.

> the for loop that copies the data contains a special case for the YUV planes.

It is actually a special case for UV only, so doing it during the YUV or the YUVA loop does not change much the issue.

apps/avifenc.c
y-guyon: This can be done in another PR. This code should be reworked to allow smaller right/bottom cells anyway.

src/codec_aom.c
```diff
@@ -216,9 +216,6 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
         image->transferCharacteristics = (avifTransferCharacteristics)codec->internal->image->tc;
         image->matrixCoefficients = (avifMatrixCoefficients)codec->internal->image->mc;
 
-        avifPixelFormatInfo formatInfo;
-        avifGetPixelFormatInfo(yuvFormat, &formatInfo);
```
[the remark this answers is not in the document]
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1184.

apps/avifenc.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1189.

src/avif.c
```diff
@@ -179,44 +180,43 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
...
+    }
+    for (int plane = firstPlane; plane <= lastPlane; ++plane) {
+        if ((plane != AVIF_CHAN_A) && !srcImage->yuvRowBytes[plane]) {
+            // plane is absent. If we're copying from a source without
+            // them, mimic the source image's state by removing our copy.
+            avifFree(dstImage->yuvPlanes[plane]);
+            dstImage->yuvPlanes[plane] = NULL;
+            dstImage->yuvRowBytes[plane] = 0;
+            continue;
+        }
```
wantehchang, line 208: I think it is cleaner to move this if statement (lines 201-208) into the allocation code above, because this is really part of the allocation code. I.e., add the following for loop after line 190:

```
        for (int yuvPlane = 1; yuvPlane < 3; ++yuvPlane) {
            if (!srcImage->yuvRowBytes[yuvPlane]) {
                // plane is absent. If we're copying from a source without
                // them, mimic the source image's state by removing our copy.
                avifFree(dstImage->yuvPlanes[yuvPlane]);
                dstImage->yuvPlanes[yuvPlane] = NULL;
                dstImage->yuvRowBytes[yuvPlane] = 0;
            }
        }
```

Then, the for loop here (which starts from line 200), which should only copy the planes, just needs to skip a null plane pointer.

wantehchang:
Yannis: I didn't know you were waiting for my review. Sorry. Could you resolve the merge conflicts in apps/avifenc.c? Thanks.

wantehchang (changes requested):
Yannis: I suggest some changes.

apps/shared/avifjpeg.c
```diff
@@ -185,11 +185,9 @@ static avifBool avifJPEGReadCopy(avifImage * avif, struct jpeg_decompress_struct
...
                     }
 
-                    avifPixelFormatInfo info;
-                    avifGetPixelFormatInfo(avif->yuvFormat, &info);
-                    uint32_t uvHeight = (avif->height + info.chromaShiftY) >> info.chromaShiftY;
-                    memset(avif->yuvPlanes[AVIF_CHAN_U], 128, avif->yuvRowBytes[AVIF_CHAN_U] * uvHeight);
-                    memset(avif->yuvPlanes[AVIF_CHAN_V], 128, avif->yuvRowBytes[AVIF_CHAN_V] * uvHeight);
+                    uint32_t uvHeight = avifImagePlaneHeight(avif, AVIF_CHAN_U);
+                    memset(avif->yuvPlanes[AVIF_CHAN_U], 128, (size_t)avif->yuvRowBytes[AVIF_CHAN_U] * uvHeight);
+                    memset(avif->yuvPlanes[AVIF_CHAN_V], 128, (size_t)avif->yuvRowBytes[AVIF_CHAN_V] * uvHeight);
```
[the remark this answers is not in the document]
y-guyon: Removed the for loop.

apps/shared/y4m.c
```diff
@@ -361,42 +361,23 @@ avifBool y4mRead(const char * inputFilename, avifImage * avif, avifAppSourceTimi
...
-    for (int plane = 0; plane < planeCount; ++plane) {
-        uint32_t planeWidth = (plane > 0) ? ((avif->width + formatInfo.chromaShiftX) >> formatInfo.chromaShiftX) : avif->width;
-        uint32_t planeHeight = (plane > 0) ? ((avif->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY) : avif->height;
-        uint32_t rowBytes = planeWidth << (avif->depth > 8);
-        uint8_t * row = avif->yuvPlanes[plane];
+    const int lastPlane = frame.hasAlpha ? AVIF_CHAN_A : AVIF_CHAN_V;
+    for (int plane = AVIF_CHAN_Y; plane <= lastPlane; ++plane) {
+        uint32_t planeHeight = avifImagePlaneHeight(avif, plane);
+        uint32_t planeWidthBytes = avifImagePlaneWidth(avif, plane) << (avif->depth > 8);
+        uint8_t * row = avifImagePlane(avif, plane);
```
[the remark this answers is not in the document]
y-guyon: > Alternatively we can check if `planeHeight` is 0.

This is already done with `for (uint32_t y = 0; y < planeHeight; ++y)`. I added comments to make that clearer.

> Note: The for loop may be able to just iterate over `plane` = 0, 1, 2, 3 blindly without checking for `frame.hasAlpha`. But I didn't look into this.

Done.

src/avif.c
```diff
@@ -179,44 +180,47 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
...
             return allocationResult;
         }
-
-        avifPixelFormatInfo formatInfo;
-        avifGetPixelFormatInfo(srcImage->yuvFormat, &formatInfo);
-        uint32_t uvHeight = (dstImage->height + formatInfo.chromaShiftY) >> formatInfo.chromaShiftY;
-        for (int yuvPlane = 0; yuvPlane < 3; ++yuvPlane) {
-            uint32_t planeHeight = (yuvPlane == AVIF_CHAN_Y) ? dstImage->height : uvHeight;
-
+        for (int yuvPlane = 1; yuvPlane < 3; ++yuvPlane) {
```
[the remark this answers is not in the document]
y-guyon: > Note: I took a closer look at the code, and I believe we can delete this for loop (lines 191-199) because of the `avifImageFreePlanes(dstImage, AVIF_PLANES_ALL)` call at line 175. But that would be a cleanup change and you don't need to make that change in this pull request.

I believe this code was originally meant for badly initialized images such as only `yuvPlanes[AVIF_CHAN_V]` being `NULL` for a 4:4:4 image. `avifImageAllocatePlanes()` only cares about the `format`, so the pointers are checked separately in `avifImageCopy()`. It should be replaced with a `if (planeHeight && !srcRow) { return AVIF_RESULT_INVALID_ARGUMENT; }` in another PR.

I removed it from `avifImageCopyAndPad()`, because `avifImageCopyAndPad()` is not a public API function.

tests/gtest/avifincrtest_helpers.cc
```diff
@@ -33,37 +33,30 @@ void ComparePartialYuva(const avifImage& image1, const avifImage& image2,
...
+    const size_t width_byte_count =
+        avifImagePlaneWidth(&image1, plane) * pixel_byte_count;
+    const uint32_t height =
+        (plane == AVIF_CHAN_Y || plane == AVIF_CHAN_A) ? row_count : uv_height;
+    const uint8_t* row1 = avifImagePlane(&image1, plane);
+    const uint8_t* row2 = avifImagePlane(&image2, plane);
+    const uint32_t row1_bytes = avifImagePlaneRowBytes(&image1, plane);
+    const uint32_t row2_bytes = avifImagePlaneRowBytes(&image2, plane);
+    for (uint32_t y = 0; y < height; ++y) {
+      ASSERT_EQ(std::memcmp(row1, row2, width_byte_count), 0);
```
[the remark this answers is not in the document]
y-guyon: Fixed by setting `uv_height` to `0` if `info.monochrome`.

src/write.c
```diff
@@ -794,55 +754,71 @@ static avifImage * avifImageCopyAndPad(const avifImage * srcImage, uint32_t dstW
...
+    }
+    for (int plane = firstPlane; plane <= lastPlane; ++plane) {
+        if ((plane != AVIF_CHAN_A) && !srcImage->yuvRowBytes[plane]) {
+            // Plane is absent. If we're copying from a source without
+            // them, mimic the source image's state by removing our copy.
+            avifFree(dstImage->yuvPlanes[plane]);
+            dstImage->yuvPlanes[plane] = NULL;
+            dstImage->yuvRowBytes[plane] = 0;
+            continue;
+        }
```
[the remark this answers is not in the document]
y-guyon: See https://github.com/AOMediaCodec/libavif/pull/1183#discussion_r1023834588.

src/write.c
```diff
@@ -800,49 +762,50 @@ static avifImage * avifImageCopyAndPad(const avifImage * srcImage, uint32_t dstW
...
+        const uint32_t dstPlaneWidth = avifImagePlaneWidth(dstImage, plane);
+        const uint32_t dstPlaneHeight = avifImagePlaneHeight(dstImage, plane); // 0 for A if no alpha and 0 for UV if 4:0:0.
+        const size_t dstPlaneWidthBytes = (size_t)dstPlaneWidth << (dstImage->depth > 8);
+
+        for (uint32_t j = 0; j < srcPlaneHeight; ++j) {
+            memcpy(dstRow, srcRow, srcPlaneWidthBytes);
+
+            // Pad columns.
+            if (dstPlaneWidth > srcPlaneWidth) {
+                if (avifImageUsesU16(srcImage)) {
```
wantehchang, line 791: Nit: Here we use `avifImageUsesU16(srcImage)`, but we use `(srcImage->depth > 8)` in line 778 and `(dstImage->depth > 8)` in line 784. These three expressions have the same value. It would be good to use the same expression in all three places, or save the value in a local variable and use the variable in all three places.

We should at least use the same expression in line 778 and here.

wantehchang (approved):
LGTM. I also realized later that the inner for loops skip the nonexistent planes because `planeHeight` is 0. Sorry about my misunderstanding.

src/avif.c
```diff
@@ -184,39 +185,36 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
...
-            uint8_t * dstAlphaRow = &dstImage->alphaPlane[j * dstImage->alphaRowBytes];
-            memcpy(dstAlphaRow, srcAlphaRow, dstImage->alphaRowBytes);
+    }
+    for (int plane = AVIF_CHAN_Y; plane <= AVIF_CHAN_A; ++plane) {
+        uint32_t planeHeight = avifImagePlaneHeight(dstImage, plane);
+        uint8_t * srcRow = avifImagePlane(srcImage, plane);
+        uint8_t * dstRow = avifImagePlane(dstImage, plane);
+        if (!dstRow) {
+            continue;
+        }
```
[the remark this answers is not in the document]
y-guyon: As I mentioned in https://github.com/AOMediaCodec/libavif/pull/1183#discussion_r1023834588, I believe this code is for `avifImage`s with `Y` and `U` not `NULL` and `V` `NULL`, for example. So checking the output of `avifImagePlaneHeight()` is not enough, because `avifImagePlaneHeight()` does not return `0` based on whether `yuvPlanes[V]` is `NULL` or not.

This way of handling bad input should be changed but this is beyond the scope of this PR.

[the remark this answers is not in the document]
wantehchang: Thanks for the explanation.

In an earlier commit (https://github.com/AOMediaCodec/libavif/pull/1183/commits/68556c5e3740827b7ad50510bf63b3753511fad5) you were checking `srcRow` (presumably because I suggested it). I am just wondering why it's better to check `dstRow`. Both checks seem equivalent to me.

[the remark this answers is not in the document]
wantehchang: I think the answer to my question is that the code allows a 4:0:0 `srcImage` to have non-null U and V planes, so we need to check `dstRow` instead.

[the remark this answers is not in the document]
y-guyon: > I think the answer to my question is that the code allows a 4:0:0 `srcImage` to have non-null U and V planes, so we need to check `dstRow` instead.

Yes, and also the `avifPlanesFlags` argument may not be `AVIF_PLANES_ALL`.

PR #1186, opened 2022-10-24
Add Exif/XMP support in PNG/JPEG export

y-guyon:
Change avifGetExifTiffHeaderOffset()'s signature.
Add avifImageGetExifOrientationFromIrotImir() and avifSetExifOrientation() in internal.h.
Add some src/*.c as source files of CMake target avif_apps, otherwise symbols are missing when building shared libraries.


CMakeLists.txt
```diff
@@ -494,7 +494,7 @@ if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND AVIF_ENABLE_GTEST))
 
     add_library(
         avif_apps STATIC apps/shared/avifjpeg.c apps/shared/iccjpeg.c apps/shared/avifpng.c apps/shared/avifutil.c
-                         apps/shared/y4m.c
+                         apps/shared/y4m.c src/diag.c src/exif.c src/stream.c src/utils.c
```
y-guyon, line 497: Alternative: link to `avif_internal` from https://github.com/AOMediaCodec/libavif/pull/1185 instead of adding source files.

apps/shared/avifpng.c
```diff
@@ -368,11 +384,33 @@ avifBool avifPNGRead(const char * inputFilename,
...
+// Does the opposite of avifCopyRawProfile(): writes a payload as a raw profile string.
+static avifBool avifBytesToRawProfile(const avifRWData * bytes, const char * profileName, avifRWData * profile)
+{
+    if (bytes->size > 99999999) {
+        return AVIF_FALSE;
+    }
+    size_t position = 1 + strlen(profileName) + 1 + 8 + 1;
+    const size_t profileLength = position + bytes->size * 2 + 1;
+    avifRWDataRealloc(profile, profileLength);
+    sprintf((char *)profile->data, "\n%s\n%08lu\n", profileName, (unsigned long)bytes->size);
```
jzern, line 399: snprintf would likely be preferred here and line 401; sprintf will produce warnings in some configurations.

apps/shared/avifpng.c
```diff
@@ -448,6 +486,41 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
+            //       the raw profile zTXt method.
+        }
+        if (avif->xmp.data && (avif->xmp.size > 0)) {
+            if (!avifBytesToRawProfile(&avif->xmp, "XMP", &xmpRawProfile)) {
+                fprintf(stderr, "Error writing PNG: XMP metadata is too big\n");
+                goto cleanup;
+            }
+            const png_text text = {
+                .compression = PNG_TEXT_COMPRESSION_zTXt,
+                .key = "Raw profile type xmp",
```
kmilos, line 515: Same for XMP: the only standardized way of storing it is in the uncompressed (and not coded) iTXt chunk w/ the "XML:com.adobe.xmp" key (see XMP spec part 3, section 1.1.5). Let's drop the raw profiles if possible.

apps/shared/avifjpeg.c
```diff
@@ -571,9 +581,79 @@ avifBool avifJPEGWrite(const char * outputFilename, const avifImage * avif, int
...
+            'I', 'I', 42,  0,         // TIFF header (little endian)
+            8,   0,   0,   0,         // offset to 0th IFD (starting at TIFF header)
+            1,   0,                   // field count
+            18,  1,                   // tag
+            3,   0,                   // type
+            1,   0,   0,   0,         // count
+            rot, 0,   0,   0,         // value offset
+            0,   0,   0,   0          // offset of next IFD (1st IFD is not recorded)
+        };
+        jpeg_write_marker(&cinfo, JPEG_APP0 + 1, exifWithJustOrientation, sizeof(exifWithJustOrientation));
```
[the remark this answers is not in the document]
y-guyon: > I don't think one can write just the Orientation

So [`paris_exif_orientation_5.jpg`](https://github.com/AOMediaCodec/libavif/tree/main/tests/data#file-paris_exif_orientation_5jpg) is invalid too. It was generated with `exiv2 -k -M "set Exif.Image.Orientation 5"`. It displays fine in Chrome.

> Leave without any Exif info then (like the PNG)?

Done. Added `TODO`s.

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+* avifImageGetExifOrientationFromIrotImir() and avifSetExifOrientation() were
+  added to internal.h.
+
+### Changed
+* Exif and XMP metadata is exported to PNG and JPEG files.
```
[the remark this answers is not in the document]
y-guyon: I changed the behavior to drop the entire XMP payload and display a warning, instead of failing the export.

The alternative would be doing as libheif: ignore any extended XMP at JPEG import, to avoid any issue at JPEG export. The downside is that the whole XMP payload of [`dog_exif_extended_xmp_icc.jpg`](https://github.com/AOMediaCodec/libavif/tree/main/tests/data#file-dog_exif_extended_xmp_iccjpg) fits into a standard XMP chunk; the small enough extended XMP part would be dropped. I guess big XMP payloads are rare enough to keep the current behavior, even if different from libheif.

apps/shared/avifjpeg.c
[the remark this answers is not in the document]
kmilos: > So [`paris_exif_orientation_5.jpg`](https://github.com/AOMediaCodec/libavif/tree/main/tests/data#file-paris_exif_orientation_5jpg) is invalid too.

Strictly speaking yes, though I haven't come across any SW that actually checks for presence of all these "mandatory" tags... As you found, it "should" work out (famous last words)...

apps/shared/avifpng.c
```diff
@@ -368,11 +384,33 @@ avifBool avifPNGRead(const char * inputFilename,
     return readResult;
 }
 
+//------------------------------------------------------------------------------
+// Writing
+
+// Does the opposite of avifCopyRawProfile(): writes a payload as a raw profile string.
+static avifBool avifBytesToRawProfile(const avifRWData * bytes, const char * profileName, avifRWData * profile)
```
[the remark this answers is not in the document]
y-guyon: At first I wanted to avoid implementing both, because currently there is no guarantee that the used libpng version has `PNG_eXIf_SUPPORTED` and/or `PNG_iTXt_SUPPORTED`. I added these but I also kept the raw profile fallback if they are not available. There is also the possibility of a `\0` in the given XMP, which is forbidden in `iTXt` (`zTXt` raw profile fallback here too).

The alternative would be requiring `PNG_eXIf_SUPPORTED` and `PNG_iTXt_SUPPORTED` at compile time, and failing PNG exports of XMP with `\0`. Bumping the minimum libpng version would not be so bad I guess? This can be done in another PR.

apps/shared/avifpng.c
```diff
@@ -462,6 +566,9 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
             rowPointers[y] = &rgb.pixels[y * rgb.rowBytes];
         }
     }
+    if (avifImageGetExifOrientationFromIrotImir(avif) != 1) {
+        // TODO(yguyon): Rotate the samples.
```
kmilos, line 570: ..or also some day construct a minimal Exif like for JPEGs - as PNG is also a "compressed" format, the mandatory tags are the same (see also overlap w/ the [pHYs chunk](https://www.w3.org/TR/png/#11pHYs)).

OTOH, there is no guarantee that all 3rd party SW will access and use the Exif info in a PNG...

y-guyon: According to the Extensions to the PNG 1.2 Specification, Version 1.5.0, section 3.7:

> It is recommended that unless a decoder has independent knowledge of the validity of the Exif data, the data should be considered to be of historical value only.

So Exif orientation should *not* be taken into account when displaying a PNG.

kmilos: Ah, good find.

So I guess this means you always (if irot/imir not 1) want to rotate the PNG output, and you always want to reset Exif Orientation to 1 (if prior Exif data exists)?

See also https://zpl.fi/exif-orientation-in-different-formats/

y-guyon: Yes.

CMakeLists.txt
jzern: This should be possible now that the change has been merged.

jzern (approved):
This looks good, I didn't look that closely at the tests or the mapping in `avifImageGetExifOrientationFromIrotImir`.

apps/shared/avifpng.c
```diff
@@ -448,6 +487,71 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
+#else
+        if (!avifBytesToRawProfile(&avif->exif, "Exif", &exif)) {
+            fprintf(stderr, "Error writing PNG: Exif metadata is too big\n");
+            goto cleanup;
+        }
+        const png_text text = {
+            .compression = PNG_TEXT_COMPRESSION_zTXt,
+            .key = "Raw profile type exif",
+            .text = (char *)exif.data,
+            .text_length = exif.size,
```
wantehchang, line 509: Older versions of Microsoft Visual Studio may not support a designated initializer. Unless we have tested this code with Visual Studio 2019 and 2017, it is best to replace this with a traditional initializer. Our CI Windows build uses Visual Studio 2022.

jzern: From 2013 onward designated initializers should work with C99 code. The issues we've seen elsewhere are in C++ where they were only added in C++20; I'm not sure what level of support VS has for those.

jzern: 2019 seems to be the first to give a warning:
... error C7555: use of designated initializers requires at least '/std:c++20'

apps/shared/avifpng.c
```diff
@@ -368,11 +384,34 @@ avifBool avifPNGRead(const char * inputFilename,
...
+static avifBool avifBytesToRawProfile(const avifRWData * bytes, const char * profileName, avifRWData * profile)
+{
+    // The width of the profile length is 8 characters.
+    if (bytes->size > 99999999) {
+        return AVIF_FALSE;
+    }
+    size_t position = 1 + strlen(profileName) + 1 + 8 + 1;
+    const size_t profileLength = position + bytes->size * 2 + 1;
+    avifRWDataRealloc(profile, profileLength);
+    snprintf((char *)profile->data, position + 1, "\n%s\n%08lu\n", profileName, (unsigned long)bytes->size);
```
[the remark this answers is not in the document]
y-guyon: I guess it needs `AVIF_FMT_ZU` and this is not convenient to prepend with zeros till 8 characters.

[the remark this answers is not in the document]
kmilos: Yeah, it's too bad that macro wasn't defined w/o the leading `%`, like the inttypes.h print macros...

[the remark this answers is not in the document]
kmilos: Actually, changing `AVIF_FMT_ZU` over to match inttypes syntax might not be a bad idea, nor a big task...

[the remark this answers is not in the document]
jzern: +1, but that can come after this lands if you want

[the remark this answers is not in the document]
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1192.

apps/shared/avifpng.c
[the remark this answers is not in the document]
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1193.

CMakeLists.txt
```diff
@@ -511,7 +511,7 @@ if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND AVIF_ENABLE_GTEST))
         avif_apps STATIC apps/shared/avifjpeg.c apps/shared/iccjpeg.c apps/shared/avifpng.c apps/shared/avifutil.c
                          apps/shared/y4m.c
     )
-    target_link_libraries(avif_apps avif ${AVIF_PLATFORM_LIBRARIES} ${PNG_LIBRARY} ${ZLIB_LIBRARY} ${JPEG_LIBRARY})
+    target_link_libraries(avif_apps avif_internal ${AVIF_PLATFORM_LIBRARIES} ${PNG_LIBRARY} ${ZLIB_LIBRARY} ${JPEG_LIBRARY})
```
wantehchang, line 514: The code in apps/shared/ is considered as a client of the libavif library and should call the public functions of libavif. So `avif_apps` should not depend on the `avif_internal` library. Another problem with this change is that if `BUILD_SHARED_LIBS` is on, then the `avifenc` and `avifdec` apps will be linked with two copies of libavif. The first copy is the shared `avif` library, and the second copy is the static `avif_internal` library (via `apps_shared`).

I haven't fully analyzed the code in this pull request, but a good strategy is to move as many of the Exif functions declared in internal.h as we can to apps/shared/ (say a new apps/shared/avifexif.c file), and export some of the remaining Exif functions as public API functions (and declare them in avif.h). This may require refactoring the Exif functions.

The `avif_internal` library should only be used by unit tests of internal functions.

y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1199.

PR #1188, opened 2022-10-25
Fix avifImageCopyAndPad()

y-guyon:
Other fields than width, height, depth and yuvFormat were not properly propagated to padded cells.
Add regression test.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -771,7 +771,13 @@ static void avifCopyAndPadPlane(uint8_t * dstPlane,
 // Same as avifImageCopy() but pads the dstImage with border pixel values to reach dstWidth and dstHeight.
 static avifImage * avifImageCopyAndPad(const avifImage * srcImage, uint32_t dstWidth, uint32_t dstHeight)
 {
-    avifImage * dstImage = avifImageCreate(dstWidth, dstHeight, srcImage->depth, srcImage->yuvFormat);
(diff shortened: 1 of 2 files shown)


src/write.c
```diff
@@ -771,7 +771,13 @@ static void avifCopyAndPadPlane(uint8_t * dstPlane,
 // Same as avifImageCopy() but pads the dstImage with border pixel values to reach dstWidth and dstHeight.
 static avifImage * avifImageCopyAndPad(const avifImage * srcImage, uint32_t dstWidth, uint32_t dstHeight)
 {
-    avifImage * dstImage = avifImageCreate(dstWidth, dstHeight, srcImage->depth, srcImage->yuvFormat);
+    avifImage * dstImage = avifImageCreateEmpty();
+    // Copy all fields but do not allocate.
+    if (avifImageCopy(dstImage, srcImage, (avifPlanesFlag)0) != AVIF_RESULT_OK) {
+        assert(AVIF_FALSE);
```
jzern, line 777: should this return NULL to be safe or do the call sites guarantee this should never fail?

jzern (approved):

src/write.c
y-guyon: Currently `avifImageCopy()` will return `AVIF_RESULT_OK` if its `planes` argument is set to `0`, but maybe in the future there will be more allocation checks so I changed it.

PR #1189, opened 2022-10-26
Allow padding in avifenc avifImageSplitGrid()

y-guyon:
So that images with dimensions that cannot be evenly split into the number of cells specified with the --grid flag can still be encoded with the right-most and bottom-most cells padded.
Add test_cmd_grid.sh.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -340,25 +340,67 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
     return AVIF_TRUE;
 }
 
-static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gridCols, uint32_t gridRows, avifImage ** gridCells)
+// Returns the best cell size for a given horizontal or vertical dimension.
+static avifBool avifTryGetBestCellSize(const char * dimensionStr, uint32_t numPixels, uint32_t numCells, avifBool isSubsampled, uint32_t * cellSize)
 {
-    if ((gridSplitImage->width % gridCols) != 0) {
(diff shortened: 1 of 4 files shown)


apps/avifenc.c
```diff
@@ -340,25 +340,67 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
...
+    assert(((uint64_t)(numCells - 1) * *cellSize < (uint64_t)numPixels) && ((uint64_t)numCells * *cellSize >= (uint64_t)numPixels));
+    return AVIF_TRUE;
+}
+
+static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gridCols, uint32_t gridRows, avifImage ** gridCells)
+{
+    uint32_t cellWidth, cellHeight;
+    avifPixelFormatInfo formatInfo;
+    avifGetPixelFormatInfo(gridSplitImage->yuvFormat, &formatInfo);
+    if (!avifTryGetBestCellSize("horizontally", gridSplitImage->width, gridCols, formatInfo.chromaShiftX, &cellWidth) ||
```
jzern, line 402: 'Try' might be a bit redundant given the function returns an error.

jzern (approved):
For display of these images are there any decoder updates needed?

y-guyon:
> For display of these images are there any decoder updates needed?

No, besides https://github.com/AOMediaCodec/libavif/pull/1188 which fixes the library, not only `avifdec`. This PR #1189 only allows the tool `avifenc` to use the library's full grid encoding capabilities.

jzern (approved):

PR #793, opened 2021-11-08
change encoder speed in gdk-pixbuf plug-in

novomesk:
I chose encoder speed 8 for the plug-in in the past.

Encoders were relatively slow at that time and speed 8 gave quite acceptable waiting time.

Today, encoders get faster and the speed settings in AOM was remapped.

So I am sure we can change the 8 at least to 7 to slightly improve visual quality of the saved file.

PR #1191, opened 2022-10-27
change encoder speed in gdk-pixbuf plug-in

novomesk:
Speed 6 is now recommended default, encoders got improved. It seems better choice now than 7.

### DIFF
--- a/contrib/gdk-pixbuf/loader.c
+++ b/contrib/gdk-pixbuf/loader.c
@@ -466,7 +466,7 @@ static gboolean avif_image_saver(FILE          *f,
     encoder->maxQuantizer = max_quantizer;
     encoder->minQuantizerAlpha = 0;
     encoder->maxQuantizerAlpha = alpha_quantizer;
-    encoder->speed = 7;
+    encoder->speed = 6;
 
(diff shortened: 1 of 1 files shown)


jzern:
This matches avifenc now, though @wantehchang I thought you had planned to change the default to 7?

novomesk:
In January, he [recommended](https://github.com/AOMediaCodec/libavif/pull/793#discussion_r784432921) speed 6.

jzern:
Sorry, I didn't get back to this yesterday. There's no reason to block the merge, I just wanted Wan-Teh to be aware of the change.

contrib/gdk-pixbuf/loader.c
```diff
@@ -466,7 +466,7 @@ static gboolean avif_image_saver(FILE          *f,
     encoder->maxQuantizer = max_quantizer;
     encoder->minQuantizerAlpha = 0;
     encoder->maxQuantizerAlpha = alpha_quantizer;
-    encoder->speed = 7;
+    encoder->speed = 6;
```
wantehchang, line 469: Yes, I recommend that we use the same speed here as the default encoder speed in avifenc.c (which is 6 right now).

PR #1192, opened 2022-10-28
Remove % from AVIF_FMT_ZU

y-guyon:
To match the inttypes.h print macros.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -1275,8 +1275,8 @@ int main(int argc, char * argv[])
     }
 
     printf("Encoded successfully.\n");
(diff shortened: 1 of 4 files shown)


kmilos:
Could also now handle https://github.com/y-guyon/libavif/blob/bc276e7edb2d7aa16a4baec2ec434966fca2a9ef/apps/shared/avifpng.c#L400

y-guyon:
> Could also now handle https://github.com/y-guyon/libavif/blob/bc276e7edb2d7aa16a4baec2ec434966fca2a9ef/apps/shared/avifpng.c#L400

The function is removed in https://github.com/AOMediaCodec/libavif/pull/1193.

PR #1193, opened 2022-10-28
Bump minimum libpng dependency to 1.6.31

y-guyon:
Remove PNG_eXIf_SUPPORTED and PNG_iTXt_SUPPORTED conditionals in avifpng.c to simplify the code and to avoid generating new PNGs with legacy raw profile chunks.
Reject PNG export with XMP data containing any null character.

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -14,6 +14,10 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if !defined(PNG_eXIf_SUPPORTED) || !defined(PNG_iTXt_SUPPORTED)
+#error "libpng 1.6.31 or above is required."
+#endif
+
 // See libpng-manual.txt, section XI.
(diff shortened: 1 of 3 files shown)


apps/shared/avifpng.c
```diff
@@ -14,6 +14,10 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if !defined(PNG_eXIf_SUPPORTED) || !defined(PNG_iTXt_SUPPORTED)
+#error "libpng 1.6.31 or above is required."
```
jzern, line 18: I suppose those could be disabled in newer releases depending on the configuration, but that should be rare.

jzern (approved):
The required release is from 2017, so this should be relatively safe and it's only in the example code. We could make PNG support optional and allow a build of avifenc to have a reduced set of input codecs.

kmilos:
I think you want 1.6.32, eXIf support was incomplete in 1.6.31 AFICT from the log...

PR #1194, opened 2022-11-04
Add clli metadata read support

madmanchan:
The purpose of this pull request and commit is to add read (decode) support for the Content Light Level Information box as described in MIAF section 7.4.4.2 and AVIF section 2.2.3. This metadata is important when tone mapping HDR images, especially when using the PQ transfer curve.

This functionality is needed to fix this Chromium bug:

https://bugs.chromium.org/p/chromium/issues/detail?id=1380533

### DIFF
--- a/apps/shared/avifutil.c
+++ b/apps/shared/avifutil.c
@@ -121,6 +121,7 @@ static void avifImageDumpInternal(const avifImage * avif, uint32_t gridCols, uin
         }
     }
     printf(" * Progressive    : %s\n", avifProgressiveStateToString(progressiveState));
+    printf(" * CLLI           : %u, %u\n", (uint32_t)avif->clli.maxCLL, (uint32_t)avif->clli.maxPALL);
 }
 
 void avifImageDump(const avifImage * avif, uint32_t gridCols, uint32_t gridRows, avifProgressiveState progressiveState)
--- a/apps/shared/avifutil.c
+++ b/apps/shared/avifutil.c
@@ -121,7 +121,9 @@ static void avifImageDumpInternal(const avifImage * avif, uint32_t gridCols, uin
         }
     }
     printf(" * Progressive    : %s\n", avifProgressiveStateToString(progressiveState));
-    printf(" * CLLI           : %u, %u\n", (uint32_t)avif->clli.maxCLL, (uint32_t)avif->clli.maxPALL);
+    if (avif->clli.maxCLL > 0 || avif->clli.maxPALL > 0) {
+        printf(" * CLLI           : %u, %u\n", (uint32_t)avif->clli.maxCLL, (uint32_t)avif->clli.maxPALL);
+    }
 }
 
 void avifImageDump(const avifImage * avif, uint32_t gridCols, uint32_t gridRows, avifProgressiveState progressiveState)
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -424,6 +424,14 @@ AVIF_API avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * cla
                                                       avifPixelFormat yuvFormat,
                                                       avifDiagnostics * diag);
 
+typedef struct avifContentLightLevelInformationBox
+{
+    // 'clli' from ISO/IEC 23000-22:2019 (MIAF) 7.4.4.2.2
+    
+    uint16_t maxCLL;           // unsigned int(16) max_content_light_level
+    uint16_t maxPALL;          // unsigned int(16) max_pic_average_light_level
+} avifContentLightLevelInformationBox;
+
 // ---------------------------------------------------------------------------
 // avifImage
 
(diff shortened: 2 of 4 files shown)


include/avif/avif.h
```diff
@@ -472,6 +480,11 @@ typedef struct avifImage
     avifImageRotation irot;
     avifImageMirror imir;
 
+    // CLLI information - Content light level information. Used to represent
```
joedrago, line 483: Style nit: Please wrap all block comments in `avif.h` to 100 characters.

Please document what the defaults/absent value (presumably 0,0) should mean, and how it should be interpreted.

include/avif/avif.h
```diff
@@ -424,6 +424,14 @@ AVIF_API avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * cla
                                                       avifPixelFormat yuvFormat,
                                                       avifDiagnostics * diag);
 
+typedef struct avifContentLightLevelInformationBox
+{
+    // 'clli' from ISO/IEC 23000-22:2019 (MIAF) 7.4.4.2.2
+    
+    uint16_t maxCLL;           // unsigned int(16) max_content_light_level
```
joedrago, line 431: Please run clang-format on all modified files.

src/read.c
```diff
@@ -1759,6 +1760,18 @@ static avifBool avifParseColourInformationBox(avifProperty * prop, uint64_t rawO
...
 
+static avifBool avifParseContentLightLevelInformationBox(avifProperty * prop, const uint8_t * raw, size_t rawLen, avifDiagnostics * diag)
+{
+    BEGIN_STREAM(s, raw, rawLen, diag, "Box[clli]");
+
+    avifContentLightLevelInformationBox *clli = &prop->u.clli;
+
+    AVIF_CHECK(avifROStreamReadU16(&s, &clli->maxCLL));
+    AVIF_CHECK(avifROStreamReadU16(&s, &clli->maxPALL));
+
```
joedrago, line 1771: Style Nit: Removing unnecessary blank line above the return.

Additionally, if (0,0) indicate the default/absent/ignore value, should this be considered an error if (0,0) is discovered here?

apps/shared/avifutil.c
```diff
@@ -121,6 +121,7 @@ static void avifImageDumpInternal(const avifImage * avif, uint32_t gridCols, uin
         }
     }
     printf(" * Progressive    : %s\n", avifProgressiveStateToString(progressiveState));
+    printf(" * CLLI           : %u, %u\n", (uint32_t)avif->clli.maxCLL, (uint32_t)avif->clli.maxPALL);
```
joedrago, line 124: If the values are the default/absent/ignore values of (0,0), should this print anything at all, or perhaps some other/additional string indicating that the CLLI is Absent? I'd like to manage enduser expectations a bit here.

joedrago (changes requested):
Thank you for your contribution. I have made some suggestions inline.

include/avif/avif.h
madmanchan: Done.  I added comment to explain that (0, 0) means unknown or unavailable.

include/avif/avif.h
madmanchan: Done.

src/read.c
madmanchan: MIAF 7.4.4.2.1 indicates that that zero values are permitted to indicate an unknown value, so I feel it should not be considered an error.

apps/shared/avifutil.c
madmanchan: I agree that (0, 0) should be ignored here.  I've added a conditional to skip the print of the CLLI unless at least one of the clli fields is non-zero.

joedrago:
Excellent, thank you for these changes. Unfortunately after seeing them, I realized I wasn't thorough enough on various subtleties in patterns in libavif to greenlight this as-is, but rather than causing you additional latency on getting this feature (as I know you're working to get it in Chrome), I'll explain the reasoning behind the changes I want alongside an actual patch proposal for them. You can apply the changes locally in your branch, and if you agree with them, push one final commit to the PR and I'll accept it.

Here's the reasoning behind the additional changes in my patch file:

* Despite not offering CLLI functionality in `avifenc` yet (which is perfectly reasonable), I thought it might surprise new libavif library endusers if they set these values in an `avifImage` and did not see it reflected in their output encodes. We could either document that this structure only has a read implementation in the header, or we could just add the property write logic (it is only a few lines of code). I chose the latter.

* The documentation for struct members in `avif.h` are supposed to describe the _meaning_ of the values, not the internal box payload structure for them. That documentation should be adjacent to the read/write implementations for those members. I've added documentation to the `avifContentLightLevelInformationBox` members directly from the HEVC standard's explanation of those fields, and I've moved the payload details to the read/write portions of the code.

* I made a minor tweak to the `clli` member struct comment in `avifImage`, letting future libavif library users know that leaving these values as `(0,0)` will avoid emitting the `clli` box entirely.

* I added a separator/header in `avif.h` for the `avifContentLightLevelInformationBox` struct to not put it under the `avifCropRect` headfer.

Again, thank you for the contribution, and I'd like to have this all resolved/merged today.

(I had to zip the patch to get GitHub to let me attach it.)
[clli_write.zip](https://github.com/AOMediaCodec/libavif/files/9954991/clli_write.zip)

joedrago (approved):

PR #1198, opened 2022-11-08
Save the return value of a function call

wantehchang:
Save the return value of avifImageGetExifOrientationFromIrotImir() in a local variable to avoid a repeated call to the same function.


jzern (approved):

y-guyon (approved):
`avifImageGetExifOrientationFromIrotImir()` is just a bunch of conditions. This looks like premature optimization to me but alright.

Edit: It is called twice only when there is an error, so it is not even an optimization. I guess it is more style-related, so please disregard my above comment.

wantehchang:
Yes! The primary reason I wrote this pull request is to make it easier to verify the same orientation value is used in those two places. The function name "avifImageGetExifOrientationFromIrotImir" is very long. (I wrote a similar pull request for libheif for the same reason: https://github.com/strukturag/libheif/pull/689.) Saving the cost of a repeated call is just a secondary concern.

PR #1199, opened 2022-11-08
Move some Exif functions to apps/shared/avifexif

y-guyon:
So that the avif_apps target does not depend on avif_internal.

### DIFF
--- a/apps/shared/avifexif.c
+++ b/apps/shared/avifexif.c
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC. All rights reserved.
+// Copyright 2022 Google LLC
 // SPDX-License-Identifier: BSD-2-Clause
 
 #include "avifexif.h"
--- a/apps/shared/avifexif.c
+++ b/apps/shared/avifexif.c
@@ -1,0 +1,63 @@
+// Copyright 2022 Google LLC. All rights reserved.
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "avifexif.h"
+#include "avif/avif.h"
+
+uint8_t avifImageGetExifOrientationFromIrotImir(const avifImage * image)
+{
+    if ((image->transformFlags & AVIF_TRANSFORM_IROT) && (image->irot.angle == 1)) {
+        if (image->transformFlags & AVIF_TRANSFORM_IMIR) {
+            if (image->imir.mode) {
+                return 7; // 90 degrees anti-clockwise then swap left and right.
+            }
+            return 5; // 90 degrees anti-clockwise then swap top and bottom.
+        }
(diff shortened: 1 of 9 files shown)


apps/shared/avifexif.c
```diff
@@ -0,0 +1,63 @@
+// Copyright 2022 Google LLC. All rights reserved.
```
jzern, line 1: `// Copyright 2022 Google LLC. All rights reserved.`

maybe we need a normalization pass here, I think this is now `// Copyright 2022 Google LLC`

apps/shared/avifexif.h
```diff
@@ -0,0 +1,10 @@
+// Copyright 2022 Google LLC. All rights reserved.
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "avif/avif.h"
+
+// Converts image->transformFlags, image->irot and image->imir to the equivalent Exif orientation value in [1:8].
+uint8_t avifImageGetExifOrientationFromIrotImir(const avifImage * image);
+
+// Attempts to parse the image->exif payload until the Exif orientation is found, then sets it to the given value.
```
jzern, line 9: This function doesn't take an avifImage*, so the comment should just say 'exif', no?

apps/shared/avifexif.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1201.

jzern (approved):

apps/shared/avifjpeg.c
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 
 #include "avifjpeg.h"
-#include "avif/internal.h"
+#include "avifexif.h"
```
wantehchang, line 5: Nit: List headers in alphabetical order.

Make the same change to apps/shared/avifpng.c.

wantehchang (approved):
LGTM. Thanks.

apps/shared/avifexif.h
```diff
@@ -0,0 +1,10 @@
+// Copyright 2022 Google LLC
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "avif/avif.h"
+
```
wantehchang, line 5: Add a header include guard and extern "C" for C++ code.

wantehchang (approved):
LGTM.

apps/shared/avifjpeg.c
jzern: To make that clearer we could add blank line:

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

PR #1204, opened 2022-11-10
Allow repetition count for animated AVIF

vigneshvg:
The HEIF specification allows the use of edit lists for specifying the number of times an animation should be repeated.

This CL does the following:
 * Adds a "--repetition-count" option to avifenc and the libavif API to write the correct repetition count using the edit list box as specified in Section 9.6 of ISO/IEC 23008-12 Part 12.
 * Adds a repetitionCount member to the avifDecoder struct for readers to be able to see the repetition count when decoding.
 * Adds two #defines one to specify INFINITE repetition and another one to specify UNKNOWN repetitions (in case of images made with older libavif versions prior to this CL, the repetition count will be deduced as UNKNOWN because there is no EditList box). Applications can choose to handle the UNKNOWN case however they want.

Compatibility check for the generated files:
 * Compliance Warden does not report any errors.
 * Chrome is able to display the files (ignoring the repetition count).

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -351,6 +351,10 @@ typedef struct avifTrack
     uint32_t premByID; // if non-zero, this track is premultiplied by Track #{premByID}
     uint32_t mediaTimescale;
     uint64_t mediaDuration;
+    uint64_t trackDuration;
+    uint64_t segmentDuration;
+    avifBool isRepeating;
+    int repetitionCount;
     uint32_t width;
     uint32_t height;
     avifSampleTable * sampleTable;
@@ -2456,17 +2460,19 @@ static avifBool avifParseTrackHeaderBox(avifTrack * track,
     uint32_t ignored32, trackID;
     uint64_t ignored64;
     if (version == 1) {
-        AVIF_CHECK(avifROStreamReadU64(&s, &ignored64)); // unsigned int(64) creation_time;
-        AVIF_CHECK(avifROStreamReadU64(&s, &ignored64)); // unsigned int(64) modification_time;
-        AVIF_CHECK(avifROStreamReadU32(&s, &trackID));   // unsigned int(32) track_ID;
-        AVIF_CHECK(avifROStreamReadU32(&s, &ignored32)); // const unsigned int(32) reserved = 0;
-        AVIF_CHECK(avifROStreamReadU64(&s, &ignored64)); // unsigned int(64) duration;
+        AVIF_CHECK(avifROStreamReadU64(&s, &ignored64));            // unsigned int(64) creation_time;
+        AVIF_CHECK(avifROStreamReadU64(&s, &ignored64));            // unsigned int(64) modification_time;
+        AVIF_CHECK(avifROStreamReadU32(&s, &trackID));              // unsigned int(32) track_ID;
+        AVIF_CHECK(avifROStreamReadU32(&s, &ignored32));            // const unsigned int(32) reserved = 0;
+        AVIF_CHECK(avifROStreamReadU64(&s, &track->trackDuration)); // unsigned int(64) duration;
     } else if (version == 0) {
-        AVIF_CHECK(avifROStreamReadU32(&s, &ignored32)); // unsigned int(32) creation_time;
-        AVIF_CHECK(avifROStreamReadU32(&s, &ignored32)); // unsigned int(32) modification_time;
-        AVIF_CHECK(avifROStreamReadU32(&s, &trackID));   // unsigned int(32) track_ID;
-        AVIF_CHECK(avifROStreamReadU32(&s, &ignored32)); // const unsigned int(32) reserved = 0;
-        AVIF_CHECK(avifROStreamReadU32(&s, &ignored32)); // unsigned int(32) duration;
+        uint32_t trackDuration;
+        AVIF_CHECK(avifROStreamReadU32(&s, &ignored32));     // unsigned int(32) creation_time;
+        AVIF_CHECK(avifROStreamReadU32(&s, &ignored32));     // unsigned int(32) modification_time;
+        AVIF_CHECK(avifROStreamReadU32(&s, &trackID));       // unsigned int(32) track_ID;
+        AVIF_CHECK(avifROStreamReadU32(&s, &ignored32));     // const unsigned int(32) reserved = 0;
+        AVIF_CHECK(avifROStreamReadU32(&s, &trackDuration)); // unsigned int(32) duration;
+        track->trackDuration = (trackDuration == AVIF_UNKNOWN_DURATION32) ? AVIF_UNKNOWN_DURATION64 : trackDuration;
     } else {
         // Unsupported version
         avifDiagnosticsPrintf(diag, "Box[tkhd] has an unsupported version [%u]", version);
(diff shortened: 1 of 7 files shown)


vigneshvg:
I looked at the files from this HEIF conformance repository: https://github.com/nokiatech/heif_conformance

There are 3 files in that repository that supports HEIC sequences with repetitions using EditList (C036.heic, C037.heic and C038.heic).

I have verified that the AVIF files produced by this CL has comparable EditList boxes to those three files.

vigneshvg:
@joedrago @leo-barnes @cconcolato please take a look at this PR and would appreciate a review from a spec and libavif API perspective.

design decision to use `repetitionCount` instead of `loopCount`: if we are simply exposing loopCount, then there is the ambiguity of what to do when the value is equal to 0 (do we loop it 0 times? which could mean not playing back anything). Whereas having the repetitionCount clarifies that case (repetitionCount == 0 means, play back exactly once and do not repeat it).

vigneshvg:
Also found a few past references to this issue:
https://github.com/AOMediaCodec/libavif/issues/766#issuecomment-925289365
https://github.com/AOMediaCodec/av1-avif/issues/73

joedrago:
I have not dug into the PR's code yet, but I'm curious what the expectations are if an `elst` box is absent, once we implement something like this. Is the expectation for the sequence to infinitely loop (maintaining the current implementation in Chrome, say), or is it that pre-existing sequences would only play once and stop, and would require an injection of an infinite `elst` box to regain the loop?

vigneshvg:
> I have not dug into the PR's code yet, but I'm curious what the expectations are if an `elst` box is absent, once we implement something like this. Is the expectation for the sequence to infinitely loop (maintaining the current implementation in Chrome, say), or is it that pre-existing sequences would only play once and stop, and would require an injection of an infinite `elst` box to regain the loop?

Yes this is handled by using a special constant `AVIF_REPTITION_COUNT_UNKNOWN`. For files created with older versions of libavif (and other tools which don't write the `elst` box), the libavif decoder will set the `repetitionCount` to this special value. It is up to the application to do whatever they want with this value. For example, in Chrome's case, it could loop infinitely when `repetitionCount` is UNKNOWN to keep backward compatibility.

src/read.c
```diff
@@ -2784,12 +2790,70 @@ static avifBool avifTrackReferenceBox(avifTrack * track, const uint8_t * raw, si
...
+
+    if (version == 1) {
+        avifROStreamReadU64(&s, &track->segmentDuration); // unsigned int(64) segment_duration;
+    } else {
+        uint32_t segmentDuration;
+        avifROStreamReadU32(&s, &segmentDuration); // unsigned int(32) segment_duration;
+        track->segmentDuration = segmentDuration;
+    }
+    if (track->segmentDuration == 0) {
+        avifDiagnosticsPrintf(diag, "Box [elst] Invalid value for segment_duration (0).");
```
wantehchang, line 2822: Nit: Delete the space in "Box [elst]".

src/read.c
```diff
@@ -2802,10 +2866,46 @@ static avifBool avifParseTrackBox(avifDecoderData * data, uint64_t rawOffset, co
...
+            track->repetitionCount = AVIF_REPETITION_COUNT_INFINITE;
+        } else {
+            // Section 9.6.1. of ISO/IEC 23008-12 Part 12: 1, the entire edit list is repeated a sufficient number of times to
+            // equal the track duration.
+            //
+            // Since libavif uses repetitionCount (which is 0-based), we subtract the value by 1 to derive the number of
+            // repetitions.
+            assert(track->segmentDuration != 0);
+            const uint64_t repetitionCount =
+                (track->trackDuration / track->segmentDuration) + (track->trackDuration % track->segmentDuration != 0) - 1;
```
wantehchang, line 2895: Nit: The `(a + b - 1) / b` expression is commonly used to calculate the ceiling of the `a / b` quotient (with normal, real-number division). Do you think `(a / b) + (a % b != 0)` is easier to understand?

vigneshvg: I believe both forms are okay. I generally prefer the `%` to circumvent the problem of `(a + b - 1)` intermediate addition potentially overflowing.

src/read.c

vigneshvg:
@joedrago can you please take a look at this PR when you have some time? :) thank you!

cconcolato:
> I looked at the files from this HEIF conformance repository: https://github.com/nokiatech/heif_conformance
> 
> There are 3 files in that repository that supports HEIC sequences with repetitions using EditList (C036.heic, C037.heic and C038.heic).
> 
> I have verified that the AVIF files produced by this CL has comparable EditList boxes to those three files.

Thank you for doing that. 
Can you clarify if you are doing the following or something different in the writer?
- if no repetition count is provided, the track is set to a non-infinite duration (equal to the sum of all sample durations) and no edit list is created
- if a repetition count value is provided but it is not infinite, the track is set to a non-infinite duration (the total duration of the repetition), an edit list is coded with: its flags & 1 = 1, with one entry, with its media time is 0, its sample rate set to 1, and its edit duration set to the sum of the durations of all samples in the movie timescale
- if an infinite repetition count is passed, the track duration is set set to infinite, an edit list is coded with: its flags & 1 = 1, with one entry, with its media time is 0, its sample rate set to 1, and its edit duration set to the sum of the durations of all samples in the movie timescale

vigneshvg:
> > I looked at the files from this HEIF conformance repository: https://github.com/nokiatech/heif_conformance
> > There are 3 files in that repository that supports HEIC sequences with repetitions using EditList (C036.heic, C037.heic and C038.heic).
> > I have verified that the AVIF files produced by this CL has comparable EditList boxes to those three files.
> 
> Thank you for doing that. Can you clarify if you are doing the following or something different in the writer?
> 

Thanks for the reply Cyril.

> * if no repetition count is provided, the track is set to a non-infinite duration (equal to the sum of all sample durations) and no edit list is created

No, in this case the writer assumes the "default" repetition count (which is infinite). So it does the same thing as the infinite case (third bullet point below). If you are probably meaning the "zero" repetition case here, then that is covered in my response below.

> * if a repetition count value is provided but it is not infinite, the track is set to a non-infinite duration (the total duration of the repetition), an edit list is coded with: its flags & 1 = 1, with one entry, with its media time is 0, its sample rate set to 1, and its edit duration set to the sum of the durations of all samples in the movie timescale

Yes this is correct.

> * if an infinite repetition count is passed, the track duration is set set to infinite, an edit list is coded with: its flags & 1 = 1, with one entry, with its media time is 0, its sample rate set to 1, and its edit duration set to the sum of the durations of all samples in the movie timescale

Yes this is correct. This is the default behavior as well (i.e.) when repetition count is not set explicitly.

Note that one of the cases that you have not mentioned here is the repetition count value of 0:
In this case, the track duration is set to the total duration of the samples, an edit list is coded with its (flags & 1) = 0, with one entry, with its media time is 0, its sample rate set to 1, and its edit duration set to the same value as the track duration.

Please let me know if any of this does not make sense from an API design/spec perspective. Thank you :)

cconcolato:
> Please let me know if any of this does not make sense from an API design/spec perspective. Thank you :)
That makes sense. I thought the default would be 0 repetition, you want it to be infinite. I'm fine with that.
Note that in the 3rd case (infinite repetition, aka default) you could also not code that edit list at all. That should result in the same. I'm just wondering if an old reader will not complain if they see an edit list.

vigneshvg:
> Note that in the 3rd case (infinite repetition, aka default) you could also not code that edit list at all. That should result in the same. I'm just wondering if an old reader will not complain if they see an edit list.

I think interpreting files without edit list should be up to the application (and not default to infinite) since there is an explicit way to signal infinite repetitions. In the new code, the libavif _reader_ has two constants to cover these two cases: one for INFINITE (for files which have an explicit edit list that points to infinite looping) and another one for UNKNOWN (for older files which have no edit list coded).

Old readers generally tend to skip unknown boxes, so coding an edit list in all cases should be okay.

joedrago:
Generally speaking, if Cyril is satisfied with the semantics of the edit list box you've chosen and Wan-Teh is satisfied with the code changes, this looks pretty reasonable to me. The two remaining things that stand out to me are:

* Why is this function pluralized: `avifParseEditsBox()`? The pattern here on these `avifParse*()` functions is to use the exact name of the box in the standard, and it looks like `[edts]` maps to `EditBox`. (I was expecting `avifParseEditBox()`.)

* I see that you have added a new call to `avifDiagnosticsPrintf()`, but this does not occur alongside a failure case as a description of the reason the calling function did not succeed, but instead as a Warning. This really isn't in the spirit of what `avifDiagnostics()` is currently designed for, and if we wanted to start emitting string warnings, we would really need to come up with fresh avifDiagnostics plumbing for that. My recommendation is (for now) to replace that warning call to `avifDiagnosticsPrintf()` with a comment documenting the chosen behavior, and perhaps open a libavif issue discussing if and how to surface warnings through the `avifDiagnostics` code.

vigneshvg:
> Generally speaking, if Cyril is satisfied with the semantics of the edit list box you've chosen and Wan-Teh is satisfied with the code changes, this looks pretty reasonable to me.

Thanks for the review. Based on Cyril's last comment, it seems like the semantics are okay.

@wantehchang do you have any comments on the code changes?

> * Why is this function pluralized: `avifParseEditsBox()`? The pattern here on these `avifParse*()` functions is to use the exact name of the box in the standard, and it looks like `[edts]` maps to `EditBox`. (I was expecting `avifParseEditBox()`.)

Oops, this is my bad. I just happened to read the `edts` box as "edits" box in my mind. When in fact it should be "edit" box. I have renamed the function.

> * I see that you have added a new call to `avifDiagnosticsPrintf()`, but this does not occur alongside a failure case as a description of the reason the calling function did not succeed, but instead as a Warning. This really isn't in the spirit of what `avifDiagnostics()` is currently designed for, and if we wanted to start emitting string warnings, we would really need to come up with fresh avifDiagnostics plumbing for that. My recommendation is (for now) to replace that warning call to `avifDiagnosticsPrintf()` with a comment documenting the chosen behavior, and perhaps open a libavif issue discussing if and how to surface warnings through the `avifDiagnostics` code.

Good point. I have updated the printf with a comment. I will file a separate issue to plumb warnings.

wantehchang:
Vignesh: I have reviewed this pull request carefully at https://aomedia-review.googlesource.com/c/libavif/+/165703 a few times. If you have made only minor changes to this pull request since my last review, I will just review it again when it is imported into Google's internal repository.

Re: Joe's comment about calling `avifDiagnosticsPrintf()` when the function does not fail: I remember I also noted this issue. libavif users will only retrieve the diagnostics message when a function fails.

src/read.c
wantehchang: If `track->trackDuration` is 0, the new expression will evaluate to 0 + 0 - 1. We should defend against this case or return an error.

vigneshvg: Ah yes.

I have added a check for trackDuration == 0.

Note that i am specifically doing it here and not where trackDuration is read. Since we only use the value in this case, we don't have to punish files which had accidentally set it to 0 without the editlist with older tools (those files would have parsed successfully so far because we never read the trackDuration value until this PR).

vigneshvg:
> Vignesh: I have reviewed this pull request carefully at https://aomedia-review.googlesource.com/c/libavif/+/165703 a few times. If you have made only minor changes to this pull request since my last review, I will just review it again when it is imported into Google's internal repository.
> 

Alright, thanks!

> Re: Joe's comment about calling `avifDiagnosticsPrintf()` when the function does not fail: I remember I also noted this issue. libavif users will only retrieve the diagnostics message when a function fails.

Yep makes sense. I have removed the call and replaced it with a code comment for now.

wantehchang (approved):
LGTM.

vigneshvg:
Thanks for the review everyone. I will go ahead and submit this PR.

PR #1206, opened 2022-11-10
Use INPUT_PNG in the lossless test

wantehchang:
After commit bc276e7edb2d7aa16a4baec2ec434966fca2a9ef we can use INPUT_PNG (paris_icc_exif_xmp.png, which has Exif and XMP metadata) in the lossless test, because DECODED_FILE_LOSSLESS will also have the Exif and XMP metadata from INPUT_PNG.


y-guyon (approved):

y-guyon:
Thanks

PR #1208, opened 2022-11-11
Require min and max quanitizers be both specified

wantehchang:
Require the min and max quantizers be both specified or both unspecified. Since the default values of the min and max quantizers are not well known (24 and 26 respectively) and may be changed, avifenc users should not set only one of them. So this new requirement is unlikely to break many existing scripts.

### DIFF
--- a/tests/test_cmd.sh
+++ b/tests/test_cmd.sh
@@ -74,6 +74,8 @@ pushd ${TMP_DIR}
   # Passing a filename starting with a dash without using -- should fail.
   "${AVIFENC}" -s 10 "${INPUT_Y4M}" "${ENCODED_FILE_WITH_DASH}" && exit 1
   "${AVIFDEC}" --info "${ENCODED_FILE_WITH_DASH}" && exit 1
+  "${AVIFENC}" -s 10 --min 24 "${INPUT_Y4M}" "${ENCODED_FILE}" && exit 1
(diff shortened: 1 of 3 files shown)


vrabaud (approved):

tests/test_cmd.sh
```diff
@@ -74,6 +74,8 @@ pushd ${TMP_DIR}
   # Passing a filename starting with a dash without using -- should fail.
   "${AVIFENC}" -s 10 "${INPUT_Y4M}" "${ENCODED_FILE_WITH_DASH}" && exit 1
   "${AVIFDEC}" --info "${ENCODED_FILE_WITH_DASH}" && exit 1
+  "${AVIFENC}" -s 10 --min 24 "${INPUT_Y4M}" "${ENCODED_FILE}" && exit 1
```
y-guyon, line 77: This should be in another comment category. Consider adding a new line and `# --min and --max must be both specified.` and `"${AVIFENC}" -s 10 --minalpha 26 "${INPUT_PNG}" "${ENCODED_FILE}" && exit 1`

y-guyon (approved):

tests/test_cmd.sh
wantehchang: Done. Note that I did not add an echo statement like the ones in lines 65 and 71. Also I used the default values of `--minalpha` and `--maxalpha` (0) in the tests (just like `--min 24` and `--max 26`), to show that even the default values don't receive special treatment.

wantehchang:
Yannis: Please review the second commit in this pull request. Thanks.

PR #1210, opened 2022-11-14
Re-order the boxes within "stbl"

vigneshvg:
The ISO-BMFF spec has a "recommendation" for how the boxes within the "stbl" box should be ordered.

Text from Section 6.2.3 of the spec:
It is *recommended* that the boxes within the Sample Table Box be in the following order: Sample Description, Time to Sample, Sample to Chunk, Sample Size, Chunk Offset.

Change the writer code to update the order of these boxes to match what is recommended in the spec.

Old order: stco, stsc, stsz, stss, stts, stsd.
New order: stsd, stts, stsc, stsz, stco, stss.

This also matches the box ordering used by other tools like ffmpeg.

Motivation behind this change:
The MP4 reader in the Android platform fails if this ordering is not followed. While that is not the right behavior (since the spec does not mandate the order, it is simply a "recommendation"), there is no harm in fixing libavif to produce AVIF animations that are consumable on Android.

Verification:
* Compliance warden has no issues with the new ordering.
* Files are playing back on Chrome like before.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -1674,63 +1674,12 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
             avifRWStreamFinishBox(&s, dref);
             avifRWStreamFinishBox(&s, dinf);
 
+            // The boxes within the "stbl" box are ordered using the following recommendation in ISO/IEC 14496-12, Section 6.2.3:
+            // 4) It is recommended that the boxes within the Sample Table Box be in the following order: Sample Description
+            // (stsd), Time to Sample (stts), Sample to Chunk (stsc), Sample Size (stsz), Chunk Offset (stco).
+            //
+            // Any boxes not listed in the above line are placed in the end (after the "stco" box).
             avifBoxMarker stbl = avifRWStreamWriteBox(&s, "stbl", AVIF_BOX_SIZE_TBD);
-
-            avifBoxMarker stco = avifRWStreamWriteFullBox(&s, "stco", AVIF_BOX_SIZE_TBD, 0, 0);
-            avifRWStreamWriteU32(&s, 1);           // unsigned int(32) entry_count;
-            avifEncoderItemAddMdatFixup(item, &s); //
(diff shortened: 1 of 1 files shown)


src/write.c
```diff
@@ -1674,64 +1674,13 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
             avifRWStreamFinishBox(&s, dref);
             avifRWStreamFinishBox(&s, dinf);
 
+            // The boxes within the "stbl" box are ordered using the following recommendation in ISO/IEC 14496-12, Section 6.2.3:
+            // 4) It is recommended that the boxes within the Sample Table Box be in the following order: Sample Description
+            // (stsd), Time to Sample (stts), Sample to Chunk (stsc), Sample Size (stsz), Chunk Offset (stco).
+            //
+            // Any boxes not listed in the above line are placed in the end (after the "stco" box).
```
wantehchang, line 1681: Just wanted to confirm that stss is the only one, right?

Does Android expect stss to be after the other boxes listed above?

wantehchang (approved):
LGTM.

src/write.c
vigneshvg: > Just wanted to confirm that stss is the only one, right?

Yes, as of now `stss` is the only box not in the list. But the spec potentially allows for other boxes that aren't in the list. So i added the note just to make sure if we add any more boxes here, it has to be added after the `stco` box.

> 
> Does Android expect stss to be after the other boxes listed above?

The Android MP4 parser code is written in such a way that if the `stsz` box appears before the `stsd` box, then the parser fails to identify the image sequences properly. It is a bug and i will try to raise an issue with the Android team and/or try to fix it.

vigneshvg:
Sorry, i landed this PR too soon. Seems like there are some failures on the workflow but they seem unrelated. I ran the workflow locally and they all pass with this PR.

wantehchang:
Vignesh: I believe the CI workflow build failures were my fault. Sorry. I just wrote a pull request to fix them: https://github.com/AOMediaCodec/libavif/pull/1211.

PR #1205, opened 2022-11-10
Build aviftest_helpers for are_images_equal.cc

wantehchang:
If are_images_equal.cc is built, enable CXX and build the aviftest_helpers library.

Fix the build failure when -DAVIF_BUILD_APPS=ON -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=OFF is passed to the cmake command.

PR #1211, opened 2022-11-14
Rename the cache files for ci-disable-gtest.yml

wantehchang:
Add "-disable-gtest" to the cache file names for ci-disable-gtest.yml.

In https://github.com/AOMediaCodec/libavif/pull/1205 I neglected to use different cache file names in the new ci-disable-gtest.yml file, or mistakenly thought that it could use the cache files for ci.yml. Since ci-disable-gtest.yml doesn't build libgav1 and GoogleTest, its cache files are missing those two dependencies.

Make a simple change to ext/googletest.cmd to force the cache files to be recreated.

### DIFF
--- a/ext/googletest.cmd
+++ b/ext/googletest.cmd
@@ -15,5 +15,5 @@ cd build
 : # on Windows:
 : # https://github.com/google/googletest/blob/main/googletest/README.md#visual-studio-dynamic-vs-static-runtimes
 cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DBUILD_GMOCK=OFF -Dgtest_force_shared_crt=ON ..
-ninja
+cmake --build .
 cd ../..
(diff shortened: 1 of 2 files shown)


vigneshvg (approved):
thanks @wantehchang.

the change looks good to me. please wait to make sure the workflows pass before submitting this.

ext/googletest.cmd
```diff
@@ -15,5 +15,5 @@ cd build
 : # on Windows:
 : # https://github.com/google/googletest/blob/main/googletest/README.md#visual-studio-dynamic-vs-static-runtimes
 cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DBUILD_GMOCK=OFF -Dgtest_force_shared_crt=ON ..
-ninja
+cmake --build .
```
wantehchang, line 18: Note that this is a simple change to force all the cache files to be recreated. The keys for the cache files depend on the contents of the ext/*.cmd files. So if we modify any of the ext/*.cmd, the keys will change, and new cache files will be created.

jzern (approved):
Let's see if we can delete the cache to avoid the extra ninja -> cmake --build change:

https://docs.github.com/en/rest/actions/cache#delete-github-actions-caches-for-a-repository-using-a-cache-key

PR #1216, opened 2022-11-17
Require src UV in avifImageCopy() if Y is not NULL

y-guyon:


vrabaud (approved):

src/avif.c
```diff
@@ -185,14 +185,9 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
...
-            if (!srcImage->yuvRowBytes[uvPlane]) {
-                // plane is absent. If we're copying from a source without
-                // them, mimic the source image's state by removing our copy.
-                avifFree(dstImage->yuvPlanes[uvPlane]);
-                dstImage->yuvPlanes[uvPlane] = NULL;
-                dstImage->yuvRowBytes[uvPlane] = 0;
-            }
+        if ((dstImage->yuvRowBytes[AVIF_CHAN_U] && !srcImage->yuvRowBytes[AVIF_CHAN_U]) ||
+            (dstImage->yuvRowBytes[AVIF_CHAN_V] && !srcImage->yuvRowBytes[AVIF_CHAN_V])) {
+            return AVIF_RESULT_INVALID_ARGUMENT;
```
wantehchang, line 190: With this error check, we should be able to remove the `if (!dstRow)` statement at lines 201-203, right?

I am just checking my understanding. It's fine to keep that if statement

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ to the avifImage struct.
   into cells.
 * Apps must be built with libpng version 1.6.32 or above.
 * Update svt.cmd/svt.sh: v1.3.0
+* avifImageCopy() does no longer accept source U and V channels to be NULL for
```
wantehchang, line 24: Nit: does no longer accept => no longer accepts

src/avif.c
```diff
@@ -185,14 +185,9 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
...
-        for (int uvPlane = AVIF_CHAN_U; uvPlane <= AVIF_CHAN_V; ++uvPlane) {
-            if (!srcImage->yuvRowBytes[uvPlane]) {
-                // plane is absent. If we're copying from a source without
-                // them, mimic the source image's state by removing our copy.
-                avifFree(dstImage->yuvPlanes[uvPlane]);
-                dstImage->yuvPlanes[uvPlane] = NULL;
-                dstImage->yuvRowBytes[uvPlane] = 0;
-            }
+        if ((dstImage->yuvRowBytes[AVIF_CHAN_U] && !srcImage->yuvRowBytes[AVIF_CHAN_U]) ||
+            (dstImage->yuvRowBytes[AVIF_CHAN_V] && !srcImage->yuvRowBytes[AVIF_CHAN_V])) {
```
wantehchang, line 189: An alternative check is to check if `srcImage` is non-4:0:0, as the changelog entry says. An advantage is that we can check that before the allocation.

```
diff --git a/src/avif.c b/src/avif.c
index a73c187..222410f 100644
--- a/src/avif.c
+++ b/src/avif.c
@@ -181,19 +181,14 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
     avifImageSetMetadataXMP(dstImage, srcImage->xmp.data, srcImage->xmp.size);
 
     if ((planes & AVIF_PLANES_YUV) && srcImage->yuvPlanes[AVIF_CHAN_Y]) {
+        if ((srcImage->yuvFormat != AVIF_PIXEL_FORMAT_YUV400) &&
+            (!srcImage->yuvRowBytes[AVIF_CHAN_U] || !srcImage->yuvRowBytes[AVIF_CHAN_V])) {
+        }
         const avifResult allocationResult = avifImageAllocatePlanes(dstImage, AVIF_PLANES_YUV);
         if (allocationResult != AVIF_RESULT_OK) {
             return allocationResult;
         }
-        for (int uvPlane = AVIF_CHAN_U; uvPlane <= AVIF_CHAN_V; ++uvPlane) {
-            if (!srcImage->yuvRowBytes[uvPlane]) {
-                // plane is absent. If we're copying from a source without
-                // them, mimic the source image's state by removing our copy.
-                avifFree(dstImage->yuvPlanes[uvPlane]);
-                dstImage->yuvPlanes[uvPlane] = NULL;
-                dstImage->yuvRowBytes[uvPlane] = 0;
-            }
-        }
     }
     if ((planes & AVIF_PLANES_A) && srcImage->alphaPlane) {
         const avifResult allocationResult = avifImageAllocatePlanes(dstImage, AVIF_PLANES_A);
```

Note: We can also check for a similar error for 4:0:0:

```
        if ((srcImage->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) &&
            (srcImage->yuvRowBytes[AVIF_CHAN_U] || srcImage->yuvRowBytes[AVIF_CHAN_V])) {
        }
```

wantehchang: On second thought, I think we should not do the second check I suggested in the previous comment (in the Note). The reason is that for a 4:0:0 image we should avoid accessing its `yuvPlanes[AVIF_CHAN_U]`, `yuvRowBytes[AVIF_CHAN_U]`, etc.

src/avif.c
y-guyon: No, see https://github.com/AOMediaCodec/libavif/pull/1183#discussion_r1026142558

src/avif.c
```diff
@@ -181,14 +181,14 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
     avifImageSetMetadataXMP(dstImage, srcImage->xmp.data, srcImage->xmp.size);
 
     if ((planes & AVIF_PLANES_YUV) && srcImage->yuvPlanes[AVIF_CHAN_Y]) {
+        if ((srcImage->yuvFormat != AVIF_PIXEL_FORMAT_YUV400) &&
+            (!srcImage->yuvRowBytes[AVIF_CHAN_U] || !srcImage->yuvRowBytes[AVIF_CHAN_V])) {
```
wantehchang, line 185: I am wondering if it would be better to test `(!srcImage->yuvPlanes[AVIF_CHAN_U] || !srcImage->yuvPlanes[AVIF_CHAN_V])` instead. We test `srcImage->yuvPlanes[AVIF_CHAN_Y]` at line 183 and `srcImage->alphaPlane` at line 193.

wantehchang (approved):
LGTM.

PR #1217, opened 2022-11-17
Use AVIF_LIBRARY_SUFFIX for libdav1d in CMake

y-guyon:
Otherwise a locally built ext/dav1d will be missing libdav1d.so in a shared build of libavif on Unix platforms.
Add a fallback to .a for Windows platforms.

Build dav1d in ci-unix-shared-local.yml.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -280,12 +280,16 @@ if(AVIF_CODEC_DAV1D)
     if(AVIF_LOCAL_DAV1D)
         set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/src/libdav1d${AVIF_LIBRARY_SUFFIX}")
         if(NOT EXISTS "${LIB_FILENAME}")
-            # On windows, meson will produce a libdav1d.a instead of the expected libdav1d.dll/.lib.
-            # See https://github.com/mesonbuild/meson/issues/8153.
-            set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/src/libdav1d.a")
-        endif()
-        if(NOT EXISTS "${LIB_FILENAME}")
-            message(FATAL_ERROR "libavif: ${LIB_FILENAME}/.a is missing, bailing out")
+            if("${AVIF_LIBRARY_SUFFIX}" STREQUAL ".a")
+                message(FATAL_ERROR "libavif: ${LIB_FILENAME} is missing, bailing out")
+            else()
+                # On windows, meson will produce a libdav1d.a instead of the expected libdav1d.dll/.lib.
+                # See https://github.com/mesonbuild/meson/issues/8153.
+                set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/src/libdav1d.a")
+                if(NOT EXISTS "${LIB_FILENAME}")
+                    message(FATAL_ERROR "libavif: ${LIB_FILENAME} (or libdav1d${AVIF_LIBRARY_SUFFIX}) is missing, bailing out")
+                endif()
+            endif()
         endif()
 
         set(AVIF_CODEC_INCLUDES
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -278,9 +278,14 @@ if(AVIF_CODEC_DAV1D)
     set(AVIF_SRCS ${AVIF_SRCS} src/codec_dav1d.c)
 
     if(AVIF_LOCAL_DAV1D)
-        set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/src/libdav1d.a")
+        set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/src/libdav1d${AVIF_LIBRARY_SUFFIX}")
         if(NOT EXISTS "${LIB_FILENAME}")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -278,9 +278,14 @@ if(AVIF_CODEC_DAV1D)
...
-        set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/src/libdav1d.a")
+        set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/src/libdav1d${AVIF_LIBRARY_SUFFIX}")
         if(NOT EXISTS "${LIB_FILENAME}")
-            message(FATAL_ERROR "libavif: ${LIB_FILENAME} is missing, bailing out")
+            # On windows, meson will produce a libdav1d.a instead of the expected libdav1d.dll/.lib.
+            # See https://github.com/mesonbuild/meson/issues/8153.
+            set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/src/libdav1d.a")
+        endif()
+        if(NOT EXISTS "${LIB_FILENAME}")
+            message(FATAL_ERROR "libavif: ${LIB_FILENAME}/.a is missing, bailing out")
```
jzern, line 288: "/.a" might not make sense since you're checking the existence of LIB_FILENAME

jzern (approved):

CMakeLists.txt
```diff
@@ -278,9 +278,14 @@ if(AVIF_CODEC_DAV1D)
     set(AVIF_SRCS ${AVIF_SRCS} src/codec_dav1d.c)
 
     if(AVIF_LOCAL_DAV1D)
-        set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/src/libdav1d.a")
+        set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/dav1d/build/src/libdav1d${AVIF_LIBRARY_SUFFIX}")
         if(NOT EXISTS "${LIB_FILENAME}")
-            message(FATAL_ERROR "libavif: ${LIB_FILENAME} is missing, bailing out")
+            # On windows, meson will produce a libdav1d.a instead of the expected libdav1d.dll/.lib.
```
kmilos, line 283: For reference, it's not only about meson - any toolchain on Windows that is not MSVC will do this by default (i.e. CMake w/ GCC/Clang, Autotools...). Just use the Cmake variables that are correctly set up for the platform, e.g.: https://gitlab.kitware.com/cmake/cmake/-/blob/ee7479fb5a457fae95e24d3b0d206d6138e5a448/Modules/Platform/Windows-GNU.cmake#L18-27

y-guyon: Thanks for the information.

> For reference, it's not only about meson - any toolchain on Windows that is not MSVC will do this by default (i.e. CMake w/ GCC/Clang, Autotools...).

Do you have any link to some documentation about that, so that I can replace `https://github.com/mesonbuild/meson/issues/8153`?

>  Just use the CMake variables that are correctly set up for the platform, e.g.: https://gitlab.kitware.com/cmake/cmake/-/blob/ee7479fb5a457fae95e24d3b0d206d6138e5a448/Modules/Platform/Windows-GNU.cmake#L18-27

Not sure I follow, are you suggesting to use `CMAKE_FIND_LIBRARY_PREFIXES` and `CMAKE_FIND_LIBRARY_SUFFIXES`? Because looking for `libdav1d${AVIF_LIBRARY_SUFFIX}` did not work.

kmilos: > Not sure I follow, are you suggesting to use `CMAKE_FIND_LIBRARY_PREFIXES` and `CMAKE_FIND_LIBRARY_SUFFIXES`? Because looking for `libdav1d${AVIF_LIBRARY_SUFFIX}` did not work.

Hm, given [it's set up correctly already](https://github.com/AOMediaCodec/libavif/blob/cd03eef10a7cc6081ee98e3911e21d977a2bf17b/CMakeLists.txt#L73-L79), I'm not sure what your problem actually is now... Maybe the `lib` prefix should not be assumed and one should also use `${AVIF_LIBRARY_PREFIX}dav1d${AVIF_LIBRARY_SUFFIX}` like for the other libraries?

y-guyon: > Maybe the `lib` prefix should not be assumed and one should also use `${AVIF_LIBRARY_PREFIX}dav1d${AVIF_LIBRARY_SUFFIX}` like for the other libraries?

From what I understood, `meson --default-library=static` will generate `libdav1d.a` instead of `dav1d.lib`.
A similar configuration issue with libgav1 is likely, because it does not use the `${AVIF_LIBRARY_PREFIX}` either.

> Is local dav1d not built using CMake?

No: https://github.com/AOMediaCodec/libavif/blob/0311709e9b7b8c46aace942e6b21207f48967e4c/ext/dav1d.cmd#L20

kmilos: Ah, ok, so it is meson + MSVC specific after all, sorry!

In that case, I guess you could iterate through the matrix of `CMAKE_FIND_LIBRARY_PREFIXES` and `CMAKE_FIND_LIBRARY_SUFFIXES`, or just handle this one known corner case like this.

y-guyon: Sure. I guess `CMAKE_FIND_LIBRARY_PREFIXES` and `CMAKE_FIND_LIBRARY_SUFFIXES` can be combined with [`find_library()`](https://cmake.org/cmake/help/latest/command/find_library.html#command:find_library) in another PR.

PR #1223, opened 2022-11-21
Clamp samples in y4mRead()

y-guyon:
Passing out-of-range values to libavif API may trigger undefined behavior. Rather than fixing the whole pipeline and maybe slowing it down, make sure the input respects the specified depth and range.

b/228859724#comment3

### DIFF
--- a/apps/shared/y4m.c
+++ b/apps/shared/y4m.c
@@ -217,8 +217,7 @@ static avifBool y4mClampSamples(avifImage * avif)
     const uint16_t maxSampleValue = (uint16_t)((1u << avif->depth) - 1u);
 
     avifBool samplesWereClamped = AVIF_FALSE;
-    // Alpha is never in limited range.
-    for (int plane = AVIF_CHAN_Y; plane <= AVIF_CHAN_V; ++plane) {
+    for (int plane = AVIF_CHAN_Y; plane <= AVIF_CHAN_A; ++plane) {
         uint32_t planeHeight = avifImagePlaneHeight(avif, plane); // 0 for UV if 4:0:0.
         uint32_t planeWidth = avifImagePlaneWidth(avif, plane);
         uint8_t * row = avifImagePlane(avif, plane);
@@ -419,7 +418,7 @@ avifBool y4mRead(const char * inputFilename, avifImage * avif, avifAppSourceTimi
     // libavif API does not guarantee the absence of undefined behavior if samples exceed the specified avif->depth.
     // Avoid that by making sure input values are within the correct range.
     if (y4mClampSamples(avif)) {
-        fprintf(stderr, "WARNING: some samples were clamped to fit into %u bits per channel per sample\n", avif->depth);
+        fprintf(stderr, "WARNING: some samples were clamped to fit into %u bits per sample\n", avif->depth);
     }
 
     result = AVIF_TRUE;
--- a/apps/shared/y4m.c
+++ b/apps/shared/y4m.c
@@ -202,71 +202,36 @@ static int y4mReadLine(FILE * inputFile, avifRWData * raw, const char * displayF
     return -1;
 }
 
-// Limits each sample value to fit into avif->depth bits and avif->yuvRange.
+// Limits each sample value to fit into avif->depth bits.
 // Returns AVIF_TRUE if any sample was clamped this way.
 static avifBool y4mClampSamples(avifImage * avif)
 {
-    avifBool samplesWereClamped = AVIF_FALSE;
-    if (avifImageUsesU16(avif)) {
-        const uint16_t maxSampleValue = (uint16_t)((1u << avif->depth) - 1u);
-        uint16_t minSampleValues[4] = { 0, 0, 0, 0 };
-        uint16_t maxSampleValues[4] = { maxSampleValue, maxSampleValue, maxSampleValue, maxSampleValue };
-        if (avif->yuvRange == AVIF_RANGE_LIMITED) {
-            minSampleValues[AVIF_CHAN_Y] = 16 << (avif->depth - 8);
-            minSampleValues[AVIF_CHAN_U] = 16 << (avif->depth - 8);
-            minSampleValues[AVIF_CHAN_V] = 16 << (avif->depth - 8);
-            maxSampleValues[AVIF_CHAN_Y] = 235 << (avif->depth - 8);
-            maxSampleValues[AVIF_CHAN_U] = 240 << (avif->depth - 8);
-            maxSampleValues[AVIF_CHAN_V] = 240 << (avif->depth - 8);
-            // Alpha is never in limited range.
-        }
-
-        for (int plane = AVIF_CHAN_Y; plane <= AVIF_CHAN_A; ++plane) {
-            uint32_t planeHeight = avifImagePlaneHeight(avif, plane); // 0 for A if no alpha and 0 for UV if 4:0:0.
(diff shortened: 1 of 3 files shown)


apps/shared/y4m.c
```diff
@@ -380,6 +451,12 @@ avifBool y4mRead(const char * inputFilename, avifImage * avif, avifAppSourceTimi
         }
     }
 
+    // libavif API does not guarantee the absence of undefined behavior if samples exceed the specified
+    // avif->depth and avif->range. Avoid that by making sure input values are within the correct range.
```
wantehchang, line 455: Typo: avif->range => avif->yuvRange

apps/shared/y4m.c
```diff
@@ -201,6 +202,76 @@ static int y4mReadLine(FILE * inputFile, avifRWData * raw, const char * displayF
...
+        const uint16_t maxSampleValue = (uint16_t)((1u << avif->depth) - 1u);
+        uint16_t minSampleValues[4] = { 0, 0, 0, 0 };
+        uint16_t maxSampleValues[4] = { maxSampleValue, maxSampleValue, maxSampleValue, maxSampleValue };
+        if (avif->yuvRange == AVIF_RANGE_LIMITED) {
+            minSampleValues[AVIF_CHAN_Y] = 16 << (avif->depth - 8);
+            minSampleValues[AVIF_CHAN_U] = 16 << (avif->depth - 8);
+            minSampleValues[AVIF_CHAN_V] = 16 << (avif->depth - 8);
+            maxSampleValues[AVIF_CHAN_Y] = 235 << (avif->depth - 8);
+            maxSampleValues[AVIF_CHAN_U] = 240 << (avif->depth - 8);
+            maxSampleValues[AVIF_CHAN_V] = 240 << (avif->depth - 8);
```
wantehchang, line 220: We need to use different values for MatrixCoefficients 0 and 8. See H.273, Equations 20-22, 41-43, and 44-46. I think Y, U, V values have the same range for MatrixCoefficients 0 and 8. It would be good to work out the details for MatrixCoefficients 8.

apps/shared/y4m.c
```diff
@@ -201,6 +202,76 @@ static int y4mReadLine(FILE * inputFile, avifRWData * raw, const char * displayF
...
+        }
+    } else {
+        assert(avif->depth == 8);
+        if (avif->yuvRange == AVIF_RANGE_FULL) {
+            return AVIF_FALSE;
+        }
+
+        // Alpha is never in limited range.
+        const uint8_t minSampleValues[3] = { 16, 16, 16 };
+        const uint8_t maxSampleValues[3] = { 235, 240, 240 };
```
wantehchang, line 251: Same here: we need to handle MatrixCoefficients 0 and 8.

apps/shared/y4m.c
```diff
@@ -201,6 +202,76 @@ static int y4mReadLine(FILE * inputFile, avifRWData * raw, const char * displayF
     return -1;
 }
 
+// Limits each sample value to fit into avif->depth bits and avif->yuvRange.
```
wantehchang, line 205: 1. Optional: Since the AV1 encoder and decoder do not care whether the samples are full range or limited range, this function can ignore `avif->yuvRange` and clamp the samples based on `avif->depth` only. An advantage is that this will make the issue I noted at line 220 and line 251 moot.

2. Another design choice is for the function to fail (rather than clamp) if it detects an out-of-range sample.

y-guyon: > 1. Optional: Since the AV1 encoder and decoder do not care whether the samples are full range or limited range, this function can ignore `avif->yuvRange` and clamp the samples based on `avif->depth` only.

Done.

> An advantage is that this will make the issue I noted at line 220 and line 251 moot.

`avifRange` should still be documented in avif.h.

> 2. Another design choice is for the function to fail (rather than clamp) if it detects an out-of-range sample.

I believe being permissive on input is convenient for most users, at the expense of behavior transparency, which is mitigated by the warning.

apps/shared/y4m.c
```diff
@@ -201,6 +202,41 @@ static int y4mReadLine(FILE * inputFile, avifRWData * raw, const char * displayF
...
+        return AVIF_FALSE;
+    }
+    assert(avif->depth < 16); // Otherwise it could be skipped too.
+
+    // AV1 encoders and decoders do not care whether the samples are full range or limited range
+    // for the internal computation: it is only passed as an informative tag, so ignore avif->yuvRange.
+    const uint16_t maxSampleValue = (uint16_t)((1u << avif->depth) - 1u);
+
+    avifBool samplesWereClamped = AVIF_FALSE;
+    // Alpha is never in limited range.
```
wantehchang, line 220: 1. Delete this comment.

2. I think the for loop needs to go up to `AVIF_CHAN_A` as in the first commit of this pull request.

apps/shared/y4m.c
```diff
@@ -380,6 +416,12 @@ avifBool y4mRead(const char * inputFilename, avifImage * avif, avifAppSourceTimi
         }
     }
 
+    // libavif API does not guarantee the absence of undefined behavior if samples exceed the specified avif->depth.
+    // Avoid that by making sure input values are within the correct range.
+    if (y4mClampSamples(avif)) {
+        fprintf(stderr, "WARNING: some samples were clamped to fit into %u bits per channel per sample\n", avif->depth);
```
wantehchang, line 422: Nit: I think "per channel" could be omitted because "sample" generally means a single number. It's the word "pixel" that can mean multiple numbers (one per channel).

tests/gtest/avify4mtest.cc
```diff
@@ -56,6 +56,42 @@ TEST_P(Y4mTest, EncodeDecode) {
...
+            << height << "_" << bit_depth << "_" << yuv_format << "_"
+            << yuv_range << "_" << create_alpha;
+
+  testutil::AvifImagePtr image = testutil::CreateImage(
+      width, height, bit_depth, yuv_format,
+      create_alpha ? AVIF_PLANES_ALL : AVIF_PLANES_YUV, yuv_range);
+  ASSERT_NE(image, nullptr);
+
+  // Insert out-of-range values on purpose if possible.
+  const uint32_t yuva8[] = {255, 0, 255, 255};
```
wantehchang, line 77: To the new version of `y4mClampSamples()`, the values 0 and 255 are no longer out of range for bit depth 8. So this is confusing for people who don't know the history of `y4mClampSamples()`.

Similarly, in the next line the value 0 is no longer out of range for bit depths 10 and 12.

This can be addressed by clarifying the comment (i.e., saying that some values are out of range for limited range) or by modifying the values in these two arrays.

tests/gtest/avify4mtest.cc
```diff
@@ -56,6 +56,42 @@ TEST_P(Y4mTest, EncodeDecode) {
...
+  const uint32_t yuva16[] = {0, (1u << 16) - 1u, 0, (1u << 16) - 1u};
+  testutil::FillImagePlain(image.get(),
+                           avifImageUsesU16(image.get()) ? yuva16 : yuva8);
+  ASSERT_TRUE(y4mWrite(file_path.str().c_str(), image.get()));
+
+  // y4mRead() should clamp the values to respect the specified depth and range.
```
wantehchang, line 83: Delete "and range".

wantehchang (approved):
LGTM. I suggest some changes.

wantehchang (approved):
LGTM.

PR #1225, opened 2022-11-21
elst: Ensure version and entry_count are valid

vigneshvg:
* Ensure that version is either 0 or 1 (this is what we do with other boxes like tkhd, mdhd, etc).
* Ensure that entry_count != 1 (to disallow an entry_count of 0).


wantehchang (approved):
LGTM.

PR #1226, opened 2022-11-22
Distinguish between sequence and progressive

wantehchang:
In avifImageDump(), improve the check for an image sequence by excluding the progressive decoding case.

### DIFF
--- a/apps/shared/avifutil.c
+++ b/apps/shared/avifutil.c
@@ -135,7 +135,7 @@ void avifImageDump(const avifImage * avif, uint32_t gridCols, uint32_t gridRows,
 void avifContainerDump(const avifDecoder * decoder)
 {
     avifImageDumpInternal(decoder->image, 0, 0, decoder->alphaPresent, decoder->progressiveState);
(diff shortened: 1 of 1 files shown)


wantehchang:
@vigneshvg Just FYI.

jzern (approved):

wantehchang:
Here are the results of manual testing. The command I ran is

```
./avifdec --info --progressive animals_00_multilayer_a1lx.avif
```

where animals_00_multilayer_a1lx.avif can be downloaded from https://github.com/AOMediaCodec/av1-avif/blob/master/testFiles/Apple/multilayer_examples/animals_00_multilayer_a1lx.avif.

Commit 3fcb1410d787ced6a48e53bd29f5d4d5e97ca1bd (before this pull request):

The output contains the following line:
```
 * Repeat Count   : 0
```

Commit 38659ea3c19dd9ba2104c9fc2b98b6500eb85dc2 (this pull request):

The ` * Repeat Count   : 0` line is not in the output.

PR #1227, opened 2022-11-22
Rename the AVIF_UNKNOWN_DURATION64/32 macros

wantehchang:
The AVIF_UNKNOWN_DURATION64/32 macros are renamed
AVIF_INDEFINITE_DURATION64/32 because their values (all 1s) are described as indefinite duration in ISO/IEC 14496-12:2022. AVIF_UNKNOWN_DURATION64/32 look similar to the
AVIF_REPETITION_COUNT_UNKNOWN macro and could be confusing.

In src/write.c, correct the comments for the fields in the EditListBox (elst).

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -1545,7 +1545,7 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
         }
         uint64_t durationInTimescales;
         if (encoder->repetitionCount == AVIF_REPETITION_COUNT_INFINITE) {
-            durationInTimescales = AVIF_UNKNOWN_DURATION64;
+            durationInTimescales = AVIF_INDEFINITE_DURATION64;
         } else {
             uint64_t loopCount = encoder->repetitionCount + 1;
             assert(framesDurationInTimescales != 0);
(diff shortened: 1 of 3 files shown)


wantehchang:
@vigneshvg Just FYI.

src/write.c
```diff
@@ -1623,9 +1623,9 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
             avifBoxMarker elst = avifRWStreamWriteFullBox(&s, "elst", AVIF_BOX_SIZE_TBD, 1, elstFlags);
             avifRWStreamWriteU32(&s, 1);                          // unsigned int(32) entry_count;
             avifRWStreamWriteU64(&s, framesDurationInTimescales); // unsigned int(64) segment_duration;
-            avifRWStreamWriteU64(&s, 0);                          // unsigned int(64) media_time;
-            avifRWStreamWriteU16(&s, 1);                          // unsigned int(16) media_rate_integer;
-            avifRWStreamWriteU16(&s, 0);                          // unsigned int(16) media_rate_fraction = 0;
+            avifRWStreamWriteU64(&s, 0);                          // int(64) media_time;
```
jzern, line 1626: I take it the fields are ints and the comment may have from the lack of signed write calls?

jzern (approved):

src/write.c
wantehchang: I don't know where Vignesh copied the comments from. I checked the following three standards:

- ISO/IEC 23008-12:2017 (HEIF)
- ISO/IEC 14496-12:2020 (ISO BMFF)
- ISO/IEC 14496-12:2022 (ISO BMFF)

In these three standards, these fields are ints. They only differ in whether `media_rate_fraction` has the `= 0`.

vigneshvg: yes, sorry these are "int" fields. 

i mistakenly copy pasted them from the above lines as unsigned int.

PR #1234, opened 2022-11-29
Bump ext/libyuv to version 1850

y-guyon:
It includes a fix for building yuv_shared locally on Mac.

### DIFF
--- a/ext/libyuv.cmd
+++ b/ext/libyuv.cmd
@@ -14,7 +14,7 @@
 git clone --single-branch https://chromium.googlesource.com/libyuv/libyuv
 
 cd libyuv
-git checkout f9fda6e7
+git checkout 4a3c79cb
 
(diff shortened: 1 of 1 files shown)


ext/libyuv.cmd
```diff
@@ -14,7 +14,7 @@
 git clone --single-branch https://chromium.googlesource.com/libyuv/libyuv
 
 cd libyuv
-git checkout f9fda6e7
+git checkout 4a3c79cb
```
wantehchang, line 17: For libyuv version update, it is best to make sure the new version is being used by Chrome. This is because Chrome's test suite provides additional test coverage of libyuv.

Chrome's src/DEPS file shows it is using this revision. So this LGTM.

wantehchang (approved):

PR #1239, opened 2022-12-05
 Fix build error with msvc x86

cgohlke:
Fixes the following warnings and compile error when building libavif-v0.11.1 with Visual Studio 2022 v17.4 for x86 (32-bit):
```
read.c
D:\Build\_build\libavif-v0.11.1\src\read.c(948): error C2220: the following warning is treated as an error
D:\Build\_build\libavif-v0.11.1\src\read.c(948): warning C4244: '-=': conversion from 'uint64_t' to 'size_t', possible loss of data
```

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -950,7 +950,7 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, const a
                     return AVIF_RESULT_BMFF_PARSE_FAILED;
                 }
                 startOffset += remainingOffset;
-                extentSize -= remainingOffset;
+                extentSize = (size_t)((uint64_t)extentSize - remainingOffset);
                 remainingOffset = 0;
             }
         }
--- a/src/read.c
+++ b/src/read.c
@@ -950,7 +950,7 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, const a
                     return AVIF_RESULT_BMFF_PARSE_FAILED;
                 }
                 startOffset += remainingOffset;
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -950,7 +950,7 @@ static avifResult avifDecoderItemMaxExtent(const avifDecoderItem * item, const a
                     return AVIF_RESULT_BMFF_PARSE_FAILED;
                 }
                 startOffset += remainingOffset;
-                extentSize -= remainingOffset;
+                extentSize = (size_t)((uint64_t)extentSize - remainingOffset);
```
y-guyon, line 953: You may not need to do the intermediate `(uint64_t)` cast. `(extentSize - remainingOffset)` should be promoted to the biggest of `sizeof(extentSize)` and `sizeof(remainingOffset)`.

y-guyon (approved):
Thanks for sending a patch!

src/read.c
wantehchang: Thanks for the pull request.

Because of the `remainingOffset >= extentSize` check at line 945, we know `remainingOffset < extentSize` here. So an alternative fix, which requires only one cast, is the following:

```
                extentSize -=(size_t)remainingOffset;
```

This cast is safe because `remainingOffset < extentSize`  and `extentSize` is a `size_t`.

wantehchang (changes requested):

src/read.c
wantehchang: Christoph: I made this change in the second commit.

wantehchang (approved):

PR #1241, opened 2022-12-06
Allocate avifReformatState tables dynamically

y-guyon:
Instead of a fixed-length array that may be larger than necessary and uses 1<<12 * 2 = 8kB of stack memory which may be limited on some environments.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -466,7 +466,11 @@ static avifBool avifCreateYUVToRGBLookUpTables(float ** unormFloatTableY, float 
             *unormFloatTableUV = *unormFloatTableY;
         } else {
             *unormFloatTableUV = avifAlloc(cpCount * sizeof(**unormFloatTableUV));
-            AVIF_CHECK(*unormFloatTableUV);
+            if (!*unormFloatTableUV) {
+                avifFree(*unormFloatTableY);
+                *unormFloatTableY = NULL;
+                return AVIF_FALSE;
+            }
             for (uint32_t cp = 0; cp < cpCount; ++cp) {
                 // Review this when implementing YCgCo limited range support.
                 (*unormFloatTableUV)[cp] = ((float)cp - state->biasUV) / state->rangeUV;
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -136,20 +136,6 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
     state->rangeY = (float)((state->yuvRange == AVIF_RANGE_LIMITED) ? (219 << (state->yuvDepth - 8)) : state->yuvMaxChannel);
     state->rangeUV = (float)((state->yuvRange == AVIF_RANGE_LIMITED) ? (224 << (state->yuvDepth - 8)) : state->yuvMaxChannel);
 
-    uint32_t cpCount = 1 << image->depth;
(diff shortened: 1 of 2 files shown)


src/reformat.c
```diff
@@ -485,8 +513,9 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image, avifRGBIm
     const float kr = state->kr;
     const float kg = state->kg;
     const float kb = state->kb;
-    const float * const unormFloatTableY = state->unormFloatTableY;
-    const float * const unormFloatTableUV = state->unormFloatTableUV;
+    float * unormFloatTableY = NULL;
+    float * unormFloatTableUV = NULL;
+    AVIF_CHECKERR(avifCreateYUVToRGBLookUpTables(&unormFloatTableY, &unormFloatTableUV, image->depth, state), AVIF_RESULT_OUT_OF_MEMORY);
```
jzern, line 518: with some synchronization on the init, this could be done once to a global, though that isn't done elsewhere in the lib currently.

y-guyon: For most use cases `avifImageYUVToRGBLibYUV()` avoids the need for lookup tables. I do not think it is worth it to save that work to a global.

On a related note, I plan to do the computation without lookup tables based on some heuristic, such as `sample_count < lut_size`, in another PR.

src/reformat.c
```diff
@@ -462,6 +448,48 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
...
+        (*unormFloatTableY)[cp] = ((float)cp - state->biasY) / state->rangeY;
+    }
+
+    if (unormFloatTableUV) {
+        if (state->mode == AVIF_REFORMAT_MODE_IDENTITY) {
+            // Just reuse the luma table since the chroma values are the same.
+            *unormFloatTableUV = *unormFloatTableY;
+        } else {
+            *unormFloatTableUV = avifAlloc(cpCount * sizeof(**unormFloatTableUV));
+            AVIF_CHECK(*unormFloatTableUV);
```
jzern, line 469: *unormFloatTableY will leak here if the check fails

jzern (approved):

src/reformat.c
```diff
@@ -462,6 +448,52 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
...
+            *unormFloatTableUV = *unormFloatTableY;
+        } else {
+            *unormFloatTableUV = avifAlloc(cpCount * sizeof(**unormFloatTableUV));
+            if (!*unormFloatTableUV) {
+                avifFree(*unormFloatTableY);
+                *unormFloatTableY = NULL;
+                return AVIF_FALSE;
+            }
+            for (uint32_t cp = 0; cp < cpCount; ++cp) {
+                // Review this when implementing YCgCo limited range support.
```
wantehchang, line 475: Please duplicate this comment before line 460, because this comment applies to both lookup tables.

(Alternatively, move this comment before the function.)

wantehchang (commented):
LGTM. I suggest a comment change.

src/reformat.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1245.

PR #1242, opened 2022-12-06
Move avifImageIsOpaque() from write.c to avif.h

y-guyon:
Make it public so that it can be used in aviftest_helpers.cc and by users.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1175,6 +1175,7 @@ AVIF_API void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const cha
 
 // Helpers
 AVIF_API avifBool avifImageUsesU16(const avifImage * image);
+AVIF_API avifBool avifImageIsOpaque(const avifImage * image);
 // channel can be an avifChannelIndex.
 AVIF_API uint8_t * avifImagePlane(const avifImage * image, int channel);
 AVIF_API uint32_t avifImagePlaneRowBytes(const avifImage * image, int channel);
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -151,7 +151,7 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
     }
 
     state->toRGBAlphaMode = AVIF_ALPHA_MULTIPLY_MODE_NO_OP;
-    if (image->alphaPlane) {
+    if (!avifImageIsOpaque(image)) {
         if (!avifRGBFormatHasAlpha(rgb->format) || rgb->ignoreAlpha) {
             // if we are converting some image with alpha into a format without alpha, we should do 'premultiply alpha' before
             // discarding alpha plane. This has the same effect of rendering this image on a black background, which makes sense.
--- a/tests/gtest/avifopaquetest.cc
+++ b/tests/gtest/avifopaquetest.cc
@@ -1,0 +1,55 @@
+// Copyright 2022 Google LLC
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "avif/avif.h"
+#include "avifpng.h"
+#include "aviftest_helpers.h"
+#include "gtest/gtest.h"
+
+namespace libavif {
+namespace {
+
+TEST(OpaqueTest, AlphaAndNoAlpha) {
+  for (bool alpha_is_opaque : {false, true}) {
+    for (int depth : {8, 10, 12}) {
+      testutil::AvifImagePtr alpha = testutil::CreateImage(
+          1, 1, depth, AVIF_PIXEL_FORMAT_YUV444, AVIF_PLANES_ALL);
+      testutil::AvifImagePtr no_alpha = testutil::CreateImage(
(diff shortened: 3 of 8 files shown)


include/avif/avif.h
```diff
@@ -1175,6 +1175,7 @@ AVIF_API void avifEncoderSetCodecSpecificOption(avifEncoder * encoder, const cha
 
 // Helpers
 AVIF_API avifBool avifImageUsesU16(const avifImage * image);
+AVIF_API avifBool avifImageIsOpaque(const avifImage * image);
```
jzern, line 1178: there should be a changelog entry for the API modification

src/reformat.c
```diff
@@ -151,7 +151,7 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
     }
 
     state->toRGBAlphaMode = AVIF_ALPHA_MULTIPLY_MODE_NO_OP;
-    if (image->alphaPlane) {
+    if (!avifImageIsOpaque(image)) {
```
wantehchang, line 154: This change needs to be reviewed carefully. I suggest excluding this change from this pull request so that we can merge the rest of the pull request immediately.

For example, this function (`avifPrepareReformatState()`) is also called by `avifImageRGBToYUV()`. In the case of `avifImageRGBToYUV()`, the `image` parameteter is the output, so we should not read its alpha pixels.

I am wondering if you made this change to prevent us from going to line 159 unnecessarily when doing `avifImageYUVToRGB()`?

y-guyon: > This change needs to be reviewed carefully. I suggest excluding this change from this pull request so that we can merge the rest of the pull request immediately.

Done.

> For example, this function (`avifPrepareReformatState()`) is also called by `avifImageRGBToYUV()`. In the case of `avifImageRGBToYUV()`, the `image` parameteter is the output, so we should not read its alpha pixels.

I missed that. It works because only `state->toRGBAlphaMode` is set and ignored by `avifImageRGBToYUV()`, but I agree alpha samples should not be read.

> I am wondering if you made this change to prevent us from going to line 159 unnecessarily when doing `avifImageYUVToRGB()`?

Yes. Opaque alpha and missing alpha should have the same output, no matter the `avifAlphaMultiplyMode`.
Unfortunately `toRGBAlphaMode != NO_OP` results in different YUV-to-RGB conversion algorithms being used (slow built-in for opaque alpha and libyuv for missing alpha). See #1246.

tests/gtest/avifopaquetest.cc
```diff
@@ -0,0 +1,55 @@
...
+    const uint32_t max_value = (1u << depth) - 1;
+    const uint32_t yuva[] = {max_value, max_value, max_value, max_value};
+    testutil::FillImagePlain(opaque_alpha.get(), yuva);
+
+    // View on YUV, to make sure the color values are identical.
+    testutil::AvifImagePtr no_alpha(avifImageCreateEmpty(), avifImageDestroy);
+    const avifCropRect rect = {0, 0, opaque_alpha->width, opaque_alpha->height};
+    ASSERT_EQ(avifImageSetViewRect(no_alpha.get(), opaque_alpha.get(), &rect),
+              AVIF_RESULT_OK);
+    avifImageFreePlanes(no_alpha.get(), AVIF_PLANES_A);
```
wantehchang, line 45: This `avifImageFreePlanes()` call is confusing. `avifImageSetViewRect()` returns an image that doesn't own the plane buffers. It's not clear (unless one inspects the code or reads the documentation) whether it is safe to call `avifImageFreePlanes()` on such an image.

I suggest replacing the `avifImageSetViewRect()` call with an `avifImageCopy()` call that only copies the YUV planes. This requires setting up the `opaque_alpha` image differently (first `testutil::FillImageGradient(opaque_alpha.get())` then fill its alpha channel with `max_value`).

tests/gtest/avifopaquetest.cc
```diff
@@ -0,0 +1,55 @@
...
+                               alpha_is_opaque ? max_value : (max_value - 1)};
+      testutil::FillImagePlain(alpha.get(), yuva);
+      testutil::FillImagePlain(no_alpha.get(), yuva);
+
+      EXPECT_EQ(testutil::AreImagesEqual(*alpha, *no_alpha), alpha_is_opaque);
+    }
+  }
+}
+
+TEST(OpaqueTest, Gradient) {
```
wantehchang, line 31: This test seems very similar to the previous test. The only differences I can see are the image size and the values of the YUV samples.

wantehchang (approved):
LGTM.

tests/gtest/avifopaquetest.cc
y-guyon: Removed the test.

tests/gtest/avifopaquetest.cc
y-guyon: Removed the test.

wantehchang (approved):
Still LGTM.

PR #1246, opened 2022-12-08
Add avifalphamulttest

y-guyon:


tests/CMakeLists.txt
```diff
@@ -75,6 +75,11 @@ if(AVIF_ENABLE_GTEST)
     target_include_directories(avifallocationtest PRIVATE ${GTEST_INCLUDE_DIRS})
     add_test(NAME avifallocationtest COMMAND avifallocationtest)
 
+    add_executable(avifalphamulttest gtest/avifalphamulttest.cc)
```
wantehchang, line 78: Nit: If we only use four characters, I suggest changing "mult" to "prem" because "prem" is the name used in the file format (HEIF?).

tests/gtest/avifalphamulttest.cc
```diff
@@ -0,0 +1,66 @@
...
+    testutil::AvifRgbImage opaque_rgb(opaque_alpha.get(), opaque_alpha->depth,
+                                      AVIF_RGB_FORMAT_RGB);
+    ASSERT_EQ(avifImageYUVToRGB(opaque_alpha.get(), &opaque_rgb),
+              AVIF_RESULT_OK);
+    testutil::AvifRgbImage no_alpha_rgb(no_alpha.get(), no_alpha->depth,
+                                        AVIF_RGB_FORMAT_RGB);
+    ASSERT_EQ(avifImageYUVToRGB(no_alpha.get(), &no_alpha_rgb), AVIF_RESULT_OK);
+
+    // YUV samples with opaque alpha and with missing alpha should be converted
+    // to the same RGB values.
```
wantehchang, line 54: This goal is reasonable, but it is also reasonable to expect that a missing alpha channel and an opaque alpha channel may be considered as different in some cases. We don't need to overly constrain our implementation if this goal becomes inconvenient.

It may be better to tolerate small differences so that we can use different conversion functions. The `avifImageIsOpaque()` check is not free. If an opaque alpha plane is uncommon, then it may not make sense to perform this check.

wantehchang (approved):
LGTM.

Thank you for adding tests for the premultiplied alpha code. The logic related to premultiplied alpha in src/reformat.c is very complicated.

I have a high-level comment about a requirement that may overly constrain our implementation options.

tests/gtest/avifalphamulttest.cc
y-guyon: > it is also reasonable to expect that a missing alpha channel and an opaque alpha channel may be considered as different in some cases.

Do you have examples of such cases?

> We don't need to overly constrain our implementation if this goal becomes inconvenient.

I was very surprised when I got different results (visible difference with a simple gradient exported to a local png). I thought there was a bug and I spent time figuring out if I did something wrong or not. I believe it would be better to fix this to avoid other people spending time on this topic too in the future. Maybe there could be a visible-enough disclaimer instead of a fix, but at the same time, using user attention for that seems unreasonable.

> It may be better to tolerate small differences so that we can use different conversion functions.

We could perform this test by disabling any use of libyuv and libsharpyuv, but I guess it may still use one of the "fast" built-in conversion functions instead of the slow generic 

> The `avifImageIsOpaque()` check is not free. If an opaque alpha plane is uncommon, then it may not make sense to perform this check.

`avifImageIsOpaque()` is kind of free for translucent images: the odds of having a nearly opaque image with translucent pixels only at the end of the buffer are fairly small. Otherwise it will return false very quickly.

For opaque images it will check all the pixels, but as you said it might be uncommon. And the benefit would be to use libyuv in that case instead of the slow built-in conversion, so probably faster overall.

wantehchang (approved):
LGTM.

PR #1212, opened 2022-11-15
Some cleanup of the new CLLI code

wantehchang:
Small changes to the new CLLI code recently added in https://github.com/AOMediaCodec/libavif/pull/1194.

Document the CLLI metadata read and write support in the CHANGELOG.

Print uint16_t (unsigned short) values using the %hu format.

Put the clli member next to the other color-related members in the avifImage struct.

PR #1248, opened 2022-12-09
Copy clli in avifImageCopyNoAlloc()

y-guyon:
Add avifcllitest.
Follow up of https://github.com/AOMediaCodec/libavif/pull/1194 and https://github.com/AOMediaCodec/libavif/pull/1212.

### DIFF
--- a/tests/gtest/avifcllitest.cc
+++ b/tests/gtest/avifcllitest.cc
@@ -15,9 +15,10 @@ TEST(ClliTest, Simple) {
   ASSERT_NE(image, nullptr);
   testutil::FillImageGradient(image.get());  // The pixels do not matter.
 
-  for (uint16_t max_content_light_level : {0, 1, 65535}) {
-    for (uint16_t max_pic_average_light_level : {0, 1, 65535}) {
-      image->clli = {max_content_light_level, max_pic_average_light_level};
+  for (int max_content_light_level : {0, 1, 65535}) {
+    for (int max_pic_average_light_level : {0, 1, 65535}) {
+      image->clli.maxCLL = max_content_light_level;
+      image->clli.maxPALL = max_pic_average_light_level;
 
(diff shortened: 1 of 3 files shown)


vrabaud (approved):

tests/gtest/avifcllitest.cc
```diff
@@ -0,0 +1,34 @@
...
+
+TEST(ClliTest, Simple) {
+  testutil::AvifImagePtr image =
+      testutil::CreateImage(/*width=*/8, /*height=*/8, /*depth=*/8,
+                            AVIF_PIXEL_FORMAT_YUV444, AVIF_PLANES_YUV);
+  ASSERT_NE(image, nullptr);
+  testutil::FillImageGradient(image.get());  // The pixels do not matter.
+
+  for (int max_content_light_level : {0, 1, 65535}) {
+    for (int max_pic_average_light_level : {0, 1, 65535}) {
```
wantehchang, line 19: Should we declare the loop index variables as `uint16_t` to avoid the casts in lines 20-21?

wantehchang (commented):
LGTM.

I audited the source tree. We should also add a command-line option to avifenc.c to set CLLI.

wantehchang (commented):
I am wondering if `avifDecoderDataFillImageGrid()` should check all grid tiles have the same clli or set `dstImage->clli` to `firstTile->image->clli`.

tests/gtest/avifcllitest.cc
wantehchang: I found the answer to my question: the initializer list `{0, 1, 65535}` has the element type `int`, and when we assign an `int` value to a `uint16_t` variable (either the loop index variable or the clli struct members) we need an explicit cast to avoid the C4244 "possible loss of data" warning under MSVC.

PR #1254, opened 2022-12-14
Handle relative vs absolute paths for .pc file variables

### LINKED ISSUE #1220: Improve pkg-config file generation
It would be good to improve the CMake generated .pc file for relocatability - namely, it should use [${prefix} etc. variables and relative paths whenever possible](https://people.freedesktop.org/~dbn/pkg-config-guide.html), but still support absolute paths when so required.

See e.g. [libjxl](https://github.com/libjxl/libjxl/blob/3f6ec60ce6890132c1e2ea9719de5981042a1d96/lib/jxl.cmake#L642-L663) and [libtiff](https://gitlab.com/libtiff/libtiff/-/blob/31d9e9f73dc8b6fa599c931d9ff058d8ea32795c/cmake/PkgConfig.cmake#L48-63)

kmilos:
Resolves #1220


CMakeLists.txt
```diff
@@ -537,6 +537,17 @@ if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
         )
     endif()
 
+    # Handle both relative and aboslute paths (e.g. NixOS) for a relocatable package
```
y-guyon, line 540: "aboslute" typo

kmilos: Ta 🤦

CMakeLists.txt
```diff
@@ -537,6 +537,17 @@ if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
         )
     endif()
 
+    # Handle both relative and absolute paths (e.g. NixOS) for a relocatable package
+    if(IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
```
wantehchang, line 541: I am not familiar with cmake, so please pardon my questions.

The original code uses `CMAKE_INSTALL_FULL_LIBDIR` and `CMAKE_INSTALL_FULL_INCLUDEDIR`. How are they different from the `CMAKE_INSTALL_INCLUDEDIR` and `CMAKE_INSTALL_LIBDIR` variables used in the new code?

Are `CMAKE_INSTALL_FULL_LIBDIR` and `CMAKE_INSTALL_FULL_INCLUDEDIR` always absolute paths?

wantehchang (approved):
LGTM.

CMakeLists.txt
jzern: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

There are a few special cases mentioned that might make a difference.

wantehchang: Thank you, James. I searched for `CMAKE_INSTALL_FULL_LIBDIR` and `CMAKE_INSTALL_FULL_INCLUDEDIR` in the cmake documentation. I should have searched for `CMAKE_INSTALL_FULL_` instead.

The description of `CMAKE_INSTALL_FULL_<dir>` seems to imply the new code is equivalent to the original code except in those special cases. Since Miloš also wrote the original code (in https://github.com/AOMediaCodec/libavif/pull/692), I trust that the new code is better. But I'd still be interested in knowning what the differences are.

jzern: I think it might come down to the special CMAKE_INSTALL_PREFIX's mentioned:
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#special-cases

If a user specified an absolute CMAKE_INSTALL_LIBDIR, though, maybe it doesn't make a difference.

kmilos: The `CMAKE_INSTALL_FULL_{INCLUDE,LIB}DIR` I did author before was indeed correct for all installs and platform, but turns out missing the nice property of being relocatable (i.e. you could not move it from the originally installed location). OTOH, `CMAKE_INSTALL_{INCLUDE,LIB}DIR` is not guaranteed to be relative, with NixOS being a notable exception (and there are perhaps more), so it cannot be used directly as e.g. just `${prefix}/@CMAKE_INSTALL_INCLUDEDIR@` placeholder, hence this acrobatics.

More background on the issue can be found e.g. here: https://github.com/libjxl/libjxl/issues/1400

PR #1259, opened 2023-01-11
Use libyuv with downshift for 12/10 -> 8 bit when possible

vigneshvg:
When converting 12-bit/10-bit YUV images into 8-bit RGB, try to use the following path first (before falling back to built-in conversion using C code): 12/10-bit YUV -> 8-bit YUV (using libyuv) -> 8-bit RGB (using libyuv).

This path is about ~40% faster than using the built-in conversion.

Reviewed by @wantehchang in: https://aomedia-review.googlesource.com/c/libavif/+/169088


src/reformat_libyuv.c
```diff
@@ -514,139 +665,57 @@ avifResult avifImageYUVToRGBLibYUV8bpc(const avifImage * image,
...
-                                         rgb->rowBytes,
-                                         matrix,
-                                         image->width,
-                                         image->height,
-                                         filter) != 0) {
-                    return AVIF_RESULT_REFORMAT_FAILED;
-                }
-                return AVIF_RESULT_OK;
-            }
+static avifResult avifImageDownscaleTo8bpcYUVAndConvertToRGBLibYUV(const avifImage * image,
```
wantehchang, line 677: Optional: Maybe change "Downscale" to "Downshift"?

"Downscale" could also mean reducing the size of the image.

wantehchang (approved):
LGTM. I suggest changing "Downscale" to "Downshift".

jzern:
The CI failure looks to be unrelated. The rav1e build step failed, but didn't report a failure. Only the later attempt to use the library broke:

```
error: failed to compile `cargo-c v0.9.15+cargo-0.67`, intermediate artifacts can be found at `/tmp/cargo-installFbhuwk`

Caused by:
  package `cargo-c v0.9.15+cargo-0.67` cannot be built because it requires rustc 1.66 or newer, while the currently active rustc version is 1.65.0
  Try re-running cargo install with `--locked`
error: no such subcommand: `cinstall`

	Did you mean `install`?

	View all installed commands with `cargo --list`
```

src/reformat_libyuv.c
jzern: or Downsample might work

vigneshvg: Changed it to downsample.

wantehchang: "Downsample" and "subsample" mean the same thing. It means reducing sampling rate, e.g., chroma subsampling.

vigneshvg: Ack. I updated it to "downshift".

wantehchang:
I can reproduce the failure of the `cargo install cargo-c` command locally because I have rustc 1.61.0 installed on my Linux computer.

Maybe we need to install rustc 1.66 or newer in our GitHub os images?

jzern:
> Maybe we need to install rustc 1.66 or newer in our GitHub os images?

Seems so or try to force an older version of rav1e temporarily.

wantehchang:
I am testing installing the stable Rust compiler in https://github.com/AOMediaCodec/libavif/pull/1261.

PR #1260, opened 2023-01-11
Run clang-format

wantehchang:


wantehchang:
@y-guyon While working on something else, I ran clang-format on avif.h and was surprised to see some changes. So I created this pull request by running clang-format on all the C/C++ files. Do you know if these changes are caused by a new version of clang-format?

wantehchang:
Yannis: Could you run clang-format in your environment and see if you get the same changes? Thanks.

All the changes are related to aligning the end-of-line `//` comments. Apparently clang-format now works harder to align those comments, even across block boundaries (curly braces `{ }`).

y-guyon:
I remember having the same issue. `clang-format` was also updated on my local environment and generated the same changes as this PR. I now locally use `clang-format-12` to force the old behavior.

I upgraded the presubmit `clang-format` version from 12 to 14 in https://github.com/AOMediaCodec/libavif/pull/1229 but the `clang-format-check` GitHub workflow still has the old behavior: [this PR failed](https://github.com/wantehchang/libavif/actions/runs/3895478822/jobs/6650844489).

`AlignTrailingComments: true` in [`.clang-format`](https://github.com/AOMediaCodec/libavif/blob/93035c173c934813b16751872fbe2f535aa7086b/.clang-format#L9) was replaced by an enum or struct in newer versions, see https://clang.llvm.org/docs/ClangFormatStyleOptions.html:

> As of clang-format 16 this option is not a bool but can be set to the options. Conventional bool options still can be parsed as before.
 
We may be able to keep the old behavior by configuring it properly, but it requires clang-format 16, so it depends on https://github.com/RafikFarhad/clang-format-github-action which depends on https://hub.docker.com/r/silkeh/clang which has clang 15 but not 16 (if that equals clang-format version numbering).

I suggest relying on `clang-format-12` for now.

wantehchang:
Yannis: Thank you for the reply. I found that there are also clang-format-13 and clang-format-14, and they all format the libavif source code in the same way. Should we use `clang-format-14` then?

y-guyon:
> Should we use `clang-format-14` then?

I did not notice a difference with `clang-format-12`, so you can use any I guess.

wantehchang:
Thank you. I will use clang-format-14 locally then.

PR #1261, opened 2023-01-11
GitHub workflows: Install stable Rust compiler

wantehchang:
Copied from
https://github.com/actions-rs/meta/blob/master/recipes/quickstart.md.

cargo-c v0.9.15+cargo-0.67 requires rustc 1.66 or newer. The rustc
installed in the ubuntu-latest os image is rustc 1.65.0.

ci-disable-gtest.yml, ci-windows.yml, and ci.yml build rav1e, so only
these three workflows are modified.

### DIFF
--- a/.github/workflows/ci-disable-gtest.yml
+++ b/.github/workflows/ci-disable-gtest.yml
@@ -22,6 +22,11 @@ jobs:
     - uses: actions/setup-python@v4
       with:
         python-version: '3.x'
+    - uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
 
     - name: Cache external dependencies
       id: cache-ext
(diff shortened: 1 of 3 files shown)


wantehchang:
The problem seems to be that the rustc installed in GitHub's `ubuntu-latest` os image is rustc 1.65.0. That's why we are only running into this problem in `ubuntu-latest`.

The rustc installed in GitHub's `macos-latest` os image is rustc 1.66.0. I bet the `windows-latest` os image has the same version of rustc.

wantehchang:
I verified that the `windows-latest` os image also has rustc 1.66.0 installed.

vigneshvg (approved):

wantehchang:
It seems that the rustc in the 'ubuntu-latest' os image has just been updated to rustc 1.66.0. So we may not need this pull request now.

wantehchang:
Hmm..., there seem to be multiple `ubuntu-latest` os images.

This one has rustc 1.66.0: https://github.com/AOMediaCodec/libavif/actions/runs/3897377112/jobs/6654989847

But this one still has rustc 1.65.0: https://github.com/AOMediaCodec/libavif/actions/runs/3897377113/jobs/6654990024

PR #1264, opened 2023-01-13
Refactor YUV->RGB conversion with libyuv

vigneshvg:
Rewrite the YUV to RGB conversion functions that use libyuv. Split the code into two logical pieces: function pointer selection and function pointer application.

The function pointer selection contains the logic to select the appropriate libyuv function pointer for all the bitdepths. In the function pointer step, we simply apply whichever function pointer was selected in the prior step (downshifting to 8bit if necessary).

This makes the code simpler to parse visually and avoid some redundant computations.

This is simply a refactoring of code. The functionality should remain unchanged.

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -42,6 +42,7 @@ unsigned int avifLibYUVVersion(void)
 
 #include <assert.h>
 #include <limits.h>
+#include <string.h>
 
 #if defined(__clang__)
 #pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


vigneshvg:
Reviewed here: https://aomedia-review.googlesource.com/c/libavif/+/169241

src/reformat_libyuv.c
```diff
@@ -335,64 +336,270 @@ avifResult avifImageRGBToYUVLibYUV8bpc(avifImage * image, const avifRGBImage * r
...
+        lcf->yuvToRgbMatrixFilter = lutYuvToRgbMatrixFilter[rgb->format][yuvFormat];
+        if (lcf->yuvToRgbMatrixFilter != NULL) {
+            return AVIF_TRUE;
+        }
+        if (!nearestNeighborFilterAllowed(rgb->chromaUpsampling)) {
+            return AVIF_FALSE;
+        }
+    }
+    lcf->yuvToRgbMatrix = lutYuvToRgbMatrix[rgb->format][yuvFormat];
+    return lcf->yuvToRgbMatrix != NULL;
```
wantehchang, line 556: Optional: We can rewrite lines 546-556 to have a similar structure as the high-bit-depth case:

```
    if (yuvFormat != AVIF_PIXEL_FORMAT_YUV444) {
        lcf->yuvToRgbMatrixFilter = lutYuvToRgbMatrixFilter[rgb->format][yuvFormat];
        if (lcf->yuvToRgbMatrixFilter != NULL) {
            return AVIF_TRUE;
        }
    }
    if (yuvFormat == AVIF_PIXEL_FORMAT_YUV444 || nearestNeighborFilterAllowed(rgb->chromaUpsampling)) {
        lcf->yuvToRgbMatrix = lutYuvToRgbMatrix[rgb->format][yuvFormat];
        if (lcf->yuvToRgbMatrix != NULL) {
            return AVIF_TRUE;
        }
    }
    return AVIF_FALSE;
```

wantehchang (approved):

src/reformat_libyuv.c
wantehchang: Note: Although I suggested the optional rewrite, I actually prefer the orginal. simpler structure of the 8-bit case, but it is a little confusing to see the high-bit-depth case look quite different from the 8-bit case.

vigneshvg: yes, while there is a difference between the two cases, the different is apparent once you think about the fact that the high bitdepth case falls through and the 8-bit case does not.

so i will leave this as-is.

src/reformat_libyuv.c
```diff
@@ -707,166 +704,113 @@ static avifResult avifImageDownshiftTo8bpcYUVAndConvertToRGBLibYUV(const avifIma
...
+        return AVIF_RESULT_NOT_IMPLEMENTED;
+    }
 
+    LibyuvConversionFunction lcf;
+    if (!getLibYUVConversionFunction(image->yuvFormat, image->depth, rgb, &lcf)) {
+        return AVIF_RESULT_NOT_IMPLEMENTED;
+    }
     avifBool isYVU = lutIsYVU[rgb->format];
     const struct YuvConstants * matrix = isYVU ? matrixYVU : matrixYUV;
-    if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) {
```
wantehchang, line 748: The special handling of `AVIF_PIXEL_FORMAT_YUV400` in the high bit depth case is also missing in the new code. The new code relies on the lookup tables having `NULL` entries for `AVIF_PIXEL_FORMAT_YUV400`, so it is still correct.

vigneshvg: yes, that table does not exist as of now. so i think this is fine.

PR #1265, opened 2023-01-18
android_jni: Make thread count configurable

vigneshvg:
Add a new function which lets the user pass the number of decoding threads as part of the JNI API.

### DIFF
--- a/android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
+++ b/android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
@@ -58,5 +58,23 @@ public class AvifDecoder {
    * @return true on success and false on failure. A few possible reasons for failure are: 1) Input
    *     was not valid AVIF. 2) Bitmap was not large enough to store the decoded image.
    */
-  public static native boolean decode(ByteBuffer encoded, int length, Bitmap bitmap);
+  public static boolean decode(ByteBuffer encoded, int length, Bitmap bitmap) {
+    return decode(encoded, length, bitmap, 0);
+  }
+
+  /**
+   * Decodes the AVIF image into the bitmap.
+   *
+   * @param encoded The encoded AVIF image. encoded.position() must be 0.
+   * @param length Length of the encoded buffer.
+   * @param bitmap The decoded pixels will be copied into the bitmap.
+   * @param threads Number of threads to be used for the AVIF decode. Zero means use number of CPU
+   *     cores as the thread count. Negative values are invalid. When this value is > 0, it is
+   *     simply mapped to the maxThreads parameter in libavif. For more details, see the
+   *     documentation for maxThreads variable in avif.h.
+   * @return true on success and false on failure. A few possible reasons for failure are: 1) Input
+   *     was not valid AVIF. 2) Bitmap was not large enough to store the decoded image. 3) Negative
+   *     value was passed for the threads parameter.
+   */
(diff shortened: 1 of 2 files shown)


android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
```diff
@@ -58,5 +58,21 @@ public static boolean isAvifImage(ByteBuffer buffer) {
...
+    return decode(encoded, length, bitmap, 0);
+  }
+
+  /**
+   * Decodes the AVIF image into the bitmap.
+   *
+   * @param encoded The encoded AVIF image. encoded.position() must be 0.
+   * @param length Length of the encoded buffer.
+   * @param bitmap The decoded pixels will be copied into the bitmap.
+   * @param threads Number of threads to be used for the AVIF decode. Zero means use number of CPU
```
jzern, line 71: There can be confusion sometimes between thread count and 'creating additional' threads.

jzern (approved):

android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
vigneshvg: added a note pointing to maxThreads variable in avif.h.

jzern: thanks, lgtm

jzern (approved):

android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
```diff
@@ -58,5 +58,22 @@ public static boolean isAvifImage(ByteBuffer buffer) {
    * @return true on success and false on failure. A few possible reasons for failure are: 1) Input
    *     was not valid AVIF. 2) Bitmap was not large enough to store the decoded image.
    */
-  public static native boolean decode(ByteBuffer encoded, int length, Bitmap bitmap);
+  public static boolean decode(ByteBuffer encoded, int length, Bitmap bitmap) {
+    return decode(encoded, length, bitmap, 0);
```
wantehchang, line 62: An alternative is to pass `android_getCpuCount()` here, for backward compatibility.

Then the new native `decode` method doesn't need to give `threads=0` a special meaning. This will avoid the potential confusion that libavif also treats `maxThreads=0` in the same way. This confusion may arise because the comment at lines 72-73 refers to the libavif documentation for `maxThreads` for more details on the `threads` parameter.

wantehchang (approved):
LGTM. I describe an alternative for your consideration.

android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
vigneshvg: > An alternative is to pass `android_getCpuCount()` here, for backward compatibility.
> 

`android_getCpuCount()` is a native function available through a dependent native library. it requires a bit more code to be able to call that from here.

> Then the new native `decode` method doesn't need to give `threads=0` a special meaning. This will avoid the potential confusion that libavif also treats `maxThreads=0` in the same way. This confusion may arise because the comment at lines 72-73 refers to the libavif documentation for `maxThreads` for more details on the `threads` parameter.

I have updated the comment to clarify that it is mapped to libavif's `maxThreads` only when the value is >= 0.

PR #1269, opened 2023-01-20
Disable ProgressiveTest.DimensionChange

wantehchang:
This is better than skip the test unconditionally.


wantehchang:
@tongyuantongyu FYI.

vigneshvg (approved):

PR #1273, opened 2023-01-25
Call color config aom_codec_control on first image

wantehchang:
The color config is part of the AV1 sequence header OBU, so the aom_codec_control() calls related to the color config only need to be called before encoding the first image.

Also if a value is the default in libaom, then skip the aom_codec_control() call. aom_codec_control() does a lot of extra work in addition to setting a config parameter, so we should try to skip unneeded aom_codec_control() calls.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -802,6 +802,30 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             }
         }
 
+        // Set color_config() in the sequence header OBU.
+        if (alpha) {
+            aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_RANGE, AOM_CR_FULL_RANGE);
+        } else {
+            // libaom's defaults are AOM_CICP_CP_UNSPECIFIED, AOM_CICP_TC_UNSPECIFIED,
+            // AOM_CICP_MC_UNSPECIFIED, AOM_CSP_UNKNOWN, and 0 (studio/limited range). Call
+            // aom_codec_control() only if the values are not the defaults.
+            if (image->colorPrimaries != AVIF_COLOR_PRIMARIES_UNSPECIFIED) {
+                aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_PRIMARIES, (int)image->colorPrimaries);
+            }
+            if (image->transferCharacteristics != AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED) {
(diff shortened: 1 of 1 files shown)


y-guyon (approved):
Is it checked somewhere that `image->colorPrimaries` and others do not change between two `aomCodecEncodeImage()` calls?

wantehchang:
I don't think libavif checks that the color config parameters do not change between two `aomCodecEncodeImage()` calls. I can add a check. I guess it can be done similar to the check for unchangeable encoder options.

Note: I also analyzed libaom source code to see how it responds to changes to the color config parameters after encoding has started. I only analyzed the `color_range` parameter. I found that `color_range` is used when writing the sequence header OBU (which we already knew about). I didn't study how libaom determines if it should write a new sequence header OBU, so I don't know if it will write a new sequence header OBU in response to a change of `color_range`. I did find that `color_range` affects the value of `clip_to_restricted_range` in `film_grain_params()`, which is part of the frame header. So it is important that applications not change `color_range` after encoding the first frame.

wantehchang:
Yannis: When I tried to implement the check that you asked about, I found that the check already exists. It's the following code in `avifEncoderAddImageInternal()` in src/write.c:

```
    if (encoder->data->items.count == 0) {
        // Make a copy of the first image's metadata (sans pixels) for future writing/validation
        const avifResult copyResult = avifImageCopy(encoder->data->imageMetadata, firstCell, 0);
        ...
    } else {
        // Another frame in an image sequence, or layer in a layered image

        // If the first image in the sequence had an alpha plane (even if fully opaque), all
        // subsequent images must have alpha as well.
        if ((imageMetadata->depth != firstCell->depth) || (imageMetadata->yuvFormat != firstCell->yuvFormat) ||
            (imageMetadata->yuvRange != firstCell->yuvRange) || (imageMetadata->colorPrimaries != firstCell->colorPrimaries) ||
            (imageMetadata->transferCharacteristics != firstCell->transferCharacteristics) ||
            (imageMetadata->matrixCoefficients != firstCell->matrixCoefficients) ||
            (imageMetadata->alphaPremultiplied != firstCell->alphaPremultiplied) ||
            (encoder->data->alphaPresent && !firstCell->alphaPlane)) {
            return AVIF_RESULT_INCOMPATIBLE_IMAGE;
        }
    }
```

So I will just add a test to exercise that check.

y-guyon:
Right, I saw that piece of code but `firstCell` made me think it only concerned grids. I should have looked further.
Thanks for investigating.

PR #1274, opened 2023-01-25
Add --target-size flag to avifenc

y-guyon:
The two first commits are avifenc refactors containing necessary changes for the third commit:
- Moving encoding code to a `avifEncodeImages()` that can be called once for each quality to try
- Encapsulating local variables into an `avifSettings` struct to avoid huge function signatures
- Caching input images (especially from `stdin`)

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -456,6 +456,18 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
     return AVIF_TRUE;
 }
 
+// Returns NULL if a memory allocation failed.
+static char * avifStrdup(const char * str)
+{
+    size_t len = strlen(str);
+    char * dup = avifAlloc(len + 1);
+    if (!dup) {
+        return NULL;
+    }
+    memcpy(dup, str, len + 1);
+    return dup;
+}
+
 static avifBool avifCodecSpecificOptionsAdd(avifCodecSpecificOptions * options, const char * keyValue)
 {
     avifBool success = AVIF_FALSE;
@@ -478,7 +490,7 @@ static avifBool avifCodecSpecificOptionsAdd(avifCodecSpecificOptions * options, 
     const char * value = strchr(keyValue, '=');
     if (value) {
         // Remove equals sign.
-        options->values[options->count] = strdup(value);
+        options->values[options->count] = avifStrdup(value + 1);
         const size_t keyLength = strlen(keyValue) - strlen(value);
         options->keys[options->count] = malloc(keyLength + 1);
         if (!options->values[options->count] || !options->keys[options->count]) {
@@ -488,9 +500,9 @@ static avifBool avifCodecSpecificOptionsAdd(avifCodecSpecificOptions * options, 
         options->keys[options->count][keyLength] = '\0';
     } else {
         // Pass in a non-NULL, empty string. Codecs can use the mere existence of a key as a boolean value.
-        options->values[options->count] = strdup("");
-        options->keys[options->count] = strdup(keyValue);
-        if (!options->keys[options->count]) {
+        options->values[options->count] = avifStrdup("");
+        options->keys[options->count] = avifStrdup(keyValue);
+        if (!options->values[options->count] || !options->keys[options->count]) {
             goto cleanup;
         }
     }
@@ -919,8 +931,8 @@ static avifBool avifEncodeImages(avifSettings * settings,
     avifRWDataFree(encoded);
     *encoded = closestEncoded;
     *ioStats = closestIoStats;
-    printf("Kept the encoded image of size %d bytes generated with color quality %d and alpha quality %d.\n",
-           (int)encoded->size,
+    printf("Kept the encoded image of size %" AVIF_FMT_ZU " bytes generated with color quality %d and alpha quality %d.\n",
+           encoded->size,
            settings->quality,
            settings->qualityAlpha);
     return AVIF_TRUE;
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -45,6 +45,13 @@ typedef struct avifInput
     avifBool useStdin;
 } avifInput;
 
+typedef struct
+{
+    char ** keys;
+    char ** values;
+    int count;
+} avifCodecSpecificOptions;
+
 static void syntax(void)
 {
     printf("Syntax: avifenc [options] input.[jpg|jpeg|png|y4m] output.avif\n");
@@ -260,7 +267,7 @@ static avifInputFile * avifInputGetNextFile(avifInput * input)
     }
     return &input->files[input->fileIndex];
 }
-static avifBool avifInputHasRemainingData(avifInput * input)
+static avifBool avifInputHasRemainingData(const avifInput * input)
 {
     if (input->useStdin) {
         return !feof(stdin);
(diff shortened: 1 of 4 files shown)


apps/avifenc.c
```diff
@@ -349,6 +356,52 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
     return AVIF_TRUE;
 }
 
+static avifBool avifCodecSpecificOptionsAdd(avifCodecSpecificOptions * options, const char * keyValue)
```
jzern, line 459: Looks like there's a mix of functions with an `avif` prefix that are static in this file. For this case I guess it was to match the parameter type like with `avifInput...`.

apps/avifenc.c
```diff
@@ -349,6 +356,52 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
...
+    }
+    if (options->count) {
+        memcpy(options->keys, oldKeys, options->count * sizeof(*options->keys));
+        memcpy(options->values, oldValues, options->count * sizeof(*options->values));
+    }
+
+    const char * value = strchr(keyValue, '=');
+    if (value) {
+        // Remove equals sign.
+        options->values[options->count] = strdup(value);
```
jzern, line 481: strdup isn't a part of C99; it's been added to C23. I guess that's a preexisting condition, though.

Did you want `value + 1` here? In the earlier code the `=` was wiped out with `*value = 0`

apps/avifenc.c
```diff
@@ -836,6 +840,92 @@ static avifBool avifEncodeImages(const avifSettings * settings,
...
+        settings->quality = closestQuality;
+    }
+    if (!settings->qualityAlphaIsConstrained) {
+        settings->qualityAlpha = closestQuality;
+    }
+    avifRWDataFree(encoded);
+    *encoded = closestEncoded;
+    *ioStats = closestIoStats;
+    printf("Kept the encoded image of size %d bytes generated with color quality %d and alpha quality %d.\n",
+           (int)encoded->size,
```
jzern, line 923: you could just specify %zu, no?

tests/test_cmd_targetsize.sh
```diff
@@ -0,0 +1,125 @@
...
+INPUT_Y4M="${TESTDATA_DIR}/kodim03_yuv420_8bpc.y4m"
+# Output file names.
+ENCODED_FILE="avif_test_cmd_targetsize_encoded.avif"
+DECODED_SMALLEST_FILE="avif_test_cmd_targetsize_decoded_smallest.png"
+DECODED_BIGGEST_FILE="avif_test_cmd_targetsize_decoded_biggest.png"
+
+# Cleanup
+cleanup() {
+  pushd ${TMP_DIR}
+    rm -- "${ENCODED_FILE}" "${DECODED_SMALLEST_FILE}" "${DECODED_BIGGEST_FILE}"
```
jzern, line 52: Thinking about this, if we were to autoformat the files this indent would be lost. I know it's used as a visual indicator, maybe wrapping the commands in {} could help. Just a thought, no change in this PR, though.

tests/test_cmd_targetsize.sh
```diff
@@ -0,0 +1,125 @@
...
+
+  "${AVIFENC}" -s 8 "${INPUT_Y4M}" -o "${ENCODED_FILE}" --target-size 0
+  "${AVIFDEC}" "${ENCODED_FILE}" "${DECODED_SMALLEST_FILE}"
+  SMALLEST_FILE_SIZE=$(wc -c < "${ENCODED_FILE}")
+
+  "${AVIFENC}" -s 8 "${INPUT_Y4M}" -o "${ENCODED_FILE}" --target-size 999999999
+  "${AVIFDEC}" "${ENCODED_FILE}" "${DECODED_BIGGEST_FILE}"
+  BIGGEST_FILE_SIZE=$(wc -c < "${ENCODED_FILE}")
+
+  test ${SMALLEST_FILE_SIZE} -lt ${DEFAULT_QUALITY_FILE_SIZE} || exit 1
```
jzern, line 69: in bash `[[...]]` would generally be preferred. This is ok if it's consistent with elsewhere.

tests/test_cmd_targetsize.sh
```diff
@@ -0,0 +1,125 @@
...
+  test ${DEFAULT_QUALITY_FILE_SIZE} -lt ${BIGGEST_FILE_SIZE} || exit 1
+
+  # The remaining tests in this file use animations that may trigger segmentation faults
+  # with libaom versions older than 3.6.0. Skip the tests if that is the case.
+  # TODO(yguyon): Investigate.
+  # The grep and cut commands are not pretty but seem to work on most platforms.
+  AOM_VERSION=$("${AVIFENC}" -V | grep -o "aom.*[0-9]*\.")        # "aom [enc/dec]:X.Y."
+  AOM_VERSION=$(echo "${AOM_VERSION}" | cut -d ":" -f 2)          # "X.Y" maybe prepended by a 'v'
+  AOM_VERSION=$(echo "${AOM_VERSION}" | grep -o "[0-9]*\.[0-9]*") # "X.Y"
```
jzern, line 102: These could all be chained together in one set of pipelines. It doesn't make much difference except maybe on windows to cut down on the number of new processes created.

jzern (commented):
No CHANGELOG entry?

apps/avifenc.c
y-guyon: > strdup isn't a part of C99; it's been added to C23. I guess that's a preexisting condition, though.

Copy-pasted `avifStrdup()` from `avif.c`.

> Did you want `value + 1` here? In the earlier code the `=` was wiped out with `*value = 0`

Done.

tests/test_cmd_targetsize.sh
y-guyon: Yes but I do not know how to keep the end-of-line comments and `echo` is used just after anyway.

apps/avifenc.c
```diff
@@ -440,10 +605,339 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
...
+                                    nextImage,
+                                    /*outDepth=*/NULL,
+                                    /*sourceIsRGB=*/NULL,
+                                    /*sourceTiming=*/NULL,
+                                    settings->chromaDownsampling)) {
+                goto cleanup;
+            }
+
+            // Verify that this frame's properties matches the first frame's properties
+            if ((firstImage->width != nextImage->width) || (firstImage->height != nextImage->height)) {
```
jzern, line 779: is there any content that can cover these conditions in the new test script?

apps/avifenc.c
```diff
@@ -440,10 +605,339 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
...
+        }
+        if (!settings->qualityAlphaIsConstrained) {
+            settings->qualityAlpha = quality;
+        }
+
+        if (!avifEncodeImagesFixedQuality(settings, input, firstFile, firstImage, gridCells, encoded, ioStats)) {
+            avifRWDataFree(&closestEncoded);
+            return AVIF_FALSE;
+        }
+        printf("Encoded image of size %d bytes.\n", (int)encoded->size);
```
jzern, line 899: this could be `AVIF_FMT_ZU` to match line 934

jzern (approved):

apps/avifenc.c
y-guyon: I do not see how this is relevant to this PR or to the new test script `test_cmd_targetsize.sh`.
I sent another PR https://github.com/AOMediaCodec/libavif/pull/1297.

PR #1276, opened 2023-01-26
Increment currentLayer only after all uses of it

wantehchang:
codec->internal->currentLayer is used in aomCodecEncodeImage() in four places. Right now we increment codec->internal->currentLayer after the second use, so the third and four uses see a value that is greater than the correct value by one.

Fix this by incrementing codec->internal->currentLayer only after all uses of it, right before we return AVIF_RESULT_OK at the end of aomCodecEncodeImage().


wantehchang:
This is a copy of https://aomedia-review.googlesource.com/c/libavif/+/169524. Please review.

@tongyuantongyu FYI.

y-guyon (approved):

PR #1277, opened 2023-01-27
Don't allow chroma sample position to change

wantehchang:
The color config (colour primaries, transfer characteristics, matrix coefficients, color range, and chroma sample position) is part of the AV1 sequence header OBU, so changes to the color config are not allowed. Check changes to chroma sample position between two aomCodecEncodeImage() calls.

Add a test with changes to chroma sample position and color range.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -1137,6 +1137,7 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
         // Another frame in an image sequence, or layer in a layered image
 
         const avifImage * imageMetadata = encoder->data->imageMetadata;
+        // Image metadata that are copied to the av1C and nclx boxes are not allowed to change.
         // If the first image in the sequence had an alpha plane (even if fully opaque), all
         // subsequent images must have alpha as well.
         if ((imageMetadata->depth != firstCell->depth) || (imageMetadata->yuvFormat != firstCell->yuvFormat) ||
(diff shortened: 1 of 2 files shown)


wantehchang:
@tongyuantongyu FYI.

jzern:
I think this change is fine, but technically I think those values could change if the sequence header was updated in an animation or video.

wantehchang:
James: You are right. We don't want the sequence header to be updated in an animation or video because these values are copied to the 'av1C' box and the 'colr' box of the 'nclx' type in the container.

Note: I analyzed libaom source code. libaom writes a sequence header only when it writes a key frame or an intra only frame. Here is an excerpt of the `av1_pack_bitstream()` function in aom/av1/encoder/bitstream.c:

```
  // write sequence header obu at each key frame or intra_only frame,
  // preceded by 4-byte size
  if (cm->current_frame.frame_type == INTRA_ONLY_FRAME ||
      cm->current_frame.frame_type == KEY_FRAME) {
    obu_header_size = av1_write_obu_header(
        level_params, &cpi->frame_header_count, OBU_SEQUENCE_HEADER, 0, data);
    obu_payload_size =
        av1_write_sequence_header_obu(cm->seq_params, data + obu_header_size);
```

After an `aom_codec_control(AV1E_SET_MATRIX_COEFFICIENTS)` call, libaom doesn't force the next frame to be a key frame. If the next frame is not encoded as a key frame or an intra only frame, libaom won't write an updated sequence header with the next frame. 

So this is another reason for libavif to disallow changes to these values.

y-guyon (approved):

wantehchang (commented):
Yannis: Thank you for the review. I made all the changes you suggested, and also added a comment to explain why we don't allow certain image metadata to change. PTAL.

PR #1278, opened 2023-02-06
Add targetSize in avifEncoder

y-guyon:
`targetSize` is an alternative to `quality` and `qualityAlpha` for setting the intended rate-distortion of the encoded image. It uses a binary search over `quality` and/or `qualityAlpha` to find the settings leading to the encoded image with a size in bytes as close to the given `targetSize` as possible.

It is imprecise because a lower `quality` does not always generate a smaller encoded payload, and because the input granularity is limited by the `quantizer` setting, which has only 64 possible integer values.

The implementation as a `--target_size` flag in `avifenc.c` can be done in another pull request. If it uses the `avifEncoder::targetSize` API though, it will lack the convenient progress notification through stdout that is implemented in https://github.com/AOMediaCodec/libavif/pull/1274, unless the API is extended somehow to handle that (unlikely to be worth it).


src/write.c
```diff
@@ -1219,10 +1224,173 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
+{
+    if (encoder->targetSize == 0) {
+        return avifEncoderAddImageInternal(encoder, gridCols, gridRows, cellImages, durationInTimescales, addImageFlags);
+    }
+
+    // Setting avifEncoder.targetSize for one call to avifEncoderAddImage() without knowing if there will
+    // be another frame is ambiguous. The user may expect the whole animation to match the targetSize, or
+    // may expect that it applies to each frame. Allowing avifEncoder.targetSize>0 for animations would also
+    // make the code more complicated, because all input frames would be stored until avifEncoderFinish().
+    // For simplicity, it is reserved to single-frame images.
```
jzern, line 1296: maybe '...reserved for...' or '...limited/restricted to...'

src/write.c
```diff
@@ -1219,10 +1224,173 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
+
+    int closestQuality = AVIF_QUALITY_DEFAULT;
+    avifRWData closestEncoded = AVIF_DATA_EMPTY;
+    size_t closestSizeDiff = 0;
+    avifIOStats closestIoStats = { 0, 0 };
+
+    int minQuality = AVIF_QUALITY_WORST; // inclusive
+    int maxQuality = AVIF_QUALITY_BEST;  // inclusive
+    while (minQuality <= maxQuality) {
+        const int quality = (minQuality + maxQuality) / 2;
```
jzern, line 1322: webp uses a different approach since it works with a lower number of passes [1]. For both I'm wondering if we could make an early guess based on the image resolution and target size to cut down on the iterations, but it's better to start with this approach.

[1]: https://chromium.googlesource.com/webm/libwebp/+/refs/tags/v1.3.0/src/enc/frame_enc.c#71

tests/gtest/aviftargetsizetest.cc
```diff
@@ -0,0 +1,147 @@
+// Copyright 2022 Google LLC
```
jzern, line 1: 2023

jzern (approved):

include/avif/avif.h
```diff
@@ -1091,6 +1091,11 @@ typedef struct avifScalingMode
 //   * NOTE: Please see the "Understanding maxThreads" comment block above
 // * Quality range: [AVIF_QUALITY_WORST - AVIF_QUALITY_BEST]
 // * Quantizer range: [AVIF_QUANTIZER_BEST_QUALITY - AVIF_QUANTIZER_WORST_QUALITY]
+// * If targetSize is set to a number of bytes strictly greater than 0, multiple values of quality and
+//   qualityAlpha are tested to find the settings leading to an encoded image with a size as close to
+//   targetSize as possible.
+//   Only modifies quality if it was not set. Only modifies qualityAlpha if it was not set.
+//   This feature is only available with AVIF_ADD_IMAGE_FLAG_SINGLE and is 7 times slower.
```
wantehchang, line 1098: I suggest moving this list item after the next list item (which is about the deprecated quantizer fields), and also moving the `targetSize` field after the `maxQuantizerAlpha` field  below. The reason is that the quality and quantizer fields are closely related, so it would be good to describe them with nothing else in between.

src/write.c
```diff
@@ -1219,10 +1224,173 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
+        } else {
+            avifRWDataFree(&encoded);
+        }
+
+        if (sizeDiff == 0) {
+            break;
+        }
+    }
+
+    // Output the used quality directly in the avifEncoder struct.
```
wantehchang, line 1373: I refrained from doing this with the `autoTiling`  feature. That was a somewhat arbitrary decision, to avoid modifying the `avifEncoder` struct that we received from the user. We should modify the `autoTiling` code to also output the used `tileRowsLog2` and `tileColsLogs` directly in the `avifEncoder` struct, and print the used values in avifenc.

wantehchang (commented):
Yannis: You can merge this pull request with James's review. I only had time to check the public API. If I have time later, I will try to review the entire pull request. But no need to wait for my review.

src/write.c
```diff
@@ -231,6 +231,9 @@ typedef struct avifEncoderData
     uint16_t primaryItemID;
     avifBool singleImage; // if true, the AVIF_ADD_IMAGE_FLAG_SINGLE flag was set on the first call to avifEncoderAddImage()
     avifBool alphaPresent;
+    // Cached encoded image from avifEncoderAddImage() when avifEncoder.targetSize>0.
```
wantehchang, line 234: Nit: Not sure if you prefer "x>0", with no spaces around the `>` sign, in the comments. There is another instance in line 1294.

src/write.c
y-guyon: As noted in https://github.com/AOMediaCodec/libavif/pull/1278#issue-1573254420, I am not sure about using the targetSize API in avifenc.c. I think displaying what is going on to stdout is important, and thus I would lean towards using both #1278 and #1274 implementations.

With that said, maybe there is no need to modify `avifEncoder.quality`. An alternative is outputting the final `quality` and `qualityAlpha` through `avifIOStats`. I did not do that at first because it gets confusing when quantizers are given, but it should be possible.

What do you think?

jzern: Outputting the achieved quality might be good. As for progress, another option would be to introduce a callback like webp to indicate progress.

joedrago:
Is this something we really want to expose at the API level, instead of just adding as an `avifenc` feature? Creating an `avifEncoder` inside the guts of an `avifEncoder` call and manually copying over each of its members feels fragile to me, and is signing us up for maintaining this block of code for every future `avifEncoder` value we add. By the implementation provided here, it is pretty clear this is the kind of operation that needs to happen one layer higher than `avifEncoder`. 

Assuming we _do_ want this to be exposed in the API, can we somehow avoid this nested `avifEncoder` creation? I see this comment:
```
    // Duplicating an avifEncoder instance that already encoded something is too involved.
```

I fully believe that there is some auditing that would need to be done to "Reset" an encoder back to having zero frames encoded (throw out items, frames, etc from `avifEncoderData` internals), but we already have cleanup paths in place in place for this kind of thing (`avifEncoderDataDestroy()`), and I imagine we could split a subset of it into a Reset() type of call which would make it ready to "try again" with fresh calls from AddImage/Finish. This would match the Reset pattern that avifDecoder offers as well. 

If we want to put it in the API, we're going to have to pay a maintenance tax _somewhere_ on "preparing a fresh encoder", and I'd rather spend that tax in maintaining the reset of the one `avifEncoder` we have, instead of instantiating a fresh one repeatedly and copying the members over.

y-guyon:
> Is this something we really want to expose at the API level, instead of just adding as an `avifenc` feature?

This was the conclusion of an internal discussion, but it is not set in stone.
The rationale was that implementing it for `avifenc` would be useful to some users, while implementing it directly in the API would benefit a larger set of users, for approximately the same amount of work. In practice, #1278 and #1274 are quite different. See https://github.com/AOMediaCodec/libavif/pull/1278#discussion_r1105551717 for a discussion on whether `avifEnc` should reuse the `targetSize` API or reimplement it.

> Creating an `avifEncoder` inside the guts of an `avifEncoder` call and manually copying over each of its members feels fragile to me, and is signing us up for maintaining this block of code for every future `avifEncoder` value we add.

This is already a concern for `avifEncoderCreate/Destroy()` and `avifEncoderBackupSettings()`, I do not think the added maintenance cost is that high.

> By the implementation provided here, it is pretty clear this is the kind of operation that needs to happen one layer higher than `avifEncoder`.

At first I considered a separate function such as `avifEncoderAddImageTargetSize()` but it would not work for grids.
On top of that, I believe that for the user it makes sense to choose the compression through `quality`, `quantizer` or `targetSize`, and it is clearer if all of these settings are next to each other, hence in `avifEncoder`.
Not accepting `avifEncoder` as input for a `targetSize` API feature means either repeating a lot of the `avifEncoder` fields, or not being able to tune these.

> Assuming we _do_ want this to be exposed in the API, can we somehow avoid this nested `avifEncoder` creation? I see this comment:
> 
> ```
>     // Duplicating an avifEncoder instance that already encoded something is too involved.
> ```
> 
> I fully believe that there is some auditing that would need to be done to "Reset" an encoder back to having zero frames encoded (throw out items, frames, etc from `avifEncoderData` internals), but we already have cleanup paths in place in place for this kind of thing (`avifEncoderDataDestroy()`), and I imagine we could split a subset of it into a Reset() type of call which would make it ready to "try again" with fresh calls from AddImage/Finish. This would match the Reset pattern that avifDecoder offers as well.
> 
> If we want to put it in the API, we're going to have to pay a maintenance tax _somewhere_ on "preparing a fresh encoder", and I'd rather spend that tax in maintaining the reset of the one `avifEncoder` we have, instead of instantiating a fresh one repeatedly and copying the members over.

It will not change anything for the user (except allowing `targetSize` for animations maybe) so I suggest we move forward with this simple solution as is, and feel free to send another PR implementing this `avifEncoder` reset afterwards. What do you think?

src/write.c
y-guyon: > Outputting the achieved quality might be good.

I assume you meant through `avifIOStats`. See #1286.

> As for progress, another option would be to introduce a callback like webp to indicate progress.

This is significantly more work than planned. But it may be necessary to be able to cancel the encoding, for native AVIF support in Photoshop for example.

tongyuantongyu:
Seeing `targetSize` in `avifEncoder` my first impression is something like target bitrate in video/audio encoder, which "magically" controls the output size without too much extra work, and serves as an alternative way to control quality. But here `targetSize` is implemented by doing encoding multiple times, which is significantly more work. This requires users to reconsider if they really want their images in specific size, at the cost of times longer of the already long encoding time. This is especially important when using API, where user is writing automatic process logic to handle lots of images, and cannot easily interrupt the process if they found it taking too long.

For me I would select some desired quality and a rather loose bpp threshold, and only re-encode with lower quality if that threshold is exceed (which should be rare). Other users may have different requirements and come up with their own trade-offs, but I believe "as close to target size as possible" is seldom the choice, other than the case of casually encoding a few images through command line.

I agree with Joe that this should not live in the `avifEncoder`. Maybe we can extract some re-usable parts like this from `avifenc` and `avifdec` into an auxiliary library?

y-guyon:
> I agree with Joe that this should not live in the `avifEncoder`.

Alright, well then unless someone has a strong opinion in favor of having it in the `avifEncoder`, I will abandon this change. I marked #1274 as ready for review.

> Maybe we can extract some re-usable parts like this from `avifenc` and `avifdec` into an auxiliary library?

The [binary search function](https://github.com/AOMediaCodec/libavif/pull/1274/commits/da709082ae75e7b69ddf74fbe7bec23cc3d0fcbe#diff-fd2436a9cdc2b36bc8bbde03350f5bd6e510eaceb44da44510e1425eb7d5b5ddR872-R911) is somewhat short, simple and custom. I doubt it is worth it to move it to `apps/shared/avifutil.h`. The heavy part is the `avifenc` refactoring.
One downside of keeping it in `avifenc` is testing: .sh tests are not convenient.

jzern:
> Alright, well then unless someone has a strong opinion in favor of having it in the `avifEncoder`, I will abandon this change. I marked #1274 as ready for review.

For what it's worth, the benefit would be to integrations that don't have a general solution like Yannis is proposing in avifenc. ImageMagick exposes this for libwebp for instance.

PR #1279, opened 2023-02-07
Check OOM in avifImageCreate, avifEncoderCreate

y-guyon:
The goal is to remove the abort() call from avifAlloc() once all libavif functions check potential out-of-memory errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ avifEncoderSetCodecSpecificOption().
 * Update aom.cmd: v3.6.0
 * Update rav1e.cmd: v0.6.3
 * avifImageCreate(), avifImageCreateEmpty(), avifEncoderCreate() and other
-  internal functions now return NULL in case of AVIF_RESULT_OUT_OF_MEMORY.
+  internal functions now return NULL if a memory allocation failed.
 * avifEncoderSetCodecSpecificOption() now returns avifResult instead of void to
   report memory allocation failures.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 There are incompatible ABI changes in this release. The clli member was added
 to the avifImage struct. The repetitionCount member was added to the avifEncoder
 and avifDecoder structs. The quality and qualityAlpha members were added to the
-avifEncoder struct.
+avifEncoder struct. Check that functions returning pointers do not return NULL
+before accessing those pointers. Check the return value of
+avifEncoderSetCodecSpecificOption().
 
 ### Added
 * Add STATIC library target avif_internal to allow tests to access functions
@@ -45,6 +47,10 @@ avifEncoder struct.
   instances.
 * Update aom.cmd: v3.6.0
 * Update rav1e.cmd: v0.6.3
+* avifImageCreate(), avifImageCreateEmpty(), avifEncoderCreate() and other
+  internal functions now return NULL in case of AVIF_RESULT_OUT_OF_MEMORY.
+* avifEncoderSetCodecSpecificOption() now returns avifResult instead of void to
+  report memory allocation failures.
 
 ## [0.11.1] - 2022-10-19
 
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -418,7 +418,7 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
             uint32_t gridIndex = gridX + (gridY * gridCols);
             avifImage * cellImage = avifImageCreateEmpty();
             if (!cellImage) {
-                fprintf(stderr, "ERROR: Cell creation failed: out-of-memory\n");
+                fprintf(stderr, "ERROR: Cell creation failed: out of memory\n");
                 return AVIF_FALSE;
             }
             gridCells[gridIndex] = cellImage;
@@ -520,7 +520,7 @@ int main(int argc, char * argv[])
     avifChromaDownsampling chromaDownsampling = AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC;
 
     if (!encoder) {
-        fprintf(stderr, "ERROR: Out-of-memory\n");
+        fprintf(stderr, "ERROR: Out of memory\n");
         returnCode = 1;
         goto cleanup;
     }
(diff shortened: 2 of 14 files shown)


include/avif/avif.h
```diff
@@ -539,6 +539,7 @@ typedef struct avifImage
     avifRWData xmp;
 } avifImage;
 
+// If avifImageCreate() or avifImageCreateEmpty() returns NULL, an AVIF_RESULT_OUT_OF_MEMORY error happened.
```
wantehchang, line 542: 1. Similar to my comment in CHANGELOG.md, in comments like this one, it would be clearer to say "an out-of-memory error happened" or "memory allocation failed".

2. Since `avifImageCreate()` has input parameters, it could fail because of invalid arguments. So it would be good to make it clearer that out-of-memory error is not the only reason for `avifImageCreate()` failure. Alternatively, we can document that `avifImageCreate()` faithfully sets the width, height, depth, and yuvFormat members of the new image to the input parameters, and this will become an API contract.

CHANGELOG.md
```diff
@@ -43,6 +45,10 @@ avifEncoder struct.
 * avifImageRGBToYUV() and avifImageYUVToRGB() handle avifImage bit depths 8, 10,
   12 and now also 16. Files read by apps/shared/ can output 16-bit avifImage
   instances.
+* avifImageCreate(), avifImageCreateEmpty(), avifEncoderCreate() and other
+  internal functions now return NULL in case of AVIF_RESULT_OUT_OF_MEMORY.
```
wantehchang, line 49: Since these functions don't report an error code, it is a little confusing to mention the error code `AVIF_RESULT_OUT_OF_MEMORY` here. I suggest we say "in case of out-of-memory error" or "in case of memory allocation error".

apps/avifenc.c
```diff
@@ -417,6 +417,10 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
         for (uint32_t gridX = 0; gridX < gridCols; ++gridX) {
             uint32_t gridIndex = gridX + (gridY * gridCols);
             avifImage * cellImage = avifImageCreateEmpty();
+            if (!cellImage) {
+                fprintf(stderr, "ERROR: Cell creation failed: out-of-memory\n");
```
wantehchang, line 421: Nit: In error messages like this, "out of memory" doesn't need to be hyphenated.
The hyphenated form "out-of-memory" is used as an adjective, such as "out-of-memory error".

wantehchang (changes requested):
Yannis: Thank you very much for writing this CL. I only had time to take a quick look today. I suggest some changes.

src/avif.c
```diff
@@ -134,7 +134,19 @@ void avifImageSetDefaults(avifImage * image)
 
 avifImage * avifImageCreate(uint32_t width, uint32_t height, uint32_t depth, avifPixelFormat yuvFormat)
 {
+    // width and height are checked by avifImageAllocatePlanes().
```
wantehchang, line 137: The callers of `avifImageCreate()` may not call `avifImageAllocatePlanes()`. They may point the `avifImage` directly to existing Y, U, V, alpha buffers. (I wonder if libavif has an example that does this.)

So this function could perform basic validation such as disallowing zero width and height.

UPDATE: I see zero width, height, and depth must be allowed because they are used when creating an empty image (line 160). So perhaps this comment can just say that width and height are checked when they are actually used. Alternatively we can document that this function doesn't validate its input parameters (which is the current behavior).

src/write.c
```diff
@@ -235,11 +235,18 @@ typedef struct avifEncoderData
...
 {
     avifEncoderData * data = (avifEncoderData *)avifAlloc(sizeof(avifEncoderData));
+    if (!data) {
+        return NULL;
+    }
     memset(data, 0, sizeof(avifEncoderData));
     data->imageMetadata = avifImageCreateEmpty();
+    if (!data->imageMetadata) {
+        goto error;
+    }
```
wantehchang, line 249: In `avifEncoderDataDestroy()`, we need to handle a null `data->imageMetadata` now:

```
static void avifEncoderDataDestroy(avifEncoderData * data)
{
    ...
    if (data->imageMetadata) {
        avifImageDestroy(data->imageMetadata);
    }
    ...
}
```

src/write.c
```diff
@@ -415,6 +425,10 @@ avifEncoder * avifEncoderCreate(void)
     encoder->scalingMode = noScaling;
     encoder->data = avifEncoderDataCreate();
     encoder->csOptions = avifCodecSpecificOptionsCreate();
+    if (!encoder->data || !encoder->csOptions) {
+        avifEncoderDestroy(encoder);
+        return NULL;
+    }
```
wantehchang, line 431: `avifEncoderDestroy()` needs to handle a null `encoder->csOptions` or `encoder->data` now:

```
void avifEncoderDestroy(avifEncoder * encoder)
{
    // NOTE: avifCodecSpecificOptionsDestroy() already allows its input to be null,
    // so this if check isn't really necessary. I probably would remove the null pointer
    // check from avifCodecSpecificOptionsDestroy() and check it here.
    if (encoder->csOptions) {
        avifCodecSpecificOptionsDestroy(encoder->csOptions);
    }
    // NOTE: It is tempting to change avifEncoderDataDestroy() to allow a null point.
    // I prefer an explicit null check at call site so that we only allow a null pointer
    // where it is expected. But this doesn't really matter.
    if (encoder->data) {
        avifEncoderDataDestroy(encoder->data);
    }
    avifFree(encoder);
}
```

NOTE: git history showed that the null pointer check in `avifCodecSpecificOptionsDestroy()` was there when Joe first added this function in https://github.com/AOMediaCodec/libavif/commit/e1097bd2. I don't know why he added the null pointer check, because it wasn't necessary and also it wasn't present in most of the destroy functions in libavif.

tests/avifyuv.c
```diff
@@ -150,6 +150,10 @@ int main(int argc, char * argv[])
                         int maxDrift = 0;
 
                         avifImage * image = avifImageCreate(dim, dim, yuvDepth, AVIF_PIXEL_FORMAT_YUV444);
+                        if (!image) {
+                            fprintf(stderr, "Out of memory\n");
```
wantehchang, line 154: Nit: existing error messages are printed to stdout using printf. See line 258:

```
                                        printf("ERROR: Encountered a drift greater than or equal to MAX_DRIFT(%d): %d\n", MAX_DRIFT, drift);
                                        return 1;
```

See also lines 491 and 509.

wantehchang (approved):
LGTM.

tests/gtest/avifimagetest.cc
```diff
@@ -0,0 +1,44 @@
+// Copyright 2022 Google LLC
```
jzern, line 1: 2023, maybe this was a holdover from last year

wantehchang (approved):
LGTM. I found that this pull request hasn't been merged. Yannis, I am not sure if you're waiting for my review. I took Thursday off and was busy on Friday. Sorry. I only reviewed the third commit "Check NULL in destroy functions".

PR #1285, opened 2023-02-15
Add support for threaded YUV to RGB conversion

vigneshvg:
Add a new function avifImageYUVToRGBThreaded which takes a maxThreads parameter and potentially uses up to that many threads to parallelize the conversion.

If there are n jobs, each job works on a portion of the image in parallel (each job will work on an image of size
width x (height / 4)).

Microbenchmarking shows that the function scales almost linearly with the number of threads on larger images (2k or above). For smaller images, performance is either neutral or slightly better.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ to the avifImage struct. The repetitionCount member was added to the avifEncoder
 and avifDecoder structs. The quality and qualityAlpha members were added to the
 avifEncoder struct. Check that functions returning pointers do not return NULL
 before accessing those pointers. Check the return value of
-avifEncoderSetCodecSpecificOption().
+avifEncoderSetCodecSpecificOption(). The maxThreads member was added to the
+avifRGBImage struct.
 
 ### Added
 * Add STATIC library target avif_internal to allow tests to access functions
@@ -26,6 +27,7 @@ avifEncoderSetCodecSpecificOption().
   initialized to the default values.
 * Add the public API function avifImageIsOpaque() in avif.h.
 * Add experimental API for progressive AVIF encoding.
+* Add API for multi-threaded YUV to RGB color conversion.
 
 ### Changed
 * Exif and XMP metadata is exported to PNG and JPEG files by default,
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -665,6 +665,9 @@ typedef struct avifRGBImage
                           // the alpha bits as if they were all 1.
     avifBool alphaPremultiplied; // indicates if RGB value is pre-multiplied by alpha. Default: false
     avifBool isFloat; // indicates if RGBA values are in half float (f16) format. Valid only when depth == 16. Default: false
+    int maxThreads; // Number of threads to be used for the YUV to RGB conversion. Note that this value is ignored for RGB to YUV
+                    // conversion. Setting this to zero has the same effect as setting it to one. Negative values are invalid.
+                    // Default: 1.
 
     uint8_t * pixels;
     uint32_t rowBytes;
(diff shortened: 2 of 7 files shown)


vigneshvg:
Reviewed in gerrit: https://aomedia-review.googlesource.com/c/libavif/+/169477

vigneshvg:
@joedrago This change was something we discussed in email long back. I will bump the email thread as well just in case :)

vigneshvg:
@joedrago This change was something we discussed in email long back. I will bump the email thread as well just in case :)

joedrago (approved):
(The gerrit review looks quite thorough, so I'll just give a general review of the design. I'm Approving this PR, so none of these thoughts require any additional code changes.)

I'm certainly on board with the overall job/threadData design, and it seems pretty easy to follow what is going on. I like the local dynamic array decl usage (a first usage in libavif, but it looks nice). The create/join flow is clear and local.

I don't love `reformat.c` having the `_WIN32` checks and OS level thread work being done in it, but seeing as this is the first and only place libavif has had to dip into threading, I won't push back on it. Asking for more here would be adding a bit more complexity than we need for now, but I hope in the future if we add more threaded work that we migrate/contain this kind of OS-level threading work into a light abstraction in its own file (`thread.c`/`worker.c`/`job.c`/`task.c`, etc). The code doesn't need to do much more than you've already shown here, but keeping the OS level magic in a single separate file helps with portability to other, weirder, non-pthread/non-windows envs. I'd ask the same thing for any other thread sync or kernel objects we want to expose to libavif, such as mutexes or condvars. I'd like to contain any OS-specific stuff into as few places as possible.

vigneshvg:
> (The gerrit review looks quite thorough, so I'll just give a general review of the design. I'm Approving this PR, so none of these thoughts require any additional code changes.)
> 
> I'm certainly on board with the overall job/threadData design, and it seems pretty easy to follow what is going on. I like the local dynamic array decl usage (a first usage in libavif, but it looks nice). The create/join flow is clear and local.
> 

Thanks for the review and approval!

> I don't love `reformat.c` having the `_WIN32` checks and OS level thread work being done in it, but seeing as this is the first and only place libavif has had to dip into threading, I won't push back on it. Asking for more here would be adding a bit more complexity than we need for now, but I hope in the future if we add more threaded work that we migrate/contain this kind of OS-level threading work into a light abstraction in its own file (`thread.c`/`worker.c`/`job.c`/`task.c`, etc). The code doesn't need to do much more than you've already shown here, but keeping the OS level magic in a single separate file helps with portability to other, weirder, non-pthread/non-windows envs. I'd ask the same thing for any other thread sync or kernel objects we want to expose to libavif, such as mutexes or condvars. I'd like to contain any OS-specific stuff into as few places as possible.

Yes agreed. If we add more threading/OS-specific components in libavif, it would be best to aggregate them in a separate file.

wantehchang:
Yes, Yannis (@y-guyon) ran into this problem with YUV420 and bilinear chroma upsampling when he implemented incremental decoding of AVIF grid images in Chrome. I am sorry I didn't think of this problem until this morning.

I suggest we add a test that detects this problem by comparing the results of single-threaded and multithreaded conversions, and turn off multithreading for this case in this pull request.

vigneshvg:
> Yes, Yannis (@y-guyon) ran into this problem with YUV420 and bilinear chroma upsampling when he implemented incremental decoding of AVIF grid images in Chrome. I am sorry I didn't think of this problem until this morning.
> 
> I suggest we add a test that detects this problem by comparing the results of single-threaded and multithreaded conversions, and turn off multithreading for this case in this pull request.

Thanks for pointing this out.

I have updated the code to disable threading for YUV420/bilinear case. I have also added tests to ensure that the single and multithreaded conversion produce the exact same output for all the cases.

Please take another look at this PR.

vigneshvg:
The issue with 420/bilinear is now resolved. This PR is ready for the final round of reviews.

wantehchang (approved):
LGTM.

vigneshvg:
Thanks for the review everyone. I will go ahead and submit this change. If there are any further comments, i will address them in follow-ups.

PR #1290, opened 2023-02-20
Fix avifIOStats at decoding

y-guyon:
Add avifiostatstest.cc.


tests/gtest/avifiostatstest.cc
```diff
@@ -0,0 +1,158 @@
...
+}
+
+//------------------------------------------------------------------------------
+
+TEST(IoStatsTest, ColorObuSize) {
+  avifIOStats io_stats;
+  ASSERT_TRUE(GetIoStatsFromEncode(io_stats, AVIF_QUALITY_DEFAULT,
+                                   AVIF_QUALITY_DEFAULT,
+                                   /*translucent=*/false));
+  EXPECT_GT(io_stats.colorOBUSize, 500u);
```
jzern, line 92: Are tests only being run with libaom? Should we say _NE(..., 0)?

jzern (approved):

PR #1291, opened 2023-02-22
Refactor read.c with avifTileGroup

y-guyon:
Add avifImageCopySamples().
Add AVIF_CHECKRES().
Add AVIF_ENCODER_NO_CHANGE.
Refactor write.c.

This should be easier to add more items besides primary color and auxiliary alpha in the future, such as in https://github.com/AOMediaCodec/libavif/pull/1215.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -763,17 +763,22 @@ error:
     return NULL;
 }
 
+// A group of AVIF tiles, such as a single tile or a grid of multiple tiles.
+typedef struct avifTileGroup
+{
+    unsigned int tileCount;
+    unsigned int decodedTileCount;
+    unsigned int firstTileIndex; // Within avifDecoderData.tiles.
+    avifImageGrid grid;
+} avifTileGroup;
+
 typedef struct avifDecoderData
 {
     avifMeta * meta; // The root-level meta box
     avifTrackArray tracks;
     avifTileArray tiles;
-    unsigned int colorTileCount;
-    unsigned int alphaTileCount;
-    unsigned int decodedColorTileCount;
-    unsigned int decodedAlphaTileCount;
-    avifImageGrid colorGrid;
-    avifImageGrid alphaGrid;
+    avifTileGroup colorTileGroup;
+    avifTileGroup alphaTileGroup;
     avifDecoderSource source;
     uint8_t majorBrand[4];                     // From the file's ftyp, used by AVIF_DECODER_SOURCE_AUTO
     avifDiagnostics * diag;                    // Shallow copy; owned by avifDecoder
@@ -818,8 +823,8 @@ static void avifDecoderDataResetCodec(avifDecoderData * data)
             tile->codec = NULL;
         }
     }
-    data->decodedColorTileCount = 0;
-    data->decodedAlphaTileCount = 0;
+    data->colorTileGroup.decodedTileCount = 0;
+    data->alphaTileGroup.decodedTileCount = 0;
 }
 
 static avifTile * avifDecoderDataCreateTile(avifDecoderData * data, uint32_t width, uint32_t height, uint8_t operatingPoint)
@@ -874,10 +879,10 @@ static void avifDecoderDataClearTiles(avifDecoderData * data)
         }
     }
     data->tiles.count = 0;
-    data->colorTileCount = 0;
-    data->alphaTileCount = 0;
-    data->decodedColorTileCount = 0;
-    data->decodedAlphaTileCount = 0;
+    data->colorTileGroup.tileCount = 0;
+    data->colorTileGroup.decodedTileCount = 0;
+    data->alphaTileGroup.tileCount = 0;
+    data->alphaTileGroup.decodedTileCount = 0;
 }
 
 static void avifDecoderDataDestroy(avifDecoderData * data)
@@ -1284,21 +1289,20 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
 // dstImage. Only the freshly decoded tiles are considered, skipping the already
 // copied or not-yet-decoded tiles.
 static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
-                                             avifImageGrid * grid,
                                              avifImage * dstImage,
-                                             unsigned int firstTileIndex,
+                                             const avifTileGroup * tileGroup,
                                              unsigned int oldDecodedTileCount,
-                                             unsigned int decodedTileCount,
                                              avifBool alpha)
 {
-    assert(decodedTileCount > oldDecodedTileCount);
+    assert(tileGroup->decodedTileCount > oldDecodedTileCount);
+    const avifImageGrid * grid = &tileGroup->grid;
 
-    avifTile * firstTile = &data->tiles.tile[firstTileIndex];
+    avifTile * firstTile = &data->tiles.tile[tileGroup->firstTileIndex];
     avifBool firstTileUVPresent = (firstTile->image->yuvPlanes[AVIF_CHAN_U] && firstTile->image->yuvPlanes[AVIF_CHAN_V]);
 
     // Check for tile consistency: All tiles in a grid image should match in the properties checked below.
-    for (unsigned int i = AVIF_MAX(1, oldDecodedTileCount); i < decodedTileCount; ++i) {
-        avifTile * tile = &data->tiles.tile[firstTileIndex + i];
+    for (unsigned int i = AVIF_MAX(1, oldDecodedTileCount); i < tileGroup->decodedTileCount; ++i) {
+        avifTile * tile = &data->tiles.tile[tileGroup->firstTileIndex + i];
         avifBool uvPresent = (tile->image->yuvPlanes[AVIF_CHAN_U] && tile->image->yuvPlanes[AVIF_CHAN_V]);
         if ((tile->image->width != firstTile->image->width) || (tile->image->height != firstTile->image->height) ||
             (tile->image->depth != firstTile->image->depth) || (tile->image->yuvFormat != firstTile->image->yuvFormat) ||
@@ -1385,11 +1389,11 @@ static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
     avifImageSetDefaults(&dstView);
     for (; rowIndex < grid->rows; ++rowIndex, colIndex = 0) {
         for (; colIndex < grid->columns; ++colIndex, ++tileIndex) {
-            if (tileIndex >= decodedTileCount) {
+            if (tileIndex >= tileGroup->decodedTileCount) {
                 // Tile is not ready yet.
                 return AVIF_TRUE;
             }
-            avifTile * tile = &data->tiles.tile[firstTileIndex + tileIndex];
+            avifTile * tile = &data->tiles.tile[tileGroup->firstTileIndex + tileIndex];
 
             avifCropRect dstViewRect = {
                 firstTile->image->width * colIndex, firstTile->image->height * rowIndex, firstTile->image->width, firstTile->image->height
@@ -3390,6 +3394,29 @@ static avifDecoderItem * avifDecoderDataFindItem(avifDecoderData * data, avifBoo
     return NULL;
 }
 
+static avifResult avifDecoderGenerateImageTiles(avifDecoder * decoder, avifTileGroup * tileGroup, avifDecoderItem * item, avifBool alpha)
+{
+    const uint32_t previousTileCount = decoder->data->tiles.count;
+    if ((tileGroup->grid.rows > 0) && (tileGroup->grid.columns > 0)) {
+        AVIF_CHECKERR(avifDecoderGenerateImageGridTiles(decoder, &tileGroup->grid, item, alpha), AVIF_RESULT_INVALID_IMAGE_GRID);
+    } else {
+        AVIF_CHECKERR(item->size != 0, AVIF_RESULT_NO_AV1_ITEMS_FOUND);
+
+        avifTile * tile = avifDecoderDataCreateTile(decoder->data, item->width, item->height, avifDecoderItemOperatingPoint(item));
+        AVIF_CHECKERR(tile, AVIF_RESULT_OUT_OF_MEMORY);
+        AVIF_CHECKERR(avifCodecDecodeInputFillFromDecoderItem(tile->input,
+                                                              item,
+                                                              decoder->allowProgressive,
+                                                              decoder->imageCountLimit,
+                                                              decoder->io->sizeHint,
+                                                              &decoder->diag),
+                      AVIF_RESULT_BMFF_PARSE_FAILED);
+        tile->input->alpha = alpha;
+    }
+    tileGroup->tileCount = decoder->data->tiles.count - previousTileCount;
+    return AVIF_RESULT_OK;
+}
+
 avifResult avifDecoderReset(avifDecoder * decoder)
 {
     avifDiagnosticsClearError(&decoder->diag);
@@ -3400,8 +3427,8 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         return AVIF_RESULT_OK;
     }
 
-    memset(&data->colorGrid, 0, sizeof(data->colorGrid));
-    memset(&data->alphaGrid, 0, sizeof(data->alphaGrid));
+    memset(&data->colorTileGroup, 0, sizeof(data->colorTileGroup));
+    memset(&data->alphaTileGroup, 0, sizeof(data->alphaTileGroup));
     avifDecoderDataClearTiles(data);
 
     // Prepare / cleanup decoded image state
@@ -3518,7 +3545,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                                                      data->diag)) {
             return AVIF_RESULT_BMFF_PARSE_FAILED;
         }
-        data->colorTileCount = 1;
+        data->colorTileGroup.tileCount = 1;
 
         if (alphaTrack) {
             avifTile * alphaTile = avifDecoderDataCreateTile(data, alphaTrack->width, alphaTrack->height, 0); // No way to set operating point via tracks
@@ -3533,7 +3560,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                 return AVIF_RESULT_BMFF_PARSE_FAILED;
             }
             alphaTile->input->alpha = AVIF_TRUE;
-            data->alphaTileCount = 1;
+            data->alphaTileGroup.tileCount = 1;
         }
 
         // Stash off sample table for future timing information
@@ -3577,7 +3604,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         if (!memcmp(colorItem->type, "grid", 4)) {
             avifROData readData;
             AVIF_CHECKRES(avifDecoderItemRead(colorItem, decoder->io, &readData, 0, 0, data->diag));
-            AVIF_CHECKERR(avifParseImageGridBox(&data->colorGrid,
+            AVIF_CHECKERR(avifParseImageGridBox(&decoder->data->colorTileGroup.grid,
                                                 readData.data,
                                                 readData.size,
                                                 decoder->imageSizeLimit,
@@ -3590,7 +3617,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         if (alphaItem && !memcmp(alphaItem->type, "grid", 4)) {
             avifROData readData;
             AVIF_CHECKRES(avifDecoderItemRead(alphaItem, decoder->io, &readData, 0, 0, data->diag));
-            AVIF_CHECKERR(avifParseImageGridBox(&data->alphaGrid,
+            AVIF_CHECKERR(avifParseImageGridBox(&decoder->data->alphaTileGroup.grid,
                                                 readData.data,
                                                 readData.size,
                                                 decoder->imageSizeLimit,
@@ -3600,10 +3627,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         }
 
         // Find Exif and/or XMP metadata, if any
-        avifResult findResult = avifDecoderFindMetadata(decoder, data->meta, decoder->image, colorItem->id);
-        if (findResult != AVIF_RESULT_OK) {
-            return findResult;
-        }
+        AVIF_CHECKRES(avifDecoderFindMetadata(decoder, data->meta, decoder->image, colorItem->id));
 
         // Set all counts and timing to safe-but-uninteresting values
         decoder->imageIndex = -1;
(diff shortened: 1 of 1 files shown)


src/write.c
```diff
@@ -481,7 +481,6 @@ static void avifEncoderBackupSettings(avifEncoder * encoder)
 static avifBool avifEncoderDetectChanges(const avifEncoder * encoder, avifEncoderChanges * encoderChanges)
 {
     const avifEncoder * lastEncoder = &encoder->data->lastEncoder;
-    *encoderChanges = 0;
```
jzern, line 484: It may be better to leave this and document that the value will always be set meaningfully. This is a minor change could be its own PR to keep the review size down in this one.

wantehchang: `avifEncoderDetectChanges()` follows a common convention that output parameters are set when the function succeeds. In this convention it is unspecified whether output parameters are set when the function fails, and the caller should only use the outputs after a successful call.

So the caller of `avifEncoderDetectChanges()` does not need to initialize the output before calling `avifEncoderDetectChanges()`.

src/read.c
```diff
@@ -1375,72 +1378,34 @@ static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
...
-                        &dstImage->yuvPlanes[AVIF_CHAN_V][(uvColOffset * pixelBytes) + ((uvRowOffset + j) * dstImage->yuvRowBytes[AVIF_CHAN_V])];
-                    memcpy(dstV, srcV, uvRowBytes);
-                }
+            const avifCropRect srcViewRect = { 0, 0, dstViewRect.width, dstViewRect.height };
+            avifImage * srcView = avifImageCreateEmpty();
+            avifImage * dstView = avifImageCreateEmpty();
+            if ((avifImageSetViewRect(dstView, dstImage, &dstViewRect) != AVIF_RESULT_OK) ||
+                (avifImageSetViewRect(srcView, tile->image, &srcViewRect) != AVIF_RESULT_OK) ||
+                (avifImageCopySamples(dstView, srcView, alpha ? AVIF_PLANES_A : AVIF_PLANES_YUV) != AVIF_RESULT_OK)) {
+                assert(AVIF_FALSE);
```
jzern, line 1408: is failure impossible in this function or should this return AVIF_FALSE?

wantehchang:
Yannis: I will review this pull request tomorrow. Sorry about the delay.

src/read.c
```diff
@@ -3389,6 +3354,63 @@ static avifResult avifDecoderFlush(avifDecoder * decoder)
     return AVIF_RESULT_OK;
 }
 
+// Searches for the primary color item or the auxiliary alpha item.
+// Returns the target item if found, or NULL.
+static avifDecoderItem * avifDecoderFindItem(avifDecoder * decoder, avifBool alpha, uint32_t parentItemId)
```
jzern, line 3359: Strictly speaking this is separate from the introduction of avifTileGroup. Adding this on its own would have made this change more digestible.

src/read.c
```diff
@@ -4034,137 +3934,99 @@ avifResult avifDecoderNextImage(avifDecoder * decoder)
...
                 return AVIF_RESULT_DECODE_ALPHA_FAILED;
             }
+            avifImageFreePlanes(decoder->image, AVIF_PLANES_ALL);
 
-            avifImageStealPlanes(decoder->image, srcAlpha, AVIF_PLANES_A);
+            decoder->image->width = src->width;
+            decoder->image->height = src->height;
+            decoder->image->depth = src->depth;
+        }
+#if 1
```
jzern, line 4003: leftover debug code?

include/avif/internal.h
```diff
@@ -29,6 +29,14 @@ extern "C" {
             return ERR;       \
     } while (0)
 
+// Forward any error to the caller now or continue execution.
+#define AVIF_CHECKRES(A)                          \
+    do {                                          \
+        const avifResult avifCheckedResult = (A); \
```
wantehchang, line 35: Nit: avifCheckedResult => checkedResult

Because of libavif's naming convention, `avifCheckedResult` unfortunately looks like a global function or a type.

We can also use a shorter name such as `result__`. The trailing underscores help avoid the -Wshadow warnings.

src/write.c
```diff
@@ -477,7 +477,7 @@ static void avifEncoderBackupSettings(avifEncoder * encoder)
 
 // This function detects changes made on avifEncoder. It returns true on success (i.e., if every
 // change is valid), or false on failure (i.e., if any setting that can't change was changed). It
-// reports detected changes in encoderChanges.
+// reports detected changes in encoderChanges, or sets it to 0 if none.
```
wantehchang, line 480: Nit/optional: Here it is clear that 0 means no detected changes. It would be more important to document that `encoderChanges` is a bitwise-OR of detected changes.

src/write.c
```diff
@@ -886,6 +886,55 @@ static int avifQualityToQuantizer(int quality, int minQuantizer, int maxQuantize
     return quantizer;
 }
 
+// Adds the items for a single cell or a grid of cells.
+// Outputs the mainItemId, being the only cell if there is exactly one, or the grid item for multiple cells.
```
wantehchang, line 890: Nit: Change "mainItem" to "primaryItem" and "Id" to "ID" to be consistent with current names. Change all occurrences.

src/write.c
```diff
@@ -886,6 +886,55 @@ static int avifQualityToQuantizer(int quality, int minQuantizer, int maxQuantize
...
+static avifResult avifEncoderAddImageItems(avifEncoder * encoder,
+                                           uint32_t gridCols,
+                                           uint32_t gridRows,
+                                           uint32_t gridWidth,
+                                           uint32_t gridHeight,
+                                           avifBool alpha,
+                                           uint16_t * mainItemId)
+{
+    const uint32_t cellCount = gridCols * gridRows;
+    const char * infeName = alpha ? "Alpha" : "Color";
```
wantehchang, line 900: Nit: I suggest adding a local variable:

```
    size_t infeNameSize = 6;
```

and use it instead of the expression `strlen(infeName) + 1` below.

If you strongly prefer, you can call the `strlen()` function:

```
    size_t infeNameSize = strlen(infeName) + 1;
```

But I think 6 is clearly correct if it is right after the `const char * infeName = alpha ? "Alpha" : "Color";` line.

src/write.c
```diff
@@ -886,6 +886,55 @@ static int avifQualityToQuantizer(int quality, int minQuantizer, int maxQuantize
...
+
+        if (cellCount > 1) {
+            item->dimgFromID = *mainItemId;
+            item->hiddenImage = AVIF_TRUE;
+        } else {
+            *mainItemId = item->id;
+        }
+    }
+
+    // Returns uint16_t * mainItemId instead of avifEncoderItem * mainItem because
```
wantehchang, line 933: This comment is confusing because there is no variable named `mainItem` in this function.

wantehchang (commented):
Yannis: I am still reviewing this pull request. This pull request contains some independent changes. It would be good to break it into smaller pull requests for easier code review. I don't want to make you do extra work, so I will only describe the independent pieces that I have identified.

I suggest we start with these two:
1. The `AVIF_CHECKRES` macro in internal.h.
2. The changes to write.c.

You can put these two in one pull request. I first publish my comments on these two.

src/write.c
```diff
@@ -1092,52 +1112,17 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
-                        primaryItem->irefType = "prem";
-                        primaryItem->irefToID = item->id;
-                    }
-                }
+            uint16_t alphaItemId;
+            AVIF_CHECKRES(avifEncoderAddImageItems(encoder, gridCols, gridRows, gridWidth, gridHeight, /*alpha=*/AVIF_TRUE, &alphaItemId));
+            avifEncoderItem * alphaItem = avifEncoderDataFindItemByID(encoder->data, alphaItemId);
+            assert(alphaItem);
+            alphaItem->irefToID = colorItemId;
+            alphaItem->irefType = "auxl";
```
wantehchang, line 1120: Nit: I suggest setting `irefToID` and `irefType` in the same order here and at lines 1124-1125. (I don't know which order is "better".)

src/read.c
y-guyon: Failure _should_ be impossible.

src/read.c
y-guyon: I changed the `0` to `1` to verify that the code inside worked properly but forgot to change the `1` back to `0`. Done.

src/write.c
y-guyon: > Nit: Change "mainItem" to "primaryItem" 

If `alpha` is true, this is not the primary item that is returned, but the "top-level" one: `the only cell if there is exactly one, or the grid item for multiple cells`. That is why I named it "main" instead of "primary".

> "Id" to "ID" to be consistent with current names. Change all occurrences.

This is against Google style but it is not even applied in this project. Done.

src/read.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1298.

include/avif/internal.h
```diff
@@ -199,6 +207,10 @@ avifBool avifImageScale(avifImage * image,
                         uint32_t imageDimensionLimit,
                         avifDiagnostics * diag);
 
+// Copies the samples from srcImage to dstImage. dstImage must be allocated.
+// srcImage and dstImage must have the same width, height, depth and subsampling.
```
wantehchang, line 211: Nit: subsampling => yuvFormat

src/read.c
```diff
@@ -1375,72 +1378,34 @@ static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
...
-                    memcpy(dstU, srcU, uvRowBytes);
-
-                    const uint8_t * srcV = &tile->image->yuvPlanes[AVIF_CHAN_V][j * tile->image->yuvRowBytes[AVIF_CHAN_V]];
-                    uint8_t * dstV =
-                        &dstImage->yuvPlanes[AVIF_CHAN_V][(uvColOffset * pixelBytes) + ((uvRowOffset + j) * dstImage->yuvRowBytes[AVIF_CHAN_V])];
-                    memcpy(dstV, srcV, uvRowBytes);
-                }
+            const avifCropRect srcViewRect = { 0, 0, dstViewRect.width, dstViewRect.height };
+            avifImage * srcView = avifImageCreateEmpty();
+            avifImage * dstView = avifImageCreateEmpty();
```
wantehchang, line 1404: `srcView` and `dstView` can be created outside the for loops and reused for all loop iterations.

Need to handle `avifImageCreateEmpty()` failures.

Need to destroy `srcView` and `dstView` when done.

src/scale.c
```diff
@@ -180,3 +182,33 @@ avifBool avifImageScale(avifImage * image,
 }
 
 #endif
+
+avifResult avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
```
wantehchang, line 186: Define `avifImageCopySamples()` in src/avif.c. This function is similar to `avifImageCopy()`, so they should be defined in the same file. This function is unrelated to scaling an image.

src/scale.c
```diff
@@ -180,3 +182,33 @@ avifBool avifImageScale(avifImage * image,
...
+{
+    for (int c = AVIF_CHAN_Y; c <= AVIF_CHAN_A; ++c) {
+        if ((!(planes & AVIF_PLANES_YUV) && (c != AVIF_CHAN_A)) || (!(planes & AVIF_PLANES_A) && (c == AVIF_CHAN_A))) {
+            continue;
+        }
```
wantehchang, line 191: Nit/optional: This if statement (lines 189-191) is a little hard to understand. I seem to remember you used an alternative solution like the following:
```
    int firstPlane = AVIF_CHAN_A;
    int lastPlane = AVIF_CHAN_V;
    if (planes & AVIF_PLANES_YUV) {
        firstPlane = AVIF_CHAN_Y;
    }
    if (planes & AVIF_PLANES_A) {
        lastPlane = AVIF_CHAN_A;
    }
    for (int c = firstPlane; c <= lastPlane; ++c) {
```
But I can't find it in the source tree. Maybe it was deleted or rewritten during the code review of the pull request.

The alternative solution is also a little hard to understand, but it is easier to parse.

src/scale.c
```diff
@@ -180,3 +182,33 @@ avifBool avifImageScale(avifImage * image,
...
+        const uint32_t planeWidth = avifImagePlaneWidth(srcImage, c);
+        const uint32_t planeHeight = avifImagePlaneHeight(srcImage, c);
+        const uint8_t * srcRow = avifImagePlane(srcImage, c);
+        uint8_t * dstRow = avifImagePlane(dstImage, c);
+        const uint32_t srcRowBytes = avifImagePlaneRowBytes(srcImage, c);
+        const uint32_t dstRowBytes = avifImagePlaneRowBytes(dstImage, c);
+        if ((planeWidth != avifImagePlaneWidth(dstImage, c)) || (planeHeight != avifImagePlaneHeight(dstImage, c)) ||
+            (!srcRow != !dstRow) || (srcImage->depth != dstImage->depth)) {
```
wantehchang, line 199: Nit: Test `srcImage->depth != dstImage->depth` outside the for loop. This check is loop invariant.

src/scale.c
```diff
@@ -180,3 +182,33 @@ avifBool avifImageScale(avifImage * image,
...
+            return AVIF_RESULT_INVALID_ARGUMENT;
+        }
+        if (!srcRow) {
+            continue;
+        }
+
+        const size_t planeWidthBytes = planeWidth * (size_t)(avifImageUsesU16(srcImage) ? 2 : 1);
```
wantehchang, line 206: Nit: Similarly the bytes-per-pixel expression is loop invariant and can be saved in a local variable outside the for loop.

Also in `avifImageCopy()` we use this form (left shift):

```
       uint32_t planeWidthBytes = avifImagePlaneWidth(dstImage, plane) << (dstImage->depth > 8);
```

Would be good to standardize on one form (either multiplication or left shift).

wantehchang (commented):
Yannis: The next independent part of this pull request is `avifImageCopySamples()` and the code that uses it. It would be good to split this part to a separate pull request.

Here are my comments on this part.

y-guyon:
> Yannis: The next independent part of this pull request is `avifImageCopySamples()` and the code that uses it. It would be good to split this part to a separate pull request.
> 
> Here are my comments on this part.

Done in https://github.com/AOMediaCodec/libavif/pull/1300.

src/read.c
```diff
@@ -3390,6 +3394,29 @@ static avifDecoderItem * avifDecoderDataFindItem(avifDecoderData * data, avifBoo
     return NULL;
 }
 
+static avifResult avifDecoderGenerateImageTiles(avifDecoder * decoder, avifTileGroup * tileGroup, avifDecoderItem * item, avifBool alpha)
```
jzern, line 3397: The parameter order is consistent with the file, I just noticed there's no this/in/out ordering that's common.

src/read.c
```diff
@@ -3958,11 +3923,56 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder,
...
+            return AVIF_RESULT_DECODE_ALPHA_FAILED;
+        }
+        avifImageFreePlanes(decoder->image, AVIF_PLANES_ALL);
+
+        decoder->image->width = src->width;
+        decoder->image->height = src->height;
+        decoder->image->depth = src->depth;
+    }
+#if 0
+    // This code is currently unnecessary as the CICP is always set by the end of avifDecoderParse().
```
jzern, line 3964: is that going to change or instead of the code should just a note be left here?

jzern (approved):

src/read.c
```diff
@@ -763,17 +763,22 @@ static avifDecoderItem * avifMetaFindItem(avifMeta * meta, uint32_t itemID)
     return NULL;
 }
 
+// A group of AVIF tiles, such as a single tile or a grid of multiple tiles.
+typedef struct avifTileGroup
```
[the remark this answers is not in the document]
wantehchang: Correction: The last sentence should read:

> Perhaps we can call this struct `avifItemTiles`, which means the tiles of an (image) item.

src/read.c
y-guyon: This code was only moved in this PR, not changed.

src/read.c
```diff
@@ -1284,21 +1289,20 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
 // dstImage. Only the freshly decoded tiles are considered, skipping the already
 // copied or not-yet-decoded tiles.
 static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
-                                             avifImageGrid * grid,
                                              avifImage * dstImage,
-                                             unsigned int firstTileIndex,
+                                             const avifTileGroup * tileGroup,
```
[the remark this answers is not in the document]
y-guyon: I do not follow. The name of the function describes its effect, not the arguments. I do not see a cost here. A shorter signature is even a benefit to me.

src/read.c
```diff
@@ -763,17 +763,22 @@ static avifDecoderItem * avifMetaFindItem(avifMeta * meta, uint32_t itemID)
     return NULL;
 }
 
+// A group of AVIF tiles, such as a single tile or a grid of multiple tiles.
```
wantehchang, line 766: Add "in an image item" to the comment.

src/read.c
```diff
@@ -1284,21 +1289,20 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
 // dstImage. Only the freshly decoded tiles are considered, skipping the already
 // copied or not-yet-decoded tiles.
 static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
-                                             avifImageGrid * grid,
                                              avifImage * dstImage,
-                                             unsigned int firstTileIndex,
+                                             const avifItemTiles * group,
```
wantehchang, line 1293: Nit: `group` => `itemTiles` or `imageTiles` or `tiles`

Also fix other occurrences.

src/read.c
```diff
@@ -3390,6 +3394,29 @@ static avifDecoderItem * avifDecoderDataFindItem(avifDecoderData * data, avifBoo
     return NULL;
 }
 
+static avifResult avifDecoderGenerateImageTiles(avifDecoder * decoder, avifItemTiles * group, avifDecoderItem * item, avifBool alpha)
```
wantehchang, line 3397: Optional: Hmm... Given this function name, perhaps the first struct name I suggested, `avifImageTiles`, isn't bad after all.

The reason I thought `avifImageTiles` was a bad name is that an `avifImage` may consist of two HEIF image items (color and alpha). The name `avifImageTiles` could refer to an `avifImage` or a HEIF image item. So it could be ambiguous.

src/read.c
```diff
@@ -3617,36 +3641,14 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
-
+        AVIF_CHECKRES(avifDecoderGenerateImageTiles(decoder, &data->color, colorItem, /*alpha=*/AVIF_FALSE));
+        if ((data->color.grid.rows == 0) || (data->color.grid.columns == 0)) {
             if (colorItem->progressive) {
                 decoder->progressiveState = AVIF_PROGRESSIVE_STATE_AVAILABLE;
+                const avifTile * colorTile = &data->tiles.tile[0];
                 if (colorTile->input->samples.count > 1) {
                     decoder->progressiveState = AVIF_PROGRESSIVE_STATE_ACTIVE;
-                    decoder->imageCount = colorTile->input->samples.count;
+                    decoder->imageCount = (int)colorTile->input->samples.count;
```
wantehchang, line 3651: Nit: I usually only add this kind of cast if the compiler warns. If you prefer this cast, please also add a cast to line 3571.

src/read.c
```diff
@@ -3958,8 +3923,53 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder,
             }
         }
 
-        ++*decodedTileCount;
+        ++group->decodedTileCount;
+    }
+    return AVIF_RESULT_OK;
+}
+
+avifResult avifDecoderOutputDecodedTiles(avifDecoder * decoder, avifItemTiles * group, avifBool alpha, unsigned int oldDecodedTileCount)
```
wantehchang, line 3931: Add `const` to `avifItemTiles * group`

src/read.c
```diff
@@ -3958,8 +3923,53 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder,
...
+        assert(group->tileCount == (group->grid.rows * group->grid.columns));
+        AVIF_CHECKERR(avifDecoderDataFillImageGrid(decoder->data, decoder->image, group, oldDecodedTileCount, alpha),
+                      AVIF_RESULT_INVALID_IMAGE_GRID);
+        return AVIF_RESULT_OK;
+    }
+
+    // Normal (most common) non-grid path.
+    assert(group->tileCount == 1);
+    avifImage * src = decoder->data->tiles.tile[group->firstTileIndex].image;
+    const avifPlanesFlag planes = alpha ? AVIF_PLANES_A : AVIF_PLANES_YUV;
```
wantehchang, line 3951: Move this to line 3972. This variable is only used there.

src/read.c
```diff
@@ -3958,8 +3923,53 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder,
...
+    if ((decoder->image->width != src->width) || (decoder->image->height != src->height) || (decoder->image->depth != src->depth)) {
+        if (alpha) {
+            avifDiagnosticsPrintf(&decoder->diag, "decoder->image does not match srcAlpha in width, height, or bit depth");
```
wantehchang, line 3954: Nit: In the diagnostic message, `srcAlpha` needs to be updated for the new code. `srcAlpha` is a variable name in the original code.

src/read.c
```diff
@@ -3976,128 +3986,48 @@ avifResult avifDecoderNextImage(avifDecoder * decoder)
...
-        (decoder->data->decodedAlphaTileCount != decoder->data->alphaTileCount)) {
+    if ((decoder->data->color.decodedTileCount != decoder->data->color.tileCount) ||
+        (decoder->data->alpha.decodedTileCount != decoder->data->alpha.tileCount)) {
         assert(decoder->allowIncremental);
         // The image is not completely decoded. There should be no error unrelated to missing bytes,
         // and at least some missing bytes.
-        assert((prepareColorTileResult == AVIF_RESULT_OK) || (prepareColorTileResult == AVIF_RESULT_WAITING_ON_IO));
-        assert((prepareAlphaTileResult == AVIF_RESULT_OK) || (prepareAlphaTileResult == AVIF_RESULT_WAITING_ON_IO));
-        assert((prepareColorTileResult != AVIF_RESULT_OK) || (prepareAlphaTileResult != AVIF_RESULT_OK));
+        assert((prepareColorTileResult == AVIF_RESULT_WAITING_ON_IO) || (prepareAlphaTileResult == AVIF_RESULT_WAITING_ON_IO));
```
wantehchang, line 4030: The new assertion seems different from the old assertions. Please confirm the new assertion is correct. (I didn't try to understand this change.)

wantehchang (approved):
LGTM. I checked the code carefully this time.

src/read.c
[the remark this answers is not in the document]
wantehchang: This isn't that important. I was just stating my first impression when I had only skimmed the diffs quickly. What I saw was:

Before the change:
```
static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
                                             avifImageGrid * grid,
```

Notice how the function name "avifDecoderDataFillImageGrid" matches the types of the first two parameters. This matching is gone after the change.

Now, looking at this more, I think the original matching was actually misleading, because the "FillImageGrid" part of the function name actually refers to "fill `avifImage * dstImage`" rather than "fill `avifImageGrid * grid`".

src/read.c
y-guyon: Done. At first I avoided that because it makes the order `tiles->grid` look weird. I also considered `item->grid` but it conflicts with `avifDecoderItem`.

src/read.c
y-guyon: I had the warning in my IDE (CLion).

src/read.c
y-guyon: It is.

src/read.c
```diff
@@ -3958,8 +3923,53 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder,
...
+
+    // Grid path.
+    if ((tiles->grid.rows > 0) && (tiles->grid.columns > 0)) {
+        assert(tiles->tileCount == (tiles->grid.rows * tiles->grid.columns));
+        AVIF_CHECKERR(avifDecoderDataFillImageGrid(decoder->data, decoder->image, tiles, oldDecodedTileCount, alpha),
+                      AVIF_RESULT_INVALID_IMAGE_GRID);
+        return AVIF_RESULT_OK;
+    }
+
+    // Normal (most common) non-grid path.
```
wantehchang, line 3948: Nit/optional: Should we preserve the full comment from the original code?

```
// Normal (most common) non-grid path. Just steal the planes from the only "tile".
```

Either is fine by me. Just wanted to confirm you meant to delete the second sentence in the original comment.

wantehchang (approved):
LGTM!

src/read.c
wantehchang: Do you know why it didn't warn about line 3571 (which is exactly the same)? Does the IDE only check the code that was modified?

src/read.c
wantehchang: I am also not very happy with `tiles`. It's hard to come up with a good, short name for this parameter.

wantehchang: One possibility is to name the struct `avifTileInfo` or `avifItemTileInfo`, and name the function parameter `info`.

src/read.c
```diff
@@ -4242,23 +4172,19 @@ uint32_t avifDecoderNearestKeyframe(const avifDecoder * decoder, uint32_t frameI
...
+    if (info->decodedTileCount == 0) {
         return 0;
     }
 
-    if ((grid->rows > 0) && (grid->columns > 0)) {
-        // Grid of AVIF tiles (not to be confused with AV1 tiles).
-        const uint32_t tileHeight = decoder->data->tiles.tile[firstTileIndex].height;
-        return AVIF_MIN((decodedTileCount / grid->columns) * tileHeight, decoder->image->height);
+    if ((info->grid.rows > 0) && (info->grid.columns > 0)) {
+        // Grid of AVIF info (not to be confused with AV1 tiles).
```
wantehchang, line 4185: AVIF info => AVIF tiles

wantehchang (approved):
LGTM. There is a find-and-replace error. Please correct that and merge this pull request. Thank you.

wantehchang:
Yannis: Are you planning to merge this pull request?

wantehchang (approved):
LGTM.

PR #1292, opened 2023-02-25
Ext: Update to dav1d 1.1.0

wantehchang:


y-guyon (approved):

jzern:
Weren't there error reports with images decoded with this version? Should it be reverted?

wantehchang:
There are meson build system changes in this version. If an application builds dav1d using an alternative build system, the meson build changes need to be ported to the alternative build system. Here is an example for Chrome's GN build system: https://chromium-review.googlesource.com/c/chromium/src/+/4288936. I also outlined how to do that for Google's internal repository (cl number 512179002).

Since libavif builds dav1d using the official meson build system, there is no problem.

PR #1294, opened 2023-02-28
Refactor write.c with avifEncoderAddImageItems()

y-guyon:
Add AVIF_CHECKRES().

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -886,6 +886,56 @@ static int avifQualityToQuantizer(int quality, int minQuantizer, int maxQuantize
     return quantizer;
 }
 
+// Adds the items for a single cell or a grid of cells.
+// Outputs the mainItemID, being the only cell if there is exactly one, or the grid item for multiple cells.
+static avifResult avifEncoderAddImageItems(avifEncoder * encoder,
+                                           uint32_t gridCols,
+                                           uint32_t gridRows,
+                                           uint32_t gridWidth,
+                                           uint32_t gridHeight,
+                                           avifBool alpha,
+                                           uint16_t * mainItemID)
+{
+    const uint32_t cellCount = gridCols * gridRows;
+    const char * infeName = alpha ? "Alpha" : "Color";
+    const size_t infeNameSize = 6;
+
+    if (cellCount > 1) {
+        avifEncoderItem * gridItem = avifEncoderDataCreateItem(encoder->data, "grid", infeName, infeNameSize, 0);
+        avifWriteGridPayload(&gridItem->metadataPayload, gridCols, gridRows, gridWidth, gridHeight);
+        gridItem->alpha = alpha;
+        gridItem->gridCols = gridCols;
+        gridItem->gridRows = gridRows;
+        gridItem->gridWidth = gridWidth;
+        gridItem->gridHeight = gridHeight;
+        *mainItemID = gridItem->id;
+    }
+
+    for (uint32_t cellIndex = 0; cellIndex < cellCount; ++cellIndex) {
+        avifEncoderItem * item = avifEncoderDataCreateItem(encoder->data, "av01", infeName, infeNameSize, cellIndex);
+        item->codec = avifCodecCreate(encoder->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
+        if (!item->codec) {
+            // Just bail out early, we're not surviving this function without an encoder compiled in
+            return AVIF_RESULT_NO_CODEC_AVAILABLE;
+        }
+        item->codec->csOptions = encoder->csOptions;
+        item->codec->diag = &encoder->diag;
+        item->alpha = alpha;
+        item->extraLayerCount = encoder->extraLayerCount;
+
+        if (cellCount > 1) {
+            item->dimgFromID = *mainItemID;
+            item->hiddenImage = AVIF_TRUE;
+        } else {
+            *mainItemID = item->id;
+        }
+    }
+
+    // The mainItemID output argument has the type uint16_t* instead of avifEncoderItem*
+    // because the pointer may be invalidated by a call to avifEncoderDataCreateItem().
+    return AVIF_RESULT_OK;
+}
+
 static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
                                               uint32_t gridCols,
                                               uint32_t gridRows,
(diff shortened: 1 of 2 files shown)


src/write.c
```diff
@@ -886,6 +886,56 @@ static int avifQualityToQuantizer(int quality, int minQuantizer, int maxQuantize
...
+        if (cellCount > 1) {
+            item->dimgFromID = *mainItemID;
+            item->hiddenImage = AVIF_TRUE;
+        } else {
+            *mainItemID = item->id;
+        }
+    }
+
+    // The mainItemID output argument has the type uint16_t* instead of avifEncoderItem*
+    // because the pointer may be invalidated by a call to avifEncoderDataCreateItem().
```
wantehchang, line 935: "the pointer" is ambiguous because both `uint16_t*` and `avifEncoderItem*` are pointers.

I guess we can change "the pointer" to "the avifEncoderItem pointer".

src/write.c
```diff
@@ -886,6 +886,56 @@ static int avifQualityToQuantizer(int quality, int minQuantizer, int maxQuantize
...
+
+        if (cellCount > 1) {
+            item->dimgFromID = *mainItemID;
+            item->hiddenImage = AVIF_TRUE;
+        } else {
+            *mainItemID = item->id;
+        }
+    }
+
+    // The mainItemID output argument has the type uint16_t* instead of avifEncoderItem*
```
wantehchang, line 934: (1) I suggest we move this comment before the function as a Note.

(2) Nit: the second type should be `avifEncoderItem**`, with two asterisks, right?

wantehchang (approved):
LGTM. Good sharing of code, at the cost of an extra `avifEncoderDataFindItemByID()` call if alpha is present (to find `alphaItem`). I think this is a good trade-off.

I suggest some comment changes.

src/write.c
```diff
@@ -886,6 +886,56 @@ static int avifQualityToQuantizer(int quality, int minQuantizer, int maxQuantize
...
+        gridItem->alpha = alpha;
+        gridItem->gridCols = gridCols;
+        gridItem->gridRows = gridRows;
+        gridItem->gridWidth = gridWidth;
+        gridItem->gridHeight = gridHeight;
+        *mainItemID = gridItem->id;
+    }
+
+    for (uint32_t cellIndex = 0; cellIndex < cellCount; ++cellIndex) {
+        avifEncoderItem * item = avifEncoderDataCreateItem(encoder->data, "av01", infeName, infeNameSize, cellIndex);
```
jzern, line 915: this return should be checked

src/write.c
```diff
@@ -886,6 +886,56 @@ static int avifQualityToQuantizer(int quality, int minQuantizer, int maxQuantize
...
+        item->codec = avifCodecCreate(encoder->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
+        if (!item->codec) {
+            // Just bail out early, we're not surviving this function without an encoder compiled in
```
jzern, line 918: and `item` cleaned up here

src/write.c
jzern: I see this is a pre-existing condition, so it could wait for a follow up

jzern (approved):

src/write.c
y-guyon: I do not think this is necessary: `avifEncoderData.items` is properly cleaned up in `avifEncoderDataDestroy()`.

src/write.c
```diff
@@ -908,31 +910,27 @@ static avifResult avifEncoderAddImageItems(avifEncoder * encoder,
...
     for (uint32_t cellIndex = 0; cellIndex < cellCount; ++cellIndex) {
         avifEncoderItem * item = avifEncoderDataCreateItem(encoder->data, "av01", infeName, infeNameSize, cellIndex);
+        AVIF_CHECKERR(item, AVIF_RESULT_OUT_OF_MEMORY);
         item->codec = avifCodecCreate(encoder->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
-        if (!item->codec) {
-            // Just bail out early, we're not surviving this function without an encoder compiled in
-            return AVIF_RESULT_NO_CODEC_AVAILABLE;
-        }
+        // Just bail out early, we're not surviving this function without an encoder compiled in
+        AVIF_CHECKERR(item->codec, AVIF_RESULT_NO_CODEC_AVAILABLE);
```
wantehchang, line 921: This comment was originally inside the if statement and was written for that location. The new code is equivalen to the following:

```
        // Just bail out early, we're not surviving this function without an encoder compiled in
        if (!item->codec) {
        }
```

In the new location, the comment needs to describe the condition for bailing out early. For example,

```
        // Just bail out early if no encoder is available, we're not surviving this function without an encoder compiled in
        AVIF_CHECKERR(item->codec, AVIF_RESULT_NO_CODEC_AVAILABLE);
```

Or we can revert to the original code. We can also just delete the comment.

src/write.c
```diff
@@ -908,31 +910,27 @@ static avifResult avifEncoderAddImageItems(avifEncoder * encoder,
         gridItem->gridRows = gridRows;
         gridItem->gridWidth = gridWidth;
         gridItem->gridHeight = gridHeight;
-        *mainItemID = gridItem->id;
+        *topLevelItemID = gridItem->id;
     }
 
     for (uint32_t cellIndex = 0; cellIndex < cellCount; ++cellIndex) {
         avifEncoderItem * item = avifEncoderDataCreateItem(encoder->data, "av01", infeName, infeNameSize, cellIndex);
+        AVIF_CHECKERR(item, AVIF_RESULT_OUT_OF_MEMORY);
```
wantehchang, line 918: Nit/optional: For something so simple, we don't need to use a macro:

```
        if (!item) {
        }
```

y-guyon: Do you prefer not to use macros? To me they provide the following benefits:
- Shorter code, easier to read
- You can put a breakpoint into it to speed up debugging, if used everywhere

At the expense of indirection, but I believe most people are used to this kind of check-ok-or-return pattern nowadays.

src/write.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1299.

src/write.c
wantehchang: Thank you. I checked the src/read.c file and found several similar uses of this macro.

src/write.c
```diff
@@ -1092,52 +1111,17 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
-                    if (encoder->data->imageMetadata->alphaPremultiplied) {
-                        avifEncoderItem * primaryItem = avifEncoderDataFindItemByID(encoder->data, encoder->data->primaryItemID);
-                        assert(primaryItem);
-                        primaryItem->irefType = "prem";
-                        primaryItem->irefToID = item->id;
-                    }
-                }
+            uint16_t alphaItemID;
+            AVIF_CHECKRES(avifEncoderAddImageItems(encoder, gridCols, gridRows, gridWidth, gridHeight, /*alpha=*/AVIF_TRUE, &alphaItemID));
+            avifEncoderItem * alphaItem = avifEncoderDataFindItemByID(encoder->data, alphaItemID);
```
wantehchang, line 1116: Just a note: I think it is possible to avoid this extra `avifEncoderDataFindItemByID()` call by moving lines 1118-1119 into the `avifEncoderAddImageItems()` function. But it will require a fixed order of the two `avifEncoderAddImageItems()` calls -- first color, then alpha -- so that `encoder->data->primaryItemID` has the correct value during the call for alpha.

PR #1297, opened 2023-03-01
Add test_cmd_animation.sh

y-guyon:


tests/test_cmd_animation.sh
```diff
@@ -0,0 +1,73 @@
...
+if [[ "$#" -ge 2 ]]; then
+  TESTDATA_DIR="$(eval echo "$2")"
+else
+  TESTDATA_DIR="$(pwd)/data"
+fi
+if [[ "$#" -ge 3 ]]; then
+  TMP_DIR="$(eval echo "$3")"
+else
+  TMP_DIR="$(mktemp -d)"
+fi
```
jzern, line 33: This could be a candidate for a shared function between the test scripts, but there's not that much duplication right now.

jzern (approved):

PR #1298, opened 2023-03-01
Add avifDecoderFindItem() in read.c

y-guyon:
Refactor read.c.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3389,6 +3389,41 @@ static avifResult avifDecoderFlush(avifDecoder * decoder)
     return AVIF_RESULT_OK;
 }
 
+// Searches for the primary color item or the auxiliary alpha item.
+// Returns the target item if found, or NULL.
+static avifDecoderItem * avifDecoderFindItem(avifDecoder * decoder, avifBool alpha, uint32_t parentItemID)
+{
+    for (uint32_t itemIndex = 0; itemIndex < decoder->data->meta->items.count; ++itemIndex) {
+        avifDecoderItem * item = &decoder->data->meta->items.item[itemIndex];
+        if (!item->size) {
+            continue;
+        }
+        if (item->hasUnsupportedEssentialProperty) {
+            // An essential property isn't supported by libavif; ignore the item.
+            continue;
+        }
+        if (memcmp(item->type, "av01", 4) && memcmp(item->type, "grid", 4)) {
+            // Probably exif or some other data.
+            continue;
+        }
+        if (item->thumbnailForID != 0) {
+            // It's a thumbnail, skip it.
+            continue;
+        }
+        if (!alpha && item->id == decoder->data->meta->primaryItemID) {
+            return item;
+        }
+        if (alpha && (item->auxForID == parentItemID)) {
+            // Is this an alpha auxiliary item of the parent item?
+            const avifProperty * auxCProp = avifPropertyArrayFind(&item->properties, "auxC");
+            if (auxCProp && isAlphaURN(auxCProp->u.auxC.auxType)) {
+                return item;
+            }
+        }
+    }
+    return NULL;
+}
+
 avifResult avifDecoderReset(avifDecoder * decoder)
 {
     avifDiagnosticsClearError(&decoder->diag);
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -3389,6 +3389,41 @@ static avifResult avifDecoderFlush(avifDecoder * decoder)
     return AVIF_RESULT_OK;
 }
 
+// Searches for the primary color item or the auxiliary alpha item.
```
wantehchang, line 3392: Document  `parentItemID`. Mention that `parentItemID` is ignored if `alpha` is false.

src/read.c
```diff
@@ -3389,6 +3389,41 @@ static avifResult avifDecoderFlush(avifDecoder * decoder)
...
+// Returns the target item if found, or NULL.
+static avifDecoderItem * avifDecoderFindItem(avifDecoder * decoder, avifBool alpha, uint32_t parentItemID)
```
wantehchang, line 3394: Optional: This function can take `avifDecoderData * data` as the first parameter and be renamed `avifDecoderDataFindItem`.

src/read.c
```diff
@@ -3389,6 +3389,41 @@ static avifResult avifDecoderFlush(avifDecoder * decoder)
...
+        }
+        if (item->hasUnsupportedEssentialProperty) {
+            // An essential property isn't supported by libavif; ignore the item.
+            continue;
+        }
+        if (memcmp(item->type, "av01", 4) && memcmp(item->type, "grid", 4)) {
+            // Probably exif or some other data.
+            continue;
+        }
+        if (item->thumbnailForID != 0) {
```
wantehchang, line 3409: Should we add `!alpha &&` to this conditional expression? The original code doesn't check this for alpha.

src/read.c
```diff
@@ -3389,6 +3389,41 @@ static avifResult avifDecoderFlush(avifDecoder * decoder)
...
+            // It's a thumbnail, skip it.
+            continue;
+        }
+        if (!alpha && item->id == decoder->data->meta->primaryItemID) {
```
wantehchang, line 3413: Nit: parenthesize `item->id == decoder->data->meta->primaryItemID` to match the style of line 3416.

src/read.c
```diff
@@ -3389,6 +3389,41 @@ static avifResult avifDecoderFlush(avifDecoder * decoder)
...
+            return item;
+        }
+        if (alpha && (item->auxForID == parentItemID)) {
+            // Is this an alpha auxiliary item of the parent item?
```
wantehchang, line 3417: Nit: parent item => parent color item ?

src/read.c
```diff
@@ -3561,103 +3596,45 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
-                                           data->diag)) {
-                    return AVIF_RESULT_INVALID_IMAGE_GRID;
-                }
-            }
-
-            colorItem = item;
-            break;
-        }
-
+        // Primary color item
```
wantehchang, line 3605: Optional: In the new code, this comment (and the comment at line 3613) can be omitted.

The original code needs a comment to summarize a long for loop. The new code is short and self-explanatory.

src/read.c
```diff
@@ -3561,103 +3596,45 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
-                alphaItem = item;
-                break;
-            }
+        // Alpha auxiliary item
+        avifDecoderItem * alphaItem = avifDecoderFindItem(decoder, /*alpha=*/AVIF_TRUE, /*parentItemID=*/colorItem->id);
+
+        // Grids
+        if (!memcmp(colorItem->type, "grid", 4)) {
+            avifROData readData;
+            AVIF_CHECKRES(avifDecoderItemRead(colorItem, decoder->io, &readData, 0, 0, decoder->data->diag));
```
wantehchang, line 3619: Nit: decoder->data => data

src/read.c
```diff
@@ -3561,103 +3596,45 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
+                                                readData.data,
+                                                readData.size,
+                                                decoder->imageSizeLimit,
+                                                decoder->imageDimensionLimit,
+                                                data->diag),
+                          AVIF_RESULT_INVALID_IMAGE_GRID);
+        }
+        if (alphaItem && !memcmp(alphaItem->type, "grid", 4)) {
+            avifROData readData;
+            AVIF_CHECKRES(avifDecoderItemRead(alphaItem, decoder->io, &readData, 0, 0, decoder->data->diag));
```
wantehchang, line 3630: Nit: decoder->data => data

src/read.c
```diff
@@ -3561,103 +3596,45 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
-                }
-
-                alphaItem = item;
-                break;
-            }
+        // Alpha auxiliary item
+        avifDecoderItem * alphaItem = avifDecoderFindItem(decoder, /*alpha=*/AVIF_TRUE, /*parentItemID=*/colorItem->id);
+
+        // Grids
+        if (!memcmp(colorItem->type, "grid", 4)) {
```
wantehchang, line 3617: Optional: We could also move this forward to line 3611 so that we first handle color and then handle alpha. If you prefer the current organization (color, alpha, grids), that's fine too.

wantehchang (approved):
LGTM. I suggest some changes.

src/read.c
y-guyon: I do not see why a thumbnail would be ok for alpha but not for color.

jzern (commented):
lgtm

PR #1300, opened 2023-03-02
Add avifImageCopySamples() to internal.h

y-guyon:
Remove testutil::CopyImageSamples().

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -89,6 +89,10 @@ void avifImageSetDefaults(avifImage * image);
 // Copies all fields that do not need to be freed/allocated from srcImage to dstImage.
 void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage);
 
+// Copies the samples from srcImage to dstImage. dstImage must be allocated.
+// srcImage and dstImage must have the same width, height, depth and yuvFormat.
+avifResult avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes);
+
 typedef struct avifAlphaParams
 {
     uint32_t width;
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -90,8 +90,8 @@ void avifImageSetDefaults(avifImage * image);
 void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage);
 
 // Copies the samples from srcImage to dstImage. dstImage must be allocated.
-// srcImage and dstImage must have the same width, height, depth and yuvFormat.
-avifResult avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes);
+// srcImage and dstImage must have the same width, height, depth, yuvFormat and yuvRange.
+void avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes);
 
 typedef struct avifAlphaParams
 {
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -90,7 +90,8 @@ void avifImageSetDefaults(avifImage * image);
 void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage);
 
 // Copies the samples from srcImage to dstImage. dstImage must be allocated.
-// srcImage and dstImage must have the same width, height, depth, yuvFormat and yuvRange.
+// srcImage and dstImage must have the same width, height, and depth.
+// If the AVIF_PLANES_YUV bit is set in planes, then srcImage and dstImage must have the same yuvFormat and yuvRange.
 void avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes);
 
 typedef struct avifAlphaParams
--- a/src/avif.c
+++ b/src/avif.c
@@ -184,8 +184,10 @@ void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage)
 
 void avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
 {
-    assert((srcImage->width == dstImage->width) && (srcImage->height == dstImage->height) && (srcImage->depth == dstImage->depth) &&
-           (srcImage->yuvFormat == dstImage->yuvFormat) && (srcImage->yuvRange == dstImage->yuvRange));
+    assert(srcImage->depth == dstImage->depth);
+    if (planes & AVIF_PLANES_YUV) {
+        assert((srcImage->yuvFormat == dstImage->yuvFormat) && (srcImage->yuvRange == dstImage->yuvRange));
+    }
     const size_t bytesPerPixel = avifImageUsesU16(srcImage) ? 2 : 1;
 
     const avifBool skipColor = !(planes & AVIF_PLANES_YUV);
@@ -206,6 +208,8 @@ void avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avif
         if (!srcRow) {
             continue;
         }
+        assert(planeWidth == avifImagePlaneWidth(dstImage, c));
+        assert(planeHeight == avifImagePlaneHeight(dstImage, c));
 
         const size_t planeWidthBytes = planeWidth * bytesPerPixel;
         for (uint32_t y = 0; y < planeHeight; ++y) {
--- a/src/avif.c
+++ b/src/avif.c
@@ -208,17 +208,41 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
             return allocationResult;
         }
     }
-    for (int plane = AVIF_CHAN_Y; plane <= AVIF_CHAN_A; ++plane) {
-        uint8_t * dstRow = avifImagePlane(dstImage, plane);
-        if (!dstRow) {
+    if (avifImageCopySamples(dstImage, srcImage, planes) != AVIF_RESULT_OK) {
+        assert(AVIF_FALSE);
+    }
+    return AVIF_RESULT_OK;
+}
+
+avifResult avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
+{
+    AVIF_CHECKERR(srcImage->depth == dstImage->depth, AVIF_RESULT_INVALID_ARGUMENT);
+    const size_t bytesPerPixel = avifImageUsesU16(srcImage) ? 2 : 1;
+
+    const avifBool skipColor = !(planes & AVIF_PLANES_YUV);
+    const avifBool skipAlpha = !(planes & AVIF_PLANES_A);
+    for (int c = AVIF_CHAN_Y; c <= AVIF_CHAN_A; ++c) {
+        const avifBool isColor = c != AVIF_CHAN_A;
+        const avifBool isAlpha = c == AVIF_CHAN_A;
+        if ((skipColor && isColor) || (skipAlpha && isAlpha)) {
             continue;
         }
-        const uint8_t * srcRow = avifImagePlane(srcImage, plane);
-        uint32_t srcRowBytes = avifImagePlaneRowBytes(srcImage, plane);
-        uint32_t dstRowBytes = avifImagePlaneRowBytes(dstImage, plane);
-        uint32_t planeWidthBytes = avifImagePlaneWidth(dstImage, plane) << (dstImage->depth > 8);
-        uint32_t planeHeight = avifImagePlaneHeight(dstImage, plane);
-        for (uint32_t j = 0; j < planeHeight; ++j) {
+
+        const uint32_t planeWidth = avifImagePlaneWidth(srcImage, c);
+        const uint32_t planeHeight = avifImagePlaneHeight(srcImage, c);
+        const uint8_t * srcRow = avifImagePlane(srcImage, c);
+        uint8_t * dstRow = avifImagePlane(dstImage, c);
+        const uint32_t srcRowBytes = avifImagePlaneRowBytes(srcImage, c);
+        const uint32_t dstRowBytes = avifImagePlaneRowBytes(dstImage, c);
+        AVIF_CHECKERR(planeWidth == avifImagePlaneWidth(dstImage, c), AVIF_RESULT_INVALID_ARGUMENT);
+        AVIF_CHECKERR(planeHeight == avifImagePlaneHeight(dstImage, c), AVIF_RESULT_INVALID_ARGUMENT);
+        AVIF_CHECKERR(!srcRow == !dstRow, AVIF_RESULT_INVALID_ARGUMENT);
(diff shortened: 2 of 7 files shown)


tests/gtest/avifgridapitest.cc
```diff
@@ -4,6 +4,7 @@
 #include <vector>
 
 #include "avif/avif.h"
+#include "avif/internal.h"
```
jzern, line 7: Is there any visibility applied when a shared object is built or are the tests always linking a static lib/objects? If the shared object is used and restricts the visible symbols like it should, this will fail to link.

src/avif.c
```diff
@@ -208,17 +208,41 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
...
-    for (int plane = AVIF_CHAN_Y; plane <= AVIF_CHAN_A; ++plane) {
-        uint8_t * dstRow = avifImagePlane(dstImage, plane);
-        if (!dstRow) {
+    if (avifImageCopySamples(dstImage, srcImage, planes) != AVIF_RESULT_OK) {
+        assert(AVIF_FALSE);
+    }
+    return AVIF_RESULT_OK;
+}
+
+avifResult avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
```
jzern, line 217: This could go before it's first use in `avifImageCopy()`, but the ordering of functions is a bit unclear since this implements things from internal/avif.h and avif.h

jzern (approved):

tests/gtest/avifgridapitest.cc
y-guyon: I see `7/25 Test  #8: avifgridapitest ... Passed` in the [build-shared-local (ubuntu-latest)](https://github.com/AOMediaCodec/libavif/actions/runs/4322639324/jobs/7545288861) presubmit so I guess it works. It means the visible symbols are not restricted as they should be? Can this be configured in CMakeLists.txt?

wantehchang: I just tested this. When I configure libavif with `cmake -DBUILD_SHARED_LIBS=ON`, two static libraries are produced:  libavif_apps.a  and libavif_internal.a. So avifgridapitest is linked with libavif_internal.a.

The `readelf` command shows `avifImageCopySamples` is not exported from the shared library:

```
$ readelf --dyn-syms libavif.so.15.0.1 | grep avifImageCopy
    61: 00000000000054da   579 FUNC    GLOBAL DEFAULT   12 avifImageCopy
```

include/avif/internal.h
```diff
@@ -89,6 +89,10 @@ void avifImageSetDefaults(avifImage * image);
 // Copies all fields that do not need to be freed/allocated from srcImage to dstImage.
 void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage);
 
+// Copies the samples from srcImage to dstImage. dstImage must be allocated.
+// srcImage and dstImage must have the same width, height, depth and yuvFormat.
```
wantehchang, line 93: The original `CopyImageSamples()` function in aviftest_helpers.cc also checks `yuvRange`. Should `avifImageCopySamples()` also check `yuvRange`?

src/avif.c
```diff
@@ -182,6 +182,43 @@ void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage)
...
+{
+    AVIF_CHECKERR(srcImage->depth == dstImage->depth, AVIF_RESULT_INVALID_ARGUMENT);
+    const size_t bytesPerPixel = avifImageUsesU16(srcImage) ? 2 : 1;
+
+    const avifBool skipColor = !(planes & AVIF_PLANES_YUV);
+    const avifBool skipAlpha = !(planes & AVIF_PLANES_A);
+    for (int c = AVIF_CHAN_Y; c <= AVIF_CHAN_A; ++c) {
+        const avifBool isColor = c != AVIF_CHAN_A;
+        const avifBool isAlpha = c == AVIF_CHAN_A;
```
wantehchang, line 194: 1. Optional: Use the commonly used name `alpha` for this boolean.

2. Replace `isColor` with `!isAlpha`.

src/avif.c
```diff
@@ -182,6 +182,43 @@ void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage)
...
+
+        const uint32_t planeWidth = avifImagePlaneWidth(srcImage, c);
+        const uint32_t planeHeight = avifImagePlaneHeight(srcImage, c);
+        const uint8_t * srcRow = avifImagePlane(srcImage, c);
+        uint8_t * dstRow = avifImagePlane(dstImage, c);
+        const uint32_t srcRowBytes = avifImagePlaneRowBytes(srcImage, c);
+        const uint32_t dstRowBytes = avifImagePlaneRowBytes(dstImage, c);
+        AVIF_CHECKERR(planeWidth == avifImagePlaneWidth(dstImage, c), AVIF_RESULT_INVALID_ARGUMENT);
+        AVIF_CHECKERR(planeHeight == avifImagePlaneHeight(dstImage, c), AVIF_RESULT_INVALID_ARGUMENT);
+        AVIF_CHECKERR(!srcRow == !dstRow, AVIF_RESULT_INVALID_ARGUMENT);
```
wantehchang, line 207: 1. Nit/optional: These expand to three if statements. Seems a little excessive.

2. I think these are equivalent to checking `width`, `height`, and `yuvFormat`, which can be checked before the for loop.

src/avif.c
```diff
@@ -208,21 +245,8 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
...
-        uint32_t dstRowBytes = avifImagePlaneRowBytes(dstImage, plane);
-        uint32_t planeWidthBytes = avifImagePlaneWidth(dstImage, plane) << (dstImage->depth > 8);
-        uint32_t planeHeight = avifImagePlaneHeight(dstImage, plane);
-        for (uint32_t j = 0; j < planeHeight; ++j) {
-            memcpy(dstRow, srcRow, planeWidthBytes);
-            srcRow += srcRowBytes;
-            dstRow += dstRowBytes;
-        }
+    if (avifImageCopySamples(dstImage, srcImage, planes) != AVIF_RESULT_OK) {
+        assert(AVIF_FALSE);
```
wantehchang, line 249: Return AVIF_FALSE. Assertions are compiled away in release builds.

Note: If `avifImageCopySamples()` is only used in places where it is obvious that `dstImage` and `srcImage` have the same properties, we can change `avifImageCopySamples()` to return void and change its checks to assertions.

src/read.c
```diff
@@ -1388,59 +1387,22 @@ static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
...
-                    memcpy(dstU, srcU, uvRowBytes);
-
-                    const uint8_t * srcV = &tile->image->yuvPlanes[AVIF_CHAN_V][j * tile->image->yuvRowBytes[AVIF_CHAN_V]];
-                    uint8_t * dstV =
-                        &dstImage->yuvPlanes[AVIF_CHAN_V][(uvColOffset * pixelBytes) + ((uvRowOffset + j) * dstImage->yuvRowBytes[AVIF_CHAN_V])];
-                    memcpy(dstV, srcV, uvRowBytes);
-                }
+            const avifCropRect srcViewRect = { 0, 0, dstViewRect.width, dstViewRect.height };
+            avifImage * srcView = avifImageCreateEmpty();
+            avifImage * dstView = avifImageCreateEmpty();
```
wantehchang, line 1401: IMPORTANT:

1. `srcView` and `dstView` can be created outside the for loops and reused for all loop iterations.

2. Need to handle `avifImageCreateEmpty()` failures.

3. Need to destroy `srcView` and `dstView` when done.

Note: Since this is internal code, we can allocate `srcView` and `dstView` on the stack. Then the second and third issues are gone. The following patch does that:

```
index febe15f..6718ca2 100644
--- a/src/read.c
+++ b/src/read.c
@@ -1379,6 +1379,10 @@ static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
     unsigned int colIndex = oldDecodedTileCount % grid->columns;
     // Only the first iteration of the outer for loop uses this initial value of colIndex.
     // Subsequent iterations of the outer for loop initializes colIndex to 0.
+    avifImage srcView;
+    avifImageSetDefaults(&srcView);
+    avifImage dstView;
+    avifImageSetDefaults(&dstView);
     for (; rowIndex < grid->rows; ++rowIndex, colIndex = 0) {
         for (; colIndex < grid->columns; ++colIndex, ++tileIndex) {
             if (tileIndex >= decodedTileCount) {
@@ -1397,12 +1401,11 @@ static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
                 dstViewRect.height = grid->outputHeight - dstViewRect.y;
             }
             const avifCropRect srcViewRect = { 0, 0, dstViewRect.width, dstViewRect.height };
-            avifImage * srcView = avifImageCreateEmpty();
-            avifImage * dstView = avifImageCreateEmpty();
-            if ((avifImageSetViewRect(dstView, dstImage, &dstViewRect) != AVIF_RESULT_OK) ||
-                (avifImageSetViewRect(srcView, tile->image, &srcViewRect) != AVIF_RESULT_OK) ||
-                (avifImageCopySamples(dstView, srcView, alpha ? AVIF_PLANES_A : AVIF_PLANES_YUV) != AVIF_RESULT_OK)) {
+            if ((avifImageSetViewRect(&dstView, dstImage, &dstViewRect) != AVIF_RESULT_OK) ||
+                (avifImageSetViewRect(&srcView, tile->image, &srcViewRect) != AVIF_RESULT_OK) ||
+                (avifImageCopySamples(&dstView, &srcView, alpha ? AVIF_PLANES_A : AVIF_PLANES_YUV) != AVIF_RESULT_OK)) {
                 assert(AVIF_FALSE);
             }
         }
     }
```

src/read.c
```diff
@@ -1388,59 +1387,22 @@ static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
...
+            if ((avifImageSetViewRect(dstView, dstImage, &dstViewRect) != AVIF_RESULT_OK) ||
+                (avifImageSetViewRect(srcView, tile->image, &srcViewRect) != AVIF_RESULT_OK) ||
+                (avifImageCopySamples(dstView, srcView, alpha ? AVIF_PLANES_A : AVIF_PLANES_YUV) != AVIF_RESULT_OK)) {
+                assert(AVIF_FALSE);
```
wantehchang, line 1405: Need to return AVIF_FALSE. Assertions are compiled away in release builds.

tests/gtest/avifgridapitest.cc
```diff
@@ -94,7 +95,7 @@ avifResult EncodeDecodeGrid(const std::vector<std::vector<Cell>>& cell_rows,
       if (result != AVIF_RESULT_OK) {
         return result;
       }
-      testutil::CopyImageSamples(**it, view.get());
+      avifImageCopySamples(/*dstImage=*/view.get(), it->get(), AVIF_PLANES_ALL);
```
wantehchang, line 98: Check return value.

wantehchang (approved):
LGTM. I suggest some changes.

include/avif/internal.h
```diff
@@ -89,6 +89,10 @@ void avifImageSetDefaults(avifImage * image);
 // Copies all fields that do not need to be freed/allocated from srcImage to dstImage.
 void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage);
 
+// Copies the samples from srcImage to dstImage. dstImage must be allocated.
+// srcImage and dstImage must have the same width, height, depth, yuvFormat and yuvRange.
```
wantehchang, line 93: I found that when copying only the alpha plane (`planes == AVIF_PLANES_A`), `yuvFormat` may be set to `yuvFormat = AVIF_PIXEL_FORMAT_NONE`. So the second sentence should be shortened to:

```
srcImage and dstImage must have the same width, height, and depth.
```

I think we can add a third sentence:

```
If the AVIF_PLANES_YUV bit is set in `planes`, then srcImage and dstImage must have the same yuvFormat and yuvRange.
```

src/avif.c
```diff
@@ -182,6 +182,40 @@ void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage)
     dstImage->imir = srcImage->imir;
 }
 
+void avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
+{
+    assert((srcImage->width == dstImage->width) && (srcImage->height == dstImage->height) && (srcImage->depth == dstImage->depth) &&
+           (srcImage->yuvFormat == dstImage->yuvFormat) && (srcImage->yuvRange == dstImage->yuvRange));
```
wantehchang, line 188: This assertion fails in Debug build. The following patch fixes the assertion failure:

```
diff --git a/src/avif.c b/src/avif.c
index 79e5082..fd6ef9c 100644
--- a/src/avif.c
+++ b/src/avif.c
@@ -184,8 +184,7 @@ void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage)
 
 void avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
 {
-    assert((srcImage->width == dstImage->width) && (srcImage->height == dstImage->height) && (srcImage->depth == dstImage->depth) &&
-           (srcImage->yuvFormat == dstImage->yuvFormat) && (srcImage->yuvRange == dstImage->yuvRange));
+    assert(srcImage->depth == dstImage->depth);
     const size_t bytesPerPixel = avifImageUsesU16(srcImage) ? 2 : 1;
 
     const avifBool skipColor = !(planes & AVIF_PLANES_YUV);
@@ -206,6 +205,8 @@ void avifImageCopySamples(avifImage * dstImage, const avifImage * srcImage, avif
         if (!srcRow) {
             continue;
         }
+        assert(planeWidth == avifImagePlaneWidth(dstImage, c));
+        assert(planeHeight == avifImagePlaneHeight(dstImage, c));
 
         const size_t planeWidthBytes = planeWidth * bytesPerPixel;
         for (uint32_t y = 0; y < planeHeight; ++y) {
```

This patch moves the checks for width and height back inside the for loops. To be conservative, it asserts srcImage and dstImage have the same width and height for the plane only when it is about to copy that plane.

Also, this patch doesn't check if yuvRange is the same when we need to copy YUV planes. That check is less important because if it is not true, we still won't have memory errors.

An alternative fix is to write the assertions according to the comment that I suggested:

```
If the AVIF_PLANES_YUV bit is set in `planes`, then srcImage and dstImage must have the same yuvFormat and yuvRange.
```

wantehchang (approved):
The changes I suggested for the width and height checks are incorrect. Sorry about my mistake.

y-guyon:
> It would be good to add a CI build in which assertions are enabled, such as a Debug build.

Done at https://github.com/AOMediaCodec/libavif/pull/1302

wantehchang (approved):
LGTM!

PR #1303, opened 2023-03-23
Use avifImagePlane*() functions in avifImageDownshiftTo8bpc

vigneshvg:
Gets rid of some repeated subsampling computations.

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -693,22 +693,18 @@ static avifResult avifImageDownshiftTo8bpc(const avifImage * image, avifImage * 
     }
     // 16384 for 10-bit and 4096 for 12-bit.
     const int scale = 1 << (24 - image->depth);
-    avifPixelFormatInfo pixelInfo;
-    avifGetPixelFormatInfo(image8->yuvFormat, &pixelInfo);
-    for (int plane = 0; plane < (pixelInfo.monochrome ? 1 : 3); ++plane) {
-        int planeWidth = image->width;
-        int planeHeight = image->height;
-        if (plane > 0) {
-            planeWidth = (planeWidth + pixelInfo.chromaShiftX) >> pixelInfo.chromaShiftX;
-            planeHeight = (planeHeight + pixelInfo.chromaShiftY) >> pixelInfo.chromaShiftY;
+    for (int plane = AVIF_CHAN_Y; plane <= AVIF_CHAN_V; ++plane) {
+        const uint32_t planeWidth = avifImagePlaneWidth(image, plane);
+        if (planeWidth == 0) {
+            continue;
         }
-        Convert16To8Plane((const uint16_t *)image->yuvPlanes[plane],
-                          image->yuvRowBytes[plane] / 2,
-                          image8->yuvPlanes[plane],
-                          image8->yuvRowBytes[plane],
+        Convert16To8Plane((const uint16_t *)avifImagePlane(image, plane),
+                          avifImagePlaneRowBytes(image, plane) / 2,
+                          avifImagePlane(image8, plane),
+                          avifImagePlaneRowBytes(image8, plane),
                           scale,
                           planeWidth,
-                          planeHeight);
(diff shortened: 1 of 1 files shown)


jzern (approved):

src/reformat_libyuv.c
```diff
@@ -693,22 +693,18 @@ static avifResult avifImageDownshiftTo8bpc(const avifImage * image, avifImage *
...
+        const uint32_t planeWidth = avifImagePlaneWidth(image, plane);
+        if (planeWidth == 0) {
+            continue;
         }
-        Convert16To8Plane((const uint16_t *)image->yuvPlanes[plane],
-                          image->yuvRowBytes[plane] / 2,
-                          image8->yuvPlanes[plane],
-                          image8->yuvRowBytes[plane],
+        Convert16To8Plane((const uint16_t *)avifImagePlane(image, plane),
+                          avifImagePlaneRowBytes(image, plane) / 2,
```
y-guyon, line 702: This is out of the scope of this PR but should `!(avifImagePlaneRowBytes(image, plane) & 1)` be checked? Could some wonky user input with weird unaligned uint16_t odd rows be forwarded through some public API call to `avifImageDownshiftTo8bpc()` ?
I do not see anything in `avif.h` that would forbid that. Maybe a comment on `yuvRowBytes` is enough.

src/reformat_libyuv.c
```diff
@@ -693,22 +693,18 @@ static avifResult avifImageDownshiftTo8bpc(const avifImage * image, avifImage *
...
+    for (int plane = AVIF_CHAN_Y; plane <= AVIF_CHAN_V; ++plane) {
+        const uint32_t planeWidth = avifImagePlaneWidth(image, plane);
+        if (planeWidth == 0) {
+            continue;
         }
-        Convert16To8Plane((const uint16_t *)image->yuvPlanes[plane],
-                          image->yuvRowBytes[plane] / 2,
-                          image8->yuvPlanes[plane],
-                          image8->yuvRowBytes[plane],
+        Convert16To8Plane((const uint16_t *)avifImagePlane(image, plane),
```
wantehchang, line 701: It is not necessary to call `avifImagePlane()` and `avifImagePlaneRowBytes()`.

vigneshvg: Yeah, i have a follow-up where i handle alpha plane also in this function. So i was just making this change to avoid some unnecessary delta in that PR.

src/reformat_libyuv.c
vigneshvg: I haven't really thought about the case where bitdepth is > 8 and stride is not a multiple of 2. I imagine there could be other functions where we just sort of presume this.

In this specific case, even if it makes its way here through the public API, the code won't crash since we are rounding down when dividing it by 2 (it won't be correct, but it won't crash either).

Let me look into this a bit more and see if i can add checks at a higher level to disallow this.

wantehchang: Yannis's question is thought-provoking. I am not suggesting we support this, but here is an example of the kind of code that should still work if yuvRowBytes[i] or alphaRowBytes is an odd number and the CPU (such as x86) allows unaligned access to uint16_t:

src/avif.c:

```
avifBool avifImageIsOpaque(const avifImage * image)
{
    if (!image->alphaPlane) {
        return AVIF_TRUE;
    }

    const uint32_t opaqueValue = (1u << image->depth) - 1u;
    const uint8_t * row = image->alphaPlane;
    for (uint32_t y = 0; y < image->height; ++y) {
        if (avifImageUsesU16(image)) {
            const uint16_t * row16 = (const uint16_t *)row;
            for (uint32_t x = 0; x < image->width; ++x) {
                if (row16[x] != opaqueValue) {
                    return AVIF_FALSE;
                }
            }
        } else { 
            ...
        }
        row += image->alphaRowBytes;
    }
    return AVIF_TRUE;
}
```

Perhaps we should audit our code and make sure yuvRowBytes[i] or alphaRowBytes is an even number before we divide it by 2.

PR #1304, opened 2023-03-23
avifrgbtoyuvtest: Add tests for YUV with alpha

vigneshvg:
Add a has_alpha parameter to YUVToRGBThreadingTests which will test the multi-threaded validity of YUV images with Alpha plane as well.

Also replace one of the rgb formats in the default test instantiation with the more commonly used RGBA (instead of ARGB).

### DIFF
--- a/tests/gtest/avifrgbtoyuvtest.cc
+++ b/tests/gtest/avifrgbtoyuvtest.cc
@@ -741,7 +741,8 @@ class YUVToRGBThreadingTest
     : public testing::TestWithParam<std::tuple<
           /*rgb_depth=*/int, /*yuv_depth=*/int,
           /*width=*/int, /*height=*/int, avifRGBFormat, avifPixelFormat,
-          /*threads=*/int, /*avoidLibYUV=*/bool, avifChromaUpsampling>> {};
+          /*threads=*/int, /*avoidLibYUV=*/bool, avifChromaUpsampling,
+          /*has_alpha=*/bool>> {};
 
 TEST_P(YUVToRGBThreadingTest, TestIdentical) {
   const int rgb_depth = std::get<0>(GetParam());
@@ -753,6 +754,7 @@ TEST_P(YUVToRGBThreadingTest, TestIdentical) {
   const int maxThreads = std::get<6>(GetParam());
   const bool avoidLibYUV = std::get<7>(GetParam());
   const avifChromaUpsampling chromaUpsampling = std::get<8>(GetParam());
+  const bool has_alpha = std::get<9>(GetParam());
 
   if (rgb_depth > 8 && rgb_format == AVIF_RGB_FORMAT_RGB_565) {
     return;
@@ -764,19 +766,20 @@ TEST_P(YUVToRGBThreadingTest, TestIdentical) {
   yuv->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601;
   yuv->yuvRange = AVIF_RANGE_FULL;
 
-  // Fill YUV planes with random values.
+  // Fill YUVA planes with random values.
   srand(0xAABBCCDD);
   const int yuv_max = (1 << yuv_depth);
-  ASSERT_EQ(avifImageAllocatePlanes(yuv.get(), AVIF_PLANES_YUV),
+  ASSERT_EQ(avifImageAllocatePlanes(
+                yuv.get(), has_alpha ? AVIF_PLANES_ALL : AVIF_PLANES_YUV),
             AVIF_RESULT_OK);
-  avifPixelFormatInfo pixel_format_info;
-  avifGetPixelFormatInfo(yuv_format, &pixel_format_info);
-  const int plane_count = pixel_format_info.monochrome ? 1 : 3;
-  for (int plane = 0; plane < plane_count; ++plane) {
+  for (int plane = AVIF_CHAN_Y; plane <= AVIF_CHAN_A; ++plane) {
+    const uint32_t plane_width = avifImagePlaneWidth(yuv.get(), plane);
+    if (plane_width == 0) continue;
+    const uint32_t plane_height = avifImagePlaneHeight(yuv.get(), plane);
+    const uint32_t rowBytes = avifImagePlaneRowBytes(yuv.get(), plane);
     uint8_t* row = avifImagePlane(yuv.get(), plane);
-    for (uint32_t y = 0; y < avifImagePlaneHeight(yuv.get(), plane);
-         ++y, row += yuv.get()->yuvRowBytes[plane]) {
-      for (uint32_t x = 0; x < avifImagePlaneWidth(yuv.get(), plane); ++x) {
+    for (uint32_t y = 0; y < plane_height; ++y, row += rowBytes) {
+      for (uint32_t x = 0; x < plane_width; ++x) {
         if (yuv_depth == 8) {
           row[x] = (uint8_t)(rand() % yuv_max);
         } else {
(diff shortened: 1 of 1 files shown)


jzern:
```
D:\a\libavif\libavif\tests\gtest\avifrgbtoyuvtest.cc(777): error C2220: the following warning is treated as an error
D:\a\libavif\libavif\tests\gtest\avifrgbtoyuvtest.cc(777): warning C4456: declaration of 'width' hides previous local declaration
D:\a\libavif\libavif\tests\gtest\avifrgbtoyuvtest.cc(750): note: see declaration of 'width'
D:\a\libavif\libavif\tests\gtest\avifrgbtoyuvtest.cc(779): warning C4456: declaration of 'height' hides previous local declaration
D:\a\libavif\libavif\tests\gtest\avifrgbtoyuvtest.cc(751): note: see declaration of 'height'
```

Looks like we should enable -Wshadow in the Linux configs.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -741,7 +741,8 @@ class YUVToRGBThreadingTest
     : public testing::TestWithParam<std::tuple<
           /*rgb_depth=*/int, /*yuv_depth=*/int,
           /*width=*/int, /*height=*/int, avifRGBFormat, avifPixelFormat,
-          /*threads=*/int, /*avoidLibYUV=*/bool, avifChromaUpsampling>> {};
+          /*threads=*/int, /*avoidLibYUV=*/bool, avifChromaUpsampling,
+          /*has_alpha*/ bool>> {};
```
jzern, line 745: this should be `/*has_alpha=*/bool` to match the others

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -764,19 +766,21 @@ TEST_P(YUVToRGBThreadingTest, TestIdentical) {
...
+                yuv.get(), has_alpha ? AVIF_PLANES_ALL : AVIF_PLANES_YUV),
             AVIF_RESULT_OK);
-  avifPixelFormatInfo pixel_format_info;
-  avifGetPixelFormatInfo(yuv_format, &pixel_format_info);
-  const int plane_count = pixel_format_info.monochrome ? 1 : 3;
-  for (int plane = 0; plane < plane_count; ++plane) {
+  for (int plane = AVIF_CHAN_Y;
+       plane <= (has_alpha ? AVIF_CHAN_A : AVIF_CHAN_V); ++plane) {
+    const uint32_t plane_width = avifImagePlaneWidth(yuv.get(), plane);
+    if (plane_width == 0) continue;
```
jzern, line 778: is this check needed given the loop index?

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -808,14 +812,15 @@ INSTANTIATE_TEST_SUITE_P(
...
+            Values(AVIF_RGB_FORMAT_RGB, AVIF_RGB_FORMAT_RGBA),
             Range(AVIF_PIXEL_FORMAT_YUV444, AVIF_PIXEL_FORMAT_COUNT),
             // Test an odd and even number for threads. Not adding all possible
             // thread values to keep the number of test instances low.
             /*threads=*/Values(2, 7),
             /*avoidLibYUV=*/Values(true, false),
             Values(AVIF_CHROMA_UPSAMPLING_FASTEST,
-                   AVIF_CHROMA_UPSAMPLING_BILINEAR)));
+                   AVIF_CHROMA_UPSAMPLING_BILINEAR),
+            /*has_alpha*/ Values(true, false)));
```
jzern, line 823: `/*has_alpha=*/Values(...`

jzern (approved):

tests/gtest/avifrgbtoyuvtest.cc
vigneshvg: yes the ordering of the planes is Y, U, V and A. this check is needed to skip U and V planes when the format is monochrome (monochrome could have alpha).

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -764,19 +766,21 @@ TEST_P(YUVToRGBThreadingTest, TestIdentical) {
...
-  ASSERT_EQ(avifImageAllocatePlanes(yuv.get(), AVIF_PLANES_YUV),
+  ASSERT_EQ(avifImageAllocatePlanes(
+                yuv.get(), has_alpha ? AVIF_PLANES_ALL : AVIF_PLANES_YUV),
             AVIF_RESULT_OK);
-  avifPixelFormatInfo pixel_format_info;
-  avifGetPixelFormatInfo(yuv_format, &pixel_format_info);
-  const int plane_count = pixel_format_info.monochrome ? 1 : 3;
-  for (int plane = 0; plane < plane_count; ++plane) {
+  for (int plane = AVIF_CHAN_Y;
+       plane <= (has_alpha ? AVIF_CHAN_A : AVIF_CHAN_V); ++plane) {
```
wantehchang, line 776: We can blindly go up to `AVIF_CHAN_A` here. If `yuv` doesn't have the alpha plane allocated, `avifImagePlaneWidth(yuv.get(), AVIF_CHAN_A)` returns 0 and line 778 will handle this case.

wantehchang (changes requested):

tests/gtest/avifrgbtoyuvtest.cc
jzern: Got it. I wasn't considering monochrome.

tests/gtest/avifrgbtoyuvtest.cc
wantehchang, line 823: Optional: In libaom I have seen `::testing::Bool()` used, which is equivalent to `Values(false, true)`. You can use `Values(true, false)` to match existing usage (such as line 820).

vigneshvg: Done. Replaced the new usage with ::testing::Bool().

There are a few other instances of Values(true, false). I will change them all to ::testing::Bool() in a separate PR.

vigneshvg:
@wantehchang i will go ahead and submit this change. if you have any further comments, i'll address them in a follow-up.

wantehchang (commented):
Vignesh: No problem. James Zern's review is sufficient. Yannis would be the best person to review tests/gtest/avifrgbtoyuvtest.cc because he's the original author.

In general my concern for this kind of change is the combinatorial explosion of tests, but I don't have a good suggestion on how to avoid exhaustive testing without sacrificing test coverage.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -753,6 +754,7 @@ TEST_P(YUVToRGBThreadingTest, TestIdentical) {
   const int maxThreads = std::get<6>(GetParam());
   const bool avoidLibYUV = std::get<7>(GetParam());
   const avifChromaUpsampling chromaUpsampling = std::get<8>(GetParam());
+  const bool has_alpha = std::get<9>(GetParam());
```
y-guyon, line 757: `has_alpha` could be deduced from `rgb_format` having an `A` component or not

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -808,14 +811,15 @@ INSTANTIATE_TEST_SUITE_P(
...
+            Values(AVIF_RGB_FORMAT_RGB, AVIF_RGB_FORMAT_RGBA),
             Range(AVIF_PIXEL_FORMAT_YUV444, AVIF_PIXEL_FORMAT_COUNT),
             // Test an odd and even number for threads. Not adding all possible
             // thread values to keep the number of test instances low.
             /*threads=*/Values(2, 7),
             /*avoidLibYUV=*/Values(true, false),
             Values(AVIF_CHROMA_UPSAMPLING_FASTEST,
-                   AVIF_CHROMA_UPSAMPLING_BILINEAR)));
+                   AVIF_CHROMA_UPSAMPLING_BILINEAR),
+            /*has_alpha=*/::testing::Bool()));
```
y-guyon, line 822: I usually avoid `::testing::Bool()` because it is harder to understand than `Values(false, true)` and it is less convenient when debugging (when you just want to set one value and not both).

y-guyon (commented):
Thanks for increasing the test coverage.

> Yannis would be the best person to review tests/gtest/avifrgbtoyuvtest.cc because he's the original author.

LGTM.

> In general my concern for this kind of change is the combinatorial explosion of tests, but I don't have a good suggestion on how to avoid exhaustive testing without sacrificing test coverage.

Fuzzing is usually a good candidate for that.

PR #1305, opened 2023-03-24
Fix avifImageSetViewRect() for monochrome odd xy

y-guyon:
avifGetPixelFormatInfo() sets chromaShiftX,Y to 1 even when there is no chroma.


maryla-uc (approved):

wantehchang (commented):
LGTM.

PR #1308, opened 2023-03-24
Skip avifPixelFormatInfo.chromaShift if monochrome

y-guyon:
Also change avifGetPixelFormatInfo() so that chromaShiftX and chromaShiftY are 0 if monochrome is set. It should not be accessed anyway but setting chroma to half the size of luma makes little sense when there is no chroma. The following comment was removed in 6ad6a2b:
  // ignored, but some codecs might use 420 for mono

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,8 @@ List of incompatible ABI changes in this release:
 * avifEncoderSetCodecSpecificOption() now returns avifResult instead of void to
   report memory allocation failures.
 * At decoding, avifIOStats now returns the same values as at encoding.
+* avifGetPixelFormatInfo() now returns 0 for chromaShiftX and chromaShiftY when
+  the format is AVIF_PIXEL_FORMAT_YUV400. Previously it returned 1.
 
 ## [0.11.1] - 2022-10-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,8 +60,6 @@ List of incompatible ABI changes in this release:
 * avifEncoderSetCodecSpecificOption() now returns avifResult instead of void to
   report memory allocation failures.
 * At decoding, avifIOStats now returns the same values as at encoding.
-* avifGetPixelFormatInfo() now returns 0 for chromaShiftX and chromaShiftY when
-  the format is AVIF_PIXEL_FORMAT_YUV400. Previously it returned 1.
 
 ## [0.11.1] - 2022-10-19
 
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -574,8 +574,10 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
     uint32_t cellWidth, cellHeight;
     avifPixelFormatInfo formatInfo;
     avifGetPixelFormatInfo(gridSplitImage->yuvFormat, &formatInfo);
-    if (!avifGetBestCellSize("horizontally", gridSplitImage->width, gridCols, formatInfo.chromaShiftX, &cellWidth) ||
-        !avifGetBestCellSize("vertically", gridSplitImage->height, gridRows, formatInfo.chromaShiftY, &cellHeight)) {
+    const avifBool isSubsampledX = !formatInfo.monochrome && formatInfo.chromaShiftX;
+    const avifBool isSubsampledY = !formatInfo.monochrome && formatInfo.chromaShiftY;
+    if (!avifGetBestCellSize("horizontally", gridSplitImage->width, gridCols, isSubsampledX, &cellWidth) ||
+        !avifGetBestCellSize("vertically", gridSplitImage->height, gridRows, isSubsampledY, &cellHeight)) {
         return AVIF_FALSE;
     }
 
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -236,6 +236,9 @@ typedef struct avifPixelFormatInfo
     int chromaShiftY;
 } avifPixelFormatInfo;
 
+// Returns the avifPixelFormatInfo depending on the avifPixelFormat.
+// info does not need to be zero-initialized. Sets all monochrome, chromaShiftX and
+// chromaShiftY fields to 1 if format is AVIF_PIXEL_FORMAT_YUV400.
 AVIF_API void avifGetPixelFormatInfo(avifPixelFormat format, avifPixelFormatInfo * info);
 
 // ---------------------------------------------------------------------------
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -237,10 +237,11 @@ typedef struct avifPixelFormatInfo
 } avifPixelFormatInfo;
 
 // Returns the avifPixelFormatInfo depending on the avifPixelFormat.
-// The chromaShiftX and chromaShiftY fields should be ignored when monochrome is AVIF_TRUE.
-// The chromaShiftX and chromaShiftY fields are set to 1 if the format is AVIF_PIXEL_FORMAT_YUV400 to allow
+// When monochrome is AVIF_TRUE, chromaShiftX and chromaShiftY are set to 1 according to the AV1 specification but they should be ignored.
+//
+// Note: This function implements the second table on page 119 of the AV1 specification version 1.0.0 with Errata 1.
+// For monochrome 4:0:0, subsampling_x and subsampling are specified as 1 to allow
 // an AV1 implementation that only supports profile 0 to hardcode subsampling_x and subsampling_y to 1.
-// This function implements the second table on page 119 of the AV1 specification version 1.0.0 with Errata 1.
 AVIF_API void avifGetPixelFormatInfo(avifPixelFormat format, avifPixelFormatInfo * info);
 
 // ---------------------------------------------------------------------------
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -237,8 +237,10 @@ typedef struct avifPixelFormatInfo
 } avifPixelFormatInfo;
 
 // Returns the avifPixelFormatInfo depending on the avifPixelFormat.
-// info does not need to be zero-initialized. Sets all monochrome, chromaShiftX and
-// chromaShiftY fields to 1 if format is AVIF_PIXEL_FORMAT_YUV400.
+// The chromaShiftX and chromaShiftY fields should be ignored when monochrome is AVIF_TRUE.
+// The chromaShiftX and chromaShiftY fields are set to 1 if the format is AVIF_PIXEL_FORMAT_YUV400 to allow
+// an AV1 implementation that only supports profile 0 to hardcode subsampling_x and subsampling_y to 1.
+// This function implements the second table on page 119 of the AV1 specification version 1.0.0 with Errata 1.
 AVIF_API void avifGetPixelFormatInfo(avifPixelFormat format, avifPixelFormatInfo * info);
 
 // ---------------------------------------------------------------------------
--- a/src/avif.c
+++ b/src/avif.c
@@ -58,8 +58,8 @@ void avifGetPixelFormatInfo(avifPixelFormat format, avifPixelFormatInfo * info)
 
         case AVIF_PIXEL_FORMAT_YUV400:
             info->monochrome = AVIF_TRUE;
-            // 4:0:0 is stored as monochrome and half chroma subsampling in each dimension in AV1
-            // according to the specification. See sections 5.5.2. and 6.4.2.
+            // 4:0:0 is stored as monochrome and the nonexistent chroma is considered as subsampled in
+            // each dimension according to the AV1 specification. See sections 5.5.2. and 6.4.2.
             info->chromaShiftX = 1;
             info->chromaShiftY = 1;
             break;
--- a/src/avif.c
+++ b/src/avif.c
@@ -58,6 +58,10 @@ void avifGetPixelFormatInfo(avifPixelFormat format, avifPixelFormatInfo * info)
 
         case AVIF_PIXEL_FORMAT_YUV400:
             info->monochrome = AVIF_TRUE;
+            // 4:0:0 is stored as monochrome and half chroma subsampling in each dimension in AV1
+            // according to the specification. See sections 5.5.2. and 6.4.2.
+            info->chromaShiftX = 1;
+            info->chromaShiftY = 1;
             break;
 
         case AVIF_PIXEL_FORMAT_NONE:
--- a/src/avif.c
+++ b/src/avif.c
@@ -58,8 +58,8 @@ void avifGetPixelFormatInfo(avifPixelFormat format, avifPixelFormatInfo * info)
 
         case AVIF_PIXEL_FORMAT_YUV400:
             info->monochrome = AVIF_TRUE;
-            // 4:0:0 is stored as monochrome and the nonexistent chroma is considered as subsampled in
-            // each dimension according to the AV1 specification. See sections 5.5.2. and 6.4.2.
+            // The nonexistent chroma is considered as subsampled in each dimension
+            // according to the AV1 specification. See sections 5.5.2. and 6.4.2.
             info->chromaShiftX = 1;
             info->chromaShiftY = 1;
             break;
--- a/src/avif.c
+++ b/src/avif.c
@@ -57,8 +57,6 @@ void avifGetPixelFormatInfo(avifPixelFormat format, avifPixelFormatInfo * info)
             break;
 
         case AVIF_PIXEL_FORMAT_YUV400:
-            info->chromaShiftX = 1;
-            info->chromaShiftY = 1;
             info->monochrome = AVIF_TRUE;
             break;
 
@@ -322,14 +320,6 @@ avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags planes)
         avifPixelFormatInfo info;
         avifGetPixelFormatInfo(image->yuvFormat, &info);
 
-        // Intermediary computation as 64 bits in case width or height is exactly UINT32_MAX.
-        const uint32_t shiftedW = (uint32_t)(((uint64_t)image->width + info.chromaShiftX) >> info.chromaShiftX);
-        const uint32_t shiftedH = (uint32_t)(((uint64_t)image->height + info.chromaShiftY) >> info.chromaShiftY);
-
-        // These are less than or equal to fullRowBytes/fullSize. No need to check overflows.
-        const size_t uvRowBytes = channelSize * shiftedW;
-        const size_t uvSize = uvRowBytes * shiftedH;
-
         image->imageOwnsYUVPlanes = AVIF_TRUE;
         if (!image->yuvPlanes[AVIF_CHAN_Y]) {
             image->yuvRowBytes[AVIF_CHAN_Y] = (uint32_t)fullRowBytes;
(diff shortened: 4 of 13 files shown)


CHANGELOG.md
```diff
@@ -60,6 +60,8 @@ List of incompatible ABI changes in this release:
 * avifEncoderSetCodecSpecificOption() now returns avifResult instead of void to
   report memory allocation failures.
 * At decoding, avifIOStats now returns the same values as at encoding.
+* avifGetPixelFormatInfo() now returns 0 for chromaShiftX and chromaShiftY when
+  the format is AVIF_PIXEL_FORMAT_YUV400. Previously it returned 1.
```
jzern, line 64: This is how they're coded in the AV1 bitstream. There is one note about MC_IDENTITY and subsampling=0:

https://aomediacodec.github.io/av1-spec/#color-config-syntax
https://aomediacodec.github.io/av1-spec/#color-config-semantics

src/avif.c
```diff
@@ -57,8 +57,6 @@ void avifGetPixelFormatInfo(avifPixelFormat format, avifPixelFormatInfo * info)
             break;
 
         case AVIF_PIXEL_FORMAT_YUV400:
-            info->chromaShiftX = 1;
-            info->chromaShiftY = 1;
```
wantehchang, line 61: As James said, this comes from the AV1 spec (page 32 and page 119). Let's add a comment here to explain this.

The rationale, I heard, is that this allows an AV1 implementation that only supports profile 0 (see seq_profile=0 on page 112 of AV1 spec) to hardcode subsampling_x=1 and subsampling_y=1. (Profile 0 only allows 4:2:0 and monochrome.)

wantehchang (changes requested):

CHANGELOG.md
y-guyon: > There is one note about MC_IDENTITY

Indeed:

This contradicts the table:

> subsampling_x	| subsampling_y |	mono_chrome |	Description
>-|-|-|-
> 1	|1|	1|	Monochrome 4:0:0

Unless MC_IDENTITY is forbidden for 4:0:0?

jzern: > Unless MC_IDENTITY is forbidden for 4:0:0?

Yes it is. You can see that in the color config syntax where monochrome is treated separately and in the case of MC_IDENTITY subsampling is forced to 0.

jzern (approved):

src/reformat.c
```diff
@@ -536,7 +536,7 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
     assert((alphaMultiplyMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP) || aPlane);
 
     for (uint32_t j = 0; j < image->height; ++j) {
-        const uint32_t uvJ = j >> state->formatInfo.chromaShiftY;
+        const uint32_t uvJ = j >> (state->formatInfo.monochrome ? 0 : state->formatInfo.chromaShiftY);
```
wantehchang, line 539: Nit: This change is not necessary because `uvJ` is only used when `hasColor` is true.

If you prefer a change, then it'd be better to use `hasColor` to match lines 567-568.

src/codec_aom.c
```diff
@@ -972,8 +972,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             bps = 16;
         }
         aomImage.bps = bps;
-        aomImage.x_chroma_shift = alpha ? 1 : codec->internal->formatInfo.chromaShiftX;
-        aomImage.y_chroma_shift = alpha ? 1 : codec->internal->formatInfo.chromaShiftY;
+        // Monochrome is handled below. Use 4:2:0 for now.
+        aomImage.x_chroma_shift = (alpha || codec->internal->formatInfo.monochrome) ? 1 : codec->internal->formatInfo.chromaShiftX;
+        aomImage.y_chroma_shift = (alpha || codec->internal->formatInfo.monochrome) ? 1 : codec->internal->formatInfo.chromaShiftY;
```
wantehchang, line 977: I think the changes to these two lines are not necessary, right?

src/scale.c
```diff
@@ -106,22 +108,15 @@ avifBool avifImageScale(avifImage * image,
...
             if (!srcYUVPlanes[i]) {
                 continue;
             }
 
             const uint32_t srcW = (i == AVIF_CHAN_Y) ? srcWidth : srcUVWidth;
             const uint32_t srcH = (i == AVIF_CHAN_Y) ? srcHeight : srcUVHeight;
-            const uint32_t dstW = (i == AVIF_CHAN_Y) ? dstWidth : dstUVWidth;
-            const uint32_t dstH = (i == AVIF_CHAN_Y) ? dstHeight : dstUVHeight;
+            const uint32_t dstW = avifImagePlaneWidth(image, AVIF_CHAN_U);
+            const uint32_t dstH = avifImagePlaneHeight(image, AVIF_CHAN_U);
```
wantehchang, line 119: IMPORTANT: Should we pass `i` instead of `AVIF_CHAN_U` as the second argument in lines 118-119?

wantehchang (commented):
Yannis: Thank you for this pull request. I'd like to read this pull request carefully. Please wait for my review.

CHANGELOG.md
wantehchang: James: The part of the color config syntax that forces subsampling to 0 in the case of MC_IDENTITY is the following:

```
        color_range = 1
        subsampling_x = 0
        subsampling_y = 0
    } else {
```

The question is: what happens if color_primaries == CP_BT_709, transfer_characteristics == TC_SRGB, matrix_coefficients == MC_IDENTITY, and mono_chrome == 1. In this case, subsampling is forced to 1 because mono_chrome is 1, and I think we will go by the following and declare the bitstream as not conforming to the spec:

apps/avifenc.c
```diff
@@ -574,8 +574,10 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
     uint32_t cellWidth, cellHeight;
     avifPixelFormatInfo formatInfo;
     avifGetPixelFormatInfo(gridSplitImage->yuvFormat, &formatInfo);
-    if (!avifGetBestCellSize("horizontally", gridSplitImage->width, gridCols, formatInfo.chromaShiftX, &cellWidth) ||
-        !avifGetBestCellSize("vertically", gridSplitImage->height, gridRows, formatInfo.chromaShiftY, &cellHeight)) {
+    const avifBool isSubsampledX = !formatInfo.monochrome && formatInfo.chromaShiftX;
+    const avifBool isSubsampledY = !formatInfo.monochrome && formatInfo.chromaShiftY;
+    if (!avifGetBestCellSize("horizontally", gridSplitImage->width, gridCols, isSubsampledX, &cellWidth) ||
+        !avifGetBestCellSize("vertically", gridSplitImage->height, gridRows, isSubsampledY, &cellHeight)) {
```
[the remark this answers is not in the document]
wantehchang: I spotted another bug fix (actually two) in tests/gtest/avifincrtest_helpers.cc. See my comment there.

Since this pull request has a lot of cleanup changes, the commit message should highlight the bug fixes. It may even make sense to move the bug fixes to a separate pull request.

include/avif/avif.h
```diff
@@ -236,6 +236,9 @@ typedef struct avifPixelFormatInfo
     int chromaShiftY;
 } avifPixelFormatInfo;
 
+// Returns the avifPixelFormatInfo depending on the avifPixelFormat.
+// info does not need to be zero-initialized. Sets all monochrome, chromaShiftX and
+// chromaShiftY fields to 1 if format is AVIF_PIXEL_FORMAT_YUV400.
```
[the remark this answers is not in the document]
wantehchang: Optional: The comment could mention that this function implement the center (second) table on page 119 of the AV1 spec.

src/codec_aom.c
```diff
@@ -1032,41 +1029,45 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
-        // Set the U plane to 0.5.
-        if (image->depth > 8) {
-            const uint16_t half = 1 << (image->depth - 1);
-            for (uint32_t j = 0; j < monoUVHeight; ++j) {
-                uint16_t * dstRow = (uint16_t *)&aomImage.planes[1][(size_t)j * aomImage.stride[1]];
-                for (uint32_t i = 0; i < monoUVWidth; ++i) {
-                    dstRow[i] = half;
+    if (monochromeRequested) {
+        if (codec->internal->monochromeEnabled) {
+            aomImage.monochrome = 1;
```
wantehchang, line 1034: I think this is a good change, but it turns out that the libaom encoder ignores this field (and also the fields we set on lines 1024-1028). What tells the libaom encoder that the image is monochrome is line 730:

```
            if (alpha || (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400)) {
                codec->internal->monochromeEnabled = AVIF_TRUE;
                cfg->monochrome = 1;  <=== THIS IS IMPORTANT
            }
```

src/codec_aom.c
```diff
@@ -972,8 +972,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             bps = 16;
         }
         aomImage.bps = bps;
-        aomImage.x_chroma_shift = alpha ? 1 : codec->internal->formatInfo.chromaShiftX;
-        aomImage.y_chroma_shift = alpha ? 1 : codec->internal->formatInfo.chromaShiftY;
+        // Monochrome is handled below. Use 4:2:0 for now.
```
wantehchang, line 975: Delete this comment?

tests/gtest/avifincrtest_helpers.cc
```diff
@@ -245,11 +245,16 @@ void EncodeRectAsIncremental(const avifImage& image, uint32_t width,
...
   avifPixelFormatInfo info;
   avifGetPixelFormatInfo(image.yuvFormat, &info);
-  const avifCropRect rect{
-      /*x=*/((image.width - width) / 2) & ~info.chromaShiftX,
-      /*y=*/((image.height - height) / 2) & ~info.chromaShiftX, width, height};
+  if (!info.monochrome) {
+    // Use even coordinates in subsampled dimensions.
+    rect.x &= ~info.chromaShiftX;
+    rect.y &= ~info.chromaShiftY;
+  }
```
wantehchang, line 257: There are two bug fixes here. 1. Check info.monochrome. 2. Use chromaShiftY with rect.y.

tests/test_cmd_grid.sh
```diff
@@ -66,6 +66,10 @@ pushd ${TMP_DIR}
   "${AVIFENC}" -s 8 "${INPUT_PNG}" --grid 2x2 -o "${ENCODED_FILE_2x2}"
   "${AVIFDEC}" "${ENCODED_FILE_2x2}" "${DECODED_FILE_2x2}"
 
+  echo "Testing monochrome grid with odd width"
```
wantehchang, line 69: Optional: add "(403)" after "odd width". Or add a comment to remind ourselves that INPUT_PNG is 403 x 302.

wantehchang (commented):
I finished my review. I suggest some changes.

src/codec_aom.c
When I first read that piece of code, I did not get why chroma_shift was set to 1 for the alpha plane, hence the hint to look below.

src/codec_aom.c
y-guyon: Technically correct. I am just applying the newly added comment `The chromaShiftX and chromaShiftY fields should be ignored when monochrome is AVIF_TRUE` for consistency.

src/codec_aom.c
wantehchang: Some of the fields of `aomImage` that we set don't really matter, in the sense that the libaom encoder actually ignores them. (I filed https://crbug.com/aomedia/3419 to document this better.) We can continue to set these fields to reasonable values, and after we have verified that the libaom encoder ignores these fields, we can re-evaluate if we want to remove these assignments.

include/avif/avif.h
```diff
@@ -236,6 +236,11 @@ typedef struct avifPixelFormatInfo
     int chromaShiftY;
 } avifPixelFormatInfo;
 
+// Returns the avifPixelFormatInfo depending on the avifPixelFormat.
+// The chromaShiftX and chromaShiftY fields should be ignored when monochrome is AVIF_TRUE.
```
wantehchang, line 240: Nit: Change this sentence to:

```
// When monochrome is AVIF_TRUE, chromaShiftX and chromaShiftY are set to 1 according to the AV1 specification but they should be ignored.
```

include/avif/avif.h
```diff
@@ -236,6 +236,11 @@ typedef struct avifPixelFormatInfo
...
+// The chromaShiftX and chromaShiftY fields are set to 1 if the format is AVIF_PIXEL_FORMAT_YUV400 to allow
+// an AV1 implementation that only supports profile 0 to hardcode subsampling_x and subsampling_y to 1.
+// This function implements the second table on page 119 of the AV1 specification version 1.0.0 with Errata 1.
```
wantehchang, line 243: Nit: Change these two sentences to the following (note the blank line and the Note):

```
//
// Note: This function implements the second table on page 119 of the AV1 specification version 1.0.0 with Errata 1.
// For monochrome 4:0:0, subsampling_x and subsampling are specified as 1 to allow
// an AV1 implementation that only supports profile 0 to hardcode subsampling_x and subsampling_y to 1.
```

Note: Another way to justify setting subsampling_x and subsampling_y to 1 for monochrome is that if we must add chroma planes (of neutral samples) to a monochrome image, then the 4:2:0 format uses the least amount of memory. For example, this is the reason behind the `avifImageCalcAOMFmt()` function in src/codec_aom.c.

src/avif.c
```diff
@@ -57,9 +57,11 @@ void avifGetPixelFormatInfo(avifPixelFormat format, avifPixelFormatInfo * info)
             break;
 
         case AVIF_PIXEL_FORMAT_YUV400:
+            info->monochrome = AVIF_TRUE;
+            // 4:0:0 is stored as monochrome and the nonexistent chroma is considered as subsampled in
```
wantehchang, line 61: Nit: "stored as monochrome" sounds a little strange. I suggest changing "4:0:0 is stored as monochrome" to "4:0:0 is monochrome", or perhaps just delete it?

src/reformat.c
```diff
@@ -536,7 +536,7 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
     assert((alphaMultiplyMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP) || aPlane);
 
     for (uint32_t j = 0; j < image->height; ++j) {
-        const uint32_t uvJ = j >> state->formatInfo.chromaShiftY;
+        const uint32_t uvJ = j >> (hasColor ? state->formatInfo.chromaShiftY : 0);
```
wantehchang, line 539: IMPORTANT: This should read (note the suggested comment):

```
        // uvJ is used only when hasColor is true.
        const uint32_t uvJ = hasColor ? (j >> state->formatInfo.chromaShiftY) : 0;
```

tests/gtest/avifscaletest.cc
```diff
@@ -0,0 +1,99 @@
...
+  const uint32_t kImageSizeLimit = std::numeric_limits<uint32_t>::max();
+  const uint32_t kImageDimensionLimit = std::numeric_limits<uint32_t>::max();
+  avifDiagnostics diag = {0};
+  ASSERT_TRUE(avifImageScale(scaled_image.get(), scaled_width, scaled_height,
+                             kImageSizeLimit, kImageDimensionLimit, &diag))
+      << diag.error;
+  EXPECT_EQ(scaled_image->width, scaled_width);
+  EXPECT_EQ(scaled_image->height, scaled_height);
+
+  ASSERT_TRUE(avifImageScale(scaled_image.get(), image->width, image->height,
```
wantehchang, line 66: Nit: It would be good to call `avifDiagnosticsClearError(&diag)` again before the second `avifImageScale()` call.

tests/gtest/avifscaletest.cc
```diff
@@ -0,0 +1,99 @@
...
+  testutil::AvifImagePtr scaled_image(avifImageCreateEmpty(), avifImageDestroy);
+  ASSERT_NE(scaled_image, nullptr);
+  ASSERT_EQ(avifImageCopy(scaled_image.get(), image.get(), AVIF_PLANES_ALL),
+            AVIF_RESULT_OK);
+
+  const uint32_t scaled_width = static_cast<uint32_t>(image->width * 0.9);
+  const uint32_t scaled_height = static_cast<uint32_t>(image->height * 2.14);
+  const uint32_t kImageSizeLimit = std::numeric_limits<uint32_t>::max();
+  const uint32_t kImageDimensionLimit = std::numeric_limits<uint32_t>::max();
+  avifDiagnostics diag = {0};
```
wantehchang, line 59: Call avifDiagnosticsClearError():

```
  avifDiagnostics diag;
  avifDiagnosticsClearError(&diag);
```

wantehchang (approved):
LGTM. I suggest some changes (mostly comments). Thank you for adding a test for avifImageScale().

src/codec_aom.c
```diff
@@ -972,6 +972,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             bps = 16;
         }
         aomImage.bps = bps;
+        // See avifImageCalcAOMFmt(). libaom doesn't have AOM_IMG_FMT_I400, so we use AOM_IMG_FMT_I420 as a substitute for monochrome.
```
wantehchang, line 975: I am sorry my previous comment about the code here wasn't clear. Please change the next two lines to what you had before:

```
        aomImage.x_chroma_shift = (alpha || codec->internal->formatInfo.monochrome) ? 1 : codec->internal->formatInfo.chromaShiftX;
        aomImage.y_chroma_shift = (alpha || codec->internal->formatInfo.monochrome) ? 1 : codec->internal->formatInfo.chromaShiftY;
```

The reason is that what you had before matches the code for alpha and AVIF_PIXEL_FORMAT_YUV400 in `avifImageCalcAOMFmt()`, so the new comment we added explains that's why we set both `x_chroma_shift` and `y_chroma_shift` to 1 if `(alpha || codec->internal->formatInfo.monochrome)` is true.

wantehchang (approved):
LGTM. Just one final adjustment (to go back to the code you had before).

Although I originally thought this change was not necessary, I verified the new code strictly follows the policy that `chromaShiftX` and `chromaShiftY` should not be used when monochrome is true. I like this. It is better than requiring the esoteric knowledge of AV1 spec's defining subsampling_x/y as 1 for monochrome.

wantehchang (approved):
LGTM. Thanks.

tests/gtest/aviftest_helpers.h
```diff
@@ -85,6 +91,8 @@ AvifImagePtr ReadImage(
         AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC,
     avifBool ignore_icc = false, avifBool ignore_exif = false,
     avifBool ignore_xmp = false);
+// Convenient wrapper around avifPNGWrite() for debugging purposes.
+bool WriteImage(const avifImage* image, const char* file_path);
```
wantehchang, line 95: I missed this new function. The function name should say the output file is a PNG file, e.g., "WritePNGImage" or "WriteImageToPNGFile".

Also, this function is not used.

y-guyon: The function could be named `WritePNG()`. I just matched `ReadImage()` for consistency.
The function is not used because it exists for debugging purposes.

wantehchang: "WritePNG" is a fine name. `ReadImage` doesn't specify the image format because it takes any image format as input (and fails if the input is of an unsupported format).

To prevent someone from removing it as an unused function, I suggest adding a comment (such as "Do not remove"), or replace "for debugging purposes" with something like "for temporary debugging code". (The temporary part is what I missed.)

PR #1309, opened 2023-03-25
avifRGBImageAllocatePixels() returns avifResult

y-guyon:
To catch out-of-memory issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ List of incompatible ABI changes in this release:
   those pointers.
 * Check the return value of avifEncoderSetCodecSpecificOption().
 * The maxThreads member was added to the avifRGBImage struct.
+* Check the return value of avifRGBImageAllocatePixels().
 
 ### Added
 * Add STATIC library target avif_internal to allow tests to access functions
@@ -60,6 +61,8 @@ List of incompatible ABI changes in this release:
 * avifEncoderSetCodecSpecificOption() now returns avifResult instead of void to
   report memory allocation failures.
 * At decoding, avifIOStats now returns the same values as at encoding.
+* avifRGBImageAllocatePixels() now returns avifResult instead of void to report
+  memory allocation failures.
 
 ## [0.11.1] - 2022-10-19
 
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -366,7 +366,10 @@ avifBool avifJPEGRead(const char * inputFilename,
         rgb.format = AVIF_RGB_FORMAT_RGB;
         rgb.chromaDownsampling = chromaDownsampling;
         rgb.depth = 8;
-        avifRGBImageAllocatePixels(&rgb);
+        if (avifRGBImageAllocatePixels(&rgb) != AVIF_RESULT_OK) {
+            fprintf(stderr, "Conversion to YUV failed: %s (out of memory)\n", inputFilename);
+            goto cleanup;
+        }
 
         int row = 0;
         while (cinfo.output_scanline < cinfo.output_height) {
@@ -557,7 +560,10 @@ avifBool avifJPEGWrite(const char * outputFilename, const avifImage * avif, int 
     rgb.format = AVIF_RGB_FORMAT_RGB;
     rgb.chromaUpsampling = chromaUpsampling;
     rgb.depth = 8;
-    avifRGBImageAllocatePixels(&rgb);
+    if (avifRGBImageAllocatePixels(&rgb) != AVIF_RESULT_OK) {
+        fprintf(stderr, "Conversion to RGB failed: %s (out of memory)\n", outputFilename);
+        goto cleanup;
+    }
     if (avifImageYUVToRGB(avif, &rgb) != AVIF_RESULT_OK) {
         fprintf(stderr, "Conversion to RGB failed: %s\n", outputFilename);
         goto cleanup;
--- a/examples/avif_example_decode_file.c
+++ b/examples/avif_example_decode_file.c
@@ -58,10 +58,15 @@ int main(int argc, char * argv[])
 
         // Alternative: set rgb.pixels and rgb.rowBytes yourself, which should match your chosen rgb.format
         // Be sure to use uint16_t* instead of uint8_t* for rgb.pixels/rgb.rowBytes if (rgb.depth > 8)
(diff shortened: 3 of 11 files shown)


tests/avifyuv.c
```diff
@@ -313,7 +319,11 @@ int main(int argc, char * argv[])
             avifRGBImage srcRGB;
             avifRGBImageSetDefaults(&srcRGB, image);
             srcRGB.depth = yuvDepth;
-            avifRGBImageAllocatePixels(&srcRGB);
+            if (avifRGBImageAllocatePixels(&srcRGB) != AVIF_RESULT_OK) {
+                avifImageDestroy(image);
+                printf("ERROR: Out of memory\n");
```
jzern, line 324: cases like this already exist, so it's fine for consistency, but I think the tests should output all errors to stderr.

jzern (approved):

src/avif.c
```diff
@@ -579,14 +579,16 @@ void avifRGBImageSetDefaults(avifRGBImage * rgb, const avifImage * image)
...
-void avifRGBImageAllocatePixels(avifRGBImage * rgb)
+avifResult avifRGBImageAllocatePixels(avifRGBImage * rgb)
 {
     if (rgb->pixels) {
         avifFree(rgb->pixels);
     }
 
     rgb->rowBytes = rgb->width * avifRGBImagePixelSize(rgb);
     rgb->pixels = avifAlloc((size_t)rgb->rowBytes * rgb->height);
+    AVIF_CHECKERR(rgb->pixels, AVIF_RESULT_OUT_OF_MEMORY);
```
wantehchang, line 590: Nit: Set `rgb->rowBytes` to 0 before returning `AVIF_RESULT_OUT_OF_MEMORY`.

A simple way to do this is to move line 588 after this line. But I think we should also set `rgb->rowBytes` to 0 at line 585.

wantehchang (commented):
Yannis: Thank you for writing this pull request. Please consider my review supplmental. James Zern's review is sufficient.

apps/shared/avifjpeg.c
```diff
@@ -366,7 +366,10 @@ avifBool avifJPEGRead(const char * inputFilename,
         rgb.format = AVIF_RGB_FORMAT_RGB;
         rgb.chromaDownsampling = chromaDownsampling;
         rgb.depth = 8;
-        avifRGBImageAllocatePixels(&rgb);
+        if (avifRGBImageAllocatePixels(&rgb) != AVIF_RESULT_OK) {
+            fprintf(stderr, "Conversion to YUV failed: %s (out of memory)\n", inputFilename);
```
wantehchang, line 370: Optional: use avifResultToString(return value) in the error message, rather than "out of memory". You do this in some files.

If you make this change, make the same change to other similar error messages.

CHANGELOG.md
```diff
@@ -60,6 +61,8 @@ List of incompatible ABI changes in this release:
 * avifEncoderSetCodecSpecificOption() now returns avifResult instead of void to
   report memory allocation failures.
 * At decoding, avifIOStats now returns the same values as at encoding.
+* avifRGBImageAllocatePixels() now returns avifResult instead of void to report
+  memory allocation failures.
```
wantehchang, line 65: Optional: Now that the function returns avifResult, it could also report invalid input parameters, so this bullet point could be written more generally. But what you have is just fine because it reflects the current implementation.

examples/avif_example_decode_file.c
```diff
@@ -58,10 +58,15 @@ int main(int argc, char * argv[])
...
+        avifResult allocationResult = avifRGBImageAllocatePixels(&rgb);
+        if (allocationResult != AVIF_RESULT_OK) {
+            fprintf(stderr, "Allocation of RGB samples failed: %s (%s)\n", inputFilename, avifResultToString(allocationResult));
+            goto cleanup;
+        }
 
-        if (avifImageYUVToRGB(decoder->image, &rgb) != AVIF_RESULT_OK) {
-            fprintf(stderr, "Conversion from YUV failed: %s\n", inputFilename);
+        avifResult convertResult = avifImageYUVToRGB(decoder->image, &rgb);
+        if (convertResult != AVIF_RESULT_OK) {
```
wantehchang, line 68: Reuse the `result` variable declared at line 25.

examples/avif_example_decode_memory.c
```diff
@@ -80,10 +80,15 @@ int main(int argc, char * argv[])
...
+        avifResult allocationResult = avifRGBImageAllocatePixels(&rgb);
+        if (allocationResult != AVIF_RESULT_OK) {
+            fprintf(stderr, "Allocation of RGB samples failed: %s (%s)\n", inputFilename, avifResultToString(allocationResult));
+            goto cleanup;
+        }
 
-        if (avifImageYUVToRGB(decoder->image, &rgb) != AVIF_RESULT_OK) {
-            fprintf(stderr, "Conversion from YUV failed: %s\n", inputFilename);
+        avifResult convertResult = avifImageYUVToRGB(decoder->image, &rgb);
+        if (convertResult != AVIF_RESULT_OK) {
```
wantehchang, line 90: Reuse the `result` variable declared at line 47.

tests/avifyuv.c
```diff
@@ -164,13 +164,19 @@ int main(int argc, char * argv[])
...
                         avifRGBImage dstRGB;
                         avifRGBImageSetDefaults(&dstRGB, image);
                         dstRGB.format = AVIF_RGB_FORMAT_RGB;
                         dstRGB.depth = rgbDepth;
-                        avifRGBImageAllocatePixels(&dstRGB);
+
+                        if ((avifRGBImageAllocatePixels(&srcRGB) != AVIF_RESULT_OK) ||
+                            (avifRGBImageAllocatePixels(&dstRGB) != AVIF_RESULT_OK)) {
+                            avifRGBImageFreePixels(&srcRGB);
+                            avifImageDestroy(image);
```
wantehchang, line 176: Optional: It takes a little thought to understand why `avifRGBImageFreePixels(&dstRGB)` is not needed. May want to rewrite this in the same way as the code below, at the cost of an extra if statement.

wantehchang (approved):
LGTM.

src/avif.c
y-guyon: I used `avifRGBImageFreePixels(rgb)` directly, it is simpler.

wantehchang (approved):
LGTM.

PR #1312, opened 2023-03-28
tests: Replace {true, false} with ::testing::Bool()

vigneshvg:


wantehchang (changes requested):
Thank you for the pull request. I request two changes.

We should only replace `Values(false, true)` or `Values(true, false)` with `Bool()`.

wantehchang (approved):
LGTM.

The commit message says "tests: Replace {true, false} with ::testing::Bool() #1312". It probably should be updated to say Values(false, true) instead, although it seems challenging to stay under 50 characters.

vigneshvg:
> LGTM.
> 
> The commit message says "tests: Replace {true, false} with ::testing::Bool() #1312". It probably should be updated to say Values(false, true) instead, although it seems challenging to stay under 50 characters.

I just replaced the true, false with false,true in the commit msg. I think that's fine.

vigneshvg:
> My previous comment was not clear. I meant that `{false, true}` is not the same as `Values(false, true)`, so the commit message could be more precise, but I am afraid that will exceed the 50-character limit.

Ah okay got it. Yeah, that would be hard to fit in 50 chars.

PR #1313, opened 2023-03-28
avifgridapitest: Use the correct pointer for comparison

vigneshvg:
The decoded image has to be compared against the source image. I think it is a typo and it originated in 90548bb1: https://github.com/AOMediaCodec/libavif/pull/1143


wantehchang (approved):
LGTM.

PR #1315, opened 2023-03-31
Implement the floorLog2() function directly

wantehchang:
The countLeadingZeros() helper function is not needed. The resulting code is essentially the same as the FloorLog2() function in the AV1 spec.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -41,20 +41,15 @@ static void writeConfigBox(avifRWStream * s, avifCodecConfigurationBox * cfg);
 // ---------------------------------------------------------------------------
 // avifSetTileConfiguration
 
-static int countLeadingZeros(uint32_t n)
-{
-    int count = 32;
-    while (n != 0) {
-        --count;
-        n >>= 1;
-    }
-    return count;
-}
-
 static int floorLog2(uint32_t n)
 {
     assert(n > 0);
-    return 31 ^ countLeadingZeros(n);
+    int count = 0;
+    while (n != 0) {
+        ++count;
+        n >>= 1;
+    }
+    return count - 1;
 }
 
 // Splits tilesLog2 into *tileDim1Log2 and *tileDim2Log2, considering the ratio of dim1 to dim2.


src/write.c
```diff
@@ -41,20 +41,15 @@ static void writeConfigBox(avifRWStream * s, avifCodecConfigurationBox * cfg);
 // ---------------------------------------------------------------------------
 // avifSetTileConfiguration
 
-static int countLeadingZeros(uint32_t n)
+static int floorLog2(uint32_t n)
 {
-    int count = 32;
+    assert(n > 0);
```
y-guyon, line 46: One comparison can be saved by using a `do {} while (n != 0)` thanks to this assertion.

y-guyon (approved):
I guess `countLeadingZeros()` was there as is because it can be easily replaced by a SIMD call, while there is little direct support for `floorLog2()`. (Just a thought based on vague SSE memories.)

wantehchang:
There are several closely-related functions: count leading zeros, floor log2, ceil log2, and bit size. The book _Hacker's Delight_ picks count leading zeros as the main function and expresses the other functions by calling count leading zeros. I think that was why I wrote it this way, but I might have been influenced by the libaom implementation since I used `31 ^` instead of the more obvious `31 -`:

```
static INLINE int get_msb(unsigned int n) {
  assert(n != 0);
  return 31 ^ __builtin_clz(n);
}
```

The compiler built-in function can be count leading zeros (e.g., GCC's `__builtin_clz`) or floor log2 (e.g., MSVC's `_BitScanReverse`).

src/write.c
wantehchang: You are right. I will keep this form because I found that it is essentially the same as the ` FloorLog2()` function in the AV1 spec. Also it is convenient to define `floorLog2(0)` as -1. This implementation supports that if we remove the assertion.

wantehchang:
I did some experiments with libaom's floor log2 function, which is named get_msb():

```
int get_msb(unsigned int n) {
  return 31 ^ __builtin_clz(n);
}
```

On x86, this function is compiled to one instruction:

```
	bsrl	%edi, %eax
	ret
```

On arm64, this function is compiled to two instructions:

```
	clz	w8, w0
	eor	w0, w8, #0x1f
	ret
```

The more obvious version, which uses subtraction:

```
int get_msb(unsigned int n) {
  return 31 - __builtin_clz(n);
}
```

gets compiled to three instructions:

```
	mov	w8, #31
	clz	w9, w0
	sub	w0, w8, w9
	ret
```

This may be why the XOR `^` operator is used, to save one instruction. XOR is commutative, so the compiler can transform the code to:

```
int get_msb(unsigned int n) {
  return __builtin_clz(n) ^ 31;
}
```

jzern:
lgtm.

PR #1320, opened 2023-04-07
Fix image8 in avifImageYUVToRGBLibYUV()

y-guyon:
The condition at the end of `avifImageYUVToRGBLibYUV()` that tests `if (!image->alphaPlane || !image->alphaRowBytes)` was potentially using a pointer to `image8` which was no longer on the stack. Fix this issue detected by sanitizers by using a temporary `image_downshifted_to_8bit` variable (replacing the previous `image8` one) and a new pointer `image8` that either points to `image` or to `image_downshifted_to_8bit`. The `image` pointer is no longer modified, `image8` is used as libyuv function args now.

Free all planes in `image_downshifted_to_8bit` and not only alpha to avoid a leak when `downshiftAlpha` is true.

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -1,6 +1,7 @@
 // Copyright 2019 Joe Drago. All rights reserved.
 // SPDX-License-Identifier: BSD-2-Clause
 
+#include "avif/avif.h"
 #include "avif/internal.h"
 
 #if !defined(AVIF_LIBYUV_ENABLED)
@@ -868,10 +869,7 @@ static avifResult avifImageDownshiftTo8bpc(const avifImage * image, avifImage * 
     image8->depth = 8;
     // downshiftAlpha will be true only if the image has an alpha plane. So it is safe to pass AVIF_PLANES_ALL here in that case.
     assert(!downshiftAlpha || image->alphaPlane);
-    avifResult result = avifImageAllocatePlanes(image8, downshiftAlpha ? AVIF_PLANES_ALL : AVIF_PLANES_YUV);
-    if (result != AVIF_RESULT_OK) {
-        return result;
-    }
+    AVIF_CHECKRES(avifImageAllocatePlanes(image8, downshiftAlpha ? AVIF_PLANES_ALL : AVIF_PLANES_YUV));
     // 16384 for 10-bit and 4096 for 12-bit.
     const int scale = 1 << (24 - image->depth);
     for (int plane = AVIF_CHAN_Y; plane <= (downshiftAlpha ? AVIF_CHAN_A : AVIF_CHAN_V); ++plane) {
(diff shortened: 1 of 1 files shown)


skal65535 (approved):

src/reformat_libyuv.c
```diff
@@ -1,6 +1,7 @@
 // Copyright 2019 Joe Drago. All rights reserved.
 // SPDX-License-Identifier: BSD-2-Clause
 
+#include "avif/avif.h"
```
wantehchang, line 4: It is not necessary to include both avif.h and internal.h. By design, internal.h includes avif.h so that internal files just need to include internal.h.

src/reformat_libyuv.c
```diff
@@ -975,85 +973,84 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb,
                                                        /*attentuate=*/0);
         *alphaReformattedWithLibYUV = AVIF_TRUE;
     } else {
-        avifImage image8;
-        avifBool inputIsHighBitDepth = image->depth > 8;
-        if (inputIsHighBitDepth) {
+        avifImage image_downshifted_to_8bit;
```
wantehchang, line 976: Nit: Rename this `imageDownshiftedTo8bit`.

src/reformat_libyuv.c
```diff
@@ -975,85 +973,84 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb,
...
-                                               image->width,
-                                               image->height,
+                                               image8->width,
+                                               image8->height,
                                                /*attenuate=*/0);
             *alphaReformattedWithLibYUV = AVIF_TRUE;
         }
-        if (inputIsHighBitDepth) {
-            avifImageFreePlanes(&image8, AVIF_PLANES_YUV);
+        if (image8 != image) {
```
wantehchang, line 1052: Nit: If we test the equivalent condition `image8 == &image_downshifted_to_8bit`, it will further reinforce why `image_downshifted_to_8bit` needs to be freed.

wantehchang (commented):
LGTM. Thank you for the fix. Vignesh or I can take care of the changes I suggested.

@vigneshvg Vignesh: FYI.

src/reformat_libyuv.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1321.

src/reformat_libyuv.c
wantehchang: There should be a way to annotate internal.h so that tools like IWYU know it can be used as a substitute for avif.h. I just took a quick look at the IWYU documentation. I think it's `IWYU pragma: export`:

https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-export

Could you test that? Thanks.

vigneshvg:
Actually, a simpler solution to this problem would just be to move the `if (!image->alphaPlane || !image->alphaRowBytes)` condition above the image8 stuff.

I will try to write a PR that does that. That makes the code slightly better without the need for the additional `image_downshifted_to_8bit`.

vigneshvg:
I have proposed an alternate fix here: https://github.com/AOMediaCodec/libavif/pull/1322

src/reformat_libyuv.c
wantehchang: I wrote https://github.com/AOMediaCodec/libavif/pull/1323.

PR #1321, opened 2023-04-07
Lint reformat_libyuv.c

y-guyon:


src/reformat_libyuv.c
```diff
@@ -973,12 +973,12 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb,
...
         *alphaReformattedWithLibYUV = AVIF_TRUE;
     } else {
-        avifImage image_downshifted_to_8bit;
+        avifImage imageDownshiftedTo8bit;
         const avifImage * image8;
         if (avifImageUsesU16(image)) {
             const avifBool downshiftAlpha = (lcf.yuvaToRgbMatrixFilter != NULL || lcf.yuvaToRgbMatrix != NULL);
-            AVIF_CHECKRES(avifImageDownshiftTo8bpc(image, &image_downshifted_to_8bit, downshiftAlpha));
-            image8 = &image_downshifted_to_8bit;
+            AVIF_CHECKRES(avifImageDownshiftTo8bpc(image, &imageDownshiftedTo8bit, downshiftAlpha));
```
wantehchang, line 980: Please change the second parameter of `avifImageDownshiftTo8bpc()` to the same name, so that the meaning of `image8` in this file doesn't diverge into two.

Optional: "imageDownshiftedTo8bit" could be shortened to "imageDownshifted", because "To8bit" is clear from the context.

wantehchang (approved):
LGTM. Thanks for the pull request. I just realized another function parameter name should be renamed after we renamed the original `image8` variable in `avifImageYUVToRGBLibYUV()`.

wantehchang (commented):
Yannis: I suggest we go with Vignesh's https://github.com/AOMediaCodec/libavif/pull/1322. Then this pull request won't be needed.

PR #1322, opened 2023-04-07
reformat_libyuv: Move alpha plane check location

vigneshvg:
The alpha plane check in avifImageYUVToRGBLibYUV potentially dereferences a pointer that is not alive. Move the condition to the beginning of the function so that this problem does not happen.

Alternative solution to
https://github.com/AOMediaCodec/libavif/pull/1320

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -1,7 +1,6 @@
 // Copyright 2019 Joe Drago. All rights reserved.
 // SPDX-License-Identifier: BSD-2-Clause
 
-#include "avif/avif.h"
 #include "avif/internal.h"
 
 #if !defined(AVIF_LIBYUV_ENABLED)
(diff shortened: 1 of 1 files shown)


src/reformat_libyuv.c
```diff
@@ -973,93 +976,87 @@ avifResult avifImageYUVToRGBLibYUV(const avifImage * image, avifRGBImage * rgb,
...
-                                               image8->height,
+                                               image->width,
+                                               image->height,
                                                /*attenuate=*/0);
             *alphaReformattedWithLibYUV = AVIF_TRUE;
         }
-        if (image8 != image) {
-            avifImageFreePlanes(&image_downshifted_to_8bit, AVIF_PLANES_ALL);
+        if (inputIsHighBitDepth) {
+            avifImageFreePlanes(&image8, AVIF_PLANES_ALL);
```
wantehchang, line 1054: Nit/optional: The end of the function could go back to the original `return (libyuvResult != 0) ? AVIF_RESULT_REFORMAT_FAILED : AVIF_RESULT_OK;` But the current if statement is fine by me!

wantehchang, line 1054: IMPORTANT: To prevent a similar bug, add `image = NULL;` after this line. Then if new code gets added that uses `image` after this point, it will get a segmentation fault.

wantehchang (approved):
LGTM.

Yannis's solution has an advantage that the meaning of `image` doesn't change throughout the function. But the current approach is also quite common; its advantage is that the same variable (`image`) is used throughout the function.  So there are pros and cons.

I suggest a change to help detect this kind of bug.

src/reformat_libyuv.c
vigneshvg: Done.

vigneshvg: Done.

vigneshvg:
@y-guyon i will go ahead and merge this PR. If you have any comments/concerns, please let me know.

wantehchang (approved):
LGTM.

PR #1324, opened 2023-04-07
Use a single decoder instance for grid images

vigneshvg:
When decoding AVIF images with grid cells, use a single decoder instance for all the cells instead of creating a decoder instance for each cell. This is okay to do since each cell must be a key frame. So they can be decoded with a single decoder instance. This change also requires each decoded cell to be copied over to the output image right away instead of being done in the end.

There are some edge cases where we will still need multiple decoder instances:
 * For animated AVIF, we will need two instances (one for the color planes and one for the alpha plane since they are both encoded as separate video sequences).
 * For grid images with multiple layers. In this case, each grid will need its own decoder instance since there would be multiple layers in each grid.
 * For grid images where the operating point of all the grids are not the same. This is a bit of a peculiar edge case that is very unlikely to happen in practice. However, the spec does not disallow this as far as i can tell. So to be safe, we will use multiple decoder instances in this case as well.

This is better because of the following reasons:
 * Eliminates the cost of creating and shutting down several decoders (and threads when threading is enabled).
 * The underlying decoder is likely to re-use its internal frame buffer and not create a new one for each cell (unless the cell dimensions vary).

On android, with 2 threads on a set of 2x2 grid images the newer implementation is about 5-10% faster when using dav1d and 25-30% faster when using libgav1.

On android, for a 5x4 grid image [1], dav1d is about 30% faster and libgav1 is about 40% faster. On desktop, both are about 10% faster.

The performance improvement gets better as the number of cells increase.

[1] https://github.com/AOMediaCodec/av1-avif/raw/master/testFiles/Microsoft/Summer_in_Tomsk_720p_5x4_grid.avif

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -646,6 +646,8 @@ static avifBool uniqueBoxSeen(uint32_t * uniqueBoxFlags, uint32_t whichFlag, con
 typedef struct avifTile
 {
     avifCodecDecodeInput * input;
+    // This may point to a codec that it owns or point to a shared codec that it does not own. In the shared case, this will
+    // point to one of avifDecoderData.codec or avifDecoderData.codecAlpha.
     struct avifCodec * codec;
     avifImage * image;
     uint32_t width;  // Either avifTrack.width or avifDecoderItem.width
@@ -780,6 +782,20 @@ typedef struct avifDecoderData
     avifTileInfo color;
     avifTileInfo alpha;
     avifDecoderSource source;
+    // When decoding AVIF images with grid, use a single decoder instance for all the tiles instead of creating a decoder instance
+    // for each tile. If that is the case, |codec| will be used by all the tiles.
+    //
+    // There are some edge cases where we will still need multiple decoder instances:
+    // * For animated AVIF with alpha, we will need two instances (one for the color planes and one for the alpha plane since they are both
+    //   encoded as separate video sequences). In this case, |codec| will be used for the color planes and |codecAlpha| will be
+    //   used for the alpha plane.
+    // * For grid images with multiple layers. In this case, each tile will need its own decoder instance since there would be
+    //   multiple layers in each tile. In this case, |codec| and |codecAlpha| are not used and each tile will have its own
+    //   decoder instance.
+    // * For grid images where the operating points of all the tiles are not the same. In this case, we each tile needs its own
+    //   decoder instance (same as above).
+    avifCodec * codec;
+    avifCodec * codecAlpha;
     uint8_t majorBrand[4];                     // From the file's ftyp, used by AVIF_DECODER_SOURCE_AUTO
     avifDiagnostics * diag;                    // Shallow copy; owned by avifDecoder
     const avifSampleTable * sourceSampleTable; // NULL unless (source == AVIF_DECODER_SOURCE_TRACKS), owned by an avifTrack
@@ -819,12 +835,23 @@ static void avifDecoderDataResetCodec(avifDecoderData * data)
             avifImageFreePlanes(tile->image, AVIF_PLANES_ALL); // forget any pointers into codec image buffers
         }
         if (tile->codec) {
-            avifCodecDestroy(tile->codec);
+            // Check if tile->codec was created separately and destroy it in that case.
+            if (tile->codec != data->codec && tile->codec != data->codecAlpha) {
+                avifCodecDestroy(tile->codec);
+            }
             tile->codec = NULL;
         }
     }
     data->color.decodedTileCount = 0;
     data->alpha.decodedTileCount = 0;
+    if (data->codec) {
+        avifCodecDestroy(data->codec);
+        data->codec = NULL;
+    }
+    if (data->codecAlpha) {
+        avifCodecDestroy(data->codecAlpha);
+        data->codecAlpha = NULL;
+    }
 }
 
 static avifTile * avifDecoderDataCreateTile(avifDecoderData * data, uint32_t width, uint32_t height, uint8_t operatingPoint)
@@ -870,7 +897,10 @@ static void avifDecoderDataClearTiles(avifDecoderData * data)
             tile->input = NULL;
         }
         if (tile->codec) {
-            avifCodecDestroy(tile->codec);
+            // Check if tile->codec was created separately and destroy it in that case.
+            if (tile->codec != data->codec && tile->codec != data->codecAlpha) {
+                avifCodecDestroy(tile->codec);
+            }
             tile->codec = NULL;
         }
         if (tile->image) {
@@ -883,6 +913,14 @@ static void avifDecoderDataClearTiles(avifDecoderData * data)
     data->color.decodedTileCount = 0;
     data->alpha.tileCount = 0;
     data->alpha.decodedTileCount = 0;
+    if (data->codec) {
+        avifCodecDestroy(data->codec);
+        data->codec = NULL;
+    }
+    if (data->codecAlpha) {
+        avifCodecDestroy(data->codecAlpha);
+        data->codecAlpha = NULL;
+    }
 }
 
 static void avifDecoderDataDestroy(avifDecoderData * data)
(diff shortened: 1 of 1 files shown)


vigneshvg:
Reviewed on gerrit here: https://aomedia-review.googlesource.com/c/libavif/+/172921

src/read.c
```diff
@@ -780,6 +782,20 @@ typedef struct avifDecoderData
     avifTileInfo color;
     avifTileInfo alpha;
     avifDecoderSource source;
+    // When decoding AVIF images with grid cells, use a single decoder instance for all the cells instead of creating a decoder
+    // instance for each cell. If that is the case, |codec| will be used by all the tiles.
```
wantehchang, line 786: Nit: Use "tiles" (instead of "cells") consistently in this paragraph.

src/read.c
```diff
@@ -780,6 +782,20 @@ typedef struct avifDecoderData
...
+    //
+    // There are some edge cases where we will still need multiple decoder instances:
+    // * For animated AVIF, we will need two instances (one for the color planes and one for the alpha plane since they are both
```
wantehchang, line 789: Nit: add "with alpha" after "animated AVIF".

src/read.c
```diff
@@ -780,6 +782,20 @@ typedef struct avifDecoderData
...
+    //   encoded as separate video sequences). In this case, |codec| will be used for the color planes and |codecAlpha| will be
+    //   used for the alpha plane.
+    // * For grid images with multiple layers. In this case, each grid will need its own decoder instance since there would be
```
wantehchang, line 792: each grid => each tile ?

src/read.c
```diff
@@ -780,6 +782,20 @@ typedef struct avifDecoderData
...
+    //   multiple layers in each grid. In this case, |codec| and |codecAlpha| are not used and each tile will have it's own
+    //   decoder instance.
+    // * For grid images where the operating point of all the grids are not the same. In this case, we each grid needs its own
```
wantehchang, line 795: point => points

all the grids => all the tiles

each grid => each tile

src/read.c
```diff
@@ -1285,134 +1323,116 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
...
-// Checks the grid consistency and copies the pixels from the tiles to the
-// dstImage. Only the freshly decoded tiles are considered, skipping the already
-// copied or not-yet-decoded tiles.
-static avifBool avifDecoderDataFillImageGrid(avifDecoderData * data,
-                                             avifImage * dstImage,
-                                             const avifTileInfo * info,
-                                             unsigned int oldDecodedTileCount,
-                                             avifBool alpha)
+// Allocates the dstImage based on the grid image requirements. Also verifies some spec compliance rules for grids.
+static avifResult avifDecoderDataAllocateGridImage(avifDecoderData * data, const avifTileInfo * info, avifImage * dstImage)
```
wantehchang, line 1327: Nit/optional: Add "Planes" to the end of the function name

src/read.c
```diff
@@ -1285,134 +1323,116 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
...
-            (tile->image->depth != firstTile->image->depth) || (tile->image->yuvFormat != firstTile->image->yuvFormat) ||
-            (tile->image->yuvRange != firstTile->image->yuvRange) || (uvPresent != firstTileUVPresent) ||
-            (tile->image->colorPrimaries != firstTile->image->colorPrimaries) ||
-            (tile->image->transferCharacteristics != firstTile->image->transferCharacteristics) ||
-            (tile->image->matrixCoefficients != firstTile->image->matrixCoefficients)) {
-            avifDiagnosticsPrintf(data->diag, "Grid image contains mismatched tiles");
-            return AVIF_FALSE;
-        }
-    }
+    avifTile * tile = &data->tiles.tile[info->firstTileIndex];
```
wantehchang, line 1330: Add `const` before `avifTile`

src/read.c
```diff
@@ -1285,134 +1323,116 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
...
         avifDiagnosticsPrintf(data->diag, "Image allocation failure");
-        return AVIF_FALSE;
+        return AVIF_RESULT_OUT_OF_MEMORY;
     }
+    return AVIF_RESULT_OK;
+}
 
-    avifPixelFormatInfo formatInfo;
-    avifGetPixelFormatInfo(firstTile->image->yuvFormat, &formatInfo);
+// Copies over the pixels from the tile into dstImage. Also verifies that the necessary properties of the tile matches that of the
```
wantehchang, line 1388: Nit: "necessary" sounds a little strange here. Maybe "relevant"?

that of => those of

src/read.c
```diff
@@ -1285,134 +1323,116 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
...
+// first tile.
```
wantehchang, line 1389: Nit: Reverse these two sentences to match the order in which the function does these two things.

For example,

> After verifying that the necessary properties of the tile matches that of the first tile, copies over the pixels from the tile into dstImage.

src/read.c
```diff
@@ -1285,134 +1323,116 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
...
+static avifBool avifDecoderDataCopyTileToImage(avifDecoderData * data,
+                                               const avifTileInfo * info,
+                                               avifImage * dstImage,
+                                               avifTile * tile,
```
wantehchang, line 1393: Declare `tile` and `firstTile` (line 1397) as const

src/read.c
```diff
@@ -3337,9 +3357,72 @@ avifResult avifDecoderParse(avifDecoder * decoder)
     return avifDecoderReset(decoder);
 }
 
-static avifCodec * avifCodecCreateInternal(avifCodecChoice choice)
+static avifCodec * avifCodecCreateInternal(avifCodecChoice choice, avifTile * tile, avifDiagnostics * diag)
```
wantehchang, line 3360: Declare `tile` as const. Alternatively, pass `operatingPoint` and `allLayers` instead.

src/read.c
```diff
@@ -3337,9 +3357,72 @@ avifResult avifDecoderParse(avifDecoder * decoder)
...
+    codec->allLayers = tile->input->allLayers;
+    return codec;
+}
+
+static avifBool avifTilesCanBeDecodedWithSameAV1Decoder(avifDecoderData * data)
+{
+    const uint8_t firstTileOperatingPoint = data->tiles.tile[0].operatingPoint;
+    const avifBool firstTileAllLayers = data->tiles.tile[0].input->allLayers;
+    for (unsigned int i = 1; i < data->tiles.count; ++i) {
+        avifTile * tile = &data->tiles.tile[i];
```
wantehchang, line 3375: Declare `tile` as const

src/read.c
```diff
@@ -3337,9 +3357,72 @@ avifResult avifDecoderParse(avifDecoder * decoder)
...
+        // met:
+        //   - The image must have exactly one layer (i.e.) decoder->imageCount == 1.
+        //   - All the tiles must have the same operating point (because AV1 decoders take operating point once at initialization
+        //     and do not allow it to be changed later).
+        //   - All the tiles must have the same value for allLayers (because AV1 decoders take allLayers once at initialization
+        //     and do not allow it to be changed later).
+        // Otherwise, we will use |tiles.count| decoder instances (one instance for each tile).
+        avifBool canUseSingleDecoder = (data->tiles.count == 1) ||
+                                       (decoder->imageCount == 1 && avifTilesCanBeDecodedWithSameAV1Decoder(data));
+        data->codecAlpha = NULL;
```
wantehchang, line 3409: Omit this, because this is done by the `avifDecoderDataResetCodec()` call.

src/read.c
```diff
@@ -3337,9 +3357,72 @@ avifResult avifDecoderParse(avifDecoder * decoder)
...
+        if (canUseSingleDecoder) {
+            data->codec = avifCodecCreateInternal(decoder->codecChoice, &decoder->data->tiles.tile[0], &decoder->diag);
+            AVIF_CHECKERR(data->codec, AVIF_RESULT_OUT_OF_MEMORY);
+            for (unsigned int i = 0; i < decoder->data->tiles.count; ++i) {
+                decoder->data->tiles.tile[i].codec = data->codec;
+            }
+        } else {
+            data->codec = NULL;
```
wantehchang, line 3417: Omit this.

src/read.c
```diff
@@ -3337,9 +3357,72 @@ avifResult avifDecoderParse(avifDecoder * decoder)
...
+            for (unsigned int i = 0; i < decoder->data->tiles.count; ++i) {
+                avifTile * tile = &decoder->data->tiles.tile[i];
```
wantehchang, line 3419: Declare `tile` as const

src/read.c
```diff
@@ -3922,17 +4006,14 @@ avifResult avifDecoderOutputDecodedTiles(avifDecoder * decoder, const avifTileIn
         // Nothing changed.
         return AVIF_RESULT_OK;
     }
-    // There is at least one newly decoded tile.
-    assert(info->decodedTileCount > oldDecodedTileCount);
-
-    // Grid path.
+    // For grid images, the output image is populated in avifDecoderDecodeTiles.
```
wantehchang, line 4009: IMPORTANT: The DecodeTiles() and OutputDecodedTiles() functions no longer do what their names imply after this change. This should be a TODO for future cleanup. It seems that these two functions should be merged.

wantehchang (approved):
LGTM. I read the new code carefully. What I didn't do was to compare the new code with the original code to make sure no errors were introduced when the code was moved. (A lot of code was moved in this pull request.)

src/read.c
```diff
@@ -780,6 +782,20 @@ typedef struct avifDecoderData
...
     avifDecoderSource source;
+    // When decoding AVIF images with grid cells, use a single decoder instance for all the cells instead of creating a decoder
+    // instance for each cell. If that is the case, |codec| will be used by all the tiles.
+    //
+    // There are some edge cases where we will still need multiple decoder instances:
+    // * For animated AVIF, we will need two instances (one for the color planes and one for the alpha plane since they are both
+    //   encoded as separate video sequences). In this case, |codec| will be used for the color planes and |codecAlpha| will be
+    //   used for the alpha plane.
+    // * For grid images with multiple layers. In this case, each grid will need its own decoder instance since there would be
+    //   multiple layers in each grid. In this case, |codec| and |codecAlpha| are not used and each tile will have it's own
```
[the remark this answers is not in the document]
vigneshvg: 1 & 3 Done.

2) I believe used is the right word, since in that case `codec` and `codecAlpha` are neither created nor used. Only `tile->codec` is used.

src/read.c

src/read.c

src/read.c

src/read.c

src/read.c

src/read.c

src/read.c

I chose to pass `tile` instead of specific parameters since some of the callers pass stuff like `&decoder->data->tile.tile[1]`. It would mean converting this into a local variable in the caller since it's too cumbersome to repeat this in the caller.

src/read.c

src/read.c

src/read.c

src/read.c
vigneshvg: This cannot be const since we are setting tile->codec in the next line.

src/read.c
vigneshvg: Added TODO In line 4090. I will look into this in a follow-up.

src/read.c
```diff
@@ -780,6 +782,20 @@ typedef struct avifDecoderData
...
     avifDecoderSource source;
+    // When decoding AVIF images with grid, use a single decoder instance for all the tiles instead of creating a decoder instance
+    // for each tile. If that is the case, |codec| will be used by all the tiles.
+    //
+    // There are some edge cases where we will still need multiple decoder instances:
+    // * For animated AVIF with alpha, we will need two instances (one for the color planes and one for the alpha plane since they are both
+    //   encoded as separate video sequences). In this case, |codec| will be used for the color planes and |codecAlpha| will be
+    //   used for the alpha plane.
+    // * For grid images with multiple layers. In this case, each tile will need its own decoder instance since there would be
+    //   multiple layers in each grid. In this case, |codec| and |codecAlpha| are not used and each tile will have its own
```
wantehchang, line 793: each grid => each tile ?

src/read.c
```diff
@@ -1285,134 +1323,116 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
...
         avifDiagnosticsPrintf(data->diag, "Image allocation failure");
-        return AVIF_FALSE;
+        return AVIF_RESULT_OUT_OF_MEMORY;
     }
+    return AVIF_RESULT_OK;
+}
 
-    avifPixelFormatInfo formatInfo;
-    avifGetPixelFormatInfo(firstTile->image->yuvFormat, &formatInfo);
+// After verifying that the relevant properties of the tile matches those of the first tile, copies over the pixels from the tile
```
wantehchang, line 1388: matches => match

src/read.c
```diff
@@ -4004,6 +4088,9 @@ avifResult avifDecoderNextImage(avifDecoder * decoder)
     // Decode all available color tiles now, then all available alpha tiles.
     const unsigned int oldDecodedColorTileCount = decoder->data->color.decodedTileCount;
     const unsigned int oldDecodedAlphaTileCount = decoder->data->alpha.decodedTileCount;
+    // TODO: For image with grid, avifDecoderDecodeTiles also populates the output. So this function and
```
wantehchang, line 4091: Nit: Add your user id. See https://google.github.io/styleguide/cppguide.html#TODO_Comments

wantehchang (approved):
LGTM.

src/read.c

src/read.c

src/read.c

vigneshvg:
Thanks for the review @wantehchang.

wantehchang (approved):
LGTM!

src/read.c
```diff
@@ -1285,134 +1323,116 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
...
+                                               avifImage * dstImage,
+                                               const avifTile * tile,
+                                               unsigned int tileIndex)
+{
+    const avifImageGrid * grid = &info->grid;
+    const avifTile * firstTile = &data->tiles.tile[info->firstTileIndex];
+    if (tile != firstTile) {
+        avifBool firstTileUVPresent = (firstTile->image->yuvPlanes[AVIF_CHAN_U] && firstTile->image->yuvPlanes[AVIF_CHAN_V]);
+
+        // Check for tile consistency. All tiles in a grid image should match the first tile in the properties checked below.
```
jzern, line 1401: This and the next line could be swapped to bring `firstTileUVPresent` and the related `uvPresent` together and allow the comment to describe the conditionals.

jzern (approved):

src/read.c
vigneshvg: Done.

PR #1334, opened 2023-04-12
avifenc: Add an experimental --progressive option

wantehchang:
If --progressive is specified, avifenc will encode a progressive, layered image. The current implementation is hardcoded to use two layers.

The --progressive option is marked as experimental because it doesn't work properly even though it produces a correctly encoded layered image.

Refactor avifEncodeImagesFixedQuality(). Move the big while loop in avifEncodeImagesFixedQuality() to a new avifEncodeRestOfImageSequence() function. Then change avifEncodeImagesFixedQuality() to call either avifEncodeRestOfImageSequence() or avifEncodeRestOfLayeredImage() depending on whether we are encoding an image sequence or a progressive, layered image.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -1520,6 +1520,7 @@ int main(int argc, char * argv[])
     // with progressive encoding.
     if (settings.progressive && ((settings.quality < 10) || (settings.qualityAlpha < 10))) {
         settings.progressive = AVIF_FALSE;
+        printf("The --progressive option was ignored because the quality is below 10.\n");
     }
 
     stdinFile.filename = "(stdin)";
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -138,6 +138,7 @@ static void syntax(void)
            AVIF_QUANTIZER_WORST_QUALITY,
            AVIF_QUANTIZER_LOSSLESS);
     printf("    --target-size S                   : Set target file size in bytes (up to 7 times slower)\n");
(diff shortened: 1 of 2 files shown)


wantehchang:
For the convenience of your review, I created a copy of this pull request in Gerrit: https://aomedia-review.googlesource.com/c/libavif/+/173342

wantehchang:
Please review. The second commit in this pull request addressed two rounds of review at https://aomedia-review.googlesource.com/c/libavif/+/173342.

apps/avifenc.c
```diff
@@ -1432,6 +1513,14 @@ int main(int argc, char * argv[])
             }
         }
     }
+    assert(settings.quality != INVALID_QUALITY);
+    assert(settings.qualityAlpha != INVALID_QUALITY);
+    // In progressive encoding we use a very low quality (2) for the base layer to ensure a small
+    // encoded size. If the target quality is close to the quality of the base layer, don't bother
+    // with progressive encoding.
+    if (settings.progressive && ((settings.quality < 10) || (settings.qualityAlpha < 10))) {
+        settings.progressive = AVIF_FALSE;
```
y-guyon, line 1522: Suggestion:

```c
printf("The --progressive flag was ignored because the quality is below 10\n");
```

y-guyon (approved):

apps/avifenc.c
wantehchang: Done.

PR #1335, opened 2023-04-12
codec_aom.c: Update min/max quantizers correctly

wantehchang:
cfg->rc_min_quantizer and cfg->rc_max_quantizer should be updated based on encoder->minQuantizer(Alpha) and encoder->maxQuantizer(Alpha) rather than cfg->rc_min_quantizer and cfg->rc_max_quantizer, because the `quantizer` parameter and encoder->minQuantizer(Alpha) and encoder->maxQuantizer(Alpha) are updated together.


wantehchang:
@tongyuantongyu @y-guyon FYI.

vigneshvg (approved):

jzern (approved):

PR #1337, opened 2023-04-12
codec_aom: Set g_limit for progressive encoding

wantehchang:
In progressive encoding, the number of frames to encode is the number of layers. Set cfg->g_limit to that number.

It's not clear if setting cfg->g_limit will actually help the libaom encoder, but more information cannot hurt.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -707,6 +707,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             cfg->kf_max_dist = 0;
         }
         if (encoder->extraLayerCount > 0) {
+            cfg->g_limit = encoder->extraLayerCount + 1;
             // For layered image, disable lagged encoding to always get output
             // frame for each input frame.
             cfg->g_lag_in_frames = 0;


wantehchang:
@tongyuantongyu @y-guyon FYI.

vigneshvg (approved):
> It's not clear if setting cfg->g_limit will actually help the libaom encoder, but more information cannot hurt.

as far as i can remember, libaom only uses this value for deciding whether to use `reduced_still_picture_header` or not.

jzern:
> > It's not clear if setting cfg->g_limit will actually help the libaom encoder, but more information cannot hurt.
> 
> as far as i can remember, libaom only uses this value for deciding whether to use `reduced_still_picture_header` or not.

https://aomedia.googlesource.com/aom/+/refs/tags/v3.6.0/aom/aom_encoder.h#435

Aside from setting `reduced_still_picture_header` when it's equal to 1, this ends up in [`frames_left`](https://aomedia.googlesource.com/aom/+/refs/tags/v3.6.0/av1/encoder/encoder.h#2610), which only affects 2 pass encoding.

wantehchang:
Thank you both for analyzing the effects of setting `g_limit`. libavif uses one-pass encoding only, so setting g_limit for progressive encoding is a no-op.

PR #1338, opened 2023-04-13
avifenc.c: Fix typo in "Encoding frame 1" message

wantehchang:
The first frame is frame 0, not frame 1.

Note: I use the %d format specifier and pass 0 as the argument so that the compiler can pool this printf string with the string of the other printf statement.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -726,7 +726,8 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
 
         uint64_t firstDurationInTimescales = firstFile->duration ? firstFile->duration : settings->outputTiming.duration;
         if (input->useStdin || (input->filesCount > 1)) {
-            printf(" * Encoding frame 1 [%" PRIu64 "/%" PRIu64 " ts]: %s\n",
+            printf(" * Encoding frame %d [%" PRIu64 "/%" PRIu64 " ts]: %s\n",
+                   0,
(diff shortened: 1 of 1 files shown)


apps/avifenc.c
```diff
@@ -726,7 +726,8 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
 
         uint64_t firstDurationInTimescales = firstFile->duration ? firstFile->duration : settings->outputTiming.duration;
         if (input->useStdin || (input->filesCount > 1)) {
-            printf(" * Encoding frame 1 [%" PRIu64 "/%" PRIu64 " ts]: %s\n",
+            printf(" * Encoding frame %d [%" PRIu64 "/%" PRIu64 " ts]: %s\n",
+                   0,
```
wantehchang, line 730: The other printf statement is at line 747. Since the two string literals are the same, the compiler can pool them.

jzern: I understand. Having a literal may look a little strange and could get cleaned up at some point.

jzern (approved):

y-guyon (approved):

wantehchang:
Yes, getting cleaned up is a risk. Since this is a minor optimization it doesn't matter either way.

Perhaps we should create a helper function for printing this message. That will ensure the messages for the first frame and subsequent frames are printed in the same way.

PR #1343, opened 2023-04-14
Do not use single decoder for non-grid images with alpha

vigneshvg:
In this case, we will steal the planes from the decoder. So invoking the same decoder for the alpha plane after having used it for the color planes will invalidate the color planes.

This bug was introduced in 39682be.


vigneshvg:
I can probably re-work the whole logic behind canUseSingleAV1Decoder into the function. i will do that in a follow-up since this fixes the bug for now.

src/read.c
```diff
@@ -3368,6 +3368,11 @@ static avifCodec * avifCodecCreateInternal(avifCodecChoice choice, const avifTil
 
 static avifBool avifTilesCanBeDecodedWithSameAV1Decoder(avifDecoderData * data)
 {
+    if (data->color.tileCount == 1 && data->alpha.tileCount == 1) {
+        // Single tile image with alpha plane. In this we each tile needs its own decoder since the planes will be "stolen".
```
y-guyon, line 3372: > In this we each tile

y-guyon (approved):
I wonder what is costlier between copying the alpha plane and not reusing the decoder instance.

vigneshvg:
> I wonder what is costlier between copying the alpha plane and not reusing the decoder instance.

I think `allocation+copy` would be more expensive for large images vs `allocation+stealing`. I will try it out to see if that is indeed the case.

The main benefit of re-using the decoder instance is to avoid repeated allocations (which can be avoided in the grid case). In the single image with alpha case, we have to allocate those planes anyway, so i don't think it will make a big difference when we re-use the decoder instance.

y-guyon (commented):
Should `decoder->image` be freed whenever `aomCodecDestroyInternal()` is called, unless `image->imageOwnsYUVPlanes` is true?

vigneshvg:
> Should `decoder->image` be freed whenever `aomCodecDestroyInternal()` is called, unless `image->imageOwnsYUVPlanes` is true?

Sorry i merged the change before i saw this comment.

I think it is already being done in `avifDecoderCleanUp` which calls `avifDecoderDataDestroy` first followed by `avifImageDestroy`.

y-guyon:
> I think it is already being done in `avifDecoderCleanip` which calls `avifDecoderDataDestroy` first followed by `avifImageDestroy` on `decoder->image`.

Yes, and `avifDecoderDataResetCodec()` calls `avifImageFreePlanes()` before `avifCodecDestroy()` too, but on `tile->image`, not directly on `decoder->image`. So I guess `avifDecoderNthImage()` failing on the second `avifDecoderNextImage()` call *could* output invalidated YUV planes (but upon error so not that bad). I have not checked that pattern thoroughly, it is just an hypothesis.

`avifDecoderReset()` calls `avifDecoderDataClearTiles()` which exhibits the same pattern: `avifCodecDestroy()` without freeing `decoder->image()`. Fortunately `avifImageDestroy(decoder->image)` is called immediately after.

wantehchang:
Vignesh, Yannis: I like the discussion. It's important to audit and test the relevant code carefully. See also a simple stealer tracking I wrote: https://github.com/AOMediaCodec/libavif/pull/1347.

PR #1344, opened 2023-04-14
Add avifbasictest

y-guyon:
Simple test that can be used as a demo or as a convenient starter template for other tests. Also useful for debugging untested or general issues.

### DIFF
--- a/tests/gtest/avifbasictest.cc
+++ b/tests/gtest/avifbasictest.cc
@@ -1,0 +1,44 @@
+// Copyright 2023 Google LLC
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "avif/avif.h"
+#include "aviftest_helpers.h"
+#include "gtest/gtest.h"
+
+namespace libavif {
+namespace {
+
+TEST(BasicTest, EncodeDecode) {
+  testutil::AvifImagePtr image =
+      testutil::CreateImage(/*width=*/12, /*height=*/34, /*depth=*/8,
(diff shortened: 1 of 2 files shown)


tests/gtest/avifbasictest.cc
```diff
@@ -0,0 +1,44 @@
...
+
+  // Verify that the input and decoded images are close.
+  ASSERT_GT(testutil::GetPsnr(*image, *decoded), 40.0);
+
+  // Uncomment the following to save the encoded image as an AVIF file.
+  //  std::ofstream("/tmp/avifbasictest.avif", std::ios::binary)
+  //      .write(reinterpret_cast<char*>(encoded.data), encoded.size);
+
+  // Uncomment the following to save the decoded image as a PNG file.
+  //  testutil::WriteImage(decoded.get(), "/tmp/avifbasictest.png");
```
wantehchang, line 40: Previously I suggested the function's name should contain "PNG" because that's the output format. Now that I have seen a call, it's clear that the function could be enhanced to infer the output format from the file name, so the "WriteImage" name is fine.

It would be good to add a "Do not remove" comment to the function's declaration to defend against cleanup changes like our internal CL 524234597.

wantehchang (approved):
LGTM.

tests/gtest/avifbasictest.cc
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1350

PR #1346, opened 2023-04-14
avifDecoderDecodeTiles Use `tile` in non-grid path

wantehchang:
The non-grid path of avifDecoderDecodeTiles() uses the expression
    decoder->data->tiles.tile[info->firstTileIndex]

I believe that is the variable `tile` (dereferenced). This requires tileIndex == 0, so add an assertion for that. That assertion can be removed in the future.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -4003,7 +4003,8 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder, uint32_t nextIma
         } else {
             // Non-grid path. Just steal the planes from the only "tile".
             assert(info->tileCount == 1);
-            avifImage * src = decoder->data->tiles.tile[info->firstTileIndex].image;
+            assert(tileIndex == 0);
+            avifImage * src = tile->image;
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -4002,7 +4002,8 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder, uint32_t nextIma
         } else {
             // Non-grid path. Just steal the planes from the only "tile".
             assert(info->tileCount == 1);
-            avifImage * src = decoder->data->tiles.tile[info->firstTileIndex].image;
+            assert(tileIndex == 0);
+            avifImage * src = tile->image;
```
wantehchang, line 4006: At line 3952, the variable `tile` is defined as follows:

```
        avifTile * tile = &decoder->data->tiles.tile[info->firstTileIndex + tileIndex];
```

So if `tileIndex == 0`, then  `decoder->data->tiles.tile[info->firstTileIndex]` is `*tile`.

y-guyon (approved):

wantehchang (commented):
Thanks for the review!

PR #1347, opened 2023-04-14
Track which images steal planes from codecs

wantehchang:
When we call codec->getNextImage() or destroy a codec, the planes stolen from the codec will become invalid. Track which images steal planes from a codec, and set their planes to NULL before we call codec->getNextImage() or after we destroy the codec.

This change would have caused avifgridapitest and avifincrtest to fail and caught b/278249578.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1397,12 +1397,9 @@ static avifBool avifDecoderDataCopyTileToImage(avifDecoderData * data,
     const avifTile * firstTile = &data->tiles.tile[info->firstTileIndex];
     if (tile != firstTile) {
         // Check for tile consistency. All tiles in a grid image should match the first tile in the properties checked below.
-        avifBool uvPresent = (tile->image->yuvPlanes[AVIF_CHAN_U] && tile->image->yuvPlanes[AVIF_CHAN_V]);
-        avifBool firstTileUVPresent = (firstTile->image->yuvPlanes[AVIF_CHAN_U] && firstTile->image->yuvPlanes[AVIF_CHAN_V]);
         if ((tile->image->width != firstTile->image->width) || (tile->image->height != firstTile->image->height) ||
             (tile->image->depth != firstTile->image->depth) || (tile->image->yuvFormat != firstTile->image->yuvFormat) ||
-            (tile->image->yuvRange != firstTile->image->yuvRange) || (uvPresent != firstTileUVPresent) ||
-            (tile->image->colorPrimaries != firstTile->image->colorPrimaries) ||
+            (tile->image->yuvRange != firstTile->image->yuvRange) || (tile->image->colorPrimaries != firstTile->image->colorPrimaries) ||
             (tile->image->transferCharacteristics != firstTile->image->transferCharacteristics) ||
             (tile->image->matrixCoefficients != firstTile->image->matrixCoefficients)) {
             avifDiagnosticsPrintf(data->diag, "Grid image contains mismatched tiles");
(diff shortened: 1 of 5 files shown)


src/read.c
```diff
@@ -1397,12 +1397,9 @@ static avifBool avifDecoderDataCopyTileToImage(avifDecoderData * data,
     const avifTile * firstTile = &data->tiles.tile[info->firstTileIndex];
     if (tile != firstTile) {
         // Check for tile consistency. All tiles in a grid image should match the first tile in the properties checked below.
-        avifBool uvPresent = (tile->image->yuvPlanes[AVIF_CHAN_U] && tile->image->yuvPlanes[AVIF_CHAN_V]);
-        avifBool firstTileUVPresent = (firstTile->image->yuvPlanes[AVIF_CHAN_U] && firstTile->image->yuvPlanes[AVIF_CHAN_V]);
         if ((tile->image->width != firstTile->image->width) || (tile->image->height != firstTile->image->height) ||
             (tile->image->depth != firstTile->image->depth) || (tile->image->yuvFormat != firstTile->image->yuvFormat) ||
-            (tile->image->yuvRange != firstTile->image->yuvRange) || (uvPresent != firstTileUVPresent) ||
-            (tile->image->colorPrimaries != firstTile->image->colorPrimaries) ||
+            (tile->image->yuvRange != firstTile->image->yuvRange) || (tile->image->colorPrimaries != firstTile->image->colorPrimaries) ||
```
wantehchang, line 1402: This change is necessary because `firstTile->image->yuvPlanes[AVIF_CHAN_U]` and `firstTile->image->yuvPlanes[AVIF_CHAN_V]` are already invalid when a shared codec decodes the second tile of a grid image. The original code only tests whether these two pointers are null or not and doesn't dereference them, but it's still bad to use invalid pointers.

With the stealer tracking, `firstTile->image->yuvPlanes[AVIF_CHAN_U]` and `firstTile->image->yuvPlanes[AVIF_CHAN_V]` become null pointers when a shared codec decodes the second tile of a grid image. So we can no longer test `(uvPresent != firstTileUVPresent)`. But I think the `(tile->image->yuvFormat != firstTile->image->yuvFormat)` test is equivalent.

wantehchang (commented):
Vignesh, Yannis: I think something like this is worthwhile.

Note, however, that stealer tracking may have its own memory errors, so this needs to be reviewed carefully. But this pull request shows we should audit the relevant code carefully because the decoder sharing breaks several assumptions of the original design.

wantehchang:
@vigneshvg @y-guyon I marked this pull request as a draft and canceled review requests. The stealer tracking code itself could introduce memory errors. I am avoiding it by verifying that all the images that may be stealers outlive the codecs. But this is brittle.

I am wondering if we should go after the root cause -- we steal planes from the codecs. For example, we can register frame buffer callbacks with the AV1 decoders so that we allocate and own the output frames. Obviously copying the planes is another solution (which Yannis mentioned in https://github.com/AOMediaCodec/libavif/pull/1343).

It also seems that in the current code, we don't really need `tile->codec` as a struct member. We may be able to replace `tile->codec` with a local variable. If we can do that, then I think this image local variable, even if it steals planes from the codec, will immediately give up the stolen planes or pass them on to `decoder->image`. Then `decoder->image` will be the only stealer and then we don't need to do stealer tracking, and erasing the dangling pointers held by the stealer will be simple.

I will move the `avifDecoderDataCopyTileToImage()` change in this pull request to a separate pull request.

y-guyon:
> The stealer tracking code itself could introduce memory errors.

I am not a big fan of custom memory checking solutions in the library itself. I guess C++ automatic memory management would solve most of these issues and save a lot of time, but it would require moving most of libavif to C++ and adding a C++ interface to AV1 codecs too, so it is not worth it.

This issue could have been prevented with sufficient testing and fuzzing coverage with sanitizers enabled. I suggest we put more effort into that, such as https://github.com/AOMediaCodec/libavif/pull/1351.

wantehchang:
Abandoned the pull request.

PR #1348, opened 2023-04-15
Don't test potentially dangling pointers

wantehchang:
In avifDecoderDataCopyTileToImage(), if a shared codec is used to decode a grid image, the yuvPlanes pointers of firstTile->image will become invalid when the shared codec decodes the second tile. Although we only test if the pointers are null and do not dereference them, it is best to avoid even testing them. Also, the yuvFormat check should cover this check.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1397,12 +1397,9 @@ static avifBool avifDecoderDataCopyTileToImage(avifDecoderData * data,
     const avifTile * firstTile = &data->tiles.tile[info->firstTileIndex];
     if (tile != firstTile) {
         // Check for tile consistency. All tiles in a grid image should match the first tile in the properties checked below.
-        avifBool uvPresent = (tile->image->yuvPlanes[AVIF_CHAN_U] && tile->image->yuvPlanes[AVIF_CHAN_V]);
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -1397,12 +1397,9 @@ static avifBool avifDecoderDataCopyTileToImage(avifDecoderData * data,
     const avifTile * firstTile = &data->tiles.tile[info->firstTileIndex];
     if (tile != firstTile) {
         // Check for tile consistency. All tiles in a grid image should match the first tile in the properties checked below.
-        avifBool uvPresent = (tile->image->yuvPlanes[AVIF_CHAN_U] && tile->image->yuvPlanes[AVIF_CHAN_V]);
-        avifBool firstTileUVPresent = (firstTile->image->yuvPlanes[AVIF_CHAN_U] && firstTile->image->yuvPlanes[AVIF_CHAN_V]);
```
wantehchang, line 1401: I found this issue with the stealer tracking code in https://github.com/AOMediaCodec/libavif/pull/1347,

y-guyon (approved):

PR #1349, opened 2023-04-16
Set `dstDepth` when input is `stdin`

sshockwave:
...otherwise the sanity checker would report
```
WARNING: [--lossless] Input depth (0) does not match output depth (8). Output might not be lossless.
```
in `--lossless` mode and `--stdin` input. Also see [the correct handling when file name is supplied](https://github.com/AOMediaCodec/libavif/blob/f8eaf32ba85828528c41f97e78da2d8f24d1f2b8/apps/shared/avifutil.c#L259).


y-guyon:
Hi sshockwave, thanks for your contribution.

PR #1350, opened 2023-04-17
Mark WriteImage() as Do not remove

y-guyon:
Only PNG is handled by WriteImage() for now so only save to file paths ending with ".png".
Add coverage of WriteImage() in avifimagetest.cc.

### DIFF
--- a/tests/gtest/aviftest_helpers.cc
+++ b/tests/gtest/aviftest_helpers.cc
@@ -8,6 +8,7 @@
 #include <cmath>
 #include <cstdint>
 #include <cstdlib>
+#include <cstring>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


tests/gtest/aviftest_helpers.cc
```diff
@@ -340,9 +341,15 @@ AvifImagePtr ReadImage(const char* folder_path, const char* file_name,
 }
 
 bool WriteImage(const avifImage* image, const char* file_path) {
-  return avifPNGWrite(file_path, image, /*requestedDepth=*/0,
-                      AVIF_CHROMA_UPSAMPLING_BEST_QUALITY,
-                      /*compressionLevel=*/0);
+  if (!image || !file_path) return false;
+  const size_t str_len = std::strlen(file_path);
```
jzern, line 345: std::string is already used in the file, but it's probably just as much work to use it here instead of cstring.

jzern (approved):

tests/gtest/aviftest_helpers.cc
y-guyon: If there was `std::string::ends_with()` I would reconsider but this is C++20 only.

tests/gtest/aviftest_helpers.cc
```diff
@@ -340,9 +341,15 @@ AvifImagePtr ReadImage(const char* folder_path, const char* file_name,
...
+  if (str_len >= 4 && !std::strncmp(file_path + str_len - 4, ".png", 4)) {
```
wantehchang, line 346: Nit: This comparison can be done with `memcmp` because both inputs are free of null characters in the first four bytes.

wantehchang (commented):
LGTM. Thanks.

PR #1351, opened 2023-04-17
Add test for all enc/dec codec pairs

y-guyon:


tests/gtest/avifcodectest.cc
```diff
@@ -0,0 +1,62 @@
...
+  // AVIF_CODEC_CHOICE_SVT requires dimensions to be at least 64 pixels.
+  testutil::AvifImagePtr image =
+      testutil::CreateImage(/*width=*/64, /*height=*/64, /*depth=*/8,
+                            AVIF_PIXEL_FORMAT_YUV420, AVIF_PLANES_ALL);
+  ASSERT_NE(image, nullptr);
+  testutil::FillImageGradient(image.get());
+
+  testutil::AvifEncoderPtr encoder(avifEncoderCreate(), avifEncoderDestroy);
+  ASSERT_NE(encoder, nullptr);
+  encoder->codecChoice = encoding_codec;
```
wantehchang, line 33: Unlike avifenc, the default encoder options of the C API tend to be "extreme" values (e.g., best quality/lossless) rather than the recommended values. So we may want to set some of the encoder options.

rav1e and SVT-AV1 don't support lossless compression (https://github.com/xiph/rav1e/issues/151 and https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1636). So a quality of 100 is still not lossless if rav1e or SVT-AV1 is used.

wantehchang (approved):
LGTM.

tests/gtest/avifcodectest.cc
```diff
@@ -0,0 +1,63 @@
...
+  ASSERT_NE(decoder, nullptr);
+  decoder->codecChoice = decoding_codec;
+  testutil::AvifImagePtr decoded(avifImageCreateEmpty(), avifImageDestroy);
+  ASSERT_NE(decoded, nullptr);
+  ASSERT_EQ(avifDecoderReadMemory(decoder.get(), decoded.get(), encoded.data,
+                                  encoded.size),
+            AVIF_RESULT_OK);
+
+  // AVIF_CODEC_CHOICE_SVT leads to more distortion than other codecs with
+  // default settings.
```
wantehchang, line 49: Nit: "default settings" probably should be updated.

Also at quality level 90, perhaps all three AV1 encoders need a larger distortion tolerance? If so, we can delete this comment.

y-guyon: Done at https://github.com/AOMediaCodec/libavif/pull/1380

PR #1354, opened 2023-04-19
Test cmd progressive

y-guyon:


tests/test_cmd_progressive.sh
```diff
@@ -0,0 +1,66 @@
...
+    rm -- "${ENCODED_FILE}" "${DECODED_FILE}"
+  popd
+}
+trap cleanup EXIT
+
+pushd ${TMP_DIR}
+  echo "Testing basic progressive"
+  "${AVIFENC}" --progressive -s 8 "${INPUT_Y4M}" -o "${ENCODED_FILE}"
+  "${AVIFDEC}" "${ENCODED_FILE}" "${DECODED_FILE}"
+  "${AVIFDEC}" --progressive "${ENCODED_FILE}" "${DECODED_FILE}"
```
vrabaud, line 63: Maybe add an image compare to be sure they are the same ?

vrabaud (approved):

tests/test_cmd_progressive.sh
y-guyon: The `are_images_equal` binary only checks for equality. We should extend it for similarity (PSNR).

PR #1358, opened 2023-04-19
Add an API lossless test.

vrabaud:


tests/gtest/aviflosslesstest.cc
```diff
@@ -0,0 +1,95 @@
...
+      const testutil::AvifImagePtr decoded_default =
+          testutil::ReadImage(temp_dir.c_str(), temp_file.c_str());
+
+      // Verify that the ground truth and decoded images are the same.
+      const bool are_images_equal =
+          testutil::AreImagesEqual(*ground_truth_image, *decoded_default);
+      if (matrix_coefficient == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
+        ASSERT_TRUE(are_images_equal);
+      } else {
+        ASSERT_FALSE(are_images_equal);
```
y-guyon, line 76: This is not lossless because `AVIF_MATRIX_COEFFICIENTS_YCGCO` uses only 8 bits, correct?

y-guyon (approved):

tests/gtest/aviflosslesstest.cc
vrabaud: Exactly, let me add a comment.

PR #1361, opened 2023-04-20
Add AV2 support

y-guyon:
Experimental. The AV2 specification is not finalized.

Handle `av02` image item type and `av2C` configuration property.
Handle `av02` track sample format.
Add `avifavmtest`.
Disable AV1 tests when all AV1 codecs are disabled.
Add presubmit `ci-unix-static-av2.yml`.

libavif can only generate AV2 files if it is built with `AVIF_CODEC_AVM` set to `ON`.
The `avifEncoder` API can only output AV2 files if `codecChoice` is explicitly set to `AVIF_CODEC_CHOICE_AVM` to avoid mistakes.
`avifenc` can only output AV2 files if `--codec=avm` is specified to avoid mistakes.
aom and avm conflict and cannot be used together in the same libavif binary. Other AV1 codecs can be enabled alongside aom.
libavif will decode any AV1 or AV2 file by default as long as the relevant codecs are enabled. The main behavior change for libavif with only AV1 codecs enabled is that before this change, items or tracks of type `av02` were ignored. After this change, conformant AV1-AVIF files with extra `av02` items will fail to decode unless avm is enabled.

This change allows decoding files with mixed AV1 and AV2 items or tracks (untested because no encoding tool for that). This decision can be changed in the future.

Note: The first commit exists just to reduce diffs. Please review the second and third commits.


jzern (commented):
This looks ok, but I'm wondering if this should have an `EXPERIMENTAL` define to avoid adding av2 support to the common code unnecessarily. Others might have some thoughts.

vigneshvg:
> So I guess you are talking about the `av02` and `av2C` boxes that are accepted during parsing, even if `AVIF_CODEC_AVM=OFF`. I do not think it adds any computational complexity, and leads to `AVIF_RESULT_NO_CODEC_AVAILABLE` anyway if the user proceeds with decoding without `AVIF_CODEC_AVM=ON`.

I didn't look at the code too closely. I have a question about the behavior of `avifDecoderParse`. Would `avifDecoderParse` return `true` for AV2 streams even when `AVIF_CODEC_AVM` is not set? If so, that would be a potentially not-okay thing to do since some API users rely on `avifDecoderParse` returning `true` to conclude if an image is valid AVIF or not.

jzern:
> So I guess you are talking about the `av02` and `av2C` boxes that are accepted during parsing, even if `AVIF_CODEC_AVM=OFF`. 

Yes, that's what I was referring to. Anything in the common code.

> I do not think it adds any computational complexity, and leads to `AVIF_RESULT_NO_CODEC_AVAILABLE` anyway if the user proceeds with decoding without `AVIF_CODEC_AVM=ON`.
> In my opinion it is rather safe to leave it as is and would avoid `#define`s cluttering the code, but I would be fine with a few `#ifdef EXPERIMENTAL` guards. If we go with the latter, should we also put all experiment-tagged features behind that define too (incremental, layers etc.)?

Making it explicit provides some documentation for the user and does avoid some code in production builds. It's not common right now, so we can get the group's opinion later.

src/codec_avm.c
```diff
@@ -0,0 +1,1098 @@
...
+    encoderChanges &= ~AVIF_ENCODER_CHANGE_SCALING_MODE;
+
+    if (!codec->internal->encoderInitialized) {
+        int aomCpuUsed = -1;
+        if (encoder->speed != AVIF_SPEED_DEFAULT) {
+            aomCpuUsed = AVIF_CLAMP(encoder->speed, 0, 9);
+        }
+
+        // aom_codec.h says: aom_codec_version() == (major<<16 | minor<<8 | patch)
+        static const int aomVersion_2_0_0 = (2 << 16);
```
jzern, line 542: this could be dropped if we put a minimum on the version of libavm. I don't know off the top of my head where it forked during this progression.

src/obu.c
```diff
@@ -123,12 +127,13 @@ static uint32_t avifBitsReadVLC(avifBits * const bits)
 }
 
 // ---------------------------------------------------------------------------
-// Variables in here use snake_case to self-document from the AV1 spec:
+// Variables in here use snake_case to self-document from the AV1 and AV2 specs:
```
jzern, line 130: this only describes AV1. You could say from the AV1 spec and the in development AV2 spec.

src/obu.c
```diff
@@ -325,12 +336,59 @@ static avifBool parseSequenceHeader(avifBits * bits, avifSequenceHeader * header
...
+
+    // See av1_read_sequence_header_beyond_av1() in avm.
+    // Other ignored fields.
     return !bits->error;
 }
 
-avifBool avifSequenceHeaderParse(avifSequenceHeader * header, const avifROData * sample)
+static avifBool avifSequenceHeaderParse(avifSequenceHeader * header,
+                                        const avifROData * sample,
+                                        avifBool (*parse)(avifBits *, avifSequenceHeader *))
```
vigneshvg, line 391: having a function pointer here seems a bit excessive.

given the caller of this function knows whether it's AV1 or AV2, can we just pass that boolean here and add a conditional to call the appropriate function?

src/write.c
```diff
@@ -1333,8 +1351,9 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
     }
 
     avifBoxMarker ftyp = avifRWStreamWriteBox(&s, "ftyp", AVIF_BOX_SIZE_TBD);
-    avifRWStreamWriteChars(&s, majorBrand, 4);                             // unsigned int(32) major_brand;
-    avifRWStreamWriteU32(&s, 0);                                           // unsigned int(32) minor_version;
+    avifRWStreamWriteChars(&s, majorBrand, 4); // unsigned int(32) major_brand;
+    // TODO(yguyon): AV2-AVIF is AVIFv2 for now (change once no longer experimental)
+    avifRWStreamWriteU32(&s, isAV2 ? 2 : 0);                               // unsigned int(32) minor_version;
```
vigneshvg, line 1356: does it say anywhere that minor_version == 2 means it's experimental?

just curious as to why the value is 2.

src/read.c
```diff
@@ -1758,9 +1780,15 @@ static avifBool avifParseContentLightLevelInformationBox(avifProperty * prop, co
...
-static avifBool avifParseAV1CodecConfigurationBox(const uint8_t * raw, size_t rawLen, avifCodecConfigurationBox * av1C, avifDiagnostics * diag)
+static avifBool avifParseCodecConfigurationBox(const uint8_t * raw,
+                                               size_t rawLen,
+                                               avifCodecConfigurationBox * config,
+                                               const char * configPropName,
+                                               avifDiagnostics * diag)
 {
-    BEGIN_STREAM(s, raw, rawLen, diag, "Box[av1C]");
+    char diagContext[] = "Box[....]";
+    memcpy(diagContext + 4, configPropName, 4); // "Box[av1C]" or "Box[av2C]"
```
vigneshvg, line 1790: nit: can be `snprintf`.

src/read.c
```diff
@@ -1771,25 +1799,29 @@ static avifBool avifParseAV1CodecConfigurationBox(const uint8_t * raw, size_t ra
...
 
-static avifBool avifParseAV1CodecConfigurationBoxProperty(avifProperty * prop, const uint8_t * raw, size_t rawLen, avifDiagnostics * diag)
+static avifBool avifParseCodecConfigurationBoxProperty(avifProperty * prop,
+                                                       const uint8_t * raw,
+                                                       size_t rawLen,
+                                                       const char * configPropName,
+                                                       avifDiagnostics * diag)
 {
-    return avifParseAV1CodecConfigurationBox(raw, rawLen, &prop->u.av1C, diag);
+    return avifParseCodecConfigurationBox(raw, rawLen, &prop->u.av1C, configPropName, diag);
```
vigneshvg, line 1824: nit (for a separate change): This function isn't doing anything. we can remove it and replace the callers directly.

src/read.c
```diff
@@ -1938,7 +1970,9 @@ static avifBool avifParseItemPropertyContainerBox(avifPropertyArray * properties
         } else if (!memcmp(header.type, "colr", 4)) {
             AVIF_CHECK(avifParseColourInformationBox(prop, rawOffset + avifROStreamOffset(&s), avifROStreamCurrent(&s), header.size, diag));
         } else if (!memcmp(header.type, "av1C", 4)) {
-            AVIF_CHECK(avifParseAV1CodecConfigurationBoxProperty(prop, avifROStreamCurrent(&s), header.size, diag));
+            AVIF_CHECK(avifParseCodecConfigurationBoxProperty(prop, avifROStreamCurrent(&s), header.size, "av1C", diag));
+        } else if (!memcmp(header.type, "av2C", 4)) {
```
vigneshvg, line 1974: should this `else if` be guarded by `AVIF_CODEC_AVM` ?

src/read.c
```diff
@@ -2043,8 +2077,8 @@ static avifBool avifParseItemPropertyAssociation(avifMeta * meta, const uint8_t
             // Copy property to item
             const avifProperty * srcProp = &meta->properties.prop[propertyIndex];
 
-            static const char * supportedTypes[] = { "ispe", "auxC", "colr", "av1C", "pasp", "clap", "irot",
-                                                     "imir", "pixi", "a1op", "lsel", "a1lx", "clli" };
+            static const char * supportedTypes[] = { "ispe", "auxC", "colr", "av1C", "av2C", "pasp", "clap",
```
vigneshvg, line 2080: should this be added only if `AVIF_CODEC_AVM` is set?

src/read.c
```diff
@@ -653,9 +653,24 @@ typedef struct avifTile
     uint32_t width;  // Either avifTrack.width or avifDecoderItem.width
     uint32_t height; // Either avifTrack.height or avifDecoderItem.height
     uint8_t operatingPoint;
+    uint8_t type[4]; // "av01" for AV1 ("av02" for AV2 if AVIF_CODEC_AVM)
```
vigneshvg, line 656: this should be a boolean (or an enum with AV1 and AV2 as possible values for future-proofing).

so that the type of the tile can be checked with a simple `==` instead of using a `memcmp` every time.

vigneshvg (changes requested):
i took a first pass at reviewing the code.

src/read.c
y-guyon: Done in `read.c` and `write.c` with `avifCodecType` enum.

src/write.c
y-guyon: > does it say anywhere that minor_version == 2 means it's experimental?

No.

> just curious as to why the value is 2.

Because major brand `avif` minor version 0 uses AV1. `avif` version 1 is confusing, so I used `avif` version 2 for AV2. Maybe it will be another brand such as `av2f` version 0 anyway.

Clarified the comment.

src/read.c
```diff
@@ -3474,7 +3522,7 @@ static avifResult avifDecoderCreateCodecs(avifDecoder * decoder)
 // If alpha is AVIF_FALSE, searches for the primary color item (parentItemID is ignored in this case).
 // If alpha is AVIF_TRUE, searches for the auxiliary alpha item whose parent item ID is parentItemID.
 // Returns the target item if found, or NULL.
-static avifDecoderItem * avifDecoderDataFindItem(avifDecoderData * data, avifBool alpha, uint32_t parentItemID)
+static avifDecoderItem * avifDecoderDataFindItem(const avifDecoderData * data, avifBool alpha, uint32_t parentItemID)
```
vigneshvg, line 3525: this change seems unrelated. please keep it out of this PR (it's already getting quite large).

src/read.c
```diff
@@ -3746,19 +3780,32 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                                                 decoder->imageDimensionLimit,
                                                 data->diag),
                           AVIF_RESULT_INVALID_IMAGE_GRID);
+            colorCodecType = avifDecoderItemGetGridCodecType(colorItem);
+            AVIF_CHECKERR(colorCodecType != AVIF_CODEC_TYPE_UNKNOWN, AVIF_RESULT_INVALID_IMAGE_GRID);
```
vigneshvg, line 3784: can you please change this to a normal if condition?

Use of `AVIF_CHECKERR` makes this a double negation and difficult to parse.

here and elsewhere. i have marked some of the places which i found.

src/read.c
```diff
@@ -3746,19 +3780,32 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
+                AVIF_CHECKRES(avifDecoderItemRead(alphaItem, decoder->io, &readData, 0, 0, data->diag));
+                AVIF_CHECKERR(avifParseImageGridBox(&data->alpha.grid,
+                                                    readData.data,
+                                                    readData.size,
+                                                    decoder->imageSizeLimit,
+                                                    decoder->imageDimensionLimit,
+                                                    data->diag),
+                              AVIF_RESULT_INVALID_IMAGE_GRID);
+                alphaCodecType = avifDecoderItemGetGridCodecType(alphaItem);
+                AVIF_CHECKERR(alphaCodecType != AVIF_CODEC_TYPE_UNKNOWN, AVIF_RESULT_INVALID_IMAGE_GRID);
```
vigneshvg, line 3804: can you please change this to a normal if condition?

Use of AVIF_CHECKERR makes this a double negation and difficult to parse.

src/write.c
```diff
@@ -1282,10 +1299,8 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
         return AVIF_RESULT_NO_CONTENT;
     }
 
-    // TODO(yguyon): Rework when AVIF_CODEC_CHOICE_AUTO can be AVM
-    assert((encoder->codecChoice != AVIF_CODEC_CHOICE_AUTO) ||
-           (strcmp(avifCodecName(encoder->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE), "avm") != 0));
-    const avifBool isAV2 = encoder->codecChoice == AVIF_CODEC_CHOICE_AVM;
+    const avifCodecType codecType = avifEncoderGetCodecType(encoder);
+    AVIF_CHECKERR(codecType != AVIF_CODEC_TYPE_UNKNOWN, AVIF_RESULT_NO_CODEC_AVAILABLE);
```
vigneshvg, line 1303: can you please change this to a normal if condition?

Use of AVIF_CHECKERR makes this a double negation and difficult to parse.

vigneshvg (changes requested):
mostly okay. just a few minor comments.

src/read.c
```diff
@@ -3418,11 +3466,11 @@ static avifBool avifTilesCanBeDecodedWithSameCodecInstance(avifDecoderData * dat
...
     const uint8_t firstTileOperatingPoint = data->tiles.tile[0].operatingPoint;
     const avifBool firstTileAllLayers = data->tiles.tile[0].input->allLayers;
-    const uint8_t * firstTileType = data->tiles.tile[0].type;
+    const avifCodecType firstTileCodecType = data->tiles.tile[0].codecType;
     for (unsigned int i = 1; i < data->tiles.count; ++i) {
         const avifTile * tile = &data->tiles.tile[i];
-        if (tile->operatingPoint != firstTileOperatingPoint || tile->input->allLayers != firstTileAllLayers ||
-            memcmp(tile->type, firstTileType, 4)) {
+        if ((tile->operatingPoint != firstTileOperatingPoint) || (tile->input->allLayers != firstTileAllLayers) ||
+            (tile->codecType != firstTileCodecType)) {
```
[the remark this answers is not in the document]
y-guyon: We do not have to support this feature, but I thought it would not hurt to do so.

I believe AV1 and HEVC tiles can be mixed in a HEIF grid if they share the same properties (subsampling etc.) but I doubt any decoder handles and tests that as of today. It is even less likely that an encoder allows that.

src/read.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1372

src/read.c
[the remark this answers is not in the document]
vigneshvg: I wasn't expecting any action on this comment, it was more of a general observation. But thanks for adding the check :)

src/read.c
vigneshvg: Yes, if it was an assertion i would certainly prefer the latter. But i don't think the analogy you have here is right. The AVIF_CHECKERR statement when expanded looks like this:

```c
if (!(colorCodecType != AVIF_CODEC_TYPE_UNKNOWN)) {
    return AVIF_FALSE;
}
```

And if we were to write it without a macro, we certainly wouldn't write it like this. :)

y-guyon: The same could be said about `assert()` if it was implemented as:
```c
#define assert(x) \
if (!(x)) {
    abort();
}
```

And I do read `AVIF_CHECKERR()` as "check this and continue if true", which is very similar to the assertion pattern. Anyway.

y-guyon:
I rebased this PR.

> The changes in read.c, write.c and obu.c look fine now. I still haven't looked at codec_avm.c. I will look into it sometime today. Sorry for the delay and thanks for the patience in working through this.

No worries, it is not urgent. Thanks for your time.

src/codec_avm.c
```diff
@@ -0,0 +1,1087 @@
...
+        int maxQuantizer;
+        if (alpha) {
+            minQuantizer = encoder->minQuantizerAlpha;
+            maxQuantizer = encoder->maxQuantizerAlpha;
+        } else {
+            minQuantizer = encoder->minQuantizer;
+            maxQuantizer = encoder->maxQuantizer;
+        }
+        // Scale from aom's [0:63] to avm's [0:255]. TODO(yguyon): Accept [0:255] directly in avifEncoder.
+        minQuantizer = AVIF_CLAMP((minQuantizer * 255 + 31) / 63, 0, 255);
```
vigneshvg, line 657: this scale + clamp is done a lot of times, can you please make it to a utility function?

vigneshvg (changes requested):

vigneshvg (approved):
lgtm. thank you.

src/codec_avm.c
```diff
@@ -0,0 +1,1090 @@
...
+    if ((aomScalingMode.h_scaling_mode != AOME_NORMAL) || (aomScalingMode.v_scaling_mode != AOME_NORMAL)) {
+        // AOME_SET_SCALEMODE only applies to next frame (layer), so we have to set it every time.
+        aom_codec_control(&codec->internal->encoder, AOME_SET_SCALEMODE, &aomScalingMode);
+    }
+
+    aom_image_t aomImage;
+    // We prefer to simply set the aomImage.planes[] pointers to the plane buffers in 'image'. When
+    // doing this, we set aomImage.w equal to aomImage.d_w and aomImage.h equal to aomImage.d_h and
+    // do not "align" aomImage.w and aomImage.h. Unfortunately this exposes a bug in libaom
+    // (https://crbug.com/aomedia/3113) if chroma is subsampled and image->width or image->height is
```
jzern, line 851: I wonder if this was carried over to avm

jzern (approved):

include/avif/avif.h
```diff
@@ -150,7 +150,7 @@ typedef enum avifResult
     AVIF_RESULT_ENCODE_COLOR_FAILED,
     AVIF_RESULT_ENCODE_ALPHA_FAILED,
     AVIF_RESULT_BMFF_PARSE_FAILED,
-    AVIF_RESULT_NO_AV1_ITEMS_FOUND,
+    AVIF_RESULT_NO_AV1_ITEMS_FOUND, // TODO(yguyon): Rename or add AVIF_RESULT_NO_AV2_ITEMS_FOUND
```
wantehchang, line 153: It seems that this error code is used to report that the primary item is missing, or the primary item or the alpha auxiliary item (non-grid) has a zero size. Whe we report this error, we haven't called `avifDecoderItemValidateAV1` (replaced by `avifDecoderItemValidateProperties` in the new code). So it seems better to rename this error code to be more generic. We can keep the old enum for backward compatibility.

When we do that, we need to explicitly assign values to the enum constants, which is a good thing to do in general when an enum type has many constants.

include/avif/internal.h
```diff
@@ -301,7 +301,24 @@ void avifCodecSpecificOptionsDestroy(avifCodecSpecificOptions * csOptions);
...
+
+// Alliance for Open Media video formats that can be used in the AVIF image format.
+typedef enum avifCodecType
+{
+    AVIF_CODEC_TYPE_UNKNOWN,
+    AVIF_CODEC_TYPE_AV1,
+#if defined(AVIF_CODEC_AVM)
+    AVIF_CODEC_TYPE_AV2, // Experimental.
+#endif
+    AVIF_CODEC_TYPE_COUNT
```
wantehchang, line 314: Don't define this `AVIF_CODEC_TYPE_COUNT` constant unless it is needed.

CMakeLists.txt
```diff
@@ -445,7 +449,46 @@ if(AVIF_CODEC_AOM)
     message(STATUS "libavif: Codec enabled: aom (${AVIF_CODEC_AOM_ENCODE_DECODE_CONFIG})")
 endif()
 
-if(NOT AVIF_CODEC_AOM AND NOT AVIF_CODEC_DAV1D AND NOT AVIF_CODEC_LIBGAV1)
+if(AVIF_CODEC_AVM)
+    message(WARNING "libavif: AV2 support with avm is experimental. Only use for testing.")
+
+    # The avm repository is a fork of aom and inherited a lot of folders, files and build artifacts named the same way.
+    # Having both dependencies at the same time generates conflicts in includes, binary lookups etc.
```
wantehchang, line 456: This is problematic. If it is not feasible to implement AV2 as a codec algorithm in the same aom_codec_ interface, the AV2 team should rename the public headers, libraries, and global functions and variables in AVM.

wantehchang (commented):
Yannis: My review is supplemental because I only checked avif.h, internal.h, and the top-level CMakelists.txt. You can merge this pull request with James and Vignesh's approval.

CMakeLists.txt
```diff
@@ -445,7 +449,46 @@ if(AVIF_CODEC_AOM)
...
+    if(AVIF_CODEC_AOM)
```
wantehchang, line 457: I found that AVM and SVT-AV1 both define a global function `compute_distortion`, so if we link with both AVM and SVT-AV1 as static libraries, we will get a "multiple definition of `compute_distortion'" linker error.

This can be avoided by declaring `compute_distortion` as a file-static function in AVM or SVT-AV1 (or both).

wantehchang: SVT-AV1 fixed this issue in https://gitlab.com/AOMediaCodec/SVT-AV1/-/commit/61ff779ea2f2b908cce00cec4cd304a06fc88a93?page=2, and it is part of their v1.5.0 release. I will update libavif to check out SVT-AV1 v1.5.0 in https://github.com/AOMediaCodec/libavif/pull/1377. Then we will be able to modify .github/workflows/ci-unix-static-av2.yml to build SVT-AV1.

src/codec_avm.c
y-guyon: No. See [avm's extend.c](https://gitlab.com/AOMediaCodec/avm/-/blob/b52eef5c9a563593edc7b928561cfce235ddbb16/av1/encoder/extend.c#L72) vs [aom's extend.c](https://aomedia.googlesource.com/aom/+/ea9a06fd1a6bf5c9c6ff71b1737b1bed99f1ac6d/av1/encoder/extend.c#124).

include/avif/internal.h
y-guyon: Done. It was to avoid the dangling comma but trailing commas seem allowed in C99.

wantehchang: I see. An alternative is to define the `AVIF_CODEC_TYPE_AV2` enum constant unconditionally, similar to how we define `AVIF_CODEC_CHOICE_AVM` in avif.h.

I don't know whether trailing commas are allowed in C99. If Clang, GCC, and MSVC all compile the new code successfully, then we can considered them as supported in practice.

jzern: Trailing commas were an addition to C99.

jzern: https://en.cppreference.com/w/c/language/enum

*...where enumerator-list is a comma-separated list (with trailing comma permitted) (since C99) of enumerator, each of which has the form:...*

wantehchang: Thank you, James.

Perhaps it's the initializer list for a struct or an array that may not have a trailing comma in C99?

CMakeLists.txt
wantehchang: Thank you. I just added a comment to that bug.

include/avif/internal.h
jzern: C90 supported that. C99 added array and struct designators.

https://en.cppreference.com/w/c/language/array_initialization

include/avif/avif.h
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1394.

CMakeLists.txt
y-guyon: > Then we will be able to modify .github/workflows/ci-unix-static-av2.yml to build SVT-AV1.

Done in https://github.com/AOMediaCodec/libavif/pull/1395.

Issue #1362, opened 2023-04-21
Android JNI - Incorrect Premultiplied Default?


joedrago:
Hello @vigneshvg --

I was doing some internal testing with a teammate and we noticed that decoded Bitmap objects coming out of AvifDecoder.decode() are flagged as premultiplied (as all Bitmaps must be in order to be drawn by Android draw calls), but in the guts of the JNI `decode()` function, we are just using the default value of `avifRGBImage.alphaPremultiplied`, which is `AVIF_FALSE`. This results in a Bitmap which contains the actual unpremultiplied pixels in them, yet the Bitmap itself is flagged as premultiplied. This causes incorrect blending.

If I'm right (please correct me if not), there are two solutions to this:

* Legislate that all Bitmap data coming out of the Java AvifDecoder is actually premultiplied. Android wants this anyway, so it'd be as simple as adding:

    rgb_image.alphaPremultiplied = AVIF_TRUE;

... after the call to `avifRGBImageSetDefaults()`. 

* Expose the `alphaPremultiplied` property in some way to Java, so consumers of this library can set it to true. I don't see a ton of value here of it ever being false, but if we're going for completeness and someone simply wants to query pixel data after decode or do post-processing before handing it over to Android to draw, that would also be fine.

Thoughts?

vigneshvg:
> Hello @vigneshvg --
> 
> I was doing some internal testing with a teammate and we noticed that decoded Bitmap objects coming out of AvifDecoder.decode() are flagged as premultiplied

Can you elaborate a bit more on what exactly you mean by "Bitmaps are flagged as premultiplied"? Because the `bitmap` is an input parameter that's taken by the `AvifDecoder.decode()` function. The JNI wrapper itself does not do anything more to the bitmap other than to copy pixels. Depending on the configuration of the bitmap, the alpha plane should be handled correctly. For ARGB_8888 and RGBA_F16, the alpha plane is not premultiplied.

I have tried loading images with alpha plane several times before and it has worked as intended. Do you know which type of bitmaps you were testing? If you can share a sample image that fails to render properly, i can take a further look. :)

> (as all Bitmaps must be in order to be drawn by Android draw calls), but in the guts of the JNI `decode()` function, we are just using the default value of `avifRGBImage.alphaPremultiplied`, which is `AVIF_FALSE`. This results in a Bitmap which contains the actual unpremultiplied pixels in them, yet the Bitmap itself is flagged as premultiplied. This causes incorrect blending.
[...]

PR #1363, opened 2023-04-21
android_jni: Set alphaPremultiplied to true

vigneshvg:
Android platform always sees the Bitmaps as premultiplied with alpha when it renders them. So set alphaPremultiplied to true when computing the pixels to be copied into the Bitmap.

https://developer.android.com/reference/android/graphics/Bitmap#setPremultiplied(boolean)

Fixes Issue #1362


vigneshvg:
Thanks @joedrago for figuring this out.

It seems the alpha blending that i remembered earlier are from rendering videos and that was using some custom GL code for the blending and wasn't using android Bitmaps.

Here is g_magenta.avif rendered in a blue background without this PR:
image

And with this PR:
image

joedrago (approved):

PR #1367, opened 2023-04-24
android_jni: Use specific files for StillImageTest

vigneshvg:
Instead of iterating over the assets directory, used a fixed set of files. This will let the tests assert for dimensions and depth.  This will also prevent the tests from accidentally passing in case the assets directory is not included correctly.

Also introduce a utility class to refactor some of the common test code.

### DIFF
--- a/android_jni/avifandroidjni/src/androidTest/java/org/aomedia/avif/android/StillImageTest.java
+++ b/android_jni/avifandroidjni/src/androidTest/java/org/aomedia/avif/android/StillImageTest.java
@@ -1,18 +1,13 @@
 package org.aomedia.avif.android;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.aomedia.avif.android.TestUtils.Image;
 
-import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.Bitmap.Config;
-import androidx.test.platform.app.InstrumentationRegistry;
 import java.io.IOException;
-import java.io.InputStream;
 import java.nio.ByteBuffer;
-import java.nio.channels.Channels;
-import java.nio.file.Paths;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import org.aomedia.avif.android.AvifDecoder.Info;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


android_jni/avifandroidjni/src/androidTest/java/org/aomedia/avif/android/StillImageTest.java
```diff
@@ -25,18 +20,33 @@
...
+    new Image("fox.profile0.8bpc.yuv420.avif", 1204, 800, 8, false, 1),
+    new Image("fox.profile0.8bpc.yuv420.monochrome.avif", 1204, 800, 8, false, 1),
+    new Image("fox.profile1.10bpc.yuv444.avif", 1204, 800, 10, false, 1),
+    new Image("fox.profile1.8bpc.yuv444.avif", 1204, 800, 8, false, 1),
+    new Image("fox.profile2.10bpc.yuv422.avif", 1204, 800, 10, false, 1),
+    new Image("fox.profile2.12bpc.yuv420.avif", 1204, 800, 12, false, 1),
+    new Image("fox.profile2.12bpc.yuv420.monochrome.avif", 1204, 800, 12, false, 1),
+    new Image("fox.profile2.12bpc.yuv422.avif", 1204, 800, 12, false, 1),
+    new Image("fox.profile2.12bpc.yuv444.avif", 1204, 800, 12, false, 1),
+    new Image("fox.profile2.8bpc.yuv422.avif", 1204, 800, 8, false, 1)
```
jzern, line 36: looks like the last two parameters are always `false, 1`. Any plans to add further coverage?

jzern (approved):

android_jni/avifandroidjni/src/androidTest/java/org/aomedia/avif/android/StillImageTest.java
vigneshvg: yes, these tests are a bit slow sometimes on emualtors. so i will probably replace some of these files with files that have alpha and change the thread count for some of the existing ones in a follow-up.

PR #1373, opened 2023-04-28
Enforce MIAF grid constraints at decoding

y-guyon:


src/read.c
```diff
@@ -1048,6 +1048,49 @@ static avifResult avifDecoderItemValidateAV1(const avifDecoderItem * item, avifD
...
+                if (!tileAv1CProp) {
+                    avifDiagnosticsPrintf(diag,
+                                          "Tile item ID %u of type '%.4s' is missing mandatory av1C property",
+                                          tile->id,
+                                          (const char *)tile->type);
+                    return AVIF_RESULT_BMFF_PARSE_FAILED;
+                }
+                // av1CProp was copied from a tile item to the grid item. Comparing tileAv1CProp with it
+                // is equivalent to comparing tileAv1CProp with the av1C property from the first tile.
+                if ((tileAv1CProp->u.av1C.seqProfile != av1CProp->u.av1C.seqProfile) ||
```
vigneshvg, line 1075: this can be a memcmp on u.av1C.

y-guyon: I know `avifCodecConfigurationBox` is only bytes for now but it seems safer to avoid `memcmp()` in case there is some padding in the future. I did not even see a `memset()` initialization on `avifCodecConfigurationBox` in read.c so just adding an unused field would generate undefined behavior.

Hopefully the compiler can optimize the comparisons into the equivalent of `memcmp()`.

vigneshvg (approved):

src/read.c
```diff
@@ -1048,6 +1048,49 @@ static avifResult avifDecoderItemValidateAV1(const avifDecoderItem * item, avifD
...
+    if (!memcmp(item->type, "grid", 4)) {
+        // MIAF (ISO 23000-22:2019), Section 7.3.11.4.1:
+        //   All input images of a grid image item shall use the same coding format, chroma sampling format, and the
+        //   same decoder configuration (see 7.3.6.2).
+        for (uint32_t i = 0; i < item->meta->items.count; ++i) {
+            const avifDecoderItem * tile = &item->meta->items.item[i];
+            if (tile->dimgForID == item->id) {
+                // The coding format is defined by the item type.
+                // avifDecoderGenerateImageGridTiles() skipped any non-AV1 tile.
+                if (memcmp(tile->type, "av01", 4)) {
```
jzern, line 1060: Would this skip a heic tile and not fail like the comment seems to imply? *All input images of a grid image item shall use the same coding format...*

Though it's shall and not SHALL...

y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1393.

PR #1376, opened 2023-05-02
Remove static for TestUtils.Image in android_jni

y-guyon:
See https://google.github.io/styleguide/javaguide.html#s3.3.4-import-class-not-static


wantehchang (approved):
LGTM. I verified that org.aomedia.avif.android.TestUtils.Image is a nested static class:

```
/** Utility class used by the instrumented tests. */
public class TestUtils {

  ...

  public static class Image {
```

So the Style Guide's recommendation that Yannis cited applies.

vigneshvg:
Thanks for fixing this @y-guyon.

PR #1379, opened 2023-05-05
Add test_cmd_avm.sh

y-guyon:
Exercize `avifenc -c avm` and `avifdec` with avm enabled.

The first commit exists for easier diff. Hide it in the GitHub reviewing interface.

### DIFF
--- a/tests/test_cmd_avm.sh
+++ b/tests/test_cmd_avm.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+# Copyright 2022 Google LLC
+#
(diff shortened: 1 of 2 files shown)


maryla-uc (approved):

tests/test_cmd_avm.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+# Copyright 2022 Google LLC
```
y-guyon, line 2: Note: I forgot to update the copyright year and squashed the branching commit. I guess it is fine.

PR #1382, opened 2023-05-09
Only find_package libyuv and libsharpyuv if not set locally.

vrabaud:
If libyuv or libsharpyuv is set locally, the variables can be overriden by find_package if another version is installed (e.g. through distribution).


y-guyon (approved):

CMakeLists.txt
```diff
@@ -241,36 +241,40 @@ if(UNIX)
...
+        if(NOT LIBYUV_VERSION)
+            message(STATUS "libavif: libyuv found, but version unknown; libyuv-based fast paths disabled.")
+        elseif(LIBYUV_VERSION LESS 1755)
+            message(STATUS "libavif: libyuv (${LIBYUV_VERSION}) found, but is too old; libyuv-based fast paths disabled.")
+        else()
+            message(STATUS "libavif: libyuv (${LIBYUV_VERSION}) found; libyuv-based fast paths enabled.")
+            if(LIBYUV_VERSION LESS 1813)
+                message(STATUS "libavif: some libyuv optimizations require at least version 1813 to work.")
+            endif()
+            set(AVIF_PLATFORM_DEFINITIONS ${AVIF_PLATFORM_DEFINITIONS} -DAVIF_LIBYUV_ENABLED=1)
```
wantehchang, line 259: This is the only line that adds `-DAVIF_LIBYUV_ENABLED=1`. So if `AVIF_LOCAL_LIBYUV` is set, then `-DAVIF_LIBYUV_ENABLED=1` will **not** be added.

CMakeLists.txt
```diff
@@ -241,36 +241,40 @@ if(UNIX)
...
-    set(AVIF_PLATFORM_DEFINITIONS ${AVIF_PLATFORM_DEFINITIONS} -DAVIF_LIBSHARPYUV_ENABLED=1)
-    set(AVIF_PLATFORM_INCLUDES ${AVIF_PLATFORM_INCLUDES} ${LIBSHARPYUV_INCLUDE_DIR})
-    set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ${LIBSHARPYUV_LIBRARY})
-else()
-    message(STATUS "libavif: libsharpyuv not found")
+if(NOT AVIF_LOCAL_LIBSHARPYUV)
+    find_package(libsharpyuv QUIET) # not required
+    if(libsharpyuv_FOUND)
+        message(STATUS "libavif: libsharpyuv found; sharp rgb to yuv conversion enabled.")
+        set(AVIF_PLATFORM_DEFINITIONS ${AVIF_PLATFORM_DEFINITIONS} -DAVIF_LIBSHARPYUV_ENABLED=1)
```
wantehchang, line 272: Similarly, if `AVIF_LOCAL_LIBSHARPYUV` is set, then `-DAVIF_LIBSHARPYUV_ENABLED=1` will **not** be added.

wantehchang (changes requested):
Vincent: This pull request needs work.

wantehchang:
Vincent: I will review this pull request tomorrow, but when I tested it on a source tree with local libyuv and libsharpyuv, the cmake command produced the following output:

```
-- The C compiler identification is GNU 12.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- The CXX compiler identification is GNU 12.2.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- AVIF_LOCAL_LIBSHARPYUV LIB_FILENAME libavif/ext/libwebp/build/libsharpyuv.a
-- libavif: Enabling warnings for GCC
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- libavif: Codec enabled: dav1d (decode)
-- libavif: Codec enabled: libgav1 (decode)
-- libavif: Codec enabled: svt (encode)
-- libavif: Codec enabled: aom (encode/decode)
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.13") 
-- Found PNG: /usr/lib/x86_64-linux-gnu/libpng.so (found suitable version "1.6.39", minimum required is "1.6.32") 
-- Found JPEG: /usr/lib/x86_64-linux-gnu/libjpeg.so (found version "62") 
-- Found local ext/googletest
-- Some tests are skipped because libsharpyuv is unavailable.
-- libavif: Using pandoc: /usr/bin/pandoc
-- Configuring done
-- Generating done
-- Build files have been written to: /usr/local/google/home/wtc/libavif-mirror.1/libavif/build
```

Note the message " Some tests are skipped because libsharpyuv is unavailable."

wantehchang:
I just fixed the AppVeyor build failure (commit 585777ead5c6dbf4a35ef0d70be7e45baaca3ad8). Please rebase your commits on the current tip of the main branch. Thanks.

wantehchang:
The message " Some tests are skipped because libsharpyuv is unavailable." was added in commit 7e70644c. I think we need to change `if(NOT libsharpyuv_FOUND)` in tests/CMakeLists.txt to test a different condition.

CMakeLists.txt
```diff
@@ -153,7 +183,19 @@ if(AVIF_LOCAL_LIBSHARPYUV)
         set(LIBSHARPYUV_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/libwebp" PARENT_SCOPE)
         set(LIBSHARPYUV_LIBRARY ${LIB_FILENAME} PARENT_SCOPE)
     endif()
+else(AVIF_LOCAL_LIBSHARPYUV)
+    find_package(libsharpyuv QUIET) # not required
+    if(libsharpyuv_FOUND)
+        message(STATUS "libavif: libsharpyuv found; sharp rgb to yuv conversion enabled.")
```
wantehchang, line 189: The `if(AVIF_LOCAL_LIBSHARPYUV)` branch should also emit a message indicating sharp rgb to yuv conversion is enabled.

vrabaud: Done

wantehchang (commented):
Vincent: I am sorry I got busy and forgot to review this pull request today. I have a question: I have been wondering why it is good to avoid calling find_package libyuv and libsharpyuv when we can. Could you explain that in the commit message?

Another question is that find_package is used with many other libraries that may be built locally, why does this pull request change only the code for libyuv and libsharpyuv? Is it because you want to try the new pattern on these two libraries first, and if the new pattern works well, you plan to extend this approach to the other libraries?

vrabaud:
I updated the description on top.
Other packages either assume the package is local and set variables by hand (e.g. *_INCLUDE_DIRS) or find_package it. For libyuv, if assumed to be local, we set the variables and then, no matter what, we find_package it thus overriding the variables.

CMakeLists.txt
```diff
@@ -140,7 +144,34 @@ if(AVIF_LOCAL_LIBYUV)
         find_package(JPEG REQUIRED)
         set(LIBYUV_LIBRARY ${LIBYUV_LIBRARY} ${JPEG_LIBRARY})
     endif()
+    set(libyuv_FOUND TRUE)
```
wantehchang, line 147: Add a message, similar to the one at line 161, to indicate that libyuv-based fast paths are enabled.

Note that the message at line 161 requires the `LIBYUV_VERSION` variable, which is not defined here, so take care if you copy that message here or move that message to the `if(libyuv_FOUND)` block at line 170.

wantehchang, line 147: Note the libyuv version check at line 157. The new code here assumes the locally-built libyuv in ext/libyuv has a new enough version. I think it is fine to assume this. I just wanted to point out this difference from the original code.

wantehchang:
> Other packages either assume the package is local and set variables by hand (e.g. *_INCLUDE_DIRS) or find_package it (e.g.
> 
> https://github.com/AOMediaCodec/libavif/blob/51395fb35d5238f995c8b2106fcb960aef3f84fc/CMakeLists.txt#L324
> 
> ). For libyuv, if assumed to be local, we set the variables and then, no matter what, we find_package it thus overriding the variables.

Thanks for the answer to my question. The AV1 codecs are indeed handled that way ("either assume the package is local and set variables by hand (e.g. *_INCLUDE_DIRS) or find_package it"). However, AVIF_LOCAL_ZLIBPNG and AVIF_LOCAL_JPEG are handled in a way similar to (but not the same as) how libyuv and libsharpyuv are handled by the original code.

wantehchang (approved):
LGTM. I suggest a change.

CMakeLists.txt
vrabaud: DOne

vrabaud: Right, I assumed the user would use the recommended commands.

vrabaud:
Thx, I had only looked into AV1 codecs. I added support for JPEG, ZLIB and PNG. We should really use ExternalProjectto transfer variables and targets properly ...

wantehchang:
Hi Vincent,

I didn't mean to ask you to also fix the JPEG, ZLIB, and PNG code in this pull request. I will need to test the new code on Linux today. (Our CI builds only test the AVIF_LOCAL_JPEG and AVIF_LOCAL_ZLIBPNG code on Windows.)

Please wait for my review. Thanks.

CMakeLists.txt
```diff
@@ -567,9 +575,13 @@ option(AVIF_ENABLE_GTEST
...
-    find_package(ZLIB REQUIRED)
-    find_package(PNG 1.6.32 REQUIRED) # 1.6.32 or above for png_get_eXIf_1()/png_set_eXIf_1() and iTXt (for XMP).
-    find_package(JPEG REQUIRED)
+    if(NOT AVIF_LOCAL_ZLIBPNG)
+        find_package(ZLIB REQUIRED)
+        find_package(PNG 1.6.32 REQUIRED) # 1.6.32 or above for png_get_eXIf_1()/png_set_eXIf_1() and iTXt (for XMP).
+    endif()
+    if(NOT AVIF_LOCAL_JPEG)
+        find_package(JPEG REQUIRED)
+    endif()
```
wantehchang, line 584: I tested the `AVIF_LOCAL_ZLIBPNG` and `AVIF_LOCAL_JPEG` changes here on my Linux desktop. These changes work, but I don't know why they work.

For example, suppose `AVIF_LOCAL_ZLIBPNG` is set to `ON`. The at line 123 we have:

```
        set(ZLIB_LIBRARY zlibstatic)
```

I verified that at line 595, the value of `${ZLIB_LIBRARY}` is still `zlibstatic`. But the `target_link_libraries()` command at line 595 somehow magically substitutes `ext/zlib/libz.a` (rather than `zlibstatic`) for `${ZLIB_LIBRARY}`. I don't know how CMake does that.

Similarly for the `PNG_LIBRARY` and `JPEG_LIBRARY` variables. They are set to `png_static` and `jpeg` at line 119 and line 130 respectively, but when these two variables are used in the `target_link_libraries()` command at line 595, they are replaced by `ext/libpng/libpng16d.a` and `ext/libjpeg/libjpeg.a` magically.

Do you know how this works?

Note: I used the following commands to remove the zlib and jpeg system headers from my Linux desktop:
```
sudo apt remove zlib1g-dev:amd64
sudo apt remove libjpeg62-turbo-dev:amd64
```

wantehchang (approved):
Vincent: LGTM. Please see my comment/question below, but I think it is fine to merge this pull request.

CMakeLists.txt
wantehchang, line 584: There is another `find_package(JPEG REQUIRED)` call at line 151:

```
149     if(BUILD_SHARED_LIBS)
150         # Fix "libyuv.so: undefined reference to `jpeg_read_raw_data'" errors.
151         find_package(JPEG REQUIRED)
152         set(LIBYUV_LIBRARY ${LIBYUV_LIBRARY} ${JPEG_LIBRARY})
153     endif()
```

It should also be guarded by `if(NOT AVIF_LOCAL_JPEG)`.

vrabaud: Good catch. It's actually better to check for the definition of JPEG_LIBRARY

wantehchang: Thanks a lot for the explanation. It all makes sense now.

I still have a lot to learn about CMake...

wantehchang: Checking for `if(NOT AVIF_LOCAL_JPEG)` is equivalent and allows simpler pattern matching with the code here. I leave the choice to you.

wantehchang (approved):
LGTM!

PR #1386, opened 2023-05-10
avifenc.c: Change error message on lossless codec

wantehchang:
Instead of saying the codec can only be AOM (or AVM) in lossless mode, say a particular codec doesn't support lossless encoding yet, with a link to its feature request for lossless support.


y-guyon (approved):
Good idea!

wantehchang:
Thanks for the review. A potential criticism of this approach is that the error message doesn't suggest how to correct the error. Since there are only 3 or 4 encoders, it should not be hard to try the other encoders one by one.

PR #1388, opened 2023-05-11
Fix lossless codec check in avifenc

y-guyon:
Otherwise using `--lossless` without aom will fail with the misleading error status `AVIF_RESULT_NO_CODEC_AVAILABLE`.

Remove redundant aom lossless check below.

Also fix `avifCodecName()` usage in `avifEncodeImagesFixedQuality()` because it could lead to printing a null pointer as a string when there is no codec available.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -222,7 +222,7 @@ if(AVIF_BUILD_APPS)
     )
 
     if(NOT AVIF_CODEC_AOM OR NOT AVIF_CODEC_AOM_ENCODE)
-        # Only aom encoder supports lossless encoding.
+        # Only aom encoder supports AV1 lossless encoding.
         set_property(TEST test_cmd_animation PROPERTY DISABLED True)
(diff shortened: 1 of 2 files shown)


vrabaud (approved):

tests/CMakeLists.txt
```diff
@@ -231,7 +231,7 @@ if(AVIF_BUILD_APPS)
             set_property(TEST test_cmd_metadata PROPERTY DISABLED True)
         endif()
 
-        # Only aom encoder supports progressive encoding.
+        # Only aom encoder supports AV1 progressive encoding.
```
wantehchang, line 234: Nit: It is AVIF, not AV1, that has progressive encoding. This can be rephrased as "Only aom encoder supports encoding AV1 spatial layers.", assuming the reader knows spatial layers are used to implement AVIF layered imaged that can be progressively decoded.

y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1390

PR #1389, opened 2023-05-11
Fix gamma for some PNG files.

### LINKED ISSUE #1139: Add support for the PNG cHRM chunk
avifenc does not support reading the gAMA chunk on PNG files. So the encoded data may not match the input file.

You can download a PNG with a gAMA chunk here:
https://github.com/eclipseo/eclipseo.github.io/blob/2f93d184577d273cac729e778d73fdcc8a93d0a8/image-comparison-web/comparisonfiles/subset1/Original/Reykjav%C3%ADk_03.png

This is easiest to see in the top right corner if you look at the woods in the background. The woods appear to be lighter than the original.
https://storage.googleapis.com/demos.webmproject.org/webp/cmp/2022_09_28/index.html#reykjavik-03*1:1&AVIF-AOM=ll&JXL=ll&subset1

Thank you @y-guyon for helping me track down the root cause of the issue.

maryla-uc:
If the PNG file contains a gAMA chunk and a cHRM chunk, and does not contain an iCCP chunk or an sRGB chunk, then convert the RGB values from the gamma specified in the gAMA chunk to the sRGB gamma curve.

Partially fixes #1139 since the cHRM chunk itself is still ignored. If the cHRM values are different from the sRGB chromaticities, there will still be a visible color change.


apps/shared/avifpng.c
```diff
@@ -196,6 +198,65 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
 
+static float avifToLinear(float v, float gamma)
+{
+    return powf(v, gamma);
+}
+
+#define SRGB_GAMMA_SLOPE 12.92f
+static float avifToSrgbGamma(float linear)
+{
+    static const float kThreshold = 0.04045f / SRGB_GAMMA_SLOPE;
```
y-guyon, line 209: Inconsistent constant naming: `SRGB_GAMMA_SLOPE` (define) vs `kThreshold` (static const)

apps/shared/avifpng.c
```diff
@@ -196,6 +198,65 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
+
+    const int rgbMaxChannel = (1 << rgb->depth) - 1;
+    const int rowWidth = rgb->rowBytes / rgbChannelBytes;
+    for (uint32_t j = 0; j < rgb->height; ++j) {
+        for (uint32_t i = 0; i < rgb->width; ++i) {
+            const uint32_t pixel_offset = (i * samplesPerPixel) + (j * rowWidth);
+            if (rgbChannelBytes > 1) {
+                uint16_t * pixels16 = (uint16_t *)rgb->pixels;
+                for (uint32_t k = 0; k < 3; ++k) {
+                    pixels16[pixel_offset + k] = (uint16_t)avifCorrectGamma(pixels16[pixel_offset + k], gamma, rgbMaxChannel);
```
y-guyon, line 241: For speed and eventually padding reasons, I recommend browsing rows with `rowBytes` as a step, then 8- or 16-bit pixels with `samplesPerPixel` as a step.

apps/shared/avifpng.c
```diff
@@ -196,6 +198,65 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
+    const int scaledUp = (int)floorf(corrected * rgbMaxChannel + 0.5f);
+    return scaledUp < 0 ? 0 : scaledUp > rgbMaxChannel ? rgbMaxChannel : scaledUp;
+}
+
+// Converts samples in the image from the given 'gAMA' value to the sRGB gamma curve.
+static void avifApplyGama(avifRGBImage * rgb, float gama)
+{
+    const float gamma = 1.f / gama;
+    assert(rgb->format == AVIF_RGB_FORMAT_RGBA || rgb->format == AVIF_RGB_FORMAT_RGB);
+    const uint32_t samplesPerPixel = rgb->format == AVIF_RGB_FORMAT_RGBA ? 4 : 3;
```
y-guyon, line 230: `avifRGBFormatChannelCount()`  
`avifRGBImagePixelSize()` may be of some use too

apps/shared/avifpng.c
```diff
@@ -347,6 +418,15 @@ avifBool avifPNGRead(const char * inputFilename,
         rowPointers[y] = &rgb.pixels[y * rgb.rowBytes];
     }
     png_read_image(png, rowPointers);
+    if (!ignoreICC && hasGama && hasChrm && !hasSrgb && !hasICC) {
+        if (!avifIsSrgbChromaticities(chrm)) {
+            fprintf(stderr,
+                    "WARNING: PNG file contains unsupported chromaticity "
+                    "information (cHRM chunk different from sRGB). Colors may "
```
y-guyon, line 425: > cHRM chunk different from sRGB

It could be misunderstood as "sRGB chunk", but there is no such chunk here.

tests/gtest/avifreadimagetest.cc
```diff
@@ -16,43 +16,64 @@ const char* data_path = nullptr;
 
 //------------------------------------------------------------------------------
 
-bool AreSamplesEqualForAllReadSettings(const char* file_name1,
-                                       const char* file_name2) {
-  constexpr bool kIgnoreMetadata = true;
+void AreSamplesEqualForAllReadSettings(const char* file_name1,
```
y-guyon, line 19: This should be renamed to `AreSamplesSimilarForAllReadSettings()` if some leeway is allowed.

tests/gtest/avifreadimagetest.cc
```diff
@@ -16,43 +16,64 @@ const char* data_path = nullptr;
...
-      "paris_icc_exif_xmp.png", "paris_icc_exif_xmp_at_end.png"));
+  AreSamplesEqualForAllReadSettings("paris_icc_exif_xmp.png",
+                                    "paris_icc_exif_xmp_at_end.png",
+                                    /*ignoreMetadata=*/true, /*min_psnr=*/99.);
+  AreSamplesEqualForAllReadSettings("gray_gama_chrm.png",
+                                    "gray_gama_applied.png",
+                                    /*ignoreMetadata=*/false, /*min_psnr=*/99.);
+  AreSamplesEqualForAllReadSettings("gray16_gama_chrm.png",
+                                    "gray16_gama_applied.png",
+                                    /*ignoreMetadata=*/false, /*min_psnr=*/65.);
```
y-guyon, line 76: Please add a comment on why these are expected to differ.

Also consider adding a `max_psnr` to make sure they differ.

apps/shared/avifpng.c
```diff
@@ -347,6 +418,15 @@ avifBool avifPNGRead(const char * inputFilename,
...
+                    "differ.\n");
+        }
+        avifApplyGama(&rgb, (float)gama);
```
y-guyon, line 428: Please add some comment linking to a bug or describing why this is necessary

y-guyon (approved):
Thoughts:
- Could we use fixed point instead of floating point?
- Why float and not double precision?
- Is there anything in libyuv that could help for speed here?
- Is there any change in how images are read/encoded with the libavif API and/or avifenc? If so, it should be described in the CHANGELOG.

apps/shared/avifpng.c
maryla-uc: Reworded.

tests/data/gray16_gama_applied.png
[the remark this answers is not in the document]
maryla-uc: Done.

maryla-uc:
Re fixed point, it might be possible but I would guess it's not trivial since there's a bunch of pow() calls. Note that this is hopefully a fairly niche use case that shouldn't be relevant for most images. I don't think libyuv has anything that can help.
 float instead of double was just because I copied code from reformat.c which was using floats.
I missed your comment about adding the functions to avifpng.h, will address it later.

tongyuantongyu:
Just my personal opinion: I feel `libavif` converting the pixel value in this case is a bit weird: Crafting an ICC profile which resemble what `gAMA` and `cHRM` specifies seems to be the more reasonable behavior. CMS library should be unnecessary, an ICC file template is probably enough for this.

apps/shared/avifpng.c
```diff
@@ -196,6 +198,70 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
+
+static int avifCorrectGamma(uint16_t v, double gamma, int rgbMaxChannel)
+{
+    const double normalized = v / (double)rgbMaxChannel;
+    const double corrected = avifToSrgbGamma(avifToLinear(normalized, gamma));
+    const int scaledUp = (int)floor(corrected * rgbMaxChannel + 0.5);
+    return scaledUp < 0 ? 0 : scaledUp > rgbMaxChannel ? rgbMaxChannel : scaledUp;
+}
+
+// Converts samples in the image from the given 'gAMA' value to the sRGB transfer curve.
```
wantehchang, line 228: Nit: Clarify that only RGB samples are converted, or equivalently. alpha samples are excluded from the conversion.

apps/shared/avifpng.c
```diff
@@ -196,6 +198,70 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
+// Gamma-compresses a linear value using the sRGB transfer curve.
+// See section 8 of https://www.color.org/srgb.pdf
+#define SRGB_GAMMA_SLOPE 12.92
+#define SRGB_GAMMA_TRESHOLD 0.0031308
+static double avifToSrgbGamma(double linear)
+{
+    if (linear <= SRGB_GAMMA_TRESHOLD) {
+        return linear * SRGB_GAMMA_SLOPE;
+    } else {
+        return pow(linear, 1. / 2.4) * 1.055 - 0.055;
```
wantehchang, line 216: Nit: 1. => 1.0

Make the same change to line 231.

See https://google.github.io/styleguide/cppguide.html#Floating_Literals

apps/shared/avifpng.c
```diff
@@ -347,6 +423,17 @@ avifBool avifPNGRead(const char * inputFilename,
         rowPointers[y] = &rgb.pixels[y * rgb.rowBytes];
     }
     png_read_image(png, rowPointers);
+    // gAMA chunk should be applied if it has an accompanying cHRM chunk, AND there is no overriding sRGB or iCCP chunk.
+    if (!ignoreICC && hasGama && hasChrm && !hasSrgb && !hasICC) {
```
wantehchang, line 427: I checked the Chrome source code in third_party/blink/renderer/platform/image-decoders/png/png_image_decoder.cc. It seems that Chrome also builds an ICC profile (in decoded form) from cHRM and gAMA. Is this correct?

apps/shared/avifpng.c
```diff
@@ -347,6 +423,17 @@ avifBool avifPNGRead(const char * inputFilename,
         rowPointers[y] = &rgb.pixels[y * rgb.rowBytes];
     }
     png_read_image(png, rowPointers);
+    // gAMA chunk should be applied if it has an accompanying cHRM chunk, AND there is no overriding sRGB or iCCP chunk.
```
wantehchang, line 426: There is a new cICP chunk. Should we also consider that? Please ask Luca Versari, who added cICP support to Chrome.

wantehchang (commented):
Maryla: Thank you for the pull request. Please treat my review as supplemental. Yannis's approval is sufficient.

apps/shared/avifpng.c
```diff
@@ -196,6 +198,70 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
+        return pow(linear, 1. / 2.4) * 1.055 - 0.055;
+    }
+}
+
+static int avifCorrectGamma(uint16_t v, double gamma, int rgbMaxChannel)
+{
+    const double normalized = v / (double)rgbMaxChannel;
+    const double corrected = avifToSrgbGamma(avifToLinear(normalized, gamma));
+    const int scaledUp = (int)floor(corrected * rgbMaxChannel + 0.5);
+    return scaledUp < 0 ? 0 : scaledUp > rgbMaxChannel ? rgbMaxChannel : scaledUp;
```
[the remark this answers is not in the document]
maryla-uc: In theory there should be no need for clamping, but I feel like it's safer in case there's a precision error or something? But I can remove the clamping if you prefer.

apps/shared/avifpng.c
maryla-uc: Thanks, I didn't know this was in the style guide!

apps/shared/avifpng.c
maryla-uc: I've heard of it but I did not include it because as far as I can tell it's only in the draft spec for now (at https://w3c.github.io/PNG-spec/ ) and it's not supported by libpng right now.

tongyuantongyu:
> The main problem is that ICC profiles are large! The smallest sRGB profile at https://github.com/saucecontrol/Compact-ICC-Profiles is 410 bytes.

If instead of sRGB curve you only need a simple gamma, then the profile can be even smaller: [gamma1.0-bt709-example.icc.zip](https://github.com/AOMediaCodec/libavif/files/11465679/gamma1.0-bt709-example.icc.zip). This is an ICC profile I crafted with gamma = 1.0 and BT709 (sRGB) primaries and only takes 373 bytes. In the same size it can represent other gamma and primaries as well.

> On the other hand, gamma correcting the values is easy and doesn't increase storage. It does have the potential drawback of introducing precision errors, but compared to wrong colors, this still seems better to me?

From user's perspective I feel this behavior a bit inconsistent. I would want either 1) always getting color management metadata preserved, 2) throwing them away because I don't care, or 3) converting whatever into sRGB, but not sometimes doing convertion and other times not. `libavif` supports for 1 and 2, and I believe is currently delegating 3 to other image processing tools instead of trying to do it itself?

tongyuantongyu:
> libavif simply copies any ICC profile that is present, or allows the user to strip it. It never creates ICC profiles from scratch.

I instead consider `gAMA` and `cHRM` as some legacy form of color profile, so categorized "generate ICC profile from them" as the "copy profile" case.

> png colors will be slightly different from the original because of the 2.2 gamma. If you then convert this png to avif, currently it will revert to the original colors, but with this patch it would have the same colors as the png

I wonder how image viewers actually dealing with `gAMA` 2.2. I vaguely remember seeing some example PNG decode code that converts whatever gamma to 2.2 with a `pow()` call and call it done. Do actual viewers also treat gamma 2.2 as sRGB, or they will do gamma 2.2 to sRGB correction? Web browsers seem to do it, by the way.

> What do you think?

Actually there's `AVIF_TRANSFER_CHARACTERISTICS_BT470M` which is precise gamma 2.2. There's also `AVIF_TRANSFER_CHARACTERISTICS_BT470BG` for gamma 2.8 and `AVIF_TRANSFER_CHARACTERISTICS_LINEAR` for gamma 1.0. These should covers most PNG tagged with `gAMA`, and probably even most PNG with `gAMA` that's not 2.2.

I would suggest an alternative plan:

- For 2.2 `gAMA` 2.2 and sRGB `cHRM`: tag AVIF as `AVIF_TRANSFER_CHARACTERISTICS_BT709` and `AVIF_TRANSFER_CHARACTERISTICS_SRGB`, and print a warning that we detected creator of this image probably intends sRGB. (*)
- match `gAMA` value with the list [2.2, 2.8, 1.0], or use `AVIF_TRANSFER_CHARACTERISTICS_SRGB` if absent
- match `cHRM` value with `avifColorPrimariesTables`, or use `AVIF_TRANSFER_CHARACTERISTICS_BT709` if absent
- if both are either matched or absent, tag AVIF using CICP, which should cover most cases
- otherwise generate an ICC Profile from `gAMA` and/or `cHRM`, which should be really rare. We may also print a warning that we are generating ICC profile from legacy PNG color management info.

(*) I actually do not have preference on whether to tag these images `AVIF_TRANSFER_CHARACTERISTICS_SRGB` or  `AVIF_TRANSFER_CHARACTERISTICS_BT470M`. That depends on how image viewers actually interpret gamma 2.2 and the two TCs.

This would avoid adding ICC Profile for the most of time, and also avoid dealing with color management at all in libavif.

maryla-uc:
Thanks for your detailed answer. I think this sounds reasonable. You are right that we can get away with using CICP for the 2.2+srgb chrm case, which covers the vast majority of cases. I think it's ok to add ICC for <0.1% of png images.

The choice between AVIF_TRANSFER_CHARACTERISTICS_BT470M and AVIF_TRANSFER_CHARACTERISTICS_SRGB is a choice between being techically correct and inferring the user's real intent...
Different browsers and image viewers render a png gamma of 2.2 differently. Chrome and Firefox use the technically correct formula, which means that a gamma of 2.2 will appear different from sRGB gamma (this is what prompted #1139). 
However, Safari shows them the same. I tried several image viewers and some of them were close to Chrome's behavior and some closer to Safari...
One option would be change the behavior of Chrome to assume sRGB if the gamma is 2.2 and the chromaticities match sRGB. We will discuss this and maybe can decide afterwards what the value CICP should be in libavif.

By the way the png_set_gamma function in libpng itself does a simple pow() correction without using the real sRGB formula. So it doesn't allow correcting from a gamma of 2.2 to sRGB. This is why e.g. libwebp has the same issue as #1139 even though it calls png_set_gamma https://github.com/webmproject/libwebp/blob/ed274371604e3321fbe9dcf9dc31dc262f8f34b1/imageio/pngdec.c#L307

If you'd like to contribute a patch for ICC profile writing feel free to do so. If needed you can get some inspriation from the jpeg xl implementation https://github.com/libjxl/libjxl/blob/91760e31a40fb811867fbe9bda6ddf73c962389e/lib/jxl/color_management.cc#L508
Thanks!

maryla-uc:
Dropping this PR since I agree with @tongyuantongyu's proposed solution. Just for information, are you looking at this right now or in the near future? There is no rush but I just want to know if I should just wait or look at it myself.

maryla-uc:
Friendly ping @tongyuantongyu as I said there's no rush but I just wanted to check if you were working on this or planning to.

tongyuantongyu:
@maryla-uc  Sorry for the late reply. Converting primaries from xy to XYZ in ICC is more complicated than I was thinking, and I was quite busy for the last few weeks, so I didn't finished implementing it yet. Fell free if you would like to work on this, or I should be able to find some time in the next week to finish it.

Here is my half-done implementation: https://github.com/tongyuantongyu/libavif/tree/png-craft-icc . It's missing the primary color conversion part. You can work on top of it if you find it helps.

Issue #1333, opened 2023-04-11
An example of XMP metadata that triggers the error "Error writing PNG: XMP metadata contains an invalid null character" in avifpng.c


wantehchang:
Yannis: Please take a look at the XMP metadata in this image and see if avifpng.c is rejecting it correctly.

Steps to reproduce:

1. Download moon.jpg from https://crbug.com/1405727.
2. avifenc moon.jpg moon.avif
3. avifdec moon.avif moon.png

The avifdec command fails with the error message:

> Error writing PNG: XMP metadata contains an invalid null character

wantehchang:
Running avifdec in the debugger, I found that the null character is at the end of the `avif->xmp.data` buffer.

wantehchang:
The XMP specification says "When XMP is encoded as UTF-8, there are no zero bytes in the XMP Packet." It seems that the XMP metadata in moon.jpg has an XMP packet that is incorrectly null-terminated.

y-guyon:
> Error writing PNG: XMP metadata contains an invalid null character

comes from

https://github.com/AOMediaCodec/libavif/blob/d48836b9e6d34a6a090e49eb3490fdf69eee90a6/apps/shared/avifpng.c#L481-L485

I assume the issue is that `avifenc` can generate a file that `avifdec` refuses. Solutions I can think of:

- Truncate the XMP chunk during PNG import when it ends by a null character (suggested solution)
- Truncate the XMP chunk during PNG export when it ends by a null character
- Truncate the XMP chunk during AVIF export when it ends by a null character
- Remove the XMP chunk during PNG import when it contains a null character
- Remove the XMP chunk during PNG export when it contains a null character
- Remove the XMP chunk during AVIF export when it contains a null character
- Replace null characters in XMP chunk by spaces or newlines or some other delimiter during AVIF export
- Remove the check during PNG export
- Add the check during PNG import

Opinions?

kmilos:
Fond it - take a look at https://github.com/Exiv2/exiv2/pull/2133 and the related issue report

wantehchang:
We need to first make sure the terminating null byte comes from moon.jpg rather than our code in `avifJPEGRead()`. That is, we should verify that libjpeg does not null-terminate the data in `marker->data` and include the appended null byte in `marker->data_length`. I believe this is the case, but it would be good for another person to doublecheck it.

[...]

PR #1391, opened 2023-05-12
Call avifFixImageXMP() during jpg,png reading

y-guyon:
Strip one null character from XMP chunks extracted from JPG and PNG files. Add regression test in avifmetadatatest.

See https://github.com/AOMediaCodec/libavif/issues/1333.


CHANGELOG.md
```diff
@@ -65,6 +65,8 @@ List of incompatible ABI changes in this release:
 * At decoding, avifIOStats now returns the same values as at encoding.
 * avifRGBImageAllocatePixels() now returns avifResult instead of void to report
   memory allocation failures.
+* avifReadImage(), avifJPEGRead() and avifPNGRead() now remove the trailing zero
+  byte from read XMP chunks, if any. See avifFixImageXMP().
```
wantehchang, line 69: Nit: avifFixImageXMP => avifImageFixXMP

apps/shared/avifutil.h
```diff
@@ -72,6 +72,9 @@ avifAppFileFormat avifReadImage(const char * filename,
                                 avifAppSourceTiming * sourceTiming,
                                 struct y4mFrameIterator ** frameIter);
 
+// Removes a single trailing null character from the image->xmp, if there is exactly one.
+void avifFixImageXMP(avifImage * image);
```
wantehchang, line 76: Nit: avifFixImageXMP => avifImageFixXMP

to follow libavif's objected-oriented function naming convention.

Alternatively, this function can take &image->xmp as input.

apps/shared/avifutil.c
```diff
@@ -277,6 +277,26 @@ avifAppFileFormat avifReadImage(const char * filename,
     return format;
 }
 
+void avifFixImageXMP(avifImage * image)
+{
+    if (!image->xmp.data) {
+        return;
+    }
```
wantehchang, line 284: Optional: Delete this if statement.

We can rely on the property that if `image->xmp.size` is nonzero, then `image->xmp.data` is non-NULL. The next if statement only dereferences `image->xmp.data` when `image->xmp.size` is nonzero.

tests/data/README.md
```diff
@@ -106,6 +106,22 @@ The goal is to have a large XMP blob so that it can only be stored as multiple e
 |   136627 | 0xffe1 APP1 |   4791 | http://ns.adobe.com/xmp/extensio |
 |          |             |        | ...                              |
 
+### File [paris_xmp_trailing_null.jpg](paris_xmp_trailing_null.jpg)
+
+![](paris_xmp_trailing_null.jpg)
+
+License: [same as libavif](https://github.com/AOMediaCodec/libavif/blob/main/LICENSE)
+
+Source: `paris_exif_xmp_icc.jpg` loaded with `avifReadImage()`, stripped from ICC and Exif, a zero byte appended to XMP,
```
wantehchang, line 115: Nit: stripped from => stripped of

tests/gtest/avifmetadatatest.cc
```diff
@@ -413,6 +414,32 @@ TEST(MetadataTest, MultipleExtendedXMPAndAlternativeGUIDTag) {
 
 //------------------------------------------------------------------------------
 
+// Regression test for https://github.com/AOMediaCodec/libavif/issues/1333.
+// Coverage for avifFixImageXMP().
```
wantehchang, line 418: Nit: avifFixImageXMP => avifImageFixXMP

tests/gtest/avifmetadatatest.cc
```diff
@@ -413,6 +414,32 @@ TEST(MetadataTest, MultipleExtendedXMPAndAlternativeGUIDTag) {
...
+  testutil::AvifImagePtr jpg =
+      testutil::ReadImage(data_path, "paris_xmp_trailing_null.jpg");
+  ASSERT_NE(jpg, nullptr);
+  ASSERT_NE(jpg->xmp.size, 0u);
+  // avifJPEGRead() should strip the trailing null character.
+  ASSERT_EQ(std::memchr(jpg->xmp.data, '\0', jpg->xmp.size), nullptr);
+
+  // Append a zero byte to see what happens when encoded with libpng.
+  avifRWDataRealloc(&jpg->xmp, jpg->xmp.size + 1);
+  jpg->xmp.data[jpg->xmp.size] = '\0';
```
wantehchang, line 429: IMPORTANT: Subtract one from `jpg->xmp.size`:

```
  jpg->xmp.data[jpg->xmp.size - 1] = '\0';
```

tests/gtest/avifmetadatatest.cc
```diff
@@ -413,6 +414,32 @@ TEST(MetadataTest, MultipleExtendedXMPAndAlternativeGUIDTag) {
...
+  testutil::WriteImage(jpg.get(),
+                       (testing::TempDir() + "xmp_trailing_null.png").c_str());
```
wantehchang, line 431: Optional: It would be better to construct the file with the trailing null byte and add it as a test data file. Then we don't need to worry about libpng export taking care of this as described by the comment at line 436.

wantehchang (approved):
LGTM. Thanks for the fix. The new test has a buffer overflow bug (off by one error).

tests/gtest/avifmetadatatest.cc
y-guyon: I believe libpng import may also strip the trailing null byte, so constructing a file to skip the export step will not bypass libpng taking of this entirely anyway.

wantehchang (approved):
LGTM.

PR #1392, opened 2023-05-12
Fix nits.

vrabaud:
As mentioned in https://github.com/AOMediaCodec/libavif/pull/999


y-guyon (approved):
I suggest a more descriptive commit name (can be replaced during squash).

wantehchang (commented):
LGTM. Thanks.

PR #1393, opened 2023-05-12
Fail upon unknown tile item type in grid

y-guyon:


src/read.c
```diff
@@ -1353,86 +1339,93 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
...
+
+        if (firstTileItem == NULL) {
+            firstTileItem = item;
+
+            // Adopt the configuration property of the first image item tile, so that it can be queried from
+            // the top-level color/alpha item during avifDecoderReset().
+            const char * configPropName = memcmp(item->type, "av02", 4) ? "av1C" : "av2C";
+            const avifProperty * srcProp = avifPropertyArrayFind(&item->properties, configPropName);
+            if (!srcProp) {
+                avifDiagnosticsPrintf(&decoder->diag, "Grid image's first tile is missing an %s property", configPropName);
```
jzern, line 1391: `tile` or its contents need to be cleaned up here and in other locations, or is the cleanup expected when the backing array is destroyed?

jzern (approved):

src/read.c
y-guyon: I see that `avifDecoderDataDestroy()` calls `avifDecoderDataClearTiles()` so it should be fine.

src/read.c
```diff
@@ -1087,58 +1087,44 @@ static avifResult avifDecoderItemValidateProperties(const avifDecoderItem * item
...
             avifDecoderItem * tile = &item->meta->items.item[i];
-            if (tile->dimgForID == item->id) {
-                // Skip the same items as avifDecoderGenerateImageGridTiles().
-                if (avifGetCodecType(item->type) == AVIF_CODEC_TYPE_UNKNOWN) {
-                    continue;
-                }
+            if (tile->dimgForID != item->id) {
+                continue;
+            }
+            // Tile item types were checked in avifDecoderGenerateImageTiles(), no need to do it here.
```
wantehchang, line 1095: Yannis: The function that checks tile item types is `avifDecoderGenerateImageGridTiles`. Since `avifDecoderGenerateImageTiles` calls `avifDecoderGenerateImageGridTiles`, this comment is still correct. But I was wondering if you meant to say "avifDecoderGenerateImageGridTiles" here.

wantehchang: Thanks for the confirmation. I thought that might be why you used "avifDecoderGenerateImageTiles" in the comment and wanted to confirm it with you.

Issue #1203, opened 2022-11-10
AVIF color grid without alpha grid looses alpha when decoded


y-guyon:
This is an image encoded with `avifenc --grid 1x2` (at cd03eef10a7cc6081ee98e3911e21d977a2bf17b using `ext/aom`):

![colorgrid_alphagrid avif](https://user-images.githubusercontent.com/51164219/201102557-104c8b31-5f51-48aa-8016-584b058cd5af.jpg)
(extension is .jpg but it is an actual AVIF file)

The pattern is the following:

```
[primary item grid]
   ^       ^     ^
   |dimg   |dimg |
   |       |     |auxl
[color] [color]  |
                 |
   [alpha aux grid]
      ^       ^
      |dimg   |dimg
      |       |
   [alpha] [alpha]
```

Here is the same image manually edited to match the following pattern:

![colorgrid_alpha_alpha avif](https://user-images.githubusercontent.com/51164219/201103527-92f8fc57-16b4-411f-9df2-d4934474fc27.jpg)

```
[primary item grid]
   ^       ^
   |dimg   |dimg
   |       |
[color] [color]
   ^       ^
   |auxl   |auxl
   |       |
[alpha] [alpha]
```

I did not find anything in ISOBMFF, HEIF and MIAF specifications regarding the invalidity of this pattern. [Compliance Warden](https://gpac.github.io/ComplianceWarden-wasm/avif.html) does not report more errors than for the image generated only with `avifenc` (even fewer errors actually because there is one less grid to complain about `construction_method=-1 on a derived image item`).

`avifdec` (at cd03eef10a7cc6081ee98e3911e21d977a2bf17b using `ext/aom`) leads to the wrongly opaque following PNG:

![colorgrid_alpha_alpha_libavif](https://user-images.githubusercontent.com/51164219/201104722-5379fa5c-ba81-419c-be14-1de368d50561.png)

There is the same issue with libheif: https://github.com/strukturag/libheif/issues/708

y-guyon:
[WD of ISO/IEC 23008-12 4th edition Image file format](https://www.mpeg.org/wp-content/uploads/mpeg_meetings/154_SantaEularia/w26186.zip) says:

> **6.3 Derivation of an output image of an image item**
>
> Auxiliary image item associations are **not** carried over from the input image items to the derived image item.
>
[...]

PR #1396, opened 2023-05-16
Allow alpha to be specified as an auxl of color tiles

vigneshvg:
Allows images of the following form to be decoded as images with alpha channel (since the spec does not disallow these):

```
[primary item grid]
   ^       ^
   |dimg   |dimg
   |       |
[color] [color]
   ^       ^
   |auxl   |auxl
   |       |
[alpha] [alpha]
```

Fixes Issue #1203.

I have tested this commit with the image attached in issue #1203 and it decodes as intended with the alpha plane.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3569,40 +3569,111 @@ static avifResult avifDecoderCreateCodecs(avifDecoder * decoder)
     return AVIF_RESULT_OK;
 }
 
-// If alpha is AVIF_FALSE, searches for the primary color item (parentItemID is ignored in this case).
-// If alpha is AVIF_TRUE, searches for the auxiliary alpha item whose parent item ID is parentItemID.
-// Returns the target item if found, or NULL.
-static avifDecoderItem * avifDecoderDataFindItem(avifDecoderData * data, avifBool alpha, uint32_t parentItemID)
+// Returns AVIF_TRUE if the item should be skipped. Items should be skipped for one of the following reasons:
+//  * Size is 0.
+//  * Has an essential property that isn't supported by libavif.
+//  * Item is Exif or similar data.
+//  * Item is a thumbnail.
+static avifBool avifDecoderDataShouldSkipItem(avifDecoderItem * item)
+{
+    return !item->size || item->hasUnsupportedEssentialProperty ||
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -3569,40 +3569,113 @@ static avifResult avifDecoderCreateCodecs(avifDecoder * decoder)
...
+                }
+            }
+        }
+        if (alphaItemCount != colorItemCount) {
+            // Not all the color items had an alpha auxiliary attached to it. Report this case as an image without alpha channel.
+            avifFree(alphaItemIndices);
+            *alphaItem = NULL;
+            return AVIF_RESULT_OK;
+        }
+        *alphaItem = avifMetaFindItem(colorItem->meta, maxItemID + 1);
```
y-guyon, line 3661: `maxItemID+1` can be replaced with `colorItem->meta->items.count`

y-guyon (approved):
Thanks for the fix.

src/read.c
vigneshvg: From what i understand, item IDs need not be consecutive (in the sample file that you made, the item ids are 1,2,3,5,6).

So i am iterating through all the items in the file and picking the largest ID and adding 1 to it to make an ID for this fake item.

src/read.c
```diff
@@ -3569,40 +3569,113 @@ static avifResult avifDecoderCreateCodecs(avifDecoder * decoder)
...
         }
-        if (alpha && (item->auxForID == parentItemID)) {
-            // Is this an alpha auxiliary item of the parent color item?
-            const avifProperty * auxCProp = avifPropertyArrayFind(&item->properties, "auxC");
-            if (auxCProp && isAlphaURN(auxCProp->u.auxC.auxType)) {
-                return item;
+        uint32_t * alphaItemIndices = avifAlloc(colorItemCount * sizeof(uint32_t));
+        if (alphaItemIndices == NULL) {
+            return AVIF_RESULT_OUT_OF_MEMORY;
+        }
```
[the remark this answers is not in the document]

wantehchang (commented):
Vignesh: LGTM. Thanks for the fix. I am wondering if we can add a test the image attached in issue #1203.

I also posted some comments in a copy of this pull request at https://aomedia-review.googlesource.com/c/libavif/+/175421.

Issue #841, opened 2022-02-09
Synchronize the placement of AV1 key frames when encoding an animated AVIF file with alpha


wantehchang:
We discovered this issue while investigating [Chrome bug 1274200](https://crbug.com/1274200). When encoding an animated AVIF file with alpha, the libavif encoder creates two AV1 encoders to encode the color planes and the alpha plane separately. This issue stems from the fact that by default, we let these two AV1 encoders pick their own key frames. So it is possible that for a particular frame, the color planes are encoded as an AV1 key frame but the alpha plane is not.

If `avifDecoderIsKeyframe()` only checks the `sync` flag of the color planes (which is the case before commit https://github.com/AOMediaCodec/libavif/commit/91bc33a6bd761f70013d7f0f041262626e1d5cad), then we may seek to a frame in which the color planes are an AV1 key frame but the alpha plane isn't an AV1 key frame, which will cause an alpha plane decode failure.

The following is the solution outlined by Joe Drago:

We can avoid this problem by detecting when we have just encoded a color frame that happens to be a key frame,
 and forcing all other tiles/planes for that specific frame to also be a key frame. There's a point where we walk all tiles in src/write.c and encode all of them. The fix would simply be to check if the first tile we encoded is a key frame,
 and if so, force it on all other tiles for this frame
 using the flag `AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME`.

TroyKomodo:
👍 Thank you <3

vigneshvg:
This issue is now fixed when using libaom.

rav1e and SVT-AV1 do not support disabled lagged encoding, so they may not sync the keyframes between the color and the alpha planes.

PR #1399, opened 2023-05-17
Use global configs for keyframe interval

vigneshvg:
Instead of computing the required keyframe interval in libavif code and passing the "force keyframe" flag on individual frames, use each encoder's global configuration parameter to set the maximum keyframe interval.

For libaom: `kf_max_dist`.
For rav1e: `key_frame_interval`.
For svt-av1: `intra_period_length` and `force_key_frames`.

Here's the output obtained from various encoders for encoding an animated file with 10 frames with different values for --keyframe. Each line is read as follows:
--keyframe value: list of key frames before => list of keyframes after.

aom:
 --keyframe 0: same before and after.
 --keyframe 1: same before and after.
 --keyframe 2: 1,2,3,4,5,6,7,8,9,10 => 1,3,5,7,9 (the old behavior was not incorrect, but new one is exact).
 --keyframe 5: 1,2,6,7 =>1,6 (the old behavior was not incorrect, but new one is exact).

rav1e:
  * no changes. same output before and after.

svt-av1:
 --keyframe 0: same before and after.
 --keyframe 1: 1 => 1,2,3,4,5,6,7,8,9,10 (old behavior was incorrect).
 --keyframe 2: 1 => 1,3,5,7,9 (old behavior was incorrect).
 --keyframe 5: 1 => 1,6 (old behavior was incorrect).

To sum up:
 * svt-av1 was not producing the intended output with --keyframe before this change.
 * libaom had some weirdness in keyframe spacing when using the force keyframe flag on each frame and that weirdness goes away when using the kf_maxdist configuration (see: https://bugs.chromium.org/p/aomedia/issues/detail?id=3445).

The `AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME` flag is retained in the codebase since there are some follow-up changes that will make use of the flag to synchronize keyframe placement between the color and alpha plane encoders (Issue #841).

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -111,7 +111,7 @@ static void syntax(void)
     printf("    --duration D                      : Set all following frame durations (in timescales) to D; default 1. Can be set multiple times (before supplying each filename)\n");
     printf("    --timescale,--fps V               : Set the timescale to V. If all frames are 1 timescale in length, this is equivalent to frames per second (Default: 30)\n");
     printf("                                        If neither duration nor timescale are set, avifenc will attempt to use the framerate stored in a y4m header, if present.\n");
-    printf("    -k,--keyframe INTERVAL            : Set the forced keyframe interval (maximum frames between keyframes). Set to 0 to disable (default).\n");
+    printf("    -k,--keyframe INTERVAL            : Set the keyframe interval (Aany set of INTERVAL consecutive frames will have at least one keyframe). Set to 0 to disable (default).\n");
     printf("    --ignore-exif                     : If the input file contains embedded Exif metadata, ignore it (no-op if absent)\n");
     printf("    --ignore-xmp                      : If the input file contains embedded XMP metadata, ignore it (no-op if absent)\n");
     printf("    --ignore-icc                      : If the input file contains an embedded ICC profile, ignore it (no-op if absent)\n");
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1138,7 +1138,8 @@ typedef struct avifEncoder
     // settings (see Notes above)
     int maxThreads;
     int speed;
-    int keyframeInterval;     // How many frames between automatic forced keyframes; 0 to disable (default).
+    int keyframeInterval;     // Any set of |keyframeInterval| consecutive frames will have at least one keyframe. When it is 0,
+                              // there is no such restriction.
     uint64_t timescale;       // timescale of the media (Hz)
     int repetitionCount;      // Number of times the image sequence should be repeated. This can also be set to
                               // AVIF_REPETITION_COUNT_INFINITE for infinite repetitions.  Only applicable for image sequences.
--- a/src/codec_rav1e.c
+++ b/src/codec_rav1e.c
@@ -172,6 +172,12 @@ static avifResult rav1eCodecEncodeImage(avifCodec * codec,
                 goto cleanup;
             }
         }
+        if (encoder->keyframeInterval > 0) {
+            // "key_frame_interval" is the maximum interval between two keyframes.
+            if (rav1e_config_parse_int(rav1eConfig, "key_frame_interval", encoder->keyframeInterval) == -1) {
+                goto cleanup;
+            }
+        }
 
         rav1e_config_set_color_description(rav1eConfig,
                                            (RaMatrixCoefficients)image->matrixCoefficients,
--- a/src/codec_svt.c
+++ b/src/codec_svt.c
@@ -161,6 +161,18 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
             svt_config->profile = HIGH_PROFILE;
         }
 
+        // In order for SVT-AV1 to force keyframes by setting pic_type to
+        // EB_AV1_KEY_PICTURE on any frame, force_key_frames has to be set.
+        svt_config->force_key_frames = TRUE;
+
+        // keyframeInterval == 1 case is handled when encoding each frame by
+        // setting pic_type to EB_AV1_KEY_PICTURE. For keyframeInterval > 1,
+        // set the intra_period_length. Eventhough setting intra_period_length
+        // to 0 should work in this case, it does not.
+        if (encoder->keyframeInterval > 1) {
+            svt_config->intra_period_length = encoder->keyframeInterval - 1;
+        }
+
         res = svt_av1_enc_set_parameter(codec->internal->svt_encoder, svt_config);
         if (res == EB_ErrorBadParameter) {
             goto cleanup;
@@ -199,7 +211,7 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
     input_buffer->pts = 0;
 
     EbAv1PictureType frame_type = EB_AV1_INVALID_PICTURE;
-    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME) {
+    if ((addImageFlags & AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME) || (encoder->keyframeInterval == 1)) {
         frame_type = EB_AV1_KEY_PICTURE;
     }
     input_buffer->pic_type = frame_type;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ List of incompatible ABI changes in this release:
 * Check the return value of avifEncoderSetCodecSpecificOption().
 * The maxThreads member was added to the avifRGBImage struct.
 * Check the return value of avifRGBImageAllocatePixels().
+* The meaning of the keyframeInterval member of avifEncoder struct has changed
+  slightly. When set to a value of "n",
+    * Before: It forces a keyframe on every nth frame.
+    * After: Any set of "n" consecutive frame will have atleast one keyframe (every nth frame may or may not be a keyframe).
 
 ### Added
 * Add STATIC library target avif_internal to allow tests to access functions
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -705,6 +705,10 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             cfg->kf_mode = AOM_KF_DISABLED;
             // Tell libaom that all frames will be key frames.
             cfg->kf_max_dist = 0;
+        } else {
+            if (encoder->keyframeInterval > 0) {
+                cfg->kf_max_dist = encoder->keyframeInterval;
+            }
         }
         if (encoder->extraLayerCount > 0) {
             cfg->g_limit = encoder->extraLayerCount + 1;
--- a/src/write.c
+++ b/src/write.c
@@ -1209,10 +1209,6 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
     // -----------------------------------------------------------------------
     // Encode AV1 OBUs
 
-    if (encoder->keyframeInterval && ((encoder->data->frames.count % encoder->keyframeInterval) == 0)) {
-        addImageFlags |= AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME;
-    }
-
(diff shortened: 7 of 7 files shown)


wantehchang:
Vignesh: Thank you for the pull request. I will take a look tomorrow.

src/codec_rav1e.c
```diff
@@ -172,6 +172,11 @@ static avifResult rav1eCodecEncodeImage(avifCodec * codec,
                 goto cleanup;
             }
         }
+        if (encoder->keyframeInterval > 0) {
+            if (rav1e_config_parse_int(rav1eConfig, "key_frame_interval", encoder->keyframeInterval) == -1) {
```
wantehchang, line 176: Nit: add a comment to document that "key_frame_interval" is the maximum interval between two keyframes.

src/codec_svt.c
```diff
@@ -161,6 +161,15 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
             svt_config->profile = HIGH_PROFILE;
         }
 
+        if (encoder->keyframeInterval > 1) {
+            svt_config->intra_period_length = encoder->keyframeInterval - 1;
+        } else if (encoder->keyframeInterval == 1) {
+            // For this case, SV1-AV1 seems to need this flag set in svt_config
```
wantehchang, line 167: Optional: "Seems to" makes this comment sound speculative. It would be good to have an SVT-AV1 team member review this so we can write this comment with a definitive tone.

wantehchang (changes requested):
In addition to the comments below, there are two typos in the commit message.

(1) In the following line:

```
 --keyframe 2: 1,2,3,4,5,6,7,8,9,19 => 1,3,5,7,9 (the old behavior was not incorrect, but new one is exact).
```

"19" seems to be a typo. Did you mean "10"?

(2) Change "kf_maxdist" to "kf_max_dist".

src/codec_rav1e.c
vigneshvg: Done.

src/codec_svt.c
vigneshvg: I have updated the comment to remove the speculation. But like you suggested, we can consult the SVT-AV1 team to make sure this is right.

src/codec_svt.c
```diff
@@ -199,7 +208,7 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
     input_buffer->pts = 0;
 
     EbAv1PictureType frame_type = EB_AV1_INVALID_PICTURE;
-    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME) {
+    if ((addImageFlags & AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME) || (encoder->keyframeInterval == 1)) {
```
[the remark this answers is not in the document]
vigneshvg: Ack.

wantehchang, line 211: @hassount Hassene:

Could you review our changes in this file? We want to set SVT-AV1's equivalent of libaom's `kf_max_dist` to implement the maximum keyframe interval.

We also have a question: the libavif encoder has an `AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME` flag to force a given animation frame to be a keyframe. We implemented that by setting `input_buffer->pic_type` to `EB_AV1_KEY_PICTURE` here, but that doesn't work. Do you know why?

hassount: @wantehchang 
I am not sure I understand what we're trying to do with this:
>Could you review our changes in this file? We want to set SVT-AV1's equivalent of libaom's kf_max_dist to implement the maximum keyframe interval.

In general, SVT has only one parameter for the key-frame setting and lines 164-165 seem to be the right way of setting it. that said to set 
`else if (encoder->keyframeInterval == 1) {`
`svt_config->intra_period_length =` should be just set to `0`.

Now as for forcing the key frames, I believe you'd need to set both `force_key_frames` set to 1 (once for the sequence) and `input_buffer->pic_type to EB_AV1_KEY_PICTURE`

vigneshvg: > @wantehchang I am not sure I understand what we're trying to do with this:
> 
> > Could you review our changes in this file? We want to set SVT-AV1's equivalent of libaom's kf_max_dist to implement the maximum keyframe interval.
> 
> In general, SVT has only one parameter for the key-frame setting and lines 164-165 seem to be the right way of setting it. that said to set `else if (encoder->keyframeInterval == 1) {` `svt_config->intra_period_length =` should be just set to `0`.
> 

i tried setting `svt_config->intra_period_length = 0` but it did not work as intended. (i.e.) the expectation is that all the frames will be keyframes in that case. but the output i got was the first frame was keyframes and subsequent frames were not.

any idea what i am missing?

> Now as for forcing the key frames, I believe you'd need to set both `force_key_frames` set to 1 (once for the sequence) and `input_buffer->pic_type to EB_AV1_KEY_PICTURE`

this is what this PR is now doing for the "all frames are key frames case".

include/avif/avif.h
```diff
@@ -1138,7 +1138,7 @@ typedef struct avifEncoder
     // settings (see Notes above)
     int maxThreads;
     int speed;
-    int keyframeInterval;     // How many frames between automatic forced keyframes; 0 to disable (default).
+    int keyframeInterval;     // Maximum interval between two keyframes; 0 to disable (default).
```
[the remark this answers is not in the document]
vigneshvg: Done.

src/codec_svt.c
hassount: @1480c1 did we change the API to remove the +1 for intra_period ?
@vigneshvg can you try svt_config->intra_period_length = 1 instead ?

vigneshvg: > @vigneshvg can you try svt_config->intra_period_length = 1 instead ?

yes, but that just outputs a keyframe every other frame. in this case, we want every frame to be a keyframe.

hassount: this is weird then. either 0 or 1 should work @1480c1 any thoughts on this ?

vigneshvg:
@wantehchang I have updated `codec_svt.c` to set force_key_frames unconditionally and documented why.

i have also updated the documentation in `avif.h` and `avifenc` to clealy define what we mean by "interval" (i.e.) the maximum allowed difference in frame indexes of any two subsequent keyframes.

Please take another look at this PR.

apps/avifenc.c
```diff
@@ -111,7 +111,7 @@ static void syntax(void)
     printf("    --duration D                      : Set all following frame durations (in timescales) to D; default 1. Can be set multiple times (before supplying each filename)\n");
     printf("    --timescale,--fps V               : Set the timescale to V. If all frames are 1 timescale in length, this is equivalent to frames per second (Default: 30)\n");
     printf("                                        If neither duration nor timescale are set, avifenc will attempt to use the framerate stored in a y4m header, if present.\n");
-    printf("    -k,--keyframe INTERVAL            : Set the forced keyframe interval (maximum frames between keyframes). Set to 0 to disable (default).\n");
+    printf("    -k,--keyframe INTERVAL            : Set the maximum keyframe interval (maximum difference between the frame indexes of any two subsequent keyframes). Set to 0 to disable (default).\n");
```
wantehchang, line 114: Nit: "subsequent" doesn't sound right. Maybe "consecutive"?

("consecutive" may imply there are no frames in between, so it is also a little confusing.)

Perhaps we can go back to the previous description for now (you can find it in patchset 3 at https://aomedia-review.googlesource.com/c/libavif/+/175462) and then try to come up with a better description later.

src/codec_svt.c
```diff
@@ -161,6 +161,17 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
             svt_config->profile = HIGH_PROFILE;
         }
 
+        // In order for SVT-AV1 to force keyframes by setting pic_type to
+        // EB_AV1_KEY_PICTURE on any frame, force_key_frames has to be set.
+        svt_config->force_key_frames = 1;
```
wantehchang, line 166: Nit/optional: 1 => TRUE

Here is the relevant code in SVT-AV1:

```
    /**
     * @brief Signal that force-key-frames is enabled.
     *
     */
    Bool force_key_frames;
```

and

```
typedef uint8_t Bool;
#define FALSE 0
#define TRUE 1
```

src/codec_svt.c
```diff
@@ -161,6 +161,17 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
...
+
+        // keyframeInterval == 1 case is handled when encoding each frame by
+        // settings pic_type to EB_AV1_KEY_PICTURE. For keyframeInterval > 1,
```
wantehchang, line 169: settings => setting

Note: It would be good to mention that setting intra_period_length to 0 doesn't work, even though it should.

I found the following comment and code in SVT-AV1:

```
            // If the Intra period length is 0, then introduce an intra for every picture
            if (scs->static_config.intra_period_length == 0)
                pcs->cra_flag = TRUE;
```

wantehchang (approved):
LGTM.

apps/avifenc.c
vigneshvg: I updated the wording based on your discussion with @joedrago.

i think the new one looks unambiguous.

src/codec_svt.c
vigneshvg: Done.

src/codec_svt.c
vigneshvg: Done.

vigneshvg:
@wantehchang i am submitting this PR based on your earlier approval. please let me know if you have any further comments and i'll address them in a follow-up.

apps/avifenc.c
```diff
@@ -111,7 +111,7 @@ static void syntax(void)
     printf("    --duration D                      : Set all following frame durations (in timescales) to D; default 1. Can be set multiple times (before supplying each filename)\n");
     printf("    --timescale,--fps V               : Set the timescale to V. If all frames are 1 timescale in length, this is equivalent to frames per second (Default: 30)\n");
     printf("                                        If neither duration nor timescale are set, avifenc will attempt to use the framerate stored in a y4m header, if present.\n");
-    printf("    -k,--keyframe INTERVAL            : Set the forced keyframe interval (maximum frames between keyframes). Set to 0 to disable (default).\n");
+    printf("    -k,--keyframe INTERVAL            : Set the keyframe interval (Aany set of INTERVAL consecutive frames will have at least one keyframe). Set to 0 to disable (default).\n");
```
wantehchang, line 114: 1. Aany => any
2. I think it is fine to have "maximum" before "keyframe interval"

src/codec_svt.c
```diff
@@ -161,6 +161,18 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
             svt_config->profile = HIGH_PROFILE;
         }
 
+        // In order for SVT-AV1 to force keyframes by setting pic_type to
+        // EB_AV1_KEY_PICTURE on any frame, force_key_frames has to be set.
+        svt_config->force_key_frames = TRUE;
+
+        // keyframeInterval == 1 case is handled when encoding each frame by
+        // setting pic_type to EB_AV1_KEY_PICTURE. For keyframeInterval > 1,
+        // set the intra_period_length. Eventhough setting intra_period_length
```
wantehchang, line 170: Eventhough => Even though

wantehchang (commented):
LGTM.

apps/avifenc.c
wantehchang: Fixed in https://github.com/AOMediaCodec/libavif/pull/1406.

src/codec_svt.c
wantehchang: Fixed in https://github.com/AOMediaCodec/libavif/pull/1406.

src/codec_svt.c
wantehchang: @hassount @1480c1 I filed https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/2076 about the intra_period_length == 0 issue discussed above.

PR #1400, opened 2023-05-18
Fix a retval of avifDecoderGenerateImageGridTiles

wantehchang:
avifDecoderGenerateImageGridTiles returns avifBool. Change a return value of AVIF_RESULT_BMFF_PARSE_FAILED (copied from a function that returns avifResult) to AVIF_FALSE.

Also move an if statement into an "else" block.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1401,9 +1401,7 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
                     decoder->imageCount = tile->input->samples.count;
                 }
             }
-        }
-
-        if (memcmp(item->type, firstTileItem->type, 4)) {
+        } else if (memcmp(item->type, firstTileItem->type, 4)) {
             // MIAF (ISO 23000-22:2019), Section 7.3.11.4.1:
             //   All input images of a grid image item shall use the same coding format [...]
             // The coding format is defined by the item type.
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -1401,9 +1401,7 @@ static avifBool avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifIma
                     decoder->imageCount = tile->input->samples.count;
                 }
             }
-        }
-
-        if (memcmp(item->type, firstTileItem->type, 4)) {
+        } else if (memcmp(item->type, firstTileItem->type, 4)) {
```
wantehchang, line 1404: This change is optional. Please let me know if you'd like me to undo this change.

This mimics the control structure of the original code (before commit 63a496c80343b67e1d5e84f984cc275a2986ee43):

```
                // The coding format is defined by the item type.
                if (firstTile == NULL) {
                    firstTile = tile;
                } else if (memcmp(tile->type, firstTile->type, 4)) {
                    avifDiagnosticsPrintf(diag,
                                          "Tile item ID %u of type '%.4s' differs from other tile type '%.4s'",
                                          tile->id,
                                          (const char *)tile->type,
                                          (const char *)firstTile->type);
                }
```

vigneshvg (approved):

wantehchang (commented):
C compilers cannot detect this bug because application-defined boolean types are simply aliases of integer types and error codes (defined as macros or enums) are converted to integers automatically. I have run into this bug several times.

y-guyon (approved):

PR #1403, opened 2023-05-19
Test AVIF_ENABLE_GTEST instead of AVIF_LOCAL_GTEST

wantehchang:
In two places we are testing AVIF_LOCAL_GTEST, which seems like a typo. Test AVIF_ENABLE_GTEST instead.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -243,7 +243,7 @@ endif()
 # AV2 tests
 
 if(AVIF_CODEC_AVM)
-    if(AVIF_LOCAL_GTEST)
+    if(AVIF_ENABLE_GTEST)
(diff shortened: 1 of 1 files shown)


tests/CMakeLists.txt
```diff
@@ -243,7 +243,7 @@ endif()
 # AV2 tests
 
 if(AVIF_CODEC_AVM)
-    if(AVIF_LOCAL_GTEST)
+    if(AVIF_ENABLE_GTEST)
```
wantehchang, line 246: Yannis: I think you meant to test `AVIF_ENABLE_GTEST` here. If it is correct to test `AVIF_LOCAL_GTEST` here, please add a comment because the reason is not obvious.

y-guyon (approved):

PR #1404, opened 2023-05-22
Check for depth coherence at the lowest level for YCgCo-R

vrabaud:
For now, the depth check is done at the level of PNG and JPEG reader.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -23,11 +23,24 @@ struct YUVBlock
 
 static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBImage * rgb, avifReformatState * state)
 {
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+    const avifBool useYCgCoRe = (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE);
+    const avifBool useYCgCoRo = (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RO);
+#else
+    const avifBool useYCgCoRe = AVIF_FALSE;
+    const avifBool useYCgCoRo = AVIF_FALSE;
+#endif
     if ((image->depth != 8) && (image->depth != 10) && (image->depth != 12) && (image->depth != 16)) {
         return AVIF_FALSE;
     }
     if ((rgb->depth != 8) && (rgb->depth != 10) && (rgb->depth != 12) && (rgb->depth != 16)) {
         return AVIF_FALSE;
+    }
+    if (useYCgCoRe || useYCgCoRo) {
+        const int bitOffset = (useYCgCoRe) ? 2 : 1;
+        if (image->depth - bitOffset != rgb->depth) {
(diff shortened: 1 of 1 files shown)


src/reformat.c
```diff
@@ -204,6 +204,16 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
         return AVIF_RESULT_NOT_IMPLEMENTED;
     }
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+    if (state.mode == AVIF_REFORMAT_MODE_YCGCO_RE || state.mode == AVIF_REFORMAT_MODE_YCGCO_RO) {
+        const int bitOffset = (state.mode == AVIF_REFORMAT_MODE_YCGCO_RE) ? 2 : 1;
+        const int maxValue = (1 << (state.yuvDepth - bitOffset)) - 1;
```
wantehchang, line 210: Nit/optional: Here we can compare `state.yuvDepth` or `image->depth` with `rgb->depth` directly and skip the calculation of `maxValue`, e.g.,

```
        if (state.yuvDepth - bitOffset != rgb->depth) {
        }
```

src/reformat.c
```diff
@@ -563,6 +572,19 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
     const uint16_t yuvMaxChannel = (uint16_t)state->yuvMaxChannel;
     const float rgbMaxChannelF = state->rgbMaxChannelF;
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+    int maxValue;
+    if (state->mode == AVIF_REFORMAT_MODE_YCGCO_RE || state->mode == AVIF_REFORMAT_MODE_YCGCO_RO) {
+        const int bitOffset = (state->mode == AVIF_REFORMAT_MODE_YCGCO_RE) ? 2 : 1;
+        maxValue = (1 << (state->yuvDepth - bitOffset)) - 1;
```
wantehchang, line 579: Nit: The variable `maxValue` is now used in a much larger scope, so its name should indicate whether it's YUV or RGB.

However, I found that we can use `state->rgbMaxChannel` in the YCgCo-R code at lines 765-767, so here we can also just do a direct comparison of bit depths as I suggested in line 210.

wantehchang (commented):
Thanks for the fix.

src/reformat.c
vrabaud: Done

src/reformat.c
vrabaud: Simpler indeed. So many members ... Done.

wantehchang (approved):
LGTM.

It just occurred to me that we may be able to check for bit depth coherence inside the `avifPrepareReformatState` function, which is called by both `avifImageRGBToYUV` and `avifImageYUVToRGB`. Would you like to try that before I merge this pull request?

vrabaud:
Indeed, that is the right spot for a check.

wantehchang (approved):
LGTM. Thanks. I will merge the pull request for you.

PR #1405, opened 2023-05-22
android_jni: Handle JNI exceptions

vigneshvg:
When calling JNI functions, the default behavior is that the native code execution will continue and a pending exception will be queued to be thrown when the control goes back to the Java layer.

This is not desirable when multiple JNI calls are being made and the second one depends on the successful execution of the first. Suppress JNI level exceptions and simply return an error from those functions instead of throwing a JNI exception in the java layer. This also avoids some SIGABRTs in the native code.

### DIFF
--- a/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
+++ b/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
@@ -158,6 +158,19 @@ int getThreadCount(int threads) {
   return (threads == 0) ? android_getCpuCount() : threads;
 }
 
+// Checks if there is a pending JNI exception that will be thrown when the
+// control returns to the java layer. If there is none, it will return false. If
+// there is one, then it will clear the pending exception and return true.
+// Whenever this function returns true, the caller should treat it as a fatal
+// error and return with a failure status at the earliest instance possible.
+bool JniExceptionCheck(JNIEnv* env) {
+  if (!env->ExceptionCheck()) {
+    return false;
+  }
+  env->ExceptionClear();
+  return true;
+}
+
 }  // namespace
 
 jint JNI_OnLoad(JavaVM* vm, void* /*reserved*/) {
@@ -175,6 +188,21 @@ FUNC(jboolean, isAvifImage, jobject encoded, int length) {
   return avifPeekCompatibleFileType(&avif);
 }
 
+#define CHECK_EXCEPTION(ret)                \
+  do {                                      \
+    if (JniExceptionCheck(env)) return ret; \
+  } while (false)
+
+#define FIND_CLASS(var, class_name, ret)         \
+  const jclass var = env->FindClass(class_name); \
+  CHECK_EXCEPTION(ret);                          \
+  if (var == nullptr) return ret
+
+#define GET_FIELD_ID(var, class_name, field_name, signature, ret)          \
+  const jfieldID var = env->GetFieldID(class_name, field_name, signature); \
+  CHECK_EXCEPTION(ret);                                                    \
+  if (var == nullptr) return ret
+
 FUNC(jboolean, getInfo, jobject encoded, int length, jobject info) {
   const uint8_t* const buffer =
       static_cast<const uint8_t*>(env->GetDirectBufferAddress(encoded));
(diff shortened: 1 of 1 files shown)


android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -175,24 +183,39 @@ FUNC(jboolean, isAvifImage, jobject encoded, int length) {
...
+
+#define FIND_CLASS(var, class_name, ret)         \
+  const jclass var = env->FindClass(class_name); \
+  CHECK_EXCEPTION(ret);                          \
+  if (var == nullptr) return ret
+
+#define GET_FIELD_ID(var, class_name, field_name, signature, ret)          \
+  const jfieldID var = env->GetFieldID(class_name, field_name, signature); \
+  CHECK_EXCEPTION(ret);                                                    \
+  if (var == nullptr) return ret
```
wantehchang, line 197: Should these three macros be put inside a `do {} while (false)` block so that they behave like a statement?

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -158,6 +158,14 @@ int getThreadCount(int threads) {
   return (threads == 0) ? android_getCpuCount() : threads;
 }
 
+bool JniExceptionCheck(JNIEnv* env) {
```
wantehchang, line 161: Please document this function.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
vigneshvg: We cannot do that here since these macros declare a local variable which has to be visible in the calling scope.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
vigneshvg: Done.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -175,24 +188,39 @@ FUNC(jboolean, isAvifImage, jobject encoded, int length) {
   return avifPeekCompatibleFileType(&avif);
 }
 
+#define CHECK_EXCEPTION(ret) \
+  if (JniExceptionCheck(env)) return ret
+
+#define FIND_CLASS(var, class_name, ret)         \
+  const jclass var = env->FindClass(class_name); \
+  CHECK_EXCEPTION(ret);                          \
+  if (var == nullptr) return ret
```
wantehchang, line 197: The null pointer checks of the return values of `env->FindClass()` and `env->GetFieldID()` are also new in this pull request, right? It would be good to mention this in the commit message.

Is it a bug that we were not checking their return values before?

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -158,6 +158,19 @@ int getThreadCount(int threads) {
   return (threads == 0) ? android_getCpuCount() : threads;
 }
 
+// Checks if there is a pending JNI exception that will be thrown when the
+// control returns to the java layer. If there is none, it will return false. If
+// there is one, then it will clear the pending exception and return false.
```
wantehchang, line 163: return false => return true

wantehchang (approved):
LGTM.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -175,24 +188,39 @@ FUNC(jboolean, isAvifImage, jobject encoded, int length) {
   return avifPeekCompatibleFileType(&avif);
 }
 
+#define CHECK_EXCEPTION(ret) \
+  if (JniExceptionCheck(env)) return ret
```
jzern, line 192: I saw your comment to Wan-Teh's question. I think this one could be wrapped in a do/while though

jzern (approved):

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
vigneshvg: Done.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
vigneshvg: Done.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
vigneshvg: > The null pointer checks of the return values of `env->FindClass()` and `env->GetFieldID()` are also new in this pull request, right? It would be good to mention this in the commit message.
> 

Updated the commit message.

> Is it a bug that we were not checking their return values before?

Yes, these functions could fail in one of two ways: throw an exception or return nullptr. This PR handles both of those cases.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -158,6 +158,19 @@ int getThreadCount(int threads) {
   return (threads == 0) ? android_getCpuCount() : threads;
 }
 
+// Checks if there is a pending JNI exception that will be thrown when the
+// control returns to the java layer. If there is none, it will return false. If
+// there is one, then it will clear the pending exception and return true.
+// Whenever this function returns true, the caller should treat it as a fatal
+// error and return with a failure status at the earliest instance possible.
```
wantehchang, line 165: Nit instance => instant ?

Or change "at the earliest instance possible" to "as early as possible".

wantehchang (approved):
LGTM.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
wantehchang: Fixed in https://github.com/AOMediaCodec/libavif/pull/1408.

PR #1411, opened 2023-05-23
Force keyframe for alpha if color is a keyframe

vigneshvg:
When the encoder supports it (as of now only libaom does), disable lagged output when encoding animated images with alpha channel. When the utput of the color planes is a keyframe, force a keyframe on the alpha channel as well.

While the spec does not enforce this, and the libavif decoder can handle files without this requirement just fine, this change is still an enhancement because most decoders look at whether or not the color frame is a keyframe to determine seek points.

Fixes: Issue #841

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -357,7 +357,9 @@ typedef avifBool (*avifCodecGetNextImageFunc)(struct avifCodec * codec,
 // avifCodecEncodeImageFunc is responsible for automatic tiling if encoder->autoTiling is set to
 // AVIF_TRUE. The actual tiling values are passed to avifCodecEncodeImageFunc as parameters.
 // Similarly, avifCodecEncodeImageFunc should use the quantizer parameter instead of
-// encoder->quality and encoder->qualityAlpha.
+// encoder->quality and encoder->qualityAlpha. If disableLaggedOutput is AVIF_TRUE, then the encoder will emit the output frame
+// without any lag (if supported). Note that disableLaggedOutput is only used by the first call to this function (which
+// initializes the encoder) and is ignored by the subsequent calls.
 //
 // Note: The caller of avifCodecEncodeImageFunc always passes encoder->data->tileRowsLog2 and
 // encoder->data->tileColsLog2 as the tileRowsLog2 and tileColsLog2 arguments. Because
@@ -372,6 +374,7 @@ typedef avifResult (*avifCodecEncodeImageFunc)(struct avifCodec * codec,
                                                int tileColsLog2,
                                                int quantizer,
                                                avifEncoderChanges encoderChanges,
+                                               avifBool disableLaggedOutput,
                                                avifAddImageFlags addImageFlags,
                                                avifCodecEncodeOutput * output);
 typedef avifBool (*avifCodecEncodeFinishFunc)(struct avifCodec * codec, avifCodecEncodeOutput * output);
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -545,6 +545,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                                       int tileColsLog2,
                                       int quantizer,
                                       avifEncoderChanges encoderChanges,
+                                      avifBool disableLaggedOutput,
                                       avifAddImageFlags addImageFlags,
                                       avifCodecEncodeOutput * output)
 {
@@ -714,6 +715,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             cfg->g_limit = encoder->extraLayerCount + 1;
             // For layered image, disable lagged encoding to always get output
             // frame for each input frame.
+            cfg->g_lag_in_frames = 0;
+        }
+        if (disableLaggedOutput) {
             cfg->g_lag_in_frames = 0;
         }
         if (encoder->maxThreads > 1) {
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -535,6 +535,7 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
                                       int tileColsLog2,
                                       int quantizer,
                                       avifEncoderChanges encoderChanges,
+                                      avifBool disableLaggedOutput,
                                       avifAddImageFlags addImageFlags,
                                       avifCodecEncodeOutput * output)
 {
@@ -634,6 +635,9 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
             cfg->g_limit = encoder->extraLayerCount + 1;
             // For layered image, disable lagged encoding to always get output
             // frame for each input frame.
+            cfg->g_lag_in_frames = 0;
+        }
+        if (disableLaggedOutput) {
             cfg->g_lag_in_frames = 0;
         }
         if (encoder->maxThreads > 1) {
--- a/src/write.c
+++ b/src/write.c
@@ -941,6 +941,35 @@ static avifCodecType avifEncoderGetCodecType(const avifEncoder * encoder)
     return avifCodecTypeFromChoice(encoder->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
 }
 
+// This function is called after every color frame is encoded. It returns AVIF_TRUE if a keyframe needs to be forced for the next
+// alpha frame to be encoded, AVIF_FALSE otherwise.
+static avifBool avifEncoderDataShouldForceKeyframeForAlpha(const avifEncoderData * data,
+                                                           const avifEncoderItem * colorItem,
+                                                           avifAddImageFlags addImageFlags)
+{
+    if (!data->alphaPresent) {
+        // There is no alpha plane.
+        return AVIF_FALSE;
+    }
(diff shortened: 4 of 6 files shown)


include/avif/internal.h
```diff
@@ -357,7 +357,8 @@ typedef avifBool (*avifCodecGetNextImageFunc)(struct avifCodec * codec,
 // avifCodecEncodeImageFunc is responsible for automatic tiling if encoder->autoTiling is set to
 // AVIF_TRUE. The actual tiling values are passed to avifCodecEncodeImageFunc as parameters.
 // Similarly, avifCodecEncodeImageFunc should use the quantizer parameter instead of
-// encoder->quality and encoder->qualityAlpha.
+// encoder->quality and encoder->qualityAlpha. If disableLaggedOutput is AVIF_TRUE, then the encoder will emit the output frame
+// without any lag (if supported).
```
wantehchang, line 361: Document that `disableLaggedOutput` is only used by the first `avifCodecEncodeImageFunc` call (which initializes the AV1 or AV2 encoder) and is ignored by subsequent `avifCodecEncodeImageFunc` calls.

include/avif/internal.h
```diff
@@ -373,6 +374,7 @@ typedef avifResult (*avifCodecEncodeImageFunc)(struct avifCodec * codec,
                                                int quantizer,
                                                avifEncoderChanges encoderChanges,
                                                avifAddImageFlags addImageFlags,
+                                               avifBool disableLaggedOutput,
```
wantehchang, line 377: Nit/optional: let `addImageFlags` be the last input parameter (a common convention).

src/codec_aom.c
```diff
@@ -688,6 +689,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         cfg->g_input_bit_depth = image->depth;
         cfg->g_w = image->width;
         cfg->g_h = image->height;
+        if (disableLaggedOutput) {
+            cfg->g_lag_in_frames = 0;
+        }
```
wantehchang, line 694: Nit: Move to line 717 or 723 to be near the other code that sets `cfg->g_lag_in_frames`.

src/codec_avm.c
```diff
@@ -536,8 +536,12 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
                                       int quantizer,
                                       avifEncoderChanges encoderChanges,
                                       avifAddImageFlags addImageFlags,
+                                      avifBool disableLaggedOutput,
                                       avifCodecEncodeOutput * output)
 {
+    // TODO(vigneshv): add support for avm lagged output when necessary. Ignore it for now.
+    (void)disableLaggedOutput;
```
wantehchang, line 543: We can use the same code as codec_aom.c.

src/write.c
```diff
@@ -941,6 +941,30 @@ static avifCodecType avifEncoderGetCodecType(const avifEncoder * encoder)
     return avifCodecTypeFromChoice(encoder->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
 }
 
+static avifBool avifEncoderDataForceKeyframeForAlpha(const avifEncoderData * data, const avifEncoderItem * colorItem, avifAddImageFlags addImageFlags)
```
wantehchang, line 944: (1) Nit: Document this function. One reason is that the function name sounds like an action, so I thought this function would force a keyframe for alpha and the boolean return value indicates success/failure.

Alternatively, add "Should" or "NeedTo" to the function name?

(2) Another useful piece of info to document, which cannot be inferred from the function name, is when this function is called.

This function is called after we have encoded a frame of the color channel and before we encode the corresponding frame of the alpha channel.

src/write.c
```diff
@@ -941,6 +941,30 @@ static avifCodecType avifEncoderGetCodecType(const avifEncoder * encoder)
...
+{
+    if (!data->alphaPresent) {
+        // There is no alpha plane.
+        return AVIF_FALSE;
+    }
+    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) {
+        // Not an animated image.
+        return AVIF_FALSE;
+    }
+    if (data->frames.count == 0) {
```
wantehchang, line 954: Nit: Document that `data->frames.count` is the number of frames that have been encoded by previous `avifEncoderAddImage()` calls. This fact is important to understanding this function, but is not obvious unless I inspect the caller's code.

src/write.c
```diff
@@ -941,6 +941,30 @@ static avifCodecType avifEncoderGetCodecType(const avifEncoder * encoder)
...
+        // This is the first frame, no need to force keyframe.
+        return AVIF_FALSE;
+    }
+    const uint32_t sampleCount = colorItem->encodeOutput->samples.count;
```
wantehchang, line 958: Nit: It would also be helpful to document that colorItem->encodeOutput->samples stores the encoded frames of the color channel that the encoder has output so far.

src/write.c
```diff
@@ -941,6 +941,30 @@ static avifCodecType avifEncoderGetCodecType(const avifEncoder * encoder)
...
+    avifBool isLaggedOutput = (sampleCount == 0) || (data->frames.count != sampleCount - 1);
```
wantehchang, line 959: Nit: The following should be equivalent:

```
    avifBool isLaggedOutput = (data->frames.count + 1 != sampleCount);
```

src/write.c
```diff
@@ -1222,6 +1246,9 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
                 cellImage = paddedCellImage;
             }
             const int quantizer = item->alpha ? encoder->data->quantizerAlpha : encoder->data->quantizer;
+            // If alpha channel is present, set disableLaggedOutput to AVIF_TRUE. If the encoder supports it, this enables
+            // avifEncoderDataForceKeyframeForAlpha to force a keyframe on the alpha channel if the color channels are a keyframe
```
wantehchang, line 1250: Nit: "if the color channels are a keyframe" should be rephrased. (There is only one color channel.)

How about "whenever the color channel has encoded a keyframe" or "whenever a keyframe has been encoded in the color channel"?

wantehchang (commented):
Vignesh: Thank you very much for fixing this issue.

I suggest some changes, mostly to clarify or add comments.

include/avif/internal.h
vigneshvg: Done.

include/avif/internal.h
vigneshvg: Done.

src/codec_aom.c
vigneshvg: Done.

src/codec_avm.c
vigneshvg: we probably can, but i don't know if libavm respects this flag and produces the intended output as of now. so i just left it as a TODO for later so that it can be verified before it is enabled.

in general, i think of codec_avm as a fork of codec_aom at that point in time. i don't think we need to keep codec_aom and codec_avm arbitrarily in sync without knowing if things actually work.

src/write.c
vigneshvg: Done.

src/write.c
vigneshvg: Done.

src/write.c
vigneshvg: I changed the variable name to `colorFramesOutputSoFar`.

src/write.c
vigneshvg: Done.

src/write.c
vigneshvg: Done.

src/write.c
```diff
@@ -941,6 +941,35 @@ static avifCodecType avifEncoderGetCodecType(const avifEncoder * encoder)
     return avifCodecTypeFromChoice(encoder->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
 }
 
+// This function is called after every color frame is encoded. It returns AVIF_TRUE if a keyframe needs to be forced for the next
+// alpha plane that's encoded, AVIF_FALSE otherwise.
```
wantehchang, line 945: alpha plane => alpha frame ?

that's encoded => to be encoded ?

src/codec_avm.c
wantehchang: I understand your point. But it will be less work to just port the codec_aom.c change now.

Note that this file already sets `cfg->g_lag_in_frames` to 0 at line 631 and line 641. So if `cfg->g_lag_in_frames == 0` doesn't work, you won't be making it worse.

wantehchang (approved):
LGTM.

src/write.c
vigneshvg: Done.

src/codec_avm.c
vigneshvg: > Note that this file already sets `cfg->g_lag_in_frames` to 0 at line 631 and line 641.

Yes, because it was forked from codec_aom that way. :)

> So if `cfg->g_lag_in_frames == 0` doesn't work, you won't be making it worse.

Anyways, done.

wantehchang (approved):
LGTM. Thanks again for fixing this bug.

jzern (commented):
lgtm

PR #1415, opened 2023-05-25
CMakeLists.txt work around zlib/CMakeLists.txt bug

wantehchang:
CMakeLists.txt work around zlib/CMakeLists.txt bug

Put an include_directories() call with the source directory of ext/zlib
before add_subdirectory(ext/zlib) to work around a zlib/CMakeLists.txt
bug fixed by https://github.com/madler/zlib/pull/818.
    
The parent project's include directories at the time of an
add_subdirectory() call are used by the child project, so the order of
include_directories() and add_subdirectory() is important.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,12 +89,19 @@ endif()
 # Any proper unix environment should ignore these entire following blocks.
 option(AVIF_LOCAL_ZLIBPNG "Build zlib and libpng by providing your own copy inside the ext subdir." OFF)
 if(AVIF_LOCAL_ZLIBPNG)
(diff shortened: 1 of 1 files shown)


joedrago (approved):

vigneshvg (approved):

jzern (approved):

wantehchang:
Thanks for the review. I  forced-pushed a new version of the commit, which has the same idea as the version that Joe and Vignesh reviewed. I made the workaround as small as possible. I also explained why the two include_directories() calls are needed by add_subdirectory(ext/libpng).

PR #1416, opened 2023-05-25
appveyor.yml: No need to run git submodule update

wantehchang:
We are no longer using git submodules.


joedrago (approved):
Yes, this is correct, although it is a harmless and near-instant command to use in any repo without submodules.

PR #1417, opened 2023-05-25
avifenc: print speed as default instead of -1

### LINKED ISSUE #149: avifenc prints "Encoding with ... speed -1, ..." if --speed default is specified
If --speed default is passed to avifenc, the 'speed' variable is assigned the value AVIF_SPEED_DEFAULT, which is -1. So the following printf statement will print "speed -1":

    printf("Encoding with AV1 codec '%s' speed [%d], color QP [%d (%s) <-> %d (%s)], alpha QP [%d (%s) <-> %d (%s)], %d worker thread(s), please wait...\n",
           avifCodecName(codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE),
           speed,
           ...

I am not sure if that is intended.

vigneshvg:
When using avifenc with `--speed d` or `--speed default` print the log message as `speed [default]` instead of `speed [-1]` (-1 is the value of `AVIF_SPEED_DEFAULT`.

Fixes #149

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -819,9 +819,15 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
     snprintf(manualTilingStr, sizeof(manualTilingStr), "tileRowsLog2 [%d], tileColsLog2 [%d]", settings->tileRowsLog2, settings->tileColsLog2);
 
     const char * const codecName = avifCodecName(settings->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
-    printf("Encoding with AV1 codec '%s' speed [%d], color quality [%d (%s)], alpha quality [%d (%s)], %s, %d worker thread(s), please wait...\n",
+    char speed_str[16];
+    if (settings->speed == AVIF_SPEED_DEFAULT) {
+        strcpy(speed_str, "default");
+    } else {
+        snprintf(speed_str, sizeof(speed_str), "%d", settings->speed);
+    }
+    printf("Encoding with AV1 codec '%s' speed [%s], color quality [%d (%s)], alpha quality [%d (%s)], %s, %d worker thread(s), please wait...\n",
            codecName ? codecName : "none",
-           settings->speed,
+           speed_str,
            settings->quality,
            qualityString(settings->quality),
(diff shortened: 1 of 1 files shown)


apps/avifenc.c
```diff
@@ -819,9 +819,15 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
     snprintf(manualTilingStr, sizeof(manualTilingStr), "tileRowsLog2 [%d], tileColsLog2 [%d]", settings->tileRowsLog2, settings->tileColsLog2);
 
     const char * const codecName = avifCodecName(settings->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
-    printf("Encoding with AV1 codec '%s' speed [%d], color quality [%d (%s)], alpha quality [%d (%s)], %s, %d worker thread(s), please wait...\n",
+    char speed_str[8];
+    if (settings->speed == AVIF_SPEED_DEFAULT) {
+        strcpy(speed_str, "default");
+    } else {
+        sprintf(speed_str, "%d", settings->speed);
+    }
```
wantehchang, line 827: IMPORTNT: Both `strcpy` and `speed_str` copy to the speed_str buffer without knowing the buffer size. I believe that settings->speed can only be -1, 0, ..., 10 here, but it is better to eliminate the concern of buffer overflow.

So please add a comment explaining why the code is safe, or replace these two calls with snprintf(), or validate settings->speed (again), or use a larger speed_str buffer that is safe for any value of `int`.

wantehchang: I think we can increase the size of the speed_str buffer to 16 so that it is large enough for any value of the "%d" format, and change the sprintf call to `snprintf(speed_str, sizeof(speed_str), "%d", settings->speed)`. Then it will be clear to human readers and static analyzers, without having to read other code, that the strcpy() and snprintf() call won't overflow the speed_str buffer.

vigneshvg: `settings->speed` is clamped between -1 and 10 when the argument is passed. so the current code is safe.

but like you said 16 probably keeps it safer from static analyzers etc. Updated.

wantehchang (approved):
LGTM. Thanks a lot for fixing this bug.

PR #1422, opened 2023-06-01
Write primaries and transfer characteritics info in decoded png.

### LINKED ISSUE #1421: CICP information lost when decoding to png/jpeg
According to https://github.com/AOMediaCodec/libavif/wiki/CICP if an avif image does not contain an ICC profile, then the first two values of the CICP triplet, i.e. the color primaries and the transfer characteristics should be used to decide how to interpret the RGB values. However, this information is not stored in the decoded png/jpeg and will cause color differences if the values differ from sRGB.

For png it's possible to use the cHRM, gAMA and/or cICP chunks.
For jpeg I'm not aware of a equivalent mechanism, the solution might be to generate an icc profile.

See an [example](https://github.com/AOMediaCodec/libavif/files/11625412/Archive.zip) avif image, and the decoded versions with `avifdec original.avif decoded.png`, at revision c74526f

maryla-uc:
If there is no ICC but the cicp values are different from sRGB, they are added to the png file using the cHRM, gAMA and cICP chunks.

The cHRM and gAMA chunks are preferred since they are older and more widely understood (although some viewers may ignore them). The new cICP chunk is added if gAMA is not enough to specify the transfer characteristics.

Fixes #1421  for png.

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -535,7 +535,7 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
             // simple gamma value. Most transfer characteristics cannot be
             // represented this way. Viewers that support the cICP chunk can use
             // that instead, but older viewers might show incorrect colors.
-            if (avifTransferCharacteristicsGamma(avif->transferCharacteristics, &gamma)) {
+            if (avifTransferCharacteristicsGetGamma(avif->transferCharacteristics, &gamma) == AVIF_RESULT_OK) {
                 png_set_gAMA(png, info, 1.0f / gamma);
             }
         }
@@ -578,7 +578,7 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
     png_write_info(png, info);
 
     // Custom chunk writing, must appear after png_write_info.
-    // With AVIF, an ICC profile takes priority over ICC, but with PNG files, CICP takes priority over ICC.
+    // With AVIF, an ICC profile takes priority over CICP, but with PNG files, CICP takes priority over ICC.
     // Therefore CICP should only be written if there is no ICC profile.
     if (!hasIcc) {
         const png_byte cicp[5] = "cICP";
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -504,38 +504,40 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
 
     png_set_IHDR(png, info, avif->width, avif->height, rgbDepth, colorType, PNG_INTERLACE_NONE, PNG_COMPRESSION_TYPE_DEFAULT, PNG_FILTER_TYPE_DEFAULT);
 
-    const avifColorPrimaries primaries = avif->colorPrimaries;
-    const avifTransferCharacteristics transfer = avif->transferCharacteristics;
     const avifBool hasIcc = avif->icc.data && (avif->icc.size > 0);
-    const avifBool hasNonSrgbPrimaries = (primaries != AVIF_COLOR_PRIMARIES_UNKNOWN &&
-                                          primaries != AVIF_COLOR_PRIMARIES_UNSPECIFIED && primaries != AVIF_COLOR_PRIMARIES_BT709);
-    const avifBool hasNonSrgbTransfer =
-        (transfer != AVIF_TRANSFER_CHARACTERISTICS_UNKNOWN && transfer != AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED &&
-         transfer != AVIF_TRANSFER_CHARACTERISTICS_SRGB);
-
-    avifBool writeCicp = AVIF_FALSE;
     if (hasIcc) {
+        // If there is an ICC profile, the CICP values are irrelevant and only the ICC profile
+        // is written. If we could extract the primaries/transfer curve from the ICC profile,
+        // then they could be written in cHRM/gAMA chunks.
         png_set_iCCP(png, info, "libavif", 0, avif->icc.data, (png_uint_32)avif->icc.size);
-    } else if (hasNonSrgbPrimaries || hasNonSrgbTransfer) {
-        float primariesCoords[8];
-        avifColorPrimariesGetValues(avif->colorPrimaries, primariesCoords);
-        png_set_cHRM(png,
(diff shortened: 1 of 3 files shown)


apps/shared/avifpng.c
```diff
@@ -503,8 +503,40 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
+                     primariesCoords[4],
+                     primariesCoords[5]);
+        float gamma;
+        if (avifTransferCharacteristicsGamma(avif->transferCharacteristics, &gamma)) {
+            png_set_gAMA(png, info, 1.0f / gamma);
+        } else {
+            // The transfer characteristics cannot be represented with a simple gAma chunk.
+            // A cICP chunk is needed, but might not be understood by all viewers since it's new.
+            writeCicp = AVIF_TRUE;
+        }
```
kmilos, line 539: Just an idea: rather than making these exclusive, it could be ok to write everything that is available, and let the PNG readers parse the color encoding chunks in order of priority in the spec and their level of support?

maryla-uc: I'm adding the cICP chunk only if it's needed as a small size optimization, but I guess we don't really care, so I could simplify this. However, even when cICP is added, I *am* still adding cHR+gAMA as a fallback as you said, so it's not exclusive.

We can't easily add other chunks in case ICC is present, because we would need to parse the ICC to get the corresponding primaries/gamma. I don't think we should use avif->colorPrimaries and avif->transferCharacteristics in this case, as they are supposed to be irrelevant when an ICC is present as explained in https://github.com/AOMediaCodec/libavif/wiki/CICP

As an aside also note that the priority between CICP and ICC is different in the two formats: ICC has priority in avif, while CICP has priority in png.

Another chunk that could also be added is the sRGB chunk, but I think all viewers just assume sRGB if no color info is present? so I don't think it's really necessary but could be added if colorPrimaries is 1 and transferCharacteristics is 13.

jzern: I'm not sure we're overly concerned about png size, so simplifying things where you can might make sense.

jzern (approved):

apps/shared/avifpng.c
maryla-uc: Fair enough. I changed the logic to write as many chunks as we can, including the sRGB chunk.

include/avif/avif.h
```diff
@@ -319,6 +319,10 @@ enum
 };
 typedef uint16_t avifTransferCharacteristics; // AVIF_TRANSFER_CHARACTERISTICS_*
 
+// If the given transfer characteristics can be expressed with a simple gamma value, sets 'gamma'
+// to that value and returns AVIF_TRUE. Returns AVIF_FALSE otherwise.
+AVIF_API avifBool avifTransferCharacteristicsGamma(avifTransferCharacteristics atc, float * gamma);
```
wantehchang, line 324: Nit: Add a "Get" to the function name, to be consistent with `avifColorPrimariesGetValues`.

Optional: It may be better for a public function to return `avifResult` to indicate success/failure.

wantehchang (approved):
LGTM. I suggest a few simple changes.

apps/shared/avifpng.c
```diff
@@ -503,8 +503,42 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
+                             primariesCoords[5]);
+            }
+            float gamma;
+            // Write the transfer characteristics IF it can be represented as a
+            // simple gamma value. Most transfer characteristics cannot be
+            // represented this way. Viewers that support the cICP chunk can use
+            // that instead, but older viewers might show incorrect colors.
+            if (avifTransferCharacteristicsGamma(avif->transferCharacteristics, &gamma)) {
+                png_set_gAMA(png, info, 1.0f / gamma);
+            }
```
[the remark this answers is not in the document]
maryla-uc: I think this is fine, in general nothing in the spec suggests that they should always be together. I do know that Chrome (and maybe Firefox?) only use them if both are present. But I don't think it hurts to write one without the other, if we can't write both for some reason.

Note that to be able to write gAMA in more cases, we could write the closest approximating gamma value for a given curve, the same way that 2.2 is used to approximate the sRGB transfer curve. The main reason I didn't do this is because I didn't know what the appropriate values would be for some of the transfer curves.

apps/shared/avifpng.c
```diff
@@ -503,8 +503,42 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
         png_set_iCCP(png, info, "libavif", 0, avif->icc.data, (png_uint_32)avif->icc.size);
+    } else {
+        const avifBool isSrgb = (avif->colorPrimaries == AVIF_COLOR_PRIMARIES_BT709) &&
+                                (avif->transferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_SRGB);
+        if (isSrgb) {
+            png_set_sRGB_gAMA_and_cHRM(png, info, PNG_sRGB_INTENT_PERCEPTUAL);
+        } else {
+            if (avif->colorPrimaries != AVIF_COLOR_PRIMARIES_UNKNOWN && avif->colorPrimaries != AVIF_COLOR_PRIMARIES_UNSPECIFIED) {
+                float primariesCoords[8];
+                avifColorPrimariesGetValues(avif->colorPrimaries, primariesCoords);
```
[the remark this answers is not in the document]
maryla-uc: I think it's fine as is for now, all colorspaces are supported right now except UNKONWN and UNDEFINED. I don't see why a colorspace wouldn't be supported.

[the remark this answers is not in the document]
wantehchang: Thank you for the reply. I didn't realize `avifColorPrimariesTables` has an entry for each of the `avifColorPrimaries` enum values.

So the only way for `avifColorPrimariesGetValues` to fail now is an invalid or a "future" `avifColorPrimaries` enum value. In that case we will use AVIF_COLOR_PRIMARIES_BT709 as a reasonable default. I was just wondering if it wuld be better to not write a cHRM chunk in that case.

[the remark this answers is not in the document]
maryla-uc: Ha I see what you mean, I hadn't considered the possibility of a new value being read by an old decoder. I agree it would be better not to write the cHRM chunk in this case. I might send a follow up patch.

PR #1426, opened 2023-06-08
Handle PNG of palette color type and tRNS chunk

wantehchang:
If a PNG file has the tRNS chunk, the tRNS chunk is converted to alpha, so avifPNGRead() should use the RGBA format rather than the RGB format.

Bug: b:286447094


apps/shared/avifpng.c
```diff
@@ -294,7 +294,9 @@ avifBool avifPNGRead(const char * inputFilename,
         png_set_expand_gray_1_2_4_to_8(png);
     }
 
+    avifBool hasTrnsChunk = AVIF_FALSE;
     if (png_get_valid(png, info, PNG_INFO_tRNS)) {
+        hasTrnsChunk = AVIF_TRUE;
```
jzern, line 299: Note in webp when there's no tRNS chunk, we check: `has_alpha = !!(color_type & PNG_COLOR_MASK_ALPHA);`

https://chromium.googlesource.com/webm/libwebp/+/refs/tags/v1.3.1-rc1/imageio/pngdec.c#294

jzern (approved):

apps/shared/avifpng.c
wantehchang: Done. Thanks for the pointer. Webp's logic is equivalent but much simpler.

I found that libwebp2's logic is more sophisticated. I didn't adopt that logic in this pull request.

jzern (approved):

PR #1427, opened 2023-06-09
avifpng.c: Validate RGB row bytes in avifPNGRead()

wantehchang:
If libavif and libpng disagree on the value of RGB row bytes, this is a bug and we should not proceed.

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -372,6 +372,11 @@ avifBool avifPNGRead(const char * inputFilename,
         fprintf(stderr, "Conversion to YUV failed: %s (out of memory)\n", inputFilename);
         goto cleanup;
     }
+    size_t rowBytes = png_get_rowbytes(png, info);
+    if (rgb.rowBytes != rowBytes) {
+        fprintf(stderr, "avifPNGRead internal error: rowBytes mismatch libavif %u vs libpng %" AVIF_FMT_ZU "\n", rgb.rowBytes, rowBytes);
+        goto cleanup;
+    }
     rowPointers = (png_bytep *)malloc(sizeof(png_bytep) * rgb.height);
     for (uint32_t y = 0; y < rgb.height; ++y) {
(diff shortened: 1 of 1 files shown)


apps/shared/avifpng.c
```diff
@@ -372,6 +372,11 @@ avifBool avifPNGRead(const char * inputFilename,
         fprintf(stderr, "Conversion to YUV failed: %s (out of memory)\n", inputFilename);
         goto cleanup;
     }
+    size_t rowBytes = png_get_rowbytes(png, info);
```
y-guyon, line 375: Can be `const`

y-guyon (approved):
Did you encounter this issue? I wonder how that can happen.

jzern:
> Did you encounter this issue? I wonder how that can happen.

I was wondering the same. My first reaction was maybe it should be an assert, though the two values are disconnected and one comes from libpng (though width does too), so this is a reasonable defense.

wantehchang:
Thanks for the review. This is not a check for invalid PNG files. This is a self check, to make sure libavif is calling png_read_image() correctly. Note that png_read_image() receives only the row buffer pointers but not the row buffer size. So it is important to know what libpng thinks the buffer size is. An assertion, compiled away in release builds, is not good enough.

This check would have detected the bug fixed by https://github.com/AOMediaCodec/libavif/pull/1426 (where the buffer size was smaller than what libpng expected), and it would also help detect a similar bug where the buffer size is larger than what libpng expects.

I will add a comment to explain the purpose of this check.

apps/shared/avifpng.c
wantehchang: Done.

wantehchang:
I addressed your review comments in the second commit. Please review. Thanks!

jzern (approved):

wantehchang:
Thank you both for the review. This may be changed again when I replace `hasAlpha` with a call to `png_get_channels()`, but I will go ahead and merge this.

PR #1431, opened 2023-06-11
avifenc: add -n --no-overwrite argument

FlyingWombat:
useful for batch processing

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -71,6 +71,7 @@ static void syntax(void)
     printf("Options:\n");
     printf("    -h,--help                         : Show syntax help\n");
     printf("    -V,--version                      : Show the version number\n");
+    printf("    --no-overwrite                    : Never overwrite existing output file\n");
     printf("    -j,--jobs J                       : Number of jobs (worker threads, default: 1. Use \"all\" to use all available cores)\n");
     printf("    -o,--output FILENAME              : Instead of using the last filename given as output, use this filename\n");
     printf("    -l,--lossless                     : Set all defaults to encode losslessly, and emit warnings when settings/input don't allow for it\n");
@@ -288,6 +289,16 @@ static avifBool avifInputAddCachedImage(avifInput * input)
     ++input->cacheCount;
     free(oldCachedImages);
     return AVIF_TRUE;
+}
+
+static avifBool fileExists(const char * filename)
+{
+    FILE * outfile = fopen(filename, "rb");
+    if (outfile) {
+        fclose(outfile);
+        return AVIF_TRUE;
+    }
+    return AVIF_FALSE;
 }
 
 static const avifInputFile * avifInputGetFile(const avifInput * input, int imageIndex)
@@ -1048,6 +1059,7 @@ int main(int argc, char * argv[])
     //     https://github.com/AOMediaCodec/libavif/issues/440
 
     int returnCode = 0;
+    avifBool noOverwrite = AVIF_FALSE;
     avifSettings settings;
     memset(&settings, 0, sizeof(settings));
     settings.codecChoice = AVIF_CODEC_CHOICE_AUTO;
@@ -1117,6 +1129,8 @@ int main(int argc, char * argv[])
         } else if (!strcmp(arg, "-V") || !strcmp(arg, "--version")) {
             avifPrintVersions();
             goto cleanup;
+        } else if (!strcmp(arg, "--no-overwrite")) {
+            noOverwrite = AVIF_TRUE;
         } else if (!strcmp(arg, "-j") || !strcmp(arg, "--jobs")) {
             NEXTARG();
             if (!strcmp(arg, "all")) {
@@ -1561,6 +1575,11 @@ int main(int argc, char * argv[])
         goto cleanup;
     }
 
+    if (noOverwrite && fileExists(outputFilename)) {
+        fprintf(stderr, "ERROR: output file %s already exists and --no-overwrite was specified\n", outputFilename);
+        goto cleanup;
+    }
+
 #if defined(_WIN32)
     if (input.useStdin) {
         setmode(fileno(stdin), O_BINARY);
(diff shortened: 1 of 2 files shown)


apps/avifenc.c
```diff
@@ -1561,6 +1564,15 @@ int main(int argc, char * argv[])
         goto cleanup;
     }
 
+    if (noOverwrite){
```
maryla-uc, line 1567: Missing space before {
(or you can just run "clang-format -style=file -i apps/avifenc.c")

apps/avifenc.c
```diff
@@ -1117,6 +1118,8 @@ int main(int argc, char * argv[])
         } else if (!strcmp(arg, "-V") || !strcmp(arg, "--version")) {
             avifPrintVersions();
             goto cleanup;
+        } else if (!strcmp(arg, "-n") || !strcmp(arg, "--no-overwrite")) {
```
maryla-uc, line 1121: Please add a description of the flag in the syntax() function. Also add a line in CHANGELOG.md

maryla-uc, line 1121: I would rather not have the short "-n" version as I think its meaning is far from obvious, and it's not a flag that will be used very often, the long version is probably enough?

apps/avifenc.c
```diff
@@ -1561,6 +1564,15 @@ int main(int argc, char * argv[])
         goto cleanup;
     }
 
+    if (noOverwrite){
+        FILE * outfile = fopen(outputFilename, "rb");
+        if (outfile) {
+            fclose(outfile);
+            printf("No Overwrite: %s\n", outputFilename);
```
maryla-uc, line 1571: Write the error message to stderr and use a more descriptive message, e.g.:
fprintf(stderr, "ERROR: output file %s already exists and --no-overwrite was specified\n", outputFilename));

Also I think this check should also be done just before writing the output file. Encodes can take a long time and the file might have been created in between. Maybe put the check in a helper function and call it here (before encoding, as early exit), and again before writing to be safe.

y-guyon: Note: `fopen()` could be used at writing time with [file access mode flag `"x"`](https://en.cppreference.com/w/c/io/fopen#File_access_flags) instead of doing a separate `fopen("r")` beforehand but unfortunately it is a C11 feature and [libavif is only C99](https://github.com/AOMediaCodec/libavif/blob/ddffb7758872f35ef6703791d353e37decb3b7d9/CMakeLists.txt#L17). Also as you mentioned encoding can be slow so an early check is necessary anyway.

apps/avifenc.c
FlyingWombat: If you want to drop the short "-n", that's not that big a deal to me. But I make use of this option more often than not (I had made a wrapper script for avifenc to do this, before making this PR). In my opinion --no-overwrite should be the default behavior, but I figured a change in default behavior would be less likely to be accepted, so I went with this.

FYI, for ffmpeg, "-n" has the same meaning as here.

maryla-uc: FYI I checked a few other tools: ffmpeg, imagemagick, cwebp, and cjxl. ffmpeg is the only one that does not overwrite by default. The other ones just overwrite and don't provide any flags. I do think it's a useful flag to have though, but I'm wondering if we might have a better use for -n in the future. I don't have a very strong opinion though.

FlyingWombat: OK , I agree. Let's drop "-n". What tipped me over to the other side is the fact that there is already a "-y" flag short for "--yuv". So, if "-n" means "no", then people might assume "-y" means "yes".

What about shorting the flag to "--no"? I don't think there will come another feature where we will want to use just "--no" to represent.

I don't have a strong opinion on this either -- the only benefit is a little less typing :)

apps/avifenc.c
```diff
@@ -290,6 +291,17 @@ static avifBool avifInputAddCachedImage(avifInput * input)
     return AVIF_TRUE;
 }
 
+static avifBool file_exists(const char * filename)
+{
+    FILE * outfile = fopen(filename, "rb");
+    if (outfile) {
+        fclose(outfile);
+        return AVIF_TRUE;
+    } else {
```
wantehchang, line 300: Nit: Omit the "else" after a return statement.

apps/avifenc.c
```diff
@@ -1561,6 +1576,11 @@ int main(int argc, char * argv[])
         goto cleanup;
     }
 
+    if (noOverwrite && file_exists(outputFilename)) {
+        fprintf(stderr, "ERROR: output file %s already exists and --no-overwrite was specified\n", outputFilename);
+        goto cleanup;
+    }
```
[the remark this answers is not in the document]
wantehchang: Suppose the specified output file pathname is in a nonexistent or nonwritable directory. If the encoding takes a long time, we will also waste time encoding before realizing that the output file cannot be opened. This is another reason for opening the output file before encoding starts.

My personal preference is to go for simpler error handling unless errors are common. I haven't inspected avifenc.c carefully to know whether error handling is simpler if we open the output file before or after encoding starts.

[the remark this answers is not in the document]
maryla-uc: Both approaches have pros and cons... But even if errors should be rare, I would find deleting a user's file in case of error a bigger problem than wasting time in case we don't have write access. (This is the scenario where out.avif already exists, I run 'avifenc' outputting to out.avif, avifenc fails midway, and out.avif is either empty, or we would have deleted it in the cleanup).
Still I don't have a super strong opinion so I'm fine to go with whichever version you prefer or find easier.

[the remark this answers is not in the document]
FlyingWombat: A common case where encoding exits part-way through is simply when the encoding job is cancelled or killed. Creating the output file before encoding, in this case, will litter the filesystem with empty files -- which is annoying to clean up.

maryla-uc (approved):

PR #1446, opened 2023-06-21
Link with ntdll.lib if rav1e is used

wantehchang:
Fix the following linker errors in the windows-latest os image:

rav1e.lib(rav1e-6457e1f48599bb43.std-391022a4250a8b9a.std.feb3b897-cgu.0.rcgu.o.rcgu.o) : error LNK2019: unresolved external symbol __imp_NtWriteFile referenced in function _ZN3std3sys7windows5stdio5write17h3c0a0f8c7c3063c9E.llvm.15018492211956286748
rav1e.lib(rav1e-6457e1f48599bb43.std-391022a4250a8b9a.std.feb3b897-cgu.0.rcgu.o.rcgu.o) : error LNK2019: unresolved external symbol __imp_RtlNtStatusToDosError referenced in function _ZN3std3sys7windows5stdio5write17h3c0a0f8c7c3063c9E.llvm.15018492211956286748

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -397,7 +397,7 @@ if(AVIF_CODEC_RAV1E)
 
     # Unfortunately, rav1e requires a few more libraries
     if(WIN32)
-        set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ws2_32.lib bcrypt.lib userenv.lib)
+        set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ws2_32.lib bcrypt.lib userenv.lib ntdll.lib)
     elseif(UNIX AND NOT APPLE)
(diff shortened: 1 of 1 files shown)


jzern (approved):

wantehchang:
When I searched for these linker errors on the Web last night, I found only one report: https://yutani.rbind.io/post/rust-1.70-and-build-failure-on-windows/

That webpage says "Rust 1.70 started to use several Windows APIs defined in `NTDLL.DLL`". I checked the logs of the "build-static (windows-latest)" CI check and it is indeed using rustc 1.70.0.

The reason we didn't see this windows-latest build failure in https://github.com/AOMediaCodec/libavif/pull/1431 is that it used the cache for external dependencies with the key Windows-1d805a71287b4063be392ebc0d58941ee93ed7ddaa0cccc01bebc2e523310e8b. For some reason the cache with that key disappeared yesterday, so we needed to build the external dependencies and ran into these linker errors.

PR #1454, opened 2023-06-28
avifenc.c: Split options into three sections

wantehchang:
Split the command-line options of avifenc into three sections:
- Standard options: --version and --help
- Basic options: --qcolor, --qalpha, and --speed
- Advanced options: everything else

Also move --no-overwrite to be next to --output because both options are related to the output file.

Bug: b:275100954

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -78,7 +78,7 @@ static void syntaxLong(void)
     printf("Syntax: avifenc [options] input.[jpg|jpeg|png|y4m] output.avif\n");
     printf("Standard options:\n");
     printf("    -h,--help                         : Show short syntax help\n");
-    printf("    --longhelp                        : Show long syntax help\n");
+    printf("    --longhelp                        : Show long syntax help (this page)\n");
     printf("    -V,--version                      : Show the version number\n");
     printf("\n");
     printf("Basic options:\n");
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -76,12 +76,27 @@ static void syntaxShort(void)
 static void syntaxLong(void)
 {
     printf("Syntax: avifenc [options] input.[jpg|jpeg|png|y4m] output.avif\n");
-    printf("Options:\n");
+    printf("Standard options:\n");
     printf("    -h,--help                         : Show short syntax help\n");
     printf("    --longhelp                        : Show long syntax help\n");
     printf("    -V,--version                      : Show the version number\n");
-    printf("    --no-overwrite                    : Never overwrite existing output file\n");
+    printf("\n");
+    printf("Basic options:\n");
+    printf("    -q,--qcolor Q                     : Set quality for color (%d-%d, where %d is lossless)\n",
+           AVIF_QUALITY_WORST,
+           AVIF_QUALITY_BEST,
+           AVIF_QUALITY_LOSSLESS);
+    printf("    --qalpha Q                        : Set quality for alpha (%d-%d, where %d is lossless)\n",
+           AVIF_QUALITY_WORST,
+           AVIF_QUALITY_BEST,
+           AVIF_QUALITY_LOSSLESS);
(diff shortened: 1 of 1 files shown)


apps/avifenc.c
```diff
@@ -76,12 +76,27 @@ static void syntaxShort(void)
 static void syntaxLong(void)
 {
     printf("Syntax: avifenc [options] input.[jpg|jpeg|png|y4m] output.avif\n");
-    printf("Options:\n");
+    printf("Standard options:\n");
```
wantehchang, line 79: The output of `avifenc --longhelp` looks like the following:

```
Syntax: avifenc [options] input.[jpg|jpeg|png|y4m] output.avif
Standard options:
    -h,--help                         : Show short syntax help
    --longhelp                        : Show long syntax help
    -V,--version                      : Show the version number

Basic options:
    -q,--qcolor Q                     : Set quality for color (0-100, where 100 is lossless)
    --qalpha Q                        : Set quality for alpha (0-100, where 100 is lossless)
    -s,--speed S                      : Encoder speed (0-10, slowest-fastest, 'default' or 'd' for codec internal defaults. default speed: 6)

Advanced options:
    -j,--jobs J                       : Number of jobs (worker threads, default: 1. Use "all" to use all available cores)
    ...

    1. <key>=<value> applies to both the color (YUV) planes and the alpha plane (if present).
    ...

Version: 0.11.1 (dav1d [dec]:1.2.0-0-g676a864, libgav1 [dec]:0.18.0, aom [enc/dec]:3.6.1, rav1e [enc]:0.6.6 (v0.6.6), svt [enc]:v1.6.0)
libyuv : available (1864)

```

jzern (approved):
Nice.

apps/avifenc.c
```diff
@@ -76,12 +76,27 @@ static void syntaxShort(void)
...
     printf("    -h,--help                         : Show short syntax help\n");
     printf("    --longhelp                        : Show long syntax help\n");
```
y-guyon, line 81: Optional: `(this page)` could be mentioned here for extra clarity

apps/avifenc.c
```diff
@@ -76,12 +76,27 @@ static void syntaxShort(void)
...
     printf("    -V,--version                      : Show the version number\n");
-    printf("    --no-overwrite                    : Never overwrite existing output file\n");
+    printf("\n");
+    printf("Basic options:\n");
+    printf("    -q,--qcolor Q                     : Set quality for color (%d-%d, where %d is lossless)\n",
```
y-guyon, line 85: Nit: I would repeat the argument to be clear:
```
-q Q,--qcolor Q
```

apps/avifenc.c
y-guyon: Optional: Putting short flags last makes it easier to read for me:
```
--help,-h
--longhelp
--version,-V
```
It has to do with alignment I guess? But the opposite might be more widespread. Increasing the gap between the flags could help as well: `-h / --help` (as in [cwebp](https://github.com/webmproject/libwebp/blob/25d94f473b10882b8bee9288d00539001b692042/examples/cwebp.c#L536)).

y-guyon (approved):

apps/avifenc.c
wantehchang: Done

wantehchang:
Yannis: Thank you for your suggestions. To keep this pull request focused, I only made the simple change and will handle the other two suggestions separately.

apps/avifenc.c
wantehchang: I looked at the man pages of a few commands such as `cp`, 'mv', `rm`, and `tar` on Mac and Linux. They list the short options before the long options. The short and long options are separated by either a comma, e.g.,

```
       -v, --verbose
              explain what is being done
```

or by a newline, e.g.,

```
       -W
       --no-warn
           Suppress warning messages.
```

PR #1448, opened 2023-06-22
Add support for the PNG gAMA, cHRM and sRGB chunks

tongyuantongyu:
Fix #1139, supersede #1389.

This implementation tries to match gAMA and cHRM values with CICP values and set TC and CP if a successful match is found, otherwise generates an ICC profile based on gAMA and cHRM values.

This implementation also handles sRGB chunk by setting sRGB CP and TC.

The gamma=2.2 and BT709 primaries case, which is usually used to tag images as sRGB, is not specially treated as sRGB, to align with how browsers handle such PNG images. This behavior is negotiable.

Introduced `--ignore-profile` flag to ignore any color profile info including ICC profile and PNG gAMA, cHRM, sRGB chunks. `--ignore-icc` is considered deprecated and treated as an alias of `--ignore-profile`.

PR #1456, opened 2023-06-30
avifpng.c: two small changes to commit bd7e0cf

wantehchang:
No need to include <math.h>. Was needed for fabs() in earleir versions of https://github.com/AOMediaCodec/libavif/pull/1448.

Move the png_read_update_info() call to its original location so as to not separate the code that sets up imgBitDepth.
---

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -10,7 +10,6 @@
 
 #include <ctype.h>
 #include <limits.h>
-#include <math.h>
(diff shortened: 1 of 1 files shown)


wantehchang:
@tongyuantongyu Yuan: Just FYI. I tried to add these changes to your pull request https://github.com/AOMediaCodec/libavif/pull/1448 but failed.

jzern (approved):

PR #1457, opened 2023-06-30
are_images_equal.cc: Run clang-format

wantehchang:


wantehchang:
@tongyuantongyu Yuan: Just FYI.

This change was made in Yuan's pull request https://github.com/AOMediaCodec/libavif/pull/1448. It would be good to look into why the CI clang-format-check did not catch this.

jzern (approved):

PR #1463, opened 2023-07-05
Fix the first two values of --cicp being ignored for some png files.

maryla-uc:
The color primaries and transfer curve provided in the --cicp flag could be overwritten by the png reading function.


apps/shared/avifpng.c
```diff
@@ -361,69 +362,71 @@ avifBool avifPNGRead(const char * inputFilename,
...
+                    avifBool generateICCResult = AVIF_FALSE;
+                    if (avif->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) {
+                        generateICCResult = avifGenerateGrayICC(&avif->icc, (float)gamma, &primaries[6]);
+                    } else {
+                        generateICCResult = avifGenerateRGBICC(&avif->icc, (float)gamma, primaries);
+                    }
+
+                    if (!generateICCResult) {
+                        fprintf(stderr,
+                                "WARNING: libavif could not generate an ICC profile for file %s. It may be caused by invalid values in the color space information. The encoded AVIF image's colors may be affected.\n",
```
jzern, line 427: strange that clang-format didn't warn or break this line

apps/shared/avifpng.c
```diff
@@ -361,69 +362,71 @@ avifBool avifPNGRead(const char * inputFilename,
...
                 }
 
-                if (!generateICCResult) {
+                if (needToGenerateICC) {
+                    avif->colorPrimaries = AVIF_COLOR_PRIMARIES_UNSPECIFIED;
+                    avif->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
                     fprintf(stderr,
-                            "WARNING: libavif could not generate an ICC profile for file %s. It may be caused by invalid values in the color space information. The encoded AVIF image's colors may be affected.\n",
+                            "INFO: legacy PNG color space information found in file %s not matching any CICP value. libavif is generating an ICC profile for it."
+                            "Use --ignore-profile to ignore color space information instead (may affect the colors of the encoded AVIF image).\n",
```
jzern, line 415: missing newline or space before this sentence.

jzern (commented):
Is there enough test coverage for this?

maryla-uc:
Added some tests!

tests/test_cmd_icc_profile.sh
```diff
@@ -90,6 +90,12 @@ pushd ${TMP_DIR}
   "${AVIFDEC}" "${ENCODED_FILE}" "${DECODED_FILE}"
   "${IMAGEMAGICK}" "${DECODED_FILE}" -profile "${SRGB_ICC}" "${CORRECTED_FILE}"
   "${ARE_IMAGES_EQUAL}" "${REFERENCE_GRAY_PNG}" "${CORRECTED_FILE}" 0 45
+
+  # Check --cicp flag works
+  "${AVIFENC}" --cicp 9/12/8 -s 8 "${INPUT_COLOR_PNG}" -o "${ENCODED_FILE}"
+  "${AVIFDEC}" "${ENCODED_FILE}" "${DECODED_FILE}" | grep "Color Primaries" | grep 9
```
jzern, line 96: I'm not familiar with the format, so maybe 2 greps are more stable if the structure changes. Otherwise .* could work, but either is fine.

jzern (approved):
github does a poor job with the whitespace changes in this diff

tests/test_cmd_icc_profile.sh
maryla-uc: Yeah I wanted to make it flexible but I realized that my version would also match e.g. "18" if grepped for "8" so I made the regex a bit more precise.

jzern (approved):

PR #1465, opened 2023-07-07
Fix CICP spec link.

maryla-uc:
Other options:
https://www.itu.int/rec/T-REC-H.273-202107-I/en (current version)
https://www.itu.int/rec/T-REC-H.273/en (no version)

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -267,7 +267,7 @@ typedef enum avifRange
 } avifRange;
 
 // ---------------------------------------------------------------------------
-// CICP enums - https://www.itu.int/rec/T-REC-H.273-201612-I/en
+// CICP enums - https://www.itu.int/rec/T-REC-H.273-201612-S/en
 
 enum
 {


y-guyon (approved):

include/avif/avif.h
```diff
@@ -267,7 +267,7 @@ typedef enum avifRange
 } avifRange;
 
 // ---------------------------------------------------------------------------
-// CICP enums - https://www.itu.int/rec/T-REC-H.273-201612-I/en
+// CICP enums - https://www.itu.int/rec/T-REC-H.273-201612-S/en
```
wantehchang, line 270: Maryla: When I visit the new URL, the page says "Status : Superseded". Perhaps we should use https://www.itu.int/rec/T-REC-H.273 or https://www.itu.int/rec/T-REC-H.273/en instead?

maryla-uc: Yes in the PR above I said this was another option. I kept that specific version because I assume it's the version that the implementation was based on at the time it was written. ...Not that there should be any significant differences with later versions? I'm ok with changing to https://www.itu.int/rec/T-REC-H.273/en 

The third option I mentioned of linking to https://www.itu.int/rec/T-REC-H.273-202107-I/en is probably not a good idea because it might break again in the future. It seems that the "-I" at the end means "In force" and it changes to "-S" when the spec becomes Superseded. At least that's my theory.

PR #1466, opened 2023-07-12
test_cmd_icc_profile.sh: Move --cicp tests earlier

wantehchang:
The --cicp tests do not use ImageMagick, so run those tests before checking for the presence of ImageMagick.

Also run popd before exiting with 0 if ImageMagick is not present.

A follow-up to the pull request
https://github.com/AOMediaCodec/libavif/pull/1463.


maryla-uc (approved):
Thanks!

PR #1467, opened 2023-07-13
Detect a libaom v3.6.0 bug with large images

wantehchang:
This bug, reported in https://crbug.com/aomedia/2871#c12, causes libaom v3.6.0 to encode bitstreams for certain large frame sizes that cannot be decoded by existing AV1 decoders.

Change the EncodingTest.ReasonableValidDimensions test to use dimensions that avoid this bug.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -703,6 +703,11 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         // https://aomedia-review.googlesource.com/c/aom/+/174421.
         static const int aomVersion_3_6_0 = (3 << 16) | (6 << 8);
         if (aomVersion == aomVersion_3_6_0) {
+            // Detect the use of levels 7.x and 8.x, which use a larger max
+            // tile area (4096 * 4608) than MAX_TILE_AREA (4096 * 2304). The
+            // larger max tile area may not result in a different bitstream
+            // (see the tile_info() function in the AV1 spec, Section 5.9.15),
+            // so this is just a necessary condition for the bug.
             if (!doesLevelMatch(image->width, image->height, 8192, 4352, 2) &&
                 (doesLevelMatch(image->width, image->height, 16384, 8704, 2) ||
                  doesLevelMatch(image->width, image->height, 32768, 17408, 2))) {
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -535,6 +535,13 @@ static avifBool avifFindAOMScalingMode(const avifFraction * avifMode, AOM_SCALIN
     return AVIF_FALSE;
 }
 
+static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelHeight, int levelDimMult)
+{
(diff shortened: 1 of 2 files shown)


src/codec_aom.c
```diff
@@ -535,6 +535,13 @@ static avifBool avifFindAOMScalingMode(const avifFraction * avifMode, AOM_SCALIN
     return AVIF_FALSE;
 }
 
+static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelHeight, int levelDimMult)
+{
+    const int64_t levelLumaPels = (int64_t)levelWidth * levelHeight;
+    const int64_t lumaPels = (int64_t)width * height;
+    return lumaPels <= levelLumaPels && width <= levelWidth * levelDimMult && height <= levelHeight * levelDimMult;
+}
```
wantehchang, line 543: This function was originally added to libaom by Debargha, a Google employee, in https://aomedia-review.googlesource.com/c/aom/+/58364. Google owns the copyright and can recontribute the code to libavif.

wantehchang (commented):
This issue recently came up again in https://github.com/strukturag/libheif/issues/790.

Although I wrote this pull request, I am not sure if we should do this. Perhaps we should instead encourage people to upgrade to libaom v3.6.1 or later.

src/codec_aom.c
```diff
@@ -689,6 +696,21 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         cfg->g_input_bit_depth = image->depth;
         cfg->g_w = image->width;
         cfg->g_h = image->height;
+
+        // Detect the libaom v3.6.0 bug described in
+        // https://crbug.com/aomedia/2871#c12. See the changes to
+        // av1/encoder/encoder.c in
+        // https://aomedia-review.googlesource.com/c/aom/+/174421.
+        static const int aomVersion_3_6_0 = (3 << 16) | (6 << 8);
```
jzern, line 704: This is OK. For your other option, were you thinking of failing at configure time or failing at runtime for all uses of 3.6.0?

jzern (approved):

src/codec_aom.c
wantehchang: I was thinking of sending email to the relevant mailing lists about this serious bug in libaom 3.6.0. I wasn't thinking of failing at configure time or failing at runtime for all uses of 3.6.0.

wantehchang:
James: Please review the second commit in the pull request. Thanks.

src/codec_aom.c
```diff
@@ -689,6 +696,26 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+        if (aomVersion == aomVersion_3_6_0) {
+            // Detect the use of levels 7.x and 8.x, which use a larger max
```
jzern, line 706: I'm not sure if it's worth mentioning the aomedia bug or CWG_C013 here.

jzern (approved):

PR #1471, opened 2023-07-13
libsharpyuv.cmd: update checkout to v1.3.1

jzern:
libsharpyuv was released as part of 1.3.0. 1.3.1 is the latest release.


jzern:
Actually, I forgot Vincent moved this past 1.3.1 to pick up the new transfer functions.

PR #1472, opened 2023-07-14
libsharpyuv.cmd: fix commit hash

jzern:
Use the hash from the final commit, not one during review; fixes
checkout failure

### DIFF
--- a/ext/libsharpyuv.cmd
+++ b/ext/libsharpyuv.cmd
@@ -12,7 +12,7 @@
 git clone --single-branch https://chromium.googlesource.com/webm/libwebp
 
 cd libwebp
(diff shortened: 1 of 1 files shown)


ext/libsharpyuv.cmd
```diff
@@ -12,7 +12,7 @@
 git clone --single-branch https://chromium.googlesource.com/webm/libwebp
 
 cd libwebp
-git checkout 048cfc
```
wantehchang, line 15: I saw the error message "error: pathspec '048cfc' did not match any file(s) known to git" when I ran this script.

wantehchang (approved):

Issue #1475, opened 2023-07-25
AVIF_CHROMA_UPSAMPLING_AUTOMATIC doesn't use kFilterBilinear with libyuv functions after commit 95d96ea6c8428e6c796b94f880b6e7a5979cb3a3


wantehchang:
A bug was introduced in commit 95d96ea6c8428e6c796b94f880b6e7a5979cb3a3 (https://github.com/AOMediaCodec/libavif/pull/1264). Here is the relevant code before and after the commit.

Before:

```
            const enum FilterMode filter = ((rgb->chromaUpsampling == AVIF_CHROMA_UPSAMPLING_FASTEST) ||
                                            (rgb->chromaUpsampling == AVIF_CHROMA_UPSAMPLING_NEAREST))
                                               ? kFilterNone
                                               : kFilterBilinear;
```

After:

```
// Only allow nearest-neighbor filter if explicitly specified or left as default.
static avifBool nearestNeighborFilterAllowed(int chromaUpsampling)
{
    return chromaUpsampling != AVIF_CHROMA_UPSAMPLING_BILINEAR && chromaUpsampling != AVIF_CHROMA_UPSAMPLING_BEST_QUALITY;
}

...

const enum FilterMode filter = nearestNeighborFilterAllowed(rgb->chromaUpsampling) ? kFilterNone : kFilterBilinear;
```

Suppose `rgb->chromaUpsampling` is `AVIF_CHROMA_UPSAMPLING_AUTOMATIC`. Before the commit, `filter` is set equal to `kFilterBilinear`. After the commit, `filter` is set equal to `kFilterNone`.

The effect of the bug is that `AVIF_CHROMA_UPSAMPLING_AUTOMATIC` doesn't use `kFilterBilinear` with libyuv functions.

I will write a fix.

PR #1476, opened 2023-07-25
Use kFilterBilinear with libyuv when allowed

wantehchang:
The condition for determining whether kFilterNone or kFilterBilinear should be used was changed incorrectly in commit 95d96ea (Refactor YUV->RGB conversion with libyuv), causing AVIF_CHROMA_UPSAMPLING_AUTOMATIC to not use kFilterBilinear with libyuv. Fix the condition by reverting to the original code.

Fix https://github.com/AOMediaCodec/libavif/issues/1475.


vigneshvg (approved):
thanks for fixing this!

PR #1477, opened 2023-07-26
Move the 4:0:0 grayscale comment from Added to the Changed.

vrabaud:


CHANGELOG.md
```diff
@@ -76,6 +75,7 @@ List of incompatible ABI changes in this release:
   memory allocation failures.
 * avifReadImage(), avifJPEGRead() and avifPNGRead() now remove the trailing zero
   byte from read XMP chunks, if any. See avifImageFixXMP().
+* Requesting lossless 4:0:0 on a grayscale input now outputs a grayscale AVIF.
```
wantehchang, line 78: I now think this can be considered a new feature. With previous releases, `avifenc -y 400 -l` outputs a lossless 4:4:4 image, so it isn't lossless 4:0:0. So this entry should be moved back to the Added section.

Of course, we can consider that as a bug and then this change will be a bug fix, suitable for the Changed section.

In any case, the new description is clearer than the old one.

wantehchang (approved):
LGTM.

PR #1482, opened 2023-07-27
Don't use Identity matrix coefficients with 4:0:0

wantehchang:
Chrome doesn't render 4:0:0 with Identity matrix coefficients correctly. Work around this Chrome issue by using matrix coefficients 6 instead with lossless 4:0:0.

Fix https://github.com/AOMediaCodec/libavif/issues/1481.


vrabaud:
Thx for the explanation in https://github.com/AOMediaCodec/libavif/pull/1460#discussion_r1275574905 !

vrabaud (approved):

wantehchang (commented):
Please review the two commits together rather than separately.

Vincent approved the first commit in this pull request before it was ready for review.

I tested this pull request with a grayscale PNG input and a monochrome y4m input, and added the second commit.

apps/avifenc.c
```diff
@@ -1676,9 +1675,18 @@ int main(int argc, char * argv[])
         goto cleanup;
     }
 
-    // Check again for y4m input (y4m input ignores input.requestedFormat and retains the format in file).
-    if ((image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) && (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV444) &&
-        (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400)) {
+    // Check again for -y auto or for y4m input (y4m input ignores input.requestedFormat and
+    // retains the format in file).
```
wantehchang, line 1679: The original comment may have been written before `-y auto` was supported. A y4m input is not the only reason `image->yuvFormat` may be different from `input.requestedFormat` now. So I updated this comment.

y-guyon (approved):

PR #1486, opened 2023-07-31
Android add image scale when libyuv is on.

WizzXu:


jzern:
The new test added fails:

```
org.aomedia.avif.android.AvifDecoderTest > testImageScale[0][test(AVD) - 11] FAILED 
```

PR #1488, opened 2023-08-02
Allow libyuv to use identity matrix with YUV 4:0:0

wantehchang:
Allow avifImageYUVToRGB() to convert YUV 4:0:0 with the identity matrix to RGB by using libyuv, similar to the Chromium CL https://chromium-review.googlesource.com/c/chromium/src/+/4727949.

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -737,8 +737,13 @@ static avifBool getLibYUVConversionFunction(avifPixelFormat yuvFormat,
 
 static void getLibYUVConstants(const avifImage * image, const struct YuvConstants ** matrixYUV, const struct YuvConstants ** matrixYVU)
 {
+    // Allow the identity matrix to be used with YUV 4:0:0. Replace the identity matrix with
+    // MatrixCoefficients 6 (BT.601).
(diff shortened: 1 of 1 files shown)


src/reformat_libyuv.c
```diff
@@ -737,8 +737,13 @@ static avifBool getLibYUVConversionFunction(avifPixelFormat yuvFormat,
 
 static void getLibYUVConstants(const avifImage * image, const struct YuvConstants ** matrixYUV, const struct YuvConstants ** matrixYVU)
 {
+    // Allow the identity matrix to be used with YUV 4:0:0. Replace the identity matrix with
+    // MatrixCoefficients 6 (BT.601).
+    const avifBool yuv400WithIdentityMatrix = (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) &&
+                                              (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY);
```
wantehchang, line 743: In the [Chromium CL](https://chromium-review.googlesource.com/c/chromium/src/+/4727949), the condition I used is just `image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400`. I am planning to change Chromium to match this condition.

vigneshvg (approved):

PR #1489, opened 2023-08-03
avifenc: add a flag to set clli

sboukortt:


apps/avifenc.c
```diff
@@ -133,6 +133,7 @@ static void syntaxLong(void)
     printf("    --clap WN,WD,HN,HD,HON,HOD,VON,VOD: Add clap property (clean aperture). Width, Height, HOffset, VOffset (in num/denom pairs)\n");
     printf("    --irot ANGLE                      : Add irot property (rotation). [0-3], makes (90 * ANGLE) degree rotation anti-clockwise\n");
     printf("    --imir MODE                       : Add imir property (mirroring). 0=top-to-bottom, 1=left-to-right\n");
+    printf("    --clli C(,A)                      : Add clli property (content light level information). C = MaxCLL, A = MaxPALL\n");
```
y-guyon, line 136: Suggestion:
```c
    printf(" --clli MAXCLL,MAXPALL : Add clli property (content light level information).\n");
```
This option is not that common so there is no need for an optional maxPALL setting. It is simpler to always request two arguments.

apps/avifenc.c
```diff
@@ -1444,6 +1447,14 @@ int main(int argc, char * argv[])
                 returnCode = 1;
                 goto cleanup;
             }
+        } else if (!strcmp(arg, "--clli")) {
+            NEXTARG();
+            settings.clliCount = parseU32List(settings.clliValues, arg);
+            if (settings.clliCount > 2) {
```
y-guyon, line 1453:
```c
        if ((settings.clliCount != 2) || (settings.clliValues[0] >= (1u << 16)) || (settings.clliValues[1] >= (1u << 16))) {
```
This way the cast below is safe.

apps/avifenc.c
```diff
@@ -1758,6 +1769,8 @@ int main(int argc, char * argv[])
         image->transformFlags |= AVIF_TRANSFORM_IMIR;
         image->imir.mode = imirMode;
     }
+    image->clli.maxCLL = settings.clliValues[0];
+    image->clli.maxPALL = settings.clliValues[1];
```
y-guyon, line 1773:
```c
    if (settings.clliCount == 2) {
        image->clli.maxCLL = (uint16_t)settings.clliValues[0];
        image->clli.maxPALL = (uint16_t)settings.clliValues[1];
    }
```

y-guyon (commented):
Thank you for your contribution. I suggest a few changes.

Adding a command line test in [test_cmd_metadata.sh](https://github.com/AOMediaCodec/libavif/blob/7a67c506bdd016b93ddabacdbbbbc04a8d41f313/tests/test_cmd_metadata.sh#L69) such as the following would be nice:
```diff
    "${AVIFENC}" "${INPUT}" -o "${ENCODED_FILE_NO_METADATA}" --ignore-xmp
    cmp "${ENCODED_FILE}" "${ENCODED_FILE_NO_METADATA}" && exit 1
+    "${AVIFENC}" "${INPUT}" -o "${ENCODED_FILE_NO_METADATA}" --clli 1,2
+    cmp "${ENCODED_FILE}" "${ENCODED_FILE_NO_METADATA}" && exit 1
  done
```

PR #1491, opened 2023-08-03
Add stream comments to internal.h

y-guyon:
[skip ci]

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -431,7 +431,7 @@ typedef size_t avifBoxMarker;
 
 typedef struct avifBoxHeader
 {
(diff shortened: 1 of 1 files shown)


include/avif/internal.h
```diff
@@ -431,14 +431,20 @@ typedef size_t avifBoxMarker;
 
 typedef struct avifBoxHeader
 {
+    // Size of the box in bytes, excluding the number of bytes read to know the size and the type.
```
wantehchang, line 434: Optional: "the number of bytes read to know the size and the type" is the size of the box header, right? It seems that "excluding the size of the box header" is clear enough.

wantehchang (approved):
LGTM. Thanks for documenting the struct members.

PR #1432, opened 2023-06-13
Experimental reduced HEIF header

y-guyon:
Remove a few hundred bytes of container boilerplate for simple images. This prototype uses a small box that expands in meaning to a full meta box with all defined items and properties.

Warning: This feature is experimental and forbidden by the current AVIF specification.

PR #1492, opened 2023-08-03
Guard avifDiagnosticsClearError() in avif.c

y-guyon:
[skip ci]

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -667,9 +667,7 @@ avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
                                              avifPixelFormat yuvFormat,
                                              avifDiagnostics * diag)
 {
-    if (diag) {
-        avifDiagnosticsClearError(diag);
-    }
+    avifDiagnosticsClearError(diag);
 
     // ISO/IEC 14496-12:2020, Section 12.1.4.1:
     //   For horizOff and vertOff, D shall be strictly positive and N may be
@@ -774,9 +772,7 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
                                              avifPixelFormat yuvFormat,
                                              avifDiagnostics * diag)
 {
-    if (diag) {
-        avifDiagnosticsClearError(diag);
-    }
+    avifDiagnosticsClearError(diag);
 
     if (!avifCropRectIsValid(cropRect, imageW, imageH, yuvFormat, diag)) {
         return AVIF_FALSE;
--- a/src/avif.c
+++ b/src/avif.c
@@ -667,7 +667,9 @@ avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
                                              avifPixelFormat yuvFormat,
                                              avifDiagnostics * diag)
 {
-    avifDiagnosticsClearError(diag);
+    if (diag) {
+        avifDiagnosticsClearError(diag);
+    }
 
     // ISO/IEC 14496-12:2020, Section 12.1.4.1:
     //   For horizOff and vertOff, D shall be strictly positive and N may be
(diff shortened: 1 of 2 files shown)


src/avif.c
```diff
@@ -667,7 +667,9 @@ avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
                                              avifPixelFormat yuvFormat,
                                              avifDiagnostics * diag)
 {
-    avifDiagnosticsClearError(diag);
+    if (diag) {
+        avifDiagnosticsClearError(diag);
+    }
```
wantehchang, line 672: Could you explain why this is needed?

If you need to call these two functions from places where you do not have a `avifDiagnostics` object, we have been working around this by declaring a local `avifDiagnostics` variable and passing it to the function.

wantehchang: Looking at this closer, I found that `avifDiagnosticsPrintf()` already ignores a null `diag` input. For API consistency it seems better to have `avifDiagnosticsClearError()` ignore a null `diag` input.

y-guyon: > Could you explain why this is needed?
> 
> If you need to call these two functions from places where you do not have a `avifDiagnostics` object, we have been working around this by declaring a local `avifDiagnostics` variable and passing it to the function.

Yes this is to avoid adding a line in the unrelated `avifconitest.cc` from #1432.

> Looking at this closer, I found that `avifDiagnosticsPrintf()` already ignores a null `diag` input. For API consistency it seems better to have `avifDiagnosticsClearError()` ignore a null `diag` input.

Done.

wantehchang: Thank you for the reply. I looked into the history of the relevant code. I found no evidence of passing a null `diag` to `avifDiagnosticsPrintf()`. Also, since we always call `avifDiagnosticsClearError()` first and it disallows a null `diag`, I think no code can possibly pass a null `diag` to `avifDiagnosticsPrintf()`. So it seems better to remove the null `diag` check from `avifDiagnosticsPrintf()`, and then just pass a local `diag` variable in `avifconitest.cc` (and use it in an error message):

```
    avifDiagnostics diag;
    ASSERT_TRUE(avifCleanApertureBoxConvertCropRect(
        &image->clap, &rect, image->width, image->height, image->yuvFormat,
        &diag)) << diag.error;
```

This way we can detect the bug of passing a null `diag` by mistake.

wantehchang (approved):
I have stated my preference. This pull request is one of two ways to make the two functions consistent about whether a null `diag` is allowed.

@joedrago The "calls to avifFileTypeIsCompatible()" part of the comment is incorrect. Joe, do you remember what you had in mind when you wrote it?

y-guyon:
> @joedrago The "calls to avifFileTypeIsCompatible()" part of the comment is incorrect. Joe, do you remember what you had in mind when you wrote it?

It was about `avifPeekCompatibleFileType()` and its `BEGIN_STREAM()` called with NULL diag argument:

https://github.com/AOMediaCodec/libavif/blob/3e6645618891a8536a623c2504d188397b2c7d46/src/read.c#L3196-L3198

src/avif.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1496

wantehchang: Thanks a lot for tracking down the code that passes a null `diag`. With your hints, I also found https://github.com/AOMediaCodec/libavif/commit/388889c2. Given the new information, I suggest we go ahead and merge this pull request.

joedrago:
My intent with `avifFileTypeIsCompatible()` was to be a very cheap way to peek into a buffer and see if it was likely to be an AVIF. If it internally creates an avifDiagnostics and throws away its value so that the entire codebase can assert that a `diag` pointer is never NULL, I think that is a good change.

PR #1493, opened 2023-08-03
Split and document avifParseCodecConfiguration()

y-guyon:
Mirror the changes of read.c to write.c too.
Also nits.

This is partly to simplify #1432.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1862,24 +1862,16 @@ static avifBool avifParseContentLightLevelInformationBox(avifProperty * prop, co
     return AVIF_TRUE;
 }
 
-static avifBool avifParseCodecConfigurationBoxProperty(avifProperty * prop,
-                                                       const uint8_t * raw,
-                                                       size_t rawLen,
-                                                       const char * configPropName,
-                                                       avifDiagnostics * diag)
+// Implementation of section 2.3.3. of AV1 Codec ISO Media File Format Binding specification v1.2.0.
+// See https://aomediacodec.github.io/av1-isobmff/v1.2.0.html#av1codecconfigurationbox-syntax.
+static avifBool avifParseCodecConfiguration(avifROStream * s, avifCodecConfigurationBox * config, const char * configPropName, avifDiagnostics * diag)
 {
-    char diagContext[] = "Box[....]";
-    snprintf(diagContext, sizeof(diagContext), "Box[%.4s]", configPropName); // "Box[av1C]" or "Box[av2C]"
-    BEGIN_STREAM(s, raw, rawLen, diag, diagContext);
-
-    avifCodecConfigurationBox * config = &prop->u.av1C;
-
     uint8_t markerAndVersion = 0;
-    AVIF_CHECK(avifROStreamRead(&s, &markerAndVersion, 1));
+    AVIF_CHECK(avifROStreamRead(s, &markerAndVersion, 1));
     uint8_t seqProfileAndIndex = 0;
-    AVIF_CHECK(avifROStreamRead(&s, &seqProfileAndIndex, 1));
+    AVIF_CHECK(avifROStreamRead(s, &seqProfileAndIndex, 1));
     uint8_t rawFlags = 0;
-    AVIF_CHECK(avifROStreamRead(&s, &rawFlags, 1));
+    AVIF_CHECK(avifROStreamRead(s, &rawFlags, 1));
 
     if (markerAndVersion != 0x81) {
         // Marker and version must both == 1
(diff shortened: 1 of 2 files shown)


jzern (approved):

wantehchang (approved):
LGTM.

I suggest some changes. Regarding the period after a section number, I wonder if this is a regional difference.

The advantages of some of the function splits are not obvious to me. Is this pull request setting up for the Slim HEIF code, so that the function splits will make more sense when the Slim HEIF code is added?

src/write.c
```diff
@@ -2199,5 +2215,15 @@ static void writeConfigBox(avifRWStream * s, avifCodecConfigurationBox * cfg, co
     // }
     avifRWStreamWriteU8(s, 0);
 
+    // According to section 2.2.1. of AV1 Image File Format specification v1.1.0,
+    // there is no need to write any OBU here.
+    // See https://aomediacodec.github.io/av1-avif/v1.1.0.html#av1-configuration-item-property.
+    // unsigned int (8) configOBUs[];
+}
+
+static void writeConfigBox(avifRWStream * s, const avifCodecConfigurationBox * cfg, const char * configPropName)
```
wantehchang, line 2224: The function split here is also uneven because almost all the code is in the writeCodecConfig() function.

src/read.c
```diff
@@ -1896,9 +1888,41 @@ static avifBool avifParseCodecConfigurationBoxProperty(avifProperty * prop,
...
+static avifBool avifParseCodecConfigurationBoxProperty(avifProperty * prop,
+                                                       const uint8_t * raw,
+                                                       size_t rawLen,
+                                                       const char * configPropName,
+                                                       avifDiagnostics * diag)
+{
+    char diagContext[] = "Box[....]";
+    snprintf(diagContext, sizeof(diagContext), "Box[%.4s]", configPropName); // "Box[av1C]" or "Box[av2C]"
+    BEGIN_STREAM(s, raw, rawLen, diag, diagContext);
+    return avifParseCodecConfiguration(&s, &prop->u.av1C, configPropName, diag);
```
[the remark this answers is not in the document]
y-guyon: As written above, because this is partly to simplify https://github.com/AOMediaCodec/libavif/pull/1432. Also it matches the [section 2.3.3](https://aomediacodec.github.io/av1-isobmff/v1.2.0.html#av1codecconfigurationbox-syntax) which is also split into two functions.

src/write.c
y-guyon: As written above, because this is partly to simplify https://github.com/AOMediaCodec/libavif/pull/1432. Also it matches the [section 2.3.3](https://aomediacodec.github.io/av1-isobmff/v1.2.0.html#av1codecconfigurationbox-syntax) which is also split into two functions.

y-guyon:
> Regarding the period after a section number, I wonder if this is a regional difference.

In the specification there is a trailing dot after section numbers so I repeated that here.

> The advantages of some of the function splits are not obvious to me. Is this pull request setting up for the Slim HEIF code, so that the function splits will make more sense when the Slim HEIF code is added?

Yes, this is partly to simplify https://github.com/AOMediaCodec/libavif/pull/1432.

src/read.c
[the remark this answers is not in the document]
wantehchang: The two functions here do not seem to match the two functions in [section 2.3.3](https://aomediacodec.github.io/av1-isobmff/v1.2.0.html#av1codecconfigurationbox-syntax).

The only thing the outer function (`avifParseCodecConfigurationBoxProperty`) does is to set up the `avifROStream s` variable, which is only used by the inner function (`avifParseCodecConfiguration`). So the outer function does essentially no parsing.

In contrast, the outer type in [section 2.3.3](https://aomediacodec.github.io/av1-isobmff/v1.2.0.html#av1codecconfigurationbox-syntax) is:

```
class AV1CodecConfigurationBox extends Box('av1C'){
  AV1CodecConfigurationRecord av1Config;
}
```

So to match [section 2.3.3](https://aomediacodec.github.io/av1-isobmff/v1.2.0.html#av1codecconfigurationbox-syntax), I would expect the outer function (`avifParseCodecConfigurationBoxProperty`) to parse the `Box('av1C')` header, but it doesn't do that.

In any case, this is fine if this helps simplify https://github.com/AOMediaCodec/libavif/pull/1432.

src/write.c
wantehchang: The two functions here also don't seem to match [section 2.3.3](https://aomediacodec.github.io/av1-isobmff/v1.2.0.html#av1codecconfigurationbox-syntax).

src/read.c
```diff
@@ -2386,7 +2410,7 @@ static avifBool avifParseItemInfoEntry(avifMeta * meta, const uint8_t * raw, siz
 
     avifDecoderItem * item = avifMetaFindItem(meta, itemID);
     if (!item) {
-        avifDiagnosticsPrintf(diag, "Box[infe] has an invalid item ID [%u]", itemID);
+        avifDiagnosticsPrintf(diag, "%s: Box[infe] with item_type %.4s has an invalid item ID [%u]", s.diagContext, itemType, itemID);
```
wantehchang, line 2413: Nit: Whether an underscore '_' is used or not used in "item_type" or "item ID" could be consistent.

src/write.c
```diff
@@ -536,6 +537,13 @@ static avifBool avifEncoderDetectChanges(const avifEncoder * encoder, avifEncode
     return AVIF_TRUE;
 }
 
+// Subset of avifEncoderWriteColorProperties() for the properties clli, pasp, clap, irot, imir.
+static void avifEncoderWriteExtendedColorProperties(avifRWStream * dedupStream,
```
wantehchang, line 541: Should the local variable `avifRWStream * s` in `avifEncoderWriteColorProperties()` also be renamed `dedupStream`?

src/read.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1502

src/write.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1503 but not so sure it is better.

wantehchang: This is for consistency. I was wondering why that `avifRWStream` variable is named `s` in one function but named `dedupStream` in the other.

I forgot how the code in these two functions works, so I am not sure if `dedupStream` is a good name. It may be better to add comments to describe `dedupStream` vs. `outputStream`.

PR #1494, opened 2023-08-03
Add limited range support to FillImageGradient()

y-guyon:
Also check clli, pasp, clap, irot and imir properties in AreImagesEqual().
Also move kSampleIcc, kSampleExif and kSampleXmp from avifmetadatatest.cc to aviftest_helpers.h.

This is partly to simplify #1432.


jzern (approved):
This is relatively small, but it could be broken up by each point in the commit message.

PR #1496, opened 2023-08-04
Forbid NULL diag in avifDiagnosticsPrintf()

y-guyon:


y-guyon:
Merged https://github.com/AOMediaCodec/libavif/pull/1492 instead.

wantehchang (commented):
Yannis: Thank you for writing this pull request. I asked you to write this pull request after I did an incomplete analysis of the code and mistakenly thought no code could possibly pass a null `diag` to `avifDiagnosticsPrintf()`. With the new information, I suggest we go with your https://github.com/AOMediaCodec/libavif/pull/1492.

wantehchang (commented):
Also, more changes would be necessary. For example, the comment and assertion in `avifROStreamStart()` would need to be deleted:

```
void avifROStreamStart(avifROStream * stream, avifROData * raw, avifDiagnostics * diag, const char * diagContext)
{
    stream->raw = raw;
    stream->offset = 0;
    stream->diag = diag;
    stream->diagContext = diagContext;

    // If diag is non-NULL, diagContext must also be non-NULL
    assert(!stream->diag || stream->diagContext);
}
```

PR #1498, opened 2023-08-04
Propagate avifResult from stream.c makeRoom()

y-guyon:

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -317,17 +317,14 @@ avifBool avifROStreamReadAndEnforceVersion(avifROStream * stream, uint8_t enforc
 // avifRWStream
 
 #define AVIF_STREAM_BUFFER_INCREMENT (1024 * 1024)
-static void makeRoom(avifRWStream * stream, size_t size)
+static avifResult makeRoom(avifRWStream * stream, size_t size)
 {
     size_t neededSize = stream->offset + size;
     size_t newSize = stream->raw->size;
     while (newSize < neededSize) {
         newSize += AVIF_STREAM_BUFFER_INCREMENT;
     }
-    if (avifRWDataRealloc(stream->raw, newSize) != AVIF_RESULT_OK) {
-        // TODO(issue #820): Return AVIF_RESULT_OUT_OF_MEMORY instead.
-        abort();
-    }
+    return avifRWDataRealloc(stream->raw, newSize);
 }
 
 void avifRWStreamStart(avifRWStream * stream, avifRWData * raw)
@@ -361,33 +358,32 @@ void avifRWStreamFinishWrite(avifRWStream * stream)
     }
 }
 
-void avifRWStreamWrite(avifRWStream * stream, const void * data, size_t size)
+avifResult avifRWStreamWrite(avifRWStream * stream, const void * data, size_t size)
 {
     assert(stream->numUsedBitsInPartialByte == 0); // Byte alignment is required.
-    if (!size) {
-        return;
+    if (size) {
+        AVIF_CHECKRES(makeRoom(stream, size));
+        memcpy(stream->raw->data + stream->offset, data, size);
+        stream->offset += size;
     }
-
-    makeRoom(stream, size);
-    memcpy(stream->raw->data + stream->offset, data, size);
-    stream->offset += size;
+    return AVIF_RESULT_OK;
 }
 
-void avifRWStreamWriteChars(avifRWStream * stream, const char * chars, size_t size)
+avifResult avifRWStreamWriteChars(avifRWStream * stream, const char * chars, size_t size)
 {
-    avifRWStreamWrite(stream, chars, size);
+    return avifRWStreamWrite(stream, chars, size);
 }
 
-avifBoxMarker avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, size_t contentSize, int version, uint32_t flags)
+avifResult avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, size_t contentSize, int version, uint32_t flags, avifBoxMarker * marker)
 {
     assert(stream->numUsedBitsInPartialByte == 0); // Byte alignment is required.
-    avifBoxMarker marker = stream->offset;
+    *marker = stream->offset;
     size_t headerSize = sizeof(uint32_t) + 4 /* size of type */;
     if (version != -1) {
         headerSize += 4;
     }
 
-    makeRoom(stream, headerSize);
+    AVIF_CHECKRES(makeRoom(stream, headerSize));
     memset(stream->raw->data + stream->offset, 0, headerSize);
     uint32_t noSize = avifHTONL((uint32_t)(headerSize + contentSize));
     memcpy(stream->raw->data + stream->offset, &noSize, sizeof(uint32_t));
@@ -400,12 +396,12 @@ avifBoxMarker avifRWStreamWriteFullBox(avifRWStream * stream, const char * type,
     }
     stream->offset += headerSize;
 
-    return marker;
+    return AVIF_RESULT_OK;
 }
 
-avifBoxMarker avifRWStreamWriteBox(avifRWStream * stream, const char * type, size_t contentSize)
+avifResult avifRWStreamWriteBox(avifRWStream * stream, const char * type, size_t contentSize, avifBoxMarker * marker)
 {
-    return avifRWStreamWriteFullBox(stream, type, contentSize, -1, 0);
+    return avifRWStreamWriteFullBox(stream, type, contentSize, -1, 0, marker);
 }
 
 void avifRWStreamFinishBox(avifRWStream * stream, avifBoxMarker marker)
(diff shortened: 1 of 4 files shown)


wantehchang:
Yannis: Thank you for writing another pull request for https://github.com/AOMediaCodec/libavif/issues/820.

I started to review this pull request this morning but got sidetracked. I will finish the review tomorrow.

But I have a quick question: I noticed that the return values of several `avifRWStreamWrite*()` calls are not checked. Is it because you want to keep this pull request at a reasonable size and deal with them in a follow-up? Or is it because at those places we know there is enough room in the stream's buffer so those calls cannot possibly fail?

y-guyon:
> I started to review this pull request this morning but got sidetracked. I will finish the review tomorrow.

Thanks, there is no rush.

> But I have a quick question: I noticed that the return values of several `avifRWStreamWrite*()` calls are not checked. Is it because you want to keep this pull request at a reasonable size and deal with them in a follow-up? Or is it because at those places we know there is enough room in the stream's buffer so those calls cannot possibly fail?

It mostly is because I forgot, but yes it is a lot of changes so I split that into multiple commits.

y-guyon:
I rebased this PR. I suggest reviewing commit per commit.

src/stream.c
```diff
@@ -276,31 +273,30 @@ void avifRWStreamFinishWrite(avifRWStream * stream)
...
 
-void avifRWStreamWriteChars(avifRWStream * stream, const char * chars, size_t size)
+avifResult avifRWStreamWriteChars(avifRWStream * stream, const char * chars, size_t size)
 {
-    avifRWStreamWrite(stream, chars, size);
+    return avifRWStreamWrite(stream, chars, size);
 }
 
-avifBoxMarker avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, size_t contentSize, int version, uint32_t flags)
+avifResult avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, size_t contentSize, int version, uint32_t flags, avifBoxMarker * marker)
```
wantehchang, line 291: Optional: This function sets `*marker` to `stream->offset`. Since the caller has access to `stream->offset` (either directly or by calling`avifRWStreamOffset(stream)`), this function doesn't need to provide the `marker` output parameter.

src/stream.c
```diff
@@ -276,31 +273,30 @@ void avifRWStreamFinishWrite(avifRWStream * stream)
...
 {
-    avifBoxMarker marker = stream->offset;
+    *marker = stream->offset;
```
wantehchang, line 293: [This issue is moot if we remove the `marker` output parameter.]

We can allow `marker` to be null if the caller doesn't need to know the stream's offset before the full box. This is the case when `contentSize` is not `AVIF_BOX_SIZE_TBD`.

I annotated the two places where this could be useful.

src/write.c
```diff
@@ -380,11 +374,11 @@ static void avifItemPropertyDedupStart(avifItemPropertyDedup * dedup)
 // assigned the next available property index, written to the output stream, and its offset/size in
 // the output stream is recorded in the dedup for future comparisons.
 //
-// This function always returns a valid 1-indexed property index for usage in a property association
-// (ipma) box later. If the most recent property was a duplicate of a previous property, the return
-// value will be the index of the original property, otherwise it will be the index of the newly
-// created property.
-static uint8_t avifItemPropertyDedupFinish(avifItemPropertyDedup * dedup, avifRWStream * outputStream)
+// Unless it returns an error, this function sets its propertyIndex argument to a valid 1-indexed
```
wantehchang, line 377: Nit/optional: Unless it returns an error => On success ?

src/write.c
```diff
@@ -380,11 +374,11 @@ static void avifItemPropertyDedupStart(avifItemPropertyDedup * dedup)
...
+// property index for usage in a property association (ipma) box later. If the most recent property
+// was a duplicate of a previous property, the return value will be the index of the original
```
wantehchang, line 379: the return value will be => *propertyIndex will be set to

src/write.c
```diff
@@ -393,7 +387,7 @@ static uint8_t avifItemPropertyDedupFinish(avifItemPropertyDedup * dedup, avifRW
         if ((property->size == newPropertySize) &&
             !memcmp(&outputStream->raw->data[property->offset], dedup->buffer.data, newPropertySize)) {
             // We've already written this exact property, reuse it
-            return property->index;
+            *propertyIndex = property->index;
```
wantehchang, line 390: IMPORTANT: We should return `AVIF_RESULT_OK` here.

This bug disables dedupication. This is why our tests cannot detect this bug.

src/write.c
```diff
@@ -402,8 +396,23 @@ static uint8_t avifItemPropertyDedupFinish(avifItemPropertyDedup * dedup, avifRW
     property->index = ++dedup->nextIndex; // preincrement so the first new index is 1 (as ipma is 1-indexed)
     property->size = newPropertySize;
     property->offset = avifRWStreamOffset(outputStream);
-    avifRWStreamWrite(outputStream, dedup->buffer.data, newPropertySize);
-    return property->index;
+    *propertyIndex = property->index;
+    AVIF_CHECKRES(avifRWStreamWrite(outputStream, dedup->buffer.data, newPropertySize));
```
wantehchang, line 400: Nit: reverse these two lines (399-400).

src/write.c
```diff
@@ -402,8 +396,23 @@ static uint8_t avifItemPropertyDedupFinish(avifItemPropertyDedup * dedup, avifRW
...
+    return AVIF_RESULT_OK;
+}
+
+static avifResult avifItemPropertyDedupFinishAndIpmaPush(avifItemPropertyDedup * dedup,
```
wantehchang, line 404: Nit/optional: This can be the new `avifItemPropertyDedupFinish()` function. We don't need two functions.

src/write.c
```diff
@@ -402,8 +396,23 @@ static uint8_t avifItemPropertyDedupFinish(avifItemPropertyDedup * dedup, avifRW
...
+                                                         avifRWStream * outputStream,
+                                                         struct ipmaArray * ipma,
+                                                         avifBool essential)
+{
+    uint8_t propertyIndex;
+    AVIF_CHECKRES(avifItemPropertyDedupFinish(dedup, outputStream, &propertyIndex));
+    AVIF_CHECKERR(ipma->count < MAX_ASSOCIATIONS, AVIF_RESULT_UNKNOWN_ERROR);
```
wantehchang, line 411: [Just a comment, no change requested.] When I reviewed the original code, I verified this property by counting the number of `ipmaPush()` calls. It's certainly better to let the code check this property.

src/write.c
```diff
@@ -590,38 +600,41 @@ static void avifEncoderWriteColorProperties(avifRWStream * outputStream,
...
+    // unsigned int(7) reserved = 0;
     avifRWStreamFinishBox(dedupStream, colr);
     if (dedup) {
-        ipmaPush(ipma, avifItemPropertyDedupFinish(dedup, outputStream), AVIF_FALSE);
+        AVIF_CHECKRES(avifItemPropertyDedupFinishAndIpmaPush(dedup, outputStream, ipma, AVIF_FALSE));
     }
 
-    avifEncoderWriteExtendedColorProperties(dedupStream, outputStream, imageMetadata, ipma, dedup);
+    AVIF_CHECKRES(avifEncoderWriteExtendedColorProperties(dedupStream, outputStream, imageMetadata, ipma, dedup));
+    return AVIF_RESULT_OK;
```
wantehchang, line 617: Nit: Replace these two lines (616-617) with `return avifEncoderWriteExtendedColorProperties(...);`

This is okay if you will be adding code between these two lines.

src/write.c
```diff
@@ -1358,6 +1383,154 @@ static size_t avifEncoderFindExistingChunk(avifRWStream * s, size_t mdatStartOff
     return 0;
 }
 
+static avifResult avifEncoderWriteMdat(avifEncoder * encoder,
+                                       avifRWStream * stream,
```
wantehchang, line 1387: Nit/optional: name this parameter `s`, which is most common in this file.

src/write.c
```diff
@@ -1440,74 +1613,79 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
...
+    AVIF_CHECKRES(avifRWStreamWriteChars(&s, "libavif", 8)); // string name; (writing null terminator)
     avifRWStreamFinishBox(&s, hdlr);
 
     // -----------------------------------------------------------------------
     // Write pitm
 
     if (encoder->data->primaryItemID != 0) {
-        avifRWStreamWriteFullBox(&s, "pitm", sizeof(uint16_t), 0, 0);
-        avifRWStreamWriteU16(&s, encoder->data->primaryItemID); //  unsigned int(16) item_ID;
+        avifBoxMarker pitm;
```
wantehchang, line 1658: We can pass `NULL` to `avifRWStreamWriteFullBox()` and avoid the `pitm` variable.

src/write.c
```diff
@@ -1836,76 +2027,90 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
...
-            avifBoxMarker dinf = avifRWStreamWriteBox(&s, "dinf", AVIF_BOX_SIZE_TBD);
-            avifBoxMarker dref = avifRWStreamWriteFullBox(&s, "dref", AVIF_BOX_SIZE_TBD, 0, 0);
-            avifRWStreamWriteU32(&s, 1);                   // unsigned int(32) entry_count;
-            avifRWStreamWriteFullBox(&s, "url ", 0, 0, 1); // flags:1 means data is in this file
+            avifBoxMarker dinf;
+            AVIF_CHECKRES(avifRWStreamWriteBox(&s, "dinf", AVIF_BOX_SIZE_TBD, &dinf));
+            avifBoxMarker dref;
+            AVIF_CHECKRES(avifRWStreamWriteFullBox(&s, "dref", AVIF_BOX_SIZE_TBD, 0, 0, &dref));
+            AVIF_CHECKRES(avifRWStreamWriteU32(&s, 1)); // unsigned int(32) entry_count;
+            avifBoxMarker url;
```
wantehchang, line 2112: We can pass `NULL` to `avifRWStreamWriteFullBox()` and avoid the `url` variable.

Do you know why we don't need to call `avifRWStreamFinishBox(&s, url)`? We pass a `contentSize` of 0 (which is equal to `AVIF_BOX_SIZE_TBD`) to `avifRWStreamWriteFullBox()` in the next line.

wantehchang (changes requested):
There are a lot of changes in this pull request. Thanks a lot for spending the effort writing it.

This pull request looks good. Please first address the comment marked with "IMPORTANT", and then address the other comments. The comments marked as "optional" can be ignored; I mentioned those alternatives just for your consideration.

src/write.c
```diff
@@ -1358,6 +1383,154 @@ static size_t avifEncoderFindExistingChunk(avifRWStream * s, size_t mdatStartOff
     return 0;
 }
 
+static avifResult avifEncoderWriteMdat(avifEncoder * encoder,
```
wantehchang, line 1386: Nit: Rename this `avifEncoderWriteMediaDataBox `. More consistent with the `avifParse*Box` function names in src/read.c.

src/write.c
y-guyon: > Do you know why we don't need to call `avifRWStreamFinishBox(&s, url)`? We pass a `contentSize` of 0 (which is equal to `AVIF_BOX_SIZE_TBD`) to `avifRWStreamWriteFullBox()` in the next line.

Yes, the box size (0) is known in advance. `AVIF_BOX_SIZE_TBD` exists just to "make it clear that the box size is currently unknown", but a `contentSize` value of 0 is valid.

src/stream.c
y-guyon: `avifRWStreamFinishBox()` requires a marker as argument so it is clearer to have `avifRWStreamWriteBox()` output one in my opinion.

src/write.c
wantehchang: Thanks for the reply.

src/stream.c
wantehchang: Yes, on second thought I agree with you. It is better to treat `marker` as an opaque token that is passed from `avifRWStreamWriteBox()` to `avifRWStreamFinishBox()`.

src/write.c
```diff
@@ -376,34 +370,43 @@ static void avifItemPropertyDedupStart(avifItemPropertyDedup * dedup)
 
 // This compares the newly written item property (in the dedup's temporary storage buffer) to
 // already-written properties (whose offsets/sizes in outputStream are recorded in the dedup). If a
-// match is found, the previous item's index is used. If this new property is unique, it is
+// match is found, the previous property index is used. If this new property is unique, it is
```
wantehchang, line 373: Nit: property => property's

src/write.c
```diff
@@ -376,34 +370,43 @@ static void avifItemPropertyDedupStart(avifItemPropertyDedup * dedup)
...
 // assigned the next available property index, written to the output stream, and its offset/size in
 // the output stream is recorded in the dedup for future comparisons.
 //
-// This function always returns a valid 1-indexed property index for usage in a property association
-// (ipma) box later. If the most recent property was a duplicate of a previous property, the return
-// value will be the index of the original property, otherwise it will be the index of the newly
+// On success, this function adds to the given ipma box the association linking the reused or newly
```
wantehchang, line 377: Nit/optional: "the association" => "an association" or "a property association" ?

Also, the sentence sounds incomplete because it leaves me wondering "linking the property with what?" Perhaps add "to the item" or "with the item" at the end of the sentence?

wantehchang (approved):
LGTM. In this round of review, I only reviewed commits 6-10 before the final rebase. I hope the merge with https://github.com/AOMediaCodec/libavif/pull/1506 did not require real changes to this pull request.

I have two minor suggestions about the comments for the `avifItemPropertyDedupFinish()` function.

wantehchang (approved):
LGTM. Thanks.

PR #1501, opened 2023-08-11
Add experimental support for reading and writing gain maps.

maryla-uc:
The feature is hidden behind the AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP compile flag.

See https://helpx.adobe.com/camera-raw/using/gain-map.html for an introduction to gain maps. Note that in this change, gain maps are not stored as auxiliary items as proposed by Adobe, but as a hidden input to a new 'tmap' (tone mapped) derived item.
An 'altr' (alternatives group) box is also added to tell the decoder that it may show the tone mapped image (preferrably) or the base image as a fall back. The primary item id still points to the base image for backward compatibility.


src/write.c
```diff
@@ -537,6 +550,35 @@ static avifBool avifEncoderDetectChanges(const avifEncoder * encoder, avifEncode
...
+                                         const avifImage * imageMetadata,
+                                         struct ipmaArray * ipma,
+                                         avifItemPropertyDedup * dedup)
+{
+    avifRWStream * s = outputStream;
+    if (dedup) {
+        assert(ipma);
+        // Use the dedup's temporary stream for box writes
+        s = &dedup->s;
+    }
```
y-guyon, line 564: Move to arg, see https://github.com/AOMediaCodec/libavif/pull/1493#discussion_r1290782821 (also https://github.com/AOMediaCodec/libavif/pull/1493#discussion_r1291856361)

Same for `avifEncoderWriteHDRProperties()` below

src/write.c
```diff
@@ -678,6 +707,36 @@ static void avifEncoderWriteExtendedColorProperties(avifRWStream * dedupStream,
     }
 }
 
+// Same as 'avifEncoderWriteColorProperties' but for properties related to Hight Dynamic Range only.
```
y-guyon, line 710: Height* Dynamic Range

src/write.c
```diff
@@ -2040,7 +2414,7 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
         // Use multiple passes to pack in the following order:
         //   * Pass 0: metadata (Exif/XMP)
         //   * Pass 1: alpha (AV1)
-        //   * Pass 2: all other item data (AV1 color)
+        //   * Pass 2: all other item data (AV1 color, gain map)
```
y-guyon, line 2417: Should gain maps come before color for the same reason as alpha?

Unrelated note: These are not really "passes" but just data ordering. Some refactoring could make it clearer.

src/write.c
```diff
@@ -2054,12 +2428,14 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
 
         for (uint32_t itemIndex = 0; itemIndex < encoder->data->items.count; ++itemIndex) {
             avifEncoderItem * item = &encoder->data->items.item[itemIndex];
-            const avifBool isGrid = (item->gridCols > 0); // Grids store their payload in metadataPayload, so use this to distinguish grid payloads from XMP/Exif
+            // Grids and tmap store their payloads in metadataPayload, so use this to distinguish grid payloads from XMP/Exif
+            const avifBool isMetadata = (!memcmp(item->type, "mime", 4) || !memcmp(item->type, "Exif", 4)) &&
+                                        (item->metadataPayload.size > 0);
```
y-guyon, line 2433: I think `&& (item->metadataPayload.size > 0)` is unnecessary here, and the comment can be removed in my opinion

src/read.c
```diff
@@ -3701,6 +3797,56 @@ static avifResult avifDecoderDataFindAlphaItem(avifDecoderData * data,
     return AVIF_RESULT_OK;
 }
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+// Finds a 'tmap' (tone mapped image item) box associated with the given 'colorItem'.
+// If found, fills 'toneMappedImageItem' and  sets 'gainMapItemID' to the id of the gain map
+// item associated with the box. Otherwise, sets 'toneMappedImageItem' to NULL.
+// Returns AVIF_RESULT_OK on success (whether or not a tmap box was found).
```
y-guyon, line 3804: Will a grid of `tmap` cells be forbidden in the specification? Otherwise, there should be a comment or TODO mentioning the lack of support for that pattern at decoding. It is fine not to support that at encoding, as discussed.

include/avif/avif.h
```diff
@@ -499,6 +504,87 @@ typedef struct avifContentLightLevelInformationBox
...
+// Gain Maps are a HIGHLY EXPERIMENTAL FEATURE. The format might still change and
+// images containing a gain map encoded with the current version of libavif might
+// not decode with a feature future version of libavif. Use are your own risk.
+// This product includes Gain Map technology under license by Adobe.
+
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+struct avifImage;
+
+// Gain map metadata, for tone mapping between SDR and HDR.
+// All float values should be positive.
```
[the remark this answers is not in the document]
maryla-uc: Changed to uint pairs, will add helper functions separately.

include/avif/avif.h
```diff
@@ -557,13 +643,20 @@ typedef struct avifImage
     // Metadata - set with avifImageSetMetadata*() before write, check .size>0 for existence after read
     avifRWData exif;
     avifRWData xmp;
+
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+    // Gain map image and metadata. If no gain map is present, avifGainMap.gainMap is NULL.
```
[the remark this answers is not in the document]
maryla-uc: Good catch, that was because I renamed the field.

src/utils.c
```diff
@@ -235,3 +234,38 @@ avifBool avifFractionSub(avifFraction a, avifFraction b, avifFraction * result)
...
+    if (v < 0) {
+        return AVIF_FALSE;
+    }
+    if (v == 0.0f) {
+        f->n = 0u;
+        f->d = 1u;
+        return AVIF_TRUE;
+    }
+
+    if ((float)(uint32_t)v == v) {
```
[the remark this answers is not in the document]
maryla-uc: Will do this in another PR.

src/utils.c
```diff
@@ -235,3 +234,38 @@ avifBool avifFractionSub(avifFraction a, avifFraction b, avifFraction * result)
     avifFractionSimplify(result);
     return AVIF_TRUE;
 }
+
+avifBool avifToFractionU(float v, avifFractionU * f)
```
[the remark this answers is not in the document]
maryla-uc: Will do this in another PR.

tests/gtest/avifgainmaptest.cc
```diff
@@ -0,0 +1,411 @@
...
+  //      .write(reinterpret_cast<char*>(encoded.data), encoded.size);
+}
+
+TEST(GainMapTest, EncodeDecodeBaseImageHdr) {
+  testutil::AvifImagePtr image =
+      testutil::CreateImage(/*width=*/12, /*height=*/34, /*depth=*/10,
+                            AVIF_PIXEL_FORMAT_YUV420, AVIF_PLANES_ALL);
+  ASSERT_NE(image, nullptr);
+  image->transferCharacteristics =
+      AVIF_TRANSFER_CHARACTERISTICS_SMPTE2084;  // PQ
```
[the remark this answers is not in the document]
maryla-uc: There are other small differences.

src/read.c
```diff
@@ -1800,6 +1822,77 @@ static avifBool avifParseImageGridBox(avifImageGrid * grid,
...
+    AVIF_CHECK(avifROStreamReadU32(s, &numerator));
+    AVIF_CHECK(avifROStreamReadU32(s, &denominator));
+    if (denominator == 0) {
+        return AVIF_FALSE;
+    }
+    *f = (float)numerator / denominator;
+    return AVIF_TRUE;
+}
+
+static avifBool avifParseFloatsAsFractions(avifROStream * s, float * f, int numFloats)
```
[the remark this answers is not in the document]
maryla-uc: It's not how I would've done it either but tihs was my interpretation of the draft spec.

src/read.c
maryla-uc: Not sure if it will be allowed. Added a comment/TODO.

tests/gtest/avifgainmaptest.cc
```diff
@@ -0,0 +1,411 @@
...
+  image->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_SRGB;
+  testutil::FillImageGradient(image.get());
+
+  testutil::AvifImagePtr gain_map =
+      testutil::CreateImage(/*width=*/6, /*height=*/17, /*depth=*/8,
+                            AVIF_PIXEL_FORMAT_YUV420, AVIF_PLANES_YUV);
+  ASSERT_NE(gain_map, nullptr);
+  testutil::FillImageGradient(gain_map.get());
+  gain_map->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_FCC;
+  // While attached to the gain map, this represents the clli information of the
```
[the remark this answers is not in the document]
maryla-uc: I meant it as "Even though"

src/read.c
```diff
@@ -3962,6 +4108,71 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
+                    return AVIF_RESULT_INVALID_IMAGE_GRID;
+                }
+            } else {
+                gainMapCodecType = avifGetCodecType(gainMapItem->type);
+                assert(gainMapCodecType != AVIF_CODEC_TYPE_UNKNOWN);
+            }
+
+            decoder->image->gainMap.image = avifImageCreateEmpty();
+
+            // Look for a colr nclx box. Other colr box types (e.g. ICC) are not supported.
```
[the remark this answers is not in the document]
maryla-uc: AVIF_RESULT_NOT_IMPLEMENTED suggests that this is a valid use case and that it's not currently implemented, but might be in the future, right?

I don't think an icc profile makes sense for the gain map, so AVIF_RESULT_NOT_IMPLEMENTED does not seem right to me.
The draft spec says `The gain map input image shall be associated with a colr item property of type nclx.` Not sure if that's supposed to imply if other types are forbidden or not (in addition to nclx).

Should I ignore other boxes (like I'm doing now, since I don't know what to do with them) or return some parsing error (AVIF_RESULT_BMFF_PARSE_FAILED ?) or return AVIF_RESULT_NOT_IMPLEMENTED assuming that maybe it's a new type of box that I don't support?

include/avif/avif.h
```diff
@@ -499,6 +504,87 @@ typedef struct avifContentLightLevelInformationBox
...
+{
+    // Gain map pixels.
+    // Used fields: width, height, depth, yufFormat, yuvRange,
+    // yuvChromaSamplePosition, yuvPlanes, yuvRowBytes, imageOwnsYUVPlanes,
+    // matrixCoefficients, transferCharacteristics. Other fields are ignored.
+    struct avifImage * image;
+
+    // Gain map metadata.
+    // For an image grid, the metadata should be identical for all cells.
+    avifGainMapMetadata metadata;
```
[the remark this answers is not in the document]
maryla-uc: Acknowledged. The gain map metadata struct is about 140 bytes. I think it's negligible compared to pixel data and other memory used by libavif/encoders/decoders. It's equivalent to less than a 7x7 pixel image with 3 channels.

tests/gtest/avifgainmaptest.cc
```diff
@@ -114,16 +134,8 @@ TEST(GainMapTest, EncodeDecodeBaseImageHdr) {
...
-  image->gainMap.metadata.hdrCapacityMin = 1.0f;
-  image->gainMap.metadata.hdrCapacityMax = 8.0f;
-  for (int c = 0; c < 3; ++c) {
-    image->gainMap.metadata.offsetSdr[c] = 1.f / (10.f * (c + 1.f));
-    image->gainMap.metadata.offsetHdr[c] = 1.f / (20.f * (c + 1.f));
-    image->gainMap.metadata.gainMapGamma[c] = 1.f / (c + 1.f);
-    image->gainMap.metadata.gainMapMin[c] = 1.f / (c + 1.f);
-    image->gainMap.metadata.gainMapMax[c] = 10 + 1.f / (c + 1.f);
-  }
+  image->gainMap.metadata = image->gainMap.metadata =
```
y-guyon, line 137: Twice `image->gainMap.metadata =`

tests/gtest/avifgridapitest.cc
```diff
@@ -76,6 +76,11 @@ avifResult EncodeDecodeGrid(const std::vector<std::vector<Cell>>& cell_rows,
   testutil::AvifImagePtr grid = testutil::CreateImage(
       static_cast<int>(image->width), static_cast<int>(image->height),
       /*depth=*/8, yuv_format, AVIF_PLANES_ALL);
+  const int num_rows = (int)cell_rows.size();
+  const int num_cols = (int)cell_rows[0].size();
+  AVIF_CHECKRES(
+      testutil::MergeGrid(num_cols, num_rows, cell_images, grid.get()));
+
```
y-guyon, line 83: Rows below should be removed now that there is `testutil::MergeGrid()`

Note: `testutil::MergeGrid()` could be made in another PR to reduce the size of this one

tests/gtest/aviftest_helpers.cc
```diff
@@ -348,6 +349,55 @@ bool AreImagesEqual(const avifRGBImage& image1, const avifRGBImage& image2) {
...
+  const uint32_t tile_height = cells[0]->height;
+  const uint32_t grid_width =
+      (grid_cols - 1) * tile_width + cells.back()->width;
+  const uint32_t grid_height =
+      (grid_rows - 1) * tile_height + cells.back()->height;
+
+  testutil::AvifImagePtr view(avifImageCreateEmpty(), avifImageDestroy);
+  if (!view) {
+    return AVIF_RESULT_OUT_OF_MEMORY;
+  }
```
y-guyon, line 375: Alternative: `AVIF_CHECKERR(view, AVIF_RESULT_OUT_OF_MEMORY);`

Below `AVIF_CHECKRES(avifImageSetViewRect(view.get(), merged, &rect));` too

y-guyon (approved):

maryla-uc:
Note I rebased this patch but had some issues with the merge and I had to add my latests fixes to the last commit, apologies for the confusing history.

tests/gtest/avifgainmaptest.cc
maryla-uc: Indeed!

tests/gtest/avifgridapitest.cc
maryla-uc: Indeed I forgot to remove them, thanks for spotting it.
I agree it could be split... but I'm afraid of spending a while fighting with git and merges again... So if it's not a strong request I'd rather leave it as is.

src/write.c
```diff
@@ -2428,23 +2406,25 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
...
+            const avifBool isMetadata = !memcmp(item->type, "mime", 4) || !memcmp(item->type, "Exif", 4);
             if (metadataPass != isMetadata) {
                 // only process metadata (XMP/Exif) payloads when metadataPass is true
                 continue;
             }
-            if (alphaPass != (item->itemCategory == AVIF_ITEM_ALPHA)) {
-                // only process alpha payloads when alphaPass is true
+            avifBool isAlphaOrGainMap = item->itemCategory == AVIF_ITEM_ALPHA;
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+            isAlphaOrGainMap = isAlphaOrGainMap || item->itemCategory == AVIF_ITEM_GAIN_MAP || !memcmp(item->type, "tmap", 4);
```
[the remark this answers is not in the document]
maryla-uc: That would work but it's not exactly the same thing since |= is a bitwise operator, and won't skip evaluating the right side if the value is already true. Not that I think the performance matters much, but I think it's better to keep |= for actual bitwise operations like flags like we do now.

include/avif/avif.h
```diff
@@ -1156,6 +1269,9 @@ typedef struct avifEncoder
     // changeable encoder settings
     int quality;
     int qualityAlpha;
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+    int qualityGainMap;
+#endif
```
wantehchang, line 1274: In anticipation of the libavif 1.0.0 release, the new member `qualityGainMap` should be added to the end of the `avifEncoder` struct, after all the current struct members.

include/avif/avif.h
```diff
@@ -994,6 +1102,11 @@ typedef struct avifDecoder
     // avifDecoderNextImage() or avifDecoderNthImage(), as decoder->image->alphaPlane won't exist yet.
     avifBool alphaPresent;
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+    // This is true when avifDecoderParse() detects a gain map.
+    avifBool gainMapPresent;
+#endif
+
```
wantehchang, line 1109: In anticipation of the libavif 1.0.0 release, the new member `gainMapPresent` should be added to the end of the `avifDecoder` struct, after all the current struct members.

include/avif/avif.h
```diff
@@ -846,6 +947,12 @@ typedef struct avifIOStats
     size_t colorOBUSize;
     // Size in bytes of the AV1 image item or track data containing alpha samples.
     size_t alphaOBUSize;
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+    // Size in bytes of the AV1 image item containing gain map samples.
+    size_t gainMapOBUSize;
```
wantehchang, line 952: Nit: Remove "OBU" from the member's name. Or perhaps change "OBU" to "Image"?

include/avif/avif.h
```diff
@@ -846,6 +947,12 @@ typedef struct avifIOStats
...
+#else
+    size_t reserved;
+#endif
```
wantehchang, line 955: IMPORTANT: If it is necessary, or very convenient, to store the gain map size in the `avifIOStats` struct, we need to prepare for this in the libavif 1.0.0 release.

There are two options:

(1) Add enough reserved fields at the end of the `avifIOStats` struct.

(2) Change the declaration of the `ioStats` member of `avifDecoder` and `avifEncoder` from

```
    avifIOStats ioStats;
```

to

```
    avifIOStats * ioStats;
```

and let `avifDecoderCreate()` and `avifEncoderCreate()` allocate the `avifIOStats` struct.

Alternatively, we can store the gain map size in the `avifDecoderData` and `avifEncoderData` structs, and add getter functions to extract the gain map size from `avifDecoder` or `avifEncoder`.

include/avif/avif.h
```diff
@@ -557,13 +650,20 @@ typedef struct avifImage
...
+    avifGainMap gainMap;
+#endif
 } avifImage;
 
 // avifImageCreate() and avifImageCreateEmpty() return NULL if arguments are invalid or if a memory allocation failed.
 AVIF_API avifImage * avifImageCreate(uint32_t width, uint32_t height, uint32_t depth, avifPixelFormat yuvFormat);
 AVIF_API avifImage * avifImageCreateEmpty(void); // helper for making an image to decode into
-AVIF_API avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes); // deep copy
-AVIF_API avifResult avifImageSetViewRect(avifImage * dstImage, const avifImage * srcImage, const avifCropRect * rect); // shallow copy, no metadata
+AVIF_API avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes); // deep copy, including gain map
```
wantehchang, line 665: Nit: The "including gain map" could be omitted. Alternatively, the new comment should also mention metadata. See line 666.

wantehchang (changes requested):
Maryla: I only reviewed the public header avif.h. I suggest some changes in anticipation of the libavif 1.0.0 release.

CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ option(BUILD_SHARED_LIBS "Build shared avif library" ON)
 option(AVIF_ENABLE_WERROR "Treat all compiler warnings as errors" ON)
 
 option(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R "Enable experimental YCgCo-R matrix code" OFF)
+option(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP "Enable experimental gain map code (for HDR images that look good both on HDR and SDR displays)" OFF)
```
wantehchang, line 37: Nit: Please add a comment somewhere (either in this file or in avif.h) to specify the ISO proposal/contribution that specifies the gain maps implemented in this pull request.

include/avif/avif.h
wantehchang: If this is the size of the output of the AV1 encoder (or the input to the AV1 decoder), then it's fine to use "OBU" in its name for consistency.

Note: It would have been more accurate to call it "Bitstream". An AV1 bitstream is divided into _open bitstream units_ (OBUs), so an OBU is just a segment of a bitstream.

maryla-uc: Removed this field.

include/avif/avif.h
maryla-uc: I removed this field. I don't have any particular need for it, I simply added it to mirror the other two fields.
I was chatting with Yannis and he was wondering if we could get rid of this struct entirely.

include/avif/avif.h
maryla-uc: Expanded the comments.

CMakeLists.txt
maryla-uc: Done in avif.h
I don't think a link exists but I added the name of the (draft) spec.

tests/gtest/avifgridapitest.cc
```diff
@@ -76,32 +76,11 @@ avifResult EncodeDecodeGrid(const std::vector<std::vector<Cell>>& cell_rows,
...
-      }
-      avifImageCopySamples(/*dstImage=*/view.get(), it->get(), AVIF_PLANES_ALL);
-      assert(!view->imageOwnsYUVPlanes);
-      ++it;
-      rect.x += rect.width;
-    }
-    rect.y += rect.height;
-  }
-  if ((rect.x != image->width) || (rect.y != image->height) ||
-      !testutil::AreImagesEqual(*image, *grid)) {
```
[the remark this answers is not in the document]
maryla-uc: Woops, indeed I removed too much, thanks for catching this.

src/write.c
```diff
@@ -630,16 +627,17 @@ static avifResult avifEncoderWriteColorProperties(avifRWStream * outputStream,
...
 
     return avifEncoderWriteExtendedColorProperties(dedupStream, outputStream, imageMetadata, ipma, dedup);
 }
 
-static avifResult avifEncoderWriteExtendedColorProperties(avifRWStream * dedupStream,
-                                                          avifRWStream * outputStream,
-                                                          const avifImage * imageMetadata,
-                                                          struct ipmaArray * ipma,
-                                                          avifItemPropertyDedup * dedup)
+// Same as 'avifEncoderWriteColorProperties' but for properties related to Hight Dynamic Range only.
```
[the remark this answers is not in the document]
maryla-uc: I thought I fixed this before but maybe it got lost in a merge :/

include/avif/avif.h
```diff
@@ -936,6 +1040,7 @@ typedef struct avifDecoder
     // If you don't actually leverage this data, it is best to ignore it here.
     avifBool ignoreExif;
     avifBool ignoreXMP;
+    // TODO(maryla): add flags to ignore the main image, or ignore the gain map if desired.
```
wantehchang, line 1043: Optional: Move this TODO comment to the end of the struct to reduce the chance of accidentally adding the new flags here.

apps/avifenc.c
```diff
@@ -1945,7 +1945,7 @@ int main(int argc, char * argv[])
         }
     }
 
-    avifIOStats ioStats = { 0, 0, 0 };
+    avifIOStats ioStats = { 0, 0 };
```
wantehchang: Maryla: It would be good to modify the following code in avifenc.c:

```
    printf("Encoded successfully.\n");
    printf(" * Color AV1 total size: %" AVIF_FMT_ZU " bytes\n", ioStats.colorOBUSize);
    printf(" * Alpha AV1 total size: %" AVIF_FMT_ZU " bytes\n", ioStats.alphaOBUSize);
```

to also print the gain map AV1 total size, and see if you find it useful to report that information.

maryla-uc: Actually, there's no way to encode a gain map with avifenc right now, only through the API, so this can be done later I think.

wantehchang (commented):
avif.h LGTM.

wantehchang (commented):
Maryla: Could you add an entry to CHANGELOG.md? Please indicate that this is an experimental feature. Thanks.

wantehchang (commented):
Maryla: In this round of review, I only checked whether the pull request is essentially a no-op if the experimental feature is disabled. Unfortunately I didn't make a complete pass through the PR. (I stopped after src/read.c, so I haven't reviewed src/write.c.)

This PR has more code and comments outside `#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)`. We should try to move as much code into `#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)` as possible. It is more difficult to deal with the comments. I suggested adding "if AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP is defned" to the comments, but that may become a little wordy.

PR #1503, opened 2023-08-17
Rename s to dedupStream in WriteColorProperties()

y-guyon:
[skip ci]

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -538,10 +538,6 @@ static avifBool avifEncoderDetectChanges(const avifEncoder * encoder, avifEncode
 }
 
 // Subset of avifEncoderWriteColorProperties() for the properties clli, pasp, clap, irot, imir.
-// outputStream is the final bitstream that will be output by the libavif encoder API.
-// dedupStream is either equal to outputStream or to dedup->s which is a temporary stream used
-// to store parts of the final bitstream; these parts may be discarded if they are a duplicate
-// of an already stored property.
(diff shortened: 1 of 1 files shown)


wantehchang (approved):
LGTM.

I would also be fine with renaming the first parameter of `avifEncoderWriteExtendedColorProperties()` as `s`. It was just a little odd to see that `avifRWStream` variable change name as it crosses function boundaries.

It may be better to add comments to explain `dedupStream` vs. `outputStream`...

y-guyon:
> It may be better to add comments to explain `dedupStream` vs. `outputStream`...

Done.

src/write.c
```diff
@@ -538,6 +538,10 @@ static avifBool avifEncoderDetectChanges(const avifEncoder * encoder, avifEncode
 }
 
 // Subset of avifEncoderWriteColorProperties() for the properties clli, pasp, clap, irot, imir.
+// outputStream is the final bitstream that will be output by the libavif encoder API.
+// dedupStream is either equal to outputStream or to dedup->s which is a temporary stream used
```
wantehchang, line 542: Nit: Add `&` before `dedup->s`.

Optional: It might be good to move the new comment closer to `avifEncoderWriteColorProperties()`, where these variables/parameters are actually used.

wantehchang (approved):
LGTM. Thank you very much for adding the new comment.

wantehchang (approved):

PR #1506, opened 2023-08-18
Add StreamReadBits() and StreamWriteBits()

y-guyon:
Simplify calling sites that were doing manual bit manipulation.
Also add VarInt versions.

https://github.com/AOMediaCodec/libavif/pull/1432 will be simplified when this PR is merged.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -453,6 +453,12 @@ typedef struct avifROStream
     // Index of the next byte in the raw stream.
     size_t offset;
 
+    // Index of the currently selected partial byte. Must be lower than offset.
+    // Only meaningful if numUsedBitsInPartialByte is not zero.
+    size_t offsetOfPartialByte;
+    // Number of bits already used in the currently selected partial byte.
+    size_t numUsedBitsInPartialByte;
+
     // Error information, if any.
     avifDiagnostics * diag;
     const char * diagContext;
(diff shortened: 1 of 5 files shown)


maryla-uc (approved):
Nice! I'm wondering if the other, whole byte writing/reading functions should check that numUsedBitsInPartialByte is 0? to make sure that padding bits were properly added? otherwise if I understand correctly, the next bit read/write will keep acting on a previous byte which seems like a weird api? but maybe that's intentional or maybe I didn't read carefully enough

y-guyon:
> Nice! I'm wondering if the other, whole byte writing/reading functions should check that numUsedBitsInPartialByte is 0?

`avifROStreamSetOffset()` already had `assert(stream->numUsedBitsInPartialByte == 0);`. I added the assertion to other functions. `offsetOfPartialByte` made no sense anymore (as described in https://github.com/AOMediaCodec/libavif/pull/1432#discussion_r1298490182) so I removed it. I fixed avifstreamtest too because it tested that pattern.

> otherwise if I understand correctly, the next bit read/write will keep acting on a previous byte which seems like a weird api

This is the behavior of [libwebp2](https://chromium.googlesource.com/codecs/libwebp2/+/c88083e84c68e6babb192d40104dcf47676a4ee1/src/common/header_enc_dec.cc#149) for what it is worth.

> but maybe that's intentional or maybe I didn't read carefully enough

Yes it was intentional. I will add it back one day if needed.

maryla-uc (approved):
Thanks for the details. I didn't realize webp2 did this. I think it can work but it's not really in the spirit of isobmff boxes. You could have bits leaking into other boxes which seems weird. Although in practice we're following what the specs say anyway, and they use padding bits. I think it's less error prone to check that padding was indeed properly added (like you did).

PR #1507, opened 2023-08-18
reformat_libyuv: Add ignore CFI annotation

vigneshvg:
libyuv is a C++ library and defines custom types (struct, enum, etc) in a namespace when conditionally compiled with a C++ compiler. When accessed from a C library like libavif, via a function pointer, this leads to signature mismatches in the CFI sanitizers since the C++ header file defines the types within the namespace and the C code has the types without the namespace. In order to avoid this CFI error, we tag some functions as an exception when being compiled with CFI enabled.

For more details on clang's CFI see:
https://clang.llvm.org/docs/ControlFlowIntegrity.html.

For a simpler example of this bug, please see:
https://github.com/vigneshvg/cpp_c_potential_cfi_bug

Sample CFI failure in Chromium: crbug.com/1473228

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -60,6 +60,18 @@ unsigned int avifLibYUVVersion(void)
 #pragma clang diagnostic pop
 #endif
 
+// libyuv is a C++ library and defines custom types (struct, enum, etc) in the libyuv namespace when the libyuv header files are
+// included by C++ code. When accessed from a C library like libavif, via a function pointer, this leads to signature mismatches
+// in the CFI sanitizers since libyuv itself, compiled as C++ code, has the types within the namespace and the C code has the
+// types without the namespace. In order to avoid this CFI error, we tag some functions as an exception when being compiled with
+// CFI enabled.  For more details on clang's CFI see: https://clang.llvm.org/docs/ControlFlowIntegrity.html. For a simpler example
+// of this bug, please see: https://github.com/vigneshvg/cpp_c_potential_cfi_bug
+#if defined(__clang__)
+#define IGNORE_CFI_ICALL __attribute__((no_sanitize("cfi-icall")))
+#else
+#define IGNORE_CFI_ICALL
+#endif
+
 //--------------------------------------------------------------------------------------------------
 // libyuv API availability management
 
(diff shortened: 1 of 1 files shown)


src/reformat_libyuv.c
```diff
@@ -60,6 +60,18 @@ unsigned int avifLibYUVVersion(void)
 #pragma clang diagnostic pop
 #endif
 
+// libyuv is a C++ library and defines custom types (struct, enum, etc) in a namespace when conditionally compiled with a C++
```
wantehchang, line 63: 1. Nit: We can be explicit and say the namespace is `libyuv`.

2. Nit: Remove "conditionally". It would be more precise to say "when the libyuv header files are included by C++ code". It is important for this comment to say it is the libyuv header files that are compiled with a C++ compiler, because the libyuv implementation (all .cc files) is always compiled with a C++ compiler.

src/reformat_libyuv.c
```diff
@@ -60,6 +60,18 @@ unsigned int avifLibYUVVersion(void)
...
+// compiler. When accessed from a C library like libavif, via a function pointer, this leads to signature mismatches in the CFI
+// sanitizers since the C++ header file defines the types within the namespace and the C code has the types without the namespace.
```
wantehchang, line 65: Nit/optional: We should change "the C++ header file defines the types within the namespace" to something like the following:

> libyuv itself, compiled as C++ code, has the types within the namespace

wantehchang (approved):
LGTM. I suggest some changes to the comments.

Note: Another simple solution would be to compile src/reformat_libyuv.c as C++ code. Then this file and libyuv itself will see libyuv's types in the libyuv namespace, and CFI will be happy. We will need to make sure this does not add a dependency on the C++ runtime library.

vigneshvg:
> LGTM. I suggest some changes to the comments.
> 
> Note: Another simple solution would be to compile src/reformat_libyuv.c as C++ code. Then this file and libyuv itself will see libyuv's types in the libyuv namespace, and CFI will be happy. We will need to make sure this does not add a dependency on the C++ runtime library.

Ack. I agree that's another solution, but would rather get away with the annotation than meddling around the compilation settings for this file alone in various environments.

src/reformat_libyuv.c
vigneshvg: Done.

src/reformat_libyuv.c
vigneshvg: Done.

wantehchang (approved):
LGTM. Thanks.

wantehchang (approved):

PR #1510, opened 2023-08-22
Add "Version 1.0.0 ends here" comments

wantehchang:
Add "Version 1.0.0 ends here" comments to the end of the avifImage, avifDecoder, and avifEncoder structs. New members in future libavif 1.x.y releases should be added to these structs after the "Version 1.0.0 ends here" comments, followed by "Version 1.x.y ends here" comments.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -552,6 +552,8 @@ typedef struct avifImage
     // Metadata - set with avifImageSetMetadata*() before write, check .size>0 for existence after read
     avifRWData exif;
     avifRWData xmp;
+
+    // Version 1.0.0 ends here.
 } avifImage;
 
 // avifImageCreate() and avifImageCreateEmpty() return NULL if arguments are invalid or if a memory allocation failed.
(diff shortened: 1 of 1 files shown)


include/avif/avif.h
```diff
@@ -552,6 +552,8 @@ typedef struct avifImage
     // Metadata - set with avifImageSetMetadata*() before write, check .size>0 for existence after read
     avifRWData exif;
     avifRWData xmp;
+
+    // Version 1.0.0 ends here.
```
jzern, line 556: Do you want to extend this note that any new members be added after this line?

jzern (approved):

y-guyon (approved):
If we are certain that new fields will be added to these structs, I am fine with adding these comments now. Otherwise I suggest waiting for new fields to be added before adding these comments, for simplicity and consistency with other structs.

wantehchang:
Yannis: This pull request adds a comment to these three structs because these structs can only be allocated by calling a Create() function and therefore may be extended in a future 1.x.y release.

Yes, we are certain that new fields will be added to these structs. https://github.com/AOMediaCodec/libavif/pull/1432 will add a new field to the `avifEncoder` struct, and https://github.com/AOMediaCodec/libavif/pull/1501 will add a new field to each of the three structs.

PR #1515, opened 2023-08-24
Hash-pin GitHub workflows, set up dependabot to keep them up-to-date

### LINKED ISSUE #1514: Hash-pin GitHub Actions used in workflows
Hey, it's Pedro (see #1328 and #1342) and I'm back with a new security suggestion!

When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v3`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

These hashes can be automatically updated by dependabot. Whenever a new version of an Action is released, you'll receive a PR updating both its hash and the version comment (see [this repo](https://github.com/pnacht/logr/pulls) as an example).

I'll send a PR pinning the Actions along with this issue.

pnacht:
Fixes #1514.

This PR hash-pins all GitHub workflows and sets up dependabot to ensure they remain updated.

Dependabot is currently configured to scan for new versions weekly, but this can be changed to monthly if you prefer. Let me know!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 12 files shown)


jzern (approved):
Thanks for the patch and the explanation in the issue. Wan-Teh, any thoughts on this?

wantehchang:
James: Let's wait until after the libavif 1.0.0 release to merge this pull request.

Based on Pedro's explanation, I think this is a good change. The version tags are still available as comments.

Scanning for new versions weekly seems like a reasonable frequency, so I'd just adopt Pedro's recommendation.

wantehchang:
James: We can merge this pull request now.

Pedro, should we squash the two commits in this pull request, or are they logically separate commits?

pnacht:
I'd say squashing is perfectly fine!

PR #1518, opened 2023-08-25
Revert "Guard diag in avifDiagnosticsClearError()"

wantehchang:
This reverts commit 173961dac588d5b7cf0d6749d1aada9050b6197a in the 1.0.0 release, to give us more time to re-evaluate https://github.com/AOMediaCodec/libavif/pull/1496.

### DIFF
--- a/src/diag.c
+++ b/src/diag.c
@@ -9,9 +9,7 @@
 
 void avifDiagnosticsClearError(avifDiagnostics * diag)
 {
-    if (diag) {
-        *diag->error = '\0';
-    }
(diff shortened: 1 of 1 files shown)


wantehchang:
While working on the changelog for the 1.0.0 release yesterday, I realized I forgot to follow up on Joe's comment in https://github.com/AOMediaCodec/libavif/pull/1492#issuecomment-1666173628. I suggest we revert the change to avifDiagnosticsClearError() in the 1.0.0 release.

y-guyon (approved):

Issue #116, opened 2020-03-19
Static build of libavif and position-independent code


novomesk:
I create a library (libqavif.so) which is a QT imageformats plug-in.
I want to make the libqavif.so stand-alone, without dynamic linking to libavif.so
I can do it when I compile everything as position independent code.
libaom has option to enable it: `-DCONFIG_PIC=1`
I can build libavif as PIC by running:
`CFLAGS="-fPIC" cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON ..`
`-fPIC` is for gcc. llc uses `-relocation-model=pic`

I think **CONFIG_PIC** option would be useful for libavif too.

vrabaud:
Thx for reporting, sorry for not keeping up: I believe this has been fixed in https://github.com/AOMediaCodec/libavif/pull/1556

PR #1530, opened 2023-08-29
Add CI for AVIF_BUILD_GDK_PIXBUF.

vrabaud:
For static, we also need to set POSITION_INDEPENDENT_CODE. We could do that by default for AVIF static (as done in libwebp and libaom).


kmilos:
Btw, it doesn't look like old CI jobs are getting cancelled when you push a new commit to a PR? It would be good to add some [concurrency control](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow) for sustainability reasons ;)

wantehchang:
@novomesk Daniel: What is the cmake command line you are using to build libavif with `-DAVIF_BUILD_GDK_PIXBUF=ON`?

Vincent: Thank you very much for writing this pull request. I suggest we try NOT to deal with position-independent code (PIC) in this PR. Here is a cmake command line that works for me locally on Linux:

```
cmake -G Ninja .. \
  -DBUILD_SHARED_LIBS=OFF \
  -DAVIF_ENABLE_WERROR=ON \
  -DAVIF_CODEC_AOM=ON \
  -DAVIF_CODEC_AOM_DECODE=OFF -DAVIF_CODEC_AOM_ENCODE=ON \
  -DAVIF_CODEC_DAV1D=ON \
  -DAVIF_BUILD_EXAMPLES=ON \
  -DAVIF_BUILD_APPS=ON \
  -DAVIF_BUILD_GDK_PIXBUF=ON
```

This requires installing the libaom and dav1d Linux packages. So ci-unix-shared-installed.yml is the most similar CI build. Could you please try modifying ci-unix-shared-installed.yml? It mght be better to add a new CI build just to test gdk-pixbuf.

novomesk:
I also discovered in the past that PIC is needed when libavif is statically linked to a library:
https://github.com/AOMediaCodec/libavif/issues/116

On Gentoo, we use dynamic linking to libavif ( -DBUILD_SHARED_LIBS=ON), so it is enough to add only `-DAVIF_BUILD_GDK_PIXBUF=ON` and the required dependencies.
https://github.com/gentoo/gentoo/blob/973a020cf85c6ed8fdd6a22f54c534babad33673/media-libs/libavif/libavif-0.10.1.ebuild#L42

vrabaud:
Simpler alternative that does not mess with PIC in https://github.com/AOMediaCodec/libavif/pull/1532

PR #1534, opened 2023-08-29
Use local copies of ScalePlane* functions when libyuv is not found

### LINKED ISSUE #709: Copy libyuv's ScalePlane*() functions to libavif
Right now the `avifImageScale()` function in libavif/src/scale.c is not implemented unless libyuv is enabled. We should copy libyuv's ScalePlane*() functions to libavif. It would suffice to just copy the C implementation.

If Google owns the copyright of those functions, I can contribute the code to libavif as a Google employee. This will simplify the source code license.

Cacodemon345:
Fixes #709.

Note: Copyright/license headers of copied-over files may be rough around the edges as I had to change them to be similar to the one in obu.c. Please note any problems regarding those. Thanks.


CMakeLists.txt
```diff
@@ -304,6 +304,10 @@ set(AVIF_CODEC_DEFINITIONS)
 set(AVIF_CODEC_INCLUDES)
 set(AVIF_CODEC_LIBRARIES)
 
+if(NOT libyuv_FOUND)
+    set(AVIF_SRCS ${AVIF_SRCS} src/scale/scale_impl.c src/scale/scale.c)
```
y-guyon, line 308: A comment here or in src/scale/README.md would be nice. It should include the libyuv version/commit at which the sources were copied from, or at least where to look it up (in avif_scale.h right now).

LICENSE
```diff
@@ -351,3 +351,32 @@ Files: android_jni/gradlew*
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+------------------------------------------------------------------------------
+
+Files: src/scale/*
```
y-guyon, line 357: I wonder whether the new folder should contain "libyuv" in its name. It would make it clear that it exclusively contains libyuv-imported code.

LICENSE
```diff
@@ -351,3 +351,32 @@ Files: android_jni/gradlew*
...
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
y-guyon, line 382: I verified that this text exactly matches https://chromium.googlesource.com/libyuv/libyuv/+/6ff369102723dea853b1642649f991d0ca7b292b/LICENSE.

src/scale.c
```diff
@@ -38,6 +23,7 @@ avifBool avifImageScale(avifImage * image,
 #if defined(__clang__)
 #pragma clang diagnostic pop
 #endif
+#endif
```
y-guyon, line 26: This was not the case at head but please comment the name of the conditional define:
```c
#endif
#endif // AVIF_LIBYUV_ENABLED
```

src/scale.c
```diff
@@ -4,25 +4,10 @@
...
-    (void)dstHeight;
-    (void)imageSizeLimit;
-    (void)imageDimensionLimit;
-    avifDiagnosticsPrintf(diag, "avifImageScale() called, but is unimplemented without libyuv!");
-    return AVIF_FALSE;
-}
-
+#include "scale/avif_scale.h"
+#define ScalePlane_12 avifScalePlane_12
+#define ScalePlane avifScalePlane
```
y-guyon, line 9: I would rather use a `#if defined(AVIF_LIBYUV_ENABLED)` below than these hard-coded aliases.
If we keep `ScalePlane()` and `ScalePlane_12()` exactly as they are in libyuv, I think it is fine to keep their original names. No need to prepend `avif`.

src/scale.c
```diff
@@ -4,25 +4,10 @@
...
-    (void)image;
-    (void)dstWidth;
-    (void)dstHeight;
-    (void)imageSizeLimit;
-    (void)imageDimensionLimit;
-    avifDiagnosticsPrintf(diag, "avifImageScale() called, but is unimplemented without libyuv!");
-    return AVIF_FALSE;
-}
-
+#include "scale/avif_scale.h"
```
y-guyon, line 7: Rather than relying on a custom header file, could we use `extern`-imported functions here? This question is only relevant if no original libyuv header containing `ScalePlane()`/`ScalePlane_12()` definitions is imported.

src/scale/avif_scale.h
y-guyon, line 1: If possible, I would avoid this "intermediate" header file.
From what I understood, the code added in this PR is exclusively used in scale.c, so `LIBYUV_VERSION` and such should be defined directly in scale.c.

src/scale/avif_scale_internal.h
y-guyon, line 1: `avif_scale_internal.h` looks like multiple libyuv files combined into one. This is also the case for `scale.c` and `scale_impl.c`. This makes it difficult to verify that it matches the original sources, and future libyuv updates will need more work to apply to libavif. On top of that, the year in the copyright boilerplate does not match all the original libyuv source files that were used. I suggest either:

1. Import the minimum set of libyuv files **as is**, keeping the original source tree hierarchy, into a well-labeled folder in libavif. This is the simplest option I see for license/copyright questions and for integrating upstream libyuv updates in the future. The downside is the potential large dead code pulled into libavif.
2. Same as 1. but with unused parts removed from files. The transformation should be thoroughly documented (which parts are kept, which parts are removed, edits, how includes were fixed etc.).
3. Assuming the number of lines to import from libyuv to handle all or most of `avifImageScale()` use cases is reasonable (say fewer than 500 line additions), copy-paste only the minimum parts of `ScalePlane_12()` and  `ScalePlane_16()` directly to src/avif.c. Looking at the size of this PR, I fear this may not be reasonable. On top of that, #709 says a Google employee should do that to avoid license/copyright complications. Otherwise, the file header license/copyright disclaimer and the LICENSE content will be messy. Did I understand that correctly @wantehchang?

src/scale/scale.c
```diff
@@ -0,0 +1,1541 @@
...
+ */
+
+#include <assert.h>
+#include <string.h>
+#include <stdint.h>
+
+#include "avif_scale_internal.h"
+
+#ifdef __cplusplus
+namespace avif_yuvscale {
```
y-guyon, line 37: Are C++ features necessary? Otherwise it would be good to stick to C, as used in all of libavif core sources.

src/scale/scale.c
y-guyon, line 1: `scale.c` already exists at `src/scale.c`. There is a file name unicity tradition in libavif.

y-guyon (changes requested):
Thank you for the contribution. Please document how the added files were imported from libyuv so that reviewers can compare with current libyuv sources rather than studying 4000 new lines. If the import process is too complex, please consider other approaches, such as the ones described in my other comment.

src/scale/scale.c
Cacodemon345: This was a leftover from original files.

tongyuantongyu:
The whole implementation of scaling in libyuv is kind of messy, and contains many code that for libavif's usage is unnecessary. Here are some suggestions:
- Remove all the `ScaleRowDown*` specializations. libavif's usage is always scaling up.
- Remove all `ScaleARGB*` and `ScaleUV*` functions. libavif only deals with planar input.

Optional further debloat (@y-guyon could you check if these are reasonable?):
- Remove the `ScalePlaneUp2*` specializations (and `ScalePlane_12`). As a backup implementation we probably don't care about performance very much.
- Only keep `ScalePlaneVertical*` and `ScalePlaneBilinearUp*`. We only uses `kFilterBox` (which reduces to `kFilterBilinear` for scale ratio > 0.5) now. We probably won't use `kFilterNone`due to quality issues, and `kFilterBox` only do scale down.

Alternatively, maybe we can porting the rescaler from `libwebp` instead. The code is cleaner, despite not so optimized for special scale ratios and only support Bilinear scaling. Related files are:

- API header: https://github.com/webmproject/libwebp/blob/main/src/utils/rescaler_utils.h
- API implementation: https://github.com/webmproject/libwebp/blob/main/src/utils/rescaler_utils.c
- Scale kernel: https://github.com/webmproject/libwebp/blob/main/src/dsp/rescaler.c

Without SIMD optimization and incremental scaling support the implementation should be able to further condensed to under 500 lines.

y-guyon:
> @y-guyon could you check if these are reasonable?

I would say so. Anyway to me the most important points are, in decreasing order: faithful license+copyright import, maintenance easiness, absence of dead code, size of imported code, behavior similarity with libyuv, performance. So which parts are kept/removed matter less than a clear import process.

> we can porting the rescaler from libwebp instead

Is that compatible with bit depths higher than 8? Also I fear that the scaled samples would be significantly different from libyuv.

---

I am starting to wonder whether just asking users to install/build the libyuv dependency is more reasonable.

tongyuantongyu:
> Is that compatible with bit depths higher than 8?

Ah, forgot that. After check the implementation uses 32bit accumulator and 32.32 fixed point arithmetic, so the only missing part is 16bit reading/writing.

src/scale/avif_scale.h
Cacodemon345: FilterMode enumeration is used inside the source files. Not feasible unless merging this with `avif_scale_internal.h`.

src/scale/avif_scale_internal.h
wantehchang: Yannis: What I meant is that if we copy libyuv code that was written by Google employees, then Google owns the copyright of that code and Google can contribute that code to libavif under libavif's license. (When done this way, the copied code can be freely mixed with libavif code.)

Although libyuv was mostly written by Frank Barchard as a Google employee, it has many other contributors. So it could be a lot of work to find a version of code that was written by Google employees only. If the libyuv license is fine (I think it is a commonly used variant of the BSD license), then the current approach is likely to be less work.

y-guyon:
> Renamed files with a prefix to indicate the corresponding libyuv file from which those originate.

As James suggested in https://github.com/AOMediaCodec/libavif/pull/1534#discussion_r1311968273, I think it is best to use a traditional "third_party" folder. What do you think of "libavif/third_party/libyuv"? I guess "libyuv_" file prefix is enough, no need to convey the meaning directly into the path words that not all libyuv files were imported. However this should be documented somewhere.

> Also cleaned up unneeded functions as per suggestion.

Thanks. Could there be precise instructions on how the files were imported, which parts were discarded etc.? It would make it easier to review by comparing with original sources. What do you think of a "libavif/third_party/libyuv/README.md"? See https://github.com/AOMediaCodec/libavif/blob/main/contrib/README.md for a similar example.

Cacodemon345:
This just adds a partial copy of the libyuv files; if this imported the entire libyuv distribution, it would make sense to put these files into `third_party`, but this only imports the parts needed for libavif's scaling functionality, so a `third_party` folder would be potentially misleading, as well as removing the `mini` part, since that signifies that these are not the full copies.

And as for precise instructions, I'm not sure those will be ultimately worth the time and effort. But the general instructions would be:
1. Keep only `ScalePlaneVertical`, `CopyPlane`, `ScalePlaneBox`, `ScalePlaneUp2_Linear`, `ScalePlaneUp2_Bilinear`, `ScalePlaneBilinearUp`, `ScalePlaneBilinearDown` and `ScalePlaneSimple` paths (and their `_16` equivalents) from libyuv.
2. `ScalePlane_12` simply calls `ScalePlane_16`, and has no specializations.
3. Thoroughly strip out anything unused and truly unneeeded by any of those functions.
4. SIMD paths are to be stripped out completely.
5. The names of the files after `libyuv_mini_` are the original filenames of libyuv source code that got imported with the prefix stripped. The headers are imported from `include/libyuv/` directory and the sources including them are imported from `source` directory in libyuv source code.
6. Update the commit hash of libyuv from where the files got imported in CMakeLists.txt.
7. `LIBYUV_API` must be removed from any and all imported functions since those do not belong in a bundled copy.

These should be enough to import libyuv into libavif, with unneeded stuff stripped of.

jzern:
> [...] but this only imports the parts needed for libavif's scaling functionality, so a `third_party` folder would be potentially misleading [...]

Having a third_party folder isolates the files based on their origin and license. The files can then be modified, subject to the license. The best approach is to bring in a pristine copy of what's needed, then follow up with a change that removes the extra parts or defines the code away. The latter would make it simpler to update the files if the upstream changes.

Cacodemon345:
Moved them to `third_party/libyuv/`.

Cacodemon345:
Added libyuv import guide as well.

third_party/libyuv/source/scale.c
```diff
@@ -0,0 +1,967 @@
...
+      // later.
+    } else if (src_height > 1 && dst_height > 1) {
+      dy = FixedDiv1(src_height, dst_height);
+    }
+    // Arbitrary scale vertically, but unscaled horizontally.
+    ScalePlaneVertical_16(src_height, dst_width, dst_height, src_stride,
+                          dst_stride, src, dst, 0, y, dy, /*bpp=*/1, filtering);
+    return;
+  }
+
```
y-guyon, line 925: The case `if (dst_width <= Abs(src_width) && dst_height <= src_height) {` is missing. In https://github.com/AOMediaCodec/libavif/pull/1534#issuecomment-1701225017 it is said that "libavif's usage is always scaling up". Please either:
- Return false in `avifImageScale()` when downscaling without the libyuv dependency, or
- Keep `ScalePlane_16()` as is, even if more code needs to be imported.

third_party/libyuv/source/scale_common.c
```diff
@@ -0,0 +1,631 @@
...
+    InterpolateRow(dst_argb, src_argb + yi * src_stride, src_stride,
+                   dst_width_words, yf);
+    dst_argb += dst_stride;
+    y += dy;
+  }
+}
+
+#define C16TO8(v, scale) clamp255(((v) * (scale)) >> 16)
+
+void Convert16To8Row_C(const uint16_t* src_y,
```
y-guyon, line 448: `Convert16To8Row_C()` is not part of [libyuv/source/scale_common.cc](https://chromium.googlesource.com/libyuv/libyuv/+/c60323de1756f489bd061e66aea9c3c73a8ef72c/source/scale_common.cc)

third_party/libyuv/README.md
```diff
@@ -0,0 +1,13 @@
+# libyuv
+This folder contains source code imported from libyuv as of `c60323de1756f489bd061e66aea9c3c73a8ef72c`, with modifications intended to keep them relatively small and simple.
```
y-guyon, line 2: Please also add an entry in CHANGELOG.md with something similar to
```
### Added

* Implement avifImageScale() fallback when libyuv is not available.
* Partial import of libyuv to third_party/libyuv (new LICENSE).
```

y-guyon (changes requested):
I checked that most changes are reasonable but the imported code is still different enough from libyuv to make it non-trivial to review.

third_party/libyuv/source/scale.c
Cacodemon345: Scaling down should still work properly as it is now without any changes; that case isn't critical for scaling down as it only deals with optimized paths, and those aren't important for a backup implementation.

third_party/libyuv/README.md
```diff
@@ -0,0 +1,13 @@
...
+# Importing from upstream
+When importing source code from upstream libyuv, the following changes must be done:
+1. Source hierachy is to be kept the same as in libyuv.
+2. The only APIs that are called by libavif for scaling are `ScalePlane` and `ScalePlane_12`. Anything else must be left out.
+3. In function `ScalePlane` and `ScalePlane_16`, only the `ScalePlaneVertical`, `CopyPlane`, `ScalePlaneBox`, `ScalePlaneUp2_Linear`, `ScalePlaneUp2_Bilinear`, `ScalePlaneBilinearUp`, `ScalePlaneBilinearDown` and `ScalePlaneSimple` paths (and their `_16` equivalents) from libyuv must be kept; any other paths are to be stripped out.
+4. `ScalePlane_12` simply calls `ScalePlane_16`, and is to have no special paths.
+5. Anything unused and truly unneeeded by any of those functions must be thoroughly stripped out.
+6. SIMD paths are to be stripped out completely.
+7. The commit hash of libyuv from where the files got imported in this README.md file must be updated if possible.
+8. `LIBYUV_API` must be removed from any and all imported functions as these files are always built static.
```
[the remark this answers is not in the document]
Cacodemon345: It's already handled by `basic_types.h` include.

CMakeLists.txt
```diff
@@ -525,6 +534,9 @@ target_include_directories(
     avif PUBLIC $<BUILD_INTERFACE:${libavif_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include> PRIVATE ${AVIF_PLATFORM_INCLUDES}
                                                                                                       ${AVIF_CODEC_INCLUDES}
 )
+if(NOT libyuv_FOUND)
+    target_include_directories(avif PRIVATE ${libavif_SOURCE_DIR}/third_party/libyuv/include/)
```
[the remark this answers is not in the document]
Cacodemon345: Done.

third_party/README.md
```diff
@@ -0,0 +1,5 @@
+# libavif third party sources
+
+Anything in this directory is maintained external to libavif by third-parties.
```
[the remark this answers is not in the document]
Cacodemon345: Done.

Cacodemon345:
Most remaining problems were handled, except for https://github.com/AOMediaCodec/libavif/pull/1534#discussion_r1314656128, see reply.

y-guyon (commented):
If you do not like the `#if 0` approach as suggested by jzern in https://github.com/AOMediaCodec/libavif/pull/1534#issuecomment-1703239675, we can still leverage GitHub PR history by having one commit in this PR that imports pristine copies of `basic_types.h`, `row.h`, `scale.c` etc. and another commit that removes the unused parts in these files. The GitHub UI will make it clear that almost nothing was modified, for future reference and easiness of review.  
Please do so if you feel comfortable with rewriting pushed commits. I see you already force pushed a few times.

This PR will be squashed before being merged into the main branch, so splitting the import and reduce steps will not pollute the main git history.

> Most remaining problems were handled, except for https://github.com/AOMediaCodec/libavif/pull/1534#discussion_r1314656128, see reply.

Please re-request review when ready.

Cacodemon345:
Did the import and reduce step (sans the renaming part; git had trouble with modified renamed files and made them `deleted` and `new file`).

y-guyon:
> Did the import and reduce step (sans the renaming part; git had trouble with modified renamed files and made them `deleted` and `new file`).

Thank you for taking the time to do that. Sorry for not being clear enough; what I meant was:

- Discard previous commits (but keeping them is fine for comment history so no worries)
- Add one commit importing only the necessary files from libyuv (libyuv.h, basic_types.h,  planar_functions.h, row.h, scale.h, scale_row.h, version.h, row_common.cc, scale_any.cc, scale.cc, scale_common.cc)
- Add one commit renaming *.cc to *.c, so that GitHub UI and git easily understand the renaming step
- Remove lines from imported files

Because as you said, currently there is no diff, just delete+new.

Cacodemon345:
Is this fine? I did it like you said.

third_party/libyuv/include/libyuv/planar_functions.h
```diff
@@ -13,1139 +13,18 @@
...
-                int width,
-                int height);
-
-#ifdef __cplusplus
-}  // extern "C"
-}  // namespace libyuv
-#endif
-
-#endif  // INCLUDE_LIBYUV_PLANAR_FUNCTIONS_H_
+#endif  // INCLUDE_LIBYUV_PLANAR_FUNCTIONS_H_
```
y-guyon, line 30: Missing new line character at end of file.
Other files too.

third_party/libyuv/include/libyuv/scale_row.h
```diff
@@ -266,22 +50,11 @@ enum FilterMode ScaleFilterReduce(int src_width,
...
-int FixedDiv1_X86(int num, int div);
-int FixedDiv1_MIPS(int num, int div);
-#ifdef HAS_FIXEDDIV_X86
-#define FixedDiv FixedDiv_X86
-#define FixedDiv1 FixedDiv1_X86
-#elif defined HAS_FIXEDDIV_MIPS
-#define FixedDiv FixedDiv_MIPS
-#define FixedDiv1 FixedDiv1_MIPS
-#else
+
```
y-guyon, line 55: Super nit: extra blank line

third_party/libyuv/source/row_common.c
```diff
@@ -14,3544 +14,32 @@
...
-  int x;
-  for (x = 0; x < width; x += 2) {
-    dst_u[0] = src_uyvy[0];
-    dst_v[0] = src_uyvy[2];
-    src_uyvy += 4;
-    dst_u += 1;
-    dst_v += 1;
-  }
-}
+#define STATIC_CAST(type, expr) (type)(expr)
```
y-guyon, line 18: I am disappointed by GitHub diff UI, it fails to consider this file as only lines removed. Local diff tool called Meld works.

third_party/libyuv/source/scale_any.c
```diff
@@ -812,267 +53,33 @@ SU2BLANY(ScaleRowUp2_Bilinear_16_Any_C,
...
+    int n = work_width & ~MASK;                                    \
+    dst_ptr[0] = src_ptr[0];                                       \
+    if (work_width > 0) {                                          \
+      if (n != 0) {                                                \
+        SIMD(src_ptr, dst_ptr + 1, n);                             \
+      }                                                            \
+      C(src_ptr + (n / 2), dst_ptr + n + 1, r);                    \
+    }                                                              \
+    dst_ptr[dst_width - 1] = src_ptr[(dst_width - 1) / 2];         \
+  }
```
y-guyon, line 70: Please keep the original order. `SUH2LANY()` should be above `SU2BLANY()`.

third_party/libyuv/source/scale_common.c
```diff
@@ -13,556 +13,23 @@
...
-namespace libyuv {
-extern "C" {
-#endif
 
-#ifdef __cplusplus
-#define STATIC_CAST(type, expr) static_cast<type>(expr)
-#else
 #define STATIC_CAST(type, expr) (type)(expr)
-#endif
+
```
y-guyon, line 22: Extra blank line.

third_party/libyuv/source/scale_common.c
```diff
@@ -1964,7 +565,54 @@ void ScaleSlope(int src_width,
...
+void CopyPlane_16(const uint16_t* src_y,
+                  int src_stride_y,
+                  uint16_t* dst_y,
+                  int dst_stride_y,
+                  int width,
+                  int height) {
+  CopyPlane((const uint8_t*)src_y, src_stride_y * 2, (uint8_t*)dst_y,
+            dst_stride_y * 2, width * 2, height);
+}
+
```
y-guyon, line 618: Please keep that in the original file (planar_functions.c)

third_party/libyuv/source/scale_common.c
```diff
@@ -1964,7 +565,54 @@ void ScaleSlope(int src_width,
 }
 #undef CENTERSTART
 
-#ifdef __cplusplus
-}  // extern "C"
-}  // namespace libyuv
-#endif
+void CopyRow_C(const uint8_t* src, uint8_t* dst, int count) {
+  memcpy(dst, src, count);
+}
```
y-guyon, line 570: Please keep that in the original file (row_common.c)

y-guyon (changes requested):
I verified that commit 9c107f3 in this PR is a subset of pristine files at commit [c60323d](https://chromium.googlesource.com/libyuv/libyuv/+/c60323de1756f489bd061e66aea9c3c73a8ef72c) in libyuv.  
Commit cd93d66 in this PR is just files renamed without changes.  
I verified that commit 3347ba4 in this PR:
- removes whole functions in row_common.c and scale_any.c
- removes whole functions and SIMD define-guarded lines in scale.c, except in `ScalePlane()` and `ScalePlane_16()` where optimized paths `ScalePlaneDown*()` are skipped (as described in third_party/README.md)
- removes whole functions and SIMD define-guarded lines in scale_common.c, except for `CopyPlane()` and `CopyPlane_16()` which are imported from planar_functions.cc and `CopyRow_C()` which is imported from row_common.cc

> Is this fine? I did it like you said.

Yes, thank you. It made the PR easier to review.  
Once the minor changes I requested are fixed, this PR should be good to be merged.

y-guyon (changes requested):
I verified the following in this PR, regarding files in third_party/libyuv compared to libyuv at commit [c60323d](https://chromium.googlesource.com/libyuv/libyuv/+/c60323de1756f489bd061e66aea9c3c73a8ef72c):
- in planar_functions.c and scale_common.c, only whole functions and SIMD define-guarded code blocks are removed,
- in scale.c, only whole functions and SIMD define-guarded code blocks are removed, except for the removed calls to `ScalePlaneDown*()` in `ScalePlane()` and `ScalePlane_16()`,
- in planar_functions.h, row.h, scale.h, scale_row.h, row_common.c and scale_any.c, only whole functions are removed,
- in libyuv.h, only includes are removed,
- basic_types.h and version.h are kept as is.

Some blank lines differences remain but it is fine.

I verified that avifscaletest succeeds locally with `-DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON` and "libyuv-based fast paths disabled". Other unrelated tests fail, they should probably be disabled in this case (in another PR).

third_party/libyuv/source/row_common.c
```diff
@@ -0,0 +1,131 @@
...
+                       int scale,
+                       int width) {
+  int x;
+  assert(scale >= 256);
+  assert(scale <= 32768);
+
+  for (x = 0; x < width; ++x) {
+    dst_y[x] = STATIC_CAST(uint8_t, C16TO8(src_y[x], scale));
+  }
+}
```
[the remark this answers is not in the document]
Cacodemon345: Done.

third_party/libyuv/source/scale.c
```diff
@@ -0,0 +1,990 @@
...
+                                   int dst_stride,
+                                   const uint8_t* src_ptr,
+                                   uint8_t* dst_ptr) {
+  void (*Scale2RowUp)(const uint8_t* src_ptr, ptrdiff_t src_stride,
+                      uint8_t* dst_ptr, ptrdiff_t dst_stride, int dst_width) =
+      ScaleRowUp2_Bilinear_Any_C;
+  int x;
+
+  // This function can only scale up by 2 times.
+  assert(src_width == ((dst_width + 1) / 2));
```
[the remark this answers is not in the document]
vigneshvg: > Remove `-Wunused-parameter` in CI for `build-android-jni`

This should be okay.

> Add libyuv dependency in CI for build-android-jni

I think this would be ideal tihng to do. We generally build android JNI code with libyuv anyways.

I don't have a strong preference and either is fine. Whichever way we go, it would also have to be done for `build-and-run-android-jni-tests`. That workflow is also likely to fail with this PR. It is not running on this PR since it is time consuming and only runs on PRs which touch the `android_jni` directory.

third_party/libyuv/source/scale_common.c
```diff
@@ -0,0 +1,563 @@
...
+#include <string.h>
+
+#include "libyuv/planar_functions.h"  // For CopyARGB
+#include "libyuv/row.h"
+#include "libyuv/scale_row.h"
+
+#define STATIC_CAST(type, expr) (type)(expr)
+
+// TODO(fbarchard): make clamp255 preserve negative values.
+static __inline int32_t clamp255(int32_t v) {
```
vigneshvg, line 23: This function is unused.

third_party/libyuv/source/scale.c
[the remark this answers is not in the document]
vigneshvg: Applying the following patch fixes the Android CI failure:

```
index 369b734..a94663b 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -321,6 +321,13 @@ if(NOT libyuv_FOUND)
         third_party/libyuv/source/row_common.c
         third_party/libyuv/source/planar_functions.c
     )
+    if(DEFINED ANDROID_ABI)
+        # When building third_party/libyuv/source/scale.c, some functions use
+        # some of the parameters only inside an assert statement. This causes
+        # unused parameter warnings when building for Android. Suppress the
+        # warning in that case.
+        add_compile_options(-Wno-unused-parameter)
+    endif()
 endif()

 if(AVIF_CODEC_DAV1D)
```

I didn't want to force push a commit onto this branch, so just leaving it as a comment here.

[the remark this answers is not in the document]
y-guyon: @Cacodemon345 Please apply vigneshvg's patch in a new commit in this PR. We can trigger the CI workflows again afterwards.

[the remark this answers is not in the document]
Cacodemon345: Done.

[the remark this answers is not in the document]
vigneshvg: Thanks, i have triggered the CI workflows on the latest version of this PR.

y-guyon (approved):

y-guyon:
(@Cacodemon345  Please rebase this PR. Rebasing is preferred but you can also merge recent head changes into this PR if you are more comfortable doing so.)

Cacodemon345:
@y-guyon I went for merging instead, since rebase at this stage is likely to induce too many conflicts.

jzern:
The top-level LICENSE file should be updated to include this new path.

y-guyon:
> The top-level LICENSE file should be updated to include this new path.

Done in https://github.com/AOMediaCodec/libavif/pull/1568.  
I thought third_party/libyuv was similar enough to ext/libyuv, which is not referenced in the top-level LICENSE file.

PR #1543, opened 2023-08-31
Add tests/oss-fuzz/repro_fuzz.cc

y-guyon:
For easy reproduction of oss-fuzz issues.


vrabaud (approved):

tests/CMakeLists.txt
```diff
@@ -47,6 +47,14 @@ if(AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
     target_link_libraries(aviftest_helpers avif_internal)
 endif()
 
+# Fuzz target without the fuzzing engine. For easy reproduction of oss-fuzz issues.
```
wantehchang, line 50: This comment seems to imply that it is easier to reproduce oss-fuzz issues without the fuzzing engine. But I suspect what you meant is that this CMakeLists.txt file have a fuzz target with the fuzzing engine, so we provide a target without the fuzzing engine as a substitute.

Also, I'm afraid that I misunderstood the purpose of repro_fuzz. The Google-internal version of repro_fuzz has additional command-line options available, e.g., for using Google's performance profiling tools. I thought that was the main reason for adding repro_fuzz. Without the additional command-line options, is repro_fuzz still more useful than avif_decode_fuzzer? Perhaps it's easier to run repro_fuzz in the debugger because it is compiled in the normal way?

wantehchang (commented):
Yannis: I just wanted to make sure that repro_fuzz is useful to people outside Google, and the comments describe its purpose (and how to use it, if not obvious) clearly. Thanks.

tests/CMakeLists.txt
```diff
@@ -47,6 +47,14 @@ if(AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
...
+add_executable(avifreprofuzz oss-fuzz/avif_decode_fuzzer.cc oss-fuzz/repro_fuzz.cc)
```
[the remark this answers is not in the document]
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/1548.

tests/CMakeLists.txt
y-guyon: > This comment seems to imply that it is easier to reproduce oss-fuzz issues without the fuzzing engine.

Yes, this is my opinion. One does not have to look up the oss-fuzz procedure, just running this target with the reproducer file is enough. No need for a dependency or to think about it.

> But I suspect what you meant is that this CMakeLists.txt file have a fuzz target with the fuzzing engine, so we provide a target without the fuzzing engine as a substitute.

This CMakeLists.txt file does not have a fuzz target with the fuzzing engine, so I wonder why you had this suspicion. `oss-fuzz/avif_decode_fuzzer.cc` can be considered dead code now in libavif without oss-fuzz, so I thought it would be good to include it in the CI.

> The Google-internal version of repro_fuzz has additional command-line options available, e.g., for using Google's performance profiling tools.

Correct.

> I thought that was the main reason for adding repro_fuzz. Without the additional command-line options, is repro_fuzz still more useful than avif_decode_fuzzer?

How do you use `avif_decode_fuzzer` as of today?

> Perhaps it's easier to run repro_fuzz in the debugger because it is compiled in the normal way?

This is also the case, yes.

wantehchang: Yannis: You are right that it is difficult to build avif_decode_fuzzer. Fortunately Google employees have a workaround: we can build avif_decode_fuzzer easily in Google's internal build system for libavif. This is why I asked about the value of repro_fuzz when you originally added it to Google's internal repository. During that code review I thought its value was mainly the additional command-line options, e.g., for using Google's performance profiling tools.

In libavif's cmake build system, I think the value of repro_fuzz is a simple way to build avif_decoder_fuzzer. Although repro_fuzz doesn't contain the fuzzing engine, we often don't need the fuzzing engine to reproduce a fuzzing bug. Note that we do need to build repro_fuzz with ASan, UBSan, or MSan if the fuzzing bug is a sanitizer warning.

Do I understand the value of repro_fuzz (both inside and outside Google) correctly?

Thanks.

wantehchang: A bit of history: avif_decode_fuzzer.cc was originally in the oss-fuzz source tree. It was a hassle to update it whenever there was a libavif API change that broke the compilation of avif_decode_fuzzer.cc. So we moved avif_decode_fuzzer.cc to the libavif source tree.

y-guyon: > Do I understand the value of repro_fuzz (both inside and outside Google) correctly?

Yes.

> A bit of history: avif_decode_fuzzer.cc was originally in the oss-fuzz source tree. It was a hassle to update it whenever there was a libavif API change that broke the compilation of avif_decode_fuzzer.cc. So we moved avif_decode_fuzzer.cc to the libavif source tree.

This is the layout recommended by oss-fuzz.

PR #1545, opened 2023-08-31
Fix bug with unhandled 'tmap' items as primary image or color.

maryla-uc:
Fixes fuzzer bug https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=61889


src/read.c
```diff
@@ -4107,6 +4106,10 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                 avifDiagnosticsPrintf(data->diag, "Box[tmap] gain map item ID %d not found", gainMapItemID);
                 return AVIF_RESULT_INVALID_TONE_MAPPED_IMAGE;
             }
+            if (avifDecoderItemShouldBeSkipped(gainMapItem)) {
+                avifDiagnosticsPrintf(data->diag, "Box[tmap] gain map item %d is not a supported image type", gainMapItemID);
+                return AVIF_RESULT_INVALID_TONE_MAPPED_IMAGE;
+            }
```
y-guyon, line 4112: Nit: An error is returned but the function is called `avifDecoderItemShouldBeSkipped()`. Maybe we should consider changing the behavior or renaming the function in the future but this is fine for now.

y-guyon (approved):

src/read.c
maryla-uc: Good point, will fix in another PR.

wantehchang (commented):
The bug fix is to revert the previous change to the `avifDecoderItemShouldBeSkipped()` function. That part LGTM. Thanks.

I didn't review the changes inside `#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)` because I am not familiar with the new gain map code.

PR #1547, opened 2023-08-31
Put more code behind AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP.

maryla-uc:
And adress some more minor comments from #1501.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -502,6 +502,7 @@ typedef struct avifContentLightLevelInformationBox
     uint16_t maxPALL;
 } avifContentLightLevelInformationBox;
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
 // ---------------------------------------------------------------------------
 // avifGainMap
 // Gain Maps are a HIGHLY EXPERIMENTAL FEATURE. The format might still change and
@@ -510,7 +511,6 @@ typedef struct avifContentLightLevelInformationBox
 // This is based on ISO/IEC JTC 1/SC 29/WG 3 m64379
 // This product includes Gain Map technology under license by Adobe.
 
-#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
 struct avifImage;
 
 // Gain map metadata, for tone mapping between SDR and HDR.
@@ -665,9 +665,11 @@ typedef struct avifImage
 // avifImageCreate() and avifImageCreateEmpty() return NULL if arguments are invalid or if a memory allocation failed.
 AVIF_API avifImage * avifImageCreate(uint32_t width, uint32_t height, uint32_t depth, avifPixelFormat yuvFormat);
 AVIF_API avifImage * avifImageCreateEmpty(void); // helper for making an image to decode into
-// Performs a deep copy of an image, including all metadata and planes, and the gain map metadata/planes if present.
+// Performs a deep copy of an image, including all metadata and planes, and the gain map metadata/planes if present
+// and AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP is defined.
 AVIF_API avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes);
-// Performs a shallow copy of a rectangular area of an image. 'dstImage' does not own the planes. The gain map if any is ignored.
+// Performs a shallow copy of a rectangular area of an image. 'dstImage' does not own the planes.
+// Ignores the gainMap field (which exists only if AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP is defined).
 AVIF_API avifResult avifImageSetViewRect(avifImage * dstImage, const avifImage * srcImage, const avifCropRect * rect);
 AVIF_API void avifImageDestroy(avifImage * image);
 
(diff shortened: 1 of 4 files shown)


include/avif/avif.h
```diff
@@ -665,9 +665,11 @@ typedef struct avifImage
 // avifImageCreate() and avifImageCreateEmpty() return NULL if arguments are invalid or if a memory allocation failed.
 AVIF_API avifImage * avifImageCreate(uint32_t width, uint32_t height, uint32_t depth, avifPixelFormat yuvFormat);
 AVIF_API avifImage * avifImageCreateEmpty(void); // helper for making an image to decode into
-// Performs a deep copy of an image, including all metadata and planes, and the gain map metadata/planes if present.
+// Performs a deep copy of an image, including all metadata and planes, and the gain map metadata/planes if present
+// and AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP is defined.
```
wantehchang, line 669: Nit/optional: The following part of this sentence could be edited:

> and the gain map metadata/planes if present and AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP is defined

I don't have a good suggestion. I think we should at least add an "if":

> and the gain map metadata/planes if present and if AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP is defined

Sorry about the nonconstructive criticism.

src/read.c
```diff
@@ -874,15 +878,18 @@ static void avifDecoderDataResetCodec(avifDecoderData * data)
         }
         if (tile->codec) {
             // Check if tile->codec was created separately and destroy it in that case.
-            if (tile->codec != data->codec && tile->codec != data->codecAlpha && tile->codec != data->codecGainMap) {
+            avifBool isOwnedCodec = tile->codec != data->codec && tile->codec != data->codecAlpha;
```
wantehchang, line 881: Nit/optional: isOwnedCodec => ownsCodec or tileOwnsCodec ?

src/read.c
```diff
@@ -942,7 +952,11 @@ static void avifDecoderDataClearTiles(avifDecoderData * data)
...
```
wantehchang, line 955: Nit/optional: isOwnedCodec => ownsCodec or tileOwnsCodec ?

src/read.c
```diff
@@ -3599,7 +3615,11 @@ static avifResult avifCodecCreateInternal(avifCodecChoice choice, const avifTile
 
 static avifBool avifTilesCanBeDecodedWithSameCodecInstance(avifDecoderData * data)
 {
-    if (data->color.tileCount == 1 && (data->alpha.tileCount == 1 || data->gainMap.tileCount == 1)) {
+    avifBool hasSingleTileAlphaOrGainMap = data->alpha.tileCount == 1;
```
wantehchang, line 3618: Nit/optional: Another nonconstructive criticism: this varibale name is very long, but I don't have a good suggestion.

src/read.c
```diff
@@ -3599,7 +3615,11 @@ static avifResult avifCodecCreateInternal(avifCodecChoice choice, const avifTile
...
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+    hasSingleTileAlphaOrGainMap = hasSingleTileAlphaOrGainMap || data->gainMap.tileCount == 1;
+#endif
+    if (data->color.tileCount == 1 && hasSingleTileAlphaOrGainMap) {
```
wantehchang, line 3622: I suggest two changes to the comments below.

Line 3623: alpha or gain map plane => alpha plane or gain map

Line 3624: alpha (or gain map) plane => alpha plane (or gain map)

src/read.c
```diff
@@ -4538,6 +4566,16 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder, uint32_t nextIma
     return AVIF_RESULT_OK;
 }
 
+static avifBool avifAWholeFrameWasDecoded(const avifDecoder * decoder)
```
wantehchang, line 4569: This function can take `const avifDecoderData * data` as input.

wantehchang, line 4569: Nit/optional: libavif often uses an object-oriented naming convention for functions. If a function's main parameter is `avifFoo`, then the function is often named `avifFooDoSomething`.

Under that naming convention, this function would be named something like `avifDecoderDataFullyDecoded` or `avifDecoderDataCompletelyDecoded`

src/read.c
```diff
@@ -4538,6 +4566,16 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder, uint32_t nextIma
...
+{
+    avifBool aWholeFrameWasDecoded = (decoder->data->color.decodedTileCount == decoder->data->color.tileCount) &&
```
wantehchang, line 4571: Nit: Shorten this local variable name, e.g., `fullyDecoded`.

wantehchang (approved):
LGTM. Thanks. I suggest a few tweaks.

src/read.c
maryla-uc: I see what you mean but I think it's not _that_ long (it's only 2 chars longer than 'canUseSingleCodecInstance' a bit lower down) and I'd rather keep it descriptive.

wantehchang: If the gain map is also an auxiliary image, then a good, slightly shorter name is `hasSingleTileAuxImage`.

The current name is also fine.

src/read.c
```diff
@@ -4538,6 +4566,16 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder, uint32_t nextIma
     return AVIF_RESULT_OK;
 }
 
+static avifBool avifDecoderDataLastFrameFullyDecoded(const avifDecoderData * data)
```
wantehchang, line 4569: Nit: Last => Current ?

If "Current" is incorrect, then we may want to change "Last" to "Previous" or "Prev". "Last" can be ambiguous because it can mean "the one before the current one" or "final".

wantehchang (approved):
LGTM.

src/read.c
maryla-uc: No the gain map is not an auxiliary image, it's a derived item with a hidden input...

src/read.c
maryla-uc: Renamed to avifDecoderDataFrameFullyDecoded because indeed it's a bit complicated which frame we're talking about. Also added a common: "Returns AVIF_FALSE if there is currently a partially decoded frame."
Ideally I would prefer a comment of the type "Returns AVIF_TRUE if..." but I thought it was simpler to explain this way.

PR #1550, opened 2023-09-01
BLD: Find rav1e shared library on Windows

carterbox:
When building rav1e as a shared library on Windows, the lib file is named `ravie.dll.lib` instead of `rav1e.lib`, so in order to find it successfully, we need to search for that name too.


cmake/Modules/Findrav1e.cmake
```diff
@@ -32,7 +32,7 @@ if(NOT RAV1E_INCLUDE_DIR)
 endif()
 
 if(NOT RAV1E_LIBRARY)
-    find_library(RAV1E_LIBRARY NAMES rav1e PATHS ${_RAV1E_LIBDIR})
+    find_library(RAV1E_LIBRARY NAMES rav1e rav1e.dll PATHS ${_RAV1E_LIBDIR})
```
wantehchang, line 35: Thank you for the patch. Could you add a short comment here to explain why we need rav1e.dll here (because the import library of rav1e.dll is named rav1e.dll.lib)?

y-guyon: Thank you for adding the comment. Could you move it inside the `if(NOT RAV1E_LIBRARY)` condition?

```diff
-# Search for both rav1e and rav1e.dll because when building shared libraries
-# for Windows, cargo-c names the libraries rav1e.dll and ravie.dll.lib
if(NOT RAV1E_LIBRARY)
+    # Search for both rav1e and rav1e.dll because when building shared libraries
+    # for Windows, cargo-c names the libraries rav1e.dll and ravie.dll.lib
    find_library(RAV1E_LIBRARY NAMES rav1e rav1e.dll PATHS ${_RAV1E_LIBDIR})
endif()
```

kmilos: This really should not be necessary, the default CMake prefix/suffix search list for Windows works for most toolchains:

https://gitlab.kitware.com/cmake/cmake/-/blob/89038367ab13e65ead7d0bb41feda3020bed9d18/Modules/Platform/Windows.cmake#L16-L17
https://gitlab.kitware.com/cmake/cmake/-/blob/89038367ab13e65ead7d0bb41feda3020bed9d18/Modules/Platform/Windows-GNU.cmake#L26-L27

etc.

I'd rather you filed an upstream bug against cargo-c to output more sane/traditional artifacts, and/or against CMake to support cargo-c specifics as well...

kmilos: Or a better workaround patch for the moment could be to append `.dll.lib` to `CMAKE_FIND_LIBRARY_SUFFIXES` (in case of `WIN32` only) before `find_library()` to make the intent clearer?

carterbox: It seems that the [developer of cargo-c is just following](https://github.com/lu-zero/cargo-c/issues/181#issuecomment-826322415) the [conventions of rustc (cargo)](https://github.com/rust-lang/cargo/blob/a5fa676731c8e570bb9be9c50d2202acea84b6b6/src/cargo/core/compiler/build_context/target_info.rs#L389), so it's probably better to open a PR with CMAKE since many libraries may be producing import libs with the suffix `.dll.lib`.

carterbox: https://gitlab.kitware.com/cmake/cmake/-/issues/25478

carterbox: A fix for this problem was merged in the CMake project. It should be available in the next release. You can either close this PR, or I have modified this PR to match the solution that was accepted in CMake. Thanks for your help!

vrabaud:
We are based on an older CMake so we need your fix anyway. It seems your fix is exactly what has been done upstream at https://gitlab.kitware.com/cmake/cmake/-/commit/f20c5c6f20aa7eab9e43919fffbf5b00281d1461 so we can merge it right?

PR #1553, opened 2023-09-04
Give meaningful names to CI workflows.

maryla-uc:
A meaningful workflow name is useful on the Actions page for example: https://github.com/AOMediaCodec/libavif/actions
On some other pages, when the job name is displayed as well, it can be redundant, but that is a minor issue.

### DIFF
--- a/.github/workflows/ci-unix-static-av2.yml
+++ b/.github/workflows/ci-unix-static-av2.yml
@@ -1,4 +1,4 @@
-name: CI
+name: CI Unix Static AV2
 on: [push, pull_request]
 
 permissions:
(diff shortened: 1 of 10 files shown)


.github/workflows/ci-unix-static-av2.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: CI Unix Static AV2
```
y-guyon, line 1: Note: This is listed twice if the job name is not shown, because of the matrix below. Still better than before.

y-guyon (approved):

.github/workflows/ci-unix-static-av2.yml
maryla-uc: Not sure which page you're referring to, but on the actions page at least, it's listed once, and when you click on it you can see the list of OSs, e.g. https://github.com/AOMediaCodec/libavif/actions/runs/6071494062
In any case there's not much we can do about that.

PR #1554, opened 2023-09-04
Add golden tests for the written AVIF metadata structure.

maryla-uc:
Depends on MP4Box (from the gpac project) to dump an AVIF file's metadata as XML.

This test catches changes in write.c that may or may not be intentional. On failure, it prints instructions on how to update the goldens if needed. The test is disabled by default, but enabled in a new CI workflow.

### DIFF
--- a/tests/test_cmd_enc_boxes_golden.sh
+++ b/tests/test_cmd_enc_boxes_golden.sh
@@ -61,9 +61,7 @@ has_errors=false
 cleanup() {
     # Only delete temp files if the test succeeded, to allow debugging in case or error.
     if ! $has_errors; then
-        find ${OUTPUT_DIR} -name '*.avif' -delete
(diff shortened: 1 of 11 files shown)


tests/test_cmd_enc_boxes_golden.sh
```diff
@@ -0,0 +1,188 @@
...
+mkdir -p "$OUTPUT_DIR"
+echo "Outputting to $OUTPUT_DIR"
+
+has_errors=false
+
+# Cleanup
+cleanup() {
+    # Only delete temp files if the test succeeded, to allow debugging in case or error.
+    if ! $has_errors; then
+        find ${OUTPUT_DIR} -name '*.avif' -delete
```
vrabaud, line 64: find ./ -type f \( -iname \*.avif -o -iname \*.xml -o -iname \*.xml.diff \)

PR #1555, opened 2023-09-05
Ignore gain maps by default.

maryla-uc:
When AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP is enabled, add two decoder options: ignoreGainMap (true by default) and ignoreColorAndAlpha (to allow decoding the gain map only).

The names were chosen to mimic ignoreExif and ignoreXMP. But maybe for gain maps, it would be better to have something that is false by default? Like 'enableGainMaps', 'decodeGainMap', etc.?

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -4661,7 +4661,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         if (decoder->ignoreGainMap) {
             // When ignoring the gain map, we still report whether one is present or not,
             // but do not fail if there was any error with the gain map.
-            decoder->gainMapPresent = (findGainMapResult == AVIF_RESULT_OK) && gainMapItem != NULL;
+            decoder->gainMapPresent = (findGainMapResult == AVIF_RESULT_OK) && (gainMapItem != NULL);
             // We also ignore the actual item and don't decode it.
             gainMapItem = NULL;
         } else {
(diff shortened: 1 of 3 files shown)


tests/gtest/avifgainmaptest.cc
```diff
@@ -351,5 +354,113 @@ TEST(GainMapTest, SequenceNotSupported) {
       << avifResultToString(result) << " " << encoder->diag.error;
 }
 
+TEST(GainMapTest, IgnoreGainMapOrColor) {
```
vrabaud, line 357: Just wondering if that test and the following could be merged as a parametrized test depending on:
ignoreGainMap, ignoreColorAndAlpha and the creation of a gain map.
Not sure if it'd be easier to read and smaller in size.

src/read.c
```diff
@@ -4559,71 +4655,17 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
-                    decoder->image->gainMap.image->clli = clliProp->u.clli;
-                }
-            }
+        avifDecoderItem * gainMapItem;
+        avifCodecType gainMapCodecType;
+        avifResult findGainMapResult = avifDecoderFindGainMapItem(decoder, colorItem, &gainMapItem, &gainMapCodecType);
+        if (decoder->ignoreGainMap) {
+            // When ignoring the gain map, we still report whether one is present or not,
+            // but do not fail if there was any error with the gain map.
+            decoder->gainMapPresent = (findGainMapResult == AVIF_RESULT_OK) && gainMapItem != NULL;
```
vrabaud, line 4664: (gainMapItem != NULL) to be constistent

vrabaud (commented):
Nits

vrabaud (approved):

PR #1558, opened 2023-09-05
Add test for animated image decoding

vigneshvg:


tests/data/README.md
```diff
@@ -362,6 +362,22 @@ Box structure of the items in this file:
 [alpha] [alpha]
 ```
 
+## Animated Images
+
+### File [star-animated-8bpc.avif](star-animated-8bpc.avif)
+
+![](star-animated-8bpc.avif)
+
+Source: [Chromium blink tests](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/images/resources/avif/star-animated-8bpc.avif;l=1;drc=c5a2c8c443872070821e51c713845651121f39fb)
```
y-guyon, line 371: Would it be possible to add the license? I am worried about [this TODO note](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/images/resources/avif/README.md;l=198).

I found [blink/LICENSE_FOR_ABOUT_CREDITS](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/LICENSE_FOR_ABOUT_CREDITS) and [LICENSE](https://source.chromium.org/chromium/chromium/src/+/main:LICENSE) when going up the source tree. Do you know if one of these apply to that file? I guess we can mention both just in case.

tests/gtest/avifdecodetest.cc
```diff
@@ -29,6 +29,26 @@ TEST(AvifDecodeTest, ColorGridAlphaNoGrid) {
...
+            AVIF_RESULT_OK);
+  ASSERT_EQ(avifDecoderParse(decoder.get()), AVIF_RESULT_OK);
+  EXPECT_EQ(decoder->alphaPresent, AVIF_FALSE);
+  EXPECT_EQ(decoder->imageCount, 5);
+  EXPECT_EQ(decoder->repetitionCount, 0);
+  for (int i = 0; i < 5; ++i) {
+    EXPECT_EQ(avifDecoderNextImage(decoder.get()), AVIF_RESULT_OK);
+  }
+}
+
```
y-guyon, line 51: Could we move that to a new avifanimtest.cc file? I like how simple avifdecodetest.cc is, and it is convenient to copy it for creating new tests. When it fails, it is also handy to have a single test case for reproduction.  
I suggest duplicating avifdecodetest.cc in a first commit and adapting it to animations in a second one, to show the minimal changes between the two.

tests/gtest/avifdecodetest.cc
```diff
@@ -29,6 +29,26 @@ TEST(AvifDecodeTest, ColorGridAlphaNoGrid) {
   EXPECT_GT(decoder->image->alphaRowBytes, 0u);
 }
 
+TEST(AvifDecodeTest, AnimatedImage) {
+  if (!testutil::Av1DecoderAvailable()) {
+    GTEST_SKIP() << "AV1 Codec unavailable, skip test.";
+  }
+  // Test case from https://github.com/AOMediaCodec/libavif/issues/1203.
```
y-guyon, line 36: Leftover

tests/gtest/avifdecodetest.cc
```diff
@@ -29,6 +29,26 @@ TEST(AvifDecodeTest, ColorGridAlphaNoGrid) {
   EXPECT_GT(decoder->image->alphaRowBytes, 0u);
 }
 
+TEST(AvifDecodeTest, AnimatedImage) {
+  if (!testutil::Av1DecoderAvailable()) {
+    GTEST_SKIP() << "AV1 Codec unavailable, skip test.";
+  }
```
y-guyon, line 35: I guess this could be moved right before the first call to `avifDecoderNextImage()` to exercise `avifDecoderParse()` when there is no AV1 codec, but at the same time it breaks the test logic flow.

y-guyon (changes requested):
Thanks a lot for adding tests, it is highly appreciated.

tests/gtest/avifdecodetest.cc
vigneshvg: Ack, while that's theoretically true, there's not much point in testing `avifDecoderParse` when no AV1 decoder is available.

tests/gtest/avifdecodetest.cc
vigneshvg: Removed.

tests/gtest/avifdecodetest.cc
vigneshvg: Done.

I didn't do the two commits thing you suggested. But the diff should be easy to see by just diffing the two files.

tests/data/README.md
vigneshvg: Good point.

Removed the file from chromium and replaced it with a dummy image generated with ffmpeg.

tests/data/README.md
```diff
@@ -362,6 +362,24 @@ Box structure of the items in this file:
...
 
+## Animated Images
+
+### File [colors-animated-8bpc.avif](colors-animated-8bpc.avif)
+
+![](colors-animated-8bpc.avif)
+
+License: [same as libavif](https://github.com/AOMediaCodec/libavif/blob/main/LICENSE)
+
+Source: Random colors generated with ffmpeg.
```
y-guyon, line 373: Optional: Add the command line so that others can reuse it in the future.

y-guyon (approved):

PR #1560, opened 2023-09-06
Remove unused includes in tilingtest,utilstest

y-guyon:


vrabaud (approved):

vrabaud (approved):
The variable CMAKE_C_INCLUDE_WHAT_YOU_USE can be defined at cmake time by giving the path of the tool. More unused headers could be fine. That could be part of the CI but I believe there are still some false positives. https://cmake.org/cmake/help/latest/prop_tgt/LANG_INCLUDE_WHAT_YOU_USE.html

PR #1546, opened 2023-08-31
Add a function to convert gain map metadata double values to fractions.

maryla-uc:

PR #1565, opened 2023-09-08
Add support for reading jpeg gain maps and converting them to AVIF.

maryla-uc:
Needs libxml2 and AVIF_ENABLE_EXPERIMENTA_GAIN_MAP to be turned on.

The parsing code is a little long because there are basically three parts:
- MPF (Multi-Picture Format) metadata parsing, to find the offset(s) of the additional image(s) in the file. These offsets are relative fo the position of the MPF segment.
- JPEG metadata segment parsing, to find the offset of the MPF segment, in order to make the image offsets absolute (unfortunately jpeglib does not provide this information)
- XMP parsing (using libxml2) for the gain map metadata


CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   in the future. Files created now might not decode in a future version.
   This feature is off by default and must be enabled with the
   AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP compilation flag.
+* Add experiment support for converting jpeg files with gain maps to AVIF
```
y-guyon, line 20: experimental*

apps/avifenc.c
```diff
@@ -2276,6 +2295,7 @@ int main(int argc, char * argv[])
                                     /*ignoreExif=*/AVIF_TRUE,
                                     /*ignoreXMP=*/AVIF_TRUE,
                                     /*allowChangingCicp=*/AVIF_FALSE,
+                                    settings.ignoreGainMap,
```
y-guyon, line 2298: `ignoreGainMap` like above?

ext/libxml2.cmd
```diff
@@ -0,0 +1,12 @@
+: # If you want to use a local build of libxml2, you must clone the libxml2 repo in this directory first, then enable CMake's AVIF_LOCAL_LIBXML2 option.
+: # The git tag below is known to work, and will occasionally be updated. Feel free to use a more recent commit.
+
+: # The odd choice of comment style in this file is to try to share this script between *nix and win32.
+
+git clone -b v2.11.5 --depth 1 https://gitlab.gnome.org/GNOME/libxml2.git
+
+mkdir -p libxml2/build.libxml2
```
y-guyon, line 8: Note: Other `ext/` scripts use "build.libavif" naming for build folders. "build.libxml2" is fine too.

tests/gtest/avifjpeggainmaptest.cc
```diff
@@ -0,0 +1,103 @@
+// Copyright 2022 Google LLC
```
y-guyon, line 1: 2023

tests/gtest/avifjpeggainmaptest.cc
```diff
@@ -0,0 +1,103 @@
...
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include <math.h>
+
+#include <array>
```
y-guyon, line 6: Unused

tests/gtest/avifjpeggainmaptest.cc
```diff
@@ -0,0 +1,103 @@
...
+                            /*ignore_gain_map=*/false);
+    ASSERT_NE(image, nullptr);
+    ASSERT_NE(image->gainMap.image, nullptr);
+    EXPECT_EQ(image->gainMap.image->width, 512u);
+    EXPECT_EQ(image->gainMap.image->height, 384u);
+    EXPECT_EQ(image->xmp.size, 0u);
+
+    EXPECT_FALSE(image->gainMap.metadata.baseRenditionIsHDR);
+    const double kEpsilon = 1e-8;
+    EXPECT_NEAR((double)image->gainMap.metadata.hdrCapacityMinN /
```
y-guyon, line 40:
```cpp
const auto& m = image->gainMap.metadata;
```
for shorter lines

tests/gtest/avifjpeggainmaptest.cc
```diff
@@ -0,0 +1,103 @@
...
+                    image->gainMap.metadata.hdrCapacityMinD,
+                1, kEpsilon);
+    EXPECT_NEAR((double)image->gainMap.metadata.hdrCapacityMaxN /
```
y-guyon, line 43: `static_cast<double>()` is recommended but `(double)` is fine

tests/gtest/avifjpeggainmaptest.cc
```diff
@@ -0,0 +1,103 @@
...
+        testutil::ReadImage(data_path, filename, AVIF_PIXEL_FORMAT_YUV444, 8,
+                            AVIF_CHROMA_DOWNSAMPLING_AUTOMATIC,
+                            /*ignore_icc=*/false, /*ignore_exif=*/false,
+                            /*ignore_xmp=*/true, /*allow_changing_cicp=*/true,
+                            /*ignore_gain_map=*/false);
+    ASSERT_NE(image, nullptr);
+    ASSERT_NE(image->gainMap.image, nullptr);
+    EXPECT_EQ(image->gainMap.image->width, 512u);
+    EXPECT_EQ(image->gainMap.image->height, 384u);
+    EXPECT_EQ(image->xmp.size, 0u);
```
y-guyon, line 36: Does this test require ignoring xmp? If so, a quick comment explaining why would be welcomed.

y-guyon (changes requested):
Still have apps/shared/avifjpeg.c to review

apps/shared/avifjpeg.c
```diff
@@ -568,18 +1149,50 @@ avifBool avifJPEGRead(const char * inputFilename,
         }
         avifImageFixXMP(avif); // Remove one trailing null character if any.
     }
+
+#if defined(AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION)
+    // The primary XMP block (for the main image) must contain a node with an hdrgm:Version field if and only if a gain map is present.
```
y-guyon, line 1154: hdrm*?

apps/shared/avifjpeg.c
```diff
@@ -568,18 +1149,50 @@ avifBool avifJPEGRead(const char * inputFilename,
...
+    if (!ignoreGainMap && avifJPEGHasGainMapXMPNode(avif->xmp.data, avif->xmp.size)) {
+        avifJPEGExtractGainMapImage(f, inputFilename, &cinfo, &avif->gainMap, requestedFormat, requestedDepth, chromaDownsampling);
```
y-guyon, line 1156: Returned value is ignored? Should it at least print a warning if the tag is found without a valid gain map?

y-guyon (changes requested):

apps/avifenc.c
maryla-uc: This is for the grid case, so I don't think we need to be concerned about animations or progressive images. These are all mutually exclusive, right?

apps/shared/avifjpeg.c
```diff
@@ -266,6 +277,568 @@ static const uint8_t * avifJPEGFindSubstr(const uint8_t * str, size_t strLength,
...
+    const long oldOffset = ftell(f);
+
+    uint32_t offset = 2; // Skip the 2 byte SOI (Start Of Image) marker.
+    if (fseek(f, offset, SEEK_SET) != 0) {
+        return AVIF_FALSE;
+    }
+    uint8_t buffer[4];
+    // Guaranteed to end when we reach the end of the file since 'offset' always increases by at least
+    // JPEG_MPF_SEGMENT_HEADER_SIZE_BYTES.
+    while (1) {
```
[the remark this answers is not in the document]
maryla-uc: Done. More fuzzer coverage would indeed be good. I'll work on adding more tests/fuzzing.

ext/libxml2.cmd
```diff
@@ -0,0 +1,12 @@
+: # If you want to use a local build of libxml2, you must clone the libxml2 repo in this directory first, then enable CMake's AVIF_LOCAL_LIBXML2 option.
+: # The git tag below is known to work, and will occasionally be updated. Feel free to use a more recent commit.
+
+: # The odd choice of comment style in this file is to try to share this script between *nix and win32.
+
+git clone -b v2.11.5 --depth 1 https://gitlab.gnome.org/GNOME/libxml2.git
```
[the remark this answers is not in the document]
maryla-uc: I added a comment with the license.

apps/shared/avifjpeg.c
```diff
@@ -266,6 +277,568 @@ static const uint8_t * avifJPEGFindSubstr(const uint8_t * str, size_t strLength,
...
+{
+    if (parentNode == NULL) {
+        return NULL;
+    }
+    for (const xmlNode * node = parentNode->children; node != NULL; node = node->next) {
+        if (node->ns != NULL && !xmlStrcmp(node->ns->href, (const xmlChar *)nameSpace) &&
+            !xmlStrcmp(node->name, (const xmlChar *)nodeName)) {
+            return node;
+        } else if (recursive && node->children != NULL) {
+            const xmlNode * descendantNode = avifJPEGFindXMLNodeByName(node->children, nameSpace, nodeName, recursive);
```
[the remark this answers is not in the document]
maryla-uc: Indeed, good catch.

apps/shared/avifjpeg.c
```diff
@@ -266,6 +277,568 @@ static const uint8_t * avifJPEGFindSubstr(const uint8_t * str, size_t strLength,
...
+    for (const xmlNode * node = rdfNode->children; node != NULL; node = node->next) {
+        // Loop through rdf::Description children.
+        // 7.4 "A single XMP packet shall be serialized using a single rdf:RDF XML element. The rdf:RDF element content
+        // shall consist of only zero or more rdf:Description elements."
+        if (node->ns && !xmlStrcmp(node->ns->href, (const xmlChar *)XML_NAME_SPACE_RDF) &&
+            !xmlStrcmp(node->name, (const xmlChar *)"Description")) {
+            // Look for the gain map version attribute: hdrm:Version="1.0"
+            for (xmlAttr * prop = node->properties; prop != NULL; prop = prop->next) {
+                if (prop->ns && !xmlStrcmp(prop->ns->href, (const xmlChar *)XML_NAME_SPACE_GAIN_MAP) &&
+                    !xmlStrcmp(prop->name, (const xmlChar *)"Version") && prop->children != NULL) {
```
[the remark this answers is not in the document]
maryla-uc: I guess you could imagine in the future having several versions in one file? So I changed to check all nodes.

apps/shared/avifjpeg.c
```diff
@@ -266,6 +277,568 @@ static const uint8_t * avifJPEGFindSubstr(const uint8_t * str, size_t strLength,
...
+    AVIF_CHECK(avifGainMapMetadataDoubleToFractions(metadata, &metadataDouble));
+
+    return AVIF_TRUE;
+}
+
+// Parses gain map metadata from an XMP payload.
+// Returns AVIF_TRUE if the gain map metadata was successfully read.
+static avifBool avifJPEGParseGainMapXMP(const uint8_t * xmpData, size_t xmpSize, avifGainMapMetadata * metadata)
+{
+    xmlDoc * document = xmlReadMemory((const char *)xmpData, (int)xmpSize, NULL, NULL, LIBXML2_XML_PARSING_FLAGS);
```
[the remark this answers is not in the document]
maryla-uc: It returns NULL from what I can tell. I added a check for that for clarity but I think it already worked because the next functions all return NULL if the input is NULL.

apps/shared/avifjpeg.c
```diff
@@ -266,6 +277,568 @@ static const uint8_t * avifJPEGFindSubstr(const uint8_t * str, size_t strLength,
...
+    }
+
+    uint32_t offsetToFirstIfd;
+    AVIF_CHECK(avifJPEGReadU32(segmentData, &offsetToFirstIfd, &offset, isBigEndian));
+    if (offsetToFirstIfd < offset) {
+        return AVIF_FALSE;
+    }
+    offset = offsetToFirstIfd;
+
+    uint16_t mpTagCount;
```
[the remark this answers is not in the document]
maryla-uc: Multi Picture

apps/shared/avifjpeg.c
```diff
@@ -266,6 +277,568 @@ static const uint8_t * avifJPEGFindSubstr(const uint8_t * str, size_t strLength,
...
+                                  avif,
+                                  requestedFormat,
+                                  requestedDepth,
+                                  chromaDownsampling,
+                                  /*ignoreColorProfile=*/AVIF_TRUE,
+                                  /*ignoreExif=*/AVIF_TRUE,
+                                  /*ignoreXMP=*/AVIF_FALSE,
+                                  /*ignoreGainMap=*/AVIF_TRUE)) {
+            continue;
+        }
```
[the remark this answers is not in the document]
maryla-uc: Indeed! I think avifImageAllocatePlanes() should be changed to call Free automatically but that's a change for another PR (and it changes the API a bit so some poeple might not agree to it).

apps/shared/avifjpeg.c
```diff
@@ -266,6 +277,568 @@ static const uint8_t * avifJPEGFindSubstr(const uint8_t * str, size_t strLength,
...
+            avifImage * image = avifImageCreateEmpty();
+
+            const avifROData mpfData = { (const uint8_t *)marker->data + tagMpf.size, marker->data_length - tagMpf.size };
+            if (!avifJPEGExtractGainMapImageFromMpf(f, inputFilename, &mpfData, image, requestedFormat, requestedDepth, chromaDownsampling)) {
+                fprintf(stderr, "Warning: failed to read gain map image\n");
+                avifImageDestroy(image);
+                return AVIF_FALSE;
+            }
+            if (!avifJPEGParseGainMapXMP(image->xmp.data, image->xmp.size, &gainMap->metadata)) {
+                fprintf(stderr, "Warning: failed to parse gain map metadata\n");
```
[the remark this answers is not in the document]
maryla-uc: Yes that's possible but if we're in this function it means the main image had some XMP indicating the presence of a gain map. So we do expect to find one. However, I can very easily imagine the gain map getting dropped during processing (e.g. if you modify the file with imagemagick probably?) so  there will probably be cases like that. Changed to a "note".

apps/shared/avifjpeg.c
maryla-uc: hdrgm is actually the one that's correct, so I fixed the instances of "hdrm".
Note that I don't actually check that the prefix is "hdrgm", just that there is a prefix associated with the namespace "http://ns.adobe.com/hdr-gain-map/1.0/". This seems more in the spirit of XML to me, where the prefix should be arbitrary. But Adobe's spec says that the prefix should be "hdrgm", not sure why, maybe for simplicity. I don't think it matters much and I expect that it will always be "hdrgm" in practice.

apps/shared/avifjpeg.c
maryla-uc: avifJPEGExtractGainMapImage already prints warnings, added a comment.

apps/avifenc.c
y-guyon: > These are all mutually exclusive, right?

Currently, yes.

y-guyon (approved):

include/avif/gainmap.h
```diff
@@ -33,7 +33,7 @@ typedef struct avifGainMapMetadataDouble
 // Converts a avifGainMapMetadataDouble to avifGainMapMetadata by converting double values
 // to the closest uint32_t fractions.
 // Returns AVIF_FALSE if some field values are < 0 or > UINT32_MAX.
-avifBool avifGainMapMetadataDoubleToFractions(avifGainMapMetadata * dst, const avifGainMapMetadataDouble * src);
+AVIF_API avifBool avifGainMapMetadataDoubleToFractions(avifGainMapMetadata * dst, const avifGainMapMetadataDouble * src);
```
wantehchang, line 36: IMPORTANT: Since `avifGainMapMetadataDoubleToFractions()` is exported from the libavif shared library (for use by avifjpeg.c), it is part of the public API. Please move this declaration to the public header avif.h.

After we move this declaration to avif.h, gainmap.h becomes an empty file and should be deleted.

Note: An alternative is to move `avifGainMapMetadataDoubleToFractions()` to avifjpeg.c (as a static function) or avifutil.{h,c}, if this function should not be in the public API of the libavif library.

maryla-uc: This was intentional. I think `avifGainMapMetadataDoubleToFractions` should be in the public API to make it easier to fill in `avifGainMapMetadata`. I also thought this type of gain map related utility function (I expect more to be added in the future) could be in a separate header file, in order not to clutter avif.h
What's the rationale for having all the public api functions in a single header file?

maryla-uc: Another option would be to #include gainmap.h from avif.h so that users only need to #include avif.h. But it's not really in the "include what you use" philosophy.

wantehchang: The rationale for having all the public api functions in avif/avif.h is simplicity. This design is quite common. For example, libpng declares all the public api functions in png.h:

https://packages.debian.org/sid/amd64/libpng-dev/filelist

Since it's not critical to add an avif/gainmap.h header, it would be better to stick to the current design.

I just wrote a pull request https://github.com/AOMediaCodec/libavif/pull/1589 to document the current libavif header inclusion policy.

maryla-uc: A single header file is only "simpler" in the sense that you only have one header file to include, but I think it makes the API more difficult to understand and navigate (the file is 1400 lines long and counting).

This single header file design also means we can't split the encoder and decoder, and even apps that only need decoding need to include the encoder (the libavif part anyway). libaom and libwebp both have separate encoding and decoding headers https://aomedia.googlesource.com/aom/+/refs/heads/main/aom/  https://chromium.googlesource.com/webm/libwebp/+/refs/heads/main/src/webp/

And of course, many more libraries have their API split into multiple header files, for example libxml2 which I used recently https://gitlab.gnome.org/GNOME/libxml2/-/tree/master/include/libxml

In particular, gainmap utilities like this are not really tied to the main libavif API and I still think it would be better not to clutter avif.h with them.
Again, if we really think it's important that users should only need to include one header, we can "export" gainmap.h from avif.h the same way internal.h exports avif.h?

wantehchang: I just realized that avif/gainmap.h is already a public header (see the second change to CMakeLists.txt in https://github.com/AOMediaCodec/libavif/pull/1546). I am sorry I missed that. I thought we were exporting a function (`avifGainMapMetadataDoubleToFractions`) declared in an internal header. That would be a bug and was why I marked my initial comment with "IMPORTANT". Sorry about my confusion.

Note: Before I went on vacation on Sept. 3, I took a quick look at https://github.com/AOMediaCodec/libavif/pull/1546 and noticed the new avif/gainmap.h header. I wrote a comment explaining the current header policy of libavif and asking whether avif/gainmap.h was an internal or public header (in that pull request, avif/gainmap.h was only used by unit tests, so I could not easily tell), but I forgot to publish my review comments.

maryla-uc: I see, I guess I wasn't clear that this was meant to be a public API from the start, I simply forgot to add AVIF_API in the first version.
The idea is to provide users an easy way to fill in the nominator/denominator fields in avifGainMapMetadata, which also happens to be useful for avifjpeg.c
So are you ok with keeping it in a separate public header then?

wantehchang: Based on my code analysis alone, I would put `avifGainMapMetadataDoubleToFractions()` in apps/shared/avifutil.{h,c} or avifjpeg.c, and wait until a user requests it to move it to the libavif library. This is a safe strategy when it is not clear whether a new utility function is useful enough to be added to the library, because the library has backward compatibility requirements now that version 1.0.0 has been released. But you are much more familiar with gain maps and may already know this utility function is generally useful. So I deter this decision to you.

As for adding a separate public header, I don't strongly prefer a single header over multiple headers; I am equally comfortable in both designs. It's simply that the current monolithic header has never bothered me (and I read/analyze code frequently), so I see no need to change. (And include-what-you-use is simple when there is only one header.) If we break away from the monolithic header, we need to do it properly.

maryla-uc: I see what you mean and I do understand the concern about API maintenainance. But I also think it's a useful utility for anybody who would want to use this API. Double (or float) values are the natural way of representing all of these metadata fields. The fact that we use fractions is just an implementation detail. Anybody who would want to use the API would have to convert floating point values to fractions, which is why I wanted to provide this utility.

Similarly, we have avifCropRectConvertCleanApertureBox() and avifCleanApertureBoxConvertCropRect() to make the clean aperture feature easier to use.

Regarding header files, I think I prefer multiple headers in general but until we make a final decision on this I guess it can go in avif.h.

This is all still behind a compile flag so nothing is set in stone anyway.

wantehchang: Do you think our public API should just take double (or float) values and then we convert them internally to fractions? Or is it better for the public API to reflect the file format faithfully?

I guess some clients may prefer to specify the values as fractions so they can control the values precisely.

maryla-uc: I wondered about this too, and at first I had double values in the API since I copied the Skia interface. But using fractions indeed gives more control and allows retrieving exactly the values that were written in the file.

Yannis suggested using fractions with a helper utility like we do for clap and I thought it made sense. This way the clap and gain map APIs are consistent in that they expose the numerator/denominator fields.

But overall I don't think any of these arguments are very strong so I'm also open to changing it back to double values if you think it's better/simpler.

wantehchang: I think double values are simpler; what I don't know is whether any of our users need to control precisely the values written to the file. We can re-evaluate this when we officially support gain maps.

PR #1567, opened 2023-09-08
Treat SYSTEM include directories as such

AMDmi3:
Include system directories with corresponding keyword. This effectively lowers their lookup priority relative to project include directories and prevents the build picking libavif includes from system installed version instead of repository version.


AMDmi3:
This fixes `0.11.1` → `1.0.1` from sources (when 0.11.1 is installed) for me.

Root CMakeLists needs similar change, but it's a bit more complex, as it requires splitting `AVIF_PLATFORM_INCLUDES` and `AVIF_CODEC_INCLUDES` into regular and SYSTEM variables. Using the opportunity to prediscuss this change, if it looks acceptable I'll submit patch for root CMakeLists.txt as well.

vrabaud:
Indeed, the main CMakeLists.txt could use SYSTEM with libyuv, PNG and JPEG

vrabaud (approved):

PR #1575, opened 2023-09-11
Add avifBreakOnError() to internal.h

y-guyon:
To debug `AVIF_CHECK*()` failures by just adding `!` to
```c
#if defined(AVIF_BREAK_ON_ERROR)
```
I have used this debugging pattern multiple times now so I thought it would be handy to have it committed. Happy to hear opinions though.

Remove `AVIF_CHECK()` from `avifPeekCompatibleFileType()` because it is expected to often fail.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -15,13 +15,13 @@ extern "C" {
 #define AVIF_MIN(a, b) (((a) < (b)) ? (a) : (b))
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
-// Used for debugging. Add ! to catch the earliest failure during encoding or decoding.
+// Used for debugging. Define AVIF_BREAK_ON_ERROR to catch the earliest failure during encoding or decoding.
 #if defined(AVIF_BREAK_ON_ERROR)
 static inline void avifBreakOnError()
 {
-    // Same mechanism as OpenCV's error() function.
-    static volatile int * p = 0;
-    *p = 0; // or replace by a breakpoint
+    // Same mechanism as OpenCV's error() function, or replace by a breakpoint.
+    int * p = NULL;
+    *p = 0;
 }
 #else
 #define avifBreakOnError()
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -19,7 +19,9 @@ extern "C" {
 #if defined(AVIF_BREAK_ON_ERROR)
 static inline void avifBreakOnError()
 {
-    abort(); // or replace by a breakpoint
+    // Same mechanisme as OpenCV's error() function.
+    static volatile int * p = 0;
+    *p = 0; // or replace by a breakpoint
 }
 #else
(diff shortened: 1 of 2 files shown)


include/avif/internal.h
```diff
@@ -15,26 +15,41 @@ extern "C" {
 #define AVIF_MIN(a, b) (((a) < (b)) ? (a) : (b))
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
+#if defined(AVIF_BREAK_ON_ERROR)
```
maryla-uc, line 18: Add a quick comment that this is used for debugging and the tip to add "!"?

maryla-uc (approved):

vrabaud:
You can also go this route so that gdb stops at that line: https://github.com/opencv/opencv/blob/70d7e83dca7824e7e9d09e03630d1134628dc70a/modules/core/src/system.cpp#L1279

include/avif/internal.h
```diff
@@ -15,26 +15,44 @@ extern "C" {
 #define AVIF_MIN(a, b) (((a) < (b)) ? (a) : (b))
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
+// Used for debugging. Add ! to catch the earliest failure during encoding or decoding.
+#if defined(AVIF_BREAK_ON_ERROR)
+static inline void avifBreakOnError()
+{
+    // Same mechanism as OpenCV's error() function.
+    static volatile int * p = 0;
```
wantehchang, line 23: `p` (a pointer) should be initialized to `NULL`.

The `static` and `volatile` keywords should be unnecessary. I often use the same code (without those two keywords) to force a crash.

include/avif/internal.h
```diff
@@ -15,26 +15,44 @@ extern "C" {
 #define AVIF_MIN(a, b) (((a) < (b)) ? (a) : (b))
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
+// Used for debugging. Add ! to catch the earliest failure during encoding or decoding.
```
wantehchang, line 18: Optional: I thought we were supposed to add a `#define AVIF_BREAK_ON_ERROR` line to enable this. If we enable it by modifying line 19, it seems clearer to rewrite line 19 as `#if 0` and ask users to change it to `#if 1` to enable it.

include/avif/internal.h
```diff
@@ -15,26 +15,44 @@ extern "C" {
...
+#if defined(AVIF_BREAK_ON_ERROR)
+static inline void avifBreakOnError()
+{
+    // Same mechanism as OpenCV's error() function.
+    static volatile int * p = 0;
+    *p = 0; // or replace by a breakpoint
```
wantehchang, line 24: Nit: This comment should be moved before line 23, because it now describes two lines (23-24). Originally it described only this line when this line was `abort();`.

PR #1579, opened 2023-09-12
build: link against static local dependencies when building a shared library

fdintino:
When building libavif as a shared library, it makes more sense for local dependencies to be statically linked. Otherwise, the shared library is not portable: it can be run from the build directory but not anywhere else. I've modified the ci-unix-shared-local workflow on a separate branch to demonstrate the issue, adding a `ninja install` and then a call to `ldd` (to show the missing links) and `avifenc --help` ([job output](https://github.com/fdintino/libavif/actions/runs/6158423138/job/16711165370#step:22:17))

```
Run ldd /usr/local/lib/libavif.so
	linux-vdso.so.1 (0x00007ffefa387000)
	libdav1d.so.6 => not found
	libaom.so.3 => /lib/x86_64-linux-gnu/libaom.so.3 (0x00007f3421ed6000)
	libyuv.so => not found
	libsharpyuv.so.0 => not found
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f3421def000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3421bc5000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f3422485000)

avifenc: error while loading shared libraries: libavif.so: cannot open shared object file: No such file or directory
Error: Process completed with exit code 127.
```

I encountered this issue in my own GHA workflow when I tried to build against the latest version of libavif, since I don't have the `sed` commands that were added to ci-unix-shared-local, and because the CMakeLists.txt was modified to _only_ allow shared local dependencies when building a shared library.


vrabaud:
I'd also be in favor of forcing static linkage to simplify CMake (which you already did).
I do not see any use case for keeping shared deps. Any opinion @wantehchang ?

fdintino:
Any updates on this?

vrabaud:
Still in favor: If we delete a folder in ext by inadvertence, the binaries still use the right versions of the deps (the `.a`).

.github/workflows/ci-disable-gtest.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
       with:
         path: ext
-        key: ${{ runner.os }}-disable-gtest-${{ hashFiles('ext/*.cmd') }}
+        key: ${{ runner.os }}-disable-gtest-${{ hashFiles('CMakeLists.txt', 'ext/*.cmd') }}
```
wantehchang, line 45: Frankie: The dependency of the `ext` cache files on CMakeLists.txt seems wrong. The contents of the `ext` directory do not depend on CMakeLists.txt. Could you explain why this is necessary?

ext/libyuv.cmd
```diff
@@ -22,6 +22,6 @@ git checkout 464c51a0
 mkdir build
 cd build
 
-cmake -G Ninja -DCMAKE_BUILD_TYPE=Release ..
+cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_SHARED_LIBS=OFF ..
```
wantehchang, line 25: Frankie: We need to make sure all the build scripts in the `ext` directory compile the respective libraries with position-independent code enabled. This pull request only made such changes to the libyuv scripts. Have you checked the scripts for the other libraries?

CMakeLists.txt
```diff
@@ -162,7 +160,7 @@ if(AVIF_LOCAL_LIBYUV)
     if(DEFINED ANDROID_ABI)
         set(AVIF_LIBYUV_BUILD_DIR "${AVIF_LIBYUV_BUILD_DIR}/${ANDROID_ABI}")
     endif()
-    set(LIB_FILENAME "${AVIF_LIBYUV_BUILD_DIR}/${AVIF_LIBRARY_PREFIX}yuv${AVIF_LIBRARY_SUFFIX}")
+    set(LIB_FILENAME "${AVIF_LIBYUV_BUILD_DIR}/${AVIF_LIBRARY_PREFIX}yuv${CMAKE_STATIC_LIBRARY_SUFFIX}")
```
wantehchang, line 163: Yannis: Could you review this pull request? It reverts some of the changes you made in https://github.com/AOMediaCodec/libavif/pull/1169.

wantehchang (commented):
Vincent: I'm sorry I missed your previous question for me on Sep 12 (because I just returned to work from a vacation). I think it is reasonable to link with the kind of library built by the scripts in the `ext` directory (which is static libraries). I've asked Yannis to review this pull request because this is closely related to his https://github.com/AOMediaCodec/libavif/pull/1169. (I will also review this pull request.)

ext/libyuv.cmd
wantehchang, line 25: Yannis: I believe `-DBUILD_SHARED_LIBS=OFF` should not be added, because libyuv's CMakeLists.txt controls the kind of library to build by the target name (yuv vs. yuv_shared). Please check this change.

fdintino: I've been building libavif as a shared library with static dependencies for a while in [pillow-avif-plugin](https://github.com/fdintino/pillow-avif-plugin), and I knew this to be a problem with libyuv. It looks like dav1d, rav1e, SVT-AV1, and libsharpyuv (by way of libwebp) all pass `-fPIC` when being built statically. libgav1, avm, and aom appear not to (and in fact, it looks like I'm also passing in `-fPIC` into aom; I don't link against libgav1 or avm). Should I add `-DCMAKE_POSITION_INDEPENDENT=ON` to all of the `.cmd` scripts, or just the ones that don't already have it as a default for static builds?

.github/workflows/ci-disable-gtest.yml
fdintino: Oh... so, it looks like I had encountered issues because the cache action is here, but then the `sed` that was replacing the cmake flags happens below. So when I removed the sed action it was still using the cached ext folder with the shared libraries. So perhaps the appropriate fix would be to modify the cache key slightly, without having it depend on the contents of CMakeLists.txt? (which as you correctly point out doesn't affect the builds in the ext directory).

fdintino: Actually, since this pull request will be merged in all at once, and since it involves changes to the ext/*.cmd files, I'll just undo these changes.

ext/libyuv.cmd
fdintino: Oh interesting. And we've only ever linked against yuv. I'll remove the flag since it doesn't do anything, but keep `- DCMAKE_POSITION_INDEPENDENT_CODE=ON`?

CMakeLists.txt
fdintino: It's useful whenever you want to distribute a library that provides bindings to libavif with a bundled build of libavif. For python and node, this is accomplished by shipping your extension with a shared library that has a modified RPATH, to get picked up at runtime.

I could be doing something wrong, but I have had difficulty linking against a static build of libavif. It seems that I need to combine the other static libraries using `LIB.EXE` or `ar`, or else add them to the link flags when I build the extension. And the libavif.pc file doesn't include `Libs.private` or `Requires.private`, so I can't use pkg-config to figure out what those link flags should be.

wantehchang: Yannis: We can still test "everything shared" in the CI. It's just that the dependent libraries such as libaom and dav1d cannot be locally built (in the `ext` directory) and need to be installed as packages.

Frankie: I am not familiar with pkg-config, but I think libavif.pc should imitate how libpng's .pc file handles the zlib dependency (assuming libpng does it correctly):
https://github.com/glennrp/libpng/blob/libpng16/libpng.pc.in

.github/workflows/ci-unix-shared-local.yml
```diff
@@ -51,15 +51,6 @@ jobs:
...
       run: convert --version
-    - name: Set shared libs
-      if: steps.cache-ext.outputs.cache-hit != 'true'
-      working-directory: ./ext
-      run: >
-        sed -i'' -e 's/-DBUILD_SHARED_LIBS=OFF/-DBUILD_SHARED_LIBS=ON/' *.cmd
-
-        sed -i'' -e 's/--default-library=static/--default-library=shared/' dav1d.cmd
-
-        sed -i'' -e 's/ninja yuv/ninja yuv_shared/' libyuv.cmd
```
wantehchang, line 62: I think we should delete this file, because it is no longer testing locally-built shared libs.

ext/libyuv.cmd
wantehchang: We only need to add `-DCMAKE_POSITION_INDEPENDENT=ON` to the `.cmd` scripts that don't already have it as a default for static builds. But it may be time-consuming to find out which `.cmd` scripts need `-DCMAKE_POSITION_INDEPENDENT=ON`. So I think it is fine to add it to all the `.cmd` scripts.

vrabaud: `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` as the library does not handle it

vrabaud: Last I checked, most dealt properly with PIC. We can just run the CI and only modify the ones who fail. Just to keep the .cmd minimal

CMakeLists.txt
vrabaud: @fdintino, just to be sure: you were building libavif as static, with dynamic deps, installing it and using pkgconfig to use it? (not CMake)

ext/libyuv.cmd
fdintino: In any event, I've already identified which don't deal with PIC (libgav1, avm, aom, and libyuv). I had added it to all based on the initial feedback, but I'll change it back to only those scripts.

.github/workflows/ci-unix-shared-local.yml
fdintino: Wouldn't we still want to test a shared build of libavif with local dependencies, regardless of how those dependencies are compiled? I think this is the only workflow dedicated specifically to that.

vrabaud: Actually yes, you are right: at least, we check that we compile against the recommended versions.

wantehchang: You are right. I was confused. Sorry.

ext/aom.cmd
```diff
@@ -14,6 +14,6 @@ cd aom
 mkdir build.libavif
 cd build.libavif
 
-cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DENABLE_DOCS=0 -DENABLE_EXAMPLES=0 -DENABLE_TESTDATA=0 -DENABLE_TESTS=0 -DENABLE_TOOLS=0 ..
+cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_BUILD_TYPE=Release -DENABLE_DOCS=0 -DENABLE_EXAMPLES=0 -DENABLE_TESTDATA=0 -DENABLE_TESTS=0 -DENABLE_TOOLS=0 ..
```
wantehchang, line 17: In aom.cmd and avm.cmd, use `-DCONFIG_PIC=1` instead of `-DCMAKE_POSITION_INDEPENDENT_CODE=ON`.

`-DCONFIG_PIC=1` will set `CMAKE_POSITION_INDEPENDENT_CODE` to `ON` and also define the `PIC` macro for libaom/libavm's assembly code.

fdintino:
Following up on the discussion about statically linking against libavif: [I've created a branch](https://github.com/fdintino/libavif/compare/ci/shared-local-static...ci/static-combine) that merges the platform and codec static libraries into libavif.a / avif.lib. I've also added a step in the CI static build workflows to test that an executable can be built without having to link against any local static dependencies. I've also created a separate branch which includes those tests but doesn't include my cmake changes, to demonstrate the issue I'm trying to fix; here are the failing [CI Unix Static](https://github.com/fdintino/libavif/actions/runs/6346140615/job/17239232397) and [CI Windows](https://github.com/fdintino/libavif/actions/runs/6346140619/job/17239232288) runs.

Depending on how folks feel about those changes, I'd be happy to open them in a separate pull request. Or I could add that as a commit to this PR.

vrabaud:
Thx a lot @fdintino for all the work: I looked at your other PR and it does look right. It is weird that CMake does not bundle static libraries by default.

fdintino:
@vrabaud okay, I've gone ahead and added that commit to this pull request. Thanks!

CMakeLists.txt
fdintino: Closing the loop on this conversation: as I noted in a comment on the PR itself, I've opted to include the commit that fixes the static library bundling as an additional commit here.

CMakeLists.txt
```diff
@@ -107,6 +105,22 @@ set(AVIF_PLATFORM_INCLUDES)
 set(AVIF_PLATFORM_SYSTEM_INCLUDES)
 set(AVIF_PLATFORM_LIBRARIES)
 
+# A macro that wraps find_package, preferring static libraries if this is a static build
+# Should only be used for libraries linked against avif, not for those only used
+# by avif_apps or tests
+macro(find_package_libavif)
```
vrabaud, line 111: Similarly, there is not official way of doing this in CMake, looking for extensions is the current right way of doing it.

vrabaud (approved):
Thx, I believe you answered all of our requests and your CMake solutions are the right ones. Thank you!

CMakeLists.txt
```diff
@@ -508,15 +515,15 @@ if(AVIF_CODEC_RAV1E)
...
+            find_package_libavif(rav1e REQUIRED)
             set(AVIF_CODEC_SYSTEM_INCLUDES ${AVIF_CODEC_SYSTEM_INCLUDES} ${RAV1E_INCLUDE_DIR})
         endif()
         set(AVIF_CODEC_LIBRARIES ${AVIF_CODEC_LIBRARIES} ${RAV1E_LIBRARIES})
     endif()
 
     # Unfortunately, rav1e requires a few more libraries
     if(WIN32)
-        set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ws2_32.lib bcrypt.lib userenv.lib ntdll.lib)
+        set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} legacy_stdio_definitions.lib ntdll.lib advapi32.lib userenv.lib ws2_32.lib bcrypt.lib msvcrt.lib)
```
wantehchang, line 526: @fdintino Why were legacy_stdio_definitions.lib, advapi32.lib, and msvcrt.lib added?

fdintino: I had missing symbols errors without them. I got this list from the windows rav1e.pc Libs.private

PR #1580, opened 2023-09-12
Treat SYSTEM include directories as such

AMDmi3:
Follow up to  #1567, this PR covers the root project.

- I've had to add SYSTEM counterparts for `AVIF_PLATFORM_INCLUDES` and `AVIF_CODEC_INCLUDES` as these may contain both local and system includes
- I've split `target_include_directories` into multiple calls for readability
- I've had to duplicate `target_include_directories` for `avif_internal` target. Before, it took properties from the main library target, but it doesn't seem to treat SYSTEM includes properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,7 @@ endif()
 
 set(AVIF_PLATFORM_DEFINITIONS)
 set(AVIF_PLATFORM_INCLUDES)
+set(AVIF_PLATFORM_SYSTEM_INCLUDES)
 set(AVIF_PLATFORM_LIBRARIES)
 
 # ---------------------------------------------------------------------------------------
@@ -105,7 +106,7 @@ if(AVIF_LOCAL_ZLIBPNG)
     set(ZLIB_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ext/zlib" CACHE PATH "zlib include dir")
     # This include_directories() call must be before add_subdirectory(ext/zlib) to work around the
     # zlib/CMakeLists.txt bug fixed by https://github.com/madler/zlib/pull/818.
-    include_directories(${ZLIB_INCLUDE_DIR})
+    include_directories(SYSTEM ${ZLIB_INCLUDE_DIR})
     add_subdirectory(ext/zlib)
     # This include_directories() call and the previous include_directories() call provide the zlib
     # include directories for add_subdirectory(ext/libpng). Because we set PNG_BUILD_ZLIB,
@@ -187,7 +188,7 @@ else(AVIF_LOCAL_LIBYUV)
 endif()
 if(libyuv_FOUND)
     set(AVIF_PLATFORM_DEFINITIONS ${AVIF_PLATFORM_DEFINITIONS} -DAVIF_LIBYUV_ENABLED=1)
-    set(AVIF_PLATFORM_INCLUDES ${AVIF_PLATFORM_INCLUDES} ${LIBYUV_INCLUDE_DIR})
+    set(AVIF_PLATFORM_SYSTEM_INCLUDES ${AVIF_PLATFORM_SYSTEM_INCLUDES} ${LIBYUV_INCLUDE_DIR})
     set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ${LIBYUV_LIBRARY})
 endif(libyuv_FOUND)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -595,13 +600,15 @@ endif()
 # The avif_internal target should not be used by external code.
 if(BUILD_SHARED_LIBS)
     add_library(avif_internal STATIC ${AVIF_SRCS})
-    # Copy most properties from the public avif library target.
+    # Copy or duplicate most properties from the public avif library target.
     target_compile_definitions(avif_internal PRIVATE "$<TARGET_PROPERTY:avif,COMPILE_DEFINITIONS>")
     target_link_libraries(avif_internal PRIVATE "$<TARGET_PROPERTY:avif,LINK_LIBRARIES>")
-    target_include_directories(
-        avif_internal PUBLIC "$<TARGET_PROPERTY:avif,INTERFACE_INCLUDE_DIRECTORIES>"
-        PRIVATE "$<TARGET_PROPERTY:avif,INCLUDE_DIRECTORIES>"
```
vrabaud, line 603: Did you try using https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_SYSTEM_INCLUDE_DIRECTORIES.html#interface-system-include-directories

AMDmi3: Yes, I've played with it a bit, but couldn't get rid of `-I` system include. Haven't investigated further, e.g. from which `avif` target property it comes from.

vrabaud (approved):

CMakeLists.txt
```diff
@@ -595,13 +600,15 @@ endif()
 # The avif_internal target should not be used by external code.
 if(BUILD_SHARED_LIBS)
     add_library(avif_internal STATIC ${AVIF_SRCS})
-    # Copy most properties from the public avif library target.
+    # Copy or duplicate most properties from the public avif library target.
```
wantehchang, line 603: "Copy" and "duplicate" have similar meanings, so it's not clear what this comment is trying to differentiate by using these two words.

By looking at the diffs, I infer that this comment considers lines 604-605 as copying (because they use `TARGET_PROPERTY:avif`) and considers lines 607-608 as duplicating (because they don't use `TARGET_PROPERTY:avif` but rather "duplicate" lines 585-586 with the library name changed). Is that correct?

Line 606 is not a duplicate of line 584. Why is that?

CMakeLists.txt
```diff
@@ -595,13 +600,15 @@ endif()
...
-    target_include_directories(
-        avif_internal PUBLIC "$<TARGET_PROPERTY:avif,INTERFACE_INCLUDE_DIRECTORIES>"
-        PRIVATE "$<TARGET_PROPERTY:avif,INCLUDE_DIRECTORIES>"
-    )
+    target_include_directories(avif_internal PUBLIC ${libavif_SOURCE_DIR}/include)
+    target_include_directories(avif_internal PRIVATE ${AVIF_PLATFORM_INCLUDES} ${AVIF_CODEC_INCLUDES})
+    target_include_directories(avif_internal SYSTEM PRIVATE ${AVIF_PLATFORM_SYSTEM_INCLUDES} ${AVIF_CODEC_SYSTEM_INCLUDES})
+    if(NOT libyuv_FOUND)
+        target_include_directories(avif_internal PRIVATE ${libavif_SOURCE_DIR}/third_party/libyuv/include/)
+    endif()
```
wantehchang, line 611: Lines 609-611 are new. Is this something that's missed in https://github.com/AOMediaCodec/libavif/pull/1534?

wantehchang (commented):
Hi Dimitry,

Thank you for the pull request. I have some questions.

CMakeLists.txt
AMDmi3: > `TARGET_PROPERTY:avif`) and considers lines 607-608 as duplicating (because they don't use `TARGET_PROPERTY:avif` but rather "duplicate" lines 585-586 with the library name changed). Is that correct?

Yes

> Line 606 is not a duplicate of line 584. Why is that?

Because internal target does not need to differentiate build and install interfaces.

CMakeLists.txt
AMDmi3: These are the same as 587-589

AMDmi3:
> Thank you for the pull request. I have some questions.

The idea behind this change is, when setting up include directories for `avif_internal` target, to switch from trying to copy these properties from `avif` target (which doesn't preserve SYSTEM flag) to specifying identical properties explicitly. Thus, all `target_include_directories` directives for `avif` are just copypasted for `avif_internal` (and that's where libyuv related lines come from) with single minor adaptation for `avif_interlan` being an internal target, thus not needing INSTALL_INTERFACE includes.

CMakeLists.txt
wantehchang: I'm sorry my previous comment was not clear. What I meant is: while the other changes in this pull request can be considered as modifications to existing code, only these three lines are additions. So I should have said "Lines 609-611 are additions." rather than "Lines 609-611 are new."

So my question was whether these three lines should have been part of https://github.com/AOMediaCodec/libavif/pull/1534.

wantehchang:
Dimitry: Thanks for the additional explanation of this pull request.

PR #1581, opened 2023-09-12
Don't add ldflags to aom library

AMDmi3:
To link with aom, only AOM_LIBRARY is needed. Don't append LDFLAGS from pkgconfig, which contain (duplicate) flags to link the very same library, also in error-prone -L/-l form.

Link with aom before: `/usr/local/lib/libaom.so  -L/usr/local/lib -laom`
Link with aom after: `/usr/local/lib/libaom.so`


vrabaud (approved):
Indeed, there is redundancy and there is nothing else in those flags according to https://aomedia.googlesource.com/aom/+/3bc9ea9e89aee0bc39bdccb591e133d0c4f788bc/build/cmake/pkg_config.cmake#62
So it seems safe to  remove _AOM_LDFLAGS.

AMDmi3:
Merge please?

vrabaud (approved):

PR #1582, opened 2023-09-12
Only ignore reproducible gain map errors

y-guyon:
Out-of-memory problems should still be propagated to the user.

Also clear any diagnostic message in case an error is returned later without setting data->diag (thus returning the wrong message).


maryla-uc (approved):
thanks

PR #1584, opened 2023-09-12
Add avifalphanoispetest for coverage

y-guyon:


tests/data/README.md
```diff
@@ -411,6 +411,14 @@ Box structure of the items in this file:
 [alpha] [alpha]
 ```
 
+### File [alpha_noispe.avif](alpha_noispe.avif)
+
+
+Source: Same pixels as `color_grid_alpha_nogrid.avif` encoded with
+`avifenc --lossless`. Manually edited with a hex editor to remove the alpha
```
wantehchang, line 419: Is it necessary to encode ths image losslessly?

tests/data/README.md
```diff
@@ -411,6 +411,14 @@ Box structure of the items in this file:
 [alpha] [alpha]
 ```
 
+### File [alpha_noispe.avif](alpha_noispe.avif)
```
wantehchang, line 414: Nit/optional: Since we are using an underscore in the file name, it would be clearer to use one more: alpha_no_ispe.avif

tests/data/README.md
wantehchang, line 419: Optional: We could encode the image with cavif-rs, with the fix to avif-serialize (https://github.com/kornelski/avif-serialize/issues/3) reverted. That was the encoder with this issue.

wantehchang (approved):
LGTM.

tests/data/README.md
y-guyon: No, besides making it possible to compare pixel-by-pixel with the other file if it becomes necessary or convenient.

tests/data/README.md
y-guyon: This is for consistency with `color_grid_alpha_nogrid.avif` which is not written `color_grid_alpha_no_grid.avif`.

wantehchang: I see. This is fine. (When I first saw "noispe", it looked like a misspelled "noise").

PR #1587, opened 2023-09-13
Allow reading the gain map metadata without reading the gain map itself.

maryla-uc:
Add 'ignoreGainMapMetadata' to 'avifDecoder'.

### DIFF
--- a/tests/gtest/avifgainmaptest.cc
+++ b/tests/gtest/avifgainmaptest.cc
@@ -358,7 +358,7 @@ TEST(GainMapTest, SequenceNotSupported) {
       << avifResultToString(result) << " " << encoder->diag.error;
 }
 
-TEST(GainMapTest, enableDecodingGainMap) {
+TEST(GainMapTest, IgnoreGainMap) {
   testutil::AvifImagePtr image =
       CreateTestImageWithGainMap(/*base_rendition_is_hdr=*/false);
   ASSERT_NE(image, nullptr);
@@ -390,7 +390,7 @@ TEST(GainMapTest, enableDecodingGainMap) {
   CheckGainMapMetadataMatches(decoded->gainMap.metadata, avifGainMapMetadata());
 }
 
-TEST(GainMapTest, enableDecodingGainMapButReadMetadata) {
+TEST(GainMapTest, IgnoreGainMapButReadMetadata) {
   testutil::AvifImagePtr image =
       CreateTestImageWithGainMap(/*base_rendition_is_hdr=*/false);
   ASSERT_NE(image, nullptr);


include/avif/avif.h
```diff
@@ -1143,11 +1143,20 @@ typedef struct avifDecoder
...
     // of this setting.
     avifBool ignoreGainMap;
+    // Do not read the gain map metadata if present (defaults to AVIF_TRUE).
+    // gainMapPresent is still set if the presence of a gain map is detected, regardless
+    // of this setting.
+    // Gain map metadata is read during avifDecoderParse(). Like Exif and XMP, this data
+    // can be (unfortunately) packed at the end of the file, which will cause
+    // avifDecoderParse() to return AVIF_RESULT_WAITING_ON_IO until it finds it.
+    // If you don't actually use this data, it's best to leave this to AVIF_TRUE (default).
+    avifBool ignoreGainMapMetadata;
```
y-guyon, line 1159: Is there a use case where `ignoreGainMapMetadata && !ignoreGainMap` makes sense? Should there be an error in this case?

tests/gtest/avifgainmaptest.cc
```diff
@@ -387,6 +390,40 @@ TEST(GainMapTest, IgnoreGainMap) {
...
+  ASSERT_EQ(result, AVIF_RESULT_OK)
+      << avifResultToString(result) << " " << decoder->diag.error;
+
+  // Verify that the input and decoded images are close.
+  EXPECT_GT(testutil::GetPsnr(*image, *decoded), 40.0);
+  // Verify that the gain map was detected...
+  EXPECT_TRUE(decoder->gainMapPresent);
+  // ... but not decoded because ignoreGainMap is true by default.
+  EXPECT_EQ(decoded->gainMap.image, nullptr);
+  // Check that the gain map metadata was not populated either.
```
y-guyon, line 422: `was populated`*?

include/avif/avif.h
```diff
@@ -1143,11 +1143,20 @@ typedef struct avifDecoder
 #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
     // This is true when avifDecoderParse() detects a gain map.
     avifBool gainMapPresent;
-    // Do not decode the gain map image or gain map metadata if present (defaults to AVIF_TRUE).
-    // Set this to AVIF_FALSE if you DO want to decode gain map pixels and metadata.
+    // Do not decode the gain map image if present (defaults to AVIF_TRUE).
+    // Set this to AVIF_FALSE if you DO want to decode gain map pixels
```
y-guyon, line 1147: This is counter-intuitive but let's keep that for consistency.

y-guyon (approved):

include/avif/avif.h
maryla-uc: I think it makes sense if you first decode the color+alpha and gain map metadata, then later decode the gain map pixels with a different decoder (I think we will do something like this in Chrome at some point)

include/avif/avif.h
maryla-uc: I agree, I did it this way for consistency but the more I use it, the more I think it's confusing. I changed both of them to "enable..."

y-guyon: Another option but unlikely better: a big enum replacing `ignoreGainMap`, `ignoreGainMapMetadata` and `ignoreColorAndAlpha`.

y-guyon (approved):

include/avif/avif.h
maryla-uc: So like a bit field? Yeah i'm not fully convinced. Maybe if we'd done this from the beginning, with ignoreExif and ignoreXMP as well.

y-guyon: I was thinking about an enum:

```c
enum avifParsingFilter {
  AVIF_PARSE_IMAGE_FEATURES,
  AVIF_PARSE_IMAGE_FEATURES_XMP_EXIF,
  AVIF_PARSE_IMAGE_FEATURES_GAIN_MAP_METADATA,
  AVIF_PARSE_ALL
}
enum avifDecodingFilter {
  AVIF_DECODE_COLOR_ALPHA,
  AVIF_DECODE_GAIN_MAP,
  AVIF_DECODE_ALL
}
```

but a bit field is likely more convenient, and individual fields are better than a bit field in my opinion anyway. So the current PR is fine.

PR #1592, opened 2023-09-18
Allow providing a custom decoder to testutil::DecodeIncrementally.

maryla-uc:
This allows testing decoders with different settings.


y-guyon (approved):
Minor note: The decoder instance being reused could hide bugs that happen only in a fresh decoder.

maryla-uc:
That's a good point. An alternative would be to pass a factory function to create decoders but it's a bit more complicated, and it seems unlikely that there would be an issue.

PR #1593, opened 2023-09-18
Add FuzzTest targets through AVIF_ENABLE_FUZZTEST

y-guyon:
Experimental Linux support only for now.  
AVIF_ENABLE_FUZZTEST requires AVIF_LOCAL_FUZZTEST for now.  
FuzzTest targets are added for transparency and for OSS-Fuzz exposure.  
The dependency on FuzzTest is added for local reproducibility convenience. The CMake part can be improved in further commits.

Not enabled in GitHub CI for now.  
`ctest` currently fails on fuzztest targets with `_NOT_BUILT` and other "no codec" failures.

Note that the new FuzzTest targets are part of tests/gtest and not tests/oss-fuzz for these reasons:
- it avoids duplicating tests/gtest/.clang-format,
- FuzzTest relies on GoogleTest so it makes sense for these targets to be under tests/gtest and among other GoogleTest targets.

It may be confusing if these targets are referenced in OSS-Fuzz configuration.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -43,8 +43,7 @@ if(AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
     enable_language(CXX)
     set(CMAKE_CXX_STANDARD 14)
     add_library(aviftest_helpers OBJECT gtest/aviftest_helpers.cc)
-    target_link_libraries(aviftest_helpers avif_apps)
-    target_link_libraries(aviftest_helpers avif_internal)
+    target_link_libraries(aviftest_helpers avif_apps avif_internal)
 endif()
 
 # Fuzz target without any fuzzing engine dependency. For easy reproduction of oss-fuzz issues.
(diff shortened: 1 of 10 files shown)


tests/CMakeLists.txt
```diff
@@ -249,6 +248,53 @@ else()
...
+        #   FUZZTEST_REPLAY=/path/to/repro_file.test
+        # and running one of the targets below.
+        # See https://github.com/google/fuzztest/blob/main/doc/quickstart-cmake.md
+        add_subdirectory(${CMAKE_SOURCE_DIR}/ext/fuzztest ${CMAKE_SOURCE_DIR}/ext/fuzztest/build.libavif EXCLUDE_FROM_ALL)
+        # FuzzTest bundles GoogleTest so no need for the targets below to link to GTEST_BOTH_LIBRARIES.
+    else()
+        message(FATAL_ERROR "fuzztest: Installed FuzzTest is not supported, please set AVIF_LOCAL_FUZZTEST=ON or AVIF_ENABLE_FUZZTEST=OFF")
+    endif()
+
+    add_executable(avif_fuzztest_dec_incr gtest/avif_fuzztest_helpers.cc gtest/avif_fuzztest_dec_incr.cc)
```
vrabaud, line 269: How about creating a macro? I guess there will be more targets to come.

tests/gtest/avif_fuzztest_dec_incr.cc
```diff
@@ -0,0 +1,98 @@
...
+    // TODO(yguyon): Increase this arbitrary threshold but decode incrementally
+    //               fewer times than as many bytes.
+    if (reference->width * reference->height * data.read_size >
+        8 * 1024 * 1024) {
+      return;
+    }
+    // decodeIncrementally() will fail if there are leftover bytes.
+    const avifRWData encoded_data = {const_cast<uint8_t*>(data.available_bytes),
+                                     data.read_size};
+    // No clue on whether encoded_data is tiled so use a lower bound of a single
```
vrabaud, line 75: Does that comment apply to  8 * 1024 * 1024 above?

tests/gtest/avif_fuzztest_dec_incr.cc
```diff
@@ -0,0 +1,98 @@
+// Copyright 2022 Google LLC
```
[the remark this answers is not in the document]
y-guyon: These were internal files created in 2022 so I just kept the copyright line.

tests/gtest/avif_fuzztest_dec_incr.cc
y-guyon: No, it applies to `max_cell_height` which is a lower bound of `cell_height` arg for `DecodeIncrementally()`.

tests/CMakeLists.txt
```diff
@@ -249,6 +248,53 @@ else()
...
+    gtest_discover_tests(avif_fuzztest_dec_incr)
+    add_test(NAME avif_fuzztest_dec_incr COMMAND avif_fuzztest_dec_incr)
+
+    add_executable(avif_fuzztest_enc_dec gtest/avif_fuzztest_helpers.cc gtest/avif_fuzztest_enc_dec.cc)
+    target_link_libraries(avif_fuzztest_enc_dec PRIVATE aviftest_helpers)
+    link_fuzztest(avif_fuzztest_enc_dec)
+    gtest_discover_tests(avif_fuzztest_enc_dec)
+    add_test(NAME avif_fuzztest_enc_dec COMMAND avif_fuzztest_enc_dec)
+
+    if(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
```
wantehchang, line 281: Should we also check `AVIF_ENABLE_EXPERIMENTAL_AVIR`?

y-guyon: When `coni` is tested in `avif_fuzztest_enc_dec_experimental`, sure. It is not the case yet.

maryla-uc (approved):

PR #1594, opened 2023-09-18
Compare box size and chunks sizes in 'coni' box

y-guyon:
Otherwise `avifParseExtendedMeta()` could be passed missing bytes.

See internal bug 300640326.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3374,6 +3374,11 @@ static avifResult avifParseCondensedImageBox(avifMeta * meta, uint64_t rawOffset
     AVIF_CHECKERR(avifParseCodecConfiguration(&s, &colorCodecConfig, "coni", diag),
                   AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(8) main_item_codec_config[];
 
+    // Make sure all extended_meta, metadata and coded chunks fit into the 'coni' box whose size is rawLen.
+    // There should be no missing nor unused byte.
+    AVIF_CHECKERR(avifROStreamOffset(&s) + extendedMetaSize + iccDataSize + alphaItemDataSize + colorItemDataSize + exifSize + xmpSize == rawLen,
+                  AVIF_RESULT_BMFF_PARSE_FAILED);
+
     // Store and update the offset for the following item extents and properties.
     // The extendedMeta field is parsed after creating the items defined by the CondensedImageBox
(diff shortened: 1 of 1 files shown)


vrabaud (approved):

src/read.c
```diff
@@ -3374,6 +3374,11 @@ static avifResult avifParseCondensedImageBox(avifMeta * meta, uint64_t rawOffset
     AVIF_CHECKERR(avifParseCodecConfiguration(&s, &colorCodecConfig, "coni", diag),
                   AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(8) main_item_codec_config[];
 
+    // Make sure all extended_meta, metadata and coded chunks fit into the 'coni' box whose size is rawLen.
+    // There should be no missing nor unused byte.
+    AVIF_CHECKERR(avifROStreamOffset(&s) + extendedMetaSize + iccDataSize + alphaItemDataSize + colorItemDataSize + exifSize + xmpSize == rawLen,
```
wantehchang, line 3379: Nit: `rawLen` is the size of the stream `s`. So we don't need to use `rawLen` explicitly and can use a suitable avifROStream function instead, e.g., `avifROStreamRemainingBytes(&s) == extendedMetaSize + iccDataSize + alphaItemDataSize + colorItemDataSize + exifSize + xmpSize`.

Also, we should check for integer overflow in the sum `extendedMetaSize + iccDataSize + alphaItemDataSize + colorItemDataSize + exifSize + xmpSize`, unless some higher-level code guarantees that that sum cannot possibly overflow. This can be done in a separate pull request.

src/read.c
```diff
@@ -3535,18 +3540,10 @@ static avifResult avifParseCondensedImageBox(avifMeta * meta, uint64_t rawOffset
     // The ExtendedMetaBox may reuse items and properties created above so it must be parsed last.
 
     if (hasExtendedMeta) {
+        assert(avifROStreamOffset(&s) + extendedMetaSize <= rawLen);
```
wantehchang, line 3543: Similarly, avoid using `rawLen` (the size of stream `s`) expliclty and use a suitable avifROStream function instead. We can say `assert(avifROStreamHasBytesLeft(&s, extendedMetaSize);`.

wantehchang (approved):
LGTM. I suggest two changes to use the size of the stream `s` instead of `rawLen`.

PR #1596, opened 2023-09-18
Expose animation track presence in avifDecoder

### LINKED ISSUE #1595: Add libavif API to allow dectection of animated AVIF
In the current libavif API, there is no simple way to conclusively detect if an AVIF image is still or animated. Some workarounds are:

1) sniff the brands in the ftyp box.
2) check if avifDecoder.imageCount > 1 (for progressive images, we will also have to check for progressiveState)

Add a simple boolean to the avifDecoder struct to expose this information.

vigneshvg:
Add a member `animationTrackPresent` to `avifDecoder` which will be set to true if an animation track is found in the input image.

Fixes #1595

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1160,6 +1160,11 @@ typedef struct avifDecoder
     // Can be useful to decode the gain map image only.
     avifBool ignoreColorAndAlpha;
 #endif
+
+    // This is true when avifDecoderParse() detects an image sequence track in the image. If this is true, the image can be
+    // decoded either as an animated image sequence or as a still image (the primary image item) by setting avifDecoderSetSource
+    // to the appropriate source.
+    avifBool imageSequenceTrackPresent;
 } avifDecoder;
 
 AVIF_API avifDecoder * avifDecoderCreate(void);
--- a/tests/gtest/avifanimationtest.cc
+++ b/tests/gtest/avifanimationtest.cc
@@ -23,11 +23,39 @@ TEST(AvifDecodeTest, AnimatedImage) {
             AVIF_RESULT_OK);
   ASSERT_EQ(avifDecoderParse(decoder.get()), AVIF_RESULT_OK);
   EXPECT_EQ(decoder->alphaPresent, AVIF_FALSE);
+  EXPECT_EQ(decoder->imageSequenceTrackPresent, AVIF_TRUE);
   EXPECT_EQ(decoder->imageCount, 5);
   EXPECT_EQ(decoder->repetitionCount, 0);
   for (int i = 0; i < 5; ++i) {
     EXPECT_EQ(avifDecoderNextImage(decoder.get()), AVIF_RESULT_OK);
   }
+}
+
+TEST(AvifDecodeTest, AnimatedImageWithSourceSetToPrimaryItem) {
+  if (!testutil::Av1DecoderAvailable()) {
+    GTEST_SKIP() << "AV1 Codec unavailable, skip test.";
+  }
+  const char* file_name = "colors-animated-8bpc.avif";
+  testutil::AvifDecoderPtr decoder(avifDecoderCreate(), avifDecoderDestroy);
+  ASSERT_NE(decoder, nullptr);
+  ASSERT_EQ(avifDecoderSetIOFile(decoder.get(),
+                                 (std::string(data_path) + file_name).c_str()),
+            AVIF_RESULT_OK);
+  ASSERT_EQ(
+      avifDecoderSetSource(decoder.get(), AVIF_DECODER_SOURCE_PRIMARY_ITEM),
+      AVIF_RESULT_OK);
+  ASSERT_EQ(avifDecoderParse(decoder.get()), AVIF_RESULT_OK);
+  EXPECT_EQ(decoder->alphaPresent, AVIF_FALSE);
+  EXPECT_EQ(decoder->imageSequenceTrackPresent, AVIF_TRUE);
+  // imageCount is expected to be 1 because we are using primary item as the
+  // preferred source.
+  EXPECT_EQ(decoder->imageCount, 1);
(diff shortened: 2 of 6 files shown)


wantehchang:
Vignesh: I remember an AVIF file with an image sequence is required to have an image item. Could you please add a test that decode an avis (note the "s") file with an `avifDecoderSetSource(decoder, AVIF_DECODER_SOURCE_PRIMARY_ITEM)` call?

Do you think `animationTrackPresent` should be true in this case?

wantehchang:
I tested your pull request. `animationTrackPresent` is true in that case. That is the answer I expected.

tests/gtest/avifanimationtest.cc
```diff
@@ -23,6 +23,7 @@ TEST(AvifDecodeTest, AnimatedImage) {
             AVIF_RESULT_OK);
   ASSERT_EQ(avifDecoderParse(decoder.get()), AVIF_RESULT_OK);
   EXPECT_EQ(decoder->alphaPresent, AVIF_FALSE);
+  EXPECT_EQ(decoder->animationTrackPresent, AVIF_TRUE);
```
wantehchang, line 26: Please add a variant of this test with

```
  ASSERT_EQ(avifDecoderSetSource(decoder.get(),
                                 AVIF_DECODER_SOURCE_PRIMARY_ITEM),
            AVIF_RESULT_OK);
```

We should have in this variant of the test:

```
  EXPECT_EQ(decoder->imageCount, 1);
```

You might want to change the for loop at the end of the test to compare `i < decoder->imageCount` instead of `i < 5`.

include/avif/avif.h
```diff
@@ -1160,6 +1160,9 @@ typedef struct avifDecoder
     // Can be useful to decode the gain map image only.
     avifBool ignoreColorAndAlpha;
 #endif
+
+    // This is true when avifDecoderParse() detects an animation track in the image.
```
y-guyon, line 1164: I suggest adding the following:
```c
// Note that a still image item may be present next to the animation track.
```
I remember `avifenc` generating `meta` and `moov` boxes by default, so I assume this is true.

y-guyon (approved):

tests/gtest/avifanimationtest.cc
vigneshvg: Done. I added a separate test.

include/avif/avif.h
vigneshvg: Yes, an AVIF animated image is always required to have a primary image item (i.e.) it should also be possible to decode the animated image as a still image file.

I have updated the comment to reflect this.

include/avif/avif.h
```diff
@@ -1160,6 +1160,11 @@ typedef struct avifDecoder
     // Can be useful to decode the gain map image only.
     avifBool ignoreColorAndAlpha;
 #endif
+
+    // This is true when avifDecoderParse() detects an animation track in the image. If this is true, the image can be decoded
```
wantehchang, line 1164: Nit: It would be better to consistently use the term "image sequence" for animations in this header. For example, this is how the decoder source is documented:

```
typedef enum avifDecoderSource
{
    // Honor the major brand signaled in the beginning of the file to pick between an AVIF sequence
    // ('avis', tracks-based) or a single image ('avif', item-based). If the major brand is neither
    // of these, prefer the AVIF sequence ('avis', tracks-based), if present.
    AVIF_DECODER_SOURCE_AUTO = 0,

    ...

    // Use the chunks inside primary/aux tracks in the moov block.
    // This is where avifs image sequences store their images.
    AVIF_DECODER_SOURCE_TRACKS

    ...
} avifDecoderSource;
```

include/avif/avif.h
```diff
@@ -1160,6 +1160,11 @@ typedef struct avifDecoder
...
+    // either as an animated image sequence or the primary image item can be retrieved as a still image by setting
+    // avifDecoderSetSource to the appropriate source.
+    avifBool animationTrackPresent;
```
wantehchang, line 1167: Nit: I wonder if we should use the term "sequence" or "image sequence" in the name of this field.

wantehchang (approved):
LGTM. I suggest using the official name "image sequence track" instead of "animation track" in comments and the name of the new field.

include/avif/avif.h
vigneshvg: Done.

include/avif/avif.h
vigneshvg: Yeah good idea. Done.

include/avif/avif.h
```diff
@@ -1160,6 +1160,11 @@ typedef struct avifDecoder
     // Can be useful to decode the gain map image only.
     avifBool ignoreColorAndAlpha;
 #endif
+
+    // This is true when avifDecoderParse() detects an image sequence track in the image. If this is true, the image can be
+    // decoded either as an animated image sequence or the primary image item can be retrieved as a still image by setting
```
wantehchang, line 1165: Nit: The "either" and "or" parts of this sentence are not of the same form.

The "either" part is "as an animated image sequence".

The "or" part is "the primary image item can be retrieved as a still image".

A possible fix is the following:

> If this is true, the image can be decoded either as an animated image sequence or as a still image (the primary image item) by setting avifDecoderSetSource to the appropriate source.

wantehchang (approved):
LGTM.

include/avif/avif.h
wantehchang: Another possible fix is to move "either" earlier:

> If this is true, either the image can be decoded as an animated image sequence or the primary image item can be retrieved as a still image by setting avifDecoderSetSource to the appropriate source.

vigneshvg: Done.

wantehchang (approved):
LGTM.

PR #1597, opened 2023-09-19
Fuzztest: switch raw ptr to unique_ptr now that fuzztest supports it.

maryla-uc:


tests/CMakeLists.txt
```diff
@@ -260,7 +260,7 @@ if(AVIF_ENABLE_GTEST AND AVIF_ENABLE_FUZZTEST)
     if(AVIF_LOCAL_FUZZTEST)
         # Run ext/fuzztest.cmd first.
         # Recommended top-level CMake options:
-        #   -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DAVIF_ENABLE_WERROR=OFF
+        #   -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -AVIF_CODEC_DAV1D=ON -DAVIF_ENABLE_WERROR=OFF
```
y-guyon, line 263: `-DAVIF_CODEC_DAV1D=ON`  
Why is that?

y-guyon (approved):

tests/CMakeLists.txt
maryla-uc: Because of https://github.com/AOMediaCodec/libavif/blob/40988c29d6bce91e762dc8b58bbac27a2a164eb2/tests/gtest/avif_fuzztest_enc_dec.cc#L55
If dav1d is requested but not available, the test will fail. Of course we can change this and only use AVIF_CODEC_CHOICE_AUTO everywhere. This is just the way it was written originally.

PR #1601, opened 2023-09-20
Add fuzzer test for encoding/decoding of a grid with gain maps.

maryla-uc:
Move some utilities to aviftest_helpers.h/.cc

### DIFF
--- a/tests/gtest/avif_fuzztest_enc_dec_incr_experimental.cc
+++ b/tests/gtest/avif_fuzztest_enc_dec_incr_experimental.cc
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2023 Google LLC
 // SPDX-License-Identifier: BSD-2-Clause
 // Encode a fuzzed image split into a grid and decode it incrementally.
 // Compare the output with a non-incremental decode.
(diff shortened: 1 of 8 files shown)


tests/gtest/avif_fuzztest_enc_dec_incr_experimental.cc
```diff
@@ -0,0 +1,127 @@
...
+      std::min(image->height, grid_rows * cell_height);
+
+  const avifImage* gain_map = image->gainMap.image;
+  if (gain_map != nullptr) {
+    std::vector<AvifImagePtr> gain_map_cells =
+        ImageToGrid(gain_map, grid_cols, grid_rows);
+    if (gain_map_cells.empty()) return;
+    ASSERT_EQ(gain_map_cells.size(), cells.size());
+    for (size_t i = 0; i < gain_map_cells.size(); ++i) {
+      cells[i]->gainMap.image = gain_map_cells[i].release();
```
y-guyon, line 52: What happens if the gain map has different dimensions than the image? Are the cells still valid?  
This is mostly by curiosity, for fuzzing it does not matter.

tests/gtest/aviftest_helpers.cc
```diff
@@ -501,5 +501,85 @@ avifIO* AvifIOCreateLimitedReader(avifIO* underlyingIO, uint64_t clamp) {
...
+      }
+    }
+  }
+  return true;
+}
+
+bool IsOpaque(const avifImage* image) {
+  return avifImageUsesU16(image) ? IsOpaque<uint16_t>(image)
+                                 : IsOpaque<uint8_t>(image);
+}
```
y-guyon, line 527: `avifImageIsOpaque()` can be used instead

y-guyon (approved):
Thanks for adding tests.

tests/gtest/aviftest_helpers.cc
maryla-uc: Fixed, thanks.

tests/gtest/avif_fuzztest_enc_dec_incr_experimental.cc
y-guyon: Thanks for the explanation. Note that the current change does not test mismatched `grid_cols` or `grid_rows` and `colorGrid+gainMapNoGrid` or `colorNoGrid+gainMapGrid` but that is fine.

y-guyon (approved):

PR #1605, opened 2023-09-20
ext: Add libyuv_android.sh

vigneshvg:
Enable building the android library with libyuv for fast YUV to RGB conversion.

It also updates the CI workflows to test the new code.

### DIFF
--- a/android_jni/README.md
+++ b/android_jni/README.md
@@ -51,7 +51,19 @@ Update [CMakeLists.txt](avifandroidjni/src/main/jni/CMakeLists.txt) as follows:
  * Set `AVIF_LOCAL_LIBGAV1` and `AVIF_CODEC_LIBGAV1` to `OFF`.
  * Set `AVIF_LOCAL_DAV1D` and `AVIF_CODEC_DAV1D` to `ON`.
 
-Step 4 - Build the JNI Wrapper and generate the AAR package
+Step 4 - Checkout and build libyuv
+
+```
+$ cd ext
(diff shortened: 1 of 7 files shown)


y-guyon (approved):

android_jni/README.md
```diff
@@ -51,7 +51,19 @@ Update [CMakeLists.txt](avifandroidjni/src/main/jni/CMakeLists.txt) as follows:
...
-Step 4 - Build the JNI Wrapper and generate the AAR package
+Step 4 - Checkout and build libyuv
+
+```
+$ cd ext
+$ ./libyuv_android.sh "${ANDROID_NDK_HOME}"
+$ cd ..
+```
+
+If you do not want to use libyuv, then update
```
wantehchang, line 62: Nit: libavif always tries to use libyuv if it is available and newer than a particular version. So `AVIF_LOCAL_LIBYUV` only controls whether libavif searches for libyuv as an installed package or locally in the ext/ directory.

If libyuv cannot be installed as a package in the Android build environment, then `AVIF_LOCAL_LIBYUV` is equivalent to whether you want to use libyuv.

vigneshvg: Yeah there is no notion of installing as a "package" in the android build environment. You will have to build each dependency (or provide a pre-built .so) and it has to be visible from cmake targets. So unsetting `AVIF_LOCAL_LIBYUV` is equivalent to telling the android build system not to use libyuv as a dependency.

PR #1608, opened 2023-09-21
Add an option to disable threading.

vrabaud:
Also use Threads::Threads independently from the platform.
The name `AVIF_USE_THREAD` is inspired by `WEBP_USE_THREAD`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,9 +352,9 @@ endif()
 
 if(AVIF_USE_THREAD)
     # Find out if we have threading available
+    set(THREADS_PREFER_PTHREAD_FLAG ON)
     find_package(Threads QUIET)
     if(TARGET Threads::Threads)
-        set(THREADS_PREFER_PTHREAD_FLAG ON)
         set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} Threads::Threads)
     endif()
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ set(LIBRARY_SOVERSION ${LIBRARY_VERSION_MAJOR})
 option(BUILD_SHARED_LIBS "Build shared avif library" ON)
 
 option(AVIF_ENABLE_WERROR "Treat all compiler warnings as errors" ON)
+option(AVIF_USE_THREAD "Enable threading support" ON)
 
 option(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R "Enable experimental YCgCo-R matrix code" OFF)
 option(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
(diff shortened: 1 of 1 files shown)


y-guyon (approved):

CMakeLists.txt
```diff
@@ -346,10 +347,16 @@ endif()
...
 if(UNIX)
+    set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} m)
+endif()
+
+if(AVIF_USE_THREAD)
     # Find out if we have threading available
-    set(THREADS_PREFER_PTHREAD_FLAG ON)
-    find_package(Threads)
-    set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} m Threads::Threads)
+    find_package(Threads QUIET)
```
kmilos, line 355: [On MINGW, this has potentially different behaviour depending on the presence of `winpthreads`](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindThreads.cmake):

The target will be available always, and `CMAKE_USE_WIN32_THREADS_INIT` will always be defined. However, if `winpthreads` is also installed, `CMAKE_THREAD_LIBS_INIT` will also not be empty and `-lpthread` or `-pthread` will be added to the target unnecessarily. (I guess the linker might be smart enough to figure out there are no symbols used though and ignore it?)

CMakeLists.txt
```diff
@@ -346,10 +347,16 @@ endif()
...
+    if(TARGET Threads::Threads)
+        set(THREADS_PREFER_PTHREAD_FLAG ON)
```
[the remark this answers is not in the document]
vrabaud: Indeed, this is used by FindThreads, it is a parameter that has to be set before: https://cmake.org/cmake/help/latest/module/FindThreads.html

CMakeLists.txt
vrabaud: I'd says so yes.

kmilos: I just checked this - in the presence of `pthread.h` from `winpthreads`, the find module actually succeeds compiling the test code and (wrongly?) sets `CMAKE_HAVE_LIBC_PTHREAD` so `CMAKE_THREAD_LIBS_INIT` is still empty as a lucky side effect on MINGW:
```
-- CMAKE_THREAD_LIBS_INIT:
-- CMAKE_USE_WIN32_THREADS_INIT:1
-- CMAKE_USE_PTHREADS_INIT:1
```

kmilos: Just for completeness: although not affecting libavif because of the native only Win32 implantation, CMake's detection is still not 100% correct on MINGW:

https://gitlab.kitware.com/cmake/cmake/-/issues/6399
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7179

If `winpthreads` is present and preferred, it should be adding `-pthread` or non-empty `CMAKE_THREAD_LIBS_INIT` to the imported target.

wantehchang (changes requested):
Hi Vincent,

libavif uses threads directly in src/reformat.c. (You can find it by searching for "pthread.h" in the src directory.) So if we add the `AVIF_USE_THREAD` option, it needs to affect the threading code in src/reformat.c.

But I'd like to take a step back and ask if libavif needs the `AVIF_USE_THREAD` option. I think it is fine for libavif to assume threads are available. Is there a customer request for a no-threads build of libavif? If so, why do they need to disable threading?

vrabaud:
@wantehchang , indeed src/reformat.c would need to change: I thought threading was only for UNIX, but we use the default windows one otherwise.
The main use case is when you do a lot of decoding at the same time (e.g. browser) or encoding (e.g. CDN), it then becomes more efficient to not split your image over different cores. That was the case for WebP at least.
For src/reformat.c, we can impose the number of threads to 1. Let me check.

wantehchang:
Vincent: Thanks for the reply. The main use case you described does not require a build option to disable threading. All the `maxThreads` struct members in avif/avif.h have a default value of 1 or 0 (which is treated the same as 1). So a client that wants to do single-threaded decoding or encoding can simply rely on the default value of `maxThreads` or can explicitly set `maxThreads` to 1.

A build option to disable threading is typically intended for two cases:
- an operating system that doesn't have threads, or
- an application doesn't want any of the complexity and overhead of mulitithreading support. In this case, the Standard C library needs to have both a single-threaded and a multithreaded version, e.g., the C runtime libraries in MSVC.

Every new build option increases the maintenance burden. These two cases don't seem important enough for libavif to support officially.

vrabaud:
Thx for the explanation @wantehchang , let's close this for now then.

PR #1609, opened 2023-09-21
Define macros for tests and fuzztests.

vrabaud:


tests/CMakeLists.txt
```diff
@@ -60,6 +60,19 @@ add_test(NAME repro_avif_decode_fuzzer COMMAND repro_avif_decode_fuzzer
...
 # GoogleTest
 
+# Adds a gtest from file TEST_NAME.cc located in the gtest folder. Extra arguments
+# are considered as extra linked libraries.
+macro(add_avif_gtest TEST_NAME USE_DATA)
+    add_executable(${TEST_NAME} gtest/${TEST_NAME}.cc)
+    target_include_directories(${TEST_NAME} PRIVATE ${GTEST_INCLUDE_DIRS})
+    target_link_libraries(${TEST_NAME} PRIVATE aviftest_helpers ${GTEST_BOTH_LIBRARIES} ${ARGN})
+    if(${USE_DATA})
+        add_test(NAME ${TEST_NAME} COMMAND ${TEST_NAME} ${CMAKE_CURRENT_SOURCE_DIR}/data/)
```
y-guyon, line 70: I wonder if we cannot pass `${CMAKE_CURRENT_SOURCE_DIR}/data/` everywhere, to simplify. Maybe it will make the default gunit_main handler fail because of an unrecognized flag.

Otherwise maybe replace `0` and `1` by `USE_DATA` or something more explicit below.

tests/CMakeLists.txt
```diff
@@ -60,6 +60,19 @@ add_test(NAME repro_avif_decode_fuzzer COMMAND repro_avif_decode_fuzzer
 ################################################################################
 # GoogleTest
 
+# Adds a gtest from file TEST_NAME.cc located in the gtest folder. Extra arguments
+# are considered as extra linked libraries.
+macro(add_avif_gtest TEST_NAME USE_DATA)
+    add_executable(${TEST_NAME} gtest/${TEST_NAME}.cc)
+    target_include_directories(${TEST_NAME} PRIVATE ${GTEST_INCLUDE_DIRS})
+    target_link_libraries(${TEST_NAME} PRIVATE aviftest_helpers ${GTEST_BOTH_LIBRARIES} ${ARGN})
```
y-guyon, line 68: I am surprised this works everywhere: `gtest_main` should not be linked in tests already defining a custom `main()`.

y-guyon (approved):

tests/CMakeLists.txt
vrabaud: Explained here: https://github.com/google/googletest/issues/2674

tests/CMakeLists.txt
vrabaud: I created 2 macros

y-guyon (approved):

tests/CMakeLists.txt
wantehchang: It would be good to avoid linking tests defining its own `main()` function with `gtest_main`.

wantehchang (commented):
Thanks you for adding the macros. This is a supplemental review.

tests/CMakeLists.txt
vrabaud: Exactly, that is why the macto add_avif_gtest_with_data was created below which links against `GTEST_LIBRARIES` and not `GTEST_BOTH_LIBRARIES`.
For now, tests defining `main` are called with a data argument.
If needs be for tests with `main` but different arguments, we can create a third macro.

wantehchang: Thank you for the explanation. I didn't know the only reason our tests need to define their own main() function is to take a data argument. The current two macros are good.

PR #1614, opened 2023-09-21
android_jni: Use %u for unsigned int

vigneshvg:
Also use \0 for null terminating a string instead of 0.


wantehchang (approved):
LGTM. Thanks.

PR #1615, opened 2023-09-22
Cancel previous CI workflows when pushing new changes, except on main.

maryla-uc:
As suggested by @kmilos in https://github.com/AOMediaCodec/libavif/pull/1530#issuecomment-1697404965
This avoids wasting resources.

The only drawback is that Github sends an email notification for each cancelled job, see https://github.com/orgs/community/discussions/13015
Gmail users can create a filter to ignore those emails.

### DIFF
--- a/.github/workflows/ci-android-emulator-tests.yml
+++ b/.github/workflows/ci-android-emulator-tests.yml
@@ -10,6 +10,11 @@ on:
 permissions:
   contents: read
 
+# Cancel the workflow if a new one is triggered from the same PR, branch, or tag, except on main.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
 jobs:
   build-and-run-android-jni-tests:
     runs-on: macos-latest
--- a/.github/workflows/ci-android-jni.yml
+++ b/.github/workflows/ci-android-jni.yml
@@ -4,6 +4,11 @@ on: [push, pull_request]
 permissions:
   contents: read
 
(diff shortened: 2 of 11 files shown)


y-guyon (approved):

.github/workflows/ci-android-emulator-tests.yml
```diff
@@ -10,6 +10,11 @@ on:
 permissions:
   contents: read
 
+# Cancel the workflow if a new one is triggered from the same PR, branch, or tag, except on main.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
kmilos, line 15: In some projects we found we had to add `github.ref || github.run_id` but can't remember why exactly now...

maryla-uc: Thanks for the input.

According to the [Github docs](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value) you should provide fallback with "||" if using a property that is not always defined.

And according to the [docs here](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), `github.ref` should be defined at least for "push" and "pull_request" events (which we use for CI workflows), if I understand the docs correctly. It also says "This is only set if a branch or tag is available for the event type" but I don't see how that could happen for a push or pull request.

Maybe there are some other events where it doesn't work but it seems fine here. Or we'll fix it if anything breaks ;)

kmilos: > Maybe there are some other events where it doesn't work but it seems fine here. Or we'll fix it if anything breaks ;)

Yeah, it could've been for the occasional manually triggered run...

PR #1622, opened 2023-09-25
Split avifReformatState into an RGB part and a YUV part.

maryla-uc:
This refactor will be useful for some gain map code later on (creating/applying gain maps).

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -158,33 +158,42 @@ typedef enum avifAlphaMultiplyMode
     AVIF_ALPHA_MULTIPLY_MODE_UNMULTIPLY
 } avifAlphaMultiplyMode;
 
-typedef struct avifReformatState
+typedef struct avifRGBSpaceInfo
+{
+    uint32_t channelBytes;
+    uint32_t pixelBytes;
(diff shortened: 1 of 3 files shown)


vrabaud (approved):

include/avif/internal.h
```diff
@@ -158,33 +158,42 @@ typedef enum avifAlphaMultiplyMode
     AVIF_ALPHA_MULTIPLY_MODE_UNMULTIPLY
 } avifAlphaMultiplyMode;
 
-typedef struct avifReformatState
+typedef struct avifRGBSpaceInfo
```
wantehchang, line 161: Nit: It is not clear what "space" means here. Especially in the `avifRGBSpaceInfo` struct, whose members are mostly sizes and offsets, one may think "space" means "buffer space". But I guess "space" means "color space", right?

This can be addressed by adding a comment or renaming the struct, e.g., `avifRGBInfo`, `avifRGBColorInfo`, `avifRGBColorSpace`, etc.

maryla-uc: Indeed I meant "color space". I will clarify.

PR #1623, opened 2023-09-25
Rename avifRGBSpaceInfo to avifRGBColorSpaceInfo.

maryla-uc:
Same for YUV.
Move 'mode' to avifYUVColorSpaceInfo.
Change argument of avifReformatStateYToUNorm and avifReformatStateUVToUNorm to avifYUVColorSpaceInfo.

The change of argument type in avifReformatStateYToUNorm and avifReformatStateUVToUNorm is in preparation for gain map code.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -21,7 +21,7 @@ struct YUVBlock
     float v;
 };
 
-static avifBool avifGetRGBSpaceInfo(const avifRGBImage * rgb, avifRGBSpaceInfo * info)
+static avifBool avifGetRGBSpaceInfo(const avifRGBImage * rgb, avifRGBColorSpaceInfo * info)
 {
     if ((rgb->depth != 8) && (rgb->depth != 10) && (rgb->depth != 12) && (rgb->depth != 16)) {
         return AVIF_FALSE;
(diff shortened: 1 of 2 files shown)


vrabaud (approved):
Maybe rename avifReformatState into avifReformatInfo? As you told me, it is definitely not a state.

src/reformat.c
```diff
@@ -21,7 +21,7 @@ struct YUVBlock
     float v;
 };
 
-static avifBool avifGetRGBSpaceInfo(const avifRGBImage * rgb, avifRGBSpaceInfo * info)
+static avifBool avifGetRGBSpaceInfo(const avifRGBImage * rgb, avifRGBColorSpaceInfo * info)
```
wantehchang, line 24: Nit: `avifGetRGBSpaceInfo` and `avifGetYUVSpaceInfo` probably shoud be renamed to match the new struct names.

src/reformat.c
```diff
@@ -194,33 +194,32 @@ static avifBool avifPrepareReformatState(const avifImage * image, const avifRGBI
...
+static int avifReformatStateYToUNorm(avifYUVColorSpaceInfo * info, float v)
 {
-    int unorm = (int)avifRoundf(v * state->yuv.rangeY + state->yuv.biasY);
-    return AVIF_CLAMP(unorm, 0, state->yuv.maxChannel);
+    int unorm = (int)avifRoundf(v * info->rangeY + info->biasY);
+    return AVIF_CLAMP(unorm, 0, info->maxChannel);
 }
 
-static int avifReformatStateUVToUNorm(avifReformatState * state, float v)
+static int avifReformatStateUVToUNorm(avifYUVColorSpaceInfo * info, float v)
```
wantehchang, line 203: Nit: These two functions, `avifReformatStateYToUNorm` and `avifReformatStateUVToUNorm`, have objected-oriented style names -- the function names start with the type of the first parameter. To preserve this naming convention (commonly used in libavif), we would need to rename them `avifYUVColorSpaceInfoYToUNorm` and `avifYUVColorSpaceInfoUVToUNorm`. But those names become quite long.

PR #1625, opened 2023-09-27
Validate some sizes avifParseCondensedImageBox

wantehchang:
avifROStreamReadVarInt() may read a uint32_t value as large as UINT32_MAX. Check the uint32_t variables before incrementing them.

Declare `offset` as uint64_t instead of size_t, because it is assigned to uint64_t struct members.


src/read.c
```diff
@@ -3409,7 +3416,7 @@ static avifResult avifParseCondensedImageBox(avifMeta * meta, uint64_t rawOffset
     // Store and update the offset for the following item extents and properties.
     // The extendedMeta field is parsed after creating the items defined by the CondensedImageBox
     // so the stream s cannot be used for keeping track of the position.
-    size_t offset = (size_t)rawOffset + avifROStreamOffset(&s) + (size_t)extendedMetaSize;
+    uint64_t offset = rawOffset + avifROStreamOffset(&s) + extendedMetaSize;
```
wantehchang, line 3419: `rawOffset` is `uint64_t`.

`avifROStreamOffset(&s)` is `size_t`, which is an unsigned 32-bit or 64-bit type.

`extendedMetaSize` is `uint32_t`.

So the two additions in this line will be done using `uint64_t`.

Note that in theory, these two additions may potentially overflow. I didn't address that in this pull request.

wantehchang (commented):
Yannis: I spent about an hour and didn't finish the validation of the sizes and offsets in `avifParseCondensedImageBox()`. It would be good if you could pick up where I left off.

y-guyon (approved):
Thanks

src/read.c
wantehchang: I checked the rest of the `avifParseCondensedImageBox()` function, including the two additions in this line. I believe there is no overflow now.

This line is safe because `avifROStreamOffset(&s) + extendedMetaSize` must be <= `rawLen`, and the caller of `avifParseCondensedImageBox()` has checked that `rawOffset + rawLen` won't overflow.

PR #1626, opened 2023-09-27
Fix comment.

maryla-uc:


y-guyon (approved):
(Note: one can include `[skip ci]` in a commit description to skip the GitHub workflows)

PR #1629, opened 2023-09-28
Remove gtest_discover_tests() from CMakeLists.txt

maryla-uc:
It doesn't seem necessary, makes tests appear twice, with one version sometimes failing supposedly with a double free().

Rename some fuzztest test cases for clarity/consistency.


vrabaud (approved):

tests/CMakeLists.txt
```diff
@@ -174,7 +174,6 @@ if(AVIF_ENABLE_GTEST AND AVIF_ENABLE_FUZZTEST)
         target_include_directories(${TEST_NAME} PRIVATE ${GTEST_INCLUDE_DIRS})
         target_link_libraries(${TEST_NAME} PRIVATE aviftest_helpers ${ARGN})
         link_fuzztest(${TEST_NAME})
-        gtest_discover_tests(${TEST_NAME})
```
wantehchang, line 177: `gtest_discover_tests()` was originally added in https://github.com/AOMediaCodec/libavif/pull/1593. Might want to ask Yannis why he added it.

maryla-uc: We talked about it, he said it was recommended in the docs.
I'm not sure why some of the tests added by gtest_discover_tests() fail. I removed almost all of the code from a test and it was still failing, so it's not something in the test code, but probably something in our set up (maybe some kind of double initialization?)
I don't think it's worth investigating as it works perfectly fine without gtest_discover_tests.

y-guyon: >> gtest_discover_tests() was originally added in https://github.com/AOMediaCodec/libavif/pull/1593. Might want to ask Yannis why he added it.

> We talked about it, he said it was recommended in the docs.

See https://github.com/google/fuzztest/blob/main/doc/quickstart-cmake.md#create-and-run-a-fuzz-test.

maryla-uc: I guess you are supposed to use that instead of add_test(). We were doing both.

PR #1630, opened 2023-09-28
Add a simple fuzzer that decodes (non incrementally) a bitstream.

maryla-uc:
Uses the avif files in libavif/tests/data as seeds.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -196,6 +196,7 @@ if(AVIF_ENABLE_GTEST AND AVIF_ENABLE_FUZZTEST)
         )
     endif()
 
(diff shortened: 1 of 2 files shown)


y-guyon (approved):

jzern:
Can oss-fuzz run these?

https://github.com/AOMediaCodec/libavif/blob/main/tests/oss-fuzz/build.sh

From:

https://github.com/google/oss-fuzz/blob/master/projects/libavif/build.sh

jzern:
I'm working bottom up through old notifications, maybe #1633 is doing some of that.

vrabaud:
@jzern , yes #1633 does that. Not merging it yet, will explain there.

PR #1632, opened 2023-09-29
Move the building script from oss-fuzz to libavif.

vrabaud:
This will ease the evolution of the fuzz targets by submitting everything locally, and not in the oss-fuzz repo.

This PR goes hand in hand withL https://github.com/google/oss-fuzz/pull/11044

### DIFF
--- a/tests/oss-fuzz/build.sh
+++ b/tests/oss-fuzz/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-# Copyright 2023 Google Inc.
+# Copyright 2020 Google Inc.
 #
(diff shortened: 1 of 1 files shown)


tests/oss-fuzz/build.sh
y-guyon, line 1: This file is taken almost verbatim from https://github.com/google/oss-fuzz/blob/master/projects/libavif/build.sh. It should keep the original copyright year.

vrabaud:
This is just a basic copy of the current upstream commands. This script will then be updated to include fuzztests.

y-guyon (approved):

PR #1633, opened 2023-09-29
Use fuzztest targets for oss-fuzz.

vrabaud:
To test locally, follow the instructions at the beginning of the file.

Also: Limit CIFuzz to when a PR is merged
This can be revisited at some point but first:
- let's make sure new oss-fuzz targets are built and created
on the site
- let's make sure we fix all fuzzer bugs if any
- let's see how long the job really takes

### DIFF
--- a/tests/oss-fuzz/build.sh
+++ b/tests/oss-fuzz/build.sh
@@ -19,21 +19,76 @@
 # https://github.com/google/oss-fuzz/blob/master/projects/libavif/build.sh
 # It builds the different fuzz targets.
 
-# build dav1d
-cd ext && bash dav1d.cmd && bash libyuv.cmd && cd ..
+# To test changes to this file:
+# - make changes and commit to your REPO
+# - run:
+#     git clone --depth=1 git@github.com:google/oss-fuzz.git
+#     cd oss-fuzz
+# - modify projects/libavif/Dockerfile to point to your REPO
+# - run:
+#     python3 infra/helper.py build_image libavif
+#     # enter 'y' and wait for everything to be downloaded
+# - run:
+#     python3 infra/helper.py build_fuzzers --sanitizer address libavif
+#     # wait for the tests to be built
+# And then run the fuzzer locally, for example:
+#     python3 infra/helper.py run_fuzzer libavif \
+#     avif_fuzztest_enc_dec_incr@EncodeDecodeAvifFuzzTest.EncodeDecodeGridValid \
+#     --sanitizer address
+
+ln -s $SRC/fuzztest $SRC/libavif/ext/fuzztest
+mkdir $SRC/libavif/ext/fuzztest/build.libavif
+cd $SRC/libavif/ext/fuzztest/build.libavif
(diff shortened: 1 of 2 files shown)


tests/oss-fuzz/build.sh
```diff
@@ -19,19 +19,76 @@
...
 
 # build libavif
 mkdir build
 cd build
-cmake -G Ninja -DBUILD_SHARED_LIBS=0 -DAVIF_CODEC_DAV1D=1 -DAVIF_LOCAL_DAV1D=1 ..
+cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=ON -DAVIF_CODEC_DAV1D=ON \
+      -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DAVIF_ENABLE_WERROR=OFF \
+      -DAVIF_LOCAL_AOM=ON -DAVIF_LOCAL_DAV1D=ON -DAVIF_LOCAL_FUZZTEST=ON \
+      -DAVIF_LOCAL_GTEST=ON -DAVIF_LOCAL_JPEG=ON -DAVIF_LOCAL_ZLIBPNG=ON \
+      -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON -DAVIF_ENABLE_FUZZTEST=ON ..
```
y-guyon, line 57: Note: No libyuv, no libsharpyuv for now.

tests/oss-fuzz/build.sh
```diff
@@ -19,19 +19,76 @@
...
+
+# build fuzztests
+# The following is taken from https://github.com/google/oss-fuzz/blob/31ac7244748ea7390015455fb034b1f4eda039d9/infra/base-images/base-builder/compile_fuzztests.sh#L59
+# Iterate the fuzz binaries and list each fuzz entrypoint in the binary. For
+# each entrypoint create a wrapper script that calls into the binaries the
+# given entrypoint as argument.
+# The scripts will be named:
+# {binary_name}@{fuzztest_entrypoint}
+FUZZ_TEST_BINARIES_OUT_PATHS=`ls ./tests/avif_fuzztest_*`
+#cp ./bin/opencv_fuzz_* $OUT/
```
y-guyon, line 74: Leftover?

y-guyon (approved):

y-guyon (approved):

tests/oss-fuzz/build.sh
vrabaud: Thx, fixed and added

tests/oss-fuzz/build.sh
vrabaud: removed

vrabaud:
Not merging yet because that increases the CI for fuzz because it launches this script, which builds a looot of dependencies. Looking into caching that,

vrabaud:
The build now "only" takes 10 minutes. Still, I am removing it from CIFuzz for now because it hangs due to the targets not being officially create on oss-fuzz. Let's wait for them to be there, and we can re-enable the CI.

vrabaud:
Link to official doc on how to only allow CI for merges: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges

let's keep the concurrency for now in case we re-enable the CI per open pull request.

y-guyon (approved):

PR #1634, opened 2023-09-29
fix: unused parameter error in third_party libyuv with release build 

fdintino:
In several functions in libyuv's scale.c, the only reference to the parameter `src_width` is an `assert` statement. When `NDEBUG` is defined (as it is on some platforms when compiling with `CMAKE_BUILD_TYPE=Release`) the assert condition is removed from the source and an unused-parameter warning/error is raised.

There were not any tests that compiled the third_party/libyuv source files. I added such a test in a separate commit to demonstrate the issue. Without the fix, [the new job fails](https://github.com/fdintino/libavif/actions/runs/6354750162/job/17261762381); with it the [job succeeds](https://github.com/fdintino/libavif/actions/runs/6354904451/job/17262117947).

### DIFF
--- a/.github/workflows/ci-unix-shared-local.yml
+++ b/.github/workflows/ci-unix-shared-local.yml
@@ -21,6 +21,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
+        libyuv: [ON, OFF]
+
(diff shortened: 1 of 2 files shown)


vrabaud:
Thx ! I just faced it this afternoon and had to go with -DAVIF_ENABLE_WERROR=OFF as other third dependencies had too many warnings too.

PR #1640, opened 2023-10-02
Bump ext/ComplianceWarden to e269736

y-guyon:
That commit contains the fix for https://github.com/gpac/ComplianceWarden/issues/68.

### DIFF
--- a/ext/compliance_warden.sh
+++ b/ext/compliance_warden.sh
@@ -5,7 +5,8 @@
 
 set -e
 
-git clone -b v33 --depth 1 https://github.com/gpac/ComplianceWarden.git
+git clone https://github.com/gpac/ComplianceWarden.git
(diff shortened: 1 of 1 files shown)


ext/compliance_warden.sh
```diff
@@ -5,7 +5,8 @@
 
 set -e
 
-git clone -b v33 --depth 1 https://github.com/gpac/ComplianceWarden.git
+git clone https://github.com/gpac/ComplianceWarden.git
+cd ComplianceWarden && git checkout e26973641f747cf3282004dcfb0747881207e748 && cd ..
```
wantehchang, line 9: Optional: We can stay inside the `ComplianceWarden` directory, do all the work, and then `cd ..` at the end of the script. This allows us to omit `ComplianceWarden/` from all the pathnames.

This can be done separately because it is unrelated to the version bump.

wantehchang (approved):
LGTM.

PR #1641, opened 2023-10-03
Add fuzzing for animation encoding then decoding.

maryla-uc:
Update version in aom.cmd to avoid a bug for high bit depth images.

### DIFF
--- a/tests/gtest/avif_fuzztest_enc_dec_anim.cc
+++ b/tests/gtest/avif_fuzztest_enc_dec_anim.cc
@@ -58,7 +58,8 @@ void EncodeDecodeAnimation(AvifImagePtr frame,
   ASSERT_EQ(result, AVIF_RESULT_OK)
       << avifResultToString(result) << " " << decoder->diag.error;
 
-  if (decoder->requestedSource == AVIF_DECODER_SOURCE_PRIMARY_ITEM) {
+  if (decoder->requestedSource == AVIF_DECODER_SOURCE_PRIMARY_ITEM ||
+      num_frames == 1) {
     ASSERT_EQ(decoder->imageCount, 1);
   } else {
     ASSERT_EQ(decoder->imageCount, num_frames);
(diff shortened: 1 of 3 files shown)


tests/gtest/avif_fuzztest_enc_dec_anim.cc
```diff
@@ -0,0 +1,100 @@
...
+constexpr int kMaxFrames = 10;
+
+FUZZ_TEST(EncodeDecodeAvifFuzzTest, EncodeDecodeAnimation)
+    .WithDomains(ArbitraryAvifImage(),
+                 fuzztest::VectorOf(
+                     fuzztest::StructOf<FrameOptions>(
+                         /*duration=*/fuzztest::InRange(1, 10),
+                         fuzztest::BitFlagCombinationOf<avifAddImageFlags>(
+                             {AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME})))
+                     .WithMinSize(2)
```
y-guyon, line 92: Why not `.WithMinSize(1)`?

tests/gtest/avif_fuzztest_enc_dec_anim.cc
```diff
@@ -0,0 +1,100 @@
...
+
+constexpr int kMaxFrames = 10;
+
+FUZZ_TEST(EncodeDecodeAvifFuzzTest, EncodeDecodeAnimation)
+    .WithDomains(ArbitraryAvifImage(),
+                 fuzztest::VectorOf(
+                     fuzztest::StructOf<FrameOptions>(
+                         /*duration=*/fuzztest::InRange(1, 10),
+                         fuzztest::BitFlagCombinationOf<avifAddImageFlags>(
+                             {AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME})))
```
y-guyon, line 91: Why not `AVIF_ADD_IMAGE_FLAG_NONE` too?

y-guyon (approved):

tests/gtest/avif_fuzztest_enc_dec_anim.cc
maryla-uc: I thought 1 frame was not really an animation but I guess it doesn't hurt to test it.

tests/gtest/avif_fuzztest_enc_dec_anim.cc
maryla-uc: Added a comment. BitFlagCombinationOf() does not accept a flag whose value is 0, but 0 is always part of the set anyway.

PR #1644, opened 2023-10-03
Always use AUTO, AOM and DAV1D as codec choices in fuzzers.

maryla-uc:
AUTO defaults to DAV1D if available, so adding AOM explicitly allows the libaom decoder to be exercised as well.


y-guyon (approved):
Thanks!

PR #1645, opened 2023-10-03
Fix ArbitraryAvifDecoder() without EXP_GAIN_MAP

y-guyon:
The API changed in https://github.com/AOMediaCodec/libavif/pull/1644.


maryla-uc (approved):
Indeed, thanks for the fix.

PR #1649, opened 2023-10-04
Remove legacy_stdio_definitions.lib dep with mingw

### LINKED ISSUE #1647: mingw-w64: legacy_stdio_definitions.lib does not exist
this library doesn't exist with mingw-w64 as far as I can tell, and causes a linking error:

```bash
/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.2.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -llegacy_stdio_definitions: No such file or directory
```

vrabaud:
This fixes #1647

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -523,7 +523,7 @@ if(AVIF_CODEC_RAV1E)
 
     # Unfortunately, rav1e requires a few more libraries
     if(WIN32)
-        set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ntdll.lib advapi32.lib userenv.lib ws2_32.lib bcrypt.lib msvcrt.lib)
+        set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ntdll.lib userenv.lib ws2_32.lib bcrypt.lib)
     elseif(UNIX AND NOT APPLE)
         set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ${CMAKE_DL_LIBS}) # for backtrace
     endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -523,7 +523,7 @@ if(AVIF_CODEC_RAV1E)
 
     # Unfortunately, rav1e requires a few more libraries
     if(WIN32)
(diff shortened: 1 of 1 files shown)


y-guyon (approved):

CMakeLists.txt
```diff
@@ -523,7 +523,7 @@ if(AVIF_CODEC_RAV1E)
 
     # Unfortunately, rav1e requires a few more libraries
     if(WIN32)
-        set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} legacy_stdio_definitions.lib ntdll.lib advapi32.lib userenv.lib ws2_32.lib bcrypt.lib msvcrt.lib)
+        set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ntdll.lib advapi32.lib userenv.lib ws2_32.lib bcrypt.lib msvcrt.lib)
```
wantehchang, line 526: Vincent: I think `msvcrt.lib` should also be removed.

If you have the time, please remove both `advapi32.lib` and `msvcrt.lib`, and only add them back if .github/workflows/ci-windows.yml fails. Thanks.

wantehchang (approved):

CMakeLists.txt
vrabaud: Sure, done

fdintino:
It looks like the only library I actually needed was advapi32.lib. My build fails after these changes ([action run here](https://github.com/fdintino/Pillow/actions/runs/6406581709/job/17391484466)) with the error:

```
avif.lib(std-43ecf4bdbb5e1e31.std.221f24cf-cgu.0.rcgu.o) : error LNK2001: unresolved external symbol __imp_OpenProcessToken
avif.lib(std-43ecf4bdbb5e1e31.std.221f24cf-cgu.0.rcgu.o) : error LNK2001: unresolved external symbol __imp_SystemFunction036
build\lib.win-amd64-cpython-39\PIL\_avif.cp39-win_amd64.pyd : fatal error LNK1120: 2 unresolved externals
```

Both of those symbols are in advapi32. Was its inclusion causing issues?

wantehchang:
Frankie: Sorry about that. The inclusion of advapi32.lib did not cause problems, but it didn't seem necessary for two reasons:
- We didn't need it before, and
- The "Check static link bundling" step in .github/workflows/ci-windows.yml still passes without it.

We will be happy to add it back. But do you know why the "Check static link bundling" step you added cannot detect this problem?

PR #1651, opened 2023-10-04
Handle out-of-memory failures in avifCodecCreate()

y-guyon:

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -1087,13 +1087,14 @@ avifCodecChoice avifCodecChoiceFromName(const char * name)
     return AVIF_CODEC_CHOICE_AUTO;
 }
 
-avifCodec * avifCodecCreate(avifCodecChoice choice, avifCodecFlags requiredFlags)
+avifResult avifCodecCreate(avifCodecChoice choice, avifCodecFlags requiredFlags, avifCodec ** codec)
 {
+    *codec = NULL;
     struct AvailableCodec * availableCodec = findAvailableCodec(choice, requiredFlags);
-    if (availableCodec) {
-        return availableCodec->create();
-    }
-    return NULL;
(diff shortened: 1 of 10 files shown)


src/read.c
```diff
@@ -4061,8 +4061,7 @@ static avifResult avifCodecCreateInternal(avifCodecChoice choice, const avifTile
         choice = AVIF_CODEC_CHOICE_AVM;
     }
 #endif
-    AVIF_CHECKERR(tile->codecType == avifCodecTypeFromChoice(choice, AVIF_CODEC_FLAG_CAN_DECODE), AVIF_RESULT_NO_CODEC_AVAILABLE);
```
vrabaud, line 4064: why remove that check?

src/avif.c
```diff
@@ -1087,13 +1087,14 @@ avifCodecChoice avifCodecChoiceFromName(const char * name)
     return AVIF_CODEC_CHOICE_AUTO;
 }
 
-avifCodec * avifCodecCreate(avifCodecChoice choice, avifCodecFlags requiredFlags)
+avifResult avifCodecCreate(avifCodecChoice choice, avifCodecFlags requiredFlags, avifCodec ** codec)
```
vrabaud, line 1090: Is the idea to move APIs (like avifAlloc) to return avifResult? I could not agree more with the nodiscard I am adding.

vrabaud (approved):

src/read.c
y-guyon: Because it is part of `avifCodecCreate()` now.

src/avif.c
y-guyon: The reason here is that `avifCodecCreate()` can fail in multiple ways (no codec available or out of memory). `avifResult` is also convenient for the calling site.

PR #1652, opened 2023-10-04
Add some missing return value checks.

maryla-uc:


vrabaud (approved):
Nice, I am working on adding an AVIF_NODISCARD that we can enable on C23 from CMake.

tests/avifyuv.c
```diff
@@ -221,8 +221,13 @@ int main(int argc, char * argv[])
                                 }
                             }
 
-                            avifImageRGBToYUV(image, &srcRGB);
-                            avifImageYUVToRGB(image, &dstRGB);
+                            if (avifImageRGBToYUV(image, &srcRGB) != AVIF_RESULT_OK || avifImageYUVToRGB(image, &dstRGB) != AVIF_RESULT_OK) {
+                                avifRGBImageFreePixels(&srcRGB);
```
vrabaud, line 225: How about a goto Error ?

maryla-uc: It might make sense here but I'm not a big fan of goto and I feel like this is an old test that's not worth spending more time on...

vrabaud: totally right, please go ahead

PR #1653, opened 2023-10-04
Add advapi32.lib back as a mingw dep.

vrabaud:


wantehchang (approved):
LGTM. Thanks. Please wait for Frankie's report back (https://github.com/AOMediaCodec/libavif/pull/1649#issuecomment-1747007919).

kmilos:
Why is there a need to list these explicitly on Windows? AFAIK, they should be included by CMake on all Windows toolchains (MSVC, and MingGW both GCC and Clang) : https://gitlab.kitware.com/search?search=advapi32&nav_source=navbar&project_id=541&group_id=415&search_code=true&repository_ref=master

fdintino:
We can close this pull request. It was indeed an issue with the rav1e release, one that is fixed in the recent weekly prereleases.

@kmilos On a fresh windows image with vs2022, my `CMAKE_C_STANDARD_LIBRARIES_INIT` is:

`advapi32.lib` `comdlg32.lib` `gdi32.lib` `kernel32.lib` `ole32.lib` `oleaut32.lib` `shell32.lib` `user32.lib` `uuid.lib` `winspool.lib`

whereas with `rustc --print native-static-libs foo.rs` I see these libraries (those not found in `CMAKE_C_STANDARD_LIBRARIES_INIT` are bolded):

`advapi32.lib` <b>`bcrypt.lib`</b> `kernel32.lib` <b>`msvcrt.lib`</b> <b>`ntdll.lib`</b> <b>`userenv.lib`</b> <b>`ws2_32.lib`</b>

msvcrt.lib is always a compile-time dependency with msvc, which leaves the four libraries added here to `AVIF_PLATFORM_LIBRARIES`

vrabaud:
Thx for checking @fdintino !

wantehchang:
Thank you, Frankie. I believe we came up with the four libraries added to `AVIF_PLATFORM_LIBRARIES` by trial and error. It's good to know why exactly those four are needed. (I only looked into why ntdll.lib was needed before because that is not commonly used for this kind of code.)

PR #1655, opened 2023-10-05
Add an option to set [[nodiscard]]

vrabaud:


apps/shared/avifjpeg.c
```diff
@@ -1172,7 +1172,7 @@ static avifBool avifJPEGReadInternal(FILE * f,
 
     if (avif->xmp.size > 0 && ignoreXMP) {
         // Clear XMP in case we read it for something else (like gain map).
-        avifImageSetMetadataXMP(avif, NULL, 0);
+        AVIF_CHECK(avifImageSetMetadataXMP(avif, NULL, 0) == AVIF_RESULT_OK);
```
maryla-uc, line 1175: (No action required) Yeah... I didn't add that one because it shouldn't fail and I thought it was fine even if it did, but it doesn't hurt to add the check.

include/avif/avif.h
```diff
@@ -48,6 +48,13 @@ extern "C" {
 #define AVIF_API
 #endif // defined(AVIF_DLL)
 
+#if defined(AVIF_ENABLE_NODISCARD) || (defined(__cplusplus) && __cplusplus >= 201700L) || \
+    (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
```
maryla-uc, line 52: why the || conditions? If we have an explicit flag to turn it on (AVIF_ENABLE_NODISCARD) it seems weird to also turn it on based on some other conditions

tests/gtest/avif_fuzztest_helpers.cc
```diff
@@ -24,8 +24,11 @@ AvifImagePtr CreateAvifImage(size_t width, size_t height, int depth,
   if (image.get() == nullptr) {
     return image;
   }
-  avifImageAllocatePlanes(image.get(),
-                          has_alpha ? AVIF_PLANES_ALL : AVIF_PLANES_YUV);
+  if (avifImageAllocatePlanes(image.get(),
+                              has_alpha ? AVIF_PLANES_ALL : AVIF_PLANES_YUV) !=
+      AVIF_RESULT_OK) {
+    return AvifImagePtr(nullptr, avifImageDestroy);
```
maryla-uc, line 30: in other places we return {nullptr, nullptr} in this case

.github/workflows/ci-unix-shared-installed.yml
```diff
@@ -97,3 +97,19 @@ jobs:
               endif()" >> CMakeLists.txt
 
         cmake . -DCMAKE_PREFIX_PATH=../install
+    - name: Prepare libavif with [[nodiscard]] (cmake)
```
maryla-uc, line 100: (no action required) just wondering if this should be in a separate workflow as intuitively it's a different kind of check, but I guess it's faster/more resource efficient to have it here?

maryla-uc (approved):
Nice improvement! There are a lot of functions in internal.h as well which could also use no_discard, such as all the stream writing ones. And probably a lot of static functions here and there. Can be done later.

include/avif/avif.h
vrabaud: I agree, I thought it would be a good thing to have it by default if possible. Maybe not.

.github/workflows/ci-unix-shared-installed.yml
vrabaud: Indeed, the idea was to minimize resources. Still, that requires to upgrade gcc and forces -Werror (which might be a problem at some point if we have some warnings we do not care about).

vrabaud:
The stream writing functions return an avifResult which is now `nodiscard` so they are taken care of. I fixed internal.h too. Indeed, static functions here and there can be fixed.

apps/shared/avifjpeg.c
vrabaud: Ack

vrabaud:
@wantehchang and @jzern , I am sure you have an opinion on that.
- imposing gcc 11 for CI should not be a problem (gcc 13.2 is the latest stable)
- adding nodiscard by default on C++17 and C23 should not be a problem, it is just a warning
- having CI with it will help us find non-checked return values. But it imposes `-Werror`  which can trigger issues we do not care about at some point. We can disable it then.

wantehchang:
Hi Vincent,

Thank you for adding `AVIF_NODISCARD`. I have made this kind of change manually a few times when we changed functions to return `avifResult`.

It would be good to also use GCC's `__attribute__(warn_unused_result))` if `[[nodiscard]]` is not available. See also this Stack Overflow article: https://stackoverflow.com/questions/45190740/ways-to-specify-nodiscard-before-c17

vrabaud:
@wantehchang , thx, I added the `__attribute__` version.

include/avif/avif.h
```diff
@@ -48,6 +48,21 @@ extern "C" {
...
+#define AVIF_NODISCARD [[nodiscard]]
+#else
+#if defined(__clang__) && defined(__has_attribute)
```
wantehchang, line 55: It seems that newer versions of GCC also have the `__has_attribute` operator. See https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html.

include/avif/avif.h
```diff
@@ -48,6 +48,21 @@ extern "C" {
...
+#if __has_attribute(warn_unused_result)
+#define AVIF_NODISCARD __attribute__((warn_unused_result))
```
wantehchang, line 57: A difference between `__attribute__((warn_unused_result))` and `[[nodiscard]]`, which I ran into, is that `__attribute__((warn_unused_result))` doesn't allow us to use a `void` cast to mark a function's result as used. In most cases this doesn't really matter.

include/avif/avif.h
vrabaud: Ah indeed, I added `defined(__clang__)` thinking it would help windows compilation.
Actually, the issue is that `cl.exe` does not do short-circuit evaluation for preprocessor with `defined(__clang__) && defined(__has_attribute) && __has_attribute(warn_unused_result)`.

include/avif/avif.h
vrabaud: Thx, good to know !

vrabaud:
@wantehchang , any more comment? Thx

include/avif/avif.h
```diff
@@ -48,6 +48,21 @@ extern "C" {
 #define AVIF_API
 #endif // defined(AVIF_DLL)
 
+#if defined(AVIF_ENABLE_NODISCARD) || (defined(__cplusplus) && __cplusplus >= 201700L) || \
```
wantehchang, line 51: Vincent: Why do you compare `__cplusplus` with `201700L` here?

If we are trying to detect C++17, we should compare  `__cplusplus` with `201703L`. See 

- https://en.wikipedia.org/wiki/C%2B%2B17
- https://en.cppreference.com/w/cpp/preprocessor/replace#Predefined_macros
- https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170.

wantehchang: I found that the comparison with `201700L` was copied to libwebp: https://chromium-review.googlesource.com/c/webm/libwebp/+/5020699

vrabaud: Indeed, bad copy/paste. I fixed it there too.

vrabaud: Fixed in https://github.com/AOMediaCodec/libavif/pull/2246

PR #1659, opened 2023-10-06
Free dedup in case of error to avoid memory leaks.

vrabaud:


y-guyon (approved):
Alternative:

```c
avifResult avifRWStreamWriteProperties(avifStream * stream, uint32_t imageWidth, ...) {
    avifBoxMarker ispe;
    AVIF_CHECKRES(avifRWStreamWriteFullBox(&dedup->s, "ispe", AVIF_BOX_SIZE_TBD, 0, 0, &ispe));
    AVIF_CHECKRES(avifRWStreamWriteU32(&dedup->s, imageWidth));  // unsigned int(32) image_width;
    ...
}

avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
{
    ...
    avifItemPropertyDedupStart(dedup);
    const avifResult propStatus = avifRWStreamWriteProperties(&dedup->s, imageWidth, ...);
    avifItemPropertyDedupDestroy(dedup);
    AVIF_CHECKRES(propStatus);
    ...
```

The new macro is fine too.

vrabaud:
There are actually not too many arguments to pass so having a function is fine.

src/write.c
```diff
@@ -2900,7 +2909,7 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
 
     avifEncoderItemReferenceArray layeredColorItems;
     avifEncoderItemReferenceArray layeredAlphaItems;
-    avifResult result = AVIF_RESULT_OK;
+    result = AVIF_RESULT_OK;
```
y-guyon, line 2912: Note that `result` is already `AVIF_RESULT_OK` here, but it is fine to enforce it to be clear.

y-guyon (approved):

src/write.c
vrabaud: If it is unassigned, it will be caught by the compiler anyway. Removing it.

PR #1662, opened 2023-10-06
Check avifAlloc() in avifSampleTableCreate()

y-guyon:
and in avifCodecDecodeInputCreate(), avifMetaCreate(), avifDecoderDataCreate(). Propagate avifResults in read.c.


vrabaud (approved):
Thx, definitely cleaner to use a status

PR #1666, opened 2023-10-06
read.c: Remove diagnostic message

vigneshvg:
`avifImageScaleWithLimit` already updates the diagnostic message on failure. So there is no need to over-write it with a less informative message in the caller.


wantehchang (approved):
LGTM. Thanks.

PR #1668, opened 2023-10-08
Correct the name of the 'infe' box

wantehchang:
It is ItemInfoEntry.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -152,7 +152,7 @@ void avifCodecEncodeOutputDestroy(avifCodecEncodeOutput * encodeOutput)
 typedef struct avifEncoderItem
 {
     uint16_t id;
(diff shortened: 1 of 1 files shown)


src/write.c
```diff
@@ -152,7 +152,7 @@ void avifCodecEncodeOutputDestroy(avifCodecEncodeOutput * encodeOutput)
 typedef struct avifEncoderItem
 {
     uint16_t id;
-    uint8_t type[4];                      // 4-character 'item_type' field in the 'infe' (item info) box
+    uint8_t type[4];                      // 4-character 'item_type' field in the 'infe' (item info entry) box
```
wantehchang, line 155: I noticed this while comparing the SlimHEIF proposal with libavif (src/write.c).

y-guyon (approved):

PR #1673, opened 2023-10-11
Rename avifStatus to avifResult in comment.

vrabaud:
[skip ci]


maryla-uc (approved):
Glad to see I'm not the only one who keeps thinking it's called avifStatus

PR #1674, opened 2023-10-11
Implement tranfer functions for gamma<->linear conversion.

maryla-uc:
Author: tongyuantongyu in https://github.com/AOMediaCodec/libavif/pull/444 with some changes by vrabaud and myself.

In particular, PQ and HLG transfer functions now map to extended SDR range where 1.0 means SDR white luminosity, and larger values mean brighter than SDR.


maryla-uc:
Changed to fix some swapped functions and document more where the implementation/formulas come from.

vrabaud (approved):

vrabaud (approved):

PR #1677, opened 2023-10-12
Add utilities to get/set RGBA values for any RGB format.

maryla-uc:
These internal utilities will be useful for gain map code like tone mapping and creating gain maps.

They might also be used to simplify some of the YUV/RGB conversion code, after checking that they don't impact performance (the code is not particularly optimized).


tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -839,5 +839,62 @@ INSTANTIATE_TEST_SUITE_P(
...
+    return;  // Invalid combination e.g. RGB565 with depth > 8.
+  }
+
+  float kEpsilon = 1.0f / color_space.maxChannelF;
+  if (rgb_format == AVIF_RGB_FORMAT_RGB_565) {
+    // Only 5 bits of information per channel except G which has 6).
+    kEpsilon = 1.0f / (1 << 5);
+  }
+
+  const float pixel_to_write[4] = {0.0f, 0.5f, 1.0f, 0.75f};
```
y-guyon, line 869: These values may be "too easy" for precision round-trip checking. Consider `FillImageGradient()` on a bigger input image with prime dimensions to generate a somewhat better covering distribution.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -839,5 +839,62 @@ INSTANTIATE_TEST_SUITE_P(
...
+  // Unused yuv image, simply needed to initialize the rgb image.
+  testutil::AvifImagePtr yuv(avifImageCreate(4, 2, 8, AVIF_PIXEL_FORMAT_YUV444),
+                             avifImageDestroy);
+
+  testutil::AvifRgbImage rgb(yuv.get(), rgb_depth, rgb_format);
+  rgb.isFloat = is_float;
+
+  avifRGBColorSpaceInfo color_space;
+  if (!avifGetRGBColorSpaceInfo(&rgb, &color_space)) {
+    return;  // Invalid combination e.g. RGB565 with depth > 8.
```
y-guyon, line 860: I recommend tweaking the `INSTANTIATE_TEST_SUITE_P` below rather than keeping this check that will silently skip invalid tests. For example I doubt that `/*is_float=*/Values(true)` works with `/*rgb_depth=*/Values(8)`.

There could also be negative checks to verify that `avifGetRGBColorSpaceInfo()` returns false when expected, but this is unnecessary in my opinion.

y-guyon (changes requested):

src/reformat.c
```diff
@@ -567,6 +567,15 @@ static void avifStoreRGB8Pixel(avifRGBFormat format, uint8_t R, uint8_t G, uint8
...
 }
 
+static void avifGetRGB565(const uint8_t * ptrR, uint8_t * R, uint8_t * G, uint8_t * B)
+{
+    // See https://docs.microsoft.com/en-us/windows/win32/directshow/working-with-16-bit-rgb
+    const uint16_t rgb656 = ((const uint16_t *)ptrR)[0];
+    *R = (uint8_t)(((rgb656 & 0xF800) >> 11) << 3);
+    *G = (uint8_t)(((rgb656 & 0x07E0) >> 5) << 2);
+    *B = (uint8_t)(((rgb656 & 0x001F)) << 3);
+}
```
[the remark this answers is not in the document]
maryla-uc: Indeed, that's microsoft's formula ¯\_(ツ)_/¯
Changed to a version like in libyuv that ORs the first bits.

src/reformat.c
```diff
@@ -1299,19 +1318,13 @@ static avifResult avifRGBImageToF16(avifRGBImage * rgb)
...
         uint16_t * pixel = pixelRowBase;
         for (uint32_t i = 0; i < rgb->width * channelCount; ++i, ++pixel) {
-            union
-            {
-                float f;
-                uint32_t u32;
-            } f16;
+            avifF16 f16;
             f16.f = *pixel * multiplier;
             *pixel = (uint16_t)(f16.u32 >> 13);
```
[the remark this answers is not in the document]
maryla-uc: I thought the current version might be a bit more efficient, not sure if it matters. I can be done later if needed.

src/reformat.c
```diff
@@ -567,6 +567,15 @@ static void avifStoreRGB8Pixel(avifRGBFormat format, uint8_t R, uint8_t G, uint8
     *ptrB = B;
 }
 
+static void avifGetRGB565(const uint8_t * ptrR, uint8_t * R, uint8_t * G, uint8_t * B)
```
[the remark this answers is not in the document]
maryla-uc: Yes you're supposed to pass a pointer to the R channel, although in practice any of them work since the offset is set to 0 for all channels.

include/avif/internal.h
```diff
@@ -193,12 +195,21 @@ typedef struct avifYUVColorSpaceInfo
...
     avifRGBColorSpaceInfo rgb;
     avifYUVColorSpaceInfo yuv;
 } avifReformatState;
 
+// Retrieves the pixel value at position (i, j) expressed as floats in [0;1]. If the image's format doesn't have alpha,
+// rgbaPixel[3] is set to 1.0f.
+void avifGetRGBAPixel(const avifRGBImage * src, int i, int j, float rgbaPixel[4], const avifRGBColorSpaceInfo * info);
+// Sets the pixel value at position (i, j) from RGBA values expressed as floats in [0;1]. If the image's format doesn't
+// support alpha, rgbaPixel[3] is ignored.
+void avifSetRGBAPixel(const avifRGBImage * dst, int i, int j, const float rgbaPixel[4], const avifRGBColorSpaceInfo * info);
```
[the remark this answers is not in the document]
maryla-uc: Possibly, todo later if needed (for now I just changed the parameter names to x/y)

src/reformat.c
```diff
@@ -1714,3 +1727,99 @@ int avifFullToLimitedUV(uint32_t depth, int v)
...
+        avifStoreRGB8Pixel(dst->format,
+                           (uint8_t)(0.5f + (rgbaPixel[0] * info->maxChannelF)),
+                           (uint8_t)(0.5f + (rgbaPixel[1] * info->maxChannelF)),
+                           (uint8_t)(0.5f + (rgbaPixel[2] * info->maxChannelF)),
+                           ptrR,
+                           ptrG,
+                           ptrB);
+        if (ptrA) {
+            *ptrA = (uint8_t)(0.5f + (rgbaPixel[3] * info->maxChannelF));
+        }
```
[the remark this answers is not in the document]
maryla-uc: Yes it's done like this evywhere in this file. Will keep as is for now and maybe we should refactor it.

tests/gtest/avifrgbtoyuvtest.cc
maryla-uc: Done with some home made values since FillImageGradient() is for a yuv image (I could convert to rgb but it adds a layer of indirection to the values)

tests/gtest/avifrgbtoyuvtest.cc
maryla-uc: Urgh you're right I meant to use 16 for the floats :/ Thanks for catching this.

src/reformat.c
```diff
@@ -569,11 +569,16 @@ static void avifStoreRGB8Pixel(avifRGBFormat format, uint8_t R, uint8_t G, uint8
 
 static void avifGetRGB565(const uint8_t * ptrR, uint8_t * R, uint8_t * G, uint8_t * B)
 {
-    // See https://docs.microsoft.com/en-us/windows/win32/directshow/working-with-16-bit-rgb
+    // References for RGB565 color conversion:
+    // * https://docs.microsoft.com/en-us/windows/win32/directshow/working-with-16-bit-rgb
+    // * http://google3/third_party/libyuv/files/source/row_common.cc;l=185;rcl=573160022
```
y-guyon, line 574: Replace internal link with GitHub permalink

tests/gtest/avifrgbtest.cc
```diff
@@ -0,0 +1,107 @@
...
+      avifSetRGBAPixel(&rgb, i, j, &color_space, pixel_to_write);
+      avifGetRGBAPixel(&rgb, i, j, &color_space, pixel_read);
+      EXPECT_NEAR(pixel_read[0], pixel_to_write[0], epsilon);
+      EXPECT_NEAR(pixel_read[1], pixel_to_write[1], epsilon);
+      EXPECT_NEAR(pixel_read[2], pixel_to_write[2], epsilon);
+      if (avifRGBFormatHasAlpha(rgb_format)) {
+        EXPECT_NEAR(pixel_read[3], pixel_to_write[3], epsilon);
+      } else {
+        EXPECT_EQ(pixel_read[3], 1.0f);
+      }
```
y-guyon, line 63: Note that the `avifImage->float->avifImage` round-trip would allow for exact value equality comparison instead of `epsilon` equivalence. It can be added later on if useful (the current test does not guarantee that there is no color drift when converting back-and-forth multiple times).

y-guyon (approved):

tests/gtest/avifrgbtoyuvtest.cc
y-guyon: There could be a `void FillImageGradient(avifRGBImage* image);` in `aviftest_helpers.h` if used in multiple places. For example `avifalphapremtest.cc` uses `FillImageGradient()` followed by `avifImageYUVToRGB()`. This is fine for now.

maryla-uc: In this case I'm not even filling an avifRGBImage, I'm filling in a float[4]. I think it's different, and any additional indirection would make it less clear what the test is testing. Also note avifRGBImage doesn't support floats but only half flaots.

tests/gtest/avifrgbtest.cc
maryla-uc: I'm sorry I don't really understand what test you have in mind when you talk about `avifImage->float->avifImage`.

tests/gtest/avifrgbtoyuvtest.cc
y-guyon: Right, I got confused with my other comment.

tests/gtest/avifrgbtest.cc
y-guyon: My bad, I meant `avifRGBImage->float->avifRGBImage`. Right now this is testing `float->avifRGBImage->float` which has some loss; the other way around should be lossless and easier to test. But I guess this test exists in this way for a reason, so please disregard this note if irrelevant.

maryla-uc: Ha thanks I understand now. The thing is, how would get an avifRGBImage in all these different RGB formats to run the test?
The simplest way is to use avifSetRGBAPixel, which essentially leads to a test like this.
I guess you could take yuv image and convert it to rgb, but again I think that muddies the waters because the yuv->rgb conversion could be using these two functions at some point.

Actually I just tried adding a test like this: `avifImage with FillImageGradient->avifRGBImage->float->avifRGBImage->avifImage` and then use GetPSNR to compare at the end. Aaaand it doesn't work because of clipping issues. FillImageGradient create YUV values that you can't get by converting from RGB. E.g. YUV(0,0,0) converts to RGB(0,135,0) but if you convert that back you get YUV(79,83,71) (for BT.601)
You can double check e.g. on https://www.mikekohn.net/file_formats/yuv_rgb_converter.php by clicking the "yuv2rgb" button then the "rgb2yuv" button. So it would need to be changed to create valid values (maybe by converting from RGB? inception...).

Anyway you did say I could disregard this and I have a headache now so I will say this test is good enough for now :)

y-guyon: > how would get an avifRGBImage in all these different RGB formats to run the test?

See a similar test at https://chromium.googlesource.com/codecs/libwebp2/+/49cb845ecae4c3856e7093eddd305b2a7c63b063/tests/test_color_precision.cc#74  
It should be adapted to each format of course (trivial for ARGB permutations, special cases for float and rgb565).

> Actually I just tried adding a test like this: `avifImage with FillImageGradient->avifRGBImage->float->avifRGBImage->avifImage` and then use GetPSNR to compare at the end. Aaaand it doesn't work because of clipping issues.

This approach could work if the comparison was made on `avifRGBImage` and not `avifImage`, but as you said it may not be bijective so some input ARGB combinations would not be tested. But as a quick additional test it could be enough. Let me try that.

> Anyway you did say I could disregard this and I have a headache now so I will say this test is good enough for now :)

Of course. This is just further thorough testing, the current one is already great. Thanks for adding tests.

Issue #1679, opened 2023-10-12
When linking avifenc with the merged static lib libavif.a, we still link with libaom.a, libdav1d.a, etc.


wantehchang:
@fdintino @vrabaud

When linking avifenc with the merged static lib libavif.a, we still link with libaom.a, libdav1d.a, etc. unnecessarily. I believe the linker won't use libaom.a, libdav1d.a, etc. because the object files in libavif.a satisfy all the symbol references.

Here is the command line on Linux:

$ ninja -v avifenc
[1/1] : && /usr/bin/c++ -g  CMakeFiles/avifenc.dir/apps/avifenc.c.o -o avifenc  libavif_apps.a  libavif.a  /home/wtc/libavif.1/libavif/ext/dav1d/build/src/libdav1d.a  /home/wtc/libavif.1/libavif/ext/libgav1/build/libgav1.a  /home/wtc/libavif.1/libavif/ext/SVT-AV1/Bin/Release/libSvtAv1Enc.a  /home/wtc/libavif.1/libavif/ext/aom/build.libavif/libaom.a  /home/wtc/libavif.1/libavif/ext/libyuv/build/libyuv.a  /home/wtc/libavif.1/libavif/ext/libwebp/build/libsharpyuv.a  -lm  -ldl  /usr/lib/x86_64-linux-gnu/libpng.so  /usr/lib/x86_64-linux-gnu/libz.so  /usr/lib/x86_64-linux-gnu/libjpeg.so && :

vrabaud:
I believe this is a duplicate of https://github.com/AOMediaCodec/libavif/issues/1679
Codec libraries are present twice: within the merged libavi.a and as their own codec libraries. They need to be removed from the dependencies. I propose https://github.com/AOMediaCodec/libavif/pull/1680

vrabaud:
Fixed in https://github.com/fdintino/libavif/commit/a8fb63ef7105783d3e014e6af93574be6bd5338b

Issue #1678, opened 2023-10-12
Merged static lib libavif.a is sometimes not updated when ext/aom/build.libavif/libaom.a changes


wantehchang:
@fdintino @vrabaud

Consider libaom commit https://source.corp.google.com/h/aomedia/aom/+/82104fc2425c1c7dc7b34329917f04de526a8873
and its parent commit https://source.corp.google.com/h/aomedia/aom/+/7e4b033152b4e7a957a4af2c63960ab58ca75d23.

First checkout libaom commit 82104fc2425c1c7dc7b34329917f04de526a8873 in ext/aom, build libaom in ext/aom/build.libavif, and then build libavif.a. Save a copy of libavif.a as libavif.a.82104fc2425c1c7dc7b34329917f04de526a8873.

Now checkout libaom commit 7e4b033152b4e7a957a4af2c63960ab58ca75d23 in ext/aom, build libaom in ext/aom/build.libavif, and then build libavif.a.

The `cmp libavif.a libavif.a.82104fc2425c1c7dc7b34329917f04de526a8873` command shows libavif.a is not updated.

Now rm libavif.a and build libavif.a. Now the `cmp libavif.a libavif.a.82104fc2425c1c7dc7b34329917f04de526a8873` command shows libavif.a is updated.

I am wondering if the `merge_static_libs()` function does not set up dependency tracking correctly.

Note that I said libavif.a is **sometimes** not updated. I haven't figured out the exact condition in which libavif.a is not updated.

This bug breaks doing a `git bisect` on libaom in ext/aom and using avifenc to determine whether a libaom commit is good or bad. I discovered this bug after spending three hours on multiple failed attempts at doing `git bisect` this way.

wantehchang:
For an example where libavif.a **is** updated, use the libaom tags v3.7.0 and v3.6.0 in the steps above.

So one may speculate that if ext/aom/build.libavif/libaom.a changes "significantly", then libavif.a will be updated.

During a "git bisect", the libaom commits will eventually get close enough such that libavif.a is not updated, and the bisection will malfunction.

vrabaud:
I cannot reproduce with the latest CMake changes.

PR #1680, opened 2023-10-13
Do not target_link_libraries when merge_static_lib

vrabaud:
Otherwise, the codec libraries appear twice in the linkage: as their own static library and within libavif.a.
This fixes https://github.com/AOMediaCodec/libavif/issues/1679. I did not try on https://github.com/AOMediaCodec/libavif/issues/1678.


CMakeLists.txt
```diff
@@ -641,7 +641,12 @@ endif()
 add_library(avif ${AVIF_SRCS})
 set_target_properties(avif PROPERTIES VERSION ${LIBRARY_VERSION} SOVERSION ${LIBRARY_SOVERSION} C_VISIBILITY_PRESET hidden)
 target_compile_definitions(avif PRIVATE ${AVIF_PLATFORM_DEFINITIONS} ${AVIF_CODEC_DEFINITIONS})
-target_link_libraries(avif PRIVATE ${AVIF_CODEC_LIBRARIES} ${AVIF_PLATFORM_LIBRARIES})
+if(BUILD_SHARED_LIBS)
+    target_link_libraries(avif PRIVATE ${AVIF_CODEC_LIBRARIES} ${AVIF_PLATFORM_LIBRARIES})
+else()
+    include(merge_static_libs)
+    merge_static_libs(avif ${AVIF_CODEC_LIBRARIES} ${AVIF_PLATFORM_LIBRARIES})
```
wantehchang, line 648: I didn't know that the decision to merge static libraries does not take the `AVIF_LOCAL_AOM`, `AVIF_LOCAL_DAV1D`, ... variables into account. I thought we would only merge libavif.a with the libaom, dav1d, ... static libraries when they were built locally in the ext/ directory.

If the libaom, dav1d, ... static libraries are installed as system packages (I don't know if system packages provide static libraries, but for this discussion let's assume they do) and `AVIF_LOCAL_AOM`, `AVIF_LOCAL_DAV1D` are equal to `OFF`, it doesn't seem appropriate to merge libavif.a with the static libraries installed on the system.

I am wondering if we should abandon the merge_static_libs approach and seek an alternative solution to the problem that Frankie needed to solve.

wantehchang (approved):
LGTM. Thanks for the fix. I tested your fix. The command line to link avifenc becomes:

[1/1] : && /usr/bin/c++ -g  CMakeFiles/avifenc.dir/apps/avifenc.c.o -o avifenc  libavif_apps.a  libavif.a  /home/wtc/libavif.1/libavif/ext/libyuv/build/libyuv.a  /home/wtc/libavif.1/libavif/ext/libwebp/build/libsharpyuv.a  -lm  -ldl  /usr/lib/x86_64-linux-gnu/libpng.so  /usr/lib/x86_64-linux-gnu/libz.so  /usr/lib/x86_64-linux-gnu/libjpeg.so && :

Note that we still link with libyuv.a and libsharpyuv.a even though they have been merged with libavif.a. But that is caused by a separate problem -- we add libyuv.a and libsharpyuv.a  to `AVIF_PLATFORM_LIBRARIES` and the `avif_apps` library is linked with `AVIF_PLATFORM_LIBRARIES`. However, `avif_apps` doesn't use libyuv and libsharpyuv.

Let's deal with one bug at a time :-)

CMakeLists.txt
fdintino: I'm not sure the behavior should be different for system static libraries vs local libraries. In either case the static library isn't really portable; whoever is linking against the library would need to know which codecs it was linked against and have ABI compatible static libraries installed on that system. And often times the build environment happens in CI, typically on a different platform than where it eventually gets used.

I think the "proper" way to do this is to include the statically-linked system libraries in Libs.private of libavif.pc. This would allow (for instance) another cmake project to call `pkg_check_modules` to get the necessary linker flags. But I'm not sure that's superior to creating a combined archive.

fdintino:
This is an alternate approach that I took

https://github.com/AOMediaCodec/libavif/compare/main...fdintino:libavif:fix/static-merge-dependency-tracking

It resolves #1678 and also handles the scenario where there is a shared library in `${AVIF_PLATFORM_LIBRARIES}` or `${AVIF_CODEC_LIBRARIES}`. This can happen if only a shared library file is available for one of the build dependencies.

fdintino:
I'll open that as a pull request so that it can host any discussion.

wantehchang:
Frankie: Thank you for your comment. I also wondered if the "proper" solution might be to add the libraries that libavif.a depend on to libavif.pc. Bu I am not familiar with pkg-config, so I don't know what is involved and the complications you alluded to.

Vincent: This pull request does not fix https://github.com/AOMediaCodec/libavif/pull/1678. (I wasn't hopeful, but I still tested it.)

fdintino:
@wantehchang could you try with #1683? Locally, if I run:

```
touch ../ext/aom/build.libavif/libaom.a
ninja -v
```

it re-archives libavif.a and then rebuilds the dependent executables.

wantehchang:
Frankie: I responded in https://github.com/AOMediaCodec/libavif/pull/1683.

PR #1681, opened 2023-10-13
Add avifGetRGBAPixel() -> avifSetRGBAPixel() test

y-guyon:
Make sure the behaviors and rounding implementations of avifGetRGBAPixel() and avifSetRGBAPixel() match exactly.


maryla-uc (approved):
thanks!

PR #937, opened 2022-05-09
Make avifImageScale function public

vigneshvg:
This is useful for apps to scale images (without having to
worry about bitdepth, subsampling, etc).

This commit renames avifImageScale to avifImageScaleWithLimit. The
latter would remain an internal only function while the former is
now part of the public ABI (moved to avif.h)

Potential workaround for issue #1484.

PR #1682, opened 2023-10-13
Add an avifApplyGainMap function.

maryla-uc:
Applies a gain map to a base image based on the display's hdr capacity.


include/avif/avif.h
```diff
@@ -625,6 +629,30 @@ typedef struct avifGainMapMetadataDouble
...
 // Returns AVIF_FALSE if some field values are < 0 or > UINT32_MAX.
 AVIF_API avifBool avifGainMapMetadataDoubleToFractions(avifGainMapMetadata * dst, const avifGainMapMetadataDouble * src);
+
+// Performs tone mapping on a base image using the provided gain map.
+// The HDR capacity (also known as HDR headroom or ratio) is the ratio of HDR to
+// SDR white brightness of the display to tone map for, in linear space.
+// 'toneMappedImage' should have the 'format' and 'depth' fields set to the desired values.
+// If non NULL, 'clli' will be filled with the light level information of the tone mapped image.
+struct avifRGBImage;
+AVIF_API avifResult avifApplyGainMap(const struct avifImage * baseImage,
```
wantehchang, line 639: Many libavif public functions follow an object-oriented naming convention, in which the function name begins with the type of the first parameter. Please see if these two functions should follow this naming convention.

include/avif/avif.h
```diff
@@ -625,6 +629,30 @@ typedef struct avifGainMapMetadataDouble
...
+                                     const avifGainMap * gainMap,
+                                     float hdrCapacity,
+                                     avifTransferCharacteristics outputTransferCharacteristics,
+                                     struct avifRGBImage * toneMappedImage,
+                                     avifContentLightLevelInformationBox * clli,
+                                     avifDiagnostics * diag);
+// Same as above but takes an avifRGBImage as input instead of avifImage.
+AVIF_API avifResult avifApplyGainMapRgb(const struct avifRGBImage * baseImage,
```
wantehchang, line 647: The "Rgb" in the function name should be all capitals "RGB".

include/avif/avif.h
```diff
@@ -625,6 +629,30 @@ typedef struct avifGainMapMetadataDouble
 // to the closest uint32_t fractions.
 // Returns AVIF_FALSE if some field values are < 0 or > UINT32_MAX.
 AVIF_API avifBool avifGainMapMetadataDoubleToFractions(avifGainMapMetadata * dst, const avifGainMapMetadataDouble * src);
+
+// Performs tone mapping on a base image using the provided gain map.
```
wantehchang, line 633: I am wondering if tone mapping is within the scope of libavif. Should we first add these functions to aviftest_helpers.* if we will only be using these functions in tests?

maryla-uc: Fair enough. I moved it to internal.h for now.
It's a lot easier to move things between avif.h and internal.h than to aviftest_helpers.

maryla-uc:
Sorry I mistakenly amended instead of adding a commit.

include/avif/avif.h
maryla-uc: I was chatting with Frank about this and he pointed out that tone mapping would be useful for things like converting an SDR+gainmap image to HDR+gainmap. It would not necessarily need to be in the API but at least in the tools? (to be discussed)
I will prepare a short doc so we can discuss what should and should not be in scope for libavif/avifenc/avifdec.

wantehchang: I'm sorry that my comment was ambiguous. I meant to say "I am wondering if tone mapping is within the scope of the libavif **library**."

Re: declaring these two functions in internal.h: The code in the apps directory cannot use the internal functions of the libavif library. So if these two functions will be used by avifdec or avifenc, they need to be either public functions of the libavif library or functions in the apps/shared directory (i.e., the apps_shared library).

If we anticipate that these two functions will be useful to most libavif clients (such as Chrome), then they should be public functions of the libavif library. Otherwise it is better to add them to the apps/shared directory.

maryla-uc: Would it make sense to try and get this submitted as is in gainmap.c/internal.h and move it later once we figure out where it should be?

wantehchang: Yes. I may not be able to review this pull request quickly, because I was just checking changes to the public header (avif.h). I can try to review this later today.

src/gainmap.c
```diff
@@ -49,4 +53,238 @@ avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, c
...
+    const double * offsetOther = gainMap->metadata.baseRenditionIsHDR ? metadata.offsetSdr : metadata.offsetHdr;
+
+    float rgbMaxLinear = 0; // Max tone mapped pixel value across R, G and B channels.
+    float rgbSumLinear = 0; // Sum of max(r, g, b) for mapped pixels.
+
+    for (uint32_t j = 0; j < height; ++j) {
+        for (uint32_t i = 0; i < width; ++i) {
+            float basePixelRGBA[4];
+            avifGetRGBAPixel(baseImage, i, j, &baseRGBInfo, basePixelRGBA);
+            if (weight == 0.0 && outputTransferCharacteristics == transferCharacteristics) {
```
wantehchang, line 203: Optional: Since this conditional is loop invariant, it would be good to check this outside the for loop and have a simpler loop for this case.

src/gainmap.c
```diff
@@ -49,4 +53,238 @@ avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, c
...
+        }
+    }
+    if (clli != NULL) {
+        // For exact CLLI value definitions, see ISO/IEC 23008-2 section D.3.35
+        // at https://standards.iso.org/ittf/PubliclyAvailableStandards/index.html
+
+        // Convert extended SDR (where 1.0 is SDR white) to nits.
+        clli->maxCLL = (uint16_t)AVIF_CLAMP(avifRoundf(rgbMaxLinear * SDR_WHITE_NITS), 0.0f, (float)INT16_MAX);
+        const float rgbAverageLinear = rgbSumLinear / (width * height);
+        clli->maxPALL = (uint16_t)AVIF_CLAMP(avifRoundf(rgbAverageLinear * SDR_WHITE_NITS), 0, (float)INT16_MAX);
```
wantehchang, line 245: In line 243 and this line, should we use `UINT16_MAX` instead of `INT16_MAX`?

src/gainmap.c
```diff
@@ -49,4 +53,238 @@ avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, c
...
+}
+
+avifResult avifImageApplyGainMap(const avifImage * baseImage,
+                                 const avifGainMap * gainMap,
+                                 float hdrCapacity,
+                                 avifTransferCharacteristics outputTransferCharacteristics,
+                                 avifRGBImage * toneMappedImage,
+                                 avifContentLightLevelInformationBox * clli,
+                                 avifDiagnostics * diag)
+{
```
wantehchang, line 263: This function also needs to call `avifDiagnosticsClearError(diag)` if you decide to keep the `diag` parameter.

src/gainmap.c
```diff
@@ -49,4 +53,238 @@ avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, c
...
+                                   baseImage->transferCharacteristics,
+                                   gainMap,
+                                   hdrCapacity,
+                                   outputTransferCharacteristics,
+                                   toneMappedImage,
+                                   clli,
+                                   diag);
+    if (res != AVIF_RESULT_OK) {
+        goto cleanup;
+    }
```
wantehchang, line 282: This if statement can be omitted. (Given the definition of this function, we are unlikely to add new code here.)

wantehchang (approved):
LGTM. I didn't review the changes in the tests/ directory.

include/avif/internal.h
```diff
@@ -644,6 +644,32 @@ typedef struct avifSequenceHeader
...
+                                 avifDiagnostics * diag);
+// Same as above but takes an avifRGBImage as input instead of avifImage.
+avifResult avifImageApplyGainMapRGB(const struct avifRGBImage * baseImage,
+                                    avifTransferCharacteristics transferCharacteristics,
+                                    const avifGainMap * gainMap,
+                                    float hdrCapacity,
+                                    avifTransferCharacteristics outputTransferCharacteristics,
+                                    struct avifRGBImage * toneMappedImage,
+                                    avifContentLightLevelInformationBox * clli,
+                                    avifDiagnostics * diag);
```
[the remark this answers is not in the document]
maryla-uc: Done. Indeed `struct` is no longer needed after the move to internal.h

src/gainmap.c
```diff
@@ -49,4 +53,238 @@ avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, c
...
+    if (hdrCapacity > hdrCapacityMin) {
+        if (hdrCapacity < hdrCapacityMax) {
+            w = (logf(hdrCapacity) - logf(hdrCapacityMin)) / (logf(hdrCapacityMax) - logf(hdrCapacityMin));
+        } else {
+            w = 1.0f;
+        }
+    }
+
+    if (metadata->baseRenditionIsHDR) {
+        w -= 1.0f;
```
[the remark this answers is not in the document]
maryla-uc: Yes this is correct, see on page 4 [here](https://helpx.adobe.com/content/dam/help/en/camera-raw/using/gain-map/jcr_content/root/content/flex/items/position/position-par/table/row-3u03dx0-column-4a63daf/download_section/download-1/Gain_Map_1_0d12.pdf) "If the Base rendition is HDR, then set the weight W = F – 1. This causes the gain map to be applied in the “inverse” direction to map HDR to SDR." (where F is the same as 'w' in this code)

src/gainmap.c
```diff
@@ -49,4 +53,238 @@ avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, c
...
+#define SDR_WHITE_NITS 203.0f
+
+avifResult avifImageApplyGainMapRGB(const avifRGBImage * baseImage,
+                                    avifTransferCharacteristics transferCharacteristics,
+                                    const avifGainMap * gainMap,
+                                    float hdrCapacity,
+                                    avifTransferCharacteristics outputTransferCharacteristics,
+                                    avifRGBImage * toneMappedImage,
+                                    avifContentLightLevelInformationBox * clli,
+                                    avifDiagnostics * diag)
```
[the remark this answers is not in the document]
maryla-uc: It's definitely simpler, but then we won't get the benefit of the diagonistics? I think the avifResult return status alone is difficult to debug... Right now avifDiagnostics is a bit annoying because it cannot be null, but I think we should also revisit this.

I saw that in [this comment](https://github.com/AOMediaCodec/libavif/pull/937#discussion_r1347560842) you said that it had to be non-null currently because "it's a useful debugging tool", but then you don't seem to think its use is warranted here or in a lot of other places.
I think making it nullable would allow the caller to decide if they want the extra diagnostics or not?

src/gainmap.c
maryla-uc: Done, I moved it earlier to avoid e.g. rescaling the gain map if we don't even need it.

include/avif/internal.h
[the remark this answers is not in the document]
wantehchang: When these declarations are moved to avif.h, we can put them after the `avifRGBImage` and `avifImage` typedefs. Alternatively we can use forward declarations of the structs and define the typedefs early:

```
struct avifImage;  // This forward declaration may not be necessary.
typedef struct avifImage avifImage;
```

src/gainmap.c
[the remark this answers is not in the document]
wantehchang: Thank you for the link to the specification.

src/gainmap.c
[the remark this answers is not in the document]
wantehchang: The main point of the `diag` parameter is to give a detailed explanation when parsing something really complicated. If the only way a function can fail is bad arguments or memory allocation failures, then it probably doesn't need the `diag` parameter. That's why I think the `diag` parameter should only be used with functions that deal with the ISO BMFF file format. We can keep the `diag` parameter if it is useful to these two functions. (I didn't analyze the diagnostics messages written by these two functions.)

We can allow `diag` to be null. But I am worried that we may see calls to those API functions that end with a `/*diag=*/NULL` argument, which seems like an eyesore. That's what I meant by "the functions will be simpler without the `diag` parameter".

wantehchang (approved):
LGTM.

src/gainmap.c
```diff
@@ -49,4 +53,257 @@ avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, c
...
+                }
+                avifSetRGBAPixel(toneMappedImage, i, j, &toneMappedPixelRGBState, basePixelRGBA);
+            }
+        }
+        goto cleanup;
+    }
+
+    if (gainMap->image->width != width || gainMap->image->height != height) {
+        rescaledGainMap = avifImageCreateEmpty();
+        res = avifImageCopy(rescaledGainMap, gainMap->image, AVIF_PLANES_YUV);
```
wantehchang, line 183: We should be able to replace `avifImageCopy` with `avifImageSetViewRect` here to avoid copying the gain map image.

Note: I considered not rescaling the gain map image, but the specification recommends against it: "It is recommended to avoid using nearest-neighbor (point sampling) for quality reasons."

maryla-uc: Done, thanks.

src/gainmap.c
```diff
@@ -49,4 +53,257 @@ avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, c
...
+    const avifTransferFunction linearToGamma = avifTransferCharacteristicsGetLinearToGammaFunction(outputTransferCharacteristics);
+
+    // Early exit if the gain map does not need to be applied.
+    if (weight == 0.0f) {
+        // Just convert from one rgb format to another.
+        for (uint32_t j = 0; j < height; ++j) {
+            for (uint32_t i = 0; i < width; ++i) {
+                float basePixelRGBA[4];
+                avifGetRGBAPixel(baseImage, i, j, &baseRGBInfo, basePixelRGBA);
+                if (outputTransferCharacteristics != transferCharacteristics) {
```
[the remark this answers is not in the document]
maryla-uc: I agree this is a loop invariant but I think it's simpler/shorter this way? Not sure if there's a performance impact.

src/gainmap.c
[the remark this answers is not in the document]
maryla-uc: Indeed apart form memory issues, errors should come from bad inputs. But even then, there are many ways in which the inputs can be wrong. I find it very frustrating to debug a simple status return value.

src/gainmap.c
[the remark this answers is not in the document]
wantehchang: Yes, let's go for simplicity. Branches in the inner loop may potentially hurt performance. There are if statements (branches) inside `avifGetRGBAPixel()` and `avifSetRGBAPixel()` too, so it doesn't seem necessary to avoid this if statement.

PR #1683, opened 2023-10-13
fix(cmake): Track dependencies for merged static archive

### LINKED ISSUE #1678: Merged static lib libavif.a is sometimes not updated when ext/aom/build.libavif/libaom.a changes

fdintino:
Fixes #1678 and #1679.

For static builds, the libavif code gets compiled into an intermediate static library target, `avif_internal`. This is then combined with any static platform and codec libraries to generate libavif.a / avif.lib.

Instead of executables linking against `avif ${AVIF_CODEC_LIBRARIES} ${AVIF_PLATFORM_LIBRARIES}`, they now link against `${AVIF_LINK_LIBRARIES}`. For shared builds this is the same as the old value, but for static builds it now includes `avif_static` (which ensures that the merge command is run) as well as any non-static libraries in `AVIF_CODEC_LIBRARIES` or `AVIF_PLATFORM_LIBRARIES`.


wantehchang:
Frankie: Thank you for writing this pull request. I tested it. It has one problem: it always executes the "Merge static libraries" step, even when nothing has changed.

Here is the output of the "ninja -v" command when nothing has changed:

$ ninja -v
[1/1] cd /home/wtc/libavif.1/libavif/build && /usr/bin/cmake -E rename /home/wtc/libavif.1/libavif/build/libavif.a /home/wtc/libavif.1/libavif/build/libavif.a.tmp && /usr/bin/cmake -E echo CREATE /home/wtc/libavif.1/libavif/build/libavif.a >script.ar && /usr/bin/cmake -E echo ADDLIB /home/wtc/libavif.1/libavif/build/libavif.a.tmp >>script.ar && /usr/bin/cmake -E echo ADDLIB /home/wtc/libavif.1/libavif/ext/dav1d/build/src/libdav1d.a >>script.ar && /usr/bin/cmake -E echo ADDLIB /home/wtc/libavif.1/libavif/ext/libgav1/build/libgav1.a >>script.ar && /usr/bin/cmake -E echo ADDLIB /home/wtc/libavif.1/libavif/ext/SVT-AV1/Bin/Release/libSvtAv1Enc.a >>script.ar && /usr/bin/cmake -E echo ADDLIB /home/wtc/libavif.1/libavif/ext/aom/build.libavif/libaom.a >>script.ar && /usr/bin/cmake -E echo ADDLIB /home/wtc/libavif.1/libavif/ext/libyuv/build/libyuv.a >>script.ar && /usr/bin/cmake -E echo ADDLIB /home/wtc/libavif.1/libavif/ext/libwebp/build/libsharpyuv.a >>script.ar && /usr/bin/cmake -E echo SAVE >>script.ar && /usr/bin/cmake -E echo END >>script.ar && /usr/bin/ar -M <script.ar && /usr/bin/cmake -E remove script.ar && /usr/bin/cmake -E remove /home/wtc/libavif.1/libavif/build/libavif.a.tmp

fdintino:
@wantehchang The behavior you were seeing was caused by the fact that we were modifying libavif.a after it had been "built"—which caused the build system to always think it had been changed when re-running a build. I think I've come up with a satisfactory solution. I'm repurposing the avif_internal target (which was formerly only used for tests with `-DBUILD_SHARED_LIBS=ON` as one of the libraries that gets merged into libavif.a.

cmake/Modules/merge_static_libs.cmake
```diff
@@ -1,33 +1,62 @@
...
+    POST_BUILD
+    COMMENT "Merge static libraries"
+    COMMAND ${CMAKE_COMMAND} -E remove $<TARGET_FILE:${target}>
+  )
+
   if(APPLE)
     add_custom_command(
       TARGET ${target}
       POST_BUILD
-      COMMENT "Merge static libraries with libtool"
```
vrabaud, line 14: Maybe keep all the comments?

CMakeLists.txt
```diff
@@ -104,6 +104,7 @@ set(AVIF_PLATFORM_DEFINITIONS)
 set(AVIF_PLATFORM_INCLUDES)
 set(AVIF_PLATFORM_SYSTEM_INCLUDES)
 set(AVIF_PLATFORM_LIBRARIES)
+set(AVIF_LINK_LIBRARIES)
```
vrabaud, line 107: Please add a comment on what this variable does: we are starting to have a lot of those ..

cmake/Modules/merge_static_libs.cmake
```diff
@@ -1,33 +1,62 @@
...
-      COMMAND ${CMAKE_COMMAND} -E rename $<TARGET_FILE:${target}> $<TARGET_FILE:${target}>.tmp
       COMMAND ${CMAKE_COMMAND} -E echo CREATE $<TARGET_FILE:${target}> >script.ar
-      COMMAND ${CMAKE_COMMAND} -E echo ADDLIB $<TARGET_FILE:${target}>.tmp >>script.ar
     )
 
     foreach(lib ${libs})
-      add_custom_command(TARGET ${target} POST_BUILD
+      add_custom_command(
+        TARGET ${target}
+        POST_BUILD
```
vrabaud, line 59: Did you run cmake-format on that file, as indicated in thee README? Thx

vrabaud (commented):
That LGTM

cmake/Modules/merge_static_libs.cmake
fdintino: Oh, I overlooked that. I made one commit that incorporated your feedback, and then a second commit where a ran cmake-format on CMakeLists.txt and merge_static_libs.cmake

vrabaud: Thx! Can you please rebase? I just ran it with https://github.com/AOMediaCodec/libavif/pull/1692. This way, it will remove some noise. Otherwise, LGTM

fdintino: Done.

wantehchang:
Hi Frankie,

Thank you very much for writing this pull request. After considering the pros and cons of the combined archive library, I have decided to abandon this approach in favor of the proper solution (to set Libs.private in libavif.pc correctly).

I think the following two properties are important:
- libavif.a and libavif.so should have the same contents.
- Consider a client that uses both libavif and libaom and links with `-lavif -laom`. If libaom.a is a newer version than the version of libaom that libavif.a was built against, the newer version of libaom should be used.

I believe you have some CI builds that assume libavif.a is a combined archive library. I will coordinate the revert of the combined archive library with you so that it won't break your CI builds.

Note: I verified that this PR fixes https://github.com/AOMediaCodec/libavif/issues/1678, but it doesn't fix https://github.com/AOMediaCodec/libavif/issues/1679. Here is an experiment:

$ rm avifenc
$ ninja -v
[1/1] : && /usr/bin/c++ -g  CMakeFiles/avifenc.dir/apps/avifenc.c.o -o avifenc  libavif_apps.a  libavif.a  /home/wtc/libavif.1/libavif/ext/dav1d/build/src/libdav1d.a  /home/wtc/libavif.1/libavif/ext/libgav1/build/libgav1.a  /home/wtc/libavif.1/libavif/ext/SVT-AV1/Bin/Release/libSvtAv1Enc.a  /home/wtc/libavif.1/libavif/ext/aom/build.libavif/libaom.a  /home/wtc/libavif.1/libavif/ext/libyuv/build/libyuv.a  /home/wtc/libavif.1/libavif/ext/libwebp/build/libsharpyuv.a  -lm  -ldl  /usr/lib/x86_64-linux-gnu/libpng.so  /usr/lib/x86_64-linux-gnu/libz.so  /usr/lib/x86_64-linux-gnu/libjpeg.so && :

fdintino:
That's certainly a possibility for system libraries, but what about local dependencies for static builds? My original reason for adding this to the build is for CI windows, where the pc file would be no help even with system static libraries.

fdintino:
In your demonstration you deleted avifenc, and then it built only the file that was deleted (no other files depend on the executable). Isn't that the expected behavior?

wantehchang:
The title of https://github.com/AOMediaCodec/libavif/issues/1679 is "When linking avifenc with the merged static lib libavif.a, we still link with libaom.a, libdav1d.a, etc."

Note that the `ninja -v `output has not only libavif.a (a combined archive library) but also libdav1d.a, libgav1.a, libSvtAv1Enc.a, and libaom.a.

fdintino:
I see. Sorry, I confused the two pull requests. Now that we have a static library that isn't combined, it was fairly easy to fix that. The linker is permissive of duplicate libraries passed into linker flags, and combined archives aren't materially different from the libraries passed in individually. For executables and shared libraries, it will only include the required symbols. You can experiment by taking the build command from ninja there and tacking on unrelated static libraries...the binary will remain the same. Changing the order of the libraries does result in non-identical binary files because the symbols are encountered in a different order, but they function the same so long as none of the libraries depend on each other. And it looks as though swapping libavif_internal.a with libavif.a has roughly the same effect: the files are not exactly identical, but they produce the same output with `nm`.

vrabaud:
I'd personally be totally ok with having the current behavior unchanged and have an option to fit your needs. Please mark this PR as draft and let's focus on the new one. Thx !!

wantehchang:
Hi Frankie,

Sorry about the late reply. I have been busy in the past few days.

Since libpng depends on zlib, it seems that libpng needs to solve a similar problem (on Windows at least). I am wondering what libpng's solution is, and whether we can apply that solution to libavif.

Re: pkg-config being unavailable on Windows: I am wondering if we can use vcpkg as a substiture for pkg-config.

fdintino:
I think you're right about pkg-config, and it's true that vcpkg is capable of taking advantage of a pkg-config file. And I think everyone is agreed that libavif.pc ought to document the shared and static dependencies. But I don't think that's incompatible with a desire to create a full-static build, so long as that's not the default. I've been working on a pull request that does both, and fixes some bugs I encountered as I did more extensive testing. I'll cross-reference once I've opened the PR.

fdintino:
libpng only has the one dependency, so if you look for instructions on how to build libpng on windows, it typically has you build that as a separate project and then has you copy over the static library file. I looked at projects like gdal and opencv, since their situation seemed closer to libavif's (projects that have multiple third-party dependencies and that are often linked to statically). In both projects the third-party libraries are vendored, and in gdal there are scripts to generate headers that prefix third-party libraries' symbols to avoid conflicts when statically linking to applications that also depend on those libraries. These vendored third-party libraries have customized CMakeLists.txt that ease integration with the project as a whole and it is all built as one big cmake project with a bunch of `add_subdirectory` commands. My approach borrows from a few other open source projects that manage to support static builds without going quite that far.

PR #1686, opened 2023-10-16
Add a --qgainmap flag.

maryla-uc:
To control the encoding quality of the gain map.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -32,12 +32,10 @@ typedef struct
     avifCodecChoice codecChoice;
     int jobs;
     int targetSize;
-    avifBool qualityIsConstrained;        // true if quality explicitly set by the user
-    avifBool qualityAlphaIsConstrained;   // true if qualityAlpha explicitly set by the user
-    avifBool qualityGainMapIsConstrained; // true if qualityGainMap explicitly set by the user
+    avifBool qualityIsConstrained;      // true if quality explicitly set by the user
+    avifBool qualityAlphaIsConstrained; // true if qualityAlpha explicitly set by the user
     int overrideQuality;
     int overrideQualityAlpha;
-    int qualityGainMap;
     avifBool progressive; // automatic layered encoding (progressive) with single input
     avifBool layered;     // manual layered encoding by specifying each layer
     int layers;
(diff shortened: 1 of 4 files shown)


apps/avifenc.c
```diff
@@ -230,6 +239,10 @@ static void syntaxLong(void)
     printf("    --ignore-profile,--ignore-icc     : If the input file contains an embedded color profile, ignore it (no-op if absent)\n");
 #if defined(AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION)
     printf("    --ignore-gain-map                 : If the input file contains an embedded gain map, ignore it (no-op if absent)\n");
+    printf("    --qgainmap Q                      : Set quality for the gain map (%d-%d, where %d is lossless)\n",
```
vrabaud, line 242: Guard with #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)

apps/avifenc.c
```diff
@@ -32,10 +32,12 @@ typedef struct
     avifCodecChoice codecChoice;
     int jobs;
     int targetSize;
-    avifBool qualityIsConstrained;      // true if quality explicitly set by the user
-    avifBool qualityAlphaIsConstrained; // true if qualityAlpha explicitly set by the user
+    avifBool qualityIsConstrained;        // true if quality explicitly set by the user
+    avifBool qualityAlphaIsConstrained;   // true if qualityAlpha explicitly set by the user
+    avifBool qualityGainMapIsConstrained; // true if qualityGainMap explicitly set by the user
```
vrabaud, line 37: Do we want to guard that with #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP) ?

apps/avifenc.c
```diff
@@ -1813,6 +1858,17 @@ int main(int argc, char * argv[])
 #if defined(AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION)
         } else if (!strcmp(arg, "--ignore-gain-map")) {
             settings.ignoreGainMap = AVIF_TRUE;
+        } else if (!strcmp(arg, "--qgainmap")) {
+            NEXTARG();
+            int qualityGainMap = atoi(arg);
+            if (qualityGainMap < AVIF_QUALITY_WORST) {
```
vrabaud, line 1864: Please use AVIF_CLAMP

apps/avifenc.c
maryla-uc: AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION implies AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP.
I changed it to use AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION everywhere in this file for simplicity/consistency and added some comments.

apps/avifenc.c
maryla-uc: It's debatable, but I think having if-guards everywhere makes the code harder to read, and this variable shouldn't cause any problems if gain maps are off. You can't use the gain map api when gain maps are off anyway.

apps/avifenc.c
maryla-uc: It's defined in internal.h which is not available in apps/

apps/avifenc.c
vrabaud: Agreed. Plus, this feature will be official one day.

apps/avifenc.c
vrabaud: Ah, I thought it was internal to the project, not the lib...

vrabaud (approved):

PR #1688, opened 2023-10-16
Remove yuvRange assertion in avifImageCopySamples

y-guyon:
Add explanation in avifDecoderDecodeTiles().
Add avifrangetest.cc.

This issue of mismatched `colr`/OBU video range values was caught by avif_fuzztest_dec.cc which triggered the following failed assertion:

https://github.com/AOMediaCodec/libavif/blob/7a07e7206bd7012de8c26249f8e49b7880897ad3/src/avif.c#L200

Related to https://github.com/AOMediaCodec/libavif/pull/1687.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1484,6 +1484,7 @@ static avifResult avifDecoderDataAllocateGridImagePlanes(avifDecoderData * data,
         dstImage->yuvFormat = tile->image->yuvFormat;
         // Keep dstImage->yuvRange which is already set to its correct value
         // (extracted from the 'colr' box if parsed or from a Sequence Header OBU otherwise).
+
         if (!data->cicpSet) {
             data->cicpSet = AVIF_TRUE;
             dstImage->colorPrimaries = tile->image->colorPrimaries;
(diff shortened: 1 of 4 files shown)


src/read.c
```diff
@@ -4984,6 +4986,25 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder, uint32_t nextIma
...
+        // Section 2.3.4 of AV1 Codec ISO Media File Format Binding v1.2.0 says:
+        //   the full_range_flag in the colr box shall match the color_range
+        //   flag in the Sequence Header OBU.
+        // See https://aomediacodec.github.io/av1-isobmff/v1.2.0.html#av1codecconfigurationbox-semantics.
+        // If a 'colr' box of colour_type 'nclx' was parsed, a mismatch between
+        // the 'colr' decoder->image->yuvRange and the AV1 OBU
+        // tile->image->yuvRange should be treated as an error.
+        // However codec_svt.c was not encoding the color_range field for
+        // multiple years, so there probably are files in the wild that will
+        // fail decoding if this is enforced. Thus this pattern is allowed.
```
wantehchang, line 4998: Yannis: SVT-AV1 is not as commonly used as libaom is with libavif, so I think we can try to be strict and only give in if a widely-used application such as Chrome gets too many decoding failures.

I only saw this pull request today and have not studied this and https://github.com/AOMediaCodec/libavif/pull/1687 in detail, so please consider this and I defer the decision to you. Thanks.

y-guyon: It seems simpler to leave it as is for now than to spend the time looking for the frequency of such invalid files. libavif can be stricter on this topic in a few years.

PR #1693, opened 2023-10-18
Allow UTF8 on windows using setlocale.

vrabaud:
This is a different approach from https://github.com/AOMediaCodec/libavif/pull/1685 that uses setlocale

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -70,6 +70,8 @@ MAIN()
     uint32_t frameIndex = 0;
     uint32_t imageSizeLimit = AVIF_DEFAULT_IMAGE_SIZE_LIMIT;
     uint32_t imageDimensionLimit = AVIF_DEFAULT_IMAGE_DIMENSION_LIMIT;
+    int returnCode = 1;
+    avifDecoder * decoder = NULL;
 
     if (argc < 2) {
         syntax();
@@ -227,7 +229,7 @@ MAIN()
             return 1;
         }
 
-        avifDecoder * decoder = avifDecoderCreate();
+        decoder = avifDecoderCreate();
         if (!decoder) {
             fprintf(stderr, "Memory allocation failure\n");
             return 1;
@@ -305,8 +307,7 @@ MAIN()
            jobs,
            (jobs == 1) ? "" : "s");
 
-    int returnCode = 1;
-    avifDecoder * decoder = avifDecoderCreate();
+    decoder = avifDecoderCreate();
     if (!decoder) {
         fprintf(stderr, "Memory allocation failure\n");
         goto cleanup;
@@ -378,7 +379,7 @@ MAIN()
     returnCode = 0;
 
 cleanup:
-    if (returnCode != 0) {
+    if (returnCode != 0 && decoder != NULL) {
         avifDumpDiagnostics(&decoder->diag);
     }
     if (decoder != NULL) {
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -14,14 +14,13 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define DEFAULT_JPEG_QUALITY 90
+#if defined(_WIN32)
+#include <locale.h>
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#endif
 
-#define NEXTARG()                                                     \
-    if (((argIndex + 1) == argc) || (argv[argIndex + 1][0] == '-')) { \
-        fprintf(stderr, "%s requires an argument.", arg);             \
-        return 1;                                                     \
-    }                                                                 \
-    arg = argv[++argIndex]
+#define DEFAULT_JPEG_QUALITY 90
 
 static void syntax(void)
 {
@@ -53,7 +52,7 @@ static void syntax(void)
     avifPrintVersions();
 }
 
-int main(int argc, char * argv[])
+MAIN()
 {
     const char * inputFilename = NULL;
     const char * outputFilename = NULL;
@@ -77,6 +76,7 @@ int main(int argc, char * argv[])
         return 1;
     }
 
+    INIT_ARGV()
     int argIndex = 1;
     while (argIndex < argc) {
         const char * arg = argv[argIndex];
@@ -384,5 +384,6 @@ cleanup:
     if (decoder != NULL) {
         avifDecoderDestroy(decoder);
     }
+    FREE_ARGV()
     return returnCode;
 }
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -386,10 +386,10 @@ MAIN()
     returnCode = 0;
 
 cleanup:
-    if (returnCode != 0 && decoder != NULL) {
-        avifDumpDiagnostics(&decoder->diag);
-    }
     if (decoder != NULL) {
+        if (returnCode != 0) {
+            avifDumpDiagnostics(&decoder->diag);
+        }
         avifDecoderDestroy(decoder);
     }
     FREE_ARGV()
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -22,6 +22,13 @@
 
 #define DEFAULT_JPEG_QUALITY 90
 
+#define NEXTARG()                                                     \
+    if (((argIndex + 1) == argc) || (argv[argIndex + 1][0] == '-')) { \
+        fprintf(stderr, "%s requires an argument.", arg);             \
+        goto cleanup;                                                 \
+    }                                                                 \
+    arg = argv[++argIndex]
+
 static void syntax(void)
 {
     printf("Syntax: avifdec [options] input.avif output.[jpg|jpeg|png|y4m]\n");
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -18,14 +18,12 @@
 // for setmode()
 #include <fcntl.h>
 #include <io.h>
+#include <locale.h>
+#define WIN32_LEAN_AND_MEAN
+// Avoid the DEFAULT_QUALITY macro redefinition warning caused by including wingdi.h.
+#define NOGDI
+#include <windows.h>
 #endif
-
-#define NEXTARG()                                                     \
-    if (((argIndex + 1) == argc) || (argv[argIndex + 1][0] == '-')) { \
-        fprintf(stderr, "%s requires an argument.", arg);             \
-        goto cleanup;                                                 \
-    }                                                                 \
-    arg = argv[++argIndex]
 
 typedef struct
 {
@@ -1382,7 +1380,7 @@ static avifBool avifEncodeImages(avifSettings * settings,
     return AVIF_TRUE;
 }
 
-int main(int argc, char * argv[])
+MAIN()
 {
     if (argc < 2) {
         syntaxShort();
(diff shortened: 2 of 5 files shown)


apps/shared/avifutil.h
```diff
@@ -10,6 +10,59 @@
...
+#if defined(_WIN32)
+#define INIT_ARGV()                                                                                           \
+    char * argvAll = NULL;                                                                                    \
+    char ** argv = NULL;                                                                                      \
+    if (setlocale(LC_ALL, ".UTF8") == NULL) {                                                                 \
+        fprintf(stderr, "setlocale failed\n");                                                                \
+        goto cleanup;                                                                                         \
+    }                                                                                                         \
+    argvAll = (char *)malloc(1024 * argc * sizeof(*argvAll));                                                 \
+    argv = (char **)malloc(argc * sizeof(*argv));                                                             \
```
wantehchang, line 45: Check for malloc() failures.

apps/shared/avifutil.h
```diff
@@ -10,6 +10,59 @@
 extern "C" {
 #endif
 
+// NEXTARG(), MAIN(), INIT_ARGV(), FREE_ARGV() for UTF8-aware command line parsing.
+#if defined(_WIN32)
```
wantehchang, line 14: Optional: It would be good to add a comment to reassure the reader that the only difference between these two implementations is `L'-'` vs. `'-'`.

apps/shared/avifutil.h
```diff
@@ -10,6 +10,59 @@
...
+#define NEXTARG()                                                     \
+    if (((argIndex + 1) == argc) || (argv[argIndex + 1][0] == '-')) { \
+        fprintf(stderr, "%s requires an argument.", arg);             \
+        goto cleanup;                                                 \
+    }                                                                 \
+    arg = argv[++argIndex]
+#endif
+
+#if defined(_WIN32)
+#define MAIN() int wmain(int argc, wchar_t * _argv[])
```
wantehchang, line 31: Nit: I suggest using `wargv` instead of `_argv`. I missed the initial underscore when I first read it.

Note: symbols that start with `_` or `__` are reserved for the C implementation. So that's another reason to avoid `_argv`.

apps/avifenc.c
```diff
@@ -1442,6 +1440,8 @@ int main(int argc, char * argv[])
     avifImage ** gridCells = NULL;
     avifImage * gridSplitImage = NULL; // used for cleanup tracking
 
+    INIT_ARGV()
```
wantehchang, line 1443: I suggest we call `INIT_ARGV()` as early as possible in the main() function.

At least, we should try to call `setlocale()` as early as possible in the main() function.

apps/shared/avifutil.h
```diff
@@ -10,6 +10,59 @@
 extern "C" {
 #endif
 
+// NEXTARG(), MAIN(), INIT_ARGV(), FREE_ARGV() for UTF8-aware command line parsing.
+#if defined(_WIN32)
+#define NEXTARG()                                                      \
+    if (((argIndex + 1) == argc) || (argv[argIndex + 1][0] == L'-')) { \
+        fprintf(stderr, "%s requires an argument.", arg);              \
+        goto cleanup;                                                  \
```
wantehchang, line 18: IMPORTANT: set `returnCode` to 1 before `goto cleanup`. Make the same change to line 25.

apps/avifdec.c
```diff
@@ -387,5 +388,6 @@ int main(int argc, char * argv[])
     if (decoder != NULL) {
         avifDecoderDestroy(decoder);
     }
+    FREE_ARGV()
```
wantehchang, line 391: IMPORTANT: At line 385, I think we also need to check `decoder` now, because we may goto cleanup from inside `NEXT_ARGV()` before `decoder` is allocated.

We can probably just move lines 385-387 right after line 388.

apps/avifenc.c
```diff
@@ -18,15 +18,13 @@
...
+#define WIN32_LEAN_AND_MEAN
+// Avoid the DEFAULT_QUALITY macro redefinition warning caused by including wingdi.h.
+#define NOGDI
+#include <windows.h>
 #endif
 
-#define NEXTARG()                                                     \
-    if (((argIndex + 1) == argc) || (argv[argIndex + 1][0] == '-')) { \
-        fprintf(stderr, "%s requires an argument.", arg);             \
-        goto cleanup;                                                 \
```
wantehchang, line 26: Hmm... I think we had a bug here. We should set `returnCode` to 1 before `goto cleanup`.

tests/test_cmd.sh
```diff
@@ -68,6 +70,8 @@ pushd ${TMP_DIR}
   "${AVIFENC}" -s 8 "${INPUT_Y4M}" -o "${ENCODED_FILE}"
   "${AVIFDEC}" "${ENCODED_FILE}" "${DECODED_FILE}"
   "${ARE_IMAGES_EQUAL}" "${INPUT_Y4M}" "${DECODED_FILE}" 0 && exit 1
+  cp ${INPUT_Y4M} ${INPUT_UTF8_Y4M}
+  "${AVIFENC}" -s 8 "${INPUT_UTF8_Y4M}" -o "${ENCODED_UTF8_FILE}"
```
wantehchang, line 74: Should we repeat lines 71-72 for the UTF8 files? This will test avifdec's UTF-8 file pathname support.

apps/shared/avifutil.h
```diff
@@ -10,6 +10,59 @@
...
+        goto cleanup;                                                                                         \
+    }                                                                                                         \
+    argvAll = (char *)malloc(1024 * argc * sizeof(*argvAll));                                                 \
+    argv = (char **)malloc(argc * sizeof(*argv));                                                             \
+    for (int i = 0; i < argc; ++i) {                                                                          \
+        argv[i] = argvAll + 1024 * i;                                                                         \
+        int rc = WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, _argv[i], -1, argv[i], 1024, NULL, NULL); \
+        if (rc == 0) {                                                                                        \
+            fprintf(stderr, "WideCharToMultiByte() failed\n");                                                \
+            goto cleanup;                                                                                     \
```
wantehchang, line 51: Optional: Since we should ideally call `INIT_ARGV()` first thing in the main() function, we can probably just free `argv` and `argvAll` and return 1, instead of goto cleanup.

This will allow us to avoid moving the `decoded` array before `INIT_ARGV()` in tests/gtest/are_images_equal.cc.

wantehchang (approved):
LGTM. Thank you very much for writing this pull request. I suggest some changes.

apps/shared/avifutil.h
wantehchang: It seems that the _WIN32 version of `NEXTARG()` is not needed. We should not use `L'-'` on Windows.

apps/avifdec.c
vrabaud: Right, it is set to NULL at first which does not trigger the destroy line 389.

apps/avifenc.c
vrabaud: Indeed, fixed in https://github.com/AOMediaCodec/libavif/pull/1699. Other paths were wrong too.

apps/avifenc.c
vrabaud: Indeed: I wanted to use the cleanup goto by default but I'll free things by hand then. Done

apps/shared/avifutil.h
vrabaud: Right, I fixed this logic in https://github.com/AOMediaCodec/libavif/pull/1699

apps/shared/avifutil.h
vrabaud: Done

apps/shared/avifutil.h
vrabaud: Done

tests/test_cmd.sh
vrabaud: Indeed, done.

apps/shared/avifutil.h
```diff
@@ -10,6 +10,55 @@
 extern "C" {
 #endif
 
+#define NEXTARG()                                                     \
+    if (((argIndex + 1) == argc) || (argv[argIndex + 1][0] == '-')) { \
+        fprintf(stderr, "%s requires an argument.", arg);             \
+        goto cleanup;                                                 \
+    }                                                                 \
+    arg = argv[++argIndex]
```
wantehchang, line 18: Optional: I think it would be easier to understand (at the cost of code duplication) if this macro is defined in avifdec.c and avifenc.c as before. The reason is that this macro uses the local variable `argIndex`.

I am just bring this up for your consideration. It is fine to leave it here.

wantehchang (commented):
Vincent: I will take a look at the new commit when I get to work. First a comment for your opinion.

apps/shared/avifutil.h
vrabaud: Done

apps/avifdec.c
```diff
@@ -378,11 +386,12 @@ int main(int argc, char * argv[])
     returnCode = 0;
 
 cleanup:
-    if (returnCode != 0) {
+    if (returnCode != 0 && decoder != NULL) {
```
wantehchang, line 389: Just wanted to confirm that you prefer to not nest these two if statements as follows:

```
    if (decoder != NULL) {
        if (returnCode != 0) {
            avifDumpDiagnostics(&decoder->diag);
        }
        avifDecoderDestroy(decoder);
    }
```

wantehchang (approved):
LGTM! Thanks a lot for all your work on this issue.

I have been trying to find answers to two questions.
1. Is it possible to configure a Windows command-line program at build time so that its main() function receives the argv arguments as UTF-8 strings and the code page is UTF-8?
2. Does the setlocale() approach require the use of Microsoft's Universal C runtime? I.e., does the setlocale() approach work with other compilers such as Clang and MINGW?

For the first question, I found a Microsoft page "[Use UTF-8 code pages in Windows apps](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page)". We can try the Fusion manifest for an unpackaged Win32 app mentioned in that page.

apps/avifdec.c
vrabaud: Done

wantehchang (approved):

vrabaud:
For 2, my understanding is that it will work with all compilers.
On Linux, the default locale is UTF8 so all is fine. Just run `locale` on the command line
On Window, it is whatever the user has, hence what has to be done here. It is just a matter of default environment, not of default library behavior from what I understand (https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-170).

apps/avifenc.c
wantehchang: Thanks a lot!

wantehchang:
Vincent: Good news (or perhaps bad news?). I found a solution that requires no changes to the source code.

Please test the following.

Go back to a revision before your fix. Reproduce the bug with `avifenc.exe`.

Change to the directory where `avifenc.exe` resides.

Save the "Fusion manifest for an unpackaged Win32 app" in the ["Use UTF-8 code pages in Windows apps](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page#examples)" page as the file `MANIFEST`.

Then run the command:
```
mt.exe -manifest MANIFEST -outputresource:avifenc.exe;#1
```

Now try `avifenc.exe` again.

PR #1694, opened 2023-10-19
Check avifAlloc() result in avifArrayPush*()

y-guyon:
Propagate `avifResult` to call sites.
Remove unused `avifArrayPushIndex()` and `avifArrayPush()`.
Rename `avifArrayPushPtr()` to `avifArrayPush()`.
Rename `avifMetaFindItem()` to `avifMetaFindOrCreateItem()`.
Fix `iref` checking item ID != 0 too late.
Add entry to `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * AVIF sequences encoded by libavif will now also have the "avio" brand when
   there is at least one track made only of AV1 keyframes.
 * Fix SVT-AV1 codec interface which was not setting video range at encoding.
+* Any item ID being 0 in an "iref" box with version 0 or 1 is now treated as an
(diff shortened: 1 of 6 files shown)


vrabaud (approved):
line 2557 is for "Fix iref checking item ID != 0 too late." right?

y-guyon:
> line 2557 is for "Fix iref checking item ID != 0 too late." right?

No, line 2640.  
In `avifParseItemReferenceBox()`, `avifROStreamReadU16/32()` is used to read an item ID.  
Previously, `avifMetaFindItem()` was checking for id!=0. I split that function into `avifCheckItemID()` and `avifMetaFindOrCreateItem()`.  
`avifMetaFindItem()` was called only if `fromID && toID`, ignoring id==0 rather than returning an error. Moving `avifCheckItemID()` closer to `avifROStreamReadU16/32()` changed the behavior.

PR #1696, opened 2023-10-19
Add CONFIG_BLOCK_256 av2C parsing path for avm

y-guyon:


src/obu.c
```diff
@@ -361,7 +369,13 @@ static avifBool parseAV2SequenceHeader(avifBits * bits, avifSequenceHeader * hea
     AVIF_CHECK(parseSequenceHeaderProfile(bits, header));
 
     // See av1_read_sequence_header() in avm.
-    AVIF_CHECK(parseSequenceHeaderFeatures(bits, header));
+    AVIF_CHECK(parseSequenceHeaderFrameMaxDimensions(bits, header));
+#if CONFIG_BLOCK_256
```
wantehchang, line 373: The `CONFIG_BLOCK_256` and `CONFIG_LR_FLEX_SYNTAX` macros are not defined by libavif's cmake build system. Does this mean the code here is intended to be compiled by alternative build systems that define these macros?

src/obu.c
```diff
@@ -361,7 +369,13 @@ static avifBool parseAV2SequenceHeader(avifBits * bits, avifSequenceHeader * hea
...
+    if (!avifBitsRead(bits, 1)) // BLOCK_256X256
+#endif
+        avifBitsRead(bits, 1); // BLOCK_128X128
```
wantehchang, line 376: Nit/optional: It is a little confusing to nest `#if` and `if` this way. We probably should duplicate some code to avoid the nesting:

```
#if CONFIG_BLOCK_256
    if (!avifBitsRead(bits, 1)) // BLOCK_256X256
        avifBitsRead(bits, 1); // BLOCK_128X128
#else
    avifBitsRead(bits, 1); // BLOCK_128X128
#endif
```

Alternatively we could consider using the solution at lines 380-387. Does any compiler warn about the empty `if` block there when `CONFIG_LR_FLEX_SYNTAX` is equal to 0?

wantehchang (approved):
LGTM.

src/obu.c
y-guyon: Just as for `CONFIG_LR_FLEX_SYNTAX`. It also helps keeping the parsing implementation in sync visually with libavm.

src/obu.c
y-guyon: I need this synced internally so I will merge now but feel free to send cosmetic changes afterwards. All approaches you mentioned are fine with me.

PR #1698, opened 2023-10-20
Add libyuv to tests/oss-fuzz/avif_decode_fuzzer

y-guyon:
Also add timing report in `tests/oss-fuzz/repro_fuzz.cc` for convenience.

Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=63360.

### DIFF
--- a/tests/oss-fuzz/repro_fuzz.cc
+++ b/tests/oss-fuzz/repro_fuzz.cc
@@ -1,11 +1,15 @@
 // Copyright 2023 Google LLC
 // SPDX-License-Identifier: BSD-2-Clause
 
+#include <chrono>
 #include <cstddef>
 #include <cstdint>
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


vrabaud (approved):

tests/oss-fuzz/repro_fuzz.cc
```diff
@@ -1,12 +1,16 @@
...
 
+#include <chrono>
 #include <cstddef>
 #include <cstdint>
 #include <fstream>
 #include <iostream>
 #include <vector>
 
+using seconds = std::chrono::duration<double>;
+using chrono = std::chrono::high_resolution_clock;
```
wantehchang, line 12: "clock" would be a clearer type alias. "chrono" looks like "std::chrono".

Also these two type aliases probably should follow the naming convention of the Google C++ Style Guide?

PR #1699, opened 2023-10-20
Simplify returnCode setting.

vrabaud:
It is now 1 by default and changed to 0 when all is fine. This also fixes a few paths that would not clean up properly.


y-guyon (approved):
Good idea. We could also use a sub-function, to `return` instead of `goto`.

PR #1700, opened 2023-10-20
Remove abort() on avifAlloc() failure

y-guyon:
See #820.

### DIFF
--- a/src/mem.c
+++ b/src/mem.c
@@ -3,10 +3,12 @@
 
 #include "avif/avif.h"
 
+#include <assert.h>
 #include <stdlib.h>
 
 void * avifAlloc(size_t size)
 {
+    assert(size != 0); // Implementation-defined. See https://en.cppreference.com/w/cpp/memory/c/malloc
     return malloc(size);
 }
 
--- a/src/mem.c
+++ b/src/mem.c
@@ -7,12 +7,7 @@
 
 void * avifAlloc(size_t size)
 {
-    void * out = malloc(size);
-    if (out == NULL) {
-        // TODO(issue #820): Remove once all calling sites propagate the error as AVIF_RESULT_OUT_OF_MEMORY.
-        abort();
-    }
-    return out;
+    return malloc(size);
 }
 
 void avifFree(void * p)
--- a/tests/gtest/avifallocationtest.cc
+++ b/tests/gtest/avifallocationtest.cc
@@ -189,10 +189,6 @@ TEST(EncodingTest, MaximumInvalidDimensions) {
 }
 
 TEST(AvifAllocTest, Extremes) {
-  void* p0 = avifAlloc(0);
-  EXPECT_NE(p0, nullptr);
-  avifFree(p0);
-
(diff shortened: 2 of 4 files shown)


tests/gtest/avifallocationtest.cc
```diff
@@ -188,5 +188,17 @@ TEST(EncodingTest, MaximumInvalidDimensions) {
                AVIF_RESULT_UNSUPPORTED_DEPTH);
 }
 
+TEST(AvifAllocTest, Extremes) {
+  void* p0 = avifAlloc(0);
```
jzern, line 192: https://man7.org/linux/man-pages/man3/malloc.3.html

*Nonportable behavior
       The behavior of these functions when the requested size is zero
       is glibc specific; other implementations may return NULL without
       setting errno, and portable POSIX programs should tolerate such
       behavior.  See realloc(3p).*

jzern (approved):

src/mem.c
```diff
@@ -3,16 +3,13 @@
...
 
 void * avifAlloc(size_t size)
 {
-    void * out = malloc(size);
-    if (out == NULL) {
-        // TODO(issue #820): Remove once all calling sites propagate the error as AVIF_RESULT_OUT_OF_MEMORY.
-        abort();
-    }
-    return out;
+    assert(size != 0); // Implementation-defined. See https://en.cppreference.com/w/cpp/memory/c/malloc
```
wantehchang, line 11: It is not an error to call malloc(0). The implementation can either return NULL or return some non-null pointer that needs to be freed.

Since avifAlloc() is a public function (this may have been a mistake, but it's done), we should document that it doesn't allow size=0. Alternatively, our implementation can choose to return NULL if size is 0 and we can document this behavior.

wantehchang (commented):
LGTM. Please see my comment in mem.c.

src/mem.c
wantehchang: I found that the code in the apps and examples directories doesn't use avifAlloc() and avifFree() (I fixed the two exceptions in https://github.com/AOMediaCodec/libavif/pull/1702), so very likely avifAlloc() and avifFree() were exported in avif.h unintentionally.

We may want to add a comment to avif.h to discourage libavif clients from using avifAlloc() and avifFree() in their code.

y-guyon: I believe deprecating `avifAlloc()` and `avifFree()` from avif.h is fine. We can keep the signatures for a few years/months to be ABI safe but `avifAlloc()` should always return `NULL` to make sure nobody uses it. Then remove them from avif.h.  
The downside is having to find new internal names for the internal.h version.

wantehchang: We don't need to remove `avifAlloc()` or `avifFree()` from avif.h. We just need to document that they are intended for internal use by the libavif library, and we disallow passing size=0 to `avifAlloc()` (so that the libavif library can report memory allocation failure whenever `avifAlloc()` returns NULL).

y-guyon: Deprecating sounds simpler to me but documenting is fine too.

PR #1702, opened 2023-10-20
Don't call avifAlloc,avifFree in apps and examples

wantehchang:
They can call malloc(), calloc(), realloc(), and free().

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -687,11 +687,11 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
     return AVIF_TRUE;
 }
 
-// Returns NULL if a memory allocation failed.
+// Returns NULL if a memory allocation failed. The return value should be freed with free().
 static char * avifStrdup(const char * str)
 {
     size_t len = strlen(str);
-    char * dup = avifAlloc(len + 1);
+    char * dup = malloc(len + 1);
     if (!dup) {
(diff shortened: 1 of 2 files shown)


vrabaud (approved):

apps/avifenc.c
```diff
@@ -687,11 +687,11 @@ static avifBool readEntireFile(const char * filename, avifRWData * raw)
     return AVIF_TRUE;
 }
 
-// Returns NULL if a memory allocation failed.
+// Returns NULL if a memory allocation failed. The return value should be freed with free().
 static char * avifStrdup(const char * str)
 {
     size_t len = strlen(str);
-    char * dup = avifAlloc(len + 1);
+    char * dup = malloc(len + 1);
```
vrabaud, line 694: entry->value uses that below and is freed through avifFree. So it should be avifAlloc-ed right?

wantehchang: avifFree() was not used in this file before this pull request. The return values of avifStrdup() are freed by free() here:

```
static void avifCodecSpecificOptionsFree(avifCodecSpecificOptions * options)
{
    while (options->count) {
        --options->count;
        free(options->keys[options->count]);
        free(options->values[options->count]);
    }
    ...
}
```

jzern (commented):
lgtm

PR #1703, opened 2023-10-20
Speed up some builds.

vrabaud:
- dav1d does not need to be compiled with tools/examples
- fuzztest tests can be builts independently from gtest tests

This will speed up the oss-fuzz build.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -75,7 +75,7 @@ macro(add_avif_gtest_with_data TEST_NAME)
     add_test(NAME ${TEST_NAME} COMMAND ${TEST_NAME} ${CMAKE_CURRENT_SOURCE_DIR}/data/)
 endmacro()
 
-if(AVIF_ENABLE_GTEST)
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -171,7 +173,7 @@ endif()
 ################################################################################
 # Experimental FuzzTest support (Linux only)
 
-if(AVIF_ENABLE_GTEST AND AVIF_ENABLE_FUZZTEST)
+if(AVIF_ENABLE_FUZZTEST)
```
wantehchang, line 176: Optional: When the "else" branch is much shorter than the "if" branch, some people recommend inverting the conditional so that the shorter branch goes first. I think that style will make the code much easier to read here, at the cost of testing a negative condition:

```
if(NOT AVIF_ENABLE_FUZZTEST)
  message(STATUS "FuzzTest targets are disabled because AVIF_ENABLE_FUZZTEST is OFF.")
else()
  ... a very long block of code ...
endif()
```

wantehchang (approved):
LGTM.

PR #1704, opened 2023-10-21
Patch alpha_video.avif to fix the iref box

vigneshvg:
The iref box contains an invalid item_id of 0. This file started failing since `1fbea62`.

This file has always had this issue but it was ignored because it is an animated AVIF and the tracks information is correct. If we try to decode it as a still image by setting `requestedSource` to primary item, then avifdec (prior to `1fbea62`) decodes it as an image with alpha (which is not correct).

Patching the file to update the reference item_id fixes the Android CI failures on this file.


y-guyon (approved):
Thank you. How did you find out? Would it be possible to add a test for this?

vigneshvg:
> Thank you. How did you find out? Would it be possible to add a test for this?

There already is a test for this since this file is being used in the Android JNI decoder tests. Unfortunately they only run on changes to the android_jni/ subdirectory since they are too slow to be run on every PR. The Android CI has been broken since #1700.

PR #1708, opened 2023-10-24
feat: FindPkg.cmake fixes; .pc deps; option to bundle system static libs

fdintino:
This pull request is meant to replace #1683.

Overall, the purpose of this pull request is to address issues that cause static builds to fail. Many of the issues are due to the `find_package` modules not returning the list of static dependencies. For instance, building with `-DBUILD_SHARED_LIBS=OFF -DAVIF_LOCAL_LIBXML2=OFF` fails because of undefined symbols from libicu and liblzma (absent any library archive merging). The local libxml2 build doesn't have this problem because it's built without those dependencies, but rhel, debian, and homebrew libxml2 distributions have them. I went through and updated the `find_package` modules to take advantage of pkg-config where it's available. There were a few other issues I encountered when trying to do a static build with non-local dependencies for which I've included fixes. An enumerated list of the changes (I think it's exhaustive but I may have missed something):

- Adds a cmake option, `AVIF_STATIC_SYSTEM_LIBRARY_MERGE`, which will include non-local static libraries in the merged static archive. By default only local libraries are bundled.
- Sets `Requires`, `Libs`, and `Libs.private` fields to their correct values in libavif.pc, omitting any libraries that are bundled in libavif.a / avif.lib from `Libs.private`.
  - The unix static test was updated to test that one can compile and link against static libavif using pkg-config ([see here](https://github.com/fdintino/libavif/blob/52732574e29af72f7f2c8c776125ab56956d72dc/.github/workflows/ci-unix-static.yml#L153))
    
- Makes fixes to the find_package modules
  - Add modules for libjpeg and libxml2
  - Fix typos in `FindPkgConfig` variable names
    - `PREFIX_INCLUDEDIR` should be `PREFIX_INCLUDE_DIRS`
    - `PREFIX_LIBDIRS` should be `PREFIX_LIBRARY_DIRS`
  - If `pkg-config` finds the package, use the `PREFIX_LIBRARIES` it returns
  - Use the `STATIC_` variants of the above variables when building a static library
  - Set link directories returned by pkg-config on relevant targets
- When building local rav1e, extracts shared/static library dependencies from the pc file's contents, which will have the correct rust system dependencies for the build platform, including any differences between msvc, cygwin, and msys2. The hard-coded list of rav1e link libraries is only set for system rav1e that wasn't found by `pkg-config`.
- Sets the `LINKER_LANGUAGE` property to "CXX" on any targets that link against libgav1. This was previously only being done if `AVIF_LOCAL_LIBGAV1` was true, but it needs to be done for non-local libgav1 packages as well, to prevent the `libstdc++` linker errors that the original conditional was intended to fix.
- Tracks dependencies in cmake for any local libraries, so that changes in the ext directory cause the merged static avif archive to be rebuilt.

### DIFF
--- a/.github/workflows/ci-unix-static.yml
+++ b/.github/workflows/ci-unix-static.yml
@@ -17,6 +17,10 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
         build-type: [Release, Debug]
+        install-type: [local, system, system-merged]
+        include:
+          - install-type: system-merged
+            static-link-flags: "-lrav1e -lSvtAv1Enc -laom -lsharpyuv -lyuv -lgav1 -ldav1d"
 
     steps:
     - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
@@ -37,7 +41,7 @@ jobs:
       uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
       with:
         path: ext
-        key: ${{ runner.os }}-${{ runner.build-type }}-${{ hashFiles('ext/*.cmd') }}
+        key: ${{ runner.os }}-${{ matrix.build-type }}-${{ hashFiles('ext/*.cmd') }}
     - name: Setup cmake
       uses: jwlawson/actions-setup-cmake@187efd9581ed20ee4e03c0dfb9ac2cd5896c4835 # v1.14.1
       with:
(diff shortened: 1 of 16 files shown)


vrabaud:
I know this is a draft but it will most likely need to be split (like for bugfixes, renamings and the core of this PR).
Is there any way the official FindJpeg and such be used and the targets only modified to fit your needs? BTW, instead of variables, could we use the official targest JPEG::JPEG, PNG::PNG ...
Would it be simpler if we had a variable per lib doing the following: e.g. AVIF_LIBYUV can only have one of those 3 values: none, local, system?

fdintino:
You know, to be honest I've been working on obscure and private libraries with cmake for long enough that I totally forgot that they have a bunch of well-maintained FindPackage modules. I think I should be able to make do with the official ones and add the pkg-config transitive static library stuff in a function or macro.

I think your suggestion is a good one about standardizing on targets and generator expressions...I hesitated to do that because I was already changing a lot and didn't want to do any heavy-handed refactoring on top of that. But we already do that for a few things and I'm happy to follow that convention for anything I'm adding or changing. And to add imported/interface library targets with the appropriate properties for existing modules such as `Findrav1e.cmake`.

I'm not sure if collapsing `AVIF_LOCAL_LIBYUV` et. al. in the way you suggest is all that beneficial, since you end up with a bunch of `if(AVIF_LIBYUV STREQUAL "local")`, which doesn't end up being much more readable or terse than the original. My gut impression is that it's also not very idiomatic, but I'm also not really sure that there is such a thing as a cmake idiom.

An idea I had, which I recall seeing thrown around in the past on some issues in this repository, is to have the projects in the `ext` directory install into a common prefix…something like `ext/build`. I think a lot of the logic around adding include directories and libraries for local dependencies could be replaced with a single prepended path to `CMAKE_PREFIX_PATH`.

Do you know the reason that only zlib, libpng, and libjpeg are built with `add_subdirectory`. I can think of a few reasons why it might be preferable to have a separate command to build local libaom, but I'm curious which ones informed the initial decision.

fdintino:
There's still value in setting the correct `Requires`, `Libs`, and `Libs.private` in the pkg-config file, but the implementation will be different enough from what we have in this PR that it makes more sense to start over.

PR #1710, opened 2023-10-25
Disable all executables for the libwebp CMake

vrabaud:
This is just to speed up the CMake, as they are not built anyway.


ext/libsharpyuv.cmd
```diff
@@ -16,6 +16,6 @@ git checkout e2c85878f6a33f29948b43d3492d9cdaf801aa54
 
 mkdir build
 cd build
-cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release ..
+cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DWEBP_BUILD_ANIM_UTILS=OFF -DWEBP_BUILD_CWEBP=OFF -DWEBP_BUILD_DWEBP=OFF -DWEBP_BUILD_GIF2WEBP=OFF -DWEBP_BUILD_IMG2WEBP=OFF -DWEBP_BUILD_VWEBP=OFF -DWEBP_BUILD_WEBPINFO=OFF -DWEBP_BUILD_LIBWEBPMUX=OFF -DWEBP_BUILD_WEBPMUX=OFF -DWEBP_BUILD_EXTRAS=OFF  ..
```
wantehchang, line 19: This file is supposed to be both a shell script and a Windows batch file. (On the other hand I remember I never successfully ran these *.cmd files successfully in the Command Prompt on Windows.) Does the `-` line continuation work in both modes? It may be safer to just let this be a very long line.

y-guyon (approved):

ext/libsharpyuv.cmd
wantehchang: My mistake. Thanks.

wantehchang (approved):

PR #1713, opened 2023-10-29
Revert "Use fuzztest targets for oss-fuzz. (#1633)"

vrabaud:
This reverts commit a12e0dc6c8a72959603797875163bcb509f98d23.

oss-fuzz does not seem ready to run fuzztest tests.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,12 +1,5 @@
 name: CI Fuzz
-on:
-  pull_request:
-    paths:
-      - 'include/**'
-      - 'src/**'
-      - 'tests/oss-fuzz/**'
-    types:
-      - closed
+on: [pull_request]
 
 permissions:
   contents: read
@@ -18,7 +11,6 @@ concurrency:
 
 jobs:
   fuzz:
-    if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     steps:
     - name: Build Fuzzers
--- a/tests/oss-fuzz/build.sh
+++ b/tests/oss-fuzz/build.sh
@@ -19,71 +19,21 @@
 # https://github.com/google/oss-fuzz/blob/master/projects/libavif/build.sh
 # It builds the different fuzz targets.
 
-# To test changes to this file:
-# - make changes and commit to your REPO
-# - run:
-#     git clone --depth=1 git@github.com:google/oss-fuzz.git
-#     cd oss-fuzz
-# - modify projects/libavif/Dockerfile to point to your REPO
-# - run:
-#     python3 infra/helper.py build_image libavif
-#     # enter 'y' and wait for everything to be downloaded
-# - run:
-#     python3 infra/helper.py build_fuzzers --sanitizer address libavif
-#     # wait for the tests to be built
-# And then run the fuzzer locally, for example:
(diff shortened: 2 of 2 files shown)


tests/oss-fuzz/build.sh
```diff
@@ -19,71 +19,21 @@
...
-#     python3 infra/helper.py run_fuzzer libavif \
-#     avif_fuzztest_enc_dec_incr@EncodeDecodeAvifFuzzTest.EncodeDecodeGridValid \
-#     --sanitizer address
-
-ln -s $SRC/fuzztest $SRC/libavif/ext/fuzztest
-
-# build dependencies
-cd ext && bash aom.cmd && bash dav1d.cmd && bash googletest.cmd && bash libjpeg.cmd && \
-      bash libsharpyuv.cmd && bash libyuv.cmd && bash zlibpng.cmd && cd ..
+# build dav1d
```
wantehchang, line 22: This builds not only dav1d but also libyuv. It may be simpler to use the original comments "build dependencies".

tests/oss-fuzz/build.sh
```diff
@@ -19,71 +19,21 @@
...
-    # Write executer script
-    echo "#!/bin/sh
-# LLVMFuzzerTestOneInput for fuzzer detection.
-this_dir=\$(dirname \"\$0\")
-chmod +x \$this_dir/$fuzz_basename
-$fuzz_basename --fuzz=$fuzz_entrypoint -- \$@" > $OUT/$TARGET_FUZZER
-    chmod +x $OUT/$TARGET_FUZZER
-  done
-done
+    $LIB_FUZZING_ENGINE libavif.a ../ext/dav1d/build/src/libdav1d.a
```
wantehchang, line 36: IMPORTANT: Add `../ext/libyuv/build/libyuv.a`.

Otherwise this will break when libavif.a is no longer a combined archive library.

.github/workflows/cifuzz.yml
```diff
@@ -1,12 +1,5 @@
 name: CI Fuzz
-on:
-  pull_request:
-    paths:
-      - 'include/**'
-      - 'src/**'
-      - 'tests/oss-fuzz/**'
-    types:
-      - closed
+on: [pull_request]
```
wantehchang, line 2: The syntax looks quite different from the original code. I am not familiar with GitHub workflow's YML files. Might want to have someone else check this change.

wantehchang (approved):
LGTM. Thanks!

.github/workflows/cifuzz.yml
wantehchang: Also the changes in this file seem unrelated to the fuzztest targets. If so, it would be good to describe these changes in the commit message (or move them to a separate pull request).

wantehchang:
I will go ahead and merge this pull request, and write a new pull request to make the changes I suggested.

.github/workflows/cifuzz.yml
wantehchang: I just realized this pull request was a pure revert of commit a12e0dc6c8a72959603797875163bcb509f98d23.

tests/oss-fuzz/build.sh
wantehchang: Done in https://github.com/AOMediaCodec/libavif/pull/1721.

PR #1714, opened 2023-10-30
avifenc: set a default gain map quality

maryla-uc:
Use the same default quality as the main image.
Print the quality setting when there is a gain map.


y-guyon (approved):
We could have a one-liner `sh` test for that in another PR.

maryla-uc:
What do you think we should check in that test? That various quality values change the size in an expectd way? Grep the printed "Encoding with AV1 codec..." string?

y-guyon:
> What do you think we should check in that test?

Just calling `avifenc` successfully.

Actually this is already done in the golden tests:

https://github.com/AOMediaCodec/libavif/blob/7aa84aec97688effa4dca0b91926dafcbac0b0d3/tests/test_cmd_enc_gainmap_boxes_golden.sh#L24

I see there is no test exercising `--qgain-map` though.

maryla-uc:
Ack. I will add more tests in another PR.

PR #1722, opened 2023-11-02
Sync third_party/libyuv to libyuv version 1879

y-guyon:
This is a draft because this change may introduce the following effect:  
Upon memory allocation failure in `third_party/libyuv`, libavif may output a scaled image buffer with uninitialized pixel values without reporting an error.  
See https://bugs.chromium.org/p/libyuv/issues/detail?id=968#c6.  
The same issue will happen when bumping `ext/libyuv`.


y-guyon:
Superseded by https://github.com/AOMediaCodec/libavif/pull/1739.

PR #1724, opened 2023-11-02
Add an avifComputeGainMap function.

maryla-uc:
Only in internal.h for now, like the tone mapping function.

### DIFF
--- a/src/gainmap.c
+++ b/src/gainmap.c
@@ -302,7 +302,7 @@ cleanup:
 // ---------------------------------------------------------------------------
 // Create a gain map.
 
-// A histogram of gain map values is used to remove outliers, whic helps with
+// A histogram of gain map values is used to remove outliers, which helps with
 // gain map accuracy/compression.
 #define NUM_HISTOGRAM_BUCKETS 1000 // Empirical value
 // Arbitrary max value, roughly equivalent to bringing SDR white to max PQ white (log2(10000/203) ~= 5.62)
@@ -311,11 +311,13 @@ cleanup:
 #define BUCKET_RANGE (BUCKET_MAX_VALUE - BUCKET_MIN_VALUE)
 #define MAX_OUTLIERS_RATIO 0.001f // 0.1%
 
+// Returns the index of the histogram bucket for a given value.
 static int avifValueToBucketIdx(float v)
 {
     v = AVIF_CLAMP(v, BUCKET_MIN_VALUE, BUCKET_MAX_VALUE);
     return AVIF_MIN((int)avifRoundf((v - BUCKET_MIN_VALUE) / BUCKET_RANGE * NUM_HISTOGRAM_BUCKETS), NUM_HISTOGRAM_BUCKETS - 1);
 }
+// Returns the lower end of the value range belonging to the given histogram bucket.
 static float avifBucketIdxToValue(int idx)
 {
     return idx * BUCKET_RANGE / NUM_HISTOGRAM_BUCKETS + BUCKET_MIN_VALUE;
@@ -341,7 +343,7 @@ static void avifFindMinMaxWithoutOutliers(int histogram[NUM_HISTOGRAM_BUCKETS], 
             break;
         }
         if (histogram[i] == 0) {
-            *rangeMin = avifBucketIdxToValue(i);
+            *rangeMin = avifBucketIdxToValue(i + 1); // +1 to get the higher end of the bucket.
         }
     }
 
@@ -398,14 +400,12 @@ avifResult avifComputeGainMapRGB(const avifRGBImage * baseRgbImage,
     // --- After this point, the function should exit with 'goto cleanup' to free allocated resources.
 
     const avifBool singleChannel = (gainMap->image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400);
-    for (int c = 0; c < 3; ++c) {
+    const int numGainMapChannels = singleChannel ? 1 : 3;
+    for (int c = 0; c < numGainMapChannels; ++c) {
         gainMapF[c] = avifAlloc(width * height * sizeof(float));
         if (gainMapF[c] == NULL) {
             res = AVIF_RESULT_OUT_OF_MEMORY;
             goto cleanup;
-        }
-        if (singleChannel) {
-            break;
         }
     }
 
@@ -436,7 +436,7 @@ avifResult avifComputeGainMapRGB(const avifRGBImage * baseRgbImage,
                 altRGBA[c] = altGammaToLinear(altRGBA[c]);
             }
 
-            for (int c = 0; c < 3; ++c) {
+            for (int c = 0; c < numGainMapChannels; ++c) {
                 float base = baseRGBA[c];
                 float alt = altRGBA[c];
                 if (singleChannel) {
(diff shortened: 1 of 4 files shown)


src/gainmap.c
```diff
@@ -286,4 +299,327 @@ avifResult avifImageApplyGainMap(const avifImage * baseImage,
...
+            }
+        }
+    }
+
+    // Find approximate min/max for each channel, discarding outliers.
+    float gainMapMinLog2[3] = { 0.0f, 0.0f, 0.0f };
+    float gainMapMaxLog2[3] = { 0.0f, 0.0f, 0.0f };
+    for (int c = 0; c < 3; ++c) {
+        avifFindMinMaxWithoutOutliers(histograms[c], &gainMapMinLog2[c], &gainMapMaxLog2[c]);
+    }
```
y-guyon, line 470: Restrict if `singleChannel`?

src/gainmap.c
```diff
@@ -286,4 +299,327 @@ avifResult avifImageApplyGainMap(const avifImage * baseImage,
...
+                    alt = y_coeffs[0] * altRGBA[0] + y_coeffs[1] * altRGBA[1] + y_coeffs[2] * altRGBA[2];
+                }
+                if (base > baseMax) {
+                    baseMax = base;
+                }
+                if (alt > altMax) {
+                    altMax = alt;
+                }
+                const float ratioLog2 =
+                    log2f((alt + (float)gainMapMetadata.alternateOffset[c]) / (base + (float)gainMapMetadata.baseOffset[c]));
```
y-guyon, line 454: I probably missed a conversation but why not use double everywhere?

src/gainmap.c
```diff
@@ -286,4 +299,327 @@ avifResult avifImageApplyGainMap(const avifImage * baseImage,
...
+        }
+    }
+
+    res = avifImageRGBToYUV(gainMapImage, &gainMapRGB);
+    if (res != AVIF_RESULT_OK) {
+        goto cleanup;
+    }
+
+    // Scale down the gain map if requested.
+    // Another way would be to scale down the source images, but it seems to perform worse.
```
y-guyon, line 545: This is "probably down" rather than "always down" here, right?

tests/gtest/avifgainmaptest.cc
```diff
@@ -933,6 +933,51 @@ class ToneMapTest
...
+      base_image, &gain_map, hdr_headroom, tone_mapped->transferCharacteristics,
+      &tone_mapped_rgb, &tone_mapped->clli, &diag);
+  ASSERT_EQ(result, AVIF_RESULT_OK)
+      << avifResultToString(result) << " " << diag.error;
+  ASSERT_EQ(avifImageRGBToYUV(tone_mapped.get(), &tone_mapped_rgb),
+            AVIF_RESULT_OK);
+  if (reference_image != nullptr) {
+    EXPECT_EQ(out_depth, (int)reference_image->depth);
+    const double psnr = testutil::GetPsnr(*reference_image, *tone_mapped);
+    printf("PSNR: %f\n", psnr);
```
y-guyon, line 964: Leftover?

y-guyon (approved):

src/gainmap.c
```diff
@@ -286,4 +299,327 @@ avifResult avifImageApplyGainMap(const avifImage * baseImage,
...
+#define MAX_OUTLIERS_RATIO 0.001f // 0.1%
+
+static int avifValueToBucketIdx(float v)
+{
+    v = AVIF_CLAMP(v, BUCKET_MIN_VALUE, BUCKET_MAX_VALUE);
+    return AVIF_MIN((int)avifRoundf((v - BUCKET_MIN_VALUE) / BUCKET_RANGE * NUM_HISTOGRAM_BUCKETS), NUM_HISTOGRAM_BUCKETS - 1);
+}
+static float avifBucketIdxToValue(int idx)
+{
+    return idx * BUCKET_RANGE / NUM_HISTOGRAM_BUCKETS + BUCKET_MIN_VALUE;
```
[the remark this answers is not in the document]
maryla-uc: Clarified that it returns the lower end. I also added a +1 in avifFindMinMaxWithoutOutliers below to make the lower bound more accurate.

src/gainmap.c
```diff
@@ -286,4 +299,327 @@ avifResult avifImageApplyGainMap(const avifImage * baseImage,
...
+        return AVIF_RESULT_NOT_IMPLEMENTED;
+    }
+
+    float * gainMapF[3] = { 0 }; // Temporary buffers for the gain map as floating point values, one per RGB channel.
+    avifRGBImage gainMapRGB;
+    memset(&gainMapRGB, 0, sizeof(gainMapRGB));
+    avifImage * gainMapImage = gainMap->image;
+
+    avifResult res = AVIF_RESULT_OK;
+    // --- After this point, the function should exit with 'goto cleanup' to free allocated resources.
```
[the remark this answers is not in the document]
maryla-uc: I know it's widespread but I thought it was useful to document when we start allocating resources and must use goto? Before this point, the input checks can do simple returns.
Not sure using a helper function is that useful? All avifComputeGainMap() below does it allocate some resources then call a helper function, and I still thought the goto was useful (as much as I hate gotos, but that's C for you).

src/gainmap.c
maryla-uc: Indeed.

tests/gtest/avifgainmaptest.cc
maryla-uc: It was intentional, yes it's debugging but I think it's useful to have it there and it doesn't hurt...

src/gainmap.c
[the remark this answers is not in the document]
y-guyon: > I know it's widespread but I thought it was useful to document when we start allocating resources and must use goto? Before this point, the input checks can do simple returns.

Fine with me.

> Not sure using a helper function is that useful?

I was not talking about a helper function, but to split the function into:

```c
avifResult avifFunctionImpl(avifImage ** image) {
    *image = avifImageCreateEmpty();
    AVIF_CHECKERR(*image != NULL, AVIF_RESULT_OOM);
    AVIF_CHECKRES(avifImageAllocate(*image));
    ...
    return AVIF_RESULT_OK;
}
avifResult avifFunction() {
    avifImage * image = NULL;
    const avifResult result = avifFunctionImpl(&image);
    if (image != NULL) {
        avifImageDestroy(image);
    }
    return result;
}
```
but this is a personal preference, `goto`s work too.

tests/gtest/avifgainmaptest.cc
y-guyon: What about `SCOPED_TRACE` then? Does it print only on failure? Also consider `std::cout` instead of `printf`.

src/gainmap.c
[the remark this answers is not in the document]
maryla-uc: I see. I don't think this solution is great either because now we have to free resources allocated by a different function. Splitting alloc and free into different functions is not intuitive and just as error prone in my opinion. But again, memory management in C is inherently error prone...

tests/gtest/avifgainmaptest.cc
maryla-uc: Yes SCOPED_TRACE only prints on failure. Changed to std::cout.

PR #1729, opened 2023-11-06
Expand CLLI documentation.

### LINKED ISSUE #1727: clli SEI documentation is missing a sentence
https://github.com/AOMediaCodec/libavif/blob/f647313b92f5399203379194d73525aa56247aad/include/avif/avif.h#L518-L520

Here, the calculation of E_Max is described, but not how the light level is calculated from it. I don't have access to the cited ISO/IEC standard, but this is the next sentence in H.274:

```
The light level corresponding to the stimulus is then defined as the CIE 1931 luminance
corresponding to equal amplitudes of E_Max for all three colour primary intensities for red, green, and blue (with appropriate
scaling to reflect the nominal luminance level associated with peak white – e.g., ordinarily scaling to associate peak white
with 10 000 candelas per square metre when vui_transfer_characteristics is equal to 16).
```

maryla-uc:
Fixes #1727

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -513,11 +513,17 @@ AVIF_NODISCARD AVIF_API avifBool avifCleanApertureBoxConvertCropRect(avifCleanAp
 typedef struct avifContentLightLevelInformationBox
 {
     // 'clli' from ISO/IEC 23000-22:2019 (MIAF) 7.4.4.2.2. The SEI message semantics written above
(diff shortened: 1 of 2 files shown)


maryla-uc:
> Should there be an additional comment mentioning that `Y luminance is assumed to be equal to E_Max`

I think the safest strategy is to quote the spec verbatim, lest we say something wrong. Very few people need to understand this in detail anyway, and if they do, they should read the specs.

Issue #1522, opened 2023-08-25
avifenc: Change the default value of --jobs to "all"


wantehchang:
The default value of the --jobs option for avifenc is 1 (single threaded). It would be good to change the default to "all" (use all available cores).

vrabaud:
That can be dangerous as it could hog ressources no? And maybe -j 100 would be counter-productive on CPUs with many cores. How about maxing to 4 or 8, and to all-1?

FrankGalligan:
My guess is the default --jobs all is fine. 

First. I'm not sure setting jobs to "all" will hog the resources so everything else is unusable. But we should be able to test this.

Second, Lets look at some use cases where avifenc may be used with --jobs set to all:
1. User encoding one file at a time. 12MP will be done in less than a second with all the defaults. So the user would not have time to switch to something else.
2. User profiling encode time. This is their main purpose, so the fastest encode is the best.
3. User encoding many photos. This is their only current task then "all" is fine.
4. User encoding many photos. They want to perform multiple tasks, then "all" may hog resources from their other task. This user may also be more likely to set their own --jobs flag.
5. Encoding service with many users. This is an advanced usage and I would assume they would set the advanced settings tailored to their use case.

So IMO the only use case that setting default to "all" may be detrimental is 4. But my guess percentage wise both 3 & 4 will be much lower compared to other use cases. Also I think 4 would be more likely to set the jobs parameter based on what they are doing on their machine.

That being said, if we set the default to "all", we can watch the issue tracker to see if we need to change this decision in the future as well.

y-guyon:
There could also be a warning displayed when running `avifenc` to remind users to explicitly set `--jobs`. Not saying we should do that but it may help.

PR #1730, opened 2023-11-07
avifenc: Change default value of --jobs to "all"

wantehchang:
Fix https://github.com/AOMediaCodec/libavif/issues/1522.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -209,7 +209,7 @@ static void syntaxLong(void)
            AVIF_SPEED_FASTEST);
     printf("\n");
     printf("Advanced options:\n");
-    printf("    -j,--jobs J                       : Number of jobs (worker threads). Use \"all\" to use all available cores (default: all)\n");
+    printf("    -j,--jobs J                       : Number of jobs (worker threads). Use \"all\" to potentially use as many cores as possible (default: all)\n");
     printf("    --no-overwrite                    : Never overwrite existing output file\n");
     printf("    -o,--output FILENAME              : Instead of using the last filename given as output, use this filename\n");
 #if defined(AVIF_ENABLE_EXPERIMENTAL_AVIR)
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -209,7 +209,7 @@ static void syntaxLong(void)
            AVIF_SPEED_FASTEST);
     printf("\n");
     printf("Advanced options:\n");
-    printf("    -j,--jobs J                       : Number of jobs (worker threads, default: 1. Use \"all\" to use all available cores)\n");
+    printf("    -j,--jobs J                       : Number of jobs (worker threads). Use \"all\" to use all available cores (default: all)\n");
     printf("    --no-overwrite                    : Never overwrite existing output file\n");
     printf("    -o,--output FILENAME              : Instead of using the last filename given as output, use this filename\n");
 #if defined(AVIF_ENABLE_EXPERIMENTAL_AVIR)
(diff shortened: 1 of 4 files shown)


src/codec_aom.c
```diff
@@ -748,7 +748,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             cfg->g_lag_in_frames = 0;
         }
         if (encoder->maxThreads > 1) {
-            cfg->g_threads = encoder->maxThreads;
+            // libaom fails if cfg->g_threads is greater than 64 threads. See MAX_NUM_THREADS in
+            // aom/aom_util/aom_thread.h.
```
vrabaud, line 752: This header is public right(encoder/encoder.h references it)? So maybe it should be included and MAX_NUM_THREADS be used (in case it changes one day).

apps/avifenc.c
```diff
@@ -209,7 +209,7 @@ static void syntaxLong(void)
            AVIF_SPEED_FASTEST);
     printf("\n");
     printf("Advanced options:\n");
-    printf("    -j,--jobs J                       : Number of jobs (worker threads, default: 1. Use \"all\" to use all available cores)\n");
+    printf("    -j,--jobs J                       : Number of jobs (worker threads). Use \"all\" to use all available cores (default: all)\n");
```
vrabaud, line 212: Maybe: "to potentially use"
Just to make sure the user understands the limitations

vrabaud: Actually, maybe: "to potentially use as many cores as possible"
In case machines have more than 64 cores ...

vrabaud (approved):

apps/avifenc.c
wantehchang: Done. Thanks for the suggestion.

vrabaud (approved):

src/codec_aom.c
```diff
@@ -748,7 +748,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+            cfg->g_threads = AVIF_MIN(encoder->maxThreads, 64);
```
fdintino, line 753: @wantehchang do you think this fix might make its way to the v1.0.x branch?

wantehchang: Frankie: The v1.0.x branch is used for bug-fix patch releases. Since this pull request is not a bug fix, it is more suitable for the next feature release (v1.1.0). However, we can consider making an exception based on customer requests.

Note: This missed the v1.0.4 release that was announced today.

fdintino: That makes sense to me, insofar as this commit changes an argument to avifenc. But I'd argue that the line I commented on here _is_ in fact a bug fix. If you were to run `avifenc --jobs 96 ...` with the aom encoder using v1.0.4 it fails with the error

> aom_codec_enc_init() failed: Invalid parameter: g_threads out of range [..MAX_NUM_THREADS]

but it succeeds on the main branch. This came to my attention because pillow-avif-plugin sets maxThreads, by default, to the number of cores, and users with 64+ cores were unable to encode using AOM (see https://github.com/fdintino/pillow-avif-plugin/pull/41 )

wantehchang: Ah, I am sorry I misunderstood you. I thought by "this fix" you meant "this pull request".

Yes, we can cherry-pick just this fix to the v1.0.x branch in preparation for the next patch release.

vrabaud: Done in https://github.com/AOMediaCodec/libavif/pull/2008

wantehchang: Thank you very much, Vincent!

PR #1731, opened 2023-11-07
Add the 'popl' flag parsing library to third_party/

maryla-uc:
This library will be used in c++ command line tools like the upcoming `avifgainmaputil`.
There are tons of command line parsers out there, I have not tried them all but chose this one because it's small but feature rich enough, and does not increase the compile time too much (for example CLI11 has more features but I found it to be much slower).

### DIFF
--- a/third_party/README.md
+++ b/third_party/README.md
@@ -26,7 +26,8 @@ When importing source code from upstream libyuv, the following changes must be d
 
 ## popl
 
-This subdirectory contains source code imported from popl https://github.com/badaix/popl/ v1.3.0 (bda5f43)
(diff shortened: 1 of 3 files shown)


third_party/README.md
```diff
@@ -23,3 +23,13 @@ When importing source code from upstream libyuv, the following changes must be d
 4. The commit hash of libyuv from where the files got imported in this README.md file must be updated.
 5. `LIBYUV_API` must be removed from any and all imported functions as these files are always built static.
 6. Replace .cc file extension by .c.
+
+## popl
+
+This subdirectory contains source code imported from popl https://github.com/badaix/popl/ v1.3.0 (bda5f43)
```
y-guyon, line 29: Missing dot at the end

y-guyon (approved):

PR #1733, opened 2023-11-07
Add a public C++ header.

vrabaud:
And have tests use it.
Headers are also cleaned (added/removed) in the tests

### DIFF
--- a/include/avif/avif_cxx.h
+++ b/include/avif/avif_cxx.h
@@ -6,7 +6,7 @@
 #ifndef AVIF_AVIF_CXX_H
 #define AVIF_AVIF_CXX_H
 
-#if !(defined(__cplusplus) || defined(c_plusplus))
+#if !defined(__cplusplus)
 #error "This a C++ only header. Use avif/avif.h for C."
 #endif
 
@@ -26,9 +26,9 @@ struct UniquePtrDeleter
 };
 
 // Use these unique_ptr to ensure the structs are automatically destroyed.
-typedef std::unique_ptr<avifEncoder, UniquePtrDeleter> EncoderPtr;
-typedef std::unique_ptr<avifDecoder, UniquePtrDeleter> DecoderPtr;
-typedef std::unique_ptr<avifImage, UniquePtrDeleter> ImagePtr;
+using EncoderPtr = std::unique_ptr<avifEncoder, UniquePtrDeleter>;
+using DecoderPtr = std::unique_ptr<avifDecoder, UniquePtrDeleter>;
+using ImagePtr = std::unique_ptr<avifImage, UniquePtrDeleter>;
 
 } // namespace libavif
 
--- a/include/avif/avif_cxx.h
+++ b/include/avif/avif_cxx.h
@@ -1,0 +1,35 @@
+// Copyright 2023 Google LLC
+// SPDX-License-Identifier: BSD-2-Clause
+
+// This is a header only library.
+
+#ifndef AVIF_AVIF_CXX_H
+#define AVIF_AVIF_CXX_H
+
+#if !(defined(__cplusplus) || defined(c_plusplus))
+#error "This a C++ only header. Use avif/avif.h for C."
+#endif
+
+#include <memory>
+
+#include "avif/avif.h"
+
+namespace libavif
+{
+
+// Struct to call the destroy functions in a unique_ptr.
+struct UniquePtrDeleter
+{
+    void operator()(avifEncoder * encoder) const { avifEncoderDestroy(encoder); }
+    void operator()(avifDecoder * decoder) const { avifDecoderDestroy(decoder); }
+    void operator()(avifImage * image) const { avifImageDestroy(image); }
+};
+
+// Use these unique_ptr to ensure the structs are automatically destroyed.
+typedef std::unique_ptr<avifEncoder, UniquePtrDeleter> EncoderPtr;
+typedef std::unique_ptr<avifDecoder, UniquePtrDeleter> DecoderPtr;
+typedef std::unique_ptr<avifImage, UniquePtrDeleter> ImagePtr;
+
+} // namespace libavif
+
+#endif // AVIF_AVIF_CXX_H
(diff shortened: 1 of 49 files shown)


maryla-uc (approved):

include/avif/avif_cxx.h
```diff
@@ -0,0 +1,35 @@
+// Copyright 2023 Google LLC
+// SPDX-License-Identifier: BSD-2-Clause
+
+// This is a header only library.
+
+#ifndef AVIF_AVIF_CXX_H
+#define AVIF_AVIF_CXX_H
+
+#if !(defined(__cplusplus) || defined(c_plusplus))
```
wantehchang, line 9: We should not test the `c_plusplus` macro. I guess that's a nonstandard macro defined by an old compiler.

Another way to see this is that this should be the opposite of the test in the C header:

```
#ifdef __cplusplus
extern "C" {
#endif
```

Note: I think this check is not necessary because compilation will fail if this header is included by C files. I guess you want a more informative error message?

include/avif/avif_cxx.h
```diff
@@ -0,0 +1,35 @@
...
+{
+    void operator()(avifEncoder * encoder) const { avifEncoderDestroy(encoder); }
+    void operator()(avifDecoder * decoder) const { avifDecoderDestroy(decoder); }
+    void operator()(avifImage * image) const { avifImageDestroy(image); }
+};
+
+// Use these unique_ptr to ensure the structs are automatically destroyed.
+typedef std::unique_ptr<avifEncoder, UniquePtrDeleter> EncoderPtr;
+typedef std::unique_ptr<avifDecoder, UniquePtrDeleter> DecoderPtr;
+typedef std::unique_ptr<avifImage, UniquePtrDeleter> ImagePtr;
```
wantehchang, line 31: Use `using` instead of `typedef`.

wantehchang (commented):
Vincent: I only reviewed the new header. Please consider my review as supplemental.

include/avif/avif_cxx.h
wantehchang, line 31: Optional: I am tempted to suggest we still add an "avif" or "Avif" prefix to these type names.  The reason is that we don't always use the fully qualified names `libavif::EncoderPtr`, `libavif::DecoderPtr`, etc. For example, we may write code inside a `namespace libavif` block:

```
namespace libavif {
  
  ...

  EncoderPtr encoder(avifEncoderCreate());
  ImagePtr image(avifImageCreateEmpty());
 
   ...

}  // namespace libavif
```

vigneshvg: Also, do we need the Ptr suffix for these variables?

wantehchang: I like the `Ptr` suffix because it suggests these types are std::unique_ptr's. We can also use the `UniquePtr` suffix, but it's a little verbose.

include/avif/avif_cxx.h
vrabaud: I removed c_plusplus
Exactly: it might as well fail with a clear message.

include/avif/avif_cxx.h
vrabaud: So Ptr is because we "might" (with big quotes because we have no idea whether it would be better, though we have done it for RGBImage and RWData in aviftest_helpers.h) create a C++ class for an encoder and we would name it `Encoder`. 
For the prefix, I think it is too verbose: either you develop within the libavif namespace (which is only done in tests) and therefore the libavif repo,  and you therefore know what you are doing and ImagePtr is obviously for an avifImage. Either you develop somewhere else, and `libavif::ImagePtr` is simply shorter and less redundant than `libavif::avifImagePtr` IMHO. WDYT?

wantehchang: Yes, let's optimize for the fully-qualified `libavif::ImagePtr` case.

Note: I wonder if we can use the `avif` namespace. `avif::ImagePtr` looks really nice. libheif's heif_cxx.h header uses the `heif` namespace. Just a thought.

vrabaud: I love the `avif` namespace: it is short and I doubt anybody uses it. Let me switch the tests to using that namespace: I am modifying all the files anyway.

vrabaud: Done

include/avif/avif_cxx.h
```diff
@@ -0,0 +1,35 @@
+// Copyright 2023 Google LLC
+// SPDX-License-Identifier: BSD-2-Clause
+
+// This is a header only library.
```
wantehchang, line 4: This comment is misleading, because this header requires the libavif library. We should delete this comment or rephrase it.

include/avif/avif_cxx.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef AVIF_AVIF_CXX_H
+#define AVIF_AVIF_CXX_H
+
+#if !defined(__cplusplus)
```
wantehchang, line 9: Optional: We can use the `#ifndef` form now.

wantehchang (approved):
LGTM. Thanks. I only reviewed CMakeLists.txt and avif_cxx.h.

include/avif/avif_cxx.h
vrabaud: You are right, let's delete it: the file name is explicit enough.

include/avif/avif_cxx.h
wantehchang: Good. As the r[eference implementation of AVIF](https://github.com/AOMediaCodec/av1-avif/wiki#reference-software), libavif should be qualified to use the `avif` namespace.

wantehchang (approved):

include/avif/avif_cxx.h
wantehchang: I guess the comment meant that there is a separate libavif_cxx library that consists of only avif_cxx.h. The reason it confused me is that I view avif_cxx.h as an additional header of the libavif library.

PR #1736, opened 2023-11-08
Move third_party/popl/LICENSE to LICENSE section

y-guyon:
See same reasoning as for `third_party/libyuv` at https://github.com/AOMediaCodec/libavif/pull/1534#issuecomment-1712458376.


maryla-uc (approved):
thanks

PR #1734, opened 2023-11-08
Add a basic version of avifgainmaputil.

maryla-uc:
It has three commands:
- help
- extractgainmap to save the gain map of a file as an image
- printmetadata to print detailed gain map metadata

Since it's written in c++ I chose to follow the same style guide as the c++ tests.
I placed it in a cpp/ subdirectory to be able to add a .clang-format file.

PR #1737, opened 2023-11-08
Only call enable_language and set CMAKE_CXX_STANDARD once.

vrabaud:
This is to make sure CMAKE_CXX_STANDARD is the smallest value that fits all needs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,7 @@ option(
 
 set(AVIF_USE_CXX OFF)
 
-if(AVIF_LOCAL_LIBGAV1)
+if(AVIF_BUILD_APPS OR AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST OR AVIF_LOCAL_LIBGAV1)
     set(AVIF_USE_CXX ON)
 endif()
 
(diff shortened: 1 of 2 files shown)


vrabaud:
That will help https://github.com/AOMediaCodec/libavif/pull/1734

CMakeLists.txt
```diff
@@ -83,7 +83,7 @@ option(
 
 set(AVIF_USE_CXX OFF)
 
-if(AVIF_LOCAL_LIBGAV1)
+if(AVIF_BUILD_APPS OR AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST OR AVIF_LOCAL_LIBGAV1)
```
maryla-uc, line 86: I know this was already the case but do you know why we enable c++ when AVIF_BUILD_APPS is on? current apps are C only afaik

maryla-uc (approved):

CMakeLists.txt
vrabaud: Nope, I was wondering too. I kept it in case there is some experimental stuff I did not know about ...
Let me remove it.

maryla-uc: Mh that's kind of weird, either are_images_equal  should not be built when only AVIF_BIULD_APPS is on, or it should be moved to the apps/ directory I think...  But that's a problem for another PR

PR #1738, opened 2023-11-08
Revert the addition of popl

maryla-uc:
Will use libargparse instead https://github.com/kmurray/libargparse which is already used internally at Google.


third_party/popl/LICENSE
y-guyon, line 1: Please remove this file.

y-guyon (approved):

y-guyon (approved):
Thanks!

PR #1739, opened 2023-11-08
Sync third_party/libyuv and propagate errors

y-guyon:
See https://github.com/AOMediaCodec/libavif/issues/820#issuecomment-1773013251.


src/scale.c
```diff
@@ -101,7 +105,13 @@ avifResult avifImageScaleWithLimit(avifImage * image,
                 const uint32_t srcStride = srcYUVRowBytes[i] / 2;
                 uint16_t * const dstPlane = (uint16_t *)image->yuvPlanes[i];
                 const uint32_t dstStride = image->yuvRowBytes[i] / 2;
-#if LIBYUV_VERSION >= 1774
+#if LIBYUV_VERSION >= 1880
+                if (ScalePlane_12(srcPlane, srcStride, srcW, srcH, dstPlane, dstStride, dstW, dstH, AVIF_LIBYUV_FILTER_MODE)) {
+                    avifDiagnosticsPrintf(diag, "Allocation of YUV planes failed in libyuv");
```
wantehchang, line 110: The "Allocation of YUV planes" part of the diagnostic message seems incorrect, because ScalePlane_12 does not allocate the YUV planes.

Also, we may want to log the nonzero return value of ScalePlane_12 in the diagnostic message. libyuv functions return -1 for invalid parameters and 1 for out-of-memory errors. (The ScalePlane* functions currently don't validate its input parameters, so the only nonzero return value is 1. But this may change in the future.)

wantehchang (commented):
Thank you for updating third_party/libyuv. I have a comment on the diagnostic messages for the failures of the ScalePlane* functions.

wantehchang (approved):
LGTM. Thanks!

PR #1741, opened 2023-11-09
Format iccjpeg.h/c

maryla-uc:
I'm not sure if the clang-format command line is supposed to be windows compatible, and if it is, if windows supports the double wildcards (if not I can remove them).
If we don't need that much compatibility, then a version with 'find' would be more future proof I think:
`find apps/ examples/ include/ src/ tests/ -regex '.*\.\(c\|cc\|h\)' | xargs clang-format -style=file -i`
Also, should contrib/ be added as well?

### DIFF
--- a/apps/shared/iccjpeg.c
+++ b/apps/shared/iccjpeg.c
@@ -18,8 +18,7 @@
  */
 
 #include "iccjpeg.h"
-#include <stdlib.h>			/* define malloc() */
-
+#include <stdlib.h> /* define malloc() */
 
 /*
  * Since an ICC profile can be larger than the maximum size of a JPEG marker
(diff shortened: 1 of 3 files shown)


apps/shared/iccjpeg.c
y-guyon, line 1: If I remember correctly, iccjpeg.h and iccjpeg.c were imported verbatim (like in [chromium](https://chromium.googlesource.com/chromium/src/third_party/+/3187afcd157cda4bb9111646324bc3473121fa85/iccjpeg/)). They also have their own LICENSE section:

https://github.com/AOMediaCodec/libavif/blob/eafb4d76adc86266b6bb917cd4f6f1b609091e6a/LICENSE#L54-L119

I guess that is why it was not modified until now.  
The license mentions "Permission is hereby granted to use, copy, modify" so I guess it is fine to reformat it. It will just be a little harder to check whether the file changed upstream (if there is still an upstream somewhere anyway).

Are you planning on modifying that file?

y-guyon (approved):
> If we don't need that much compatibility, then a version with 'find' would be more future proof I think

Is `find` supported on windows?

Why not `clang-format -style=file -i **/*.h **/*.c **/*.cc` rather than `find`?

> Also, should contrib/ be added as well?

It has a dedicated LICENSE section:

https://github.com/AOMediaCodec/libavif/blob/eafb4d76adc86266b6bb917cd4f6f1b609091e6a/LICENSE#L123-L146

which mentions modifying the code, so I guess it is fine.

apps/shared/iccjpeg.c
maryla-uc: Ha ok, I didn't notice that, I confused it with iccmaker.
In that case we should probably move it to third_party/?
I'm not planning on modifying it. I was just playing with the clang-format command since I'm adding other files.

With regard to your other questions:

> Is `find` supported on windows?

I don't know (probably not?), I was suggesting this in case we didn't care about windows compatibility

> Why not `clang-format -style=file -i **/*.h **/*.c **/*.cc` rather than `find`?

This might match too much, like third_party/?
Also it seems that for the double asterisk to work proplery, you need the "globstar" option enabled, which is off by default (it's indeed off on my machine), otherwise it works like a single star. So it's better to stick to single * for compatibility it seems;

wantehchang (commented):
Yannis is right. These two files are third-party code. We should preserve their original formatting to make it easier to compare our copy with the upstream.

We can move these two files to the new third_party/ directory.

PR #1743, opened 2023-11-09
Don't require crop width,height be even for 4:2:0

wantehchang:
These requirements were removed in ISO/IEC 23000-22:2019/Amd. 2:2021 from Section 7.3.6.7. This is one reason most of the clap properties seen by Chrome were considered invalid.

Bug: b/308458941

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * API calls now return AVIF_RESULT_OUT_OF_MEMORY instead of aborting on memory
   allocation failure.
 * avifenc: Change the default value of the --jobs option from 1 to "all".
+* Update avifCropRectConvertCleanApertureBox() to the revised requirements in
+  ISO/IEC 23000-22:2019/Amd. 2:2021 Section 7.3.6.7.
 
 ## [1.0.1] - 2023-08-29
 
--- a/src/avif.c
+++ b/src/avif.c
@@ -653,16 +653,15 @@ static avifBool overflowsInt32(int64_t x)
 static avifBool avifCropRectIsValid(const avifCropRect * cropRect, uint32_t imageW, uint32_t imageH, avifPixelFormat yuvFormat, avifDiagnostics * diag)
 
 {
-    // ISO/IEC 23000-22:2019/DAM 2:2021, Section 7.3.6.7:
+    // ISO/IEC 23000-22:2019/Amd. 2:2021, Section 7.3.6.7:
     //   The clean aperture property is restricted according to the chroma
     //   sampling format of the input image (4:4:4, 4:2:2:, 4:2:0, or 4:0:0) as
     //   follows:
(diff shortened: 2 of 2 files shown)


vigneshvg (approved):

tests/gtest/avifclaptest.cc
```diff
@@ -65,6 +65,15 @@ constexpr InvalidClapPropertyParam kInvalidClapPropertyTestParams[] = {
...
      AVIF_PIXEL_FORMAT_YUV420,
      {330, 1, 385, 1, static_cast<uint32_t>(-308), 1, 103, 1}},
+    // pcX = -1/2 + (99 - 1)/2 = 48.5
+    // pcY = -1/2 + (99 - 1)/2 = 48.5
+    // leftmost = 48.5 - (99 - 1)/2 = -0.5 (not an integer)
+    // topmost = 48.5 - (99 - 1)/2 = -0.5 (not an integer)
+    {99,
+     99,
+     AVIF_PIXEL_FORMAT_YUV420,
+     {99, 1, 99, 1, static_cast<uint32_t>(-1), 2, static_cast<uint32_t>(-1),
```
y-guyon, line 75: This is fine for consistency but please consider `uint32_t{-1}`.

Also the topic was already discussed but this unsigned-to-signed API should be reworked in my opinion, especially since the specification was updated. Is switching from `unsigned` to `signed` for some fields in `avifCleanApertureBox` a breaking API/ABI change? We should at least add a comment nearby this struct in avif.h to explain.

y-guyon (approved):

tests/gtest/avifclaptest.cc
wantehchang: This is the compiler warning for `uint32_t{-1}`:

```
libavif/tests/gtest/avifclaptest.cc:75:32: error: narrowing conversion of ‘-1’ from ‘int’ to ‘uint32_t’ {aka ‘unsigned int’} [-Wnarrowing]
   75 |      {99, 1, 99, 1, uint32_t{-1}, 2, static_cast<uint32_t>(-1),
      |                                ^
```

wantehchang: > Also the topic was already discussed but this unsigned-to-signed API should be reworked in my opinion, especially since the specification was updated.

When you said "the specification was updated", were you referring to ISO BMFF or MIAF?

The eight fields in the clean aperture property still have the `unsigned int(32)` type in the latest versions of ISO BMFF and MIAF.

> We should at least add a comment nearby this struct in avif.h to explain.

Please see https://github.com/AOMediaCodec/libavif/pull/1749.

y-guyon: > The eight fields in the clean aperture property still have the `unsigned int(32)` type in the latest versions of ISO BMFF and MIAF.

Indeed, I was mistaken. The version with `signed horizOffN` and `signed vertOffN` I was looking at is only a draft. Maybe it will get published some time soon.

> Please see #1749.

Thanks.

PR #1744, opened 2023-11-10
feat(cmake): Use FetchContent and ExternalProject for local deps

fdintino:
This replaces the ad-hoc scripts in the ext folder with modules that use FetchContent and ExternalProject to fetch and build the dependencies as part of the overall project.

This is a fairly radical change, so I would not take offense if it is rejected out of hand. That said, I find the developer experience for this approach very appealing. I think #1683 and #1708 were maybe trying to solve the wrong problem: static linking is hard, and trying to solve for it generally (by, for instance, trying to account for all the variations of system static libraries) is maybe not worth the time and complexity required. Better to make the build process easier and more robust for library consumers.

There is a lot of duplicated code in these local build modules that could be refactored into functions and macros, but I held off on doing any of that work—in part because I've already invested a lot of time in an approach that might not get any traction, but also because I think the mechanics are clearer without additional abstraction.

I tested out this approach with my python library [here](https://github.com/fdintino/pillow-avif-plugin/commit/4c246ebf614023555f87eefe1ebcc266b1ea57ce). A lot of workarounds that would have to be implemented separately (supporting arm64 cross-builds, for instance, or incorporating @wantehchang's aom alpine patch) can instead be shared and maintained with libavif itself.

I feel pretty confident in the code here and have done a fair bit of testing, but I'm marking this as draft while I await general feedback. At the very least the build documentation would need to be updated before this was ready to merge.


tests/gtest/avifavmtest.cc
```diff
@@ -53,9 +53,9 @@ TEST_P(AvmTest, EncodeDecode) {
     decoder->codecChoice = av1_codec;
     // AVIF_RESULT_NO_CODEC_AVAILABLE is expected because av1_codec is not
     // enabled or because we are trying to decode an AV2 file with an AV1 codec.
-    ASSERT_EQ(avifDecoderReadMemory(decoder.get(), decoded.get(), encoded.data,
+    ASSERT_NE(avifDecoderReadMemory(decoder.get(), decoded.get(), encoded.data,
```
fdintino, line 56: These tests are broken in the main branch, but because of a typo in the github actions workflow it was never being executed. dav1d in particular was returning `AVIF_RESULT_DECODE_COLOR_FAILED`, not `AVIF_RESULT_NO_CODEC_AVAILABLE`.

.github/workflows/ci-unix-shared-installed.yml
```diff
@@ -110,6 +100,8 @@ jobs:
         -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON
         -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON -DAVIF_LOCAL_GTEST=ON
         -DAVIF_BUILD_GDK_PIXBUF=ON -DCMAKE_INSTALL_PREFIX=./install
+        -DENABLE_AVX2=OFF
```
fdintino, line 103: A small percentage of macOS github runners don't have avx2 support but advertise that they do, causing "Illegal instruction: 4" errors in AOM tests.

vrabaud (commented):
Ok, so here are a few comments:
- first, thank you, this is very welcome (the ext folder is definitely not CMake friendly and has grown too much). Using FetchContent will definitely clean our CMake, make it more future proof, more compatible with static libraries and cross compilation
- there are unrelated fixes in this PR, you should make other PRs to get those fixes in first
- the Local*.cmake files could be made more user friendly: have comments that explain what you should change if you want to tweak those (e.g. have libaom with specific flags), set a version number variable on top so that it is easy to update. I believe some variables are useless define (e.g. *_INCLUDE_DIRS). Maybe we should focus on having proper targets that we can then use in the libavif CMake.
- I believe we should first have  PR with Local*.cmake for dependencies we do not want to tweak: definitely gtest, libxml2, zlibpng, jpeg, maybe libsharpyuv, libyuv, rav1e, dav1d. We can then focus on the ones we usually tweak as developers (libaom, gav1)

.github/workflows/ci-disable-gtest.yml
```diff
@@ -37,12 +37,12 @@ jobs:
         toolchain: stable
         override: true
 
-    - name: Cache external dependencies
-      id: cache-ext
-      uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
-      with:
-        path: ext
-        key: ${{ runner.os }}-disable-gtest-${{ hashFiles('ext/*.cmd') }}
+    - name: Install cargo-c
```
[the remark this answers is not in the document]
fdintino: I've set up LocalRav1e to build the cargo-c crate as part of the cmake project  only if cargo-c isn't already available. It takes longer to build cargo-c than all of the other dependencies combined, so it seemed like a sensible optimization to use the binaries provided by the project releases. Particularly since the question of how to cache the external dependency builds in CI with FetchContent/ExternalProject is unresolved.

cmake/Modules/LocalGoogletest.cmake
```diff
@@ -0,0 +1,34 @@
...
+    set(BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS_ORIG} CACHE BOOL "-" FORCE)
+endif()
+message(CHECK_PASS "fetched")
+
+set_target_properties(gtest gtest_main PROPERTIES AVIF_LOCAL ON)
+
+add_library(GTest::gtest ALIAS gtest)
+add_library(GTest::gtest_main ALIAS gtest_main)
+
+set(GTEST_INCLUDE_DIRS ${googletest_SOURCE_DIR}/googletest/include)
```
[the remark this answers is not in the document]
fdintino: yes. I was aiming to provide the same variables as find_package but I agree that it's unnecessary.

cmake/Modules/LocalDav1d.cmake
```diff
@@ -0,0 +1,106 @@
...
+
+    set(DAV1D_FOUND ON PARENT_SCOPE)
+    set(DAV1D_LIBRARY dav1d::dav1d PARENT_SCOPE)
+    set(DAV1D_LIBRARY_DIRS ${install_prefix}/lib PARENT_SCOPE)
+    set(DAV1D_INCLUDE_DIR ${install_prefix}/include PARENT_SCOPE)
+    set(DAV1D_VERSION ${AVIF_LOCAL_DAV1D_TAG} PARENT_SCOPE)
+    set_target_properties(dav1d::dav1d PROPERTIES FOLDER "ext/dav1d")
+endfunction()
+
+avif_build_local_dav1d()
```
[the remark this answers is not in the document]
fdintino: To scope the variables. Otherwise it would pollute the global namespace with a bunch of dav1d-specific variables. What I've seen in the cmake source itself as an alternative is to create macros that clear and then unset the variables set in a block of code.

cmake/Modules/LocalAom.cmake
```diff
@@ -0,0 +1,133 @@
...
+    if("${AOM_CONFIG_H}" MATCHES "CONFIG_LIBYUV 0")
+        string(REPLACE "CONFIG_LIBYUV 0" "CONFIG_LIBYUV 1" AOM_CONFIG_H "${AOM_CONFIG_H}")
+        file(WRITE ${AOM_BINARY_DIR}/config/aom_config.h "${AOM_CONFIG_H}")
+    endif()
+    target_link_libraries(aom PRIVATE $<TARGET_FILE:yuv::yuv>)
+endif()
+
+set_property(TARGET aom PROPERTY AVIF_LOCAL ON)
+
+set(AOM_INCLUDE_DIR "${libaom_SOURCE_DIR}")
```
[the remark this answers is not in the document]
fdintino: Right. I'm not using any include_dir variables, everything is based on interface include directories. Likewise for transitive library dependencies.

.github/workflows/ci-disable-gtest.yml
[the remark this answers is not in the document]
fdintino: I was trying to keep the size of this PR down, but I think ideally it should be an option to FetchContent the pre-built binaries provided by cargo-c, and even rav1e, configurable with options.

cmake/Modules/LocalDav1d.cmake
```diff
@@ -0,0 +1,106 @@
+if(NOT AVIF_LOCAL_DAV1D_TAG)
+    set(AVIF_LOCAL_DAV1D_TAG "1.2.1")
+endif()
+if(NOT AVIF_LOCAL_DAV1D_REPO)
+    set(AVIF_LOCAL_DAV1D_REPO "https://code.videolan.org/videolan/dav1d.git")
+endif()
+
+function(avif_build_local_dav1d)
+    set(source_dir "${AVIF_SOURCE_DIR}/ext/dav1d")
```
fdintino, line 9: In this PR I opted to put the source dirs in the same place, but the default behavior is for the source to also be in the build directory. The rules for determining the directories are in the cmake docs at the end of [this section](https://cmake.org/cmake/help/latest/module/ExternalProject.html#directory-options). Is there a preference for these directory options?

tests/gtest/avifavmtest.cc
fdintino: This is now in #1748

.github/workflows/ci-unix-shared-installed.yml
fdintino: I can't come up with a way to conditionally set `-DENABLE_AVX2` for macOS in aom.cmd in a way that's also compatible with sh and cmd.exe...short of creating a separate aom.sh file. So this might need to wait until aom is built as a subdirectory.

maryla-uc: Is there a way to factor this out into a separate small PR? These "ILLEGAL" errors are really annoying. I tried creating a PR that just adds these two lines to the yml files but it seems that ENABLE_AVX2 is not recognized by cmake. I guess it's tied to other changes in this PR?

And a few more questions on this topic:
- This sounds like an issue with github runners that we're working around. Do you know if there's an issue open somewhere?
- Why is the CMAKE_OSX_DEPLOYMENT_TARGET flag needed?
- In practice, I'm fairly sure I've only ever seen the ci-unix-static one fail. I doubt it's just a fluke. Maybe we don't need the changes in the other two workflows?

Thanks!

fdintino: The `ENABLE_AVX2` flag is actually in AOM (see [here](https://aomedia.googlesource.com/aom/+/e2318cdca95519318af93205fe6d9156b5353924/build/cmake/aom_config_defaults.cmake#53)). In this pull request, I am adding libaom with `add_subdirectory` to incorporate it into the larger cmake project, which allows one to pass `-DENABLE_AVX2=OFF`. Whereas on the libavif main branch it gets built in a separate step by running aom.cmd. I can't think of a way to pass arguments to that script that would work for windows and unix, and we presumably don't want to disable avx2 for everyone just to get around this issue.

As for your observation: I _have_ seen this error in all of the workflows that use macos, but it does seem to be much more frequent for ci-unix-static. And there is a pattern I've noticed: the builds seem to fail for macOS 12.7 but not for 12.6.9 (you can see the version by expanding the "Operating system" output in the "Set up job" step). If you use tmate to open a shell and run `sysctl machdep.cpu`, you'll see that a 12.6.9 runner always has AVX2 and a 12.7 runner does not. Anecdotally it seems like the ci-unix-static gets assigned the 12.7 worker more often. I can only speculate as to why that is.

I found an issue opened a year ago to runner-images on the subject of AVX2 in macos runners: https://github.com/actions/runner-images/issues/6845. The user who closed that issue references a comment on a separate issue https://github.com/actions/runner-images/issues/5609#issuecomment-1137149036, which reads:

> Unfortunately, at this time out agents are not guaranteed to have AVX2 CPU instruction, and only the way to enable the feature for you on permanent basis is to use self-hosted agents(where it is possible to control the underlying hardware).

I don't know if it's worth raising a new issue, since it seems like the 12.6.9 runners have AVX2 pretty reliably, and the 12.7 runners represent a regression of sorts. 

FWIW, AOM actually does cpu feature detection before it uses its own avx2 optimizations, but when building it will add the `-mavx2` compile flag so long as the compiler supports it, and that enables avx2 optimizations in glibc. The failures in these tests all seem to be caused by a call to the `round` function, which has a `vpbroadcastq` instruction in the SIMD-optimized version.

fdintino: The CMAKE_OSX_TARGET is a red herring. Before I figured out that the hardware was to blame, I had hoped that targeting a specific OS X version might resolve this apparently discrepancy between 12.6.9 and 12.7.

maryla-uc: Thanks for all the extra context, I understand the issue much better now.
Can we maybe use a `sed` command to add  `-DENABLE_AVX2=OFF` to aom.cmd in the CI workflows, similar to what we used to do for shared libs here? https://github.com/AOMediaCodec/libavif/commit/12e066686892df1c8201cfb0d8d6c68ad248c872#diff-cf7515a55275bc6077268c78cd03a0f56422f2e21e35040f907cb7965aed4f14L45
It's not *pretty* but I'll take any fix at this point.

fdintino: yes, I think that ought to work.

fdintino: Just make sure to do it before the actions/cache call

fdintino:
Yes, I think so. What are your thoughts on my question above about source_dir? https://github.com/AOMediaCodec/libavif/pull/1744#discussion_r1389640067

cmake/Modules/LocalDav1d.cmake
vrabaud: I'd say the only reason to keep it in ext would be for legacy. ext makes it clearer to:
- tweak build parameters (but now that will be in the Local*cmake files)
- patch the repos
If we modify the source of a FetchContent, do they get reset anyway by CMake? Or is the only way to patch the sources through the CMake file?

fdintino: There are a few ways to prevent cmake from overwriting local changes to source repos. One option is to set `FETCHCONTENT_SOURCE_DIR_<PackageName>`; it allows you to override the source directory, and FetchContent will use whatever is there instead of fetching it. You could also make `DOWNLOAD_COMMAND` a no-op `if(EXISTS)`, either by default or based on a user option. If we ended up going that route maybe we would only put SOURCE_DIR in ext if that flag is present.

vrabaud: Ok, so as long as it is documented, I think we do not care. My preference would be to check if there is anything in ext: if so, take it, otherwise FetchContent it in build.
A big use case we have is to find bugs in libaom/dav1d: we might need to go down those repos and modify them and/or use a specific tag.

vrabaud:
I believe I am done with all the CMake changes I wanted to put in so you should be ok rebasing. Thx!

fdintino:
Sorry for the delay. It probably won't be until next week that I'm able to come back to this to rebase and clean up this PR.

vrabaud:
Thx for all the updates! should I start reviewing?

fdintino:
@vrabaud this PR is ready for another review. Sorry for the long gaps in my responses; I'm out on leave from work with a newborn at home, so my attention is a bit divided.

I've reworked the LocalPackage.cmake modules so that they're compatible with the existing ext scripts: if the library artifact is found in the expected location, the behavior is unchanged from what it is now. I've also set it up so that FetchContent will use the source directory in `ext` if it exists (overriding any downloads or git clones). I'm agnostic about how best to handle the transition from the ext scripts (if at all); I leave it up to you and the other maintainers to determine what the desired behavior should be. One motivation for keeping compatibility with the existing ext scripts is to avoid integrating the CI fuzz build flags into the fetchcontent builds, and also to put off the work of incorporating changes like #2034 into the rebase.

cmake/Modules/LocalLibsharpyuv.cmake
```diff
@@ -1,10 +1,62 @@
+set(AVIF_LOCAL_LIBSHARPYUV_GIT_TAG v1.3.2)
+
 set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/libwebp/build/libsharpyuv${CMAKE_STATIC_LIBRARY_SUFFIX}")
-if(NOT EXISTS "${LIB_FILENAME}")
-    message(FATAL_ERROR "libavif(AVIF_LIBSHARPYUV=LOCAL): ${LIB_FILENAME} is missing, bailing out")
-endif()
+if(EXISTS "${LIB_FILENAME}")
```
vrabaud, line 4: All those checks for a static lib are for backward compatibility right? In case somebody has built those libs locally. We could decide that this is not supported which would result in less code right ?

cmake/Modules/LocalRav1e.cmake
```diff
@@ -1,8 +1,113 @@
...
+    )
+    if(EXISTS "${AVIF_SOURCE_DIR}/ext/rav1e")
+        message(STATUS "libavif(AVIF_CODEC_RAV1E=LOCAL): ext/rav1e found; using as FetchContent SOURCE_DIR")
+        set(FETCHCONTENT_SOURCE_DIR_RAV1E "${AVIF_SOURCE_DIR}/ext/rav1e")
+        message(CHECK_START "libavif(AVIF_CODEC_RAV1E=LOCAL): configuring rav1e")
+    else()
+        message(CHECK_START "libavif(AVIF_CODEC_RAV1E=LOCAL): fetching and configuring rav1e")
+    endif()
+
+    find_program(CARGO_CINSTALL cargo-cinstall HINTS "$ENV{HOME}/.cargo/bin")
```
vrabaud, line 26: maybe those 5 lines can be moved to below with `if(NOT TARGET cargo-cinstall)`

CMakeLists.txt
```diff
@@ -716,8 +733,8 @@ if(AVIF_BUILD_APPS)
         )
     endif()
     if(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
-        include(LocalLibargparse)
-        if(TARGET libargparse)
+        if(AVIF_LOCAL_LIBARGPARSE)
```
vrabaud, line 736: Why not do `include(LocalLibargparse)` and assume `AVIF_LOCAL_LIBARGPARSE` is `ON` by default?

vrabaud (commented):
Mostly ok, please rebase.

.github/workflows/ci-unix-static.yml
```diff
@@ -35,18 +35,28 @@ jobs:
...
-      uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
-      with:
-        path: ext
-        key: ${{ runner.os }}-${{ matrix.build-type }}-${{ hashFiles('ext/*.cmd', 'ext/svt.sh') }}
     - name: Setup cmake
       uses: jwlawson/actions-setup-cmake@d06b37b47cfd043ec794ffa3e40e0b6b5858a7ec # v1.14.2
       with:
-        # CMake version 3.17 is required to build libwebp (which libsharpyuv is part of) on macOS.
-        cmake-version: '3.17.x'
+        # CMake version 3.18 is required to build libxml2
```
vrabaud, line 58: no need to bump here, libxml2 is not built in this CI (it is not set as a CMake argument)

CMakeLists.txt
fdintino: That makes sense. I suppose it's implied by `AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP`.

fdintino:
The CI Windows build is failing, but I don't think that was introduced by my changes. Judging by this output from a recently run build on the main branch, it looks like this test was just being skipped previously:

```
CMake Warning at cmake/Modules/LocalLibargparse.cmake:9 (message):
  D:/a/libavif/libavif/ext/libargparse/build/argparse.lib is missing, not
  building avifgainmaputil, please run ext/libargparse.cmd
```

fdintino:
@vrabaud Any updates on this? I see there are some version bumps on the main branch that need to be incorporated here, but I've held off on including that in a rebase while awaiting further feedback. I'm happy to push that if it's helpful for review.

vrabaud:
@fdintino , sorry for the late review. I believe it is good as is. We decided to keep the legacy way of building things through ext scripts for now. We can remove that code later. Please sync the d'EPS to their latest version. Thx again for this tremendous work !

vrabaud:
Thx again ! Let's see how it goes with the core developers. We will also revisit all the static lib open bugs.

y-guyon:
It seems the CI workflows have been slow to complete since this PR have been merged.  
I believe it is due to the "Build libavif (ninja)" step that runs `rustc` now.

vrabaud:
I am enabling cache in `build/_deps` in https://github.com/AOMediaCodec/libavif/pull/2103, let's see how that goes.

y-guyon:
I also have weird issues on the CI in my fork:

https://github.com/y-guyon/libavif/actions/runs/8662720594/job/23755243272#step:15:47

```
Running `rustc --crate-name crossbeam_utils ...`
error: error parsing manifest: TOML parse error at line 18908, column 26
      |
18908 | [[pkg.rust.target.x86_64-
      |                          ^
invalid table header
expected `.`, `]]`

error: could not compile `crossbeam-utils` (lib)
```

PR #1748, opened 2023-11-10
ci: Fix a typo in the ci-unix-static-av2 workflow so AV1+AV2 tests run 

fdintino:
A typo in the ci-unix-static-av2  GHA workflow prevented it from executing tests where AVM is enabled alongside other AV1 codecs.

### DIFF
--- a/.github/workflows/ci-unix-static-av2.yml
+++ b/.github/workflows/ci-unix-static-av2.yml
@@ -86,9 +86,9 @@ jobs:
         cmake .. -G Ninja
         -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF
         -DAVIF_CODEC_AVM=ON -DAVIF_LOCAL_AVM=ON
-        -DAVIF_CODEC_DAV1D=${{ runner.also-enable-av1-codecs }} -DAVIF_LOCAL_DAV1D=ON
-        -DAVIF_CODEC_RAV1E=${{ runner.also-enable-av1-codecs }} -DAVIF_LOCAL_RAV1E=ON
-        -DAVIF_CODEC_SVT=${{ runner.also-enable-av1-codecs }} -DAVIF_LOCAL_SVT=ON
+        -DAVIF_CODEC_DAV1D=${{ matrix.also-enable-av1-codecs }} -DAVIF_LOCAL_DAV1D=ON
+        -DAVIF_CODEC_RAV1E=${{ matrix.also-enable-av1-codecs }} -DAVIF_LOCAL_RAV1E=ON
+        -DAVIF_CODEC_SVT=${{ matrix.also-enable-av1-codecs }} -DAVIF_LOCAL_SVT=ON
         -DAVIF_LOCAL_LIBYUV=ON
         -DAVIF_LOCAL_LIBSHARPYUV=ON
         -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON


vrabaud (approved):
Thx!

src/read.c
```diff
@@ -4924,6 +4924,16 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         }
     }
 
+#if defined(AVIF_CODEC_AVM)
+    if (decoder->codecChoice != AVIF_CODEC_CHOICE_AUTO) {
+        if (colorCodecType == AVIF_CODEC_TYPE_AV2 && decoder->codecChoice != AVIF_CODEC_CHOICE_AVM) {
+            return AVIF_RESULT_NO_CODEC_AVAILABLE;
+        }
+        if (colorCodecType == AVIF_CODEC_TYPE_AV1 && decoder->codecChoice == AVIF_CODEC_CHOICE_AVM) {
+            return AVIF_RESULT_NO_CODEC_AVAILABLE;
```
wantehchang, line 4933: The use of `AVIF_RESULT_NO_CODEC_AVAILABLE` for this error condition seems wrong.

This error condition is that the wrong type of codec is chosen for the image. `AVIF_RESULT_NO_CODEC_AVAILABLE` means there is no codec (of the correct type) available.

Also, `avifDecoderParse()` is designed to not involve the codec. I.e., one can build libavif with no codecs at all and `avifDecoderParse()` must still work in such a build. I think this check should be moved to `avifDecoderNextImage()` (or some function it calls).

And then, ideally this check should return some other error code to indicate the wrong type of codec was chosen for the image.

fdintino: I don't have an opinion on what it should be. I wrote this PR to fix a test that is currently failing https://github.com/AOMediaCodec/libavif/blob/main/tests/gtest/avifavmtest.cc#L54-L58

fdintino: I had assumed, apparently wrongly, that the test assertion was correct and the code was wrong.

fdintino: Your point about `avifDecoderParse` is a good one. I’d be happy to move it there. Or close this pull request and have the test fail but be skipped in CI. Let me know.

fdintino: I think my replies don't strike the tone I had intended, so let me try to give some context and pose the question in plainer terms:

In the course of experimenting with ExternalProject and FetchContent for the external dependencies, I had to make changes to ci-unix-static-av2.yml. The GHA linter in my editor flagged that `runner.also-enable-av1-codecs` is not a valid variable—and plainly it was meant to be `matrix.also-enable-av1-codecs`, so I updated it. Without that change, `DAVIF_CODEC_DAV1D` et. al. were never set to `ON`, and the pertinent test cases never got run in CI. But after I fixed that type, the tests failed. It wasn't until I ran the test with gdb locally that I realized I hadn't caused the issue.

I don't anticipate using the experimental AVM feature any time soon, so I'm not all that invested in its implementation. My only concern is that I don't introduce any regressions in the course of making changes to the build, so I _would_ like to have these tests working. Your feedback is very thoughtful, but it doesn't prescribe a clear path forward, which is the cause of the annoyed tone in my replies. As I see it, there are several options that resolve the CI issue:

1. I change the test condition from `ASSERT_EQ(..., AVIF_RESULT_NO_CODEC_AVAILABLE)` to `ASSERT_NE(..., AVIF_RESULT_OK)`, or else comment out, disable, or remove the problematic assertion. The comments around the test assertion are quite clear about what the author intended the behavior to be, but I don't disagree with your arguments against it. This is actually what I had originally done in #1744 (i.e., change to an `ASSERT_NE`) before I took a stab at addressing the issue. The current behavior would remain, to be fixed at a later date (if at all).
2. I move the `avifResult` return to `avifDecoderNextImage`; somewhere near where it calls `avifDecoderCreateCodecs` would probably make the most sense. The check could perhaps be made more robust by using the `availableCodecs` struct in avif.c, which has a member declaring which `avifCodecType` a given codec supports. If this return value should not be `AVIF_RESULT_NO_CODEC_AVAILABLE`, I'd ask that you suggest the name for the new `avifResult` enum value.
3. I close this pull request and open an issue, and leave it to you or someone else to fix it. The fact that I put this check in the wrong function shows that maybe I'm not the best person to make this fix anyway.

As I said, my only interest is in assuring I don't introduce any regressions to the AVM feature. I defer to your judgment on how I should move forward.

y-guyon: Thanks for your detailed answer and your tentative fix.

> leave it to you or someone else to fix it

Done at https://github.com/AOMediaCodec/libavif/pull/1753.

Please only keep the changes to `.github/workflows/ci-unix-static-av2.yml` in this PR and merge it once the presubmit tests pass.

vrabaud:
The PR summary needs to be siimplified now. Thx!

y-guyon (approved):

PR #1749, opened 2023-11-10
Add/update comments about avifCleanApertureBox

wantehchang:


vigneshvg (approved):

include/avif/avif.h
```diff
@@ -428,30 +428,36 @@ typedef uint32_t avifTransformFlags;
...
+    // 'pasp' from ISO/IEC 14496-12:2022 12.1.4.3
 
     // define the relative width and height of a pixel
     uint32_t hSpacing;
     uint32_t vSpacing;
 } avifPixelAspectRatioBox;
 
+// NOTE: The members of the avifCleanApertureBox struct are declared as uint32_t to match the
+// unsigned int(32) type used in ISO/IEC 14496-12:2022 faithfully. However, ISO/IEC 14496-12:2022
+// 12.1.4.1 clearly interprets these values as signed int(32) and talk about them being strictly
```
y-guyon, line 440:
```diff
-talk
+talks
```

include/avif/avif.h
```diff
@@ -428,30 +428,36 @@ typedef uint32_t avifTransformFlags;
...
+// positive, positive, or negative. Cast these struct members to int32_t before use.
```
y-guyon, line 441:
```diff
-positive, positive, or negative.
+positive, zero, or negative.
```

include/avif/avif.h
y-guyon, line 440:
```diff
-these values
+horizOffN and vertOffN
```

y-guyon (approved):

include/avif/avif.h
wantehchang: I think all eight members of `CleanApertureBox`, rather than just `horizOffN` and `vertOffN`, are interpreted as `signed int(32)`. Only `horizOffN` and `vertOffN` are allowed to be negative.

One evidence is how ISO BMFF talks about these eight values in 12.1.4.1. Another evidence is the definition of the clean aperture atom in QuickTime, which I understand is the predecessor of ISO BMFF:
https://developer.apple.com/documentation/quicktime-file-format/clean_aperture

libavif casts all eight members of the `avifCleanApertureBox` struct to `int32_t` before use. For example,

```
avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
                                             const avifCleanApertureBox * clap,
                                             uint32_t imageW,
                                             uint32_t imageH,
                                             avifPixelFormat yuvFormat,
                                             avifDiagnostics * diag)
{
    ...

    const int32_t widthN = (int32_t)clap->widthN;
    const int32_t widthD = (int32_t)clap->widthD;
    const int32_t heightN = (int32_t)clap->heightN;
    const int32_t heightD = (int32_t)clap->heightD;
    const int32_t horizOffN = (int32_t)clap->horizOffN;
    const int32_t horizOffD = (int32_t)clap->horizOffD;
    const int32_t vertOffN = (int32_t)clap->vertOffN;
    const int32_t vertOffD = (int32_t)clap->vertOffD;
```

See also the definition of the `avifFraction` type, which is used with `avifCleanApertureBox`:

```
typedef struct avifFraction 
{
    int32_t n;
    int32_t d;
} avifFraction;
```

include/avif/avif.h
```diff
@@ -428,30 +428,39 @@ typedef uint32_t avifTransformFlags;
...
     uint32_t hSpacing;
     uint32_t vSpacing;
 } avifPixelAspectRatioBox;
 
+// NOTE: The members of the avifCleanApertureBox struct are declared as uint32_t to match the
+// unsigned int(32) type used in ISO/IEC 14496-12:2022 faithfully. However, ISO/IEC 14496-12:2022
+// 12.1.4.1 clearly interprets these values as signed int(32) and talks about them being strictly
+// positive, positive, or negative. Cast these struct members to int32_t before use. See also the
+// clean aperture extension in the QuickTime File Format:
+// https://developer.apple.com/documentation/quicktime-file-format/clean_aperture
```
wantehchang, line 443: Another thing we can say in this comment is to encourage people to use the `avifCropRect` struct instead and use the two conversion functions to convert between `avifCropRect` and `avifCleanApertureBox`. Do you think that's a good idea?

wantehchang (commented):
Thanks for the review! PTAL.

include/avif/avif.h
y-guyon: Yes.

include/avif/avif.h
y-guyon: On the ISOBMFF draft I have, this is the new definition:

```c
class CleanApertureBox extends Box('clap')
{
   unsigned int(32) cleanApertureWidthN;
   unsigned int(32) cleanApertureWidthD;
   unsigned int(32) cleanApertureHeightN;
   unsigned int(32) cleanApertureHeightD;
   signed int(32) horizOffN;
   unsigned int(32) horizOffD;
   signed int(32) vertOffN;
   unsigned int(32) vertOffD;
}
```
so this is why I suggested that edit.

On top of that, the definition you mentioned at https://github.com/AOMediaCodec/libavif/pull/1749/commits/41f1839b58fa9648480888d091b90f7b72e6b0f5#r1391480366 makes it useless to cast `D`, `cleanApertureWidth` and `cleanApertureHeight` to `signed` because they shall not be negative.

Anyway suggesting using `avifCropRect` is better.

wantehchang: Do you know if the change to the `CleanApertureBox` in the ISOBMFF draft was discussed in an issue? I'd like to take a look at the discussion. Thanks.

It is unusual to mix signed and unsigned types like that. I'd like to find out if the clean aperture definition in the QuickTime file format was mentioned in the discussion.

It is okay to use a signed type for a variable that cannot be negative. For example, the `width` and `height` in many APIs are of the signed `int` type.

y-guyon: > Do you know if the change to the `CleanApertureBox` in the ISOBMFF draft was discussed in an issue? I'd like to take a look at the discussion. Thanks.

See https://github.com/MPEGGroup/FileFormat/issues/41 and https://mpeg.expert/software/MPEG/Systems/FileFormat/isobmff/-/issues/101.

> It is okay to use a signed type for a variable that cannot be negative. For example, the `width` and `height` in many APIs are of the signed `int` type.

To use a signed type for a variable, sure, but to explicitly request to cast it is confusing.

PR #1752, opened 2023-11-13
Use a manifest on Windows for UTF8 handling

vrabaud:
This first reverts https://github.com/AOMediaCodec/libavif/pull/1693 (except for test_cmd.sh) and applies a manifest on Windows platforms.

CMake handles manifests as source files on Windows but most likely only for MSVC (cf https://cmake.org/cmake/help/v3.4/release/3.4.html#other).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -784,16 +784,27 @@ endif()
 
 if(AVIF_BUILD_APPS)
     add_executable(avifenc apps/avifenc.c)
-    if(AVIF_USE_CXX)
-        set_target_properties(avifenc PROPERTIES LINKER_LANGUAGE "CXX")
-    endif()
-    target_link_libraries(avifenc avif_apps)
     add_executable(avifdec apps/avifdec.c)
-    if(AVIF_USE_CXX)
-        set_target_properties(avifdec PROPERTIES LINKER_LANGUAGE "CXX")
-    endif()
-    target_link_libraries(avifdec avif_apps)
-
+    foreach(PROG IN ITEMS avifenc avifdec)
+        if(AVIF_USE_CXX)
+            set_target_properties(${PROG} PROPERTIES LINKER_LANGUAGE "CXX")
+        endif()
+        target_link_libraries(${PROG} avif_apps)
+        if(WIN32)
+            # Add a manifest file to deal with UTF8.
+            if(CMAKE_C_COMPILER_ID MATCHES "MSVC")
+                target_sources(${PROG} PUBLIC ${CMAKE_SOURCE_DIR}/src/utf8.manifest)
+            else()
+                add_custom_command(
+                    TARGET ${PROG}
+                    POST_BUILD
+                    COMMAND "mt.exe" -manifest \"${CMAKE_SOURCE_DIR}\\src\\utf8.manifest\"
+                            -inputresource:\"$<TARGET_FILE:${PROG}>\"\;\#1 -outputresource:\"$<TARGET_FILE:${PROG}>\"\;\#1
+                    COMMENT "Adding manifest..."
+                )
+            endif()
+        endif()
+    endforeach()
     if(NOT SKIP_INSTALL_APPS AND NOT SKIP_INSTALL_ALL)
         install(
             TARGETS avifenc avifdec
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -14,18 +14,12 @@
 #include <stdlib.h>
 #include <string.h>
 
-#if defined(_WIN32)
-#include <locale.h>
-#define WIN32_LEAN_AND_MEAN
-#include <windows.h>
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -793,7 +793,18 @@ if(AVIF_BUILD_APPS)
...
     target_link_libraries(avifdec avif_apps)
-
+    if(WIN32)
+        foreach(PROG IN LISTS avifenc avifdec)
+            add_custom_command(
+                TARGET ${PROG}
+                POST_BUILD
+                COMMAND "mt.exe" -manifest \"${CMAKE_SOURCE_DIR}\\src\\utf8.manifest\"
+                        -inputresource:\"$<TARGET_FILE:${PROG}>\"\;\#1 -outputresource:\"$<TARGET_FILE:${PROG}>\"\;\#1
+                COMMENT "Adding manifest..."
```
wantehchang, line 803: When I tested this, I didn't need to use add_custom_command(). It sufficed to just add utf8.manifest as a source file.

I will try to find my changes.

CMakeLists.txt
```diff
@@ -784,16 +784,27 @@ endif()
...
+        target_link_libraries(${PROG} avif_apps)
+        if(WIN32)
+            # Add a manifest file to deal with UTF8.
+            if(CMAKE_C_COMPILER_ID MATCHES "MSVC")
+                target_sources(${PROG} PUBLIC ${CMAKE_SOURCE_DIR}/src/utf8.manifest)
+            else()
+                add_custom_command(
+                    TARGET ${PROG}
+                    POST_BUILD
+                    COMMAND "mt.exe" -manifest \"${CMAKE_SOURCE_DIR}\\src\\utf8.manifest\"
```
kmilos, line 801: I don't think this can work on MinGW, there is no mt.exe...

CMakeLists.txt
```diff
@@ -784,16 +784,27 @@ endif()
...
-    target_link_libraries(avifdec avif_apps)
-
+    foreach(PROG IN ITEMS avifenc avifdec)
+        if(AVIF_USE_CXX)
+            set_target_properties(${PROG} PROPERTIES LINKER_LANGUAGE "CXX")
+        endif()
+        target_link_libraries(${PROG} avif_apps)
+        if(WIN32)
+            # Add a manifest file to deal with UTF8.
+            if(CMAKE_C_COMPILER_ID MATCHES "MSVC")
```
kmilos, line 795: `if(MSVC)` simply

CMakeLists.txt
vrabaud: Can MINGW access the Windows SDK and therefore "mt.exe"?
It seems adding the manifest as source (or at least the .rc) works, cf https://doc.magnum.graphics/magnum/platforms-windows.html . Can you confirm this is the way to go? I have no experience with manifest files, @wantehchang just found this cleaner solution for handling UTF8.

kmilos: I don't think one can assume MinGW users/devs will have access to the Windows SDK.

I might be completely wrong here, but perhaps a conversation is to be had w/ the MinGW (or is it Cygwin?) project to ship a pre-built _UTF8_ object manifest that can be statically linked in then, [like the existing ("regular"?) one](https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-windows-default-manifest?repo=ucrt64)?

kmilos: Furthermore, the https://cygwin.com/git/?p=cygwin-apps/windows-default-manifest.git;a=tree also seems to wrap that default manifest into a .rc file, and resource files can be dealt with w/o a problem in MinGW toolchains (windres is shipped). Maybe that could be a temporary workaround?

kmilos: Maybe take a [lead ](https://blog.r-project.org/2022/11/07/issues-while-switching-r-to-utf-8-and-ucrt-on-windows/index.html)from [RStudio](https://github.com/rstudio/rstudio/blob/a1d6917afc68965cfb7a964a9afcfcb04da0e7d7/src/cpp/session/CMakeLists.txt#L466)?

wantehchang: Miloš: Thank you very much for the research.

The UTF-8 support for libavif is simpler than the UTF-8 support for RStudio, because the libavif library itself has only one function, `avifIOCreateFileReader()`, that takes a `filename` input parameter, and that function faithfully passes `filename` to `fopen()` with no processing. So we can document that the libavif library honors the code pages selected by the application, and we just need to solve the UTF-8 support problem for the three command-line programs, avifdec, avifenc, and are_images_equal. And it is fine to solve the UTF-8 support problem on Windows 10 and later only so that we can use the simple solution of the manifest file.

vrabaud:
Go with the cleaner https://github.com/AOMediaCodec/libavif/pull/1900

PR #1756, opened 2023-11-14
Remove potential out of bound access to alphaItemIndices

vigneshvg:
It is possible to craft a file that has more alpha auxiliary items than color items and trigger an out of bound access into alphaItemIndices in the for loop.

Fix is to allocate alphaItemIndices to be the same size as meta.items.count so that the existing inequality check can cover both the cases.

Also, update a diagnostic error message to cover all cases (i.e.) there can be more grids than necessary as well.


src/read.c
```diff
@@ -4247,7 +4247,7 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
         *isAlphaItemInInput = AVIF_FALSE;
         return AVIF_RESULT_OK;
     }
-    uint32_t * alphaItemIndices = avifAlloc(colorItemCount * sizeof(uint32_t));
+    uint32_t * alphaItemIndices = avifAlloc(meta->items.count * sizeof(uint32_t));
```
vigneshvg, line 4250: an alternative solution is to check for `alphaItemCount` not exceeding `colorItemCount` in line 4262. but that involves repeating the code for the failure return. so i chose the simpler approach. the overhead should be minimal since this is an extreme case anyways.

wantehchang (approved):
Thanks for the fix. I have some questions.

src/read.c
wantehchang: The correctness of this solution requires all the items in the `meta->items` array have distinct IDs. I believe that is the case, because the array is constructed by `avifMetaFindOrCreateItem()`. Please double check.

src/read.c
```diff
@@ -4266,7 +4266,8 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
         }
     }
     if (alphaItemCount != colorItemCount) {
-        // Not all the color items had an alpha auxiliary attached to it. Report this case as an image without alpha channel.
+        // Number of color items does not match the number of alpha auxiliary items. Report this case as an image without alpha
+        // channel.
```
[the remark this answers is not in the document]
vigneshvg: yes thats a good point. i have updated the inner loop to ensure that each color grid tile has exactly one alpha grid tile.

actually, i just realized that another reason for failure is if the outer loop runs more than `colorItemCount` times. because in the current code, we never validate that there are only as many color items as claimed in the grid config available. so i have added that validation as well to `avifDecoderItemReadAndParse`.

these two changes together will ensure that `colorItemCount == alphaItemCount` at the end of the for loop here. so i have added that as an `assert.

vigneshvg:
The latest commit seems to be causing test failures. I will look at them tomorrow morning.

vigneshvg:
> The latest commit seems to be causing test failures. I will look at them tomorrow morning.

This is now fixed. PTAL.

src/read.c
```diff
@@ -4257,21 +4267,31 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+                        *alphaItem = NULL;
+                        *isAlphaItemInInput = AVIF_FALSE;
+                        return AVIF_RESULT_INVALID_IMAGE_GRID;
+                    }
                     alphaItemIndices[alphaItemCount++] = j;
+                    seenAlphaForCurrentItem = AVIF_TRUE;
                 }
             }
+            if (!seenAlphaForCurrentItem) {
+                // No alpha auxiliary item was found for the current item. Treat this as an image without alpha.
```
wantehchang, line 4286: Could you remind me why we should treat this as an image without alpha, rather than return an error?

src/read.c
```diff
@@ -4257,21 +4267,31 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
                 if (avifDecoderItemIsAlphaAux(auxlItem, item->id)) {
+                    if (seenAlphaForCurrentItem) {
+                        // Multiple items are claiming to be the alpha auxiliary of the current item.
+                        avifFree(alphaItemIndices);
+                        *alphaItem = NULL;
+                        *isAlphaItemInInput = AVIF_FALSE;
+                        return AVIF_RESULT_INVALID_IMAGE_GRID;
+                    }
                     alphaItemIndices[alphaItemCount++] = j;
+                    seenAlphaForCurrentItem = AVIF_TRUE;
```
wantehchang, line 4282: Should we also verify that `auxlItem->dimgForID == 0`?

Note that we will set `auxlItem->dimgForID` to the fake alpha grid item at line 4306 below. So we should make sure we don't overwrite an existing value. Also `auxlItem->dimgForID == 0` is implied by this diagram:

```
[primary item grid]
   ^       ^
   |dimg   |dimg
   |       |
[color] [color]
   ^       ^
   |auxl   |auxl
   |       |
[alpha] [alpha]
```

src/read.c
```diff
@@ -1441,7 +1441,7 @@ static avifResult avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifI
 
     if (tilesAvailable != grid->rows * grid->columns) {
         avifDiagnosticsPrintf(&decoder->diag,
-                              "Grid image of dimensions %ux%u requires %u tiles, and only %u were found",
+                              "Grid image of dimensions %ux%u requires %u tiles, but %u were found",
```
wantehchang, line 1444: Is the check here redundant now that we validate `grid->rows * grid->columns` in `avifDecoderItemReadAndParse`?

wantehchang (approved):
LGTM.

src/read.c
vigneshvg: i thought about this as well, the only subtle difference is that this check happens for all grid items (not just those found in the actual input).

so there is no harm in keeping this. or if you prefer i can change this one into an assert and document why that is sufficient.

src/read.c
vigneshvg: When the original change was made, that was libavif's behavior. And there seems to be some test files this way. So let's not change the behavior in this PR.

I am okay treating this as an error, but we should update the test files to ensure we don't break anything.

src/read.c
vigneshvg: Done.

src/read.c
```diff
@@ -4257,21 +4267,38 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+                        *isAlphaItemInInput = AVIF_FALSE;
+                        return AVIF_RESULT_INVALID_IMAGE_GRID;
+                    }
+                    if (auxlItem->dimgForID != 0) {
+                        // Alpha auxiliary is dimg for another item. This is invalid.
+                        avifFree(alphaItemIndices);
+                        *alphaItem = NULL;
+                        *isAlphaItemInInput = AVIF_FALSE;
+                        return AVIF_RESULT_INVALID_IMAGE_GRID;
+                    }
```
wantehchang, line 4287: Nit: merge these two if statements (lines 4274-4287)?

Note: I am tempted to suggest we use goto statements to share the cleanup code. But let's be conservative and focus on the bug fix in this pull request.

wantehchang (approved):
LGTM.

src/read.c
vigneshvg: > Nit: merge these two if statements (lines 4274-4287)?
> 

Done.

> Note: I am tempted to suggest we use goto statements to share the cleanup code. But let's be conservative and focus on the bug fix in this pull request.

Yes, i was planning on doing this in a follow-up anyway.

wantehchang (approved):

vigneshvg:
@wantehchang i rebased this PR after PR #1757 was merged. please take another look.

wantehchang (approved):
LGTM.

PR #1757, opened 2023-11-14
Do not store item pointers until all items are created

vigneshvg:
Calling avifMetaFindOrCreateItem() could invalidate all the existing item pointers that are being stored by the caller (since the function could resize the item array).

This patch fixes avifDecoderReset by storing the item indices instead of item pointers until all the items are either created or found.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -781,6 +781,8 @@ static avifResult avifCheckItemID(const char * boxFourcc, uint32_t itemID, avifD
     return AVIF_RESULT_OK;
 }
 
+// CAUTION: This function could potentially resize the meta->items array thereby invalidating all existing pointers that are being
+// stored locally. So if this function is being called, exercise caution in the caller to not use invalid pointers.
 static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avifDecoderItem ** item)
 {
     *item = NULL;
@@ -1914,11 +1916,12 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
 // isItemInInput must be false if the item is a made-up structure
 // (and thus not part of the parseable input bitstream).
 static avifResult avifDecoderItemReadAndParse(const avifDecoder * decoder,
-                                              avifDecoderItem * item,
+                                              int itemIndex,
                                               avifBool isItemInInput,
                                               avifImageGrid * grid,
                                               avifCodecType * codecType)
 {
+    avifDecoderItem * item = &decoder->data->meta->items.item[itemIndex];
     if (!memcmp(item->type, "grid", 4)) {
         if (isItemInInput) {
             avifROData readData;
@@ -4186,8 +4189,8 @@ static avifBool avifDecoderItemShouldBeSkipped(const avifDecoderItem * item)
            (avifGetCodecType(item->type) == AVIF_CODEC_TYPE_UNKNOWN && memcmp(item->type, "grid", 4)) || item->thumbnailForID != 0;
 }
 
-// Returns the primary color item if found, or NULL.
-static avifDecoderItem * avifMetaFindColorItem(avifMeta * meta)
+// Returns the index of the primary color item if found, or -1.
+static int avifMetaFindColorItem(avifMeta * meta)
 {
     for (uint32_t itemIndex = 0; itemIndex < meta->items.count; ++itemIndex) {
         avifDecoderItem * item = &meta->items.item[itemIndex];
@@ -4195,10 +4198,10 @@ static avifDecoderItem * avifMetaFindColorItem(avifMeta * meta)
             continue;
         }
         if (item->id == meta->primaryItemID) {
-            return item;
+            return itemIndex;
         }
     }
-    return NULL;
+    return -1;
 }
 
 // Returns AVIF_TRUE if item is an alpha auxiliary item of the parent color
@@ -4211,39 +4214,40 @@ static avifBool avifDecoderItemIsAlphaAux(avifDecoderItem * item, uint32_t color
     return auxCProp && isAlphaURN(auxCProp->u.auxC.auxType);
 }
 
-// Finds the alpha item whose parent item is colorItem and sets it in the alphaItem output parameter. Returns AVIF_RESULT_OK on
-// success. Note that *alphaItem can be NULL even if the return value is AVIF_RESULT_OK. If the colorItem is a grid and the alpha
-// item is represented as a set of auxl items to each color tile, then a fake item will be created and *isAlphaItemInInput will be
-// set to AVIF_FALSE. In this case, the alpha item merely exists to hold the locations of the alpha tile items. The data of this
-// item need not be read and the pixi property cannot be validated. Otherwise, *isAlphaItemInInput will be set to AVIF_TRUE when
-// *alphaItem is not NULL.
+// Finds the alpha item whose parent item is the color item and sets its index in the alphaItemIndex output parameter. Returns
+// AVIF_RESULT_OK on success. Note that *alphaItemIndex can be -1 even if the return value is AVIF_RESULT_OK. If the color item is
+// a grid and the alpha item is represented as a set of auxl items to each color tile, then a fake item will be created and
+// *isAlphaItemInInput will be set to AVIF_FALSE. In this case, the alpha item merely exists to hold the locations of the alpha
+// tile items. The data of this item need not be read and the pixi property cannot be validated. Otherwise, *isAlphaItemInInput
+// will be set to AVIF_TRUE when *alphaItemIndex is not -1.
 static avifResult avifMetaFindAlphaItem(avifMeta * meta,
-                                        const avifDecoderItem * colorItem,
+                                        int colorItemIndex,
                                         const avifTileInfo * colorInfo,
-                                        avifDecoderItem ** alphaItem,
+                                        int * alphaItemIndex,
                                         avifTileInfo * alphaInfo,
                                         avifBool * isAlphaItemInInput)
 {
+    const avifDecoderItem * colorItem = &meta->items.item[colorItemIndex];
     for (uint32_t itemIndex = 0; itemIndex < meta->items.count; ++itemIndex) {
         avifDecoderItem * item = &meta->items.item[itemIndex];
         if (avifDecoderItemShouldBeSkipped(item)) {
             continue;
         }
         if (avifDecoderItemIsAlphaAux(item, colorItem->id)) {
-            *alphaItem = item;
+            *alphaItemIndex = itemIndex;
             *isAlphaItemInInput = AVIF_TRUE;
             return AVIF_RESULT_OK;
         }
     }
     if (memcmp(colorItem->type, "grid", 4)) {
-        *alphaItem = NULL;
+        *alphaItemIndex = -1;
         *isAlphaItemInInput = AVIF_FALSE;
         return AVIF_RESULT_OK;
     }
     // If color item is a grid, check if there is an alpha channel which is represented as an auxl item to each color tile item.
     uint32_t colorItemCount = colorInfo->grid.rows * colorInfo->grid.columns;
     if (colorItemCount == 0) {
-        *alphaItem = NULL;
+        *alphaItemIndex = -1;
         *isAlphaItemInInput = AVIF_FALSE;
         return AVIF_RESULT_OK;
     }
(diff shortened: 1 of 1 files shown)


vigneshvg:
I have audited the code for all calls into `avifMetaFindOrCreateItem` and found that the only other function with this potential issue is `avifParseCondensedImageBox`. I will look into it and address that in a separate PR.

maryla-uc (approved):
Thanks for the fix!
To avoid this issue in the future maybe we could have a compile mode where avifArrayPush always resizes the array (only increases its capacity by 1) therefore alwyas invalidating pointers, that use it in tests?

y-guyon:
Thanks for this PR, I encountered this issue multiple times in the past when refactoring the code, it is a classic blunder of `read.c` to me.  
I see that local pointers are used in most places for convenience, so risks remain despite passing indices instead of pointers as function arguments. Since pointers were more handy than indices, please consider using a fixed-size array or an array of pointers instead. I believe in C++ it would be a `std::vector<std::unique_ptr<avifDecoderItem>>` or a privately encapsulated `std::vector<avifDecoderItem>` with a public getter.

vigneshvg:
> Thanks for this PR, I encountered this issue multiple times in the past when refactoring the code, it is a classic blunder of `read.c` to me. 

i agree.

> I see that local pointers are used in most places for convenience, so risks remain despite passing indices instead of pointers as function arguments.

I have audited the code for all uses of this function and there are only two places where we create new items on the go while potentially storing pointers to existing items. This PR fixes one of them (other one is `avifParseCondensedImageBox `). Every where else we only store read-only pointers which don't modify the items array. So it is safe to use them.

>  Since pointers were more handy than indices, please consider using a fixed-size array or an array of pointers instead. I believe in C++ it would be a `std::vector<std::unique_ptr<avifDecoderItem>>` or a privately encapsulated `std::vector<avifDecoderItem>` with a public getter.

I don't think a fixed size array can be used for items because the theoretical number of items can be too large. We could consider using a fixed size array of pointers but that seems like a bigger change and i will leave that to a separate PR.

src/read.c
```diff
@@ -4712,6 +4726,14 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         }
 #endif // AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
 
+        // When searching for alpha and gainmap items, it is possible that the data->meta->items array is re-sized in calls to
+        // avifMetaFindOrCreateItem. So it is not safe to store pointers into that array until all the items are either found or
+        // created. From this point onwards, we do not create any more items, so it is safe to store the pointers.
```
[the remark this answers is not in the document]
vigneshvg: Done.

src/read.c
```diff
@@ -781,6 +781,8 @@ static avifResult avifCheckItemID(const char * boxFourcc, uint32_t itemID, avifD
     return AVIF_RESULT_OK;
 }
 
+// CAUTION: This function could potentially resize the meta->items array there by invalidating all existing pointers that are
```
wantehchang, line 784: Nit: there by => thereby

src/read.c
```diff
@@ -4211,39 +4214,40 @@ static avifBool avifDecoderItemIsAlphaAux(avifDecoderItem * item, uint32_t color
     return auxCProp && isAlphaURN(auxCProp->u.auxC.auxType);
 }
 
-// Finds the alpha item whose parent item is colorItem and sets it in the alphaItem output parameter. Returns AVIF_RESULT_OK on
-// success. Note that *alphaItem can be NULL even if the return value is AVIF_RESULT_OK. If the colorItem is a grid and the alpha
-// item is represented as a set of auxl items to each color tile, then a fake item will be created and *isAlphaItemInInput will be
-// set to AVIF_FALSE. In this case, the alpha item merely exists to hold the locations of the alpha tile items. The data of this
-// item need not be read and the pixi property cannot be validated. Otherwise, *isAlphaItemInInput will be set to AVIF_TRUE when
-// *alphaItem is not NULL.
+// Finds the alpha item whose parent item is colorItem and sets its index in the alphaItemId output parameter. Returns
```
wantehchang, line 4217: Replace `colorItem` and `alphaItemId` with the new parameter names.

Perhaps `colorItem` can be replaced by "color item" (i.e., a description rather than a parameter name).

src/read.c
```diff
@@ -4211,39 +4214,40 @@ static avifBool avifDecoderItemIsAlphaAux(avifDecoderItem * item, uint32_t color
...
+// AVIF_RESULT_OK on success. Note that *alphaItemIndex can be -1 even if the return value is AVIF_RESULT_OK. If the colorItem is
```
wantehchang, line 4218: `colorItem` needs to be replaced by `colorItemIndex` or "color item".

src/read.c
```diff
@@ -4268,24 +4272,29 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+    const avifResult result = avifMetaFindOrCreateItem(meta, maxItemID + 1, &alphaItem); // Create new empty item.
     if (result != AVIF_RESULT_OK) {
         avifFree(alphaItemIndices);
         *isAlphaItemInInput = AVIF_FALSE;
         return result;
     }
-    memcpy((*alphaItem)->type, "grid", 4); // Make it a grid and register alpha items as its tiles.
-    (*alphaItem)->width = colorItem->width;
-    (*alphaItem)->height = colorItem->height;
+    // avifMetaFindOrCreateItem() could invalidate all existing item pointers. So repopulate the colorItem pointer.
```
wantehchang, line 4286: Nit: "repopulate" implies filling out the members of a struct. Perhaps "reassign" or "reset"?

src/read.c
```diff
@@ -4268,24 +4272,29 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+    memcpy(alphaItem->type, "grid", 4); // Make it a grid and register alpha items as its tiles.
+    alphaItem->width = colorItem->width;
+    alphaItem->height = colorItem->height;
     for (uint32_t i = 0; i < alphaItemCount; ++i) {
         avifDecoderItem * item = &meta->items.item[alphaItemIndices[i]];
-        item->dimgForID = (*alphaItem)->id;
+        item->dimgForID = alphaItem->id;
     }
     avifFree(alphaItemIndices);
+    *alphaItemIndex = meta->items.count - 1;
```
wantehchang, line 4297: Nit/optional: It seems clearer to do this assignment immediately after the `avifMetaFindOrCreateItem()` call. We can probably wait until we know the call is successful, after line 4285.

src/read.c
```diff
@@ -4342,41 +4351,49 @@ static avifResult avifDecoderDataFindToneMappedImageItem(const avifDecoderData *
...
+        avifDecoderDataFindToneMappedImageItem(data, &data->meta->items.item[colorItemIndex], &toneMappedImageItemTmp, &gainMapItemID));
     if (!toneMappedImageItemTmp) {
         return AVIF_RESULT_OK;
     }
 
     assert(gainMapItemID != 0);
-    avifDecoderItem * gainMapItemTmp;
-    AVIF_CHECKRES(avifMetaFindOrCreateItem(data->meta, gainMapItemID, &gainMapItemTmp));
-    if (avifDecoderItemShouldBeSkipped(gainMapItemTmp)) {
+    avifDecoderItem * gainMapItem;
```
wantehchang, line 4381: Nit: I wonder if the original code uses a `gainMapItemTmp` local variable so that the output parameter `*gainMapItem` is only assigned a value when the function returns AVIF_RESULT_OK. If so, the new code should do the same with the `*gainMapItemIndex` output parameter.

src/read.c
```diff
@@ -4712,6 +4728,14 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         }
 #endif // AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
 
+        // When searching for alpha and gainmap items, it is possible that the data->meta->items array is re-sized in calls to
+        // avifMetaFindOrCreateItem. So it is not safe to store pointers into that array until all the items are either found or
```
wantehchang, line 4732: Nit: "that array" is not clear. Replaced it with "the mainItems array".

src/read.c
```diff
@@ -4342,41 +4351,49 @@ static avifResult avifDecoderDataFindToneMappedImageItem(const avifDecoderData *
...
         avifDiagnosticsPrintf(data->diag, "Box[tmap] gain map item %d is not a supported image type", gainMapItemID);
         return AVIF_RESULT_INVALID_TONE_MAPPED_IMAGE;
     }
+    for (uint32_t itemIndex = 0; itemIndex < data->meta->items.count; ++itemIndex) {
+        if (gainMapItem->id == data->meta->items.item[itemIndex].id) {
+            *gainMapItemIndex = itemIndex;
+            break;
+        }
+    }
+    assert(*gainMapItemIndex >= 0);
```
wantehchang, line 4393: Nit/optional: It is unfortunate that we have to search in the `data->meta->items` array to find the item index. We can consider adding a variant of the `avifMetaFindOrCreateItem` function that outputs an item index rather than an item pointer.

wantehchang (approved):
LGTM. I suggest some changes. I will take a look at this CL again.

src/read.c
```diff
@@ -4342,41 +4351,49 @@ static avifResult avifDecoderDataFindToneMappedImageItem(const avifDecoderData *
     return AVIF_RESULT_OK;
 }
 
-// Finds a 'tmap' (tone mapped image item) box associated with the given 'colorItem',
+// Finds a 'tmap' (tone mapped image item) box associated with the given colorItem at index 'colorItemIndex',
```
wantehchang, line 4354: Nit: the given colorItem => the color item

src/read.c
wantehchang: The answer to my question is Yes. Please see the comment at line 4429:

```
    // Only set the output parameters after everything has been validated.
    *toneMappedImageItem = toneMappedImageItemTmp;
    return AVIF_RESULT_OK;
```

So `*gainMapItemIndex` should receive the same treatment as `*gainMapItem` in the original code.

src/read.c

src/read.c

src/read.c

src/read.c

src/read.c

src/read.c
vigneshvg: Done.

src/read.c
vigneshvg: Done.

src/read.c

src/read.c
vigneshvg: Ack, i thought about this as well. But since this is the only use, i will leave this as-is.

src/read.c
```diff
@@ -4211,39 +4214,40 @@ static avifBool avifDecoderItemIsAlphaAux(avifDecoderItem * item, uint32_t color
     return auxCProp && isAlphaURN(auxCProp->u.auxC.auxType);
 }
 
-// Finds the alpha item whose parent item is colorItem and sets it in the alphaItem output parameter. Returns AVIF_RESULT_OK on
-// success. Note that *alphaItem can be NULL even if the return value is AVIF_RESULT_OK. If the colorItem is a grid and the alpha
-// item is represented as a set of auxl items to each color tile, then a fake item will be created and *isAlphaItemInInput will be
-// set to AVIF_FALSE. In this case, the alpha item merely exists to hold the locations of the alpha tile items. The data of this
-// item need not be read and the pixi property cannot be validated. Otherwise, *isAlphaItemInInput will be set to AVIF_TRUE when
-// *alphaItem is not NULL.
+// Finds the alpha item whose parent item is the color item and sets its index in the alphaItemId output parameter. Returns
```
wantehchang, line 4217: alphaItemId => alphaItemIndex

src/read.c
```diff
@@ -4292,7 +4301,7 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
 }
 
 #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
-// Finds a 'tmap' (tone mapped image item) box associated with the given 'colorItem'.
+// Finds a 'tmap' (tone mapped image item) box associated with the given color item.
```
wantehchang, line 4304: Optional: Undo this change (i.e., change `color item` back to `'colorItem'`) ?

src/read.c
wantehchang: The change I suggested was not made.

src/read.c
```diff
@@ -4712,6 +4730,14 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         }
 #endif // AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
 
+        // When searching for alpha and gainmap items, it is possible that the data->meta->items array is re-sized in calls to
+        // avifMetaFindOrCreateItem. So it is not safe to store pointers into data->meta->items array until all the items are
```
wantehchang, line 4734: Nit: add "the" before "data->meta->items array"

I am sorry I misunderstood what "that array" referred to. You can change this back to "that array".

wantehchang (approved):
LGTM. Just some comment nits.

src/read.c
vigneshvg: Done.

src/read.c
vigneshvg: Done.

src/read.c
vigneshvg: sorry i made it in the wrong place. fixed it now.

src/read.c
vigneshvg: I added `the` and left it as data->meta->items.

PR #1758, opened 2023-11-14
Set some dependency options to be multi-choice.

vrabaud:
The current options are complex:
- you want a local AOM, you need to set to variables (AVIF_CODEC_AOM and AVIF_LOCAL_AOM to ON)
- they can be confusing: some work in best effort, e.g. if you do not set AVIF_LOCAL_LIBYUV, it tries to find libyuv if not, oh well, too bad let's move on.
- you cannot disable some options: if libyuv is found, use it! Oh wait, I actually never asked for it (causing issues like https://github.com/AOMediaCodec/libavif/issues/1732)

That's why I propose to set dependency options to: OFF: it will never be used whether it is findable or in ext LOCAL: it has to use the version in ext/ and will fail if not found SYSTEM: it will find_package it and fail if not found

An AUTO version (that uses a dependency if found) should not be added as it will create more confusion for developers or package building.

This is compatible with `ccmake` or `cmake-gui`.


vrabaud:
@fdintino , @wantehchang , what do you think?

fdintino:
I like it. I've also seen projects accept paths for these options, to force find_package to look in a specific prefix.

CMakeLists.txt
```diff
@@ -234,11 +236,11 @@ if(libyuv_FOUND)
     set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ${LIBYUV_LIBRARY})
 endif(libyuv_FOUND)
 
-set_local_system_option("LIBSHARPYUV" "OFF" "Use libsharpyuv.")
+set_local_or_system_option("LIBSHARPYUV" "OFF" "Use libsharpyuv.")
 if(AVIF_LIBSHARPYUV STREQUAL "LOCAL")
     set(LIB_FILENAME "${CMAKE_CURRENT_SOURCE_DIR}/ext/libwebp/build/libsharpyuv${CMAKE_STATIC_LIBRARY_SUFFIX}")
     if(NOT EXISTS "${LIB_FILENAME}")
-        message(FATAL_ERROR "libavif(AVIF_LOCAL_LIBSHARPYUV): ${LIB_FILENAME} is missing, bailing out")
+        message(FATAL_ERROR "libavif(AVIF_LIBSHARPYUV): ${LIB_FILENAME} is missing, bailing out")
```
wantehchang, line 243: `AVIF_LIBSHARPYUV=LOCAL` ?

CMakeLists.txt
```diff
@@ -197,7 +199,7 @@ if(AVIF_LIBYUV STREQUAL "LOCAL")
     set(LIB_FILENAME "${AVIF_LIBYUV_BUILD_DIR}/${AVIF_LIBRARY_PREFIX}yuv${CMAKE_STATIC_LIBRARY_SUFFIX}")
 
     if(NOT EXISTS "${LIB_FILENAME}")
-        message(FATAL_ERROR "libavif(AVIF_LOCAL_LIBYUV): ${LIB_FILENAME} is missing, bailing out")
+        message(FATAL_ERROR "libavif(AVIF_LIBYUV): ${LIB_FILENAME} is missing, bailing out")
```
wantehchang, line 202: `AVIF_LIBYUV=LOCAL` ?

wantehchang (approved):
I am sorry I missed the review request.

Could you describe this change in the changelog?

wantehchang:
Vincent: In addition to the changelog, please also update https://github.com/AOMediaCodec/libavif/blob/main/README.md.

I think this is a good simplification of the cmake options. Thanks.

CMakeLists.txt
```diff
@@ -40,40 +40,55 @@ option(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
...
+    elseif(DEFINED AVIF_${VAR})
+        set(DEFAULT ${AVIF_${VAR}})
+    endif()
+    set(AVIF_${VAR} ${DEFAULT} CACHE STRING ${TEXT} FORCE)
+    set_property(CACHE AVIF_${VAR} PROPERTY STRINGS OFF LOCAL SYSTEM)
+endmacro()
+macro(set_codec_option CODEC NAME ENCDEC EXTRA)
+    # Deal with the older way of setting options.
+    if(DEFINED AVIF_CODEC_${CODEC})
+        set(AOM_CODEC_DEFAULT ${AVIF_CODEC_${CODEC}})
```
fdintino, line 56: Since this is used by more than aom, I'm guessing that this variable name is a mistake or got held over from an earlier version of this macro?

README.md
```diff
@@ -51,8 +52,9 @@ list of requirements.
 
 If you want to statically link any codec into your local (static) build of
 libavif, building using one of these scripts and then enabling the associated
-`AVIF_LOCAL_*` is a convenient method, but you must make sure to disable
-`BUILD_SHARED_LIBS` in CMake to instruct it to make a static libavif library.
+`AVIF_LOCAL_*` and setting `AVIF_CODEC_*` to `LOCAL` is a convenient method,
```
wantehchang, line 55: The new text mentions both the old and new ways, connected by "and". I think we just need to mention the new way ("setting `AVIF_CODEC_*` to `LOCAL`"). If we mention both, they should be connected by "or" because one of them is sufficient.

wantehchang (commented):
LGTM.

README.md
vrabaud: Thx, fixed in https://github.com/AOMediaCodec/libavif/pull/1795 I believe

PR #1763, opened 2023-11-15
Do not store potentially invalid pointers

vigneshvg:
Calling avifMetaFindItem() could invalidate all the existing item pointers that are being stored by the caller (since the function could resize the item array).

This patch fixes avifDecoderReset by updating the locally stored pointer after call to avifMetaFindItem.

This is a manual cherry-pick of commit `2041109` (PR #1757) on to the `v1.0.x` branch.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -769,6 +769,8 @@ static void avifMetaDestroy(avifMeta * meta)
     avifFree(meta);
 }
 
+// CAUTION: This function could potentially resize the meta->items array thereby invalidating all existing pointers that are being
+// stored locally. So if this function is being called, exercise caution in the caller to not use invalid pointers.
 static avifDecoderItem * avifMetaFindItem(avifMeta * meta, uint32_t itemID)
 {
     if (itemID == 0) {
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -3614,17 +3616,20 @@ static avifBool avifDecoderItemIsAlphaAux(avifDecoderItem * item, uint32_t color
     return auxCProp && isAlphaURN(auxCProp->u.auxC.auxType);
 }
 
-// Finds the alpha item whose parent item is colorItem and sets it in the alphaItem output parameter. Returns AVIF_RESULT_OK on
-// success. Note that *alphaItem can be NULL even if the return value is AVIF_RESULT_OK. If the colorItem is a grid and the alpha
+// Finds the alpha item whose parent item is color item and sets it in the alphaItem output parameter. Returns AVIF_RESULT_OK on
+// success. Note that *alphaItem can be NULL even if the return value is AVIF_RESULT_OK. If the color item is a grid and the alpha
```
wantehchang, line 3620: Nit: It seems clearer to replace `color item` by `*colorItemPtr` in these two lines (3619-3620).

Also would be good to clarify that `*colorItemPtr` is an in/out parameter.

src/read.c
```diff
@@ -3614,17 +3616,20 @@ static avifBool avifDecoderItemIsAlphaAux(avifDecoderItem * item, uint32_t color
...
 // item is represented as a set of auxl items to each color tile, then a fake item will be created and *isAlphaItemInInput will be
 // set to AVIF_FALSE. In this case, the alpha item merely exists to hold the locations of the alpha tile items. The data of this
 // item need not be read and the pixi property cannot be validated. Otherwise, *isAlphaItemInInput will be set to AVIF_TRUE when
-// *alphaItem is not NULL.
+// *alphaItem is not NULL. If the data->meta->items array is resized, then the value in *colorItemPtr could be invalid. This
```
wantehchang, line 3624: Nit: be invalid => "become invalid" or "be invalidated"

src/read.c
```diff
@@ -3614,17 +3616,20 @@ static avifBool avifDecoderItemIsAlphaAux(avifDecoderItem * item, uint32_t color
...
+// function also resets *colorItemPtr to the right value if an alpha item was found and added the data->meta->items array.
```
wantehchang, line 3625: Nit: added => added to ?

src/read.c
```diff
@@ -3907,6 +3916,14 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
             return AVIF_RESULT_MISSING_IMAGE_ITEM;
         }
+        int colorItemIndex = -1;
+        for (uint32_t i = 0; i < data->meta->items.count; ++i) {
+            if (colorItem->id == data->meta->items.item[i].id) {
+                colorItemIndex = i;
+                break;
+            }
+        }
+        assert(colorItemIndex >= 0);
```
wantehchang, line 3926: I suggest we move this block of code into `avifDecoderDataFindAlphaItem` so that we don't need to add a new `colorItemIndex` parameter to `avifDecoderDataFindAlphaItem`.

wantehchang (commented):
Thanks for the pull request. I believe the code is correct. I suggest some comment nit fixes and a tweak that hopefully simplifies the code a little bit.

src/read.c
vigneshvg: Done.

I think the newly added sentence a the end of the paragraph makes it obvious that this is an in/out param.

src/read.c
vigneshvg: DOne.

src/read.c
vigneshvg: Done.

src/read.c
vigneshvg: Done.

vigneshvg: Done.

src/read.c
```diff
@@ -3614,17 +3616,20 @@ static avifBool avifDecoderItemIsAlphaAux(avifDecoderItem * item, uint32_t color
...
+// on success. Note that *alphaItem can be NULL even if the return value is AVIF_RESULT_OK. If the *colorItemPtr is a grid and the
+// alpha item is represented as a set of auxl items to each color tile, then a fake item will be created and *isAlphaItemInInput
+// will be set to AVIF_FALSE. In this case, the alpha item merely exists to hold the locations of the alpha tile items. The data
+// of this item need not be read and the pixi property cannot be validated. Otherwise, *isAlphaItemInInput will be set to
+// AVIF_TRUE when *alphaItem is not NULL. If the data->meta->items array is resized, then the value in *colorItemPtr could become
+// invalid. This function also resets *colorItemPtr to the right value if an alpha item was found and added to the data->meta->items
+// array.
 static avifResult avifDecoderDataFindAlphaItem(avifDecoderData * data,
-                                               avifDecoderItem * colorItem,
+                                               avifDecoderItem ** colorItemPtr,
```
wantehchang, line 3628: [Just a comment, no change requested] The names of these two parameters, which have the same type, are now inconsistent; one has `Ptr`, the other doesn't. This inconsistency comes from the original code, where it is the types that are inconsistent. I understand that the new code needs to add `Ptr` to the parameter so that we can use the `colorItem` name as a local variable.

wantehchang (approved):
LGTM. Thanks!

src/read.c
vigneshvg: yeah, i agree. but this is the simplest way to do it without adding more diff noise to this PR.

PR #1764, opened 2023-11-16
Remove potential out of bound access to alphaItemIndices

vigneshvg:
It is possible to craft a file that has more alpha auxiliary items than color items and trigger an out of bound access into alphaItemIndices in the for loop.

Fix is to ensure that each color grid item has exactly one alpha grid item. Also, ensure that there are exactly the same number of color grids as informed in the grid config before trying to find the alpha item.

Also, update a diagnostic error message to cover all cases (i.e.) there can be more grids than necessary as well.

This is a manual cherry-pick of commit `6d62963` (PR #1756) on to v1.0.x branch.


wantehchang (approved):
LGTM. Thanks!

PR #1768, opened 2023-11-16
Check for NULL input in avifIOMemoryReaderRead()

y-guyon:
The check does not hurt and prevents ubsan errors in fuzzed targets with fuzzing engines that allow NULL to be passed as the first argument to LLVMFuzzerTestOneInput().

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -35,6 +35,11 @@ static avifResult avifIOMemoryReaderRead(struct avifIO * io, uint32_t readFlags,
     avifIOMemoryReader * reader = (avifIOMemoryReader *)io;
 
     // Sanitize/clamp incoming request
+    if (reader->rodata.data == NULL) {
+        // NULL input is definitely an issue. Also prevent the offset addition below to
+        // trigger an undefined behavior sanitizer error (happens even with offset zero).
+        return AVIF_RESULT_IO_ERROR;
+    }
     if (offset > reader->rodata.size) {
         // The offset is past the end of the buffer.
         return AVIF_RESULT_IO_ERROR;


src/io.c
```diff
@@ -35,6 +35,11 @@ static avifResult avifIOMemoryReaderRead(struct avifIO * io, uint32_t readFlags,
     avifIOMemoryReader * reader = (avifIOMemoryReader *)io;
 
     // Sanitize/clamp incoming request
+    if (reader->rodata.data == NULL) {
+        // NULL input is definitely an issue. Also prevent the offset addition below to
+        // trigger an undefined behavior sanitizer error (happens even with offset zero).
+        return AVIF_RESULT_IO_ERROR;
+    }
```
wantehchang, line 42: We need to follow the specification of `avifIOReadFunc` here:

```
// This function should return a block of memory that *must* remain valid until another read call to
// this avifIO struct is made (reusing a read buffer is acceptable/expected).
//
// * If offset exceeds the size of the content (past EOF), return AVIF_RESULT_IO_ERROR.
// * If offset is *exactly* at EOF, provide a 0-byte buffer and return AVIF_RESULT_OK.
// * If (offset+size) exceeds the contents' size, it must truncate the range to provide all
//   bytes from the offset to EOF.
// * If the range is unavailable yet (due to network conditions or any other reason),
//   return AVIF_RESULT_WAITING_ON_IO.
// * Otherwise, provide the range and return AVIF_RESULT_OK.
typedef avifResult (*avifIOReadFunc)(struct avifIO * io, uint32_t readFlags, uint64_t offset, size_t size, avifROData * out);
```

If `reader->rodata.data` is `NULL`, `reader->rodata.size` is 0, and `offset` is 0, it seems more correct to follow the second bullet point rather than the first bullet point:

```
// * If offset is *exactly* at EOF, provide a 0-byte buffer and return AVIF_RESULT_OK.
```

wantehchang: The issue here is whether `{ NULL, 0 }` is a valid representation of a 0-byte memory buffer. I think it is. But if we decide to treat `{ NULL, 0 }` as an invalid memory buffer, or more generally treat 0-byte memory buffers as invalid, then it is better to handle that early in `avifIOMemoryReaderCreate()`, rather than in `avifIOMemoryReaderRead()`.

y-guyon: Done at https://github.com/AOMediaCodec/libavif/pull/1779.

---

I thought about `avifIOCreateMemoryReader()` as well but this is the current definition:

```c
// Returns NULL if the reader cannot be allocated.
AVIF_API avifIO * avifIOCreateMemoryReader(const uint8_t * data, size_t size);
```

The signature would need to be changed to return an `avifResult` and this is a breaking change.

wantehchang: I see. The existing Create functions cannot return error information. But they can still check for invalid input parameters, for example,

```
AVIF_NODISCARD AVIF_API avifImage * avifImageCreate(uint32_t width, uint32_t height, uint32_t depth, avifPixelFormat yuvFormat);
```

If we can handle a 0-byte input buffer properly, then it is better to not treat it as invalid input.

Issue #1769, opened 2023-11-16
Random "ILLEGAL" errors on CI runs on macos


maryla-uc:
A small percentage of macOS github runners don't have avx2 support but advertise that they do, causing random "Illegal instruction: 4" errors in AOM tests.

See discussion here https://github.com/AOMediaCodec/libavif/pull/1744#discussion_r1388819364

jzern:
> One small correction to the PR description: it looks as though the runners are actually advertising their SIMD support correctly, and AOM is using the correct SIMD optimizations for the hardware (I was mistaken about that). The -mavx2 compile flag is added to aom regardless of the host machine so long as the compiler supports it.

This flag should only be used with files containing avx2 code. The availability of avx2 is detected at runtime by libaom. Do you see -mavx2 being universally added or is the round() crash in a libaom avx2 module?

fdintino:
This is the line that usually triggers the illegal instruction error:

https://aomedia.googlesource.com/aom/+/d0f3147e5242edf52b896a65b23af47006e9e84f/av1/encoder/ratectrl.c#2506

```c
rc->avg_frame_bandwidth =
  (int)round(oxcf->rc_cfg.target_bandwidth / cpi->framerate);
````

Its pretty prosaic, not part of a libaom avx2 module. Just incidentally because the round function has an avx2-optimized version in glibc.

To drive home the fact that aom isn't really at fault , here is aminimal script that will cause the illegal instruction error on one of these macos runners:

```c
#include <math.h>
#include <stdio.h>

int main(int argc, char * argv[]) {
  int x = 2, y = 1;
  printf("%u\n", (int)round(x / y));
}
```

compile that with `clang -o out -O0 -g -mavx2 src.c` , then run lldb with the commands:

```lldb
lldb --file ~/out
(lldb) disassemble -n main
```

And you'll see a bunch of AVX2-specific instuctions.

The `HAVE_AVX2` flag in AOM, which controls the addition of the `-mavx2` CFLAG, doesn't check the host CPU. It just checks whether the compiler supports the flag. And clang supports that flag regardless of the host's capabilities.

wantehchang:
In [my experiment](https://github.com/AOMediaCodec/libavif/pull/1783), I saw no evidence of libaom's non-avx2 source files being compiled with -mavx2. Here is the compiler command line for ratectrl.c:

[...]

PR #1770, opened 2023-11-16
Disable AVX2 on CI workflows that run on macos.

maryla-uc:
Workaround for #1769


wantehchang:
This has been replaced by https://github.com/AOMediaCodec/libavif/pull/1784.

PR #1772, opened 2023-11-16
fix: android jni unused-parameter

fdintino:
The more stringent error and warning compiler flags in the root CMakeLists.txt are set with `add_compile_options` and so, as directory properties, they don't get applied to the avif_android target. If `-Wall -Werror` or `/W4 /WX` are enabled for avif_android, the compiler fails on `-Werror,-Wunused-parameter`. This issue is fixed by defining a macro, `IGNORE_UNUSED_JNIPARAMETERS`, that is called at the top of any functions that don't reference the `env` or `thiz` parameters.

Note: this file appears to not have had clang-format run on it, so I didn't do any reformatting.

### DIFF
--- a/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
+++ b/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
@@ -24,7 +24,7 @@
   JNIEXPORT RETURN_TYPE Java_org_aomedia_avif_android_AvifDecoder_##NAME( \
       JNIEnv* env, jobject thiz, ##__VA_ARGS__)
 
-#define IGNORE_UNUSED_JNIPARAMETERS \
+#define IGNORE_UNUSED_JNI_PARAMETERS \
   (void) env; \
   (void) thiz
 
@@ -237,7 +237,7 @@ jint JNI_OnLoad(JavaVM* vm, void* /*reserved*/) {
 }
 
 FUNC(jboolean, isAvifImage, jobject encoded, int length) {
-  IGNORE_UNUSED_JNIPARAMETERS;
+  IGNORE_UNUSED_JNI_PARAMETERS;
   const uint8_t* const buffer =
       static_cast<const uint8_t*>(env->GetDirectBufferAddress(encoded));
   const avifROData avif = {buffer, static_cast<size_t>(length)};
@@ -260,7 +260,7 @@ FUNC(jboolean, isAvifImage, jobject encoded, int length) {
   if (var == nullptr) return ret
 
 FUNC(jboolean, getInfo, jobject encoded, int length, jobject info) {
-  IGNORE_UNUSED_JNIPARAMETERS;
+  IGNORE_UNUSED_JNI_PARAMETERS;
   const uint8_t* const buffer =
       static_cast<const uint8_t*>(env->GetDirectBufferAddress(encoded));
   AvifDecoderWrapper decoder;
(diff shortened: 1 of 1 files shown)


android_jni/avifandroidjni/src/main/jni/CMakeLists.txt
```diff
@@ -35,6 +35,8 @@ add_subdirectory(../../../../.. build)
 
 add_library("avif_android" SHARED "libavif_jni.cc")
 
+target_compile_options(avif_android PRIVATE $<TARGET_PROPERTY:avif,COMPILE_OPTIONS>)
```
wantehchang, line 38: Vignesh: I didn't review the change to this file. The changes to libavif_jni.cc seem like a good idea.

vigneshvg:
Thank you, i will take a look at this PR.

android_jni/avifandroidjni/src/main/jni/CMakeLists.txt
fdintino: This change isn't actually necessary and I'd be fine with removing it. It makes it so that this warning causes errors on build, but that's not strictly necessary. If something like #1778 gets merged then this would be redundant.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -19,10 +19,10 @@
 #define FUNC(RETURN_TYPE, NAME, ...)                                      \
   extern "C" {                                                            \
   JNIEXPORT RETURN_TYPE Java_org_aomedia_avif_android_AvifDecoder_##NAME( \
-      JNIEnv* env, jobject thiz, ##__VA_ARGS__);                          \
+      __VA_ARGS__);                                                       \
```
vigneshvg, line 22: hmm, i don't like the idea of repeating the jni env parameters in all the functions. it pushes the actual argument list to the back and it's a mix of named and unnamed parameters.

how about this instead: keep this macro as is. add a new macro like so:
```
#define IGNORE_UNUSED_JNIPARAMETERS \
  (void) env; \
  (void) thiz
```

and add this as the first line to each of the functions? this way it is uniform how we handle it in all the functions (it doesn't really matter if we use these variables in these functinos or not since we know what we are ignoring here).

vigneshvg (changes requested):

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
fdintino: It was between that approach and the one I have here, and I didn't have strong opinion either way. I'd be fine with adding an ignore parameter macro

fdintino: Done.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -24,6 +24,10 @@
   JNIEXPORT RETURN_TYPE Java_org_aomedia_avif_android_AvifDecoder_##NAME( \
       JNIEnv* env, jobject thiz, ##__VA_ARGS__)
 
+#define IGNORE_UNUSED_JNIPARAMETERS \
```
vigneshvg, line 27: nit: `IGNORE_UNUSED_JNI_PARAMETERS` (JNI and PARAMETERS are to be separated with _).

vigneshvg (approved):
thank you! lgtm with one minor comment.

vigneshvg (approved):

vigneshvg:
thank you for the PR!

PR #1779, opened 2023-11-17
Return OK in IOMemRead() if rodata.data is NULL

y-guyon:


maryla-uc (approved):

wantehchang (approved):
LGTM. Thanks.

src/io.c
```diff
@@ -49,7 +44,9 @@ static avifResult avifIOMemoryReaderRead(struct avifIO * io, uint32_t readFlags,
         size = (size_t)availableSize;
     }
 
-    out->data = reader->rodata.data + offset;
+    // Prevent the offset addition to trigger an undefined behavior
```
wantehchang, line 47: Nit: "from triggering" sounds better.

PR #1782, opened 2023-11-18
Add a decoder test that parses empty data

wantehchang:
Add a test for the following pull requests:
https://github.com/AOMediaCodec/libavif/pull/1768
https://github.com/AOMediaCodec/libavif/pull/1779

### DIFF
--- a/tests/gtest/avifdecodetest.cc
+++ b/tests/gtest/avifdecodetest.cc
@@ -1,6 +1,9 @@
 // Copyright 2023 Google LLC
 // SPDX-License-Identifier: BSD-2-Clause
 
+#include <iostream>
+#include <string>
+
 #include "avif/avif.h"
 #include "aviftest_helpers.h"
 #include "gtest/gtest.h"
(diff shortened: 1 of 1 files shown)


y-guyon (approved):
Thanks for adding tests!

tests/gtest/avifdecodetest.cc
```diff
@@ -30,6 +33,14 @@ TEST(AvifDecodeTest, ColorGridAlphaNoGrid) {
   EXPECT_GT(decoder->image->alphaRowBytes, 0u);
 }
 
+TEST(AvifDecodeTest, ParseEmptyData) {
+  DecoderPtr decoder(avifDecoderCreate());
+  ASSERT_NE(decoder, nullptr);
+  ASSERT_EQ(avifDecoderSetIOMemory(decoder.get(), nullptr, 0), AVIF_RESULT_OK);
+  // No ftyp box was seen.
+  ASSERT_EQ(avifDecoderParse(decoder.get()), AVIF_RESULT_INVALID_FTYP);
```
wantehchang, line 41: The test only allows the error code returned by the current implementation here, but we may want to give the implementation some flexibility to return any other reasonable error code in the future.

I guess what's important is that `avifDecoderParse()` should fail (the error code should not be `AVIF_RESULT_OK`). In addition, it would be good to check that the `avifIO::read()` call succeeds (`avifIO::read()` should not return `AVIF_RESULT_IO_ERROR`), but I don't know if we can reliably determine that from the return value of `avifDecoderParse()`.

PR #1784, opened 2023-11-18
svt.sh: Pass disable-native to build.sh

wantehchang:
Disable the use of -march=native.

Hopefully this will fix
https://github.com/AOMediaCodec/libavif/issues/1769.


wantehchang:
From the help message of the build.sh script in SVT-AV1:

```
--enable-*/--disable-* options shown are not the default
For each enable-*, there is a disable-* option, and vice versa.

...
    --disable-native,   Disable the use of -march=native
    disable-native
```

Since the --disable-native option is shown, it is not the default.

y-guyon (approved):

PR #1793, opened 2023-11-21
Fix avifgainmaputil trying to create images with pixel format 0.

maryla-uc:


vrabaud (approved):
Not sure you need a test for that.

maryla-uc:
So the thing is, there is a test, but it's not currently run in the CI, I will fix that in another PR (it was kind of on my todo list but it slipped my mind).

PR #1801, opened 2023-11-23
Make the gainMap member of avifImage a pointer.

maryla-uc:
Allows adding members in the future without breaking the ABI.

### DIFF
--- a/apps/avifgainmaputil/printmetadata_command.cc
+++ b/apps/avifgainmaputil/printmetadata_command.cc
@@ -3,6 +3,7 @@
 
 #include "printmetadata_command.h"
 
+#include <cassert>
 #include <iomanip>
 
 #include "avif/avif_cxx.h"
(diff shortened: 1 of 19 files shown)


src/write.c
```diff
@@ -1208,8 +1208,10 @@ static avifResult avifValidateGrid(uint32_t gridCols,
     const avifImage * bottomRightCell = cellImages[cellCount - 1];
 #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
     if (validateGainMap) {
-        firstCell = firstCell->gainMap.image;
-        bottomRightCell = bottomRightCell->gainMap.image;
+        assert(firstCell->gainMap && firstCell->gainMap->image);
```
vrabaud, line 1211: Just wondering if that can never happen and whether  AVIF_CHECKRES should be used

vrabaud (approved):
Just wondering if there should be function that checks for gainMap and gainMap->image and returns NULL or the image to avoid doing the double condition every time. Then again, that would be more to maintain so no.

apps/avifgainmaputil/printmetadata_command.cc
```diff
@@ -57,13 +57,13 @@ avifResult PrintMetadataCommand::Run() {
               << decoder->diag.error << ")\n";
     return result;
   }
-  if (!decoder->gainMapPresent) {
+  if (!decoder->gainMapPresent || decoder->image->gainMap == nullptr) {
```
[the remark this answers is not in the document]
maryla-uc: In theory no. Changed to an assert.

src/write.c
maryla-uc: No the caller should only call the function with validateGainMap if there is a gain map. I think defensive programming is not useful in this instance.

Issue #1755, opened 2023-11-13
Building in MSYS2/MinGW with GCC 13.2 fails: undefined reference to `WinMain@16'


LigH-de:
Compiling libavif project in [media-autobuild suite](https://github.com/m-ab-s/media-autobuild_suite), a set of scripts for a MSYS2/MinGW build environment which keeps itself up-to-date before compiling ffmpeg including a lot of codecs.

```
CPPFLAGS: 
CFLAGS: -D_FORTIFY_SOURCE=2 -fstack-protector-strong -mtune=generic -O2 -pipe -D__USE_MINGW_ANSI_STDIO=1 -mthreads
CXXFLAGS: -D_FORTIFY_SOURCE=2 -fstack-protector-strong -mtune=generic -O2 -pipe -D__USE_MINGW_ANSI_STDIO=1
LDFLAGS: -D_FORTIFY_SOURCE=2 -fstack-protector-strong -mtune=generic -O2 -pipe -D__USE_MINGW_ANSI_STDIO=1 -static-libgcc -static-libstdc++
ninja -j4 
{...}
[37/38] Linking CXX executable avifdec.exe
FAILED: avifdec.exe 
[...]

PR #1802, opened 2023-11-23
Add -municode for MINGW

vrabaud:
Fixes https://github.com/AOMediaCodec/libavif/issues/1755


y-guyon (approved):

CMakeLists.txt
```diff
@@ -299,7 +299,10 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang")
 elseif(CMAKE_C_COMPILER_ID MATCHES "GNU")
     message(STATUS "libavif: Enabling warnings for GCC")
     add_compile_options(-Wall -Wextra)
-elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")
+elseif(MINGW)
+    add_compile_options(-municode)
+    add_compile_definitions(UNICODE _UNICODE)
```
wantehchang, line 304: Delete this line.

Alternatively, add this line for all compilers on Windows. I suggested this before based on [GCC documentation of -municode](https://gcc.gnu.org/onlinedocs/gcc/x86-Windows-Options.html#index-municode):

> `-municode`
> 
> This option is available for MinGW-w64 targets. It causes the `UNICODE` preprocessor macro to be predefined, and chooses Unicode-capable runtime startup code.

The rationale is that if GCC defines the `UNICODE` preprocessor macro (when `-municode` is specified), then we should define that macro for other compilers on Windows, otherwise MINGW and other compilers will differ in this aspect.

But I now think we need to verify exactly what preprocessor macros `-municode` defines before we do this. So I suggest we just delete this line for now.

wantehchang (approved):
LGTM. I suggest a change.

CMakeLists.txt
```diff
@@ -299,7 +299,10 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang")
 elseif(CMAKE_C_COMPILER_ID MATCHES "GNU")
     message(STATUS "libavif: Enabling warnings for GCC")
     add_compile_options(-Wall -Wextra)
-elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")
+elseif(MINGW)
```
wantehchang, line 302: We should test this. I remember MINGW is GCC, so I am worried that it will match the GCC check at line 299 instead.

vrabaud: Definitely, let me add a MINGW workflow from https://github.com/msys2/setup-msys2

wantehchang: If MINGW is identified as GCC, we will need to handle it as follows:

```
elseif(CMAKE_C_COMPILER_ID MATCHES "GNU")
    message(STATUS "libavif: Enabling warnings for GCC")
    add_compile_options(-Wall -Wextra)
    if(MINGW)
       add_compile_options(-municode)
    endif()
elseif(MSVC)
```

wantehchang (changes requested):

CMakeLists.txt
wantehchang: On the Web I found at least two pages that say `-municode` can be used as a linker option. Here is a patch file to do that:

```
index 12e0a10..6b05fb7 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -802,11 +802,17 @@ if(AVIF_BUILD_APPS)
     if(AVIF_USE_CXX)
         set_target_properties(avifenc PROPERTIES LINKER_LANGUAGE "CXX")
     endif()
+    if(MINGW)
+        target_link_options(avifenc PRIVATE -municode)
+    endif()
     target_link_libraries(avifenc avif_apps)
     add_executable(avifdec apps/avifdec.c)
     if(AVIF_USE_CXX)
         set_target_properties(avifdec PROPERTIES LINKER_LANGUAGE "CXX")
     endif()
+    if(MINGW)
+        target_link_options(avifdec PRIVATE -municode)
+    endif()
     target_link_libraries(avifdec avif_apps)

     if(NOT SKIP_INSTALL_APPS AND NOT SKIP_INSTALL_ALL)
diff --git a/tests/CMakeLists.txt b/tests/CMakeLists.txt
index d174e05..b8d4b10 100644
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -223,6 +223,9 @@ if(AVIF_BUILD_APPS)
     # When building apps, test the avifenc/avifdec.
     # 'are_images_equal' is used to make sure inputs/outputs are unchanged.
     add_executable(are_images_equal gtest/are_images_equal.cc)
+    if(MINGW)
+        target_link_options(are_images_equal PRIVATE -municode)
+    endif()
     target_link_libraries(are_images_equal aviftest_helpers)
     add_test(NAME test_cmd COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/test_cmd.sh ${CMAKE_BINARY_DIR}
                                    ${CMAKE_CURRENT_SOURCE_DIR}/data

```

wantehchang: Some webpages also say if `wmain()` is defined in a C++ file, it needs to be declared as `extern "C"` for the g++ in MinGW. In my testing I don't need to make that change to are_images_equal.cc.

kmilos: FYI, MINGW is defined for both GCC and Clang. See

https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Windows-GNU.cmake
https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Windows-Clang.cmake

In the CI, you can test both [UCRT64 and CLANG64 environments](https://www.msys2.org/docs/environments/).

.github/workflows/ci-mingw.yml
```diff
@@ -0,0 +1,88 @@
...
+      with:
+        update: true
+        install: >-
+          aom
+          cmake
+          dav1d
+          gcc
+          git
+          libyuv
+          ninja
```
kmilos, line 38: To install packages for the appropriate environment, it's best to use the "pacboy" command install of "install", and and ":p" to the package name: https://github.com/msys2/setup-msys2#pacboy

This won't work as is (w/o the prefix), as it'll try to install packages in the base MSYS environment, and they don't exit there...

.github/workflows/ci-mingw.yml
```diff
@@ -0,0 +1,89 @@
...
+      uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
+      with:
+        path: ext
+        key: ${{ runner.sys }}-mingw
+    - name: Print cmake version
+      run: cmake --version
+    - name: Build GoogleTest
+      if: steps.cache-ext.outputs.cache-hit != 'true'
+      working-directory: ./ext
+      run: bash -e googletest.cmd
```
kmilos, line 51: Just in case you want to make the CI faster, MSYS2 ships [googletest](https://packages.msys2.org/base/mingw-w64-googletest) already...

.github/workflows/ci-mingw.yml
```diff
@@ -0,0 +1,89 @@
+# This is a copy of ci-unix-shared-local.yml for building shared libraries
+# with an additional build configuration (using installed deps and dav1d).
+
+name: CI MINGW
```
kmilos, line 4: "MinGW" 😉

.github/workflows/ci-mingw.yml
```diff
@@ -0,0 +1,90 @@
...
+    steps:
+    - uses: msys2/setup-msys2@v2
+      with:
+        msystem: ${{matrix.sys}}
+        update: true
+        install: >-
+          cmake
+          gcc
+          git
+          ninja
```
kmilos, line 36: It really is "git" only here. The rest belong to their respective environments (so in the "pacboy" option below).

kmilos: Also, instead of "gcc", do request just "cc:p" (covers both GCC and Clang via a metapackage).

.github/workflows/ci-mingw.yml
```diff
@@ -0,0 +1,90 @@
...
+        pacboy: >-
+          aom
+          dav1d
+          libyuv
```
kmilos, line 40: Missing ":p" suffix on each line.

vrabaud: Done

.github/workflows/ci-mingw.yml
```diff
@@ -0,0 +1,90 @@
...
+      fail-fast: false
+      matrix:
+        sys:
+          - mingw64
+          - mingw32
+          - ucrt64
+          - clang64
+
+    steps:
+    - uses: msys2/setup-msys2@v2
```
kmilos, line 28: Might be missing the default shell to "msys2" (which is bash in essence): https://github.com/msys2/setup-msys2#default-shell

.github/workflows/ci-mingw.yml
vrabaud: Done

.github/workflows/ci-mingw.yml
vrabaud: Right, but I thought (just like on Ubuntu), that it was better to compile gtest from source to make sure the same compile flags are passed?

.github/workflows/ci-mingw.yml
vrabaud: Done

vrabaud:
Ok, I added a working CI.
@kmilos , it is now failing for pure code reasons: `-municode` is set, `wmain` is used, but `setlocale(LC_ALL, ".UTF8")` does not work. Any way you can help me with that? Thx!

.github/workflows/ci-mingw.yml
vrabaud: Right, but I thought (just like on Ubuntu), that it was better to compile gtest from source to make sure the same compile flags are passed?

.github/workflows/ci-mingw.yml
vrabaud: Done, let's be right :)

kmilos: I guess it depends on whether libavif changes the compile options significantly from the ones cmake sets...

kmilos:
> `setlocale(LC_ALL, ".UTF8")` does not work

@vrabaud It seems to not fail in UCRT64 and CLANG64 environments (as they use the modern UCRT), but obviously fail in the legacy MINGW32/64 ones as they use MSVCRT where .utf8 locales are non-existent. **_In MINGW32/64 (i.e. MSVCRT), there's no way around not using wstring/wchat_t and W APIs for i18n._**

I guess what you need to decide/say is - libavif is dropping support for MSVCRT (MINGW32/64 and anything before Windows 10 1903 and VS 2015), and will have only UCRT builds w/ UTF-8, so default on VS 2015 and later, and UCRT64+CLANG64 (and possibly CLANG32 if you really want to cover x86).

vrabaud:
Thx @kmilos for the explanation. How about:
- I remove `-municode` for MINGW32/64
- I also check for `__MINGW32__` and `__MINGW64__` not being defined when defining INIT_ARGV with setlocale in apps/shared/avifutil.h?
- I disable the UTF8 test for MINGW32/64

This way, MINGW32/64 is supported, not just UTF8

kmilos:
> * I also check for `__MINGW32__` and `__MINGW64__` not being defined when defining INIT_ARGV with setlocale in apps/shared/avifutil.h?

These are defined for all MinGW environments (and it's redundant to check for `__MINGW64__` like it is redundant to check for `_WIN64` if you already check for `_WIN32`). They really just define the platform, not the actual toolchain details. I guess you really want to check for `_UCRT` as the runtime C library being used (by either MSVC or MinGW toolchains).

Good idea for the other two points!

vrabaud:
Indeed, the only thing I found so far (from https://discourse.cmake.org/t/how-to-check-that-the-current-compiler-is-gcc-msys-when-using-ninja-generator/6698/5) is to run `gcc -dumpmachine`.

kmilos:
> Indeed, the only thing I found so far (from https://discourse.cmake.org/t/how-to-check-that-the-current-compiler-is-gcc-msys-when-using-ninja-generator/6698/5) is to run `gcc -dumpmachine`.

Again, not the same thing. There's no problem telling if you're building for `MINGW` vs `MSYS`/`CYGWIN` (those all exist in CMake as variables already, so there's no need for such acrobatics). You can also tell GCC vs Clang w/ `CMAKE_{C,CXX}_COMPILER_ID`. But AFAIK you can't tell between MSVCRT and UCRT? You can only e.g. check the `MSYSTEM` environment variable perhaps...

Btw, `gcc -dumpmachine` in the UCRT64 environment returns the same as in MINGW64: `x86_64-w64-mingw32`

vrabaud:
Thx, @kmilos, that works ! Still I had to disable mingw32: some errors appeared that seem unrelated: https://github.com/vrabaud/libavif/actions/runs/6985029492/job/19008737494#step:3:79
mingw64 is fine though.

tests/CMakeLists.txt
```diff
@@ -220,13 +220,20 @@ endif()
...
     # When building apps, test the avifenc/avifdec.
     # 'are_images_equal' is used to make sure inputs/outputs are unchanged.
     add_executable(are_images_equal gtest/are_images_equal.cc)
     target_link_libraries(are_images_equal aviftest_helpers)
-    add_test(NAME test_cmd COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/test_cmd.sh ${CMAKE_BINARY_DIR}
-                                   ${CMAKE_CURRENT_SOURCE_DIR}/data
-    )
+    if(MINGW)
+        if("$ENV{MSYSTEM}" MATCHES "(MINGW32|MINGW64)")
+            set_tests_properties(test_cmd PROPERTIES ENVIRONMENT "AVIF_TEST_NO_UTF8=1")
```
wantehchang, line 232: Nit: Many people recommend against defining a boolean flag in the negative sense. To follow that recommendation, we will need define the `AVIF_TEST_UTF8` boolean flag, set it to 1 by default, and set it to 0 here.

wantehchang:
Vincent: I assume this pull request has stabilized. I will review it tomorrow (Thursday).

.github/workflows/ci-mingw.yml
```diff
@@ -0,0 +1,79 @@
...
+          libpng:p
+          libyuv:p
+          ninja:p
+          zlib:p
+    - name: Cache external dependencies
+      id: cache-ext
+      uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
+      with:
+        path: ext
+        key: ${{ matrix.sys }}-mingw-${{ hashFiles('ext/*.cmd') }}
```
wantehchang, line 54: Optional: We can just hash `ext/googletest.cmd` in the current setup. Then the cache file can still be used when aom.cmd, dav1d.cmd, etc. are updated.

But if we also build, say, libsharpyuv locally, this will be broken. So maybe this isn't a good idea.

.github/workflows/ci-mingw.yml
```diff
@@ -0,0 +1,79 @@
...
+    - name: Print cmake version
+      run: cmake --version
+    - name: Build GoogleTest
```
wantehchang, line 57: Optional: Would be good to build libsharpyuv as well. Can do this in a follow-up pull request.

wantehchang (approved):
LGTM. Thank you very much for all your work on this.

CMakeLists.txt
```diff
@@ -809,6 +809,10 @@ if(AVIF_BUILD_APPS)
         set_target_properties(avifdec PROPERTIES LINKER_LANGUAGE "CXX")
     endif()
     target_link_libraries(avifdec avif_apps)
+    if(MINGW AND NOT "$ENV{MSYSTEM}" MATCHES "(MINGW32|MINGW64)")
```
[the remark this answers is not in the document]
kmilos: > I guess MINGW32|MINGW64 don't have UCRT?

Correct, those are "legacy" environments linking against MSVCRT.

apps/shared/avifutil.h
```diff
@@ -11,17 +11,17 @@ extern "C" {
...
 #define MAIN() int main(int argc, char * argv[])
 #endif
 
-#if defined(_WIN32)
+#if defined(_WIN32) && defined(_UCRT)
 #ifdef __cplusplus
 #define INIT_ARGV()                                                                                           \
     if (setlocale(LC_ALL, ".UTF8") == NULL) {                                                                 \
-        fprintf(stderr, "setlocale failed\n");                                                                \
+        fprintf(stderr, "C++ setlocale failed\n");                                                            \
```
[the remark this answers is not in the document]
vrabaud: Done

apps/shared/avifutil.h
```diff
@@ -62,7 +62,7 @@ extern "C" {
 #else
 #define INIT_ARGV()
 #endif
-#if defined(_WIN32) && !defined(__cplusplus)
+#if defined(_WIN32) && !defined(__cplusplus) && defined(_UCRT)
```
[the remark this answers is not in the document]
vrabaud: Done

apps/shared/avifutil.h
```diff
@@ -11,17 +11,17 @@ extern "C" {
 #endif
 
 // MAIN(), INIT_ARGV(), FREE_ARGV() for UTF8-aware command line parsing.
-#if defined(_WIN32)
+#if defined(_WIN32) && defined(_UCRT)
```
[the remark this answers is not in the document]
vrabaud: Done

tests/CMakeLists.txt
```diff
@@ -220,13 +220,21 @@ endif()
...
     # When building apps, test the avifenc/avifdec.
     # 'are_images_equal' is used to make sure inputs/outputs are unchanged.
     add_executable(are_images_equal gtest/are_images_equal.cc)
     target_link_libraries(are_images_equal aviftest_helpers)
-    add_test(NAME test_cmd COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/test_cmd.sh ${CMAKE_BINARY_DIR}
-                                   ${CMAKE_CURRENT_SOURCE_DIR}/data
-    )
+    set_tests_properties(test_cmd PROPERTIES ENVIRONMENT "AVIF_TEST_UTF8=1")
+    if(MINGW)
+        if("$ENV{MSYSTEM}" MATCHES "(MINGW32|MINGW64)")
```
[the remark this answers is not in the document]
vrabaud: Done

CMakeLists.txt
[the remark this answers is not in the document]
vrabaud: Done

.github/workflows/ci-mingw.yml
vrabaud: Done

CMakeLists.txt
```diff
@@ -431,7 +431,7 @@ endif()
 # /usr/local/include.
 set(CMAKE_FIND_FRAMEWORK LAST)
 
-if(UNIX)
+if(UNIX OR MINGW)
```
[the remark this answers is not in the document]
vrabaud: Done in commit message.

PR #1808, opened 2023-11-27
Do not store colorproperties until alpha item is found

vigneshvg:
colorProperties could be pointing to a dangling pointer if findAlphaItem() resizes the meta.items array.


vigneshvg:
> We should add tests for these fixes.

Ack. I will do it in a separate PR (or come up with a fix for the underlying problem). I will go ahead and merge this PR since it fixes crbug.com/1504792.

PR #1812, opened 2023-11-28
Add metadata about the alternate image to the gainMap struct.

maryla-uc:
In avif files, write the following boxes:
- for the tmap derived item: colr (icc and/or cicp), pixi (optional), clli (optional)
- fot the gain map image: colr (cicp) where color_primaries and transfer_characteristics must be 2

Follows the latest draft specs.

### DIFF
--- a/apps/avifgainmaputil/swapbase_command.cc
+++ b/apps/avifgainmaputil/swapbase_command.cc
@@ -18,21 +18,25 @@ avifResult ChangeBase(avifImage& image, avifImage* swapped) {
       image.gainMap->metadata.alternateHdrHeadroomD;
   const bool tone_mapping_to_sdr = (headroom == 0.0f);
 
-  if (image.gainMap->altColorPrimaries != AVIF_COLOR_PRIMARIES_UNSPECIFIED ||
-      image.gainMap->altTransferCharacteristics !=
-          AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED) {
-    swapped->colorPrimaries = image.gainMap->altColorPrimaries;
-    swapped->transferCharacteristics =
-        image.gainMap->altTransferCharacteristics;
-    swapped->matrixCoefficients = image.gainMap->altMatrixCoefficients;
-  } else {
-    // If there is no cicp on the gain map, use the same values as the old base
-    // image, except for the transfer characteristics.
+  swapped->colorPrimaries = image.gainMap->altColorPrimaries;
+  if (swapped->colorPrimaries == AVIF_COLOR_PRIMARIES_UNSPECIFIED) {
+    // Default to the same primaries as the base image if unspecified.
     swapped->colorPrimaries = image.colorPrimaries;
+  }
+
+  swapped->transferCharacteristics = image.gainMap->altTransferCharacteristics;
+  if (swapped->transferCharacteristics ==
+      AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED) {
+    // Default to PQ for HDR and sRGB for SDR if unspecified.
     const avifTransferCharacteristics transfer_characteristics =
         tone_mapping_to_sdr ? AVIF_TRANSFER_CHARACTERISTICS_SRGB
                             : AVIF_TRANSFER_CHARACTERISTICS_SMPTE2084;
     swapped->transferCharacteristics = transfer_characteristics;
+  }
+
+  swapped->matrixCoefficients = image.gainMap->altMatrixCoefficients;
+  if (swapped->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED) {
+    // Default to the same matrix as the base image if unspecified.
     swapped->matrixCoefficients = image.matrixCoefficients;
   }
 
--- a/apps/avifgainmaputil/swapbase_command.cc
+++ b/apps/avifgainmaputil/swapbase_command.cc
@@ -18,20 +18,13 @@ avifResult ChangeBase(avifImage& image, avifImage* swapped) {
       image.gainMap->metadata.alternateHdrHeadroomD;
   const bool tone_mapping_to_sdr = (headroom == 0.0f);
 
-  // The gain map's cicp values are those of the 'tmap' item and describe the
-  // image obtained by fully applying the gain map. See ISO/IEC JTC 1/SC 29/WG 3
-  // m64379v1 4.1.1: A tmap derived item shall be associated with a colr item
-  // property. This property describes the colour properties of the
-  // reconstructed image if the gain map input item is fully applied according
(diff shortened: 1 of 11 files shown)


apps/avifgainmaputil/swapbase_command.cc
```diff
@@ -18,20 +18,13 @@ avifResult ChangeBase(avifImage& image, avifImage* swapped) {
...
-  // reconstructed image if the gain map input item is fully applied according
-  // to ISO/AWI 214961.
-  if (image.gainMap->image->transferCharacteristics !=
-          AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED ||
-      image.gainMap->image->colorPrimaries !=
-          AVIF_COLOR_PRIMARIES_UNSPECIFIED) {
-    swapped->colorPrimaries = image.gainMap->image->colorPrimaries;
+  if (image.gainMap->altColorPrimaries != AVIF_COLOR_PRIMARIES_UNSPECIFIED ||
+      image.gainMap->altTransferCharacteristics !=
+          AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED) {
```
y-guyon, line 23: `matrixCoefficients` look like they were forgotten from the condition but I guess not, maybe add a comment why?

apps/avifgainmaputil/tonemap_command.cc
```diff
@@ -90,66 +79,121 @@ avifResult TonemapCommand::Run() {
...
+    }
   }
+
+  // Determine output pixel format.
   avifPixelFormat pixel_format =
       (avifPixelFormat)arg_image_read_.pixel_format.value();
+  const avifPixelFormat alt_yuv_format =
+      (image->gainMap->altPlaneCount == 1)
+          ? AVIF_PIXEL_FORMAT_YUV400
+          : std::min(image->yuvFormat, image->gainMap->image->yuvFormat);
```
y-guyon, line 159: `std::min()` on an enum is weird but it works.  
Maybe add a comment `// Favor least subsampled format` to avoid having to look at the `avifPixelFormat` definition? Or is it better to leave it like this to avoid disconnection between `avifPixelFormat` order and the comment?

src/read.c
```diff
@@ -4363,6 +4363,54 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+    for (uint32_t propertyIndex = 0; propertyIndex < properties->count; ++propertyIndex) {
+        avifProperty * prop = &properties->prop[propertyIndex];
+
+        if (!memcmp(prop->type, "colr", 4)) {
+            if (prop->u.colr.hasICC) {
+                if (colrICCSeen) {
+                    return AVIF_RESULT_BMFF_PARSE_FAILED;
+                }
+                avifROData iccRead;
+                const avifResult readResult = io->read(io, 0, prop->u.colr.iccOffset, prop->u.colr.iccSize, &iccRead);
```
y-guyon, line 4388: `AVIF_CHECKRES`

src/read.c
```diff
@@ -4467,26 +4515,39 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
...
+                                              /*cicpSet=*/NULL));
+
         const avifProperty * clliProp = avifPropertyArrayFind(&toneMappedImageItemTmp->properties, "clli");
         if (clliProp) {
-            decoder->image->gainMap->image->clli = clliProp->u.clli;
+            gainMap->altCLLI = clliProp->u.clli;
+        }
+
+        const avifProperty * pixiProp = avifPropertyArrayFind(&toneMappedImageItemTmp->properties, "pixi");
+        if (pixiProp && pixiProp->u.pixi.planeCount > 0) {
```
y-guyon, line 4547: Should `pixiProp->u.pixi.planeCount == 0` be an error?

src/write.c
```diff
@@ -251,6 +256,9 @@ static avifEncoderData * avifEncoderDataCreate()
     }
     memset(data, 0, sizeof(avifEncoderData));
     data->imageMetadata = avifImageCreateEmpty();
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+    data->altImageMetadata = avifImageCreateEmpty();
```
y-guyon, line 260: Missing
```c
    if (!data->altImageMetadata) {
        goto error;
    }
```

src/write.c
```diff
@@ -1356,6 +1400,10 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
     if (hasGainMap) {
         AVIF_CHECKRES(avifValidateImageBasicProperties(firstCell->gainMap->image));
         AVIF_CHECKRES(avifValidateGrid(gridCols, gridRows, cellImages, /*validateGainMap=*/AVIF_TRUE, &encoder->diag));
+        if (firstCell->gainMap->image->colorPrimaries != 2 || firstCell->gainMap->image->transferCharacteristics != 2) {
```
y-guyon, line 1403: Any reason why this is `2` instead of `AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED`?

src/write.c
```diff
@@ -947,6 +960,24 @@ size_t avifEncoderGetGainMapSizeBytes(avifEncoder * encoder)
 {
     return encoder->data->gainMapSizeBytes;
 }
+
+// Sets altImageMetadata's metadata values to represent the "alternate" image after applying the gain map to the base image.
```
y-guyon, line 964: Maybe use "as if" instead of "after" because at first I thought the function would "apply the gain map to the base image".

y-guyon (approved):

src/read.c
```diff
@@ -4544,7 +4541,11 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
         }
 
         const avifProperty * pixiProp = avifPropertyArrayFind(&toneMappedImageItemTmp->properties, "pixi");
-        if (pixiProp && pixiProp->u.pixi.planeCount > 0) {
+        if (pixiProp) {
+            if (pixiProp->u.pixi.planeCount == 0) {
+                avifDiagnosticsPrintf(data->diag, "Box[pixi] of tmap item contains invalid plane count [%u]", pixiProp->u.pixi.planeCount);
```
y-guyon, line 4546: Note that the specification does not forbid `num_channels` to be 0. I meant it as libavif should ignore or reject an item tagged with such as property.  
The current error message is fine though.

y-guyon (approved):

apps/avifgainmaputil/swapbase_command.cc
maryla-uc: Yeah it was intentional but a bit weird indeed. Reworked.

src/write.c
maryla-uc: Indeed, thanks.

PR #1813, opened 2023-11-28
Limit the possible number of threads to fuzz on.

vrabaud:
Force: auto (0), 1 thread, multiple threads (2), and maximum (64).
This will limit a parameter to fuzz upon and hence improve fuzzing.


vrabaud:
Locally, I've had more luck finding bugs by reducing the parameter space by an order of magnitude this way. Let's wait to see if the latest changes find bugs without this change.

Issue #1816, opened 2023-11-28
are_images_equal and many tests are linked with both libavif and libavif_internal


wantehchang:
I found that are_images_equal and many tests are linked with both libavif and libavif_internal. This is more noticeable when `BUILD_SHARED_LIBS` is ON (the default), because when `BUILD_SHARED_LIBS` is OFF, libavif_internal is an alias of libavif.

```
if(BUILD_SHARED_LIBS)
    add_library(avif_internal STATIC ${AVIF_SRCS})
    ...
else()
    add_library(avif_internal ALIAS avif)
    ...
endif()
```

Steps to reproduce:
```
$ cmake -G Ninja .. -DAVIF_CODEC_AOM=SYSTEM -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON -DAVIF_BUILD_TESTS=ON -DAVIF_LOCAL_GTEST=ON
$ ninja -v
```

Here are two representative linker command lines on Linux:

> /usr/bin/c++   tests/CMakeFiles/aviftest_helpers.dir/gtest/aviftest_helpers.cc.o tests/CMakeFiles/are_images_equal.dir/gtest/are_images_equal.cc.o -o tests/are_images_equal  -Wl,-rpath,/home/wtc/libavif.1/libavif/build  libavif_apps.a  libavif.so.16.0.2  /usr/lib/x86_64-linux-gnu/libpng.so  /usr/lib/x86_64-linux-gnu/libz.so  /usr/lib/x86_64-linux-gnu/libjpeg.so  libavif_internal.a  /usr/lib/x86_64-linux-gnu/libaom.so  -lm

[...]

PR #1819, opened 2023-11-29
Do not depend on both avif and avif_internal

vrabaud:
This properly fixes https://github.com/AOMediaCodec/libavif/issues/1816

In a follow-up PR, I can also move `internal.h` to `./src/` (or an `include_private` folder).

The header will still be accessed from targets if we `target_include_directories` them with `$<TARGET_PROPERTY:avif,INCLUDE_DIRECTORIES>` (which contains both public and private directories).

This will be less confusing for the user ("it's internal but is it still public?") and will enable us to split internal.h in several common-meaning headers.


include/avif/internal.h
```diff
@@ -89,9 +89,9 @@ typedef float (*avifTransferFunction)(float);
 // and can go beyond 1.0 for HDR transfer characteristics:
 // - For AVIF_TRANSFER_CHARACTERISTICS_SMPTE2084 (PQ), the linear range is [0.0, 10000/203]
 // - For AVIF_TRANSFER_CHARACTERISTICS_HLG, the linear range is [0.0, 1000/203]
-avifTransferFunction avifTransferCharacteristicsGetGammaToLinearFunction(avifTransferCharacteristics atc);
+AVIF_API avifTransferFunction avifTransferCharacteristicsGetGammaToLinearFunction(avifTransferCharacteristics atc);
```
wantehchang, line 92: We should not export the private functions declared in internal.h. Even though we don't provide the internal.h header to our clients, they can still call these functions if these functions are exported.

Note: In some Unix platforms (including Linux), we can give exported symbols a version (which is a string), which allows us to mark certain exported functions as private exports by using a version with "Private" in its name. But I don't think this kind of symbol versioning feature is available on Windows, and it may not even be available on all Unix platforms.

wantehchang (dismissed):
Vincent: Thanks for writing a fix for https://github.com/AOMediaCodec/libavif/issues/1816 quickly. But we should only export public functions.

fdintino:
This all looks good to me.

tests/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ endforeach()
 
 if(AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
     add_library(aviftest_helpers OBJECT gtest/aviftest_helpers.cc)
-    target_link_libraries(aviftest_helpers avif_apps avif_internal)
+    target_link_libraries(aviftest_helpers avif_apps_internal avif_internal)
```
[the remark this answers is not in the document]
vrabaud: Right, but as `aviftest_helpers` uses `avif_internal`, it needs to be linked to `avif_apps_internal`. We could also create an `aviftest_helpers_internal`. Or split avif.c: internal.h is only used here for `AVIF_CHECKERR`, `AVIF_CHECRES` and `avifImageCopySamples`. We could envision a small utils library that contain non-avif specific stuff (like the `AVIF_CHECK*`) and an image library that would just deal with the basic functions for images (like `avifImageCopySamples`) which is half of avif.c.

include/avif/internal.h
```diff
@@ -656,6 +656,10 @@ AVIF_NODISCARD avifBool avifSequenceHeaderParse(avifSequenceHeader * header, con
 #define AVIF_INDEFINITE_DURATION64 UINT64_MAX
 #define AVIF_INDEFINITE_DURATION32 UINT32_MAX
 
+#if defined(AVIF_BUILDING_SHARED_LIBS) && defined(AVIF_BUILDING_STATIC_LIBS)
+#error "Your target is linking against avif and avif_internal: only one should be chosen"
+#endif
```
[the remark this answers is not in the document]
vrabaud: Nothing subtle, just a habit.

CMakeLists.txt
```diff
@@ -511,18 +511,27 @@ if(NOT AVIF_CODEC_AOM_ENABLED
     message(WARNING "libavif: No decoding library is enabled.")
 endif()
 
-set_target_properties(avif PROPERTIES VERSION ${LIBRARY_VERSION} SOVERSION ${LIBRARY_SOVERSION} C_VISIBILITY_PRESET hidden)
-target_include_directories(avif PUBLIC $<BUILD_INTERFACE:${libavif_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include>)
+set_target_properties(avif_obj PROPERTIES C_VISIBILITY_PRESET hidden)
```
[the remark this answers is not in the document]
vrabaud: Actually, this is where it is necessary. Removing below.

CMakeLists.txt
```diff
@@ -531,23 +540,12 @@ endif()
...
-        set_target_properties(avif_internal PROPERTIES LINKER_LANGUAGE "CXX")
-    endif()
-    if(NOT AVIF_LIBYUV_ENABLED)
-        target_include_directories(avif_internal PRIVATE ${libavif_SOURCE_DIR}/third_party/libyuv/include/)
-    endif()
-    if(AVIF_ENABLE_COMPLIANCE_WARDEN)
-        target_include_directories(avif_internal PRIVATE ${libavif_SOURCE_DIR}/ext/ComplianceWarden/src/utils/)
-    endif()
+    # Define the following to avoid linking against the two at the same time.
+    target_compile_definitions(avif PUBLIC $<BUILD_INTERFACE:AVIF_BUILDING_SHARED_LIBS>)
```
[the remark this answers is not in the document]
vrabaud: Good catch. I also renamed `AVIF_BUILD_STATIC_LIBS` to `AVIF_USING_STATIC_LIBS`.

CMakeLists.txt
```diff
@@ -624,30 +622,44 @@ if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE
...
             add_compile_definitions(AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION)
-            target_link_libraries(avif_apps LibXml2::LibXml2)
+            target_link_libraries(avif_apps_obj LibXml2::LibXml2)
         else()
             message(STATUS "libavif: libxml2 not found; avifenc will ignore any gain map in jpeg files")
         endif()
     endif()
+
+    # Main avif_apps library.
+    add_library(avif_apps)
```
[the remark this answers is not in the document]
vrabaud: Thx for catching it! all fine now:
```
   linux-vdso.so.1 (0x00007ffc0cbb4000)
        libavif.so.16 (0x00007f0248600000)
        libpng16.so.16 => /lib/x86_64-linux-gnu/libpng16.so.16 (0x00007f0248e55000)
        libjpeg.so.62 => /lib/x86_64-linux-gnu/libjpeg.so.62 (0x00007f024856d000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f024848e000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f02482ac000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f0248e34000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f0248eda000)
```

tests/CMakeLists.txt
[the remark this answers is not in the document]
wantehchang: I discussed the `aviftest_helpers` issue in https://github.com/AOMediaCodec/libavif/pull/1836. The use of `AVIF_CHECKERR` and `AVIF_CHECRES` in aviftest_helpers.cc can be easily avoided. And we can make `avifImageCopySamples` a public function.

CMakeLists.txt
```diff
@@ -511,18 +511,27 @@ if(NOT AVIF_CODEC_AOM_ENABLED
...
+    target_include_directories(avif_obj PRIVATE ${libavif_SOURCE_DIR}/third_party/libyuv/include/)
 endif()
 if(AVIF_ENABLE_COMPLIANCE_WARDEN)
-    target_include_directories(avif PRIVATE ${libavif_SOURCE_DIR}/ext/ComplianceWarden/src/utils/)
+    target_include_directories(avif_obj PRIVATE ${libavif_SOURCE_DIR}/ext/ComplianceWarden/src/utils/)
 endif()
 set(AVIF_PKG_CONFIG_EXTRA_CFLAGS "")
 if(BUILD_SHARED_LIBS)
-    target_compile_definitions(avif PUBLIC AVIF_DLL PRIVATE AVIF_BUILDING_SHARED_LIBS)
+    target_compile_definitions(avif_obj PUBLIC AVIF_DLL PRIVATE AVIF_BUILDING_SHARED_LIBS)
```
wantehchang, line 525: Nit/optional: When building a file in src/, the compiler command line is:

> /usr/bin/cc -DAVIF_BUILDING_SHARED_LIBS -DAVIF_CODEC_AOM=1 -DAVIF_CODEC_AOM_DECODE=1 -DAVIF_CODEC_AOM_ENCODE=1 -DAVIF_DLL  ...

It would be good if `-DAVIF_BUILDING_SHARED_LIBS` and `-DAVIF_DLL` could be next to each other. I wonder if the current gap between them is caused by line 547.

CMakeLists.txt
```diff
@@ -135,13 +135,13 @@ else()
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 endif()
 
-add_library(avif)
+add_library(avif_obj OBJECT)
 
-# Adds <target> to avif's public link libraries for build, and adds
+# Adds <target> to avif_obj's public link libraries for build, and adds
```
wantehchang, line 140: It seems that the `avif_target_link_library` function (line 143) should operate on the avif library rather than the avif_obj library.

For example, this function talks about installation, but we don't install the avif_obj library.

CMakeLists.txt
```diff
@@ -511,18 +511,27 @@ if(NOT AVIF_CODEC_AOM_ENABLED
     message(WARNING "libavif: No decoding library is enabled.")
 endif()
 
-set_target_properties(avif PROPERTIES VERSION ${LIBRARY_VERSION} SOVERSION ${LIBRARY_SOVERSION} C_VISIBILITY_PRESET hidden)
-target_include_directories(avif PUBLIC $<BUILD_INTERFACE:${libavif_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include>)
+set_target_properties(avif_obj PROPERTIES C_VISIBILITY_PRESET hidden)
+set_property(TARGET avif_obj PROPERTY POSITION_INDEPENDENT_CODE ON)
```
wantehchang, line 515: This doesn't seem to be in the original code. This is a good change. I am just wondering if it needs to be part of this pull request.

If this change can stand alone, it is better to do it separately.

CMakeLists.txt
```diff
@@ -531,23 +540,12 @@ endif()
...
-        set_target_properties(avif_internal PROPERTIES LINKER_LANGUAGE "CXX")
-    endif()
-    if(NOT AVIF_LIBYUV_ENABLED)
-        target_include_directories(avif_internal PRIVATE ${libavif_SOURCE_DIR}/third_party/libyuv/include/)
-    endif()
-    if(AVIF_ENABLE_COMPLIANCE_WARDEN)
-        target_include_directories(avif_internal PRIVATE ${libavif_SOURCE_DIR}/ext/ComplianceWarden/src/utils/)
-    endif()
+    # Define the following to avoid linking against the two at the same time.
+    target_compile_definitions(avif PUBLIC $<BUILD_INTERFACE:AVIF_DLL>)
```
wantehchang, line 547: This seems redundant with line 525 above. Can we delete this?

If we delete this line, we need to update the comment at line 546.

CMakeLists.txt
```diff
@@ -624,30 +622,44 @@ if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE
     endif()
 
     add_library(
-        avif_apps STATIC apps/shared/avifexif.c apps/shared/avifjpeg.c apps/shared/avifpng.c apps/shared/avifutil.c
-                         apps/shared/iccmaker.c apps/shared/y4m.c third_party/iccjpeg/iccjpeg.c
+        avif_apps_obj OBJECT $<TARGET_OBJECTS:avif_obj> apps/shared/avifexif.c apps/shared/avifjpeg.c apps/shared/avifpng.c
+                             apps/shared/avifutil.c apps/shared/iccmaker.c apps/shared/y4m.c third_party/iccjpeg/iccjpeg.c
     )
-    target_link_libraries(avif_apps avif PNG::PNG ZLIB::ZLIB JPEG::JPEG)
+    set_property(TARGET avif_apps_obj PROPERTY POSITION_INDEPENDENT_CODE ON)
```
wantehchang, line 628: This change does not seem necessary.

If this is necessary or desirable, then we should also compile avifdec.c and avifenc.c as position-independent code, right?

Like my comment at line 515 above, if this change can stand alone, it is better to make this change in a separate pull request.

CMakeLists.txt
```diff
@@ -624,30 +622,44 @@ if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE
...
+    target_link_libraries(avif_apps_obj avif_obj PNG::PNG ZLIB::ZLIB JPEG::JPEG)
```
wantehchang, line 629: IMPORTANT: I think this line causes the .c files in avif_apps_obj, such as avifjpeg.c, to be compiled with -DAVIF_DLL when -DBUILD_SHARED_LIBS=ON is specified.

Then, avifjpeg.c.o, compiled with -DAVIF_DLL, is added to both libavif_apps.a and libavif_apps_internal.a.

But the object files in libavif_apps_internal.a should be compiled with -DAVIF_USING_STATIC_LIBS, not -DAVIF_DLL.

wantehchang (commented):
Vincent: I inspected the build log. This is much better

High-level comments:
1. The position-independent code changes seem unrelated to fixing the dual avif/avif_internal linking bug.
2. Please see the comment marked with "IMPORTANT".

CMakeLists.txt
vrabaud: From what I see, definitions are sorted in alphabetical order. This is what we have right now on main:
```
/usr/bin/cc -DAVIF_BUILDING_SHARED_LIBS -DAVIF_CODEC_AOM=1 -DAVIF_CODEC_AOM_DECODE=1 -DAVIF_CODEC_AOM_ENCODE=1 -DAVIF_DLL -Davif_EXPORTS
```

CMakeLists.txt
vrabaud: Both would work as they get transmitted transitively. Still, because of the install, I switch to `avif` for the `INSTALL_INTERFACE`

CMakeLists.txt
vrabaud: This is due to the fact that avif_obj is now kindof like a static library. So, even if you build a shared lib, you need that. I set CMAKE_POSITION_INDEPENDENT_CODE once and for all at the beginning.

CMakeLists.txt
vrabaud: Well, that is already the case:
https://github.com/AOMediaCodec/libavif/blob/15fa36c4c1b1beb409ec7ef44e41fadf8a38b944/CMakeLists.txt#L524
and https://github.com/AOMediaCodec/libavif/blob/15fa36c4c1b1beb409ec7ef44e41fadf8a38b944/CMakeLists.txt#L630

It does not matter that much as avif_internal is not meant to be distributed. This only affects which symbols are visible and it has to be used at compile time: we would definitely need to compile the lib twice then.

wantehchang: Thanks for the explanation. On Windows, when `-DBUILD_SHARED_LIBS=ON` is specified, if we compile avifjpeg.c with `-DAVIF_DLL,` then avifjpeg.c will see the `AVIF_API` macro defined as `__declspec(dllimport)`, and I believe a call to `avifImageAllocatePlanes` will be recorded as a reference to the symbol `__imp_avifImageAllocatePlanes` in the object file. If so, that reference can only be resolved by linking with the avif DLL import library. If we link the object file with only the avif_internal static library, then I believe we will get an unresolved symbol linker error for `__imp_avifImageAllocatePlanes`.

In the current code, since we either link with the avif DLL import library or link with both the avif DLL import library and the avif_internal static library, we will not see this unresolved symbol linker error. But in the previous version of this pull request, we would link the tests with only the avif_internal static library. Then the tests that use avif_apps would get the unresolved symbol linker error, I think. (I haven't verified this on my Windows laptop though.)

wantehchang: I just tested this pull request on Windows. I verified that by compiling avifjpeg.c with `-DAVIF_DLL`, we recorded references to symbols such as `__imp_avifRGBImageSetDefaults` in the object file (avifjpeg.c.obj). But I was surprised to find that the static library avif_internal.lib can also satisfy the references to the `__imp_*` symbols. I don't know why.

So I think it is fine that files in avif_apps_obj are compiled only once, with `-DAVIF_DLL`.

CMakeLists.txt
```diff
@@ -624,30 +620,43 @@ if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE
...
+
+    # Main avif_apps library.
+    add_library(avif_apps STATIC)
+    target_link_libraries(avif_apps PUBLIC avif PRIVATE avif_apps_obj)
+    target_include_directories(avif_apps INTERFACE apps/shared)
+
+    # avif_apps_internal is to use when linking to avif_internal.
+    if(BUILD_SHARED_LIBS)
+        add_library(avif_apps_internal STATIC)
+        target_link_libraries(avif_apps_internal PUBLIC avif_internal PRIVATE avif_apps_obj)
```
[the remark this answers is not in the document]
wantehchang: I tested this on Windows. I found that this causes no problems because the static library avif_internal.lib can also satisfy the references to the `__imp_avif*` symbols. So I think we can ignore this issue.

I am a little confused now -- do we still need both the avif_apps and avif_apps_internal libraries? Or can we get by with just the avif_apps library?

CMakeLists.txt
```diff
@@ -127,21 +127,22 @@ else()
     set(XCRUN)
 endif()
 
+# This is also needed to get shared libraries (e.g. pixbufloader-avif) to compile against a static libavif.
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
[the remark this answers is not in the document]
vrabaud: Well, it only makes sense because of the object library that requires it for shared builds

CMakeLists.txt
```diff
@@ -511,18 +512,25 @@ if(NOT AVIF_CODEC_AOM_ENABLED
...
+    target_include_directories(avif_obj PRIVATE ${libavif_SOURCE_DIR}/ext/ComplianceWarden/src/utils/)
 endif()
 set(AVIF_PKG_CONFIG_EXTRA_CFLAGS "")
 if(BUILD_SHARED_LIBS)
-    target_compile_definitions(avif PUBLIC AVIF_DLL PRIVATE AVIF_BUILDING_SHARED_LIBS)
+    target_compile_definitions(avif_obj PUBLIC AVIF_DLL PRIVATE AVIF_BUILDING_SHARED_LIBS)
     set(AVIF_PKG_CONFIG_EXTRA_CFLAGS " -DAVIF_DLL")
+endif()
+
+# Main library.
```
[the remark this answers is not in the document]
vrabaud: Done

CMakeLists.txt
```diff
@@ -531,23 +539,11 @@ endif()
...
-    if(AVIF_USE_CXX)
-        set_target_properties(avif_internal PROPERTIES LINKER_LANGUAGE "CXX")
-    endif()
-    if(NOT AVIF_LIBYUV_ENABLED)
-        target_include_directories(avif_internal PRIVATE ${libavif_SOURCE_DIR}/third_party/libyuv/include/)
-    endif()
-    if(AVIF_ENABLE_COMPLIANCE_WARDEN)
-        target_include_directories(avif_internal PRIVATE ${libavif_SOURCE_DIR}/ext/ComplianceWarden/src/utils/)
-    endif()
+    # Define the following to avoid linking against the two at the same time.
```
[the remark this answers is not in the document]
vrabaud: Done

tests/gtest/aviftest_helpers.cc
```diff
@@ -15,13 +15,88 @@
...
+
+// Forward any error to the caller now or continue execution.
+#define AVIF_CHECKRES(A)              \
+  do {                                \
+    const avifResult result__ = (A);  \
+    if (result__ != AVIF_RESULT_OK) { \
+      avifBreakOnError();             \
+      return result__;                \
+    }                                 \
+  } while (0)
```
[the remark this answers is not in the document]
vrabaud: Done

tests/gtest/aviftest_helpers.cc
```diff
@@ -15,13 +15,88 @@
...
+
+static void avifImageCopySamples(avifImage* dstImage, const avifImage* srcImage,
```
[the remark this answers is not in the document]
vrabaud: Done

tests/CMakeLists.txt
```diff
@@ -117,7 +124,7 @@ if(AVIF_ENABLE_GTEST)
     target_link_libraries(avifincrtest aviftest_helpers avifincrtest_helpers)
     add_test(NAME avifincrtest COMMAND avifincrtest ${CMAKE_CURRENT_SOURCE_DIR}/data/)
 
-    add_avif_gtest(avifcolrtest)
+    add_avif_internal_gtest(avifcolrtest)
```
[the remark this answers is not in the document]
vrabaud: They actually only need it for macros. I put the macro in avif*test_helpers.h and I was able to remove the internal header in a few more tests.

CMakeLists.txt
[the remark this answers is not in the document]
vrabaud: Indeed, I fixed it and `AVIF_DLL` is not defined for avif_apps anymore.

tests/gtest/avifincrtest_helpers.h
```diff
@@ -11,6 +11,41 @@
...
+#define AVIF_CHECKRES(A)              \
+  do {                                \
+    const avifResult result__ = (A);  \
+    if (result__ != AVIF_RESULT_OK) { \
+      avifBreakOnError();             \
+      return result__;                \
+    }                                 \
+  } while (0)
+//------------------------------------------------------------------------------
+
```
wantehchang, line 48: Delete this.

Among the files that include this header, only avifincrtest_helpers.cc uses the `AVIF_CHECK*` macros. But avifincrtest_helpers.cc also includes aviftest_helpers.h.

Note: Some files (e.g., avifgainmaptest.cc) include both this header and aviftest_helpers.h. I am surprised that compilers allow duplicate macro definitions. I guess if the definitions are exactly the same, then they are allowed.

tests/gtest/aviftest_helpers.cc
```diff
@@ -15,13 +15,58 @@
...
 #include "avif/avif_cxx.h"
-#include "avif/internal.h"
 #include "avifpng.h"
 #include "avifutil.h"
 
 namespace avif {
 namespace testutil {
 
+//------------------------------------------------------------------------------
+// Duplicated from internal.h
```
wantehchang, line 25: Update this comment. Perhaps just say CopyImageSamples is a copy of avifImageCopySamples?

tests/gtest/aviftest_helpers.h
```diff
@@ -17,6 +17,50 @@
...
+// Forward any error to the caller now or continue execution.
+#define AVIF_CHECKRES(A)              \
+  do {                                \
+    const avifResult result__ = (A);  \
+    if (result__ != AVIF_RESULT_OK) { \
+      avifBreakOnError();             \
+      return result__;                \
+    }                                 \
+  } while (0)
+//------------------------------------------------------------------------------
```
wantehchang, line 62: Nit: Move the macro definitions outside the avif namespace to make it clear that namespaces don't apply to macros.

Later, we should use a different prefix for these macros.

wantehchang (approved):
LGTM. Thanks a lot!

CMakeLists.txt
```diff
@@ -139,7 +139,7 @@ add_library(avif_obj OBJECT)
 add_library(avif)
 
 # Adds <target> to avif_obj's public link libraries for build, and adds
-# the library as an install link library for export if a consumer
+# the 'target' library as an install link library for export if a consumer
```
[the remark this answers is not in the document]
vrabaud: <> indeed

tests/gtest/avifincrtest_helpers.h
vrabaud: Indeed, that's how C works: you are allowed to redefine the macro as long as it is the same:

tests/gtest/aviftest_helpers.h
vrabaud: Done

CMakeLists.txt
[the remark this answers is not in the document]
vrabaud: Done

CMakeLists.txt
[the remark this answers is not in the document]
vrabaud: I did not add it to the CHANGELOG as it seems too minor.

PR #1823, opened 2023-11-29
Revert pointer storage fixes

vigneshvg:
After PR #1817, there is no need to safeguard against potential dangling pointers. So restore the old code which is simpler and does not have unnecessary intermediaries (like storing the index instead of pointers).

This PR manually reverts the following commits:
7845153645cfe245de5add94fb07c227c2d16402
2041109967c1746178c736e0a1504d9c97b51a5c


wantehchang (approved):
LGTM. It would be good to have Maryla take a look because she reviewed one of the commits that is reverted.

maryla-uc (approved):

PR #1824, opened 2023-11-30
fix(svt): Don't clamp enc_mode to [0,8]

fdintino:
SVT-AV1 added encoder modes 9-12 in v0.9. Because speed is still clamped to max 8 it's not possible to use any of these faster presets.

This still would prevent users from selecting M11 or M12, but addressing that would require broader changes.


vrabaud:
I guess we should keep the old behavior for older versions:
```c++
#if(SVT_AV1_CHECK_VERSION(0,9,0)
            svt_config->enc_mode = (int8_t)encoder->speed;
#else
            int speed = AVIF_CLAMP(encoder->speed, 0, 8);
            svt_config->enc_mode = (int8_t)speed;
#endif
```

vrabaud (approved):

PR #1829, opened 2023-12-01
Replace atoi() by y4mReadUnsignedInt() in y4m.c

y-guyon:
atoi() could lead to out-of-bounds access.

### DIFF
--- a/apps/shared/y4m.c
+++ b/apps/shared/y4m.c
@@ -7,9 +7,14 @@
 
 #include <assert.h>
 #include <inttypes.h>
+#include <limits.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+
+#include "avif/avif.h"
+#include "avifutil.h"
 
 #define Y4M_MAX_LINE_SIZE 2048 // Arbitrary limit. Y4M headers should be much smaller than this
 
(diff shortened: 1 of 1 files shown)


vrabaud (approved):

apps/shared/y4m.c
```diff
@@ -132,6 +137,23 @@ static avifBool y4mColorSpaceParse(const char * formatString, struct y4mFrameIte
...
+        return -1;
+    }
+    int64_t value = 0;
+    while (start != end && *start >= '0' && *start <= '9') {
+        value = value * 10 + (*(start++) - '0');
+        if (value > INT_MAX) {
+            return -1;
+        }
+    }
+    return (int)value;
```
wantehchang, line 154: IMPORTANT: Return -1 if the input contains no decimal digits.

This is harmless because the caller also treats 0 as an error.

apps/shared/y4m.c
```diff
@@ -7,10 +7,15 @@
...
 #include <assert.h>
 #include <inttypes.h>
+#include <limits.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
+#include "avif/avif.h"
+#include "avifutil.h"
```
wantehchang, line 17: Optional: I remember IWYU allows foo.c to rely on the headers included by foo.h. This is why these two headers were not included. It's fine to follow the IWYU policy strictly.

PR #1830, opened 2023-12-01
Allow for several directories in TEST_DATA_DIR (and rename it)

vrabaud:
This can be useful when analyzing different corpuses


jzern:
Should the format of this environment variable be documented somewhere like the README?

PR #1834, opened 2023-12-01
Fix y4mReadUnsignedInt() when no digit char

y-guyon:


vrabaud (approved):

apps/shared/y4m.c
```diff
@@ -141,10 +140,10 @@ static avifBool y4mColorSpaceParse(const char * formatString, struct y4mFrameIte
 // Returns -1 in case of failure.
 int y4mReadUnsignedInt(const char * start, const char * end)
 {
-    if (start >= end) {
+    if (start >= end || *start < '0' || *start > '9') {
         return -1;
     }
-    int64_t value = 0;
+    int64_t value = *(start++) - '0';
```
wantehchang, line 146: Optional: I think we can also initialize `value` to 0 as before and change the while loop to a do-while loop.

wantehchang (approved):

PR #1837, opened 2023-12-02
Convert avifDecoderItemArray into an array of pointers

wantehchang:
Port commit 06799082864459d6bb31ae3612dc753035c61ce1 and commit 380fb999169ecaa280853e86901303fb35e79f13 to the v1.0.x branch.

Instead of storing the struct directly in a dynamic array, store it as pointers that are stored in a dynamic array. This way when the dynamic array is resized, the pointers will still be valid and be copied over and there will be no dangling pointers.


wantehchang:
Vignesh, Yannis: I manually ported Vignesh's bug fix to the v1.0.x release branch. Please review the changes to `avifMetaFindItem()` carefully. The changes to that function on the v1.0.x branch are simpler because we don't need to handle the failure of `avifAlloc()` and `avifArrayPushPtr()`.

y-guyon (approved):

vigneshvg (approved):

Issue #1845, opened 2023-12-04
Fully static libavif build fails with local zlib


y-guyon:
To reproduce:

```sh
git clone https://github.com/AOMediaCodec/libavif.git && \
cd libavif && \
git checkout c44a43ba943c1cf60a200a7bea2c1dd774ae547c && \
cd ext && \
./libjpeg.cmd && \
./zlibpng.cmd && \
cd .. && \
mkdir build && \
cd build && \
cmake .. \
 -DAVIF_BUILD_APPS=ON \
 -DAVIF_LOCAL_JPEG=ON \
 -DAVIF_LOCAL_ZLIBPNG=ON \
 -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ \
 -DCMAKE_EXE_LINKER_FLAGS="-static-libgcc -static -static-libstdc++ -fPIC" \
 -DCMAKE_C_FLAGS_RELEASE="-static -fPIC" \
 -DCMAKE_CXX_FLAGS_RELEASE="-static -fPIC" \
 -DBUILD_SHARED_LIBS=OFF && \
cmake --build .
```

Output:
```
[113%] Linking C shared library libz.so
[113%] Built target zlib
[114%] Building C object ext/zlib/CMakeFiles/example.dir/test/example.c.o
[114%] Linking C executable example
clang: warning: argument unused during compilation: '-static-libstdc++' [-Wunused-command-line-argument]
/usr/bin/ld: attempted static link of dynamic object `libz.so.1.3'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [ext/zlib/CMakeFiles/example.dir/build.make:98: ext/zlib/example] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:379: ext/zlib/CMakeFiles/example.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
```

PR #1848, opened 2023-12-04
Only build zlib/example when building tests

vrabaud:
This fixes https://github.com/AOMediaCodec/libavif/issues/1845


y-guyon (approved):
Thank you

PR #1862, opened 2023-12-08
Fix reading alt image metadata when enableDecodingGainMap is off.

maryla-uc:
This simply moves the metadata reading outside of the `if(decoder->enableDecodingGainMap)`, and inside of an `if (decoder->enableParsingGainMapMetadata)` instead.
+ adds a test.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -4518,7 +4518,7 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
     }
 
     if (decoder->enableParsingGainMapMetadata) {
-        avifGainMap * gainMap = decoder->image->gainMap;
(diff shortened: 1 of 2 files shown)


src/read.c
```diff
@@ -4557,6 +4546,24 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
         }
     }
 
+    if (decoder->enableDecodingGainMap) {
+        avifGainMap * gainMap = decoder->image->gainMap;
```
vrabaud, line 4550: `avifGainMap * const gainMap` and above ? skal style :)

vrabaud (approved):

PR #1868, opened 2023-12-11
Gainmap computation: fix handling of extreme values.

maryla-uc:
Gain map values outside of [-6, +6] were being clamped.

### DIFF
--- a/src/gainmap.c
+++ b/src/gainmap.c
@@ -5,6 +5,7 @@
 #include <assert.h>
 #include <float.h>
 #include <math.h>
+#include <stdio.h>
 #include <string.h>
 
 #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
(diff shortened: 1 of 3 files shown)


src/gainmap.c
```diff
@@ -314,29 +315,43 @@ avifResult avifImageApplyGainMap(const avifImage * baseImage,
...
-// Arbitrary max value, roughly equivalent to bringing SDR white to max PQ white (log2(10000/203) ~= 5.62)
-#define BUCKET_MAX_VALUE 6.0f
-#define BUCKET_MIN_VALUE -BUCKET_MAX_VALUE
-#define BUCKET_RANGE (BUCKET_MAX_VALUE - BUCKET_MIN_VALUE)
-#define MAX_OUTLIERS_RATIO 0.001f // 0.1%
+#define MAX_OUTLIERS_RATIO 0.001f  // 0.1%
 
 // Returns the index of the histogram bucket for a given value.
-static int avifValueToBucketIdx(float v)
+static int avifValueToBucketIdx(float v, float bucketMin, float bucketMax)
```
y-guyon, line 321: For extra clarity, indicate that `bucketMax` is inclusive in the comment?

src/gainmap.c
```diff
@@ -314,29 +315,43 @@ avifResult avifImageApplyGainMap(const avifImage * baseImage,
...
+    for (int i = 0; i < numPixels; ++i) {
+        min = AVIF_MIN(min, gainMapF[i]);
+        max = AVIF_MAX(max, gainMapF[i]);
+    }
+
+    const float bucketMin = min;
+    // Use a bucket range of constant size 12 so that each bucket has a fixed size.
+    // 12 is an arbitrary empirical value, equivalent to a [-6, +6] range, where
+    // +6 is roughly equivalent to bringing SDR white to max PQ white (log2(10000/203) ~= 5.62).
+    const float bucketMax = min + 12;
```
y-guyon, line 348: Why not use [0:12] rather than [min:min+12]? It seems weird to fix the relative maximum but not the absolute minimum.

Also [-6:+6] is mentioned but the used range is [min:min+12]. Is min expected to be around -6?

y-guyon (approved):

y-guyon (approved):

src/gainmap.c
maryla-uc: It was intentional but the intention was not very clear :) I reworked the code to make it clearer hopefully that the idea is to have buckets of fixed size.

PR #1869, opened 2023-12-12
List AVIF_COLOR_PRIMARIES_SRGB after BT709

wantehchang:
This order suggests that AVIF_COLOR_PRIMARIES_BT709 is the main name of the constant and AVIF_COLOR_PRIMARIES_SRGB is an alias. The association of color_primaries value 1 with BT.709 is well known.


include/avif/avif.h
```diff
@@ -311,8 +311,8 @@ enum
     // This is actually reserved, but libavif uses it as a sentinel value.
     AVIF_COLOR_PRIMARIES_UNKNOWN = 0,
 
-    AVIF_COLOR_PRIMARIES_SRGB = 1,
     AVIF_COLOR_PRIMARIES_BT709 = 1,
+    AVIF_COLOR_PRIMARIES_SRGB = 1,
```
wantehchang, line 315: Note that I didn't reorder `	AVIF_TRANSFER_CHARACTERISTICS_PQ` and `AVIF_TRANSFER_CHARACTERISTICS_SMPTE2084` for the transfer_characteristics value 16. I think we can make an exception for that constant because PQ is better known than SMPTE ST 2084.

y-guyon (approved):

PR #1871, opened 2023-12-12
Add missing result check and fix build warning.

maryla-uc:


tests/gtest/avifgainmaptest.cc
```diff
@@ -1358,7 +1358,9 @@ TEST(FindMinMaxWithoutOutliers, AllSame) {
     std::vector<float> values(kNumValues, v);
 
     float min, max;
-    avifFindMinMaxWithoutOutliers(values.data(), kNumValues, &min, &max);
+    ASSERT_EQ(
+        avifFindMinMaxWithoutOutliers(values.data(), kNumValues, &min, &max),
```
vrabaud, line 1362: Maybe make that function nodiscard?

vrabaud (approved):

tests/gtest/avifgainmaptest.cc
maryla-uc: It returns un avifResult so it's already  nodiscard.
Not sure why it didn't show up in presubmit.

PR #1873, opened 2023-12-12
Add support for creating gain maps for image with different primaries.

maryla-uc:


src/gainmap.c
```diff
@@ -382,11 +441,59 @@ avifResult avifFindMinMaxWithoutOutliers(const float * gainMapF, int numPixels,
...
 
+static const float kEpsilon = 1e-10f;
+
+// Decides which of 'basePrimaries' or 'altPrimaries' should be used for doing gain map math.
+// The other image (base or alternate) will be converted to this color space before computing
+// the ratio between the two images.
+// If a pixel color is outside of the target color space, some of the converted channel values will be negative.
+// This should be avoided, as the negative values must either be clamped or offset before computing the log2()
+// (since log2 only works on > 0 values). But a large offset causes artefacts when partially applying the gain map.
+// Therefore we want to do gain map math in the larger of the two color spaces.
```
y-guyon, line 452: "largest"? Not sure

src/gainmap.c
```diff
@@ -436,8 +547,68 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
+                    if (useBaseColorSpace) {
+                        rgba[c] = altGammaToLinear(rgba[c]);
+                    } else {
+                        rgba[c] = baseGammaToLinear(rgba[c]);
+                    }
+                }
+                avifLinearRGBConvertColorSpace(rgba, rgbConversionCoeffs);
+                for (int c = 0; c < 3; ++c) {
+                    if (rgba[c] < channelMin[c]) {
+                        channelMin[c] = AVIF_MIN(channelMin[c], rgba[c]);
```
y-guyon, line 592: Just `channelMin[c] = rgba[c];` or drop the condition

src/gainmap.c
```diff
@@ -436,8 +547,68 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
+                    }
+                }
+            }
+        }
+
+        for (int c = 0; c < 3; ++c) {
+            // Large offsets cause artefacts when partially applying the gain map, so set a max (empirical) offset value.
+            // If hte offset is clamped, some gain map values will get be clamped.
```
y-guyon, line 600: the  
-be

tests/gtest/avifgainmaptest.cc
```diff
@@ -1217,139 +1232,194 @@ INSTANTIATE_TEST_SUITE_P(
             /*out_rgb_format=*/AVIF_RGB_FORMAT_RGB,
             /*reference=*/"", /*min_psnr=*/0.0f, /*max_psnr=*/0.0f)));
 
-class CreateGainMapTest : public testing::TestWithParam<std::tuple<
-                              /*downscaling=*/int, /*gain_map_depth=*/int,
-                              /*gain_map_format=*/avifPixelFormat,
-                              /*min_psnr=*/float, /*max_psnr=*/float>> {};
+class CreateGainMapTest
+    : public testing::TestWithParam<std::tuple<
+          /*image1_name=*/std::string, /*image1_name=*/std::string,
```
y-guyon, line 1237: image2_name

tests/gtest/avifgainmaptest.cc
```diff
@@ -1217,139 +1232,194 @@ INSTANTIATE_TEST_SUITE_P(
...
 
-  const float hdr_headroom = (float)gain_map->metadata.alternateHdrHeadroomN /
-                             gain_map->metadata.alternateHdrHeadroomD;
+  const float image1_headroom = (float)gain_map->metadata.baseHdrHeadroomN /
+                                gain_map->metadata.baseHdrHeadroomD;
+  float image2_headroom = (float)gain_map->metadata.alternateHdrHeadroomN /
+                          gain_map->metadata.alternateHdrHeadroomD;
+
+  if (image1_headroom == image2_headroom) {
+    gain_map->metadata.alternateHdrHeadroomN += 1;
```
y-guyon, line 1285: This seems arbitrary. Is it that useless to run that test with `image1_headroom == image2_headroom`?

y-guyon (approved):

src/gainmap.c
```diff
@@ -126,6 +127,14 @@ avifResult avifRGBImageApplyGainMap(const avifRGBImage * baseImage,
 
     const uint32_t width = baseImage->width;
     const uint32_t height = baseImage->height;
+
+    const avifBool useBaseColorSpace = gainMap->metadata.useBaseColorSpace;
+    const avifColorPrimaries gainMapMathPrimaries =
+        (useBaseColorSpace || (gainMap->altColorPrimaries == AVIF_COLOR_PRIMARIES_UNSPECIFIED)) ? baseColorPrimaries
+                                                                                                : gainMap->altColorPrimaries;
```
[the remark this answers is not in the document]
maryla-uc: Because we are applying the gain map that is given to us, so we use the information in the gain map metadata.
avifChooseColorSpaceForGainMapMath() is used when creating a gain map, to find the best color space to do the math in. After that, the color space if fixed, and the results will be wrong if you use the other color space.

src/gainmap.c
maryla-uc: I believe both are ok but some people consider the -er verison to be more correct when comparing two things.
According to https://www.merriam-webster.com/grammar/can-you-use-a-superlative-of-two it doesn't matter that much.

tests/gtest/avifgainmaptest.cc
maryla-uc: Added a comment. It is indeed arbitrary, but the  tone mapping will behave unexpectedly (for the test) if the two values are equal (it will never tone map).

y-guyon: > it will never tone map

I meant it could test for that rather than change the input values, but this may already be part of another test.

y-guyon (approved):

tests/gtest/avifgainmaptest.cc
maryla-uc: Done in #1876

PR #1874, opened 2023-12-12
Change some AVIF_COLOR_PRIMARIES_SRGB to BT709

wantehchang:
In these places the use of AVIF_COLOR_PRIMARIES_BT709 is clearer.

### DIFF
--- a/src/obu.c
+++ b/src/obu.c
@@ -290,7 +290,8 @@ static avifBool parseSequenceHeaderColorConfig(avifBits * bits, avifSequenceHead
         header->av1C.chromaSubsamplingX = 1;
         header->av1C.chromaSubsamplingY = 1;
         header->yuvFormat = AVIF_PIXEL_FORMAT_YUV400;
-    } else if (header->colorPrimaries == AVIF_COLOR_PRIMARIES_BT709 && header->transferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_SRGB &&
+    } else if (header->colorPrimaries == AVIF_COLOR_PRIMARIES_BT709 &&
+               header->transferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_SRGB &&
                header->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
         header->range = AVIF_RANGE_FULL;
         header->av1C.chromaSubsamplingX = 0;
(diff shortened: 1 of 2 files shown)


y-guyon (approved):

src/obu.c
```diff
@@ -290,7 +290,7 @@ static avifBool parseSequenceHeaderColorConfig(avifBits * bits, avifSequenceHead
         header->av1C.chromaSubsamplingX = 1;
         header->av1C.chromaSubsamplingY = 1;
         header->yuvFormat = AVIF_PIXEL_FORMAT_YUV400;
-    } else if (header->colorPrimaries == AVIF_COLOR_PRIMARIES_SRGB && header->transferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_SRGB &&
+    } else if (header->colorPrimaries == AVIF_COLOR_PRIMARIES_BT709 && header->transferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_SRGB &&
```
wantehchang, line 293: Yannis: This change is optional. Please let me know if you prefer to not change this.

Here we are checking for sRGB, so the use of `AVIF_COLOR_PRIMARIES_SRGB` is appropriate. But it may be clearer to use `AVIF_COLOR_PRIMARIES_BT709` here because one is likely to compare this code with the following in the AV1 spec:

```
        color_range = 1
        subsampling_x = 0
        subsampling_y = 0
    } else {
```

src/reformat_libyuv.c
```diff
@@ -785,7 +785,7 @@ static void getLibYUVConstants(const avifImage * image, const struct YuvConstant
                 break;
             case AVIF_MATRIX_COEFFICIENTS_CHROMA_DERIVED_NCL:
                 switch (image->colorPrimaries) {
-                    case AVIF_COLOR_PRIMARIES_SRGB:
+                    case AVIF_COLOR_PRIMARIES_BT709:
```
wantehchang, line 788: It is clearer to use `AVIF_COLOR_PRIMARIES_BT709` in these two places because of the `kYuvF709Constants` in line 791 and line 849.

src/obu.c
y-guyon: I do not mind, in the WebP team we are more familiar with sRGB than BT709 but if libavif is closer to AV1 than to an image format in your opinion, that is fine with me.

y-guyon (approved):

src/obu.c
wantehchang: The purpose of this file (src/obu.c) is to parse the AV1 sequence heaser OBU. To check the correctness of the code in this file, one needs to compare it with the AV1 spec. That's why I think it would be clearer to match the AV1 spec in the constant's name.

PR #1875, opened 2023-12-13
Set image->gainMap->altTransferCharacteristics if value is 2.

maryla-uc:
Right now it's not really needed because it's already set in avifjpeg.c but this makes the logic/intention clearer and might be more future proof.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -2286,20 +2286,20 @@ MAIN()
         // interpreted as SRGB anyway.
         image->colorPrimaries = AVIF_COLOR_PRIMARIES_SRGB;
         image->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_SRGB;
+    }
 
 #if defined(AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION)
-        if (image->gainMap && !image->gainMap->altICC.size) {
-            if (image->gainMap->altColorPrimaries == AVIF_COLOR_PRIMARIES_UNSPECIFIED) {
-                // Assume the alternate image has the same primaries as the base image.
-                image->gainMap->altColorPrimaries = image->colorPrimaries;
-            }
-            if (image->gainMap->altTransferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED) {
-                // Assume the alternate image is PQ HDR.
-                image->gainMap->altTransferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_PQ;
-            }
+    if (image->gainMap && !image->gainMap->altICC.size) {
+        if (image->gainMap->altColorPrimaries == AVIF_COLOR_PRIMARIES_UNSPECIFIED) {
+            // Assume the alternate image has the same primaries as the base image.
+            image->gainMap->altColorPrimaries = image->colorPrimaries;
         }
-#endif // AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION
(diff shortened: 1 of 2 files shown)


apps/avifenc.c
```diff
@@ -2285,12 +2285,20 @@ MAIN()
...
         image->colorPrimaries = AVIF_COLOR_PRIMARIES_SRGB;
+        image->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_SRGB;
+
 #if defined(AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION)
-        if (image->gainMap && image->gainMap->image) {
-            image->gainMap->altColorPrimaries = AVIF_COLOR_PRIMARIES_SRGB;
+        if (image->gainMap && !image->gainMap->altICC.size) {
+            if (image->gainMap->altColorPrimaries == AVIF_COLOR_PRIMARIES_UNSPECIFIED) {
+                // Assume the alternate image has the same primaries as the base image.
+                image->gainMap->altColorPrimaries = image->colorPrimaries;
```
wantehchang, line 2294: [Note: This comment also applies to the other file.]

Because of line 2287 above, this is equivalent to setting `altColorPrimaries` to `AVIF_COLOR_PRIMARIES_SRGB`.

Line 2298 below sets `altTransferCharacteristics` to `AVIF_TRANSFER_CHARACTERISTICS_PQ` But PQ is usually used with color prrimaries 9 (`AVIF_COLOR_PRIMARIES_BT2020` or `AVIF_COLOR_PRIMARIES_BT2100`).

Should we set `altColorPrimaries` to `AVIF_COLOR_PRIMARIES_BT2100` here?

Or perhaps this should not be nested inside the if statement that starts at line 2282?

apps/avifgainmaputil/convert_command.cc
```diff
@@ -64,7 +64,19 @@ avifResult ConvertCommand::Run() {
     // If there is no ICC and no CICP, assume sRGB by default.
     image->colorPrimaries = AVIF_COLOR_PRIMARIES_SRGB;
     image->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_SRGB;
-    image->gainMap->altColorPrimaries = AVIF_COLOR_PRIMARIES_SRGB;
+    if (image->gainMap && !image->gainMap->altICC.size) {
```
wantehchang, line 67: Nit/optional: Use the same form (`size == 0` or `!size`) to test if ICC size is 0 at line 60 and here.

wantehchang (approved):
LGTM. I have some questions.

apps/avifenc.c
maryla-uc: > Should we set altColorPrimaries to AVIF_COLOR_PRIMARIES_BT2100 here?

No, if altColorPrimaries is not set we should assume that it's the same primaries as the base image. In practice right now the only way we can get a gain map is though a jpeg-r file, which does not allow specifying different color primaries for the alternate image, it's always the same as the base image. 
In avifjpeg.c we already set altColorPrimaries to the same value as the image
https://github.com/AOMediaCodec/libavif/blob/e0778af38864bba88f2c6ff2cd30fc7abf007500/apps/shared/avifjpeg.c#L1143
But the issue is that avifenc.c only sets the default value (sRGB) here, after we are done reading the image. When we go through the line above in avifjpeg.c, colorPrimaries is 2, so altColorPrimaries also ends up being 2.
So the idea is to propagate the new value to the altColorPrimaries field. This is why the if is nested, but I don't think it matters much i practice and perhaps it's more confusing this way so I changed it.

wantehchang (approved):
LGTM. Thanks for the reply.

Can the PQ transfer characteristics function be used with BT.709 (sRGB) color primaries? Just curious.

PR #1878, opened 2023-12-14
When converting a jpeg-r file with ICC, copy it to altICC as well.

maryla-uc:
We assume that the ICC applies to both the base and alternate images.


tests/data/paris_exif_xmp_icc_gainmap_bigendian.jpg
y-guyon, line 1: Add it to README.md?

y-guyon (approved):

y-guyon (approved):

PR #1881, opened 2023-12-14
Fix avifDiagnosticsPrintf doc

### LINKED ISSUE #631: The comment for avifDiagnostics seems to contradict the code in avifDiagnosticsPrintf
In include/avif/avif.h, the comment for `struct avifDiagnostics` says:

```
    // Upon receiving an error from any non-const libavif API call, if the toplevel structure used
    // in the API call (avifDecoder, avifEncoder) contains a diag member, this buffer may be
    // populated with a NULL-terminated, freeform error string explaining the most recent error in
    // more detail. It will be cleared at the beginning of every non-const API call.
```

Note that it says "the most recent error". But the code in `avifDiagnosticsPrintf()` does not overwrite any existing error message:

```
void avifDiagnosticsPrintf(avifDiagnostics * diag, const char * format, ...)
{
    if (!diag) {
        // It is possible this is NULL (e.g. calls to avifFileTypeIsCompatible())
        return;
    }
    if (*diag->error) {
        // There is already a detailed error set.
        return;
    }

    va_list args;
    va_start(args, format);
    vsnprintf(diag->error, AVIF_DIAGNOSTICS_ERROR_BUFFER_SIZE, format, args);
    diag->error[AVIF_DIAGNOSTICS_ERROR_BUFFER_SIZE - 1] = '\0';
    va_end(args);
}
```

Joe, is the comment wrong? Or should we remove the `if (*diag->error)` statement in `avifDiagnosticsPrintf()`?

vrabaud:
This fixes #631

[skip ci]

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -396,7 +396,7 @@ typedef struct avifDiagnostics
 {
     // Upon receiving an error from any non-const libavif API call, if the toplevel structure used
     // in the API call (avifDecoder, avifEncoder) contains a diag member, this buffer may be
-    // populated with a NULL-terminated, freeform error string explaining the first error in
+    // populated with a NULL-terminated, freeform error string explaining the first encountered error in
     // more detail. It will be cleared at the beginning of every non-const API call.
     //
     // Note: If an error string contains the "[Strict]" prefix, it means that you encountered an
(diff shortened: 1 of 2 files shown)


y-guyon (approved):

include/avif/avif.h
```diff
@@ -396,7 +396,7 @@ typedef struct avifDiagnostics
 {
     // Upon receiving an error from any non-const libavif API call, if the toplevel structure used
     // in the API call (avifDecoder, avifEncoder) contains a diag member, this buffer may be
-    // populated with a NULL-terminated, freeform error string explaining the most recent error in
+    // populated with a NULL-terminated, freeform error string explaining the first error in
```
wantehchang, line 399: Optional: Maybe add "encountered" after "first"?

wantehchang (approved):
LGTM. Thanks.

PR #1883, opened 2023-12-14
Fix gain map memory leak and null ptr dereference

y-guyon:


src/avif.c
```diff
@@ -1170,6 +1170,7 @@ void avifGainMapDestroy(avifGainMap * gainMap)
     if (gainMap->image) {
         avifImageDestroy(gainMap->image);
     }
+    avifRWDataFree(&gainMap->altICC);
```
maryla-uc, line 1173: uh, actually this should be moved to  avifGainMapDestroy() and the call below should be avifGainMapDestroy() as well

maryla-uc: sorry I misread the github diff this is actually what you're doing -_- nevermind!

maryla-uc (approved):

Issue #385, opened 2020-10-23
Finish coverage implementation


joedrago:
The goal of the coverage code was to get as close to 100% coverage of read.c as possible for testing. To finish this, two big things must be added:

* `aviftest`'s IO tests must have a mechanism where it is legal for a file to fail (perhaps a naming convention on the loaded file that it is known to fail)
* A ton of carefully crafted AVIFs that will test each path in read.c (this is the tough one)

vrabaud:
We are now at 40%. Coverage is complex as the current one cannot combine the gtest coverage, the executable coverage (executed through shell scripts) and the fuzzing coverage. So closing for now as this is an ever going progress.

PR #1885, opened 2023-12-18
Get coverage to cover all test files.

vrabaud:
To get the coverage, just run:
```
cmake ../ -DAVIF_ENABLE_COVERAGE=ON -DAVIF_CODEC_AOM=LOCAL -DAVIF_BUILD_TESTS=ON -DAVIF_LOCAL_GTEST=ON -DAVIF_BUILD_APPS=ON -DAVIF_CODEC_AOM_DECODE=ON -DAVIF_CODEC_AOM_ENCODE=ON
make -j avif_coverage
```

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -8,27 +8,25 @@ enable_testing()
 ################################################################################
 # C tests and tools
 
+set(COVERAGE_TARGETS)
+# Macro to register a test for coverage. The first argument is the target name.
+# Other arguments, like data path, can be added.
+macro(register_test_for_coverage TEST_NAME)
+    add_custom_target(
+        ${TEST_NAME}_coverage
+        COMMAND ${CMAKE_COMMAND} -E env "LLVM_PROFILE_FILE=${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}.profraw"
+                $<TARGET_FILE:${TEST_NAME}> ${ARGN}
+    )
+    list(APPEND COVERAGE_TARGETS ${TEST_NAME})
+endmacro()
+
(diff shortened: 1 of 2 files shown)


vrabaud:
Tests cover 40% of lines and 60% of functions.
https://github.com/AOMediaCodec/libavif/issues/385 wanted to get to 100% of coverage in read,c, we are at 40%. Then again, coverage only takes into account the tests that are executables, like gtest, not shell tests. The fuzzer is also not taken into account.

tests/CMakeLists.txt
```diff
@@ -8,27 +8,25 @@ enable_testing()
...
+# Macro to register a test for coverage. The first argument is the target name.
+# Other arguments, like data path, can be added.
+macro(register_test_for_coverage TEST_NAME)
+    add_custom_target(
+        ${TEST_NAME}_coverage
+        COMMAND ${CMAKE_COMMAND} -E env "LLVM_PROFILE_FILE=${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}.profraw"
+                $<TARGET_FILE:${TEST_NAME}> ${ARGN}
+    )
+    list(APPEND COVERAGE_TARGETS ${TEST_NAME})
+endmacro()
```
y-guyon, line 21: Note: This only applies to clang. This should still build because it only adds an unused environment target.

y-guyon (approved):

tests/CMakeLists.txt
vrabaud: Indeed. I now disable `AVIF_ENABLE_COVERAGE` when it is not available.
I also do not define those targets if disabled, to lighten the list of targets.

PR #1894, opened 2023-12-21
Get fuzztest tests to build for oss-fuzz

vrabaud:
This now works! I had missed two defines in the first attempt. Thx @DavidKorczynski  for the help!
The first commit is the cherry-picked original one, so that we keep track of what I had missed.


tests/oss-fuzz/build.sh
```diff
@@ -50,8 +51,9 @@ cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_DAV1D
       -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DAVIF_ENABLE_WERROR=OFF \
       -DAVIF_LOCAL_FUZZTEST=ON \
       -DAVIF_LOCAL_GTEST=ON -DAVIF_LOCAL_JPEG=ON -DAVIF_LOCAL_LIBSHARPYUV=ON \
-      -DAVIF_LIBYUV=LOCAL -DAVIF_LOCAL_ZLIBPNG=ON \
-      -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=OFF -DAVIF_ENABLE_FUZZTEST=ON ..
+      -DAVIF_LIBYUV=LOCAL \-DAVIF_LOCAL_ZLIBPNG=ON \
```
wantehchang, line 54: Delete the `\` in the middle of this line.

tests/oss-fuzz/build.sh
```diff
@@ -50,8 +51,9 @@ cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_DAV1D
...
+      -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=OFF -DAVIF_ENABLE_FUZZTEST=ON \
+      -DFUZZTEST_COMPATIBILITY_MODE=libfuzzer ..
```
wantehchang, line 56: Optional: Move the `..` to line 49, to follow "cmake" immediately.

tests/oss-fuzz/build.sh
```diff
@@ -41,6 +41,7 @@ ln -s $SRC/fuzztest $SRC/libavif/ext/fuzztest
 # build dependencies
 cd ext && bash aom.cmd && bash dav1d.cmd && bash googletest.cmd && bash libjpeg.cmd && \
       bash libsharpyuv.cmd && bash libyuv.cmd && bash zlibpng.cmd && cd ..
+export CXXFLAGS="${CXXFLAGS} -DFUZZTEST_COMPATIBILITY_MODE"
```
wantehchang, line 44: Just curious: This `FUZZTEST_COMPATIBILITY_MODE` doesn't need to be defined as "libfuzzer"?

tests/oss-fuzz/build.sh
```diff
@@ -19,23 +19,78 @@
...
 
 # build fuzzer
 $CXX $CXXFLAGS -std=c++11 -I../include \
     ../tests/oss-fuzz/avif_decode_fuzzer.cc -o $OUT/avif_decode_fuzzer \
     $LIB_FUZZING_ENGINE libavif.a ../ext/dav1d/build/src/libdav1d.a \
-    ../ext/libyuv/build/libyuv.a
+    ../ext/libyuv/build/libyuv.a ../ext/aom/build.libavif/libaom.a
+
+# build fuzztests
+# The following is taken from https://github.com/google/oss-fuzz/blob/31ac7244748ea7390015455fb034b1f4eda039d9/infra/base-images/base-builder/compile_fuzztests.sh#L59
```
y-guyon, line 67: Thankfully this is the same license as this file

y-guyon (approved):

tests/oss-fuzz/build.sh
```diff
@@ -19,23 +19,78 @@
...
+      bash libsharpyuv.cmd && bash libyuv.cmd && bash zlibpng.cmd && cd ..
+export CXXFLAGS="${CXXFLAGS} -DFUZZTEST_COMPATIBILITY_MODE"
 
 # build libavif
 mkdir build
 cd build
-cmake -G Ninja -DBUILD_SHARED_LIBS=OFF \
-    -DAVIF_CODEC_DAV1D=LOCAL -DAVIF_LIBYUV=LOCAL ..
+cmake -G Ninja -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_DAV1D=LOCAL \
+      -DAVIF_CODEC_AOM_DECODE=ON -DAVIF_CODEC_AOM_ENCODE=ON \
```
wantehchang, line 50: Set `AVIF_CODEC_AOM_DECODE` to `OFF`.

tests/oss-fuzz/build.sh
```diff
@@ -19,23 +19,78 @@
...
+      -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DAVIF_ENABLE_WERROR=OFF \
+      -DAVIF_LOCAL_FUZZTEST=ON \
+      -DAVIF_LOCAL_GTEST=ON -DAVIF_LOCAL_JPEG=ON -DAVIF_LOCAL_LIBSHARPYUV=ON \
+      -DAVIF_LIBYUV=LOCAL \-DAVIF_LOCAL_ZLIBPNG=ON \
+      -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=OFF -DAVIF_ENABLE_FUZZTEST=ON \
+      -DFUZZTEST_COMPATIBILITY_MODE=libfuzzer ..
+
 ninja
 
 # build fuzzer
```
wantehchang, line 60: Update this comment to say "build decode fuzzer"? Now we have multiple fuzzers.

tests/oss-fuzz/build.sh
vrabaud: Duh, done

tests/oss-fuzz/build.sh
vrabaud: Not here no, this is just to compile some code. This is what is done with bazel is oss-fuzz now: https://github.com/google/fuzztest/blob/9c6d264c9bd77c2dcff17424e7e40ccb49f8e09d/bazel/setup_configs.sh#L116
The CMake counterpart does need to be defined as libfuzzer though (it is the only one that is supported for now)

tests/oss-fuzz/build.sh
vrabaud: It is a google project so the license is good :)

tests/oss-fuzz/build.sh
vrabaud: I am doing it but we want to test all stable decoding libraries no ? I know libaom has its own fuzzers but we could kill two birds with one stone by letting it ON no?

tests/oss-fuzz/build.sh
vrabaud: Done

tests/oss-fuzz/build.sh
```diff
@@ -19,23 +19,78 @@
...
+
 ninja
 
-# build fuzzer
+# build decode fuzzer
 $CXX $CXXFLAGS -std=c++11 -I../include \
     ../tests/oss-fuzz/avif_decode_fuzzer.cc -o $OUT/avif_decode_fuzzer \
     $LIB_FUZZING_ENGINE libavif.a ../ext/dav1d/build/src/libdav1d.a \
-    ../ext/libyuv/build/libyuv.a
+    ../ext/libyuv/build/libyuv.a ../ext/aom/build.libavif/libaom.a
```
wantehchang, line 64: I just realized that we should link with the sharpyuv library. Is libavif.a still a combined archive library?

vrabaud: libavif.a is a combined archive and that target actually compiles and runs (it's the only fuzzer running on oss-fuzz for now). Once we are sure that the other ones are running there, this target will be converted to fuzztest and this hand-crafted compilation will be removed.

wantehchang: My previous comment wasn't clear. I meant that this target must link with the sharpyuv library when libavif.a is no longer a combined archive library, so we should link with the sharpyuv library now in preparation for that.

vrabaud: I agree, but we should be able to switch that binary in the next few days :) The builds were successful, we just need to make sure the reports are created properly (that will happen after today is over). So far, the fuzztest tests have not appeared on https://oss-fuzz.com/fuzzer-stats?project=libavif&fuzzer=libFuzzer&job=libfuzzer_asan_libavif&group_by=by-fuzzer

Issue #1896, opened 2023-12-22
"setlocale failed" on Windows


y-guyon:
#### Issue

`avifenc.exe 1.PNG 1.avif` fails when `avifenc.exe` is downloaded from https://ci.appveyor.com/project/louquillio/libavif/builds/48799163 (built with `Visual Studio 2022 Developer Command Prompt v17.8.2` according to the the log and run in `Visual Studio 2022 Developer Command Prompt v17.1.0` prompt):

```
setlocale failed
```

#### Expected behavior

Building and running `avifenc` locally worked, see https://github.com/AOMediaCodec/libavif/issues/1889#issuecomment-1866046537 (`Visual Studio 2022 Developer Command Prompt v17.1.0` prompt).

#### Environment

Windows Server 2016 Datacenter (Google Compute Engine Virtual Machine, Intel Xeon x64)

y-guyon:
@louquillio

kmilos:
[I guess one would need to check UCRT is being used rather than MSVCRT](https://learn.microsoft.com/en-us/cpp/porting/upgrade-your-code-to-the-universal-crt?view=msvc-170) in order for `setlocale` to work. Also, [Windows 10 1903 or later is a minimum requirement](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-170#utf-8-support) for UTF-8 locales.

vrabaud:
How can you figure that out in Windows given a binary? What is the equivalent of the Unix' ldd?
From the logs, it seems CMake is doing the default which should be UCRT right?

kmilos:
> How can you figure that out in Windows given a binary? What is the equivalent of the Unix' ldd?

Yep, either using MSYS2 tools equivalents (ldd/ntldd), or something like https://github.com/lucasg/Dependencies or https://www.dependencywalker.com/

UCRT should be the default since VS 2015 according to MS, but I haven't used it in a very long time...

wantehchang:
We can use the "dumpbin /dependents avifenc.exe" command.

Running that command on a local build I did on 2023-10-27, I get:

```
  Image has the following dependencies:

    KERNEL32.dll
    VCRUNTIME140D.dll
    ucrtbased.dll
```

Running that command on avifenc.exe I downloaded from AppVeyoer, I get:

```
  Image has the following dependencies:

    KERNEL32.dll
    VCRUNTIME140.dll
[...]

PR #1900, opened 2023-12-23
Windows: add a manifest to use UTF-8 code pages

wantehchang:
As of Windows Version 1903 (May 2019 Update), we can use the ActiveCodePage property in the fusion manifest for unpackaged apps to force a process to use UTF-8 as the process code page. This method requires no changes to the source code.

https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page#set-a-process-code-page-to-utf-8


kmilos:
Please add it via a .rc resource file instead so MinGW is supported as well? See e.g. how RStudio did it: https://github.com/rstudio/rstudio/blob/6f7af1eab02568be3de327190a96fc5dba2dc247/src/cpp/desktop/CMakeLists.txt#L308-L328

wantehchang:
I somehow forgot that Vincent wrote https://github.com/AOMediaCodec/libavif/pull/1752 to do this. I am sorry!

Miloš: Thank you for the reminder. I didn't forget your suggestion. In the first commit of this pull request, I was aiming to make the initial solution pass the tests on all platforms. In the second commit, I tried wrapping the manifest in a resource file. It apparently works in all variants of MinGW, whether they use UCRT or MSVCRT. So the only downside of this solution is that it requires Windows 10 version 1903 instead of 1803. But I think this is a worthwhile trade-off.

wantehchang:
I wrote this pull request originally when I looked into how to fix https://github.com/AOMediaCodec/libavif/issues/1896.

kmilos:
Thanks Wan-Teh for working on supporting a variety of Windows toolchains.

apps/avifdec.c
```diff
@@ -59,7 +53,7 @@ static void syntax(void)
     avifPrintVersions();
 }
 
-MAIN()
+int main(int argc, char * argv[])
```
vrabaud, line 56: I guess we use `char*argv[]` for C and `char**argv` for C++?

CMakeLists.txt
```diff
@@ -668,20 +668,27 @@ endif()
 
 if(AVIF_BUILD_APPS)
     add_executable(avifenc apps/avifenc.c)
+    if(WIN32)
```
vrabaud, line 671: Could be factorized in a macro. Then again. `tests/` could not access it. Will refactor later.

vrabaud:
Less code, no UTF-8 hack, LGTM !

apps/avifdec.c
wantehchang: I don't know of such a difference between C and C++. I always use `char * argv[]`, but in this pull request I took care to go back to the original code, which exists for avifdec.c, avifenc.c, and are_images_equal.cc.

I don't know why I chose `char ** argv` for avifgainmaputil.cc. I may have copied it from are_images_equal.cc without thinking.

CMakeLists.txt
wantehchang: Thank you. After we know this code works, we can consider using it with all executable programs, including the test programs.

apps/utf8.manifest
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
```
wantehchang, line 2: I copied this file from https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page#examples. The original file has this line:

```
  <assemblyIdentity type="win32" name="..." version="6.0.0.0"/>
```

I noticed that the name "..." and version "6.0.0.0" in this line look like placeholders. I think we are supposed to replace them with the name and version of avifdec, avifenc, and are_images_equal.

According to Microsoft documentation, the `assemblyIdentity` element is required in an application manifest:
https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests

However, I deleted this line and our tests still pass. So I suggest we try going without the `assemblyIdentity` element.

wantehchang (commented):
Vincent: Please review. Thanks!

When I wrote this pull request, I somehow forgot that you already wrote https://github.com/AOMediaCodec/libavif/pull/1752. I am really sorry.

apps/utf8.manifest
vrabaud: Thx for clarifying

vrabaud (approved):

PR #1904, opened 2023-12-25
pkgconf: add UAVIF_DLL for static linking with pkgconf

Biswa96:
allows for the usage of `pkgconf --static --cflags libavif' to produce
    the proper CFLAGS for static linking. Relies on pkgconf rather than pkg-config.

This is as similar as https://gitlab.com/AOMediaCodec/SVT-AV1/-/merge_requests/1911

@kmilos @1480c1 @mechakotik

### DIFF
--- a/libavif.pc.cmake
+++ b/libavif.pc.cmake
@@ -8,3 +8,4 @@ Description: Library for encoding and decoding .avif files
 Version: @PROJECT_VERSION@
 Libs: -L${libdir} -lavif
 Cflags: -I${includedir}@AVIF_PKG_CONFIG_EXTRA_CFLAGS@
+Cflags.private: -UAVIF_DLL


vrabaud:
Thx! If it's only for static linking, how about changing the code to:
```CMake
if(BUILD_SHARED_LIBS)
    target_compile_definitions(avif_obj PRIVATE AVIF_DLL AVIF_BUILDING_SHARED_LIBS)
    set(AVIF_PKG_CONFIG_EXTRA_CFLAGS " -DAVIF_DLL")
    set(AVIF_PKG_CONFIG_EXTRA_CFLAGS_PRIVATE "")
else()
    set(AVIF_PKG_CONFIG_EXTRA_CFLAGS "")
    set(AVIF_PKG_CONFIG_EXTRA_CFLAGS_PRIVATE " -UAVIF_DLL")
endif()
```

Biswa96:
> If it's only for static linking, how about changing the code to:

Yeah, this is a fix for static linking but the `Cflags.private` is only used when `-static` option is passed to pkgconf. So, having `Cflags.private` in pkgconfig file for shared library would not cause any issue. The svt-av1 project also has same change.

This is required in msys2/mingw packages. The projects provides both shared and static library in one package.

vrabaud:
Why even define a variable then? Just put ` -UAVIF_DLL` right in `libavif.pc.cmake` no?

Biswa96:
> Just put ` -UAVIF_DLL` right in `libavif.pc.cmake` no?

Certainly, it is indeed a possibility. My approach was based on the definition of the `AVIF_PKG_CONFIG_EXTRA_CFLAGS` variable. If you prefer, I can remove the newly created variable as per your request.

vrabaud:
Yes please, remove it: there is a `AVIF_PKG_CONFIG_EXTRA_CFLAGS` because it can be overwritten when `BUILD_SHARED_LIBS` is on. But `AVIF_PKG_CONFIG_EXTRA_CFLAGS_PRIVATE` is only set once so we might as well set it for good in `libavif.pc.cmake`.

kmilos:
Please leave the private variable in even when building shared. In MSYS2 we shop both static and shared libs, but only one .pc (from the shared build run).

vrabaud:
@kmilos , so it's ok to only have it in `libavif.pc.cmake` then, no need for `AVIF_PKG_CONFIG_EXTRA_CFLAGS_PRIVATE` right?

Biswa96:
I have added `-UAVIF_DLL` in libavif.pc.cmake.

PR #1905, opened 2023-12-28
pkgconf: Add private libraries in Requires.private for static linking

Biswa96:
@kmilos @1480c1 @mechakotik

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,8 @@ option(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
 )
 option(AVIF_ENABLE_EXPERIMENTAL_AVIR "Enable experimental reduced header" OFF)
 
+set(AVIF_PKG_CONFIG_EXTRA_REQUIRES_PRIVATE "")
+
 function(set_local_or_system_option VAR DEFAULT TEXT)
     # Deal with the older way of setting options.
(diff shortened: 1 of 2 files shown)


vrabaud:
Thx. How do you choose the libs to link to? Why not libyuv?

Biswa96:
> How do you choose the libs to link to?

* For shared linking:

```
$ pkgconf -libs libavif
-lavif
```

* For static linking:

```
$ pkgconf -libs -static libavif
-lavif -lsharpyuv -ldav1d -lrav1e -lkernel32 -ladvapi32 -lbcrypt -lkernel32 -lntdll -luserenv -lws2_32 -lkernel32 -lws2_32 -lkernel32 -lSvtAv1Enc -laom -lm
```

> Why not libyuv?

libyuv does not provide any pkgconfig file. So, it can not be added in Requires.private but could be added in Libs.private. I was not sure about the later one - will be added in next commit.

vrabaud:
Thx, makes sense.

fdintino:
These libraries should be omitted from Requires.private and Libs.private if they are local, since in that case they get included in the archive file. If someone else doesn't get around to addressing that I'll open a PR myself next week to fix that.

Biswa96:
Sure, please create a PR. I did not find the conditions for local dependencies.

PR #1915, opened 2024-01-08
Disable fuzztest flags for coverage.

vrabaud:
This should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65340

### DIFF
--- a/tests/oss-fuzz/build.sh
+++ b/tests/oss-fuzz/build.sh
@@ -68,7 +68,7 @@ $CXX $CXXFLAGS -std=c++11 -I../include \
     ../ext/libyuv/build/libyuv.a ../ext/aom/build.libavif/libaom.a
 
 # Restrict fuzztest tests to the only compatible fuzz engine: libfuzzer.
-if [ "$FUZZING_ENGINE" == "libfuzzer" ]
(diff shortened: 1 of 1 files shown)


tests/oss-fuzz/build.sh
```diff
@@ -41,19 +41,25 @@ ln -s $SRC/fuzztest $SRC/libavif/ext/fuzztest
 # build dependencies
 cd ext && bash aom.cmd && bash dav1d.cmd && bash googletest.cmd && bash libjpeg.cmd && \
       bash libsharpyuv.cmd && bash libyuv.cmd && bash zlibpng.cmd && cd ..
-export CXXFLAGS="${CXXFLAGS} -DFUZZTEST_COMPATIBILITY_MODE"
 
 # build libavif
 mkdir build
 cd build
+export CXXFLAGS="${CXXFLAGS}"
```
y-guyon, line 48: Is this line doing anything?

y-guyon (approved):

tests/oss-fuzz/build.sh
vrabaud: Removed

PR #1916, opened 2024-01-08
Encode an anim instead of the same frame X times

y-guyon:


tests/gtest/avif_fuzztest_helpers.cc
```diff
@@ -69,6 +69,41 @@ ImagePtr CreateAvifImage16b(size_t width, size_t height, int depth,
                          reinterpret_cast<const uint8_t*>(samples.data()));
 }
 
+std::vector<ImagePtr> CreateAvifAnim8b(size_t num_frames, size_t width,
+                                       size_t height,
+                                       avifPixelFormat pixel_format,
+                                       bool has_alpha,
+                                       const std::vector<uint8_t>& samples) {
+  std::vector<ImagePtr> frames;
```
vrabaud, line 77: Maybe encapsulate the function in template in the .cc (with depth as a parameter and called here with `/*depth*/=8`)? Just in case the function becomes more complex? Not necessary,

tests/gtest/avif_fuzztest_helpers.h
```diff
@@ -92,20 +108,67 @@ inline auto ArbitraryAvifImage16b() {
...
+inline auto ArbitraryAvifAnim8b() {
+  return fuzztest::FlatMap(
+      [](size_t num_frames, size_t width, size_t height,
+         avifPixelFormat pixel_format, bool has_alpha) {
+        return fuzztest::Map(
+            CreateAvifAnim8b, fuzztest::Just(num_frames), fuzztest::Just(width),
+            fuzztest::Just(height), fuzztest::Just(pixel_format),
+            fuzztest::Just(has_alpha),
+            fuzztest::Arbitrary<std::vector<uint8_t>>().WithSize(GetNumSamples(
+                num_frames, width, height, pixel_format, has_alpha)));
```
vrabaud, line 131: That will force all frames to have alpha or none to have alpha. Maybe we should have a vector of bools?
Same for width and height but I actually do not know whether AVIF supports animations with frames of different size: maybe you can run this test locally to see if it passes with frames of different size?

vrabaud (approved):

y-guyon (commented):
As discussed, let's wait until internal b/avif_fuzztest_enc_dec_anim are verified as fixed internally, or create another target if it takes too much time.

vrabaud:
With libaom updated to 3.8.1, we should be good to merge this right?

wantehchang:
Vincent: I didn't review this pull request. Does this pull request need a bug fix in libaom 3.8.1?

vrabaud:
I thought all internal bugs depending on this fuzzer had been fixed in 3.8.1 but there is still b/319140742. Once it is fixed, we can change this fuzzer.

Issue #1889, opened 2023-12-20
libpng warning: iCCP: known incorrect sRGB profile


atbsi:
libpng warning: iCCP: known incorrect sRGB profile

I've come across an image that won't convert correctly when batch converting images.
I'm not sure if it's a bug or if this is a normal reaction.
Note that this PNG image converts properly to .jxl.

win 10 64
Version: 1.0.3 (dav1d [dec]:1.2.1-0-g8a6f054, aom [enc/dec]:3.8.0)
libyuv : available (1864)

CL:
avifenc.exe  "D:\1.PNG" "D:\out.avif"

https://github.com/atbsi/pic/releases/download/1.1/1.PNG

y-guyon:
Thank you for the report.

`avifenc -q 90 -s 9 1.IMG 1.avif` (at 995f89f) results in:

![1q90 avif](https://github.com/AOMediaCodec/libavif/assets/51164219/a5e0ce0d-4a04-4ad4-92e5-7f47b60e9e8f)

I cannot perceive the visual difference with the original PNG when viewing both in Chrome.  
Please note that it is 12-bit encoded because the input PNG is 16-bit.

> I've come across an image that won't convert correctly when batch converting images.

What do you mean by "won't convert correctly"?  
If this is only about the warning message, "libpng warning: iCCP: known incorrect sRGB profile" is printed by libpng, a dependency of libavif used to read PNG files. I did not search whether this can be silenced, and even so I am not sure it should be.

> Note that this PNG image converts properly to .jxl.

Using `cjxl`? I wonder what library `libjxl` uses as a dependence to read PNG files.

atbsi:
请使用我的所使用的命令行来重现。我没有使用任何参数。(当然，我测试使用你的参数后，转换正常了。)
"无法正常转换"，这也需要解释？我有几千张图片都用CLI转换生成了avif图片，只有这张没生成，并且报错，我反馈"无法正确转换"，这有任何难以理解的地方吗？
我说这个图片可以转换为 .jxl，是要说明这个图片本身没有品题，其他软件可以正常读取转换它。

y-guyon:
Translated with Google Translate:

> Please use the command line I used to reproduce. I'm not using any parameters. (Of course, after I tested using your parameters, the conversion was normal.)
[...]

PR #1921, opened 2024-01-10
appveyor.yml: only build on v1.0.x branch

jzern:
This should hopefully avoid creating unstable artifacts from PRs.

See issue #1889.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,8 @@
 image: Visual Studio 2022
 configuration: Release
+branches:
+  only:
+    - v1.0.x
 skip_commits:
   files:
     - '.github/*'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,8 +1,6 @@
 image: Visual Studio 2022
 configuration: Release
-branches:
-  only:
-    - v1.0.x
+skip_non_tags: true
 skip_commits:
   files:
     - '.github/*'


appveyor.yml
```diff
@@ -1,5 +1,8 @@
 image: Visual Studio 2022
 configuration: Release
+branches:
+  only:
+    - v1.0.x
```
jzern, line 5: `skip_branch_with_pr` [1] sounds like it might be helpful, but reading through some comments it will execute a build for the first push [2]. The option used here limits the branches considered [3].

I don't know if there's a way to conditionally create artifacts. The deploy section [4] does allow restricting it to a branch, but I don't know where that's actually deploying to. I don't see artifacts as part of this project's release tag.

[1]: https://www.appveyor.com/docs/appveyor-yml/
[2]: https://help.appveyor.com/discussions/questions/18437-skip_branch_with_pr-setting-doesnt-work
[3]: https://www.appveyor.com/docs/branches/#white--and-blacklisting
[4]: https://www.appveyor.com/docs/deployment/github/

jzern:
This is just a guess after a quick read through the docs.

appveyor.yml
vrabaud: How about we only build on tags by setting `skip_non_tags: true` ?
Also, the deploy sections is filled but nothing is set on the download page for the releases (we only have the tar.gz of the sources).
The doc for config is here https://www.appveyor.com/docs/deployment/github/ but it all seems fine.

y-guyon (approved):

appveyor.yml
jzern: > How about we only build on tags by setting `skip_non_tags: true` ?

That sounds like it will work.

> Also, the deploy sections is filled but nothing is set on the download page for the releases (we only have the tar.gz of the sources). The doc for config is here https://www.appveyor.com/docs/deployment/github/ but it all seems fine.

It wasn't clear where those were deployed to. I'm guessing `APPVEYOR_REPO_TAG: true` limited the builds, but I'm not sure why the tags were missed (if I understand that flag correctly).

After https://github.com/AOMediaCodec/libavif/pull/1923 and the follow up work is completed we can just delete this file. The older binaries on Appveyor will be deleted after 90 days.

PR #1923, opened 2024-01-10
Add ci-windows-artifacts

y-guyon:


y-guyon:
See https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts for downloading `avifenc.exe` and `avifdec.exe`.

.github/workflows/ci-windows-artifacts.yml
```diff
@@ -0,0 +1,83 @@
...
+      working-directory: ./ext
+      run: ./aom.cmd
+    - name: Build dav1d
+      if: steps.cache-ext.outputs.cache-hit != 'true'
+      working-directory: ./ext
+      run: ./dav1d.cmd
+    - name: Build libyuv
+      if: steps.cache-ext.outputs.cache-hit != 'true'
+      working-directory: ./ext
+      run: ./libyuv.cmd
```
wantehchang, line 49: Thank you for adding this CI workflow.

We should use clang-cl to build libyuv. The assembly code in libyuv is written in the GCC inline assembly format. Visual C++ doesn't support GCC inline assembly.

In appveyor.yml this is done as follows:

```
 - set "CC=clang-cl" && set "CXX=clang-cl"
 - libyuv.cmd
 - set "CC=" && set "CXX="
```

y-guyon: I was wondering why these lines were there. Thanks for the explanation.

Done in https://github.com/AOMediaCodec/libavif/pull/1930.

PR #1929, opened 2024-01-11
Fix GitHub workflow cache key conflict

y-guyon:


vrabaud (approved):

.github/workflows/ci-windows-artifacts.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
       with:
         path: ext
-        key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}
+        key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}-somedeps
```
wantehchang, line 30: We can also consider building everything from scratch in the Windows artifacts workflow.

y-guyon: Since it will not save computing resources with the current setup, I do not think this is worth it.

PR #1930, opened 2024-01-11
Use clang-cl to build libyuv

y-guyon:


vrabaud (approved):

.github/workflows/ci-windows-artifacts.yml
```diff
@@ -46,7 +43,18 @@ jobs:
     - name: Build libyuv
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
-      run: ./libyuv.cmd
+      # Use clang-cl to build libyuv. The assembly code in libyuv is written in the
+      # GCC inline assembly format. Visual C++ doesn't support GCC inline assembly.
+      run: >
+        set "CC=clang-cl"
+
+        set "CXX=clang-cl"
```
wantehchang, line 51: The default shell for GitHub actions on Windows is the PowerShell:
https://github.blog/changelog/2019-10-17-github-actions-default-shell-on-windows-runners-is-changing-to-powershell/

PowerShell seems to use a different syntax to set environment variables. (The `set FOO=bar` syntax is for the Command Prompt.)

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.4#using-the-variable-syntax

I wonder how we can verify clang-cl is used.

.github/workflows/ci-windows-artifacts.yml
```diff
@@ -46,7 +43,18 @@ jobs:
     - name: Build libyuv
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
-      run: ./libyuv.cmd
+      # Use clang-cl to build libyuv. The assembly code in libyuv is written in the
+      # GCC inline assembly format. Visual C++ doesn't support GCC inline assembly.
+      run: >
+        set "CC=clang-cl"
```
wantehchang, line 49: Nit: The quotes can be omitted. I remember the quotes are used in appveyor.yml because we connect two `set` commands with `&&`:

```
set "CC=clang-cl" && set "CXX=clang-cl"
```

wantehchang (dismissed):

.github/workflows/ci-windows-artifacts.yml
wantehchang: I found that this isn't working.

When it works, you should see something like this from the cmake command in libyuv.cmd:

```
-- The C compiler identification is Clang 16.0.5 with MSVC-like command-line
-- The CXX compiler identification is Clang 16.0.5 with MSVC-like command-line
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe - skipped
```

But in the build log, I see

```
-- The C compiler identification is MSVC 19.37.32826.1
-- The CXX compiler identification is MSVC 19.37.32826.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.37.32822/bin/HostX64/x64/cl.exe - skipped
```

wantehchang: I wonder if we should just add a `shell: cmd` line to this step.

.github/workflows/ci-windows-artifacts.yml
wantehchang: OK, in PowerShell these two lines should read:

```
$Env:CC = 'clang-cl'

$Env:CXX = 'clang-cl'
```

.github/workflows/ci-windows-artifacts.yml
```diff
@@ -46,7 +43,18 @@ jobs:
...
+
+        set "CXX=clang-cl"
+
+        ./libyuv.cmd
+
+        set "CC="
+
+        set "CXX="
```
wantehchang, line 57: In PowerShell, these two lines should read:

```
$Env:CC = ''

$Env:CXX = ''
```

We may be able to omit these if the environment variables affect only the current step.

.github/workflows/ci-windows-artifacts.yml
y-guyon: It seems to work now, the log shows

```
-- The C compiler identification is Clang 16.0.6 with MSVC-like command-line
-- The CXX compiler identification is Clang 16.0.6 with MSVC-like command-line
```

.github/workflows/ci-windows-artifacts.yml
y-guyon: Thank you for the PowerShell syntax research.

> We may be able to omit these if the environment variables affect only the current step.

It works as is so I am keeping them.

.github/workflows/ci-windows-artifacts.yml
wantehchang: Yes, this is using clang-cl (specifically, the clang-cl shipped with Visual Studio 2022). Thanks!

PR #1931, opened 2024-01-11
Fix AVM header parsing

vrabaud:


wantehchang (approved):
LGTM.

I remember we only need to parse the AV1 sequence header OBU when an AVIF file doesn't have a colr box of the nclx type. If that's the case, perhaps in AVIF with AV2 we should mandate a colr box of the nclx type.

PR #1932, opened 2024-01-11
Convert color conversion internals to matrices stored as 1D array

vrabaud:
This is to prevent passing M[3][3] and make sure they are const, cf https://c-faq.com/ansi/constmismatch.html

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -104,14 +104,14 @@ void avifColorPrimariesComputeYCoeffs(avifColorPrimaries colorPrimaries, float c
 AVIF_NODISCARD avifBool avifColorPrimariesComputeRGBToXYZD50Matrix(avifColorPrimaries colorPrimaries, double coeffs[9]);
 // Computes a row-major conversion matrix (stored as an array) from XYZ with a D50 white point to RGB.
 AVIF_NODISCARD avifBool avifColorPrimariesComputeXYZD50ToRGBMatrix(avifColorPrimaries colorPrimaries, double coeffs[9]);
-// Computes the RGB->RGB conversion matrix to convert from one set of RGB primaries to another.
+// Computes the RGB->RGB conversion matrix (stored as an array) to convert from one set of RGB primaries to another.
 AVIF_NODISCARD avifBool avifColorPrimariesComputeRGBToRGBMatrix(avifColorPrimaries srcColorPrimaries,
                                                                 avifColorPrimaries dstColorPrimaries,
-                                                                double coeffs[3][3]);
+                                                                double coeffs[9]);
 // Converts the given linear RGB pixel from one color space to another using the provided coefficients.
 // The coefficients can be obtained with avifColorPrimariesComputeRGBToRGBMatrix().
 // The output values are not clamped and may be < 0 or > 1.
-void avifLinearRGBConvertColorSpace(float rgb[4], const double coeffs[3][3]);
+void avifLinearRGBConvertColorSpace(float rgb[4], const double coeffs[9]);
 
 #define AVIF_ARRAY_DECLARE(TYPENAME, ITEMSTYPE, ITEMSNAME) \
     typedef struct TYPENAME                                \
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -100,10 +100,10 @@ avifTransferFunction avifTransferCharacteristicsGetLinearToGammaFunction(avifTra
 // Computes the RGB->YUV conversion coefficients kr, kg, kb, such that Y=kr*R+kg*G+kb*B.
 void avifColorPrimariesComputeYCoeffs(avifColorPrimaries colorPrimaries, float coeffs[3]);
 
-// Computes a conversion matrix from RGB to XYZ with a D50 white point.
-AVIF_NODISCARD avifBool avifColorPrimariesComputeRGBToXYZD50Matrix(avifColorPrimaries colorPrimaries, double coeffs[3][3]);
-// Computes a conversion matrix from XYZ with a D50 white point to RGB.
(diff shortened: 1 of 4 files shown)


src/colrconvert.c
```diff
@@ -160,13 +173,23 @@ avifBool avifColorPrimariesComputeRGBToRGBMatrix(avifColorPrimaries srcColorPrim
...
+    avifMatMul(xyzToDstRGB, srcRGBToXYZ, coeffsTmp);
+    coeffs[0][0] = coeffsTmp[0];
+    coeffs[0][1] = coeffsTmp[1];
+    coeffs[0][2] = coeffsTmp[2];
+    coeffs[1][0] = coeffsTmp[3];
+    coeffs[1][1] = coeffsTmp[4];
+    coeffs[1][2] = coeffsTmp[5];
+    coeffs[2][0] = coeffsTmp[6];
+    coeffs[2][1] = coeffsTmp[7];
+    coeffs[2][2] = coeffsTmp[8];
```
wantehchang, line 192: This kind of conversion is not ideal.

include/avif/internal.h
```diff
@@ -100,10 +100,10 @@ avifTransferFunction avifTransferCharacteristicsGetLinearToGammaFunction(avifTra
...
 void avifColorPrimariesComputeYCoeffs(avifColorPrimaries colorPrimaries, float coeffs[3]);
 
-// Computes a conversion matrix from RGB to XYZ with a D50 white point.
-AVIF_NODISCARD avifBool avifColorPrimariesComputeRGBToXYZD50Matrix(avifColorPrimaries colorPrimaries, double coeffs[3][3]);
-// Computes a conversion matrix from XYZ with a D50 white point to RGB.
-AVIF_NODISCARD avifBool avifColorPrimariesComputeXYZD50ToRGBMatrix(avifColorPrimaries colorPrimaries, double coeffs[3][3]);
+// Computes a row-major conversion matrix (stored as an array) from RGB to XYZ with a D50 white point.
+AVIF_NODISCARD avifBool avifColorPrimariesComputeRGBToXYZD50Matrix(avifColorPrimaries colorPrimaries, double coeffs[9]);
+// Computes a row-major conversion matrix (stored as an array) from XYZ with a D50 white point to RGB.
+AVIF_NODISCARD avifBool avifColorPrimariesComputeXYZD50ToRGBMatrix(avifColorPrimaries colorPrimaries, double coeffs[9]);
```
wantehchang, line 106: Why didn't you change line 110 to `double coeffs[9]`?

vrabaud: Simply to minimize the change

src/colrconvert.c
vrabaud: Done

y-guyon (approved):

wantehchang:
Thank you for writing this pull request. The new code is not as readable as the original code, so I think we should ask in the C/C++ discussion group at Google about this problem. The two dimensional array is not exactly the same as https://c-faq.com/ansi/constmismatch.html. Hopefully there is a better solution.

I am also wondering why none of the compilers we use regularly warn about this. Is it because we do not enable a certain warning option?

wantehchang:
I found a Stack Overflow question about this issue:
https://stackoverflow.com/questions/28062095/pass-a-two-dimensional-array-to-a-function-of-constant-parameter

tests/gtest/avifcolrconverttest.cc
```diff
@@ -15,36 +15,38 @@ namespace {
...
   // http://brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html
   const double kEpsilon = 0.00015;
   ExpectMatrixNear(coeffs,
-                   {{{0.4360747, 0.3850649, 0.1430804},
-                     {0.2225045, 0.7168786, 0.0606169},
-                     {0.0139322, 0.0971045, 0.7141733}}},
+                   {{
+                       0.4360747, 0.3850649, 0.1430804,  // row 0
+                       0.2225045, 0.7168786, 0.0606169,  // row 1
+                       0.0139322, 0.0971045, 0.7141733   // row 2
```
wantehchang, line 43: If the original still works, the original code seems to be a better initializer for `std::array<std::array<double, 3>, 3>` (see line 19).

This comment also applies to the second arguments for the `ExpectMatrixNear()` calls below.

wantehchang (approved):

wantehchang:
Vincent: Removing the const from the function prorotypes is fine by me.

My suggestion was to wrap the two-dimensional array in a struct, e.g.,

```
typedef struct {
    double x[3][3];
} avifMat;
```

But this requires more changes to the code, because `coeffs[i][j]` will become `coeffs.d[i][j]`.

vrabaud:
Right, but the problem would be the same: a const avifMat offers no const guarantee on the data in `x`.

vrabaud:
Closing in favor of https://github.com/AOMediaCodec/libavif/pull/1946

wantehchang:
> Right, but the problem would be the same: a const avifMat offers no const guarantee on the data in `x`.

Are you sure? I guess you may be thinking of a struct like `avifRWData`:

```
typedef struct avifRWData
{
    uint8_t * data;
    size_t size;
} avifRWData;
```

then the following is allowed:

```
    uint8_t data[16];
    const avifRWData rw = { data, sizeof(data) };
    rw.data[0] = 0;
```

On the other hand, if we have a struct like the following:

```
typedef struct FixedSizedData16
{
    uint8_t data[16];
} FixedSizedData16;
```

then the following is not allowed:

```
    const FixedSizedData16 fixed;
    fixed.data[0] = 0; 
```

vrabaud:
I thought the same issue would pop with 2d but it indeed works.

PR #1933, opened 2024-01-12
appveyor.yml: A test to build libaom with clang-cl

wantehchang:


wantehchang:
This is just an experiment.

PR #1937, opened 2024-01-13
Improve commands for Build everything from scratch

wantehchang:
Add libsharpyuv.

Also use the old AVIF_LOCAL_AOM, AVIF_LOCAL_LIBYUV, and AVIF_LOCAL_LIBSHARPYUV cmake variables because libavif v1.0.3 doesn't support the new -DAVIF_CODEC_AOM=LOCAL method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,10 +95,11 @@ git clone -b v1.0.3 https://github.com/AOMediaCodec/libavif.git
 cd libavif/ext
 ./aom.cmd
 ./libyuv.cmd
+./libsharpyuv.cmd
 ./libjpeg.cmd
 ./zlibpng.cmd
 cd ..
-cmake -S . -B build -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=LOCAL -DAVIF_LIBYUV=LOCAL -DAVIF_LOCAL_JPEG=ON -DAVIF_LOCAL_ZLIBPNG=ON -DAVIF_BUILD_APPS=ON
+cmake -S . -B build -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON -DAVIF_LOCAL_LIBYUV=ON -DAVIF_LOCAL_LIBSHARPYUV=ON -DAVIF_LOCAL_JPEG=ON -DAVIF_LOCAL_ZLIBPNG=ON -DAVIF_BUILD_APPS=ON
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -95,10 +95,11 @@ git clone -b v1.0.3 https://github.com/AOMediaCodec/libavif.git
 cd libavif/ext
 ./aom.cmd
 ./libyuv.cmd
+./libsharpyuv.cmd
 ./libjpeg.cmd
 ./zlibpng.cmd
 cd ..
-cmake -S . -B build -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=LOCAL -DAVIF_LIBYUV=LOCAL -DAVIF_LOCAL_JPEG=ON -DAVIF_LOCAL_ZLIBPNG=ON -DAVIF_BUILD_APPS=ON
+cmake -S . -B build -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON -DAVIF_LOCAL_LIBYUV=ON -DAVIF_LOCAL_LIBSHARPYUV=ON -DAVIF_LOCAL_JPEG=ON -DAVIF_LOCAL_ZLIBPNG=ON -DAVIF_BUILD_APPS=ON
```
wantehchang, line 102: I tested this command by removing the -dev packages for libaom, zlib, libpng, etc. on my Linux computer. I will test this command on my Windows laptop when I get home.

The cmake command at line 85 has the same issue with libavif v1.0.3: `-DAVIF_CODEC_AOM=SYSTEM` is treated the same as `-DAVIF_CODEC_AOM=ON`.

vrabaud: `-DAVIF_CODEC_AOM=ON` should be proposed to the v1.0.x branch then

wantehchang: This PR is for the main branch. The build instructions are not in the v1.0.x branch.

Should I remove the `-b v1.0.3` option from the `git clone` commands in line 83 and line 94?

y-guyon: > Should I remove the `-b v1.0.3` option from the `git clone` commands in line 83 and line 94?

Sorry, I forgot about that part. This PR can be submitted as is then, and we should use the new flag syntax upon the next release.

y-guyon (approved):

README.md
wantehchang: Thanks for the review. A good solution is to provide build instructions for both the latest release and the main branch.

PR #1938, opened 2024-01-15
Add imageSizeLimit to avif_fuzztest_dec_incr

y-guyon:
to avoid out-of-memory issues.

[BUG=oss-fuzz:65738](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65738)

### DIFF
--- a/tests/gtest/avif_fuzztest_dec_incr.cc
+++ b/tests/gtest/avif_fuzztest_dec_incr.cc
@@ -61,6 +61,9 @@ void DecodeIncr(const std::string& arbitrary_bytes, bool is_persistent,
   DecoderPtr decoder(avifDecoderCreate());
   ASSERT_NE(decoder.get(), nullptr);
   avifDecoderSetIO(decoder.get(), &io);
+  // OSS-Fuzz limits the allocated memory to 2560 MB.
+  decoder->imageSizeLimit =
+      2560u * 1024 * 1024 / AVIF_MAX_AV1_LAYER_COUNT / sizeof(uint16_t);
 
   if (avifDecoderRead(decoder.get(), reference.get()) == AVIF_RESULT_OK) {
     // Avoid timeouts by discarding big images decoded many times.


vrabaud:
Maybe worth having it as a function usable by other fuzztest tests.

vrabaud (approved):

tests/gtest/avif_fuzztest_dec_incr.cc
```diff
@@ -61,6 +61,9 @@ void DecodeIncr(const std::string& arbitrary_bytes, bool is_persistent,
   DecoderPtr decoder(avifDecoderCreate());
   ASSERT_NE(decoder.get(), nullptr);
   avifDecoderSetIO(decoder.get(), &io);
+  // OSS-Fuzz limits the allocated memory to 2560 MB.
+  decoder->imageSizeLimit =
+      2560u * 1024 * 1024 / AVIF_MAX_AV1_LAYER_COUNT / sizeof(uint16_t);
```
wantehchang, line 66: I think we should divide by `AVIF_PLANE_COUNT_YUV` (3) instead of `AVIF_MAX_AV1_LAYER_COUNT` (4).

Note that `allowProgressive` is false (the default), so even if we get a layered image, it will be decoded into a single frame.

y-guyon: > I think we should divide by `AVIF_PLANE_COUNT_YUV` (3) instead of `AVIF_MAX_AV1_LAYER_COUNT` (4).

Alpha can also be allocated, right?

> Note that `allowProgressive` is false (the default), so even if we get a layered image, it will be decoded into a single frame.

Good to know, it could be a comment.

wantehchang: You are right. Then it's the name of the constant `AVIF_MAX_AV1_LAYER_COUNT` that's wrong. We can use 4 directly or use `(AVIF_PLANE_COUNT_YUV + 1)`, with a comment saying "Y, U, V, and alpha planes".

PR #1940, opened 2024-01-15
Fix avif_fuzztest_enc_dec_anim for AVIF_ADD_IMAGE_FLAG_SINGLE

vrabaud:
BUG=oss-fuzz:65696

### DIFF
--- a/tests/gtest/avif_fuzztest_enc_dec_anim.cc
+++ b/tests/gtest/avif_fuzztest_enc_dec_anim.cc
@@ -35,12 +35,27 @@ void EncodeDecodeAnimation(ImagePtr frame,
   int total_duration = 0;
 
   // Encode.
-  for (const FrameOptions& options : frame_options) {
+  for (size_t i = 0; i < frame_options.size(); ++i) {
+    const FrameOptions& options = frame_options[i];
     total_duration += options.duration;
     const avifResult result = avifEncoderAddImage(
         encoder.get(), frame.get(), options.duration, options.flags);
-    ASSERT_EQ(result, AVIF_RESULT_OK)
-        << avifResultToString(result) << " " << encoder->diag.error;
+    // AVIF_ADD_IMAGE_FLAG_SINGLE can only be used on one image.
+    bool should_not_use_image_flag_single =
+        (frame_options[0].flags & AVIF_ADD_IMAGE_FLAG_SINGLE) && i > 0;
+    // AVIF_ADD_IMAGE_FLAG_SINGLE can only be used on the first image.
+    // It also cannot be set for layered images.
+    should_not_use_image_flag_single |=
+        (options.flags & AVIF_ADD_IMAGE_FLAG_SINGLE) &&
+        (i > 0 || encoder->extraLayerCount > 0);
+    if (should_not_use_image_flag_single) {
+      ASSERT_NE(result, AVIF_RESULT_OK)
+          << avifResultToString(result) << " " << encoder->diag.error;
+      return;
+    } else {
+      ASSERT_EQ(result, AVIF_RESULT_OK)
+          << avifResultToString(result) << " " << encoder->diag.error;
+    }
   }
   AvifRwData encoded_data;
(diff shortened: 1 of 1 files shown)


tests/gtest/avif_fuzztest_enc_dec_anim.cc
```diff
@@ -35,12 +35,27 @@ void EncodeDecodeAnimation(ImagePtr frame,
...
+    // AVIF_ADD_IMAGE_FLAG_SINGLE can only be used on one image.
+    bool should_not_use_image_flag_single =
+        (frame_options[0].flags & AVIF_ADD_IMAGE_FLAG_SINGLE) && i > 0;
+    // AVIF_ADD_IMAGE_FLAG_SINGLE can only be used on the first image.
+    // It also cannot be set for layered images.
+    should_not_use_image_flag_single |=
+        (options.flags & AVIF_ADD_IMAGE_FLAG_SINGLE) &&
+        (i > 0 || encoder->extraLayerCount > 0);
+    if (should_not_use_image_flag_single) {
+      ASSERT_NE(result, AVIF_RESULT_OK)
```
y-guyon, line 52: Optional: Check EQ with ENCODE_COLOR_FAILED or INVALID_ARG

y-guyon (approved):

tests/gtest/avif_fuzztest_enc_dec_anim.cc
vrabaud: Done

tests/gtest/avif_fuzztest_enc_dec_anim.cc
```diff
@@ -35,10 +35,26 @@ void EncodeDecodeAnimation(ImagePtr frame,
...
+      return;
+    }
+    // AVIF_ADD_IMAGE_FLAG_SINGLE can only be used on the first image.
+    // It also cannot be set for layered images.
+    if ((options.flags & AVIF_ADD_IMAGE_FLAG_SINGLE) &&
+        (i > 0 || encoder->extraLayerCount > 0)) {
+      ASSERT_EQ(result, AVIF_RESULT_INVALID_ARGUMENT)
+          << avifResultToString(result) << " " << encoder->diag.error;
+      return;
+    }
```
wantehchang, line 56: Vincent: At line 105, we have:

```
            fuzztest::StructOf<FrameOptions>(
                /*duration=*/fuzztest::InRange<uint64_t>(1, 10),
                // BitFlagCombinationOf() requires non zero flags as input.
                // AVIF_ADD_IMAGE_FLAG_NONE (0) is implicitly part of the set.
                fuzztest::BitFlagCombinationOf<avifAddImageFlags>(
                    {AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME})))
```

How can `AVIF_ADD_IMAGE_FLAG_SINGLE` possibly be set in `options.flags`?

wantehchang: When I ran the fuzz test with the reproducer test case, I got the following `frame_options`:

```
argument 1: {{2, 7}, {3, 0}}
```

7 is an invalid value for `avifAddImageFlags`, which is a combination of the following bits:

```
typedef enum avifAddImageFlag
{
    AVIF_ADD_IMAGE_FLAG_NONE = 0,

    // Force this frame to be a keyframe (sync frame).
    AVIF_ADD_IMAGE_FLAG_FORCE_KEYFRAME = (1 << 0),

    // Use this flag when encoding a single frame, single layer image.
    // Signals "still_picture" to AV1 encoders, which tweaks various compression rules.
    // This is enabled automatically when using the avifEncoderWrite() single-image encode path.
    AVIF_ADD_IMAGE_FLAG_SINGLE = (1 << 1)
} avifAddImageFlag;
```

This seems like a bug in the FuzzTest framework!

If so, I think it would be better to work around the bug by clearing the unwanted bits in `flags`.

vrabaud: You are right: we now us a fuzztest version with a fix so let's revert this "fix" and let's see where that takes us.

PR #1941, opened 2024-01-15
Add imageSizeLimit arg to avifutil avifReadImage()

y-guyon:
Use AVIF_DEFAULT_IMAGE_SIZE_LIMIT in places where the image is encoded as AVIF afterwards anyway. Use 2^32-1 in other places for a reasonable threshold.

Avoids out-of-memory issues in fuzztest targets.

No entry in CHANGES.md as it should not impact the set of valid workflows in the public API or tools, besides the returned error code or message.

[BUG=oss-fuzz:65700](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65700)

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -582,7 +582,7 @@ static avifBool avifInputReadImage(avifInput * input,
         if (feof(stdin)) {
             return AVIF_FALSE;
         }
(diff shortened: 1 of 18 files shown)


vrabaud (approved):
I guess there is no way the height can be 0 (in all the frame/grid configurations)

Issue #1925, opened 2024-01-11
FAILED while building


marchLex:
Hi! Hope you'll help me dealing with a problem.

Using ninja I got next error

```
[1/30] Building C object CMakeFiles/avif_obj.dir/src/colrconvert.c.o
FAILED: CMakeFiles/avif_obj.dir/src/colrconvert.c.o
/usr/bin/cc -DAVIF_CODEC_AOM=1 -DAVIF_CODEC_AOM_DECODE=1 -DAVIF_CODEC_AOM_ENCODE=1 -I../include -I../third_party/libyuv/include -isystem ../ext/aom -O2 -DNDEBUG -fPIC -fvisibility=hidden   -Wall -Wextra -Werror -pthread -std=gnu99 -MD -MT CMakeFiles/avif_obj.dir/src/colrconvert.c.o -MF CMakeFiles/avif_obj.dir/src/colrconvert.c.o.d -o CMakeFiles/avif_obj.dir/src/colrconvert.c.o   -c ../src/colrconvert.c
../src/colrconvert.c: In function ‘avifColorPrimariesComputeRGBToXYZD50Matrix’:
../src/colrconvert.c:112:5: error: passing argument 1 of ‘avifVecMul’ from incompatible pointer type [-Werror]
     avifVecMul(rgbPrimariesInv, whitePointXYZ, rgbCoefficients);
     ^
../src/colrconvert.c:76:13: note: expected ‘const double (*)[3]’ but argument is of type ‘double (*)[3]’
 static void avifVecMul(const double M[3][3], const double x[3], double y[3])
             ^
../src/colrconvert.c:118:5: error: passing argument 2 of ‘avifMatMul’ from incompatible pointer type [-Werror]
     avifMatMul(rgbPrimaries, rgbCoefficientsMat, rgbXYZ);
     ^
../src/colrconvert.c:48:13: note: expected ‘const double (*)[3]’ but argument is of type ‘double (*)[3]’
 static void avifMatMul(const double A[3][3], const double B[3][3], double C[3][3])
             ^
../src/colrconvert.c:136:5: error: passing argument 1 of ‘avifMatMul’ from incompatible pointer type [-Werror]
     avifMatMul(adaptation, avifBradford, tmp);
     ^
../src/colrconvert.c:48:13: note: expected ‘const double (*)[3]’ but argument is of type ‘double (*)[3]’
 static void avifMatMul(const double A[3][3], const double B[3][3], double C[3][3])
             ^
[...]

PR #1946, opened 2024-01-17
Remove const in const double**

vrabaud:
We cannot keep the const meaning unless we switch to storing the 2d array as a 1d array which make the code harder to read. cf https://c-faq.com/ansi/constmismatch.html


y-guyon (approved):

vrabaud:
Let's merge this simplest solution as approved by @wantehchang at https://github.com/AOMediaCodec/libavif/pull/1932#issuecomment-1894361991
It works as mentions by @sezero at https://github.com/AOMediaCodec/libavif/issues/1925#issuecomment-1895864221

Issue #1943, opened 2024-01-17
build error with `comparison of unsigned expression < 0 is always false`


sezero:
I hit this error when building both 1.0.x and the main branches using gcc-4.4.7.

Here's the output from a 1.0.x build:

```
cc1: warnings being treated as errors
src/avif.c: In function ‘avifImageCreate’:
src/avif.c:144: error: comparison of unsigned expression < 0 is always false

cc1: warnings being treated as errors
src/reformat.c: In function ‘avifPrepareReformatState’:
src/reformat.c:52: error: comparison of unsigned expression < 0 is always false
```

y-guyon:
Is https://github.com/AOMediaCodec/libavif/pull/1944 enough?

What is your compiling environment?

Also feel free to set `-DAVIF_ENABLE_WERROR=OFF` when using CMake.

sezero:
> Is #1944 enough?

That just transforms errors into `comparison of unsigned expression >= 0 is always true`

> What is your compiling environment?

gcc-4.4.7, as I had mentioned, on CentOS 6.10 (yes, EOL, I know.)
 
> Also feel free to set `-DAVIF_ENABLE_WERROR=OFF` when using CMake.

That is the workaround I am currently using

y-guyon:
> That just transforms errors into `comparison of unsigned expression >= 0 is always true`

What about https://github.com/AOMediaCodec/libavif/pull/1945?

sezero:
> What about #1945?

Yes, the `(int)` cast does the trick (#1925 is still in the way for me to build main branch, but the int cast solution for this particular issue works.)

PR #1949, opened 2024-01-17
Use AVIF_CHECKERR() in avifImageCreate()

y-guyon:
Cherry-pick of 1c3de07e440623dd2bd0b38fd3c76b6e69b8cbdc and 8ca862da585c5cac95d0803f4670b2dab4627823 (mostly as an exercise).

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -137,18 +137,11 @@ void avifImageSetDefaults(avifImage * image)
 avifImage * avifImageCreate(uint32_t width, uint32_t height, uint32_t depth, avifPixelFormat yuvFormat)
 {
     // width and height are checked when actually used, for example by avifImageAllocatePlanes().
-    if (depth > 16) {
-        // avifImage only supports up to 16 bits per sample. See avifImageUsesU16().
-        return NULL;
-    }
-    if ((yuvFormat < AVIF_PIXEL_FORMAT_NONE) || (yuvFormat > AVIF_PIXEL_FORMAT_YUV400)) {
-        return NULL;
-    }
+    AVIF_CHECKERR(depth <= 16, NULL); // avifImage only supports up to 16 bits per sample. See avifImageUsesU16().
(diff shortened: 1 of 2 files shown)


sezero:
Fixes #1943 in v1.0.x for me.

src/avif.c
```diff
@@ -137,18 +137,12 @@ void avifImageSetDefaults(avifImage * image)
...
-    if (depth > 16) {
-        // avifImage only supports up to 16 bits per sample. See avifImageUsesU16().
-        return NULL;
-    }
-    if ((yuvFormat < AVIF_PIXEL_FORMAT_NONE) || (yuvFormat > AVIF_PIXEL_FORMAT_YUV400)) {
-        return NULL;
-    }
+    AVIF_CHECKERR(depth <= 16, NULL); // avifImage only supports up to 16 bits per sample. See avifImageUsesU16().
+    // Cast to silence "comparison of unsigned expression is always true" warning.
+    AVIF_CHECKERR((int)yuvFormat >= AVIF_PIXEL_FORMAT_NONE && yuvFormat < AVIF_PIXEL_FORMAT_COUNT, NULL);
```
wantehchang, line 142: Since `yuvFormat` is of the `avifPixelFormat` enum type and `AVIF_PIXEL_FORMAT_NONE` is a constant in that enum type, this compiler warning is incorrect.

As reported in https://github.com/AOMediaCodec/libavif/issues/1943, the compiler that warns about this is gcc-4.4.7 on CentOS 6.10. That's a very old version of gcc, and [CentOS 6 reached the end of support in November 2020](https://wiki.centos.org/FAQ(2f)General.html#What_is_the_support_.27.27end_of_life.27.27_for_each_CentOS_release.3F). We don't need to add a workaround for this compiler.

If we want to support this old compiler, please add a comment noting the compiler and OS version, so that we know when it is safe to remove the workaround.

y-guyon:
Superseded by https://github.com/AOMediaCodec/libavif/pull/1970

Issue #1947, opened 2024-01-17
`-Wignored-attributes` error with old clang


sezero:
With clang 3.4.2, I got the following error:

```
In file included from /tmp/libavif/src/alpha.c:4:
In file included from /tmp/libavif/include/avif/internal.h:7:
/tmp/libavif/include/avif/avif.h:157:14: error: 'warn_unused_result' attribute only applies to functions, methods, and classes [-Werror,-Wignored-attributes]
typedef enum AVIF_NODISCARD avifResult
             ^
/tmp/libavif/include/avif/avif.h:57:39: note: expanded from macro 'AVIF_NODISCARD'
#define AVIF_NODISCARD __attribute__((warn_unused_result))
                                      ^
1 error generated.
```

Seems to be after https://github.com/AOMediaCodec/libavif/pull/1655

Removing `AVIF_NODISCARD` from `avifResult` enum makes the build to succeed.

y-guyon:
Thanks for the report.

Does setting `-DAVIF_ENABLE_WERROR=OFF` solve the issue?

sezero:
Yes, it does for build of libavif itself. (The thing is though, the public header is the offender.)

vrabaud:
Using godbolt, I could figure out clang 3.9.0 was the first one to introduce the behavior for enums.

PR #1951, opened 2024-01-18
Only define no_discard for clang >= 3.9.0

vrabaud:
godbolt shows that's when it was also define for enums. This fixes https://github.com/AOMediaCodec/libavif/issues/1947

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -52,7 +52,7 @@ extern "C" {
     (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
 #define AVIF_NODISCARD [[nodiscard]]
 #else
-#if defined(__clang__) && defined(__has_attribute)
+#if defined(__clang__) && defined(__has_attribute) && ((__clang_major__ << 8) | __clang_minor__) >= ((3 << 8) | 9)
 #if __has_attribute(warn_unused_result)
 #define AVIF_NODISCARD __attribute__((warn_unused_result))
 #else


y-guyon (approved):

include/avif/avif.h
```diff
@@ -52,7 +52,7 @@ extern "C" {
     (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
 #define AVIF_NODISCARD [[nodiscard]]
 #else
-#if defined(__clang__) && defined(__has_attribute)
+#if defined(__clang__) && defined(__has_attribute) && ((__clang_major__ << 8) | __clang_minor__) >= ((3 << 8) | 9)
```
wantehchang, line 55: Vincent: Thank you for using finding the first version of clang that supports the `warn_unused_result` attribute.

It is strange that `__has_attribute(warn_unused_result)` is not working correctly for clang 3.4.2.

Optional: `((3 << 8) | 9)` can be written as `0x309` or `0x0309`.

jzern: > Vincent: Thank you for using finding the first version of clang that supports the `warn_unused_result` attribute.
> 
> It is strange that `__has_attribute(warn_unused_result)` is not working correctly for clang 3.4.2.
> 

I think `__has_attribute` is working, but earlier versions of clang didn't allow the attribute to be used on enums (gcc still doesn't I believe). I didn't think adding it to enums was strictly necessary.

> Optional: `((3 << 8) | 9)` can be written as `0x309` or `0x0309`.

wantehchang: James: You are right. I forgot what the bug was about. Thanks.

Yes, another solution would be to individually mark the public functions that return `avifResult` as `AVIF_NODISCARD`.

Vincent: I suggest adding a comment to explain the clang version >= 3.9 test is to check whether the `warn_unused_result` attribute can be used on enums. This is not obvious.

vrabaud: Comment added in https://github.com/AOMediaCodec/libavif/pull/1954

PR #463, opened 2020-12-31
fix avis encode bug on Deduplication

nju04zq:
Change-Id: Ic5a2a781d08f00d878a160d8a7aae4f0a563a62f

PR #1956, opened 2024-01-19
Properly check if an identical chunk has been written before.

vrabaud:
Fixes internal bugs: b/321189607 and b/321189607

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -1820,7 +1820,10 @@ static avifResult avifEncoderWriteMediaDataBox(avifEncoder * encoder,
                 avifEncodeSample * sample = &item->encodeOutput->samples.sample[0];
                 size_t total_size = sample->data.size;
                 for (uint32_t sampleIndex = 1; sampleIndex < item->encodeOutput->samples.count; ++sampleIndex) {
-                    total_size += item->encodeOutput->samples.sample[sampleIndex].data.size;
+                    const avifEncodeSample * const sample = &item->encodeOutput->samples.sample[sampleIndex];
+                    AVIF_CHECKERR(sample->data.data == item->encodeOutput->samples.sample[0].data.data + total_size,
+                                  AVIF_RESULT_UNKNOWN_ERROR);
+                    total_size += sample->data.size;
                 }
                 chunkOffset = avifEncoderFindExistingChunk(s, mdatStartOffset, sample->data.data, total_size);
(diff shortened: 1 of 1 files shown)


src/write.c
```diff
@@ -1818,7 +1818,11 @@ static avifResult avifEncoderWriteMediaDataBox(avifEncoder * encoder,
             // Deduplication - See if an identical chunk to this has already been written
             if (item->encodeOutput->samples.count > 0) {
                 avifEncodeSample * sample = &item->encodeOutput->samples.sample[0];
-                chunkOffset = avifEncoderFindExistingChunk(s, mdatStartOffset, sample->data.data, sample->data.size);
+                size_t total_size = sample->data.size;
+                for (uint32_t sampleIndex = 1; sampleIndex < item->encodeOutput->samples.count; ++sampleIndex) {
+                    total_size += item->encodeOutput->samples.sample[sampleIndex].data.size;
```
y-guyon, line 1823: Is the data contiguous?

Suggestion:

```c
AVIF_CHECKERR(
  item->encodeOutput->samples.sample[sampleIndex].data.data ==
    item->encodeOutput->samples.sample[0].data.data + total_size,
  AVIF_RESULT_UNKNOWN_ERROR);
```

(not just an assertion, for extra security)

y-guyon (approved):
This piece of code was written 4 years ago but maybe someone more familiar with tracks than me could review.

src/write.c
vrabaud: Well, the data is written contiguously a few lines below but I'll add a check.
The problem happened starting with https://github.com/AOMediaCodec/libavif/pull/463

vrabaud: Actually, we do not care right? The stream is already formed so as long as we have a match of size total_size, we are good and the deduplication works.

vrabaud: I missed it's on the input, that makes sense. Done

y-guyon (approved):

vigneshvg (approved):

wantehchang:
@vrabaud Vincent: The first comment in this pull request says "Fixes internal bugs: b/321189607 and b/321189607", but the two bug numbers are the same. Is there a second bug that this pull request fixes?

vrabaud:
Nope, just a mistake on my end ...

PR #1960, opened 2024-01-19
Add asan as a dependency by default for fuzztest tests.

vrabaud:
This is because dav1d does it by default apparently and we need it in case we build other sanitizers (e.g. coverage)


vrabaud:
[DavidKorczynski](https://github.com/DavidKorczynski), that's what you had in mind right?

y-guyon (approved):

DavidKorczynski (approved):
Yes, this LGTM and will make fuzztest fuzzers appear in the coverage reports

PR #1964, opened 2024-01-22
Actually use a local fuzztest repo and bump its version.

vrabaud:
This is after discussion at:
https://github.com/google/oss-fuzz/issues/11528

### DIFF
--- a/ext/fuzztest.cmd
+++ b/ext/fuzztest.cmd
@@ -9,8 +9,8 @@
 
 git clone https://github.com/google/fuzztest.git
 cd fuzztest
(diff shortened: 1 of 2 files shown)


y-guyon (approved):

DavidKorczynski (approved):
awesome!

I'd be interested in whether https://github.com/AOMediaCodec/libavif/pull/1960 is still needed. Am not sure if it is, but in case it is it may be worth building fuzztest as part of the Dockerfile (as opposed to `build.sh` from OSS-Fuzz's perspective) assuming the added `-fsanitize=address` somewhere in building the dependencies is due to environment flag set by OSS-Fuzz.

PR #1967, opened 2024-01-22
Simpler itemID finding for alpha.

vrabaud:
Also better handle the case where there is no possible value and newItemID wraps around.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -4402,24 +4402,18 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
     }
     assert(alphaItemCount == colorItemCount);
     // Find an unused ID.
-    avifResult result;
-    if (meta->items.count >= UINT32_MAX - 1) {
-        // In the improbable case where all IDs are used.
-        result = AVIF_RESULT_DECODE_ALPHA_FAILED;
-    } else {
-        uint32_t newItemID = 0;
-        avifBool isUsed;
-        do {
+    avifResult result = AVIF_RESULT_DECODE_ALPHA_FAILED;
+    for (uint32_t i = 0, newItemID = 1; i < meta->items.count; ++i) {
+        if (meta->items.item[i]->id == newItemID) {
             ++newItemID;
-            isUsed = AVIF_FALSE;
-            for (uint32_t i = 0; i < meta->items.count; ++i) {
-                if (meta->items.item[i]->id == newItemID) {
-                    isUsed = AVIF_TRUE;
-                    break;
-                }
-            }
-        } while (isUsed && newItemID != 0);
-        result = avifMetaFindOrCreateItem(meta, newItemID, alphaItem); // Create new empty item.
+        } else {
+            // ISO/IEC 23008-12, First edition, 2017-12, Section 9.3.1:
+            //   Each ItemPropertyAssociation box shall be ordered by increasing item_ID, and there shall
+            //   be at most one association box for each item_ID, in any ItemPropertyAssociation box.
+            // A free spot id is found.
+            result = avifMetaFindOrCreateItem(meta, newItemID, alphaItem); // Create new empty item.
+            break;
+        }
     }
     if (result != AVIF_RESULT_OK) {
         avifFree(alphaItemIndices);
--- a/src/read.c
+++ b/src/read.c
@@ -714,6 +714,11 @@ typedef struct avifMeta
     // and are then further modified/updated as new information for an item's ID is parsed.
     avifDecoderItemArray items;
 
+    // Used when the colorItem is a grid and the alpha item is represented as a set of auxl items
+    // to each color tile. The data of this item need not be read and the pixi property cannot be
+    // validated.
+    avifDecoderItem * alphaItem;
+
     // Any ipco boxes explained above are populated into this array as a staging area, which are
     // then duplicated into the appropriate items upon encountering an item property association
     // (ipma) box.
@@ -772,6 +777,7 @@ static void avifMetaDestroy(avifMeta * meta)
         }
         avifFree(item);
     }
+    avifFree(meta->alphaItem);
     avifArrayDestroy(&meta->items);
     avifArrayDestroy(&meta->properties);
     avifRWDataFree(&meta->idat);
@@ -787,6 +793,26 @@ static avifResult avifCheckItemID(const char * boxFourcc, uint32_t itemID, avifD
     return AVIF_RESULT_OK;
 }
 
+static void avifCreateItem(avifDecoderItem ** item)
+{
+    *item = (avifDecoderItem *)avifAlloc(sizeof(avifDecoderItem));
+    if (*item == NULL) {
+        return;
+    }
+    memset(*item, 0, sizeof(avifDecoderItem));
+
+    if (!avifArrayCreate(&(*item)->properties, sizeof(avifProperty), 16)) {
+        avifFree(*item);
+        *item = NULL;
+        return;
+    }
+    if (!avifArrayCreate(&(*item)->extents, sizeof(avifExtent), 1)) {
+        avifArrayDestroy(&(*item)->properties);
+        avifFree(*item);
+        *item = NULL;
+    }
+}
+
 static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avifDecoderItem ** item)
 {
     *item = NULL;
@@ -799,29 +825,12 @@ static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avi
         }
     }
 
+    avifCreateItem(item);
+    AVIF_CHECKERR(item != NULL, AVIF_RESULT_OUT_OF_MEMORY);
     avifDecoderItem ** itemPtr = (avifDecoderItem **)avifArrayPush(&meta->items);
     AVIF_CHECKERR(itemPtr != NULL, AVIF_RESULT_OUT_OF_MEMORY);
-    *item = (avifDecoderItem *)avifAlloc(sizeof(avifDecoderItem));
-    if (*item == NULL) {
-        avifArrayPop(&meta->items);
-        return AVIF_RESULT_OUT_OF_MEMORY;
-    }
-    memset(*item, 0, sizeof(avifDecoderItem));
 
     *itemPtr = *item;
-    if (!avifArrayCreate(&(*item)->properties, sizeof(avifProperty), 16)) {
-        avifFree(*item);
-        *item = NULL;
-        avifArrayPop(&meta->items);
-        return AVIF_RESULT_OUT_OF_MEMORY;
-    }
-    if (!avifArrayCreate(&(*item)->extents, sizeof(avifExtent), 1)) {
-        avifArrayDestroy(&(*item)->properties);
-        avifFree(*item);
-        *item = NULL;
-        avifArrayPop(&meta->items);
-        return AVIF_RESULT_OUT_OF_MEMORY;
-    }
     (*item)->id = itemID;
     (*item)->meta = meta;
     return AVIF_RESULT_OK;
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -4402,24 +4402,18 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
-                    isUsed = AVIF_TRUE;
-                    break;
-                }
-            }
-        } while (isUsed && newItemID != 0);
-        result = avifMetaFindOrCreateItem(meta, newItemID, alphaItem); // Create new empty item.
+        } else {
+            // ISO/IEC 23008-12, First edition, 2017-12, Section 9.3.1:
+            //   Each ItemPropertyAssociation box shall be ordered by increasing item_ID, and there shall
+            //   be at most one association box for each item_ID, in any ItemPropertyAssociation box.
```
y-guyon, line 4412: Are `meta->items` always created in this order? Is the `ItemPropertyAssociation` box parsed first?

I remember most places in `read.c` referencing items calling `avifMetaFindOrCreateItem()`, so depending on which box is parsed first, items may not be in order.

If they are in order, please add
```c
AVIF_CHECKERR(i == 0 || meta->items.item[i-1]->id < meta->items.item[i]->id, ERROR);
```

y-guyon (changes requested):

src/read.c
```diff
@@ -787,6 +793,26 @@ static avifResult avifCheckItemID(const char * boxFourcc, uint32_t itemID, avifD
     return AVIF_RESULT_OK;
 }
 
+static void avifCreateItem(avifDecoderItem ** item)
```
wantehchang, line 796: Nit: Return the newly-created item as the return value?

```
static avifDecoderItem * avifCreateItem(void)
```

src/read.c
```diff
@@ -799,29 +825,12 @@ static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avi
         }
     }
 
+    avifCreateItem(item);
+    AVIF_CHECKERR(*item != NULL, AVIF_RESULT_OUT_OF_MEMORY);
```
wantehchang, line 829: IMPORTANT: Line 831 should be rewritten as an if statement so that we can free `*item` before returning an error:

```
    if (itemPtr == NULL) {
        avifFree(*item);
        *item = NULL;
        AVIF_RESULT_OUT_OF_MEMORY);
    }
```

src/read.c
```diff
@@ -4401,31 +4410,15 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
-                    isUsed = AVIF_TRUE;
-                    break;
-                }
-            }
-        } while (isUsed && newItemID != 0);
-        result = avifMetaFindOrCreateItem(meta, newItemID, alphaItem); // Create new empty item.
-    }
-    if (result != AVIF_RESULT_OK) {
+    // Create a temporary item;
+    avifCreateItem(alphaItem);
```
wantehchang, line 4414: IMPORTANT: Check whether `*alphaItem` is a null pointer.

It seems that line 4416 is meant to check that, but the check is incorrect (it should check `*alphaItem`, not  `alphaItem`) and too late (`*alphaItem` is already dereferenced at line 4415).

src/read.c
```diff
@@ -714,6 +714,11 @@ typedef struct avifMeta
     // and are then further modified/updated as new information for an item's ID is parsed.
     avifDecoderItemArray items;
 
+    // Used when the colorItem is a grid and the alpha item is represented as a set of auxl items
+    // to each color tile. The data of this item need not be read and the pixi property cannot be
+    // validated.
```
wantehchang, line 719: Document that `alphaItem->id` is 0.

IMPORTANT: Ideally `alphaItem->id` should not be used because it is an invalid value. But it is used at line 4427 in src/read.c:

```
    for (uint32_t i = 0; i < alphaItemCount; ++i) {
        avifDecoderItem * item = meta->items.item[alphaItemIndices[i]];
        item->dimgForID = (*alphaItem)->id;  <=== USED HERE!
    }
```

This can lead to bugs. Note that `dimgForID` is documented as follows:

```
    uint32_t dimgForID;            // if non-zero, this item is an input of derived Item #{dimgForID}
```

wantehchang (commented):
Vincent: I guess you are trying to make it unnecessary to find an itemID for the fake alpha item. But the itemID of the fake alpha item is still used at src/read.c:4427.

src/read.c
vrabaud: Done

src/read.c
vrabaud: Right, I was testing, sorry about the noise.

src/read.c
vrabaud: Outdated code

src/read.c
vrabaud: Outdated code

src/read.c
```diff
@@ -799,6 +799,16 @@ static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avi
         }
     }
 
+    uint32_t prevID = 0;
+    // ISO/IEC 23008-12, First edition, 2017-12, Section 9.3.1:
+    //   Each ItemPropertyAssociation box shall be ordered by increasing item_ID, and there shall
+    //   be at most one association box for each item_ID, in any ItemPropertyAssociation box.
+    for (uint32_t i = 0; i < meta->items.count; ++i) {
+        AVIF_CHECKERR(meta->items.item[i]->id > prevID, AVIF_RESULT_BMFF_PARSE_FAILED);
```
wantehchang, line 807: It would be good to add a diagnostic message. (We won't be able to use the `AVIF_CHECKERR` macro.)

src/read.c
```diff
@@ -799,6 +799,16 @@ static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avi
...
+        prevID = meta->items.item[i]->id;
+    }
+    AVIF_CHECKERR(itemID > prevID, AVIF_RESULT_BMFF_PARSE_FAILED);
```
wantehchang, line 810: If this is the only place where we add new items to `meta->items`, then we can assume the function maintains this property as an invariant, and the only validation we need to check is that `itemID` be greater than the last item ID in `meta->items`:

```
    if (meta->items.count != 0) {
        const uint32_t lastID = meta->items.item[meta->items.count - 1]->id;
        // Also log a diagnostic message!
        AVIF_CHECKERR(itemID > lastID, AVIF_RESULT_BMFF_PARSE_FAILED);
    }
```

src/read.c
```diff
@@ -4403,23 +4413,12 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
     assert(alphaItemCount == colorItemCount);
     // Find an unused ID.
     avifResult result;
-    if (meta->items.count >= UINT32_MAX - 1) {
-        // In the improbable case where all IDs are used.
+    const uint32_t lastID = meta->items.count ? meta->items.item[meta->items.count - 1]->id : 0;
```
wantehchang, line 4416: Nit/optional: Update the "Find an unused ID" comment? Now we aren't really finding an unused ID.

src/read.c
vrabaud: Done (i had to change all callers)

src/read.c
vrabaud: Indeed, let's keep it simple.

src/read.c
vrabaud: Done

src/read.c
```diff
@@ -787,7 +787,7 @@ static avifResult avifCheckItemID(const char * boxFourcc, uint32_t itemID, avifD
     return AVIF_RESULT_OK;
 }
 
-static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avifDecoderItem ** item)
+static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avifDiagnostics * diag, avifDecoderItem ** item)
```
wantehchang, line 790: Nit: By convention `avifDiagnostics * diag` is usually the last parameter.

src/read.c
```diff
@@ -799,6 +799,15 @@ static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avi
         }
     }
 
+    if (meta->items.count != 0) {
+        // ISO/IEC 23008-12, First edition, 2017-12, Section 9.3.1:
+        //   Each ItemPropertyAssociation box shall be ordered by increasing item_ID, and there shall
+        //   be at most one association box for each item_ID, in any ItemPropertyAssociation box.
+        const uint32_t lastID = meta->items.item[meta->items.count - 1]->id;
+        avifDiagnosticsPrintf(diag, "The added itemID [%u] does not preserve the itemID order", itemID);
+        AVIF_CHECKERR(itemID > lastID, AVIF_RESULT_BMFF_PARSE_FAILED);
```
wantehchang, line 808: IMPORTANT: Strictly speaking, what you do is correct. Like `errno`, the diagnostic message should only be used when a function fails. So it is fine to log a diagnostic message before the check for error. But ideally we should log a diagnostic message only when a function fails.

To do this, we cannot use the `AVIF_CHECKERR` macro. We need to use an if statement:

```
        if (itemID <= lastID) {
            avifBreakOnError();
            avifDiagnosticsPrintf(diag, "The added itemID [%u] does not preserve the itemID order", itemID);
        }
```

src/read.c
```diff
@@ -4401,25 +4411,14 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
         }
     }
     assert(alphaItemCount == colorItemCount);
-    // Find an unused ID.
+    // Figure out the last used itemID.
     avifResult result;
-    if (meta->items.count >= UINT32_MAX - 1) {
-        // In the improbable case where all IDs are used.
+    const uint32_t lastID = meta->items.count ? meta->items.item[meta->items.count - 1]->id : 0;
```
wantehchang, line 4416: At this point, we know `meta->items.count` is nonzero. Reason: The `if` statement at line 4376 ensures `colorItemCount` is nonzero. The assertion at line 4413 implies `alphaItemCount` is nonzero. If `alphaItemCount` is nonzero, then we must have entered the for loop at line 4384. So `meta->items.count` is nonzero.

So this can be simplified to:

```
    const uint32_t lastID = meta->items.item[meta->items.count - 1]->id;
```

wantehchang (approved):

src/read.c
vrabaud: Done

src/read.c
vrabaud: Good catch, done.

src/read.c
vrabaud: Cleaner indeed, done

src/read.c
```diff
@@ -4341,6 +4353,7 @@ static avifBool avifDecoderItemIsAlphaAux(const avifDecoderItem * item, uint32_t
 static avifResult avifMetaFindAlphaItem(avifMeta * meta,
                                         const avifDecoderItem * colorItem,
                                         const avifTileInfo * colorInfo,
+                                        avifDiagnostics * diag,
```
wantehchang, line 4356: Nit: It would be good to move the `diag` parameter to be the last. I am sorry I missed this one in my previous review.

src/read.c
```diff
@@ -4401,25 +4414,14 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
     }
     assert(alphaItemCount == colorItemCount);
-    // Find an unused ID.
+    // Figure out the last used itemID.
     avifResult result;
-    if (meta->items.count >= UINT32_MAX - 1) {
-        // In the improbable case where all IDs are used.
+    const uint32_t lastID = meta->items.item[meta->items.count - 1]->id;
+    if (lastID == UINT32_MAX) {
+        // In the improbable case where the last ID is the maximum one, ids cannot be kept ordered.
```
wantehchang, line 4421: We should log a diagnostic message here. I thought of this in my previous review but somehow forgot to comment about it. Sorry.

wantehchang (commented):
LGTM. I suggest two more changes.

src/read.c
vrabaud: I saw that after too, done in https://github.com/AOMediaCodec/libavif/pull/1976 (with your comment below.

src/read.c
vrabaud: Done in https://github.com/AOMediaCodec/libavif/pull/1976

PR #1973, opened 2024-01-26
Get ubsan fuzztest test to build.

vrabaud:
- add missing flags that would trigger undefined reference to asan
- update fuzztest to not get an undefined reference to LLVMFuzzerRunDriver
- use normal flags for projects built outside of libavif (dav1d with meson is problematic with sanitizer flags so let's minimize risks)

### DIFF
--- a/ext/fuzztest.cmd
+++ b/ext/fuzztest.cmd
@@ -9,7 +9,8 @@
 
 git clone https://github.com/google/fuzztest.git
 cd fuzztest
-: # There is no tagged release as of 2024/01/26. Pick the earliest commit that contains a needed bug fix.
+: # There is no tagged release as of 2024/01/26. Pick the earliest commit that fixes the
+: # undefined reference to LLVMFuzzerRunDriver when building ubsan tests.
 git checkout a53a2083e7df08749ea26b5960c05a9bffa186c2
 
 : # fuzztest is built by the main CMake project through add_subdirectory as recommended at:
--- a/ext/fuzztest.cmd
+++ b/ext/fuzztest.cmd
@@ -9,8 +9,8 @@
 
 git clone https://github.com/google/fuzztest.git
 cd fuzztest
-: # There is no tagged release as of 2024/01/22. Pick the earliest commit that contains a needed bug fix.
-git checkout 8f82f46e14cc2484ddecf6d55e46243923f9a036
+: # There is no tagged release as of 2024/01/26. Pick the earliest commit that contains a needed bug fix.
+git checkout a53a2083e7df08749ea26b5960c05a9bffa186c2
 
 : # fuzztest is built by the main CMake project through add_subdirectory as recommended at:
 : # https://github.com/google/fuzztest/blob/main/doc/quickstart-cmake.md
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -199,6 +199,7 @@ if(AVIF_ENABLE_FUZZTEST)
                 "fuzztest: Installed FuzzTest is not supported, please set AVIF_LOCAL_FUZZTEST=ON or AVIF_ENABLE_FUZZTEST=OFF"
         )
     endif()
+    fuzztest_setup_fuzzing_flags()
 
     add_avif_fuzztest(avif_fuzztest_dec)
     add_avif_fuzztest(avif_fuzztest_dec_incr gtest/avifincrtest_helpers.cc)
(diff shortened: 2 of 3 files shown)


tests/oss-fuzz/build.sh
```diff
@@ -43,14 +43,13 @@ export CFLAGS=""
 export CXXFLAGS=""
 
 # fuzz flags are problematic with meson (hence no dav1d) and no point in fuzzing fuzztest.
```
jzern, line 45: The last part of the sentence needs an update.

vrabaud: Done

ext/fuzztest.cmd
```diff
@@ -9,8 +9,8 @@
 
 git clone https://github.com/google/fuzztest.git
 cd fuzztest
-: # There is no tagged release as of 2024/01/22. Pick the earliest commit that contains a needed bug fix.
-git checkout 8f82f46e14cc2484ddecf6d55e46243923f9a036
+: # There is no tagged release as of 2024/01/26. Pick the earliest commit that contains a needed bug fix.
```
wantehchang, line 12: It would be good to say specifically what the needed bug fix is. Or just say something like "Pick a stable snapshot."

tests/oss-fuzz/build.sh
```diff
@@ -42,15 +42,15 @@ export ORIG_CXXFLAGS="$CXXFLAGS"
 export CFLAGS=""
 export CXXFLAGS=""
 
-# fuzz flags are problematic with meson (hence no dav1d) and no point in fuzzing fuzztest.
-cd ext && bash dav1d.cmd && bash fuzztest.cmd && cd ..
+# Use normal flags for projects built outside of libavif (e.g. dav1d with meson is
```
wantehchang, line 45: "projects built outside of libavif" also include libaom, but aom.cmd is not in this list (line 47).

Should we move aom.cmd to this list?

tests/oss-fuzz/build.sh
```diff
@@ -42,15 +42,15 @@ export ORIG_CXXFLAGS="$CXXFLAGS"
...
+# problematic with sanitizer flags)
+cd ext && bash dav1d.cmd && bash libsharpyuv.cmd && bash libyuv.cmd && cd ..
 
 export CFLAGS=$ORIG_CFLAGS
 export CXXFLAGS=$ORIG_CXXFLAGS
 
 # build dependencies
-cd ext && bash aom.cmd && bash libjpeg.cmd && bash libsharpyuv.cmd && \
-      bash libyuv.cmd && bash zlibpng.cmd && cd ..
+cd ext && bash aom.cmd && bash fuzztest.cmd && bash libjpeg.cmd && bash zlibpng.cmd && cd ..
```
wantehchang, line 53: Move the "build dependencies" comment before line 39 ("# Reset compile flags to build some deps without fuzzer flags."). Or update this comment to say that these dependencies are built as subprojects of libavif.

ext/fuzztest.cmd
vrabaud: I put the reason. Then again, this is a moving target. It's the latest stable for us. Let's hope they start versioning.

tests/oss-fuzz/build.sh
vrabaud: Yes and no. I reduced the list to only have dav1d which truly does not work here with sanitizer flags.
We could add the ones we build outside (like aom, sharpyuv,libyuv) so that the fuzzing info is not present (for speed or so that they do not pollute the coverage) but we are deeply involved with those projects so we might as well put them.

tests/oss-fuzz/build.sh
vrabaud: I rephrased that and actually have dav1d be on its own because of meon.

vrabaud:
As you saw, this is still a bit fragile and ad-hoc but it will get us further (ubsan tests are not building right now)

tests/CMakeLists.txt
```diff
@@ -199,6 +199,7 @@ if(AVIF_ENABLE_FUZZTEST)
                 "fuzztest: Installed FuzzTest is not supported, please set AVIF_LOCAL_FUZZTEST=ON or AVIF_ENABLE_FUZZTEST=OFF"
         )
     endif()
+    fuzztest_setup_fuzzing_flags()
```
wantehchang, line 202: Just curious: did you write this function before and it was not used until now?

wantehchang (approved):
LGTM. Thanks.

tests/CMakeLists.txt
vrabaud: This is actually part of the fuzztest framework and is part of the doc https://github.com/google/fuzztest/blob/main/doc/quickstart-cmake.md. It actually helps for ubsan tests that also require to be linked to asan but was not necessary for the main tests I cared about so far (pure asan tests). Let's get them all now :)

PR #1985, opened 2024-02-01
Add link to wiki in README.md

y-guyon:


README.md
```diff
@@ -112,7 +112,10 @@ Debug and Release are available on
 
 ## Development Notes
 
-libavif is written in C99.
+Please check the [wiki](https://github.com/AOMediaCodec/libavif/wiki) for extra
+resources on libavif, such as the Release Checklist.
+
+Core libavif is written in C99. Most of the tests are written in C++14.
```
wantehchang, line 118: Nit: We can replace "Core libavif" with "The libavif library". This leaves the two main apps avifenc and avifdec open to be converted to C++ code.

We could also say "The C code in libavif is written in C99. The C++ code is written in C++14."

wantehchang (approved):

PR #1993, opened 2024-02-05
Revert "Simpler itemID finding for alpha. (#1967)"

vigneshvg:
This reverts commit 4378794207f7a244bc36730c63dcc2bc188b6b3c.

Validating the ipma requirement of increasing item_ids inside avifMetaFindOrCreateItem is incorrect since that function is called when parsing several other boxes where new items could be added (for e.g. iloc).

A sample avif file that is broken by this change:
https://g-issues.chromium.org/issues/323735410#comment4

BUG=chromium:323735410

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -787,7 +787,7 @@ static avifResult avifCheckItemID(const char * boxFourcc, uint32_t itemID, avifD
     return AVIF_RESULT_OK;
 }
 
-static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avifDecoderItem ** item, avifDiagnostics * diag)
+static avifResult avifMetaFindOrCreateItem(avifMeta * meta, uint32_t itemID, avifDecoderItem ** item)
 {
     *item = NULL;
     AVIF_ASSERT_OR_RETURN(itemID != 0);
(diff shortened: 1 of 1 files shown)


wantehchang:
I confirm the 'ipma' item_ID ordering requirement is no longer in ISO/IEC 23008-12:2022. However, I found the requirement in ISO/IEC 14496-12:2022, the last paragraph of Section 8.11.14.1 on page 91:

> Each ItemPropertyAssociationBox shall be ordered by increasing item_ID, and there shall be at most
> one occurrence of a given item_ID, in the set of ItemPropertyAssociationBox boxes. The version 0
> should be used unless 32-bit item_ID values are needed; similarly, flags should be equal to 0 unless
> there are more than 127 properties in the ItemPropertyContainerBox. There shall be at most one
> ItemPropertyAssociationBox with a given pair of values of version and flags.

vigneshvg:
> I confirm the 'ipma' item_ID ordering requirement is no longer in ISO/IEC 23008-12:2022. However, I found the requirement in ISO/IEC 14496-12:2022, the last paragraph of Section 8.11.14.1 on page 91:
> 
> > Each ItemPropertyAssociationBox shall be ordered by increasing item_ID, and there shall be at most
> > one occurrence of a given item_ID, in the set of ItemPropertyAssociationBox boxes. The version 0
> > should be used unless 32-bit item_ID values are needed; similarly, flags should be equal to 0 unless
> > there are more than 127 properties in the ItemPropertyContainerBox. There shall be at most one
> > ItemPropertyAssociationBox with a given pair of values of version and flags.

Thanks for finding this @wantehchang. So the validation that we have already in `avifParseItemPropertyAssociation` is correct. We will just have to update the comment about the spec section which it is enforcing.

wantehchang (approved):
LGTM. You may need to edit the commit message rather than the first comment of the pull request. Or use the "Squash and rebase" method, which will give you a chance to edit the final commit message.

sezero:
v1.0.x branch had cherry-picked the reverted commit: Will it be reverted in v1.0.x branch as well?

vigneshvg:
> v1.0.x branch had cherry-picked the reverted commit: Will it be reverted in v1.0.x branch as well?

Yep, that is the plan.

wantehchang:
I will revert commit 5ff8f49de0537fd9c6ec23fa7489b435a5e07b32 and cherry-pick commit 59f1c9af777d517a781e2bec933dd492d72b9976 on the v1.0.x branch.

PR #1994, opened 2024-02-06
Use libjpeg-turbo in ext/libjpeg.cmd

y-guyon:
[BUG=oss-fuzz:66492](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=66492)
[BUG=oss-fuzz:66434](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=66434)
[BUG=oss-fuzz:66425](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=66425)


.gitignore
```diff
@@ -9,6 +9,7 @@
 /ext/gpac
 /ext/libargparse
 /ext/libjpeg
+/ext/libjpeg-turbo
```
wantehchang, line 12: Should we delete the `/ext/libjpeg` line?

ext/libjpeg.cmd
```diff
@@ -1,7 +1,10 @@
 : # If you want to use a local build of jpeg, you must clone the repos in this directory first,
-: # then enable CMake's AVIF_LOCAL_JPEG.
-: # This git tag isn't likely to move much, as libjpeg isn't actively developed anymore.
+: # then set CMake's AVIF_JPEG=LOCAL.
+: # The git tags below are known to work, and will occasionally be updated. Feel free to use a more recent commit.
```
wantehchang, line 3: Nit: There is only one git tag in this file, so "The git tag below is known to work" sounds better.

.gitignore
y-guyon: I thought I could leave it for now. If users update their local repositories instead of cloning fresh ones, it will avoid an eventual `libjpeg` entry popping into `git status`.

cmake/Modules/LocalJpeg.cmake
```diff
@@ -1,7 +1,18 @@
-add_subdirectory(${AVIF_SOURCE_DIR}/ext/libjpeg EXCLUDE_FROM_ALL)
+set(LIB_FOLDER "${AVIF_SOURCE_DIR}/ext/libjpeg-turbo/build.libavif")
```
wantehchang, line 1: Nit/optional: "DIR" is more common than "FOLDER" (e.g., line 13).

.gitignore
wantehchang: Seeing the `ext/libjpeg` entry in the `git status` output could remind the user to delete the directory. That seems good.

wantehchang (approved):
LGTM. Thanks.

.gitignore
y-guyon: > Seeing the `ext/libjpeg` entry in the `git status` output could remind the user to delete the directory.

Good point.  
The case I had in mind was switching between the `main` and `v1.0.x` branches while keeping both `ext/libjpeg` and `ext/libjpeg-turbo`.

> That seems good.

Not sure what this refers to.

cmake/Modules/LocalJpeg.cmake
```diff
@@ -1,7 +1,18 @@
-add_subdirectory(${AVIF_SOURCE_DIR}/ext/libjpeg EXCLUDE_FROM_ALL)
+set(LIB_DIR "${AVIF_SOURCE_DIR}/ext/libjpeg-turbo/build.libavif")
+if(WIN32)
```
kmilos, line 2: Can we make this MSVC specific please? MinGW toolcjains have no problem distinguishing static .a from stufb .dll.a files and don't need the "static" suffix workaaround.

wantehchang: Did our MinGW CI builds fail when we were checking `if(WIN32)` here?

.gitignore
wantehchang: I meant it seems good to have the `ext/libjpeg` entry in the `git status` output, because it could remind the user to delete the directory.

wantehchang (approved):

cmake/Modules/LocalJpeg.cmake
wantehchang: Thank you.

PR #1974, opened 2024-01-29
Simpler itemID finding for alpha.

vrabaud:
Backport https://github.com/AOMediaCodec/libavif/commit/4378794207f7a244bc36730c63dcc2bc188b6b3c

PR #1996, opened 2024-02-06
Cherry pick unused item

wantehchang:


wantehchang:
@vrabaud Vincent: FYI.

@vigneshvg @y-guyon Vignesh, Yannis: The two commits in this pull request will be merged as separate commits. The first commit is a revert of https://github.com/AOMediaCodec/libavif/pull/1974. The second commit is a cherry-pick of https://github.com/AOMediaCodec/libavif/pull/1922. You only need to review the second commit.

src/read.c
```diff
@@ -3707,25 +3695,31 @@ static avifResult avifDecoderDataFindAlphaItem(avifDecoderData * data,
...
+            ++newItemID;
+            isUsed = AVIF_FALSE;
+            for (uint32_t i = 0; i < colorItem->meta->items.count; ++i) {
+                if (colorItem->meta->items.item[i]->id == newItemID) {
+                    isUsed = AVIF_TRUE;
+                    break;
+                }
+            }
+        } while (isUsed && newItemID != 0);
+        *alphaItem = avifMetaFindItem(colorItem->meta, newItemID); // Create new empty item.
```
y-guyon, line 3716: This is fine because

https://github.com/AOMediaCodec/libavif/blob/e6edfd7f04ca7fd9305401d04a7a853208cc73c4/src/read.c#L773-L777

y-guyon (approved):
Thank you

src/read.c
wantehchang: Yes. If `newItemID` is 0, `avifMetaFindItem(colorItem->meta, newItemID)` fails and returns `NULL`. So we will set `result` to `AVIF_RESULT_OUT_OF_MEMORY` in the next line. `AVIF_RESULT_OUT_OF_MEMORY` is not the right error code for this condition, but at least we will fail rather than keep going.

However, if all the item IDs in the `colorItem->meta->items` array are nonzero and distinct (I believe this is an invariant that we maintain), then `newItemID` cannot possibly be zero. So we could add `assert(newItemID != 0)` after the do-while loop to indicate our belief.

Note: I suggested removing `&& newItemID != 0` from the condition of the do-while loop, but Vincent was uncomfortable with a potential infinite loop if we fail to maintain the invariant.

y-guyon:
Related: https://github.com/AOMediaCodec/libavif/pull/1993

src/read.c
wantehchang: I just realized that my previous comment is wrong:

> However, if all the item IDs in the `colorItem->meta->items` array are nonzero and distinct (I believe this is an invariant that we maintain), then `newItemID` cannot possibly be zero. [...]

All we need to prove `newItemID` cannot be zero is that `colorItem->meta->items.count < UINT32_MAX - 1`. If there are fewer than `UINT32_MAX - 1` elements in the `colorItem->meta->items` array, then there must be a nonzero `uint32_t` value that is not an item ID in the `colorItem->meta->items` array.

Since the `colorItem->meta->items.count < UINT32_MAX - 1` condition is guaranteed by the check at line 3700, the only way the do-while loop may loop infinitely is that the loop itself has a bug. The do-while loop is short and can be reviewed thoroughly. So I think the `&& newItemID != 0` check can be safely omitted.

PR #1981, opened 2024-01-31
Update the version to 1.0.4 (devel flag on)

y-guyon:
See #1850 for the corresponding pull request for the libavif 1.0.3 release.

PR #2001, opened 2024-02-08
Remove duplicate bug in CHANGELOG.md

wantehchang:


wantehchang:
This is the main branch's equivalent of https://github.com/AOMediaCodec/libavif/pull/1991.

y-guyon (approved):
I forgot that when rebasing #1981, thank you.

Note that it was also done in #1990.

PR #2009, opened 2024-02-13
Speed rgbtoyuvtest up

y-guyon:
Please review only the second commit.


tests/gtest/avifrgbtoyuvthreadingtest.cc
```diff
@@ -0,0 +1,124 @@
+// Copyright 2022 Google LLC
```
vrabaud, line 1: 2024

y-guyon: The file was copied in the first commit so we should keep the copyright as is I think.

vrabaud:
Please keep the removed test: maybe DISABLE them.

y-guyon:
> Please keep the removed tests: maybe DISABLE them.

They are already tested with min/max rgb values:

https://github.com/AOMediaCodec/libavif/blob/b10d2697e9ed2fb09cb722335ff4342c353612b8/tests/gtest/avifrgbtoyuvtest.cc#L296-L349

I do not think it is necessary to add a DISABLED test suite instantiation.

vrabaud (approved):

PR #2010, opened 2024-02-13
Add tests/test_all_configurations.sh

y-guyon:
Can be used as a tool to run sanitized tests before tagging a release.  
Can be heavily improved but I suggest to commit it as is and send PRs afterwards.


tests/test_all_configurations.sh
```diff
@@ -0,0 +1,97 @@
...
+  mkdir build_${BUILD_TYPE}
+  pushd build_${BUILD_TYPE}
+    cmake .. \
+     -DAVIF_BUILD_APPS=ON -DAVIF_BUILD_EXAMPLES=ON \
+     -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON -DAVIF_LOCAL_GTEST=ON \
+     -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON \
+     -DAVIF_CODEC_DAV1D=ON -DAVIF_LOCAL_DAV1D=ON \
+     -DAVIF_LOCAL_LIBYUV=ON -DAVIF_LOCAL_LIBSHARPYUV=ON \
+     -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DBUILD_SHARED_LIBS=OFF \
+     -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
```
wantehchang, line 54: It doesn't seem necessary to force the use of Clang.

If we want to do that, we should set the appropriate environment variables (`CC` and `CXX`?) so that Clang is also used to build the dependencies in the ext/ directory, in particular googletest.

tests/test_all_configurations.sh
```diff
@@ -0,0 +1,97 @@
...
+  popd
+
+  mkdir build_${BUILD_TYPE}
+  pushd build_${BUILD_TYPE}
+    cmake .. \
+     -DAVIF_BUILD_APPS=ON -DAVIF_BUILD_EXAMPLES=ON \
+     -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON -DAVIF_LOCAL_GTEST=ON \
+     -DAVIF_CODEC_AOM=ON -DAVIF_LOCAL_AOM=ON \
+     -DAVIF_CODEC_DAV1D=ON -DAVIF_LOCAL_DAV1D=ON \
+     -DAVIF_LOCAL_LIBYUV=ON -DAVIF_LOCAL_LIBSHARPYUV=ON \
```
wantehchang, line 52: For the main branch, these lines should use the new form (`-DAVIF_CODEC_AOM=LOCAL`, etc.). I guess you want to cherry-pick this to the v1.0.x release branch and then change this file to the new form?

tests/test_all_configurations.sh
```diff
@@ -0,0 +1,97 @@
...
+     -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DBUILD_SHARED_LIBS=OFF \
+     -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
+    cmake --build . --parallel
+    ctest . -j $(nproc)
+  popd
+
+  # The memory sanitizer triggers too many use-of-uninitialized-value errors
+  # (possibly false positives because standard libraries were not instrumented).
```
wantehchang, line 60: Yes.

Also, we may need to build libaom and dav1d with assembly code disabled. (SIMD intrinsics are okay.)

tests/test_all_configurations.sh
```diff
@@ -0,0 +1,97 @@
...
+  for SANITIZER in address thread undefined; do
+    pushd ext
+      rm -rf aom
+      CC=clang CXX=clang++ CFLAGS=-fsanitize=${SANITIZER} CXXFLAGS=-fsanitize=${SANITIZER} LDFLAGS=-fsanitize=${SANITIZER} \
+      ./aom.cmd
```
wantehchang, line 65: For libaom, the best way is to patch aom.cmd so that we pass `-DSANITIZE=${SANITIZER}` to the cmake command. This should replace the `CFLAGS`, `CXXFLAGS`, and `LDFLAGS` overrides.

tests/test_all_configurations.sh
```diff
@@ -0,0 +1,97 @@
...
+      rm -rf dav1d
+      CC=clang CXX=clang++ CFLAGS=-fsanitize=${SANITIZER} CXXFLAGS=-fsanitize=${SANITIZER} LDFLAGS=-fsanitize=${SANITIZER} \
+      ./dav1d.cmd
```
wantehchang, line 68: I remember the dav1d build system also has a config flag for the sanitizers. I suggest we imitate how the tests/oss-fuzz/build.sh script builds these dependencies.

tests/test_all_configurations.sh
```diff
@@ -0,0 +1,97 @@
...
+      rm -rf libyuv
+      # libyuv is not instrumented with ubsan because it generates linker errors.
+      # CC=clang CXX=clang++ CFLAGS=-fsanitize=${SANITIZER} CXXFLAGS=-fsanitize=${SANITIZER} LDFLAGS=-fsanitize=${SANITIZER} \
```
wantehchang, line 71: I figured out a solution to this problem. Please see https://github.com/AOMediaCodec/libavif/pull/2038.

We may need to cherry-pick that pull request to the v1.0.x release branch, in case we need to create v1.0.5.

wantehchang (approved):
Yannis: Sorry about the late review. You can check in this version first and improve it in follow-up pull requests.

tests/test_all_configurations.sh
y-guyon: Yes. `tests/oss-fuzz/build.sh` changed since I wrote this PR.

tests/test_all_configurations.sh
y-guyon: Why is it "the best way"?

tests/test_all_configurations.sh
y-guyon: > Also, we may need to build libaom and dav1d with assembly code disabled. (SIMD intrinsics are okay.)

I did not do that because msan is not used.

tests/test_all_configurations.sh
wantehchang: It is the best way in the sense that it is how the libaom team builds libaom for sanitizers:
https://aomedia.googlesource.com/aom/+/master/README.md#sanitizers

So it is guaranteed to work. Perhaps it would be better to call it the "official way".

tests/test_all_configurations.sh
```diff
@@ -0,0 +1,97 @@
...
+  for SANITIZER in address thread undefined; do
+    pushd ext
+      rm -rf aom
+      CC=clang CXX=clang++ CFLAGS=-fsanitize=${SANITIZER} CXXFLAGS=-fsanitize=${SANITIZER} LDFLAGS=-fsanitize=${SANITIZER} \
+      ./aom.cmd
+      rm -rf dav1d
+      cp dav1d.cmd dav1d_sanitized.cmd
+      sed -i 's/meson setup \(.*\) \.\./meson setup \1 '"-Db_sanitize=${SANITIZER} -Db_lundef=false"' ../g' dav1d_sanitized.cmd
+      CC=clang CXX=clang++ CFLAGS=-fsanitize=${SANITIZER} CXXFLAGS=-fsanitize=${SANITIZER} LDFLAGS=-fsanitize=${SANITIZER} \
+      ./dav1d_sanitized.cmd
```
wantehchang, line 69: The previous line (line 68) just needs to set `CC` and `CXX` now, so we can just say:

```
      CC=clang CXX=clang++ ./dav1d_sanitized.cmd
``

wantehchang (approved):

tests/test_all_configurations.sh
y-guyon: It does not seem to be done in https://github.com/AOMediaCodec/libavif/blob/main/tests/oss-fuzz/build.sh so leaving it as is for now.

PR #2013, opened 2024-02-15
Fix std::min() conflicting types in fuzz_dec_incr

y-guyon:
See b/308013905

### DIFF
--- a/tests/gtest/avif_fuzztest_dec_incr.cc
+++ b/tests/gtest/avif_fuzztest_dec_incr.cc
@@ -3,6 +3,7 @@
 // Compare non-incremental and incremental decode results of an arbitrary byte
 // sequence.
 
+#include <algorithm>
 #include <cstdint>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


vrabaud (approved):

tests/gtest/avif_fuzztest_dec_incr.cc
```diff
@@ -37,8 +38,10 @@ avifResult AvifIoRead(struct avifIO* io, uint32_t read_flags, uint64_t offset,
     return AVIF_RESULT_IO_ERROR;
   }
   out->data = data->available_bytes + offset;
-  out->size = std::min(size, data->available_size - offset);
-  data->read_size = std::max(data->read_size, offset + out->size);
+  out->size =
+      std::min(size, data->available_size - static_cast<size_t>(offset));
+  data->read_size =
+      std::max(data->read_size, static_cast<size_t>(offset) + out->size);
```
wantehchang, line 44: I verified that it is safe to cast `offset` to `size_t` (because of the `data->available_size < offset` check at line 37) and the `static_cast<size_t>(offset) + out->size` addition won't overflow (because the sum is at most `data->available_size`).

PR #2018, opened 2024-02-19
Build AOM and dav1d differently for msan.

vrabaud:
That should fix the oss-fuzz builds, cf https://oss-fuzz-build-logs.storage.googleapis.com/log-c171740d-4c8e-42bf-b2dc-55d83472d8e7.txt

### DIFF
--- a/tests/oss-fuzz/build.sh
+++ b/tests/oss-fuzz/build.sh
@@ -36,20 +36,53 @@
 #     avif_fuzztest_enc_dec_incr@EncodeDecodeAvifFuzzTest.EncodeDecodeGridValid \
 #     --sanitizer address
 
-# Reset compile flags to build dav1d without fuzzer flags. The meson build system
-# is problematic with sanitizer flags.
+# Build dav1d with sanitizer flags.
+# This duplicates dav1d.cmd but includes extra flags: -Db_sanitize=$SANITIZER -Db_lundef=false,
+# and -Denable_asm=false for msan
+cd ext
+git clone -b 1.4.0 --depth 1 https://code.videolan.org/videolan/dav1d.git
+cd dav1d
+mkdir build
+cd build
+export DAV1D_FLAGS="--default-library=static --buildtype release -Denable_tools=false -Denable_tests=false -Db_sanitize=$SANITIZER -Db_lundef=false"
+if [ "$SANITIZER" == "memory" ]
+then
+  export DAV1D_FLAGS="${DAV1D_FLAGS} -Denable_asm=false"
+fi
+meson setup ${DAV1D_FLAGS} ..
+ninja
+cd ../../..
+
+# Reset compile flags to build libyuv without fuzzer flags.
 export ORIG_CFLAGS="$CFLAGS"
 export ORIG_CXXFLAGS="$CXXFLAGS"
 export CFLAGS=""
 export CXXFLAGS=""
 
-cd ext && bash dav1d.cmd && bash libyuv.cmd && cd ..
+cd ext && bash libyuv.cmd && cd ..
 
 export CFLAGS=$ORIG_CFLAGS
 export CXXFLAGS=$ORIG_CXXFLAGS
 
+# Build libaom with sanitizer flags.
+# This duplicates aom.cmd but adds extra flags: -DAOM_TARGET_CPU=generic for msan.
+if [ "$SANITIZER" == "memory" ]
+then
+  cd ext
+  git clone -b v3.8.1 --depth 1 https://aomedia.googlesource.com/aom
+  cd aom
+  mkdir build.libavif
+  cd build.libavif
+  # Use -DAOM_TARGET_CPU=generic to disable assembly.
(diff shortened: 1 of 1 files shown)


tests/oss-fuzz/build.sh
```diff
@@ -36,20 +36,53 @@
 #     avif_fuzztest_enc_dec_incr@EncodeDecodeAvifFuzzTest.EncodeDecodeGridValid \
 #     --sanitizer address
 
-# Reset compile flags to build dav1d without fuzzer flags. The meson build system
-# is problematic with sanitizer flags.
+# Build dav1d with sanitizer flags.
+# This duplicates dav1d.cmd but includes extra flags: -Db_sanitize=$SANITIZER -Db_lundef=false,
+# and -Denable_asm=false for msan
+cd ext
+git clone -b 1.4.0 --depth 1 https://code.videolan.org/videolan/dav1d.git
```
y-guyon, line 43: We could also apply a patch to `dav1d.cmd` before running it, to keep most of the instructions, including the version. We would not forget to update it here too.  
However this is a temporary solution if I understood correctly so it is not worth the hassle.

y-guyon (approved):

tests/oss-fuzz/build.sh
vrabaud: Indeed, that is temporary until https://github.com/AOMediaCodec/libavif/pull/1744 but let's do it anyway.

y-guyon (approved):

tests/oss-fuzz/build.sh
```diff
@@ -36,21 +36,36 @@
 #     avif_fuzztest_enc_dec_incr@EncodeDecodeAvifFuzzTest.EncodeDecodeGridValid \
 #     --sanitizer address
 
-# Reset compile flags to build dav1d without fuzzer flags. The meson build system
-# is problematic with sanitizer flags.
+# Reset compile flags to build libyuv without fuzzer flags.
```
wantehchang, line 39: If we need to build libyuv without fuzzer flags, add a commet to explain why.

tests/oss-fuzz/build.sh
```diff
@@ -36,21 +36,36 @@
...
 export CXXFLAGS=""
 
-cd ext && bash dav1d.cmd && bash libyuv.cmd && cd ..
+cd ext && bash libyuv.cmd && cd ..
 
 export CFLAGS=$ORIG_CFLAGS
 export CXXFLAGS=$ORIG_CXXFLAGS
 
+# Build dav1d with sanitizer flags.
+# Adds extra flags: -Db_sanitize=$SANITIZER -Db_lundef=false, and -Denable_asm=false for msan
```
wantehchang, line 51: If the `-Db_lundef=false` flag is necessary for only certain sanitizers, it would be good to specifically check for those sanitizers.

But when I searched for the error message "undefined reference to `__afl_area_ptr'" on the Web and looked at the results, I think this linker error is caused by compiling dav1d with afl-clang-fast and compiling libavif with clang.

afl-clang-fast comes from the oss-fuzz environment:

```
Step #3 - "compile-afl-address-x86_64": ---------------------------------------------------------------
Step #3 - "compile-afl-address-x86_64": CC=/src/aflplusplus/afl-clang-fast
Step #3 - "compile-afl-address-x86_64": CXX=/src/aflplusplus/afl-clang-fast++
Step #3 - "compile-afl-address-x86_64": CFLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope 
Step #3 - "compile-afl-address-x86_64": CXXFLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope  -stdlib=libc++
Step #3 - "compile-afl-address-x86_64": RUSTFLAGS=--cfg fuzzing -Zsanitizer=address -Cdebuginfo=1 -Cforce-frame-pointers
Step #3 - "compile-afl-address-x86_64": ---------------------------------------------------------------
```

while clang comes from the `-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++` cmake options at line 80.

Could you try removing `-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++` and removing `-Db_lundef=false`?

tests/oss-fuzz/build.sh
vrabaud: There was a good reason at some point: it simply did not build. It builds now with the common fuzzer flags so let's give it a try, thx for the PR.

tests/oss-fuzz/build.sh
wantehchang: Vincent: I tested the changes I suggested locally. They passed the `check_build` step. To be safe, I will make the changes in two steps.

Here is the first pull request to remove `-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++` (https://github.com/AOMediaCodec/libavif/pull/2028), which I am confident is correct (unless I miss something subtle).

vrabaud: Indeed, it is now fine

PR #2022, opened 2024-02-20
Compare the item contentType string using strcmp()

wantehchang:
In avifDecoderFindMetadata(), compare the item contentType string using strcmp() instead of memcmp(). item->contentType.contentType is a null-terminated string, so the bytes after the terminating null character are uninitialized.

Add the AVIF_CONTENT_TYPE_XMP_SIZE macro, defined as the size of AVIF_CONTENT_TYPE_XMP including the terminating null character.

BUG=oss-fuzz:66848

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -33,9 +33,6 @@
 // }
 static const size_t VISUALSAMPLEENTRY_SIZE = 78;
 
-static const char xmpContentType[] = AVIF_CONTENT_TYPE_XMP;
-static const size_t xmpContentTypeSize = sizeof(xmpContentType);
-
 // The only supported ipma box values for both version and flags are [0,1], so there technically
 // can't be more than 4 unique tuples right now.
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -33,9 +33,6 @@
 // }
 static const size_t VISUALSAMPLEENTRY_SIZE = 78;
 
-static const char xmpContentType[] = AVIF_CONTENT_TYPE_XMP;
-static const size_t xmpContentTypeSize = sizeof(xmpContentType);
```
wantehchang, line 37: Yannis: It seems that we can use `AVIF_CONTENT_TYPE_XMP` and `sizeof(AVIF_CONTENT_TYPE_XMP)` directly. It doesn't seem necessary to copy `AVIF_CONTENT_TYPE_XMP` to a static variable (`xmpContentType`).

I hope I didn't miss something subtle...

PR #2024, opened 2024-02-21
Build libyuv with fuzzer flags

wantehchang:
This fixes false positive MSan use-of-uninitialized-value warnings when libyuv functions are involved.

With this change, all dependencies are built with fuzzer flags.

BUG=oss-fuzz:66845
BUG=oss-fuzz:66846

### DIFF
--- a/tests/oss-fuzz/build.sh
+++ b/tests/oss-fuzz/build.sh
@@ -36,17 +36,6 @@
 #     avif_fuzztest_enc_dec_incr@EncodeDecodeAvifFuzzTest.EncodeDecodeGridValid \
 #     --sanitizer address
 
-# Reset compile flags to build libyuv without fuzzer flags.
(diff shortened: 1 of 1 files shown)


jzern (approved):

wantehchang:
Vincent: If this pull request looks good to you, please merge it for me. Thanks.

vrabaud:
Thx, let's see if that now works.

wantehchang:
Vincent: I think I know why you did not build libyuv with fuzzer flags. When running `python infra/helper.py build_fuzzers` with `--sanitizer undefined`, I got linker errors when linking the aviftest and avifyuv test programs.

I think this is the same issue discussed in https://github.com/google/oss-fuzz/issues/713. The final suggestion there is to link the C program with a C++ library using the C++ compiler. I did that in https://github.com/AOMediaCodec/libavif/pull/2038.

PR #2028, opened 2024-02-21
Do not set CMAKE_C_COMPILER and CMAKE_CXX_COMPILER

wantehchang:
See https://google.github.io/oss-fuzz/getting-started/new-project-guide/#Requirements

Passing the -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ options to cmake ignores the CC and CXX environment variables set by oss-fuzz.

PR #2029, opened 2024-02-22
Remove -Db_lundef=false from DAV1D_EXTRA_FLAGS

wantehchang:
The -Db_lundef=false flag does not seem necessary if we use the compilers specified in the oss-fuzz environment as the linkers.

This pull request depends on
https://github.com/AOMediaCodec/libavif/pull/2028.

### DIFF
--- a/tests/oss-fuzz/build.sh
+++ b/tests/oss-fuzz/build.sh
@@ -37,10 +37,10 @@
 #     --sanitizer address
 
 # Build dav1d with sanitizer flags.
-# Adds extra flags: -Db_sanitize=$SANITIZER -Db_lundef=false, and -Denable_asm=false for msan
+# Adds extra flags: -Db_sanitize=$SANITIZER, and -Denable_asm=false for msan
 if [ "$SANITIZER" != "coverage" ] && [ "$SANITIZER" != "introspector" ]
 then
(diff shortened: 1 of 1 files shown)


vrabaud:
It does not seem necessary but there is a warning at the end of the meson setup otherwise:
```
../tests/meson.build:31: WARNING: Trying to use address sanitizer on Clang with b_lundef.
This will probably not work.
Try setting b_lundef to false instead.

dav1d 1.4.0

  User defined options
    buildtype      : release
    default_library: static
    b_sanitize     : address
    enable_tests   : false
    enable_tools   : false

Found ninja-1.11.1.git.kitware.jobserver-1 at /usr/local/bin/ninja
```

vrabaud (approved):

vrabaud:
Please rebase because of https://github.com/AOMediaCodec/libavif/pull/2030

wantehchang:
Vincent: Thank you. I didn't know `-Db_lundef=false` was suggested by a warning message from meson setup. Then we should keep it. I will look into it.

wantehchang:
Vincent: I looked into this. I documented my findings in https://github.com/AOMediaCodec/libavif/pull/2035. I will abandon this pull request.

PR #2030, opened 2024-02-22
Get fuzz tests to build in coverage and introspector mode.

vrabaud:
They are not proper sanitizers so dav1d is confused and does not build


y-guyon (approved):

y-guyon:
List of current possible values:

https://github.com/google/oss-fuzz/blob/12e97b9c3334b6e6406bc0abe45b5e30703849d1/infra/constants.py#L37-L46

Issue #2033, opened 2024-02-22
ci-windows failure "unknown file: error: SEH exception with code 0xc000001d thrown in the test body"


y-guyon:
Note that [ci-windows](https://github.com/y-guyon/libavif/actions/workflows/ci-windows.yml) has been failing since https://github.com/AOMediaCodec/libavif/pull/2032.

It may be unrelated to that PR because the [previous passing test](https://github.com/AOMediaCodec/libavif/actions/runs/8007780983/job/21872700753) used the cache while this PR used a [full workflow](https://github.com/AOMediaCodec/libavif/actions/runs/8007654445/job/21872284438), so it may be cache-related instead.

The errors seem consistent across across the forks of this repository:
```
unknown file: error: SEH exception with code 0xc000001d thrown in the test body.

The following tests FAILED:
	 11 - avifchangesettingtest (Failed)
	 17 - avifgainmaptest (Failed)
	 24 - aviflosslesstest (Failed)
	 39 - avify4mtest (Failed)
	 46 - test_cmd_progressive (Failed)
```

y-guyon:
A possibly unrelated log entry is

```
Error: Svt[error]: deinit called without sending EOS!
```

jzern:
I think that's a code for illegal instruction. Visual Studio 2022 has an issue starting at 17.8.0 which might cause AVX-512 instructions to be emitted in non-AVX-512 code. This was [labeled fixed in 17.10 Preview 1](https://developercommunity.visualstudio.com/t/Illegal-Instruction-AVX512/10511452).

wantehchang:
Thanks for filing this bug report.

We can consider compiling libaom with clang-cl until this Visual Studio 2022 bug is fixed. I will write a pull request to try that. (I remember I wrote an experimental pull request to do this before.)

wantehchang:
I found my experimental pull request: https://github.com/AOMediaCodec/libavif/pull/1933. It was for AppVeyor.

PR #2034, opened 2024-02-22
Compile libaom and libyuv with clang-cl

wantehchang:
The libaom change is temporary, until the Visual Studio 2022 bug is fixed.

The libyuv change is permanent.

Fix https://github.com/AOMediaCodec/libavif/issues/2033.

### DIFF
--- a/.github/workflows/ci-windows-artifacts.yml
+++ b/.github/workflows/ci-windows-artifacts.yml
@@ -26,7 +26,7 @@ jobs:
       uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
       with:
         path: ext
-        key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}-releasedeps
+        key: ${{ runner.os }}-${{ hashFiles('ext/*.cmd') }}-releasedeps-v2
     - name: Print cmake version
       run: cmake --version
     - uses: ilammy/setup-nasm@13cbeb366c45c4379d3478cdcbadd8295feb5028 # v1.5.1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-windows.yml
```diff
@@ -62,6 +62,12 @@ jobs:
       if: steps.cache-ext.outputs.cache-hit != 'true'
       working-directory: ./ext
       run: ./aom.cmd
+      # Visual Studio 2022 has an issue starting at 17.8.0 which might cause
+      # AVX-512 instructions to be emitted in non-AVX-512 code. See
+      # https://github.com/AOMediaCodec/libavif/issues/2033#issuecomment-1960062751.
+      env:
```
jzern, line 68: order doesn't matter, but to me `run:` makes sense to be last, since the other fields affect it.

jzern (approved):

wantehchang:
James, Yannis: Please review the new version of this pull request.

Note that I added `-v2` to the names of the cache files. Should we try to delete the old cache files so that we don't need to rename the cache files?

.github/workflows/ci-windows-artifacts.yml
```diff
@@ -38,26 +38,25 @@ jobs:
...
-        $Env:CXX = 'clang-cl'
-
-        ./libyuv.cmd
-
-        $Env:CC = ''
-
-        $Env:CXX = ''
+      env:
+        CC: clang-cl
+        CXX: clang-cl
```
wantehchang, line 59: Yannis: I found the `env` key, documented at https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow.

wantehchang:
I verified that only libaom and libyuv were compiled with clang-cl; everything else was compiled with MSVC:
https://github.com/AOMediaCodec/libavif/actions/runs/8009956432/job/21879809755?pr=2034

jzern (approved):

y-guyon:
> Note that I added `-v2` to the names of the cache files. Should we try to delete the old cache files so that we don't need to rename the cache files?

I do not think it matters much.

PR #2035, opened 2024-02-22
ext/dav1d.cmd: Add -Db_lundef=false to comments

wantehchang:
The need to use -Db_lundef=false with the sanitizers in Clang is documented in the following pages:
https://clang.llvm.org/docs/AddressSanitizer.html#usage
https://clang.llvm.org/docs/MemorySanitizer.html#usage

If -Db_sanitize=<sanitizer> is used without -Db_lundef=false and the compiler is Clang, meson emits a warning message like this:
  ../tests/meson.build:31: WARNING: Trying to use address sanitizer on
  Clang with b_lundef.
  This will probably not work.
  Try setting b_lundef to false instead.

Note: The change to ext/dav1d.cmd allows us to remove the "-v2" suffix (added in https://github.com/AOMediaCodec/libavif/pull/2034) from the cache file names.

### DIFF
--- a/.github/workflows/ci-windows-artifacts.yml
+++ b/.github/workflows/ci-windows-artifacts.yml
@@ -26,7 +26,7 @@ jobs:
       uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
       with:
         path: ext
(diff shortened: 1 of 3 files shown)


wantehchang:
@vrabaud @jzern Please review.

@y-guyon Just FYI. Note the removal of the "-v2" suffix from the cache file names used in the Windows CI workflows.

vrabaud (approved):

PR #2032, opened 2024-02-22
Bump fuzztest

vrabaud:
BUG=oss-fuzz:66560
b/324876907

PR #2036, opened 2024-02-23
Revert to an older fuzztest version

vrabaud:


wantehchang (approved):
LGTM. Thanks.

wantehchang:
For future reference: This pull request fixes the problem reported in https://github.com/AOMediaCodec/libavif/pull/2032#issuecomment-1960652819.

PR #2041, opened 2024-02-26
Export EXTRA_CMAKE_FLAGS to remove the "unbound variable" error

vrabaud:
BUG=oss-fuzz:67018


tests/oss-fuzz/build.sh
```diff
@@ -75,6 +75,8 @@ if [ "$FUZZING_ENGINE" == "libfuzzer" ]
 then
   export CXXFLAGS="${CXXFLAGS} -DFUZZTEST_COMPATIBILITY_MODE"
   export EXTRA_CMAKE_FLAGS="-DAVIF_ENABLE_FUZZTEST=ON -DFUZZTEST_COMPATIBILITY_MODE=libfuzzer"
+else
+  export EXTRA_CMAKE_FLAGS=""
```
y-guyon, line 79: No need to revert `CXXFLAGS`?

y-guyon (approved):

tests/oss-fuzz/build.sh
vrabaud: nope, it is already given by the fuzzer (with all the -fsanitize ...)

jzern: I don't think you need to export EXTRA_CMAKE_FLAGS; no subprocess depends on the value. You can just assign it to something.

If CXXFLAGS is already exported, assigning to it will update the exported value as well. Though it doesn't hurt to be explicit.

vrabaud: Fixed in https://github.com/AOMediaCodec/libavif/pull/2044

PR #2042, opened 2024-02-26
Add --icc flag to avifdec

y-guyon:
Same meaning as in avifenc.
Move readEntireFile() from avifenc to avifutil.


vrabaud (approved):
Please add a changelog entry

PR #2044, opened 2024-02-28
Do not export env variables, just assign to them.

vrabaud:


jzern:
The commit message made me first think all the variables are environment variables, but only CXXFLAGS is here. To distinguish the two you could make the non-constant/environment variables lowercase, but that's a cosmetic for another change.

jzern (approved):

tests/oss-fuzz/build.sh
```diff
@@ -40,13 +40,14 @@ set -eu
 
 # Build dav1d with sanitizer flags.
 # Adds extra flags: -Db_sanitize=$SANITIZER -Db_lundef=false, and -Denable_asm=false for msan
+DAV1D_EXTRA_FLAGS=""
 if [ "$SANITIZER" != "coverage" ] && [ "$SANITIZER" != "introspector" ]
 then
-  export DAV1D_EXTRA_FLAGS="-Db_sanitize=$SANITIZER -Db_lundef=false"
+  DAV1D_EXTRA_FLAGS="-Db_sanitize=$SANITIZER -Db_lundef=false"
```
wantehchang, line 46: Optional: Since `DAV1D_EXTRA_FLAGS` has been initialized an empty string, we can append to it here.

tests/oss-fuzz/build.sh
```diff
@@ -73,10 +74,10 @@ mkdir build
 cd build
 if [ "$FUZZING_ENGINE" == "libfuzzer" ]
 then
-  export CXXFLAGS="${CXXFLAGS} -DFUZZTEST_COMPATIBILITY_MODE"
-  export EXTRA_CMAKE_FLAGS="-DAVIF_ENABLE_FUZZTEST=ON -DFUZZTEST_COMPATIBILITY_MODE=libfuzzer"
+  CXXFLAGS="${CXXFLAGS} -DFUZZTEST_COMPATIBILITY_MODE"
+  EXTRA_CMAKE_FLAGS="-DAVIF_ENABLE_FUZZTEST=ON -DFUZZTEST_COMPATIBILITY_MODE=libfuzzer"
```
wantehchang, line 78: Optional: We can also initialize `EXTRA_CMAKE_FLAGS` to an empty string before the if statement, and append to it here. This will allow us to get rid of the else branch.

wantehchang (approved):
Thanks for the cleanup. I agree that the commit message could have a summary of the changes below the subject line.

tests/oss-fuzz/build.sh
vrabaud: Done

tests/oss-fuzz/build.sh
vrabaud: Done

PR #2047, opened 2024-03-04
Rename imageSizeLimit to sizeLimit in avifJPEGRead

y-guyon:
Also add the `((uint64_t)standardXMPSize + totalExtendedXMPSize) > SIZE_MAX` check back in case `sizeLimit` changes.

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -48,14 +48,14 @@ static void my_error_exit(j_common_ptr cinfo)
 
 // An internal function used by avifJPEGReadCopy(), this is the shared libjpeg decompression code
 // for all paths avifJPEGReadCopy() takes.
-static avifBool avifJPEGCopyPixels(avifImage * avif, uint32_t imageSizeLimit, struct jpeg_decompress_struct * cinfo)
+static avifBool avifJPEGCopyPixels(avifImage * avif, uint32_t sizeLimit, struct jpeg_decompress_struct * cinfo)
 {
     cinfo->raw_data_out = TRUE;
     jpeg_start_decompress(cinfo);
 
     avif->width = cinfo->image_width;
     avif->height = cinfo->image_height;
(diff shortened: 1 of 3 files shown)


apps/shared/avifjpeg.c
```diff
@@ -996,11 +996,11 @@ static avifBool avifJPEGReadInternal(FILE * f,
                     goto cleanup;
                 }
 
-                if (marker->data_length - AVIF_JPEG_EXIF_HEADER_LENGTH > imageSizeLimit) {
+                if (marker->data_length - AVIF_JPEG_EXIF_HEADER_LENGTH > sizeLimit) {
```
wantehchang, line 999: In line 1001 and line 1079, change "px" to "bytes" in the error messages.

apps/shared/avifjpeg.h
```diff
@@ -11,7 +11,8 @@ extern "C" {
 #endif
 
 // Decodes the jpeg file at path 'inputFilename' into 'avif'.
-// At most imageSizeLimit pixels will be read or an error returned.
+// At most sizeLimit pixels will be read or an error returned. At most sizeLimit
+// bytes of Exif or XMP metadata will be read or an error returned.
```
wantehchang, line 15: [This is just a comment.] Thank you for documenting the two purposes of `sizeLimit`. Using `sizeLimit` for both the (uncompressed) image area and the Exif/XMP metadata size still has a conceptual difficulty: these two quantities have different units. The image area is in pixels, and the Exif/XMP metadata size is in bytes. In addition to the comment, we can clarify this by noting that `sizeLimit` ultimately limits the amount of memory used by the function.

apps/shared/avifjpeg.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2048

PR #2050, opened 2024-03-12
Experimental AVIF Sample Transform

y-guyon:
Allow encoding and decoding images of more than 12 bits, the limit of the AV1 format. This commit experiments 16-bit AVIF images, made of one primary lossless 8-bit image and one hidden lossy/lossless bit depth extension 8-bit image. The two are combined thanks to a Sample Transform derived image item. Alpha is supported.

Warning: This feature is experimental and not covered by the current AVIF specification.


src/read.c
```diff
@@ -1500,41 +1517,58 @@ static avifResult avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifI
     return AVIF_RESULT_OK;
 }
 
-// Allocates the dstImage based on the grid image requirements. Also verifies some spec compliance rules for grids.
-static avifResult avifDecoderDataAllocateGridImagePlanes(avifDecoderData * data, const avifTileInfo * info, avifImage * dstImage)
+// Allocates the dstImage. Also verifies some spec compliance rules for grids, if relevant.
+static avifResult avifDecoderDataAllocateImagePlanes(avifDecoderData * data, const avifTileInfo * info, avifImage * dstImage)
```
wantehchang, line 1521: The changes to this function are an example of general cleanup/improvements that can be reviewed and checked in first.

wantehchang (commented):
Yannis: I made a first pass through the pull request, checking only that the pull request is essentially a no-op if `AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM` is not defined.

I noticed some general cleanup or improvements in src/read.c and src/write.c. I suggest moving those changes and the new `avifIsAlpha()` function to a separate pull request that can be reviewed and checked in first.

The first version of  `avifIsAlpha()` function would be simply:

```
avifBool avifIsAlpha(avifItemCategory itemCategory)
{
    if (itemCategory == AVIF_ITEM_ALPHA) {
        return AVIF_TRUE;
    }
    return AVIF_FALSE;
}
```

src/write.c
```diff
@@ -2206,8 +2528,9 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
         avifEncoderItem * item = &encoder->data->items.item[itemIndex];
         const avifBool isGrid = (item->gridCols > 0);
         const avifBool isToneMappedImage = !memcmp(item->type, "tmap", 4);
+        const avifBool isSampleTransformImage = !memcmp(item->type, "sato", 4);
         memset(&item->ipma, 0, sizeof(item->ipma));
-        if (!item->codec && !isGrid && !isToneMappedImage) {
+        if (!item->codec && !isGrid && !isToneMappedImage && !isSampleTransformImage) {
```
wantehchang, line 2533: Optional: If would be good to avoid the `isToneMappedImage` and `isSampleTransformImage` code completely when the respective experimental feature flag is disabled. Here is one way to do that:

```
        const avifBool isGrid = (item->gridCols > 0);
        // Whether there is ipma to write for this item
        avifBool hasIpmaToWrite = item->codec || isGrid;
        const avifBool isToneMappedImage = !memcmp(item->type, "tmap", 4);
        if (isToneMappedImage) {
            hasIpmaToWrite = AVIF_TRUE;
        }
#endif
#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
        const avifBool isSampleTransformImage = !memcmp(item->type, "sato", 4);
        if (isSampleTransformImage) {
            hasIpmaToWrite = AVIF_TRUE;
        }
#endif
        memset(&item->ipma, 0, sizeof(item->ipma));
        if (!hasIpmaToWrite) {
            continue;
        }
```

src/read.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2056, https://github.com/AOMediaCodec/libavif/pull/2057, https://github.com/AOMediaCodec/libavif/pull/2058, https://github.com/AOMediaCodec/libavif/pull/2059

wantehchang: Thanks a lot!

y-guyon:
Rebased.

src/write.c
y-guyon: Done.

wantehchang:
Yannis: I have just started reviewing this pull request. I may not be able to post comments until tomorrow. I will mainly verify that the pull request is essentially a no-op if `AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM` is not defined.

include/avif/avif.h
```diff
@@ -706,6 +710,39 @@ AVIF_NODISCARD AVIF_API avifBool avifGainMapMetadataFractionsToDouble(avifGainMa
 
 #endif // AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
 
+// ---------------------------------------------------------------------------
+
+#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
+// Sample Transforms are a HIGHLY EXPERIMENTAL FEATURE. The format might still
+// change and images containing a sample transform item encoded with the current
+// version of libavif might not decode with a future version of libavif.
+// Use are your own risk.
```
wantehchang, line 719: Typo: are => at

include/avif/avif.h
```diff
@@ -706,6 +710,39 @@ AVIF_NODISCARD AVIF_API avifBool avifGainMapMetadataFractionsToDouble(avifGainMa
...
+// This is based on a proposal from the Alliance for Open Media.
+
+typedef enum avifSampleTransformRecipe
+{
+    AVIF_SAMPLE_TRANSFORM_NONE,
+    // Encode the 8 most significant bits of each input image sample losslessly
+    // into one base image. The remaining least 8 significant bits are encoded
```
wantehchang, line 726: Nit/optional: one => a

Nit: least 8 => 8 least

include/avif/avif.h
```diff
@@ -706,6 +710,39 @@ AVIF_NODISCARD AVIF_API avifBool avifGainMapMetadataFractionsToDouble(avifGainMa
...
+    // in a separate hidden image item. The two are combined at decoding into
+    // one image with the same bit depth as the original image.
+    // It is backward compatible in the sense that only the base image may be
+    // decoded (ignoring the hidden image item), leading to a valid image but
```
wantehchang, line 730: Nit/optional: "only the base image may be decoded" could be misinterpreted as "only the base image is allowed to be decoded".

One way to make this clearer is to rephrase it as "it is possible to decode only the base image".

include/avif/avif.h
```diff
@@ -706,6 +710,39 @@ AVIF_NODISCARD AVIF_API avifBool avifGainMapMetadataFractionsToDouble(avifGainMa
...
+    // with precision loss (16-bit samples truncated to the 8 most significant
+    // bits).
+    AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_8B_8B,
```
wantehchang, line 733: Make the same changes to the comments for this enum constant.

include/avif/avif.h
```diff
@@ -1458,6 +1495,11 @@ typedef struct avifEncoder
 #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
     int qualityGainMap; // changeable encoder setting
 #endif
+
+#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
+    // Perform extra steps at encoding and decoding to extend AV1 features as bundled additional image items.
```
wantehchang, line 1500: Nit/optional: change "as" to "with" or "using"?

Also, did you mean "AV1 features" or "AVIF features"?

src/write.c
```diff
@@ -1119,7 +1178,11 @@ static avifResult avifEncoderAddImageItems(avifEncoder * encoder,
 #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
                             : (itemCategory == AVIF_ITEM_GAIN_MAP) ? infeNameGainMap
 #endif
-                                                                   : infeNameColor;
+#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
+                            : (itemCategory >= AVIF_SAMPLE_TRANSFORM_MIN_CATEGORY && itemCategory <= AVIF_SAMPLE_TRANSFORM_MAX_CATEGORY)
+                                ? infeNameSampleTransform
+#endif
+                                : infeNameColor;
```
wantehchang, line 1185: The cascaded ternary `? : `operators are difficult to parse. This would be clearer if rewritten as if-else-if statements.

src/write.c
```diff
@@ -1381,6 +1635,8 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
     }
 
     if (hasGainMap) {
+        // AVIF supports 16-bit images through sample transforms used as bit depth extensions,
```
wantehchang, line 1638: Optional: Should we add "TODO" as a reminder?

wantehchang (commented):
I only reviewed up to avif.h carefully. The rest of the files I only verified that the changes are a no-op if the experimental flag is disabled.

I suggest some small changes (mostly comments).

include/avif/avif.h
y-guyon: I meant "AV1 features". These new "AVIF features" extend the AV1 ones.

include/avif/avif.h
```diff
@@ -706,6 +710,38 @@ AVIF_NODISCARD AVIF_API avifBool avifGainMapMetadataFractionsToDouble(avifGainMa
...
+    AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_8B_8B,
+    // Encode the 12 most significant bits of each input image sample losslessly
+    // into a base image. The remaining 4 least significant bits are encoded in
+    // a separate hidden image item. The two are combined at decoding into one
+    // image with the same bit depth as the original image. It is backward
+    // compatible in the sense that it is possible to decode only the base image
+    // (ignoring the hidden image item), leading to a valid image but with
+    // precision loss (16-bit samples truncated to the 12 most significant
+    // bits).
+    AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_4B
```
wantehchang, line 741: Just curious: are you planning to implement the residual bit depth extension?

include/avif/internal.h
```diff
@@ -156,6 +156,78 @@ void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage);
...
+    AVIF_SAMPLE_TRANSFORM_MSB = 12,       // S = L<=0 ? 0 : floor(log2(L))
+    AVIF_SAMPLE_TRANSFORM_POW = 13,       // S = L==0 ? 0 : pow(L, R)
+    AVIF_SAMPLE_TRANSFORM_MIN = 14,       // S = L<=R ? L : R
+    AVIF_SAMPLE_TRANSFORM_MAX = 15,       // S = L<=R ? R : L
+    AVIF_SAMPLE_TRANSFORM_RESERVED
+} avifSampleTransformTokenType;
+
+typedef struct avifSampleTransformToken
+{
+    uint8_t value;               // avifSampleTransformTokenType
```
wantehchang, line 198: Nit/optional: Why is this not named `type`?

include/avif/internal.h
```diff
@@ -156,6 +156,78 @@ void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage);
...
+    int32_t constant;            // If value is AVIF_SAMPLE_TRANSFORM_CONSTANT.
```
wantehchang, line 199: Optional: Document that only 32-bit (bit_depth=2) constant is supported.

src/read.c
```diff
@@ -2006,6 +2023,107 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
 }
 #endif // AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
+static avifResult avifParseSampleTransformTokens(avifROStream * s, avifSampleTransformExpression * expression)
```
wantehchang, line 2027: Document that this function assumes `bit_depth` is 2 (32 bits).

src/read.c
```diff
@@ -2006,6 +2023,107 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
...
+    AVIF_CHECK(avifROStreamRead(s, &tokenCount, /*size=*/1)); // unsigned int(8) token_count;
+    AVIF_CHECKERR(avifArrayCreate(expression, sizeof(expression->tokens[0]), tokenCount), AVIF_RESULT_OUT_OF_MEMORY);
+
+    for (uint32_t t = 0; t < tokenCount; ++t) {
+        avifSampleTransformToken * token = (avifSampleTransformToken *)avifArrayPush(expression);
+        AVIF_ASSERT_OR_RETURN(token != NULL);
+
+        AVIF_CHECK(avifROStreamRead(s, &token->value, /*size=*/1)); // unsigned int(8) token;
+        if (token->value == AVIF_SAMPLE_TRANSFORM_CONSTANT) {
+            // TODO(yguyon): Verify two's complement representation is guaranteed here.
```
wantehchang, line 2039: I think we have verified this.

src/read.c
```diff
@@ -2006,6 +2023,107 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
...
+            uint32_t constant;
+            AVIF_CHECK(avifROStreamReadU32(s, &constant)); // signed int(1<<(bit_depth+3)) constant;
+            token->constant = *(int32_t *)&constant;       // maybe =(int32_t)constant; is enough
```
wantehchang, line 2042: Yes, `=(int32_t)constant;` is enough. Alternatively, we can do this:

```
    AVIF_CHECK(avifROStreamReadU32(s, (uint32_t *)&token->constant)); // signed int(1<<(bit_depth+3)) constant;
```

src/read.c
```diff
@@ -2048,6 +2166,11 @@ static avifResult avifDecoderItemReadAndParse(const avifDecoder * decoder,
         *codecType = avifGetCodecType(item->type);
         AVIF_ASSERT_OR_RETURN(*codecType != AVIF_CODEC_TYPE_UNKNOWN);
     }
+    // Note: If AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM is defined, backward-incompatible files
```
wantehchang, line 2169: Optional: Should this be a TODO comment?

src/read.c
```diff
@@ -4981,6 +5136,92 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
+                    continue;
+                }
+                if (avifDecoderItemShouldBeSkipped(inputImageItem)) {
+                    avifDiagnosticsPrintf(data->diag, "Box[sato] input item %d is not a supported image type", inputImageItem->id);
+                    return AVIF_RESULT_DECODE_SAMPLE_TRANSFORM_FAILED;
+                }
+                // Input image item order is important because input image items are indexed according to this order.
+                AVIF_CHECKERR(inputImageItem->dimgIdx == data->sampleTransformNumInputImageItems, AVIF_RESULT_NOT_IMPLEMENTED);
+
+                avifItemCategory * category = &data->sampleTransformInputImageItems[data->sampleTransformNumInputImageItems];
```
wantehchang, line 5158: IMPORTANT: return an error if `data->sampleTransformNumInputImageItems >= AVIF_SAMPLE_TRANSFORM_MAX_NUM_INPUT_IMAGE_ITEMS`.

src/read.c
```diff
@@ -5361,6 +5619,62 @@ static avifBool avifDecoderDataFrameFullyDecoded(const avifDecoderData * data)
...
+                avifImageStealPlanes(dstImage, dstImageWithCorrectDepth, AVIF_PLANES_ALL);
+            }
+        }
+        avifImageDestroy(dstImageWithCorrectDepth);
+        return result;
+    }
+
+    for (avifBool alpha = AVIF_FALSE; alpha <= decoder->alphaPresent ? AVIF_TRUE : AVIF_FALSE; ++alpha) {
+        const avifImage * inputImages[AVIF_SAMPLE_TRANSFORM_MAX_NUM_INPUT_IMAGE_ITEMS];
+        for (uint32_t i = 0; i < decoder->data->sampleTransformNumInputImageItems; ++i) {
```
wantehchang, line 5650: Please verify that `decoder->data->sampleTransformNumInputImageItems <= AVIF_SAMPLE_TRANSFORM_MAX_NUM_INPUT_IMAGE_ITEM` if we reach here. I didn't have time to verify that.

src/write.c
```diff
@@ -1115,11 +1174,20 @@ static avifResult avifEncoderAddImageItems(avifEncoder * encoder,
...
+        infeName = infeNameGainMap;
+#endif
+#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
+    } else if (itemCategory >= AVIF_SAMPLE_TRANSFORM_MIN_CATEGORY && itemCategory <= AVIF_SAMPLE_TRANSFORM_MAX_CATEGORY) {
+        infeName = infeNameSampleTransform;
 #endif
-                                                                   : infeNameColor;
+    } else {
+        infeName = infeNameColor;
+    }
```
wantehchang, line 1190: Nit/optional: If this is implemented in a function, then we can use return statements and avoid the `else`. For example,

```
const char * getInfeName(avifItemCategory itemCategory)
{
    if (avifIsAlpha(itemCategory)) {
        return infeNameAlpha;
    }
    if (itemCategory == AVIF_ITEM_GAIN_MAP) {
        return infeNameGainMap;
    }
#endif
#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
    ...
#endif
    return infeNameColor;
}
```

wantehchang (approved):
LGTM. I verified that the pull request is essentially a no-op if the experimental feature is disabled.

As real code review, I have reviewed up to avif.c. I also did a little checking in read.c.

include/avif/avif.h
y-guyon: It would be safer to do so before finalizing the specification proposal. Unfortunately there are other priorities.

src/read.c
y-guyon: Is there some public resource I can cite here?

src/read.c
y-guyon: > Yes, `=(int32_t)constant;` is enough.

https://en.cppreference.com/w/c/language/conversion#Integer_conversions seems to say it is implementation-defined. Should we consider it as widespread as two's complement representation?

> Alternatively, we can do this:
> 
> ```
>     AVIF_CHECK(avifROStreamReadU32(s, (uint32_t *)&token->constant)); // signed int(1<<(bit_depth+3)) constant;
> ```

I do not think it is equivalent depending on the platform's endianness. Should we consider it as widespread as two's complement representation?

src/read.c
wantehchang: No. We have verified this by private email with Cyril Concolato and Emmanuel Thomas. There is a clarification in the current draft 14496-34: “Signed integer values shall be represented using the two’s-complement representation.”

src/read.c
wantehchang: The reason casting a `uint32_t` value greater than `INT32_MAX` to `int32_t` is implementation-defined is that there are three representations of signed integers in C. In practice, the only widely-used repesentation is two's complement.

include/avif/internal.h
```diff
@@ -156,6 +156,79 @@ void avifImageCopyNoAlloc(avifImage * dstImage, const avifImage * srcImage);
...
+    AVIF_SAMPLE_TRANSFORM_POW = 13,       // S = L==0 ? 0 : pow(L, R)
+    AVIF_SAMPLE_TRANSFORM_MIN = 14,       // S = L<=R ? L : R
+    AVIF_SAMPLE_TRANSFORM_MAX = 15,       // S = L<=R ? R : L
+    AVIF_SAMPLE_TRANSFORM_RESERVED
+} avifSampleTransformTokenType;
+
+typedef struct avifSampleTransformToken
+{
+    uint8_t type;                // avifSampleTransformTokenType
+    int32_t constant;            // If value is AVIF_SAMPLE_TRANSFORM_CONSTANT.
```
wantehchang, line 199: Nit: value => type

src/read.c
```diff
@@ -2006,6 +2023,108 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
 }
 #endif // AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
+// bit-depth is assumed to be 2 (32-bit).
```
wantehchang, line 2027: Nit: bit-depth => bit_depth ?

src/read.c
```diff
@@ -2048,6 +2167,11 @@ static avifResult avifDecoderItemReadAndParse(const avifDecoder * decoder,
         *codecType = avifGetCodecType(item->type);
         AVIF_ASSERT_OR_RETURN(*codecType != AVIF_CODEC_TYPE_UNKNOWN);
     }
+    // TODO(yguyon): If AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM is defined, backward-incompatible
+    //               files with a primary 'sato' Sample Transform derived image item could be
+    //               handled here (compared to backward-compatible files with a 'sato' item in the
+    //               same 'altr' group as the primary regular color item which are handled in
+    //               avifDecoderDataFindSampleTransformImageItem() below).
```
wantehchang, line 2174: Nit: Lines 2171-2174 do not need to be aligned with the colon (:) in line 2170.

wantehchang (approved):

src/read.c
y-guyon: I prefer it that way. Also some editors such as CLion only highlight the aligned part because they have no way of knowing where the TODO ends otherwise.

https://google.github.io/styleguide/cppguide.html#TODO_Comments does not cover that topic.

wantehchang (approved):

src/read.c
```diff
@@ -2013,6 +2030,108 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
 }
 #endif // AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
+// bit_depth is assumed to be 2 (32-bit).
+static avifResult avifParseSampleTransformTokens(avifROStream * s, avifSampleTransformExpression * expression)
+{
+    uint8_t tokenCount;
+    AVIF_CHECK(avifROStreamRead(s, &tokenCount, /*size=*/1)); // unsigned int(8) token_count;
+    AVIF_CHECKERR(avifArrayCreate(expression, sizeof(expression->tokens[0]), tokenCount), AVIF_RESULT_OUT_OF_MEMORY);
```
wantehchang, line 2039: Reject `tokenCount == 0` now? We should not pass 0 as the `initialCapacity` argument to `avifArrayCreate()`.

Alternatively, pass `AVIF_MAX(1, tokenCount)` as the `initialCapacity` argument to `avifArrayCreate()` and let the `avifSampleTransformExpressionIsValid()` call reject it.

src/read.c
```diff
@@ -2013,6 +2030,108 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
...
+
+    for (uint32_t t = 0; t < tokenCount; ++t) {
+        avifSampleTransformToken * token = (avifSampleTransformToken *)avifArrayPush(expression);
+        AVIF_ASSERT_OR_RETURN(token != NULL);
```
wantehchang, line 2043: `AVIF_ASSERT_OR_RETURN` is not appropriate here. This should return `AVIF_RESULT_OUT_OF_MEMORY`.

src/read.c
```diff
@@ -2013,6 +2030,108 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
...
+{
+    const avifProperty * pixiProp = avifPropertyArrayFind(&item->properties, "pixi");
+    if (!pixiProp) {
+        avifDiagnosticsPrintf(diag, "Item ID %u of type '%.4s' is missing mandatory pixi property", item->id, (const char *)item->type);
+        return AVIF_RESULT_BMFF_PARSE_FAILED;
+    }
+    for (uint8_t i = 0; i < pixiProp->u.pixi.planeCount; ++i) {
+        if (pixiProp->u.pixi.planeDepths[i] != pixiProp->u.pixi.planeDepths[0]) {
+            avifDiagnosticsPrintf(diag,
+                                  "Item ID %u of type '%.4s' depth specified by pixi property [%u] is not supported",
```
wantehchang, line 2096: I think the error message should instead say that two depths in the pixi property are different, and also print both `pixiProp->u.pixi.planeDepths[0]` and `pixiProp->u.pixi.planeDepths[i]`.

src/read.c
```diff
@@ -4992,6 +5148,94 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         }
 #endif // AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
+        // AVIF_ITEM_SAMPLE_TRANSFORM (not used through mainItems because not a coded item (well grids neither but it's different)).
```
wantehchang, line 5152: Nit: "well grids neither" is not clear. Does it mean "grids are not used through mainItems either" or "grids are not a coded item either"?

src/read.c
```diff
@@ -5372,6 +5633,63 @@ static avifBool avifDecoderDataFrameFullyDecoded(const avifDecoderData * data)
...
+        for (uint32_t i = 0; i < decoder->data->sampleTransformNumInputImageItems; ++i) {
+            avifItemCategory category = decoder->data->sampleTransformInputImageItems[i];
+            if (category == AVIF_ITEM_COLOR) {
+                inputImages[i] = decoder->image;
+            } else {
+                AVIF_ASSERT_OR_RETURN(category >= AVIF_ITEM_SAMPLE_TRANSFORM_INPUT_0_COLOR &&
+                                      category < AVIF_ITEM_SAMPLE_TRANSFORM_INPUT_0_COLOR +
+                                                     AVIF_SAMPLE_TRANSFORM_MAX_NUM_EXTRA_INPUT_IMAGE_ITEMS);
+                if (alpha) {
+                    category += AVIF_ITEM_SAMPLE_TRANSFORM_INPUT_0_ALPHA - AVIF_ITEM_SAMPLE_TRANSFORM_INPUT_0_COLOR;
```
wantehchang, line 5674: Replace right-hand side with `AVIF_SAMPLE_TRANSFORM_MAX_NUM_EXTRA_INPUT_IMAGE_ITEMS`.

src/sampletransform.c
```diff
@@ -0,0 +1,377 @@
...
+                    } else {
+                        AVIF_ASSERT_OR_RETURN(stackSize >= 2);
+                        stack[stackSize - 2] =
+                            avifSampleTransformOperation32bTwoOperands(stack[stackSize - 2], stack[stackSize - 1], token->type);
+                        stackSize--; // Pop two and push one.
+                    }
+                }
+                AVIF_ASSERT_OR_RETURN(stackSize == 1);
+                // Fit to 'pixi'-defined range. TODO(yguyon): Take avifRange into account.
+                stack[0] = stack[0] <= minValue ? minValue : stack[0] >= maxValue ? maxValue : stack[0];
```
wantehchang, line 308: Use `AVIF_CLAMP`?

wantehchang (approved):
Yannis: I checked parts of read.c and sampletransform.c. This is probably the best I can do by the end of this week. You can go ahead and merge this pull request.

src/read.c
y-guyon: I disagree, `avifArrayCreate()` was just called with sufficient `initialCapacity`. But your suggestion is easier to maintain I guess.

wantehchang: You are right. I missed that point. Sorry. This should be an assertion.

Note: This array can be a fixed-sized dynamically-allocated array. It's not necessary to use avifArray.

PR #2056, opened 2024-03-15
Add note about buffer copy in avifDecoderRead()

y-guyon:
I think tiles are allocated by libavif and the buffers lent to the codecs.  
If not used afterwards, they could be moved instead of copied.

More of a question than a change request. Could be a `TODO: investigate`.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -5589,6 +5589,8 @@ avifResult avifDecoderRead(avifDecoder * decoder, avifImage * image)
     if (result != AVIF_RESULT_OK) {
         return result;
     }
+    // If decoder->image owns its planes, their ownership could be transferred here instead of copied,
+    // unless the user reuses the decoder instance, which is unknown yet.
     return avifImageCopy(image, decoder->image, AVIF_PLANES_ALL);
 }
 
--- a/src/read.c
+++ b/src/read.c
@@ -5589,8 +5589,12 @@ avifResult avifDecoderRead(avifDecoder * decoder, avifImage * image)
     if (result != AVIF_RESULT_OK) {
         return result;
     }
-    // If decoder->image owns its planes, their ownership could be transferred here instead of copied,
-    // unless the user reuses the decoder instance, which is unknown yet.
+    // If decoder->image->imageOwnsYUVPlanes and the decoder instance is not used after this call,
+    // the ownership of the samples in decoder->image could be transferred here instead of copied.
+    // However most codec_*.c implementations allocate the output buffer themselves and return a
+    // view, unless some postprocessing is applied (container-level grid reconstruction for
+    // example), so the first condition rarely holds.
+    // The second condition does not hold either: it is not required by the documentation in avif.h.
     return avifImageCopy(image, decoder->image, AVIF_PLANES_ALL);
 }
 
--- a/src/read.c
+++ b/src/read.c
@@ -5589,8 +5589,8 @@ avifResult avifDecoderRead(avifDecoder * decoder, avifImage * image)
     if (result != AVIF_RESULT_OK) {
         return result;
     }
-    // If decoder->image->imageOwnsYUVPlanes and the decoder instance is not used after this call,
-    // the ownership of the samples in decoder->image could be transferred here instead of copied.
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -5589,6 +5589,8 @@ avifResult avifDecoderRead(avifDecoder * decoder, avifImage * image)
     if (result != AVIF_RESULT_OK) {
         return result;
     }
+    // If decoder->image owns its planes, their ownership could be transferred here instead of copied,
+    // unless the user reuses the decoder instance, which is unknown yet.
```
wantehchang, line 5593: I looked into this. I found that this function is documented in avif.h as follows:

```
// Simple interfaces to decode a single image, independent of the decoder afterwards (decoder may be destroyed).
AVIF_API avifResult avifDecoderRead(avifDecoder * decoder, avifImage * image); // call avifDecoderSetIO*() first
AVIF_API avifResult avifDecoderReadMemory(avifDecoder * decoder, avifImage * image, const uint8_t * data, size_t size);
AVIF_API avifResult avifDecoderReadFile(avifDecoder * decoder, avifImage * image, const char * filename);
```

That comment in avif.h implies `image` must stay valid if `decoder` is reused or destroyed. So we must copy `decoder->image` to `image` here.

Note: I am not sure how to reuse a decoder instance. Could you explain what you mean by that?

So we can add a comment here to explain why we must copy the planes and cannot transfer the ownership of the planes from `decoder->image` to `image`.

y-guyon: > I looked into this.

Thank you for that.

> That comment in avif.h implies `image` must stay valid if `decoder` is reused or destroyed. So we must copy `decoder->image` to `image` here.

I understand it as: the `image` samples can still be accessed by the user if the `decoder` is destroyed first. So a copy is not mandatory, it could be a move, right?

> Note: I am not sure how to reuse a decoder instance. Could you explain what you mean by that?

Not sure myself either. What should happen if a user calls `avifDecoderRead()` twice in a row, for a regular still non-layered image? What if it is an animation?

I was telling myself that maybe for an image where the user could do something with the `decoder` instance afterwards (such as calling `avifDecoderRead()` again for a valid use case such as an animation), copying the `image` makes sense, but for a simple still image this copy could be avoided.

wantehchang: If the caller of `avifDecoderRead()` doesn't use `decoder->image` after the call and `decoder->image` owns the planes, you are right that we can move the planes from `decoder->image` to `image`.

The first condition is not required by the documentation. Although callers of `avifDecoderRead()` are unlikely to use `decoder->image` after the call, it would be good to assume that they may do that.

The second condition does not hold for the libaom and dav1d decoders (check that by searching for "imageOwnsYUVPlanes" in src/codec_aom.c and src/codec_dav1d.c).

In summary, I suggest we not touch the `avifDecoderRead()` function. Callers that want to avoid the image copying should not use this function. They should call `avifDecoderParse()` and `avifDecoderNextImage()` and then use `decoder->image`.

wantehchang: Also, `avifDecoderRead()` is apparently designed to decode a still image. Since `avifDecoderRead()` calls `avifDecoderParse()` internally, calling `avifDecoderRead()` again on the same decoder instance will return the first image again, if I understand it correctly.

src/read.c
```diff
@@ -5589,6 +5589,12 @@ avifResult avifDecoderRead(avifDecoder * decoder, avifImage * image)
     if (result != AVIF_RESULT_OK) {
         return result;
     }
+    // If decoder->image->imageOwnsYUVPlanes and the decoder instance is not used after this call,
```
wantehchang, line 5592: Nit: Add "is true" after "decoder->image->imageOwnsYUVPlanes".

"the decoder instance is not used" can be more specific: "decoder->image is not used".

src/read.c
```diff
@@ -5589,6 +5589,12 @@ avifResult avifDecoderRead(avifDecoder * decoder, avifImage * image)
...
+    // the ownership of the samples in decoder->image could be transferred here instead of copied.
```
wantehchang, line 5593: Optional: "planes" seems better than "samples".

wantehchang (approved):
LGTM. Thanks for adding the comment.

PR #2058, opened 2024-03-15
Check that >0 pixi planes have the same depth

y-guyon:


vrabaud (approved):

src/read.c
```diff
@@ -2222,12 +2222,19 @@ static avifBool avifParsePixelInformationProperty(avifProperty * prop, const uin
 
     avifPixelInformationProperty * pixi = &prop->u.pixi;
     AVIF_CHECK(avifROStreamRead(&s, &pixi->planeCount, 1)); // unsigned int (8) num_channels;
-    if (pixi->planeCount > MAX_PIXI_PLANE_DEPTHS) {
+    if (pixi->planeCount < 1 || pixi->planeCount > MAX_PIXI_PLANE_DEPTHS) {
```
wantehchang, line 2225: Optional: Since `pixi->planeCount` is an unsigned type, I probably would test `pixi->planeCount == 0` (as in the original code, at line 4612).

wantehchang (commented):
Good change. Thanks!

src/read.c
y-guyon: That is fine with me.

PR #2061, opened 2024-03-15
ci-unix-static-sanitized.yml for macos

y-guyon:


wantehchang (approved):

.github/workflows/ci-unix-static-sanitized.yml
```diff
@@ -6,10 +6,11 @@ permissions:
 
 jobs:
   build-static-sanitized:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
```
wantehchang, line 9: Ideally, we should test macOS on Arm. Not sure if there is a virtual machine for that.

y-guyon: It is not listed at https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories.

wantehchang: macos-14 [Beta] has the M1 CPU. I am testing it in https://github.com/AOMediaCodec/libavif/pull/2064.

PR #2060, opened 2024-03-15
Use env vars in ci-unix-static-sanitized.yml

y-guyon:

PR #2062, opened 2024-03-15
Fix env indent in ci-unix-static-sanitized

y-guyon:


y-guyon:
Note that it was broken in #2060 but the PR UI did not list the CI workflow at all.

vrabaud (approved):

PR #2064, opened 2024-03-18
Add macos-14

wantehchang:
macos-14 [Beta] has the M1 CPU.

### DIFF
--- a/.github/workflows/ci-unix-static-sanitized.yml
+++ b/.github/workflows/ci-unix-static-sanitized.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, macos-14]
         sanitizer: [ undefined ] # TODO(yguyon): Add address, memory, thread
 
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-unix-static-sanitized.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-14]
```
vrabaud, line 13: How about using [ubuntu-latest, macos-latest, macos-14] to make sure we do not miss when macos-latest switched to macos-15 (in a veeeery long time)

wantehchang: Vincent: Thanks for the review. This pull request is not ready. I have changed it to a draft (work in progress).

I replaced macos-latest with macos-14 because I thought that from libavif's perspective, macos-latest and ubuntu-latest (both use Intel CPUs) are similar enough so that we don't need to test on both. macos-14 is documented to use the M1 CPU, so it will allow us to run our Arm code under sanitizers.

But it seems that macos-14 still uses an Intel CPU but have cross-compilers that generate Arm code. So that makes it challenging to get it to work. So I am not sure if I will be able to get this pull request to work.

wantehchang: I have accepted your suggestion.

wantehchang:
I am abandoning this pull request.

PR #2070, opened 2024-03-20
Pass the -DSANITIZE=<type> cmake option to libaom

wantehchang:
This is the official way to build libaom with sanitizer support: https://aomedia.googlesource.com/aom/+/master/README.md#sanitizers

### DIFF
--- a/.github/workflows/ci-unix-static-sanitized.yml
+++ b/.github/workflows/ci-unix-static-sanitized.yml
@@ -41,11 +41,10 @@ jobs:
       - name: Build aom
         if: steps.cache-ext.outputs.cache-hit != 'true'
         working-directory: ./ext
-        run: ./aom.cmd
-        env:
-          CFLAGS: -fsanitize=${{ matrix.sanitizer }}
-          CXXFLAGS: -fsanitize=${{ matrix.sanitizer }}
-          LDFLAGS: -fsanitize=${{ matrix.sanitizer }}
+        run: >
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-unix-static-sanitized.yml
```diff
@@ -41,11 +41,10 @@ jobs:
       - name: Build aom
         if: steps.cache-ext.outputs.cache-hit != 'true'
         working-directory: ./ext
-        run: ./aom.cmd
-        env:
-          CFLAGS: -fsanitize=${{ matrix.sanitizer }}
-          CXXFLAGS: -fsanitize=${{ matrix.sanitizer }}
-          LDFLAGS: -fsanitize=${{ matrix.sanitizer }}
+        run: >
+          sed -i -e 's/cmake -G Ninja \(.*\) \.\./cmake -G Ninja \1 -DSANITIZE=${{ matrix.sanitizer }} ../g' aom.cmd
```
wantehchang, line 45: This line is based on line 52 below:

```
          sed -i -e 's/meson setup \(.*\) \.\./meson setup \1 '"-Db_sanitize=${{ matrix.sanitizer }} -Db_lundef=false"' ../g' dav1d.cmd
```

Note that I removed the double quotes and single quotes in the middle of the sed script. They do not seem necessary for this line. Are they necessary for line 45?

y-guyon: > Are they necessary for line 45?

No idea, probably not. I took that from:

https://github.com/AOMediaCodec/libavif/blob/f8ad918b155ae324f7b43eb74cf7fe4127ccfce7/tests/oss-fuzz/build.sh#L52

y-guyon (approved):

PR #2072, opened 2024-03-20
Use double quotes "" to allow substitution

wantehchang:
Also remove an extraneous space in tests/oss-fuzz/build.sh.

### DIFF
--- a/.github/workflows/ci-unix-static-sanitized.yml
+++ b/.github/workflows/ci-unix-static-sanitized.yml
@@ -49,7 +49,7 @@ jobs:
         if: steps.cache-ext.outputs.cache-hit != 'true'
         working-directory: ./ext
         run: >
-          sed -i -e 's/meson setup \(.*\) \.\./meson setup \1 '"-Db_sanitize=${{ matrix.sanitizer }} -Db_lundef=false"' ../g' dav1d.cmd
+          sed -i -e 's/meson setup \(.*\) \.\./meson setup \1 -Db_sanitize=${{ matrix.sanitizer }} -Db_lundef=false ../g' dav1d.cmd
 
           ./dav1d.cmd
       - name: Build libyuv
--- a/tests/oss-fuzz/build.sh
+++ b/tests/oss-fuzz/build.sh
@@ -49,7 +49,7 @@ if [ "$SANITIZER" == "memory" ]
 then
   DAV1D_EXTRA_FLAGS="${DAV1D_EXTRA_FLAGS} -Denable_asm=false"
 fi
-sed -i 's/meson setup \(.*\) \.\./meson setup \1 '"${DAV1D_EXTRA_FLAGS}"' ../g' ./ext/dav1d.cmd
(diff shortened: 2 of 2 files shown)


.github/workflows/ci-unix-static-sanitized.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         if: steps.cache-ext.outputs.cache-hit != 'true'
         working-directory: ./ext
         run: >
-          sed -i -e 's/meson setup \(.*\) \.\./meson setup \1 '"-Db_sanitize=${{ matrix.sanitizer }} -Db_lundef=false"' ../g' dav1d.cmd
+          sed -i -e 's/meson setup \(.*\) \.\./meson setup \1 -Db_sanitize=${{ matrix.sanitizer }} -Db_lundef=false ../g' dav1d.cmd
```
wantehchang, line 52: James pointed out that the `-i` option is not portable. I did not address that issue in this pull request.

The `-i` option is not in the Single Unix specification:
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html

And it is different between Linux and macOS.

Linux:

```
  -i[SUFFIX], --in-place[=SUFFIX]
                 edit files in place (makes backup if SUFFIX supplied)
```

macOS:

```
     -i extension
             Edit files in-place similarly to -I, but treat each file
             independently from other files.  In particular, line numbers in
             each file start at 1, the “$” address matches the last line of
             the current file, and address ranges are limited to the current
             file.  (See Sed Addresses.) The net result is as though each file
             were edited by a separate sed instance.
```

y-guyon: Yes, I had to add `-e` to make it work on macos.

y-guyon (approved):

.github/workflows/ci-unix-static-sanitized.yml
wantehchang: Thanks for the confirmation. James belives (and I agree) that on macOS, `sed -i -e script` is parsed as `sed`, `-i -e`, `script`. So `-e` is the argument to the `-i` option, i.e., `-e` is the extension of the backup file before `sed` edits the file in place.

tests/oss-fuzz/build.sh
```diff
@@ -49,7 +49,7 @@ if [ "$SANITIZER" == "memory" ]
 then
   DAV1D_EXTRA_FLAGS="${DAV1D_EXTRA_FLAGS} -Denable_asm=false"
 fi
-sed -i 's/meson setup \(.*\) \.\./meson setup \1 '"${DAV1D_EXTRA_FLAGS}"' ../g' ./ext/dav1d.cmd
+sed -i "s/meson setup \(.*\) \.\./meson setup \1${DAV1D_EXTRA_FLAGS} ../g" ./ext/dav1d.cmd
```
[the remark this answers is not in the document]
vrabaud: That's fine

PR #2074, opened 2024-03-21
Don't request version 2.15.05 with setup-nasm

wantehchang:
The current default version of nasm is 2.16.01:
https://github.com/ilammy/setup-nasm


jzern (approved):
The commit message made me think there was a bug with this version at first, but I think you mean don't force the version because there's a newer one.

wantehchang:
I will edit the commit message as you suggested.

PR #2080, opened 2024-03-29
Fix avifParseItemLocationBox() itemReferenceIndex

y-guyon:
Revert bug introduced with StreamReadBits(): https://github.com/AOMediaCodec/libavif/pull/1506/files#diff-d4b71621e3fb7422ad72c18339089f9626a0c48a95d4b71483bb18351bcd58c4
Read item_reference_index even if construction_method!=2 to be accurate with the ISOBMFF specification.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1744,6 +1744,7 @@ static avifResult avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw,
 {
     BEGIN_STREAM(s, raw, rawLen, diag, "Box[iloc]");
 
+    // Section 8.11.3.2 of ISO/IEC 14496-12.
     uint8_t version;
     AVIF_CHECKERR(avifROStreamReadVersionAndFlags(&s, &version, NULL), AVIF_RESULT_BMFF_PARSE_FAILED);
     if (version > 2) {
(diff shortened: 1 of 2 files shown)


y-guyon:
There is the same issue with v1.0.x:

https://github.com/AOMediaCodec/libavif/blob/8892ff40459e94d5ce06b4337e09ac7c07b18c29/src/read.c#L1661-L1665

I do not think it is necessary to cherry pick the fix. It should just wrongly reject/accept some rare files.

vigneshvg (approved):

src/read.c
```diff
@@ -1830,10 +1838,8 @@ static avifResult avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw,
         uint16_t extentCount;                                                                // unsigned int(16) extent_count;
         AVIF_CHECKERR(avifROStreamReadU16(&s, &extentCount), AVIF_RESULT_BMFF_PARSE_FAILED); //
         for (int extentIter = 0; extentIter < extentCount; ++extentIter) {
-            // If extent_index is ever supported, this spec must be implemented here:
-            // ::  if (((version == 1) || (version == 2)) && (index_size > 0)) {
-            // ::      unsigned int(index_size*8) extent_index;
-            // ::  }
+            uint64_t itemReferenceIndex; // unsigned int(index_size*8) item_reference_index; (ignored unless construction_method=2)
+            AVIF_CHECKERR(avifROStreamReadUX8(&s, &itemReferenceIndex, indexSize), AVIF_RESULT_BMFF_PARSE_FAILED);
```
wantehchang, line 1842: Nit/optional: It would be clearer to put this call inside `if (indexSize > 0)`. The corresponding spec is:

```
    if (((version == 1) || (version == 2)) && (index_size > 0)) {
         unsigned int(index_size*8) item_reference_index;
    }
```

But perhaps it is that spec that should omit the `&& (index_size > 0)` check.

src/read.c
```diff
@@ -1805,8 +1813,8 @@ static avifResult avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw,
             uint8_t constructionMethod;
             AVIF_CHECKERR(avifROStreamReadBits8(&s, &constructionMethod, /*bitCount=*/4),
                           AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(4) construction_method;
-            if ((constructionMethod != 0 /* file */) && (constructionMethod != 1 /* idat */)) {
-                // construction method item(2) unsupported
+            if (constructionMethod != 0 /* file offset */ && constructionMethod != 1 /* idat offset */) {
+                // construction method item(2) unsupported (item offset)
```
wantehchang, line 1817: Nit: The new comment "(item offset)" is redundant. I suggest just adding "offset" after the first "item":

```
    // construction method item offset(2) unsupported
```

or

```
    // construction method 2 (item offset) unsupported
```

src/read.c
wantehchang: Thanks. While reviewing https://github.com/AOMediaCodec/libavif/pull/2091, I discovered a subtle point about `item_reference_index` -- its value is 1 rather than 0 when `item_size` is 0. Here is the excerpt from ISO/IEC 14496-12:2022 Section 8.11.3.1, page 82:

> The `item_reference_index` is only used for the method `item_offset`; it indicates the 1-based index of the item reference with `referenceType 'iloc'` linked from this item. If `index_size` is 0, then the value 1 is implied; the value 0 is reserved.

Calling `avifROStreamReadUX8(&s, &itemReferenceIndex, indexSize)` with `indexSize=0` will set `itemReferenceIndex` to 0. So `item_reference_index` needs different handling anyway.

PR #2086, opened 2024-04-02
Run cifuzz for changes in ext/

vrabaud:


y-guyon (approved):

.github/workflows/cifuzz.yml
```diff
@@ -3,6 +3,7 @@ on:
   pull_request:
     paths:
       - 'include/**'
+      - 'ext/**'
```
wantehchang, line 6: Nit: List "ext" before "include"

vrabaud: Fixed in https://github.com/AOMediaCodec/libavif/pull/2092

PR #2088, opened 2024-04-03
Remove spaces at the end of lines

wantehchang:


vigneshvg (approved):

jzern:
`git diff --check` in the CI might catch this with the defaults or `core.whitespace` could be updated in `.git/config`.

PR #2089, opened 2024-04-04
Add .vscode to .gitignore

y-guyon:
[skip ci]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,6 @@
 /ext/SVT-AV1
 /ext/zlib
 .clangd/
+.vscode/
 cscope.*
 tags


vrabaud (approved):

kmilos:
As you can't know all possible environments developers use, stuff like this really belongs to local user configuration and their global `~/.gitignore`.

y-guyon:
> As you can't know all possible environments developers use, stuff like this really belongs to local user configuration and their global `~/.gitignore` or `~/.config/git/ignore`.

I agree but VSCode is widespread enough now and it does not hurt much.

.gitignore
```diff
@@ -19,5 +19,6 @@
 /ext/SVT-AV1
 /ext/zlib
 .clangd/
```
Azizjon8822, line 21: Дубликат #

PR #2093, opened 2024-04-07
Document avifROStreamReadUX8 behavior for factor=0

wantehchang:


include/avif/internal.h
```diff
@@ -592,7 +592,8 @@ AVIF_NODISCARD avifBool avifROStreamReadU16(avifROStream * stream, uint16_t * v)
 AVIF_NODISCARD avifBool avifROStreamReadU16Endianness(avifROStream * stream, uint16_t * v, avifBool littleEndian);
 AVIF_NODISCARD avifBool avifROStreamReadU32(avifROStream * stream, uint32_t * v);
 AVIF_NODISCARD avifBool avifROStreamReadU32Endianness(avifROStream * stream, uint32_t * v, avifBool littleEndian);
-AVIF_NODISCARD avifBool avifROStreamReadUX8(avifROStream * stream, uint64_t * v, uint64_t factor); // Reads a factor*8 sized uint, saves in v
+// Reads a factor*8 sized uint, saves in v. If factor is 0, reads nothing and saves 0 in v.
```
wantehchang, line 595: I wonder if we should change `factor*8 sized uint` to what is used in the spec, `unsigned int(factor*8)`.

y-guyon (approved):

PR #2097, opened 2024-04-11
Do not use AVIF_LOCAL for libxml2 and gtest.

vrabaud:
Switch some old AVIF_LOCAL_*.


vrabaud:
This only fixes the Windows CI (and actually the ci-static one too, though it was passing).

y-guyon (approved):
Wondering if one or two CI workflows should keep the old flags as long as they are supported.

vrabaud:
I think deprecated code should not be used except on purpose. I just added a CI file to make sure deprecated flags are still working though.

y-guyon (approved):
Thanks

PR #2103, opened 2024-04-12
Enable CI cache in build/_deps

vrabaud:


.github/workflows/ci-disable-gtest.yml
```diff
@@ -47,6 +47,12 @@ jobs:
         CARGO_C_FILE: cargo-c-x86_64-unknown-linux-musl.tar.gz
       run: |
         curl -L $LINK/$CARGO_C_FILE | tar xz -C ~/.cargo/bin
+    - name: Cache external dependencies
+      id: cache-ext
+      uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      with:
+        path: ext, build/_deps
+        key: ${{ runner.os }}-disable-gtest-${{ hashFiles('cmake/Modules/*', 'ext/*.cmd', 'ext/*.sh') }}
```
y-guyon, line 55: There is no point in creating a cache-entry if no step is skipped thanks to it (guarded by `if: steps.cache-ext.outputs.cache-hit != 'true'`)

vrabaud: If the hash is the same, the folder 'build/_deps' will be restored from cache. CMake will then automagically use it (it will not re-download/compile whatever is in there). At least, that is my understanding. Let's see if I am missing more files than just 'build/_deps'.

y-guyon (approved):

Issue #2105, opened 2024-04-12
Findrav1e.cmake ignores library dependencies listed in the generated pkgconfig file


1480c1:
librav1e declares a dependency to `-lntdll` in their .pc file, and it's presumably populated into `RAV1E_LIBRARIES`, but it is not included in the link target `rav1e::rav1e`, which means it's never actually used in the linking.

```bash
$ cat /local64/lib/pkgconfig/rav1e.pc
prefix=D:/gcc/local64
exec_prefix=${prefix}
libdir=${prefix}/lib
includedir=${prefix}/include

Name: rav1e
Description: The fastest and safest AV1 encoder
Version: 0.7.0
Libs: -L${libdir} -lrav1e
Cflags: -I${includedir}/rav1e
Libs.private:  -lkernel32 -ladvapi32 -lbcrypt -lkernel32 -lntdll -luserenv -lws2_32 -lkernel32 -lws2_32 -lkernel32 -lntdll -lkernel32 -Wl,--allow-multiple-definition
```

```c++
[38/39] Linking CXX executable avifdec.exe
FAILED: avifdec.exe 
[...]

PR #2106, opened 2024-04-12
Add missing library deps to rav1e::rav1e on Windows.

vrabaud:


cmake/Modules/Findrav1e.cmake
```diff
@@ -64,5 +64,10 @@ if(RAV1E_LIBRARY)
         endif()
         set_target_properties(rav1e::rav1e PROPERTIES IMPORTED_LOCATION "${RAV1E_LIBRARY}" IMPORTED_SONAME rav1e)
         target_include_directories(rav1e::rav1e INTERFACE ${RAV1E_INCLUDE_DIR})
+        if(WIN32)
+            target_link_libraries(rav1e::rav1e INTERFACE ntdll.lib userenv.lib ws2_32.lib bcrypt.lib)
+        elseif(UNIX AND NOT APPLE)
+            target_link_libraries(rav1e::rav1e INTERFACE ${CMAKE_DL_LIBS}) # for backtrace
+        endif()
```
wantehchang, line 71: For future maintainers, it would be good to cite the source of this information. This is not obvious.

I am also wondering why we didn't need this before. Does this fix a bug that affects our clients but doesn't affect our own build?

wantehchang (approved):

cmake/Modules/Findrav1e.cmake
```diff
@@ -64,5 +64,10 @@ if(RAV1E_LIBRARY)
         endif()
         set_target_properties(rav1e::rav1e PROPERTIES IMPORTED_LOCATION "${RAV1E_LIBRARY}" IMPORTED_SONAME rav1e)
         target_include_directories(rav1e::rav1e INTERFACE ${RAV1E_INCLUDE_DIR})
+        if(WIN32)
+            target_link_libraries(rav1e::rav1e INTERFACE ntdll.lib userenv.lib ws2_32.lib bcrypt.lib)
```
1480c1, line 68: You might want to consider checking the contents of `_RAV1E_LIBRARIES` as on my local machine, it produced
```cmake
_RAV1E_LIBRARIES:INTERNAL=rav1e;kernel32;advapi32;bcrypt;kernel32;ntdll;userenv;ws2_32;kernel32;ws2_32;kernel32;ntdll;kernel32
```

Also, I'm not sure if the .lib suffix is necessary, but I don't think it hurts either.

1480c1:
This patch works for fixing #2105

cmake/Modules/Findrav1e.cmake
fdintino: I doubt it will be useful, but at one point I had written a version of this cmake module that uses the link libraries from pkg_check_modules if they're available: https://github.com/fdintino/libavif/blob/fe6be1adddaf504b03ce18c5aaa4b9ad83b07a70/cmake/Modules/Findrav1e.cmake. If anyone should ever open an issue caused by their local rust native libraries being different from this list, this is one approach that could solve it.

Also, because pkg-config support is not guaranteed with windows, I had written a function to parse the requirements from the `.pc` file. I didn't end up needing it for local builds because you can get the list of libraries with corrosion from `Rust_CARGO_TARGET_LINK_NATIVE_LIBS`, but here is what I had:

```cmake
function(avif_parse_rav1e_pc_libs RAV1E_PC_PATH OUTPUT_VARIABLE)
    if(BUILD_SHARED_LIBS)
        set(RAV1E_PC_LIB_REGEX "Libs: -L\\$\\{libdir\\} -lrav1e ([^\n]+)\n")
    else()
        set(RAV1E_PC_LIB_REGEX "Libs\\.private: +(.+)\n")
    endif()
    file(READ ${RAV1E_PC_PATH} RAV1E_PC_CONTENTS)
    string(REGEX MATCH "${RAV1E_PC_LIB_REGEX}" _ ${RAV1E_PC_CONTENTS})
    if(CMAKE_MATCH_1)
        string(STRIP ${CMAKE_MATCH_1} RAV1E_LIBS_PRIVATE)
        string(REPLACE " " ";" RAV1E_LIBS_PRIVATE ${RAV1E_LIBS_PRIVATE})
        set(${OUTPUT_VARIABLE} ${${OUTPUT_VARIABLE}} ${RAV1E_LIBS_PRIVATE} PARENT_SCOPE)
    endif()
endfunction()

cmake/Modules/Findrav1e.cmake
vrabaud: These instructions were simply taken from the main CMake. We need to unify that.
We did not see it because this happens when rav1e is installed. Our CI only has local rav1e. I'll try adding one with installed packages on Window.s

cmake/Modules/Findrav1e.cmake
vrabaud: The .lib are here because I copy/pasted from the main CMake. I'll try cleaning that up in a following PR. Let's first merge this as it works.

vrabaud:
I added a CI for installed packages on Windows in https://github.com/AOMediaCodec/libavif/pull/2110

PR #2115, opened 2024-04-15
Limit the CI jobs running on pull_request.

vrabaud:
Only trigger the ones recompiling dependencies locally when ./ext or CMake is changed.
This effectively prunes all the jobs taking more than 5 minutes.

### DIFF
--- a/.github/workflows/ci-disable-gtest.yml
+++ b/.github/workflows/ci-disable-gtest.yml
@@ -5,7 +5,13 @@
 #   * Disables gtest.
 
 name: CI Disable GTest
-on: [push, pull_request]
+on:
+  push
+  pull_request:
+    paths:
+      - '**CMakeLists.txt**'
+      - 'cmake/**'
+      - 'ext/**'
 
 permissions:
   contents: read
--- a/.github/workflows/ci-disable-gtest.yml
+++ b/.github/workflows/ci-disable-gtest.yml
@@ -6,7 +6,7 @@
 
 name: CI Disable GTest
 on:
-  push
+  push:
   pull_request:
     paths:
       - '**CMakeLists.txt'
--- a/.github/workflows/ci-disable-gtest.yml
+++ b/.github/workflows/ci-disable-gtest.yml
@@ -9,7 +9,7 @@ on:
   push
   pull_request:
     paths:
-      - '**CMakeLists.txt**'
+      - '**CMakeLists.txt'
       - 'cmake/**'
       - 'ext/**'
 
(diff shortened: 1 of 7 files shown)


.github/workflows/ci-disable-gtest.yml
```diff
@@ -5,7 +5,13 @@
 #   * Disables gtest.
 
 name: CI Disable GTest
-on: [push, pull_request]
+on:
+  push
+  pull_request:
+    paths:
+      - '**CMakeLists.txt**'
```
jzern, line 12: The trailing '**' might not be necessary.

.github/workflows/ci-unix-static-av2.yml
```diff
@@ -1,5 +1,10 @@
 name: CI Unix Static AV2
-on: [push, pull_request]
+  push
```
jzern, line 2: where is this documented? Shouldn't these be under 'on:'?

.github/workflows/ci-disable-gtest.yml
vrabaud: Done

.github/workflows/ci-unix-static-av2.yml
vrabaud: They should, just a bad copy/paste that happens to work for some reason.

.github/workflows/cifuzz.yml
```diff
@@ -4,8 +4,6 @@ on:
     paths:
       - 'cmake/**'
       - 'ext/**'
-      - 'include/**'
-      - 'src/**'
```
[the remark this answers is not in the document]
vrabaud: It is slow, that's why I removed it. Then again, I had forgotten the push. I put it back.

.github/workflows/ci-disable-gtest.yml
```diff
@@ -5,7 +5,13 @@
 #   * Disables gtest.
 
 name: CI Disable GTest
-on: [push, pull_request]
+on:
+  push
```
jzern, line 9: I didn't see an example or reference for merging two sections this way (no trailing ':') here.

jzern (approved):

.github/workflows/ci-disable-gtest.yml
vrabaud: I saw the no trailing `:` here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push . I added `:` just to be more YAML like .

PR #2120, opened 2024-04-17
Add avifRange ranges in avif.h

y-guyon:
Taken from libaom.

[skip ci]

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -305,7 +305,7 @@ typedef enum avifChromaSamplePosition
 
 typedef enum avifRange
 {
-    // AVIF_RANGE_LIMITED is only applicable to YUV planes. RGB and alpha planes are always full range.
+    // avifRange is only applicable to YUV planes. RGB and alpha planes are always full range.
     AVIF_RANGE_LIMITED = 0, /**<- Y  [16..235],  UV  [16..240]  (bit depth 8) */
                             /**<- Y  [64..940],  UV  [64..960]  (bit depth 10) */
                             /**<- Y [256..3760], UV [256..3840] (bit depth 12) */
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -305,12 +305,13 @@ typedef enum avifChromaSamplePosition
 
 typedef enum avifRange
 {
+    // AVIF_RANGE_LIMITED is only applicable to YUV planes. RGB and alpha planes are always full range.
     AVIF_RANGE_LIMITED = 0, /**<- Y  [16..235],  UV  [16..240]  (bit depth 8) */
                             /**<- Y  [64..940],  UV  [64..960]  (bit depth 10) */
                             /**<- Y [256..3760], UV [256..3840] (bit depth 12) */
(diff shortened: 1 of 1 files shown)


include/avif/avif.h
```diff
@@ -305,8 +305,12 @@ typedef enum avifChromaSamplePosition
...
 typedef enum avifRange
 {
-    AVIF_RANGE_LIMITED = 0,
-    AVIF_RANGE_FULL = 1
+    AVIF_RANGE_LIMITED = 0, /**<- Y  [16..235],  UV  [16..240]  (bit depth 8) */
+                            /**<- Y  [64..940],  UV  [64..960]  (bit depth 10) */
+                            /**<- Y [256..3760], UV [256..3840] (bit depth 12) */
+    AVIF_RANGE_FULL = 1     /**<- YUV/RGB [0..255]  (bit depth 8) */
+                            /**<- YUV/RGB [0..1023] (bit depth 10) */
+                            /**<- YUV/RGB [0..4095] (bit depth 12) */
```
wantehchang, line 313: I think it is better to remove "/RGB" from this comment. I think "/RGB" means that R, G, B are always in full range, but this is not obvious from reading the comment.

Note: There is a special case for limited range: Identity (MatrixCoefficients=0) and YCgCo (MatrixCoefficients=8). In this special case, the limited range is the same for all three channels (G,B,R, or Y,Cg,Co).

wantehchang: Perhaps we can write the comments as follows:

```
    AVIF_RANGE_LIMITED = 0, /**<- Y  [16..235],  UV  [16..240]  (bit depth 8) */
                            /**<- Y  [64..940],  UV  [64..960]  (bit depth 10) */
                            /**<- Y [256..3760], UV [256..3840] (bit depth 12) */
    AVIF_RANGE_FULL = 1     /**<- [0..255]  (bit depth 8) */
                            /**<- [0..1023] (bit depth 10) */
                            /**<- [0..4095] (bit depth 12) */
```

And add another comment that says "avifRange is not applicable to RGB and alpha planes." or "avifRange is only applicable to YUV planes. RGB and alpha planes are always full range."

include/avif/avif.h
```diff
@@ -305,8 +305,13 @@ typedef enum avifChromaSamplePosition
 
 typedef enum avifRange
 {
-    AVIF_RANGE_LIMITED = 0,
-    AVIF_RANGE_FULL = 1
+    // AVIF_RANGE_LIMITED is only applicable to YUV planes. RGB and alpha planes are always full range.
```
wantehchang, line 308: Nit: The first word of this comment should be "avifRange" (i.e., the enum type), not "AVIF_RANGE_LIMITED". It means the issue of limited vs. full range is only applicable to YUV planes.

wantehchang (approved):

include/avif/avif.h
wantehchang: This comment is intended to explain why the `avifRange` type is only used for the `yuvRange` member of the `avifImage` struct and is not used with the `avifRGBImage` struct or with the alpha plane in the `avifImage` struct. That's why it is clearer to use "avifRange" in this comment.

PR #2124, opened 2024-04-18
Check that the header size fits in the stream.

vrabaud:


vigneshvg (approved):

src/read.c
```diff
@@ -3231,6 +3231,7 @@ avifBool avifPeekCompatibleFileType(const avifROData * input)
     if (memcmp(header.type, "ftyp", 4)) {
         return AVIF_FALSE;
     }
+    AVIF_CHECK(avifROStreamHasBytesLeft(&s, header.size));
```
vigneshvg, line 3234: this branch calls `avifROStreamReadBoxHeader` (unlike main, which calls avifROStreamReadBoxHeaderPartial`).

`avifROStreamReadBoxHeader`, already checks that there must be at least `header.size` bytes left. so we don't need this patch in this branch.

vigneshvg (changes requested):

src/read.c
vrabaud: Indeed, no need for that PR>

PR #2131, opened 2024-04-24
ext: update to  libaom v3.9.0

jeromejj:
Updates the libaom version checked out in the ext and docker scripts
from v3.8.2 to v3.9.0.


vrabaud:
Thx, please also bump here too: https://github.com/AOMediaCodec/libavif/blob/b2ecd8e45af92cbd121e8efd13ae63ec9c96c749/tests/docker/build.sh#L29 and https://github.com/AOMediaCodec/libavif/blob/b2ecd8e45af92cbd121e8efd13ae63ec9c96c749/cmake/Modules/LocalAom.cmake#L1

jeromejj:
> Thx, please also bump here too:
> 
> https://github.com/AOMediaCodec/libavif/blob/b2ecd8e45af92cbd121e8efd13ae63ec9c96c749/tests/docker/build.sh#L29
> 
> and
> https://github.com/AOMediaCodec/libavif/blob/b2ecd8e45af92cbd121e8efd13ae63ec9c96c749/cmake/Modules/LocalAom.cmake#L1

These were bumped in the last commit update.

jzern:
> FWIW, seeing a build failure on MSYS2 CLANGARM64: ...

This change hasn't landed. Is the issue with libaom or libavif? I hit an authentication error when following the link.

jzern:
> It's an issue w/ bumping libaom to 3.9.0 for MSYS2.You can alternatively try the links via https://packages.msys2.org/queue (while they're still there)?
> 

Lots of _FORTIFY_SOURCE spam unfortunately. The failure I see is:

```
2024-04-24T18:14:40.7741548Z error: Incorrect size for aom_highbd_convolve8_horiz_sve epilogue: 28 bytes of instructions in range, but .seh directives corresponding to 24 bytes
```

As a workaround you could disable the code with `-DENABLE_SVE=0`. Seems like it might be an issue with clang and those intrinsics. What version is installed there?

> P.S. libavif CI does not and cannot test the CLANGARM64 build as there are no runners available. MSYS2 CI does have a self-hosted one.

Good to know.

jzern:
> > What version is installed there?
> 
> Clang 18.1.4 currently.

Thanks. This code isn't built in Chromium, which would use something closer to the tip of the tree, due to how the source is configured. So I can't say whether the issue is specific to that version.
You can file a bug on libaom: https://bugs.chromium.org/p/aomedia/issues/entry?template=Defect+report+from+user. Though this may end up better as a llvm bug.

kleisauke:
Any code with SVE intrinsics can't be used for Windows targets, see: https://github.com/llvm/llvm-project/issues/64278#issuecomment-2081649539.

jzern:
> Any code with SVE intrinsics can't be used for Windows targets, see: [llvm/llvm-project#64278 (comment)](https://github.com/llvm/llvm-project/issues/64278#issuecomment-2081649539).

Thank you for the pointer. As noted here and in those comments `-DENABLE_SVE=0` is a workaround. I don't remember seeing an error report from Martin, but maybe I missed it. Some detection could be added to libaom's CMakeLists.txt to simplify this.

PR #2132, opened 2024-04-24
FIx CI for macOS

vrabaud:
- use system CMake
- install imagemagick (which is not installed by default anymore)
- fix "use of logical '||' with constant operand" warning

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -99,8 +99,6 @@ avifBool avifGetYUVColorSpaceInfo(const avifImage * image, avifYUVColorSpaceInfo
 #if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
     const avifBool useYCgCo = (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RE) ||
                               (image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO_RO);
-#else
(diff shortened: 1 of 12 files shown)


src/reformat.c
```diff
@@ -112,7 +110,12 @@ avifBool avifGetYUVColorSpaceInfo(const avifImage * image, avifYUVColorSpaceInfo
     // YCgCo performs limited-full range adjustment on R,G,B but the current implementation performs range adjustment
     // on Y,U,V. So YCgCo with limited range is unsupported.
     if ((image->matrixCoefficients == 3 /* CICP reserved */) ||
-        ((image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO || useYCgCo) && (image->yuvRange == AVIF_RANGE_LIMITED)) ||
+        ((image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_YCGCO
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
```
jzern, line 114: `-Wconstant-logical-operand` has been around a while. I wonder if it got enabled by default in mac's clang.

jzern (approved):

src/reformat.c
vrabaud: That's my guess too. This code will soon be out of experimental anyway so it does not hurt to clean it now.

PR #2135, opened 2024-04-25
Handle unknown top level boxes with size == 0

vigneshvg:
In avifParse() right now we skip over unknown top level boxes with
size == 0 incorrectly.

The correct behavior is to simply fail on them because if we reach
that point it means that:
1) there are no more boxes left to parse (since the current unknown
   box goes on until the end of stream).
2) we haven't found all the necessary boxes for the parsing to be
   considered a succeess (either ftyp or moov or meta was not yet
   seen).


y-guyon (approved):
Note: Not sure `return AVIF_RESULT_OK;` at line 4070 can be reached.

PR #2137, opened 2024-04-26
read.c: Clamp CICP enum values to valid range

vigneshvg:
If invalid values are seen, replace them with UNSPECIFIED. This
change does not affect the color conversion paths in anyway, it just
changes what is reported in the public API.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2222,9 +2222,20 @@ static avifBool avifParseColourInformationBox(avifProperty * prop, uint64_t rawO
         colr->iccOffset = rawOffset + avifROStreamOffset(&s);
         colr->iccSize = avifROStreamRemainingBytes(&s);
     } else if (!memcmp(colorType, "nclx", 4)) {
-        AVIF_CHECK(avifROStreamReadU16(&s, &colr->colorPrimaries));          // unsigned int(16) colour_primaries;
+        AVIF_CHECK(avifROStreamReadU16(&s, &colr->colorPrimaries)); // unsigned int(16) colour_primaries;
+        if (colr->colorPrimaries == 3 ||
+            (colr->colorPrimaries > AVIF_COLOR_PRIMARIES_DCI_P3 && colr->colorPrimaries < AVIF_COLOR_PRIMARIES_EBU3213) ||
+            colr->colorPrimaries > AVIF_COLOR_PRIMARIES_EBU3213) {
+            colr->colorPrimaries = AVIF_COLOR_PRIMARIES_UNSPECIFIED;
+        }
         AVIF_CHECK(avifROStreamReadU16(&s, &colr->transferCharacteristics)); // unsigned int(16) transfer_characteristics;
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -2222,9 +2222,20 @@ static avifBool avifParseColourInformationBox(avifProperty * prop, uint64_t rawO
         colr->iccOffset = rawOffset + avifROStreamOffset(&s);
         colr->iccSize = avifROStreamRemainingBytes(&s);
     } else if (!memcmp(colorType, "nclx", 4)) {
-        AVIF_CHECK(avifROStreamReadU16(&s, &colr->colorPrimaries));          // unsigned int(16) colour_primaries;
+        AVIF_CHECK(avifROStreamReadU16(&s, &colr->colorPrimaries)); // unsigned int(16) colour_primaries;
+        if (colr->colorPrimaries == 3 ||
+            (colr->colorPrimaries > AVIF_COLOR_PRIMARIES_DCI_P3 && colr->colorPrimaries < AVIF_COLOR_PRIMARIES_EBU3213) ||
+            colr->colorPrimaries > AVIF_COLOR_PRIMARIES_EBU3213) {
```
y-guyon, line 2228: Consider adding `AVIF_COLOR_PRIMARIES_LAST` too

y-guyon (approved):

src/read.c
vigneshvg: yeah i considered it but decided not to do it because it's not very useful here since there are other holes in the range anwyays. so the check won't be simple.

i also didn't wanna update the pubic ABI simply for such a corner case check. if you strongly feel otherwise please let me know and i can add it.

src/read.c
```diff
@@ -2222,9 +2222,20 @@ static avifBool avifParseColourInformationBox(avifProperty * prop, uint64_t rawO
...
+        }
         AVIF_CHECK(avifROStreamReadU16(&s, &colr->transferCharacteristics)); // unsigned int(16) transfer_characteristics;
-        AVIF_CHECK(avifROStreamReadU16(&s, &colr->matrixCoefficients));      // unsigned int(16) matrix_coefficients;
+        if (colr->transferCharacteristics == 3 || colr->transferCharacteristics > AVIF_TRANSFER_CHARACTERISTICS_HLG) {
+            colr->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
+        }
+        AVIF_CHECK(avifROStreamReadU16(&s, &colr->matrixCoefficients)); // unsigned int(16) matrix_coefficients;
+        if (colr->matrixCoefficients == 3 || colr->matrixCoefficients >= AVIF_MATRIX_COEFFICIENTS_LAST) {
+            colr->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
+        }
```
wantehchang, line 2238: I am afraid that this pull request and the subsequent https://github.com/AOMediaCodec/libavif/pull/2139 should be reverted. The reserved values may become valid in the future. The decoder should report the reserved values faithfully.

vigneshvg: Done in https://github.com/AOMediaCodec/libavif/pull/2151

PR #2139, opened 2024-04-29
Validate CICP in all code paths

vigneshvg:
Color Primaries, Transfer Characteristics and Matrix Coefficients
are read from the bitstream in multiple places. Make sure they are
validated in all the paths.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1519,6 +1519,22 @@ static avifResult avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifI
     return AVIF_RESULT_OK;
 }
 
+static void clampCicpToValidRange(avifColorPrimaries * colorPrimaries,
+                                  avifTransferCharacteristics * transferCharacteristics,
+                                  avifMatrixCoefficients * matrixCoefficients)
+{
+    if (*colorPrimaries == 3 || (*colorPrimaries > AVIF_COLOR_PRIMARIES_DCI_P3 && *colorPrimaries < AVIF_COLOR_PRIMARIES_EBU3213) ||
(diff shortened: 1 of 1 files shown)


jzern (approved):

src/read.c
```diff
@@ -1519,6 +1519,22 @@ static avifResult avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifI
...
+    if (*colorPrimaries == 3 || (*colorPrimaries > AVIF_COLOR_PRIMARIES_DCI_P3 && *colorPrimaries < AVIF_COLOR_PRIMARIES_EBU3213) ||
+        *colorPrimaries > AVIF_COLOR_PRIMARIES_EBU3213) {
+        *colorPrimaries = AVIF_COLOR_PRIMARIES_UNSPECIFIED;
+    }
+    if (*transferCharacteristics == 3 || *transferCharacteristics > AVIF_TRANSFER_CHARACTERISTICS_HLG) {
+        *transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
+    }
+    if (*matrixCoefficients == 3 || *matrixCoefficients >= AVIF_MATRIX_COEFFICIENTS_LAST) {
+        *matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
+    }
```
wantehchang, line 1535: I am afraid that this pull request is incorrect. We should report the values faithfully. At least, we should report the reserved values faithfully, because they may become valid in the future.

vigneshvg: Done in https://github.com/AOMediaCodec/libavif/pull/2151.

PR #2142, opened 2024-05-01
obu: Do not parse more than obu_size bytes

vigneshvg:
When trying to parse a sequence header, parse only obu_size bytes.

The current code ignores obu_size and parses as long as it sees
data and could end up over-reading into other obus.


wantehchang (approved):
LGTM. Thanks.

This code comes from dav1d/src/obu.c. I checked the dav1d code. I think it is correct.

PR #2143, opened 2024-05-01
Further cleanup in avifParseSampleDescriptionBox()

wantehchang:
Rename the `remainingBytes` variable as `sampleEntryBytes`. The `remainingBytes` name made sense when the variable stored the return value of avifROStreamRemainingBytes(&s). The variable now stores the value of sampleEntryHeader.size, so the name `sampleEntryBytes` is more appropriate.

Remove an extraneous pair of parentheses around a conditional expression.

Replace an instance of sampleEntryHeader.size with `sampleEntryBytes`.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3259,7 +3259,7 @@ static avifResult avifParseSampleDescriptionBox(avifSampleTable * sampleTable,
             return AVIF_RESULT_OUT_OF_MEMORY;
         }
         memcpy(description->format, sampleEntryHeader.type, sizeof(description->format));
-        size_t sampleEntryBytes = sampleEntryHeader.size;
+        const size_t sampleEntryBytes = sampleEntryHeader.size;
         if (avifGetCodecType(description->format) != AVIF_CODEC_TYPE_UNKNOWN) {
             if (sampleEntryBytes < VISUALSAMPLEENTRY_SIZE) {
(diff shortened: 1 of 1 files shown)


wantehchang:
PTAL. This is a follow-up to https://github.com/AOMediaCodec/libavif/pull/2140.

src/read.c
```diff
@@ -3259,20 +3259,20 @@ static avifResult avifParseSampleDescriptionBox(avifSampleTable * sampleTable,
             return AVIF_RESULT_OUT_OF_MEMORY;
         }
         memcpy(description->format, sampleEntryHeader.type, sizeof(description->format));
-        size_t remainingBytes = sampleEntryHeader.size;
-        if ((avifGetCodecType(description->format) != AVIF_CODEC_TYPE_UNKNOWN)) {
-            if (remainingBytes < VISUALSAMPLEENTRY_SIZE) {
+        size_t sampleEntryBytes = sampleEntryHeader.size;
```
vigneshvg, line 3262: nit: we could also mark this as `const` while we are at it.

vigneshvg (approved):

src/read.c
wantehchang: Done

vigneshvg (approved):

y-guyon (approved):

PR #2144, opened 2024-05-02
obu: Validate obu_forbidden_bit in sequence header

vigneshvg:
Section 6.2.2 of AV1 Specification: obu_forbidden_bit must be set to 0.

### DIFF
--- a/src/obu.c
+++ b/src/obu.c
@@ -413,7 +413,10 @@ avifBool avifSequenceHeaderParse(avifSequenceHeader * header, const avifROData *
         avifBitsInit(&bits, obus.data, obus.size);
 
         // obu_header()
-        avifBitsRead(&bits, 1); // obu_forbidden_bit
+        const uint8_t obu_forbidden_bit = avifBitsRead(&bits, 1);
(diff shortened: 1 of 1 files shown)


src/obu.c
```diff
@@ -413,7 +413,10 @@ avifBool avifSequenceHeaderParse(avifSequenceHeader * header, const avifROData *
         avifBitsInit(&bits, obus.data, obus.size);
 
         // obu_header()
-        avifBitsRead(&bits, 1); // obu_forbidden_bit
+        const uint8_t obu_forbidden_bit = avifBitsRead(&bits, 1);
```
wantehchang, line 416: Nit: The return type of `avifBitsRead()` is `uint32_t`, so it is simpler to use the `uint32_t` type here (to avoid a type conversion).

wantehchang (commented):
LGTM.

src/obu.c
vigneshvg: Done in https://github.com/AOMediaCodec/libavif/pull/2146.

PR #2145, opened 2024-05-02
obu: Handle size == 0 in avifBitsInit

vigneshvg:
If size == 0, eof has to be set to 1 in avifBitsInit for the
rest of the code to work as intended. Otherwise the first byte
is read unconditionally which is incorrect.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=68568

### DIFF
--- a/src/obu.c
+++ b/src/obu.c
@@ -66,7 +66,7 @@ static void avifBitsInit(avifBits * const bits, const uint8_t * const data, cons
     bits->bitsLeft = 0;
     bits->state = 0;
     bits->error = 0;
-    bits->eof = 0;
+    bits->eof = (size == 0);
 }
 
 static void avifBitsRefill(avifBits * const bits, const uint32_t n)


src/obu.c
```diff
@@ -439,6 +439,11 @@ avifBool avifSequenceHeaderParse(avifSequenceHeader * header, const avifROData *
             return AVIF_FALSE;
 
         if (obu_type == 1) { // Sequence Header
+            if (obu_size == 0) {
+                // This has to be checked for explicitly because avifBitsInit
+                // does not validate whether there is at least one byte or not.
+                return AVIF_FALSE;
+            }
```
wantehchang, line 446: I suggest including <assert.h> and adding `assert(size != 0)` to the beginning of `avifBitsInit`.

Alternatively, we can change line 69 from

```
    bits->eof = 0;
```

to

```
    bits->eof = (size == 0);
```

wantehchang (approved):

src/obu.c
vigneshvg: Good point. I have updated the code to set eof to size == 0 since that is a better fix.

wantehchang (approved):

PR #2146, opened 2024-05-02
obu: Use uint32_t for obu_forbidden_bits

vigneshvg:
This avoids a type conversion.

### DIFF
--- a/src/obu.c
+++ b/src/obu.c
@@ -413,7 +413,7 @@ avifBool avifSequenceHeaderParse(avifSequenceHeader * header, const avifROData *
         avifBitsInit(&bits, obus.data, obus.size);
 
         // obu_header()
-        const uint8_t obu_forbidden_bit = avifBitsRead(&bits, 1);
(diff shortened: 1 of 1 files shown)


wantehchang (approved):
LGTM. Thanks!

vigneshvg:
This also fixes CI failures such as this one: https://github.com/AOMediaCodec/libavif/actions/runs/8931900623/job/24534782963

vigneshvg:
> Were the CI actions not run on #2144?

They ran and passed. But it seems like the windows CI bots only run on push (and on PRs only when certain files are changed).

I don't know why they are set up this way, may be it is too slow to run on every PR?

https://github.com/AOMediaCodec/libavif/blob/main/.github/workflows/ci-windows.yml#L16
https://github.com/AOMediaCodec/libavif/blob/main/.github/workflows/ci-windows-installed.yml#L9

wantehchang:
Thank you. I will propose to run ci-windows-installed.yml on every PR.

wantehchang:
See https://github.com/AOMediaCodec/libavif/pull/2149.

PR #2148, opened 2024-05-03
read.c: Validate that there is at least one trak box

vigneshvg:
When parsing the 'moov' box, ensure that there is at least one
'trak' box that was successfully parsed. If not, report it as a
failure.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3530,6 +3530,7 @@ static avifResult avifParseMovieBox(avifDecoderData * data,
 {
     BEGIN_STREAM(s, raw, rawLen, data->diag, "Box[moov]");
 
+    avifBool hasTrak = AVIF_FALSE;
     while (avifROStreamHasBytesLeft(&s, 1)) {
         avifBoxHeader header;
         AVIF_CHECKERR(avifROStreamReadBoxHeader(&s, &header), AVIF_RESULT_BMFF_PARSE_FAILED);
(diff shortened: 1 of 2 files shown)


wantehchang (commented):
Note to self: `decoder->imageSequenceTrackPresent` was added in commit 5410b23f7.

src/read.c
```diff
@@ -4050,7 +4050,7 @@ static avifResult avifParse(avifDecoder * decoder)
             AVIF_CHECKRES(
                 avifParseMovieBox(data, boxOffset, boxContents.data, boxContents.size, decoder->imageSizeLimit, decoder->imageDimensionLimit));
             moovSeen = AVIF_TRUE;
-            decoder->imageSequenceTrackPresent = AVIF_TRUE;
+            decoder->imageSequenceTrackPresent = (decoder->data->tracks.count > 0);
```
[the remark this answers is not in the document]
vigneshvg: Done. We don't need the `hasTrak` variable. we can simply check tracks.count since avifParseTrackBox always creates a track on success.

Also added a test.

src/read.c
```diff
@@ -3541,6 +3541,10 @@ static avifResult avifParseMovieBox(avifDecoderData * data,
 
         AVIF_CHECKERR(avifROStreamSkip(&s, header.size), AVIF_RESULT_BMFF_PARSE_FAILED);
     }
+    if (data->tracks.count == 0) {
```
wantehchang, line 3544: It would be good to add a comment to indicate that this simple check depends on the fact that there is at most one "moov" box:

> Box Type: `'moov'`
> Container: File
> Mandatory: Yes
> Quantity: Exactly one

The check I suggested does not depend on this requirement.

src/read.c
```diff
@@ -3541,6 +3541,10 @@ static avifResult avifParseMovieBox(avifDecoderData * data,
...
+        avifDiagnosticsPrintf(data->diag, "moov box did not contain any tracks");
```
wantehchang, line 3545: Nit/optional: We can use the present tense "does" instead of the past tense "did".

wantehchang (approved):

src/read.c

src/read.c
vigneshvg: I have updated the code to use a local boolean like you suggested earlier.

wantehchang (approved):
LGTM. Thanks.

Note that your previous version is also correct. It's just that it requires the precondition that `avifParseMovieBox()` is called only when we have not seen any movie ("moov") box.

PR #2152, opened 2024-05-04
Fix size 0 handling in avifParse()

vigneshvg:
header.size in avifParse is not the raw value that is found in
the input stream. It is calculated in avifROStreamReadBoxHeaderPartial.

So when handling size == 0 in avifParse(), we cannot use header.size
since it may be 0 in legit cases (where the box itself is empty).

This PR adds a new member to avifBoxHeader which stores the actual
raw size found in the input stream and uses that in avifParse().

Also handle the isSizeZeroBox in avifPeekCompatibleFileType.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -647,6 +647,11 @@ typedef size_t avifBoxMarker;
 
 typedef struct avifBoxHeader
 {
+    // If set to AVIF_TRUE, it means that the box goes on until the end of the
+    // stream. So, |size| must be set to the number of bytes left in the input
+    // stream. If set to AVIF_FALSE, |size| indicates the size of the box in
+    // bytes, excluding the box header.
+    avifBool isSizeZeroBox;
     // Size of the box in bytes, excluding the box header.
     size_t size;
 
--- a/src/read.c
+++ b/src/read.c
@@ -3995,8 +3995,8 @@ static avifResult avifParse(avifDecoder * decoder)
 #endif
             boxOffset = parseOffset;
             size_t sizeToRead;
-            if (header.size == 0) {
-                // A size of 0 means the box body goes till the end of the file.
+            if (header.isSizeZeroBox) {
+                // The box body goes till the end of the file.
                 if (decoder->io->sizeHint != 0 && decoder->io->sizeHint - parseOffset < SIZE_MAX) {
                     sizeToRead = decoder->io->sizeHint - parseOffset;
                 } else {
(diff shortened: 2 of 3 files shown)


y-guyon (approved):

vigneshvg:
I've updated the PR to use a boolean for checking if the box is of size 0 (instead of storing the raw_size). PTAL. Thanks!

wantehchang:
Vignesh: Thanks for the fix. I will take a look at this pull request now.

include/avif/internal.h
```diff
@@ -647,6 +647,10 @@ typedef size_t avifBoxMarker;
 
 typedef struct avifBoxHeader
 {
+    // If set to AVIF_TRUE, then |size| will be set to 0 and it means that the
```
wantehchang, line 650: The comment `|size| will be set to 0` is inaccurate because of line 4013 in src/read.c:

```
4012             if (header.isSizeZeroBox) {
4013                 header.size = boxContents.size;
```

One way to make this comment accurate would be to set a local variable, say `boxSize`, to either `header.size` or `boxContents.size` and use `boxSize` instead of `header.size` in line 4025 in src/read.c. Another way would be to update the comment to also describe what we do in line 4013 in src/read.c.

src/read.c
```diff
@@ -4009,15 +4009,15 @@ static avifResult avifParse(avifDecoder * decoder)
...
+            if (header.isSizeZeroBox) {
                 header.size = boxContents.size;
             } else if (boxContents.size != header.size) {
                 // A truncated box, bail out
                 return AVIF_RESULT_TRUNCATED_DATA;
             }
         } else if (header.size > (UINT64_MAX - parseOffset)) {
             return AVIF_RESULT_BMFF_PARSE_FAILED;
-        } else if (header.size == 0) {
+        } else if (header.isSizeZeroBox) {
```
wantehchang, line 4020: I suggest reversing the checks at line 4018 and line 4020. We should use `header.size` only when `header.isSizeZeroBox` is false.

wantehchang (commented):
IMPORTANT: We also need to fix size 0 handling in `avifPeekCompatibleFileType()`. (I found this using Code Search. It would be good to analyze the relevant code using Code Search.)

src/read.c
vigneshvg: Done.

vigneshvg:
> IMPORTANT: We also need to fix size 0 handling in `avifPeekCompatibleFileType()`. (I found this using Code Search. It would be good to analyze the relevant code using Code Search.)

Done, i have updated `avifPeekCompatibleFileType()` as well.

src/read.c
```diff
@@ -4126,9 +4126,10 @@ avifBool avifPeekCompatibleFileType(const avifROData * input)
     if (!avifROStreamReadBoxHeaderPartial(&s, &header, /*topLevel=*/AVIF_TRUE) || memcmp(header.type, "ftyp", 4)) {
         return AVIF_FALSE;
     }
-    if (header.size == 0) {
-        // The size of the 'ftyp' box is 0, which means it goes till the end of the file.
-        // Either there is no brand requiring anything in the file but a FileTypebox (so not AVIF), or it is invalid.
+    if (header.isSizeZeroBox || header.size == 0) {
+        // * If header.isSizeZeroBox is true, the 'ftyp' box goes on till the end of the file. This cannot be valid AVIF since
+        //   there are no other boxes left for the image data.
+        // * If header.size == 0, the 'ftyp' box is empty. So there are no brands identifying as AVIF.
```
wantehchang, line 4132: We let the `header.size == 0` case fail inside the `avifParseFileTypeBox()` call at line 4139 as before. No need to handle it as a special case here, just like we don't handle `header.size` equal to 1, 2, ..., 7 as special cases.

wantehchang (approved):
LGTM. I suggest a simple change to `avifPeekCompatibleFileType()`.

Also, we may want to mention the fix to `avifPeekCompatibleFileType()` in the commit message.

src/read.c
vigneshvg: Done, i have restored the comment here also to it's previous state.

vigneshvg:
> LGTM. I suggest a simple change to `avifPeekCompatibleFileType()`.
> 
> Also, we may want to mention the fix to `avifPeekCompatibleFileType()` in the commit message.

Done.

wantehchang (approved):

PR #2154, opened 2024-05-07
read.c: Early return on unsupported iref version

vigneshvg:
We do not support iref boxes with version > 1. Instead of breaking
from the first iteration of each nested loop, simply return early
when version is greater than 1.


src/read.c
```diff
@@ -2879,6 +2879,10 @@ static avifResult avifParseItemReferenceBox(avifMeta * meta, const uint8_t * raw
 
     uint8_t version;
     AVIF_CHECKERR(avifROStreamReadVersionAndFlags(&s, &version, NULL), AVIF_RESULT_BMFF_PARSE_FAILED);
+    if (version > 1) {
+        // iref versions > 1 are not supported. skip it.
```
wantehchang, line 2883: Nit: Capitalize "skip" because it is after a period.

wantehchang (approved):

src/read.c

wantehchang (approved):

PR #2155, opened 2024-05-07
read.c: Fix hdlr duplicate detection

vigneshvg:
'hdlr' box has to be the first box inside meta and it can occur
only once. The current duplicate detection for 'hdlr' is not
correct as it is inside the `if (firstBox)` block.

Update the code to check for duplicate `hdlr` occurrence outside
of the `if (firstBox)` block.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2966,7 +2966,6 @@ static avifResult avifParseMetaBox(avifMeta * meta, uint64_t rawOffset, const ui
 
         if (firstBox) {
             if (!memcmp(header.type, "hdlr", 4)) {
-                AVIF_CHECKERR(uniqueBoxSeen(&uniqueBoxFlags, 0, "meta", "hdlr", diag), AVIF_RESULT_BMFF_PARSE_FAILED);
                 AVIF_CHECKERR(avifParseHandlerBox(avifROStreamCurrent(&s), header.size, diag), AVIF_RESULT_BMFF_PARSE_FAILED);
                 firstBox = AVIF_FALSE;
             } else {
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -2966,14 +2966,16 @@ static avifResult avifParseMetaBox(avifMeta * meta, uint64_t rawOffset, const ui
...
                 AVIF_CHECKERR(avifParseHandlerBox(avifROStreamCurrent(&s), header.size, diag), AVIF_RESULT_BMFF_PARSE_FAILED);
                 firstBox = AVIF_FALSE;
             } else {
                 // hdlr must be the first box!
                 avifDiagnosticsPrintf(diag, "Box[meta] does not have a Box[hdlr] as its first child box");
                 return AVIF_RESULT_BMFF_PARSE_FAILED;
             }
+        } else if (!memcmp(header.type, "hdlr", 4)) {
+            avifDiagnosticsPrintf(diag, "Box[meta] contains a duplicate unique box of type 'hdlr'");
+            return AVIF_RESULT_BMFF_PARSE_FAILED;
```
wantehchang, line 2978: Nit: I suggest renumbering the second arguments passed to the `uniqueBoxSeen()` calls below to start from 0, otherwise people who read this code may be wondering why `whichFlag=0` is not used.

Alternatively, we could continue to use `uniqueBoxSeen(&uniqueBoxFlags, 0, "meta", "hdlr", diag)` to detect duplicate hdlr boxes. It would require two calls to `uniqueBoxSeen(&uniqueBoxFlags, 0, "meta", "hdlr", diag)` -- keep the original call, and add a new call here.

```
        } else if (!memcmp(header.type, "hdlr", 4)) {
            AVIF_CHECKERR(uniqueBoxSeen(&uniqueBoxFlags, 0, "meta", "hdlr", diag), AVIF_RESULT_BMFF_PARSE_FAILED);
            // Not reached.
```

wantehchang (commented):
Thank you very much for the fix. I suggest a change.

src/read.c
vigneshvg: Done, updated the other calls to reduce `whichFlags` by 1.

wantehchang (approved):

PR #2162, opened 2024-05-18
GetSeedDataDirs: allow TEST_DATA_DIRS to be -D defined

jzern:
This allows build/test environments to set the path if they are unable
influence the environment. This is based on the implementations in
libvpx and libaom.

Bug: https://crbug.com/329464135

### DIFF
--- a/tests/gtest/avif_fuzztest_helpers.cc
+++ b/tests/gtest/avif_fuzztest_helpers.cc
@@ -187,11 +187,13 @@ class Environment : public ::testing::Environment {
 
 //------------------------------------------------------------------------------
 
+#ifdef TEST_DATA_DIRS
 // Helper macros to ensure TEST_DATA_DIRS is a quoted string.
-// NOTE: TEST_DATA_DIRS MUST NOT be a quoted string before stringification
+// NOTE: TEST_DATA_DIRS must NOT be a quoted string before stringification
 // otherwise it will be quoted twice, resulting in the use of an incorrect path.
 #define AVIF_TO_STRING(S) #S
 #define AVIF_STRINGIFY(S) AVIF_TO_STRING(S)
+#endif
 
 std::vector<std::string> GetSeedDataDirs() {
   const char* var = std::getenv("TEST_DATA_DIRS");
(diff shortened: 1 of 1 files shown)


jzern:
Looks like ubuntu-latest might need some additional packages now.

tests/gtest/avif_fuzztest_helpers.cc
```diff
@@ -187,9 +187,20 @@ ::testing::Environment* SetEnv(const char* name, const char* value) {
 
 //------------------------------------------------------------------------------
 
+// Helper macros to ensure TEST_DATA_DIRS is a quoted string.
+// NOTE: TEST_DATA_DIRS MUST NOT be a quoted string before stringification
+// otherwise it will be quoted twice, resulting in the use of an incorrect path.
+#define AVIF_TO_STRING(S) #S
+#define AVIF_STRINGIFY(S) AVIF_TO_STRING(S)
```
y-guyon, line 194: Could be in the `#ifdef TEST_DATA_DIRS` block below because it is used only there

tests/gtest/avif_fuzztest_helpers.cc
```diff
@@ -187,9 +187,20 @@ ::testing::Environment* SetEnv(const char* name, const char* value) {
 
 //------------------------------------------------------------------------------
 
+// Helper macros to ensure TEST_DATA_DIRS is a quoted string.
+// NOTE: TEST_DATA_DIRS MUST NOT be a quoted string before stringification
```
y-guyon, line 191: `TEST_DATA_DIRS must NOT` is easier to read

y-guyon (approved):
Very minor comments, feel free to submit as is to save time

tests/gtest/avif_fuzztest_helpers.cc
jzern: Done.

tests/gtest/avif_fuzztest_helpers.cc
jzern: Done.

PR #2164, opened 2024-05-20
Switch from gcc 13 to gcc 12 in ubuntu-latest

wantehchang:
gcc-13 was removed from the ubuntu-22.04 os image. See https://github.com/actions/runner-images/issues/8555#issuecomment-2115818597

### DIFF
--- a/.github/workflows/ci-disable-gtest.yml
+++ b/.github/workflows/ci-disable-gtest.yml
@@ -32,7 +32,7 @@ jobs:
         include:
           - runs-on: ubuntu-latest
             compiler: gcc
-            gcc: 13
+            gcc: 12
 
     steps:
     - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
--- a/.github/workflows/ci-linux-golden-tests.yml
+++ b/.github/workflows/ci-linux-golden-tests.yml
@@ -23,7 +23,7 @@ jobs:
         include:
           - runs-on: ubuntu-latest
             compiler: gcc
-            gcc: 13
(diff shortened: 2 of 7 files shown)


wantehchang:
@vrabaud @y-guyon FYI.

There are two other options:
https://github.com/actions/runner-images/issues/9866#issuecomment-2114583692

vigneshvg (approved):

jzern:
lgtm.

wantehchang:
Note that the ubuntu-latest alias is still pointing to ubuntu-22.04. It is ubuntu-22.04 itself that changed.

vrabaud:
Why even mention a specific version ? Doesn't the default gcc work anyway ?

vrabaud:
gcc is indeed at 11.2 by default: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md
Ubuntu 24.04 is deployed: https://github.com/actions/runner-images?tab=readme-ov-file#available-images so we might as well use it: https://github.com/AOMediaCodec/libavif/pull/2165

wantehchang:
Vincent:

The use of a specific version of gcc started in the very first version of these CI workflow files (.github/workflows/ci.yml):
https://github.com/AOMediaCodec/libavif/commit/0f8fcf8db

I guess the reason was what Miloš said, to use the newest version of gcc in the os image, rather than the default version.

The ubuntu-24.04 os image still has the "beta" status. That's why I didn't consider it yesterday. I probably would wait until ubuntu-24.04 goes out of beta, but we can certainly volunteer to be a beta tester.

PR #2166, opened 2024-05-22
Use composite actions

vrabaud:
Based on:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow

### DIFF
--- a/.github/actions/setup-common/action.yml
+++ b/.github/actions/setup-common/action.yml
@@ -35,9 +35,8 @@ runs:
       shell: bash
     - name: Set up ninja
       uses: seanmiddleditch/gha-setup-ninja@8b297075da4cd2a5f1fd21fe011b499edf06e9d2 # v4
-
     - name: Set up nasm
-      if: ${{ inputs.codec-aom == 'LOCAL' }}
+      if: ${{ inputs.codec-aom == 'LOCAL' || inputs.codec-dav1d == 'LOCAL' }}
       uses: ilammy/setup-nasm@13cbeb366c45c4379d3478cdcbadd8295feb5028 # v1.5.1
     - name: Set up meson
       if: ${{ inputs.codec-dav1d == 'LOCAL' }}
--- a/.github/actions/setup-common/action.yml
+++ b/.github/actions/setup-common/action.yml
@@ -1,0 +1,52 @@
+name: 'Common setup for all OSes'
+description: 'Installs common dependencies'
+inputs:
+  codec-aom:
+    description: 'Can take the values: OFF, LOCAL, SYSTEM'
+    default: 'OFF'
+  codec-dav1d:
+    description: 'Can take the values: OFF, LOCAL, SYSTEM'
+    default: 'OFF'
+  codec-rav1e:
+    description: 'Can take the values: OFF, LOCAL, SYSTEM'
+    default: 'OFF'
+  recent-cmake:
+    description: 'Can take the values: true, false. Only useful on Linux'
+    default: 'false'
+runs:
+  using: "composite"
+  steps:
+    - name: Set up Python
+      uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
+      with:
+        python-version: '3.x'
+    - name: Set up CMake < 3.18
+      if: ${{ runner.os == 'Linux' && inputs.recent-cmake == 'false' }}
+      uses: jwlawson/actions-setup-cmake@802fa1a2c4e212495c05bf94dba2704a92a472be # v2.0.2
+      with:
+        cmake-version: '3.13.x'
+    - name: Set up CMake >= 3.18
+      if: ${{ runner.os == 'Linux' && inputs.recent-cmake == 'true' }}
+      uses: jwlawson/actions-setup-cmake@802fa1a2c4e212495c05bf94dba2704a92a472be # v2.0.2
+      with:
+        cmake-version: '3.18.x'
+    - name: Print CMake version
+      run: cmake --version
+      shell: bash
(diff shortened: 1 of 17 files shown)


.github/workflows/ci-unix-static-sanitized.yml
```diff
@@ -26,55 +26,57 @@ jobs:
...
-        id: cache
+        id: setup_macos
         with:
-          extra-key: ${{ matrix.sanitizer }}
+          codec-aom: 'LOCAL'
+          codec-dav1d: 'LOCAL'
+          codec-rav1e: 'OFF'
+          extra-cache-key: ${{ matrix.sanitizer }}
+      - id: cache-hit
+        run: echo "hit=$${{ (runner.os == 'Linux' && steps.setup_linux.outputs.ext-cache-hit != 'true') || (runner.os == 'macOS' && steps.setup_macos.outputs.ext-cache-hit != 'true') }}" >> "$GITHUB_OUTPUT"
```
wantehchang, line 46: IMPORTANT: The variable "hit" means ext-cache-hit is not true. Should the variable be renamed "miss" and the id  renamed "cache-miss"?

.github/actions/setup-common/action.yml
```diff
@@ -0,0 +1,52 @@
...
+      with:
+        cmake-version: '3.13.x'
+    - name: Set up CMake >= 3.18
+      if: ${{ runner.os == 'Linux' && inputs.recent-cmake == 'true' }}
+      uses: jwlawson/actions-setup-cmake@802fa1a2c4e212495c05bf94dba2704a92a472be # v2.0.2
+      with:
+        cmake-version: '3.18.x'
+    - name: Print CMake version
+      run: cmake --version
+      shell: bash
```
wantehchang, line 35: Do we need to specify the shell? It should not matter what the shell is.

This coment also applies to the other "shell: bash" lines in this pull request except where noted.

.github/actions/setup-linux/action.yml
```diff
@@ -0,0 +1,78 @@
...
+    - uses: ./.github/actions/cache
+      id: cache
+      with:
+        extra-key: ${{ inputs.extra-cache-key }}
+        use-rust: ${{ inputs.codec-rav1e == 'LOCAL' }}
+
+    - name: Set GCC & G++ compiler
+      if: ${{ inputs.gcc-version != '' }}
+      run: echo "CC=gcc-${{ inputs.gcc-version }}" >> $GITHUB_ENV && echo "CXX=g++-${{ inputs.gcc-version }}" >> $GITHUB_ENV
+      shell: bash
```
wantehchang, line 78: Here it matters which shell we are using, but bash is very likely the default.

wantehchang (commented):
Vincent: I have some questions and suggested changes.

.github/actions/setup-linux/action.yml
```diff
@@ -0,0 +1,78 @@
...
+  extra-cache-key:
+    description: 'Extra cache key to use in the cache name. Useful when several caches are used in one workflow.'
+    default: ''
+  gcc-version:
+    description: 'Can be empty, in which case CC and CXX are not overriden'
+    default: ''
+  libxml2:
+    default: 'OFF'
+  libyuv:
+    default: 'OFF'
```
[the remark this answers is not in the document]
vrabaud: Yes, done.

.github/workflows/ci-unix-static-sanitized.yml
```diff
@@ -26,55 +26,57 @@ jobs:
...
+
       - name: Build aom
-        if: steps.cache.outputs.ext-cache-hit != 'true'
+        if: $${{ steps.cache-hit.outputs.hit }}
```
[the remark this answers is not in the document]
vrabaud: Done

.github/workflows/ci-unix-static-sanitized.yml
vrabaud: Done

.github/actions/setup-common/action.yml
vrabaud: We actually do need to specify it in composite actions: https://github.com/orgs/community/discussions/18597

.github/actions/setup-linux/action.yml
vrabaud: Any shell matters as long as GITHUB_ENV is filled.

wantehchang: My previous comment may not be clear. I meant that in this step the shell has to be bash because the command uses bash's syntax.

.github/workflows/ci-unix-static-sanitized.yml
```diff
@@ -26,55 +26,57 @@ jobs:
...
-      - uses: ./.github/actions/cache
-        id: cache
+        id: setup_macos
         with:
-          extra-key: ${{ matrix.sanitizer }}
+          codec-aom: 'LOCAL'
+          codec-dav1d: 'LOCAL'
+          codec-rav1e: 'OFF'
+          extra-cache-key: ${{ matrix.sanitizer }}
+      - id: cache-miss
```
wantehchang, line 45: I am sorry that my previous comment did not say that we could also fix this by defining the  output `hit` correctly here and testing `hit` correctly below. Then we could still use the id `cache-hit` and the output `hit`. I think this would be better because it would be more consistent.

wantehchang (approved):

.github/workflows/ci-unix-static-sanitized.yml
vrabaud: You are right, let's keep the conventional `hit` word.

PR #2169, opened 2024-05-23
Rely on CMake to build the build folder.

vrabaud:
This is useful in case the build folder already exists due to caching. "mkdir -p" could be used on Unix (not Windows) but it is easier to use a common command.

### DIFF
--- a/.github/workflows/ci-disable-gtest.yml
+++ b/.github/workflows/ci-disable-gtest.yml
@@ -73,9 +73,7 @@ jobs:
 
     - name: Prepare libavif (cmake)
       run: >
-        mkdir build && cd build
-
-        cmake .. -G Ninja
+        cmake .. -G Ninja -S ./ -B build
         -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF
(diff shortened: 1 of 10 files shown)


.github/workflows/ci-disable-gtest.yml
```diff
@@ -73,9 +73,7 @@ jobs:
 
     - name: Prepare libavif (cmake)
       run: >
-        mkdir build && cd build
-
-        cmake .. -G Ninja
+        cmake .. -G Ninja -S ./ -B build
```
y-guyon, line 76: Remove `..`

`-S .` is enough

y-guyon (approved):

.github/workflows/ci-disable-gtest.yml
```diff
@@ -73,9 +73,7 @@ jobs:
 
     - name: Prepare libavif (cmake)
       run: >
-        mkdir build && cd build
-
-        cmake .. -G Ninja
+        cmake -G Ninja -S ./ -B build
```
wantehchang, line 76: The `/` after `.` is not necessary. We don't have `/` after `build`.

vrabaud: Fixed in https://github.com/AOMediaCodec/libavif/pull/2176

PR #2172, opened 2024-05-23
Properly check for cache hits.

vrabaud:
`steps.cache-ext.outputs.cache-hit != 'true'` is actually always true because `steps.cache-ext.outputs.cache-hit` does not exist anymore, as the step `cache-hit` is now hidden in the cache action. The `cache-hit` information has to be sent back from the cache action.

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -4,6 +4,9 @@ inputs:
   use-rust:
     description: 'Whether rust is used'
     default: false
+outputs:
+  ext-cache-hit:
+    value: ${{ steps.cache-ext.outputs.cache-hit }}
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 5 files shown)


y-guyon (approved):
I am surprised this change is needed, it was working as is at some point.

wantehchang (changes requested):
Vincent: Searching for "cache-hit" in https://github.com/actions/cache, I found that we are testing `cache-hit` in the same way as the sample code.

Are you sure this pull request is necessary?

vrabaud:
Let's put this one on the back burner, maybe I thgought it was necessary because of a cache miss.

vrabaud:
The cache hit was not checked properly: the answer from the action has to be sent back to the main workflow.

y-guyon (approved):

PR #2173, opened 2024-05-23
Rust fixes

vrabaud:
- switch deprecated actions-rs/toolchain, cf https://github.com/actions-rs/toolchain/issues/216
- do not download cargo: it does not always work (e.g. macOS has several architectures but there is only one download), it is complex code, and we have a proper way to install it through CMake cf
https://github.com/AOMediaCodec/libavif/blob/c21943e1e07ced29a22d1a59b0c4748e00b84c67/cmake/Modules/LocalRav1e.cmake#L49

### DIFF
--- a/.github/workflows/ci-disable-gtest.yml
+++ b/.github/workflows/ci-disable-gtest.yml
@@ -42,21 +42,12 @@ jobs:
     - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
       with:
         python-version: '3.x'
-    - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
-      with:
-        profile: minimal
-        toolchain: stable
-        override: true
+    - name: Set up rust
+      uses: dtolnay/rust-toolchain@stable
     - name: Install dependencies
       run: |
         sudo apt update
         sudo apt install imagemagick libjpeg-turbo8-dev libpng-dev
-    - name: Install cargo-c
-      env:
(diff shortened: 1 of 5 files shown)


y-guyon (approved):

.github/workflows/ci-disable-gtest.yml
```diff
@@ -42,21 +42,12 @@ jobs:
     - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
       with:
         python-version: '3.x'
-    - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
-      with:
-        profile: minimal
-        toolchain: stable
-        override: true
+    - name: Set up rust
+      uses: dtolnay/rust-toolchain@stable
```
y-guyon, line 46:
```yml
uses: dtolnay/rust-toolchain@d8352f6b1d2e870bc5716e7a6d9b65c4cc244a1a
```

y-guyon: Maybe take the last commit from the `stable` branch so https://github.com/dtolnay/rust-toolchain/commit/d388a4836fcdbde0e50e395dc79a2670ccdef13f

vrabaud: Done

vrabaud: It is not possible to stick to a commit. Reverting the rust dependency. What really matters in this PR is the cargo install.

y-guyon (approved):

PR #2174, opened 2024-05-23
Use common code for caching

vrabaud:
Also fixes:
- the cargo cache needs to be called before CMake
- the cargo cache hash needs to depend on existing file (otheriwse the value is just the empty string which creates invalid cache hits)
- the cache can be split between ext./ and build/_deps to trigger its refresh less often
- the hash keys need to take the architecture into account (as macOS can have multiple architecures)

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -1,14 +1,14 @@
 name: 'Cache for all OSes'
 description: 'Needs to be called before any call to ext/.cmd|sh or cmake'
 inputs:
-  codec-rav1e:
-    description: 'Can take the values: OFF, LOCAL, SYSTEM'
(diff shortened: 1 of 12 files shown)


y-guyon (approved):
Thanks!

vrabaud:
To answer some concerns: the number of caches is not limited, just the overall size (cf https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy). This PR increases the number of caches (by not sharing them between CI, which is more future proof in case CMake parameters change in one CI and not the other), but we should be fine with the 10G limit. Let's keep track of it at https://github.com/AOMediaCodec/libavif/actions/caches though.

PR #2175, opened 2024-05-24
Actually cache all of ~/.cargo

vrabaud:


y-guyon (approved):
You may need to change the key or to delete the cache manually

vrabaud:
Caches should not be deleted manually anymore. If caches fail, there is a bug, Just fixed it in https://github.com/AOMediaCodec/libavif/pull/2172

PR #2177, opened 2024-05-27
Finer grain cache.

vrabaud:
- only consider the current workflow file (and not all of them) to rebuild a cache.
- fix action files that were not taken into account for the key
- allow for workflows to specialize the cache (useful in case of matrix)

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -4,27 +4,42 @@ inputs:
   use-rust:
     description: 'Whether rust is used'
     default: false
+  extra-key:
+    description: 'Extra cache key to use in the cache name. Useful when several caches are used in one workflow.'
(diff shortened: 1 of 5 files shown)


y-guyon (approved):

.github/actions/cache/action.yml
```diff
@@ -4,27 +4,42 @@ inputs:
   use-rust:
     description: 'Whether rust is used'
     default: false
+  extra-key:
+    description: 'Extra cache key to use in the cache name. Useful when several caches are used in one workflow.'
```
wantehchang, line 8: Nit: In the second sentence "Useful when ...", we could mention this is true when matrix is used.

wantehchang (commented):
LGTM.

PR #2178, opened 2024-05-27
Properly check for cargo cache hit.

vrabaud:
Boolean are not booleans in composite actions, cf
https://github.com/actions/runner/issues/2238


wantehchang (commented):
LGTM.

PR #2181, opened 2024-05-30
read.c: Treat version 1 as version 0 for stsd box

vigneshvg:
According to 8.5.2.3 of ISO/IEC 14496-12:
version is set to zero. A version number of 1 shall be treated as
a version of 0.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -3227,7 +3227,15 @@ static avifResult avifParseSampleDescriptionBox(avifSampleTable * sampleTable,
 {
     BEGIN_STREAM(s, raw, rawLen, diag, "Box[stsd]");
 
-    AVIF_CHECKERR(avifROStreamReadAndEnforceVersion(&s, 0), AVIF_RESULT_BMFF_PARSE_FAILED);
+    uint8_t version;
+    AVIF_CHECKERR(avifROStreamReadVersionAndFlags(&s, &version, NULL), AVIF_RESULT_BMFF_PARSE_FAILED);
+
+    // Section 8.5.2.3 of ISO/IEC 14496-12:
+    //   version is set to zero. A version number of 1 shall be treated as a version of 0.
+    if (version != 0 && version != 1) {
+        avifDiagnosticsPrintf(diag, "Box[stsd]: Expecting box version 0 or 1, got version %u", version);
+        return AVIF_RESULT_BMFF_PARSE_FAILED;
+    }
 
     uint32_t entryCount;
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -3227,7 +3227,15 @@ static avifResult avifParseSampleDescriptionBox(avifSampleTable * sampleTable,
...
     BEGIN_STREAM(s, raw, rawLen, diag, "Box[stsd]");
 
-    AVIF_CHECKERR(avifROStreamReadAndEnforceVersion(&s, 0), AVIF_RESULT_BMFF_PARSE_FAILED);
+    uint8_t version;
+    AVIF_CHECK(avifROStreamReadVersionAndFlags(&s, &version, NULL));
+
+    // Section 8.5.2.3 of ISO/IEC 14496-12:
+    //   version is set to zero. A version number of 1 shall be treated as a version of 0.
+    if (version != 0 && version != 1) {
+        avifDiagnosticsPrintf(diag, "Box[stsd] Expecting box version 0 or 1, got version %u", version);
```
wantehchang, line 3236: Nit: Add a colon `:` after "Box[stsd]".

src/read.c
```diff
@@ -3227,7 +3227,15 @@ static avifResult avifParseSampleDescriptionBox(avifSampleTable * sampleTable,
 {
     BEGIN_STREAM(s, raw, rawLen, diag, "Box[stsd]");
 
-    AVIF_CHECKERR(avifROStreamReadAndEnforceVersion(&s, 0), AVIF_RESULT_BMFF_PARSE_FAILED);
+    uint8_t version;
+    AVIF_CHECK(avifROStreamReadVersionAndFlags(&s, &version, NULL));
```
wantehchang, line 3231: IMPORTANT: Change `AVIF_CHECK` to `AVIF_CHECKERR(..., AVIF_RESULT_BMFF_PARSE_FAILED)`, as in the original code.

wantehchang (commented):
Vignesh: I suggest two changes.

@y-guyon We should think about whether an unsupported version should be reported as an error or silently ignored. (This pull request follows our current policy of reporting an unsupported version as an error.)

src/read.c
vigneshvg: Done.

src/read.c
vigneshvg: Done.

wantehchang (approved):
LGTM. Thanks!

y-guyon:
> @y-guyon We should think about whether an unsupported version should be reported as an error or silently ignored. (This pull request follows our current policy of reporting an unsupported version as an error.)

I do not see any problem with keeping the current policy. It can also be on a case-by-case basis once we encounter issues requiring more leniency.

PR #2190, opened 2024-06-03
read.c: Return avifResult from avifParseSampleTableBox

vigneshvg:
One of the error paths is currently returning `AVIF_FALSE` whereas
the function is supposed to return `avifResult`.

This is important because we may allow parsing of erroneous files.


y-guyon (approved):

wantehchang (commented):
LGTM.

PR #2191, opened 2024-06-03
read.c: Fail if gainmap properties have more than one nclx

vigneshvg:
The current code simply adopts the first nclx property of the
gainmap item. The right behavior seems to be to fail if there are
multiple nclx properties (similar to what we do for the color image
and the tonemap image).

HEIF Section 6.5.5.1:
```
When two ColourInformationBoxes are associated with an image
item, one shall have a colour_type value of 'rICC' or 'prof'
(providing either restricted or unrestricted ICC profiles
respectively) and the other one shall have a colour_type value of
'nclx' with colour_primaries equal to 2 and
transfer_characteristics equal to 2 (2 indicating "unspecified",
since these data are supplied by the ICC profile instead).
```

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -4677,6 +4677,48 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
     return AVIF_RESULT_OK;
 }
 
+// On success, this function returns AVIF_RESULT_OK and does the following:
+// * If a nclx property was found in |properties|:
+//   - Set |*colorPrimaries|, |*transferCharacteristics|, |*matrixCoefficients|
+//     and |*yuvRange|.
+//   - If cicpSet is not NULL, set |*cicpSet| to AVIF_TRUE.
+// This function fails if more than one nclx property is found in |properties|.
+// The output parameters may be populated even in case of failure and must be
+// ignored.
+static avifResult avifReadColorNclxProperty(const avifPropertyArray * properties,
+                                            avifColorPrimaries * colorPrimaries,
+                                            avifTransferCharacteristics * transferCharacteristics,
+                                            avifMatrixCoefficients * matrixCoefficients,
+                                            avifRange * yuvRange,
+                                            avifBool * cicpSet)
+{
+    avifBool colrNCLXSeen = AVIF_FALSE;
+    for (uint32_t propertyIndex = 0; propertyIndex < properties->count; ++propertyIndex) {
+        avifProperty * prop = &properties->prop[propertyIndex];
+        if (!memcmp(prop->type, "colr", 4) && prop->u.colr.hasNCLX) {
+            if (colrNCLXSeen) {
+                return AVIF_RESULT_BMFF_PARSE_FAILED;
+            }
+            colrNCLXSeen = AVIF_TRUE;
+            if (cicpSet != NULL) {
+                *cicpSet = AVIF_TRUE;
+            }
+            *colorPrimaries = prop->u.colr.colorPrimaries;
+            *transferCharacteristics = prop->u.colr.transferCharacteristics;
+            *matrixCoefficients = prop->u.colr.matrixCoefficients;
+            *yuvRange = prop->u.colr.range;
+        }
+    }
+    return AVIF_RESULT_OK;
+}
+
+// On success, this function returns AVIF_RESULT_OK and does the following:
+// * If a colr property was found in |properties|:
+//   - Read the icc data into |icc| from |io|.
+//   - Sets the CICP values as documented in avifReadColorNclxProperty().
+// This function fails if more than one icc or nclx property is found in
+// |properties|. The output parameters may be populated even in case of failure
+// and must be ignored (and the |icc| object may need to be freed).
 static avifResult avifReadColorProperties(avifIO * io,
                                           const avifPropertyArray * properties,
                                           avifRWData * icc,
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -4677,6 +4677,33 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
     return AVIF_RESULT_OK;
 }
 
+static avifResult avifReadColorNclxProperty(const avifPropertyArray * properties,
```
wantehchang, line 4680: We should document what this function does to the output parameters when it returns `AVIF_RESULT_OK`.

Similarly, the `avifReadColorProperties()` function at line 4707 should be documented.

vigneshvg: Done.

src/read.c
```diff
@@ -4677,6 +4677,42 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
     return AVIF_RESULT_OK;
 }
 
+// On success, this function returns AVIF_RESULT_OK and does the following:
+// * If a nclx property was found in |properties|:
```
wantehchang, line 4681: Document under what conditions this functions fails, and that the output parameters may be set when the function fails.

src/read.c
```diff
@@ -4677,6 +4677,42 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+            *yuvRange = prop->u.colr.range;
+        }
+    }
+    return AVIF_RESULT_OK;
+}
+
+// On success, this function returns AVIF_RESULT_OK and does the following:
+// * If a colr property was found in |properties|:
+//   - Read the icc data into |icc| from |io|.
+//   - Sets the CICP values as documented in avifReadColorNclxProperty().
```
wantehchang, line 4715: Document under what conditions this functions fails, and that the output parameters may be set when the function fails. Therefore, it is important to free `icc` even when the function fails.

wantehchang (approved):

src/read.c
vigneshvg: Done. I don't think `icc` needs to be freed in case of failure since we are simply populating pointers. There is no allocation done in this function.

src/read.c
vigneshvg: Done.

src/read.c
```diff
@@ -4677,6 +4677,48 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
     return AVIF_RESULT_OK;
 }
 
+// On success, this function returns AVIF_RESULT_OK and does the following:
+// * If a nclx property was found in |properties|:
+//   - Set |*colorPrimaries|, |*transferCharacteristics|, |*matrixCoefficients|
+//     and |*yuvRange|.
+//   - If cicpSet is not NULL, set |*cicpSet| to AVIF_TRUE.
+// This function fails if more than one nclx property is found in |properties|.
+// The output pointers may be populated even in case of failure and must be
```
wantehchang, line 4686: Optional: output pointers => output parameters ?

src/read.c
```diff
@@ -4677,6 +4677,48 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+    }
+    return AVIF_RESULT_OK;
+}
+
+// On success, this function returns AVIF_RESULT_OK and does the following:
+// * If a colr property was found in |properties|:
+//   - Read the icc data into |icc| from |io|.
+//   - Sets the CICP values as documented in avifReadColorNclxProperty().
+// This function fails if more than one icc or nclx property is found in
+// |properties|. The output pointers may be populated even in case of failure
```
wantehchang, line 4720: Optional: output pointers => output parameters ?

src/read.c
wantehchang: Memory is allocated in the `avifRWDataSet(icc, iccRead.data, iccRead.size)` call. The  `avifRWDataSet()` function is defined as follows:

```
avifResult avifRWDataSet(avifRWData * raw, const uint8_t * data, size_t len)
{
    if (len) {
        AVIF_CHECKRES(avifRWDataRealloc(raw, len));
        memcpy(raw->data, data, len);
    } else {
        avifRWDataFree(raw);
    }
    return AVIF_RESULT_OK;
}
```

wantehchang (approved):

src/read.c
vigneshvg: Done.

src/read.c
```diff
@@ -4688,37 +4730,23 @@ static avifResult avifReadColorProperties(avifIO * io,
...
-                colrICCSeen = AVIF_TRUE;
-                AVIF_CHECKRES(avifRWDataSet(icc, iccRead.data, iccRead.size));
-            }
-            if (prop->u.colr.hasNCLX) {
-                if (colrNCLXSeen) {
-                    return AVIF_RESULT_BMFF_PARSE_FAILED;
+        if (!memcmp(prop->type, "colr", 4) && prop->u.colr.hasICC) {
+            if (colrICCSeen) {
+                if (icc->size > 0) {
+                    avifRWDataFree(icc);
```
wantehchang, line 4740: It is not enogh to free `icc` here, because this function may also return an error from line 4745 or line 4747 (via the `AVIF_CHECKRES` macro).

It is better to just document that the caller needs to free `icc` whether this functions succeeds or fails at line 4721, and then verify that the callers do that. I verified that is the case because `icc` will eventually be freed when we destroy the decoder.

vigneshvg: Done, i have updated the comment and reverted the latest change.

wantehchang (approved):

PR #2195, opened 2024-06-06
Fix grid item progressive propagation logic

vigneshvg:
Set the top level grid item's progressive to be true only if all
the grid items have progressive set to true (and not just the
first one).

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1426,6 +1426,7 @@ static avifResult avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifI
     // Count number of dimg for this item, bail out if it doesn't match perfectly
     unsigned int tilesAvailable = 0;
     avifDecoderItem * firstTileItem = NULL;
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -1504,8 +1502,13 @@ static avifResult avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifI
                                   (const char *)firstTileItem->type);
             return AVIF_RESULT_INVALID_IMAGE_GRID;
         }
+        progressive = progressive && item->progressive;
```
wantehchang, line 1505: Nit/optional: I usualy write this as:

```
        if (!item->progressive) {
            progressive = AVIF_FALSE;
        }
```

But your version makes it clear that we are AND'ing all the `item->progressive` values.

wantehchang (approved):

src/read.c
vigneshvg: Done.

wantehchang (approved):

PR #2197, opened 2024-06-10
Signal gain map transformative properties

y-guyon:


y-guyon:
This PR does not check for properties tied to the tone mapping derived image item, but it could.
Combining chained same-type transformative properties into one should be doable, although probably a hassle to implement.
I suggest returning NOT_IMPLEMENTED when both tmap and input items have transformative properties for now.

y-guyon:
I noticed that `AVIF_TRANSFORM_PASP` is part of `avifImage::transformFlags` but `pasp` is a Descriptive item property according to the HEIF spec,while `clap`, `irot`, and `imir` are transformative properties.
Maybe there is no need to check `pasp` equality in this PR.

include/avif/avif.h
```diff
@@ -799,6 +799,7 @@ typedef struct avifImage
     //
     // To encode any of these boxes, set the values in the associated box, then enable the flag in
     // transformFlags. On decode, only honor the values in boxes with the associated transform flag set.
+    // These also appy to gainMap->image, if any.
```
wantehchang, line 802: Misspelling: appy => apply

Should we add "gainMap->image->transformFlags is always AVIF_TRANSFORM_NONE" to this comment?

src/write.c
```diff
@@ -2684,8 +2684,13 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
         } else if (item->itemCategory == AVIF_ITEM_GAIN_MAP) {
             // Gain map specific properties
 
-            // Write just the colr nclx box
+            // Write the colr nclx box.
             AVIF_CHECKRES(avifEncoderWriteNclxProperty(&dedup->s, s, itemMetadata, &item->ipma, dedup));
+
+            // Also write the transformative properties. They have to match those of the primary item.
+            AVIF_CHECKRES(avifEncoderWriteExtendedColorProperties(&dedup->s, s, imageMetadata, &item->ipma, dedup));
+            // Do not expect them to be explicitly set by the user.
```
wantehchang, line 2692: Nit: expect => allow

wantehchang (commented):
The commit message should have more detail. For example, it should mention the transformative properties of the gain map item have to match those of the primary item.

We should add a comment to document where this requirement comes from. Does this requirement come from a (draft) standard?

src/write.c
```diff
@@ -2684,8 +2684,13 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
...
+            AVIF_CHECKERR(itemMetadata->transformFlags == AVIF_TRANSFORM_NONE, AVIF_RESULT_ENCODE_GAIN_MAP_FAILED);
```
wantehchang, line 2693: Nit: Move this check before line 2691. This requires merging the comment at line 2692 into the comment at line 2690.

wantehchang (approved):

src/write.c
```diff
@@ -2684,8 +2684,43 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
...
+
+            // 'clap' is treated as 'irot'/'imir', although it could differ between the base and
+            // gain map image items if these have different dimensions.
+            if (imageMetadata->transformFlags & AVIF_TRANSFORM_CLAP) {
+                AVIF_CHECKERR(imageMetadata->width != itemMetadata->width || imageMetadata->height != itemMetadata->height,
+                              AVIF_RESULT_NOT_IMPLEMENTED);
+            }
+
+            // 'pasp' is not a transformative property but it is assumed to apply to the gain map
+            // in the same way as the transformative properties above.
```
wantehchang, line 2717: Nit: It would be good to point out that this is because `avifTransformFlags` has a `AVIF_TRANSFORM_PASP` bit flag.

src/read.c
```diff
@@ -5429,6 +5444,40 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         decoder->image->transformFlags |= AVIF_TRANSFORM_IMIR;
         decoder->image->imir = imirProp->u.imir;
     }
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+    if (gainMapProperties != NULL) {
+        // The transformative properties of the gain map item have to match those of the primary item.
```
wantehchang, line 5449: Note: I am sorry this comment is verbose. The first sentence is the key point. Please feel free to address this issue in other ways; it is not necessary to accept my suggestion.

I still think we need to clarify whether this requirement comes from a (draft) standard or libavif imposes this requirement based on the justification in `avifRWStreamWriteProperties()`.

It is fine to refer to a draft standard because the gain map code is guarded by the `AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP` macro.

However, the proposed commit message implies this requirement is imposed by libavif. We can add the proposed commit message (slightly edited) as a comment here:

```
libavif requires the bitstream contain the same pasp, clap, irot, imir
properties for both the base and gain map image items used as input to
the tone-mapped derived image item. libavif also requires the tone-mapped
derived image item itself not be associated with these properties. This is
enforced at encoding. Other patterns are rejected at decoding.
```

This comment could also be added to line 4868.

src/read.c
```diff
@@ -5429,6 +5444,40 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
+        irotProp = avifPropertyArrayFind(gainMapProperties, "irot");
+        AVIF_CHECKERR(!irotProp == !(decoder->image->transformFlags & AVIF_TRANSFORM_IROT), AVIF_RESULT_DECODE_GAIN_MAP_FAILED);
+        if (irotProp) {
+            AVIF_CHECKERR(irotProp->u.irot.angle == decoder->image->irot.angle, AVIF_RESULT_DECODE_GAIN_MAP_FAILED);
+        }
+        imirProp = avifPropertyArrayFind(gainMapProperties, "imir");
+        AVIF_CHECKERR(!imirProp == !(decoder->image->transformFlags & AVIF_TRANSFORM_IMIR), AVIF_RESULT_DECODE_GAIN_MAP_FAILED);
+        if (imirProp) {
+            AVIF_CHECKERR(imirProp->u.imir.axis == decoder->image->imir.axis, AVIF_RESULT_DECODE_GAIN_MAP_FAILED);
+        }
```
wantehchang, line 5478: Optional: It would be good to log a diagnostic message in `decoder->diag` if these checks fail. One way to do this and still use the `AVIF_CHECKERR` macro is to move these checks to a helper function and to call the helper function here and log a diagnostic message if the helper function fails.

vigneshvg (approved):

src/write.c
```diff
@@ -2684,8 +2684,43 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
...
+            // - Associate 'irot'/'imir' with the 'tmap' derived image item only.
+            //   If so, decoding only the base image would give a different orientation than
+            //   decoding the tone-mapped image.
+            // - Wrap the base image in an 'iden' derived image item and associate 'irot'/'imir'
+            //   with the 'tmap' and 'iden' derived image items. 'iden' is not currently supported
+            //   by libavif, reducing the backward compatibility of this solution.
+            // - Associate 'irot'/'imir' with the base and gain map image items.
+            //   These transformative properties are supposed to be applied at decoding on
+            //   image items before these are used as input to a derived image item.
+            //   This is the implemented solution in libavif.
```
wantehchang, line 2704: Add: "libavif imposes this requirement at decoding", or "libavif only accepts this pattern at decoding", or a similar comment.

src/write.c
```diff
@@ -2684,8 +2684,43 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
...
+            // For the orientation, it could be done in multiple ways:
+            // - Bake the orientation in the base and gain map images.
+            //   This does not allow for orientation changes without recompression.
+            // - Associate 'irot'/'imir' with the 'tmap' derived image item only.
+            //   If so, decoding only the base image would give a different orientation than
+            //   decoding the tone-mapped image.
+            // - Wrap the base image in an 'iden' derived image item and associate 'irot'/'imir'
+            //   with the 'tmap' and 'iden' derived image items. 'iden' is not currently supported
+            //   by libavif, reducing the backward compatibility of this solution.
+            // - Associate 'irot'/'imir' with the base and gain map image items.
```
wantehchang, line 2701: Add or clarify that we do not associate 'irot'/'imir' with the 'tmap' derived image item.

wantehchang (approved):
Yannis: The proposed commit message looks good. Thanks! Most of my comments are about clarifying the source of the checks we perform at decoding.

src/read.c
```diff
@@ -4895,6 +4895,43 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
...
+        avifDiagnosticsPrintf(&decoder->diag, "Rotation property mismatch between input items of tone-mapping derived image item");
+        return AVIF_RESULT_DECODE_GAIN_MAP_FAILED;
+    }
+    const avifProperty * imirProp = avifPropertyArrayFind(gainMapProperties, "imir");
+    if (!imirProp != !(image->transformFlags & AVIF_TRANSFORM_IMIR) || (imirProp && imirProp->u.imir.axis != image->imir.axis)) {
+        avifDiagnosticsPrintf(&decoder->diag, "Mirroring property mismatch between input items of tone-mapping derived image item");
+        return AVIF_RESULT_DECODE_GAIN_MAP_FAILED;
+    }
+    return AVIF_RESULT_OK;
+}
```
wantehchang, line 4934: Optional: Your version has more precise diagnostic messages. The following is what I meant, which preserves the use of the `AVIF_CHECKERR` macro in your original code:

```
static avifResult aviDecoderCheckGainMapProperties(avifDecoder * decoder, const avifPropertyArray * gainMapProperties)
{
    const avifImage * image = decoder->image;
    // ***** NOTE: COPIED VERBATIM FROM THE ORIGINAL CODE *****
        // libavif requires the bitstream contain the same pasp, clap, irot, imir
        // ...
        paspProp = avifPropertyArrayFind(gainMapProperties, "pasp");
        AVIF_CHECKERR(!paspProp == !(decoder->image->transformFlags & AVIF_TRANSFORM_PASP), AVIF_RESULT_DECODE_GAIN_MAP_FAILED);
        if (paspProp) {
            AVIF_CHECKERR(paspProp->u.pasp.hSpacing == decoder->image->pasp.hSpacing, AVIF_RESULT_DECODE_GAIN_MAP_FAILED);
            AVIF_CHECKERR(paspProp->u.pasp.vSpacing == decoder->image->pasp.vSpacing, AVIF_RESULT_DECODE_GAIN_MAP_FAILED);
        }
        ...
        imirProp = avifPropertyArrayFind(gainMapProperties, "imir");
        AVIF_CHECKERR(!imirProp == !(decoder->image->transformFlags & AVIF_TRANSFORM_IMIR), AVIF_RESULT_DECODE_GAIN_MAP_FAILED);
        if (imirProp) {
            AVIF_CHECKERR(imirProp->u.imir.axis == decoder->image->imir.axis, AVIF_RESULT_DECODE_GAIN_MAP_FAILED);
        }
    return AVIF_RESULT_OK;
}
```

And then the caller at line 5489 does the following:

```
        avifResult result = aviDecoderCheckGainMapProperties(decoder, gainMapProperties);
            avifDiagnosticsPrintf(&decoder->diag, "Transformative property mismatch between input items of tone-mapping derived image item");
            return result;
        }
```

We could replace "Transformative" with "pasp, clap, irot, or imir" in the diagnostic message.

wantehchang (approved):
LGTM. Thanks!

src/read.c
y-guyon: I prefer to keep `avifDiagnosticsPrintf()` close to the source of the `avifResult`.

PR #2198, opened 2024-06-10
Use avifDecoderItemShouldBeSkipped when harvesting ispe

vigneshvg:
This function has the same conditions as the ones that are
explicitly re-written with one extra condition (thumbnailForID
!= 0). This extra condition is okay to be checked too in this
case since we anyway do not process such items.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -4350,6 +4350,17 @@ static avifResult avifDecoderPrepareSample(avifDecoder * decoder, avifDecodeSamp
     return AVIF_RESULT_OK;
 }
 
+// Returns AVIF_TRUE if the item should be skipped. Items should be skipped for one of the following reasons:
+//  * Size is 0.
+//  * Has an essential property that isn't supported by libavif.
+//  * Item is not a single image or a grid.
+//  * Item is a thumbnail.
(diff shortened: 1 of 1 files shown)


wantehchang (approved):

wantehchang:
@y-guyon Please take a look if you have time.

@vigneshvg You can merge this pull request first. If Yannis suggests any changes, we can address them after the merge.

src/read.c
```diff
@@ -4350,6 +4350,17 @@ static avifResult avifDecoderPrepareSample(avifDecoder * decoder, avifDecodeSamp
     return AVIF_RESULT_OK;
 }
 
+// Returns AVIF_TRUE if the item should be skipped. Items should be skipped for one of the following reasons:
+//  * Size is 0.
+//  * Has an essential property that isn't supported by libavif.
+//  * Item is not a single image or a grid.
```
y-guyon, line 4356: Should `tmap` be included too?  
If so, `avifDecoderItemShouldBeSkipped()` could also be used in `avifDecoderDataFindToneMappedImageItem()`.

y-guyon (approved):

src/read.c
vigneshvg: sure, let me look into this in a follow up.

PR #2200, opened 2024-06-11
Invoke ninja with -C option

wantehchang:
Make ninja the last command in the ext/*.cmd script/batch file, so that
the failure of the ninja command makes the script/batch file fail.
    
The only exception is ext/svt.cmd. It needs special handling.
    
Also verified all the ext/*.sh scripts have "set -e" as the first
command.


wantehchang:
Vincent: You pointed out this issue in https://github.com/AOMediaCodec/libavif/pull/2194#issuecomment-2160698961:

> libaom is built from ext and it fails (the CI should stop here ...)

As https://github.com/AOMediaCodec/libavif/actions/runs/9469382119/job/26087916277?pr=2200 demonstrates, with this pull request the "Build aom" step failed because the ninja command failed.

vrabaud:
We might as well have this behavior for all .cmd/.sh in ./ext right?

wantehchang:
Yes, I will make the same change to the other .cmd/.sh in ./ext and re-request your review.

wantehchang:
@vrabaud Please review. Note the updated commit message.

@jzern I am not sure how to make ext/svt.cmd fail after any command fails. ext/svt.cmd uses a special trick to execute ext/svt.sh instead when it detect that it runs on Unix, so we can probably treat it as a pure DOS batch file.

jzern:
> @vrabaud Please review. Note the updated commit message.
> 
> @jzern I am not sure how to make ext/svt.cmd fail after any command fails. ext/svt.cmd uses a special trick to execute ext/svt.sh instead when it detect that it runs on Unix, so we can probably treat it as a pure DOS batch file.

`if ERRORLEVEL 1 exit /b 1` after each command you think can fail should work.

wantehchang:
James: Thank you for the tip. I will not deal with ext/svt.cmd in this pull request. Please review the changes to the other ext/*.cmd files in this pull request. I have tested them on both Linux and Windows.

jzern (approved):
The whole mkdir/cd/cmake/cd could probably be simplified to `cmake -S ... -B ...` in another commit.

PR #2201, opened 2024-06-11
Add an "Upgrade clang" step

wantehchang:
In the current windows-latest (windows-2022) image, the installed Clang is version 16, but the Visual Studio STL requires Clang version 17 or later. The "Upgrade clang" workaround is recommended in https://github.com/actions/runner-images/issues/10001#issuecomment-2153764664.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -45,6 +45,8 @@ jobs:
         codec-dav1d: 'LOCAL'
         codec-rav1e: 'LOCAL'
 
+    - name: Upgrade clang
+      run: choco upgrade llvm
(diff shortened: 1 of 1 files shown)


vrabaud:
No need for that if you remove it from the path in https://github.com/AOMediaCodec/libavif/pull/2202 right ?

wantehchang:
Yes, we only need either this pull request or https://github.com/AOMediaCodec/libavif/pull/2202. This pull request is simpler, but the upgrade takes a long time. So I also prefer https://github.com/AOMediaCodec/libavif/pull/2202.

wantehchang:
I abandoned this pull request.

PR #2202, opened 2024-06-11
Remove standalone clang from PATH

wantehchang:
Use the clang in Visual Studio. Recommended in
https://github.com/actions/runner-images/issues/10001#issuecomment-2150541007.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -48,7 +48,10 @@ jobs:
     - name: Build aom
       if: steps.setup.outputs.ext-cache-hit != 'true'
       working-directory: ./ext
-      run: ./aom.cmd
+      run: |
+        # Remove standalone clang from PATH.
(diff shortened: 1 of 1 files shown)


wantehchang:
@vrabaud Vincent: I can't reproduce the avifcodectest failure on my Windows laptop. I use the clang-cl in Visual Studio, and this pull request emulates my setup. The only difference is that I disable the rav1e codec (because I don't have Rust installed).

vrabaud (approved):

wantehchang:
I abandoned this pull request. After 02816a353713fd1a5bbebf4d5818d33822bb42ae, we are not compiling libaom with clang-cl. Although we are still compiling libyuv with clang-cl, libyuv doesn't use the C++ STL. (libyuv is essentially C code compiled with C++ compilers.) So we no longer need this workaround.

Although it is arguably safer to use the clang bundled with Visual Studio, I think it is good to test libavif with a standalone installation of clang.

PR #2203, opened 2024-06-11
Work around a bug in the windows-latest os image

wantehchang:
The bug causes access violation in avifcodectest when libgav1 is used. The error message is:
  error: SEH exception with code 0xc0000005 thrown in the test body.


wantehchang:
@vrabaud Vincent: FYI.

vigneshvg (approved):

PR #2204, opened 2024-06-11
Do not compile libaom with clang-cl on Windows

wantehchang:
The Visual Studio 2022 bug has been fixed. Remove the workaround added in commit 877a2742671849dd393f9b54cd39328c63c6ee4e.


jzern (approved):
We have nightly tests running using the Microsoft compiler which pass, so this should be safe.

PR #2206, opened 2024-06-12
Revert workaround for bug in windows-latest image

wantehchang:
The bug has been fixed in the windows-2022 runner image version 20240610.1.0.


wantehchang:
https://github.com/AOMediaCodec/libavif/actions/runs/9485469475/job/26137474856?pr=2206 shows the windows-2022 runner image is the new version 20240610.1.0, and the version of clang-cl used to compile libyuv is "Clang 18.1.6 with MSVC-like command-line", which is newer than the version 17 required by the MSVC STL.

vigneshvg (approved):

Issue #2207, opened 2024-06-14
oss-fuzz is pinned at an old builder


jzern:
This was done as a part of an [infra update](https://github.com/google/oss-fuzz/commit/a2c60af93357f2053a45816fcdefdfc02206b64f).

From the [Dockerfile](https://github.com/google/oss-fuzz/blob/a2c60af93357f2053a45816fcdefdfc02206b64f/projects/libavif/Dockerfile#L17):

```
FROM gcr.io/oss-fuzz-base/base-builder@sha256:19782f7fe8092843368894dbc471ce9b30dd6a2813946071a36e8b05f5b1e27e
# ! This project was pinned after a clang bump. Please remove the pin, Try to fix any build warnings and errors, as well as runtime errors
```

The link errors are related to `LLVMFuzzerRunDriver`:

```
FAILED: tests/avif_fuzztest_dec_incr 
...
/usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x25
/src/libavif/ext/fuzztest/build.libavif/fuzztest/libfuzztest_compatibility_mode.a(compatibility_mode.cc.o): in function `fuzztest::internal::FuzzTestExternalEngineAdaptor::RunInFuzzingMode(int*, char***, fuzztest::internal::Configuration const&)':
compatibility_mode.cc:(.text+0x118f): undefined reference to `LLVMFuzzerRunDriver'
```

I think the first one is just noise, `-gdwarf-4` would probably silence it. The difference is that `libclang_rt.fuzzer_no_main-x86_64.a` isn't being supplied with the build using clang 18. The difference in command line options is below.

```
$ diff /tmp/old.txt /tmp/new.txt 
0a1
> 
6a8,13
> -Wno-error=enum-constexpr-conversion
> -Wno-error=incompatible-function-pointer-types
> -Wno-error=int-conversion
> -Wno-error=deprecated-declarations
> -Wno-error=implicit-function-declaration
> -Wno-error=implicit-int
60d66
< /usr/local/lib/clang/15.0.0/lib/linux/libclang_rt.fuzzer_no_main-x86_64.a
```

vrabaud:
The issue is that libclang_rt.fuzzer_no_main-x86_64.a is now named libclang_rt.fuzzer_no_main.a
fuzztest should soon have a bug fix.

PR #2209, opened 2024-06-19
Bump fuzztest

vrabaud:
This seems to fix the oss-fuzz build pinned at an old builder: https://github.com/AOMediaCodec/libavif/issues/2207


y-guyon (approved):

vrabaud:
Ignoring the CI breakages as this PR is only for fuzz tests, and the corresponding CI passes.

PR #2212, opened 2024-06-19
Fix Python conflict with homebrew

vrabaud:
- delete the Python symlinks before reinstalling Python through meson
- install meson using the proper package managers (as pip now requires running in a virtualenc on macOS) 
- use the setup-linux action on Android (because meson installation is now moved to setup-linux)

### DIFF
--- a/.github/actions/setup-macos/action.yml
+++ b/.github/actions/setup-macos/action.yml
@@ -19,6 +19,14 @@ outputs:
 runs:
   using: "composite"
   steps:
+    # github actions overwrites brew's python. Force it to reassert itself, by running in a separate step.
+    # Borrowed from https://github.com/mesonbuild/meson/blob/master/.github/workflows/macos.yml#L87-L92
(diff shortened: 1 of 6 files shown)


y-guyon (approved):

.github/actions/setup-macos/action.yml
```diff
@@ -19,6 +19,17 @@ outputs:
 runs:
   using: "composite"
   steps:
+    # github actions overwrites brew's python. Force it to reassert itself, by running in a separate step.
+    # Borrowed from https://github.com/mesonbuild/meson/blob/master/.github/workflows/macos.yml#L87-L92
```
jzern, line 23: `blob/master` will change. It would be better to add a stable commit with links like this.

vrabaud: Fixed in https://github.com/AOMediaCodec/libavif/pull/2224

PR #2215, opened 2024-06-20
Add sato BIT_DEPTH_EXTENSION_12B_8B_OVERLAP_4B

y-guyon:
Sample Transform derived image item support for residual bit depth extension.

### DIFF
--- a/src/sampletransform.c
+++ b/src/sampletransform.c
@@ -137,6 +137,33 @@ avifResult avifSampleTransformRecipeToExpression(avifSampleTransformRecipe recip
         return AVIF_RESULT_OK;
     }
 
+    if (recipe == AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_8B_OVERLAP_4B) {
+        // reference_count is two: one 12-bit input image and one 8-bit input image.
+        //   (base_sample << 4) + hidden_sample
+        // Note: Both base_sample and hidden_sample are encoded lossily or losslessly. hidden_sample overlap
+        //       with base_sample by 4 bits to alleviate the loss caused by the quantization of base_sample.
+        AVIF_CHECKERR(avifArrayCreate(expression, sizeof(avifSampleTransformToken), 7), AVIF_RESULT_OUT_OF_MEMORY);
+
+        // The base image represents the 12 most significant bits of the reconstructed, bit-depth-extended output image.
+        // Left shift the base image (which is also the primary item, or the auxiliary alpha item of the primary item)
+        // by 4 bits. This is equivalent to multiplying by 2^4.
+        AVIF_ASSERT_OR_RETURN(avifPushConstant(expression, 16));
+        AVIF_ASSERT_OR_RETURN(avifPushInputImageItem(expression, 1));
+        AVIF_ASSERT_OR_RETURN(avifPushOperator(expression, AVIF_SAMPLE_TRANSFORM_PRODUCT));
+
+        // The second image represents the offset to apply to the shifted base image to retrieve
+        // the original image, with some loss due to quantization.
+        AVIF_ASSERT_OR_RETURN(avifPushInputImageItem(expression, 2));
+        AVIF_ASSERT_OR_RETURN(avifPushOperator(expression, AVIF_SAMPLE_TRANSFORM_SUM));
+
+        // The second image is offset by 128 to have unsigned values to encode.
+        // Correct that last to always work with unsigned values in the operations above.
+        AVIF_ASSERT_OR_RETURN(avifPushConstant(expression, 128));
+        AVIF_ASSERT_OR_RETURN(avifPushOperator(expression, AVIF_SAMPLE_TRANSFORM_DIFFERENCE));
+        // Sample values are clamped to [0:1<<depth[ at that point.
+        return AVIF_RESULT_OK;
+    }
+
     return AVIF_RESULT_INVALID_ARGUMENT;
 }
 
@@ -144,7 +171,8 @@ avifResult avifSampleTransformExpressionToRecipe(const avifSampleTransformExpres
 {
     *recipe = AVIF_SAMPLE_TRANSFORM_NONE;
     const avifSampleTransformRecipe kAllRecipes[] = { AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_8B_8B,
-                                                      AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_4B };
+                                                      AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_4B,
+                                                      AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_8B_OVERLAP_4B };
     for (size_t i = 0; i < sizeof(kAllRecipes) / sizeof(kAllRecipes[0]); ++i) {
         avifSampleTransformRecipe candidateRecipe = kAllRecipes[i];
         avifSampleTransformExpression candidateExpression = { 0 };
--- a/src/sampletransform.c
+++ b/src/sampletransform.c
@@ -140,7 +140,7 @@ avifResult avifSampleTransformRecipeToExpression(avifSampleTransformRecipe recip
     if (recipe == AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_8B_OVERLAP_4B) {
         // reference_count is two: one 12-bit input image and one 8-bit input image.
         //   (base_sample << 4) + hidden_sample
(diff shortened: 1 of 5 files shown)


src/write.c
```diff
@@ -1325,6 +1326,48 @@ static avifResult avifImageCreateAllocate(avifImage ** sampleTransformedImage, c
     return avifImageAllocatePlanes(*sampleTransformedImage, planes);
 }
 
+static avifResult avifEncoderDecodeSatoBaseImage(const avifEncoder * encoder,
+                                                 const avifImage * original,
+                                                 uint32_t numBits,
+                                                 avifPlanesFlag planes,
+                                                 avifCodec ** codec,
```
wantehchang, line 1333: It is better to make `codec` a local variable in this function instead of an in/out parameter. This function needs to call `avifCodecDestroy(codec)` whether it succeeds or fails. This means we won't be able to use the `AVIF_CHECKERR` macro after `codec` has been created and before it is destroyed.

src/write.c
```diff
@@ -1325,6 +1326,48 @@ static avifResult avifImageCreateAllocate(avifImage ** sampleTransformedImage, c
...
+        AVIF_ASSERT_OR_RETURN(item->encodeOutput != NULL); // TODO: Support grids?
+        AVIF_ASSERT_OR_RETURN(item->encodeOutput->samples.count == 1);
+        AVIF_ASSERT_OR_RETURN(item->encodeOutput->samples.sample[0].data.size != 0);
+        AVIF_ASSERT_OR_RETURN(sample.data.size == 0); // There should be only one base item.
+        sample.data.data = item->encodeOutput->samples.sample[0].data.data;
+        sample.data.size = item->encodeOutput->samples.sample[0].data.size;
+    }
+    AVIF_ASSERT_OR_RETURN(sample.data.size != 0); // There should be at least one base item.
+
+    // Used as an argument to avifCodecGetNextImageFunc only for a few fields.
```
wantehchang, line 1357: [This is just a comment, or rather a note to myself. No change requested.]

We need to think about how to make it easier and safer to do this.

I think we can add `maxThreads` and `imageSizeLimit` fields to the `avifCodec` struct. And after `codec` is created, the owner can set those two fields. Then, the `getNextImage` method does not need the `decoder` input parameter.

src/write.c
```diff
@@ -1325,6 +1326,48 @@ static avifResult avifImageCreateAllocate(avifImage ** sampleTransformedImage, c
...
+    avifDecoder decoder;
+    memset(&decoder, 0, sizeof(decoder));
+    decoder.maxThreads = encoder->maxThreads;
+    decoder.imageSizeLimit = AVIF_DEFAULT_IMAGE_SIZE_LIMIT;
+
+    AVIF_CHECKRES(avifCodecCreate(AVIF_CODEC_CHOICE_AUTO, AVIF_CODEC_FLAG_CAN_DECODE, codec));
```
wantehchang, line 1363: We need to set `codec->diag`. We can use a local variable, or use `encoder->diag`.

Note: The current `getNextImage` methods don't seem to use `codec->diag`, so this is mainly to be future-proof.

tests/gtest/avif16bittest.cc
```diff
@@ -115,6 +119,16 @@ INSTANTIATE_TEST_SUITE_P(
         /*quality=*/
         testing::Values(AVIF_QUALITY_LOSSLESS)));
 
+INSTANTIATE_TEST_SUITE_P(
+    ResidualBitDepthExtension, SampleTransformTest,
```
wantehchang, line 123: Nit: Add "s" to "ResidualBitDepthExtension" to make it plural? See "BitDepthExtensions" at line 113.

wantehchang (commented):
Yannis: I have verified the pull request is essentially a no-op if the Sample Transforms experimental feature is disabled. I have only skimmed through the code in this round.

src/write.c
y-guyon: That was the point: to be able to use `AVIF_CHECKRES()` within `avifEncoderDecodeSatoBaseImage()` and just having to handle `avifImageDestroy(decodedBaseImage)` and `avifCodecDestroy(codec)` in a single location, at its call site.

I added a comment to `avifEncoderDecodeSatoBaseImage()`.

wantehchang: In C code this kind of resource cleanup is best handled by `goto` statements. (This is an acceptable use of `goto`.)

I understand the debugger break in the `AVIF_CHECKRES()` macro is helpful, but trying to use `AVIF_CHECKRES()` in this function leaks the implementation detail to the caller. This is also why `(*codec)->diag` cannot point to a local variable, and we had to remove `const` from the `avifEncoder * encoder` parameters of three functions in the call chain.

src/write.c
```diff
@@ -1325,7 +1326,52 @@ static avifResult avifImageCreateAllocate(avifImage ** sampleTransformedImage, c
     return avifImageAllocatePlanes(*sampleTransformedImage, planes);
 }
 
-static avifResult avifEncoderCreateSatoImage(const avifEncoder * encoder,
+// Finds the encoded base image and decodes it.
+// Callers of this function must free *codec and *decodedBaseImage if not null, in case of success or not.
```
wantehchang, line 1330: Nit: Change "in case of success or not" to "whether the function succeeds or not"

src/sampletransform.c
```diff
@@ -137,14 +137,42 @@ avifResult avifSampleTransformRecipeToExpression(avifSampleTransformRecipe recip
         return AVIF_RESULT_OK;
     }
 
+    if (recipe == AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_8B_OVERLAP_4B) {
+        // reference_count is two: one 12-bit input image and one 8-bit input image.
+        //   (base_sample << 4) + hidden_sample
+        // Note: Both base_sample and hidden_sample are encoded lossily or losslessly. hidden_sample overlap
```
wantehchang, line 143: Nit: overlap => overlaps

src/write.c
```diff
@@ -1325,7 +1326,52 @@ static avifResult avifImageCreateAllocate(avifImage ** sampleTransformedImage, c
...
+        AVIF_ASSERT_OR_RETURN(item->encodeOutput != NULL); // TODO: Support grids?
+        AVIF_ASSERT_OR_RETURN(item->encodeOutput->samples.count == 1);
+        AVIF_ASSERT_OR_RETURN(item->encodeOutput->samples.sample[0].data.size != 0);
+        AVIF_ASSERT_OR_RETURN(sample.data.size == 0); // There should be only one base item.
+        sample.data.data = item->encodeOutput->samples.sample[0].data.data;
+        sample.data.size = item->encodeOutput->samples.sample[0].data.size;
+    }
+    AVIF_ASSERT_OR_RETURN(sample.data.size != 0); // There should be at least one base item.
+
+    // Used as an argument to avifCodecGetNextImageFunc only for a few fields.
```
wantehchang, line 1359: Nit: Change this comment to something like "avifCodecGetNextImageFunc uses only a few fields of its decoder argument."?

src/write.c
```diff
@@ -1376,11 +1421,48 @@ static avifResult avifEncoderCreateSatoImage(const avifEncoder * encoder,
...
+            AVIF_CHECKRES(avifImageApplyImgOpConst(*sampleTransformedImage, image, AVIF_SAMPLE_TRANSFORM_DIVIDE, 16, planes));
+        } else {
+            AVIF_CHECKRES(avifImageCreateAllocate(sampleTransformedImage, image, 8, planes));
+            avifCodec * codec = NULL;
+            avifImage * decodedBaseImage = NULL;
+            avifResult result = avifEncoderDecodeSatoBaseImage(encoder, image, 12, planes, &codec, &decodedBaseImage);
+            if (result == AVIF_RESULT_OK) {
+                // decoded = main*16+hidden-128 so hidden = clamp_8b(original-main*16+128). Postfix notation.
+                const avifSampleTransformToken tokens[] = { { AVIF_SAMPLE_TRANSFORM_INPUT_IMAGE_ITEM_INDEX, 0, /*inputImageItemIndex=*/1 },
+                                                            { AVIF_SAMPLE_TRANSFORM_INPUT_IMAGE_ITEM_INDEX, 0, /*inputImageItemIndex=*/2 },
```
wantehchang, line 1438: Just wanted to check my understanding: is image item index 1 the `original` image and image item index 2 the `main` image in the comment at line 1436??

Are these indexes for the `inputImageItems` array at line 1444?

wantehchang (approved):
LGTM. Thanks!

src/write.c
wantehchang: I just realized that for better error reporting we should set `(*codec)->diag` to `&encoder->diag`, so we have to remove `const` from the `avifEncoder * encoder` parameters of those three functions in any case.

src/write.c
y-guyon: Yes, `image` is the "original" image with index 1 and `decodedBaseImage` is the "main" image with index 2.

PR #2217, opened 2024-06-20
Use pip to install meson on windows, not vcpkg

y-guyon:


vrabaud (approved):

y-guyon:
It does not make the CI Windows workflow pass but at least it seems to fix the meson part.

PR #2220, opened 2024-06-21
Upgrade to Corrosion v0.4.10

wantehchang:
We need the bug fix https://github.com/corrosion-rs/corrosion/pull/511.

### DIFF
--- a/cmake/Modules/LocalRav1e.cmake
+++ b/cmake/Modules/LocalRav1e.cmake
@@ -1,5 +1,5 @@
 set(AVIF_LOCAL_RAV1E_GIT_TAG v0.7.1)
-set(AVIF_LOCAL_CORROSION_GIT_TAG v0.4.4)
+set(AVIF_LOCAL_CORROSION_GIT_TAG v0.4.10)
 set(AVIF_LOCAL_CARGOC_GIT_TAG v0.9.27)
 
 set(RAV1E_LIB_FILENAME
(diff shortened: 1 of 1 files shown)


wantehchang:
@fdintino Please take a look. Thanks!

cmake/Modules/LocalRav1e.cmake
```diff
@@ -97,19 +97,10 @@ else()
...
-        list(TRANSFORM RAV1E_LIBRARIES REPLACE "^/" "-")
-        # Remove msvcrt from RAV1E_LIBRARIES since it's linked by default
-        list(REMOVE_ITEM RAV1E_LIBRARIES "msvcrt.lib" "-lmsvcrt")
-    endif()
-
     add_library(rav1e::rav1e STATIC IMPORTED)
     add_dependencies(rav1e::rav1e rav1e)
-    target_link_libraries(rav1e::rav1e INTERFACE "${RAV1E_LIBRARIES}")
+    target_link_libraries(rav1e::rav1e INTERFACE "${Rust_CARGO_TARGET_LINK_NATIVE_LIBS}")
+    target_link_options(rav1e::rav1e INTERFACE "${Rust_CARGO_TARGET_LINK_OPTIONS}")
```
wantehchang, line 103: I am not sure if we should use `INTERFACE` or `PRIVATE` here.

vrabaud: INTERFACE is the way to go because we are creating a library from scratch. Setting PRIVATE dependencies would not change anything as it is a shell library. INTERFACE ensures anybody depending on rav1e:rav1e will also use the dep.

vrabaud:
Nice finding and solution !

vrabaud (approved):

jzern (approved):

cmake/Modules/LocalRav1e.cmake
wantehchang: Since `Rust_CARGO_TARGET_LINK_OPTIONS` is an empty list on non-Windows platforms, I wonder if we should guard the `target_link_options()` call in an if statement:

```
    if (Rust_CARGO_TARGET_LINK_OPTIONS)
        target_link_options(rav1e::rav1e INTERFACE "${Rust_CARGO_TARGET_LINK_OPTIONS}")
    endif()
```

I am also not sure if we need to quote `${Rust_CARGO_TARGET_LINK_NATIVE_LIBS}` and `${Rust_CARGO_TARGET_LINK_OPTIONS}` in the `target_link_libraries()` and `target_link_options()` calls.

cmake/Modules/LocalRav1e.cmake
```diff
@@ -1,5 +1,5 @@
 set(AVIF_LOCAL_RAV1E_GIT_TAG v0.7.1)
-set(AVIF_LOCAL_CORROSION_GIT_TAG v0.4.4)
+set(AVIF_LOCAL_CORROSION_GIT_TAG v0.4.10)
```
wantehchang, line 2: Note that I know Corrosion v0.5.0 is available. The reason I did not upgrade to v0.5.0 in this pull request is that I didn't spend the time understanding the breaking change mentioned in the release notes. We can upgrade to v0.5.0 in a follow-up pull request.

wantehchang (commented):
Thank you both for the review!

cmake/Modules/LocalRav1e.cmake
jzern: > Since `Rust_CARGO_TARGET_LINK_OPTIONS` is an empty list on non-Windows platforms, I wonder if we should guard the `target_link_options()` call in an if statement:
> 
> ```
>     if (Rust_CARGO_TARGET_LINK_OPTIONS)
>         target_link_options(rav1e::rav1e INTERFACE "${Rust_CARGO_TARGET_LINK_OPTIONS}")
>     endif()
> ```
> 

That seems fine.

> I am also not sure if we need to quote `${Rust_CARGO_TARGET_LINK_NATIVE_LIBS}` and `${Rust_CARGO_TARGET_LINK_OPTIONS}` in the `target_link_libraries()` and `target_link_options()` calls.

Probably not given it would be a single item or ';' separated list. Though I think quoting is usually safest.

wantehchang: Thanks for the reply. I did some experiments and found that both of the following are harmless:

```
add_library(wtc wtc.c)
target_link_options(wtc INTERFACE)
target_link_options(wtc INTERFACE "")
```

So we can just call `target_link_options()` unconditionally here.

PR #2222, opened 2024-06-21
Upgrade to Corrosion v0.5.0

wantehchang:


cmake/Modules/LocalRav1e.cmake
```diff
@@ -1,5 +1,5 @@
 set(AVIF_LOCAL_RAV1E_GIT_TAG v0.7.1)
-set(AVIF_LOCAL_CORROSION_GIT_TAG v0.4.10)
+set(AVIF_LOCAL_CORROSION_GIT_TAG v0.5.0)
```
wantehchang, line 2: I tried updating cargo-c to the latest v0.10.0 release, but there was a build failure on Windows. So I removed that change from this pull request.

jzern (approved):

PR #2223, opened 2024-06-21
Upgrade to cargo-c v0.9.32

wantehchang:


cmake/Modules/LocalRav1e.cmake
```diff
@@ -1,6 +1,6 @@
 set(AVIF_LOCAL_RAV1E_GIT_TAG v0.7.1)
 set(AVIF_LOCAL_CORROSION_GIT_TAG v0.5.0)
-set(AVIF_LOCAL_CARGOC_GIT_TAG v0.9.27)
+set(AVIF_LOCAL_CARGOC_GIT_TAG v0.9.32)
```
wantehchang, line 3: This is the highest version of cargo-c we can upgrade to without a build failure (on Windows).

wantehchang: This is the build error message on Windows:

```
  Installing debugging information
Error: CliError { error: Some(failed to copy `D:\a\libavif\libavif\build\_deps\rav1e-src\target\x86_64-pc-windows-msvc\release\rav1e.pdb` to `D:/a/libavif/libavif/build/ext/rav1e\usr\bin\rav1e.pdb`

Caused by:
    The system cannot find the file specified. (os error 2)), exit_code: 101 }
```

It may be related to the second bullet point in the cargo-c v0.10.0 release notes:
https://github.com/lu-zero/cargo-c/releases/tag/v0.10.0

vrabaud: Maybe add a comment saying why we do not bump further (we did that for libpng)

wantehchang: Done in commit 2.

PR #2226, opened 2024-06-25
Remove gainmap backward direction

ccameron-chromium:
This did not end up in the ISO 21496-1 specification.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -630,16 +630,13 @@ typedef struct avifGainMapMetadata
     //
     // If 'H' is the display's current log2-encoded HDR capacity (HDR to SDR ratio),
     // then the weight 'w' to apply the gain map is computed as follows:
-    // f = clamp((H - hdrCapacityMin) /
-    //           (hdrCapacityMax - hdrCapacityMin), 0, 1);
-    // w = backwardDirection ? f * -1 : f;
+    // f = clamp((H - baseHdrHeadroom) /
+    //           (alternateHdrHeadroom - baseHdrHeadroom), 0, 1);
+    // w = sign(alternateHdrHeadroom - baseHdrHeadroom) * f
     uint32_t baseHdrHeadroomN;
     uint32_t baseHdrHeadroomD;
     uint32_t alternateHdrHeadroomN;
     uint32_t alternateHdrHeadroomD;
-
-    // True if the gain map should be applied in reverse, see weight formula above.
-    avifBool backwardDirection;
 
     // True if tone mapping should be performed in the color space of the
     // base image. If false, the color space of the alternate image should
@@ -702,7 +699,6 @@ typedef struct avifGainMapMetadataDouble
     double alternateOffset[3];
     double baseHdrHeadroom;
     double alternateHdrHeadroom;
-    avifBool backwardDirection;
     avifBool useBaseColorSpace;
 } avifGainMapMetadataDouble;
 
--- a/src/gainmap.c
+++ b/src/gainmap.c
@@ -22,7 +22,6 @@ avifBool avifGainMapMetadataDoubleToFractions(avifGainMapMetadata * dst, const a
     }
     AVIF_CHECK(avifDoubleToUnsignedFraction(src->baseHdrHeadroom, &dst->baseHdrHeadroomN, &dst->baseHdrHeadroomD));
     AVIF_CHECK(avifDoubleToUnsignedFraction(src->alternateHdrHeadroom, &dst->alternateHdrHeadroomN, &dst->alternateHdrHeadroomD));
-    dst->backwardDirection = src->backwardDirection;
     dst->useBaseColorSpace = src->useBaseColorSpace;
     return AVIF_TRUE;
 }
@@ -50,7 +49,6 @@ avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, c
     }
     dst->baseHdrHeadroom = (double)src->baseHdrHeadroomN / src->baseHdrHeadroomD;
     dst->alternateHdrHeadroom = (double)src->alternateHdrHeadroomN / src->alternateHdrHeadroomD;
-    dst->backwardDirection = src->backwardDirection;
     dst->useBaseColorSpace = src->useBaseColorSpace;
     return AVIF_TRUE;
 }
@@ -81,11 +79,8 @@ static float avifGetGainMapWeight(float hdrHeadroom, const avifGainMapMetadataDo
         // This case is not handled in the specification and does not make practical sense.
         return 0.0f;
     }
-    float w = AVIF_CLAMP((hdrHeadroom - baseHdrHeadroom) / (alternateHdrHeadroom - baseHdrHeadroom), 0.0f, 1.0f);
-    if (metadata->backwardDirection) {
-        w *= -1.0f;
-    }
-    return w;
+    const float w = AVIF_CLAMP((hdrHeadroom - baseHdrHeadroom) / (alternateHdrHeadroom - baseHdrHeadroom), 0.0f, 1.0f);
+    return (alternateHdrHeadroom < baseHdrHeadroom) ? -w : w;
 }
 
 // Linear interpolation between 'a' and 'b' (returns 'a' if w == 0.0f, returns 'b' if w == 1.0f).
@@ -517,8 +512,6 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
     const avifBool colorSpacesDiffer = (baseColorPrimaries != altColorPrimaries);
     avifColorPrimaries gainMapMathPrimaries;
     AVIF_CHECKRES(avifChooseColorSpaceForGainMapMath(baseColorPrimaries, altColorPrimaries, &gainMapMathPrimaries));
-    const avifBool useBaseColorSpace = (gainMapMathPrimaries == baseColorPrimaries);
-
     const int width = baseRgbImage->width;
     const int height = baseRgbImage->height;
 
@@ -549,6 +542,7 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
 
     avifGainMapMetadataDouble gainMapMetadata;
     avifGainMapMetadataSetDefaults(&gainMapMetadata);
+    gainMapMetadata.useBaseColorSpace = (gainMapMathPrimaries == baseColorPrimaries);
 
     float (*baseGammaToLinear)(float) = avifTransferCharacteristicsGetGammaToLinearFunction(baseTransferCharacteristics);
     float (*altGammaToLinear)(float) = avifTransferCharacteristicsGetGammaToLinearFunction(altTransferCharacteristics);
@@ -557,7 +551,7 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
 
     double rgbConversionCoeffs[3][3];
     if (colorSpacesDiffer) {
-        if (useBaseColorSpace) {
+        if (gainMapMetadata.useBaseColorSpace) {
             if (!avifColorPrimariesComputeRGBToRGBMatrix(altColorPrimaries, baseColorPrimaries, rgbConversionCoeffs)) {
                 avifDiagnosticsPrintf(diag, "Unsupported RGB color space conversion");
                 res = AVIF_RESULT_NOT_IMPLEMENTED;
@@ -581,11 +575,12 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
         float channelMin[3] = { 0 };
         for (int j = 0; j < height; ++j) {
             for (int i = 0; i < width; ++i) {
-                avifGetRGBAPixel(useBaseColorSpace ? altRgbImage : baseRgbImage, i, j, useBaseColorSpace ? &altRGBInfo : &baseRGBInfo, rgba);
+                avifGetRGBAPixel(gainMapMetadata.useBaseColorSpace ? altRgbImage : baseRgbImage, i, j,
+                                 gainMapMetadata.useBaseColorSpace ? &altRGBInfo : &baseRGBInfo, rgba);
 
                 // Convert to linear.
                 for (int c = 0; c < 3; ++c) {
-                    if (useBaseColorSpace) {
+                    if (gainMapMetadata.useBaseColorSpace) {
                         rgba[c] = altGammaToLinear(rgba[c]);
                     } else {
                         rgba[c] = baseGammaToLinear(rgba[c]);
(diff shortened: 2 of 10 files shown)


tests/gtest/avifgainmaptest.cc
```diff
@@ -47,7 +46,6 @@ void CheckGainMapMetadataMatches(const avifGainMapMetadata& lhs,
 
 avifGainMapMetadata GetTestGainMapMetadata(bool base_rendition_is_hdr) {
   avifGainMapMetadata metadata = {};
-  metadata.backwardDirection = base_rendition_is_hdr;
```
wantehchang, line 50: The `bool base_rendition_is_hdr` parameter is not used now. Should we remove it?

include/avif/avif.h
```diff
@@ -630,17 +630,14 @@ typedef struct avifGainMapMetadata
...
+    // f = clamp((H - baseHdrHeadroom) /
+    //           (alternateHdrHeadroom - baseHdrHeadroom), 0, 1);
+    // w = sign(alternateHdrHeadroom - baseHdrHeadroom) * f
     uint32_t baseHdrHeadroomN;
     uint32_t baseHdrHeadroomD;
     uint32_t alternateHdrHeadroomN;
     uint32_t alternateHdrHeadroomD;
 
-    // True if the gain map should be applied in reverse, see weight formula above.
-    avifBool backwardDirection;
```
vigneshvg, line 642: i know it's behind a flag, but may worth adding a CHANGELOG entry for this "public" ABI change.

src/gainmap.c
```diff
@@ -82,8 +80,8 @@ static float avifGetGainMapWeight(float hdrHeadroom, const avifGainMapMetadataDo
         return 0.0f;
     }
     float w = AVIF_CLAMP((hdrHeadroom - baseHdrHeadroom) / (alternateHdrHeadroom - baseHdrHeadroom), 0.0f, 1.0f);
-    if (metadata->backwardDirection) {
-        w *= -1.0f;
+    if (alternateHdrHeadroom < baseHdrHeadroom) {
+        w = -w;
     }
     return w;
```
vigneshvg, line 86: optional: `w` can be `const` and this can be a ternary return:

```
return (alternateHdrHeadroom < baseHdrHeadroom) ? -w : w;
```

vigneshvg:
The CI failure seems to be related? can you please take a look at that too.

ccameron-chromium:
> The CI failure seems to be related? can you please take a look at that too.

Yeah, the failure is from my change. I meant to send out the email asking "how do I test stuff" before sending the PR to you!

include/avif/avif.h
wantehchang: In this case it is not necessary to add a CHANGELOG entry, because the gain map code has not appeared in any libavif release yet.

tests/gtest/avifgainmaptest.cc
ccameron-chromium: Removed.

include/avif/avif.h
ccameron-chromium: Added.

src/gainmap.c
ccameron-chromium: Done.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 ## [Unreleased]
 
 ### Added since 1.0.0
+* Remove backwardDirection from `avifGainMapMetadata`.
```
wantehchang, line 12: Please remove this item. The new experimental gain map feature is covered by the next item (line 13). Since this new feature hasn't appeared in a libavif release, we don't need to document changes to this new feature.

src/gainmap.c
```diff
@@ -658,6 +655,21 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
 
+    // Compute the headroom for the base and alterate images. The gainmap is the log-ratio of the
+    // higher-headroom (more HDR) image to the lower-headroom (less HDR) image. Reverse the gainmap
+    // if this is the case.
+    gainMapMetadata.baseHdrHeadroom = log2f(AVIF_MAX(baseMax, kEpsilon));
+    gainMapMetadata.alternateHdrHeadroom = log2f(AVIF_MAX(altMax, kEpsilon));
+    if (gainMapMetadata.baseHdrHeadroom > gainMapMetadata.alternateHdrHeadroom) {
+        for (int j = 0; j < height; ++j) {
+            for (int i = 0; i < width; ++i) {
+                for (int c = 0; c < numGainMapChannels; ++c) {
```
wantehchang, line 666: Nit: Since `c` is the index for the first dimension of the `gainMapF` array, it would be better to let `c` be the outer loop. See the nested for loop at line 697 for an example.

tests/gtest/avifgainmaptest.cc
```diff
@@ -44,7 +44,7 @@ void CheckGainMapMetadataMatches(const avifGainMapMetadata& lhs,
   }
 }
 
-avifGainMapMetadata GetTestGainMapMetadata(bool base_rendition_is_hdr) {
+avifGainMapMetadata GetTestGainMapMetadata() {
```
wantehchang, line 47: IMPORTANT: I wonder if we should keep the `base_rendition_is_hdr` boolean parameter and let it control `metadata.baseHdrHeadroomN/D` and `metadata.alternateHdrHeadroomN/D`.

Right now `baseHdrHeadroom` is hardcoded to be 0 (0/1) and `alternateHdrHeadroom` is hardcoded to be 3 (6/2). So `alternateHdrHeadroom < baseHdrHeadroom` is always false. This seems incorrect.

src/gainmap.c
```diff
@@ -658,6 +655,21 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
         }
     }
 
+    // Compute the headroom for the base and alterate images. The gainmap is the log-ratio of the
+    // higher-headroom (more HDR) image to the lower-headroom (less HDR) image. Reverse the gainmap
+    // if this is the case.
```
wantehchang, line 660: I assume it is this change that fixed the test failures on Windows. Do you know why the tests only fail on Windows?

Seel also my comment (marked with IMPORTANT) in tests/gtest/avifgainmaptest.cc. I wonder if that is why the tests failed on Windows. Still, it is strange that the tests did not fail on all platforms.

ccameron-chromium: Yes, this is why the test failed. I just assumed that only AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP was defined on the Windows machine. I developed the patch locally on Linux, and the test failed locally for me (which would be consistent with only AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP being defined on the Windows bot).

tests/gtest/avifgainmaptest.cc
ccameron-chromium: Changed this to swap base/alt headroom if base_rendition_is_hdr.

CHANGELOG.md
ccameron-chromium: Removed.

include/avif/avif.h
ccameron-chromium: Removed.

src/gainmap.c
```diff
@@ -660,6 +655,21 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
+    gainMapMetadata.baseHdrHeadroom = log2f(AVIF_MAX(baseMax, kEpsilon));
+    gainMapMetadata.alternateHdrHeadroom = log2f(AVIF_MAX(altMax, kEpsilon));
```
wantehchang, line 662: Move lines 683-690 here to be with lines 658-662 as before. These lines "fill in the gain map's metadata,", so they should stay together.

tests/gtest/avif_fuzztest_enc_dec_experimental.cc
```diff
@@ -24,7 +24,6 @@ void CheckGainMapMetadataMatches(const avifGainMapMetadata& actual,
   // 'expecteed' is the source struct which has arbitrary data and booleans
   // values can contain any value, but the decoded ('actual') struct should
   // be 0 or 1.
```
wantehchang, line 26: Delete this comment. It describes the line that we deleted.

tests/gtest/avifgainmaptest.cc
```diff
@@ -47,12 +46,15 @@ void CheckGainMapMetadataMatches(const avifGainMapMetadata& lhs,
...
-  metadata.backwardDirection = base_rendition_is_hdr;
   metadata.useBaseColorSpace = true;
   metadata.baseHdrHeadroomN = 0;
   metadata.baseHdrHeadroomD = 1;
   metadata.alternateHdrHeadroomN = 6;
   metadata.alternateHdrHeadroomD = 2;
+  if (base_rendition_is_hdr) {
+    std::swap(metadata.baseHdrHeadroomN, metadata.alternateHdrHeadroomN);
+    std::swap(metadata.baseHdrHeadroomD, metadata.alternateHdrHeadroomD);
+  }
```
wantehchang, line 57: Should we also swap `baseOffset` and `alternateOffset` if `base_rendition_is_hdr` is true?

tests/gtest/avifgainmaptest.cc
```diff
@@ -884,7 +886,6 @@ TEST(GainMapTest, ConvertMetadata) {
   metadata_double.alternateOffset[1] = 0.0;
   metadata_double.baseHdrHeadroom = 1.0;
   metadata_double.alternateHdrHeadroom = 10.0;
```
wantehchang, line 888: The original code has `backwardDirection = AVIF_TRUE`. To match that, should we make sure `baseHdrHeadroom > alternateHdrHeadroom`?

wantehchang (approved):
Chris: I took a closer look and have some additional comments/questions. PTAL. Thanks!

src/gainmap.c
ccameron-chromium: I disagree here. We populate `gainMapMetadata` all over the function. The offsets are computed up on line 603/605.

This is really the most natural place to populate `gainMapMetadata.alternateHdrHeadroom` and `gainMapMetadata.baseHdrHeadroom`. The reason for this is that
* Whether or not the gainmap is negated depends on whether or not depends on these values
* Whether or not the gainmap is negated affects how to compute `gainMapMinLog2` and `gainMapMaxLog2`.
* We can't populate `gainMapMetadata.gainMapMin` until we've computed `gainMapMinLog2`
So it would be fairly awkward to 

I've changed this function to no longer say "Fill in the gain map's metadata", but rather to say "Populate the gain map metdata's min and max fields", and I've separated it (just with a newline and comment) from the part that converts to fractions.

I also moved populating of `gainMapMetadata.useBaseColorSpace` to where `gainMapMetadata` is initialized, and removed the local `useBaseColorSpace` variable.

Let me know if this still feels unreasonable to you.

tests/gtest/avifgainmaptest.cc
ccameron-chromium: No, this was needed in the JPEG code because they explicitly say "SDR offset" and "HDR offset" in the XMP. When formulated as "base offset" and "alternate offset", no conversion is needed.

tests/gtest/avifgainmaptest.cc
ccameron-chromium: No, that combination (where base is SDR and altr is HDR, but the gainmap "goes in reverse") is no longer supported in the spec.

src/gainmap.c
wantehchang: Thank you for the explanation. Your rearrangement of the code and new comments make sense.

tests/gtest/avifgainmaptest.cc
wantehchang: Thanks for the explanation. Did you mean we store JPEG's "SDR offset" in `metadata.baseOffset` and store JPEG's "HDR offset" in `metadata.alternateOffset`?

I wonder if we should document that because it is not obvious.

src/gainmap.c
```diff
@@ -660,6 +655,22 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
         }
     }
 
+    // Populate the gain map metadata's headrooms.
+    gainMapMetadata.baseHdrHeadroom = log2f(AVIF_MAX(baseMax, kEpsilon));
+    gainMapMetadata.alternateHdrHeadroom = log2f(AVIF_MAX(altMax, kEpsilon));
+
+    // The gainmap is the log-ratio of the higher-headroom (more HDR) image to the lower-headroom
+    // (less HDR) image. Reverse the gainmap if this is the case.
```
wantehchang, line 663: It is not clear what "this is the case" means. Could you clarify that? Should we say "if this is **not** the case"?

src/gainmap.c
```diff
@@ -660,6 +655,22 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
+    if (baseMax > altMax) {
```
wantehchang, line 664: I am curious why you changed this check to compare `baseMax` and `altMax` instead of `baseHdrHeadroom` and  `alternateHdrHeadroom`?

src/gainmap.c
```diff
@@ -670,16 +681,14 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
         }
     }
 
-    // Fill in the gain map's metadata.
+    // Populate in the gain map metadata's min and max values.
```
wantehchang, line 684: Remove "in".

wantehchang (approved):
LGTM. Thanks! Just two small suggested changes to the comments, and a question about a check.

tests/gtest/avifgainmaptest.cc
ccameron-chromium: I was referring to [this code](https://source.chromium.org/chromium/chromium/src/+/main:third_party/libavif/src/apps/shared/avifjpeg.c;l=619;drc=c3809cce7333810f66b79f5319b799422463897e) where we read SDR into "base" and then [swap them](https://source.chromium.org/chromium/chromium/src/+/main:third_party/libavif/src/apps/shared/avifjpeg.c;l=643;drc=c3809cce7333810f66b79f5319b799422463897e) if the base is HDR.

src/gainmap.c
ccameron-chromium: That was an accident (I was playing with using this instead of the `gainMapMetadata` parameters, to see if it was reasonable to populate them all at the end... this ended up staying by accident). Changed it back.

src/gainmap.c
ccameron-chromium: Done.

wantehchang (approved):

PR #2228, opened 2024-06-25
Upgrade to cargo-c v0.10.0 + fix for cargo-c#383

wantehchang:


cmake/Modules/LocalRav1e.cmake
```diff
@@ -1,8 +1,6 @@
 set(AVIF_LOCAL_RAV1E_GIT_TAG v0.7.1)
 set(AVIF_LOCAL_CORROSION_GIT_TAG v0.5.0)
-# v0.9.32 is the highest version of cargo-c we can upgrade to without a build
-# failure (on Windows). See https://github.com/AOMediaCodec/libavif/pull/2223.
-set(AVIF_LOCAL_CARGOC_GIT_TAG v0.9.32)
+set(AVIF_LOCAL_CARGOC_GIT_TAG v0.10.0)
```
wantehchang, line 3: Hi @amyspark,

We (libavif) get a build error on Windows when we upgrade to cargo-c v0.10.0:

https://github.com/AOMediaCodec/libavif/actions/runs/9668355268/job/26672314397#step:6:1924

It is apparently related to your change "[Implement installing MSVC debug symbols](https://github.com/lu-zero/cargo-c/commit/19ec991ebfcb32697b9099537e0a16af289fed01)" in cargo-c v0.10.0. Could you take a look at this and see if you know what might be wrong?

I will try to reproduce the build error locally on my Windows laptop when I go home.

Thank you for your help!

amyspark: Hi! From the looks of it I never thought of the case when Rust could *not* generate the PDB file on linking. However, I couldn't find what profile libavif's building rav1e with. Do you know if libavif is setting `strip` to something other than `none`, or alternatively, where it's selecting the build profile for rav1e?

wantehchang: Hi @amyspark,

Thanks for taking a look! I am not familiar with Rust or rav1e, but you can find the relevant code by searching for "rav1e" in https://github.com/AOMediaCodec/libavif/blob/main/cmake/Modules/LocalRav1e.cmake.

I haven't had a chance to reproduce this build error on my Windows laptop.

wantehchang: I reproduced the error locally. There is no rav1e.pdb. I tested with both the release and debug profile. The contents of the output directory is:

```
[.]                    [..]                   .cargo-lock            [.fingerprint]         [build]
cargo-c-rav1e.cache    [deps]                 [examples]             [include]              [incremental]
rav1e-uninstalled.pc   rav1e.d                rav1e.h                rav1e.lib              rav1e.pc
```

wantehchang: It seems that rav1e.pdb is not generated if `--library-type=staticlib` is passed to `cargo cinstall`.

wantehchang: In https://github.com/lu-zero/cargo-c/issues/279, we have:

> tl;dr: if you install foo.dll you must also install foo.pdb next to it. Same with foo.exe. You can ignore static libraries.

Perhaps https://github.com/lu-zero/cargo-c/commit/19ec991ebfcb32697b9099537e0a16af289fed01 should ignore static libraries?

amyspark: One could still set on the `release` profile to not generate debugging information for shared libraries, so I'd go with making this step skippable if the PDB file is not present.

wantehchang: You are right. It is best to assume the PDB file may not be present. Could you write a pull request for cargo-c? Thanks.

wantehchang: Hi @amyspark, I filed a bug report for cargo-c v0.10.0: https://github.com/lu-zero/cargo-c/issues/383

jzern (approved):

PR #2229, opened 2024-06-26
Update low-overhead image format prototype

y-guyon:
Use MetaBox with version 1 instead of MinimizedImageBox.


src/read.c
```diff
@@ -3607,198 +3559,267 @@ static avifProperty * avifDecoderItemAddProperty(avifDecoderItem * item, const a
...
+    } else if (hasExplicitCicp) {
+        AVIF_CHECKERR(avifROStreamReadBits8(s, &matrixCoefficients, 8), AVIF_RESULT_BMFF_PARSE_FAILED); // bit(8) matrix_coefficients;
+    } else {
+        matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601; // 6
     }
 
-    uint32_t hasExplicitCodecTypes;
-    AVIF_CHECKERR(avifROStreamReadBits(&s, &hasExplicitCodecTypes, 1), AVIF_RESULT_BMFF_PARSE_FAILED); // bit(1) has_explicit_codec_types;
+    char infeType[] = "av01";
+    char codecConfigType[] = "av1C";
```
wantehchang, line 3643: Declare these as `uint8_t` arrays of size 4 and delete the `(uint8_t *)` casts in line 3645 and line 3646.

See the `type` field in the `avifDecoderItem` struct:
```
typedef struct avifDecoderItem
{
    ...
    uint8_t type[4];
    ...
} avifDecoderItem;
```

The default values can be set using `memcpy()` calls. Also, the default values can be set in a new `else` branch at line 3654, similar to how we set the default values for `colorPrimaries`, `transferCharacteristics`, etc.

wantehchang (commented):
Thanks for updating the prototype. I haven't reviewed the pull request. Just a comment on a commit that have looked at.

wantehchang (commented):
I have reviewed all the files except src/read.c and src/write.c. I will review those two files tomorrow.

tests/gtest/aviftest_helpers.cc
```diff
@@ -230,15 +231,30 @@ bool AreImagesEqual(const avifImage& image1, const avifImage& image2,
...
+        image1.clap.heightN != image2.clap.heightN ||
+        image1.clap.heightD != image2.clap.heightD ||
+        image1.clap.horizOffN != image2.clap.horizOffN ||
+        image1.clap.horizOffD != image2.clap.horizOffD ||
+        image1.clap.vertOffN != image2.clap.vertOffN ||
+        image1.clap.vertOffD != image2.clap.vertOffD))) {
+    return false;
+  }
+  // Different rotation and mirroring properties could lead to the same
+  // orientation. Convert to Exif orientation for an easy comparison.
```
wantehchang, line 255: IMPORTANT: This is also true for the fractions in the `clap` property. If we allow different but equivalent representations, we can call `avifFractionSimplify()` to simplify the fractions before comparing them.

On the other hand, if `AreImagesEqual()` means "the images are identical" rather than "the images are equivalent", then we should require all struct members to be equal. It seems that this is now `AreImagesEqual()` is used by the tests. Please double check.

src/write.c
```diff
@@ -2306,46 +2252,62 @@ static avifBool avifEncoderIsMinimizedImageBoxCompatible(const avifEncoder * enc
...
             colorItem = item;
-            continue; // The primary item can be stored in the MinimizedImageBox.
+            if (item->encodeOutput->samples.count != 1 || item->encodeOutput->samples.sample[0].data.size > (1 << 28)) {
+                return AVIF_FALSE;
+            }
+            continue; // The primary item can be stored in the MetaBox with version 1.
         }
         if (item->itemCategory == AVIF_ITEM_ALPHA && item->irefToID == encoder->data->primaryItemID) {
-            continue; // The alpha auxiliary item can be stored in the MinimizedImageBox.
+            if (item->encodeOutput->samples.count != 1 || item->encodeOutput->samples.sample[0].data.size >= (1 << 28)) {
```
wantehchang, line 2291: We use `>` in line 2285 but `>=` here. Should these two conditional expressions be exactly the same? Or is the maximum size of the alpha item smaller?

src/write.c
```diff
@@ -2400,129 +2358,218 @@ static avifResult avifEncoderWriteMinimizedImageBox(avifEncoder * encoder, avifR
...
+                                       : image->yuvFormat == AVIF_PIXEL_FORMAT_YUV422 ? 2
+                                                                                      : 3;
+
+    const avifColorPrimaries defaultColorPrimaries = hasIcc ? AVIF_COLOR_PRIMARIES_UNSPECIFIED : AVIF_COLOR_PRIMARIES_SRGB;
+    const avifTransferCharacteristics defaultTransferCharacteristics = hasIcc ? AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED
+                                                                              : AVIF_TRANSFER_CHARACTERISTICS_SRGB;
+    const avifMatrixCoefficients defaultMatrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601;
+    const avifBool hasExplicitCicp = image->colorPrimaries == defaultColorPrimaries ||
+                                     image->transferCharacteristics == defaultTransferCharacteristics ||
+                                     image->matrixCoefficients == defaultMatrixCoefficients;
```
wantehchang, line 2376: IMPORTANT: The value of `hasExplicitCicp` seems inverted. Please double check.

src/write.c
```diff
@@ -2400,129 +2358,218 @@ static avifResult avifEncoderWriteMinimizedImageBox(avifEncoder * encoder, avifR
...
-    }
+    const avifBool fullRange = image->yuvRange == AVIF_RANGE_FULL;
+    const uint32_t pixelFormat = image->depth == 8    ? AVIF_METAV1_PIXEL_FORMAT_UINT8
+                                 : image->depth == 10 ? AVIF_METAV1_PIXEL_FORMAT_UINT10
+                                                      : AVIF_METAV1_PIXEL_FORMAT_UINT12;
+
+    // AV1 does not explicitly signal horizontal chroma siting (could be CSP_RESERVED though).
+    const avifBool isHorizontallyCentered = AVIF_FALSE;
+    const avifBool isVerticallyCentered = image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420 &&
+                                          image->yuvChromaSamplePosition == AVIF_CHROMA_SAMPLE_POSITION_VERTICAL;
```
wantehchang, line 2386: IMPORTANT: We should review the values of `isHorizontallyCentered` and `isVerticallyCentered`.

I assume they are only used for 4:2:0. For 4:2:0 AVIF, it is better to assume the "center" chroma sample position unless `image->yuvChromaSamplePosition` is `VERTICAL` or `COLOCATED`.

src/write.c
```diff
@@ -2400,129 +2358,218 @@ static avifResult avifEncoderWriteMinimizedImageBox(avifEncoder * encoder, avifR
...
-        AVIF_CHECKRES(avifRWStreamWriteVarInt(s, (uint32_t)alphaData->size));  // varint alpha_item_data_size;
-        AVIF_ASSERT_OR_RETURN(alphaData->size != 0);
-        AVIF_CHECKRES(avifRWStreamWriteVarInt(s, 4)); // varint alpha_item_codec_config_size;
-    }
+    const avifBool fullRange = image->yuvRange == AVIF_RANGE_FULL;
+    const uint32_t pixelFormat = image->depth == 8    ? AVIF_METAV1_PIXEL_FORMAT_UINT8
+                                 : image->depth == 10 ? AVIF_METAV1_PIXEL_FORMAT_UINT10
+                                                      : AVIF_METAV1_PIXEL_FORMAT_UINT12;
+
+    // AV1 does not explicitly signal horizontal chroma siting (could be CSP_RESERVED though).
```
wantehchang, line 2383: This comment seems wrong. If the AV1 `chroma_sample_position` is `CSP_VERTICAL` or `CSP_COLOCATED`, then chroma is NOT horizontally centered.

wantehchang (commented):
This is the first round of review. I skimmed through src/read.c and src/write.c, mainly to verify that if the experimental feature is disabled, the changes are effectively a no-op. I will take another look tomorrow.

Could you email me a copy of the draft proposal? I can try to compare the code with the draft proposal. (I may not have time to do that though.)

wantehchang (commented):
Vincent: Please review the changes related to the removal of the `aviftest_helpers_internal` library in tests/. Thanks!

tests/CMakeLists.txt
```diff
@@ -134,8 +132,8 @@ if(AVIF_ENABLE_GTEST)
     add_avif_gtest_with_data(aviflosslesstest)
     add_avif_gtest_with_data(avifmetadatatest)
 
-    if(AVIF_ENABLE_EXPERIMENTAL_MINI)
-        add_avif_gtest(avifminitest)
+    if(AVIF_ENABLE_EXPERIMENTAL_METAV1)
+        add_avif_gtest(avifmetav1test)
```
vrabaud, line 136: That should probably become `add_avif_internal_gtest`

tests/CMakeLists.txt
```diff
@@ -42,9 +42,7 @@ endforeach()
 
 if(AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
     add_library(aviftest_helpers OBJECT gtest/aviftest_helpers.cc)
-    target_link_libraries(aviftest_helpers avif_apps avif)
-    add_library(aviftest_helpers_internal OBJECT gtest/aviftest_helpers.cc)
-    target_link_libraries(aviftest_helpers_internal avif_apps_internal avif_internal)
+    target_link_libraries(aviftest_helpers avif_apps_internal avif_internal)
```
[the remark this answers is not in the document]
vrabaud: Exactly. Those should not be changed. If there is a problem, it's the wrong macro that is called.

wantehchang:
I wrote: 
> Could you email me a copy of the draft proposal?

Please ignore this request. I downloaded a copy of w23988 from https://www.mpeg.org/wp-content/uploads/mpeg_meetings/146_Rennes/w23988.zip.

src/write.c
```diff
@@ -2400,129 +2358,218 @@ static avifResult avifEncoderWriteMinimizedImageBox(avifEncoder * encoder, avifR
...
+    const avifBool hasAlpha = alphaItem != NULL;
+    const avifBool alphaIsPremultiplied = encoder->data->imageMetadata->alphaPremultiplied;
+    const avifBool hasHdr = AVIF_FALSE; // Not implemented.
+    const avifBool hasIcc = image->icc.size != 0;
+    const uint32_t chromaSubsampling = image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400   ? 0
+                                       : image->yuvFormat == AVIF_PIXEL_FORMAT_YUV420 ? 1
+                                       : image->yuvFormat == AVIF_PIXEL_FORMAT_YUV422 ? 2
+                                                                                      : 3;
+
+    const avifColorPrimaries defaultColorPrimaries = hasIcc ? AVIF_COLOR_PRIMARIES_UNSPECIFIED : AVIF_COLOR_PRIMARIES_SRGB;
```
wantehchang, line 2370: Optional: Use the original enum name `AVIF_COLOR_PRIMARIES_BT709` rather than the alias `AVIF_COLOR_PRIMARIES_SRGB`.

src/write.c
wantehchang: I understand this now. This is because one has `minus_one` and the other does not:

```
    unsigned int(item_data_size_size) main_item_data_size_minus_one;
    if (has_alpha)
        unsigned int(item_data_size_size) alpha_item_data_size;
```

src/write.c
```diff
@@ -2400,129 +2358,218 @@ static avifResult avifEncoderWriteMinimizedImageBox(avifEncoder * encoder, avifR
...
+        AVIF_CHECKRES(avifRWStreamWriteBits(s, image->transferCharacteristics, 8)); // bit(8) transfer_characteristics;
+    }
+    if (chromaSubsampling == 0) {
+        AVIF_CHECKERR(image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED, AVIF_RESULT_ENCODE_COLOR_FAILED);
+    } else if (hasExplicitCicp) {
+        AVIF_CHECKRES(avifRWStreamWriteBits(s, image->matrixCoefficients, 8)); // bit(8) matrix_coefficients;
+    }
+
+    AVIF_CHECKRES(avifRWStreamWriteBits(s, codecConfigSize, codecConfigSizeSize)); // unsigned int(codec_config_size_size) main_item_codec_config_size;
+    AVIF_CHECKRES(avifRWStreamWriteBits(s, (uint32_t)colorData->size - 1, itemDataSizeSize)); // unsigned int(item_data_size_size) main_item_data_size;
```
wantehchang, line 2440: Search for `(uint32_t)` casts. Make sure the value is <= `UINT32_MAX` before casting.

wantehchang, line 2440: Typo: main_item_data_size => main_item_data_size_minus_one

wantehchang (commented):
I read src/read.c and src/write.c again. Here are the comments from the second round of code review.

src/exif.c
```diff
@@ -142,6 +142,48 @@ avifResult avifImageExtractExifOrientationToIrotImir(avifImage * image)
     return AVIF_RESULT_OK;
 }
 
+uint8_t avifImageIrotImirToExifOrientation(const avifImage * image)
```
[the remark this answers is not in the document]
y-guyon: Thank you for the snippet. It would require checking `axis` and `angle` values before using them as array indices though.

Keeping the current version as it is already done.

src/read.c
```diff
@@ -2955,11 +2921,30 @@ static avifResult avifParseItemReferenceBox(avifMeta * meta, const uint8_t * raw
...
-    AVIF_CHECKERR(avifROStreamReadAndEnforceVersion(&s, 0), AVIF_RESULT_BMFF_PARSE_FAILED);
+    uint8_t version;
+    uint32_t flags;
+    AVIF_CHECKERR(avifROStreamReadVersionAndFlags(&s, &version, &flags), AVIF_RESULT_BMFF_PARSE_FAILED);
+
+#if defined(AVIF_ENABLE_EXPERIMENTAL_METAV1)
+    if (version == 1) {
+        // Experimental box.
+        AVIF_CHECKRES(avifParseMetaBoxV1(&s, meta, rawOffset, flags, diag));
+        return AVIF_RESULT_OK;
```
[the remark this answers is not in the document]
y-guyon: I would rather keep `AVIF_CHECKRES` for `AVIF_BREAK_ON_ERROR` to report the issue as early as possible, in case `avifParseMetaBoxV1()` has a direct `return` statement.

src/read.c
```diff
@@ -3872,23 +3901,51 @@ static avifResult avifParseMinimizedImageBox(avifMeta * meta, uint64_t rawOffset
...
-        offset += alphaItemDataSize;
-        alphaItem->size = alphaItemDataSize;
+        alphaExtent->offset = rawOffset + avifROStreamOffset(s);
+        alphaExtent->size = (size_t)alphaItemDataSize;
+        AVIF_CHECKERR(avifROStreamSkip(s, alphaExtent->size), AVIF_RESULT_BMFF_PARSE_FAILED);
+        alphaItem->size = alphaExtent->size;
     }
 
+    // if (has_hdr && has_gainmap && gainmap_item_data_size > 0)
+    //     unsigned int(8) gainmap_item_data[gainmap_item_data_size];
```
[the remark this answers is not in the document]
y-guyon: Already done in the earlier `if (hasHdr)` check above.

src/write.c
y-guyon: This should be enforced by `avifEncoderIsMetaBoxV1Compatible()`.

tests/CMakeLists.txt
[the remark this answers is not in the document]
y-guyon: Reverted to the original `AreImagesEqual()` behavior.

src/write.c
y-guyon: I went for `CSP_UNKNOWN` meaning unknown or centered.

src/read.c
```diff
@@ -3607,198 +3559,270 @@ static avifProperty * avifDecoderItemAddProperty(avifDecoderItem * item, const a
...
+    const uint32_t orientation = (flags & 0x070000) >> 16;                   // orientation (3 bits)
+    const avifBool hasExplicitCodecTypes = (flags & 0x080000) != 0;          // has_explicit_codec_types
+    const uint32_t dimensionSize = ((flags & 0x100000) != 0) ? 7 : 15;       // dimension_size
+    const uint32_t codecConfigSizeSize = ((flags & 0x200000) != 0) ? 3 : 12; // codec_config_size_size
+    const uint32_t itemDataSizeSize = ((flags & 0x400000) != 0) ? 15 : 28;   // item_data_size_size
+    const uint32_t metadataSizeSize = ((flags & 0x800000) != 0) ? 10 : 20;   // metadata_size_size
+
+    // Make sure the total MetaBox size is coherent with the flags.
+    const size_t maxHeaderSize = 128; // To be tuned.
+    AVIF_CHECKERR(s->raw->size < maxHeaderSize + 3 * (1 << codecConfigSizeSize) + 3 * (1 << itemDataSizeSize) + 3 * (1 << metadataSizeSize),
```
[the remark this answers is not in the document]
y-guyon: I removed this check. The one after padding and codec configs is more precise and simpler.

include/avif/avif.h
```diff
@@ -210,9 +210,9 @@ typedef enum avifHeaderFormat
 {
     // AVIF file with an "avif" brand, a MetaBox and all its required boxes for maximum compatibility.
     AVIF_HEADER_FULL,
-#if defined(AVIF_ENABLE_EXPERIMENTAL_MINI)
-    // AVIF file with an "mif3" brand and a MinimizedImageBox to reduce the encoded file size.
-    // This is based on the m64572 "Condensed image item" MPEG proposal for HEIF.
+#if defined(AVIF_ENABLE_EXPERIMENTAL_METAV1)
+    // AVIF file with an "mif3" brand and a MetaBox with version 1 to reduce the encoded file size.
```
wantehchang, line 214: Nit: change "an" to "a" before "mif3".

include/avif/internal.h
```diff
@@ -450,6 +462,10 @@ AVIF_NODISCARD avifBool avifAreGridDimensionsValid(avifPixelFormat yuvFormat,
 // image->imir on success. Returns AVIF_RESULT_INVALID_EXIF_PAYLOAD on failure.
 avifResult avifImageExtractExifOrientationToIrotImir(avifImage * image);
 
+// Returns the Exif orientation in [1-8] as defined in JEITA CP-3451C section 4.6.4.A Orientation
+// corresponding to image->irot and image->imir.
+uint8_t avifImageIrotImirToExifOrientation(const avifImage * image);
```
wantehchang, line 467: This function is only used in src/write.c now. Move this function to src/write.c as a static function, or put this function inside `#if defined(AVIF_ENABLE_EXPERIMENTAL_METAV1)`.

src/exif.c
```diff
@@ -142,6 +142,52 @@ avifResult avifImageExtractExifOrientationToIrotImir(avifImage * image)
...
```
wantehchang, line 145: Move this function to src/write.c as a static function, or put this function inside `#if defined(AVIF_ENABLE_EXPERIMENTAL_METAV1)`.

src/read.c
```diff
@@ -3607,198 +3559,276 @@ static avifProperty * avifDecoderItemAddProperty(avifDecoderItem * item, const a
...
+                                   : chromaSubsampling == 2 ? AVIF_PIXEL_FORMAT_YUV422
+                                                            : AVIF_PIXEL_FORMAT_YUV444;
+    if (colorItem->metaV1PixelFormat == AVIF_PIXEL_FORMAT_YUV422) {
+        // In AV1, the chroma_sample_position syntax element is not present for the YUV 4:2:2 format.
+        // Assume that AV1 uses the same 4:2:2 chroma sample location as HEVC and VVC (colocated).
+        AVIF_CHECKERR(!isHorizontallyCentered, AVIF_RESULT_BMFF_PARSE_FAILED);
+        // isVerticallyCentered: Ignored unless chroma_subsampling is 1.
+        colorItem->metaV1ChromaSamplePosition = AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN;
+    } else if (colorItem->metaV1PixelFormat == AVIF_PIXEL_FORMAT_YUV420 && isHorizontallyCentered) {
+        // There is no way to describe this with AV1's chroma_sample_position enum besides CSP_UNKNOWN or CSP_RESERVED.
```
wantehchang, line 3807: Nit/optional: The "or CSP_RESERVED" part is a little confusing. I suggest removing it. Perhaps we could add a second sentence "There is a proposal to assign the reserved value 3 (CSP_RESERVED) to the center chroma sample position."

src/read.c
```diff
@@ -3607,198 +3559,276 @@ static avifProperty * avifDecoderItemAddProperty(avifDecoderItem * item, const a
...
+        colorItem->metaV1ChromaSamplePosition = AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN;
+    } else if (colorItem->metaV1PixelFormat == AVIF_PIXEL_FORMAT_YUV420) {
+        colorItem->metaV1ChromaSamplePosition = isVerticallyCentered ? AVIF_CHROMA_SAMPLE_POSITION_VERTICAL
+                                                                     : AVIF_CHROMA_SAMPLE_POSITION_COLOCATED;
+    } else {
```
wantehchang, line 3812: Nit: I suggest restructuring the second and third cases (lines 3806-3812) as follows, so that it is clearer what the final "else" case is.

```
    } else if (colorItem->metaV1PixelFormat == AVIF_PIXEL_FORMAT_YUV420) {
        if (isHorizontallyCentered) {
            // There is no way to describe this with AV1's chroma_sample_position enum besides CSP_UNKNOWN or CSP_RESERVED.
            colorItem->metaV1ChromaSamplePosition = AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN;
        } else {
            colorItem->metaV1ChromaSamplePosition = isVerticallyCentered ? AVIF_CHROMA_SAMPLE_POSITION_VERTICAL
                                                                         : AVIF_CHROMA_SAMPLE_POSITION_COLOCATED;
        }
    } else {       
```

src/read.c
```diff
@@ -4046,59 +4096,64 @@ static avifResult avifParse(avifDecoder * decoder)
...
-            miniSeen = AVIF_TRUE;
-#endif // AVIF_ENABLE_EXPERIMENTAL_MINI
         }
 
-#if defined(AVIF_ENABLE_EXPERIMENTAL_MINI)
-        if (ftypSeen && !needsMini && miniSeen) {
-            // The 'mini' box should be ignored if there is no 'mif3' brand, but libavif allows reading them in any order.
+#if defined(AVIF_ENABLE_EXPERIMENTAL_METAV1)
+        if (ftypSeen && !needsMetaV1 && metaV1Seen) {
+            // The 'meta' box with version 1 box should be ignored if there is no 'mif3' brand, but libavif allows reading them in any order.
```
wantehchang, line 4132: Nit: Delete "box" after "version 1".

src/write.c
```diff
@@ -2400,129 +2360,230 @@ static avifResult avifEncoderWriteMinimizedImageBox(avifEncoder * encoder, avifR
...
+                                       : image->yuvFormat == AVIF_PIXEL_FORMAT_YUV422 ? 2
+                                                                                      : 3;
+
+    const avifColorPrimaries defaultColorPrimaries = hasIcc ? AVIF_COLOR_PRIMARIES_UNSPECIFIED : AVIF_COLOR_PRIMARIES_BT709;
+    const avifTransferCharacteristics defaultTransferCharacteristics = hasIcc ? AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED
+                                                                              : AVIF_TRANSFER_CHARACTERISTICS_SRGB;
+    const avifMatrixCoefficients defaultMatrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601;
+    const avifBool hasExplicitCicp = image->colorPrimaries != defaultColorPrimaries ||
+                                     image->transferCharacteristics != defaultTransferCharacteristics ||
+                                     image->matrixCoefficients != defaultMatrixCoefficients;
```
wantehchang, line 2378: Nit: Comparing with the spec in the proposal, I think the last comparison (line 2378) should be:

```
(image->yuvFormat != AVIF_PIXEL_FORMAT_YUV400 && image->matrixCoefficients != defaultMatrixCoefficients)
```

tests/gtest/avifmetav1test.cc
```diff
@@ -59,10 +59,10 @@ TEST_P(AvifMetaBoxV1Test, SimpleOpaque) {
   }
   image->transformFlags = create_transform_flags;
   if (create_transform_flags & AVIF_TRANSFORM_IROT) {
-    image->irot.angle = 2;
+    image->irot.angle = 1;
   }
   if (create_transform_flags & AVIF_TRANSFORM_IMIR) {
-    image->imir.axis = 1;
+    image->imir.axis = 0;
```
wantehchang, line 65: I am just curious as to why you changed `image->irot.angle` from 2 to 1 and changed `image->imir.axis` from 1 to 0.

wantehchang (approved):
LGTM. Thanks a lot for updating the prototype to the latest proposal.

I agree with how you handled `isHorizontallyCentered` and `isVerticallyCentered` in both src/read.c and src/write.c. We can adjust it if other people have suggestions.

You can merge the pull request after addressing the comments.

wantehchang (commented):
LGTM!

src/write.c
```diff
@@ -2372,7 +2372,8 @@ static avifResult avifEncoderWriteMetaBoxV1(avifEncoder * encoder, avifRWStream
     const avifColorPrimaries defaultColorPrimaries = hasIcc ? AVIF_COLOR_PRIMARIES_UNSPECIFIED : AVIF_COLOR_PRIMARIES_BT709;
     const avifTransferCharacteristics defaultTransferCharacteristics = hasIcc ? AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED
                                                                               : AVIF_TRANSFER_CHARACTERISTICS_SRGB;
-    const avifMatrixCoefficients defaultMatrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601;
+    const avifMatrixCoefficients defaultMatrixCoefficients =
+        image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400 ? AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED : AVIF_MATRIX_COEFFICIENTS_BT601;
```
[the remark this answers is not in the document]
y-guyon: > Suppose the user sets `colour_primaries=1`, `transfer_characteristics=13`, `matrix_coefficients=6` for a monochrome image with no ICC profile

This is forbidden:

https://github.com/AOMediaCodec/libavif/blob/b1b0c54a3413b0029cc2f879943cc60a99ff1371/src/write.c#L2443-L2447

tests/gtest/avifmetav1test.cc
y-guyon: I do not think this is an issue. I believe `irot`/`imir` are a suboptimal way to represent the orientation. Wasting another bit in SlimHEIF to carry that pattern over is not worth it in my opinion.

Alternatively we could just forbid `irot`/`imir` combinations that are not part of the Exif->`irot`/`imir` mapping defined in the spec draft. I am fine with keeping the current implementation.

wantehchang: I just realized that `flags` in FullBox is 24 bits, not 32 bits. So we have already used up all 24 bits in `flags`.

Your answer to my other comment also made me think that our encoder could forbid the `irot/imir` combinations that do not "roundtrip".

PR #2230, opened 2024-06-26
Update the URL for the cargo-c v0.10.0 bug

wantehchang:
I filed a bug report for cargo-c v0.10.0. Reference the bug report in the comment.


jzern (approved):

wantehchang:
Thank you for the review. The cargo-c v0.10.0 bug has just been fixed. So I will just upgrade to the commit of the bug fix instead. See https://github.com/AOMediaCodec/libavif/pull/2228.

PR #2231, opened 2024-06-26
Do not use GIT_SHALLOW ON with a hash GIT_TAG

wantehchang:


cmake/Modules/LocalRav1e.cmake
```diff
@@ -51,7 +51,7 @@ else()
             cargoc
             GIT_REPOSITORY https://github.com/lu-zero/cargo-c.git
             GIT_TAG "${AVIF_LOCAL_CARGOC_GIT_TAG}"
-            GIT_SHALLOW ON
+            GIT_SHALLOW OFF
```
wantehchang, line 54: The build fails after the pull request https://github.com/AOMediaCodec/libavif/pull/2228 was merged. The error messages are:

```
[1/9] Creating directories for 'cargoc-populate'
[1/9] Performing download step (git clone) for 'cargoc-populate'
Cloning into 'cargoc-src'...
fatal: unable to read tree (a128f7763e81ac688164d9f885f89770a53a0ada)
CMake Error at cargoc-subbuild/cargoc-populate-prefix/tmp/cargoc-populate-gitclone.cmake:40 (message):
  Failed to checkout tag: 'a128f7763e81ac688164d9f885f89770a53a0ada'
```

My theory for why this did not fail when the pull request https://github.com/AOMediaCodec/libavif/pull/2228 was tested is that the git hash was the tip of the tree at that time. Now there is a new commit after this git hash:
https://github.com/lu-zero/cargo-c/commits/master/

wantehchang: I found this documented in https://cmake.org/cmake/help/latest/module/ExternalProject.html#git:

> If `GIT_SHALLOW` is enabled then `GIT_TAG` works only with branch names and tags. A commit hash is not allowed.

jzern (approved):

PR #2235, opened 2024-06-27
Update gainmap metadata to current ISO 21496-1 draft

ccameron-chromium:
The ISO blob now includes a `minimum_version` and `writer_version`, the flags have been shuffled a bit, and there is no more concept of common denominator.


tests/gtest/avifgainmaptest.cc
```diff
@@ -776,7 +776,7 @@ TEST(GainMapTest, NoGainMap) {
   EXPECT_EQ(decoded->gainMap, nullptr);
 }
 
-TEST(GainMapTest, DecodeGainMapGrid) {
+TEST(GainMapTest, DISABLED_DecodeGainMapGrid) {
```
wantehchang, line 779: Please add (TODO?) comments to explain why these tests are disabled.

src/read.c
```diff
@@ -1968,52 +1968,36 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
...
+    AVIF_CHECK(avifROStreamReadU16(&s, &minimumVersion));
+    if (minimumVersion != 0) {
+        avifDiagnosticsPrintf(diag, "Box[tmap] has unsupported minimum version [%u]", minimumVersion);
+        return AVIF_FALSE;
+    }
+    uint16_t writerVersion;
+    AVIF_CHECK(avifROStreamReadU16(&s, &writerVersion));
+
     uint8_t flags;
     AVIF_CHECK(avifROStreamRead(&s, &flags, 1)); // unsigned int(8) flags;
```
y-guyon, line 1981: `avifROStreamReadBits()` can be used directly if more convenient.

Also `unsigned int(8) flags` is not part of the spec (unless there is a more recent draft than the one I have).

src/read.c
```diff
@@ -1968,52 +1968,36 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
         return AVIF_FALSE;
     }
 
+    uint16_t minimumVersion;
+    AVIF_CHECK(avifROStreamReadU16(&s, &minimumVersion));
```
y-guyon, line 1972: If possible it would be good to keep original names verbatim, for example:

```c
AVIF_CHECK(avifROStreamReadU16(&s, &minimumVersion)); // unsigned int(16) minimum_version;
```

It makes it easier to find something from the spec. Also for consistency with the rest of the file.

y-guyon (approved):
LGTM. Feel free to submit as is (with the tests enabled though) and I can send another PR with my suggestions.

src/read.c
ccameron-chromium: Updated to use the bit stream and names for the bits.

src/read.c
ccameron-chromium: Done.

tests/gtest/avifgainmaptest.cc
ccameron-chromium: Added a TODO for the issues.

src/read.c
```diff
@@ -1968,52 +1968,41 @@ static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, cons
...
+    }
+    uint16_t writerVersion;
+    AVIF_CHECK(avifROStreamReadU16(&s, &writerVersion)); // unsigned int(16) writer_version;
+
+    uint32_t isMultichannel;
+    AVIF_CHECK(avifROStreamReadBits(&s, &isMultichannel, 1)); // unsigned int(1) is_multichannel;
+    const uint8_t channelCount = isMultichannel ? 3 : 1;
+
+    uint32_t useBaseColorSpace;
+    AVIF_CHECK(avifROStreamReadBits(&s, &useBaseColorSpace, 1)); // unsigned int(1) use_base_colour_space;
```
wantehchang, line 1985: Should we store the value in `metadata->useBaseColorSpace`?

ccameron-chromium: Yes (can't read it directly because of the type, but stored the result).

Also updated all other fields to be commented.

wantehchang (approved):
LGTM. Thanks!

@vrabaud FYI.

PR #2237, opened 2024-06-29
avifRWStreamWriteBits: Check `v` for valid range

wantehchang:
Make sure the input parameter `v` can be represented in `bitCount` bits. This was checked by an assertion. Replace the assertion by an error return.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -732,7 +732,7 @@ avifResult avifRWStreamWriteU32(avifRWStream * stream, uint32_t v);
 avifResult avifRWStreamWriteU64(avifRWStream * stream, uint64_t v);
 avifResult avifRWStreamWriteZeros(avifRWStream * stream, size_t byteCount);
 // The following functions can write non-aligned bits.
-avifResult avifRWStreamWriteBits(avifRWStream * stream, uint32_t v, size_t bitCount);
+AVIF_NODISCARD avifResult avifRWStreamWriteBits(avifRWStream * stream, uint32_t v, size_t bitCount);
 avifResult avifRWStreamWriteVarInt(avifRWStream * stream, uint32_t v);
 
 // This is to make it clear that the box size is currently unknown, and will be determined later (with a call to avifRWStreamFinishBox)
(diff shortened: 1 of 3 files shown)


src/stream.c
```diff
@@ -479,7 +479,9 @@ avifResult avifRWStreamWriteZeros(avifRWStream * stream, size_t byteCount)
 
 avifResult avifRWStreamWriteBits(avifRWStream * stream, uint32_t v, size_t bitCount)
 {
-    assert(((uint64_t)v >> bitCount) == 0); // (uint32_t >> 32 is undefined behavior)
+    if (bitCount < 32 && (v >> bitCount) != 0) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
```
wantehchang, line 484: If `bitCount >= 32`, then all `uint32_t` values can be represented, so no check is needed.

Should we disallow `bitCount == 0`?

vrabaud: The standard probably allows it no? Let's keep it for now.

vrabaud (approved):

src/stream.c
y-guyon: Optional: Use `AVIF_ASSERT_OR_RETURN`. To me `AVIF_RESULT_INVALID_ARGUMENT` carries the meaning of an error from the user, not an implementation bug.

> Should we disallow `bitCount == 0`?

I do not mind enforcing it or not.

y-guyon (approved):
Thank you for writing a test.

include/avif/internal.h
```diff
@@ -732,7 +732,7 @@ avifResult avifRWStreamWriteU32(avifRWStream * stream, uint32_t v);
 avifResult avifRWStreamWriteU64(avifRWStream * stream, uint64_t v);
 avifResult avifRWStreamWriteZeros(avifRWStream * stream, size_t byteCount);
 // The following functions can write non-aligned bits.
-avifResult avifRWStreamWriteBits(avifRWStream * stream, uint32_t v, size_t bitCount);
+AVIF_NODISCARD avifResult avifRWStreamWriteBits(avifRWStream * stream, uint32_t v, size_t bitCount);
```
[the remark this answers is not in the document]
vrabaud: I thought there was a compiler restriction I had missed because there are other `AVIF_NODISCARD avifResult` in the file but no. Indeed, `avifResult` is enough.

src/stream.c
wantehchang: Thanks for the suggestion. I found that `AVIF_ASSERT_OR_RETURN` makes the new test call `abort()` in the Debug build, so we can't use `AVIF_ASSERT_OR_RETURN` with a test for error checking. I also found that `avifRWStreamWriteVarInt()` returns `AVIF_RESULT_INVALID_PARAMETER` for a similar error. So in the end I just converted the if statement to use `AVIF_CHECKERR`.

wantehchang (commented):
Thanks for the review!

PR #2242, opened 2024-07-03
Put AVIF_NODISCARD after AVIF_API

wantehchang:
It looks more consistent to have AVIF_API at the beginning of a function declaration. It also looks nicer to have AVIF_NODISCARD right before the function return type.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -11,6 +11,14 @@
 extern "C" {
 #endif
 
+// Define the internal macro AVIF_USE_NODISCARD if AVIF_NODISCARD will be defined as [[nodiscard]].
+// In this case, also use the standard [[...]] attribute syntax for GCC's visibility("default")
+// attribute to avoid compilation errors.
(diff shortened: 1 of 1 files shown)


wantehchang:
Here is a small test that shows the compilation errors I referred to in the comment before the `AVIF_USE_NODISCARD` macro definition.

```
$ cat foo.c
__attribute__((visibility("default"))) [[nodiscard]] int Add(int x, int y) {
  return x + y;
}
$ gcc-14 -Wall -Wextra -c foo.c
foo.c:1:1: warning: ‘nodiscard’ attribute ignored [-Wattributes]
    1 | __attribute__((visibility("default"))) [[nodiscard]] int Add(int x, int y) {
      | ^~~~~~~~~~~~~
foo.c:1:54: error: expected identifier or ‘(’ before ‘int’
    1 | __attribute__((visibility("default"))) [[nodiscard]] int Add(int x, int y) {
      |                                                      ^~~
$ cat bar.c
[[gnu::visibility("default")]] [[nodiscard]] int Add(int x, int y) {
  return x + y;
}
$ gcc-14 -Wall -Wextra -c bar.c
$
```

vrabaud (approved):

PR #2244, opened 2024-07-09
Remove old "Add any new members after this line."

wantehchang:
We only need "Add any new members after this line." after the latest version.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1287,7 +1287,7 @@ typedef struct avifDecoder
     // Internals used by the decoder
     struct avifDecoderData * data;
 
-    // Version 1.0.0 ends here. Add any new members after this line.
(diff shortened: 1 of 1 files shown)


include/avif/avif.h
```diff
@@ -1287,7 +1287,7 @@ typedef struct avifDecoder
     // Internals used by the decoder
     struct avifDecoderData * data;
 
-    // Version 1.0.0 ends here. Add any new members after this line.
+    // Version 1.0.0 ends here.
```
wantehchang, line 1290: Line 1297 below says:

```
    // Version 1.1.0 ends here. Add any new members after this line.
```

So this line should not also say "Add any new members after this line."

vrabaud (approved):

PR #2245, opened 2024-07-09
Use proper API for figuring out configPropName

roticv:
There is avifGetCodecType and avifGetConfigurationPropertyName and so there is no need for manual memcpy. This change will make supporting future codecs in libavif easier.


vrabaud:
And the old version was also not using an `#if defined(AVIF_CODEC_AVM)`. Thx !

wantehchang:
> And the old version was also not using an `#if defined(AVIF_CODEC_AVM)`. Thx !

After this pull request was merged, the uses of "av02" and "av2C" are all inside `#if defined(AVIF_CODEC_AVM)`. (Note: the only exception, which tests the wrong macro, is fixed in https://github.com/AOMediaCodec/libavif/pull/2247.)

PR #2246, opened 2024-07-09
Fix __cplusplus version

vrabaud:
cf https://en.cppreference.com/w/cpp/preprocessor/replace#Predefined_macros


wantehchang (approved):
LGTM. Thanks!

PR #2248, opened 2024-07-10
Fix errors when lib built as C++

vrabaud:


src/reformat.c
```diff
@@ -502,7 +502,7 @@ static avifBool avifCreateYUVToRGBLookUpTables(float ** unormFloatTableY, float
     const size_t cpCount = (size_t)1 << depth;
 
     assert(unormFloatTableY);
-    *unormFloatTableY = avifAlloc(cpCount * sizeof(**unormFloatTableY));
+    *unormFloatTableY = (float *)avifAlloc(cpCount * sizeof(**unormFloatTableY));
```
wantehchang, line 505: Optional: When there is a cast for the return value, I think using the `float` type in `sizeof` is more readable:

```
    *unormFloatTableY = (float *)avifAlloc(cpCount * sizeof(float));
```

Also at line 516.

wantehchang (approved):
LGTM. Thanks!

src/reformat.c
vrabaud: agreed. Done.

src/reformat.c
```diff
@@ -502,7 +502,7 @@ static avifBool avifCreateYUVToRGBLookUpTables(float ** unormFloatTableY, float
     const size_t cpCount = (size_t)1 << depth;
 
     assert(unormFloatTableY);
-    *unormFloatTableY = (float *)avifAlloc(cpCount * sizeof(**unormFloatTableY));
+    *unormFloatTableY = (float *)avifAlloc(cpCount * sizeof(float));
```
wantehchang, line 505: Please make the same change to line 516. Thanks!

wantehchang (changes requested):

src/reformat.c
vrabaud: DOne

wantehchang (approved):
LGTM. Thanks!

PR #2249, opened 2024-07-10
Rename the AVIF_ENABLE_WUNUSED_RESULT cmake option

wantehchang:
Rename it AVIF_ENABLE_NODISCARD, which matches the corresponding C macro's name.


CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ set(LIBRARY_SOVERSION ${LIBRARY_VERSION_MAJOR})
 option(BUILD_SHARED_LIBS "Build shared avif library" ON)
 
 option(AVIF_ENABLE_WERROR "Treat all compiler warnings as errors" OFF)
-option(AVIF_ENABLE_WUNUSED_RESULT "Add [[nodiscard]] to some functions. CMake must be at least 3.21 to force C23" OFF)
+option(AVIF_ENABLE_NODISCARD "Add [[nodiscard]] to some functions. CMake must be at least 3.21 to force C23" OFF)
```
wantehchang, line 49: The `AVIF_ENABLE_WUNUSED_RESULT` option name is also confusing because it seems to imply the option enables the GCC `-Wunused-result` option, which has a different meaning:
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wunused-result

vrabaud (approved):

PR #2251, opened 2024-07-11
Fix default qualityAlpha calculation bug

wantehchang:
When calculating qualityAlpha from minQuantizerAlpha and maxQuantizerAlpha, use their values rather than whether or not they are set (a boolean value).

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -2130,7 +2130,7 @@ int main(int argc, char * argv[])
                 if (fileSettings->minQuantizerAlpha.set) {
                     assert(fileSettings->maxQuantizerAlpha.set);
                     if (!fileSettings->qualityAlpha.set) {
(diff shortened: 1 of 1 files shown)


apps/avifenc.c
```diff
@@ -2130,7 +2130,7 @@ int main(int argc, char * argv[])
                 if (fileSettings->minQuantizerAlpha.set) {
                     assert(fileSettings->maxQuantizerAlpha.set);
                     if (!fileSettings->qualityAlpha.set) {
-                        const int quantizerAlpha = (fileSettings->minQuantizerAlpha.set + fileSettings->maxQuantizerAlpha.set) / 2;
+                        const int quantizerAlpha = (fileSettings->minQuantizerAlpha.value + fileSettings->maxQuantizerAlpha.value) / 2;
```
wantehchang, line 2133: Please compare this with line 2117 above.

```
                        const int quantizer = (fileSettings->minQuantizer.value + fileSettings->maxQuantizer.value) / 2;
```

vrabaud (approved):

PR #2252, opened 2024-07-11
Add the meson version in the cache keys

vrabaud:


.github/actions/cache/action.yml
```diff
@@ -25,6 +28,16 @@ runs:
...
       shell: bash
+    - name: Generate meson cache key
+      if: ${{ inputs.use-meson == 'true' }}
+      run: |
+        echo "MESON_KEY=`meson --version`" >> $GITHUB_ENV
+      shell: bash
+    - name: Generate empty meson cache key
+      if: ${{ inputs.use-meson != 'true' }}
+      run: |
+        echo "MESON_KEY=\"0\"" >> $GITHUB_ENV
```
wantehchang: Optional: The quotes around 0 don't seem necessary, because `meson --version` cannot be 0. But the quotes do make the format different. I wonder if we can just leave the right-hand side empty:

```
      run: |
        echo "MESON_KEY=" >> $GITHUB_ENV
```

.github/actions/cache/action.yml
```diff
@@ -1,12 +1,15 @@
...
-    default: false
   extra-key:
     description: 'Extra cache key to use in the cache name. Useful when several caches are used in one workflow.'
     default: ''
+  use-meson:
+    description: 'Whether meson is used'
+    default: false
+  use-rust:
+    description: 'Whether rust is used'
+    default: false
```
wantehchang, line 12: Could you check .github/workflows/ci-mingw.yml? It also uses `.github/actions/cache`:

```
    - uses: ./.github/actions/cache
      with:
        extra-key: ${{ matrix.sys }}
```

Note: I used this `find` command to find the files that use `.github/actions/cache`:

```
find .github -name "*" -a -type f | xargs grep github/actions/cache
```

.github/actions/cache/action.yml
```diff
@@ -25,6 +28,16 @@ runs:
...
       shell: bash
+    - name: Generate meson cache key
+      if: ${{ inputs.use-meson == 'true' }}
+      run: |
+        echo "MESON_KEY=`meson --version`" >> $GITHUB_ENV
+      shell: bash
+    - name: Generate empty meson cache key
+      if: ${{ inputs.use-meson != 'true' }}
+      run: |
+        echo "MESON_KEY=\"none\"" >> $GITHUB_ENV
```
wantehchang, line 39: Nit: Are the quotes around `none` necessary?

In line 34, we don't quote the output of `meson --version`.

Also, it would be more readable to include `meson-` in the value of `MESON_KEY`:

```
        echo "MESON_KEY=\"meson-none\"" >> $GITHUB_ENV
```

.github/actions/cache/action.yml
```diff
@@ -44,9 +57,9 @@ runs:
...
       with:
         path: ext
-        key: ext-${{ runner.os }}-${{ runner.arch }}-${{ github.job }}-${{ inputs.extra-key }}-${{ hashFiles('.github/action/**', steps.workflow-info.outputs.path, 'ext/*.cmd', 'ext/*.sh') }}
+        key: ext-${{ runner.os }}-${{ runner.arch }}-${{ github.job }}-${{ inputs.extra-key }}-${{ hashFiles('.github/action/**', steps.workflow-info.outputs.path, 'ext/*.cmd', 'ext/*.sh') }}-${{ env.MESON_KEY }}
     - name: Cache external dependencies in build/_deps
       uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
       with:
         path: build/_deps
-        key: deps-${{ runner.os }}-${{ runner.arch }}-${{ github.job }}-${{ inputs.extra-key }}-${{ hashFiles('.github/action/**', steps.workflow-info.outputs.path, 'CMakeLists.txt', 'cmake/Modules/*') }}
+        key: deps-${{ runner.os }}-${{ runner.arch }}-${{ github.job }}-${{ inputs.extra-key }}-${{ hashFiles('.github/action/**', steps.workflow-info.outputs.path, 'CMakeLists.txt', 'cmake/Modules/*') }}-${{ env.MESON_KEY }}
```
wantehchang, line 65: Nit/optional: It looks nicer to let the `hashFiles` output remain the last component of the key.

.github/actions/cache/action.yml
```diff
@@ -25,6 +28,16 @@ runs:
         echo "Current workflow path is: ${MIDDLE}"
         echo "path=${MIDDLE}" >> "$GITHUB_OUTPUT"
       shell: bash
+    - name: Generate meson cache key
+      if: ${{ inputs.use-meson == 'true' }}
+      run: |
+        echo "MESON_KEY=`meson --version`" >> $GITHUB_ENV
```
wantehchang, line 34: I wonder if we should include `meson-` in the value of `MESON_KEY` to make it more reasable:

```
        echo "MESON_KEY=meson-`meson --version`" >> $GITHUB_ENV
```

vrabaud: Agreed.

vrabaud:
And we get:
https://github.com/AOMediaCodec/libavif/actions/runs/9891928986/job/27323410187?pr=2252
```
Cache not found for input keys: deps-Linux-X64-build-shared-local--meson-1.3.2-37182331cb2576a444df073acd2f13521500d04795c0df3c9668991c0322b92c
```
https://github.com/AOMediaCodec/libavif/actions/runs/9891928990/job/27323410157?pr=2252
```
Cache not found for input keys: deps-Windows-X64-build-windows-installed--meson-none-7032e832ebbacac7781b6a8cd4d6450e83d6df7168abb4b3f035c69a8f21fbcf
```

.github/actions/cache/action.yml
wantehchang: Please ignore this comment. I didn't realize this comment was still posted. I thought it wasn't saved. Sorry!

wantehchang (approved):
LGTM. Thanks!

PR #2257, opened 2024-07-11
Revert "Put AVIF_NODISCARD after AVIF_API"

wantehchang:
This reverts commit 508f00f7904d1c21ebb21ca5acbc9f7461e47951.

The commit broke compilation with clang-cl on Windows:
  'nodiscard' attribute cannot be applied to types

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -11,14 +11,6 @@
 extern "C" {
 #endif
 
-// Define the internal macro AVIF_USE_NODISCARD if AVIF_NODISCARD will be defined as [[nodiscard]].
-// In this case, also use the standard [[...]] attribute syntax for GCC's visibility("default")
-// attribute to avoid compilation errors.
-#if defined(AVIF_ENABLE_NODISCARD) || (defined(__cplusplus) && __cplusplus >= 201703L) || \
-    (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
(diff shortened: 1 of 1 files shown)


vrabaud (approved):

wantehchang:
Here is a simple test that demonstrates the clang-cl compilation error and a potential fix:

```
C:\src\tmp>type foo.cc
struct Foo {
  int x;
  int y;
};

__declspec(dllimport) [[nodiscard]] Foo* CreateFoo(int x, int y);

C:\src\tmp>clang-cl /W4 /c foo.cc
foo.cc(6,25): error: 'nodiscard' attribute cannot be applied to types
    6 | __declspec(dllimport) [[nodiscard]] Foo* CreateFoo(int x, int y);
      |                         ^
1 error generated.

C:\src\tmp>type bar.cc
struct Foo {
  int x;
  int y;
};

[[gnu::dllimport]] [[nodiscard]] Foo* CreateFoo(int x, int y);

C:\src\tmp>clang-cl /W4 /c bar.cc

C:\src\tmp>
```

I found that clang-cl and clang accept `[[gnu::dllimport]]`. Before we re-attempt this, it would be good to set up a CI workflow for clang-cl (and perhaps also clang) on Windows.

vrabaud:
clang-cl CI added in https://github.com/AOMediaCodec/libavif/pull/2270

PR #2259, opened 2024-07-11
Fix ci-android-emulator-tests

vrabaud:
ci-android-emulator-tests actually ran on macOS but with the arm64 only architecture and the emulation happening, it is easier to stick to x86 and therefore linux.


vrabaud:
vigneshvg@, can you please advise on the failures? Thx. Setting `api-level >= 30`  triggers failures.

vigneshvg:
I have created another PR to address this issue: https://github.com/AOMediaCodec/libavif/pull/2283

I will close this one.

PR #2262, opened 2024-07-12
Remove the docker-based test

vrabaud:
Remove the docker-based test


tests/docker/build.sh
```diff
@@ -19,7 +19,7 @@ set -e
 
 # build env
 apt update
-DEBIAN_FRONTEND="noninteractive" apt install -y build-essential libjpeg-dev libpng-dev libssl-dev ninja-build cmake pkg-config git perl vim meson cargo cargo-c nasm
+DEBIAN_FRONTEND="noninteractive" apt install -y build-essential cargo cargo-c cmake git libjpeg-dev libpng-dev libssl-dev libyuv-dev meson nasm ninja-build perl pkg-config vim
```
wantehchang, line 22: The cmake command at line 74 should be updated to say `-DAVIF_CODEC_AOM=SYSTEM`, etc. We can do that in a separate pull request.

wantehchang (approved):

tests/docker/build.sh
vrabaud: Indeed. BTW, what is the point of that file: shouldn't we compile the dependencies in the libavif work tree like we do now and simply install libavif? We could also install the packages for the dependencies.

wantehchang: This tests building and installing the AV1 codecs and libavif from source code in a brand new Linux computer. We can delete this test.

vrabaud:
Now that we have a good CI, we might as well use it. We'll revive the docker if there is ever a need for it.

PR #2265, opened 2024-07-12
fix: do not disable CONFIG_RUNTIME_CPU_DETECT in cross-arch aom builds

fdintino:
Currently, LocalAom.cmake disables CONFIG_RUNTIME_CPU_DETECT for cross-architecture builds. This doesn't make sense, and I can't recall why I added it. Without runtime cpu detection, a cross-architecture macOS arm64 build gets compiled with SVE and I8MM Neon intrinsics, and since those aren't supported on apple silicon it's segfaulting with illegal instruction errors (see https://github.com/fdintino/pillow-avif-plugin/issues/59).


wantehchang (commented):
Frankie: Thanks for the fix. Could you please add an entry to CHANGELOG.md under `### Changed since 1.1.0`?

cmake/Modules/LocalAom.cmake
```diff
@@ -91,9 +91,6 @@ else()
     set(ENABLE_TESTDATA 0 CACHE INTERNAL "")
     set(ENABLE_TESTS 0 CACHE INTERNAL "")
     set(ENABLE_TOOLS 0 CACHE INTERNAL "")
-    if(NOT CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL CMAKE_SYSTEM_PROCESSOR)
```
jzern, line 94: I wonder if this should have been `if(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL CMAKE_SYSTEM_PROCESSOR)`; checking if the host matches the target system. Either way it's a little unsafe if the library or binaries are distributed to other systems which may not have the same baseline as the build host.

jzern (approved):

wantehchang (approved):

fdintino:
I remember now why I was disabling cpu detection here: aom didn't have cpu detection for AOM until 3.7.0, and I started working on these cmake changes prior to its release.

PR #2266, opened 2024-07-12
Reapply "Put AVIF_NODISCARD after AVIF_API"

wantehchang:
This reverts commit 53fe96de1b7707fe830d213828cfaebd71793a8a.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -11,6 +11,14 @@
 extern "C" {
 #endif
 
+// Define the internal macro AVIF_USE_NODISCARD if AVIF_NODISCARD will be defined as [[nodiscard]].
+// In this case, also use the standard [[...]] attribute syntax for GCC's visibility("default")
+// attribute to avoid compilation errors.
+#if defined(AVIF_ENABLE_NODISCARD) || (defined(__cplusplus) && __cplusplus >= 201703L) || \
+    (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
+#define AVIF_USE_NODISCARD 1
+#endif
+
 // ---------------------------------------------------------------------------
 // Export macros
 
@@ -31,7 +39,11 @@ extern "C" {
 #define AVIF_HELPER_EXPORT __declspec(dllexport)
 #define AVIF_HELPER_IMPORT __declspec(dllimport)
 #elif defined(__GNUC__) && __GNUC__ >= 4
+#if defined(AVIF_USE_NODISCARD)
+#define AVIF_HELPER_EXPORT [[gnu::visibility("default")]]
+#else
 #define AVIF_HELPER_EXPORT __attribute__((visibility("default")))
+#endif
 #define AVIF_HELPER_IMPORT
 #else
 #define AVIF_HELPER_EXPORT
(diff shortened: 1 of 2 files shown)


wantehchang (commented):
The first commit in the pull request is pure relanding. The second commit is the clang-cl compilation error fix. I am testing it in Chrome at https://chromium-review.googlesource.com/c/chromium/src/+/5704377.

include/avif/avif.h
```diff
@@ -28,10 +36,19 @@ extern "C" {
 // For static build, AVIF_API is always defined as nothing.
 
 #if defined(_WIN32)
+#if defined(AVIF_USE_NODISCARD) && (defined(__GNUC__) || defined(__clang__))
+#define AVIF_HELPER_EXPORT [[gnu::dllexport]]
+#define AVIF_HELPER_IMPORT [[gnu::dllimport]]
```
wantehchang, line 41: These two GCC attributes are document at https://gcc.gnu.org/onlinedocs/gcc/Microsoft-Windows-Function-Attributes.html.

jzern: Aren't these considered valid only with C23 or gnu17?

https://gcc.gnu.org/onlinedocs/gcc/Attribute-Syntax.html#Attribute-Syntax

wantehchang: The definition of the `AVIF_USE_NODISCARD` macro at line 19 detects C++17 or C23:

```
#if defined(AVIF_ENABLE_NODISCARD) || (defined(__cplusplus) && __cplusplus >= 201703L) || \
    (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
#define AVIF_USE_NODISCARD 1
#endif
```

C++17 is the first C++ version that supports the `nodiscard` attribute.

C23 is the first C version that supports the standard `[[...]]` attribute syntax.

jzern: I'm sorry I missed the use of `AVIF_USE_NODISCARD` in this section and only read the second half.

jzern (approved):

vrabaud (approved):

wantehchang (commented):
@vrabaud Vincent: Would you like to take this pull request in libavif v1.1.1?

The net change from v1.1.0 is only the second commit in this pull request, so it is not a big change to v1.1.0. But we will need to import it into Chrome as testing before the v1.1.1 release.

vrabaud:
I think it is a good change, let's put it in. I am trying to get a clang-cl CI

wantehchang (commented):
Vincent: Thanks! I also need to update the changelog entry.

include/avif/avif.h
```diff
@@ -11,6 +11,14 @@
 extern "C" {
 #endif
 
+// Define the internal macro AVIF_USE_NODISCARD if AVIF_NODISCARD will be defined as [[nodiscard]].
+// In this case, also use the standard [[...]] attribute syntax for GCC's visibility("default")
+// attribute and MSVC's dllexport/dllimport attributes to avoid compilation errors.
+#if defined(AVIF_ENABLE_NODISCARD) || (defined(__cplusplus) && __cplusplus >= 201703L) || \
+    (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
+#define AVIF_USE_NODISCARD 1
```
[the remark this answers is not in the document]
jzern: That makes sense unless the idea was to use this in internal functions too.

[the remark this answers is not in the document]
wantehchang: Done

jzern (approved):

PR #2268, opened 2024-07-15
Read byte-unaligned infe in meta v1

y-guyon:
avifROStreamRead() requires byte alignment.


wantehchang (approved):
LGTM.

@vrabaud Vincent: I think we should merge this bug fix duirng the code review for the v1.1.1 patch release.

vrabaud:
I guess that deserves a CHANGELOG entry though?

wantehchang:
Yes, this fix should be documented in the changelog. We can point out that it fixes a bug in the experimental feature `AVIF_ENABLE_EXPERIMENTAL_METAV1`.

Issue #2264, opened 2024-07-12
v1.1.0 CMake config leaks shared library dependencies into consumers


jbeich:
Regressed by #1778. Found via sdl2_image and kf5-kimageformats. Affects CMake-based consumers on [FreeBSD](https://wiki.freebsd.org/WarnerLosh/UsrLocal#Library_paths), other BSDs and maybe even Nix due to using different prefix for each package. To reproduce install libavif and its dependencies outside of `/usr` and `/usr/local` thus not part of default C/C++ compiler search path.

```
$ pkg-config --libs dav1d
-L/custom_prefix/lib -ldav1d
$ pkg-config --libs libyuv
-L/custom_prefix/lib -lyuv
$ pkg-config --libs aom
-L/custom_prefix/lib -laom

$ cat foo.c
int main() {}
$ cat CMakeLists.txt
cmake_minimum_required(VERSION 3.28)
project(foo C)
add_executable(foo foo.c)
find_package(libavif)
target_link_libraries(foo avif)
$ cmake -B /tmp/foo_build
$ cmake --build /tmp/foo_build
[...]
ld: error: unable to find library -lyuv
ld: error: unable to find library -ldav1d
ld: error: unable to find library -laom
cc: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error code 1
```

jbeich:
Also, this is about libavif built against `SYSTEM` dependencies rather than `LOCAL`.

wantehchang:
I think it is caused by this code in the `avif_target_link_library` function in `CMakeLists.txt`:

```
    # The transitive dependency is an export link library if it is a shared
    # library, or if it is a static library in a static build
    get_target_property(target_type ${target} TYPE)
    if(target_type STREQUAL "SHARED_LIBRARY" OR NOT BUILD_SHARED_LIBS)
        target_link_libraries(avif PUBLIC $<INSTALL_INTERFACE:${install_target}>)
    endif()
```

The `target_type STREQUAL "SHARED_LIBRARY"` condition seems incorrect.

wantehchang:
Jan: Could you attach your `libavif.pc` file? Thanks.

jbeich:
libavif.pc looks mostly correct: `-lyuv` without `-L` is a bug (should use libyuv.pc instead) but it works fine as long as libavif and libyuv share prefix.

```
$ pkg-config libavif --libs
-L/tmp/test_prefix/lib -lavif
[...]

PR #2269, opened 2024-07-15
Only INSTALL_INTERFACE for static.

vrabaud:
This fixes https://github.com/AOMediaCodec/libavif/issues/2264

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,8 +156,8 @@ add_library(avif_obj OBJECT)
 add_library(avif)
 
 # Adds <target> to avif_obj's public link libraries for build, and adds
-# the <target> library as an install link library for export if a consumer
-# would need to include that library alongside libavif when linking.
+# the <target> library as an install link library for export in case a consumer
+# needs to include that library alongside libavif when statically linking.
 function(avif_target_link_library target)
     target_link_libraries(avif_obj PUBLIC $<BUILD_INTERFACE:${target}>)
     get_target_property(target_is_local ${target} AVIF_LOCAL)
@@ -168,9 +168,7 @@ function(avif_target_link_library target)
     if(NOT install_target)
         set(install_target ${target})
     endif()
-    # The transitive dependency is an export link library if it is a shared
-    # library, or if it is a static library in a static build
-    get_target_property(target_type ${target} TYPE)
+    # The transitive dependency is an export link library if it is a static library in a static build.
     if(NOT BUILD_SHARED_LIBS)
         target_link_libraries(avif PUBLIC $<INSTALL_INTERFACE:${install_target}>)
     endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -171,7 +171,7 @@ function(avif_target_link_library target)
     # The transitive dependency is an export link library if it is a shared
     # library, or if it is a static library in a static build
     get_target_property(target_type ${target} TYPE)
-    if(target_type STREQUAL "SHARED_LIBRARY" OR NOT BUILD_SHARED_LIBS)
+    if(NOT BUILD_SHARED_LIBS)
```
wantehchang, line 174: 1) Also delete line 173. It sets the `target_type` variable, which is now unused.

2) Please review and update the comments at line 158 and line 171.

It seems that the comment at line 158 is not affected by the fix, although I don't really understand what "for export"in line 159 means. It seems to be in contrast to "for build" in line 158, but if that's the case, then line 159 should read "the <target> library to avif's public link library for install if a consumer". I.e., the two comment lines describe line 162:

```
    target_link_libraries(avif_obj PUBLIC $<BUILD_INTERFACE:${target}>)
```

respectively.

and line 175:

```
        target_link_libraries(avif PUBLIC $<INSTALL_INTERFACE:${install_target}>)
```

The comment at line 171 is affected by this change, so it should be updated. It probably just needs to say we need a transitive dependency only in a static build, or we can simply delete the comment. (Ideally the comment should explain why we do not need a transitive dependency in a shared build.)

vrabaud: I believe the new comments I added are right.

CMakeLists.txt
```diff
@@ -168,10 +168,8 @@ function(avif_target_link_library target)
     if(NOT install_target)
         set(install_target ${target})
     endif()
-    # The transitive dependency is an export link library if it is a shared
-    # library, or if it is a static library in a static build
-    get_target_property(target_type ${target} TYPE)
-    if(target_type STREQUAL "SHARED_LIBRARY" OR NOT BUILD_SHARED_LIBS)
+    # The transitive dependency is an export link library if it is a static library in a static build.
+    if(NOT BUILD_SHARED_LIBS)
```
wantehchang, line 172: I also wanted to note that we can't test this code (lines 172-174) right now, because in a static build, the cmake config files are not installed.

I tried to test this code last week but could not make it work. Another difficulty is that in a static build, the library we install is `avif_static`, not `avif`:

```
    if(BUILD_SHARED_LIBS)
        set(LIBAVIF_INSTALL_TARGET avif)
    else()
        set(LIBAVIF_INSTALL_TARGET avif_static)
    endif()
```

where `avif_static` is the merged archive library:

```
    merge_static_libs(avif_static avif)
```

I am wondering if `avif_static` does not inherit the link libraries of `avif` that we set here.

CMakeLists.txt
vrabaud: Done

wantehchang (approved):
LGTM. Thanks!

CMakeLists.txt
vrabaud: I am playing with an installed static CI right now to debug this.

PR #2270, opened 2024-07-15
Add clang-cl CI

vrabaud:
For default, CMake gives:
```
-- The C compiler identification is MSVC 19.40.33811.0
```
For clang-cl:
```
-- The C compiler identification is Clang 18.1.6 with MSVC-like command-line
```


.github/workflows/ci-windows-installed.yml
```diff
@@ -28,6 +30,13 @@ jobs:
         codec-aom: 'SYSTEM'
         codec-dav1d: 'SYSTEM'
         codec-rav1e: 'SYSTEM'
+        extra-cache-key: ${{ matrix.compiler }}
+    - name: Leave compiler as default
+      if: matrix.compiler == 'default'
+      run: echo "CMAKE_AVIF_FLAGS=\"\"" >> $GITHUB_ENV
+    - name: Set clang-cl as compiler
+      if: matrix.compiler == 'clang-cl'
+      run: echo "CMAKE_AVIF_FLAGS=-DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl" >> "$GITHUB_ENV"
```
wantehchang, line 39: Nit: I suggest avoiding the `CMAKE_` prefix because cmake owns that namespace. For example, we can name this variable `AVIF_CMAKE_FLAGS`.

.github/workflows/ci-windows-installed.yml
```diff
@@ -20,6 +20,8 @@ jobs:
     runs-on: windows-latest
     strategy:
       fail-fast: false
+      matrix:
+        compiler: [default, clang-cl]
```
wantehchang, line 24: Another option is to use the `env` key:

https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow

```
      env:
        CC: clang-cl
        CXX: clang-cl
```

.github/workflows/ci-windows-installed.yml
vrabaud: Done

.github/workflows/ci-windows-installed.yml
vrabaud: I wanted to do so except that I do not know what to put for MSVC. And setting env variables conditionally is roughly the same as what I did.

vrabaud:
There are many warnings I disabled, they seem harmless. We can fix them one by one later.

vrabaud:
https://github.com/AOMediaCodec/libavif/pull/2275 needs to be merged first to get this PR passing.

wantehchang:
Vincent: I have merged https://github.com/AOMediaCodec/libavif/pull/2278. Please rebase this pull request to the tip of the main branch.

.github/workflows/ci-windows-installed.yml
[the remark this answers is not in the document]
vrabaud: done

.github/workflows/ci-windows-installed.yml
```diff
@@ -50,6 +63,7 @@ jobs:
         mv ${{ github.workspace }}\tmp\rav1e-windows-msvc-sdk\include\* ${{ github.workspace }}\vcpkg\installed\x64-windows-release\include
         mv ${{ github.workspace }}\tmp\rav1e-windows-msvc-sdk\lib\rav1e* ${{ github.workspace }}\vcpkg\installed\x64-windows-release\lib
         mv ${{ github.workspace }}\tmp\rav1e-windows-msvc-sdk\lib\pkgconfig\* ${{ github.workspace }}\vcpkg\installed\x64-windows-release\lib\pkgconfig
+    - uses: ilammy/msvc-dev-cmd@v1
```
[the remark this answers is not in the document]
wantehchang: Can we replace egor-tensin/vs-shell with ilammy/msvc-dev-cmd? We should not need both of them.

Do you know why we can build libyuv using clang-cl in ci-windows.yml and ci-windows-artifacts.yml without ilammy/msvc-dev-cmd?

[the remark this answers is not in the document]
vrabaud: Yes, so we want to call CMake with two different compilers. There are several solutions:
- set up two jobs, running conditionally on matrix.compiler. I did not do that because the CMake command is quite long
- have one job and set the compilers through `CC` and `CXX`. But I do not know what to put for `CC` for `msvc`. Setting it as empty creates an error: CMake believes there is no compiler. And env variables cannot be set conditionally, only jobs. The jobs you mention use `CC`.
- have one job and set the compilers through CMake. That is what I am doing here and it works.

Let me try to remove that line so that you can see the error.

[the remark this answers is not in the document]
vrabaud: It now works without that line. Perfect ! Let's remove it.

wantehchang (approved):
LGTM. Thanks a lot for all the hard work on setting this up!

I still think we can switch to ilammy/msvc-dev-cmd later. Based on the time of its last commit, it is newer than egor-tensin/vs-shell.

.github/workflows/ci-windows-installed.yml
```diff
@@ -28,6 +30,17 @@ jobs:
         codec-aom: 'SYSTEM'
         codec-dav1d: 'SYSTEM'
         codec-rav1e: 'SYSTEM'
+        extra-cache-key: ${{ matrix.compiler }}
+    - name: Leave compiler as default
+      if: matrix.compiler == 'msvc'
+      run: |
+        echo "AVIF_CMAKE_C_COMPILER=" >> $env:GITHUB_ENV
+        echo "AVIF_CMAKE_CXX_COMPILER=" >> $env:GITHUB_ENV
```
wantehchang, line 38: Note: We can deal with this later.

This generates two cmake warnings:

```
CMake Warning:
  Ignoring empty string ("") provided on the command line.


CMake Warning:
  Ignoring empty string ("") provided on the command line.
```

See the log file at https://github.com/AOMediaCodec/libavif/actions/runs/9976134198/job/27567560349?pr=2270.

vrabaud: I guess we can do nothing and the CMake instruction will simply convert non-existent variables.

wantehchang: There are other warnings in the log file.

This one should be easy to fix:

```
  Manually-specified variables were not used by the project:

    AVIF_ENABLE_EXPERIMENTAL_AVIR
```

@y-guyon Yannis, Is that the old name for `AVIF_ENABLE_EXPERIMENTAL_METAV1`?

This one I don't understand:

```
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
Hint: The project() command has not yet been called.  It sets up system-specific search paths.
```

vrabaud: For the last warning, it's because we do `include(FindPkgConfig)` before calling `project()` in CMake.

wantehchang: Thanks. I will write a pull request to fix that.

Issue #2271, opened 2024-07-15
Encoder and decoder support of large image size


wantehchang:
This issue tracks the work to allow the libavif encoder and decoder to support larger image size.

Note that it requires an AV1 codec that supports large frame size. For example, this is libaom's issue on encoder support of large frame size: https://b.corp.google.com/issues/353371270

wantehchang:
@jzern @vrabaud @y-guyon 

Here is a representative example of (unsigned) integer overflow, taken from avifpng.c. Note that `rgb.rowBytes` is of the `uint32_t` type.

```
     for (uint32_t y = 0; y < rgb.height; ++y) {
         rowPointers[y] = &rgb.pixels[y * rgb.rowBytes];
     }
```

There are three ways to fix this.

1) Add a `size_t` cast to `y` or `rgb.rowBytes`:

```
     for (uint32_t y = 0; y < rgb.height; ++y) {
         rowPointers[y] = &rgb.pixels[(size_t)y * rgb.rowBytes];
     }
```

2) Declare the loop index variable `y` as `size_t`:

```
     for (size_t y = 0; y < rgb.height; ++y) {
         rowPointers[y] = &rgb.pixels[y * rgb.rowBytes];
     }
```

3) Replace the multiplication with addition of a pointer variable with `rgb.rowBytes`:

```
     uint8_t * rgbRow = rgb.pixels;
     for (uint32_t y = 0; y < rgb.height; ++y) {
         rowPointers[y] = rgbRow;
         rgbRow += rgb.rowBytes;
     }
```

Which solution do you prefer?

I prefer solution 3, because it is also a form of optimization ("strength reduction"). It requires more changes.

y-guyon:
I prefer solution 3 too.

PR #2272, opened 2024-07-16
avifenc: Allow large images to be encoded

wantehchang:
Pass UINT32_MAX as the imageSizeLimit argument to avifReadImage().

Bug: https://github.com/AOMediaCodec/libavif/issues/2271


wantehchang (commented):
Yannis: Please review this pull request carefully. I remember you added the `imageSizeLimit` parameter to `avifReadImage()` to avoid out-of-memory errors in fuzzing tests. Also,  `avifReadImage()` passes that parameter to `avifJPEGRead()` as the `sizeLimit` parameter, which is also used to limit the size of Exif and XMP metadata. So I don't fully understand the implications of this pull request.

apps/avifenc.c
```diff
@@ -611,7 +611,7 @@ static avifBool avifInputReadImage(avifInput * input,
                                                             ignoreXMP,
                                                             allowChangingCicp,
                                                             ignoreGainMap,
-                                                            AVIF_DEFAULT_IMAGE_SIZE_LIMIT,
```
y-guyon, line 614: Should it also be changed at line 585?

y-guyon (approved):
I think this change is fine. Metadata >256MB makes little sense but this is not worth adding a parameter to `avifReadImage()`.

apps/avifenc.c
wantehchang: Done. Thanks.

I found that `AVIF_DEFAULT_IMAGE_SIZE_LIMIT` is also used in two places in `avifgainmaputil`. I didn't look at them to see if either of them should be changed to `UINT32_MAX`.

wantehchang (commented):
Thanks for the review. I also added a changelog entry. PTAL.

y-guyon (approved):

PR #2275, opened 2024-07-16
Get libavif to build with clang-cl.

vrabaud:
Some clang-cl warnings are also ignored for now.


src/write.c
```diff
@@ -248,7 +248,7 @@ typedef struct avifEncoderData
 static void avifEncoderDataDestroy(avifEncoderData * data);
 
 // Returns NULL if a memory allocation failed.
-static avifEncoderData * avifEncoderDataCreate()
+static avifEncoderData * avifEncoderDataCreate(void)
```
wantehchang, line 251: Some of these warning fixes are good, such as adding `void` for an empty parameter list, and fixing unintended "const casts" (e.g., at line 1005 below).

wantehchang:
I wrote: 
> There are more warnings than I expected. It's also strange that we don't get these warnings when we compile with clang on Linux or Mac.

I figured out why. We pass `-Wall -Wextra -Wshorten-64-to-32` to clang-cl. Since clang-cl aims to emulate MSVC, apparently it treats `-Wall` as MSVC's `/Wall` option.

For example, https://clang.llvm.org/docs/UsersManual.html#id11 says:

>   /W4                     Enable -Wall and -Wextra
>   /Wall                   Enable -Weverything

So `-Wall` for clang-cl maps to `-Weverything` for clang.

For genuine MSVC, `/Wall` means:
https://learn.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-170

> Displays all warnings displayed by `/W4` and all other warnings that `/W4` doesn't include—for example, warnings that are off by default. For more information, see [Compiler warnings that are off by default](https://learn.microsoft.com/en-us/cpp/preprocessor/compiler-warnings-that-are-off-by-default?view=msvc-170).

So I think my suggestion of using `/W4` for clang-cl is correct, because it maps to `Enable -Wall and -Wextra` for clang.

apps/shared/avifutil.c
```diff
@@ -412,7 +412,7 @@ void avifDumpDiagnostics(const avifDiagnostics * diag)
 
 // Windows
 
-#include <windows.h>
+#include <Windows.h>
```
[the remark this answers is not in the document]
vrabaud: Otherwise, we get the warning:
```
D:\a\libavif\libavif\apps\shared\avifutil.c(415,10): error: non-portable path to file '<Windows.h>'; specified path differs in case from file name on disk [-Werror,-Wnonportable-system-include-path]
  415 | #include <windows.h>
      |          ^~~~~~~~~~~
      |          <Windows.h>
```

apps/shared/avifjpeg.c
```diff
@@ -297,13 +297,13 @@ static uint32_t avifJPEGReadUint32LittleEndian(const uint8_t * src)
 // Reads a 2-byte unsigned integer in big-endian format from the raw bitstream src.
 static uint16_t avifJPEGReadUint16BigEndian(const uint8_t * src)
 {
-    return ((uint32_t)src[0] << 8) | ((uint32_t)src[1] << 0);
+    return (uint16_t)(((uint16_t)src[0] << 8) | ((uint16_t)src[1] << 0));
```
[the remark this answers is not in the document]
vrabaud: done

include/avif/internal.h
```diff
@@ -558,17 +558,17 @@ struct avifCodecInternal;
...
+    AVIF_ENCODER_CHANGE_MAX_QUANTIZER = (1 << 1),
+    AVIF_ENCODER_CHANGE_MIN_QUANTIZER_ALPHA = (1 << 2),
+    AVIF_ENCODER_CHANGE_MAX_QUANTIZER_ALPHA = (1 << 3),
+    AVIF_ENCODER_CHANGE_TILE_ROWS_LOG2 = (1 << 4),
+    AVIF_ENCODER_CHANGE_TILE_COLS_LOG2 = (1 << 5),
+    AVIF_ENCODER_CHANGE_QUANTIZER = (1 << 6),
+    AVIF_ENCODER_CHANGE_QUANTIZER_ALPHA = (1 << 7),
+    AVIF_ENCODER_CHANGE_SCALING_MODE = (1 << 8),
+
+    AVIF_ENCODER_CHANGE_CODEC_SPECIFIC = (1 << 30)
```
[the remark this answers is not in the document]
wantehchang: Thank you! I didn't realize that enumeration constants in C always have the `int` type.

Given this, we should also change the `avifEncoderChanges` typedef at line 573 to `int`:

```
typedef int avifEncoderChanges;
```

because `avifEncoderChanges` represents a bitwise-OR of these enumeration constants.

Note: There are several typedefs of this kind in the public avif.h header, but we can't change them (woud be incompatible/breaking changes).

src/read.c
```diff
@@ -5451,7 +5451,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
             if (avifIsAlpha((avifItemCategory)c) && !isAlphaItemInInput) {
                 // In this case, the made up grid item will not have an associated pixi property. So validate everything else
                 // but the pixi property.
-                strictFlags &= ~AVIF_STRICT_PIXI_REQUIRED;
+                strictFlags &= (avifStrictFlags)(~AVIF_STRICT_PIXI_REQUIRED);
```
wantehchang, line 5454: Nit: Please cast first, then invert:

```
                // but the pixi property.
                strictFlags &= ~(avifStrictFlags)AVIF_STRICT_PIXI_REQUIRED;
```

src/read.c
```diff
@@ -5361,7 +5361,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                 if (!foundItem) {
                     AVIF_CHECKERR(numExtraInputImageItems < AVIF_SAMPLE_TRANSFORM_MAX_NUM_EXTRA_INPUT_IMAGE_ITEMS,
                                   AVIF_RESULT_NOT_IMPLEMENTED);
-                    *category = AVIF_ITEM_SAMPLE_TRANSFORM_INPUT_0_COLOR + numExtraInputImageItems;
+                    *category = (avifItemCategory)(AVIF_ITEM_SAMPLE_TRANSFORM_INPUT_0_COLOR + numExtraInputImageItems);
```
wantehchang, line 5364: @y-guyon Yannis, please see this comment.

Nit: This line and the next line are inconsistent.

In this line, `*category` is of the `avifItemCategory` type, so we need a `avifItemCategory` cast.

In the next line, `alphaCategory` is of the `uint32_t` type (see line 5351), so we don't need a cast.

I wonder if we should declare `alphaCategory` as `avifItemCategory`.

apps/shared/avifutil.c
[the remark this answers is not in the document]
vrabaud: reverted

include/avif/internal.h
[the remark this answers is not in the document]
vrabaud: done

src/read.c
vrabaud: done

src/read.c
```diff
@@ -5441,7 +5441,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
             if (avifIsAlpha((avifItemCategory)c) && !isAlphaItemInInput) {
                 // In this case, the made up grid item will not have an associated pixi property. So validate everything else
                 // but the pixi property.
-                strictFlags &= ~AVIF_STRICT_PIXI_REQUIRED;
+                strictFlags &= ~(avifStrictFlags)(AVIF_STRICT_PIXI_REQUIRED);
```
wantehchang, line 5444: Nit: Omit the parentheses around `AVIF_STRICT_PIXI_REQUIRED`. Alternatively, we can add parentheses after casting:

```
                strictFlags &= ~((avifStrictFlags)AVIF_STRICT_PIXI_REQUIRED);
```

wantehchang (approved):

CMakeLists.txt
```diff
@@ -298,6 +298,24 @@ if(AVIF_ENABLE_WERROR)
...
+                   -Wno-deprecated-declarations
+                   -Wno-double-promotion
+                   -Wno-float-equal
+                   -Wno-implicit-int-float-conversion
+                   -Wno-sign-conversion
+                   -Wno-switch-default
+                   -Wno-switch-enum
+                   -Wno-unsafe-buffer-usage>
+        )
+    endif()
```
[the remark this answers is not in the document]
wantehchang: We were passing `-Wall -Wextra` to clang-cl. clang-cl treats `-` and `/` interchangeably, so `-Wall` is treated the same as `/Wall`.

clang-cl maps MSVC's `/Wall` to Clang's `-Weverything`. clang-cl maps MSVC's `/W4` to Clang's `-Wall -Wextra`, which enables fewer warnings than Clang's `-Weverything`.  If there are any warnings in `-Weverything` that you find useful, we can enable them for clang-cl.

src/read.c
vrabaud: done

src/read.c
y-guyon: I guess I kept the most convenient type for each variable at the time but feel free to align them for consistency.

wantehchang: Done in https://github.com/AOMediaCodec/libavif/pull/2332.

PR #2278, opened 2024-07-16
Support compiling with clang-cl

wantehchang:
Instead of testing
    CMAKE_C_COMPILER_ID MATCHES "MSVC"
test the cmake variable MSVC.

Exclude /source-charset:utf-8 /execution-charset:us-ascii from clang-cl because clang-cl does not support /execution-charset:us-ascii and /source-charset:utf-8 is the only /source-charset option clang-cl supports.

Define _CRT_NONSTDC_NO_WARNINGS and _CRT_SECURE_NO_WARNINGS to disable deprecation warnings about POSIX function names and less secure functions in Microsoft CRT. I don't know why genuine MSVC doesn't warn about these functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The changes are relative to the previous release, unless the baseline is specifi
   disabled for non-native builds.
 * Fix CMake config shared library leaks
   https://github.com/AOMediaCodec/libavif/issues/2264.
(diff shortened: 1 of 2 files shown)


jzern (approved):

wantehchang:
Thanks for the review. I added a changelog entry.

PR #2280, opened 2024-07-17
setup-linux: Use -y with apt commands

vigneshvg:
Some apt install/update commands could get stuck waiting on yes/no
confirmation. Always use -y with apt to ensure that we do not get
stuck.

### DIFF
--- a/.github/actions/setup-linux/action.yml
+++ b/.github/actions/setup-linux/action.yml
@@ -36,43 +36,43 @@ runs:
   steps:
     - name: Install non-library dependencies
       run: |
-        sudo apt update
-        sudo apt install imagemagick libjpeg-turbo8-dev libpng-dev
+        sudo apt update -y
+        sudo apt install -y imagemagick libjpeg-turbo8-dev libpng-dev
       shell: bash
     - name: Install libaom library
       if: ${{ inputs.codec-aom == 'SYSTEM' }}
       run:
-        sudo apt install libaom-dev
+        sudo apt install -y libaom-dev
       shell: bash
     - name: Install libdav1d library
       if: ${{ inputs.codec-dav1d == 'SYSTEM' }}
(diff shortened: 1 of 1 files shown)


wantehchang (approved):
Please wait for Vincent's approval.

Vincent: I think we can merge this pull request before the v1.1.1 release.

.github/actions/setup-linux/action.yml
```diff
@@ -36,43 +36,43 @@ runs:
   steps:
     - name: Install non-library dependencies
       run: |
-        sudo apt update
-        sudo apt install imagemagick libjpeg-turbo8-dev libpng-dev
+        sudo apt update -y
+        sudo apt install -y imagemagick libjpeg-turbo8-dev libpng-dev
```
wantehchang, line 40: @vigneshvg @y-guyon We are using brew and vcpkg on Mac and Windows. I wonder if they have a similar option.

I searched for "-y" and "yes" in the [brew man page](https://docs.brew.sh/Manpage) but didn't find anything. Same for the [vcpkg install doc page](https://learn.microsoft.com/en-us/vcpkg/commands/install#options).

vigneshvg: > @vigneshvg @y-guyon We are using brew and vcpkg on Mac and Windows. I wonder if they have a similar option.
> 
> I searched for "-y" and "yes" in the [brew man page](https://docs.brew.sh/Manpage) but didn't find anything. Same for the [vcpkg install doc page](https://learn.microsoft.com/en-us/vcpkg/commands/install#options).

yeah i was just searching for it as well. it seems like bre and vcpkg assume -y by default and they don't ask for any user prompts.

vrabaud:
Sure, let's merge. have we seen cases on the CI where it actually freezes?

PR #2281, opened 2024-07-17
Move the project() call earlier

wantehchang:
Call project() after the cmake_minimum_required() and cmake_policy() calls, but before everything else.

Fix the following warning message from
.github/workflows/ci-windows-installed.yml:

  -- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
  Hint: The project() command has not yet been called.  It sets up
  system-specific search paths.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,8 @@ if(POLICY CMP0135)
     cmake_policy(SET CMP0135 NEW) # valid for DOWNLOAD_EXTRACT_TIMESTAMP option in CMake 3.24 and later
 endif()
 
+project(libavif LANGUAGES C VERSION 1.1.0)
+
 # The root directory of the avif source
 set(AVIF_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
 
@@ -24,8 +26,6 @@ include(ExternalProject)
 include(FetchContent)
 include(FindPkgConfig)
 include(AvifExternalProjectUtils)
-
-project(libavif LANGUAGES C VERSION 1.1.0)
 
 # Set C99 as the default
 set(CMAKE_C_STANDARD 99)


CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ if(POLICY CMP0135)
     cmake_policy(SET CMP0135 NEW) # valid for DOWNLOAD_EXTRACT_TIMESTAMP option in CMake 3.24 and later
 endif()
 
+project(libavif LANGUAGES C VERSION 1.1.0)
+
```
wantehchang, line 18: [CMake documentation on `project()`](https://cmake.org/cmake/help/latest/command/project.html#usage) says:

> **Note:**  Call the `project()` command near the top of the top-level `CMakeLists.txt`, but after calling [cmake_minimum_required()](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required).

But I did not find guidance in CMake documentation on whether `project()` should be called before or after `cmake_policy()`. In this pull request I preserve the current order of calling `project()` after `cmake_policy()`.

wantehchang: Yes, [CMake documentation on cmake_minimum_required()](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html) says it should be the first call in the top-level CMakeLists.txt file:

> **Note:**  Call the `cmake_minimum_required()` command at the beginning of the top-level `CMakeLists.txt` file even before calling the [project()](https://cmake.org/cmake/help/latest/command/project.html#command:project) command.

Do you recommend we call `project()` before `make_policy()`?

jzern: Thinking about it a little, I think it would only matter if one of the policies affected `project()` and you wanted to keep the grouping. So think the order you have makes sense.

vrabaud (approved):

PR #2282, opened 2024-07-17
ci: Run on push only in the `main` branch

vigneshvg:
This avoids excessive CI runs when pushing changes to other
branches on forks.

### DIFF
--- a/.github/workflows/ci-android-jni.yml
+++ b/.github/workflows/ci-android-jni.yml
@@ -1,5 +1,9 @@
 name: CI Android JNI
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
 
 permissions:
   contents: read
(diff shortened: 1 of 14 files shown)


wantehchang (commented):
Vignesh: Thanks for the improvement. I am not familiar with the "run on" settings, so I am deferring to Vincent  and Yannis.

We need to make sure we run these on release branches. Is there a way to specify that?

We can assume our release branches follow the naming convension `v*.*.x`.

y-guyon:
I disabled GitHub actions on my fork some time ago to avoid this issue. Local tests and PR tests are enough for me.

> We need to make sure we run these on release branches.

I agree. Maybe restricting to base repository branches would be better but that may be difficult to setup.

vrabaud:
I agree, we should keep that for release branches and tags:
```
branches
  - main
  - 'v*.*.x'
 tags
  - v*.*.x
```
cf https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags

PR #2284, opened 2024-07-17
Remove the MSVC /wd4996 option

wantehchang:
The MSVC /wd4996 option disabled the C4996 warning on deprecated functions. We no longer need it now that we turn off deprecation warnings by defining the _CRT_NONSTDC_NO_WARNINGS and _CRT_SECURE_NO_WARNINGS macros.

Note: This explains why clang-cl emitted the deprecation warnings but MSVC didn't. Apparently clang-cl silently ignores the MSVC /wd4996 option.

Edit the comment for the _CRT_SECURE_NO_WARNINGS macro slightly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,7 +268,6 @@ if(MSVC)
         PUBLIC $<BUILD_INTERFACE:
                /W4 # For clang-cl, /W4 enables -Wall and -Wextra
                /wd4324 # Disable: structure was padded due to alignment specifier
-               /wd4996 # Disable: potentially unsafe stdlib methods
                >
     )
     # clang-cl documentation says:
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -342,7 +341,7 @@ endif()
 if(MSVC)
     # Disable deprecation warnings about POSIX function names such as setmode (replaced by the ISO C and C++ conformant name _setmode).
     add_compile_definitions(_CRT_NONSTDC_NO_WARNINGS)
-    # Disable deprecation warnings about less secure CRT functions such as fopen (replaced by fopen_s).
+    # Disable deprecation warnings about unsafe CRT library functions such as fopen (replaced by fopen_s).
```
wantehchang, line 344: The C4996 warning and these two macros are documented in https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=msvc-170

vrabaud (approved):

PR #2287, opened 2024-07-18
Change egor-tensin/vs-shell to ilammy/msvc-dev-cmd

wantehchang:


wantehchang:
Vincent: Is it possible to find out which of these is more widely used?

My current decision is based on how recent is the last commit.
- egor-tensin/vs-shell: Nov 11, 2021
- ilammy/msvc-dev-cmd: Mar 30, 2024

vrabaud:
You are totally right. Egor has no issues, pull requests and it has failing tests at HEAD soooo ... ilammy seems more future proof

wantehchang:
Thanks for the confirmation. I also found that github provides the number of stars and a "Used by" number. ilammy/msvc-dev-cmd is much more widely used.

PR #2288, opened 2024-07-18
FOR TESTING ONLY: test windows clang-cl workflow

wantehchang:
clang-cl compilation should fail.


vrabaud:
It passes so I guess we need more warning flags?

wantehchang:
I am also puzzled why it passed. I think it is because Chrome enables C++ version >= 17. We enable C++14.

I just checked Google's [Foundational C++ Support table](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md). It supports C++ >= 14. So we should not enable C++17 yet. Our only option is to set `-DAVIF_ENABLE_NODISCARD=ON` in the clang-cl CI workflow. Not sure if we should do that. I will do that as an experiment in this test pull request.

For now, let's just import into Chrome frequently.

wantehchang:
This pull request was just a test. Test completed.

PR #2289, opened 2024-07-18
Set Policy CMP0169 to OLD

wantehchang:
Allow `FetchContent_Populate(<name>)` to be called directly.


wantehchang:
Here is an example of the cmake warning messages:

```
-- libavif(AVIF_LOCAL_LIBARGPARSE): compiled library not found at D:/a/libavif/libavif/ext/libargparse/build/argparse.lib; using FetchContent
-- libavif(AVIF_LOCAL_LIBARGPARSE): fetching and configuring libargparse
CMake Warning (dev) at C:/Program Files/CMake/share/cmake-3.30/Modules/FetchContent.cmake:1953 (message):
  Calling FetchContent_Populate(libargparse) is deprecated, call
  FetchContent_MakeAvailable(libargparse) instead.  Policy CMP0169 can be set
  to OLD to allow FetchContent_Populate(libargparse) to be called directly
  for now, but the ability to call it with declared details will be removed
  completely in a future version.
Call Stack (most recent call first):
  cmake/Modules/AvifExternalProjectUtils.cmake:3 (FetchContent_Populate)
  cmake/Modules/LocalLibargparse.cmake:28 (avif_fetchcontent_populate_cmake)
  CMakeLists.txt:759 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- libavif(AVIF_LOCAL_LIBARGPARSE): fetching and configuring libargparse - complete
-- libavif(AVIF_GTEST=LOCAL): compiled library not found in ext/googletest; using FetchContent
-- libavif(AVIF_LOCAL_GTEST): fetching and configuring googletest
CMake Warning (dev) at C:/Program Files/CMake/share/cmake-3.30/Modules/FetchContent.cmake:1953 (message):
  Calling FetchContent_Populate(googletest) is deprecated, call
  FetchContent_MakeAvailable(googletest) instead.  Policy CMP0169 can be set
  to OLD to allow FetchContent_Populate(googletest) to be called directly for
  now, but the ability to call it with declared details will be removed
  completely in a future version.
Call Stack (most recent call first):
  cmake/Modules/AvifExternalProjectUtils.cmake:3 (FetchContent_Populate)
  cmake/Modules/LocalGTest.cmake:44 (avif_fetchcontent_populate_cmake)
  CMakeLists.txt:198 (include)
  tests/CMakeLists.txt:90 (check_avif_option)
This warning is for project developers.  Use -Wno-dev to suppress it.

```

vrabaud (approved):

CMakeLists.txt
```diff
@@ -14,6 +14,12 @@ if(POLICY CMP0135)
     cmake_policy(SET CMP0135 NEW) # valid for DOWNLOAD_EXTRACT_TIMESTAMP option in CMake 3.24 and later
 endif()
 
+# New in CMake version 3.30. FetchContent_Populate(<name>) is deprecated, call
+# FetchContent_MakeAvailable(<name>) instead.
```
jzern, line 18: `FetchContent_MakeAvailable` was introduced in 3.14. A version check could be used to call it instead to guard against the removal of this policy.

https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable

wantehchang: Thanks for the suggestion. I did not mention that I tried to do the right thing. Our code in cmake/Modules/AvifExternalProjectUtils.cmake does the following (irrelevant details omitted):

```
macro(avif_fetchcontent_populate_cmake name)
    if(NOT ${name}_POPULATED)
        FetchContent_Populate(${name})
        ...
        add_subdirectory(${${name}_SOURCE_DIR} ${${name}_BINARY_DIR} EXCLUDE_FROM_ALL)
        ...
    endif()
endmacro()
```

Note that we pass `EXCLUDE_FROM_ALL` to the `add_subdirectory()` call. To replace the code above with a `FetchContent_MakeAvailable()` call, we will need to pass `EXCLUDE_FROM_ALL` to all the `FetchContent_Declare()` calls. But passing `EXCLUDE_FROM_ALL` to `FetchContent_Declare()` is a new feature in CMake 3.28:

https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_declare

Therefore, the cmake version check we need is 3.28, not 3.14, and we will need to add a new function that takes care of the `FetchContent_Declare()` + `FetchContent_MakeAvailable()` sequence. At that point I concluded I should just set the `CMP0169` policy to `OLD`.

wantehchang: James: We can consider requiring CMake 3.16 as the minimum version, based on the [Google Foundational C++ support table](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md).

Issue #2277, opened 2024-07-16
Broad meaning of the cmake variable AVIF_USE_CXX makes certain conditional tests inaccurate


wantehchang:
The cmake variable AVIF_USE_CXX means whether C++ is used in any part of libavif, e.g., in the library, in the apps (specifically avifgainmaputil), or in the tests:

```
set(AVIF_USE_CXX OFF)                  
                                                                              
if(AVIF_BUILD_APPS
   OR AVIF_ENABLE_FUZZTEST
   OR AVIF_ENABLE_GTEST                   
   OR AVIF_CODEC_LIBGAV1 STREQUAL "LOCAL"                                 
   OR AVIF_CODEC_LIBGAV1 STREQUAL "SYSTEM"                                      )                                          
    set(AVIF_USE_CXX ON)                                     
endif()
```

This broad meaning of `AVIF_USE_CXX` makes certain conditional tests inaccurate. Here is an example of an inaccurate conditional test:

```
if(BUILD_SHARED_LIBS)
    target_compile_definitions(avif INTERFACE AVIF_DLL)
    if(AVIF_USE_CXX)                   
        set_target_properties(avif PROPERTIES LINKER_LANGUAGE "CXX")          
    endif()       
endif()
```

The `avif` library does not contain C++ code (unless `AVIF_ENABLE_COMPLIANCE_WARDEN` is ON). Suppose we use libaom and dav1d as the AV1 codecs (so libgav1 is not used), and we do not use libyuv and libsharpyuv. Then the `avif` library does not depend on any C++ library. However, if we enable `AVIF_BUILD_APPS` or `AVIF_ENABLE_GTEST`, then `AVIF_USE_CXX` is `ON`. But in this case, the `avif` shared library should not be linked as C++ code.

One solution is to add a new cmake variable that specifically indicates whether the `avif` library contains C++ code or needs to be linked as C++.

Issue #2274, opened 2024-07-16
v1.1.0 failed to build with aom -DCONFIG_TUNE_BUTTERAUGLI or -DCONFIG_TUNE_VMAF


Dzhami-Jame-ALI:
Since v1.1.0, libavif failed to build with [aom-DCONFIG_TUNE_BUTTERAUGLI] and [aom-DCONFIG_TUNE_VMAF] (aom without butteraugli or vmaf works fine with libavif, and I can still use ssim). The libjxl-butteraugli and vmaf library is pre-compiled, can be recognized and built into aom (v3.9.1). However, libavif failed to compile after building aom with butteraugli or vmaf:
```
...
vmaf.c:(.text+0x2b5): undefined reference to `vmaf_picture_alloc'
/usr/bin/ld: vmaf.c:(.text+0x2e1): undefined reference to `vmaf_picture_alloc'
/usr/bin/ld: vmaf.c:(.text+0x3cd): undefined reference to `vmaf_read_pictures'
/usr/bin/ld: vmaf.c:(.text+0x3e5): undefined reference to `vmaf_read_pictures'
/usr/bin/ld: vmaf.c:(.text+0x3f7): undefined reference to `vmaf_picture_unref'
/usr/bin/ld: vmaf.c:(.text+0x401): undefined reference to `vmaf_picture_unref'
/usr/bin/ld: vmaf.c:(.text+0x417): undefined reference to `vmaf_score_at_index'
/usr/bin/ld: /usr/local/lib/libaom.a(vmaf.c.o): in function `aom_read_vmaf_image':
vmaf.c:(.text+0x52d): undefined reference to `vmaf_picture_alloc'
/usr/bin/ld: vmaf.c:(.text+0x559): undefined reference to `vmaf_picture_alloc'
/usr/bin/ld: vmaf.c:(.text+0x647): undefined reference to `vmaf_read_pictures'
/usr/bin/ld: vmaf.c:(.text+0x659): undefined reference to `vmaf_picture_unref'
/usr/bin/ld: vmaf.c:(.text+0x663): undefined reference to `vmaf_picture_unref'
/usr/bin/ld: /usr/local/lib/libaom.a(vmaf.c.o): in function `aom_calc_vmaf_at_index':
vmaf.c:(.text+0x72c): undefined reference to `vmaf_score_at_index'
/usr/bin/ld: /usr/local/lib/libaom.a(vmaf.c.o): in function `aom_flush_vmaf_context':
[...]

PR #2290, opened 2024-07-18
Clarify the definition of AVIF_USE_CXX.

vrabaud:
Only set `AVIF_USE_CXX` when linking to C++ libraries: libyuv, libgav1 or ComplianceWarden 

This fixes https://github.com/AOMediaCodec/libavif/issues/2277


.github/workflows/ci-android-emulator-tests.yml
```diff
@@ -21,20 +17,13 @@ concurrency:
 
 jobs:
   build-and-run-android-jni-tests:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
wantehchang, line 20: Undo the changes to this file. It looks like a merge conflict resolved incorrectly.

CMakeLists.txt
```diff
@@ -632,13 +629,14 @@ endif()
...
     get_target_property(AOM_INTERFACE_LINK_LIBRARIES aom INTERFACE_LINK_LIBRARIES)
-    if(AOM_INTERFACE_LINK_LIBRARIES MATCHES vmaf)
-        # vmaf only affects the avifenc and avifdec targets below, not the avif library nor examples targets.
-        set(AVIF_USE_CXX ON)
-    endif()
 endif()
 
-if(AVIF_USE_CXX)
+if(AVIF_USE_CXX
+   OR AVIF_BUILD_APPS
```
wantehchang, line 635: Optional: Right now avifgainmaputil is the only app that uses C++, and it is an experimental feature. So this line can be changed to:

```
   OR (AVIF_BUILD_APPS AND AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
```

CMakeLists.txt
```diff
@@ -133,17 +133,9 @@ option(
     OFF
 )
 
+# Whether the libavif library uses c++ indirectly (e.g. through linking to libgav1).
```
wantehchang, line 136: I suggest we rename the variable to reflect the new, more specific meaning. The new name should have "LIB" in the name, e.g., "AVIF_LIB_USE_CXX" or "AVIF_USE_CXX_IN _LIB" or "AVIF_LINK_LIB_AS_CXX".

CMakeLists.txt
```diff
@@ -571,6 +565,20 @@ if(NOT AVIF_CODEC_AOM_ENABLED
...
+   OR (AVIF_CODEC_AOM_ENABLED AND (AOM_INTERFACE_LINK_LIBRARIES MATCHES vmaf))
+)
+    enable_language(CXX)
+    if(AVIF_ENABLE_NODISCARD)
+        set(CMAKE_CXX_STANDARD 17)
+    else()
+        set(CMAKE_CXX_STANDARD 14)
+    endif()
+endif()
+
```
wantehchang, line 581: Delete the if statement at lines 648-650 because the variable `AOM_INTERFACE_LINK_LIBRARIES` is unused now:

```
if(AVIF_CODEC_AOM_ENABLED)
    get_target_property(AOM_INTERFACE_LINK_LIBRARIES aom INTERFACE_LINK_LIBRARIES)
endif()
```

CMakeLists.txt
```diff
@@ -571,6 +565,20 @@ if(NOT AVIF_CODEC_AOM_ENABLED
     message(WARNING "libavif: No decoding library is enabled.")
 endif()
 
+if(AVIF_USE_CXX
+   OR (AVIF_BUILD_APPS AND AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+   OR AVIF_ENABLE_FUZZTEST
+   OR AVIF_ENABLE_GTEST
+   OR (AVIF_CODEC_AOM_ENABLED AND (AOM_INTERFACE_LINK_LIBRARIES MATCHES vmaf))
```
wantehchang, line 572: IMPORTANT: With the new meaning of `AVIF_USE_CXX` defined at line 136, I would expect that this case falls under the new definition of `AVIF_USE_CXX` and therefore should be handled by setting `AVIF_USE_CXX` to `ON`.

However, the original comment has a comment (which may have been deleted accidentally in this pull request):

```
if(AVIF_CODEC_AOM_ENABLED)
    get_target_property(AOM_INTERFACE_LINK_LIBRARIES aom INTERFACE_LINK_LIBRARIES)
    if(AOM_INTERFACE_LINK_LIBRARIES MATCHES vmaf)
        # vmaf only affects the avifenc and avifdec targets below, not the avif library nor examples targets.
        set(AVIF_USE_CXX ON)
    endif()
endif()
```

If that comment is accurate, then we do NOT need to link the libavif library using the C++ compiler. So we need to look into this. If we are not sure if that comment is wrong, it is best to assume the comment is accurate and preserve the original behavior.

vrabaud: Bad copy paste: this will be fixed once https://github.com/AOMediaCodec/libavif/issues/2274 is fixed

CMakeLists.txt
vrabaud: AVIF_LIB_USE_CXX ssems like the best one to me

wantehchang (approved):
LGTM. Thanks! I suggest some small changes.

wantehchang:
I tested this pull request (both without and with my suggested changes) using the following commands:
```
cd ext
./aom.cmd
cd ..
mkdir build
cd build
cmake .. -G Ninja -DAVIF_CODEC_AOM=LOCAL -DAVIF_LIBYUV=OFF -DAVIF_BUILD_APPS=ON
ninja
```

Before:

```
[34/40] Linking CXX shared library libavif.so.16.1.0
...
[39/40] Linking CXX executable avifdec
[40/40] Linking CXX executable avifenc
```

After:

```
[34/40] Linking C shared library libavif.so.16.1.0
...
[39/40] Linking C executable avifdec
[40/40] Linking C executable avifenc
```

CMakeLists.txt
wantehchang: I looked up the source of that comment. The comment was originally added in https://github.com/AOMediaCodec/libavif/commit/ef6da2a78 and updated in https://github.com/AOMediaCodec/libavif/commit/47e58af14. I believe the comment was added to describe the existing code. The code was originally added in https://github.com/AOMediaCodec/libavif/commit/778b77fa0. I think the code is incorrect. The comment accurately describes the code. That's why I find the comment confusing and seeminly incorrect.

This pull request corrects the code and deletes the comment (which becomes inaccurate because the code it described has changed). That's good.

CMakeLists.txt
```diff
@@ -133,16 +133,8 @@ option(
...
-
-if(AVIF_BUILD_APPS
-   OR AVIF_ENABLE_FUZZTEST
-   OR AVIF_ENABLE_GTEST
-   OR AVIF_CODEC_LIBGAV1 STREQUAL "LOCAL"
-   OR AVIF_CODEC_LIBGAV1 STREQUAL "SYSTEM"
-)
-    set(AVIF_USE_CXX ON)
-endif()
+# Whether the libavif library uses c++ indirectly (e.g. through linking to libgav1).
```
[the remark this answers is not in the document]
vrabaud: done

CMakeLists.txt
```diff
@@ -543,6 +537,12 @@ if(AVIF_CODEC_AOM_ENABLED)
 
     message(STATUS "libavif: Codec enabled: aom (${AVIF_CODEC_AOM_ENCODE_DECODE_CONFIG})")
     set(AVIF_PKG_CONFIG_EXTRA_REQUIRES_PRIVATE "${AVIF_PKG_CONFIG_EXTRA_REQUIRES_PRIVATE} aom")
+
+    get_target_property(AOM_INTERFACE_LINK_LIBRARIES aom INTERFACE_LINK_LIBRARIES)
+    if(AOM_INTERFACE_LINK_LIBRARIES MATCHES vmaf)
+        # vmaf only affects the avifenc and avifdec targets below, not the avif library nor examples targets.
```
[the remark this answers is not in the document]
vrabaud: done

CMakeLists.txt
```diff
@@ -571,6 +571,20 @@ if(NOT AVIF_CODEC_AOM_ENABLED
     message(WARNING "libavif: No decoding library is enabled.")
 endif()
 
+if(AVIF_LIB_USE_CXX
+   OR (AVIF_BUILD_APPS AND AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+   OR AVIF_ENABLE_FUZZTEST
+   OR AVIF_ENABLE_GTEST
```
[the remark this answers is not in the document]
vrabaud: let's do it here.

wantehchang (commented):
I reviewed the last (fifth) commit ("Fix nits"). LGTM. Thanks!

Issue #2276, opened 2024-07-16
In CMakeLists.txt, AVIF_ENABLE_NODISCARD probably should set CMAKE_CXX_STANDARD to 17


wantehchang:
In CMakeLists.txt, `AVIF_ENABLE_NODISCARD` sets `CMAKE_C_STANDARD` to 23:

```
if(AVIF_ENABLE_NODISCARD)
    if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.21.0)
        set(CMAKE_C_STANDARD 23)                  
        set_property(TARGET avif_obj PROPERTY C_STANDARD 23)
    else()
```

It probably should also set `CMAKE_CXX_STANDARD` to 17, because the `nodiscard` attribute was added in C++17:
https://en.cppreference.com/w/cpp/language/attributes/nodiscard

Note: GCC 13.2.0 and Clang 16.0.6 support `[[nodiscard]]` when the `-std=gnu++14` flag is used. I guess `[[nodiscard]]` is considered a GNU extension in C++14 mode.

vrabaud:
Let's not try to figure out the right flags with which version. Let's stick to C++17, it is easier.

PR #2291, opened 2024-07-18
Set C++ version to 17 when using [[no_discard]]

vrabaud:
17 was introduced in CMake 3.8
https://cmake.org/cmake/help/v3.8/prop_tgt/CXX_STANDARD.html

This fixes https://github.com/AOMediaCodec/libavif/issues/2276


CMakeLists.txt
```diff
@@ -314,6 +314,7 @@ if(AVIF_ENABLE_WERROR)
 endif()
 
 if(AVIF_ENABLE_NODISCARD)
+    set(CMAKE_CXX_STANDARD 17)
```
wantehchang, line 317: This is too early, before the following code at line 642:

```
if(AVIF_USE_CXX)
    enable_language(CXX)
    set(CMAKE_CXX_STANDARD 14)
endif()
```

It seems that we should change that code to:

```
if(AVIF_USE_CXX)
    enable_language(CXX)
    if(AVIF_ENABLE_NODISCARD)
        set(CMAKE_CXX_STANDARD 17)
    else()
        set(CMAKE_CXX_STANDARD 14)
    endif()
endif()
```

wantehchang (approved):

PR #2292, opened 2024-07-18
Use add_compile_options() to set warning options

wantehchang:
The warning options should be set on all targets. Use add_compile_options() instead of target_compile_options(avif_obj) to set them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,12 +269,9 @@ endif()
 include(CheckCCompilerFlag)
 if(MSVC)
     message(STATUS "libavif: Enabling warnings for MSVC")
-    target_compile_options(
-        avif_obj
-        PUBLIC $<BUILD_INTERFACE:
-               /W4 # For clang-cl, /W4 enables -Wall and -Wextra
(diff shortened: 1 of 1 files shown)


fdintino:
I limited the compiler warnings to `avif_obj` because otherwise it throws errors when compiling the dependencies as cmake subprojects. I think I also figured that making them public compile flags would cause them to be used for all source files in the repository, but I could have been mistaken in that.

wantehchang:
Frankie: Thanks for your comment. I also found from the CI workflow failures that we cannot use `add_compile_options()` because we should not impose our warning options on the dependencies.

So we may need to save our warning options in a variable and set them on our targets individually.

wantehchang:
I am abandoning this experimental pull request. It is replaced by https://github.com/AOMediaCodec/libavif/pull/2327. I am encapsulating compiler warning options in a CMake interface library named `avif_enable_warnings`, modeled after the `enable_rtti` example at https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#interface-libraries.

PR #2295, opened 2024-07-18
avifgainmaputil: Fix MSVC type conversion warnings

wantehchang:
Fix MSVC C4244 warnings on type conversions, possible loss of data.

### DIFF
--- a/apps/avifgainmaputil/combine_command.cc
+++ b/apps/avifgainmaputil/combine_command.cc
@@ -98,10 +98,13 @@ avifResult CombineCommand::Run() {
   }
 
   const int downscaling = std::max<int>(1, arg_downscaling_);
-  const uint32_t gain_map_width = std::max<uint32_t>(
-      std::round((float)base_image->width / downscaling), 1u);
-  const uint32_t gain_map_height = std::max<uint32_t>(
-      std::round((float)base_image->height / downscaling), 1u);
+  const uint32_t gain_map_width = std::max(
+      static_cast<uint32_t>(std::round((float)base_image->width / downscaling)),
+      1u);
+  const uint32_t gain_map_height =
+      std::max(static_cast<uint32_t>(
+                   std::round((float)base_image->height / downscaling)),
+               1u);
   std::cout << "Creating a gain map of size " << gain_map_width << " x "
             << gain_map_height << "\n";
 
--- a/apps/avifgainmaputil/combine_command.cc
+++ b/apps/avifgainmaputil/combine_command.cc
@@ -97,8 +97,8 @@ avifResult CombineCommand::Run() {
         arg_alternate_cicp_.value().matrix_coefficients;
   }
 
-  const int downscaling = std::max<int>(1, arg_downscaling_);
-  const int rounding = downscaling / 2;
+  const uint32_t downscaling = std::max<int>(1, arg_downscaling_);
+  const uint32_t rounding = downscaling / 2;
   const uint32_t gain_map_width =
       std::max((base_image->width + rounding) / downscaling, 1u);
   const uint32_t gain_map_height =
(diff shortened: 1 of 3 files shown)


wantehchang:
Here are the MSVC warning messages:

```
D:\a\libavif\libavif\apps\avifgainmaputil\combine_command.cc(102): warning C4244: 'argument': conversion from 'float' to 'const _Ty', possible loss of data
        with
        [
            _Ty=uint32_t
        ]
D:\a\libavif\libavif\apps\avifgainmaputil\combine_command.cc(104): warning C4244: 'argument': conversion from 'float' to 'const _Ty', possible loss of data
        with
        [
            _Ty=uint32_t
        ]
```

apps/avifgainmaputil/tonemap_command.cc
```diff
@@ -52,7 +52,7 @@ avifResult TonemapCommand::Run() {
   avifContentLightLevelInformationBox clli_box = {};
   bool clli_set = false;
   if (arg_clli_str_.value().size() > 0) {
-    std::vector<uint32_t> clli;
+    std::vector<uint16_t> clli;
```
wantehchang, line 55: At lines 62-63 these are assigned to `clli_box.maxCLL` and `clli_box.maxPALL`, both of which are `uint16_t`.

So declaring the `clli` vector as `uint16_t` fixed the warnings.

If the `uint16_t` type is too small, we can change lines 62-63 to:

```
    if (clli[0] > UINT16_MAX || clli[1] > UINT16_MAX) {
      std::cerr << "Huge clli values > 2^16 - 1\n";
    }
    clli_box.maxCLL = static_cast<uint16_t>(clli[0]);
    clli_box.maxPALL = static_cast<uint16_t>(clli[1]);
```

vigneshvg: i think this the current cast is fine.

vigneshvg (approved):

apps/avifgainmaputil/combine_command.cc
```diff
@@ -98,10 +98,13 @@ avifResult CombineCommand::Run() {
...
-      std::round((float)base_image->width / downscaling), 1u);
-  const uint32_t gain_map_height = std::max<uint32_t>(
-      std::round((float)base_image->height / downscaling), 1u);
+  const uint32_t gain_map_width = std::max(
+      static_cast<uint32_t>(std::round((float)base_image->width / downscaling)),
+      1u);
+  const uint32_t gain_map_height =
+      std::max(static_cast<uint32_t>(
+                   std::round((float)base_image->height / downscaling)),
+               1u);
```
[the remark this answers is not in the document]
wantehchang: With the casts, this code is quite verbose. I just realized that we can do the rounded division using integer arithmetic. Please see the second commit in this pull request.

apps/avifgainmaputil/combine_command.cc
```diff
@@ -98,13 +98,11 @@ avifResult CombineCommand::Run() {
   }
 
   const int downscaling = std::max<int>(1, arg_downscaling_);
-  const uint32_t gain_map_width = std::max(
-      static_cast<uint32_t>(std::round((float)base_image->width / downscaling)),
-      1u);
+  const int rounding = downscaling / 2;
+  const uint32_t gain_map_width =
+      std::max((base_image->width + rounding) / downscaling, 1u);
```
y-guyon, line 103: `width` is `uint32_t`, `rounding` and `downscaling` should be unsigned too, no?

y-guyon (approved):

apps/avifgainmaputil/combine_command.cc
wantehchang: Done. When `int` and `unsigned int` are used in the same operation, `int` is implicitly converted to `unsigned int`. It's good to use the desired type explicitly.

y-guyon: I remember compiler warnings on Windows complaining about signed/unsigned mismatch. I understand it is fine according to the C standard though.

wantehchang: Compilers warn about signed/unsigned mismatch in comparisons. I have not seem compiler warnings about signed/unsigned mismatch in arithmetic operations (`+`, `-`, `*`, `/`).

PR #2296, opened 2024-07-18
Use -std=gnu2x -Wunused-result with Clang,GCC only

wantehchang:
Do not use those compiler options with MSVC.


CMakeLists.txt
```diff
@@ -325,9 +325,13 @@ if(AVIF_ENABLE_NODISCARD)
     else()
         unset(CMAKE_C_STANDARD)
         set_property(TARGET avif_obj PROPERTY C_STANDARD)
-        target_compile_options(avif_obj PUBLIC $<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:C>:-std=gnu2x>>)
+        if(CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID MATCHES "GNU")
+            target_compile_options(avif_obj PUBLIC $<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:C>:-std=gnu2x>>)
```
jzern, line 329: What needed `-std=gnu2x` as opposed to C++14/17?

jzern (approved):

CMakeLists.txt
wantehchang: This is the C standard,  not the C++ standard. The `[[...]]` attribute syntax was added in C23, so `[[nodiscard]]` requires C23. The `-std=gnu2x` option is intended to request C23.

Note: In C++, The `[[...]]` attribute syntax was added in C++11, and the `nodiscard` attribute was added in C++17, so `[[nodiscard]]` requires C++17.

I just realized that we should use the equivalent [`/std:clatest`](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170) option for MSVC. Please see the second commit.

jzern: Thanks, I didn't look at the context closely enough and assumed it was c++ related.

jzern (approved):
`/std:clatest` looks good, the only other options are `/std:c11` or `/std:c17`.

wantehchang:
Yes. I verified that the only C standard options are `/std:c11`, `/std:c17`, and `/std:clatest`, and only `/std:clatest` supports `[[nodiscard]]`.

CMakeLists.txt
```diff
@@ -325,9 +325,15 @@ if(AVIF_ENABLE_NODISCARD)
...
+        elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")
+            target_compile_options(avif_obj PUBLIC $<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:C>:/std:clatest>>)
+        endif()
+    endif()
+    if(CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID MATCHES "GNU")
+        target_compile_options(avif_obj PUBLIC $<BUILD_INTERFACE:-Wunused-result>)
```
wantehchang, line 335: [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wunused-result) says:

> `-Wno-unused-result`
> Do not warn if a caller of a function marked with attribute `warn_unused_result` (see [Declaring Attributes of Functions](https://gcc.gnu.org/onlinedocs/gcc/Function-Attributes.html)) does not use its return value. The default is `-Wunused-result`.

[Clang documentation](https://clang.llvm.org/docs/DiagnosticsReference.html#wunused-result) says:

> **-Wunused-result**
> 
> This diagnostic is enabled by default.

Since the default is `-Wunused-result`, can we delete this?

jzern: That sounds fine to me. It's safe to leave, but I doubt the default would be changed.

CMakeLists.txt
```diff
@@ -325,9 +325,15 @@ if(AVIF_ENABLE_NODISCARD)
     else()
         unset(CMAKE_C_STANDARD)
         set_property(TARGET avif_obj PROPERTY C_STANDARD)
-        target_compile_options(avif_obj PUBLIC $<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:C>:-std=gnu2x>>)
+        if(CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID MATCHES "GNU")
```
vrabaud, line 328: Can also be written as:
```CMake
target_compile_options(avif_obj PUBLIC 
     $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:GNU>>:
          -std=gnu2x>
     $<$<CXX_COMPILER_ID:MSVC>:
          /std:clatest>)
```

CMakeLists.txt
vrabaud: yes, please delete

CMakeLists.txt
wantehchang: Thanks for the suggestion. This is harder to read. Also it omits the `$<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:C>` part used in the current code. Is that not neeeded?

CMakeLists.txt
wantehchang: Done.

PR #2298, opened 2024-07-18
Revert "Put AVIF_NODISCARD after AVIF_API" again

wantehchang:
This reverts commit da268fb447f3b9998ee00af7121f740d8c890fe7.

I tested this with Visual Studio 2022 Version 17.10.3. It does not work. I must have made a mistake when I tested this with Visual Studio last time.

I found two webpages that support my test results:
https://stackoverflow.com/questions/67502217/warning-c5240-nodiscard-attribute-is-ignored-in-this-syntactic-position https://github.com/microsoft/STL/issues/1702

I think this issue is settled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ The changes are relative to the previous release, unless the baseline is specifi
 
 ### Changed since 1.1.0
 * In avif.h, change "AVIF_API AVIF_NODISCARD" back to "AVIF_NODISCARD AVIF_API"
-  to fix clang-cl and MSVC compilation errors on Windows.
+  to fix clang-cl and MSVC compilation errors in the shared library build on
+  Windows.
 * Fix -DAVIF_GTEST=SYSTEM, https://github.com/AOMediaCodec/libavif/issues/2258.
 * Fix infe_type and codec_config_type wrongly read as byte-aligned fields in the
   experimental feature AVIF_ENABLE_EXPERIMENTAL_METAV1.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ The changes are relative to the previous release, unless the baseline is specifi
 ## [Unreleased]
 
 ### Changed since 1.1.0
-* In avif.h, fix clang-cl compilation errors of "AVIF_API AVIF_NODISCARD" on
-  Windows.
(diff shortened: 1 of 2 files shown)


wantehchang:
Here is my latest test of a standalone file.

The compiler version:

```
C:\src>cl
Microsoft (R) C/C++ Optimizing Compiler Version 19.40.33811 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

usage: cl [ option... ] filename... [ /link linkoption... ]

```

The bad code:

```
C:\src>type bad.c
__declspec(dllimport) [[nodiscard]] int Add(int x, int y);

__declspec(dllexport) [[nodiscard]] int Double(int x) {
  return Add(x, x);
}
C:\src>cl /std:clatest /W4 /c bad.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.40.33811 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

bad.c
bad.c(1): error C2059: syntax error: 'type'
bad.c(3): error C2059: syntax error: 'type'

```

The good code:

```
C:\src>type good.c
[[nodiscard]] __declspec(dllimport) int Add(int x, int y);

[[nodiscard]] __declspec(dllexport) int Double(int x) {
  return Add(x, x);
}
C:\src>cl /std:clatest /W4 /c good.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.40.33811 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

good.c

```

wantehchang:
I forgot to request code review. PTAL. Thanks!

vrabaud:
How come the CI does not catch it?

vrabaud (approved):

wantehchang:
To catch the compilation error (either MSVC or clang-cl), we need a shared library build (`-DBUILD_SHARED_LIBS=ON`).

There are some two build errors in our shared library build. One is easy to fix (two gtest-based tests need to link with the`avif_internal` library). The other is a strange linker unresolved symbol error that I cannot reproduce locally. I haven't figured out what caused it, but one of my experiments worked, so at least we have a workaround. (I call it a workaround rarher than a fix because I cannot explain why it works.)

Finally, we cannot run the tests in a shared library build on Windows because the tests and `avif.dll` are in different directories, so the tests cannot find `avif.dll`. We should be able to fix this by setting the `Path` envirnonment variable before running the tests, or by copying `avif.dll` to the tests directory.

jzern (commented):
lgtm

Issue #2300, opened 2024-07-19
CMake fails in tests/CMakeLists.txt "No known features for CXX compiler" if -DAVIF_BUILD_TESTS=ON -DAVIF_BUILD_APPS=OFF


wantehchang:
Steps to reproduce:

On a Linux computer, run the cmake command:

cmake .. -G Ninja -DAVIF_CODEC_AOM=LOCAL -DAVIF_LIBYUV=OFF -DAVIF_BUILD_TESTS=ON -DAVIF_GTEST=LOCAL -DAVIF_BUILD_APPS=OFF

It fails with these messages:

```
-- libavif(AVIF_GTEST=LOCAL): compiled library not found in ext/googletest; using FetchContent
-- libavif(AVIF_LOCAL_GTEST): fetching and configuring googletest
-- Found Python3: /usr/bin/python3 (found version "3.11.8") found components: Interpreter 
-- libavif(AVIF_LOCAL_GTEST): fetching and configuring googletest - complete
-- Some tests are skipped because libsharpyuv is unavailable.
-- FuzzTest targets are disabled because AVIF_ENABLE_FUZZTEST is OFF.
-- Configuring done (26.6s)
CMake Error in tests/CMakeLists.txt:
  No known features for CXX compiler

  ""

  version .


CMake Generate step failed.  Build files cannot be regenerated correctly.
```

If I run the cmake command again, it succeeds, and I can build and run the tests (using `ctest`) successfully.

If I change `-DAVIF_BUILD_APPS=OFF` to `-DAVIF_BUILD_APPS=ON`, it also works.

wantehchang:
Here is a variant of this bug. Run the command:

```
cmake .. -G Ninja -DAVIF_CODEC_AOM=LOCAL -DAVIF_LIBYUV=OFF -DAVIF_BUILD_TESTS=ON -DAVIF_GTEST=LOCAL
```

I get these CMake error messages:

```
-- Configuring done (0.9s)
CMake Error: CMake can not determine linker language for target: aviftest_helpers
CMake Error: CMake can not determine linker language for target: aviftest_helpers_internal
CMake Error: CMake can not determine linker language for target: avifincrtest_helpers
CMake Error: CMake can not determine linker language for target: avifallocationtest
CMake Error: CMake can not determine linker language for target: avifalphanoispetest
...
CMake Error: CMake can not determine linker language for target: avifutilstest
CMake Error: CMake can not determine linker language for target: avify4mtest
[...]

PR #2302, opened 2024-07-22
 Fix "No known features for CXX compiler" CMake error

vrabaud:
Fix "No known features for CXX compiler" CMake error
    
- move relevant options at the top. AVIF_ENABLE_GTEST defaults
   to ON so it should be defined/set before it is checked
- update AVIF_ENABLE_FUZZTEST doc string to fix https://github.com/AOMediaCodec/libavif/issues/2314
- isolate repro_avif_decode_fuzzer build with CMAKE_CXX_COMPILER_LOADED

This fixes https://github.com/AOMediaCodec/libavif/issues/2300

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,14 +122,28 @@ set_local_or_system_option(
     "GTEST" OFF
     "Build the GoogleTest framework by providing your own copy of the repo in ext/googletest (see Local Builds in README)"
 )
+option(AVIF_BUILD_APPS "Build avif apps." OFF)
+option(AVIF_BUILD_TESTS "Build avif tests." OFF)
 option(AVIF_LOCAL_FUZZTEST
        "Build the Google FuzzTest framework by providing your own copy of the repo in ext/fuzztest (see Local Builds in README)"
        OFF
 )
-
 option(
     AVIF_ENABLE_COMPLIANCE_WARDEN
     "Check all avifEncoderFinish() output for AVIF specification compliance. Depends on gpac/ComplianceWarden which can be added with ext/compliance_warden.sh"
+    OFF
+)
+option(
+    AVIF_ENABLE_GOLDEN_TESTS
+    "Build tests that compare encoding outputs to golden files. Needs AVIF_BUILD_APPS=ON, and depends on MP4box which can be built with ext/mp4box.sh"
+    OFF
+)
+option(AVIF_ENABLE_GTEST
+       "Build avif C++ tests, which depend on GoogleTest. Requires GoogleTest. Has no effect unless AVIF_BUILD_TESTS is ON." ON
+)
+option(
+    AVIF_ENABLE_FUZZTEST
+    "Build avif fuzztest targets. Requires Google FuzzTest. Has no effect unless AVIF_BUILD_TESTS is ON."
     OFF
 )
 
@@ -646,22 +660,6 @@ endif()
 if(NOT SKIP_INSTALL_ALL)
     include(GNUInstallDirs)
 endif()
-
-option(AVIF_BUILD_APPS "Build avif apps." OFF)
-option(AVIF_BUILD_TESTS "Build avif tests." OFF)
-option(
-    AVIF_ENABLE_GOLDEN_TESTS
-    "Build tests that compare encoding outputs to golden files. Needs AVIF_BUILD_APPS=ON, and depends on MP4box which can be built with ext/mp4box.sh"
-    OFF
-)
-option(AVIF_ENABLE_GTEST
-       "Build avif C++ tests, which depend on GoogleTest. Requires GoogleTest. Has no effect unless AVIF_BUILD_TESTS is ON." ON
-)
-option(
-    AVIF_ENABLE_FUZZTEST
-    "Build avif fuzztest targets. Requires Google FuzzTest. Has no effect unless AVIF_BUILD_TESTS and AVIF_ENABLE_GTEST are ON."
-    OFF
-)
 
 if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST)))
     if(AVIF_ZLIBPNG STREQUAL "OFF")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,10 +124,6 @@ set_local_or_system_option(
 )
 option(AVIF_BUILD_APPS "Build avif apps." OFF)
 option(AVIF_BUILD_TESTS "Build avif tests." OFF)
-option(AVIF_LOCAL_FUZZTEST
-       "Build the Google FuzzTest framework by providing your own copy of the repo in ext/fuzztest (see Local Builds in README)"
-       OFF
-)
 option(
     AVIF_ENABLE_COMPLIANCE_WARDEN
     "Check all avifEncoderFinish() output for AVIF specification compliance. Depends on gpac/ComplianceWarden which can be added with ext/compliance_warden.sh"
@@ -135,16 +131,18 @@ option(
 )
 option(
     AVIF_ENABLE_GOLDEN_TESTS
-    "Build tests that compare encoding outputs to golden files. Needs AVIF_BUILD_APPS=ON, and depends on MP4box which can be built with ext/mp4box.sh"
+    "Build tests that compare encoding outputs to golden files. Needs AVIF_BUILD_APPS=ON and AVIF_BUILD_TESTS=ON, and depends on MP4box which can be built with ext/mp4box.sh"
     OFF
 )
 option(AVIF_ENABLE_GTEST
        "Build avif C++ tests, which depend on GoogleTest. Requires GoogleTest. Has no effect unless AVIF_BUILD_TESTS is ON." ON
 )
-option(
-    AVIF_ENABLE_FUZZTEST
-    "Build avif fuzztest targets. Requires Google FuzzTest. Has no effect unless AVIF_BUILD_TESTS is ON."
-    OFF
+option(AVIF_ENABLE_FUZZTEST "Build avif fuzztest targets. Requires Google FuzzTest. Has no effect unless AVIF_BUILD_TESTS is ON."
+       OFF
+)
+option(AVIF_LOCAL_FUZZTEST
+       "Build the Google FuzzTest framework by providing your own copy of the repo in ext/fuzztest (see Local Builds in README)"
+       OFF
 )
 
 # Whether the libavif library uses c++ indirectly (e.g. through linking to libyuv).
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -47,7 +47,7 @@ if(AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
     target_link_libraries(aviftest_helpers_internal avif_apps_internal avif_internal)
 endif()
 
-if(CMAKE_CXX_COMPILER_VERSION)
+if(CMAKE_CXX_COMPILER_LOADED)
     # Fuzz target without any fuzzing engine dependency. For easy reproduction of oss-fuzz issues.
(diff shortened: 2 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -47,14 +47,17 @@ if(AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
...
 
-# Fuzz target without any fuzzing engine dependency. For easy reproduction of oss-fuzz issues.
-add_executable(repro_avif_decode_fuzzer oss-fuzz/avif_decode_fuzzer.cc oss-fuzz/repro_fuzz.cc)
-set_target_properties(repro_avif_decode_fuzzer PROPERTIES LINKER_LANGUAGE "CXX")
-target_link_libraries(repro_avif_decode_fuzzer avif)
-# The test below exists for coverage and as a usage example: repro_avif_decode_fuzzer [reproducer file path]
-add_test(NAME repro_avif_decode_fuzzer COMMAND repro_avif_decode_fuzzer
-                                               ${CMAKE_CURRENT_SOURCE_DIR}/data/color_grid_alpha_nogrid.avif
-)
+if(AVIF_ENABLE_FUZZTEST)
```
wantehchang, line 50: This change doesn't seem necessary. `repro_avif_decode_fuzzer` isn't a fuzz test. It is a utility program for reproducing a fuzz test error without building a real fuzz test (linked with a fuzzing engine). So it is not clear whether it should be controlled by `AVIF_ENABLE_FUZZTEST`.

CMakeLists.txt
```diff
@@ -122,28 +122,34 @@ set_local_or_system_option(
     "GTEST" OFF
     "Build the GoogleTest framework by providing your own copy of the repo in ext/googletest (see Local Builds in README)"
 )
+
```
wantehchang, line 125: Does this pull request also fix or partially address https://github.com/AOMediaCodec/libavif/issues/2277? I am wondering if this pull request replaces https://github.com/AOMediaCodec/libavif/pull/2290.

CMakeLists.txt
```diff
@@ -571,6 +576,15 @@ if(NOT AVIF_CODEC_AOM_ENABLED
     message(WARNING "libavif: No decoding library is enabled.")
 endif()
 
+if(AVIF_BUILD_APPS
+   OR AVIF_CODEC_LIBGAV1_ENABLED
+   OR AVIF_ENABLE_COMPLIANCE_WARDEN
+   OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST))
```
wantehchang, line 582: Optional: If the documentation for `AVIF_ENABLE_FUZZTEST` at lines 145-149 is accurate:

```
option(
    AVIF_ENABLE_FUZZTEST
    "Build avif fuzztest targets. Requires Google FuzzTest. Has no effect unless AVIF_BUILD_TESTS and AVIF_ENABLE_GTEST are ON."
    OFF
)
```

then the condition `AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST` can simplified to `AVIF_ENABLE_GTEST`.

Since this is an independent issue, you don't need to look into it in this pull request.

CMakeLists.txt
```diff
@@ -571,6 +576,15 @@ if(NOT AVIF_CODEC_AOM_ENABLED
     message(WARNING "libavif: No decoding library is enabled.")
 endif()
 
+if(AVIF_BUILD_APPS
```
wantehchang, line 579: IMPORTANT: Keep line 151 with this if statement. Alternatively, use an if-else statement.

CMakeLists.txt
vrabaud: Almost but not quite: the bug still happens with #2290 2290

CMakeLists.txt
vrabaud: No, because `AVIF_USE_CXX` is set to ON above (e.g. with Compliance Warden)

CMakeLists.txt
wantehchang: By "the bug still happens with #2290", were you referring to https://github.com/AOMediaCodec/libavif/issues/2277 or https://github.com/AOMediaCodec/libavif/issues/2300?

I was asking about https://github.com/AOMediaCodec/libavif/issues/2277.

CMakeLists.txt
wantehchang: You are right. Thanks.

wantehchang:
I will review this pull request after https://github.com/AOMediaCodec/libavif/pull/2290 is merged and this is rebased.

wantehchang:
Vincent: To speed up code review, you can ask Yannis to review this when this is rebased.

CMakeLists.txt
wantehchang: I looked into this. I believe the documentation for `AVIF_ENABLE_FUZZTEST` is incorrect. It should say "Has no effect unless AVIF_BUILD_TESTS is ON." You can ignore this comment.

CMakeLists.txt
vrabaud: With the rebase, this PR only fixes #2300

tests/CMakeLists.txt
vrabaud: I changed the build condition to AVIF_LIB_USE_CXX

CMakeLists.txt
```diff
@@ -122,16 +122,30 @@ set_local_or_system_option(
...
        OFF
 )
-
 option(
     AVIF_ENABLE_COMPLIANCE_WARDEN
     "Check all avifEncoderFinish() output for AVIF specification compliance. Depends on gpac/ComplianceWarden which can be added with ext/compliance_warden.sh"
     OFF
 )
+option(
+    AVIF_ENABLE_GOLDEN_TESTS
```
wantehchang, line 137: Optional: (Can do this in a separate CL.) Should we declare the `AVIF_ENABLE_GOLDEN_TESTS` option with `cmake_dependent_option()` and make it depend on `AVIF_BUILD_TESTS AND AVIF_BUILD_APPS`?

CMakeLists.txt
```diff
@@ -122,16 +122,30 @@ set_local_or_system_option(
...
+    "Build tests that compare encoding outputs to golden files. Needs AVIF_BUILD_APPS=ON, and depends on MP4box which can be built with ext/mp4box.sh"
```
wantehchang, line 138: Code inspection shows (it's also obvious) that this also needs `AVIF_BUILD_TESTS=ON`.

It actually also requires `AVIF_CODEC_AOM_ENCODE`, but that can be considered as implementation detail.

CMakeLists.txt
```diff
@@ -122,16 +122,30 @@ set_local_or_system_option(
...
+    OFF
+)
+option(AVIF_ENABLE_GTEST
```
wantehchang, line 141: Optional: (Can do this in a separate CL.) Should we declare the `AVIF_ENABLE_GTEST` and `AVIF_ENABLE_FUZZTEST` options with `cmake_dependent_option()` and make them depend on `AVIF_BUILD_TESTS`?

CMakeLists.txt
```diff
@@ -122,16 +122,30 @@ set_local_or_system_option(
...
+       "Build avif C++ tests, which depend on GoogleTest. Requires GoogleTest. Has no effect unless AVIF_BUILD_TESTS is ON." ON
+)
+option(
+    AVIF_ENABLE_FUZZTEST
+    "Build avif fuzztest targets. Requires Google FuzzTest. Has no effect unless AVIF_BUILD_TESTS is ON."
+    OFF
+)
```
wantehchang, line 148: Nit: Move the `AVIF_LOCAL_FUZZTEST` option here, because it is related to the `AVIF_ENABLE_FUZZTEST` option.

tests/CMakeLists.txt
```diff
@@ -47,14 +47,17 @@ if(AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
...
 
-# Fuzz target without any fuzzing engine dependency. For easy reproduction of oss-fuzz issues.
-add_executable(repro_avif_decode_fuzzer oss-fuzz/avif_decode_fuzzer.cc oss-fuzz/repro_fuzz.cc)
-set_target_properties(repro_avif_decode_fuzzer PROPERTIES LINKER_LANGUAGE "CXX")
-target_link_libraries(repro_avif_decode_fuzzer avif)
-# The test below exists for coverage and as a usage example: repro_avif_decode_fuzzer [reproducer file path]
-add_test(NAME repro_avif_decode_fuzzer COMMAND repro_avif_decode_fuzzer
-                                               ${CMAKE_CURRENT_SOURCE_DIR}/data/color_grid_alpha_nogrid.avif
-)
+if(AVIF_LIB_USE_CXX)
```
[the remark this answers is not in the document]
vrabaud: I was trying to minimize the C++ usage for no special reason, let's enable C++ when `AVIF_BUILD_TESTS` is `ON`.

CMakeLists.txt
vrabaud: sure

CMakeLists.txt
vrabaud: we can also add a dependent option on that

CMakeLists.txt
vrabaud: that is probably cleaner indeed, though the doc is right here at least (compare to `AVIF_ENABLE_GOLDEN_TESTS`)

CMakeLists.txt
vrabaud: done. This one could also be made dependent, or actually be removed by adding the proper `LocalFuzztest.cmake` in './cmake/`.

tests/CMakeLists.txt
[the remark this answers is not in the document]
wantehchang: Understood. Another idea that just occurred to me is to use the guard `AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST` (and undo the change to the top-level CMakeLists.txt) to minimize the C++ usage. The downside is that it is not clear why we build `repro_avif_decode_fuzzer` under this condition, so we will need to add a comment to explain the guard.

wantehchang (approved):

vrabaud:
I ended up guarding the test with `if(CMAKE_CXX_COMPILER_VERSION)`.
Note, if enabling AOM:
```
rm -fr * && cmake .. -G Ninja -DAVIF_CODEC_AOM=LOCAL -DAVIF_LIBYUV=OFF -DAVIF_BUILD_TESTS=ON -DAVIF_GTEST=LOCAL -DAVIF_BUILD_APPS=OFF
```
CXX is enabled indirectly by libaom's CMake.
Not if you only enable dav1d

tests/CMakeLists.txt
```diff
@@ -47,8 +47,7 @@ if(AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
     target_link_libraries(aviftest_helpers_internal avif_apps_internal avif_internal)
 endif()
 
-if(AVIF_LIB_USE_CXX)
-    # TODO(vrabaud): merge with oss-fuzz.
+if(CMAKE_CXX_COMPILER_VERSION)
```
wantehchang, line 50: This is a clever solution. I found a Stack Overflow page on this issue:
https://stackoverflow.com/questions/32389273/detect-project-language-in-cmake

vrabaud: perfect ! `CMAKE_CXX_COMPILER_LOADED` is what I was looking for

wantehchang: `CMAKE_CXX_COMPILER_LOADED` is not documented. It may be safer to use `CMAKE_CXX_COMPILER_VERSION` (which is not a good match but is documented) or `ENABLED_LANGUAGES`.

wantehchang: I was wrong. It is documented! https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LOADED.html

wantehchang (approved):

wantehchang (approved):

PR #2303, opened 2024-07-22
Link Sample Transform tests with avif_internal lib

wantehchang:


tests/CMakeLists.txt
```diff
@@ -94,8 +94,8 @@ endif()
 
 if(AVIF_ENABLE_GTEST)
     if(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
-        add_avif_gtest_with_data(avif16bittest)
-        add_avif_gtest(avifsampletransformtest)
+        add_avif_internal_gtest_with_data(avif16bittest)
+        add_avif_internal_gtest(avifsampletransformtest)
```
wantehchang, line 98: Yannis: This means none of the CI workflows test the experimental features in a shared library build. Could you suggest a suitable CI workflow where we can do this? Or do we need to add a new CI workflow?

wantehchang:
To reproduce this bug, run these commands:

```
cmake .. -G Ninja -DBUILD_SHARED_LIBS=ON -DAVIF_CODEC_AOM=LOCAL \
    -DAVIF_LIBYUV=OFF -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON \
    -DAVIF_GTEST=LOCAL -DAVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM=ON

ninja
```

Note: `-DBUILD_SHARED_LIBS=ON` is the default and could be omitted. I added it for clarity. `-DAVIF_ENABLE_GTEST=ON` is also the default. I added it to work around https://github.com/AOMediaCodec/libavif/issues/2300.

tests/CMakeLists.txt
y-guyon: Since it is in `ci-unix-static.yml` I guess it makes sense to add it to `ci-unix-shared-local.yml`. I expect the other experimental flags to be untested in a shared library build too.

y-guyon (approved):

jzern (approved):

tests/CMakeLists.txt
wantehchang: Thanks for the suggestion. I will do that in https://github.com/AOMediaCodec/libavif/pull/2307 so that I can first show that these tests have linker errors in shared library builds when the experimental features are enabled.

vrabaud (approved):

wantehchang:
I verified this fix in https://github.com/AOMediaCodec/libavif/pull/2307.

Before this fix:
https://github.com/AOMediaCodec/libavif/actions/runs/10048504681

After this fix:
https://github.com/AOMediaCodec/libavif/actions/runs/10048720740

PR #2305, opened 2024-07-22
Move gainmap function definitions to gainmap.c

wantehchang:
Move two gainmap function definitions (avifGainMapCreate and avifGainMapDestroy).

This is a good change on its own, but the reason I wanted to move them is to work around a strange unresolved symbol linker error on Windows if I enable -DBUILD_SHARED_LIBS=ON in ci-windows-installed.yml:

  avif_apps_internal.lib(avifjpeg.c.obj) : error LNK2019: unresolved
  external symbol __imp_avifGainMapMetadataDoubleToFractions referenced
  in function avifJPEGParseGainMapXMPProperties

  tests\avif16bittest.exe : fatal error LNK1120: 1 unresolved externals

See https://github.com/AOMediaCodec/libavif/actions/runs/10046011390.

I have debugged this for a long time but haven't found what is wrong. I also cannot reproduce this unresolved symbol linker error locally.

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -1163,29 +1163,3 @@ void avifCodecVersions(char outBuffer[256])
         append(&writePos, &remainingLen, availableCodecs[i].version());
     }
 }
-
-#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
-avifGainMap * avifGainMapCreate(void)
-{
-    avifGainMap * gainMap = (avifGainMap *)avifAlloc(sizeof(avifGainMap));
-    if (!gainMap) {
-        return NULL;
-    }
-    memset(gainMap, 0, sizeof(avifGainMap));
-    gainMap->altColorPrimaries = AVIF_COLOR_PRIMARIES_UNSPECIFIED;
-    gainMap->altTransferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
-    gainMap->altMatrixCoefficients = AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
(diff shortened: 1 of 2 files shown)


jzern (approved):

wantehchang:
@vrabaud Vincent: Just FYI. This is the workaround I told you about in Chat.

James: Thanks for the review. Here is a link to a test of ci-windows-installed.yml with `-DBUILD_SHARED_LIBS=ON` and with this workaround applied:
https://github.com/AOMediaCodec/libavif/actions/runs/10046453219

You will see that the builds succeeded and we were failing in tests (because the tests linked with avif.dll cannot find avif.dll).

In the test build I also passed `-v` to the `ninja` command. It showed that we linked `avif16bittest.exe` with `avif_apps_internal.lib` and `avif_internal.lib`, in that order.

I also added a command to list the object files in the `avif_internal.lib` static library. It verified that `gainmap.c.obj` is in `avif_internal.lib`.

So everything looks correct. I don't know what is wrong.

I said earlier that I cannot reproduce this unresolved symbol linker error locally. I should clarify that I did not install the dependencies as vcpkg packages. I built them locally from source code. The ci-windows-installed.yml workflow uses a huge CMake toolchain file (presumably to help cmake find the vcpkg packages). My current speculation is that the linker error is caused by that CMake toolchain file.

wantehchang:
I found that Chrome's third_party/libavif/BUILD.gn does not compile src/gainmap.c even though it defines `AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP` when compiling libavif. Apparently the code that was originally in src/gainmap.c is used by apps/shared/avifjpeg.c only, not needed by Chrome. I will look into that and write a follow-up pull request.

PR #2306, opened 2024-07-22
Don't add avif_obj's object files to avif_apps_obj

wantehchang:
Do not add the object files in the avif_obj library to the avif_apps_obj
library. avif_apps_obj should contain only the object files compiled
from source files in the apps/shared/ and from
third_party/iccjpeg/iccjpeg.c.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -684,8 +684,8 @@ if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE
     endif()
 
     add_library(
-        avif_apps_obj OBJECT $<TARGET_OBJECTS:avif_obj> apps/shared/avifexif.c apps/shared/avifjpeg.c apps/shared/avifpng.c
-                             apps/shared/avifutil.c apps/shared/iccmaker.c apps/shared/y4m.c third_party/iccjpeg/iccjpeg.c
+        avif_apps_obj OBJECT apps/shared/avifexif.c apps/shared/avifjpeg.c apps/shared/avifpng.c apps/shared/avifutil.c
+                             apps/shared/iccmaker.c apps/shared/y4m.c third_party/iccjpeg/iccjpeg.c
     )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -719,7 +719,7 @@ if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE
     # avif_apps_internal is to use when linking to avif_internal.
     if(BUILD_SHARED_LIBS)
         add_library(avif_apps_internal STATIC)
-        target_link_libraries(avif_apps_internal PUBLIC avif_internal PRIVATE avif_apps_obj)
+        target_link_libraries(avif_apps_internal PRIVATE avif_apps_obj PUBLIC avif_internal)
```
wantehchang, line 722: The order in which `avif_apps_ob` and `avif_internal` are listed may not matter. I can revert this change if you prefer.

jzern: Having them in alphabetical order is fine. I wonder how consistent other entries are.

jzern (approved):

jzern (approved):

vrabaud (approved):

wantehchang:
I found that without this fix, the `avif_apps` static library does NOT contain the object files in the `avif_obj` OBJECT library. For example, in a static library build on Linux:

```
$ ar t libavif_apps.a
avifexif.c.o
avifjpeg.c.o
avifpng.c.o
avifutil.c.o
iccmaker.c.o
y4m.c.o
iccjpeg.c.o
```

and in a shared library build on Linux:

```
$ ar t libavif_apps.a
avifexif.c.o
avifjpeg.c.o
avifpng.c.o
avifutil.c.o
iccmaker.c.o
y4m.c.o
iccjpeg.c.o
$ ar t libavif_apps_internal.a
avifexif.c.o
avifjpeg.c.o
avifpng.c.o
avifutil.c.o
iccmaker.c.o
y4m.c.o
iccjpeg.c.o
```

So it seems that cmake is smart enough to exclude the object files in `avif_obj` from the `avif_apps` static library.

PR #2309, opened 2024-07-22
Add ci-windows-shared-local.yml

wantehchang:
Add a CI workflow for building shared libraries with local dependencies on Windows.


.github/workflows/ci-windows-shared-local.yml
```diff
@@ -0,0 +1,59 @@
+# Do a shared library build with local dependencies.
+
+name: CI Windows Shared Local
+on: [push, pull_request]
```
y-guyon, line 4: Is it intentional to run it everytime or should it be reserved to run on merge only?  
The intent is to avoid a long list of presubmit in the PR GitHub UI and to save resources.

.github/workflows/ci-windows-shared-local.yml
```diff
@@ -0,0 +1,59 @@
...
+        -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON
+        -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_DAV1D=LOCAL
+        -DAVIF_JPEG=LOCAL -DAVIF_LIBSHARPYUV=LOCAL -DAVIF_LIBXML2=LOCAL
+        -DAVIF_LIBYUV=LOCAL -DAVIF_ZLIBPNG=LOCAL
+        -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON
+        -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON -DAVIF_GTEST=LOCAL
+        -DAVIF_ENABLE_EXPERIMENTAL_YCGCO_R=ON
+        -DAVIF_ENABLE_EXPERIMENTAL_GAIN_MAP=ON
+        -DAVIF_ENABLE_EXPERIMENTAL_METAV1=ON
+        -DAVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM=ON
```
y-guyon, line 52: Just a thought: we could have a CI with a matrix testing all combinations of these experimental flags

.github/workflows/ci-windows-shared-local.yml
```diff
@@ -0,0 +1,59 @@
...
+    - name: Set clang-cl as compiler
+      if: matrix.compiler == 'clang-cl'
+      run: |
+        echo "AVIF_CMAKE_C_COMPILER=-DCMAKE_C_COMPILER=clang-cl" >> $env:GITHUB_ENV
+        echo "AVIF_CMAKE_CXX_COMPILER=-DCMAKE_CXX_COMPILER=clang-cl" >> $env:GITHUB_ENV
+
+    - name: Prepare libavif (cmake)
+      run: >
+        cmake -G Ninja -S . -B build
+        -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON
```
y-guyon, line 43: Should `Debug` be tested too? I remember difficulties when linking differently built dependencies and wonder if that should be exercised. Probably not worth the effort.

.github/workflows/ci-windows-shared-local.yml
```diff
@@ -0,0 +1,59 @@
...
+        echo "AVIF_CMAKE_CXX_COMPILER=" >> $env:GITHUB_ENV
+    - name: Set clang-cl as compiler
+      if: matrix.compiler == 'clang-cl'
+      run: |
+        echo "AVIF_CMAKE_C_COMPILER=-DCMAKE_C_COMPILER=clang-cl" >> $env:GITHUB_ENV
+        echo "AVIF_CMAKE_CXX_COMPILER=-DCMAKE_CXX_COMPILER=clang-cl" >> $env:GITHUB_ENV
+
+    - name: Prepare libavif (cmake)
+      run: >
+        cmake -G Ninja -S . -B build
```
y-guyon, line 42: By curiosity, why are we specifying the generator? Why ninja?

CMakeLists.txt
```diff
@@ -594,6 +594,14 @@ if(BUILD_SHARED_LIBS)
     if(AVIF_USE_CXX)
         set_target_properties(avif PROPERTIES LINKER_LANGUAGE "CXX")
     endif()
+    if(WIN32 AND AVIF_BUILD_TESTS)
+        add_custom_command(
+            TARGET avif
+            POST_BUILD
+            COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_BINARY_DIR}/avif.dll ${CMAKE_CURRENT_BINARY_DIR}/tests
+            COMMENT "Copying avif.dll to the tests binary directory"
```
y-guyon, line 602: I assume this is a common pattern.

y-guyon (approved):

.github/workflows/ci-windows-shared-local.yml
vrabaud: It handles multicore machines automatically

vrabaud:
Why not do a matrix in ci-windows.yml instead ? By adding
```
build type: [Debug, Release]
shared_libs: [OFF, ON]
```

.github/workflows/ci-windows-shared-local.yml
```diff
@@ -0,0 +1,59 @@
...
+        compiler: [msvc, clang-cl]
+
+    steps:
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+    - uses: ./.github/actions/setup-windows
+      with:
+        codec-aom: 'LOCAL'
+        codec-dav1d: 'LOCAL'
+        extra-cache-key: ${{ matrix.compiler }}
+    - name: Leave compiler as default
```
vrabaud, line 29: Setting those variables should probably be put in .github/actions/setup-windows

.github/workflows/ci-windows-shared-local.yml
y-guyon: `cmake --build [path] --parallel` too no?

.github/workflows/ci-windows-shared-local.yml
wantehchang: I didn't consider this issue. I copied this line from some other file (most likely ci-windows-installed.yml).

Please advise what we should use here. Thanks!

y-guyon: For example:

https://github.com/AOMediaCodec/libavif/blob/676a6cb82c0261cf5474c086f8e94f0f65ad4645/.github/workflows/ci-unix-static.yml#L3-L9

wantehchang:
I am changing this pull request to a draft because there seem to be linker warnings, and there is an unresolved symbol linker error if experimental features are enabled.

.github/workflows/ci-windows-shared-local.yml
wantehchang: I am familiar with Ninja from working on Chrome. So it is a personal preference. The Meson build tool (used by dav1d) also uses Ninja.

Ninja defaults to running commands in parallel, so `ninja` should be equivalent to `cmake --build . --parallel`. In a script it doesn't matter, but if we are typing the commands, `ninja` is much shorter than `cmake --build . --parallel` or `cmake --build . -j`.

.github/workflows/ci-windows-shared-local.yml
wantehchang: We can looking into adding `Debug` later. Right now I am focusing on getting a clean (no compiler or linker warnings) `Release` build.

CMakeLists.txt
wantehchang: I moved this change to its own pull request (https://github.com/AOMediaCodec/libavif/pull/2328). In that pull request I added a comment to answer your question.

wantehchang (commented):
I looked at the linker warnings carefully. I found that those linker warnings are only emitted when the linker error I reported in https://github.com/AOMediaCodec/libavif/issues/2339 is there. So it should be fine to add this CI workflow as long as we avoid the linker error by turning off the gain map experimental feature. We can enable the other three experimental features.

.github/workflows/ci-windows-shared-local.yml
wantehchang: Done. Thanks!

wantehchang (commented):
I removed the v1.1.1 milestone from this pull request.

y-guyon (approved):

Issue #2311, opened 2024-07-23
Grid image arranges tiles in wrong order if IDs in the `iref` are not sequential


farindk:
libavif apparently assumes that the tile IDs of a grid image are sequential. If that is not the case, the tiles are composed into the output image in an incorrect arrangement.

Here is an [example image](https://cloud.dirk-farin.de/s/8nW38ym2DQjHBdt) that is decoded by libavif like this:

![image](https://github.com/user-attachments/assets/d5215bff-6c15-458f-835f-5ed47ae764ec)

y-guyon:
> libavif apparently assumes that the tile IDs of a grid image are sequential. If that is not the case, the tiles are composed into the output image in an incorrect arrangement.

The tile IDs should not matter at all. HEIF spec ISO/IEC 23008-12 says:

```
The input images are inserted in row-major order, top-row first, left to right, in the order of
SingleItemTypeReferenceBox of type 'dimg' for this derived image item within the ItemReferenceBox. 
```

However libavif does not respect that pattern as of today (libavif is used in Chrome for example):
![2345 avif](https://github.com/user-attachments/assets/45342ea8-eaba-41f3-941c-208635454d4b)

![5432 avif](https://github.com/user-attachments/assets/0c31f46f-12e8-49d7-9839-e67a6a512efe)

y-guyon:
Not sure what GIMP 2.10.36 uses under the hood but it opens the out-of-order `dimg` association file correctly.

y-guyon:
(Thank you for taking the time to report this issue by the way.)

wantehchang:
To test the fix for this issue, we also need an image grid in which multiple grid tiles reference the same image item. (In the extreme case, all grid tiles reference the same image item.)

@farindk Dirk: do yo happen to have such an image available?

y-guyon:
> To test the fix for this issue, we also need an image grid in which multiple grid tiles reference the same image item. (In the extreme case, all grid tiles reference the same image item.)

I manually edited
https://github.com/AOMediaCodec/libavif/blob/676a6cb82c0261cf5474c086f8e94f0f65ad4645/tests/data/sofa_grid1x5_420.avif
to repeat the same tile twice in `dimg`:
![sofa_grid1x5_420_dimg_repeat avif](https://github.com/user-attachments/assets/98fb5c42-0ad0-4243-a2cd-3eafd8ea50c1)

The image does not work in Chrome (libavif).  
[...]

PR #2312, opened 2024-07-23
Fix out-of-order dimg grid associations

y-guyon:
Fixes https://github.com/AOMediaCodec/libavif/issues/2311.

A simpler alternative would have been to refuse to decode out-of-order `dimg` associations.


y-guyon:
Note that the current implementation may not be as optimal as it could be for the following scenario:

- A part of an encoded grid image with out-of-order `dimg` associations is received at `timeA`
- libavif decodes all the in-order `numO` tiles among the `numA` available tiles
- The rest of the file is received at `timeB`
- libavif decodes all of the remaining tiles
- the image is fully decoded with a delay corresponding to `min(timeB - timeA, timeRequiredToDecodeTheNumOTiles - timeRequiredToDecodeTheNumATiles)` compared to eager decoding of every available tile at any time

This only concerns the unlikely out-of-order `dimg` associations use case and is not worth the fixing effort in my opinion.

tests/gtest/avifdimgtest.cc
```diff
@@ -0,0 +1,80 @@
...
+  ASSERT_NE(avif.size, 0u);
+  ImagePtr reference(avifImageCreateEmpty());
+  ASSERT_NE(reference, nullptr);
+  DecoderPtr decoder(avifDecoderCreate());
+  ASSERT_NE(decoder, nullptr);
+  ASSERT_EQ(avifDecoderReadMemory(decoder.get(), reference.get(), avif.data,
+                                  avif.size),
+            AVIF_RESULT_OK);
+
+  // Change the order of the 'dimg' associations.
```
vigneshvg, line 33: this bit twiddling is non-trivial. could we just add a new test file with the swapped dimg ordering?

vigneshvg (commented):
i will take a look at the rest of the PR in a while.

src/read.c
```diff
@@ -1420,17 +1420,42 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
     return AVIF_CODEC_TYPE_UNKNOWN;
 }
 
-static avifResult avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifImageGrid * grid, avifDecoderItem * gridItem, avifItemCategory itemCategory)
+// Fills the dimgIdxToItemIdx array with a mapping from each 0-based tile index in the 'dimg' reference
+// to its corresponding 0-based index in the avifMeta::items array.
+static avifResult avifFillDimgIdxToItemIdxArray(uint32_t * dimgIdxToItemIdx, avifImageGrid * grid, avifDecoderItem * gridItem)
```
wantehchang, line 1425: 1) Declare `grid` as `const avifImageGrid *`.

2) Declare `gridItem` as `const avifDecoderItem *`.

src/read.c
```diff
@@ -1420,17 +1420,42 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
...
+    const uint32_t itemIndexNotSet = UINT32_MAX;
+    for (uint32_t dimgIdx = 0; dimgIdx < numTiles; ++dimgIdx) {
+        dimgIdxToItemIdx[dimgIdx] = itemIndexNotSet;
+    }
     for (uint32_t i = 0; i < gridItem->meta->items.count; ++i) {
-        avifDecoderItem * item = gridItem->meta->items.item[i];
-        if (item->dimgForID != gridItem->id) {
-            continue;
+        if (gridItem->meta->items.item[i]->dimgForID == gridItem->id) {
+            avifDecoderItem * tileItem = gridItem->meta->items.item[i];
```
wantehchang, line 1434: Declare `tileItem` as `const avifDecoderItem *`.

Alternative we just need to declare a variable for `gridItem->meta->items.item[i]->dimgIdx`.

src/read.c
```diff
@@ -1420,17 +1420,42 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
...
+            AVIF_CHECKERR(tileItem->dimgIdx < numTiles, AVIF_RESULT_BMFF_PARSE_FAILED);
+            AVIF_CHECKERR(dimgIdxToItemIdx[tileItem->dimgIdx] == itemIndexNotSet, AVIF_RESULT_BMFF_PARSE_FAILED);
+            dimgIdxToItemIdx[tileItem->dimgIdx] = i;
```
wantehchang, line 1437: As in the original code, we can declare a `tilesAvailable` counter and increment it here. Then we can replace the for loop at line 1440 with a simple if statement. (See also my comment at line 1441.)

src/read.c
```diff
@@ -1420,17 +1420,42 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
...
         }
+    }
+    for (uint32_t dimgIdx = 0; dimgIdx < numTiles; ++dimgIdx) {
+        AVIF_CHECKERR(dimgIdxToItemIdx[dimgIdx] != itemIndexNotSet, AVIF_RESULT_BMFF_PARSE_FAILED);
```
wantehchang, line 1441: Since the number of tiles has been verified in `avifDecoderItemReadAndParse()`, this check should not fail. We can omit this check or change it to an assertion.

src/read.c
```diff
@@ -4670,25 +4688,29 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
                         // * Multiple items are claiming to be the alpha auxiliary of the current item.
                         // * Alpha auxiliary is dimg for another item.
-                        avifFree(alphaItemIndices);
+                        avifFree(dimgIdxToAlphaItemIdx);
                         *alphaItem = NULL;
                         *isAlphaItemInInput = AVIF_FALSE;
                         return AVIF_RESULT_INVALID_IMAGE_GRID;
                     }
-                    alphaItemIndices[alphaItemCount++] = j;
+                    if (item->dimgIdx >= tileCount || dimgIdxToAlphaItemIdx[item->dimgIdx] != itemIndexNotSet) {
```
wantehchang, line 4696: It seems that the `dimgIdxToAlphaItemIdx[item->dimgIdx] != itemIndexNotSet` check is redundant with the `seenAlphaForCurrentItem` check and can be omitted. Please double check.

src/read.c
```diff
@@ -4670,25 +4688,29 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+                        avifFree(dimgIdxToAlphaItemIdx);
+                        *isAlphaItemInInput = AVIF_FALSE;
```
wantehchang, line 4698: Should we add `*alphaItem = NULL;` here? See line 4669 and line 4692.

src/read.c
```diff
@@ -4670,25 +4688,29 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+                        return AVIF_RESULT_BMFF_PARSE_FAILED;
+                    }
+                    dimgIdxToAlphaItemIdx[item->dimgIdx] = j;
```
wantehchang, line 4701: As in the original code, I suggest we declare an `alphaItemCount` counter and increment it here. Then, after line 4713, we can check if `alphaItemCount != tileCount` and fail early. This allows us to delete the if statement at line 4743.

src/read.c
```diff
@@ -4710,18 +4732,24 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
         result = avifMetaFindOrCreateItem(meta, newItemID, alphaItem); // Create new empty item.
     }
     if (result != AVIF_RESULT_OK) {
-        avifFree(alphaItemIndices);
+        avifFree(dimgIdxToAlphaItemIdx);
```
wantehchang, line 4735: (This is a pre-existing issue.) Should we add `*alphaItem = NULL;` here?

`result` may be set to an error code at either line 4718 or line 4732. If we go through line 4718, then we have not assigned a value to `*alphaItem`.

src/read.c
```diff
@@ -4710,18 +4732,24 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
     memcpy((*alphaItem)->type, "grid", 4); // Make it a grid and register alpha items as its tiles.
     (*alphaItem)->width = colorItem->width;
     (*alphaItem)->height = colorItem->height;
-    for (uint32_t i = 0; i < alphaItemCount; ++i) {
-        avifDecoderItem * item = meta->items.item[alphaItemIndices[i]];
-        item->dimgForID = (*alphaItem)->id;
+    for (uint32_t dimgIdx = 0; dimgIdx < tileCount; ++dimgIdx) {
+        if (dimgIdxToAlphaItemIdx[dimgIdx] == itemIndexNotSet) {
+            avifFree(dimgIdxToAlphaItemIdx);
+            *isAlphaItemInInput = AVIF_FALSE;
```
wantehchang, line 4745: It is not cleae whether we should add `*alphaItem = NULL;` here. We can avoid this issue by performing this check early as I suggested in the comment at line 4701.

src/read.c
```diff
@@ -4710,18 +4732,24 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+            return AVIF_RESULT_BMFF_PARSE_FAILED;
+        }
+        avifDecoderItem * alphaTileItem = meta->items.item[dimgIdxToAlphaItemIdx[dimgIdx]];
+        alphaTileItem->dimgForID = (*alphaItem)->id;
+        alphaTileItem->dimgIdx = dimgIdx;
```
wantehchang, line 4750: This line is not in the original code. Is this a bug fix?

src/read.c
```diff
@@ -5017,7 +5045,15 @@ static avifResult avifDecoderGenerateImageTiles(avifDecoder * decoder, avifTileI
 {
     const uint32_t previousTileCount = decoder->data->tiles.count;
     if ((info->grid.rows > 0) && (info->grid.columns > 0)) {
-        AVIF_CHECKRES(avifDecoderGenerateImageGridTiles(decoder, &info->grid, item, itemCategory));
+        // The number of tiles was verified in avifDecoderItemReadAndParse().
+        uint32_t * dimgIdxToItemIdx = (uint32_t *)avifAlloc((size_t)info->grid.rows * info->grid.columns * sizeof(uint32_t));
```
wantehchang, line 5049: Nit: The `size_t` cast could be omitted. It is not used in line 4674.

Since both `info->grid.rows` and `info->grid.columns` are <= 256, the result of this expression fits in `uint32_t`.

tests/CMakeLists.txt
```diff
@@ -123,12 +124,7 @@ if(AVIF_ENABLE_GTEST)
     add_avif_gtest(avifgridapitest)
     add_avif_gtest_with_data(avifilocextenttest)
     add_avif_gtest(avifimagetest)
-
-    add_executable(avifincrtest gtest/avifincrtest.cc)
-    target_link_libraries(avifincrtest aviftest_helpers avifincrtest_helpers)
-    add_test(NAME avifincrtest COMMAND avifincrtest ${CMAKE_CURRENT_SOURCE_DIR}/data/)
-    register_test_for_coverage(avifincrtest ${CMAKE_CURRENT_SOURCE_DIR}/data/)
-
+    add_avif_gtest_with_data(avifincrtest avifincrtest_helpers)
```
wantehchang, line 127: Optional: This cleanup change is independent of the bug fix and can be moved to a separate pull request.

src/read.c
```diff
@@ -1420,17 +1420,42 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
     return AVIF_CODEC_TYPE_UNKNOWN;
 }
 
-static avifResult avifDecoderGenerateImageGridTiles(avifDecoder * decoder, avifImageGrid * grid, avifDecoderItem * gridItem, avifItemCategory itemCategory)
+// Fills the dimgIdxToItemIdx array with a mapping from each 0-based tile index in the 'dimg' reference
+// to its corresponding 0-based index in the avifMeta::items array.
```
wantehchang, line 1424: IMPORTANT: Please check this question carefully because I am not familiar with this.

I see `avifDecoderItem::dimgIdx` documented as follows:

```
    // If dimgForId is non-zero, this is the zero-based index of this item in the list of Item #{dimgForID}'s dimg
    // Note that if the same item appears multiple times in the dimg box, this is the last index for this item.
    uint32_t dimgIdx;
```

Please verify that our code can handle the case where "the same item appears multiple times in the dimg box".

If I understand it correctly, in this case the `dimgIdxToItemIdx` array will have some entries left unset (equal to the special value `itemIndexNotSet`).

I wonder if `avifParseItemReferenceBox()` should store more information so that we can handle this case.

wantehchang (commented):
Yannis: Here is my first round of review. I saved some style/preference comments for the next round. I did not review the new avifdimgtest.cc file, but I reviewed the changes to read.c carefully.

Please check my first comment in read.c (marked with "IMPORTANT") carefully. If I am right, this pull request needs more work.

src/read.c
y-guyon: It seems there is no impact of doing so in case of error.

src/read.c
y-guyon: It seems there is no impact of doing so in case of error.

src/read.c
y-guyon: No. Previously `dimgIdx` was not used for tile ordering and it was unnecessary to set it to this made-up value.

tests/CMakeLists.txt
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2317

src/read.c
y-guyon: See https://github.com/AOMediaCodec/libavif/pull/2315

src/read.c
y-guyon: It is not the same check: `dimg` vs `auxl`

src/read.c
wantehchang: You are right. This function usually sets `*alphaItem` to `NULL` and `*isAlphaItemInInput` to `AVIF_FALSE` when it returns an error, but it isn't consistent. The callers of this function ignore the output parameters when the function fails. So this function can actually follow the common policy if not touching the output parameters when it fails. We can write a CL to do that later.

For now, let's follow the current policy of setting the output parameters to `NULL` and `AVIF_FALSE` before returning an error.

src/read.c
```diff
@@ -1420,17 +1420,43 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
...
+static avifResult avifFillDimgIdxToItemIdxArray(uint32_t * dimgIdxToItemIdx, const avifImageGrid * grid, const avifDecoderItem * gridItem)
```
wantehchang, line 1425: Optional: The `grid` parameter is only used in the expression `grid->rows * grid->columns`. So we can replace the `grid` parameter with a `uint32_t numExpectedTiles` parameter, which is the size of the `dimgIdxToItemIdx` array.

src/read.c
```diff
@@ -1420,17 +1420,43 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
...
+        dimgIdxToItemIdx[dimgIdx] = itemIndexNotSet;
+    }
+    uint32_t numTiles = 0;
     for (uint32_t i = 0; i < gridItem->meta->items.count; ++i) {
-        avifDecoderItem * item = gridItem->meta->items.item[i];
-        if (item->dimgForID != gridItem->id) {
-            continue;
+        if (gridItem->meta->items.item[i]->dimgForID == gridItem->id) {
+            const uint32_t tileItemDimgIdx = gridItem->meta->items.item[i]->dimgIdx;
+            AVIF_CHECKERR(tileItemDimgIdx < numExpectedTiles, AVIF_RESULT_BMFF_PARSE_FAILED);
```
wantehchang, line 1436: Optional: We can use the more specific `AVIF_RESULT_INVALID_IMAGE_GRID` error code in this function.

src/read.c
```diff
@@ -1420,17 +1420,43 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
...
         }
+    }
+    // The number of tiles has been verified in avifDecoderItemReadAndParse().
+    AVIF_ASSERT_OR_RETURN(numTiles == numExpectedTiles);
+    return AVIF_RESULT_OK;
+}
+
+// Creates the tiles and associate them to the items in the order of the 'dimg' association.
+static avifResult avifDecoderGenerateImageGridTiles(avifDecoder * decoder,
+                                                    avifImageGrid * grid,
```
wantehchang, line 1449: Declare `grid` as `const avifImageGrid *`.

Optional: The `grid` parameter is only used in the expression `grid->rows * grid->columns`. So we can replace the `grid` parameter with a `uint32_t numTiles` parameter, listed last, after the `dimgIdxToItemIdx` parameter.

src/read.c
```diff
@@ -4666,29 +4686,33 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
-                avifFree(alphaItemIndices);
+                avifFree(dimgIdxToAlphaItemIdx);
                 *alphaItem = NULL;
                 *isAlphaItemInInput = AVIF_FALSE;
                 return AVIF_RESULT_OK;
             }
         }
     }
-    AVIF_ASSERT_OR_RETURN(alphaItemCount == colorItemCount);
+    AVIF_ASSERT_OR_RETURN(alphaItemCount == tileCount);
```
wantehchang, line 4715: IMPORTANT: This should not be an assertion. This condition can be false if the input file is invalid.

This should read:

```
    if (alphaItemCount != tileCount) {
        avifFree(dimgIdxToAlphaItemIdx);
        *alphaItem = NULL;
        return AVIF_RESULT_INVALID_IMAGE_GRID;  // or AVIF_RESULT_BMFF_PARSE_FAILED
    }
```

src/read.c
```diff
@@ -4710,18 +4734,20 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
         return result;
     }
     memcpy((*alphaItem)->type, "grid", 4); // Make it a grid and register alpha items as its tiles.
     (*alphaItem)->width = colorItem->width;
     (*alphaItem)->height = colorItem->height;
-    for (uint32_t i = 0; i < alphaItemCount; ++i) {
-        avifDecoderItem * item = meta->items.item[alphaItemIndices[i]];
-        item->dimgForID = (*alphaItem)->id;
+    for (uint32_t dimgIdx = 0; dimgIdx < tileCount; ++dimgIdx) {
+        AVIF_ASSERT_OR_RETURN(dimgIdxToAlphaItemIdx[dimgIdx] != itemIndexNotSet);
```
wantehchang, line 4745: IMPORTANT: We cannot use the `AVIF_ASSERT_OR_RETURN` macro here, because it doesn't free `dimgIdxToAlphaItemIdx` and set the output parameters to `NULL` and `AVIF_FALSE`. We need to use an if statement. Alternatively, use `assert()`.

wantehchang (approved):
LGTM. Thanks. Please evaluate the optional suggestions and only accept the ones you agree with. There are two important changes that should be made.

src/read.c
y-guyon: `dimgIdxToAlphaItemIdx[dimgIdx] != itemIndexNotSet` is guaranteed by `alphaItemCount != tileCount` above. I agree an `assert()` would be enough here, but the point of `AVIF_ASSERT_OR_RETURN()` was to replace as many `assert()` as possible, to be safe. I believe a leak is preferable to an out-of-bounds array access.

wantehchang (approved):

src/read.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2330.

PR #2315, opened 2024-07-24
Detect iref repetition before ItemReadAndParse()

y-guyon:
Related to https://github.com/AOMediaCodec/libavif/issues/2311#issuecomment-2247206691.


src/read.c
```diff
@@ -2900,6 +2896,12 @@ static avifResult avifParseItemReferenceBox(avifMeta * meta, const uint8_t * raw
                 avifDecoderItem * dimg;
                 AVIF_CHECKRES(avifMetaFindOrCreateItem(meta, toID, &dimg));
 
+                // Section 8.11.12.1 of ISO/IEC 14496-12:
+                //   The items linked to are then represented by an array of to_item_IDs;
+                //   within a given array, a given value shall occur at most once.
+                AVIF_CHECKERR(dimg->dimgForID != fromID, AVIF_RESULT_INVALID_IMAGE_GRID);
+                // A given value may occur more than once within different arrays but this is not supported by libavif.
```
wantehchang, line 2903: Nit: Change "more than once within different arrays" to "within multiple arrays"?

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Fix CMake config shared library leaks
   https://github.com/AOMediaCodec/libavif/issues/2264.
 * Fix clang-cl compilation.
+* Report files with an item used in multiple 'dimg' arrays with
```
wantehchang, line 24: Nit/optional: Would "boxes" be more specific than "arrays"?

wantehchang (approved):
LGTM. Thanks. (I assume this pull request is ready for review.)

I didn't check the new test. It would be good to wait for Vignesh's review.

vigneshvg (approved):

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Fix CMake config shared library leaks
   https://github.com/AOMediaCodec/libavif/issues/2264.
 * Fix clang-cl compilation.
+* Report files with an item used in multiple 'dimg' boxes with
+  AVIF_RESULT_NOT_IMPLEMENTED instead of AVIF_RESULT_INVALID_IMAGE_GRID.
```
wantehchang, line 25: Should we delete "instead of AVIF_RESULT_INVALID_IMAGE_GRID"? It implies that we were detecting this condition before, but the checks added in this pull request are new.

Should we also document that we now reject images that are not compliant with "within a given array, a given value shall occur at most once"?

wantehchang (approved):

CHANGELOG.md
y-guyon: > Should we delete "instead of AVIF_RESULT_INVALID_IMAGE_GRID"? It implies that we were detecting this condition before, but the checks added in this pull request are new.

Such files were failing with `AVIF_RESULT_INVALID_IMAGE_GRID` before this PR, so this sentence looks accurate to me.

> Should we also document that we now reject images that are not compliant with "within a given array, a given value shall occur at most once"?

These images were already rejected with `AVIF_RESULT_INVALID_IMAGE_GRID`. So there is no change in behavior.  
I did not try to create such a file but maybe one could have created a grid with twice the same tile and another extra unused tile, so that the check `dimgItemCount == grid->rows * grid->columns` would still pass. In this very specific case, maybe a CHANGELOG entry is useful, but this is really niche.

wantehchang: Did you mean these images are rejected by the following check in `avifDecoderItemReadAndParse()`?
```
            // Validate that there are exactly the same number of dimg items to form the grid.
            uint32_t dimgItemCount = 0;
            for (uint32_t i = 0; i < item->meta->items.count; ++i) {
                if (item->meta->items.item[i]->dimgForID == item->id) {
                    ++dimgItemCount;
                }
            }
            if (dimgItemCount != grid->rows * grid->columns) {
            }
```

So the new checks added in this pull request reject these images earlier (before avifDecoderItemReadAndParse()), and in one case with a different error code?

Thanks for the clarification.

PR #2318, opened 2024-07-24
Compile avif_apps_internal without -DAVIF_DLL

wantehchang:
In shared library builds (-DBUILD_SHARED_LIBS=ON), compile the source
files in the avif_apps_internal static library without -DAVIF_DLL.

Also link avif_obj with the m and Threads::Threads libraries in the
PRIVATE scope instead of the PUBLIC scope.

Fix https://github.com/AOMediaCodec/libavif/issues/2339.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -455,7 +455,7 @@ if(UNIX OR MINGW)
     # Find out if we have threading available
     set(THREADS_PREFER_PTHREAD_FLAG ON)
     find_package(Threads)
-    target_link_libraries(avif_obj PUBLIC m Threads::Threads)
+    target_link_libraries(avif_obj PRIVATE m Threads::Threads)
 endif()
 
 if(NOT AVIF_LIBYUV_ENABLED)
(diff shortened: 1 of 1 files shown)


wantehchang:
@vrabaud FYI.

CMakeLists.txt
```diff
@@ -676,45 +676,43 @@ if(AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE
...
     endif()
 
+    set(AVIF_APPS_SRCS apps/shared/avifexif.c apps/shared/avifjpeg.c apps/shared/avifpng.c apps/shared/avifutil.c
+                       apps/shared/iccmaker.c apps/shared/y4m.c third_party/iccjpeg/iccjpeg.c
+    )
+
+    macro(add_avif_apps_library suffix)
+        add_library(avif_apps${suffix} STATIC ${AVIF_APPS_SRCS})
+        target_link_libraries(avif_apps${suffix} PUBLIC avif${suffix} PRIVATE PNG::PNG ZLIB::ZLIB JPEG::JPEG avif_enable_warnings)
+        if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
jzern, line 695: Is this generic enough? I'm wondering if BSD variants would need this too.

jzern (approved):

CMakeLists.txt
jzern: Maybe the find_library method might be the safest.

PR #2304, opened 2024-07-22
fix: link against transitive libaom library dependencies

fdintino:
fixes #2274

PR #2324, opened 2024-07-25
Reapply fdintino's fix for libaom shared lib case

wantehchang:
Revert the change to cmake/Modules/Findaom.cmake in https://github.com/AOMediaCodec/libavif/pull/2304/commits/a125ed83ac97458404daed3e5285afe368b3c037


wantehchang:
@fdintino Frankie: Could you create your own pull request using this pull request? I verified that this fixes both the libaom static library and shared library cases. Thanks, ans sorry about the inconvenience.

wantehchang:
I found that the change in this pull request is unnecessary if libvmaf.so is on the run-time linker's search path, e.g, by installing libvmaf.so in `/usr/lib` or `/lib` or by setting the `LD_LIBRARY_PATH` environment variable properly. So I am going to consider it a libvmaf installation error if libvmaf.so is NOT on the run-time linker's search path. I will abandon this pull request.

PR #2328, opened 2024-07-26
Copy avif.dll to tests binary directory on Windows

wantehchang:
This is a simple way to allow are_images_equal.exe and the test programs find avif.dll. Simpler than adding the directory where avif.dll resides to the Path environment variable.


wantehchang:
I have tested this change in https://github.com/AOMediaCodec/libavif/pull/2309.

jzern (approved):

Issue #2322, opened 2024-07-25
MSVC: avifsampletransformtest.cc(153): warning C4244: 'argument': conversion from 'const avifSampleTransformTokenType' to 'uint8_t', possible loss of data


wantehchang:
Please see https://github.com/AOMediaCodec/libavif/actions/runs/10099659232:

```
[74/162] Building CXX object tests\CMakeFiles\avifsampletransformtest.dir\gtest\avifsampletransformtest.cc.obj
FAILED: tests/CMakeFiles/avifsampletransformtest.dir/gtest/avifsampletransformtest.cc.obj 
C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1440~1.338\bin\Hostx64\x64\cl.exe  /nologo /TP -DAVIF_ENABLE_EXPERIMENTAL_GAIN_MAP -DAVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION -DAVIF_ENABLE_EXPERIMENTAL_METAV1 -DAVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM -DAVIF_ENABLE_EXPERIMENTAL_YCGCO_R -D_CRT_NONSTDC_NO_WARNINGS -D_CRT_SECURE_NO_WARNINGS -ID:\a\libavif\libavif\apps\shared -ID:\a\libavif\libavif\include -external:ID:\a\libavif\libavif\build\_deps\googletest-src\googletest\include -external:ID:\a\libavif\libavif\build\_deps\googletest-src\googletest -external:W0 /DWIN32 /D_WINDOWS /GR /EHsc /MD /O2 /Ob2 /DNDEBUG -std:c++14 /W4 /wd4324 /WX /showIncludes /Fotests\CMakeFiles\avifsampletransformtest.dir\gtest\avifsampletransformtest.cc.obj /Fdtests\CMakeFiles\avifsampletransformtest.dir\ /FS -c D:\a\libavif\libavif\tests\gtest\avifsampletransformtest.cc
D:\a\libavif\libavif\tests\gtest\avifsampletransformtest.cc(153): error C2220: the following warning is treated as an error
D:\a\libavif\libavif\tests\gtest\avifsampletransformtest.cc(153): warning C4244: 'argument': conversion from 'const avifSampleTransformTokenType' to 'uint8_t', possible loss of data
[...]

PR #2329, opened 2024-07-26
Fix AddOperator() in sato test

y-guyon:
Fixes https://github.com/AOMediaCodec/libavif/issues/2322.

### DIFF
--- a/tests/gtest/avifsampletransformtest.cc
+++ b/tests/gtest/avifsampletransformtest.cc
@@ -28,7 +28,7 @@ class AvifExpression : public avifSampleTransformExpression {
     token.type = AVIF_SAMPLE_TRANSFORM_INPUT_IMAGE_ITEM_INDEX;
     token.inputImageItemIndex = inputImageItemIndex;
   }
-  void AddOperator(uint8_t op) {
+  void AddOperator(avifSampleTransformTokenType op) {
     avifSampleTransformToken& token = AddToken();
     token.type = op;
   }


vrabaud (approved):

tests/gtest/avifsampletransformtest.cc
```diff
@@ -28,7 +28,7 @@ class AvifExpression : public avifSampleTransformExpression {
     token.type = AVIF_SAMPLE_TRANSFORM_INPUT_IMAGE_ITEM_INDEX;
     token.inputImageItemIndex = inputImageItemIndex;
   }
-  void AddOperator(uint8_t op) {
+  void AddOperator(avifSampleTransformTokenType op) {
```
wantehchang, line 31: (This is an unrelated issue.) I noticed that the local variable `token` declared at line 60 shadows the data member `token` of the base class `avifSampleTransformExpression`. It would be good to rename the local variable.

wantehchang, line 31: Thank you for the pull request. I am worried that this did not fix the compiler warning. (Compiler warnings are not enabled on tests yet. Sorry about the confusion.) I will test this using a draft pull request of mine later (https://github.com/AOMediaCodec/libavif/pull/2318).

The warning we need to fix is:

```
warning C4244: 'argument': conversion from 'const avifSampleTransformTokenType' to 'uint8_t', possible loss of data
```

Since `token.type` is a uint8_t, the assignment `token.type = op;` is now a conversion from 'const avifSampleTransformTokenType' to 'uint8_t', so I am worried that MSVC will warn about the assignment.

y-guyon: The data member is named `tokens` in `avifSampleTransformExpression`:

https://github.com/AOMediaCodec/libavif/blob/40b5589b003ca11a63572a470ac8744404fec2f9/include/avif/internal.h#L204

So there is no shadowing.

y-guyon: Sorry, it was an oversight on my end. Fixed in https://github.com/AOMediaCodec/libavif/pull/2331.

wantehchang: You are right. I made a mistake when using Code Search, but I don't know what I did wrong :-). Sorry!

PR #2332, opened 2024-07-26
read.c: Declare alphaCategory as avifItemCategory

wantehchang:
This way *category and alphaCategory have the same type, and code that uses them looks consistent.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -5377,7 +5377,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                               AVIF_RESULT_NOT_IMPLEMENTED);
                 avifItemCategory * category = &data->sampleTransformInputImageItems[data->sampleTransformNumInputImageItems];
                 avifBool foundItem = AVIF_FALSE;
-                uint32_t alphaCategory = AVIF_ITEM_CATEGORY_COUNT;
+                avifItemCategory alphaCategory = AVIF_ITEM_CATEGORY_COUNT;
                 for (int c = AVIF_ITEM_COLOR; c < AVIF_ITEM_CATEGORY_COUNT; ++c) {
                     if (mainItems[c] && inputImageItem->id == mainItems[c]->id) {
                         *category = c;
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -5377,7 +5377,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                               AVIF_RESULT_NOT_IMPLEMENTED);
                 avifItemCategory * category = &data->sampleTransformInputImageItems[data->sampleTransformNumInputImageItems];
                 avifBool foundItem = AVIF_FALSE;
-                uint32_t alphaCategory = AVIF_ITEM_CATEGORY_COUNT;
+                avifItemCategory alphaCategory = AVIF_ITEM_CATEGORY_COUNT;
```
wantehchang, line 5380: (This question is unrelated to this CL.) I wonder why clang-cl (with `-Wall`, which is treated as Clang's `-Weverything`) did not warn about line 5383 but warned about line 5393 (before the cast was added).

vrabaud: What was the warning ?

vrabaud (approved):

src/read.c
wantehchang: I don't know. Although I reviewed your pull request https://github.com/AOMediaCodec/libavif/pull/2275, I didn't see the original warnings.

When I added clang-cl support, I only tested `/W4` and did not test `-Wall -Wextra`.

In any case, this question isn't that important. I was just curious.

wantehchang: Microsoft documents that in MSVC "a variable declared as **`enum`** is an **`int`**":
https://learn.microsoft.com/en-us/cpp/c-language/enum-type?view=msvc-170

Since clang-cl emulates MSVC, this may be why clang-cl  apparently uses a signed type as the underlying type for the `avifItemCategory` enum type. Regular Clang may use an unsigned type instead because it emulates GCC, which says "Normally, the type is `unsigned int` if there are no negative values in the enumeration, otherwise `int`."
https://gcc.gnu.org/onlinedocs/gcc/Structures-unions-enumerations-and-bit-fields-implementation.html

y-guyon (approved):

Issue #2340, opened 2024-07-29
In libavif v1.1.0, only the source files in the avif_obj and avif_apps_obj OBJECT libraries are compiled with warnings enabled


wantehchang:
In libavif v1.1.0, only the source files in the avif_obj and avif_apps_obj OBJECT libraries are compiled with warnings enabled.

The source files in the examples, apps, and tests are NOT compiled with warnings enabled.

wantehchang:
I verified with a local build on Linux that this bug has been fixed [except the fuzztest targets](https://github.com/AOMediaCodec/libavif/issues/2350) in tests/CMakeLists.txt.

I agree that the new method is tedious and it is easy to miss when we add a new library or executable target. But no solution for this issue is perfect. I probably would prefer `add_compile_options()` if we could prevent it from affecting the external dependencies that we incoporate using `add_subdirectory()`.

We can adjust the new method, such as changing the `avif_enable_warnings` interface library to a function or macro that we call on each target, or adding back the `$<BUILD_INTERFACE>` generator expressions.

PR #2344, opened 2024-07-30
Enable warnings in tests

wantehchang:
Note that the fuzztest targets are not compiled with warnings enabled because the FuzzTest and Abseil headers have compiler warnings.

Part 2 of the fix for
https://github.com/AOMediaCodec/libavif/issues/2340.


wantehchang:
@vrabaud FYI.

jzern (approved):

PR #2351, opened 2024-07-31
Do not define AVIF_ENABLE_NODISCARD for avif.h

wantehchang:
```
Make the AVIF_ENABLE_NODISCARD cmake option require C23 so that avif.h
can define the AVIF_NODISCARD macro as [[nodiscard]] without checking
the AVIF_ENABLE_NODISCARD macro. This requires changing avif.h to test
__STDC_VERSION__ >= 202000L, which is the value of __STDC_VERSION__ in
recent versions of GCC and Clang when -std=c2x or -std=gnu2x is
specified.

In general avif.h should only test compiler predefined macros. The only
exception is AVIF_DLL (ignoring the experimental feature macros.)

Fix https://github.com/AOMediaCodec/libavif/issues/2352.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,8 +33,16 @@ include(FetchContent)
 include(FindPkgConfig)
 include(AvifExternalProjectUtils)
 
+option(AVIF_ENABLE_NODISCARD "Add [[nodiscard]] to some functions. CMake must be at least 3.21 to request C23." OFF)
+
 # Set C99 as the default
-set(CMAKE_C_STANDARD 99)
+if(AVIF_ENABLE_NODISCARD AND CMAKE_VERSION VERSION_GREATER_EQUAL 3.21.0)
+    # [[nodiscard]] requires C23. The supported value 23 for CMAKE_C_STANDARD
+    # was added in CMake version 3.21.
+    set(CMAKE_C_STANDARD 23)
+else()
+    set(CMAKE_C_STANDARD 99)
+endif()
 
 # SOVERSION scheme: MAJOR.MINOR.PATCH
 #   If there was an incompatible interface change:
@@ -52,7 +60,6 @@ set(LIBRARY_SOVERSION ${LIBRARY_VERSION_MAJOR})
 option(BUILD_SHARED_LIBS "Build shared avif library" ON)
 
 option(AVIF_ENABLE_WERROR "Treat all compiler warnings as errors" OFF)
-option(AVIF_ENABLE_NODISCARD "Add [[nodiscard]] to some functions. CMake must be at least 3.21 to force C23" OFF)
 
 option(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R "Enable experimental YCgCo-R matrix code" OFF)
 option(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -33,8 +33,16 @@ include(FetchContent)
 include(FindPkgConfig)
 include(AvifExternalProjectUtils)
 
+option(AVIF_ENABLE_NODISCARD "Add [[nodiscard]] to some functions. CMake must be at least 3.21 to request C23." OFF)
```
wantehchang, line 36: This is copied from line 55 in the original code, with "force C23" changed to "request C23."

CMakeLists.txt
```diff
@@ -333,22 +340,6 @@ endif()
 
 target_link_libraries(avif_obj PRIVATE avif_enable_warnings)
 
-if(AVIF_ENABLE_NODISCARD)
-    if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.21.0)
-        set(CMAKE_C_STANDARD 23)
-        set_property(TARGET avif_obj PROPERTY C_STANDARD 23)
-    else()
```
wantehchang, line 340: If we redefine `AVIF_ENABLE_NODISCARD` to be a request rather than a requirement, we don't need this `else` block.

CMakeLists.txt
```diff
@@ -333,22 +340,6 @@ endif()
...
-        unset(CMAKE_C_STANDARD)
-        set_property(TARGET avif_obj PROPERTY C_STANDARD)
-        if(CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID MATCHES "GNU")
-            target_compile_options(avif_obj PUBLIC $<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:C>:-std=gnu2x>>)
-        elseif(CMAKE_C_COMPILER_ID MATCHES "MSVC")
-            target_compile_options(avif_obj PUBLIC $<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:C>:/std:clatest>>)
-        endif()
-    endif()
-    target_compile_definitions(avif_obj PUBLIC $<BUILD_INTERFACE:AVIF_ENABLE_NODISCARD=1>)
```
wantehchang, line 349: The definition of the `AVIF_ENABLE_NODISCARD` macro for avif.h is also deleted.

wantehchang:
@vrabaud FYI.

jzern (approved):
This makes sense.

The decaying is unfortunate:

https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#id8

But it looks like [`CMAKE_C_STANDARD_REQUIRED`](https://cmake.org/cmake/help/latest/variable/CMAKE_C_STANDARD_REQUIRED.html) might work if we wanted to abort when the option was specified, but the version not available.

wantehchang:
> But it looks like [`CMAKE_C_STANDARD_REQUIRED`](https://cmake.org/cmake/help/latest/variable/CMAKE_C_STANDARD_REQUIRED.html) might work if we wanted to abort when the option was specified, but the version not available.

Thank you for the suggestion. I implemented this in the second commit. I also rewrote the commit message and pasted it at the top of this issue. Please take a look. Thanks!

CMakeLists.txt
```diff
@@ -33,8 +33,19 @@ include(FetchContent)
...
-set(CMAKE_C_STANDARD 99)
+if(AVIF_ENABLE_NODISCARD)
+    # [[nodiscard]] requires C23.
+    if(CMAKE_VERSION VERSION_LESS 3.21.0)
+        message(FATAL_ERROR "CMake must be at least 3.21 to force C23, bailing out")
+    endif()
+    set(CMAKE_C_STANDARD 23)
+    set(CMAKE_C_STANDARD_REQUIRED ON)
+else()
+    set(CMAKE_C_STANDARD 99)
```
jzern, line 47: I would think this would be required too, but it's highly unlikely anyone would try this with a c90 only compiler (or force those flags)

jzern (approved):

CMakeLists.txt
wantehchang: Yes, we assume the compiler supports C99.

PR #2354, opened 2024-08-01
Fix overflows in avifRGBImageAllocatePixels()

wantehchang:
Calculate rowBytes in uint32_t. Only the allocation size needs to be size_t.

Also make sure it is safe to cast various rowBytes fields to ptrdiff_t. We need to do this when subtracting rowBytes from a pointer to go back one row.

Part of the fix to https://github.com/AOMediaCodec/libavif/issues/2271.

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -352,15 +352,21 @@ avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags planes)
     if (image->width == 0 || image->height == 0) {
         return AVIF_RESULT_INVALID_ARGUMENT;
     }
-    const size_t channelSize = avifImageUsesU16(image) ? 2 : 1;
-    if (image->width > SIZE_MAX / channelSize) {
+    const uint32_t channelSize = avifImageUsesU16(image) ? 2 : 1;
+    if (image->width > UINT32_MAX / channelSize) {
         return AVIF_RESULT_INVALID_ARGUMENT;
     }
-    const size_t fullRowBytes = channelSize * image->width;
-    if ((fullRowBytes > UINT32_MAX) || (image->height > SIZE_MAX / fullRowBytes)) {
+    const uint32_t fullRowBytes = channelSize * image->width;
+#if UINT32_MAX > PTRDIFF_MAX
+    // Make sure it is safe to cast image->yuvRowBytes[i] or image->alphaRowBytes to ptrdiff_t.
+    if (fullRowBytes > PTRDIFF_MAX) {
         return AVIF_RESULT_INVALID_ARGUMENT;
     }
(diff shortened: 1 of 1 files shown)


maryla-uc (approved):

src/avif.c
```diff
@@ -352,27 +352,33 @@ avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags planes)
...
-    const size_t fullRowBytes = channelSize * image->width;
-    if ((fullRowBytes > UINT32_MAX) || (image->height > SIZE_MAX / fullRowBytes)) {
+    const uint32_t fullRowBytes = channelSize * image->width;
+#if UINT32_MAX > PTRDIFF_MAX
+    // Make sure it is safe to cast image->yuvRowBytes[i] or image->alphaRowBytes to ptrdiff_t.
+    if (fullRowBytes > PTRDIFF_MAX) {
         return AVIF_RESULT_INVALID_ARGUMENT;
     }
-    const size_t fullSize = fullRowBytes * image->height;
+#endif
```
wantehchang, line 365: @jzern I added this check yesterday (because I will need to cast `rowBytes` (what libaom calls `stride`) to `ptrdiff_t` when I fix some overflows in src/reformat.c. But I am now wondering if this check is necessary. If we have allocated the plane buffer successfully, it implies that `rowBytes` is <= `SIZE_MAX`. So in a reasonable C implentation, this should also imply that `rowBytes` is <= `PTRDIFF_MAX`. Do you agree?

cppreference.com describes this [unlikely case](https://en.cppreference.com/w/c/types/ptrdiff_t):

> If an array is so large (greater than [`PTRDIFF_MAX`](https://en.cppreference.com/w/c/types/limits) elements, but equal to or less than [`SIZE_MAX`](https://en.cppreference.com/w/c/types/limits) bytes), that the difference between two pointers may not be representable as `ptrdiff_t`, the result of subtracting two such pointers is undefined.

wantehchang: I have another idea: we can avoid this unlikely case by changing the check below to use `PTRDIFF_MAX`:

```
    if (image->height > PTRDIFF_MAX / fullRowBytes) {
    }
```

This seems to match the [following paragraph in cppreference.com](https://en.cppreference.com/w/c/types/ptrdiff_t):

> For char arrays shorter than [`PTRDIFF_MAX`](https://en.cppreference.com/w/c/types/limits), `ptrdiff_t` acts as the signed counterpart of [`size_t`](https://en.cppreference.com/w/c/types/size_t): it can store the size of the array of any type and is, on most platforms, synonymous with `intptr_t`).

jzern: Using `PTRDIFF_MAX` sounds reasonable.

Issue #2350, opened 2024-07-31
Consider enabling compiler warnings in add_avif_fuzztest() in tests/CMakeLists.txt


wantehchang:
@jzern @vrabaud @y-guyon 

In https://github.com/AOMediaCodec/libavif/pull/2344 I enabled compiler warnings in the tests/ directory, but I had to exclude the `add_avif_fuzztest()` macro because there are compiler warnings in the FuzzTest and Abseil headers.

It would be good to look into whether it is possible to enable compiler warnings in `add_avif_fuzztest()`, e.g., by turning off the warnings in the FuzzTest and Abseil headers.

I am surprised that Abseil headers have compiler warnings when compiled with `-Wall -Wextra`. The [Abseil Compiler Flags](https://abseil.io/docs/cpp/platforms/compilerflags#abseil-warning-flags) page says they compile with `-Wall -Wextra` under GCC but turn off `-Wsign-compare` (which is one of the warnings I saw). So we can first suppress the compiler warnings in Abseil headers under the guidance of the Abseil Compiler Flags page, and then see what we need to do with the compiler warnings in FuzzTest headers.

vrabaud:
In CMake 3.25 and above, we can add the SYSTEM property to the add_subdirectory of the fuzztest folder. fuzztest only recommends to use an add_subdirectory but we could try to FetchContent it as imported targets are SYSTEM I believe.

wantehchang:
We ignore the compiler warnings when FuzzTest itself is compiled. This issue is about the compiler warnings when our fuzztest source files are compiled (because they include FuzzTest headers).

wantehchang:
Vincent: I misunderstood your comment. Sorry. I think adding the SYSTEM property to the add_subdirectory() call of the fuzztest folder is a good solution. Thanks for the suggestion. I am testing it in https://github.com/AOMediaCodec/libavif/pull/2356.

PR #2356, opened 2024-08-01
Enable compiler warnings in fuzztest targets

wantehchang:
Pass the SYSTEM argument to add_subdirectory(ext/fuzztest) to treat 
FuzzTest headers as system headers. This suppresses compiler warnings in
FuzzTest headers.

Fix https://github.com/AOMediaCodec/libavif/issues/2350.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -198,10 +198,12 @@ if(AVIF_ENABLE_FUZZTEST)
         #   FUZZTEST_REPLAY=/path/to/repro_file.test
         # and running one of the targets below.
         # See https://github.com/google/fuzztest/blob/main/doc/quickstart-cmake.md
-        # Note: The SYSTEM argument for add_subdirectory() is new in CMake
-        # version 3.25. There are compiler warnings in FuzzTest headers. Add
-        # the ext/fuzztest subdirectory with the SYSTEM directory property so
-        # that warnings in its headers are suppressed.
+        # Note: There are compiler warnings in the FuzzTest headers. Add the
+        # ext/fuzztest subdirectory with the SYSTEM directory property set to
+        # true so that warnings in its headers are suppressed.
+        if(CMAKE_VERSION VERSION_LESS 3.25.0)
+            message(FATAL_ERROR "CMake must be at least 3.25 to pass the SYSTEM argument to add_subdirectory(), bailing out")
+        endif()
         add_subdirectory(${AVIF_SOURCE_DIR}/ext/fuzztest ${AVIF_SOURCE_DIR}/ext/fuzztest/build.libavif EXCLUDE_FROM_ALL SYSTEM)
     else()
         message(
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -201,7 +198,11 @@ if(AVIF_ENABLE_FUZZTEST)
         #   FUZZTEST_REPLAY=/path/to/repro_file.test
         # and running one of the targets below.
         # See https://github.com/google/fuzztest/blob/main/doc/quickstart-cmake.md
-        add_subdirectory(${AVIF_SOURCE_DIR}/ext/fuzztest ${AVIF_SOURCE_DIR}/ext/fuzztest/build.libavif EXCLUDE_FROM_ALL)
+        # Note: The SYSTEM argument for add_subdirectory() is new in CMake
+        # version 3.25. There are compiler warnings in FuzzTest headers. Add
+        # the ext/fuzztest subdirectory with the SYSTEM directory property so
+        # that warnings in its headers are suppressed.
+        add_subdirectory(${AVIF_SOURCE_DIR}/ext/fuzztest ${AVIF_SOURCE_DIR}/ext/fuzztest/build.libavif EXCLUDE_FROM_ALL SYSTEM)
```
wantehchang, line 205: Note that the use of `SYSTEM` is not guarded by a CMake version check here. This passed our CI workflows. I guess CMake only parses the conditional branches based on the conditions.

wantehchang:
@vrabaud FYI. This uses the `SYSTEM` directory property that you suggested.

tests/CMakeLists.txt
jzern: I was wondering if you wanted to add `cmake_minimum_required(VERSION 3.25)` or check `CMAKE_VERSION` when this was enabled.

wantehchang: Do you recommend I add a cmake version check even though this passed our CI workflows?

jzern: I don't see any workflows that set `AVIF_LOCAL_FUZZTEST`. If an older version of cmake is used and that set, I'd expect at least a warning, but I haven't tested this. Let me check.

jzern: With 3.21:

```
CMake Error at tests/CMakeLists.txt:205 (add_subdirectory):
  add_subdirectory called with incorrect number of arguments


CMake Error at tests/CMakeLists.txt:212 (fuzztest_setup_fuzzing_flags):
  Unknown CMake command "fuzztest_setup_fuzzing_flags".
```

A cmake notice might be a little more understandable to the user.

jzern (approved):
I tested with cmake 3.21, it works as expected. Building with 3.29.6 works.

tests/CMakeLists.txt
wantehchang: > I don't see any workflows that set `AVIF_LOCAL_FUZZTEST`.

This is hidden in tests/oss-fuzz/build.sh. It is apparently invoked by cifuzz.yml, but I haven't figured out how it is done.

PR #2360, opened 2024-08-01
Fix Clang's -Wmissing-field-initializers warnings

wantehchang:
Fix -Wmissing-field-initializers warnings in avif_fuzztest_dec_incr.cc.


wantehchang:
The warnings were discovered by https://github.com/AOMediaCodec/libavif/pull/2356.

PR #2361, opened 2024-08-01
Run cifuzz.yml on pull requests for CMakeLists.txt

wantehchang:
All the other CI workflows that run on pull requests for 'cmake/**' also run on pull requests for '**CMakeLists.txt'. So cifuzz.yml is the only exception.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - '.github/workflows/cifuzz.yml'
+      - '**CMakeLists.txt'
       - 'cmake/**'
       - 'ext/**'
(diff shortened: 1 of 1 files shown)


y-guyon (approved):

y-guyon:
> All the other CI workflows that run on pull requests for 'cmake/**' also run on pull requests for '**CMakeLists.txt'. So cifuzz.yml is the only exception.

This may be expected. Most CI workflows make sure the targets build and the tests pass. Fuzzing on the other hand is focused on finding implementation bugs. Up to you.

wantehchang:
Our build system is comprised of `'**CMakeLists.txt'` and `'cmake/**'`, so it makes sense to give them the same treatment.

PR #2362, opened 2024-08-01
In setup-common action, bump recent-cmake to 3.21

wantehchang:
Set recent-cmake to true in ci-unix-shared-installed.yml because it sets -DAVIF_ENABLE_NODISCARD=ON, which needs cmake 3.21 to force C23.


wantehchang:
@vrabaud FYI.

jzern (approved):

Issue #2365, opened 2024-08-02
CMAKE_CXX_FLAGS_RELEASE is empty if -DAVIF_CODEC_AOM=LOCAL -DAVIF_LIBYUV=OFF is specified


wantehchang:
@fdintino @vrabaud 

To reproduce this bug, run the following command on Linux. Make sure the ext/aom/ directory doesn't exist.

```
rm -rf build
mkdir build
cd build
cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DAVIF_CODEC_AOM=LOCAL -DAVIF_BUILD_APPS=ON -DAVIF_BUILD_TESTS=ON -DAVIF_GTEST=LOCAL -DAVIF_LIBYUV=OFF
ninja -v
```

Note that C++ files such as aviftest_helpers.cc and avifincrtest.cc are compiled without the `-O3 -DNDEBUG` flags commonly used in the Release build configuration.

Note that the problem is gone if I change `-DAVIF_LIBYUV=OFF` to `-DAVIF_LIBYUV=LOCAL`.

My experimets showed that this bug only occurs when we use FetchContent to build libaom locally. I think this bug is caused by the code in cmake/Modules/LocalAom.cmake. It touches variables like `CMAKE_C_FLAGS_RELEASE`. Seeing how complicated that code is, I think it would be better to build libaom using ExternalProject.

You can apply this patch and your will see that the `CMAKE_CXX_FLAGS_RELEASE` variable is empty.

```
diff --git a/tests/CMakeLists.txt b/tests/CMakeLists.txt
index ec972dff..3e22775d 100644
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -42,6 +42,8 @@ endforeach()
 
 if(AVIF_ENABLE_FUZZTEST OR AVIF_ENABLE_GTEST OR AVIF_BUILD_APPS)
     add_library(aviftest_helpers OBJECT gtest/aviftest_helpers.cc)
+    message(STATUS "WTC WTC WTC: CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
+    message(STATUS "WTC WTC WTC: CMAKE_CXX_FLAGS_RELEASE: ${CMAKE_CXX_FLAGS_RELEASE}")
     target_link_libraries(aviftest_helpers PUBLIC avif_apps avif)
     target_link_libraries(aviftest_helpers PRIVATE avif_enable_warnings)
[...]

PR #2367, opened 2024-08-05
Fix empty CMAKE_CXX_FLAGS_RELEASE if libaom local

wantehchang:
Fix empty CMAKE_CXX_FLAGS_RELEASE if libaom is built locally using FetchContent and AVIF_LIBYUV is OFF.

libaom sets CMAKE_CXX_FLAGS_RELEASE and other flags as CACHE variables. To undo the changes libaom makes, unset the variables as cache variables so that they can be reset to their original values as normal variables.

Fix https://github.com/AOMediaCodec/libavif/issues/2365.

### DIFF
--- a/cmake/Modules/LocalAom.cmake
+++ b/cmake/Modules/LocalAom.cmake
@@ -143,9 +143,11 @@ else()
         endif()
         avif_set_aom_compile_options(aom ${_aom_config})
 
+        # Restore the variables.
         foreach(_config_setting CMAKE_C_FLAGS CMAKE_CXX_FLAGS CMAKE_EXE_LINKER_FLAGS)
             foreach(_config_type DEBUG RELEASE MINSIZEREL RELWITHDEBINFO)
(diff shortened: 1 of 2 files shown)


cmake/Modules/LocalAom.cmake
```diff
@@ -145,7 +145,8 @@ else()
 
         foreach(_config_setting CMAKE_C_FLAGS CMAKE_CXX_FLAGS CMAKE_EXE_LINKER_FLAGS)
             foreach(_config_type DEBUG RELEASE MINSIZEREL RELWITHDEBINFO)
-                set(${_config_setting}_${_config_type} ${${_config_setting}_${_config_type}_ORIG} CACHE STRING "" FORCE)
+                unset(${_config_setting}_${_config_type} CACHE)
```
jzern, line 148: It might be helpful to document these two loops as restoring the variables.

jzern (approved):

cmake/Modules/LocalAom.cmake
wantehchang: Done.

Issue #2345, opened 2024-07-31
Investigate whether we still need to compile with /source-charset:utf-8 /execution-charset:us-ascii on Windows


wantehchang:
@tongyuantongyu Yuan: I need your review of this issue.

You added the MSVC compiler flags `/source-charset:utf-8 /execution-charset:us-ascii` to libavif. I am wondering if we should still compile libavif with these flags for three reasons:

1. While adding clang-cl compilation support, I found that clang-cl does not support `/execution-charset:us-ascii`.
2. The vcpkg libavif package [removes these flags](https://github.com/microsoft/vcpkg/blob/cacf5994341f27e9a14a7b8724b0634b138ecb30/ports/libavif/disable-source-utf8.patch).
3. avifenc.exe and avifdec.exe are now [configured to use the UTF-8 code pages](https://github.com/AOMediaCodec/libavif/commit/3ec01cefd1ddd266a622d5e114a0888581b68f4a).

It seems that the third reason alone should make it unnecessary to compile libavif with `/execution-charset:us-ascii`. I am not familiar with internationalization issues on Windows, so I would appreciate your expert opinion. Thanks!

tongyuantongyu:
This flag was added to ensure we don't use non-ASCII characters in strings in the source code. Here are the two reasons:
- Some terminals may not be able to render non-ASCII characters, or render it correctly, potentially creating confusion to the user.
- Some programmatic way of calling subprocesses may assume stdout is using the active codepage and fail to decode.

So the flag is more like a canary, than to solve some Windows-specific issue.

wantehchang:
Yuan: Thank you for the reply. Using non-ASCII characters in strings in the source code is very unlikely. The only reasons I can think of are:
- (A programming error) copy and paste texts from a PDF or rich text file that has non-ASCII quotation marks or dashes.
- Personal names (especially the European Latin character sets).
- The copyright, trademark, or register trademark symbols.

All of these reasons are unlikely. So I will probably just remove these flags or try to improve the comments for them.

I am afraid that none of the libavif maintainers understand the purpose of these flags, so after the recent rework of the build system, the flags are only used when compiling the source files in the avif and avif_apps libraries. It is a maintenance burden.

wantehchang:
[...]

PR #2368, opened 2024-08-05
Use /execution-charset:us-ascii only with /WX

wantehchang:
Use the MSVC /source-charset:utf-8 /execution-charset:us-ascii options only when warnings are treated as errors (/WX) so that we never substitute a question mark (?) for a non-ASCII character in string or character literals.

Update the comments for /source-charset:utf-8
/execution-charset:us-ascii basec on improved understranding of these options.

Related to https://github.com/AOMediaCodec/libavif/issues/2345.


wantehchang:
@tongyuantongyu Yuan: Please do a preliminary review of this pull request. Because of the substitution of a question mark (?) for a non-ASCII character in string or character literals, my first thought is that these options should only be used when warnings are treated as errors. So that's what this pull request does.

You also made a good point that for compatibility with clang-cl and with non-Windows platforms, we should use the `/utf-8` option instead. We can do that next, but it's good to give this a try. However, if you or my coworkers don't understand my comment in this pull request, then I will switch to the `/utf-8` option directly.

tongyuantongyu:
I prefer to always add `/source-charset:utf-8` whether or not `/WX` is used. `/source-charset:utf-8` is more to "tell MSVC to treat source code in UTF-8" than "validate that the source code contains only characters that can be represented in UTF-8", otherwise MSVC will default to use ANSI code page, decoding source code wrongly, and emit C4819 when a decode error is encounted.

My intention to add `/source-charset:utf-8` was to make MSVC decode source code correctly. My intention to add `/execution-charset:us-ascii` was to ensure we don't force user to deal with i18n and character encodings when using CLI. If you agree with my intention, let's add these to the comments; or we need to recheck whether we really want the two flags.

Your new comments explained many possible misunderstandings a user or new maintainer may have. Can you move the explanations to a new paragraph, so the first paragraph explains what we want to do and what the flags do, and the second paragraph explains more details (like "check the strings you added" when seeing C4566) and resolves misunderstandings (like it won't interfere escape sequences and won't prevent printing user provided UTF-8 strings)?

PR #2370, opened 2024-08-06
Remove decoder param of avifCodecGetNextImageFunc

wantehchang:
Remove the decoder parameter of avifCodecGetNextImageFunc so that the avifCodec struct is independent of the avifDecoder struct.

Add the maxThreads and imageSizeLimit fields to the avifCodec struct so that they remain available to avifCodecGetNextImageFunc.

This allows us to remove the avifDecoder local variable in avifEncoderDecodeSatoBaseImage().

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -5711,7 +5711,9 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder, uint32_t nextIma
         }
 
         avifBool isLimitedRangeAlpha = AVIF_FALSE;
-        if (!tile->codec->getNextImage(tile->codec, decoder, sample, avifIsAlpha(tile->input->itemCategory), &isLimitedRangeAlpha, tile->image)) {
+        tile->codec->maxThreads = decoder->maxThreads;
+        tile->codec->imageSizeLimit = decoder->imageSizeLimit;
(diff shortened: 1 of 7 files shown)


src/read.c
```diff
@@ -5711,7 +5711,9 @@ static avifResult avifDecoderDecodeTiles(avifDecoder * decoder, uint32_t nextIma
         }
 
         avifBool isLimitedRangeAlpha = AVIF_FALSE;
-        if (!tile->codec->getNextImage(tile->codec, decoder, sample, avifIsAlpha(tile->input->itemCategory), &isLimitedRangeAlpha, tile->image)) {
+        tile->codec->maxThreads = decoder->maxThreads;
+        tile->codec->imageSizeLimit = decoder->imageSizeLimit;
```
wantehchang, line 5715: To match the current behavior, I copy these two fields from `decoder` before each `getNextImage()` call.  In theory this allows users to change `decoder->maxThreads` or `decoder->imageSizeLimit` before decoding each image. Not sure if this actually works, because I believe all the codecs use these two fields in the first `getNextImage()` call only (when the underlying decoder instance is created).

An alternative is to copy these two fields only when `tile->codec` is created.

y-guyon (approved):
Thank you

PR #2372, opened 2024-08-06
Check if it is safe to call libyuv or libsharpyuv

wantehchang:
The width, height, and stride parameters of libyuv and libsharpyuv functions are all of the int type. The corresponding fields in libavif are of the uint32_t type. So we need to check if it is safe to pass the libavif uint32_t values to libyuv and libsharpyuv.

Related to https://github.com/AOMediaCodec/libavif/issues/2271.

### DIFF
--- a/src/reformat_libsharpyuv.c
+++ b/src/reformat_libsharpyuv.c
@@ -4,11 +4,17 @@
 #include "avif/internal.h"
 
 #if defined(AVIF_LIBSHARPYUV_ENABLED)
+#include <limits.h>
 #include <sharpyuv/sharpyuv.h>
 #include <sharpyuv/sharpyuv_csp.h>
 
 avifResult avifImageRGBToYUVLibSharpYUV(avifImage * image, const avifRGBImage * rgb, const avifReformatState * state)
 {
+    // The width, height, and stride parameters of SharpYuvConvertWithOptions()
+    // and SharpYuvConvert() are all of the int type.
+    if (rgb->width > INT_MAX || rgb->height > INT_MAX || rgb->rowBytes > INT_MAX || image->yuvRowBytes[AVIF_CHAN_Y] > INT_MAX) {
+        return AVIF_RESULT_NOT_IMPLEMENTED;
+    }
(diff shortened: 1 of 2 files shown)


src/reformat_libsharpyuv.c
```diff
@@ -9,6 +9,11 @@
 
 avifResult avifImageRGBToYUVLibSharpYUV(avifImage * image, const avifRGBImage * rgb, const avifReformatState * state)
 {
+    // The width, height, and stride parameters of SharpYuvConvertWithOptions()
+    // and SharpYuvConvert() are all of the int type.
+    if (rgb->width > INT_MAX || rgb->height > INT_MAX || rgb->rowBytes > INT_MAX || image->yuvRowBytes[AVIF_CHAN_Y] > INT_MAX) {
+        return AVIF_RESULT_NOT_IMPLEMENTED;
+    }
```
wantehchang, line 16: Rather than checking this only when we are calling into libyuv or libsharpyuv, I am wondering if we should check this in the public `avifImageRGBToYUV()` and `avifImageYUVToRGB()` functions.

y-guyon (approved):

src/reformat_libsharpyuv.c
y-guyon: We can do both to be safe. I prefer to keep the check close to where the values are used.

wantehchang: Yannis: Thank you for your comment. I also just realized that the maximum width and height that AVIF can achieve may be smaller than INT_MAX. We calculated the maximum width and height of AVIF before (assuming max AV1 frame size of 64K x 64K and the maximum HEIF/MIAF image grid), but I can't find that email thread.

On the other hand, the `avifImageRGBToYUV()` and `avifImageYUVToRGB()` functions are really independent of AVIF, so it may be good to let them "be all they can be" and only impose the limits related to YUV<->RGB conversions.

I'd like to think about this a bit more, so I won't merge this pull request right away.

wantehchang:
I didn't have time to work on this more yesterday, so I will merge this version of the pull request first.

PR #2373, opened 2024-08-06
Divide rowBytes by 2 if casting pixels to uint16_t

wantehchang:
When calling the libyuv function HalfFloatPlane(), divide rgb->rowBytes by 2 because rgb->pixels is cast to uint16_t *.

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -1132,9 +1132,9 @@ avifResult avifRGBImageToF16LibYUV(avifRGBImage * rgb)
 {
     const float scale = 1.0f / ((1 << rgb->depth) - 1);
     const int result = HalfFloatPlane((const uint16_t *)rgb->pixels,
-                                      rgb->rowBytes,
+                                      rgb->rowBytes / 2,
                                       (uint16_t *)rgb->pixels,
-                                      rgb->rowBytes,
+                                      rgb->rowBytes / 2,
                                       scale,
                                       rgb->width * avifRGBFormatChannelCount(rgb->format),
                                       rgb->height);


src/reformat_libyuv.c
```diff
@@ -1132,9 +1132,9 @@ avifResult avifRGBImageToF16LibYUV(avifRGBImage * rgb)
 {
     const float scale = 1.0f / ((1 << rgb->depth) - 1);
     const int result = HalfFloatPlane((const uint16_t *)rgb->pixels,
-                                      rgb->rowBytes,
+                                      rgb->rowBytes / 2,
                                       (uint16_t *)rgb->pixels,
-                                      rgb->rowBytes,
+                                      rgb->rowBytes / 2,
```
wantehchang, line 1137: Vignesh: This code was originally added in https://github.com/AOMediaCodec/libavif/commit/fb17c6a28. I think this is a bug. Do you know why this bug hasn't been reported?

The `HalfFloatPlane()` function does the following:
```
  for (y = 0; y < height; ++y) {
    HalfFloatRow(src_y, dst_y, scale, width);
    src_y += src_stride_y;
    dst_y += dst_stride_y;
  }
```

So its stride parameters should be in units of `uint16_t`.

vigneshvg: This libyuv function is an exception where it expects stride to be in `uint8_t` units.

If you go up a few lines above the snippet that you pasted, you can see that libyuv converts the stride to `uint16_t` units internally.

```
src_stride_y >>= 1;
dst_stride_y >>= 1;
```

We should definitely document this so that it does not confuse the readers.

wantehchang: Vignesh: Thank you very much! I will abandon this pull request and write a new one to add a comment. I should also add a comment to the libyuv header about this.

PR #2375, opened 2024-08-07
Add comment about stride params of HalfFloatPlane

wantehchang:


src/reformat_libyuv.c
```diff
@@ -1131,6 +1131,7 @@ avifResult avifRGBImageUnpremultiplyAlphaLibYUV(avifRGBImage * rgb)
 avifResult avifRGBImageToF16LibYUV(avifRGBImage * rgb)
 {
     const float scale = 1.0f / ((1 << rgb->depth) - 1);
+    // Note: Unlike CopyPlane_16, HalfFloatPlane wants the stride in bytes.
```
vigneshvg, line 1134: it is also unlike all the other 16bit conversion functions which expect the stride in `uint16_t`.

so the comment can just be about `HalfFloatPlane` without singling out `CopyPlane_16`.

```
// Note: HalfFloatPlane requires the stride to be in bytes.
```

wantehchang: Done. Thanks!

PR #2376, opened 2024-08-07
AVIF_ENABLE_EXPERIMENTAL_MINI instead of METAV1

y-guyon:
Update the experimental prototype implementation to the latest specification draft.


wantehchang (approved):
LGTM. Thanks again for revising the spec and the implementation!

I only skimmed through the pull request and made sure it is a no-op if the experimental feature is disabled. I'd like to read it again and compare with the spec, because that's usually when I actually review the spec carefully, but I think it is fine to merge this PR first.

PR #2378, opened 2024-08-09
Empty HandlerBox name instead of "libavif"

y-guyon:
Empty HandlerBox name by default.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -216,8 +216,6 @@ typedef enum avifHeaderFormat
 {
     // AVIF file with an "avif" brand, a MetaBox and all its required boxes for maximum compatibility.
     AVIF_HEADER_FULL,
-    // Same as AVIF_HEADER_FULL but the name field of the HandlerBox is empty instead of "libavif".
-    AVIF_HEADER_DEFAULT,
 #if defined(AVIF_ENABLE_EXPERIMENTAL_MINI)
     // AVIF file with a "mif3" brand and a MinimizedImageBox to reduce the encoded file size.
     // This is based on the w24144 "Low-overhead image file format" MPEG proposal for HEIF.
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -216,6 +216,8 @@ typedef enum avifHeaderFormat
 {
     // AVIF file with an "avif" brand, a MetaBox and all its required boxes for maximum compatibility.
     AVIF_HEADER_FULL,
+    // Same as AVIF_HEADER_FULL but the name field of the HandlerBox is empty instead of "libavif".
+    AVIF_HEADER_DEFAULT,
 #if defined(AVIF_ENABLE_EXPERIMENTAL_MINI)
     // AVIF file with a "mif3" brand and a MinimizedImageBox to reduce the encoded file size.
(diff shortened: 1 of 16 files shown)


include/avif/avif.h
```diff
@@ -216,6 +216,8 @@ typedef enum avifHeaderFormat
 {
     // AVIF file with an "avif" brand, a MetaBox and all its required boxes for maximum compatibility.
     AVIF_HEADER_FULL,
+    // Same as AVIF_HEADER_FULL but the name field of the HandlerBox is empty instead of "libavif".
```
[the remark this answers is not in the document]
wantehchang: @leo-barnes Leo: This is the libavif  HandlerBox name issue I mentioned at the end of the AOMedia Image Container subgroup meeting today.

Yannis: I suggest we make this change without adding a new AVIF header format option.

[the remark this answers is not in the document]
leo-barnes: I can confirm that Apple platforms create HEIF files with empty handler name.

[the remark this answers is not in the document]
y-guyon: > Yannis: I suggest we make this change without adding a new AVIF header format option.

Done. PTAL.

src/write.c
```diff
@@ -771,6 +771,18 @@ static avifResult avifEncoderWriteExtendedColorProperties(avifRWStream * dedupSt
     return AVIF_RESULT_OK;
 }
 
+static avifResult avifEncoderWriteHandlerBox(avifRWStream * s, const char handler_type[4])
```
wantehchang, line 774: Nit: The function name should start with `avifRWStream` (the type of the first/main parameter) intead of `avifEncoder`.

wantehchang, line 774: Nit: handler_type => handlerType

tests/gtest/avifheadertest.cc
```diff
@@ -0,0 +1,39 @@
...
+  ImagePtr image = testutil::CreateImage(12, 34, 8, AVIF_PIXEL_FORMAT_YUV420,
+                                         AVIF_PLANES_ALL);
+  ASSERT_NE(image, nullptr);
+  testutil::FillImageGradient(image.get());
+  const testutil::AvifRwData avif = testutil::Encode(image.get());
+  ASSERT_NE(avif.data, nullptr);
+
+  // Make sure the HandlerBox is as small as possible, meaning its name field is
+  // empty.
+  const uint8_t* kHdlr = reinterpret_cast<const uint8_t*>("hdlr");
```
wantehchang, line 23: Nit: The way `kHdlr` is defined here, it is a variable, not a constant. (It is not even a const variable.) So it should be named like a variable:

```
    const uint8_t* const hdlr = reinterpret_cast<const uint8_t*>("hdlr");
```

Alternatively, we can define it as a constant:

```
    const uint8_t kHdlr[] = { 'h', 'd', 'l', 'r' };
```

tests/gtest/avifheadertest.cc
```diff
@@ -0,0 +1,39 @@
...
+  uint8_t* hdlr_position =
```
wantehchang, line 24: Nit: add `const`:

```
  const uint8_t* hdlr_position =
```

wantehchang (approved):
LGTM. Thanks. I suggest some changes.

PR #2379, opened 2024-08-09
Fix sizeof call in memset

maryla-uc:


y-guyon (approved):

src/gainmap.c
```diff
@@ -55,7 +55,7 @@ avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, c
 
 static void avifGainMapMetadataSetDefaults(avifGainMapMetadataDouble * metadata)
 {
-    memset(metadata, 0, sizeof(avifGainMapMetadata));
+    memset(metadata, 0, sizeof(*metadata));
```
wantehchang, line 58: The function name can begin with "avifGainMapMetadataDouble" to follow the naming convention,.

wantehchang (approved):
LGTM. We can look at our Coverity defects to see if it detected this bug.

wantehchang (approved):

wantehchang:
I checked our Coverity defects. It did not seem to detect this error. Strange.

maryla-uc:
I assume the bug was not caught because the two structs happen to have the exact same size.

Issue #2369, opened 2024-08-06
Check version numbers of boxes related to the ISO gain map experimental feature


wantehchang:
@maryla-uc @vigneshvg @y-guyon @ccameron-chromium

Please audit our ISO gain map code and make sure we check the version numbers of the boxes related to ISO gain maps.

I took a look at the `avifParseToneMappedImageBox()` function in src/read.c. We check `version` and `minimum_version` but ignore `writer_version`. Should we check `writer_version`?

Is there any other box we need to check?

ccameron-chromium:
`writer_version` is added for future compatibility, where `writer_version>0` with `minimum_version=0` would signal new but backwards-compatible features. It is indeed to be ignored.

wantehchang:
Alternatively we can restrict that check to `writer_version=0`. That may be a better fix:

```
    return writerVersion != 0 || avifROStreamRemainingBytes(&s) == 0;
```

wantehchang:
Another thing to check: If `avifParseToneMappedImageBox()` sees a tmap box of an unsupported newer version, I think the caller of `avifParseToneMappedImageBox()` should gracefully ignore the tmap box (as if it were not there) rather than return an error.

ccameron-chromium:
> Another thing to check: If `avifParseToneMappedImageBox()` sees a tmap box of an unsupported newer version, I think the caller of `avifParseToneMappedImageBox()` should gracefully ignore the tmap box (as if it were not there) rather than return an error.

Yes, I noticed this behavior, and I agree that it should be ignored and not fatal for the image decode.

maryla-uc:
Thanks for looking into this Wan-Teh, I agree with your interpretation and proposed changes.
I have a couple of questions:
- What about the '[version](https://github.com/AOMediaCodec/libavif/blob/47f154ae4cdefbdb7f9d86c0017acfe118db260e/src/read.c#L1961)' tag that is read just before 'minimum_version'? Should we also simply ignore the tmap if the version is unknown, like 'minimum_version'?
- If there is a gainmap but we can't read it because it's an unknown version, should we still report that a gainmap in present? (See '[gainMapPresent](https://github.com/AOMediaCodec/libavif/blob/47f154ae4cdefbdb7f9d86c0017acfe118db260e/src/read.c#L5338)')

wantehchang:
Hi Maryla,

[...]

PR #2380, opened 2024-08-09
Ignore gain maps with unsupported version

maryla-uc:
Issue #2369


src/read.c
```diff
@@ -5384,8 +5395,19 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
+            if (isUnsupportedVersion) {
+                // Forget about the gain map.
+                toneMappedImageItem = NULL;
+                mainItems[AVIF_ITEM_GAIN_MAP] = NULL;
+                avifGainMapDestroy(decoder->image->gainMap);
+                decoder->image->gainMap = NULL;
+                decoder->gainMapPresent = AVIF_FALSE;
+            } else {
+                AVIF_CHECKERR(tmapParsingRes, AVIF_RESULT_INVALID_TONE_MAPPED_IMAGE);
+            }
```
wantehchang, line 5410: Nit: I suggest restructuring the if-else statement so that we first check the return value `tmapParsingRes` and then check the output parameter `isUnsupportedVersion`. This requires nested if statements, but it makes it clearer that `isUnsupportedVersion` is effectively an error/reason code that we check only when the function fails.

```
            if (!tmapParsingRes) {
                if (isUnsupportedVersion) {
                    // Forget about the gain map.
                    toneMappedImageItem = NULL;
                    mainItems[AVIF_ITEM_GAIN_MAP] = NULL;
                    avifGainMapDestroy(decoder->image->gainMap);
                    decoder->image->gainMap = NULL;
                    decoder->gainMapPresent = AVIF_FALSE;
                } else {
                }
            }
```

Note:  the current version also has a nested if statement hidden inside the `AVIF_CHECKERR` macro.

src/read.c
```diff
@@ -1953,26 +1953,34 @@ static avifBool avifParseImageGridBox(avifImageGrid * grid,
 
 #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
 
-static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata, const uint8_t * raw, size_t rawLen, avifDiagnostics * diag)
+static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata,
```
wantehchang, line 1956: Optional/future: An alternative to adding the output parameter `isUnsupportedVersion` is to change the return type to `avifResult` and use the error code `AVIF_RESULT_NOT_IMPLEMENTED` to indicate an unsupported version.

This requires changing all the `AVIF_CHECK` macros used in the function.

include/avif/avif.h
```diff
@@ -1312,16 +1312,21 @@ typedef struct avifDecoder
     // Version 1.1.0 ends here. Add any new members after this line.
 
 #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
-    // This is true when avifDecoderParse() detects a gain map.
+    // If 'enableParsingGainMapMetadata' is false: this is true when avifDecoderParse()
```
wantehchang, line 1315: Nit: change "this" to "gainMapPresent" in these comments.

include/avif/avif.h
```diff
@@ -1312,16 +1312,21 @@ typedef struct avifDecoder
...
+    // detects a gain map.
+    // If 'enableParsingGainMapMetadata' is true: this is true when avifDecoderParse()
+    // detects a gain map whose metadata is a supported version.
```
wantehchang, line 1318: Nit/optional: It seems clearer to move the `gainMapPresent` field after the `enableParsingGainMapMetadata` field.

We can move `gainMapPresent` after `ignoreColorAndAlpha`, so that we first declare the input options, and then declare the output options.

include/avif/avif.h
```diff
@@ -1312,16 +1312,21 @@ typedef struct avifDecoder
...
+    // If 'enableParsingGainMapMetadata' is also true, the gain map is only decoded if
+    // the gan map's metadata is a supported version.
+    // 'gainMapPresent' is still set if the presence of a gain map is detected, regardless
     // of this setting.
     avifBool enableDecodingGainMap;
     // Enable parsing the gain map metadata if present (defaults to AVIF_FALSE).
-    // gainMapPresent is still set if the presence of a gain map is detected, regardless
-    // of this setting.
+    // This setting can affect the value of 'gainMapPresent', see the description of
+    // 'gainMapPresent' above.
```
wantehchang, line 1329: If you adjusted the declaration order, change "above" to "below".

src/read.c
```diff
@@ -1953,26 +1953,34 @@ static avifBool avifParseImageGridBox(avifImageGrid * grid,
...
+                                            const uint8_t * raw,
+                                            size_t rawLen,
+                                            avifBool * isUnsupportedVersion,
```
wantehchang, line 1959: Nit: Document that when the function fails, the output parameter `isUnsupportedVersion` indicates whether the metadata is an unsupported version or is invalid.

wantehchang (approved):
LGTM. Thank you very much for working through the version checking and adding tests. It's good that you made the version check future-proof (by adding the `supportedMetadataVersion` variable).

Nit: The commit message could mention the change to the check of remaining data.

src/read.c
wantehchang, line 1956: I wonder if we should add a changelog item about this change to gain map version check.

src/read.c
```diff
@@ -1953,61 +1953,35 @@ static avifBool avifParseImageGridBox(avifImageGrid * grid,
 
 #if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
 
-static avifBool avifParseToneMappedImageBox(avifGainMapMetadata * metadata,
-                                            const uint8_t * raw,
-                                            size_t rawLen,
-                                            avifBool * isUnsupportedVersion,
-                                            avifDiagnostics * diag)
+static avifBool avifParseGainMapMetadta(avifGainMapMetadata * metadata, avifROStream * s)
```
wantehchang, line 1956: Nit/optional: Add the `writerVersion` parameter to be future-proof. It will be ignored in the current implementation, so we will need to add a `(void)` cast.

We can also wait until this is really needed.

wantehchang (approved):

src/read.c
wantehchang: The remaining data check could be moved to this function. Conceptually it seems to belong in this function.

If so, then this function will use the `writerVersion` parameter. But we will also need to pass `supportedMetadataVersion` to this function or make it a constant in the file scope, so I'm not sure if this is a good idea.

maryla-uc: Thanks for the suggestions, but I think we can keep things simple (and therefore more readable) for now.

src/read.c
maryla-uc: Done. I thought about that too but thought that it would make all the checks annoying, but actually extracting a sub function solves this issue. I think it's indeed cleaner now.

wantehchang (approved):

wantehchang (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Renamed AVIF_ENABLE_EXPERIMENTAL_METAV1 to AVIF_ENABLE_EXPERIMENTAL_MINI and
   updated the experimental reduced header feature to the latest specification
   draft.
+* Ignore gain maps with unsupported metadata.
```
wantehchang, line 18: We should also mention we handle writer_version > 0 correctly. (The previous remaining data check would fail if new fields were added in writer_version > 0.) This can be done in a follow-up PR.

wantehchang: Done in https://github.com/AOMediaCodec/libavif/pull/2382.

leo-barnes:
> @maryla-uc I notice that `supported_gainmap_writer_version_with_extra_bytes.avif` is not actually a valid `tmap` file. The `ispe` property associated with the `tmap` has width and height set to 0.
> 
> ```
>       [ 8] ('ispe' "Image Spatial Extents", size = 20, offset = 493) {
>         Version: 0, Flags: 0x000000
>         Image Width: 0
>         Image Height: 0
>       }
> ```

Filed a new issue for it: https://github.com/AOMediaCodec/libavif/issues/2391

PR #2381, opened 2024-08-09
Add custom codec callback in avifEncoder

y-guyon:


wantehchang (commented):
Yannis: I only took a cursory look at this pull request. The commit message has the subject line "Add custom codec callback in avifEncoder". That is confusing, because these callbacks do not seem to provide access to a custom AV1 codec. Should we change "custom codec" to "custom encode data"?

y-guyon:
> Yannis: I only took a cursory look at this pull request.

Thank you. This PR is a proof-of-concept to kickstart a discussion, not a request to merge code.

> The commit message has the subject line "Add custom codec callback in avifEncoder". That is confusing, because these callbacks do not seem to provide access to a custom AV1 codec.

I did not mention AV1. This could be used for any kind of coded samples (AV2?) but to be compliant with the AVIF specification it would have to be AV1 samples indeed.

> Should we change "custom codec" to "custom encode data"?

To me the piece of software that generates encoded data is a codec, so "custom codec callback" is accurate. Whether that is used to feed already compressed samples or to actually encode pixels on the spot, it is opaque to the libavif API.

include/avif/avif.h
```diff
@@ -1442,6 +1442,40 @@ typedef struct avifScalingMode
...
+// Returns an error otherwise.
+// TODO: output sample should be avifEncodeSampleArray
+typedef avifResult (*avifEncoderCustomEncodeImageFunc)(struct avifEncoder * encoder,
+                                                       const avifImage * image,
+                                                       avifBool alpha,
+                                                       int tileRowsLog2,
+                                                       int tileColsLog2,
+                                                       int quantizer,
+                                                       avifBool disableLaggedOutput,
+                                                       avifAddImageFlags addImageFlags);
```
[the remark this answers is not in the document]
y-guyon: Done. I think `avifEncoderCustomEncodeImageArgs` is extensible.

y-guyon:
Note that `avifCodecEncodeImageFunc` can use `avifEncoderCustomEncodeImageArgs` too. It can be implemented in another PR to simplify the review of this PR.

y-guyon:
I tagged the new `avifEncoder` fields as "Experimental feature" as there may be overlooked edge cases.

The new API can be fuzzed in another PR once the design is settled.

include/avif/avif.h
```diff
@@ -1524,6 +1547,22 @@ typedef enum avifAddImageFlag
...
+typedef struct avifEncoderCustomEncodeImageArgs
+{
+    // Encoding settings requested by the avifEncoder instance for the current AV1 coded image item or sequence.
+    avifAddImageFlags addImageFlags;
+    int quantizer;    // AV1 quality setting in range [AVIF_QUANTIZER_BEST_QUALITY:AVIF_QUANTIZER_WORST_QUALITY].
+    int tileRowsLog2; // Logarithm in base 2 of the number of AV1 tile rows.
+    int tileColsLog2; // Logarithm in base 2 of the number of AV1 tile columns.
+
+    // Description of the current AV1 coded image item or sequence.
+    avifBool isAlpha; // True if the current AV1 image item or sequence holds the translucency layer.
```
maryla-uc, line 1560: Maybe use an enum?

include/avif/avif.h
```diff
@@ -1524,6 +1547,22 @@ typedef enum avifAddImageFlag
...
 typedef uint32_t avifAddImageFlags;
 
+// Arguments passed to avifEncoderCustomEncodeImageFunc by the avifEncoder instance.
+typedef struct avifEncoderCustomEncodeImageArgs
+{
+    // Encoding settings requested by the avifEncoder instance for the current AV1 coded image item or sequence.
+    avifAddImageFlags addImageFlags;
+    int quantizer;    // AV1 quality setting in range [AVIF_QUANTIZER_BEST_QUALITY:AVIF_QUANTIZER_WORST_QUALITY].
+    int tileRowsLog2; // Logarithm in base 2 of the number of AV1 tile rows.
+    int tileColsLog2; // Logarithm in base 2 of the number of AV1 tile columns.
```
maryla-uc, line 1557: You would need to have the coordinates of the current tile being encoded, no?

include/avif/avif.h
```diff
@@ -1414,15 +1414,31 @@ AVIF_API avifResult avifDecoderNthImageMaxExtent(const avifDecoder * decoder, ui
...
+// Returns AVIF_RESULT_NO_CONTENT if the AV1 codec encoding pipeline should be run for that element.
+// Returns an error otherwise.
+typedef avifResult (*avifEncoderCustomEncodeImageFunc)(struct avifEncoder * encoder,
+                                                       const avifImage * image,
+                                                       const struct avifEncoderCustomEncodeImageArgs * args);
+// Only called if avifEncoderCustomEncodeImageFunc returned AVIF_RESULT_OK.
+// Returns AVIF_RESULT_OK every time it outputs an AV1 sample.
+// Returns AVIF_RESULT_NO_IMAGES_REMAINING once all samples were output.
+// Returns an error otherwise.
+typedef avifResult (*avifEncoderCustomEncodeFinishFunc)(struct avifEncoder * encoder, avifROData * sample);
```
maryla-uc, line 1440: How does this function know which image's samples to output?

include/avif/avif.h
y-guyon: Added `avifEncoderCustomEncodeImageItemType`.

include/avif/avif.h
y-guyon: Added `gridRow`/`gridColumn` in `avifEncoderCustomEncodeImageItem`.

include/avif/avif.h
y-guyon: Done through `avifEncoderCustomEncodeImageItem`.

y-guyon:
I had another go at it to answer Maryla's comments but the added API starts to be large. I wonder if this is still worth it. Thoughts?

s1ddok:
@y-guyon do you think this PR leads us to be able to demux AVIFs? I'm looking for a way to access raw av1 encoded packets to decode on them on GPU. previously you pointed to this PR as the one that can potentially enable it. do you think it is feasible?

y-guyon:
> do you think this PR leads us to be able to demux AVIFs?

No, this PR only allows already encoded AV1 OBU sequences to be passed to libavif at encoding. I believe you would need something similar but on the decoding side.

maryla-uc:
Other libraries should be able to do that. FFmpeg's libavformat, or gpac's libgpac?

Note that if the image is a grid, you will have to get each grid cell item data and stitch up the image yourself. Same thing if there is alpha, that will be a different item. Not sure libavformat can get you the different item data, but libgpac should be able to, although I haven't actually tried either.

Issue #2077, opened 2024-03-25
Roll YCgCo-R out


vrabaud:
Since it has been approved (but not published yet as of 25/03/2024, cf https://www.itu.int/rec/T-REC-H.273), we can rool YCgCo-R.
We first need to roll the decoder out by default, and once there are enough clients, we can roll the encoder out by default.

vrabaud:
Waiting on https://www.itu.int/rec/T-REC-H.273 to be public. Looking at the draft at https://jvet-experts.org/doc_end_user/documents/30_Antalya/wg11/JVET-AD1008-v1.zip (JVET-AF1008), it seems the enums have changed to 16 and 17.

Note that equations E-14 and E-15 have one error and they need to be subtractions instead of additions (BitdepthY - 2 and BitdepthY - 1).

wantehchang:
[H.273](https://www.itu.int/rec/T-REC-H.273) (09/23) has been published. However, it is already in the suspended state, because it will be replaced by H.273 (07/24).

In H.273 (09/23) the values of YCgCo-Re and YCgCo-Ro are 16 and 17, respectively. So we will need to change `AVIF_MATRIX_COEFFICIENTS_YCGCO_RE` and `AVIF_MATRIX_COEFFICIENTS_YCGCO_RO` to the new values.

wantehchang:
After consulting with other members of the AOMedia Image Container subgroup, we will go ahead and promote YCgCo-Re and YCgCo-Ro to libavif's stable API based on [H.273](https://www.itu.int/rec/T-REC-H.273) (09/23).

I suggest we plan to make this change early next week, after the Chrome 129 release branch is created. However, if this is low risk, we don't need to wait.

@vigneshvg Vignesh: do we need to change CrabbyAvif at the same time?

vigneshvg:
> @vigneshvg Vignesh: do we need to change CrabbyAvif at the same time?

yes, there probably has to be some minor changes. i will take a look at it and make the changes.

y-guyon:
Freely available H.273 V4 (07/24) mentions:
| Value | Matrix coefficients |
|-|-|
|16 |YCgCo-Re|
|17| YCgCo-Ro|

See https://www.itu.int/rec/T-REC-H.273-202407-I/en.

novomesk:
Hello,
I am encountering an interesting issue in Google Chrome Version 143.0.7499.170 (Official Build) (64-bit)

[...]

PR #2384, opened 2024-08-12
Update AVIF_MATRIX_COEFFICIENTS_YCGCO_RE and _RO

y-guyon:
to the values defined by H.273 (09/23).

See https://github.com/AOMediaCodec/libavif/issues/2077.


wantehchang (approved):
LGTM. Thank you, Yannis.

@vrabaud FYI.

include/avif/avif.h
```diff
@@ -400,8 +400,8 @@ enum
     AVIF_MATRIX_COEFFICIENTS_CHROMA_DERIVED_CL = 13,
     AVIF_MATRIX_COEFFICIENTS_ICTCP = 14,
 #if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
-    AVIF_MATRIX_COEFFICIENTS_YCGCO_RE = 15,
-    AVIF_MATRIX_COEFFICIENTS_YCGCO_RO = 16,
+    AVIF_MATRIX_COEFFICIENTS_YCGCO_RE = 16,
```
wantehchang, line 403: Optional: It would be good to add a comment on what value 15 is. But I see value 3 is also missing and we don't have a comment for that. So it should be fine to not add a comment.

y-guyon: Are 3 and 15 both "Reserved" in the latest edition?

wantehchang: I checked H.273 (09/23). 3 is reserved, and 15 is IPT-C2.

Issue #2391, opened 2024-08-15
supported_gainmap_writer_version_with_extra_bytes.avif is invalid


leo-barnes:
@maryla-uc
I notice that supported_gainmap_writer_version_with_extra_bytes.avif is not actually a valid tmap file. The ispe property associated with the tmap has width and height set to 0.

      [ 8] ('ispe' "Image Spatial Extents", size = 20, offset = 493) {
        Version: 0, Flags: 0x000000
        Image Width: 0
        Image Height: 0
      }

leo-barnes:
The `tmap` is also missing a `pixi`. This is not mandated as necessary by the spec, but is a nice to have.

leo-barnes:
Ah. And another issue. The `altr` group ID is the same as the base item ID. Group and item ID's are not allowed to have collisions.

leo-barnes:
The `ispe`, `pixi` and ID collision issues are present in all of the files.

For `unsupported_gainmap_minimum_version.avif` there is an additional issue in that the `minimum_version` (99) is greater than `writer_version` (0). I'm not sure if the spec explicitly states that `minimum_version <= writer_version`, but having it be the opposite doesn't really make sense.

wantehchang:
Leo: Thank you very much for inspecting the test files and reporting the issues!

ISO/CD 21496-1 says:

> `writer_version` is the version of the metadata structure used by the writer. This shall be greater than
or equal to `minimum_version`.

libavif enforces this requirement:

```
    uint16_t writerVersion;
    AVIF_CHECKERR(avifROStreamReadU16(&s, &writerVersion), AVIF_RESULT_INVALID_TONE_MAPPED_IMAGE); // unsigned int(16) writer_version;
    AVIF_CHECKERR(writerVersion >= minimumVersion, AVIF_RESULT_INVALID_TONE_MAPPED_IMAGE);
```

The file name "unsupported_gainmap_minimum_version.avif" suggests that this test file is intended to test how the application handles an unsupported minimum_version. I wonder if Maryla generated the file so that it could be used to test the handling of two error conditions. Her [notes for this test file](unsupported_gainmap_minimum_version.avif) says:

> Contains a gain map with the `minimum_version` field set to 99 in the tmap box. `version` and `writer_version` are 0.

[...]

PR #2396, opened 2024-08-19
Fix group id conflicting with item id. #2391

maryla-uc:
Some test files still need to be updated, this will be done in another PR.


src/write.c
```diff
@@ -3294,7 +3294,14 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
     // Write grpl/altr box
 
     if (encoder->data->alternativeItemIDs.count) {
-        AVIF_CHECKRES(avifWriteAltrGroup(&s, &encoder->data->alternativeItemIDs));
+        // ISO/IEC FDIS 14496-12:2020 8.18.3.3:
+        // group_id is a non-negative integer assigned to the particular grouping that shall not be equal to any
+        // group_id value of any other EntityToGroupBox, any item_ID value of the hierarchy level
+        // (file, movie. or track) that contains the GroupsListBox, or any track_ID value (when the
+        // GroupsListBox is contained in the file level).
+        ++encoder->data->lastItemID;
```
y-guyon, line 3302: Check for overflows

y-guyon (approved):

src/write.c
maryla-uc: Done

PR #2398, opened 2024-08-19
Check for ftyp precedence in avifParse()

y-guyon:


maryla-uc (approved):

src/read.c
```diff
@@ -4111,13 +4111,33 @@ static avifResult avifParse(avifDecoder * decoder)
...
+        avifBool isMini = AVIF_FALSE;
+        if (!isNonSkippableVariableLengthBox && !memcmp(header.type, "mini", 4)) {
+            isMini = AVIF_TRUE;
+            isNonSkippableVariableLengthBox = AVIF_TRUE;
+        }
 #endif
+
+        if (!isFtyp && (isNonSkippableVariableLengthBox || !memcmp(header.type, "free", 4) || !memcmp(header.type, "mdat", 4))) {
+            // Section 6.3.4 of ISO/IEC 14496-12:
+            //   The FileTypeBox shall occur before any variable-length box (e.g. movie, free space, media data).
```
wantehchang, line 4136: For AVIF, it may be fine to require the FileTypeBox be the first box in the file. The only example given in ISO/IEC 14496-12 that may go before the FileTypeBox is a file signature, but AVIF doesn't have a file signature.

PR #2399, opened 2024-08-19
Change gain map parsing API.

maryla-uc:
Now gainMapPresent is only populated if enableParsingGainMapMetadata is true. The combination of enableDecodingGainMap=true and enableParsingGainMapMetadata=false is no longer allowed (returns AVIF_RESULT_INVALID_ARGUMENT).

See discussion in https://github.com/AOMediaCodec/libavif/issues/2369

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Ignore gain maps with unsupported metadata. Handle gain maps with
   writer_version > 0 correctly.
   The combination of settings enableParsingGainMapMetadata=false with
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ The changes are relative to the previous release, unless the baseline is specifi
   draft.
 * Ignore gain maps with unsupported metadata. Handle gain maps with
   writer_version > 0 correctly.
+  The combination of settings enableParsingGainMapMetadata=false with
+  enableDecodingGainMap=true is no longer allowed and returns and invalid argument
```
y-guyon, line 21: an*

y-guyon (approved):

PR #2400, opened 2024-08-19
Improve 'mini' read and write of float and brand

y-guyon:
Check minor_version field of FileTypeBox.
Return proper error for samples in floating point format.
Verify codec config length.

This PR requires #2398 to be merged first.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -2472,13 +2472,13 @@ static avifResult avifEncoderWriteMiniBox(avifEncoder * encoder, avifRWStream * 
     AVIF_CHECKRES(avifRWStreamWriteBits(s, 0, 2)); // bit(2) version = 0;
 
     // flags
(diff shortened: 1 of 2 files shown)


src/write.c
```diff
@@ -2500,7 +2500,7 @@ static avifResult avifEncoderWriteMiniBox(avifEncoder * encoder, avifRWStream *
 
     if (floatFlag) {
         // bit(2) bit_depth_log2_minus4;
-        return AVIF_RESULT_NOT_IMPLEMENTED;
+        return AVIF_RESULT_INTERNAL_ERROR;
```
maryla-uc, line 2503: AVIF_ASSERT_OR_RETURN? (same below if you change it)

maryla-uc (approved):

PR #2407, opened 2024-08-22
Also check the type "skip" for a FreeSpaceBox

wantehchang:
The type of a FreeSpaceBox can be "free" or "skip".

A follow-up to https://github.com/AOMediaCodec/libavif/pull/2398.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -4153,7 +4153,8 @@ static avifResult avifParse(avifDecoder * decoder)
         }
 #endif
 
-        if (!isFtyp && (isNonSkippableVariableLengthBox || !memcmp(header.type, "free", 4) || !memcmp(header.type, "mdat", 4))) {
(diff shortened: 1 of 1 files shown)


y-guyon (approved):

src/read.c
```diff
@@ -4153,7 +4153,8 @@ static avifResult avifParse(avifDecoder * decoder)
         }
 #endif
 
-        if (!isFtyp && (isNonSkippableVariableLengthBox || !memcmp(header.type, "free", 4) || !memcmp(header.type, "mdat", 4))) {
+        if (!isFtyp && (isNonSkippableVariableLengthBox || !memcmp(header.type, "free", 4) || !memcmp(header.type, "skip", 4) ||
+                        !memcmp(header.type, "mdat", 4))) {
```
wantehchang, line 4157: For AVIF, it may be fine to require the FileTypeBox be the first box in the file. The only example given in ISO/IEC 14496-12 that may go before the FileTypeBox is a file signature, but AVIF doesn't have a file signature.

PR #2006, opened 2024-02-12
Merge avif_decode_fuzzer into avif_fuzztest_dec.cc

vrabaud:
This is the last non-fuzztest test to be migrated.
This removes the last old API test: cifuzz.yml is therefore not needed anymore.

PR #2410, opened 2024-08-23
Add a presubmit workflow that runs fuzztest tests

maryla-uc:


.github/workflows/ci-linux-shared-local-fuzztest.yml
```diff
@@ -0,0 +1,68 @@
...
+        -DAVIF_ENABLE_FUZZTEST=ON
+        -DAVIF_LOCAL_FUZZTEST=ON
+        -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
+        -DAVIF_ENABLE_WERROR=ON
+    - name: Build libavif (ninja)
+      working-directory: ./build
+      run: ninja
+    - name: Run AVIF Tests
+      working-directory: ./build
+      run: ctest -j $(getconf _NPROCESSORS_ONLN) --output-on-failure
```
y-guyon, line 68: Should this be limited to fuzz tests only?

y-guyon (commented):
I do not see any fuzz test in the presubmit of this PR besides `repro_avif_decode_fuzzer`. Does `ctest` work with FuzzTest targets?

Is there an overlap on what is supposed to be tested with .github/workflows/cifuzz.yml?

.github/workflows/ci-linux-shared-local-fuzztest.yml
maryla-uc: Done

maryla-uc:
I do see a bunch of tests starting with "avif_fuzztest_" in the output?

There is no overlap with .github/workflows/cifuzz.yml, this one is ossfuzz and runs only the fuzzer in tests/oss-fuzz (which this one does not run) (yes this is all very simple...)

vrabaud:
Indeed, because CIFuzz does not handle fuzztest (cf https://github.com/google/oss-fuzz/issues/11599), this CI is useful. We should remove CIFuzz altogether, this way, https://github.com/AOMediaCodec/libavif/pull/2006 could be merged too.

y-guyon (approved):
> I do see a bunch of tests starting with "avif_fuzztest_" in the output?

Indeed, GitHub UI + Chrome search failed me. The logs are probably loaded lazily.

.github/workflows/ci-linux-shared-local-fuzztest.yml
```diff
@@ -0,0 +1,71 @@
...
+  build-shared-local:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-24.04]
+        libyuv: [OFF]
+        include:
+          - runs-on: ubuntu-24.04
+            compiler: gcc
```
vrabaud, line 28: no need to specify this as clang is necessary for fuzztest (cf `CMAKE_C_COMPILER=clang`)

.github/workflows/ci-linux-shared-local-fuzztest.yml
```diff
@@ -0,0 +1,71 @@
...
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  build-shared-local:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-24.04]
+        libyuv: [OFF]
```
vrabaud, line 25: it's probably worth adding LOCAL too for libyuv

vrabaud (approved):

PR #2411, opened 2024-08-23
avifgainmaputil: ignore the alt image icc when --ignore-profile is set

maryla-uc:


apps/avifgainmaputil/imageio.cc
```diff
@@ -163,6 +163,11 @@ avifResult ReadAvif(avifDecoder* decoder, const std::string& input_filename,
   }
   if (ignore_profile) {
     avifRWDataFree(&decoder->image->icc);
+#if defined(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
```
wantehchang, line 166: Delete this. If this file is compiled, `AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP` must be defined.

maryla-uc: Ha yes good point!

y-guyon (approved):

Issue #2363, opened 2024-08-01
Fix the limitation of recent-cmake in the setup-common GitHub action


wantehchang:
The setup-common GitHub action installs the minimum required version of cmake (currently 3.13) by default. If the `recent-cmake` option is set to `'true'`, it installs a recent version of cmake (currently 3.18). If a new use case needs a newer cmake version, we need to bump the `recent-cmake` version, e.g., in https://github.com/AOMediaCodec/libavif/pull/2362. This is inconvenient.

There are two possible improvements:

1. Replace the `recent-cmake` option with a `cmake-version` option, which defaults to the minimum required version of cmake for libavif.
2. Replace the `recent-cmake` option with a `latest-cmake` option, which installs the latest version of cmake.

vrabaud:
For 1, you would still need to figure out a version that works with all the features you need for your test (e.g. C23, fuzztest ...) and this is not trivial. I believe having it done once and for all is easier.

For 2, we could end up with different CMake versions depending on the platform, and we might face CMake bugs in the latest releases.

Actually, we use `recent-cmake: true` everywhere where it matters except in `ci-android-emulator-tests.yml` ...

According to https://github.com/actions/runner-images?tab=readme-ov-file, CMake 3.30 is used on all images: we should probably use the system one by default, and have a test for the old CMake 3.13 that we support (though it should now be 3.16 if we follow https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md). Let me make a PR for that.

PR #2413, opened 2024-08-27
Replace recent-cmake action variable with oldest-cmake

vrabaud:
All tests actually use a recent CMake.
A new test in ci-unix-shared-installed.yml is added to test for the oldest supported CMake (3.13).
This fixes https://github.com/AOMediaCodec/libavif/issues/2363

### DIFF
--- a/.github/actions/setup-common/action.yml
+++ b/.github/actions/setup-common/action.yml
@@ -10,7 +10,7 @@ inputs:
   codec-rav1e:
     description: 'Can take the values: OFF, LOCAL, SYSTEM'
     default: 'OFF'
-  recent-cmake:
+  oldest-cmake:
     description: 'Can take the values: true, false. Only useful on Linux'
     default: 'false'
 runs:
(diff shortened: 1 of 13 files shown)


.github/actions/setup-common/action.yml
```diff
@@ -10,7 +10,7 @@ inputs:
   codec-rav1e:
     description: 'Can take the values: OFF, LOCAL, SYSTEM'
     default: 'OFF'
-  recent-cmake:
+  oldest-cmake:
```
wantehchang, line 13: Optional: It would be good to document that by "oldest" we mean the minimum required version (i.e., the version specified in the `cmake_minimum_required()` call in CMakeLists.txt).

.github/workflows/ci-unix-shared-installed.yml
```diff
@@ -14,15 +14,19 @@ concurrency:
...
         os: [ubuntu-24.04, macos-latest]
+        oldest-cmake: [false, true]
         include:
-          - runs-on: ubuntu-24.04
+          - os: ubuntu-24.04
             compiler: gcc
             gcc: 14
+        exclude:
+          - os: macos-latest
+            oldest-cmake: true
```
wantehchang, line 28: Optional: I can't quite figure out what the `include` and `exclude` do. A comment for each of them may be useful.

.github/workflows/ci-unix-shared-installed.yml
```diff
@@ -84,6 +87,8 @@ jobs:
 
         cmake . -DCMAKE_PREFIX_PATH=../install
     - name: Prepare libavif with [[nodiscard]] (cmake)
+      # CMake 3.21 is needed to force C23 for [[nodiscard]].
+      if: runner.oldest-cmake == 'false'
```
wantehchang, line 91: Note: This is a comment, not a request for change.

Here is an example where a `cmake-version` option would be better. We may be able to check if `cmake-version` is >= 3.21. The current check of `runner.oldest-cmake == 'false'` will become incorrect if our minimum required version is bumped to 3.21.

wantehchang (approved):
LGTM. Thanks!

.github/actions/setup-common/action.yml
vrabaud: done

.github/workflows/ci-unix-shared-installed.yml
vrabaud: Rules are indeed tricky: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow . Done

.github/workflows/ci-unix-shared-installed.yml
vrabaud: Indeed, I just removed one CMake version in `.github/actions/setup-common/action.yml`. Imposing a version in the condition means we need to install two CMake for the same job and dynamically choose one or the other. Or we split the job. I believe I  chose the least confusing solution that is easy for future maintenance: when we bump the minimum CMake version, we might delete oldest-cmake altogether (but as you mention, we will indeed to check all usages of `oldest-cmake` in the scripts).

PR #2078, opened 2024-03-25
Allow YCgCoRe and YCgCoRo decoding.

vrabaud:
Now that the https://www.itu.int/rec/T-REC-H.273 update is to be published, we can allow for decoding by default.
Encoding will come once decoding is sufficiently deployed.

This is the first phase of
https://github.com/AOMediaCodec/libavif/issues/2077

PR #2418, opened 2024-08-29
Clean YCgCo-R a bit.

vrabaud:
- update references
- add tests

This is to simplify https://github.com/AOMediaCodec/libavif/pull/2078.

It is possible the P in the URL is for `pre-published`: we might need to update it once fully published.

### DIFF
--- a/tests/gtest/avifrgbtoyuvtest.cc
+++ b/tests/gtest/avifrgbtoyuvtest.cc
@@ -366,11 +366,10 @@ TEST(RGBToYUVTest, AllMatrixCoefficients) {
                      AVIF_MATRIX_COEFFICIENTS_SMPTE240,
                      AVIF_MATRIX_COEFFICIENTS_YCGCO,
                      AVIF_MATRIX_COEFFICIENTS_BT2020_NCL,
-                     AVIF_MATRIX_COEFFICIENTS_CHROMA_DERIVED_NCL
+                     AVIF_MATRIX_COEFFICIENTS_CHROMA_DERIVED_NCL,
 #if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
-                     ,
                      AVIF_MATRIX_COEFFICIENTS_YCGCO_RE,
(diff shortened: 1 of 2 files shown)


tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -358,26 +358,46 @@ TEST(RGBToYUVTest, AllMatrixCoefficients) {
...
+                     AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED,
+                     AVIF_MATRIX_COEFFICIENTS_FCC,
+                     AVIF_MATRIX_COEFFICIENTS_BT470BG,
+                     AVIF_MATRIX_COEFFICIENTS_BT601,
+                     AVIF_MATRIX_COEFFICIENTS_SMPTE240,
+                     AVIF_MATRIX_COEFFICIENTS_YCGCO,
+                     AVIF_MATRIX_COEFFICIENTS_BT2020_NCL,
+                     AVIF_MATRIX_COEFFICIENTS_CHROMA_DERIVED_NCL
+#if defined(AVIF_ENABLE_EXPERIMENTAL_YCGCO_R)
+                     ,
```
wantehchang, line 371: Move this comma (,) to the previous line, if my suggestion below works.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -358,26 +358,46 @@ TEST(RGBToYUVTest, AllMatrixCoefficients) {
...
+                     AVIF_MATRIX_COEFFICIENTS_YCGCO_RE,
+                     AVIF_MATRIX_COEFFICIENTS_YCGCO_RO
```
wantehchang, line 373: Add a comma (,) to the end of this line. I think it is allowed in both C++ and C now. Does it not work?

wantehchang (approved):

tests/gtest/avifrgbtoyuvtest.cc
vrabaud: it works indeed. Done

PR #2419, opened 2024-08-29
Update the H273 URL to something that exists.

vrabaud:
I believe the `S` is for `superseded`.

### DIFF
--- a/src/colr.c
+++ b/src/colr.c
@@ -108,7 +108,7 @@ struct avifMatrixCoefficientsTable
     const float kb;
 };
 
-// https://www.itu.int/rec/T-REC-H.273-201612-I/en
+// https://www.itu.int/rec/T-REC-H.273-201612-S
 static const struct avifMatrixCoefficientsTable matrixCoefficientsTables[] = {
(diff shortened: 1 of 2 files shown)


wantehchang (approved):
LGTM. Thanks.

If the formulas are not renumbered in new versions, ww can also point to the top-level URL https://www.itu.int/rec/T-REC-H.273, which should be stable. But I don't know if these standards aim to keep the section numbers and formula numbers stable.

src/colr.c
```diff
@@ -108,7 +108,7 @@ struct avifMatrixCoefficientsTable
     const float kb;
 };
 
-// https://www.itu.int/rec/T-REC-H.273-201612-I/en
+// https://www.itu.int/rec/T-REC-H.273-201612-S
```
wantehchang, line 111: I am curious: in France is this URL redirected to https://www.itu.int/rec/T-REC-H.273-201612-S/fr?

vrabaud: yes it does (from my personal account)

vrabaud:
I did not want to check the most recent version (especially because there can be more in the future ...)

PR #2422, opened 2024-08-30
Add support for all profiles of SVT-AV1

FrankGalligan:
SVT-AV1 profiles (speed) can be set from 0 to 13, while libaom and Rav1e speed settings range is 0-10. This CL adds support to set the speed paramter up to 13.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -113,7 +113,10 @@ typedef int avifBool;
 
 #define AVIF_SPEED_DEFAULT -1
 #define AVIF_SPEED_SLOWEST 0
-#define AVIF_SPEED_FASTEST 10
+
+// Libaom and Rav1e fastest speed setting is 10, while SVT-AV1 fastest profile is 13.
+#define AVIF_SPEED_LIBAOM_RAV1E_FASTEST 10
+#define AVIF_SPEED_FASTEST 13
 
 // This value is used to indicate that an animated AVIF file has to be repeated infinitely.
 #define AVIF_REPETITION_COUNT_INFINITE -1
--- a/src/codec_svt.c
+++ b/src/codec_svt.c
@@ -160,7 +160,8 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         }
         if (encoder->speed != AVIF_SPEED_DEFAULT) {
 #if SVT_AV1_CHECK_VERSION(0, 9, 0)
-            svt_config->enc_mode = (int8_t)encoder->speed;
+            int speed = AVIF_CLAMP(encoder->speed, 0, AVIF_SPEED_FASTEST);
+            svt_config->enc_mode = (int8_t)speed;
 #else
             int speed = AVIF_CLAMP(encoder->speed, 0, 8);
             svt_config->enc_mode = (int8_t)speed;
--- a/tests/gtest/avif16bittest.cc
+++ b/tests/gtest/avif16bittest.cc
@@ -42,7 +42,7 @@ TEST_P(SampleTransformTest, Avif16bit) {
 
   EncoderPtr encoder(avifEncoderCreate());
   ASSERT_NE(encoder, nullptr);
-  encoder->speed = AVIF_SPEED_FASTEST;
+  encoder->speed = AVIF_SPEED_LIBAOM_RAV1E_FASTEST;
   encoder->quality = quality;
   encoder->qualityAlpha = quality;
   encoder->sampleTransformRecipe = recipe;
--- a/tests/gtest/avif_fuzztest_helpers.h
+++ b/tests/gtest/avif_fuzztest_helpers.h
@@ -187,7 +187,7 @@ inline auto ArbitraryAvifEncoder() {
   const auto tile_cols_log2 = fuzztest::InRange(0, 6);
   // Fuzz only a small range of 'speed' values to avoid slowing down the fuzzer
   // too much. The main goal is to fuzz libavif, not the underlying AV1 encoder.
-  const auto speed = fuzztest::InRange(6, AVIF_SPEED_FASTEST);
+  const auto speed = fuzztest::InRange(6, AVIF_SPEED_LIBAOM_RAV1E_FASTEST);
   return fuzztest::Map(CreateAvifEncoder, codec_choice, max_threads,
                        min_quantizer, max_quantizer, min_quantizer_alpha,
                        max_quantizer_alpha, tile_rows_log2, tile_cols_log2,
--- a/tests/gtest/avifallocationtest.cc
+++ b/tests/gtest/avifallocationtest.cc
@@ -137,7 +137,7 @@ void TestEncoding(uint32_t width, uint32_t height, uint32_t depth,
   // Try to encode.
   EncoderPtr encoder(avifEncoderCreate());
   ASSERT_NE(encoder, nullptr);
-  encoder->speed = AVIF_SPEED_FASTEST;
+  encoder->speed = AVIF_SPEED_LIBAOM_RAV1E_FASTEST;
   testutil::AvifRwData encoded_avif;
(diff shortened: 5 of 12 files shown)


y-guyon (commented):
I did not check thoroughly but this change looks good overall.

This does not look like a widespread use case so I wonder if silently allowing up to 13 in the libavif API when SVT-AV1 is selected without changing the value of `AVIF_SPEED_FASTEST` is safer.

An alternative approach would be to allow up to 13 for all codecs but to make 11,12,13 the same as 10 for libaom and rav1e. This avoids the need for a separate `AVIF_SPEED_LIBAOM_RAV1E_FASTEST` constant. It should just be written in a comment somewhere.

Another alternative approach could be a custom `key` passed to `avifEncoderSetCodecSpecificOption()` to avoid changing the current API or constant values. I do not expect speeds 11,12,13 to be widely used anyway.

FrankGalligan:
So you can actually encode in 11-13 with SVT-AV1 in the current code. You can pass `-a preset=13` to avifenc, but I think the output would confuse people:
`Encoding with codec 'svt' speed [6], color quality [60 (Medium)], alpha quality [100 (Lossless)], tileRowsLog2 [0], tileColsLog2 [0], 128 worker thread(s), please wait...`

I think I like your first alternative approach.

apps/avifenc.c
```diff
@@ -202,6 +202,7 @@ static void syntaxLong(void)
     printf("    -s,--speed S                      : Encoder speed (%d-%d, slowest-fastest, 'default' or 'd' for codec internal defaults. default speed: 6)\n",
            AVIF_SPEED_SLOWEST,
            AVIF_SPEED_FASTEST);
+    printf("                                        libaom & rav1e range 0-10\n");
```
y-guyon, line 205: libaom range seems to be 0-9:

https://github.com/AOMediaCodec/libavif/blob/db72bb1313a637f9c3aeaf56e3bac43f22254d2d/src/codec_aom.c#L589

src/codec_svt.c
```diff
@@ -160,7 +160,8 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         }
         if (encoder->speed != AVIF_SPEED_DEFAULT) {
 #if SVT_AV1_CHECK_VERSION(0, 9, 0)
-            svt_config->enc_mode = (int8_t)encoder->speed;
+            int speed = AVIF_CLAMP(encoder->speed, 0, AVIF_SPEED_FASTEST);
+            svt_config->enc_mode = (int8_t)speed;
```
y-guyon, line 164: This could be moved outside of the `#if` now

include/avif/avif.h
```diff
@@ -113,7 +113,9 @@ typedef int avifBool;
 
 #define AVIF_SPEED_DEFAULT -1
 #define AVIF_SPEED_SLOWEST 0
-#define AVIF_SPEED_FASTEST 10
+
+// Libaom and Rav1e fastest speed setting is 10, while SVT-AV1 fastest profile is 13.
+#define AVIF_SPEED_FASTEST 13
```
y-guyon, line 118: This may be a breaking change for some users who relied on this constant's value to never change. It should at least be described in CHANGES.md.

y-guyon (commented):
> So you can actually encode in 11-13 with SVT-AV1 in the current code. You can pass -a preset=13 to avifenc, but I think the output would confuse people

What about (also or instead) fixing stdout for this use case then? Because this PR does not prevent people from using `-a preset=13` and have a confusing console output.

include/avif/avif.h
FrankGalligan: Done

src/codec_svt.c
FrankGalligan: Done

apps/avifenc.c
FrankGalligan: Done

FrankGalligan:
> > So you can actually encode in 11-13 with SVT-AV1 in the current code. You can pass -a preset=13 to avifenc, but I think the output would confuse people
> 
> What about (also or instead) fixing stdout for this use case then? Because this PR does not prevent people from using `-a preset=13` and have a confusing console output.

I'm not sure I want to do this as then we may need to establish precedence rules. E.g. if you specify preset and speed. For simplicity I think it is better this way.

y-guyon:
I am a bit uneasy about changing the value of a define in a public header.  
@wantehchang Is this common practice?

Vincent suggested just changing `avifenc` behavior, because the API currently allows for SVT speed 11+.

FrankGalligan:
So what about keeping `#define AVIF_SPEED_FASTEST 10` as is. Adding `#define AVIF_SPEED_FASTEST_SVT-AV1 13` and changing avifenc to clamp the speed to  `AVIF_SPEED_FASTEST_SVT-AV1`?

wantehchang:
@FrankGalligan @y-guyon I am sorry I forgot to answer Yannis's question. I will take a look tomorrow.

But very quickly, this is very likely a gray area. That is, changing the value of `AVIF_SPEED_FASTEST` (a macro defined in a public header) is an incompatible change strictky speaking, but under a practical definition of backward compatibility, I suspect it won't break any application. But I need to think about this more.

wantehchang:
I have two proposals that avoid changing `AVIF_SPEED_FASTEST`. I recommend the first proposal.

Proposal 1: Use `avifEncoder::codecChoice` to select which speed scale to use.

- `AVIF_CODEC_CHOICE_AUTO` (the default) selects the current libavif speed scale of 0..10. `AVIF_SPEED_SLOWEST` and `AVIF_SPEED_FASTEST` only apply in this mode.
- Any other `AVIF_CODEC_CHOICE_*` value selects the original speed scale of the explicitly chosen codec. This is a change of behavior.
- In the `avifenc` command line interface, the `-c / --codec` option selects which speed scale to use.

I think this proposal is easy to understand even though the link between `codecChoice` and `speed` may not be obvious at first. It does not require adding a new `avifEncoder` option, so I prefer this proposal.

Proposal 2: Add an `avifBool useCodecSpeed` option to `avifEncoder`.

- If `useCodecSpeed` is false (the default), use the current libavif speed scale of 0..10. `AVIF_SPEED_SLOWEST` and `AVIF_SPEED_FASTEST` only apply in this mode. This is the current behavior.
- If `useCodecSpeed` is true, use the original speed scale of the codec in use. This is new behavior.
- In the `avifenc` command line interface, `useCodecSpeed` is indicated by adding the `codec:` prefix to the value for the `-s / --speed` option. For example, `-s 6` means speed 6 in the libavif 0..10 speed scale, and `-s codec:6` means speed 6 in the original speed scale of the codec in use.

This proposal is fully backward compatible, but requires adding a new `avifEncoder` option.

FrankGalligan:
Proposal 2 seems too complicated and do not think that is good to implement.

So for proposal 1 to get SVT-AV1 to encode with speed 13 you would need to set this command:
`avifenc -c svt -s 13 temp.png temp.avif`

And if you used this command, even with svt the only encoder:
`avifenc -s 13 temp.png temp.avif`

You would get speed 10 instead of 13.


How about proposal 3
* Keep `AVIF_SPEED_FASTEST` as is. I.e. 10
* Change `codec_svt.c` to clamp from 0 to 13. 
* Change avifenc.c to remove the check for `if (settings.speed > AVIF_SPEED_FASTEST) { settings.speed = AVIF_SPEED_FASTEST; }` 
* * The library specific clamps will still happen in their wrapper code. 
* No other changes.

 With proposal 3 you will get speed 13 with both of these commands:
`avifenc -c svt -s 13 temp.png temp.avif`

And if you used this command, even with svt the only encoder:
`avifenc -s 13 temp.png temp.avif`

FrankGalligan:
I also changed the avifenc help text to output the encoder speed range as 0-13

Issue #2423, opened 2024-08-31
Can not cross-compile for Windows


vtorri:
libavif 1.1.1, libyuv, libaom and dav1d are installed

cmake + make invokation:

```
rm -rf builddir && mkdir builddir && cd builddir

cmake \
    -DCMAKE_TOOLCHAIN_FILE=../cross_toolchain.txt \
    -DCMAKE_INSTALL_PREFIX=/opt/libavif \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=TRUE \
    -DAVIF_CODEC_AOM=SYSTEM \
    -DAVIF_CODEC_DAV1D=SYSTEM \
    -DAVIF_BUILD_APPS:BOOL=OFF \
    -G "Unix Makefiles" \
    .. > ../../config.log 2>&1

make -j  install > ../../make.log 2>&1
```

cmake output:

```
-- The C compiler identification is GNU 12.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/x86_64-w64-mingw32-gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.8.0") 
-- Checking for module 'libyuv'
--   Found libyuv, version 1840
-- Found libyuv: /home/vtorri/ewpi_64/bin/libyuv.dll (found version "1840") 
-- libavif: libyuv (1895) found; libyuv-based fast paths enabled.
-- libavif: Enabling warnings for GCC
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Checking for module 'dav1d'
--   Found dav1d, version 1.4.3
-- Found dav1d: /home/vtorri/ewpi_64/bin/libdav1d.dll (found version "1.4.3") 
-- libavif: Codec enabled: dav1d (decode)
-- Checking for module 'aom'
--   Found aom, version 3.9.1
-- Found aom: /home/vtorri/ewpi_64/bin/libaom.dll (found version "3.9.1") 
-- libavif: Codec enabled: aom (encode/decode)
-- The CXX compiler identification is GNU 12.2.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
[...]

PR #2425, opened 2024-09-01
Add IMPORTED_IMPLIB property to each CMake Module

y-guyon:
Apply cmake-format.

Fixes https://github.com/AOMediaCodec/libavif/issues/2423.


cmake/Modules/Findaom.cmake
```diff
@@ -51,6 +51,6 @@ if(AOM_LIBRARY)
     else()
         add_library(aom SHARED IMPORTED GLOBAL)
     endif()
-    set_target_properties(aom PROPERTIES IMPORTED_LOCATION "${AOM_LIBRARY}")
+    set_target_properties(aom PROPERTIES IMPORTED_LOCATION "${AOM_LIBRARY}" IMPORTED_IMPLIB "${AOM_LIBRARY}")
```
wantehchang, line 54: Yannis: Thank you for writing a fix. Note that this comment is just a comment, not a request for change.

According to CMake documentation, `IMPORTED_LOCATION` is the pathname of the DLL and `IMPORTED_IMPLIB` is the pathname of the import library. So the file names are different (.dll vs. .lib) and the files may be installed in different subdirectories (bin vs. lib, as @vtorri noted in https://github.com/AOMediaCodec/libavif/issues/2423#issuecomment-2323234955).

Perhaps the CMake documentation about this is incomplete and setting `IMPORTED_LOCATION` and `IMPORTED_IMPLIB` to the same value also works. Perhaps setting `IMPORTED_IMPLIB` to  the pathname of some file is better than not setting it. I don't know. I am just reporting what I found in the CMake documentation.

Here are the relevant excerpts:

https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED_LOCATION.html

> **IMPORTED_LOCATION**
> 
> Full path to the main file on disk for an IMPORTED target.
> 
> Set this to the location of an IMPORTED target file on disk. [...] For DLLs this is the location of the .dll part of the library. [...]

https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED_IMPLIB.html

> **IMPORTED_IMPLIB**
>
> Full path to the import library for an IMPORTED target.
> 
> This property may be set:
> 
> - On DLL platforms, to the location of the .lib part of the DLL.

https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html#importing-libraries

> On Windows, a .dll and its .lib import library may be imported together:
> 
> ```
> add_library(bar SHARED IMPORTED)
> set_property(TARGET bar PROPERTY
>              IMPORTED_LOCATION "c:/path/to/bar.dll")
> set_property(TARGET bar PROPERTY
>              IMPORTED_IMPLIB "c:/path/to/bar.lib")
> add_executable(myexe src1.c src2.c)
> target_link_libraries(myexe PRIVATE bar)
> ```

wantehchang: I found a comment in a CMake issue that apparently matches what @vtorri reported in https://github.com/AOMediaCodec/libavif/issues/2423#issuecomment-2323236316 and explains why this pull request works:

https://gitlab.kitware.com/cmake/cmake/-/issues/22446#note_1444129

> On a related note, MinGW can directly link with DLLs, without the need for an import library. It would be great if, when building with a MinGW toolchain, `IMPORTED_IMPLIB` were optional and `IMPORTED_LOCATION` were used if it is absent. (As a workaround, specifying the path to the DLL as `IMPORTED_IMPLIB` seems to work, but only by accident.)

vrabaud (approved):

cmake/Modules/Findaom.cmake
y-guyon: Thank you for investigating.

PR #2426, opened 2024-09-10
Allow odd clap dimensions and offsets

y-guyon:
The constraint in MIAF was replaced by an upsampling step before cropping.

Notes:
- `avifdec` does not crop anything when decoding.
- the libavif API only outputs YUV samples along with a `clap` field in `avifImage`. It is the user's duty to upsample, hence the added output argument in `avifCropRectFromCleanApertureBox()`.

See https://github.com/AOMediaCodec/av1-avif/pull/202#issuecomment-2339926253.

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -347,6 +347,24 @@ int main(int argc, char * argv[])
     printf("Image details:\n");
     avifImageDump(decoder->image, 0, 0, decoder->progressiveState);
 
+    if (decoder->image->transformFlags & AVIF_TRANSFORM_CLAP) {
+        avifCropRect cropRect;
+        if (avifCropRectFromCleanApertureBox(&cropRect,
+                                             &decoder->image->clap,
+                                             decoder->image->width,
+                                             decoder->image->height,
+                                             &decoder->diag)) {
+            if (cropRect.x != 0 || cropRect.y != 0 || cropRect.width != decoder->image->width ||
+                cropRect.height != decoder->image->height) {
+                // TODO: Implement, see https://github.com/AOMediaCodec/libavif/issues/2427
+                fprintf(stderr, "Warning: Clean Aperture values were ignored, the output image was NOT cropped\n");
+            }
+        } else {
+            // Should happen only if AVIF_STRICT_CLAP_VALID is disabled.
+            fprintf(stderr, "Warning: Invalid Clean Aperture values\n");
+        }
+    }
+
     if (ignoreICC && (decoder->image->icc.size > 0)) {
         printf("[--ignore-icc] Discarding ICC profile.\n");
         // This cannot fail.
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -349,17 +349,11 @@ int main(int argc, char * argv[])
 
     if (decoder->image->transformFlags & AVIF_TRANSFORM_CLAP) {
         avifCropRect cropRect;
-        if (avifCropRectFromCleanApertureBox(&cropRect,
-                                             &decoder->image->clap,
-                                             decoder->image->width,
-                                             decoder->image->height,
-                                             &decoder->diag)) {
-            if (cropRect.x != 0 || cropRect.y != 0 || cropRect.width != decoder->image->width ||
-                cropRect.height != decoder->image->height) {
-                // TODO: Implement, see https://github.com/AOMediaCodec/libavif/issues/2427
-                fprintf(stderr, "Warning: Clean Aperture values were ignored, the output image was NOT cropped\n");
-            }
-        } else {
+        if (!avifCropRectFromCleanApertureBox(&cropRect,
+                                              &decoder->image->clap,
+                                              decoder->image->width,
+                                              decoder->image->height,
+                                              &decoder->diag)) {
             // Should happen only if AVIF_STRICT_CLAP_VALID is disabled.
             fprintf(stderr, "Warning: Invalid Clean Aperture values\n");
         }
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -2316,9 +2316,10 @@ int main(int argc, char * argv[])
 
         // Validate clap
         avifCropRect cropRect;
+        avifBool upsampleBeforeCropping;
         avifDiagnostics diag;
         avifDiagnosticsClearError(&diag);
-        if (!avifCropRectConvertCleanApertureBox(&cropRect, &image->clap, image->width, image->height, image->yuvFormat, &diag)) {
+        if (!avifCropRectFromCleanApertureBox(&cropRect, &upsampleBeforeCropping, &image->clap, image->width, image->height, image->yuvFormat, &diag)) {
             fprintf(stderr,
                     "ERROR: Invalid clap: width:[%d / %d], height:[%d / %d], horizOff:[%d / %d], vertOff:[%d / %d] - %s\n",
                     (int32_t)image->clap.widthN,
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -388,7 +388,7 @@ static avifBool strpre(const char * str, const char * prefix)
     return strncmp(str, prefix, strlen(prefix)) == 0;
 }
 
-static avifBool convertCropToClap(uint32_t srcW, uint32_t srcH, avifPixelFormat yuvFormat, uint32_t clapValues[8])
+static avifBool convertCropToClap(uint32_t srcW, uint32_t srcH, uint32_t clapValues[8])
 {
     avifCleanApertureBox clap;
     avifCropRect cropRect;
@@ -399,13 +399,12 @@ static avifBool convertCropToClap(uint32_t srcW, uint32_t srcH, avifPixelFormat 
 
     avifDiagnostics diag;
     avifDiagnosticsClearError(&diag);
-    avifBool convertResult = avifCleanApertureBoxConvertCropRect(&clap, &cropRect, srcW, srcH, yuvFormat, &diag);
+    avifBool convertResult = avifCleanApertureBoxFromCropRect(&clap, &cropRect, srcW, srcH, &diag);
     if (!convertResult) {
         fprintf(stderr,
-                "ERROR: Impossible crop rect: imageSize:[%ux%u], pixelFormat:%s, cropRect:[%u,%u, %ux%u] - %s\n",
+                "ERROR: Impossible crop rect: imageSize:[%ux%u], cropRect:[%u,%u, %ux%u] - %s\n",
                 srcW,
                 srcH,
-                avifPixelFormatToString(yuvFormat),
                 cropRect.x,
                 cropRect.y,
                 cropRect.width,
@@ -2298,7 +2297,7 @@ int main(int argc, char * argv[])
         image->pasp.vSpacing = settings.paspValues[1];
     }
     if (cropConversionRequired) {
-        if (!convertCropToClap(image->width, image->height, image->yuvFormat, settings.clapValues)) {
+        if (!convertCropToClap(image->width, image->height, settings.clapValues)) {
             goto cleanup;
         }
         settings.clapValid = AVIF_TRUE;
@@ -2332,7 +2331,6 @@ int main(int argc, char * argv[])
                     diag.error);
             goto cleanup;
         }
-        // avifCropRectRequiresUpsampling() does not need to be called for clap validation.
     }
     if (irotAngle != 0xff) {
         image->transformFlags |= AVIF_TRANSFORM_IROT;
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -2316,10 +2316,9 @@ int main(int argc, char * argv[])
 
         // Validate clap
         avifCropRect cropRect;
-        avifBool upsampleBeforeCropping;
         avifDiagnostics diag;
         avifDiagnosticsClearError(&diag);
-        if (!avifCropRectFromCleanApertureBox(&cropRect, &upsampleBeforeCropping, &image->clap, image->width, image->height, image->yuvFormat, &diag)) {
+        if (!avifCropRectFromCleanApertureBox(&cropRect, &image->clap, image->width, image->height, &diag)) {
             fprintf(stderr,
                     "ERROR: Invalid clap: width:[%d / %d], height:[%d / %d], horizOff:[%d / %d], vertOff:[%d / %d] - %s\n",
                     (int32_t)image->clap.widthN,
@@ -2333,6 +2332,7 @@ int main(int argc, char * argv[])
                     diag.error);
             goto cleanup;
         }
+        // avifCropRectRequiresUpsampling() does not need to be called for clap validation.
     }
     if (irotAngle != 0xff) {
         image->transformFlags |= AVIF_TRANSFORM_IROT;
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -557,7 +557,8 @@ AVIF_NODISCARD AVIF_API avifBool avifCleanApertureBoxFromCropRect(avifCleanApert
                                                                   uint32_t imageH,
                                                                   avifDiagnostics * diag);
 // If this function returns true, the image must be upsampled from 4:2:0 or 4:2:2 to 4:4:4 before
-// Clean Aperture values are applied.
+// Clean Aperture values are applied. This can be done by converting the avifImage to RGB using
+// avifImageYUVToRGB() and only using the cropRect region of the avifRGBImage.
 AVIF_NODISCARD AVIF_API avifBool avifCropRectRequiresUpsampling(const avifCropRect * cropRect, avifPixelFormat yuvFormat);
 
 // Deprecated. Use avifCropRectFromCleanApertureBox() instead.
@@ -1309,6 +1310,10 @@ typedef struct avifDecoder
     // legal to call avifImageYUVToRGB() on this in between calls to avifDecoderNextImage(), but use
     // avifImageCopy() if you want to make a complete, permanent copy of this image's YUV content or
     // metadata.
+    //
+    // For each field among clap, irot and imir, if the corresponding avifTransformFlag is set, the
+    // transform must be applied before rendering or converting the image, or forwarded along as
+    // attached metadata.
     avifImage * image;
 
     // Counts and timing for the current image in an image sequence. Uninteresting for single image files.
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -544,14 +544,10 @@ typedef struct avifCropRect
 
 // These will return AVIF_FALSE if the resultant values violate any standards, and if so, the output
 // values are not guaranteed to be complete or correct and should not be used.
-// If upsampleBeforeCropping is true, the image must be upsampled from 4:2:0 or 4:2:2 to 4:4:4 before
-// Clean Aperture values are applied.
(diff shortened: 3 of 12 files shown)


wantehchang:
> @wantehchang Do you plan on submitting #1749 first?

Done. The reason I had not submitted https://github.com/AOMediaCodec/libavif/pull/1749 is that I am not sure whether only some of the members of the avifCleanApertureBox struct will be changed to int32_t in the next version of ISO BMFF. But it is fine to submit https://github.com/AOMediaCodec/libavif/pull/1749 first because it describes how libavif interprets the members of the avifCleanApertureBox struct now. We can update the comment when the next version of ISO BMFF is released.

src/avif.c
```diff
@@ -722,6 +699,16 @@ avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
...
 
+    // ISO/IEC 23000-22:2024, Section 7.3.6.7:
+    // The clean aperture property is restricted according to the chroma sampling format of the input image
+    // (4:4:4, 4:2:2, 4:2:0, or 4:0:0) as follows:
+    // - cleanApertureWidth and cleanApertureHeight shall be integers;
+    // - If any of the following conditions hold true, the image is first implicitly upsampled to 4:4:4:
+    //   - chroma is subsampled horizontally (i.e., 4:2:2 and 4:2:0) and cleanApertureWidth is odd
+    //   - chroma is subsampled horizontally (i.e., 4:2:2 and 4:2:0) and left-most pixel is on an odd position
+    //   - chroma is subsampled vertically (i.e., 4:2:0) and cleanApertureHeight is odd
+    //   - chroma is subsampled vertically (i.e., 4:2:0) and topmost line is on an odd position
```
wantehchang, line 710: For AVIF, the first and third conditions are too strict. They are equivalent to disallowing odd widths for 4:2:2 and 4:2:0 and disallowing odd heights for 4:2:0, but AV1 supports those two cases. In libavif we can consider ignoring the first and third conditions.

src/avif.c
```diff
@@ -820,9 +827,10 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
                                              avifPixelFormat yuvFormat,
                                              avifDiagnostics * diag)
 {
+    (void)yuvFormat;
```
wantehchang, line 830: If we want to preserve the old behavior, we cannot ignore `yuvFormat`.

It would be useful to move lines 804-806 to a new function that takes `const avifCropRect * cropRect` and `avifPixelFormat yuvFormat` as input and returns the `upsampleBeforeCropping` boolean. Then we can call that new function here (to preserve the old behavior) and at line 804.

src/avif.c
```diff
@@ -810,7 +797,27 @@ avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
...
+}
+
+avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
+                                             const avifCleanApertureBox * clap,
+                                             uint32_t imageW,
+                                             uint32_t imageH,
+                                             avifPixelFormat yuvFormat,
+                                             avifDiagnostics * diag)
+{
+    // Keep the same pre-deprecation behavior.
```
wantehchang, line 817: This does not keep the old behavior exactly. The difference is the first and third conditions at lines 706-710. So to be precise, the comment should say "most of the pre-deprecation behavior".

If we ignore the first and third conditions as I suggested in my comment at line 710, then we keep the old behavior exactly.

wantehchang (commented):
I need to think about this issue more to see if we should take a different approach. Here are some comments on the current approach in the pull request.

y-guyon:
> I need to think about this issue more to see if we should take a different approach in this pull request.

@wantehchang Did you find another approach?

src/avif.c
```diff
@@ -810,7 +797,27 @@ avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
...
     cropRect->width = (uint32_t)clapW;
     cropRect->height = (uint32_t)clapH;
-    return avifCropRectIsValid(cropRect, imageW, imageH, yuvFormat, diag);
+    if (!avifCropRectIsValid(cropRect, imageW, imageH, diag)) {
+        return AVIF_FALSE;
+    }
+
+    *upsampleBeforeCropping = ((yuvFormat == AVIF_PIXEL_FORMAT_YUV420 || yuvFormat == AVIF_PIXEL_FORMAT_YUV422) &&
+                               (cropRect->width % 2 || cropRect->x % 2)) ||
+                              (yuvFormat == AVIF_PIXEL_FORMAT_YUV420 && (cropRect->height % 2 || cropRect->y % 2));
```
wantehchang, line 806: An alternative design is to remove the `yuvFormat` input parameter and the `upsampleBeforeCropping` output parameter from this function (`avifCropRectFromCleanApertureBox`), and then add another function that takes `const avifCropRect * cropRect` and `yuvFormat` as input parameters, and returns `upsampleBeforeCropping` as the return value.

Note that `yuvFormat` and `upsampleBeforeCropping` are only used in this statement (lines 804-806), so this statement can be cleanly moved to a separate function.

Another alternative is for libavif to upsample to 4:4:4 and crop automatically when necessary so that libavif users don't need to worry about this. But I am not sure if this is a good idea.

y-guyon: > add another function that takes `const avifCropRect * cropRect` and `yuvFormat` as input parameters, and returns `upsampleBeforeCropping` as the return value

The issue with this new function and with the current `avifImage::clap` are that users can easily ignore them. The point of the "mandatory" `upsampleBeforeCropping` argument was to force users to at least consider upsampling (because there is no other function available to get a rect). But they can just ignore `clap` anyway so it has little effect.

> Another alternative is for libavif to upsample to 4:4:4 and crop automatically when necessary so that libavif users don't need to worry about this. But I am not sure if this is a good idea.

I would lean towards that as a default behavior, and why not add an "ignoreClap" setting in `avifDecoder` to keep the old behavior back.  
However existing files with `clap` will no longer display the same, in places where `clap` is ignored by users of libavif.

src/avif.c
```diff
@@ -722,6 +699,16 @@ avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
     //   positive or negative. For cleanApertureWidth and cleanApertureHeight,
     //   N shall be positive and D shall be strictly positive.
 
+    // ISO/IEC 23000-22:2024, Section 7.3.6.7:
```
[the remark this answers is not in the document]
y-guyon: Added "FDIS" in the comment.

I am fine with waiting till it is published, although I guess it would not hurt to merge this PR now. As you see fit.

src/avif.c
y-guyon: Done.

src/avif.c
y-guyon: Done.

src/avif.c
y-guyon: > An alternative design is to remove the `yuvFormat` input parameter and the `upsampleBeforeCropping` output parameter from this function (`avifCropRectFromCleanApertureBox`), and then add another function that takes `const avifCropRect * cropRect` and `yuvFormat` as input parameters, and returns `upsampleBeforeCropping` as the return value.
> 
> Note that `yuvFormat` and `upsampleBeforeCropping` are only used in this statement (lines 804-806), so this statement can be cleanly moved to a separate function.

Done.

src/avif.c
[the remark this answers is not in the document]
y-guyon: It is now "Under publication". I wonder if it will be numbered as ISO/IEC 23000-22:2025 or just "2nd edition".

wantehchang:
Yannis: I will review this pull request tomorrow. Sorry about the delay.

apps/avifdec.c
```diff
@@ -347,6 +347,24 @@ int main(int argc, char * argv[])
...
+        avifCropRect cropRect;
+        if (avifCropRectFromCleanApertureBox(&cropRect,
+                                             &decoder->image->clap,
+                                             decoder->image->width,
+                                             decoder->image->height,
+                                             &decoder->diag)) {
+            if (cropRect.x != 0 || cropRect.y != 0 || cropRect.width != decoder->image->width ||
+                cropRect.height != decoder->image->height) {
+                // TODO: Implement, see https://github.com/AOMediaCodec/libavif/issues/2427
+                fprintf(stderr, "Warning: Clean Aperture values were ignored, the output image was NOT cropped\n");
```
wantehchang, line 360: I believe that avifdec.c also ignores the imir and irot properties. Should we also emit warning messages about imir and irot?

apps/avifenc.c
```diff
@@ -2305,6 +2305,7 @@ int main(int argc, char * argv[])
                     diag.error);
             goto cleanup;
         }
+        // avifCropRectRequiresUpsampling() does not need to be called for clap validation.
```
wantehchang, line 2308: Optional: This comment could be omitted. `avifCropRectRequiresUpsampling()` is a function intended for the decoder, so it seems irrelevant in avifenc.c.

include/avif/avif.h
```diff
@@ -484,6 +484,8 @@ typedef struct avifPixelAspectRatioBox
 typedef struct avifCleanApertureBox
 {
     // 'clap' from ISO/IEC 14496-12:2022 12.1.4.3
+    // Note that ISO/IEC 23000-22:2024 7.3.6.7 requires the decoded image to be upsampled to 4:4:4 before
+    // clean aperture is applied if a clean aperture size or offset is odd in a subsampled dimension.
```
wantehchang, line 488: IMPORTANT: AVIF can subsample an odd width or height. This comment should clarify that we relax this requirement for AVIF, similar to the comment at src/avif.c:885:

```
    // AV1 supports odd dimensions with chroma subsampling in those directions, so only look for x and y.
```

src/avif.c
```diff
@@ -898,7 +870,46 @@ avifBool avifCropRectConvertCleanApertureBox(avifCropRect * cropRect,
...
+        //   - If chroma is subsampled vertically (i.e., 4:2:0),
+        //     the topmost line of the clean aperture shall be even numbers.
+
+        if (avifCropRectRequiresUpsampling(cropRect, yuvFormat)) {
+            avifDiagnosticsPrintf(diag, "[Strict] crop rect X and Y offsets must be even due to this image's YUV subsampling");
+            return AVIF_FALSE;
+        }
+        return AVIF_TRUE;
+    }
+    return AVIF_FALSE;
```
wantehchang, line 912: Nit: Use early return for both checks:

```
    if (!avifCropRectFromCleanApertureBox(cropRect, clap, imageW, imageH, diag)) {
        return AVIF_FALSE;
    }
    // Keep the same pre-deprecation behavior.

    // ISO/IEC 23000-22:2019/Amd. 2:2021, Section 7.3.6.7:
    //   - If chroma is subsampled horizontally (i.e., 4:2:2 and 4:2:0),
    //     the leftmost pixel of the clean aperture shall be even numbers;
    //   - If chroma is subsampled vertically (i.e., 4:2:0),
    //     the topmost line of the clean aperture shall be even numbers.

    if (avifCropRectRequiresUpsampling(cropRect, yuvFormat)) {
        avifDiagnosticsPrintf(diag, "[Strict] crop rect X and Y offsets must be even due to this image's YUV subsampling");
        return AVIF_FALSE;
    }
    return AVIF_TRUE;
```

src/avif.c
```diff
@@ -908,9 +919,10 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
                                              avifPixelFormat yuvFormat,
                                              avifDiagnostics * diag)
 {
+    (void)yuvFormat;
```
wantehchang, line 922: We probably should also deprecate this function and add a new version of this function, say `avifCleanApertureBoxFromCropRect`, that doesn't have the `yuvFormat` parameter.

tests/gtest/avifclaptest.cc
```diff
@@ -129,27 +131,46 @@ constexpr ValidClapPropertyParam kValidClapPropertyTestParams[] = {
      100,
      AVIF_PIXEL_FORMAT_YUV420,
      {99, 1, 99, 1, static_cast<uint32_t>(-1), 2, static_cast<uint32_t>(-1), 2},
-     {0, 0, 99, 99}},
+     {0, 0, 99, 99},
+     false},
+    // pcX = -1/2 + (100 - 1)/2 = 49
+    // pcY = -1/2 + (100 - 1)/2 = 49
+    // leftmost = 49 - (99 - 1)/2 = 0
+    // topmost = 49 - (99 - 1)/2 = 0
```
wantehchang, line 139: This comment is exactly the same as the comment for the previous test case (lines 126-129), so this comment must be incorrect.

I think this comment should read:

```
    // pcX = 1/2 + (100 - 1)/2 = 50
    // pcY = 1/2 + (100 - 1)/2 = 50
    // leftmost = 50 - (99 - 1)/2 = 1
    // topmost = 50 - (99 - 1)/2 = 1
```

tests/gtest/avifclaptest.cc
```diff
@@ -129,27 +131,46 @@ constexpr ValidClapPropertyParam kValidClapPropertyTestParams[] = {
...
       << diag.error;
+  const avifBool upsample_before_cropping =
+      avifCropRectRequiresUpsampling(&crop_rect, param.yuv_format);
   EXPECT_EQ(crop_rect.x, param.expected_crop_rect.x);
   EXPECT_EQ(crop_rect.y, param.expected_crop_rect.y);
   EXPECT_EQ(crop_rect.width, param.expected_crop_rect.width);
   EXPECT_EQ(crop_rect.height, param.expected_crop_rect.height);
+  EXPECT_EQ(upsample_before_cropping, param.expected_upsample_before_cropping);
+
+  // Deprecated function coverage.
```
wantehchang, line 169: This coverage is insufficient. If `avifCropRectConvertCleanApertureBox()` returns true, we should also check `crop_rect`. Here is one way to do that:

```
  const avifBool success =
    avifCropRectConvertCleanApertureBox(&crop_rect, &param.clap,
                                        param.width, param.height,
                                        param.yuv_format, &diag);
  EXPECT_EQ(success, !upsample_before_cropping);
  if (success) {
    EXPECT_EQ(crop_rect.x, param.expected_crop_rect.x);
    EXPECT_EQ(crop_rect.y, param.expected_crop_rect.y);
    EXPECT_EQ(crop_rect.width, param.expected_crop_rect.width);
    EXPECT_EQ(crop_rect.height, param.expected_crop_rect.height);
  }
```

apps/avifdec.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2564.

wantehchang (approved):
LGTM. Thanks!

include/avif/avif.h
```diff
@@ -542,18 +546,26 @@ typedef struct avifCropRect
...
+                                                                  uint32_t imageH,
+                                                                  avifDiagnostics * diag);
+AVIF_NODISCARD AVIF_API avifBool avifCleanApertureBoxFromCropRect(avifCleanApertureBox * clap,
+                                                                  const avifCropRect * cropRect,
+                                                                  uint32_t imageW,
+                                                                  uint32_t imageH,
+                                                                  avifDiagnostics * diag);
+// If this function returns true, the image must be upsampled from 4:2:0 or 4:2:2 to 4:4:4 before
+// Clean Aperture values are applied.
+AVIF_NODISCARD AVIF_API avifBool avifCropRectRequiresUpsampling(const avifCropRect * cropRect, avifPixelFormat yuvFormat);
```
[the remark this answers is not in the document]
y-guyon: > "requires upsampling" boolean as an output parameter (as you did before)

We went away from that approach already, and it does not cover cropping and rotation.

> upsample the image for the users

I began writing a PR for that and quickly gave up. It requires a lot of changes, especially because the whole implementation revolves around `avifDecoder::image`. Since it is part of the public API, it must stay there, and it must be the buffer to crop/rotate/upsample. However it is also used by the underlying codec, so it cannot be cropped/rotated/upsampled (at least till the last frame is decoded or the codec is destroyed).  
So some internal `avifDecoderData::image` must be introduced, and `avifDecoder->image` is either a copy or a view of it, depending on what is doable.  
Upsampling can be done in multiple ways and has to take into account chroma citing metadata. `avifDecoderDecodedRowCount()` must also be adapted to cropping/rotation.

A cheaper alternative could be a helper function that returns a new image object with all the necessary transforms applied, and users must call that. But it will not go well with incremental support and may introduce unnecessary buffer copies.

Yet another path is to add a public API `avifDecoder::imageWithTransformsApplied`. This is roughly the same amount of work as the internal `avifDecoderData::image` but with fewer risks of breaking something in the existing API.

Overall it felt like a lot of work compared to just writing in avif.h that users must take care of that themselves, which I agree is not great, but also already the current situation.

wantehchang (approved):

include/avif/avif.h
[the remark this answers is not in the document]
wantehchang: Thanks a lot for the analysis.

Do you think we should provide an `avifImageUpsample()` function? Or should we just document that one can also convert YUV to RGB and then crop the RGB image?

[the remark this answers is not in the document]
y-guyon: > Do you think we should provide an `avifImageUpsample()` function?

In Chrome, samples can be rendered directly without being converted to RGB first, right? If so, it would be useful at least for this kind of use case.  
Is orientation and cropping both applied in Chrome?

> Or should we just document that one can also convert YUV to RGB and then crop the RGB image?

That sounds useful to mention, even if libavif offers an alternative in the future. Done.

Also added another comment about applying transform next to `avifDecoder::image`.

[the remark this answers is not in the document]
wantehchang: Chrome's blink::AVIFImageDecoder can provide either YUV samples or RGB samples to its callers. Cropping is done by blink::AVIFImageDecoder. (It requires that the top left corner of the clean aperture be located at (0, 0).) Orientation is applied by the callers of blink::AVIFImageDecoder.

src/avif.c
```diff
@@ -965,6 +936,68 @@ avifBool avifCleanApertureBoxConvertCropRect(avifCleanApertureBox * clap,
...
+    // Keep the same pre-deprecation behavior.
+
+    // ISO/IEC 23000-22:2019/Amd. 2:2021, Section 7.3.6.7:
+    //   - If chroma is subsampled horizontally (i.e., 4:2:2 and 4:2:0),
+    //     the leftmost pixel of the clean aperture shall be even numbers;
+    //   - If chroma is subsampled vertically (i.e., 4:2:0),
+    //     the topmost line of the clean aperture shall be even numbers.
+
+    if (avifCropRectRequiresUpsampling(cropRect, yuvFormat)) {
+        avifDiagnosticsPrintf(diag, "[Strict] crop rect X and Y offsets must be even due to this image's YUV subsampling");
```
wantehchang, line 972: Should we add an `avifDiagnostics * diag` parameter to `avifCropRectRequiresUpsampling()` so that it can add the diagnostic message itself?

wantehchang (approved):
LGTM. The new comments look good. Thanks!

src/avif.c
wantehchang: Please ignore this question. `avifCropRectRequiresUpsampling()` can fail for only one reason. It doesn't need to return an error code or diagnostic message.

PR #2429, opened 2024-09-12
Add lsel (layer selection) property to progressive images.

maryla-uc:
My understanding from the discussions in https://github.com/AOMediaCodec/libavif/pull/939 and the AVIF specification is that libavif treats the absence of the 'lsel' property the same as 'lsel' being present with the value 0xFFFF, but this is done for backward compatibility only and encoders are supposed to add the 'lsel' property on layered images.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -2892,7 +2892,6 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
 
         if (item->extraLayerCount > 0) {
             // Layered Image Indexing Property
-
             avifItemPropertyDedupStart(dedup);
             avifBoxMarker a1lx;
             AVIF_CHECKRES(avifRWStreamWriteBox(&dedup->s, "a1lx", AVIF_BOX_SIZE_TBD, &a1lx));
@@ -2921,6 +2920,19 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
             }
             avifRWStreamFinishBox(&dedup->s, a1lx);
             AVIF_CHECKRES(avifItemPropertyDedupFinish(dedup, s, &item->ipma, AVIF_FALSE));
+
+            avifItemPropertyDedupStart(dedup);
+            avifBoxMarker lsel;
+            AVIF_CHECKRES(avifRWStreamWriteBox(&dedup->s, "lsel", AVIF_BOX_SIZE_TBD, &lsel));
+            // Layer Selection Property
+            // Section 2.3.1 of of AV1 Image File Format specification v1.1.0:
+            // The value 0xFFFF is reserved for a special meaning. If a lsel property is associated with an AV1
+            // Image Item but its layer_id value is set to 0xFFFF, the renderer is free to render either only
+            // the output image of the highest spatial layer, or to render all output images of all the intermediate
+            // layers and the highest spatial layer, resulting in a form of progressive decoding.
+            AVIF_CHECKRES(avifRWStreamWriteU16(&dedup->s, 0xFFFF));
+            avifRWStreamFinishBox(&dedup->s, lsel);
+            AVIF_CHECKRES(avifItemPropertyDedupFinish(dedup, s, &item->ipma, AVIF_TRUE));
         }
     }
     return AVIF_RESULT_OK;
--- a/src/write.c
+++ b/src/write.c
@@ -2925,7 +2925,7 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
             avifBoxMarker lsel;
             AVIF_CHECKRES(avifRWStreamWriteBox(&dedup->s, "lsel", AVIF_BOX_SIZE_TBD, &lsel));
             // Layer Selection Property
-            // Section 2.3.1 of of AV1 Image File Format specification v1.1.0:
+            // Section 2.3.1 of AV1 Image File Format specification v1.1.0:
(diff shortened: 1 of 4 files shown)


src/write.c
```diff
@@ -2921,6 +2920,19 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
             }
             avifRWStreamFinishBox(&dedup->s, a1lx);
             AVIF_CHECKRES(avifItemPropertyDedupFinish(dedup, s, &item->ipma, AVIF_FALSE));
+
+            avifItemPropertyDedupStart(dedup);
+            avifBoxMarker lsel;
+            AVIF_CHECKRES(avifRWStreamWriteBox(&dedup->s, "lsel", AVIF_BOX_SIZE_TBD, &lsel));
+            // Layer Selection Property
+            // Section 2.3.1 of of AV1 Image File Format specification v1.1.0:
```
vrabaud, line 2928: "of of"

tongyuantongyu:
I'd like to mention that Firefox was rejecting AVIF images with 'lsel' property, so I didn't add 'lsel' property to images when I was implementing progressive encoding.

I just tested that Firefox (131.0b5) still rejects AVIF images with 'lsel' property, unless you set `image.avif.compliance_strictness` to `0` in `about:config` (because 'lsel' is an essential property but is [unsupported](https://github.com/mozilla/mp4parse-rust/blob/master/mp4parse/src/lib.rs#L3126) by Firefox, so per spec the decoding fails).

Currently `avifenc` produces progressive images that Firefox accepts (because 'a1lx' is non-essential, so it can be ignored), but adding 'lsel' will prevent these images from being decoded even non-progressively.

I would suggest landing this change at least after we ensure major decoder implementations all have supported 'lsel'. I know the following decoders do not support 'lsel', but there may be more:
- [mp4parse-rust](https://github.com/mozilla/mp4parse-rust) (Firefox)
- Windows with AV1 Video Extension
  -  It rejects images with only 'a1lx' as well.

And even after that, this will still upset early adopters of progressive AVIFs because images will appear broken on old version of browsers, instead of losing progressive but still work.

maryla-uc:
Thanks @tongyuantongyu for the valuable background information.
I agree with you that it's best not to add this property for now. I will add some comments in the code instead, and ideally we should work on getting this property supported in major decoders...

src/write.c
```diff
@@ -2921,6 +2920,19 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
...
+            AVIF_CHECKRES(avifRWStreamWriteBox(&dedup->s, "lsel", AVIF_BOX_SIZE_TBD, &lsel));
+            // Layer Selection Property
+            // Section 2.3.1 of AV1 Image File Format specification v1.1.0:
+            // The value 0xFFFF is reserved for a special meaning. If a lsel property is associated with an AV1
+            // Image Item but its layer_id value is set to 0xFFFF, the renderer is free to render either only
+            // the output image of the highest spatial layer, or to render all output images of all the intermediate
+            // layers and the highest spatial layer, resulting in a form of progressive decoding.
+            AVIF_CHECKRES(avifRWStreamWriteU16(&dedup->s, 0xFFFF));
+            avifRWStreamFinishBox(&dedup->s, lsel);
+            AVIF_CHECKRES(avifItemPropertyDedupFinish(dedup, s, &item->ipma, AVIF_TRUE));
```
wantehchang, line 2935: The code has changed slightly. Here is the patch for the current code:

```
index fc68e95d..23bb966d 100644
--- a/src/write.c
+++ b/src/write.c
@@ -3080,8 +3080,18 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
             AVIF_CHECKRES(avifRWStreamFinishBox(&dedup->s, a1lx));
             AVIF_CHECKRES(avifItemPropertyDedupFinish(dedup, s, &item->associations, /*essential=*/AVIF_FALSE));
 
-            // We don't add an 'lsel' property since many decoders do not support it and will reject the image,
-            // see https://github.com/AOMediaCodec/libavif/pull/2429
+            avifItemPropertyDedupStart(dedup);
+            avifBoxMarker lsel;
+            AVIF_CHECKRES(avifRWStreamWriteBox(&dedup->s, "lsel", AVIF_BOX_SIZE_TBD, &lsel));
+            // Layer Selection Property
+            // Section 2.3.1 of AV1 Image File Format specification v1.1.0:
+            // The value 0xFFFF is reserved for a special meaning. If a lsel property is associated with an AV1
+            // Image Item but its layer_id value is set to 0xFFFF, the renderer is free to render either only
+            // the output image of the highest spatial layer, or to render all output images of all the intermediate
+            // layers and the highest spatial layer, resulting in a form of progressive decoding.
+            AVIF_CHECKRES(avifRWStreamWriteU16(&dedup->s, 0xFFFF));
+            AVIF_CHECKRES(avifRWStreamFinishBox(&dedup->s, lsel));
+            AVIF_CHECKRES(avifItemPropertyDedupFinish(dedup, s, &item->associations, /*essential=*/AVIF_TRUE));
         }
 
         // Write out any opaque properties from avifImageAddOpaqueProperty() or avifImageAddUUIDProperty().
```

wantehchang:
Re: https://github.com/AOMediaCodec/libavif/pull/2429#issuecomment-2346484758

Firefox 155 (expected to be released on 2026-09-01) will support an 'lsel' property whose layer_id is 0xFFFF. See https://bugzilla.mozilla.org/show_bug.cgi?id=2055813#c8.

Issue #2412, opened 2024-08-25
Use SVT-AV1 PSY fork with --tune 4


gitoss:
The PSY fork of SVT-AV1 features a new --tune 4 which is designed to improve still picture (i.e. AVIF) encoding:

"--tune 4 - A new Tune called Still Picture has been introduced for AVIF encoding, with promising gains observed over aomenc, aomenc 4:4:4, and mainline SVT-AV1" https://github.com/AOMediaCodec/libavif/releases

It would be nice to have libavif binaries use this fork and tune out of the box (if SVT-AV1 encoder is selected). 

There are no indepennded benchmarks / image comparisons though as far as I can tell. https://github.com/gianni-rosato/svt-av1-psy/commit/393cf6dfe4d5315370ad429e891f784940ed7926

y-guyon:
Thank you for the suggestion.

We may want to wait for other benchmarks to confirm the gains and/or for the fork to be merged into the base repository before using it in libavif.

juliobbv-p:
Hi @y-guyon! I wanted to follow up on this issue with some updates to reassess the feature.

Regarding benchmarks, we just released our first results, which can be found on [this page](https://svt-av1-psy.com/avif/). Depending on the image, SVT-AV1-PSY gets between 5-15% gains over aomenc on the [CID22 validation set](https://cloudinary.com/labs/cid22) and the gb82 photographic dataset on the SSIMU2 metric.

As an example, here's [a visual comparison](https://slow.pics/c/oqnUcU1J) for a typical photo image. The SV1-AV1-PSY encoded image (147 KB) is 90.5% the percent the size of aomenc's (162 KB) for comparable image quality, and overall better consistency. For the record, the aomenc image took 3x as long to encode on my machine.

SVT-AV1-PSY also recently gained the ability to encode images with odd dimensions, and with sizes as small as 4x4 px (from the previous 64x64 px), which are especially useful for images as there's no longer a need to crop or pad them for the encoder to accept them.

vrabaud:
Very nice ! Thanks for the info ! Is there anything we should merge in libavif? I am thinking about https://github.com/gianni-rosato/libavif/commit/d53aa456d54716e4afd2f0b475ef97c6789e866e#diff-29748a2db41273018a16b71ccb60bcd7b632b86c78cc69bb88784702122ebdde

y-guyon:
Here is another benchmark:  
[...]

PR #2432, opened 2024-09-14
fix: correct SVT-AV1 defaults to avoid crash

gianni-rosato:
Occaisonally, I have observed SVT-AV1 builds on Linux crash due to these two options not being configured properly for CRF encoding. This should fix that issue.

### DIFF
--- a/src/codec_svt.c
+++ b/src/codec_svt.c
@@ -152,6 +152,12 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         }
         svt_config->qp = quantizer;
 
+        // Set gop constraint rc to 0, otherwise encoder won't run
+        svt_config->gop_constraint_rc = 0;
+
+        // Set target socket to -1 (both sockets)
+        svt_config->target_socket = -1;
+
         if (tileRowsLog2 != 0) {
             svt_config->tile_rows = tileRowsLog2;
         }


maryla-uc:
Hi,
Thank you for the pull request. Do you have an example file/command line to reproduce the crash? Are you sure this is the right fix? The values you are setting seem to be the default values according to the [API](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/Source/API/EbSvtAv1Enc.h?ref_type=heads#L893) and the [initialization function](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/Source/Lib/Globals/enc_settings.c?ref_type=heads#L949) (same goes for target_socket) so this change should be a noop.

gianni-rosato:
I'll try to reproduce this without this PR. I had a build a couple of weeks ago that was causing issues, and I was confused as well due to the API documentation. Let me get back to you.

vrabaud:
Thx, this is related to https://github.com/AOMediaCodec/libavif/issues/2412#issuecomment-2447138086 right ? to get psy in.

gianni-rosato:
This is related to my fork of libavif designed to avoid issues when using SVT-AV1-PSY. This PR might be stale though.

vrabaud:
Is that still needed at head or can we close this?

wantehchang:
@gitoss Thanks for the report.

Closed the pull request based on what gitoss reported.

PR #2437, opened 2024-09-23
Bump AVM to research-8.0.0

vrabaud:


.github/workflows/ci-unix-static-av2.yml
```diff
@@ -39,6 +39,10 @@ jobs:
         gcc-version: ${{ matrix.gcc }}
         libyuv: 'LOCAL'
 
+    - name: Install cpuinfo
+      run:
+        sudo apt install -y cpuinfo
```
wantehchang, line 44: Why do we need to install cpuinfo here? It would be good to mention this in the commit message.

cmake/Modules/LocalAom.cmake
```diff
@@ -89,13 +89,15 @@ else()
     endif()
 
     if(AVIF_CODEC_AVM)
+        # TODO(vrabaud) remove the PATCH_COMMAND once upstream is fixed.
```
wantehchang, line 92: Optional: It would be good to link to a merge request for the AVM upstream.

wantehchang (commented):
Thank you for writing this pull request!

wantehchang (commented):
Vincent: We need to remove `#if CONFIG_LR_IMPROVEMENTS` and the corresponding `#endif` from src/obu.c.

.github/workflows/ci-unix-static-av2.yml
vrabaud: This is not needed anymore, removing

cmake/Modules/LocalAom.cmake
vrabaud: No need for a patch anymore as I disable tensorflow. I followed-up with and filed an issue: https://gitlab.com/AOMediaCodec/avm/-/issues/504

y-guyon (approved):

PR #2438, opened 2024-09-24
Add avifROStreamSkipBits()

y-guyon:
Rename avifROStreamReadBitsX() to avifROStreamReadBitsUX().
Add avifROStreamReadBitsU16().
Refactor avifParseContentLightLevelInformation() to support unaligned fields.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -146,19 +146,39 @@ avifBool avifROStreamReadU64(avifROStream * stream, uint64_t * v)
     return AVIF_TRUE;
 }
 
-// Override of avifROStreamReadBits() for convenient uint8_t output.
-avifBool avifROStreamReadBits8(avifROStream * stream, uint8_t * v, size_t bitCount)
+avifBool avifROStreamSkipBits(avifROStream * stream, size_t bitCount)
 {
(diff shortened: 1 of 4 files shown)


src/stream.c
```diff
@@ -146,19 +146,39 @@ avifBool avifROStreamReadU64(avifROStream * stream, uint64_t * v)
...
-    *v = (uint8_t)v32;
+    const size_t num_bytes = (bitCount + 7) / 8;
+    AVIF_CHECK(avifROStreamSkip(stream, num_bytes));
+    stream->numUsedBitsInPartialByte = bitCount % 8;
+    return AVIF_TRUE;
+}
+
+avifBool avifROStreamReadBitsU8(avifROStream * stream, uint8_t * v, size_t bitCount)
+{
+    assert(bitCount <= sizeof(*v) * 8);
```
wantehchang, line 165: Optional: It seems that these assertions should be replaced by error checking (returning `AVIF_FALSE` on failure).

y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2457.

PR #2440, opened 2024-09-24
Implement the HDR part of MinimizedImageBox

y-guyon:
Write and parse tone mapping and gain map elements in SlimHEIF.

Fixes #2393.

PR #2439, opened 2024-09-24
Fix return type of avifWriteToneMappedImagePayload

y-guyon:
Also refactor avifWriteToneMappedImagePayload() into avifWriteGainmapMetadata() which will be used by the HDR part of SlimHEIF.
Add avifGainmapMetadataSize() which will be used by the HDR part of SlimHEIF.
Add spec comments and references.

PR #2441, opened 2024-09-25
Only generate valid fuzzed avifGainMapMetadata

y-guyon:
Also add avifGainMapMetadataValidate() to verify it is valid before encoding.

This is related to the fuzz failure in https://github.com/AOMediaCodec/libavif/pull/2440.  
This fuzz failure was introduced by https://github.com/AOMediaCodec/libavif/pull/2439 which changed the way avifGainMapMetadata was encoded (`avifRWStreamWriteBits(useBaseColorSpace)` instead of a packed `avifRWStreamWriteU8(flags)`).


tests/gtest/avif_fuzztest_helpers.cc
```diff
@@ -169,6 +169,42 @@ DecoderPtr AddGainMapOptionsToDecoder(DecoderPtr decoder,
   // is no longer the case when this option is on.
   return decoder;
 }
+
+ImagePtr AddGainMapToImage(
+    ImagePtr image, ImagePtr gain_map, int32_t gainMapMinN0,
```
maryla-uc, line 174: snake case?

maryla-uc (approved):

Issue #2442, opened 2024-09-25
Failed to build libavif with Clang GNU driver on Windows


AlifianK:
Failed to build libavif with Clang GNU driver on WIndows, MSVC and clang-cl are not affected.
```
cmake -G Ninja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_FLAGS='-flto=full -march=native' -DCMAKE_CXX_FLAGS='-flto=full -march=native' -DCMAKE_BUILD_TYPE=Release -DAVIF_CODEC_DAV1D=LOCAL -DAVIF_LIBYUV=LOCAL -DAVIF_LIBSHARPYUV=LOCAL -DAVIF_JPEG=LOCAL -DAVIF_ZLIBPNG=LOCAL -DAVIF_CODEC_AOM=LOCAL -DAVIF_BUILD_APPS=ON ..
```
```
-- The C compiler identification is Clang 19.1.0 with GNU-like command-line
...
[83/84] Linking CXX executable avifenc.exe
FAILED: avifenc.exe
[...]

PR #2443, opened 2024-09-25
List utf8.manifest as normal source file for MSVC

wantehchang:
List utf8.manifest as normal source file for MSVC

For toolchains targeting the MSVC ABI, CMake has first-class support for
embedding manifests. Manual embedding through a .rc source is needed
only when targeting the GNU ABI on Windows (MinGW).

Suggested by Brad King.

Fix https://github.com/AOMediaCodec/libavif/issues/2442.


jzern (approved):

wantehchang:
@jzern James: I updated this pull request to the fix that Brad King suggested in https://github.com/AOMediaCodec/libavif/issues/2442#issuecomment-2377364754. Please review again. Thanks!

@bradking Brad: Please make sure I implemented your suggestion correctly. Thanks!

bradking:
Yes, the diff as of commit 4362ac61487b19e62e79017396ca5c23f98c80fa is what I suggested, thanks.

wantehchang:
@kmilos Miloš: I'd appreciate it if you could also take a look. Thanks!

@bradking My concern with the `if(MSVC) elseif(MINGW)` form is that it will silently ignore the utf8.manifest file if neither `MSVC` nor `MINGW` is true. I wonder if we should use the  `if(MSVC) else()` or  `if(MINGW) else()` form instead:

bradking:
MSVC and MINGW are the only two ABI possibilities inside `if(WIN32)`, but if you prefer, `if(MSVC) .manifest else() .rc endif()` is probably best.

CMakeLists.txt
```diff
@@ -718,17 +718,12 @@ endif()
...
-        # don't use the /MANIFEST:NO linker option:
-        #   /MANIFEST:EMBED,ID=1" failed (exit code 1123) with the following
-        #   output:
-        #   CVTRES : fatal error CVT1100: duplicate resource.  type:MANIFEST,
-        #   name:1, language:0x0409
-        if(NOT MINGW)
-            target_link_options(avifenc PRIVATE /MANIFEST:NO)
+        if(MSVC)
+            target_sources(avifenc PRIVATE apps/utf8.manifest)
+        elseif(MINGW)
```
jzern, line 723: I leave making this an `else()` as Brad suggested up to you.

jzern (approved):

PR #2444, opened 2024-09-26
Refactor gain map API

maryla-uc:
avifGainMapMetadata and avifGainMapMetadataDouble were not future proof. Instead of adding Create()/Destroy() functions, simply inline avifGainMapMetadata inside avifGainMap and remove avifGainMapMetadataDouble.

Add avifSignedFraction (int32/uint32) and avifUnsignedFraction (uint32/uint32), in addition to the existing avifFraction (int32/int32). It's unfortunate that avifFraction uses a int32 denominator, but I don't think this can be changed now. This is not strictly necessary but makes the API a bit neater.

Expose avifDoubleToSignedFraction and avifDoubleToUnsignedFraction in avif.h
This is not strictly necessary either, since users should rarely have to fill in the fraction values themselves, and even if they do, they can use a naive method like setting the denominator to a large fixed value. But they are convenient and avifjpeg.c does use them right now. However, I'm open to removing them from the API, or putting them behind AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP (although this define should go away eventually).


src/avif.c
```diff
@@ -262,7 +262,23 @@ avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifP
             dstImage->gainMap = avifGainMapCreate();
             AVIF_CHECKERR(dstImage->gainMap, AVIF_RESULT_OUT_OF_MEMORY);
         }
-        dstImage->gainMap->metadata = srcImage->gainMap->metadata;
+        for (int c = 0; c < 3; ++c) {
+            dstImage->gainMap->gainMapMin[c].n = srcImage->gainMap->gainMapMin[c].n;
+            dstImage->gainMap->gainMapMin[c].d = srcImage->gainMap->gainMapMin[c].d;
```
y-guyon, line 267: Would `dstImage->gainMap->gainMapMin[c] = srcImage->gainMap->gainMapMin[c];` work?

src/gainmap.c
```diff
@@ -9,71 +9,43 @@
...
-        AVIF_CHECK(avifDoubleToUnsignedFraction(src->gainMapGamma[i], &dst->gainMapGammaN[i], &dst->gainMapGammaD[i]));
-        AVIF_CHECK(avifDoubleToSignedFraction(src->baseOffset[i], &dst->baseOffsetN[i], &dst->baseOffsetD[i]));
-        AVIF_CHECK(avifDoubleToSignedFraction(src->alternateOffset[i], &dst->alternateOffsetN[i], &dst->alternateOffsetD[i]));
-    }
-    AVIF_CHECK(avifDoubleToUnsignedFraction(src->baseHdrHeadroom, &dst->baseHdrHeadroomN, &dst->baseHdrHeadroomD));
-    AVIF_CHECK(avifDoubleToUnsignedFraction(src->alternateHdrHeadroom, &dst->alternateHdrHeadroomN, &dst->alternateHdrHeadroomD));
-    dst->useBaseColorSpace = src->useBaseColorSpace;
-    return AVIF_TRUE;
+        gainMap->gainMapMin[i].n = 1;
+        gainMap->gainMapMin[i].d = 1;
```
y-guyon, line 16: `gainMap->gainMapMin[i] = avifFrac{1, 1};`?

src/avif.c
```diff
@@ -1192,17 +1208,17 @@ avifGainMap * avifGainMapCreate(void)
...
-    gainMap->metadata.baseHdrHeadroomD = 1;
-    gainMap->metadata.alternateHdrHeadroomD = 1;
+        gainMap->gainMapMin[i].d = 1;
+        gainMap->gainMapMax[i].d = 1;
+        gainMap->gainMapGamma[i].d = 1;
+        gainMap->baseOffset[i].d = 1;
+        gainMap->alternateOffset[i].d = 1;
+    }
+    gainMap->baseHdrHeadroom.d = 1;
+    gainMap->alternateHdrHeadroom.d = 1;
```
y-guyon, line 1221: Should this be replaced by `avifGainMapSetDefaults()`?

src/gainmap.c
```diff
@@ -9,71 +9,43 @@
...
+    }
+    gainMap->baseHdrHeadroom.n = 0;
+    gainMap->baseHdrHeadroom.d = 1;
+    gainMap->alternateHdrHeadroom.n = 0;
+    gainMap->alternateHdrHeadroom.d = 1;
+    gainMap->useBaseColorSpace = AVIF_TRUE;
 }
 
-avifBool avifGainMapMetadataFractionsToDouble(avifGainMapMetadataDouble * dst, const avifGainMapMetadata * src)
+static float toFloat(int numerator, int denominator)
```
y-guyon, line 33: No implicit cast issue? `denominator` should always be `uint32_t` at least

src/gainmap.c
```diff
@@ -9,71 +9,43 @@
...
 // ---------------------------------------------------------------------------
 // Apply a gain map.
 
 // Returns a weight in [-1.0, 1.0] that represents how much the gain map should be applied.
-static float avifGetGainMapWeight(float hdrHeadroom, const avifGainMapMetadataDouble * metadata)
+static float avifGetGainMapWeight(float hdrHeadroom, const avifGainMap * gainMap)
 {
-    const float baseHdrHeadroom = (float)metadata->baseHdrHeadroom;
-    const float alternateHdrHeadroom = (float)metadata->alternateHdrHeadroom;
+    const float baseHdrHeadroom = toFloat(gainMap->baseHdrHeadroom.n, gainMap->baseHdrHeadroom.d);
```
y-guyon, line 47: What about two `static float toFloat(avifFrac frac)` instead

src/gainmap.c
```diff
@@ -250,9 +217,21 @@ avifResult avifRGBImageApplyGainMap(const avifRGBImage * baseImage,
 
     float rgbMaxLinear = 0; // Max tone mapped pixel value across R, G and B channels.
     float rgbSumLinear = 0; // Sum of max(r, g, b) for mapped pixels.
-    const float gammaInv[3] = { 1.0f / (float)metadata.gainMapGamma[0],
-                                1.0f / (float)metadata.gainMapGamma[1],
-                                1.0f / (float)metadata.gainMapGamma[2] };
+    const float gammaInv[3] = { toFloat(gainMap->gainMapGamma[0].d, gainMap->gainMapGamma[0].n),
+                                toFloat(gainMap->gainMapGamma[1].d, gainMap->gainMapGamma[1].n),
+                                toFloat(gainMap->gainMapGamma[2].d, gainMap->gainMapGamma[2].n) };
```
y-guyon, line 222: Add a comment that it is 1/gamma at decoding as mentioned in avif.h

src/gainmap.c
```diff
@@ -707,15 +694,12 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
-    // All of gainMapMetadata has been populated now (except for gamma which is left to the default
-    // value), so convert to the fraction form in which it will be stored.
-    if (!avifGainMapMetadataDoubleToFractions(&gainMap->metadata, &gainMapMetadata)) {
-        res = AVIF_RESULT_UNKNOWN_ERROR;
-        goto cleanup;
+        if (!avifDoubleToSignedFraction(gainMapMinLog2[singleChannel ? 0 : c], &gainMap->gainMapMin[c]) ||
+            !avifDoubleToSignedFraction(gainMapMaxLog2[singleChannel ? 0 : c], &gainMap->gainMapMax[c]) ||
+            !avifDoubleToSignedFraction(alternateOffset[c], &gainMap->alternateOffset[c]) ||
+            !avifDoubleToSignedFraction(baseOffset[c], &gainMap->baseOffset[c])) {
+            goto cleanup;
```
y-guyon, line 701: No longer need for `res = AVIF_RESULT_UNKNOWN_ERROR;`?

tests/gtest/avif_fuzztest_helpers.cc
```diff
@@ -189,22 +189,48 @@ ImagePtr AddGainMapToImage(
...
-      {base_offset_d0, base_offset_d1, base_offset_d2},
-      {alternate_offset_n0, alternate_offset_n1, alternate_offset_n2},
-      {alternate_offset_d0, alternate_offset_d1, alternate_offset_d2},
-      base_hdr_headroom_n,
-      base_hdr_headroom_d,
-      alternate_hdr_headroom_n,
-      alternate_hdr_headroom_d,
-      use_base_color_space};
+
+  image->gainMap->gainMapMin[0].n = gain_map_min_n0;
```
y-guyon, line 193: This function could take `avifFrac` arguments and there could be a custom domain for these, but from my experience it is a pain to get it to compile

tests/gtest/avifgainmaptest.cc
```diff
@@ -309,25 +305,24 @@ TEST(GainMapTest, EncodeDecodeMetadataSameDenominator) {
...
-    image->gainMap->metadata.alternateOffsetN[c] = 3;
-    image->gainMap->metadata.alternateOffsetD[c] = 4;
-    image->gainMap->metadata.gainMapGammaN[c] = 5;
-    image->gainMap->metadata.gainMapGammaD[c] = 6;
-    image->gainMap->metadata.gainMapMinN[c] = 7;
-    image->gainMap->metadata.gainMapMinD[c] = 8;
-    image->gainMap->metadata.gainMapMaxN[c] = 9;
-    image->gainMap->metadata.gainMapMaxD[c] = 10;
+    image->gainMap->baseOffset[c].n = 1;
+    image->gainMap->baseOffset[c].d = 2;
```
y-guyon, line 317: `=avifFrac{1,2}` here maybe too

y-guyon (approved):

include/avif/avif.h
```diff
@@ -426,14 +426,33 @@ typedef struct avifDiagnostics
...
 typedef struct avifFraction
 {
     int32_t n;
     int32_t d;
 } avifFraction;
 
+typedef struct avifSignedFraction
+{
+    int32_t n;
+    uint32_t d;
```
wantehchang, line 440: Optional: We can consider using the `avifFraction` type, unless we need to allow a denominator greater than `INT32_MAX`. The current implementation of `avifDoubleToUnsignedFractionImpl()` may use a denominator as large as `UINT32_MAX`, so this suggestion may not work.

src/utils.c
```diff
@@ -269,20 +269,20 @@ static avifBool avifDoubleToUnsignedFractionImpl(double v, uint32_t maxNumerator
     return AVIF_TRUE;
 }
 
-avifBool avifDoubleToSignedFraction(double v, int32_t * numerator, uint32_t * denominator)
+avifBool avifDoubleToSignedFraction(double v, avifSignedFraction * fraction)
 {
     uint32_t positive_numerator;
-    if (!avifDoubleToUnsignedFractionImpl(fabs(v), INT32_MAX, &positive_numerator, denominator)) {
+    if (!avifDoubleToUnsignedFractionImpl(fabs(v), INT32_MAX, &positive_numerator, &fraction->d)) {
```
wantehchang, line 275: In `avifDoubleToUnsignedFractionImpl()`, the variable `maxD` is declared as `uint64_t`, but it is only used in a comparison with `newD`, which is a `double`. So `maxD` could be declared as `double`. If `maxD` should have an integer type, `uint32_t` is sufficient.

include/avif/avif.h
maryla-uc: The wire format uses uint32_t for the denominator. So when decoding a file, the value of the denominator might be larger than INT32_MAX, and we have to be able to store it, unless we're ok with modifying the values to make them fit in a int32/int32 fraction.
I agree it's unfortunate to have so many fraction types, and avifFraction and avifSignedFraction are indeed very similar.

wantehchang (commented):
Maryla: Please consider my review as supplemental. I only reviewed the changes to avif.h. They look good.

wantehchang:
@vigneshvg FYI.

@maryla-uc Maryla: Could you or Vignesh port the changes to the public header avif.h to CrabbyAvif? Thanks.

Issue #2447, opened 2024-09-28
MSVC build error C2220 due to C4232


sezero:
This happened in CI in our fork. E.g.: 

```
D:\a\libavif\libavif\src\reformat_libyuv.c(536): error C2220: the following warning is treated as an error
D:\a\libavif\libavif\src\reformat_libyuv.c(536): warning C4232: nonstandard extension used: 'lutYuv400ToRgbMatrix': address of dllimport 'I400ToARGBMatrix' is not static, identity not guaranteed
```

Full error log:
<details>

```
[12/159] Building C object CMakeFiles\avif_obj.dir\src\reformat_libyuv.c.obj
FAILED: CMakeFiles/avif_obj.dir/src/reformat_libyuv.c.obj 
C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1441~1.341\bin\Hostx64\x64\cl.exe  /nologo -DAVIF_CODEC_AOM=1 -DAVIF_CODEC_AOM_DECODE=1 -DAVIF_CODEC_AOM_ENCODE=1 -DAVIF_CODEC_DAV1D=1 -DAVIF_CODEC_RAV1E=1 -DAVIF_ENABLE_EXPERIMENTAL_GAIN_MAP -DAVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION -DAVIF_ENABLE_EXPERIMENTAL_MINI -DAVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM -DAVIF_ENABLE_EXPERIMENTAL_YCGCO_R -DAVIF_LIBYUV_ENABLED=1 -D_CRT_NONSTDC_NO_WARNINGS -D_CRT_SECURE_NO_WARNINGS -ID:\a\libavif\libavif\include -external:ID:\a\libavif\libavif\vcpkg\installed\x64-windows-release\include -external:ID:\a\libavif\libavif\vcpkg\installed\x64-windows-release\include\rav1e -external:W0 /DWIN32 /D_WINDOWS /MD /O2 /Ob2 /DNDEBUG /source-charset:utf-8 /execution-charset:us-ascii /W4 /wd4324 /WX /showIncludes /FoCMakeFiles\avif_obj.dir\src\reformat_libyuv.c.obj /FdCMakeFiles\avif_obj.dir\ /FS -c D:\a\libavif\libavif\src\reformat_libyuv.c
[...]

PR #2449, opened 2024-09-30
Disable the MSVC C4232 warning

wantehchang:
The way we use the addresses of the libyuv functions in src/reformat_libyuv.c, the addresses of those functions do not need to be "static", which I think means fixed from execution to execution. As long as the addresses of those functions are fixed during an execution, that is fine for src/reformat_libyuv.c.

Fix https://github.com/AOMediaCodec/libavif/issues/2447.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,8 +293,11 @@ add_library(avif_enable_warnings INTERFACE)
 if(MSVC)
     message(STATUS "libavif: Enabling warnings for MSVC")
     target_compile_options(
-        avif_enable_warnings INTERFACE /W4 # For clang-cl, /W4 enables -Wall and -Wextra
-                                       /wd4324 # Disable: structure was padded due to alignment specifier
+        avif_enable_warnings
+        INTERFACE /W4 # For clang-cl, /W4 enables -Wall and -Wextra
+                  /wd4232 # Disable: address of dllimport 'dllimport' is not static,
+                          # identity not guaranteed
+                  /wd4324 # Disable: structure was padded due to alignment specifier
     )
     # Disable deprecation warnings about POSIX function names such as setmode (replaced by the ISO C and C++ conformant name _setmode).
     # Disable deprecation warnings about unsafe CRT library functions such as fopen (replaced by fopen_s).


CMakeLists.txt
```diff
@@ -293,8 +293,11 @@ add_library(avif_enable_warnings INTERFACE)
 if(MSVC)
     message(STATUS "libavif: Enabling warnings for MSVC")
     target_compile_options(
-        avif_enable_warnings INTERFACE /W4 # For clang-cl, /W4 enables -Wall and -Wextra
-                                       /wd4324 # Disable: structure was padded due to alignment specifier
+        avif_enable_warnings
+        INTERFACE /W4 # For clang-cl, /W4 enables -Wall and -Wextra
+                  /wd4232 # Disable: address of dllimport 'dllimport' is not static,
+                          # identity not guaranteed
+                  /wd4324 # Disable: structure was padded due to alignment specifier
```
wantehchang, line 300: cmake-format reformatted lines 296-300 here. The only real change is the addition of lines 298-299.

wantehchang (commented):
In the [MSVC warning C4232 documentation](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4232?view=msvc-170), the difficult part is what it means for the address of a function to be "static". I interpret it to mean the address of a function is fixed from execution to execution. Please see if you agree that is what "static" means in this context. Thanks!

vigneshvg (approved):

CMakeLists.txt
```diff
@@ -293,8 +293,11 @@ add_library(avif_enable_warnings INTERFACE)
 if(MSVC)
     message(STATUS "libavif: Enabling warnings for MSVC")
     target_compile_options(
-        avif_enable_warnings INTERFACE /W4 # For clang-cl, /W4 enables -Wall and -Wextra
-                                       /wd4324 # Disable: structure was padded due to alignment specifier
+        avif_enable_warnings
+        INTERFACE /W4 # For clang-cl, /W4 enables -Wall and -Wextra
+                  /wd4232 # Disable: address of dllimport 'dllimport' is not static,
```
jzern, line 298: Though unlikely it sounds like this would be an error if `/Za` was used. Though I guess the alternative would be to manually load the functions.

jzern (commented):
lgtm

CMakeLists.txt
jzern: Or have a wrapper as you suggested with the totw reference in the bug.

wantehchang: Thanks for taking a look at the MSVC C4232 warning documentation. The wrapper would also fix the CFI issue, so it seems like the best long-term solution.

PR #2453, opened 2024-10-01
Check that the gain map metadata is valid on decoding.

maryla-uc:
All fraction denominators must be != 0, and the gamma numerator must also be != 0. Files with invalid gain map metadata are rejected.


maryla-uc:
Plot twist: the breakage was unrelated to this commit?

PR #2456, opened 2024-10-01
Use runner.os in artifact archive name

y-guyon:
"Linux-artifacts.zip" and "macOS-artifacts.zip" should be prettier than "ubuntu-latest-artifacts.zip" and "macos-latest-artifacts.zip".

See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#runner-context.


y-guyon:
I wonder if `runner.arch` should be part of the name of the archive too.

maryla-uc (approved):

PR #2459, opened 2024-10-02
Disable JPEG finding in libyuv

vrabaud:
This is to prevent extra files in libyuv to be built and have a dependency order problem.

### DIFF
--- a/cmake/Modules/LocalLibyuv.cmake
+++ b/cmake/Modules/LocalLibyuv.cmake
@@ -34,6 +34,7 @@ else()
         GIT_REPOSITORY "https://chromium.googlesource.com/libyuv/libyuv"
         BINARY_DIR "${LIBYUV_BINARY_DIR}"
         GIT_TAG "${AVIF_LOCAL_LIBYUV_TAG}"
+        PATCH_COMMAND sed -i.bak -e "s:find_package.*(.*JPEG.*)::" CMakeLists.txt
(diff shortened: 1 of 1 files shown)


maryla-uc (approved):
Thank you!!

vrabaud:
The windows CI problem is unrelated.

cmake/Modules/LocalLibyuv.cmake
```diff
@@ -34,6 +34,7 @@ else()
         GIT_REPOSITORY "https://chromium.googlesource.com/libyuv/libyuv"
         BINARY_DIR "${LIBYUV_BINARY_DIR}"
         GIT_TAG "${AVIF_LOCAL_LIBYUV_TAG}"
+        PATCH_COMMAND sed -i.bak -e "s:find_package.*(.*JPEG.*)::" CMakeLists.txt
```
wantehchang, line 37: Vincent: Are you planning to submit a patch to libyuv to handle this properly?

Looking at the other build systems in the libyuv source tree, I think we should add a LIBYUV_DISABLE_JPEG option to libyuv's CMakeLists.txt file.

vrabaud: No need actually, I reverted that patch: there is actually an official way to make find_package fail, cf https://github.com/AOMediaCodec/libavif/pull/2473

PR #2472, opened 2024-10-08
android_jni: Set threads to 2 instead of CPU count

vigneshvg:
Empirically, on Android devices with more than 1 core, it is
almost always better to use 2 threads than to use
number_of_cpu_cores threads.

Update the public java API function documentation to reflect this
change.

The new behavior is:
* negative values map to number of cpu cores.
* zero maps to 1 or 2 depending on single core or multi core devices.
* value >=0 maps to value.

This change ensures that users who use the library via JNI take the
optimal path when using the default value of 0.

### DIFF
--- a/android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
+++ b/android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
@@ -110,12 +110,11 @@ public class AvifDecoder {
    * @param bitmap The decoded pixels will be copied into the bitmap.
    *     If the bitmap dimensions do not match the decoded image's dimensions,
    *               then the decoded image will be scaled to match the bitmap's dimensions.
-   * @param threads Number of threads to be used for the AVIF decode. Zero means use number of CPU
-   *     cores as the thread count. Negative values are invalid. When this value is > 0, it is
-   *     simply mapped to the maxThreads parameter in libavif. For more details, see the
-   *     documentation for maxThreads variable in avif.h.
-   * @return true on success and false on failure. A few possible reasons for failure are: 1) Input
-   *     was not valid AVIF. 2) Negative value was passed for the threads parameter.
+   * @param threads Number of threads to be used for the AVIF decode. Zero means use the library
+   *     determined optimal value as the thread count. Negative values mean use the number of CPU
+   *     cores as the thread count. For more details, see the documentation for maxThreads variable
+   *     in avif.h.
+   * @return true on success and false on failure.
    */
   public static native boolean decode(ByteBuffer encoded, int length, Bitmap bitmap, int threads);
 
--- a/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
+++ b/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
@@ -6,6 +6,7 @@
 #include <cpu-features.h>
 #include <jni.h>
 
+#include <algorithm>
 #include <cstdio>
 #include <memory>
 #include <new>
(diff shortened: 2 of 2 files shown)


android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
```diff
@@ -110,10 +110,11 @@ public static boolean decode(ByteBuffer encoded, int length, Bitmap bitmap) {
    * @param bitmap The decoded pixels will be copied into the bitmap.
    *     If the bitmap dimensions do not match the decoded image's dimensions,
    *               then the decoded image will be scaled to match the bitmap's dimensions.
-   * @param threads Number of threads to be used for the AVIF decode. Zero means use number of CPU
-   *     cores as the thread count. Negative values are invalid. When this value is > 0, it is
-   *     simply mapped to the maxThreads parameter in libavif. For more details, see the
-   *     documentation for maxThreads variable in avif.h.
+   * @param threads Number of threads to be used for the AVIF decode. Zero means use the library
+   *     determined optimal value as the thread count. Negative values mean use number of CPU cores
+   *     as the thread count. When this value is > 0, it is simply mapped to the maxThreads
```
wantehchang, line 115: The sentence "When this value is > 0, it is simply mapped to the maxThreads parameter in libavif." is a little confusing. It seems to imply that the maxThreads parameter in libavif is only set when the value of the `threads` parameter is > 0.

wantehchang:
In Chrome's avif_image_decoder.cc, I also set libavif's maxThreads parameter to the hardcoded value of 2. I have two reasons:

1. Almost all computers are multicore today, so I just assume a multicore computer. Not using more than two cores avoids the performance core vs. efficiency core issue we observed when working on libgav1 multithreading.
2. Chrome is likely to decode many images at the same time, so it does not make sense to use many cores for each image.

android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
vigneshvg: Removed the sentence about > 0.

vigneshvg:
> In Chrome's avif_image_decoder.cc, I also set libavif's maxThreads parameter to the hardcoded value of 2. I have two reasons:
> 
> 1. Almost all computers are multicore today, so I just assume a multicore computer. Not using more than two cores avoids the performance core vs. efficiency core issue we observed when working on libgav1 multithreading.
> 2. Chrome is likely to decode many images at the same time, so it does not make sense to use many cores for each image.

Yes, this change is also along the same lines. The only difference is that there are several low-end Android devices in the market which are single core. So i chose min(core_count, 2) as the default value instead of always 2.

android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
```diff
@@ -110,10 +110,10 @@ public static boolean decode(ByteBuffer encoded, int length, Bitmap bitmap) {
    * @param bitmap The decoded pixels will be copied into the bitmap.
    *     If the bitmap dimensions do not match the decoded image's dimensions,
    *               then the decoded image will be scaled to match the bitmap's dimensions.
-   * @param threads Number of threads to be used for the AVIF decode. Zero means use number of CPU
-   *     cores as the thread count. Negative values are invalid. When this value is > 0, it is
-   *     simply mapped to the maxThreads parameter in libavif. For more details, see the
-   *     documentation for maxThreads variable in avif.h.
+   * @param threads Number of threads to be used for the AVIF decode. Zero means use the library
+   *     determined optimal value as the thread count. Negative values mean use number of CPU cores
```
wantehchang, line 114: Previously negative values were invalid. I guess you want to provide a way to specify "use the number of CPU cores", so you now use negative values for that purpose?

wantehchang, line 114: Nit: Add "the" before "number of CPU cores".

wantehchang (approved):
LGTM.

android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java
vigneshvg: yes that is correct.

vigneshvg: Done. Also removed the comment about failure because of negative values to threads.

wantehchang (approved):

PR #2473, opened 2024-10-09
Do not allow libyuv to find_package JPEG

vrabaud:


maryla-uc (approved):

cmake/Modules/LocalLibyuv.cmake
```diff
@@ -29,17 +29,26 @@ else()
     if(ANDROID_ABI)
         set(LIBYUV_BINARY_DIR "${LIBYUV_BINARY_DIR}/${ANDROID_ABI}")
     endif()
+
+    # unset JPEG_FOUND so that libyuv does not find it
```
wantehchang, line 33: Nit: Thanks for the fix. It would good to explain why we need to do this.

cmake/Modules/LocalLibyuv.cmake
```diff
@@ -29,17 +29,26 @@ else()
...
+    set(JPEG_FOUND_ORIG ${JPEG_FOUND})
```
wantehchang, line 34: If `JPEG_FOUND` is originally undefined, won't line 48 cause it to be set to an empty string (which is a false constant)?

If no code tests the `DEFINED JPEG_FOUND` condition, then this is fine.

vrabaud: Internal CMake does not care about JPEG_FOUND being set to empty at least. If you look into FindJPEG.cmake on your machine, it looks for libraries and includes to fill the JPEG_FOUND variable, whether it has been set before or not using find_package_handle_standard_args.

So we are safe

PR #2474, opened 2024-10-09
Update parseAV2SequenceHeader() for AVM CWG-E103

wantehchang:
In https://gitlab.com/AOMediaCodec/avm/-/merge_requests/1183, the syntax elements for the max frame width and height and color_config() were moved to the beginning of the sequence header OBU. parseAV2SequenceHeader() needs to be updated if AVM has this change. Since this change has not appeared in an AVM release tag, add the AVIF_ENABLE_CWG_E103 macro to indicate whether AVM has this change.

To facilitate the changes to parseAV2SequenceHeader(), split parseSequenceHeaderProfile() into two functions:
  - parseSequenceHeaderProfile() parses only seq_profile
  - parseSequenceHeaderLevelIdxAndTier parses seq_level_idx and seq_tier and change parseSequenceHeaderColorConfig() to not parse separate_uv_delta_q.

### DIFF
--- a/src/obu.c
+++ b/src/obu.c
@@ -141,7 +141,11 @@ static avifBool parseSequenceHeaderProfile(avifBits * bits, avifSequenceHeader *
         return AVIF_FALSE;
     }
     header->av1C.seqProfile = (uint8_t)seq_profile;
+    return !bits->error;
+}
 
+static avifBool parseSequenceHeaderLevelIdxAndTier(avifBits * bits, avifSequenceHeader * header)
+{
     uint32_t still_picture = avifBitsRead(bits, 1);
     header->reduced_still_picture_header = (uint8_t)avifBitsRead(bits, 1);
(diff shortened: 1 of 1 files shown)


y-guyon (approved):

src/obu.c
```diff
@@ -255,6 +259,7 @@ static avifBool parseSequenceHeaderEnabledFeatures(avifBits * bits, avifSequence
     return !bits->error;
 }
 
+// Note: Does not parse separate_uv_delta_q.
```
wantehchang, line 262: It may be better to fork this function into an AV1 and an AV2 version because I am planning to change the `chroma_sample_position` syntax element in AV2. I will try that in a follow-up pull request.

wantehchang (commented):
Yannis: Thanks for the review. I annotated some of the changes but forgot to publish them. Sorry! Here they are.

PR #2479, opened 2024-10-17
Make sure the gain map metadata is towards the start of the file.

maryla-uc:
The 'tmap' item data containing the gain map metadata was mistakenly being put at the end of the file.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -2195,8 +2195,8 @@ static avifResult avifEncoderWriteMediaDataBox(avifEncoder * encoder,
     const size_t mdatStartOffset = avifRWStreamOffset(s);
     for (uint32_t itemPasses = 0; itemPasses < 3; ++itemPasses) {
         // Use multiple passes to pack in the following order:
-        //   * Pass 0: metadata (Exif/XMP)
-        //   * Pass 1: alpha, gain map (AV1)
+        //   * Pass 0: metadata (Exif/XMP/gain map metadata)
+        //   * Pass 1: alpha, gain map image (AV1)
         //   * Pass 2: all other item data (AV1 color)
         //
         // See here for the discussion on alpha coming before color:
@@ -2215,7 +2215,8 @@ static avifResult avifEncoderWriteMediaDataBox(avifEncoder * encoder,
                 // this item has nothing for the mdat box
                 continue;
             }
-            const avifBool isMetadata = !memcmp(item->type, "mime", 4) || !memcmp(item->type, "Exif", 4);
+            const avifBool isMetadata = !memcmp(item->type, "mime", 4) || !memcmp(item->type, "Exif", 4) ||
+                                        !memcmp(item->type, "tmap", 4);
             if (metadataPass != isMetadata) {
                 // only process metadata (XMP/Exif) payloads when metadataPass is true
                 continue;
--- a/tests/gtest/avifincrtest_helpers.cc
+++ b/tests/gtest/avifincrtest_helpers.cc
@@ -80,8 +80,8 @@ void ComparePartialYuva(const avifImage& image1, const avifImage& image2,
 // byte_count were given to the decoder, for an image of height rows, split into
 // cells of cell_height rows.
 uint32_t GetMinDecodedRowCount(uint32_t height, uint32_t cell_height,
-                               bool has_alpha, size_t available_byte_count,
-                               size_t byte_count,
+                               bool has_alpha, bool has_gain_map,
+                               size_t available_byte_count, size_t byte_count,
                                bool enable_fine_incremental_check) {
   // The whole image should be available when the full input is.
   if (available_byte_count >= byte_count) {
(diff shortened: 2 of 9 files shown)


tests/gtest/avifincrtest_helpers.cc
```diff
@@ -332,6 +331,13 @@ avifResult DecodeIncrementally(const avifRWData& encoded_avif,
     data.available.size = std::min(data.available.size + step, data.full_size);
     parse_result = avifDecoderParse(decoder);
   }
+  if (data.available.size == data.full_size) {
+    // Can happen if the data is in 'idat', or if some metadata is at the end of
+    // the file. But ideally this should be avoided.
+    printf(
+        "WARNING: had to provide the whole file for avifDecoderParse() to "
+        "succeed\n");
+  }
```
y-guyon, line 340: I would check this only if an arg given to `DecodeIncrementally()` is true and return an error or abort here. This way it can be silent for fuzzed encoded bytes but strongly enabled for encoded bytes generated by libavif.

tests/gtest/avifincrtest_helpers.cc
```diff
@@ -347,13 +353,30 @@ avifResult DecodeIncrementally(const avifRWData& encoded_avif,
       return AVIF_RESULT_INVALID_ARGUMENT;
     }
     const uint32_t decoded_row_count = avifDecoderDecodedRowCount(decoder);
-    AVIF_CHECKERR(decoded_row_count >= previously_decoded_row_count,
-                  AVIF_RESULT_INVALID_ARGUMENT);
+    if (decoded_row_count < previously_decoded_row_count) {
+      printf("ERROR: decoded row count decreased from %d to %d\n",
+             previously_decoded_row_count, decoded_row_count);
+      return AVIF_RESULT_INVALID_ARGUMENT;
```
y-guyon, line 359: Consider using `AVIF_CHECKERR(false, AVIF_RESULT_INVALID_ARGUMENT);` to keep the finest break point on `AVIF_BREAK_ON_ERROR`.

Same in other places.

tests/gtest/avifincrtest_helpers.cc
```diff
@@ -347,13 +353,30 @@ avifResult DecodeIncrementally(const avifRWData& encoded_avif,
...
     const uint32_t min_decoded_row_count = GetMinDecodedRowCount(
         reference.height, cell_height, reference.alphaPlane != nullptr,
-        data.available.size, data.full_size, enable_fine_incremental_check);
-    AVIF_CHECKERR(decoded_row_count >= min_decoded_row_count,
-                  AVIF_RESULT_INVALID_ARGUMENT);
+        has_gain_map, data.available.size, data.full_size,
+        enable_fine_incremental_check);
+    if (decoded_row_count != previously_decoded_row_count) {
+      printf("bytes %zu decoded %d lines (expected min %d)\n",
+             data.available.size, decoded_row_count, min_decoded_row_count);
```
y-guyon, line 371: Leftover?

src/write.c
```diff
@@ -2195,8 +2195,8 @@ static avifResult avifEncoderWriteMediaDataBox(avifEncoder * encoder,
     const size_t mdatStartOffset = avifRWStreamOffset(s);
     for (uint32_t itemPasses = 0; itemPasses < 3; ++itemPasses) {
         // Use multiple passes to pack in the following order:
-        //   * Pass 0: metadata (Exif/XMP)
-        //   * Pass 1: alpha, gain map (AV1)
+        //   * Pass 0: metadata (Exif/XMP/gain map metadata)
```
y-guyon, line 2198: Unrelated to this PR: Shouldn't Exif/XMP be at the end?

y-guyon (approved):

tests/gtest/avifincrtest_helpers.cc
```diff
@@ -104,15 +104,14 @@ uint32_t GetMinDecodedRowCount(uint32_t height, uint32_t cell_height,
...
-  // Alpha, if any, is assumed to be located before the other planes and to
-  // represent at most 50% of the payload.
-  if (has_alpha) {
-    if (available_byte_count <= (byte_count / 2)) {
-      return 0;
-    }
-    available_byte_count -= byte_count / 2;
-    byte_count -= byte_count / 2;
+  // Extra planes (alpha, gain map), if any, are assumed to be located before
+  // the other planes and to represent at most 50% of the payload.
```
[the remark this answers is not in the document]
maryla-uc: Yeah I changed it a bit randomly, but after a while I realized the problem was not this formula but actually this part:

```C
  // There is no valid AV1 payload smaller than 10 bytes, so all but one cell
  // should be decoded if at most 10 bytes are missing.
  if ((available_byte_count + 10) >= byte_count) {
```

Because my image had tiles that were all identical, and so all the tiles pointed to the same blob in `mdat` and were all decoded at once. So I fixed it by changing the test image.
I left this just in case, although looking at it again my change was wrong (and I changed it again), but I'm not sure this code is really exercised right now.

src/write.c
maryla-uc: It's the same as the gain map metadata: if we put it at the end of the file, avifDecoderParse() will block untill it receives the whole file, unless `ignoreExif` and `ignoreXMP` are set. In the documentation for those fields we say `some of these payloads are (unfortunately) packed at the end of the file` so we consider this to be a bad thing.

PR #2497, opened 2024-10-31
Fix avifROStreamSkipBits bug for incomplete byte

wantehchang:
Fix a bug in avifROStreamSkipBits if
bitCount < 8 - stream->numUsedBitsInPartialByte.

Bug: b:376733153

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -153,6 +153,9 @@ avifBool avifROStreamSkipBits(avifROStream * stream, size_t bitCount)
         const size_t padding = AVIF_MIN(8 - stream->numUsedBitsInPartialByte, bitCount);
         stream->numUsedBitsInPartialByte = (stream->numUsedBitsInPartialByte + padding) % 8;
         bitCount -= padding;
+        if (stream->numUsedBitsInPartialByte != 0) {
+            return AVIF_TRUE;
+        }
     }
     const size_t num_bytes = (bitCount + 7) / 8;
     AVIF_CHECK(avifROStreamSkip(stream, num_bytes));
--- a/tests/gtest/avifstreamtest.cc
+++ b/tests/gtest/avifstreamtest.cc
@@ -167,6 +167,32 @@ TEST(StreamTest, Roundtrip) {
   EXPECT_FALSE(avifROStreamSkip(&ro_stream, /*byteCount=*/1));
 }
 
+TEST(StreamTest, SkipBits) {
+  const uint8_t data[40] = {};
+  avifROData ro_data = {data, sizeof(data)};
+  avifDiagnostics diag;
+  avifDiagnosticsClearError(&diag);
+  avifROStream ro_stream;
+  avifROStreamStart(&ro_stream, &ro_data, &diag, "diagContext");
+
+  EXPECT_TRUE(avifROStreamSkip(&ro_stream, 32));
+  EXPECT_EQ(avifROStreamOffset(&ro_stream), 32);
+  EXPECT_EQ(ro_stream.numUsedBitsInPartialByte, 0);
+
+  uint32_t unused;
+  EXPECT_TRUE(avifROStreamReadBitsU32(&ro_stream, &unused, 5));
+  EXPECT_EQ(avifROStreamOffset(&ro_stream), 33);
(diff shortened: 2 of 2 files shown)


tests/gtest/avifstreamtest.cc
```diff
@@ -167,6 +167,32 @@ TEST(StreamTest, Roundtrip) {
...
+  avifROStreamStart(&ro_stream, &ro_data, &diag, "diagContext");
+
+  EXPECT_TRUE(avifROStreamSkip(&ro_stream, 32));
+  EXPECT_EQ(avifROStreamOffset(&ro_stream), 32);
+  EXPECT_EQ(ro_stream.numUsedBitsInPartialByte, 0);
+
+  uint32_t unused;
+  EXPECT_TRUE(avifROStreamReadBitsU32(&ro_stream, &unused, 5));
+  EXPECT_EQ(avifROStreamOffset(&ro_stream), 33);
+  EXPECT_EQ(ro_stream.numUsedBitsInPartialByte, 5);
```
wantehchang, line 185: This unit test simulates the reproducer testcase for b/376733153.

Before the `avifROStreamSkipBits(s, 1)` call, the contents of `*s` are:

```
(gdb) p *s
$8 = {raw = 0x7fffffff7f88, offset = 33, numUsedBitsInPartialByte = 5, 
  diag = 0x73fb7fe173e0, diagContext = 0x7ffee6c18e06 "Box[mini]"}
```

tests/gtest/avifstreamtest.cc
```diff
@@ -167,6 +167,32 @@ TEST(StreamTest, Roundtrip) {
...
+
+  EXPECT_TRUE(avifROStreamSkipBits(&ro_stream, 1));
```
wantehchang, line 187: Before the bug fix, we get an assertion failure in this `avifROStreamSkipBits()` call:

```
avifstreamtest: libavif/src/stream.c:57: avifROStreamSkip: Assertion `stream->numUsedBitsInPartialByte == 0' failed.
Aborted
```

tests/gtest/avifstreamtest.cc
```diff
@@ -167,6 +167,32 @@ TEST(StreamTest, Roundtrip) {
...
+  EXPECT_EQ(avifROStreamOffset(&ro_stream), 33);
+  EXPECT_EQ(ro_stream.numUsedBitsInPartialByte, 6);
+
+  EXPECT_TRUE(avifROStreamSkipBits(&ro_stream, 2));
+  EXPECT_EQ(avifROStreamOffset(&ro_stream), 33);
+  EXPECT_EQ(ro_stream.numUsedBitsInPartialByte, 0);
```
wantehchang, line 193: In this case, the stream's offset must not be incremented. It should still be 33. But the stream's `numUsedBitsInPartialByte` should become 0 rather than 8.

vigneshvg (approved):

vigneshvg:
Thank you for the fix and the explanation comments.

wantehchang:
Yannis: I will go ahead and submit this. It would be good if you could do a post-commit review. Thanks.

src/stream.c
```diff
@@ -153,6 +153,9 @@ avifBool avifROStreamSkipBits(avifROStream * stream, size_t bitCount)
         const size_t padding = AVIF_MIN(8 - stream->numUsedBitsInPartialByte, bitCount);
         stream->numUsedBitsInPartialByte = (stream->numUsedBitsInPartialByte + padding) % 8;
         bitCount -= padding;
+        if (stream->numUsedBitsInPartialByte != 0) {
```
y-guyon, line 156: I think `bitCount == 0` would be easier to understand but that is the proper fix.

y-guyon (commented):
LGTM. Thanks for the fix and the test.

src/stream.c
wantehchang: Thanks for the suggestion. It is also more efficient (avoids unnecessary work) if `stream->numUsedBitsInPartialByte == 0` and `bitCount == 0`. I already have a test for that condition in this pull request.

I switched to this test in https://github.com/AOMediaCodec/libavif/pull/2499.

PR #2505, opened 2024-11-14
Fix static libs INTERFACE_INCLUDE_DIRECTORIES

WangWeiLin-MV:
Fix issue https://github.com/microsoft/vcpkg/issues/42112, error with `fatal error: 'avif/avif.h' file not found` that target `avif` (i.e. `avif_static`) lost `INTERFACE_INCLUDE_DIRECTORIES` for static build.

Let `merge_static_libs.cmake` copy `INTERFACE_INCLUDE_DIRECTORIES` too.

### DIFF
--- a/cmake/Modules/merge_static_libs.cmake
+++ b/cmake/Modules/merge_static_libs.cmake
@@ -52,6 +52,11 @@ function(merge_static_libs target in_target)
     set(source_file ${CMAKE_CURRENT_BINARY_DIR}/${target}_depends.c)
     add_library(${target} STATIC ${source_file})
 
+    get_target_property(include_dirs ${in_target} INTERFACE_INCLUDE_DIRECTORIES)
+    if(include_dirs)
+        target_include_directories(${target} PUBLIC ${include_dirs})
+    endif()
+
     avif_collect_deps(${in_target} lib_deps)
 
     foreach(lib ${lib_deps})


vrabaud (approved):

vrabaud:
Thx !

cmake/Modules/merge_static_libs.cmake
```diff
@@ -63,6 +63,10 @@ function(merge_static_libs target in_target)
         if(link_dirs)
             target_link_directories(${target} PUBLIC ${link_dirs})
         endif()
+        get_target_property(include_dirs ${lib} INTERFACE_INCLUDE_DIRECTORIES)
```
vrabaud, line 66: That is for the deps which is not needed as avif doe snot publicly depend on other headers.
Please try the following outside this for loop instead.
```
get_target_property(include_dirs ${in_target} INTERFACE_INCLUDE_DIRECTORIES)
if(include_dirs)
  target_include_directories(${target} PUBLIC ${include_dirs})
endif()
```

WangWeiLin-MV: Thanks for the comment. Updated.

vrabaud:
@WangWeiLin-MV , does that new version of the patch actually fix the original bug ?

WangWeiLin-MV:
@vrabaud Yes, in my tests, it was generated correctly in `packages/libavif_x64-windows-static/share/libavif/libavif-config.cmake`

```cmake
add_library(avif STATIC IMPORTED)

set_target_properties(avif PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
)
```

vrabaud:
Great, thx !

PR #2420, opened 2024-08-29
Add properties to avifImage

y-guyon:

PR #2510, opened 2024-11-27
props: write out any properties set on the image

bradh:
This complements the opaque property reading done under https://github.com/AOMediaCodec/libavif/pull/2420

There are a couple of issues I'd like guidance on:

 - the AV2 CI test failure
 - whether it is OK to carry over unsupported properties when writing out a new file

An example of the unsupported property might be a `udes` that applied to the original image, but might not be appropriate in a modified version of that image. I'm unsure of the best way to handle this.


src/write.c
```diff
@@ -3090,6 +3090,18 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
             // We don't add an 'lsel' property since many decoders do not support it and will reject the image,
             // see https://github.com/AOMediaCodec/libavif/pull/2429
         }
+        for (size_t i = 0; i < itemMetadata->numProperties; i++) {
+            avifItemPropertyDedupStart(dedup);
+            avifImageItemProperty * prop = &(itemMetadata->properties[i]);
+            avifBoxMarker propMarker;
+            AVIF_CHECKRES(avifRWStreamWriteBox(&dedup->s, (const char *)prop->boxtype, AVIF_BOX_SIZE_TBD, &propMarker));
```
y-guyon, line 3097: I wonder if we should check for custom box types (for example forbid `meta`). Otherwise it would be possible to generate invalid files with the libavif API.

With or without this check, we should fuzz this new API. I think a new dedicated fuzz target would be easy. It needs to check that encoding any API-compliant custom prop garbage is a success, and that decoding the encoded output is a success when expected (that is the hard part to determine) or a graceful error.  
Maybe an approach similar to `ArbitraryAvifImageWithGainMap()` would be a good fit.

include/avif/avif.h
```diff
@@ -862,6 +862,10 @@ AVIF_API avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags p
 AVIF_API void avifImageFreePlanes(avifImage * image, avifPlanesFlags planes);           // Ignores already-freed planes
 AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, avifPlanesFlags planes);
 
+// Add arbitrary (opaque) properties to the image
```
y-guyon, line 865: There should be a comment about how to add a `FullBox` (first four bytes of `data`).

Alternatively there could be yet another function with `version` and `flags` arguments but I guess it would be overkill and it would not match the decoding API.

Note that `FullBox`es are described in `avifImageItemProperty` comments but these functions do not use the `avifImageItemProperty` struct.

include/avif/avif.h
```diff
@@ -862,6 +862,10 @@ AVIF_API avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags p
...
+AVIF_API avifResult avifImageAddOpaqueProperty(avifImage * image, const uint8_t boxtype[4], const uint8_t * data, size_t dataSize);
+AVIF_API avifResult avifImageAddUUIDProperty(avifImage * image, const uint8_t uuid[16], const uint8_t * data, size_t dataSize);
```
y-guyon, line 867: Did you consider using only one signature leveraging the existing `avifImageItemProperty` definition?

```c
AVIF_API avifResult avifImageAddProperty(avifImage * image, const avifImageItemProperty * property);
```

The downside is the need for `avifImageItemPropertyCreate()` and `avifImageItemPropertyDestroy()` so it may be worse.

Alternatively it could reuse the struct members but not the struct itself:

```c
// Appends an avifImageItemProperty to the image properties.
// The arguments have the same semantics as the fields of avifImageItemProperty.
AVIF_API avifResult avifImageAddProperty(avifImage * image, const uint8_t boxtype[4], const uint8_t usertype[16], avifROData boxPayload);
```

tests/gtest/avifpropertytest.cc
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 
 #include <cstring>
+#include <fstream>
```
y-guyon, line 5: Unused as is, please remove

tests/gtest/avifpropertytest.cc
```diff
@@ -45,6 +46,66 @@ TEST(AvifPropertyTest, Parse) {
...
+  testutil::FillImageGradient(image.get());
+
+  EncoderPtr encoder(avifEncoderCreate());
+  ASSERT_NE(encoder, nullptr);
+  testutil::AvifRwData encoded;
+  avifResult result = avifEncoderWrite(encoder.get(), image.get(), &encoded);
+  ASSERT_EQ(result, AVIF_RESULT_OK) << avifResultToString(result);
+  // Uncomment the following to save the encoded image as an AVIF file.
+  // std::ofstream("/tmp/avifproperties.avif", std::ios::binary)
+  //    .write(reinterpret_cast<char*>(encoded.data), encoded.size);
```
y-guyon, line 76: I would remove this comment

tests/gtest/avifpropertytest.cc
```diff
@@ -45,6 +46,66 @@ TEST(AvifPropertyTest, Parse) {
...
+
+  ImagePtr decoded(avifImageCreateEmpty());
+  ASSERT_NE(decoded, nullptr);
+  DecoderPtr decoder(avifDecoderCreate());
+  ASSERT_NE(decoder, nullptr);
+  result = avifDecoderReadMemory(decoder.get(), decoded.get(), encoded.data,
+                                 encoded.size);
+  ASSERT_EQ(result, AVIF_RESULT_OK) << avifResultToString(result);
+  EXPECT_EQ(decoder->imageSequenceTrackPresent, AVIF_FALSE);
```
y-guyon, line 85: Unrelated to this test, please remove

tests/gtest/avifpropertytest.cc
```diff
@@ -45,6 +46,66 @@ TEST(AvifPropertyTest, Parse) {
...
+
+  const avifImageItemProperty& uuidProp = decoder->image->properties[2];
+  EXPECT_EQ(std::string(uuidProp.boxtype, uuidProp.boxtype + 4), "uuid");
+  EXPECT_EQ(std::vector<uint8_t>(uuidProp.usertype, uuidProp.usertype + 16),
+            std::vector<uint8_t>(uuid, uuid + 16));
+  EXPECT_EQ(
+      std::vector<uint8_t>(uuidProp.boxPayload.data,
+                           uuidProp.boxPayload.data + uuidProp.boxPayload.size),
+      uuid_data);
+}
```
y-guyon, line 108: Missing blank line

It would be good to test custom props on gainmaps too.

include/avif/avif.h
bradh: I did think about that, but it seemed somewhat cleaner to have two different methods. It just makes it harder to use incorrectly.

tests/gtest/avifpropertytest.cc
bradh: Removed. I had kept it based on a similar entry in the encoder test.

tests/gtest/avifpropertytest.cc
bradh: Removed.

tests/gtest/avifpropertytest.cc
bradh: Removed.

tests/gtest/avifpropertytest.cc
bradh: I added the blank line.

I don't have enough spec information or understanding to do the gainmap work without more guidance. Should it work just the same? Just set it on `altImageMetadata` ?

include/avif/avif.h
bradh: I've added some extra usage notes.

src/write.c
bradh: I added a note to the API documentation about not adding known properties.

The other parts are still to be done.

y-guyon: > I added a note to the API documentation about not adding known properties.

"Known properties" is likely too vague as a criterion.

I would still prefer the encoder to check for validity. I can think of ways:
- Check that there is no forbidden box type (an approximate list should be rather quick to gather, for example [here](https://aomediacodec.github.io/av1-avif/#avif-boxes) up to level 2)
- Check that the box type makes sense (`[a-zA-Z0-9! ]{4}` or something)
- Check that the uuid makes sense (is that possible?)
- Check that there is no property with the same box type assigned twice to the same item (assuming nobody needs that besides `colr` which is set by libavif)
- Call `avifParse()` within `avifEncode()` when custom properties are set (a bit too much of a change maybe)

@wantehchang What do you think?

> The other parts are still to be done.

You meant in this PR or in another one?

include/avif/avif.h
```diff
@@ -862,6 +862,17 @@ AVIF_API avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags p
...
 
+// Add arbitrary (opaque) properties to the image.
+// Use of these functions for known properties and boxes (e.g. ispe or meta) will likely result in
+// invalid files, and should be avoided.
+// If creating a ItemFullProperty, the version and flags values should be provided as the first four bytes of
+// the data argument, and accounted for in the dataSize.
+AVIF_API avifResult avifImageAddOpaqueProperty(avifImage * image, const uint8_t boxtype[4], const uint8_t * data, size_t dataSize);
+// This version adds an ItemProperty (or ItemFullProperty if version and flags are provided in data argument), using
+// the user extension (uuid) mechanism, see ISO/IEC 14496-12:2022 Section 4.2. The box type is set to 'uuid' and
+// does not need to be specified.
```
y-guyon, line 873: // The box type is set to 'uuid' ~and does not need to be specified~

tests/gtest/avifpropertytest.cc
```diff
@@ -45,6 +45,63 @@ TEST(AvifPropertyTest, Parse) {
...
+  testutil::AvifRwData encoded;
+  avifResult result = avifEncoderWrite(encoder.get(), image.get(), &encoded);
+  ASSERT_EQ(result, AVIF_RESULT_OK) << avifResultToString(result);
+
+  ImagePtr decoded(avifImageCreateEmpty());
+  ASSERT_NE(decoded, nullptr);
+  DecoderPtr decoder(avifDecoderCreate());
+  ASSERT_NE(decoder, nullptr);
+  result = avifDecoderReadMemory(decoder.get(), decoded.get(), encoded.data,
+                                 encoded.size);
```
y-guyon, line 79: `avifDecoderParse()` would be enough here.

If `avifDecoderReadMemory()` is kept, it would be cleaner to check `decoded` than `decoder->image`.

tests/gtest/avifpropertytest.cc
y-guyon: `avifImageAddOpaqueProperty(image->gainMap->image)` should work for encoding.  
`image->gainMap` can be created this way:

https://github.com/AOMediaCodec/libavif/blob/378c646247cb7d286dda55b28f6451cf289383a5/tests/gtest/avifgainmaptest.cc#L113

At decoding `decoder->image->gainMap->image` should be tested too.

y-guyon (changes requested):

src/write.c
bradh: > > The other parts are still to be done.
> 
> You meant in this PR or in another one?

I mean the other parts of this comment. 

I'm doubtful about the value of checking for known boxes on encode, but will try it. My concern is that we're trying to protect programmers from themselves. They can always make junk files (e.g. by changing the results of the encode before writing it out) and we're potentially making the encode slower for  limited protection.

I would like clarification on why you only wanted up to level 2 though - if we're going to check properties, why not check for adding known properties?  Possibly that would be caught by the fourth point (no duplicates), but I'm -0 on that, because there are likely uses for that (e.g. `udes` with different languages is supported in 23008-12, and maybe similar use of that pattern in other documents).

There are pretty limited options for checking a UUID, but I could see how much hassle that is.

Instead of requiring `avifParse()`, we could suggest that as an option for users. Again, trying to limit the impact on encode performance.

include/avif/avif.h
bradh: Fixed

src/write.c
bradh: I've added argument checks and associated test code. The checks are at set time, so the failure result is closer to the problem.

I still need to add documentation about using `avifParse`.

y-guyon (changes requested):
Sorry for the churn. All these open questions are part of why I did not write the encoder part at the same time as the decoder part.

src/properties.c
```diff
@@ -0,0 +1,54 @@
...
+};
+
+static const size_t numKnownProperties = sizeof(avifKnownProperties) / sizeof(avifKnownProperties[0]);
+
+static const size_t FOURCC_BYTES = 4;
+static const size_t UUID_BYTES = 16;
+
+static const uint8_t ISO_UUID_SUFFIX[12] = { 0x00, 0x01, 0x00, 0x10, 0x80, 0x00, 0x00, 0xAA, 0x00, 0x38, 0x9b, 0x71 };
+
+avifBool avifIsKnownPropertyType(const uint8_t boxtype[4])
```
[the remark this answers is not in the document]
bradh: Would you like that here, or in the call site?

src/properties.c
```diff
@@ -0,0 +1,54 @@
...
+}
+
+avifBool avifIsValidUUID(const uint8_t uuid[16])
+{
+    for (size_t i = 0; i < numKnownProperties; i++) {
+        if ((memcmp(avifKnownProperties[i].fourcc, uuid, FOURCC_BYTES) == 0) &&
+            (memcmp(ISO_UUID_SUFFIX, uuid + FOURCC_BYTES, UUID_BYTES - FOURCC_BYTES) == 0)) {
+            return AVIF_FALSE;
+        }
+    }
```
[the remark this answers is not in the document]
bradh: Yes. Added comment on the source.

src/properties.c
```diff
@@ -0,0 +1,54 @@
...
+{
+    for (size_t i = 0; i < numKnownProperties; i++) {
+        if (memcmp(avifKnownProperties[i].fourcc, boxtype, FOURCC_BYTES) == 0) {
+            return AVIF_TRUE;
+        }
+    }
+    return AVIF_FALSE;
+}
+
+avifBool avifIsValidUUID(const uint8_t uuid[16])
```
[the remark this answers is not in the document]
bradh: Added notes.

tests/gtest/avifpropinternaltest.cc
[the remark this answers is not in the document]
bradh: Fixed. Had missed adding that to the commit.

tests/gtest/avifpropinternaltest.cc
```diff
@@ -0,0 +1,54 @@
...
+#include "avif/internal.h"
+#include "aviftest_helpers.h"
+#include "gtest/gtest.h"
+
+namespace avif {
+namespace {
+
+TEST(InternalPropertiesTest, KnownFound) {
+  const uint8_t FTYP[4]{'f', 't', 'y', 'p'};
+  ASSERT_EQ(avifIsKnownPropertyType(FTYP), AVIF_TRUE);
```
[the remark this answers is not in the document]
bradh: Fixed (multiple locations, `ASSERT_TRUE` and `ASSERT_FALSE`).

tests/gtest/avifpropinternaltest.cc
```diff
@@ -0,0 +1,54 @@
+// Copyright 2023 Google LLC
```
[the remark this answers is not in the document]
bradh: Fixed.

tests/gtest/avifpropertytest.cc
bradh: I'm not sure I understand. Did you want this change:

```
-  result = avifDecoderReadMemory(decoder.get(), decoded.get(), encoded.data,
-                                 encoded.size);
-  ASSERT_EQ(result, AVIF_RESULT_OK) << avifResultToString(result);
+  ASSERT_EQ(avifDecoderSetIOMemory(decoder.get(), encoded.data, encoded.size),
+            AVIF_RESULT_OK);
+  ASSERT_EQ(avifDecoderParse(decoder.get()), AVIF_RESULT_OK);
```
?

y-guyon: Yes.

Or:

```diff
-ASSERT_EQ(decoder->image->numProperties, 3u);
+ASSERT_EQ(decoded->numProperties, 3u);
...
-const avifImageItemProperty& abcd = decoder->image->properties[0];
+const avifImageItemProperty& abcd = decoded->properties[0];
etc.
```

src/properties.c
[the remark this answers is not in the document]
y-guyon: Good question. I think the call site makes more sense because the function is `avifIsKnownPropertyType()` and not `avifVerifyPropertyType()`.

tests/gtest/avifpropertytest.cc
bradh: Thanks. Fixed.

src/properties.c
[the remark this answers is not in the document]
bradh: Fixed.

include/avif/avif.h
```diff
@@ -824,7 +824,7 @@ typedef struct avifImage
 
     // Other properties attached to this image item (primary or gainmap).
     // At decoding: Forwarded here as opaque byte sequences by the avifDecoder.
-    // At encoding: Ignored.
+    // At encoding: Written by the avifEncoder.
```
y-guyon, line 827: Suggestion:

```diff
-// At encoding: Written by the avifEncoder.
+// At encoding: Set using avifImageAddOpaqueProperty() or avifImageAddUUIDProperty() and written by the avifEncoder.
```

tests/gtest/avifpropertytest.cc
```diff
@@ -45,6 +45,67 @@ TEST(AvifPropertyTest, Parse) {
...
+  testutil::FillImageGradient(image.get());
+
+  EncoderPtr encoder(avifEncoderCreate());
+  ASSERT_NE(encoder, nullptr);
+  testutil::AvifRwData encoded;
+  avifResult result = avifEncoderWrite(encoder.get(), image.get(), &encoded);
+  ASSERT_EQ(result, AVIF_RESULT_OK) << avifResultToString(result);
+
+  ImagePtr decoded(avifImageCreateEmpty());
+  ASSERT_NE(decoded, nullptr);
```
y-guyon, line 79: Unused

bradh: Removed.

include/avif/avif.h
bradh: Applied.

src/write.c
```diff
@@ -3090,6 +3090,18 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
             // We don't add an 'lsel' property since many decoders do not support it and will reject the image,
             // see https://github.com/AOMediaCodec/libavif/pull/2429
         }
+        for (size_t i = 0; i < itemMetadata->numProperties; i++) {
+            avifItemPropertyDedupStart(dedup);
+            avifImageItemProperty * prop = &(itemMetadata->properties[i]);
```
wantehchang, line 3095: Nit: Omit the parentheses. This kind of expression (`&foo->bar[i]`) is very common in libavif.

wantehchang, line 3095: Nit: I think `prop` can be declared as `const`: `const avifImageItemProperty * prop = ...`

include/avif/avif.h
```diff
@@ -824,7 +824,7 @@ typedef struct avifImage
 
     // Other properties attached to this image item (primary or gainmap).
     // At decoding: Forwarded here as opaque byte sequences by the avifDecoder.
-    // At encoding: Ignored.
+    // At encoding: Set using avifImageAddOpaqueProperty() or avifImageAddUUIDProperty() and written by the avifEncoder.
```
wantehchang, line 827: If the order of properties is important, we should document that these properties are written in the order in which they are added to the image, and after all the other properties generated by avifEncoder itself.

include/avif/avif.h
```diff
@@ -862,6 +862,16 @@ AVIF_API avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags p
...
 
+// Add arbitrary (opaque) properties to the image.
+// Use of these functions for known properties and boxes (e.g. ispe or meta) will likely result in
+// invalid files, and should be avoided.
+// If creating a ItemFullProperty, the version and flags values should be provided as the first four bytes of
+// the data argument, and accounted for in the dataSize.
+AVIF_API avifResult avifImageAddOpaqueProperty(avifImage * image, const uint8_t boxtype[4], const uint8_t * data, size_t dataSize);
+// This version adds an ItemProperty (or ItemFullProperty if version and flags are provided in data argument), using
+// the user extension (uuid) mechanism, see ISO/IEC 14496-12:2022 Section 4.2. The box type is set to 'uuid'.
+AVIF_API avifResult avifImageAddUUIDProperty(avifImage * image, const uint8_t uuid[16], const uint8_t * data, size_t dataSize);
```
wantehchang, line 873: It would be good to add a Note or Warning to these two functions to warn users that these are advanced features intended only for people familiar with the ISO BMFF and HEIF file formats.

wantehchang (commented):
Brad: Please consider my review as supplemental. Yannis is the primary reviewer of this pull request.

include/avif/avif.h
```diff
@@ -862,6 +862,16 @@ AVIF_API avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags p
 AVIF_API void avifImageFreePlanes(avifImage * image, avifPlanesFlags planes);           // Ignores already-freed planes
 AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, avifPlanesFlags planes);
 
+// Add arbitrary (opaque) properties to the image.
+// Use of these functions for known properties and boxes (e.g. ispe or meta) will likely result in
+// invalid files, and should be avoided.
+// If creating a ItemFullProperty, the version and flags values should be provided as the first four bytes of
```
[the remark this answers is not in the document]
bradh: Fixed.

src/properties.c
```diff
@@ -0,0 +1,68 @@
+// Copyright 2024 Brad Hards. All rights reserved.
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "avif/internal.h"
```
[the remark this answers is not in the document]
bradh: Fixed.

src/properties.c
```diff
@@ -0,0 +1,68 @@
...
+#include <string.h>
+
+struct avifKnownProperty
+{
+    const uint8_t fourcc[4];
```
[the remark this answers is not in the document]
bradh: Fixed.

src/write.c
bradh: Fixed

include/avif/avif.h
bradh: Partly fixed. I've noted the part about the order of addition.

In doing this, I realised that there is a problem with doing it after "all the other properties". ISO/IEC 23008-12:2022 Section 6.5 requires that descriptive properties come before transformative properties. So adding at the end might or might not work depending on whether the reader actually enforces the part about ignoring descriptive properties after the first transformative or unrecognized property, whichever is earlier; and whether there are any transformative properties.

Separate issue here: `a1lx` might not be written in the  correct order - not sure I'm completely following that part of the code though.

Options:
1. Document that only descriptive properties are allowed, and move the addition up to before the transformative properties.
2. Use different structures for descriptive and transformative properties, and document they are added at the end of the descriptive and transformative properties (respectively) added by avifEncoder.

I lean to Option 1. 

Suggestions on other options? Preferences on implementation?

include/avif/avif.h
bradh: Added (slightly above).

src/write.c
bradh: Fixed.

src/properties.c
```diff
@@ -0,0 +1,68 @@
...
+    { { 'n', 'd', 'w', 't' } }, { { 'a', '1', 'o', 'p' } }, { { 'l', 's', 'e', 'l' } }, { { 'a', '1', 'l', 'x' } },
+    { { 'c', 'm', 'i', 'n' } }, { { 'c', 'm', 'e', 'x' } }, { { 'i', 'p', 'm', 'a' } }, { { 'i', 'r', 'e', 'f' } },
+    { { 'a', 'u', 'x', 'l' } }, { { 't', 'h', 'm', 'b' } }, { { 'd', 'i', 'm', 'g' } }, { { 'p', 'r', 'e', 'm' } },
+    { { 'c', 'd', 's', 'c' } }, { { 'g', 'r', 'p', 'l' } }, { { 'a', 'l', 't', 'r' } }, { { 's', 't', 'e', 'r' } },
+    { { 'm', 'd', 'a', 't' } },
+};
+
+static const size_t numKnownProperties = sizeof(avifKnownProperties) / sizeof(avifKnownProperties[0]);
+
+static const size_t FOURCC_BYTES = 4;
```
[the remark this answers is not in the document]
bradh: I thought it made the code a little easier to read.

The other usages were in array bounds, which is only supported by GCC as an extension. 

I can change it to be a `#define` if that would be preferred.

src/properties.c
```diff
@@ -0,0 +1,68 @@
...
+
+#include "avif/internal.h"
+#include <string.h>
+
+struct avifKnownProperty
+{
+    const uint8_t fourcc[4];
+};
+
+static const struct avifKnownProperty avifKnownProperties[] = {
```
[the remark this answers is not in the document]
bradh: Fixed

src/write.c
bradh: > With or without this check, we should fuzz this new API. I think a new dedicated fuzz target would be easy. It needs to check that encoding any API-compliant custom prop garbage is a success, and that decoding the encoded output is a success when expected (that is the hard part to determine) or a graceful error.
> Maybe an approach similar to `ArbitraryAvifImageWithGainMap()` would be a good fit.

I can't say it was easy. However I have added the fuzzer.

bradh:
I believe I have addressed all the comments.

I'd particularly like review on the fuzzer and the gain map related code. 

I'd like feedback on the ordering issue, see https://github.com/AOMediaCodec/libavif/pull/2510#discussion_r1884688034

include/avif/avif.h
y-guyon: > In doing this, I realised that there is a problem with doing it after "all the other properties". 

Good catch!

> ISO/IEC 23008-12:2022 Section 6.5 requires that descriptive properties come before transformative properties. So adding at the end might or might not work depending on whether the reader actually enforces the part about ignoring descriptive properties after the first transformative or unrecognized property, whichever is earlier; and whether there are any transformative properties.

It says "readers shall ignore". We should not care about non-compliant implementations, but in practice I agree it is best to make sure it works everywhere. libavif probably does not check that.  
Since these properties are descriptive and should not impact the output sample values, I wonder what is the difference between ignoring them and, well, not ignoring them.

> Separate issue here: `a1lx` might not be written in the correct order - not sure I'm completely following that part of the code though.

I think https://github.com/AOMediaCodec/libavif/pull/2524 would fix that.

> Options:
> 
> 1. Document that only descriptive properties are allowed, and move the addition up to before the transformative properties.
> 2. Use different structures for descriptive and transformative properties, and document they are added at the end of the descriptive and transformative properties (respectively) added by avifEncoder.
> 
> I lean to Option 1.
> 
> Suggestions on other options? Preferences on implementation?

Assuming we go with https://github.com/AOMediaCodec/libavif/pull/2524:
- The descriptive custom props impl can be inserted in write.c between the libavif-handled descriptive properties and the libavif-handled transformative properties.
- The public API can then allow both descriptive and transformative custom props, as long as they are in this order.
- Note that the spec says "Writers should arrange the descriptive properties specified in 6.5 prior to any other properties", so HEIF props come first, before non-HEIF descriptive props. 
  - This may be too much of a detail for the user to respect. I would avoid mentioning that in avif.h.
  - `a1lx` is specified in [av1-avif](https://aomediacodec.github.io/av1-avif/), not in HEIF. So if a user adds a HEIF property that is not handled by libavif, and `a1lx` is written, this constraint will not be fulfilled. I think we can ignore this constraint because it is fairly niche and only a "should". A comment explaining this part would be nice (not necessarily in the header though).

y-guyon (changes requested):
Thank you for adding the gainmap test and fuzz test! The PR looks good, just a few nits.

include/avif/avif.h
```diff
@@ -853,6 +854,19 @@ AVIF_API avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags p
 AVIF_API void avifImageFreePlanes(avifImage * image, avifPlanesFlags planes);           // Ignores already-freed planes
 AVIF_API void avifImageStealPlanes(avifImage * dstImage, avifImage * srcImage, avifPlanesFlags planes);
 
+// Add arbitrary (opaque) properties to the image.
+// Note: This is an advanced usage, intended for users with specific requirements who are familiar with the
+// HEIF and ISO BMFF standards. Use of these functions for known properties and boxes (e.g. ispe or meta)
```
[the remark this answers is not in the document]
bradh: Fixed.

tests/gtest/avif_fuzztest_properties.cc
```diff
@@ -0,0 +1,115 @@
+// Copyright 2022 Google LLC
```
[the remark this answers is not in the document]
bradh: Fixed.

tests/gtest/avif_fuzztest_properties.cc
```diff
@@ -0,0 +1,115 @@
...
+  for (TestProp testProp : testProps) {
+    if (testProp.fourcc == UUID_4CC) {
+      uint8_t uuid[16];
+      for (int i = 0; i < 16; i++) {
+        uuid[i] = testProp.uuid[i];
+      }
+      ASSERT_EQ(
+          avifImageAddUUIDProperty(image.get(), uuid, testProp.body.data(),
+                                   testProp.body.size()),
+          AVIF_RESULT_OK);
```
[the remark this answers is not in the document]
bradh: Fixed.

I thought the MSVC compiler objected to this, but appears to be OK in CI.

include/avif/avif.h
bradh: > * The descriptive custom props impl can be inserted in write.c between the libavif-handled descriptive properties and the libavif-handled transformative properties.
> 
> * The public API can then allow both descriptive and transformative custom props, as long as they are in this order.

That isn't going to work for all cases, because `ispe` has to be the first transformative property (see ISO/IEC 23008-12:2022 Section 6.5.3.1).

src/read.c
```diff
@@ -5938,6 +5938,17 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
+            for (size_t i = 0; i < gainMapProperties->count; ++i) {
+                const avifProperty * property = &gainMapProperties->prop[i];
+                if (property->isOpaque) {
+                    AVIF_CHECKRES(avifImagePushProperty(decoder->image->gainMap->image,
+                                                        property->type,
+                                                        property->u.opaque.usertype,
+                                                        property->u.opaque.boxPayload.data,
+                                                        property->u.opaque.boxPayload.size));
+                }
+            }
```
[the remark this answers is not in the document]
bradh: Fixed.

include/avif/avif.h
bradh: It would probably be easier if #2524 went in before trying to update this PR to reflect that.

y-guyon: Done!

y-guyon: I also filed https://github.com/AOMediaCodec/libavif/issues/2525.

bradh: Thanks. I've rebased against main branch, which predictably caused merge conflicts. I've moved up the opaque property writing to be before transformative properties (as suggested), and added documentation about the ordering.

y-guyon (approved):

bradh:
@y-guyon Thanks for all the help on this one - much appreciated.

y-guyon:
> @y-guyon Thanks for all the help on this one - much appreciated.

At least as many thanks to you then!

PR #2511, opened 2024-11-28
Turn gain maps on in ci-unix-shared-installed.yml

vrabaud:
The default values for the install fix are taken from: https://cmake.org/cmake/help/v3.31/command/install.html#signatures


maryla-uc (approved):
Thank you!!
This ensures that we'll be able to turn on the gain map code by default.

PR #2513, opened 2024-12-02
Turn gain map flag on by default.

maryla-uc:
Remove 'EXPERIMENTAL' from its name.


wantehchang (commented):
Thanks for the pull request.

Can we remove the `AVIF_ENABLE_GAIN_MAP` option?

Remove `EXPERIMENTAL` from `AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION`

.github/workflows/ci-fuzztest.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON
           -DAVIF_BUILD_TESTS=ON -DAVIF_ENABLE_GTEST=ON -DAVIF_GTEST=LOCAL
           -DAVIF_ENABLE_EXPERIMENTAL_YCGCO_R=ON
-          -DAVIF_ENABLE_EXPERIMENTAL_GAIN_MAP=ON
+          -DAVIF_ENABLE_GAIN_MAP=ON
```
wantehchang, line 53: Should we delete this (and similar lines in other *.yml files) since this is the default?

CMakeLists.txt
```diff
@@ -272,11 +273,11 @@ if(AVIF_LIBSHARPYUV_ENABLED)
 endif(AVIF_LIBSHARPYUV_ENABLED)
 
 set_local_or_system_option(
-    "LIBXML2" "OFF" "Build libxml2 by providing your own copy inside the ext subdir. \
-libxml2 is used when AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP is ON"
+    "LIBXML2" "OFF" "Use libxml2. When enabled and AVIF_ENABLE_GAIN_MAP is ON, \
+    allows converting jpeg files with gain maps to avif using avifenc"
```
wantehchang, line 277: Nit: Use "JPEG" and "AVIF" (capitalized) in help messages.

tests/CMakeLists.txt
```diff
@@ -271,7 +271,7 @@ if(AVIF_BUILD_APPS)
                                               ${CMAKE_CURRENT_SOURCE_DIR}/data
     )
 
-    if(AVIF_ENABLE_AVIFGAINMAPUTIL AND AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION)
+    if(AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION)
```
wantehchang, line 274: IMPORTANT: It seems that this should still be guarded by `AVIF_ENABLE_GAIN_MAP`.

Please try building with `-DAVIF_ENABLE_GAIN_MAP=OFF -DAVIF_BUILD_APPS=ON`.

tests/CMakeLists.txt
```diff
@@ -223,7 +223,7 @@ if(AVIF_ENABLE_FUZZTEST)
     add_avif_fuzztest(avif_fuzztest_read_image)
     add_avif_fuzztest(avif_fuzztest_yuvrgb)
 
-    if(AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP)
+    if(AVIF_ENABLE_GAIN_MAP)
```
wantehchang, line 226: I wonder if we should remove `_experimental` from these two fuzztest names. Note: Simply removing `_experimental` will conflict with other fuzztest names. Need to look into this.

tests/data/README.md
```diff
@@ -643,7 +643,7 @@ libavif and ignored by the tests.
 License: [same as libavif](https://github.com/AOMediaCodec/libavif/blob/main/LICENSE)
 
 Source : created from `seine_sdr_gainmap_srgb.jpg` converted to avif with avifenc built
-with AVIF_ENABLE_EXPERIMENTAL_GAIN_MAP and AVIF_ENABLE_LIBXML2 then:
+with AVIF_ENABLE_GAIN_MAP and AVIF_ENABLE_LIBXML2 then:
```
wantehchang, line 646: Nit: `AVIF_ENABLE_LIBXML2` no longer exists. I guess it has been renamed. Please look into this.

wantehchang (commented):
I read the whole pull request. Other than whether `AVIF_ENABLE_GAIN_MAP` should be kept or removed, it looks good. Here are some more comments.

tests/CMakeLists.txt
wantehchang: I understand why this is correct now. `AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION` implies `AVIF_ENABLE_GAIN_MAP`, so it is not necessary to test `AVIF_ENABLE_GAIN_MAP` here.

.github/workflows/ci-fuzztest.yml
maryla-uc: We can, I left it for now in case we wanted to easily revert the default to OFF. I kept the changes in this PR to the minimum: changing default and renaming the constant. But we can definitely make more changes.

tests/CMakeLists.txt
maryla-uc: The fuzzers can indeed be merged but this can be done in another PR I think.

maryla-uc:
I kept the AVIF_ENABLE_GAIN_MAP option for now just in case we needed to revert it to OFF for some reason, and to keep this PR simple (even though the rename means that there are still quite a few changes).
I'm happy to remove it though. It can be done in another PR.

maryla-uc:
As for AVIF_ENABLE_EXPERIMENTAL_JPEG_GAIN_MAP_CONVERSION, since it's still off by default, I kept the "experimental" name for now. Again, can be changed later.

Issue #2514, opened 2024-12-02
The libavif encoder adds the MA1B or MA1A brand incorrectly


wantehchang:
In libavif/src/write.c, the code that adds the MA1B or MA1A brand is as follows:

```
    if ((imageMetadata->depth == 8) || (imageMetadata->depth == 10)) {     //
        if (imageMetadata->yuvFormat == AVIF_PIXEL_FORMAT_YUV420) {        //
            AVIF_CHECKRES(avifRWStreamWriteChars(&s, "MA1B", 4));          // ... compatible_brands[]
        } else if (imageMetadata->yuvFormat == AVIF_PIXEL_FORMAT_YUV444) { //
            AVIF_CHECKRES(avifRWStreamWriteChars(&s, "MA1A", 4));          // ... compatible_brands[]
        }
    }
```

This code is incorrect because it does not check the AV1 profile or the AV1 level of the AV1 bitstreams.

wantehchang:
The attached patch should fix the code that adds the MA1B brand:

```
diff --git a/src/write.c b/src/write.c
index cf4e1c79..52829208 100644
--- a/src/write.c
+++ b/src/write.c
@@ -3134,6 +3134,7 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
     // -----------------------------------------------------------------------
     // Harvest configuration properties from sequence headers
 
+    avifBool isMA1B = AVIF_TRUE;
     for (uint32_t itemIndex = 0; itemIndex < encoder->data->items.count; ++itemIndex) {
         avifEncoderItem * item = &encoder->data->items.item[itemIndex];
         if (item->encodeOutput->samples.count > 0) {
@@ -3142,6 +3143,11 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
             AVIF_CHECKERR(avifSequenceHeaderParse(&sequenceHeader, (const avifROData *)&firstSample->data, codecType),
                           avifGetErrorForItemCategory(item->itemCategory));
             item->av1C = sequenceHeader.av1C;
+            //  The MA1B brand: The AV1 profile shall be the Main Profile and
+            //  the level shall be 5.1 or lower.
+            if (item->av1C.seqProfile != 0 || item->av1C.seqLevelIdx0 > 13) {
+                isMA1B = AVIF_FALSE;
+            }
         }
     }
 
[...]

PR #2515, opened 2024-12-02
Write the MA1B brand correctly

wantehchang:
A partial fix for https://github.com/AOMediaCodec/libavif/issues/2514.


wantehchang (commented):
Yannis: To help code review, I split the pull request into two commits. Commit 1 is the actual change. Commit 2 is clang-format only.

src/write.c
```diff
@@ -3142,6 +3143,11 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
             AVIF_CHECKERR(avifSequenceHeaderParse(&sequenceHeader, (const avifROData *)&firstSample->data, codecType),
                           avifGetErrorForItemCategory(item->itemCategory));
             item->av1C = sequenceHeader.av1C;
+            // The MA1B brand: The AV1 profile shall be the Main Profile and the level shall be 5.1
+            // or lower.
+            if (item->av1C.seqProfile != 0 || item->av1C.seqLevelIdx0 > 13) {
+                isMA1B = AVIF_FALSE;
+            }
```
wantehchang, line 3150: Yannis: I am not that familiar with this code. It is important that this for loop (which starts at line 3138) is executed for both image items and image sequences, and that we enter the `if (item->encodeOutput->samples.count > 0)` block (which starts at line 3140) at least once. Otherwise we will incorrectly report that `isMA1B` is true (the initial value). Please check this. Thanks!

src/write.c
```diff
@@ -3226,24 +3232,25 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
...
+        AVIF_CHECKRES(avifRWStreamWriteChars(&s, "iso8", 4));          // ... compatible_brands[]
+    }                                                                  //
+    AVIF_CHECKRES(avifRWStreamWriteChars(&s, "mif1", 4));              // ... compatible_brands[]
+    AVIF_CHECKRES(avifRWStreamWriteChars(&s, "miaf", 4));              // ... compatible_brands[]
+    if (isMA1B) {                                                      //
+        AVIF_CHECKRES(avifRWStreamWriteChars(&s, "MA1B", 4));          // ... compatible_brands[]
+    }                                                                  //
+    if ((imageMetadata->depth == 8) || (imageMetadata->depth == 10)) { //
+        if (imageMetadata->yuvFormat == AVIF_PIXEL_FORMAT_YUV444) {    //
+            AVIF_CHECKRES(avifRWStreamWriteChars(&s, "MA1A", 4));      // ... compatible_brands[]
```
y-guyon, line 3253: Are the constraints listed in https://aomediacodec.github.io/av1-avif/v1.1.0.html#advanced-profile always true here?  
Note that they differ for items and sequences.

src/write.c
y-guyon: Here is my understanding:
- A track exists in `avifEncoder` only if the corresponding item is present too, likely with the first frame of the sequence being used as the image item data.
- There is no change of profile/level within a single sequence.

Assuming this is correct, the for loop and the `MA1B` check look good.

PR #2520, opened 2024-12-13
Change the way brew python is updated

vrabaud:


.github/actions/setup-macos/action.yml
```diff
@@ -22,17 +22,14 @@ outputs:
...
+    - name: Force update all brew python
       run: |
-        find /usr/local/bin -lname '*/Library/Frameworks/Python.framework/2to3*' \
-          -o -lname '*/Library/Frameworks/Python.framework/idle3*' \
-          -o -lname '*/Library/Frameworks/Python.framework/pydoc3*' \
-          -o -lname '*/Library/Frameworks/Python.framework/python3*' -delete
-        sudo rm -rf /Library/Frameworks/Python.framework/
-        brew install --force python3 && brew unlink python3 && brew unlink python3 && brew link --overwrite python3
+        brew update
+        # GHA messes with links https://github.com/actions/setup-python/issues/577 so we just force replace their installs' with brew's
```
wantehchang, line 28: Nit: Delete the `'` after "installs"

wantehchang: Hmm... Perhaps it should be changed to `their install's`?

wantehchang: Or perhaps it is short for `their installations' links`? I guess we can leave it unchanged.

PR #2526, opened 2024-12-20
Fail early when trying to convert a jpeg+gainmap image without libxml2.

maryla-uc:


vrabaud (approved):

apps/avifgainmaputil/convert_command.cc
```diff
@@ -31,6 +31,12 @@ ConvertCommand::ConvertCommand()
 }
 
 avifResult ConvertCommand::Run() {
+#if !defined(AVIF_ENABLE_JPEG_GAIN_MAP_CONVERSION)
+  std::cout << "JPEG gainmap conversion unavailable because avifgainmaputil "
+               "was not built with libxml2.\n";
+  return AVIF_RESULT_NOT_IMPLEMENTED;
+#endif
```
wantehchang, line 38: I suggest enclosing the rest of the function in a `#else` block. Otherwise, some compilers may warn that the code afterthe return statement (line 37) is not reachable.

PR #2527, opened 2024-12-20
Remove AVIF_ENABLE_GTEST cmake option.

maryla-uc:
There is no need to have two AVIF_ENABLE_GTEST and AVIF_GTEST options.
Now setting AVIF_GTEST=LOCAL or AVIF_GTEST=SYSTEM automatically turns ont gtest tests.


CMakeLists.txt
```diff
@@ -127,8 +127,12 @@ cmake_dependent_option(
 
 set_local_or_system_option(
     "GTEST" OFF
-    "Build the GoogleTest framework by providing your own copy of the repo in ext/googletest (see Local Builds in README)"
+    "Use the GoogleTest framework. Enables avif C++ tests that depend on GoogleTest. Has no effect unless AVIF_BUILD_TESTS is ON."
 )
+if(NOT AVIF_GTEST STREQUAL "OFF")
+    set(AVIF_ENABLE_GTEST ON)
```
vrabaud, line 133: no need to define an extra variable anymore: just replace the two instances of it below with AVIF_GTEST. Same in tests/CMakeLists.txt

maryla-uc: I though it was easier like this (and fewer changes) than checking `NOT AVIF_GTEST STREQUAL "OFF"` everywhere, no? Or does `if(AVIF_GTEST)` work? (I didn't think so since it's a string). I'm happy to change it though if you think it's better.

vrabaud: Hmm, I thought it would work with `"OFF"`, https://cmake.org/cmake/help/latest/command/if.html#constant but maybe you can try and see if it works. If not, if it is equal to `"OFF"`, just set it to `OFF`

maryla-uc: That does seem to work. TIL!

README.md
```diff
@@ -96,9 +96,8 @@ first run the associated script in the `ext/` subdirectory.
 A few tests written in C can be built by enabling the `AVIF_BUILD_TESTS` CMake
 option.
 
-The remaining tests can be built by enabling the `AVIF_BUILD_TESTS` and
-`AVIF_ENABLE_GTEST` CMake options. They require GoogleTest
-(`-DAVIF_GTEST=SYSTEM` or `-DAVIF_GTEST=LOCAL`).
+The remaining tests can be built by enabling `AVIF_BUILD_TESTS` and setting
+`AVIF_GTEST` to `SYSTEM` or `LOCAL`.
```
wantehchang, line 100: Optional: We can consider keeping the comment "They require GoogleTest."

CMakeLists.txt
```diff
@@ -572,9 +570,7 @@ if(NOT AVIF_CODEC_AOM_ENABLED
     message(WARNING "libavif: No decoding library is enabled.")
 endif()
 
-if(AVIF_LIB_USE_CXX OR AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST
-                                                                                                         OR AVIF_ENABLE_GTEST))
-)
+if(AVIF_LIB_USE_CXX OR AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_ENABLE_FUZZTEST OR AVIF_GTEST)))
```
wantehchang, line 573: Do you think `AVIF_ENABLE_FUZZTEST` can receive the same treatment?

wantehchang (commented):
Good cleanup!

CMakeLists.txt
miwucs: Yes I'm working on it.

vrabaud: Indeed, we talked about it but I wanted to add more stuff at the same time.
I fixed it in https://github.com/AOMediaCodec/libavif/pull/2529 (and fuzztest is now downloadable from CMake just like any other dependency)

PR #2528, opened 2024-12-20
Move the comment on 'pasp' forward

wantehchang:
It was added to the middle of a paragraph. It should go before that paragraph.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -3037,11 +3037,12 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
                               AVIF_RESULT_NOT_IMPLEMENTED);
             }
 
-            // Based on the explanation above, 'clap', 'irot' and 'imir' have to match between the base and
-            // gain map image items in the container part of the encoded file.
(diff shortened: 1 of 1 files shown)


src/write.c
```diff
@@ -3037,11 +3037,12 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
...
             }
 
-            // Based on the explanation above, 'clap', 'irot' and 'imir' have to match between the base and
-            // gain map image items in the container part of the encoded file.
             // 'pasp' is not a transformative property (despite AVIF_TRANSFORM_PASP being part of
             // avifTransformFlag) but it is assumed to apply to the gain map in the same way as
             // the transformative properties above.
+
+            // Based on the explanation above, 'clap', 'irot', 'imir' and 'pasp' have to match between the base and
+            // gain map image items in the container part of the encoded file.
```
wantehchang, line 3045: Note that I also added `'pasp'` to line 3044.

y-guyon (approved):

PR #2529, opened 2024-12-23
Fuzztest clean-ups

vrabaud:
- merge AVIF_ENABLE_FUZZTEST and AVIF_LOCAL_FUZZTEST
- allow for fuzztest be built from CMake


CMakeLists.txt
```diff
@@ -142,9 +142,6 @@ option(
     "Build tests that compare encoding outputs to golden files. Needs AVIF_BUILD_APPS=ON and AVIF_BUILD_TESTS=ON, and depends on MP4box which can be built with ext/mp4box.sh"
     OFF
 )
-option(AVIF_ENABLE_FUZZTEST "Build avif fuzztest targets. Requires Google FuzzTest. Has no effect unless AVIF_BUILD_TESTS is ON."
-       OFF
-)
```
wantehchang, line 147: I am wondering why you didn't use the `AVIF_FUZZTEST=SYSTEM/LOCAL/OFF` convention for FuzzTest.

vrabaud: `SYSTEM` is most likely not available and even if it were, our dependency on fuzztest is so weak that it's better we handle it ourselves with `LOCAL` only: we know our dependency works with a particular sha1+patches. I did not try with more recent versions (as long as the CMake issues are not fixed, there is no point in trying anything newer).

wantehchang: The new meaning of `AVIF_LOCAL_FUZZTEST` is subtly different from the meaning of the deprecated `AVIF_LOCAL_XXX` options. `AVIF_LOCAL_XXX` is only used when XXX is enabled, and `AVIF_LOCAL_XXX` toggles between "local XXX" and "system XXX":

```
function(set_local_or_system_option VAR DEFAULT TEXT)
    # Deal with the older way of setting options.
    if(DEFINED AVIF_LOCAL_${VAR})
        if(AVIF_LOCAL_${VAR})
            set(DEFAULT "LOCAL")
        else()
            set(DEFAULT "SYSTEM")
        endif()
        message(WARNING "Setting AVIF_LOCAL_${VAR} is deprecated. " "Set AVIF_${VAR} to ${DEFAULT} instead.")
```

After this CL, `AVIF_LOCAL_FUZZTEST` will toggle between "local FuzzTest" and "FuzzTest disabled".

It would be better if we could support `AVIF_FUZZTEST=LOCAL/OFF` instead. We can disallow setting it to `SYSTEM`. This would still be closer to the `AVIF_XXX=SYSTEM/LOCAL/OFF` convention.

wantehchang (commented):
Vincent: This is a very incomplete review. I mainly wanted to respond to the discussion on `AVIF_FUZZTEST=SYSTEM/LOCAL/OFF`.

cmake/Modules/LocalFuzztest.cmake
```diff
@@ -0,0 +1,32 @@
...
+
+if(EXISTS "${FUZZTEST_SOURCE_DIR}")
+    message(STATUS "libavif(AVIF_LOCAL_FUZZTEST=ON): folder found at ${FUZZTEST_SOURCE_DIR}")
+    set(FUZZTEST_BINARY_DIR "${FUZZTEST_SOURCE_DIR}/build.libavif")
+else()
+    message(STATUS "libavif(AVIF_LIBYUV=LOCAL): compiled library not found at ${LIB_FILENAME}; using FetchContent")
+
+    message(CHECK_START "libavif(AVIF_LOCAL_FUZZTEST=ON): configuring fuzztest")
+
+    set(FUZZTEST_SOURCE_DIR "${FETCHCONTENT_BASE_DIR}/fuzztest-src")
```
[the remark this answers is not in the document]
vrabaud: It makes the code shorter and this variable is actually used in avif_fetchcontent_populate_cmake called at the end of that script.

cmake/Modules/LocalFuzztest.cmake
```diff
@@ -0,0 +1,32 @@
+set(AVIF_LOCAL_FUZZTEST_TAG "078ea0871cc96d3a69bad406577f176a4fa14ae9")
+
+set(FUZZTEST_SOURCE_DIR "${AVIF_SOURCE_DIR}/ext/fuzztest")
+
+if(EXISTS "${FUZZTEST_SOURCE_DIR}")
+    message(STATUS "libavif(AVIF_LOCAL_FUZZTEST=ON): folder found at ${FUZZTEST_SOURCE_DIR}")
+    set(FUZZTEST_BINARY_DIR "${FUZZTEST_SOURCE_DIR}/build.libavif")
+else()
+    message(STATUS "libavif(AVIF_LIBYUV=LOCAL): compiled library not found at ${LIB_FILENAME}; using FetchContent")
```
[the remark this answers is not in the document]
vrabaud: Done

tests/CMakeLists.txt
```diff
@@ -169,42 +169,35 @@ endif()
...
 
-if(AVIF_ENABLE_FUZZTEST)
+if(AVIF_LOCAL_FUZZTEST)
     # Adds a fuzztest from file TEST_NAME.cc located in the gtest folder. Extra arguments
     # are considered as extra source files.
     macro(add_avif_fuzztest TEST_NAME)
         add_executable(${TEST_NAME} gtest/${TEST_NAME}.cc ${ARGN})
         # FuzzTest bundles GoogleTest so no need to link to gtest librairies.
-        target_link_libraries(${TEST_NAME} PRIVATE avif_fuzztest_helpers aviftest_helpers_internal avif_enable_warnings)
+        target_link_libraries(${TEST_NAME} PRIVATE avif_fuzztest_helpers aviftest_helpers_internal)
```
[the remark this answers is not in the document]
vrabaud: Yes definitely: I'll try a newer version of fuzztest at some point to see if they are gone.

CMakeLists.txt
vrabaud: Mostly, the old way of setting dependencies should be removed :)
Done

tests/CMakeLists.txt
```diff
@@ -170,42 +170,38 @@ endif()
...
 
-if(AVIF_ENABLE_FUZZTEST)
+if(AVIF_FUZZTEST)
     # Adds a fuzztest from file TEST_NAME.cc located in the gtest folder. Extra arguments
     # are considered as extra source files.
     macro(add_avif_fuzztest TEST_NAME)
         add_executable(${TEST_NAME} gtest/${TEST_NAME}.cc ${ARGN})
         # FuzzTest bundles GoogleTest so no need to link to gtest librairies.
-        target_link_libraries(${TEST_NAME} PRIVATE avif_fuzztest_helpers aviftest_helpers_internal avif_enable_warnings)
+        target_link_libraries(${TEST_NAME} PRIVATE avif_fuzztest_helpers aviftest_helpers_internal)
```
wantehchang, line 179: Nit: Add a comment to explain why `avif_enable_warnings` is not here.

wantehchang (approved):

tests/CMakeLists.txt
```diff
@@ -170,42 +170,39 @@ endif()
...
 
-if(AVIF_ENABLE_FUZZTEST)
+if(AVIF_FUZZTEST)
     # Adds a fuzztest from file TEST_NAME.cc located in the gtest folder. Extra arguments
     # are considered as extra source files.
     macro(add_avif_fuzztest TEST_NAME)
         add_executable(${TEST_NAME} gtest/${TEST_NAME}.cc ${ARGN})
         # FuzzTest bundles GoogleTest so no need to link to gtest librairies.
-        target_link_libraries(${TEST_NAME} PRIVATE avif_fuzztest_helpers aviftest_helpers_internal avif_enable_warnings)
+        # avif_enable_warnings is not added because it triggers to many warnings in fuzztest.
```
wantehchang, line 179: Typo: to => too

Issue #2476, opened 2024-10-14
Add a comment to avifCheckItemID() about the source of the nonzero item_ID requirement


wantehchang:
@maryla-uc @vigneshvg @vrabaud @y-guyon FYI.

In src/read.c, the `avifCheckItemID()` function requires that the value of `item_ID` be nonzero. We should add a comment about the source of this requirement.

I am having trouble finding this requirement in ISOBMFF, HEIF, and MIAF. The closest thing I can find is the following paragraph in ISO/IEC 14496-12:2022 (ISOBMFF), Clause 8.11.1.1 on the definition of the MetaBox:

> Derived specifications may further restrict the criteria for uniqueness: unique among the `item_IDs` in
> both file and movie-level boxes, or unique within that set extended with the `track_ID` of the tracks in a
> movie box. The `item_ID` value of 0 should not be used, and shall not be used when the set is extended to
> include `track_IDs`.

But the last sentence uses "should not" in the initial, general part of the statement, and uses "shall not" only in the more specific part of the statement.

PR #2539, opened 2025-01-02
Add item ID 0 comment in avifCheckItemID()

y-guyon:
Fixes https://github.com/AOMediaCodec/libavif/issues/2476.


maryla-uc (approved):

wantehchang (commented):
LGTM. Thanks!

PR #2541, opened 2025-01-02
Add code coverage CI.

maryla-uc:
Add a cmake option.
Document in readme.

After this is merged I will delete the seriously outdated instructions in the wiki https://github.com/AOMediaCodec/libavif/wiki/Coverage

### DIFF
--- a/.github/workflows/ci-linux-coverage.yml
+++ b/.github/workflows/ci-linux-coverage.yml
@@ -1,0 +1,62 @@
+# Workflow that builds libabvif with aom and dav1d, runs tests and generates coverage report.
+
+name: CI Linux Code Coverage
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+# Cancel the workflow if a new one is triggered from the same PR, branch, or tag, except on main.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  build-shared-run-golden-tests:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-24.04]
+
+    steps:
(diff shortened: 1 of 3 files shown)


maryla-uc:
Note the CI is not strictly required, but it makes sure the coverage stuff actually works, and uploads the coverage report as a github artifact that can be downloaded e.g. here https://github.com/AOMediaCodec/libavif/actions/runs/12584719944?pr=2541

vrabaud:
We could also upload those to [https://app.codecov.io/ .](https://about.codecov.io/tool/github-actions/) and have a little badge on top of our README.

maryla-uc:
Indeed, that looks nice! I think it can be done as a second step (in another PR).

maryla-uc:
I tried setting up codecov.io but I think you need to be a member of the AOMediaCodec organization?
Basically you need to log into codecov.io with your github account, and it should show a list of github repos you can configure. But in my case it only shows my own fork of libavif (among other forks I have), not AOMediaCodec/libavif. @wantehchang would you have the necessary permissions to set it up? See their quick start guide: https://docs.codecov.com/docs/quick-start

wantehchang:
Maryla: I am not sure if I am a member of the AOMediaCodec organization. https://github.com/AOMediaCodec says "This organization has no public members. You must be a member to see who’s a part of this organization." I cannot view the "Members" in https://github.com/orgs/AOMediaCodec/people. I guess this implies I am not a member?

wantehchang:
Maryla: I logged into codecov.io with my github account. I also can only configure my personal forks of projects. Note that I selected "personal use" when my codecov.io account was created. Not sure if that matters.

maryla-uc:
Thanks for trying it. I see the same as you on the AOMediaCodec page, so I guess none of us are members.

jzern:
We got to the point where we could try to configure the AOMediaCodec organization (https://github.com/apps/codecov/installations/select_target), but trying to add a repository fails because no one is an organization owner. @agrange are you an owner of https://github.com/AOMediaCodec?

PR #2548, opened 2025-01-03
avifenc: make alpha quality be the same as the color quality by default.

maryla-uc:
If only the color quality is set, alpha quality is automatically set to the same value.

The old behavior was alpha quality set to 100 (lossless) but this generates very large file and is not needed in most cases.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -807,7 +807,6 @@ static avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gr
 
 #define INVALID_QUALITY (-1)
 #define DEFAULT_QUALITY 60 // Maps to a quantizer (QP) of 25.
-#define DEFAULT_QUALITY_ALPHA AVIF_QUALITY_LOSSLESS
 #define DEFAULT_QUALITY_GAIN_MAP DEFAULT_QUALITY
 #define PROGRESSIVE_WORST_QUALITY 10 // Not doing auto automatic layered encoding below this quality
 #define PROGRESSIVE_START_QUALITY 2  // First layer use this quality
@@ -925,12 +924,6 @@ static avifBool avifEncodeRestOfImageSequence(avifEncoder * encoder,
     while ((nextFile = avifInputGetFile(input, imageIndex)) != NULL) {
         uint64_t nextDurationInTimescales = nextFile->duration ? nextFile->duration : settings->outputTiming.duration;
 
-        printf(" * Encoding frame %d [%" PRIu64 "/%" PRIu64 " ts]: %s\n",
-               imageIndex,
-               nextDurationInTimescales,
-               settings->outputTiming.timescale,
-               nextFile->filename);
-
         if (nextImage) {
             avifImageDestroy(nextImage);
         }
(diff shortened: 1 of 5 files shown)


y-guyon (approved):

apps/avifenc.c
```diff
@@ -1898,6 +1906,15 @@ int main(int argc, char * argv[])
...
 
+    // Alpha quality defaults to the same value as (color) quality until the first time it's explicitly set.
+    for (int i = 0; i < input.filesCount; ++i) {
+        avifInputFileSettings * fileSettings = &input.files[i].settings;
+        if (fileSettings->qualityAlpha.set) {
+            break;
+        }
+        fileSettings->qualityAlpha = fileSettings->quality;
+    }
+
```
wantehchang, line 1917: Maryla: Thank you for making this change! I tried to review it but found that the code in avifenc that handles the command-line options has become very complicated.

I am wondering if the deprecated --minalpha and --maxalpha options should also default to the deprecated --min and --max options. But I can't figure out the default values for these four options.

maryla-uc: I agree the code is quite complex!
I double checked and this change does work for --min --max too. If --minalpha and --maxalpha are not specified, then the alpha quality will be the same as the color quality. This can be seen in `test_cmd.sh`:

```shell
  "${AVIFENC}" -s 10 --min 15 --max 25 "${INPUT_Y4M}" "${ENCODED_FILE}" > "${OUT_MSG}"
  grep " color quality \[68 " "${OUT_MSG}"
  grep " alpha quality \[68 " "${OUT_MSG}"
```

wantehchang: Thanks for looking into it. With `--min 15 --max 25`, we will use their midpoint 20 as the quantizer, and convert the quantizer to quality with the formula `quality = ((63 - quantizer) * 100 + 31) / 63`. So we will get quality 68, which matches your test.

y-guyon:
Should it be the same behavior with the libavif API and not only `avifenc`?

maryla-uc:
The goal of this PR was to no longer have lossles alpha by default.
The API never defaulted to lossless alpha. In the API, alpha quality defaults to `-1` (AVIF_QUALITY_DEFAULT) which produces a medium quality. Color also defaults to this medium quality, so by default they have the same value. But if one is changed by the API user and not the other, then they will no longer have the same value.
We could change the API so that `if (qualityAlpha == -1) qualityAlpha = quality`
But I'm not sure if it's ok to change the API behavior in this way.
@wantehchang do you have an opinion?

wantehchang:
Whether changing the default value of an option breaks backward compatibility is a gray area. I just checked avif.h and the default value of `qualityAlpha` is not documented. This means no API users can deliberately depend on the documented default value of `qualityAlpha`. So it should be fine to change the default value of `qualityAlpha`.

It is a gray area because when an API user leaves an option unset, it is ambiguous whether
- the user knows what the default value is and assumes the option has that default value, or
- the user doesn't care about the option.

If it is the former reason, changing the default value of the option may break the user.

If it is the latter reason, we can change the default value of the option because the user doesn't care about it.

maryla-uc:
Thanks for your input, that makes sense! In the case of alpha quality, when users are not setting it, I doubt that "some default medium value" is what they really want, especially if they are changing the color quality. It seems more likely that they were not aware of this other quality setting or did not think about images with alpha.
I think it makes sense to change the behavior here.

wantehchang:
Maryla: I agree with your analysis.

PR #2549, opened 2025-01-06
Remove LOCAL from logs as this was part of the old convention

vrabaud:


maryla-uc (approved):

cmake/Modules/LocalGTest.cmake
```diff
@@ -27,15 +27,15 @@ else()
     if(EXISTS "${AVIF_SOURCE_DIR}/ext/googletest")
         message(STATUS "libavif(AVIF_GTEST=LOCAL): ext/googletest found; using as FetchContent SOURCE_DIR")
         set(FETCHCONTENT_SOURCE_DIR_GOOGLETEST "${AVIF_SOURCE_DIR}/ext/googletest")
-        message(CHECK_START "libavif(AVIF_LOCAL_GTEST): configuring googletest")
+        message(CHECK_START "libavif(AVIF_GTEST): configuring googletest")
```
wantehchang, line 30: Should we change `AVIF_LOCAL_GTEST` to `AVIF_GTEST=LOCAL`? See lines 26 and 28.

vrabaud: Done in https://github.com/AOMediaCodec/libavif/pull/2587

PR #2553, opened 2025-01-06
Bump fuzztest

vrabaud:
This is to get better minimizers and error messages. The build status should be checked closely in the next few days.


ext/fuzztest.cmd
```diff
@@ -9,8 +9,8 @@
 
 git clone https://github.com/google/fuzztest.git
 cd fuzztest
-: # There is no tagged release as of 2024/06/21. Pick the latest commit that works.
-git checkout 078ea0871cc96d3a69bad406577f176a4fa14ae9
+: # There is no tagged release as of 2024/01/06. Pick the latest commit that works.
```
y-guyon, line 12: 2025

y-guyon (approved):

ext/fuzztest.cmd
vrabaud: fixed

PR #2557, opened 2025-01-06
Update to SVT-AV1 v2.3.0

wantehchang:

PR #2561, opened 2025-01-07
Use CMake's compiler when checking for static link bundling

vrabaud:
This should fix issues seen in https://github.com/AOMediaCodec/libavif/pull/2557

### DIFF
--- a/.github/workflows/ci-unix-static-av2.yml
+++ b/.github/workflows/ci-unix-static-av2.yml
@@ -61,7 +61,7 @@ jobs:
         run: ctest -j $(getconf _NPROCESSORS_ONLN) --output-on-failure
       - name: Check static link bundling
         run: |
-          cc -o avifenc  -I./apps/shared -I./third_party/iccjpeg -I./include apps/avifenc.c \
+          $CC -o avifenc  -I./apps/shared -I./third_party/iccjpeg -I./include apps/avifenc.c \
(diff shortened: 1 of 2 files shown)


y-guyon (approved):

.github/workflows/ci-unix-static-av2.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         run: ctest -j $(getconf _NPROCESSORS_ONLN) --output-on-failure
       - name: Check static link bundling
         run: |
-          cc -o avifenc  -I./apps/shared -I./third_party/iccjpeg -I./include apps/avifenc.c \
+          $CC -o avifenc  -I./apps/shared -I./third_party/iccjpeg -I./include apps/avifenc.c \
```
wantehchang, line 64: (This comment is just a discussion, not a request for change.)

Note the use of the C compiler `$CC` and the `-lstdc++` linker option. This may not be portable because there may be other C++ runtime code we need to link with to support the C++ code in build/libavif.a. Ideally we should compile the .c file into .o files, and then use the C++ compiler `$CXX` to link all the .o files and build/libavif.a into the avifenc executable.

wantehchang, line 64: Vincent: Thanks for the fix. I am curious how you figured out this is the problem.

vrabaud: Simple: on OSX 14, this cannot run because cc does not exist :)

PR #2562, opened 2025-01-07
Set CC on macOS

vrabaud:
This seems necessary on some runners ...

### DIFF
--- a/.github/workflows/ci-unix-static-av2.yml
+++ b/.github/workflows/ci-unix-static-av2.yml
@@ -59,9 +59,14 @@ jobs:
       - name: Run AVIF Tests
         working-directory: ./build
         run: ctest -j $(getconf _NPROCESSORS_ONLN) --output-on-failure
+      - name: Set CC on macOS
+        if: ${{ runner.os == 'macOS' }}
+        run:
+          echo "CC=cc" >> $GITHUB_ENV
+        shell: bash
       - name: Check static link bundling
         run: |
-          $CC -o avifenc  -I./apps/shared -I./third_party/iccjpeg -I./include apps/avifenc.c \
+          ${{ env.CC }} -o avifenc  -I./apps/shared -I./third_party/iccjpeg -I./include apps/avifenc.c \
             apps/shared/*.c third_party/iccjpeg/iccjpeg.c build/libavif.a \
             -lpng -ljpeg -lz -lm -ldl -lstdc++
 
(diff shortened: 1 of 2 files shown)


y-guyon (approved):

.github/workflows/ci-unix-static-av2.yml
```diff
@@ -59,9 +59,14 @@ jobs:
       - name: Run AVIF Tests
         working-directory: ./build
         run: ctest -j $(getconf _NPROCESSORS_ONLN) --output-on-failure
+      - name: Set CC on macOS
+        if: ${{ runner.os == 'macOS' }}
+        run:
+          echo "CC=cc" >> $GITHUB_ENV
```
wantehchang, line 65: This solution seems a little ad hoc. I wonder if we should add a testing CMakeLists.txt file that links a simple program with the combined libavif archive library. This will allow us to replace the command at lines 69-71 with a cmake command.

vrabaud: This is exactly what we do when we build the lib as static in this CI: avif becomes avif_static and the tests use it. I'd say this test is useless.

wantehchang: This is the command line used to link `avifcolrtest` with the libavif static library:

> [943/949] : && /usr/bin/c++ -g  tests/CMakeFiles/aviftest_helpers_internal.dir/gtest/aviftest_helpers.cc.o tests/CMakeFiles/avifcolrtest.dir/gtest/avifcolrtest.cc.o -o tests/avifcolrtest  lib/libgtest.a  lib/libgtest_main.a  libavif_apps.a  /usr/lib/x86_64-linux-gnu/libpng.so  /usr/lib/x86_64-linux-gnu/libz.so  /usr/lib/x86_64-linux-gnu/libjpeg.so  libavif_internal.a  _deps/libyuv-build/libyuv.a  _deps/libwebp/libsharpyuv.a  _deps/dav1d-install/lib/libdav1d.a  -ldl  _deps/libgav1-build/libgav1.a  _deps/svt-build/libSvtAv1Enc.a  _deps/aom-build/libaom.a  -lm  _deps/libyuv-build/libyuv.a  lib/libgtest.a && :

I don't know if libavif_internal.a is the combined archive library. Even if it is, we are linking with libdav1d.a, libgav1.a, libSvtAv1Enc.a, libaom.a, etc. So this command line fails to verify that the combined archive library contains not only libavif but also its AV1 codec dependencies.

PR #2563, opened 2025-01-07
Switch macos-13 to macos-latest in CI

vrabaud:
The CI on macos-13 has been slower to trigger, probably due to it being phased out (macos-15 is out now).

### DIFF
--- a/.github/workflows/ci-unix-static.yml
+++ b/.github/workflows/ci-unix-static.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-24.04, macos-13]
+        os: [ubuntu-24.04, macos-latest]
         build-type: [Release, Debug]
         include:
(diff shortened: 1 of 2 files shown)


y-guyon (approved):

.github/workflows/ci-unix-static.yml
```diff
@@ -78,6 +78,7 @@ jobs:
       - name: Check static link bundling
         run: |
           ${{ env.CC }} -o avifenc  -I./apps/shared -I./third_party/iccjpeg -I./include apps/avifenc.c \
+            -I/opt/homebrew/include/ -L/opt/homebrew/lib \
```
wantehchang, line 81: One way to avoid this is to build a simple test program instead of avifenc.c. Then we don't need to compile `apps/shared/*.c third_party/iccjpeg/iccjpeg.c` and link with -lpng -ljpeg -lz.

PR #2566, opened 2025-01-08
Check FuzzTest WithSize() actual size

y-guyon:
Because of the recent false positive possibly attributed to FuzzTest domains not respecting constraints.

Can be reverted when no longer useful.


wantehchang:
The [FuzzTest bug fix](https://github.com/google/fuzztest/commit/8fdc787e8c64066619b69f3ee9d168c00aeddab2) mentions the `Filter()` function. avif_fuzztest_properties.cc is the only test that uses `Filter()`. So only avif_fuzztest_properties.cc needs to check this.

wantehchang:
I just saw the FuzzTest update in https://github.com/AOMediaCodec/libavif/pull/2565. This pull request should be unnecessary now.

PR #2569, opened 2025-01-13
Fix some static analysis minor defects

vrabaud:
- fix comparison of uint64_t with SIZE_MAX for platforms where uint64_t and size_t are different
- do not compare two floats for equality


src/gainmap.c
```diff
@@ -45,12 +45,13 @@ static float avifGetGainMapWeight(float hdrHeadroom, const avifGainMap * gainMap
...
     const float alternateHdrHeadroom = avifUnsignedFractionToFloat(gainMap->alternateHdrHeadroom);
-    if (baseHdrHeadroom == alternateHdrHeadroom) {
+    const float diffHdrHeadroom = alternateHdrHeadroom - baseHdrHeadroom;
+    if (diffHdrHeadroom == 0.f) {
         // Do not apply the gain map if the HDR headroom is the same.
         // This case is not handled in the specification and does not make practical sense.
         return 0.0f;
     }
-    const float w = AVIF_CLAMP((hdrHeadroom - baseHdrHeadroom) / (alternateHdrHeadroom - baseHdrHeadroom), 0.0f, 1.0f);
+    const float w = AVIF_CLAMP((hdrHeadroom - baseHdrHeadroom) / diffHdrHeadroom, 0.0f, 1.0f);
```
wantehchang, line 54: I think this Coverity defect (464731 [Division or modulo by float zero](https://scan9.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_DIVIDE_BY_ZERO)) is a Coverity bug. We don't need to change our code for a Coverity bug if the current code is clearer. We can classify the Coverity defect as a false positive and ignore it in the Coverity UI.

Have you verified (with a Coverity analysis) that this change fixes the Coverity defect?

wantehchang (approved):
Thanks for the fixes.  I suggest a change to the `UINT64_MAX != SIZE_MAX` tests.

PR #2570, opened 2025-01-14
Move entries from Changed to Added in CHANGELOG.md

y-guyon:
It is not very useful to keep these added/changed/deleted categories in my opinion. Could we just have one section per version?


maryla-uc (approved):

wantehchang (approved):

wantehchang:
Re: the Added/Changed/Removed categories: I did a web search for "changelog style". It seems that Joe followed a common convention. For example,

- https://keepachangelog.com/en/1.1.0/
- https://common-changelog.org/

y-guyon:
> Re: the Added/Changed/Removed categories: I did a web search for "changelog style". It seems that Joe followed a common convention. For example,
> 
> * https://keepachangelog.com/en/1.1.0/
> * https://common-changelog.org/

Thank you for looking into it.

Issue #2525, opened 2024-12-20
Ignore descriptive properties associated after non-descriptive ones


y-guyon:
Section 6.5.1 of HEIF says:

> Readers shall allow and ignore descriptive properties following the first transformative or unrecognized property, whichever is earlier, in the sequence associating properties with an item.

- I do not think libavif respects that constraint.
- I am not sure what is the definition of an "unrecognized property".
- I guess "ignoring a descriptive property" means treating it like it was not there.  
  Because descriptive properties should not impact the rendering of the decoded image, the effect of ignoring one can only be:
  - treating the file as invalid from a missing mandatory descriptive property (such as `ispe`),
  - not exposing the same image metadata from the libavif API.

bradh:
I think there could be a distinction between ignoring a property at the application level and ignoring it at the library level. Or maybe this needs to go to MPEG?

y-guyon:
> I think there could be a distinction between ignoring a property at the application level and ignoring it at the library level.

Do you have another meaning in mind than "ignore as if absent"? Otherwise I do not really see the app/lib distinction.

bradh:
> > I think there could be a distinction between ignoring a property at the application level and ignoring it at the library level.
> 
> Do you have another meaning in mind than "ignore as if absent"? Otherwise I do not really see the app/lib distinction.

I do. I'm thinking about the geospatial case as an example, where the library might treat those properties as opaque. An application that wants to put imagery on a map can use the geospatial-related properties. An application that doesn't care about the spatial (or temporal) aspects can ignore the properties.  Same for special production information, or Dublin Core type properties.  

In those cases, the library isn't ignoring the descriptive properties after the first one it doesn't recognise. It can pull out whatever properties are useful at the library level, and pass the rest up. 

By contrast, an application is actually ignoring the properties.

Still seems like a non-useful requirement though. 

Side rabbit-hole topic: this might impact on the `mini` property ordering. I don't have a current copy, but it would be worth making sure that the non-HDR properties come first.

y-guyon:
>   * not exposing the same image metadata from the libavif API.

Actually, as described in https://github.com/MPEGGroup/FileFormat/issues/111#issuecomment-2589369192, all unrecognized non-essential properties can be exposed from the libavif API, as long as they are associated before a transformative property.

y-guyon:
[...]

PR #2571, opened 2025-01-14
Ignore descr props after transf props

y-guyon:
Also reject non-essential transformative properties.

Fixes https://github.com/AOMediaCodec/libavif/issues/2525.


src/read.c
```diff
@@ -2731,15 +2731,21 @@ static avifResult avifParseItemPropertyAssociation(avifMeta * meta, const uint8_
...
+        avifBool transformativePropertySeen = AVIF_FALSE;
         for (uint8_t associationIndex = 0; associationIndex < associationCount; ++associationIndex) {
             uint8_t essential;
             AVIF_CHECKERR(avifROStreamReadBitsU8(&s, &essential, /*bitCount=*/1), AVIF_RESULT_BMFF_PARSE_FAILED); // bit(1) essential;
             uint32_t propertyIndex;
             AVIF_CHECKERR(avifROStreamReadBitsU32(&s, &propertyIndex, /*bitCount=*/propertyIndexIsU15 ? 15 : 7),
                           AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(7/15) property_index;
 
+            // ISO/IEC 14496-12 Section 8.11.14.3:
+            //   0 indicating that no property is associated (the essential indicator shall also be 0)
```
wantehchang, line 2743: [This is a question, not a request for change.]

It seems that an association with `property_index=0` is useless and can be omitted. Do you know what it is used for?

src/read.c
```diff
@@ -2756,6 +2762,23 @@ static avifResult avifParseItemPropertyAssociation(avifMeta * meta, const uint8_
             // Copy property to item
             const avifProperty * srcProp = &meta->properties.prop[propertyIndex];
 
+            // ISO/IEC 23000-22 Section 6.5.1:
```
wantehchang, line 2765: 6.5.1 is not in ISO/IEC 23000-22:2019 (the [current published version](https://www.iso.org/standard/74417.html)).

This requirement is found in ISO/IEC 23000-22:2019/Amd. 2:2021 Section 7.3.9. Note the amendment 2.

Include the version/edition number of ISO/IEC 23008-12 that you are quoting from.

Note: In line 2818 you have Section 7.3.9, which is correct:

```
                        // MIAF: Section 7.3.9: "All transformative properties associated with coded and derived
                        //                      images shall be marked as essential"
```

src/read.c
```diff
@@ -2756,6 +2762,23 @@ static avifResult avifParseItemPropertyAssociation(avifMeta * meta, const uint8_
...
+            //   All transformative properties associated with coded and derived images shall be marked as essential,
+            //   and shall be from the set defined in 7.3.6.7 or the applicable MIAF profile. No other essential
+            //   transformative property shall be associated with such images.
+            const avifBool isTransformative = !memcmp(srcProp->type, "clap", 4) || !memcmp(srcProp->type, "irot", 4) ||
+                                              !memcmp(srcProp->type, "imir", 4);
+            // ISO/IEC 23008-12 Section 3.1.29:
```
wantehchang, line 2771: 3.1.29 => 3.1.28

Alternatively, include the version/edition number of ISO/IEC 23008-12.

In ISO/IEC 23008-12:2022 (the [current published version](https://www.iso.org/standard/83650.html)), the term "item property" is defined in 3.1.28.

src/read.c
```diff
@@ -2756,6 +2762,23 @@ static avifResult avifParseItemPropertyAssociation(avifMeta * meta, const uint8_
...
+            //   item property: descriptive or transformative information
+            const avifBool isDescriptive = !isTransformative;
+            // ISO/IEC 23008-12 Section 7.3.9 :
```
wantehchang, line 2774: 7.3.9 => 6.5.1

Alternatively, include the version/edition number of ISO/IEC 23008-12.

In ISO/IEC 23008-12:2022 (the [current published version](https://www.iso.org/standard/83650.html)), this sentence is in 6.5.1.

src/read.c
```diff
@@ -2790,7 +2813,15 @@ static avifResult avifParseItemPropertyAssociation(avifMeta * meta, const uint8_
                         "a1op",
 
                         // HEIF: Section 6.5.11.1: "essential shall be equal to 1 for an 'lsel' item property."
-                        "lsel"
+                        "lsel",
+
+                        // MIAF: Section 7.3.9: "All transformative properties associated with coded and derived
+                        //                      images shall be marked as essential"
```
wantehchang, line 2819: We probably should add the version/edition number "2019/Amd. 2:2021" after "MIAF". Note the amendment 2.

In the original ISO/IEC 23000-22:2019 Section 7.3.9, this reads:

> All transformative properties associated with coded and derived images required or conditionally required by this document shall be marked as essential

src/read.c
```diff
@@ -2809,8 +2840,23 @@ static avifResult avifParseItemPropertyAssociation(avifMeta * meta, const uint8_
...
+                }
             } else {
+                AVIF_ASSERT_OR_RETURN(!isTransformative);
+
                 if (essential) {
+                    // ISO/IEC 23008-12 Section 10.2.1:
+                    //   Under any brand, the primary item (or an alternative if alternative support is required)
+                    //   shall be processable by a reader implementing only the required features of that brand.
+                    //   Specifically, given that each brand has a set of properties that a reader is required to
+                    //   support: the item shall not have properties that are marked as essential and are outside
```
wantehchang, line 2855: [This should be fixed in HEIF.]

The colon (:) after "support" should be a comma (,).

src/read.c
```diff
@@ -2809,8 +2840,23 @@ static avifResult avifParseItemPropertyAssociation(avifMeta * meta, const uint8_
...
+                    //   this set.
+                    // Assuming this rule also applies to items whose primary item depends on (such as the cells
```
wantehchang, line 2857: Nit: Assuming => Assume ?

Nit: The word "whose" sounds strange here. Should it be changed to "the"?

src/read.c
```diff
@@ -2809,8 +2840,23 @@ static avifResult avifParseItemPropertyAssociation(avifMeta * meta, const uint8_
                 avifProperty * dstProp = (avifProperty *)avifArrayPush(&item->properties);
                 AVIF_CHECKERR(dstProp != NULL, AVIF_RESULT_OUT_OF_MEMORY);
                 *dstProp = *srcProp;
+
+                if (isTransformative) {
+                    transformativePropertySeen = AVIF_TRUE;
+                }
```
wantehchang, line 2846: Optional: I think it is clearer to move this if statement (lines 2844-2846) to the end of the inner for loop's body, between line 2873 and line 2874. This makes it clear that this if statement is executed in every iteration of the inner for loop.

wantehchang (approved):
I checked the texts quoted from the standards. Some of the section numbers are wrong. Perhaps you are using unpublished drafts of the new versions of the standards.

src/read.c
y-guyon: No but I could imagine the same purpose as a `FreeBox`: lazy writer implementation with a fixed number of associations that are just toggled on or off.

src/read.c
y-guyon: I think I just mistakenly swapped the section numbers of HEIF and MIAF, see https://github.com/AOMediaCodec/libavif/pull/2571#discussion_r1915750730.

src/read.c
wantehchang: This depends on whether section/clause numbers may change in a new version of the standard. Do you know if these standards try to keep their section/clause numbers fixed?

CHANGELOG.md
```diff
@@ -55,6 +55,8 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Deprecate avifCropRectConvertCleanApertureBox() and
   avifCleanApertureBoxConvertCropRect(). Replace them with
   avifCropRectFromCleanApertureBox() and avifCleanApertureBoxFromCropRect().
+* Ignore descriptive properties associated after transformative ones.
+* Reject non-essential transformative properties.
```
wantehchang, line 59: @farindk FYI.

We should audit the AVIF encoders in libavif and libheif and make sure they generate an `ItemPropertyAssociationBox` that meet the requirements checked in this pull request:

- All transformative properties (clap, irot, imir) associated with coded and derived images shall be marked as essential.
- If `property_index` is 0, `essential` shall also be 0.
- Should descriptive properties be listed before transformative properties in the `ItemPropertyAssociationBox`? Otherwise they will be ignored by readers.

farindk: > All transformative properties (clap, irot, imir) associated with coded and derived images shall be marked as essential.

This should always be the case in libheif.

> If property_index is 0, essential shall also be 0.

libheif never generates property_index=0. Is there any potential use case for this that I have missed?

> Should descriptive properties be listed before transformative properties in the ItemPropertyAssociationBox? Otherwise they will be ignored by readers.

I will have to check this: https://github.com/strukturag/libheif/issues/1443

wantehchang: Dirk: Thanks for the quick reply!

> > If property_index is 0, essential shall also be 0.
> 
> libheif never generates property_index=0. Is there any potential use case for this that I have missed?

I also asked this question. Please see Yannis's reply in https://github.com/AOMediaCodec/libavif/pull/2571#discussion_r1916243157.

> > Should descriptive properties be listed before transformative properties in the ItemPropertyAssociationBox? Otherwise they will be ignored by readers.
> 
> I will have to check this: [strukturag/libheif#1443](https://github.com/strukturag/libheif/issues/1443) BTW: what is the rationale for this rule? I cannot think about any case where it would be important to have the descriptive items before the transformative items.

This comes from the following in ISO/IEC 23008-12:2022 Section 6.5.1:

> Readers shall allow and ignore descriptive properties following the first transformative or unrecognized property, whichever is earlier, in the sequence associating properties with an item.

Note that it says "allow and ignore", so it isn't a hard rule. I don't know what is the rationale is.

farindk: > > Readers shall allow and ignore descriptive properties following the first transformative or unrecognized property, whichever is earlier, in the sequence associating properties with an item.
> 
> Note that it says "allow and ignore", so it isn't a hard rule. I don't know what is the rationale is.

But if a reader may ignore properties that are in the "wrong order", it effectively forces the writer to use the "correct order".

I also think that "Readers shall allow and ignore descriptive properties following the first transformative or unrecognized property" does not work well in practice because when encoders add new descriptive properties that are unknown to old readers, readers may ignore all descriptive properties following that new property, even if the new property was not essential (but a successive old property may be essential). I.e. images may be decoded differently by adding a non-essential property.

I thought that this is what the `essential` flag is for. When the reader sees an essential property that it does not understand, it knows that the decoded output might be incorrect. Non-essential properties can be ignored.
The separation of transformative and descriptive properties on-the-other hand makes little sense to me. Even more so since that means that the set of transformative properties may never be extended in future standards as old decoders have no way to classify them as transformative.

tests/data/README.md
```diff
@@ -35,6 +35,27 @@ License: [same as libavif](https://github.com/AOMediaCodec/libavif/blob/main/LIC
 Source: `avifenc circle-trns-after-plte.png` with custom properties added in
 `avifRWStreamWriteProperties()`: FullBox `1234`, Box `abcd` and Box `uuid`.
 
+### File [clap_irot_imir_non_essential.avif](clap_irot_imir_non_essential.avif)
```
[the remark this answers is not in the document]
y-guyon: > Here is the Compliance Warden report. Does it meet your expectations?

Yes. Thank you for checking.

> The only issue I see is that Compliance Warden should use the Amendment 2 version of [miaf][Rule #25] Section 7.3.9.

Section 7.3.9 is part of w18260 "Revised text of ISO/IEC FDIS 23000-22 Multi-Image Application Format". Are you saying Amd2 was published since then and there is no need to refer to an FDIS output document?

tests/data/README.md
```diff
@@ -35,6 +35,27 @@ License: [same as libavif](https://github.com/AOMediaCodec/libavif/blob/main/LIC
...
+
+![](clap_irot_imir_non_essential.avif)
+
+License: [same as libavif](https://github.com/AOMediaCodec/libavif/blob/main/LICENSE)
+
+Source: File generated with `TransformTest.ClapIrotImir` from
+`aviftransformtest.cc`, where `clap`, `irot` and `imir` are tagged as
+non-essential.
+
+### File [clop_irot_imor.avif](clop_irot_imor.avif)
```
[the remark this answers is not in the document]
y-guyon: > Here is the Compliance Warden report. It reports no errors. Does this meet your expectations?

Yes.

CHANGELOG.md
y-guyon: > > > Readers shall allow and ignore descriptive properties following the first transformative or unrecognized property, whichever is earlier, in the sequence associating properties with an item.
> >
> > Note that it says "allow and ignore", so it isn't a hard rule. I don't know what is the rationale is.
> 
> But if a reader may ignore properties that are in the "wrong order", it effectively forces the writer to use the "correct order".

It is not "may" but "shall ignore". To me it is a hard rule.

> I.e. images may be decoded differently by adding a non-essential property.

I disagree: MIAF* forbids non-essential transformative properties, and descriptive properties _should not_ impact decoding.

*and maybe HEIF? that is unclear to me because it depends on profiles if I remember correctly.

farindk: The `colr` box, for example, is a descriptive property. Still, it affects how the image will be displayed.
If the file now contains a descriptive property that the reader does not understand before the `colr`, the rule asks the reader to drop the `colr` box and show the image with wrong colors.

You may have the objection against the above because HEIF also says that "Writers should arrange the descriptive properties specified in 6.5 prior to any other properties in the sequence associating properties with an item.".
However, with every standard version, the set of properties in 6.5 is extended. Thus, we might have this situation:

A writer with HEIF-v2 knowledge writes this:
- `mdcv` - added in HEIF-v2
- `colr`  -  from HEIF-v1

A reader with only HEIF-v1 knowledge sees this as:
- `mdcv` - unknown property
- `colr` - known, but I have to drop it since it follows an unknown property

tests/data/README.md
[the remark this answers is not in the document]
wantehchang: Section 7.3.9 is in both ISO/IEC 23000-22:2019 and ISO/IEC 23000-22:2019/Amd. 2:2021.

In ISO/IEC 23000-22:2019, the paragraph reads:

> All transformative properties associated with coded and derived images required or conditionally required by this document shall be marked as essential, and shall be from the set that are permitted by this document or the applicable profile. No other essential transformative property shall be associated with such images.

In ISO/IEC 23000-22:2019/Amd. 2:2021, the paragraph is modified and reads:

> All transformative properties associated with coded and derived images shall be marked as essential,
> and shall be from the set defined in 7.3.6.7 or the applicable MIAF profile. No other essential
> transformative property shall be associated with such images.

So Compliance Warden quotes the older version of the paragraph in the error message. It is better to quote the newer version of the paragraph in the error message.

In general Compliance Warden should not refer to an FDIS output document.

CHANGELOG.md
wantehchang: Dirk: I agree with your analysis.

Yannis: Should we clarify in the source code comment that the following is what we actually do?

> Readers shall allow and ignore descriptive properties following the first transformative in the sequence associating properties with an item.

Although there is already a clarifying comment:

```
// No need to check for unrecognized properties as they cannot be transformative according to MIAF.
```

it explains why we don't need to check for unrecognized transformative properties, but it doesn't explain why we don't need to check for unrecognized descriptive properties.

y-guyon: > The `colr` box, for example, is a descriptive property. Still, it affects how the image will be displayed. If the file now contains a descriptive property that the reader does not understand before the `colr`, the rule asks the reader to drop the `colr` box and show the image with wrong colors.

Should `colr` be a transformative property then? Changing that would break a lot of files and implementations.

> You may have the objection against the above because HEIF also says that "Writers should arrange the descriptive properties specified in 6.5 prior to any other properties in the sequence associating properties with an item.". However, with every standard version, the set of properties in 6.5 is extended. Thus, we might have this situation:
> 
> A writer with HEIF-v2 knowledge writes this:
> 
> * `mdcv` - added in HEIF-v2
> * `colr`  -  from HEIF-v1
> 
> A reader with only HEIF-v1 knowledge sees this as:
> 
> * `mdcv` - unknown property
> * `colr` - known, but I have to drop it since it follows an unknown property

I filed a GitHub issue for MPEG.

wantehchang: `colr` should not be a descriptive property. It probably should be marked as essential. Since `colr` is widely supported, whether it is marked as essential or not doesn't really matter in practice.

farindk: Yes, this also seems to be non-consistent. HEIF says in 6.5.1 that "Descriptive properties are non-essential". But in the next section 6.5.2, it says that the decoder configuration properties are descriptive and essential.

According to HEIF, `colr` is not essential, but in all files I looked at, it was marked as essential.

In fact, many of the descriptive properties affect how the decoded image should look like. Also boxes like `pasp`. Thus, we might need some clarification what `essential` really means.

wantehchang: Dirk: Do you have the access to view https://github.com/MPEGGroup/FileFormat/issues/113? The following problematic requirment in ISO/IEC 23008-12:2022 (HEIF) Clause 6.5.1 will be removed in Amd2 of HEIF:
> Readers shall allow and ignore descriptive properties following the first transformative or unrecognized property, whichever is earlier, in the sequence associating properties with an item.

farindk: Yes, I know https://github.com/MPEGGroup/FileFormat/issues/113, but the point of my last comment is a bit different:

if all properties that influence how a decoded image will look like, should be marked `essential`, then more properties should be essential. Not just `colr` (which is usually marked `essential` despite the standard telling otherwise), but also, for example `pasp`, which might need a stretching at the decoder to be displayed in the correct aspect ratio.

y-guyon: My understanding:
- Transformative properties may impact the decoded output buffer (the pixel values). That is why MIAF mandates them as essential. Maybe HEIF should do that too but currently it explicitly mentions that they can be optionally applied by a player.
- Descriptive properties may affect how the decoded output buffer is rendered, but not change the pixel values within the buffer.
- Non-impacting, informational metadata is better located in Exif/XMP.

> Yes, this also seems to be non-consistent. HEIF says in 6.5.1 that "Descriptive properties are non-essential, unless stated otherwise in their specification".

There is also a note:

_"NOTE Each property descriptor definition below specifies whether the property is mandatory, i.e. whether the property is required to be present in the file. Note that when a property is mandatory, it might or might not be essential."_

My understanding:
- `essential=1` means the player has to understand and apply the property.
- A mandatory property for files means the whole file is invalid if the property is not in it. It may or may not be `essential`, that is orthogonal.
- A mandatory property for players means the player has to understand it (which makes `essential` only mean apply the property).

So it is somewhat clear to me. I agree it could be constrained further to remove the possibility of non-essential `colr`, which makes rendering non-deterministic depending on the player's choice.  
This is aligned with the permissive spirit of HEIF in my opinion. I suggest filing an issue to have all render-impacting descriptive properties be essential in MIAF, much like transformative properties.

Issue #2572, opened 2025-01-15
Multiple options are ignored with --stdin ("Trailing options with update suffix has no effect") using --output


gitoss:
This is an extension of https://github.com/AOMediaCodec/libavif/issues/2504 - but as a newer ticket because the problem is much more broader.

I guess the cause of this is the change to "The following options can optionally have a :u (or :update) suffix like `-q:u Q`, to apply only to input files appearing after the option:"

As a result, all affected options (--qcolor, --qalpha, ... --advanced) are ignored when there is no input file at all - i.e. --stdin. To reproduce:

1. Get some animated input file like https://jpegxl.info/images/anim-icos.gif
 
2a. ffmpeg -loglevel quiet -i anim-icos.gif -pix_fmt yuv444p -f yuv4mpegpipe anim-icos.y4m
2b. avifenc --codec aom --speed 4 --qcolor 90 --advanced deltaq-mode=4 anim-icos.y4m --output anim-icos.avif
```
Successfully loaded: anim-icos.y4m
AVIF to be written: (Lossy)
 * Resolution     : 128x128
 * Bit Depth      : 8
 * Format         : YUV444
 * Alpha          : Absent
 * Range          : Limited
 * Color Primaries: 1
 * Transfer Char. : 13
 * Matrix Coeffs. : 6
 * ICC Profile    : Absent
 * XMP Metadata   : Absent
 * Exif Metadata  : Absent
 * Transformations: None
 * Progressive    : Unavailable
 * Gain map       : Absent
Encoding with initial settings: codec 'aom' speed [4], color quality [90 (High)], alpha quality [90 (High)], tileRowsLog2 [0], tileColsLog2 [0], 12 worker thread(s), please wait...
 * Encoding frame 1 [1/20 ts] color quality [90 (High)], alpha quality [90 (High)]: anim-icos.y4m
```
=> please note: if you change to deltaq-mode=6 there's an error from aom

3. ffmpeg -loglevel quiet -i anim-icos.gif -pix_fmt yuv444p -f yuv4mpegpipe - | avifenc --codec aom --qcolor 90 --advanced deltaq-mode=4 --stdin --output anim-icos.avif

```
**WARNING: Trailing options with update suffix has no effect. Place them before the input you intend to apply to.**
AVIF to be written: (Lossy)
 * Resolution     : 128x128
 * Bit Depth      : 8
 * Format         : YUV444
 * Alpha          : Absent
 * Range          : Limited
[...]

PR #2574, opened 2025-01-17
Treat avifenc --stdin as regular file input arg

y-guyon:
Consider --stdin to be a positional argument as if it was the name of the y4m file path containing the y4m standard input contents.

Fixes https://github.com/AOMediaCodec/libavif/issues/2572.


tests/test_cmd_stdin.sh
```diff
@@ -0,0 +1,112 @@
...
+  # WARNING: Trailing options with update suffix has no effect. Place them before the input you intend to apply to.
+  "${AVIFENC}" -s 10 --stdin "${ENCODED_FILE_STDIN}" -q:u 90 < "${INPUT_Y4M}"
+  cmp --silent "${ENCODED_FILE_REGULAR}" "${ENCODED_FILE_STDIN}" && exit 1
+
+  # ERROR: there cannot be any other input if --stdin is specified
+  "${AVIFENC}" --stdin && exit 1
+  "${AVIFENC}" --stdin "${ENCODED_FILE_STDIN}" "${ENCODED_FILE_STDIN}" && exit 1
+  "${AVIFENC}" "${ENCODED_FILE_STDIN}" --stdin "${ENCODED_FILE_STDIN}" && exit 1
+  "${AVIFENC}" "${ENCODED_FILE_STDIN}" "${ENCODED_FILE_STDIN}" --stdin && exit 1
+
```
wantehchang, line 99: There is trailing whitespace in this line.

wantehchang:
@tongyuantongyu Yuan: Could you also take a look at this pull request if you have time? Thanks.

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ The changes are relative to the previous release, unless the baseline is specifi
   avifCropRectFromCleanApertureBox() and avifCleanApertureBoxFromCropRect().
 * Ignore descriptive properties associated after transformative ones.
 * Reject non-essential transformative properties.
+* Treat avifenc --stdin as a regular positional file path argument.
```
wantehchang, line 60: I haven't read the pull request. I have a question: where is the position of this fictitious file path argument?

Here is my expectation:

- If the output file path is specified without using the `-o` option, the fictitious input file path argument is right before the output file path argument.
- If the `-o` option is used, the fictitious input file path argument is at the end of the command line.

Is this what you implemented?

tongyuantongyu:
> @tongyuantongyu Yuan: Could you also take a look at this pull request if you have time? Thanks.

LGTM, though I have a small question: while we are there, can we support `-` as an alias of `--stdin`?

wantehchang (approved):
LGTM. Thanks a lot for the fix.

@tongyuantongyu Yuan: Thank you for the review!

apps/avifenc.c
```diff
@@ -215,7 +222,7 @@ static void syntaxLong(void)
     printf("                                        For JPEG, auto honors the JPEG's internal format, if possible. For grayscale PNG, auto defaults to 400. For all other cases, auto defaults to 444\n");
     printf("    -p,--premultiply                  : Premultiply color by the alpha channel and signal this in the AVIF\n");
     printf("    --sharpyuv                        : Use sharp RGB to YUV420 conversion (if supported). Ignored for y4m or if output is not 420.\n");
-    printf("    --stdin                           : Read y4m frames from stdin instead of files; no input filenames allowed, must set before offering output filename\n");
+    printf("    --stdin                           : Read y4m frames from stdin instead of files; no input filenames allowed\n");
```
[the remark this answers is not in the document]
y-guyon: > Just wondering: why is "must set before offering output filename" removed?

Because it is allowed.

> The new code allows `--stdin` to be set after the output filename, so "must set before offering output filename" is incorrect for the new code. Does the original code also allow `--stdin` to be set after the output filename?

I did not try it but I do not see how it could detect that.  
The only place where `--stdin` is parsed from the command line arguments just flips a boolean:

https://github.com/AOMediaCodec/libavif/blob/7ae69dbea82f8dc9775e1678d4fa69877ec16066/apps/avifenc.c#L1475-L1477

which is not checked until all other arguments were parsed.

> Note: When `--stdin` is treated as a regular positional file path argument, it seems cleaner if we disallow `output.avif ... --stdin`. But if the original code allows `output.avif ... --stdin`, the new code will need to allow it.

I was thinking about these cases:
```sh
avifenc --stdin -c out.avif < in.y4m # Explicit
avifenc --stdin    out.avif < in.y4m # Expected
avifenc -c out.avif --stdin < in.y4m # Allowed
avifenc    out.avif --stdin < in.y4m # Why not? It keeps "--stdin" close to "< in.yam".
```

CHANGELOG.md
y-guyon: It makes no sense to use `:u` with `--stdin` because no other input is allowed. So how `:u` behaves with `--stdin` does not matter, as long as nothing is silently ignored.

I find it simpler to understand and to implement if `--stdin` can just be replaced with `/path/to/in.y4m` with identical behavior. I do not think it has to be mentioned in `--help` because there is no other input file name allowed, so it does not really matter where `--stdin` is.

Happy to revisit if needed though.

wantehchang:
Hi Yannis,

On Fri, Jan 24, 2025 at 3:17 AM Yannis Guyon ***@***.***> wrote:
>
> I was thinking about these cases:
>
> avifenc --stdin -c out.avif < in.y4m # Explicit
> avifenc --stdin    out.avif < in.y4m # Expected
> avifenc -c out.avif --stdin < in.y4m # Allowed
> avifenc    out.avif --stdin < in.y4m # Why not? It keeps "--stdin" close to "< in.yam".

There are two typos in this comment. Could you please edit the comment
and change "-c" to "-o"?

Thanks,
Wan-Teh

CHANGELOG.md
wantehchang: I agree that it's easier to understand if `--stdin` can be replaced with `/path/to/in.y4m` with identical behavior. This simple rule is only mentioned in the changelog. It would be good to document this in the help message.

Also, this rule needs to make an exception for the `avifenc out.avif --stdin` case. The exception might be worth documenting.

y-guyon: > It would be good to document this in the help message.

I disagree. It does not provide any benefit as long as "No other input is allowed": since there is a single input and a single output, the order of the other arguments should not matter. The same behavior as if replaced by a file path is somewhat expected, and if it is not the case, it does not really matter either because there is a single input and a single output.  
So describing that in the help message would just bring confusion in my opinion.

PR #2575, opened 2025-01-17
Remove filter in avif_fuzztest_properties

y-guyon:
Avoid failures due to "Ineffective use of Filter()".

### DIFF
--- a/tests/gtest/avif_fuzztest_properties.cc
+++ b/tests/gtest/avif_fuzztest_properties.cc
@@ -25,24 +25,35 @@ struct TestProp {
 };
 
 void EncodeDecode(ImagePtr image, EncoderPtr encoder, DecoderPtr decoder,
-                  const std::vector<TestProp>& testProps) {
+                  const std::vector<TestProp>& test_props) {
   ImagePtr decoded_image(avifImageCreateEmpty());
   ASSERT_NE(image.get(), nullptr);
   ASSERT_NE(encoder.get(), nullptr);
   ASSERT_NE(decoder.get(), nullptr);
   ASSERT_NE(decoded_image.get(), nullptr);
 
-  for (const TestProp& testProp : testProps) {
+  std::vector<TestProp> valid_test_properties;
+  for (const TestProp& testProp : test_props) {
     if (testProp.fourcc == std::array<uint8_t, 4>{'u', 'u', 'i', 'd'}) {
-      ASSERT_EQ(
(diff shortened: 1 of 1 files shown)


maryla-uc (approved):

y-guyon:
See https://github.com/google/fuzztest/blob/5bbbddfc241c8c87902d4a80cda5697dd8c20199/fuzztest/internal/domains/filter_impl.h#L103-L110

tests/gtest/avif_fuzztest_properties.cc
```diff
@@ -74,22 +85,14 @@ inline auto ArbitraryProp() {
   auto fourcc = fuzztest::Arbitrary<std::array<uint8_t, 4>>();
   auto uuid = fuzztest::Arbitrary<std::array<uint8_t, 16>>();  // ignored
   auto body = fuzztest::Arbitrary<std::vector<uint8_t>>();
-  // Don't return known properties.
-  return fuzztest::Filter(
-      [](const TestProp& prop) {
-        return !avifIsKnownPropertyType(prop.fourcc.data());
```
wantehchang, line 80: A random fourcc is unlikely to be a known property type. So this Filter should be okay.

tests/gtest/avif_fuzztest_properties.cc
```diff
@@ -74,22 +85,14 @@ inline auto ArbitraryProp() {
...
+  return fuzztest::StructOf<TestProp>(fourcc, uuid, body);
 }
 
 inline auto ArbitraryUUIDProp() {
   auto fourcc = fuzztest::Just(std::array<uint8_t, 4>{'u', 'u', 'i', 'd'});
   auto uuid = fuzztest::Arbitrary<std::array<uint8_t, 16>>();
   auto body = fuzztest::Arbitrary<std::vector<uint8_t>>();
-  // Don't use invalid UUIDs.
-  return fuzztest::Filter(
-      [](const TestProp& prop) { return avifIsValidUUID(prop.uuid.data()); },
```
wantehchang, line 91: I guess a random 16-byte is likely to be an invalid UUID more than 90% of the time in the first 100 tries. It may make sense to replace line 87 (`auto uuid = fuzztest::Arbitrary<std::array<uint8_t, 16>>();`) with custom code that generates a valid UUID.

wantehchang (commented):
Thanks for the fix!

PR #2578, opened 2025-01-20
Fix assert failure when the gainmap has a constant values.

maryla-uc:
Add a defensive AVIF_CLAMP on the range to make sure it cannot be negative (although it shouldn't happen).
Add more tests.

See discussion in #2573

### DIFF
--- a/src/gainmap.c
+++ b/src/gainmap.c
@@ -707,16 +707,22 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
         const float range = AVIF_MAX(gainMapMaxLog2[c] - gainMapMinLog2[c], 0.0f);
         const float gainMapGamma = avifUnsignedFractionToFloat(gainMap->gainMapGamma[c]);
 
-        for (int j = 0; j < height; ++j) {
-            for (int i = 0; i < width; ++i) {
-                // Remap [min; max] range to [0; 1]
-                if (range == 0.0f) {
+        if (range == 0.0f) {
+            for (int j = 0; j < height; ++j) {
+                for (int i = 0; i < width; ++i) {
                     // If the range is 0, the gain map values will be multiplied by zero when tonemapping so the values
                     // don't matter, but we still need to make sure that gainMapF is in [0,1].
                     gainMapF[c][j * width + i] = 0.0f;
-                } else {
-                    const float v = AVIF_CLAMP(gainMapF[c][j * width + i], gainMapMinLog2[c], gainMapMaxLog2[c]);
-                    gainMapF[c][j * width + i] = AVIF_CLAMP(powf((v - gainMapMinLog2[c]) / range, gainMapGamma), 0.0f, 1.0f);
+                }
+            }
+        } else {
+            // Remap [min; max] range to [0; 1]
+            for (int j = 0; j < height; ++j) {
+                for (int i = 0; i < width; ++i) {
+                    float v = gainMapF[c][j * width + i];
+                    v = AVIF_CLAMP(v, gainMapMinLog2[c], gainMapMaxLog2[c]);
+                    v = powf((v - gainMapMinLog2[c]) / range, gainMapGamma);
+                    gainMapF[c][j * width + i] = AVIF_CLAMP(v, 0.0f, 1.0f);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/gainmap.c
```diff
@@ -704,17 +704,20 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
-                const float v = AVIF_CLAMP(gainMapF[c][j * width + i], gainMapMinLog2[c], gainMapMaxLog2[c]);
-                gainMapF[c][j * width + i] = powf((v - gainMapMinLog2[c]) / range, gainMapGamma);
+                if (range == 0.0f) {
+                    // If the range is 0, the gain map values will be multiplied by zero when tonemapping so the values
+                    // don't matter, but we still need to make sure that gainMapF is in [0,1].
+                    gainMapF[c][j * width + i] = 0.0f;
+                } else {
+                    const float v = AVIF_CLAMP(gainMapF[c][j * width + i], gainMapMinLog2[c], gainMapMaxLog2[c]);
+                    gainMapF[c][j * width + i] = AVIF_CLAMP(powf((v - gainMapMinLog2[c]) / range, gainMapGamma), 0.0f, 1.0f);
+                }
```
wantehchang, line 720: Optional: It would be more efficient to have two nested for loops so that the inner loop doesn't have an if statement:

```
        if (range == 0.0f) {
            for (int j = 0; j < height; ++j) {
                for (int i = 0; i < width; ++i) {
                    // If the range is 0, the gain map values will be multiplied by zero when tonemapping so the values
                    // don't matter, but we still need to make sure that gainMapF is in [0,1].
                    gainMapF[c][j * width + i] = 0.0f;
                }
            }
        } else {
            const float gainMapGamma = avifUnsignedFractionToFloat(gainMap->gainMapGamma[c]);

            for (int j = 0; j < height; ++j) {
                for (int i = 0; i < width; ++i) {
                    // Remap [min; max] range to [0; 1]
                    const float v = AVIF_CLAMP(gainMapF[c][j * width + i], gainMapMinLog2[c], gainMapMaxLog2[c]);
                    gainMapF[c][j * width + i] = AVIF_CLAMP(powf((v - gainMapMinLog2[c]) / range, gainMapGamma), 0.0f, 1.0f);
                }
            }
        }
```

src/gainmap.c
```diff
@@ -704,17 +704,20 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
                 // Remap [min; max] range to [0; 1]
-                const float v = AVIF_CLAMP(gainMapF[c][j * width + i], gainMapMinLog2[c], gainMapMaxLog2[c]);
-                gainMapF[c][j * width + i] = powf((v - gainMapMinLog2[c]) / range, gainMapGamma);
+                if (range == 0.0f) {
+                    // If the range is 0, the gain map values will be multiplied by zero when tonemapping so the values
+                    // don't matter, but we still need to make sure that gainMapF is in [0,1].
+                    gainMapF[c][j * width + i] = 0.0f;
+                } else {
+                    const float v = AVIF_CLAMP(gainMapF[c][j * width + i], gainMapMinLog2[c], gainMapMaxLog2[c]);
+                    gainMapF[c][j * width + i] = AVIF_CLAMP(powf((v - gainMapMinLog2[c]) / range, gainMapGamma), 0.0f, 1.0f);
```
wantehchang, line 719: Since `AVIF_CLAMP` is a macro, the expression `powf((v - gainMapMinLog2[c]) / range, gainMapGamma)` will be evaluated multiple times unless the compiler eliminates the common subexpressions. It would be good to store the result of `powf((v - gainMapMinLog2[c]) / range, gainMapGamma)` in a local variable so that we don't rely on this compiler optimization.

Note: The `AVIF_MAX(gainMapMaxLog2[c] - gainMapMinLog2[c], 0.0f)` expression at line 707 has the same issue, but it is less serious.

wantehchang (approved):
I only reviewed src/gainmap.c and didn't review the tests.

src/gainmap.c
```diff
@@ -704,17 +704,26 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
+        if (range == 0.0f) {
+            for (int j = 0; j < height; ++j) {
+                for (int i = 0; i < width; ++i) {
+                    // If the range is 0, the gain map values will be multiplied by zero when tonemapping so the values
+                    // don't matter, but we still need to make sure that gainMapF is in [0,1].
+                    gainMapF[c][j * width + i] = 0.0f;
+                }
+            }
+        } else {
+            // Remap [min; max] range to [0; 1]
```
wantehchang, line 719: Optional: The declaration of `gainMapGamma` at line 708 could be moved into this `else` block because `gainMapGamma` is only used in this block.

wantehchang (approved):

PR #2580, opened 2025-01-20
Check that the gain map max is >=  gain map min.

### LINKED ISSUE #2573: Require that gainMap->gainMapMax[c] be greater than or equal to gainMap->gainMapMin[c]

maryla-uc:
Fixes #2573

### DIFF
--- a/tests/gtest/avif_fuzztest_enc_dec.cc
+++ b/tests/gtest/avif_fuzztest_enc_dec.cc
@@ -49,12 +49,12 @@ void EncodeDecodeValid(ImagePtr image, EncoderPtr encoder, DecoderPtr decoder) {
   const avifResult encoder_result =
       avifEncoderWrite(encoder.get(), image.get(), &encoded_data);
   ASSERT_EQ(encoder_result, AVIF_RESULT_OK)
-      << avifResultToString(encoder_result);
+      << avifResultToString(encoder_result) << " " << encoder->diag.error;
 
   const avifResult decoder_result = avifDecoderReadMemory(
       decoder.get(), decoded_image.get(), encoded_data.data, encoded_data.size);
   ASSERT_EQ(decoder_result, AVIF_RESULT_OK)
-      << avifResultToString(decoder_result);
+      << avifResultToString(decoder_result) << " " << decoder->diag.error;
 
   EXPECT_EQ(decoded_image->width, image->width);
   EXPECT_EQ(decoded_image->height, image->height);
--- a/tests/gtest/avif_fuzztest_enc_dec.cc
+++ b/tests/gtest/avif_fuzztest_enc_dec.cc
@@ -49,12 +49,12 @@ void EncodeDecodeValid(ImagePtr image, EncoderPtr encoder, DecoderPtr decoder) {
   const avifResult encoder_result =
       avifEncoderWrite(encoder.get(), image.get(), &encoded_data);
   ASSERT_EQ(encoder_result, AVIF_RESULT_OK)
-      << avifResultToString(encoder_result) << " " << encoder->diag.error;
+      << avifResultToString(encoder_result) << ": " << encoder->diag.error;
 
   const avifResult decoder_result = avifDecoderReadMemory(
       decoder.get(), decoded_image.get(), encoded_data.data, encoded_data.size);
   ASSERT_EQ(decoder_result, AVIF_RESULT_OK)
-      << avifResultToString(decoder_result) << " " << decoder->diag.error;
+      << avifResultToString(decoder_result) << ": " << decoder->diag.error;
 
   EXPECT_EQ(decoded_image->width, image->width);
   EXPECT_EQ(decoded_image->height, image->height);
--- a/tests/gtest/avif_fuzztest_helpers.h
+++ b/tests/gtest/avif_fuzztest_helpers.h
@@ -168,8 +168,8 @@ inline auto ArbitraryAvifAnim() {
   return fuzztest::OneOf(ArbitraryAvifAnim8b(), ArbitraryAvifAnim16b());
 }
 
-// Generates two signed fractions where the first one is smaller than the second
-// one.
+// Generates two signed fractions where the first one is smaller than or equal
+// to the second one.
 inline auto ArbitraryMinMaxSignedFraction() {
   return fuzztest::FlatMap(
       [](int32_t max_n, uint32_t max_d) {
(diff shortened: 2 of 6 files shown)


wantehchang:
Maryla: I am really sorry. I remember I reviewed this pull request but somehow got distracted and forgot to come back. I will review it tomorrow.

maryla-uc:
No worries, I completely forgot about this PR as well...

tests/gtest/avif_fuzztest_enc_dec.cc
```diff
@@ -49,12 +49,12 @@ void EncodeDecodeValid(ImagePtr image, EncoderPtr encoder, DecoderPtr decoder) {
   const avifResult encoder_result =
       avifEncoderWrite(encoder.get(), image.get(), &encoded_data);
   ASSERT_EQ(encoder_result, AVIF_RESULT_OK)
-      << avifResultToString(encoder_result);
+      << avifResultToString(encoder_result) << " " << encoder->diag.error;
```
wantehchang, line 52: Nit: change `" "` to `": "` so that the output looks like `ERROR_CODE: error message`.

Also in line 57.

tests/gtest/avif_fuzztest_helpers.h
```diff
@@ -167,13 +168,31 @@ inline auto ArbitraryAvifAnim() {
   return fuzztest::OneOf(ArbitraryAvifAnim8b(), ArbitraryAvifAnim16b());
 }
 
-// TODO: Try StructOf<Metadata>(StructOf<uint32_t[3]>())?
+// Generates two signed fractions where the first one is smaller than the second
+// one.
```
wantehchang, line 172: I think the first one is smaller than or equal to the second one, right?

tests/gtest/avif_fuzztest_helpers.h
```diff
@@ -167,13 +168,31 @@ inline auto ArbitraryAvifAnim() {
...
+inline auto ArbitraryMinMaxSignedFraction() {
+  return fuzztest::FlatMap(
+      [](int32_t max_n, uint32_t max_d) {
+        return fuzztest::Map(
+            [max_n, max_d](int32_t min_n) {
+              // For simplicity, use the same denominator for both fractions.
+              // This does not cover all possible fractions but makes it easy
+              // to gurantee that the fraction is smaller.
```
wantehchang, line 180: Nit: gurantee => guarantee

Add "first" before "fraction".

tests/gtest/avif_fuzztest_helpers.h
```diff
@@ -167,13 +168,31 @@ inline auto ArbitraryAvifAnim() {
...
+              return std::pair<avifSignedFraction, avifSignedFraction>(
+                  {min_n, max_d}, {max_n, max_d});
+            },
+            fuzztest::InRange<int32_t>(std::numeric_limits<int32_t>::min(),
+                                       max_n));
+      },
+      fuzztest::Arbitrary<int32_t>(),
+      fuzztest::InRange<uint32_t>(1, std::numeric_limits<uint32_t>::max()));
```
wantehchang, line 188: Optional: Use `fuzztest::NonZero<uint32_t>()` instead?

If this change is correct, make the same change to the other instances below. Or you can make these change in a separate pull request.

wantehchang (approved):
LGTM. Thanks!

tests/gtest/avif_fuzztest_helpers.h
```diff
@@ -188,38 +206,34 @@ ImagePtr AddGainMapToImage(
...
-      fuzztest::InRange<uint32_t>(1, std::numeric_limits<uint32_t>::max()),
-      fuzztest::Arbitrary<bool>());
+      AddGainMapToImage, ArbitraryAvifImage(),
+      /*gain_map=*/ArbitraryAvifImage(),
+      /*gain_map_min_max0=*/ArbitraryMinMaxSignedFraction(),
+      /*gain_map_min_max1=*/ArbitraryMinMaxSignedFraction(),
+      /*gain_map_min_max2=*/ArbitraryMinMaxSignedFraction(),
+      /*gain_map_gamma_n0=*/fuzztest::NonZero<uint32_t>(),
+      /*gain_map_gamma_n1=*/fuzztest::NonZero<uint32_t>(),
+      /*gain_map_gamma_n2=*/fuzztest::NonZero<uint32_t>(),
```
wantehchang, line 216: Should these three numerators use `Arbitrary` instead of `NonZero`?

tests/gtest/avif_fuzztest_helpers.h
```diff
@@ -167,13 +168,30 @@ inline auto ArbitraryAvifAnim() {
...
+// Generates two signed fractions where the first one is smaller than or equal
+// to the second one.
+inline auto ArbitraryMinMaxSignedFraction() {
+  return fuzztest::FlatMap(
+      [](int32_t max_n, uint32_t max_d) {
+        return fuzztest::Map(
+            [max_n, max_d](int32_t min_n) {
+              // For simplicity, use the same denominator for both fractions.
+              // This does not cover all possible fractions but makes it easy
+              // to guarantee that the first fraction is smaller.
```
wantehchang, line 180: Nit: Add "than or equal to the second one", or change "smaller" to "not bigger".

tests/gtest/avif_fuzztest_helpers.h
maryla-uc: No this is working as intended. Gamma must be >0.

PR #2591, opened 2025-02-03
Add AVIF_ENABLE_EXPERIMENTAL_EXTENDED_PIXI

y-guyon:
Add experimental support for PixelInformationProperty syntax from HEIF 3rd Ed. Amd2 behind a compilation flag.


src/read.c
```diff
@@ -2566,6 +2669,63 @@ static avifBool avifParsePixelInformationProperty(avifProperty * prop, const uin
...
+            if (pixi->subsamplingFlag[i]) {
+                AVIF_CHECK(avifROStreamReadBitsU8(&s, &pixi->subsamplingType[i], /*bitCount=*/4)); // unsigned int(4) subsampling_type;
+                AVIF_CHECK(avifROStreamReadBitsU8(&s, &pixi->subsamplingLocation[i], /*bitCount=*/4)); // unsigned int(4) subsampling_location;
+            }
+
+            // ISO/IEC 23008-12:2024/CDAM 2:2025 section 6.5.6.3:
+            //   This field indicates the contents of the channel. A value of 0 indicates colour/grayscale. A value of
+            //   1 indicates alpha. A value of 2 indicates depth. Values 3-7 are reserved for future use. At most one
+            //   channel shall have a channel_idc of 1.
+            if (channelIdc != 0) {
```
maryla-uc, line 2690: this is just not implemented, right? (since it doesn't match the comment just above it)
ideally it should lead to a different return status.
(same for unsupported component_format below?)

src/read.c
```diff
@@ -367,6 +371,66 @@ static uint32_t avifCodecConfigurationBoxGetDepth(const avifCodecConfigurationBo
     return 8;
 }
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
+uint8_t avifCodecConfigurationBoxGetSubsamplingType(const avifCodecConfigurationBox * av1C, uint8_t channelIndex)
+{
+    // See ISO/IEC 23008-12:2024/CDAM 2:2025 section 6.5.6.3.
+    if (channelIndex == 0) {
+        return 0; // 4:4:4
```
maryla-uc, line 379: could you use an enum? just a private one in this file would make things clearer I think but up to you

src/read.c
```diff
@@ -367,6 +371,66 @@ static uint32_t avifCodecConfigurationBoxGetDepth(const avifCodecConfigurationBo
     return 8;
 }
 
+#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
+uint8_t avifCodecConfigurationBoxGetSubsamplingType(const avifCodecConfigurationBox * av1C, uint8_t channelIndex)
+{
+    // See ISO/IEC 23008-12:2024/CDAM 2:2025 section 6.5.6.3.
```
maryla-uc, line 377: Is this the right reference? this method deals with the av1C box, not sure this bit of the spec is relevant, but maybe it's because I don't have the latest version of the spec

tests/gtest/avifpixitest.cc
```diff
@@ -0,0 +1,107 @@
...
+                             &encoded_regular_pixi),
+            AVIF_RESULT_OK);
+
+  testutil::AvifRwData encoded_extended_pixi;
+  EncoderPtr encoder_extended_pixi(avifEncoderCreate());
+  ASSERT_NE(encoder_extended_pixi, nullptr);
+  encoder_extended_pixi->speed = AVIF_SPEED_FASTEST;
+  encoder_extended_pixi->headerFormat = AVIF_HEADER_FULL_WITH_EXTENDED_PIXI;
+  ASSERT_EQ(avifEncoderWrite(encoder_extended_pixi.get(), image.get(),
+                             &encoded_extended_pixi),
```
maryla-uc, line 46: maybe check that encoded_extended_pixi is bigger than encoded_regular_pixi to make sure it's doing something?

maryla-uc (approved):

maryla-uc (approved):
thanks for the changes!

include/avif/avif.h
```diff
@@ -220,6 +220,11 @@ typedef enum avifHeaderFormat
     // WARNING: Experimental feature. Produces files that are incompatible with older decoders.
     AVIF_HEADER_REDUCED,
 #endif
+#if defined(AVIF_ENABLE_EXPERIMENTAL_EXTENDED_PIXI)
+    // Use the full syntax of the PixelInformationProperty from HEIF 3rd edition Amendment 2.
+    // WARNING: Experimental feature. Produces files that may be incompatible with older decoders.
+    AVIF_HEADER_FULL_WITH_EXTENDED_PIXI,
```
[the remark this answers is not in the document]
y-guyon: Another alternative: https://github.com/AOMediaCodec/libavif/pull/2612

PR #2595, opened 2025-02-05
Update MinimizedImageBox syntax

y-guyon:


y-guyon:
Feel free to wait for MPEG 149's "Text of ISO/IEC 23008-12 3rd edition DAM 2 Low-overhead image file format" document to be available at https://www.mpeg.org/standards/MPEG-H/12/.

wantehchang:
Yannis: I will try to review it tomorrow. If it turns out to be difficult to review without the spec, I will wait for the spec to be available.

src/write.c
```diff
@@ -2591,21 +2591,25 @@ static avifResult avifEncoderWriteMiniBox(avifEncoder * encoder, avifRWStream *
         hasExplicitCodecTypes = AVIF_FALSE;
     }
 
-    uint32_t smallDimensionsFlag = image->width <= (1 << 7) && image->height <= (1 << 7);
-    const uint32_t codecConfigSize = 4; // 'av1C' always uses 4 bytes.
+    uint32_t largeDimensionsFlag = image->width > (1 << 7) || image->height > (1 << 7); // _minus1
```
wantehchang, line 2594: Optional: The `// _minus1` comment may be unclear to people unfamiliar with the code. I suggest changing the comment to `// _minus1 is encoded`.

I wonder if we should rewrite the conditions to use `image->width - 1` and `image->height - 1` instead (this requires verifying that `image->width` and `image->height` are nonzero):

```
    uint32_t largeDimensionsFlag = image->width - 1 >= (1 << 7) || image->height - 1 >= (1 << 7);
```

If you make this change, make the same change to the conditions with a `// _minus1` comment below.

src/write.c
```diff
@@ -2591,21 +2591,25 @@ static avifResult avifEncoderWriteMiniBox(avifEncoder * encoder, avifRWStream *
...
 
     if (hasGainmap) {
         AVIF_ASSERT_OR_RETURN(image->gainMap != NULL && image->gainMap->image != NULL);
         gainmapMetadataSize = avifGainMapMetadataSize(image->gainMap);
         AVIF_ASSERT_OR_RETURN(gainmapData != NULL);
 
-        smallDimensionsFlag &= image->gainMap->image->width <= (1 << 7) && image->gainMap->image->height <= (1 << 7);
-        fewItemDataBytesFlag &= gainmapData->size < (1 << 15);
-        fewMetadataBytesFlag &= encoder->data->altImageMetadata->icc.size <= (1 << 10) && gainmapMetadataSize <= (1 << 10);
+        largeDimensionsFlag &= image->gainMap->image->width > (1 << 7) || image->gainMap->image->height > (1 << 7); // _minus1
```
wantehchang, line 2609: IMPORTANT: Change `&=` to `|=` (three instances).

src/write.c
```diff
@@ -2591,21 +2591,25 @@ static avifResult avifEncoderWriteMiniBox(avifEncoder * encoder, avifRWStream *
...
+        largeItemDataFlag &= gainmapData->size >= (1 << 15);
+        largeMetadataFlag &= encoder->data->altImageMetadata->icc.size > (1 << 10) // _minus1
+                             || gainmapMetadataSize >= (1 << 10);
```
wantehchang, line 2612: IMPORTANT: The original code is `gainmapMetadataSize <= (1 << 10)`. If we invert the condition, the new code should be `gainmapMetadataSize > (1 << 10)`.

Is the original code wrong?

src/write.c
```diff
@@ -2824,8 +2845,8 @@ static avifResult avifEncoderWriteMiniBox(avifEncoder * encoder, avifRWStream *
         AVIF_CHECKRES(avifRWStreamWrite(s, image->xmp.data, image->xmp.size)); // unsigned int(8) xmp_data[xmp_data_size_minus1 + 1];
     }
 
-    const size_t expectedChunkBytes = (hasAlpha ? codecConfigSize : 0) + (hasGainmap ? codecConfigSize : 0) + codecConfigSize +
-                                      image->icc.size + (hasGainmap ? tmapIccSize : 0) + (hasGainmap ? gainmapMetadataSize : 0) +
+    const size_t expectedChunkBytes = codecConfigSize + alphaCodecConfigSize + gainmapCodecConfigSize + image->icc.size +
```
wantehchang, line 2848: Optional/future: This long summation probably should check for integer (`size_t`) overflow. Alternatively we can check for the expected `avifRWStreamOffset(s)` after each `avifRWStreamWrite()` call.

wantehchang (approved):
I understood the changes without looking at the updated draft spec. I think we can set the target milestone of this pull request to v1.2.0.

src/write.c
y-guyon: > Is the original code wrong?

Yes, or it was based on a draft that signaled `gainmap_metadata_size_minus1` which is no longer the case.

src/write.c
y-guyon: This check is already dubious in my opinion. I keep it as a cheap verification but any added complexity would make it not worth it.

> This long summation probably should check for integer (`size_t`) overflow.

It should not be needed as `mini` total size (and thus the sum of its field sizes) is guaranteed to fit in uint32_t per design.

> Alternatively we can check for the expected `avifRWStreamOffset(s)` after each `avifRWStreamWrite()` call.

This would only verify that reading something advance its offset by that many bytes, not that we did not forget something, right?

src/write.c
y-guyon: > I wonder if we should rewrite the conditions to use `image->width - 1` and `image->height - 1` instead (this requires verifying that `image->width` and `image->height` are nonzero):
> 
> ```
>     uint32_t largeDimensionsFlag = image->width - 1 >= (1 << 7) || image->height - 1 >= (1 << 7);
> ```
> 
> If you make this change, make the same change to the conditions with a `// _minus1` comment below.

Done.

src/write.c
wantehchang: That's correct. I didn't know this check is a double check. I thought it was just checking that the final offset was expected and could be replaced by incremental checks.

src/write.c
```diff
@@ -2591,21 +2591,34 @@ static avifResult avifEncoderWriteMiniBox(avifEncoder * encoder, avifRWStream *
         hasExplicitCodecTypes = AVIF_FALSE;
     }
 
-    uint32_t smallDimensionsFlag = image->width <= (1 << 7) && image->height <= (1 << 7);
-    const uint32_t codecConfigSize = 4; // 'av1C' always uses 4 bytes.
+    // _minus1 is encoded for these fields.
+    AVIF_ASSERT_OR_RETURN(image->width != 0 && image->height != 0);
```
wantehchang, line 2595: Since `image` comes from user input, if `image->width` or `image->height` is 0, it is an invalid input argument rather than an internal error. So `AVIF_RESULT_INVALID_ARGUMENT` seems more appropriate than `AVIF_RESULT_INTERNAL_ERROR` unless we expect libavif to have validated `image->width` and `image->height` before reaching here.

y-guyon: I think it is enforced by

https://github.com/AOMediaCodec/libavif/blob/c6807414018046e280ed1f5f24d2ce1a04b16ca3/src/write.c#L1719

and

https://github.com/AOMediaCodec/libavif/blob/c6807414018046e280ed1f5f24d2ce1a04b16ca3/src/write.c#L3159

y-guyon:
[Text of ISO/IEC 23008-12 3rd edition DAM 2 Low-overhead image file format](https://www.mpeg.org/wp-content/uploads/mpeg_meetings/149_Geneva/w24745.zip) is now available.

src/write.c
y-guyon: Thank you for the analysis on `colorData->size`. To me an issue with a codec is still an internal libavif error because libavif dependencies are the responsibility of libavif, unless we allow user-defined codecs such as in https://github.com/AOMediaCodec/libavif/pull/2381.

If `AVIF_RESULT_INTERNAL_ERROR` is improper, I would suggest returning `avifGetErrorForItemCategory()` instead, as is done for other codec errors. What do you think?

y-guyon: > a malfunctioning codec could call `avifCodecEncodeOutputAddSample()` with a zero-sized sample

Should this be caught somewhere? Can a zero-sized sample be valid in regular HEIF?

PR #2597, opened 2025-02-06
update libyuv to d32d19ccf (1904)

jzern:
This adds RAWToI444 and RAWToJ444 functions which will be used in
reformat_libyuv.c in a follow up.

Bug: b:345259429

### DIFF
--- a/cmake/Modules/LocalLibyuv.cmake
+++ b/cmake/Modules/LocalLibyuv.cmake
@@ -1,4 +1,4 @@
-set(AVIF_LIBYUV_TAG "a6a2ec654b1be1166b376476a7555c89eca0c275")
+set(AVIF_LIBYUV_TAG "d32d19ccf2711d8bdf43a68dc5f8daedb7965c62")
 
 set(AVIF_LIBYUV_BUILD_DIR "${AVIF_SOURCE_DIR}/ext/libyuv/build")
 # If ${ANDROID_ABI} is set, look for the library under that subdirectory.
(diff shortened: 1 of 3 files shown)


cmake/Modules/LocalLibyuv.cmake
```diff
@@ -1,4 +1,4 @@
-set(AVIF_LIBYUV_TAG "a6a2ec654b1be1166b376476a7555c89eca0c275")
+set(AVIF_LIBYUV_TAG "d32d19ccf2711d8bdf43a68dc5f8daedb7965c62")
```
wantehchang, line 1: This commit has not been rolled into Chrome. My roll attempt was reverted because it broke some tests that compare the conversion results with golden images: https://chromium-review.googlesource.com/c/chromium/src/+/6230822

So we could consider updating to the parent of this commit first.

Note: This commit was rolled into Google's internal repository last night and it has not been reverted.

wantehchang (approved):

cmake/Modules/LocalLibyuv.cmake
jzern: > This commit has not been rolled into Chrome. My roll attempt was reverted because it broke some tests that compare the conversion results with golden images: https://chromium-review.googlesource.com/c/chromium/src/+/6230822
> 
> So we could consider updating to the parent of this commit first.
> 
> Note: This commit was rolled into Google's internal repository last night and it has not been reverted.

Thanks for pointing that out. I rolled back to the parent.

wantehchang (approved):

jzern:
The test failure is unrelated:

```
curl: (35) schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
```

(while downloading rav1e)

PR #2599, opened 2025-02-07
WIP: Allow AOM_HAVE_TUNE_IQ

vrabaud:


cmake/Modules/LocalAom.cmake
```diff
@@ -1,4 +1,4 @@
-set(AVIF_AOM_GIT_TAG v3.11.0)
+set(AVIF_AOM_GIT_TAG 2454213a72f194522c98610ef17057c226ffe391)
```
wantehchang, line 1: ext/aom.cmd also needs to be updated.

src/codec_aom.c
```diff
@@ -871,8 +874,14 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         if (!avifProcessAOMOptionsPostInit(codec, alpha)) {
             return AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION;
         }
-        if (!codec->internal->tuningSet) {
-            if (aom_codec_control(&codec->internal->encoder, AOME_SET_TUNING, AOM_TUNE_SSIM) != AOM_CODEC_OK) {
+        if (!lossless && !codec->internal->tuningSet) {
+#if defined(AOM_HAVE_TUNE_IQ)
+            if (aomUsage == AOM_USAGE_ALL_INTRA &&
+                aom_codec_control(&codec->internal->encoder, AOME_SET_TUNING, AOM_TUNE_IQ) != AOM_CODEC_OK) {
```
wantehchang, line 880: We are not ready to make `AOM_HAVE_TUNE_IQ` the default in libavif yet. But we can think about how to make the new tuning mode more discoverable.

For example, if we release libavif v1.2.0 after libaom v3.12.0 is released, we can mention the libaom codec-specific option tune=iq in the libavif v1.2.0 release notes.

wantehchang (commented):
Vincent: Thanks for looking into this.

wantehchang:
Vincent: I removed the v1.2.0 milestone from this pull request. It should not require any change to libavif to use the new tune=iq codec-specific option in libaom v3.12.0.

vrabaud:
Ok, let's wait for setting it by default then.

Issue #2589, opened 2025-02-01
MinGW+GCC fails possibly due to changes in SVT-AV1


LigH-de:
Related to https://github.com/m-ab-s/media-autobuild_suite/issues/2816#issuecomment-2628730268

Building libavif binaries in the [media-autobuild suite](https://github.com/m-ab-s/media-autobuild_suite) started to fail about the same time as there was [an issue with ffmpeg](https://github.com/m-ab-s/media-autobuild_suite/issues/2819) reported.

I guess the reason may be a change in SVT-AV1.

LigH-de:
[logs.zip](https://github.com/user-attachments/files/18627218/logs.zip) collected by M-AB-S while building

wantehchang:
Thanks for the bug report. I think the following are the relevant SVT-AV1 commits:
https://gitlab.com/AOMediaCodec/SVT-AV1/-/commit/56957ecb43a96c1db015599dfcd41859a66fab16
https://gitlab.com/AOMediaCodec/SVT-AV1/-/commit/1a9c1f11945f7f4496e1baf85b848b20c906675d
https://gitlab.com/AOMediaCodec/SVT-AV1/-/commit/988e930c1083ce518ead1d364e3a486e9209bf73

wantehchang:
@LigH-de We probably will wait until SVT-AV1 v3.0 is released to fix this because it is tricky to fix this while the SVT-AV1 team is cleanup up its public API. It will be easier if we can handle this with an API version check.

vrabaud:
I believe this is fixed, now that libavif depends on SVT-AV1 3.0.1

PR #2607, opened 2025-02-11
Make code compatible with SVT-AV1 3.0

vrabaud:


vrabaud:
@wantehchang , we can probably merge the C code without bumping the version for now. That will fix https://github.com/AOMediaCodec/libavif/issues/2589 . If more API changes happen, we will clean them along the way

src/codec_svt.c
```diff
@@ -125,14 +129,18 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         svt_config->is_16bit_pipeline = image->depth > 8;
 #endif
 
+#if !SVT_AV1_CHECK_VERSION(1, 5, 0)
```
wantehchang, line 132: I verified that in SVT-AV1 v1.5.0, the comment for `high_dynamic_range_input` changed from `Enable writing of HDR metadata in the bitstream` to `Currently unused`. See https://gitlab.com/AOMediaCodec/SVT-AV1/-/commit/33e0fdfe3.

But even in SVT-AV1 v1.0.0, `high_dynamic_range_input` is unused. So I suggest we test `!SVT_AV1_CHECK_VERSION(3, 0, 0)` instead here.

src/codec_svt.c
```diff
@@ -114,7 +114,11 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         // See https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1697.
         memset(svt_config, 0, sizeof(EbSvtAv1EncConfiguration));
 
+#if SVT_AV1_CHECK_VERSION(3, 0, 0)
+        res = svt_av1_enc_init_handle(&codec->internal->svt_encoder, svt_config);
+#else
         res = svt_av1_enc_init_handle(&codec->internal->svt_encoder, NULL, svt_config);
+#endif
```
wantehchang, line 121: There is another change in SVT-AV1 v3.0. The type of `svt_config->force_key_frames` changed from `Bool` (defined in  EbSvtAv1.h) to `bool` (defined in <stdbool.h>).

So we should add `#include <stdbool.h>` and change the line

```
        svt_config->force_key_frames = TRUE;
```

to

```
        svt_config->force_key_frames = true;
```

src/codec_svt.c
```diff
@@ -125,14 +129,18 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
...
                                                 image->transferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_SMPTE2084 &&
                                                 image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_BT2020_NCL);
+#endif
 
         svt_config->source_width = image->width;
         svt_config->source_height = image->height;
-        svt_config->logical_processors = encoder->maxThreads;
+#if SVT_AV1_CHECK_VERSION(3, 0, 0)
+        svt_config->level_of_parallelism = encoder->maxThreads;
+#endif
```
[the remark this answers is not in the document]
vrabaud: done

src/codec_svt.c
vrabaud: done

src/codec_svt.c
vrabaud: Good catch, done

vrabaud:
Actually, we just need to fix whatever is in here: https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/.gitlab/workflows/linux/libavif_api_job_fix.patch?ref_type=heads

wantehchang:
Yes, I consulted [libavif_api_job_fix.patch](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/.gitlab/workflows/linux/libavif_api_job_fix.patch?ref_type=heads) when I reviewed your pull request.

src/codec_svt.c
wantehchang: Disclosure: This suggestion is not mine. It comes from [libavif_api_job_fix.patch](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/.gitlab/workflows/linux/libavif_api_job_fix.patch?ref_type=heads) :-)

vrabaud:
Merging the .c fix will break their CI because SVT_AV1_VERSION_MAJOR has not been bumped. Let's wait for the last minute before our 1.2 release.

src/codec_svt.c
```diff
@@ -125,14 +132,20 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         svt_config->is_16bit_pipeline = image->depth > 8;
 #endif
 
+#if !SVT_AV1_CHECK_VERSION(3, 0, 0)
         // Follow comment in svt header: set if input is HDR10 BT2020 using SMPTE ST2084 (PQ).
         svt_config->high_dynamic_range_input = (image->depth == 10 && image->colorPrimaries == AVIF_COLOR_PRIMARIES_BT2020 &&
                                                 image->transferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_SMPTE2084 &&
                                                 image->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_BT2020_NCL);
+#endif
```
wantehchang, line 140: It just occurred to me that since `high_dynamic_range_input` seems to be never used, we might as well just delete it. We can submit this version first and confirm with the SVT-AV1 team that `high_dynamic_range_input` is never used.

cmake/Modules/LocalSvt.cmake
```diff
@@ -1,4 +1,4 @@
-set(AVIF_SVT_GIT_TAG "v2.3.0")
+set(AVIF_SVT_GIT_TAG "38bc2f92f0360f8280858f40f3229f796c70b9c5")
```
wantehchang, line 1: Did you forget to undo the changes to this file?

wantehchang (changes requested):

cmake/Modules/LocalSvt.cmake
vrabaud: That is just to make sure the CI passes with 3.0.0.
If we merge the .c file, it will work with 2.3.0 but not with HEAD (which still has not bumped its version), that is why I suggest to wait for upstream to to bump their version, even for an RC

src/codec_svt.c
vrabaud: It got used at some point right? Many eons ago.

src/codec_svt.c
```diff
@@ -9,6 +9,9 @@
 
 #include <stdint.h>
 #include <string.h>
+#if SVT_AV1_CHECK_VERSION(3, 0, 0)
+#include <stdbool.h>
+#endif
```
wantehchang, line 14: Optional: We can include <stdbool.h> and use the constant `true` unconditionally.

wantehchang (approved):
LGTM.

src/codec_svt.c
wantehchang: @1480c1 Christopher: Does `svt_config->high_dynamic_range_input` have an effect in any SVT-AV1 release? If not, then we will just delete this assignment statement rather than keep it under a `#if !SVT_AV1_CHECK_VERSION(3, 0, 0)` test.

wantehchang (commented):
@hassount @1480c1 Please review. Thanks!

src/codec_svt.c
1480c1: As far as I know, it has never had an effect since SVT-AV1's public release.

src/codec_svt.c
```diff
@@ -175,7 +188,11 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
 
         // In order for SVT-AV1 to force keyframes by setting pic_type to
         // EB_AV1_KEY_PICTURE on any frame, force_key_frames has to be set.
+#if SVT_AV1_CHECK_VERSION(3, 0, 0)
+        svt_config->force_key_frames = true;
+#else
         svt_config->force_key_frames = TRUE;
+#endif
```
1480c1, line 195: I believe in the case of [c99, section 7.16][1], `true` should always expand to `1`, and we always used `1` for `TRUE`, so you should be able to unconditionally just use `true` with no problem regardless of SVT-AV1's version.

[1]: https://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf

cmake/Modules/LocalSvt.cmake
```diff
@@ -57,6 +58,11 @@ else()
     set(CMAKE_OUTPUT_DIRECTORY_ORIG "${CMAKE_OUTPUT_DIRECTORY}")
     set(CMAKE_OUTPUT_DIRECTORY "${SVT_BINARY_DIR}" CACHE INTERNAL "")
 
+    # For now, this creates compilation issues:
+    # _deps/svt-build/libSvtAv1Enc.a: error adding symbols: file format not recognized
+    # clang: error: linker command failed with exit code 1 (use -v to see invocation)
+    set(SVT_AV1_LTO OFF)
```
1480c1, line 63: As a small nit, it might be possible to instead determine this based on the value of `CMAKE_INTERPROCEDURAL_OPTIMIZATION`.
However, this review comment might be unnecessary as it should be that if the user defines `CMAKE_INTERPROCEDURAL_OPTIMIZATION` for libavif, SVT-AV1 should also inherit it, and `SVT_AV1_LTO` should be a one-way variable (it will set `CMAKE_INTERPROCEDURAL_OPTIMIZATION` to `ON` if `SVT_AV1_LTO` is on and LTO is supported, it does not ever set it to `OFF`). So if the user does set `CMAKE_INTERPROCEDURAL_OPTIMIZATION`, this will only disable that LTO support check and not actually disable LTO.

src/codec_svt.c
vrabaud: Sorry, too used to work with C89 in WebP :)

src/codec_svt.c
vrabaud: done

src/codec_svt.c
vrabaud: done

cmake/Modules/LocalSvt.cmake
vrabaud: I wanted to add a proper option for libavif in a different PR but I can already check for CMAKE_INTERPROCEDURAL_OPTIMIZATION indeed.

cmake/Modules/LocalSvt.cmake
```diff
@@ -57,6 +57,12 @@ else()
     set(CMAKE_OUTPUT_DIRECTORY_ORIG "${CMAKE_OUTPUT_DIRECTORY}")
     set(CMAKE_OUTPUT_DIRECTORY "${SVT_BINARY_DIR}" CACHE INTERNAL "")
 
+    if(CMAKE_INTERPROCEDURAL_OPTIMIZATION)
+        set(SVT_AV1_LTO ON)
+    else()
+        set(SVT_AV1_LTO OFF)
+    endif()
```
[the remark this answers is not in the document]
1480c1: This looks to be in line with my comment, so this is good.

PR #2609, opened 2025-02-11
Update test_all_configurations.sh

y-guyon:


tests/test_all_configurations.sh
```diff
@@ -21,74 +21,54 @@
...
-  # The memory sanitizer triggers too many use-of-uninitialized-value errors
-  # (possibly false positives because standard libraries were not instrumented).
-  for SANITIZER in address thread undefined; do
-    pushd ext
-      rm -rf aom
-      CC=clang CXX=clang++ CFLAGS=-fsanitize=${SANITIZER} CXXFLAGS=-fsanitize=${SANITIZER} LDFLAGS=-fsanitize=${SANITIZER} \
-      ./aom.cmd
-      rm -rf dav1d
-      cp dav1d.cmd dav1d_sanitized.cmd
-      sed -i 's/meson setup \(.*\) \.\./meson setup \1 '"-Db_sanitize=${SANITIZER} -Db_lundef=false"' ../g' dav1d_sanitized.cmd
```
vrabaud, line 66: This is weird: we have to patch meson to make it aware we are using sanitizers in the CI, cf https://github.com/AOMediaCodec/libavif/blob/c6807414018046e280ed1f5f24d2ce1a04b16ca3/.github/workflows/ci-unix-static-sanitized.yml#L70

y-guyon: This is mostly to check that everything builds correctly and that all tests pass, so if dav1d is not sanitized in this script that's alright in my opinion. It is already tested in the CI as you mentioned.

It would be good to have a CMake option for that in libavif/CMakeLists.txt, if feasible.

PR #2611, opened 2025-02-11
Mark the avifGainMap struct as extensible

wantehchang:


wantehchang (commented):
Maryla: I am not sure if this pull request is correct. If the avifGainMap struct doesn't need to be extended in the future, I will abandon this pull request. Thanks!

maryla-uc (approved):

PR #2618, opened 2025-02-13
Switch dav1d's source to the official github mirror

vrabaud:
This will help when the server is down.

### DIFF
--- a/cmake/Modules/LocalDav1d.cmake
+++ b/cmake/Modules/LocalDav1d.cmake
@@ -8,8 +8,8 @@ function(avif_build_local_dav1d)
     else()
         message(STATUS "libavif(AVIF_CODEC_DAV1D=LOCAL): ext/dav1d not found, fetching")
         set(source_dir "${FETCHCONTENT_BASE_DIR}/dav1d-src")
-        list(APPEND download_step_args GIT_REPOSITORY https://code.videolan.org/videolan/dav1d.git GIT_TAG ${AVIF_DAV1D_TAG}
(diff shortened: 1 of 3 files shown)


y-guyon (approved):

vrabaud:
We are already not pulling the official repo for libxml2

cmake/Modules/LocalDav1d.cmake
```diff
@@ -8,8 +8,8 @@ function(avif_build_local_dav1d)
     else()
         message(STATUS "libavif(AVIF_CODEC_DAV1D=LOCAL): ext/dav1d not found, fetching")
         set(source_dir "${FETCHCONTENT_BASE_DIR}/dav1d-src")
-        list(APPEND download_step_args GIT_REPOSITORY https://code.videolan.org/videolan/dav1d.git GIT_TAG ${AVIF_DAV1D_TAG}
-             GIT_SHALLOW ON
+        list(APPEND download_step_args GIT_REPOSITORY https://github.com/videolan/dav1d.git GIT_TAG ${AVIF_DAV1D_TAG} GIT_SHALLOW
```
wantehchang, line 11: Is dav1d's master repository often down?

wantehchang (approved):

cmake/Modules/LocalDav1d.cmake
vrabaud: it was for several hours when I created that PR. It was never down before that. We can wait and see if it happens again.

PR #2625, opened 2025-02-18
Handle animations with audio tracks.

maryla-uc:
HEIF allows animations to have audio tracks, subtitle tracks etc. Previously, these files would yield and error because the track header for these tracks has width/height of zero. Now the size is only checked for tracks with a known video handler type ('pict', 'vide', 'auxv').

Moreover, we now only use tracks with a handler type of 'pict' for the main track (as specified in HEIF). Previously, the handler type was not checked.


tests/data/README.md
```diff
@@ -830,6 +830,16 @@ An animated AVIF image file with the following attributes:
...
 
+### File [colors-animated-8bpc-audio.avif](colors-animated-8bpc-audio.avif)
+
+![](colors-animated-8bpc-audio.avif)
+
+License: [same as libavif](https://github.com/AOMediaCodec/libavif/blob/main/LICENSE)
+
+Same as `colors-animated-8bpc.avif` but with an audio track added using MP4Box:
+
+`MP4Box -add audio.aac colors-animated-8bpc-audio.avif`
```
y-guyon, line 841: Where does `audio.aac` comes from?

y-guyon (approved):

y-guyon (approved):

src/read.c
```diff
@@ -3322,7 +3318,12 @@ static avifResult avifParseMetaBox(avifMeta * meta, uint64_t rawOffset, const ui
 
         if (firstBox) {
             if (!memcmp(header.type, "hdlr", 4)) {
-                AVIF_CHECKERR(avifParseHandlerBox(avifROStreamCurrent(&s), header.size, diag), AVIF_RESULT_BMFF_PARSE_FAILED);
+                uint8_t handlerType[4];
+                AVIF_CHECKERR(avifParseHandlerBox(avifROStreamCurrent(&s), header.size, handlerType, diag), AVIF_RESULT_BMFF_PARSE_FAILED);
+                if (memcmp(handlerType, "pict", 4) != 0) {
```
y-guyon, line 3323: Optional/nit: Could be good to mention the section and the spec mandating this

maryla-uc: Done in #2627

maryla-uc:
Note that after some internal discussion I will revert the check for the "pict" handler type for tracks to avoid the risk of breaking existing files. One such example file is https://github.com/web-platform-tests/wpt/blob/master/webcodecs/four-colors-flip.avif at commit 88004c4 which according to the readme was created with:

```
ffmpeg -i four-colors-flip.gif -vcodec libaom-av1 -crf 16 four-colors-flip.mp4
mp4box -add-image ref:primary:tk=1:samp=1 -ab avis -ab avif -ab miaf -brand avis four-colors-flip.mp4 -out four-colors-flip.avif
mp4box -edits 1=r four-colors-flip.avif
```

So it's basically a video file with extra brands and an image added. Therefore it uses the "vide" handler type instead of "pict" as it should.

PR #2626, opened 2025-02-18
Remove sample transform constraint on dimg order.

maryla-uc:
Previously, the dimg had to be specified in the same order as the items in the infe box.

Clarify max items supported by format vs max items supported by implementation.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -6009,6 +6009,24 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         AVIF_CHECKRES(avifDecoderDataFindSampleTransformImageItem(data, &sampleTransformItem));
         if (sampleTransformItem != NULL) {
             AVIF_ASSERT_OR_RETURN(data->sampleTransformNumInputImageItems == 0);
+
+            for (uint32_t i = 0; i < data->meta->items.count; ++i) {
+                avifDecoderItem * inputImageItem = data->meta->items.item[i];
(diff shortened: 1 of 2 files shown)


src/read.c
```diff
@@ -6019,12 +6037,9 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                     avifDiagnosticsPrintf(data->diag, "Box[sato] input item %u is not a supported image type", inputImageItem->id);
                     return AVIF_RESULT_DECODE_SAMPLE_TRANSFORM_FAILED;
                 }
-                // Input image item order is important because input image items are indexed according to this order.
-                AVIF_CHECKERR(inputImageItem->dimgIdx == data->sampleTransformNumInputImageItems, AVIF_RESULT_NOT_IMPLEMENTED);
```
y-guyon, line 6023: This is no longer checked?

y-guyon (commented):
Thanks!

src/read.c
y-guyon: I missed the `avifItemCategory * category = &data->sampleTransformInputImageItems[inputImageItem->dimgIdx];` part, thanks!

y-guyon (approved):

src/read.c
y-guyon: > I don't see why this was needed in the first place

Just to make sure items were in the same order are their `dimg` index, meaning not having to bother thinking how to implement it otherwise.

PR #2629, opened 2025-02-19
Disable testing on dependencies by default

vrabaud:


y-guyon (approved):

y-guyon:
Note this is necessary when building with VSCode which sets `-DBUILD_TESTING=ON` and leads to the following error without this patch:

```
[cmake] CMake Error at build/_deps/googletest-src/googletest/cmake/internal_utils.cmake:165 (add_library):
[cmake]   add_library cannot create target "gtest" because another target with the
[cmake]   same name already exists.  The existing target is a static library created
[cmake]   in source directory
[cmake]   "libavif/build/_deps/svt-src/third_party/googletest".  See
[cmake]   documentation for policy CMP0002 for more details.
```

PR #2630, opened 2025-02-19
Use new SVT 3.0.0 EbSvtAv1EncConfiguration::lossless

y-guyon:
Add test.


src/codec_svt.c
```diff
@@ -205,6 +205,11 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         }
 #endif
 
+#if SVT_AV1_CHECK_VERSION(3, 0, 0)
+        svt_config->lossless = quantizer == AVIF_QUANTIZER_LOSSLESS;
```
wantehchang, line 209: apps/avifenc.c has the following code:

```
    // Check global lossless parameters and set to default if needed.
    if (lossless) {
        ...
        // Codec.
        const char * codecName = avifCodecName(settings.codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
        ...
        } else if (codecName && !strcmp(codecName, "svt")) {
            fprintf(stderr, "SVT-AV1 doesn't support lossless encoding yet: https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1636\n");
            goto cleanup;
        }
```

We should update that check with an SVT-AV1 version check. I am not sure how to get the SVT-AV1 version number in apps/avifenc.c ...

tests/gtest/avifsvttest.cc
```diff
@@ -0,0 +1,125 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include <vector>
```
wantehchang, line 4: Also include <tuple>.

src/codec_svt.c
```diff
@@ -205,6 +205,11 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
...
+        svt_config->avif = (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) != 0;
```
wantehchang, line 210: I verified that src/codec_aom.c sets the equivalent fields under the same conditions. So these two lines are correct.

tests/gtest/avifsvttest.cc
```diff
@@ -0,0 +1,125 @@
...
+  DecoderPtr decoder(avifDecoderCreate());
+  ASSERT_NE(decoder, nullptr);
+  ImagePtr decoded(avifImageCreateEmpty());
+  ASSERT_NE(decoded, nullptr);
+  ASSERT_EQ(avifDecoderReadMemory(decoder.get(), decoded.get(), encoded.data,
+                                  encoded.size),
+            AVIF_RESULT_OK);
+
+  if (quality == AVIF_QUALITY_LOSSLESS) {
+    // Lossless is not supported in SVT-AV1 2.3.0.
```
wantehchang, line 53: I assume we set `svt_config->qp` to 0 in this case. We should ask the SVT-AV1 team what happens if `svt_config->qp` is set to 0 in SVT-AV1 2.3.0.

wantehchang (approved):
LGTM. I didn't check the tests carefully.

src/codec_svt.c
y-guyon: Thanks!

tests/gtest/avifsvttest.cc
y-guyon: As shown by the test, it encodes something but with loss.

tests/gtest/avifsvttest.cc
y-guyon: I assume you meant `#include <tuple>`.

src/codec_svt.c
y-guyon: Done.

src/codec_svt.c
y-guyon: As described in https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/2245, `avif` mode cannot be used in libavif for now.

y-guyon:
Should this be part of the libavif 1.2.0 release, now that only the new `lossless` mode is enabled and not the new `avif` mode?

wantehchang (approved):

tests/gtest/avifsvttest.cc
wantehchang: Yes. Thanks for the correction.

tests/gtest/avifsvttest.cc
wantehchang: Thanks!

tests/CMakeLists.txt
```diff
@@ -152,6 +152,7 @@ if(AVIF_GTEST)
     add_avif_gtest_with_data(avifsize0test)
     add_avif_internal_gtest(avifstreamtest)
     add_avif_gtest(avifsvttest)
+    target_include_directories(avifsvttest PRIVATE ${SVT_INCLUDE_DIR}) # To be able to check the SVT-AV1 version in test.
```
wantehchang, line 155: We should only do this if SVT-AV1 is used. Maybe check if `SVT_INCLUDE_DIR` is defined or nonempty?

Also, we don't need to build avifsvttest if SVT-AV1 is not used. See my comment at tests/gtest/avifsvttest.cc:37.

tests/gtest/avifsvttest.cc
```diff
@@ -53,7 +82,9 @@ TEST_P(SvtAv1Test, EncodeDecodeStillImage) {
     // Lossless is not supported in SVT-AV1 2.3.0.
     EXPECT_GT(testutil::GetPsnr(*image, *decoder->image), 60.0);
     // The following should pass starting with SVT-AV1 version 3.0.0.
-    // EXPECT_TRUE(testutil::AreImagesEqual(*image, *decoded));
+    if (kSvtAv1SupportsLossless) {
+      EXPECT_TRUE(testutil::AreImagesEqual(*image, *decoded));
+    }
```
wantehchang, line 87: Optional: I wonder if lines 82-87 should become an if-else statement:

```
    if (kSvtAv1SupportsLossless) {
      EXPECT_TRUE(testutil::AreImagesEqual(*image, *decoded));
    } else {
      EXPECT_GT(testutil::GetPsnr(*image, *decoder->image), 60.0);
    }
```

Similarly for lines 140-145 below.

tests/gtest/avifsvttest.cc
```diff
@@ -1,12 +1,41 @@
...
+constexpr bool kSvtAv1SupportsLossless = true;
+#else
+constexpr bool kSvtAv1SupportsLossless = false;
+#endif
+#else
+constexpr bool kSvtAv1SupportsLossless = false;
+#endif
+#else
+constexpr bool kSvtAv1SupportsLossless = false;
+#endif
```
wantehchang, line 37: Optional: If we build this file (avifsvttest.cc) only when SVT-AV1 is used, then we don't need to check `__has_include("svt-av1/EbSvtAv1Enc.h")`. That will get rid of the two `#if`'s at lines 11-12.

wantehchang (approved):

src/codec_svt.c
wantehchang: apps/avifenc.c is not modified in this PR, so this is not done.

It is fine to just add a TODO comment.

y-guyon: Sorry, I discussed checking the SVT version in `avifsvttest.cc` with Vincent and thought your comment was the same thing.

tests/gtest/avifsvttest.cc
y-guyon: Done.

tests/gtest/avifsvttest.cc
y-guyon: Done.

tests/CMakeLists.txt
y-guyon: Went with `if(AVIF_CODEC_SVT_ENABLED)`.

src/codec_svt.c
y-guyon: I went with checking values returned by `avifCodecVersions()` in `avifenc.c` and by `avifCodecVersionSvt()` in `avifsvttest.c` rather than adapting CMakeLists.txt or the libavif public API.

wantehchang: What you implemented is what I would have done. But James suggested we move the version check into src/codec_svt.c and let the encoding fail rather than produce a lossy output incorrectly. I am sorry I forgot to post James's suggestion last night.

I will review your solution first today.

y-guyon: We thought about that. Unfortunately calling libavif with SVT at `AVIF_QUALITY_BEST` should not result in a failure, right? Or we have to change the value of `AVIF_QUALITY_LOSSLESS` so that it differs from `AVIF_QUALITY_BEST` etc.

jzern: > We thought about that. Unfortunately calling libavif with SVT at `AVIF_QUALITY_BEST` should not result in a failure, right? Or we have to change the value of `AVIF_QUALITY_LOSSLESS` so that it differs from `AVIF_QUALITY_BEST` etc.

That's unfortunate. I guess this may be the only option then.

tests/gtest/avifsvttest.cc
```diff
@@ -0,0 +1,129 @@
...
+#include "aviftest_helpers.h"
+#include "gtest/gtest.h"
+
+namespace avif {
+namespace {
+
+bool SvtAv1SupportsLossless() {
+  const char* version = avifCodecVersionSvt();  // "vX.Y.Z" expected.
+  if (version == nullptr || version[0] != 'v') return false;
+  const int major_version = std::atoi(version + 1);
```
jzern, line 18: add `<cstdlib> for `std::atoi`

src/codec_svt.c
wantehchang: libavif quality 100 (or equivalently quantizer 0) means lossless. So `AVIF_QUALITY_LOSSLESS` is just a synomym for `AVIF_QUALITY_BEST`.

You can see this in the `avifenc` help message:

```
    -q,--qcolor Q                     : Quality for color in 0..100 where 100 is lossless
    --qalpha Q                        : Quality for alpha in 0..100 where 100 is lossless
    ...
    --qgain-map Q                     : Quality for the gain map in 0..100 where 100 is lossless
```

If libavif is called with SVT-AV1 v2.3.0 at `AVIF_QUALITY_BEST`, a failure is expected. Producing a lossily-encoded output is arguably a bug.

apps/avifenc.c
```diff
@@ -2009,8 +2009,22 @@ int main(int argc, char * argv[])
...
-            fprintf(stderr, "SVT-AV1 doesn't support lossless encoding yet: https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1636\n");
-            goto cleanup;
+            char versions[256];
+            avifCodecVersions(versions);
+            const char svtVersionPrefix[] = "svt [enc]:v";
+            const char * svtVersionPrefixPos = strstr(versions, svtVersionPrefix);
+            if (svtVersionPrefixPos == NULL) {
+                // Make sure the syntax returned by avifCodecVersions() did not change.
+                assert(strstr(versions, "svt") == NULL && strstr(versions, "SVT") == NULL);
+                // Let the encode fail later because SVT was not included in the build.
```
wantehchang, line 2019: Optional: We probably should return an error here, because the libavif library is apparently malfunctioning.

tests/gtest/avifsvttest.cc
```diff
@@ -0,0 +1,129 @@
...
+class SvtAv1Test : public testing::TestWithParam<std::tuple</*quality=*/int>> {
+};
+
+TEST_P(SvtAv1Test, EncodeDecodeStillImage) {
+  const int quality = std::get<0>(GetParam());
+
+  ASSERT_NE(avifCodecName(AVIF_CODEC_CHOICE_SVT, AVIF_CODEC_FLAG_CAN_ENCODE),
+            nullptr);
+  if (!testutil::Av1DecoderAvailable()) {
+    GTEST_SKIP() << "Codec unavailable, skip test.";
```
wantehchang, line 32: Nit: Codec => Decoder

tests/gtest/avifsvttest.cc
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+TEST_P(SvtAv1Test, EncodeDecodeSequence) {
+  const int quality = std::get<0>(GetParam());
+
+  ASSERT_NE(avifCodecName(AVIF_CODEC_CHOICE_SVT, AVIF_CODEC_FLAG_CAN_ENCODE),
+            nullptr);
+  if (!testutil::Av1DecoderAvailable()) {
+    GTEST_SKIP() << "Codec unavailable, skip test.";
```
wantehchang, line 75: Nit: Codec => Decoder

wantehchang (approved):

apps/avifenc.c
y-guyon: It might just be `avifCodecVersions()` that returns another string, for example "svt-av1 [enc]:v1.2.3" instead of the expected "svt [enc]:v1.2.3". Catching that in the CI with an assertion is likely enough, no need to be stricter as the only side effect is allowing lossy encoding with -q 100 for SVT < 3.0.0.

y-guyon: Also `avifenc -c svt --lossless` fails with "Matrix coefficients have to be identity, YCgCo-Re, or YCgCo-Ro in lossless mode".  
`avifenc -c svt --lossless --cicp x/x/0` fails with "matrixCoefficients may not be set to identity (0) when subsampling".  
So only `--cicp x/x/16" works which is quite niche.

src/codec_svt.c
y-guyon: The API does not mention lossless:

https://github.com/AOMediaCodec/libavif/blob/bdba09d6429c0f19dbeb487d8042ab6fb4a2d868/include/avif/avif.h#L1541-L1542

`avifenc -c svt -q 100` could be forbidden with SVT < 3.0.0 indeed.

wantehchang: We can move the definition of the `AVIF_QUALITY_LOSSLESS` macro around to make it clear that it is the same as `AVIF_QUALITY_BEST`: https://github.com/AOMediaCodec/libavif/pull/2638

tests/test_cmd_lossless.sh
```diff
@@ -68,6 +72,17 @@ pushd ${TMP_DIR}
   "${AVIFENC}" -y 400 -s 10 -l "${INPUT_GRAY_PNG}" -o "${ENCODED_FILE}"
   "${AVIFDEC}" "${ENCODED_FILE}" "${DECODED_FILE_LOSSLESS}"
   "${ARE_IMAGES_EQUAL}" "${INPUT_GRAY_PNG}" "${DECODED_FILE_LOSSLESS}" 0
+
+  # SVT-AV1 supports 4:2:0 lossless starting with version 3.0.0.
```
wantehchang, line 76: Nit: This comment can be misinterpreted to mean that SVT-AV1 supported 4:4:4 lossless before version 3.0.0. It would be good to avoid this misinterpretation. Should we remove "4:2:0"?

tests/test_cmd_lossless.sh
```diff
@@ -68,6 +72,17 @@ pushd ${TMP_DIR}
...
+  if "${AVIFENC}" --help | grep 'svt \['; then
```
wantehchang, line 77: Nit: --help => --version

tests/test_cmd_lossless.sh
```diff
@@ -68,6 +72,17 @@ pushd ${TMP_DIR}
...
+    # Depending on the version, either succeeds or prints an error message.
+    if "${AVIFENC}" -c svt --lossless --cicp x/x/16 -s 9 "${INPUT_Y4M}" "${ENCODED_FILE}" 2> "${OUT_MSG}"; then
```
wantehchang, line 79: x/x/16 => 2/2/16

Note: I am surprised that the `x` does not result in an error. It was caused by the use of the `atoi()` function in `parseU32List()`.  https://en.cppreference.com/w/c/string/byte/atoi says "If no conversion can be performed, `​0​` is returned." I observed this in the debugger. "x/x/16" is parsed into {0, 0, 16}.

Another problem with `parseU32List()` is the use of the `strtok()` function. `strtok()` is not thread-safe. See https://en.cppreference.com/w/c/string/byte/strtok:

> Each call to `strtok` modifies a static variable: is not thread safe.

Current implementations of `strtok()` are likely to make it thread-safe by replacing the static variable with a thread-specific variable.

wantehchang (approved):

apps/avifenc.c
wantehchang: These two error messages are correct. They are confusing because it is the default matrix coefficients or chroma subsampling format that causes the error. I guess we can clarify them by printing the matrix coefficients and the chroma subsampling format.

y-guyon: The former is guarded by `settings.cicpExplicitlySet` so it is not confusing.

The latter is mentioning `matrixCoefficients may not be set to identity (0) when subsampling` so stating 4:2:0 or 4:2:2 does not bring much. Monochrome is checked above.

PR #2535, opened 2024-12-31
Deprecate min/max quantizer fields/flags.

maryla-uc:
They are still supported but no longer documented and quality fields are recommended instead.
avifenc prints a warning when they are used.

PR #2631, opened 2025-02-20
Make alpha and gain map quality default to the same as color in API.

maryla-uc:


src/codec_aom.c
```diff
@@ -783,6 +783,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         }
         minQuantizer = AVIF_CLAMP(minQuantizer, 0, 63);
         maxQuantizer = AVIF_CLAMP(maxQuantizer, 0, 63);
+        // Make sure that maxQuantizer >= quantizer >= minQuantizer.
+        minQuantizer = AVIF_MIN(minQuantizer, quantizer);
+        maxQuantizer = AVIF_MAX(maxQuantizer, quantizer);
```
wantehchang, line 788: Please check src/codec_rav1e.c and src/codec_svt.c and see if they need the same change.

Why didn't we need this change before? Or is this change a bug fix?

src/write.c
```diff
@@ -1831,11 +1831,24 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
...
     encoder->data->quantizer = avifQualityToQuantizer(encoder->quality, encoder->minQuantizer, encoder->maxQuantizer);
-    encoder->data->quantizerAlpha = avifQualityToQuantizer(encoder->qualityAlpha, encoder->minQuantizerAlpha, encoder->maxQuantizerAlpha);
-    encoder->data->quantizerGainMap =
-        avifQualityToQuantizer(encoder->qualityGainMap, AVIF_QUANTIZER_BEST_QUALITY, AVIF_QUANTIZER_WORST_QUALITY);
+    // If alpha quality, and min and max quantizer have their default values, default to the same quality as color.
+    if (encoder->qualityAlpha == AVIF_QUALITY_DEFAULT && encoder->minQuantizerAlpha == AVIF_QUANTIZER_BEST_QUALITY &&
+        encoder->maxQuantizerAlpha == AVIF_QUANTIZER_WORST_QUALITY) {
+        encoder->data->quantizerAlpha = encoder->data->quantizer;
+        encoder->minQuantizerAlpha = encoder->minQuantizer;
+        encoder->maxQuantizerAlpha = encoder->maxQuantizer;
```
wantehchang, line 1841: We probably can leave `encoder->minQuantizerAlpha` and `encoder->maxQuantizerAlpha` unchanged.

src/codec_aom.c
maryla-uc: This is a bugfix. Previously, specifying --min that was > --max would result in an error.
Specifying a --qcolor value that would result in a quantizer that was not in the [min,max] range would also result in an error for animations with speed >= 7 (which is when we switch to CBR mode). This is very unintuitive given that users don't even know what quantizer a given quality maps to.
In both cases the error was
`aom_codec_enc_init() failed: Invalid parameter: rc_min_quantizer out of range [..cfg->rc_max_quantizer]`

I did test with svt and rav1d as well and even though we pass the min/max quantizer values, they seemed to have no effect whatsoever and I did not get any errors even with invalid values (e.g. min > max), which is why I didn't make any changes there. This is probably due to how we parameter them and the min/max might have an effect with other values, for example if we used CQP (constant quantization parameter) instead of CRF (constant rate factor) in svt.

This error made me realize that min/max were actually used, whereas when I deprecated them in #2535 I was under the impression that they were just for backward compatibility, to compute an equivalent quality factor.
But actually in some cases they do have an effect when specified in addition to --qcolor. However, this is only when using libaom AND speed >= 7 AND encoding an animation. Even in that case, we modify them with `minQuantizer = AVIF_MAX(quantizer - 4, minQuantizer); maxQuantizer = AVIF_MIN(quantizer + 4, maxQuantizer);` which ensures that the range is at most quantizer+/-4. So the min/max you provide only have an effet if they're in [q-3;q+3], in which case their effect is to constrain even more the qp range. I think this is all extremely confusing...

I was wondering if I should un-deprecate these parameters/flags but I chatted with Yannis and his opinion was that just because they sometimes do something doesn't mean it's something useful, and maybe they should be turned into codec-specific advanced options. Indeed I think only power users who want high control over how animations are encoded would need to use them, so this might be a good option. The question is again whether it's ok to change the behavior, but given their current effect which is very limited and confusing, I think it is.

Right now most people who use this flag set them to 0 and 63 because that's what you see in a lot of example command lines. These are the default values, so it's a noop. In other cases there might be a small behavior change, but only in limited cases (animations using CBR mode).

But for now however I just fixed the errors.

src/write.c
maryla-uc: As explained above right now min/max quantizer actually do something (sometimes) and should be kept in sync I believe.

src/codec_aom.c
maryla-uc: I updated the PR to add the same bugfix when  updating min/max. I think avm would need the same fixes but I'm a bit confused because it seems that min can be negative but the rc_min_quantizer is still a uint? So I left it as is for now. Maybe these ranges should be fixed earlier in write.c actually... Or maybe as I suggested we just get rid of them :)

wantehchang: Both min and max are in the range 0..63, so they cannot be negative.

The code is complicated because it tries its best to be backward compatible if the (newer) quality option is not set. Also, some rate control modes in libaom such as CBR and VBR use a quantizer range rather than a single quantizer. This is why we have to calculate rc_min_quantizer and rc_max_quantizer from the user-specified quantizer (which comes from quaity) +/- 4.

src/write.c
wantehchang: I looked into this. We should delete these two lines (1840-1841).

We should not modify `encoder->minQuantizerAlpha` and `encoder->maxQuantizerAlpha` because they are owned by the caller/user. `encoder->data->quantizerAlpha`, on the other hand, is an internal data member that we own.

src/write.c
```diff
@@ -1831,11 +1831,24 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
     }
 
     // -----------------------------------------------------------------------
-    // Map quality and qualityAlpha to quantizer and quantizerAlpha
+    // Map quality settings to quantizer values.
     encoder->data->quantizer = avifQualityToQuantizer(encoder->quality, encoder->minQuantizer, encoder->maxQuantizer);
-    encoder->data->quantizerAlpha = avifQualityToQuantizer(encoder->qualityAlpha, encoder->minQuantizerAlpha, encoder->maxQuantizerAlpha);
-    encoder->data->quantizerGainMap =
-        avifQualityToQuantizer(encoder->qualityGainMap, AVIF_QUANTIZER_BEST_QUALITY, AVIF_QUANTIZER_WORST_QUALITY);
+    // If alpha quality, and min and max quantizer have their default values, default to the same quality as color.
```
wantehchang, line 1836: Nit: min and max quantizer => alpha min and max quantizers

Alternatively, we can say "If alpha quality, min quantizer, and max quantizer ..."

src/codec_aom.c
```diff
@@ -783,6 +783,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         }
         minQuantizer = AVIF_CLAMP(minQuantizer, 0, 63);
         maxQuantizer = AVIF_CLAMP(maxQuantizer, 0, 63);
+        // Make sure that maxQuantizer >= quantizer >= minQuantizer.
```
wantehchang, line 786: We can alternatively return an error if `maxQuantizer >= quantizer >= minQuantizer` does not hold. Do you think it is better to fix it up transparently?

wantehchang, line 786: Add the following comment:

> If quantizer conflicts with minQuantizer and maxQuantizer, quantizer has precedence.

Note: The conflict could also be resolved by giving minQuantizer and maxQuantizer precedence:

```
        quantizer = AVIF_CLAMP(quantizer, minQuantizer, quantizer);
```

Since minQuantizer and maxQuantizer are deprecated, it doesn't make sense to give them precedence.

src/codec_aom.c
```diff
@@ -921,6 +924,11 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 quantizerUpdated = AVIF_TRUE;
             }
         }
+        if (quantizerUpdated) {
+            // Make sure that maxQuantizer >= quantizer >= minQuantizer.
```
wantehchang, line 928: Nit: Change `maxQuantizer` and `minQuantizer` to `rc_max_quantizer` and `rc_min_quantizer`


Add the following comment:

> If quantizer conflicts with rc_min_quantizer and rc_max_quantizer, quantizer has precedence.

src/write.c
```diff
@@ -1831,11 +1831,24 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
     }
 
     // -----------------------------------------------------------------------
-    // Map quality and qualityAlpha to quantizer and quantizerAlpha
+    // Map quality settings to quantizer values.
```
wantehchang, line 1834: Optional: I suggest moving the changes to codec_aom.c to a separate PR and keeping only the changes to this file for the v1.2.0 release.

I assume the changes to this file are sufficient to accomplish the goal of the PR: Make alpha and gain map quality default to the same as color in API.

wantehchang (approved):

src/codec_aom.c
wantehchang: If this eventually leads to an error, we can also fail early with a more meaningful diagnostic message.

y-guyon:
Note that the current CHANGELOG entry covers both API and `avifenc`:

https://github.com/AOMediaCodec/libavif/blob/96fc68a8c174faa8847c39633b3574631e632cb7/CHANGELOG.md?plain=1#L67

and for gain maps, it went out of experimental status since the last release:

https://github.com/AOMediaCodec/libavif/blob/96fc68a8c174faa8847c39633b3574631e632cb7/CHANGELOG.md?plain=1#L12-L13

so it is unnecessary to describe the gain map default value change (it is considered an addition).

PR #2635, opened 2025-02-24
Disable PNG tRNS before PLTE test

y-guyon:
https://github.com/AOMediaCodec/libavif/pull/2630 failed with the unrelated error:

```
-- Found PNG: /opt/homebrew/lib/libpng.dylib (found suitable version "1.6.47", minimum required is "1.6.32")
```
```
[ RUN      ] PngTest.RgbColorTypeWithTrnsBeforePlte
libpng warning: PLTE: out of place
/Users/runner/work/libavif/libavif/tests/gtest/avifreadimagetest.cc:93: Failure
Expected equality of these values:
  image->alphaPlane
    Which is: 0x11930a900
  nullptr
    Which is: (nullptr)

[  FAILED  ] PngTest.RgbColorTypeWithTrnsBeforePlte (0 ms)
```

This seems to match libpng 1.6.47 CHANGELOG:  
https://github.com/pnggroup/libpng/blob/738f5e743ccc59de62a7b789460636f6f9d12c93/CHANGES#L6243-L6246

Note that this test was added in https://github.com/AOMediaCodec/libavif/pull/1435.

### DIFF
--- a/tests/gtest/avifreadimagetest.cc
+++ b/tests/gtest/avifreadimagetest.cc
@@ -84,7 +84,10 @@ TEST(PngTest, RgbColorTypeWithTrnsAfterPlte) {
 // Verify we can read a PNG file with PNG_COLOR_TYPE_RGB and a tRNS chunk
 // before a PLTE chunk. libpng considers the tRNS chunk as invalid and ignores
 // it, so the decoded image should have no alpha.
-TEST(PngTest, RgbColorTypeWithTrnsBeforePlte) {
+// This test is disabled because the behavior seemed to have changed starting
+// with libpng 1.6.47.
+// See https://github.com/pnggroup/libpng/blob/libpng16/CHANGES#L6243-L6246.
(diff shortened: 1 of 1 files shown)


maryla-uc (approved):
Should we just delete the test...?

y-guyon:
> Should we just delete the test...?

Maybe we can disable it depending on the libpng version in the future.

wantehchang:
Yannis: Thank you for looking up the history of this test. It is useful to keep this test, but the test needs to be relaxed to allow the decoded image to have an alpha channel. (We don't need to go to the extent of checking the libpng version.) The important thing is that we can decode the PNG image with no memory errors similar to those reported in the libwebp2 bug b/165819892.

I will write a pull request to upgrade libpng to 1.6.47 and another pull request to relax this test and re-enable it.

wantehchang:
I found that we cannot upgrade libpng until a zlib > 1.3.1 is released.

PR #2638, opened 2025-02-24
Make it clear that the best quality is lossless

wantehchang:
Define AVIF_QUALITY_LOSSLESS right after AVIF_QUALITY_BEST to make it clear that AVIF_QUALITY_LOSSLESS is the same as AVIF_QUALITY_BEST.


jzern (approved):

include/avif/avif.h
```diff
@@ -101,9 +101,9 @@ typedef int avifBool;
 #define AVIF_DEFAULT_IMAGE_COUNT_LIMIT (12 * 3600 * 60)
 
 #define AVIF_QUALITY_DEFAULT -1
-#define AVIF_QUALITY_LOSSLESS 100
 #define AVIF_QUALITY_WORST 0
 #define AVIF_QUALITY_BEST 100
+#define AVIF_QUALITY_LOSSLESS 100
```
jzern, line 106: Note line 108 and 109 are in the opposite order. This group makes sense as it goes worst->best->lossless

wantehchang: Yes. I considered this and decided to keep worst and best next to each other, like line 109 and 110.

vrabaud:
Should we have a different value for AVIF_QUALITY_LOSSLESS? Right now, when using the API with rav1e and AVIF_QUALITY_LOSSLESS, it is not lossless, just their highest quality (for avifenc, we do check for lossless and error out though https://github.com/AOMediaCodec/libavif/blob/640df130534f696edbf291cde87f9245d9b20e1c/apps/avifenc.c#L2009 ).

PR #2642, opened 2025-02-25
Update README build instructions

y-guyon:


maryla-uc (approved):

maryla-uc (approved):

.github/workflows/ci-windows-artifacts.yml
```diff
@@ -56,6 +56,7 @@ jobs:
 
       - name: Prepare libavif (cmake)
         run: >
+          CFLAGS=-static CXXFLAGS=-static LDFLAGS=-static
```
wantehchang, line 59: IMPORTANT: This should be examined. Does MSVC support the `-static` compiler or linker option? It seems unlikely.

wantehchang: I see you are removing this in https://github.com/AOMediaCodec/libavif/pull/2648.

PR #2647, opened 2025-02-25
Disable ext-partitions and uneven-4way in avm

y-guyon:


src/codec_avm.c
```diff
@@ -760,6 +760,11 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
         if (!avifProcessAOMOptionsPostInit(codec, alpha)) {
             return AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION;
         }
+        // Disabling these two gives 1.19% PSNR YUV loss in All-Intra config, but encode will be ~4X faster.
+        if (aom_codec_set_option(&codec->internal->encoder, "enable-ext-partitions", "0") != AOM_CODEC_OK ||
+            aom_codec_set_option(&codec->internal->encoder, "enable-uneven-4way-partitions", "0") != AOM_CODEC_OK) {
+            return AVIF_FALSE;
```
wantehchang, line 766: IMPORTANT: Return an error code, such as `AVIF_RESULT_UNKNOWN_ERROR`.

`AVIF_FALSE` has the same value as `AVIF_RESULT_OK` (0). This is a limitation of C (integer can be converted to an enum type automatically).

PR #2648, opened 2025-02-25
Remove -static in ci-windows-artifacts workflow

y-guyon:


y-guyon:
For Windows, prepending `CFLAGS=-static` does not work. I wonder if it should be [`/WHOLEARCHIVE`](https://learn.microsoft.com/en-us/cpp/build/reference/wholearchive-include-all-library-object-files?view=msvc-170).  
Is `-static` cross-platform somehow?

On top of that change, we should use the GitHub workflow `env:` syntax to set CFLAGS etc.

Anyway the example command line in README is not cross-platform and needs to be adapted I guess:
https://github.com/AOMediaCodec/libavif/blob/31d13993420f34b5af1d11d69d9f82844d98e645/README.md?plain=1#L170
If it is not cross-platform, we may as well use line breaks.

wantehchang (approved):

wantehchang:
No, `-static` is not cross-platform. It is Unix-specific. I don't know if there is a MSVC linker option equivalent to `-static`.

`/WHOLEARCHIVE` is equivalent to the Unix linker option `--whole-archive`. It means all object files in the static library will be included in the link, whether the object file is needed to resolve a reference or not.

PR #2652, opened 2025-02-27
Specify target platform version in binary artifacts

y-guyon:


y-guyon:
Versions were arbitrarily selected in https://en.wikipedia.org/wiki/Microsoft_Windows_SDK and https://en.wikipedia.org/wiki/MacOS_version_history.

wantehchang:
Yannis: Thanks a lot for writing this pull request. I asked James to review this.

I suggest we follow this OS support table: https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md

y-guyon:
I do not fully grasp the implications of passing this flag or not and using a newer or older version.

Could you describe what could go wrong in these scenarios:
- Binary compiled with `CMAKE_OSX_DEPLOYMENT_TARGET=X` and run on platform version < X (the OS just refuses to run that binary, right?)
- Binary compiled with `CMAKE_OSX_DEPLOYMENT_TARGET=X` and run on platform version >= X (all is fine)
- Binary compiled without `CMAKE_OSX_DEPLOYMENT_TARGET` (I guess CMake assigns some value X depending on the environment at compile time and this scenario is equivalent to the two others)

I understood that setting a target version makes sure no system call introduced in a later platform version is used. If that is the only impact, why not use the oldest possible target version?  
From what I saw, setting `CMAKE_OSX_DEPLOYMENT_TARGET` does not change the SDK version used to compile libavif anyway.

Same questions as above for `CMAKE_VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION` instead of `CMAKE_OSX_DEPLOYMENT_TARGET` if the behavior is different on Windows.

---

Should we use `CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION`, `CMAKE_VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION`, or both?

jzern:
> I do not fully grasp the implications of passing this flag or not and using a newer or older version.
> 
> Could you describe what could go wrong in these scenarios:
> 
> * Binary compiled with `CMAKE_OSX_DEPLOYMENT_TARGET=X` and run on platform version < X (the OS just refuses to run that binary, right?)

This sets `-mmacosx-min-version` / `-mmacos-min-version`, so in general that would be my expectation, unless no new functions were used and you got lucky in the shared object dependencies.

> * Binary compiled with `CMAKE_OSX_DEPLOYMENT_TARGET=X` and run on platform version >= X (all is fine)
> * Binary compiled without `CMAKE_OSX_DEPLOYMENT_TARGET` (I guess CMake assigns some value X depending on the environment at compile time and this scenario is equivalent to the two others)
> 
> I understood that setting a target version makes sure no system call introduced in a later platform version is used. If that is the only impact, why not use the oldest possible target version?

Security / bug fixes in the library functions would be the main reason. But if the goal is to support back to some version it's what you're left with.

> From what I saw, setting `CMAKE_OSX_DEPLOYMENT_TARGET` does not change the SDK version used to compile libavif anyway.
> 

https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html. So for the most part just `-mmacos-version-min` will be set. It sounds like in cases where multiple versions of the SDK were installed this might also influence that.

> Same questions as above for `CMAKE_VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION` instead of `CMAKE_OSX_DEPLOYMENT_TARGET` if the behavior is different on Windows.
> 
> Should we use `CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION`, `CMAKE_VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION`, or both?

It sounds like `CMAKE_VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION` won't be used for non-WindowsStore projects:

https://cmake.org/cmake/help/latest/prop_tgt/VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION.html#prop_tgt:VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION

[`CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION`](https://cmake.org/cmake/help/latest/variable/CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION.html) will choose a SDK version. I guess if you wanted to support the oldest versions of Windows 10, you could specify it, but I feel like this might be something to address if there's a compatibility issue reported by a user.

.github/workflows/ci-windows-artifacts.yml
```diff
@@ -64,6 +64,7 @@ jobs:
           -DAVIF_JPEG=LOCAL -DAVIF_ZLIBPNG=LOCAL
           -DAVIF_BUILD_EXAMPLES=OFF -DAVIF_BUILD_APPS=ON
           -DAVIF_BUILD_TESTS=OFF -DAVIF_ENABLE_WERROR=ON
+          -DCMAKE_VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION=5
```
jzern, line 67: I would guess you want `10` at least, Windows 8 is EOL

.github/workflows/ci-macos-artifacts.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           -DAVIF_BUILD_EXAMPLES=OFF -DAVIF_BUILD_APPS=ON
           -DAVIF_BUILD_TESTS=OFF -DAVIF_ENABLE_WERROR=ON
           -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
+          -DCMAKE_OSX_DEPLOYMENT_TARGET=10
```
jzern, line 37: 11 (from the compatibility matrix) or 10.15 might make sense.

y-guyon:
> It sounds like `CMAKE_VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION` won't be used for non-WindowsStore projects:
> 
> https://cmake.org/cmake/help/latest/prop_tgt/VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION.html#prop_tgt:VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION

It is unclear to me if "the target platform minimum version will not be specified for the project" when:
- `VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION` is not specified, or
- it is not a WindowsStore project.

> [`CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION`](https://cmake.org/cmake/help/latest/variable/CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION.html) will choose a SDK version. I guess if you wanted to support the oldest versions of Windows 10, you could specify it, but I feel like this might be something to address if there's a compatibility issue reported by a user.

It is unclear to me what "might be something to address if there's a compatibility issue reported by a user".  
Should we NOT use `CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION` until a user complains, or should we DO use `CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION` set to a recent version and downgrade it once a user needs it?

jzern:
> It is unclear to me what "might be something to address if there's a compatibility issue reported by a user".
> Should we NOT use `CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION` until a user complains, or should we DO use `CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION` set to a recent version and downgrade it once a user needs it?

I was thinking we would leave things as is and wait for a report about incompatibility. That might be a little fragile if we move the builder to Windows 11. We can try what you have and see if there are any issues.

jzern (approved):

jzern (commented):
Actually I wonder if `CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION` only affects the Visual Studio generator; NInja is being used now.

jzern:
With `-G Ninja`:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION
```

y-guyon:
Are you suggesting dropping `-G Ninja` in general?  
I myself often favor omitting the generator selection argument and use `cmake --build` instead of `make` or `ninja`.  
Do you know why we select `ninja` instead of letting CMake pick something?

jzern:
> Are you suggesting dropping `-G Ninja` in general? I myself often favor omitting the generator selection argument and use `cmake --build` instead of `make` or `ninja`. Do you know why we select `ninja` instead of letting CMake pick something?

If we want to go forward with the Visual Studio variables we will need to use the Visual Studio project generator. I'm guessing ninja was used because locally it will be faster since it will parallelize on the file level, not project. For something like there's less of a need for it.

.github/workflows/ci-windows-artifacts.yml
```diff
@@ -56,14 +56,15 @@ jobs:
...
           -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_AOM_ENCODE=ON
           -DAVIF_CODEC_AOM_DECODE=OFF -DAVIF_CODEC_DAV1D=LOCAL
           -DAVIF_LIBYUV=LOCAL -DAVIF_LIBSHARPYUV=LOCAL
           -DAVIF_JPEG=LOCAL -DAVIF_ZLIBPNG=LOCAL
           -DAVIF_BUILD_EXAMPLES=OFF -DAVIF_BUILD_APPS=ON
           -DAVIF_BUILD_TESTS=OFF -DAVIF_ENABLE_WERROR=ON
+          -DCMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION=10
       - name: Build libavif (ninja)
         working-directory: ./build
         run: ninja
```
jzern, line 70: Now this will need an update to `cmake --build` or `msbuild`, no?

y-guyon: Yes, but I started doing that in another PR https://github.com/AOMediaCodec/libavif/pull/2665, which may be too much of a change for no tangible benefit? Let me know if you have an opinion on that.

jzern: > Yes, but I started doing that in another PR #2665, which may be too much of a change for no tangible benefit? Let me know if you have an opinion on that.

I'm OK with doing that assuming the performance of the presubmits was similar.

.github/workflows/ci-windows-artifacts.yml
```diff
@@ -56,23 +56,23 @@ jobs:
...
+          -DCMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION=10
+      - name: Build libavif
+        run: cmake --build build --parallel 4
       - name: Archive artifacts
         uses: thedoctor0/zip-release@b57d897cb5d60cb78b51a507f63fa184cfe35554 # 0.7.6
         with:
           type: "zip"
           filename: "windows-artifacts.zip"
-          directory: "build"
+          directory: "build/Debug"
```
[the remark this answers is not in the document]
y-guyon: > Why not build/Release?

I tried that at first but it was empty.

> With `cmake --build` I think something like `--config=Release` should work.

That worked, thanks.

jzern (approved):

Issue #2659, opened 2025-03-03
Compilation failure on some aarch64 for Pillow


vrabaud:
All explained here: https://github.com/python-pillow/Pillow/pull/5201#discussion_r1977622504
It requires patching libyuv. Let's wait in case libyuv gets patched quickly.

PR #2660, opened 2025-03-03
fix: patch libyuv CMakeLists for compatibility with gcc 10

fdintino:
This provides a resolution for #2659 until a fix gets committed for [libyuv issue 399856238](https://libyuv.issues.chromium.org/issues/399856238).

Here is a [successful Pillow manylinux 2014 build](https://github.com/fdintino/Pillow/actions/runs/13637057763/job/38118334872) using this patch.

### DIFF
--- a/ext/libyuv.patch
+++ b/ext/libyuv.patch
@@ -1,8 +1,8 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
-index 5df76856..1ac840fb 100644
+index 5df76856..eae0b729 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -112,7 +112,7 @@ if(NOT MSVC)
+@@ -112,13 +112,13 @@ if(NOT MSVC)
        ${ly_src_dir}/rotate_neon64.cc
        ${ly_src_dir}/row_neon64.cc
        ${ly_src_dir}/scale_neon64.cc)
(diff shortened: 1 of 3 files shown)


jzern:
cc: @fbarchard

vrabaud:
Thx! Let's go with this for now.

wantehchang:
I converted ext/libyuv.patch to a CL for libyuv:
https://chromium-review.googlesource.com/c/libyuv/libyuv/+/6405834

ext/libyuv.patch
```diff
@@ -0,0 +1,20 @@
...
+       ${ly_src_dir}/scale_neon64.cc)
+-    target_compile_options(${ly_lib_name}_neon64 PRIVATE -march=armv8-a+dotprod+i8mm)
++    target_compile_options(${ly_lib_name}_neon64 PRIVATE -march=armv8.2-a+dotprod+i8mm)
+     list(APPEND ly_lib_parts $<TARGET_OBJECTS:${ly_lib_name}_neon64>)
+ 
+     # Enable AArch64 SVE kernels.
+     add_library(${ly_lib_name}_sve OBJECT
+       ${ly_src_dir}/row_sve.cc)
+-    target_compile_options(${ly_lib_name}_sve PRIVATE -march=armv9-a+sve2)
++    target_compile_options(${ly_lib_name}_sve PRIVATE -march=armv8.5-a+sve2)
```
wantehchang, line 17: @fdintino Frankie: In the review of the [libyuv CL](https://chromium-review.googlesource.com/c/libyuv/libyuv/+/6405834), George Stead suggested that we use `-march=armv8.5-a+i8mm+sve2` instead. Could you please test that and see if it works with gcc 10?

fdintino: Sure, though I won't be able to get to it until late this evening.

fdintino: `-march=armv8.5-a+i8mm+sve2` works fine.

wantehchang: Frankie: Thanks a lot for verifying it works!

PR #2665, opened 2025-03-04
Use cmake --build instead of direct ninja tool

y-guyon:
Use the default generator instead of specifying Ninja when possible.
Keep -G Ninja for Windows to avoid generating binaries in a sub Debug directory.


wantehchang (commented):
I remember we discussed this issue before. The reason I chose Ninja is that it is cross-platform and faster than some other build tools such as GNU make. Is the use of Ninja causing problems? Or are we concerned that CMake may not support the Ninja generator as well as the default generator?

y-guyon:
> I remember we discussed this issue before. The reason I chose Ninja is that it is cross-platform and faster than some other build tools such as GNU make. Is the use of Ninja causing problems? Or are we concerned that CMake may not support the Ninja generator as well as the default generator?

Thank you for sharing the context.

Advantages of using `cmake --build` instead of directly `ninja`:
- It does not depend on the generator used.

Cons:
- `-j N` or `--parallel N` must be specified in case of another generator than ninja (failures on macos when unlimited threads).

Advantages of not specifying the generator:
- More generators are tested, not just ninja.
- No specific dependency on ninja, cmake can use any available generator.

Cons:
- Slower (I thought ninja was only faster when rebuilding which is irrelevant in CI)
- Needs some tuning and extra testing to support more generators (binaries in a subfolder with MSVC for example)

So for now I suggest going with `cmake -G ninja` and `cmake --build`. What do you think?  
Not a big deal anyway, I can also drop this PR.

.github/workflows/ci-disable-gtest.yml
```diff
@@ -47,17 +47,16 @@ jobs:
 
       - name: Prepare libavif (cmake)
         run: >
-          cmake -G Ninja -S . -B build
+          cmake -S . -B build
```
wantehchang, line 50: The `CMAKE_BUILD_TYPE` variable is not used by a multi-config generator such as Visual Studio. I remember `-DCMAKE_BUILD_TYPE=Release` will cause a cmake warning to be emitted if the Visual Studio generator is used.

wantehchang:
> So for now I suggest going with `cmake -G ninja` and `cmake --build`. What do you think? Not a big deal anyway, I can also drop this PR.

Yes, this is fine by me. `cmake --build` vs. `ninja -C` is a matter of preference for abstraction vs. concreteness, or whether one uses CMake for everything or just for build system configuration and generation. (I am not familiar with the failures on macos when unlimited threads that you mentioned.)

.github/workflows/ci-disable-gtest.yml
y-guyon: I went with `cmake --build build --config Release`. It does not seem to print a warning for single-config generators.

The alternative is specifying `cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_CONFIGURATION_TYPES=Release` everywhere but that will always emit a warning.

ext/dav1d.cmd
```diff
@@ -3,22 +3,17 @@
 
 : # The odd choice of comment style in this file is to try to share this script between *nix and win32.
 
-: # meson and ninja must be in your PATH.
+: # meson must be in your PATH.
```
wantehchang, line 6: Undo this change? Meson requires Ninja:
https://mesonbuild.com/Getting-meson.html#dependencies

wantehchang, line 6: Undo this change? Meson requires Ninja. See https://mesonbuild.com/Getting-meson.html#dependencies and https://code.videolan.org/videolan/dav1d#compile.

ext/dav1d.cmd
```diff
@@ -3,22 +3,17 @@
...
-
 : # macOS might require: -Dc_args=-fno-stack-check
 : # Build with asan: -Db_sanitize=address -Db_lundef=false
 : # Build with msan: -Db_sanitize=memory -Db_lundef=false -Denable_asm=false
 : # Build with ubsan: -Db_sanitize=undefined -Db_lundef=false
-meson setup --default-library=static --buildtype release -Denable_tools=false -Denable_tests=false ..
-cd ../..
-ninja -C dav1d/build
+meson setup --default-library=static --buildtype release -Denable_tools=false -Denable_tests=false dav1d/build dav1d
+meson compile -C dav1d/build
```
wantehchang, line 19: Optional: I think it is fine to use the `ninja -C dav1d/build` command as before. The instructions in the dav1d website also use the ninja command:
https://code.videolan.org/videolan/dav1d#compile
https://code.videolan.org/videolan/dav1d#build-documentation

ext/dav1d_android.sh
```diff
@@ -16,9 +16,7 @@ if [ $# -ne 1 ]; then
   exit 1
 fi
 git clone -b 1.5.1 --depth 1 https://code.videolan.org/videolan/dav1d.git
-cd dav1d
-mkdir build
-cd build
+mkdir dav1d/build
```
wantehchang, line 19: IMPORTANT: Should we remove line 33 `cd "${abi}"`?

ext/dav1d_android.sh
```diff
@@ -32,12 +30,11 @@ ABI_LIST=("armeabi-v7a" "arm64-v8a" "x86" "x86_64")
...
 for i in "${!ABI_LIST[@]}"; do
   abi="${ABI_LIST[i]}"
-  mkdir "${abi}"
   cd "${abi}"
   PATH=$PATH:${android_bin} meson setup --default-library=static --buildtype release \
     --cross-file="../../package/crossfiles/${ARCH_LIST[i]}-android.meson" \
-    -Denable_tools=false -Denable_tests=false ../..
-  PATH=$PATH:${android_bin} ninja
+    -Denable_tools=false -Denable_tests=false "dav1d/build/${abi}" dav1d
+  PATH=$PATH:${android_bin} meson compile -C "dav1d/build/${abi}"
```
wantehchang, line 37: Optional: Same here. I think it is fine to use the `ninja -C "dav1d/build/${abi}"` command.

wantehchang (approved):

.github/workflows/ci-macos-artifacts.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
       - name: Prepare libavif (cmake)
         run: >
-          cmake -G Ninja -S . -B build
+          cmake -S . -B build
```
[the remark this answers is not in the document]
y-guyon: I thought it was necessary for `CMAKE_OSX_DEPLOYMENT_TARGET` to work but I mixed it with MSVC.

ext/dav1d.cmd
y-guyon: I guess if we use `cmake` instead of `ninja`, we may as well use `meson` instead of `ninja`? Not sure I understand the nuance.

wantehchang: It is fine to use `meson compile`. I am not that familiar with meson. I was just pointing out that the official dav1d build instructions still use `ninja`.

wantehchang (approved):

PR #2669, opened 2025-03-05
Do not allow to keep the ICC when converting from RGB to gray

vrabaud:


vrabaud:
Indeed, this API is internal and only used by tests and exposed in avifenc. I changed the message.

PR #2670, opened 2025-03-05
avidec: output all frames of animations when passed `--index all`

maryla-uc:
Also refactor avifdec.c to remove duplication between --info mode and normal mode.

#2634

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -15,6 +15,7 @@
 #include <string.h>
 
 #define DEFAULT_JPEG_QUALITY 90
+#define DECODE_ALL_FRAMES -1
 
 #define NEXTARG()                                                     \
     if (((argIndex + 1) == argc) || (argv[argIndex + 1][0] == '-')) { \
@@ -54,39 +55,32 @@ static void syntax(void)
     avifPrintVersions();
 }
 
-int avifWriteToFile(avifAppFileFormat outputFormat,
-                    const char * outputFilename,
-                    avifImage * image,
-                    avifBool rawColor,
-                    int jpegQuality,
-                    int pngCompressionLevel,
-                    int requestedDepth,
-                    int chromaUpsampling)
+avifBool avifWriteToFile(avifAppFileFormat outputFormat,
+                         const char * outputFilename,
+                         avifImage * image,
+                         avifBool rawColor,
+                         int jpegQuality,
+                         int pngCompressionLevel,
+                         int requestedDepth,
+                         int chromaUpsampling)
 {
     if (outputFormat == AVIF_APP_FILE_FORMAT_Y4M) {
         if (image->icc.size || image->exif.size || image->xmp.size) {
             fprintf(stderr, "Warning: metadata dropped when saving to y4m.\n");
         }
-        if (!y4mWrite(outputFilename, image)) {
-            return AVIF_FALSE;
-        }
+        return y4mWrite(outputFilename, image);
     } else if (outputFormat == AVIF_APP_FILE_FORMAT_JPEG) {
         // Bypass alpha multiply step during conversion
         if (rawColor) {
             image->alphaPremultiplied = AVIF_TRUE;
         }
-        if (!avifJPEGWrite(outputFilename, image, jpegQuality, chromaUpsampling)) {
-            return AVIF_FALSE;
-        }
+        return avifJPEGWrite(outputFilename, image, jpegQuality, chromaUpsampling);
     } else if (outputFormat == AVIF_APP_FILE_FORMAT_PNG) {
-        if (!avifPNGWrite(outputFilename, image, requestedDepth, chromaUpsampling, pngCompressionLevel)) {
-            return AVIF_FALSE;
-        }
+        return avifPNGWrite(outputFilename, image, requestedDepth, chromaUpsampling, pngCompressionLevel);
     } else {
         fprintf(stderr, "Unsupported output file extension: %s\n", outputFilename);
         return AVIF_FALSE;
     }
-    return AVIF_TRUE;
 }
 
 int main(int argc, char * argv[])
@@ -213,7 +207,7 @@ int main(int argc, char * argv[])
         } else if (!strcmp(arg, "--index")) {
             NEXTARG();
             if (!strcmp(arg, "all")) {
-                frameIndex = -1;
+                frameIndex = DECODE_ALL_FRAMES;
             } else {
                 frameIndex = (uint32_t)atoi(arg);
             }
@@ -360,10 +354,10 @@ int main(int argc, char * argv[])
     }
 
     if (infoOnly && !frameIndexSpecified) {
-        frameIndex = -1; // Decode all frames by default in 'info only' mode.
+        frameIndex = DECODE_ALL_FRAMES; // Decode all frames by default in 'info only' mode.
     }
 
-    const avifBool decodeAllFrames = frameIndex == -1;
+    const avifBool decodeAllFrames = frameIndex == DECODE_ALL_FRAMES;
     int currIndex = decodeAllFrames ? 0 : frameIndex;
     while (AVIF_TRUE) {
         result = decodeAllFrames ? avifDecoderNextImage(decoder) : avifDecoderNthImage(decoder, frameIndex);
@@ -417,7 +411,7 @@ int main(int argc, char * argv[])
         }
 
         if (decodeAllFrames) {
-            // Create filename for individual frames, in the form path/to/output-000.ext
+            // Create filename for individual frames, in the form path/to/output-0000000000.ext
             char * lastDot = strrchr(outputFilename, '.');
             const size_t dotPos = (lastDot != NULL) ? (size_t)(lastDot - outputFilename) : strlen(outputFilename);
             const char * extension = (lastDot != NULL) ? lastDot + 1 : "";
@@ -457,7 +451,10 @@ int main(int argc, char * argv[])
         if (decodeAllFrames) {
             result = AVIF_RESULT_OK;
         } else {
-            fprintf(stderr, "ERROR: Frame at index %d requested but the file does not contain enough frames\n", frameIndex);
+            fprintf(stderr,
+                    "ERROR: Frame at index %d requested but the file does not contain enough frames (signalled frame count: %d)\n",
+                    frameIndex,
+                    decoder->imageCount);
             goto cleanup;
         }
     }
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -10,6 +10,7 @@
 
 #include <assert.h>
 #include <inttypes.h>
+#include <math.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


apps/avifdec.c
```diff
@@ -54,6 +54,41 @@ static void syntax(void)
     avifPrintVersions();
 }
 
+int avifWriteToFile(avifAppFileFormat outputFormat,
```
y-guyon, line 57: Could be `avifBool` instead of `int`

apps/avifdec.c
```diff
@@ -54,6 +54,41 @@ static void syntax(void)
...
+                    int requestedDepth,
+                    int chromaUpsampling)
+{
+    if (outputFormat == AVIF_APP_FILE_FORMAT_Y4M) {
+        if (image->icc.size || image->exif.size || image->xmp.size) {
+            fprintf(stderr, "Warning: metadata dropped when saving to y4m.\n");
+        }
+        if (!y4mWrite(outputFilename, image)) {
+            return AVIF_FALSE;
+        }
```
y-guyon, line 72: Why not just `return y4mWrite(outputFilename, image)` directly?

apps/avifdec.c
```diff
@@ -337,78 +333,139 @@ int main(int argc, char * argv[])
...
-        printf("[--icc] Setting ICC profile: %s\n", iccOverrideFilename);
-        result = avifImageSetProfileICC(decoder->image, iccOverride.data, iccOverride.size);
-        avifRWDataFree(&iccOverride);
+    }
+
+    if (infoOnly && !frameIndexSpecified) {
+        frameIndex = -1; // Decode all frames by default in 'info only' mode.
+    }
+
+    const avifBool decodeAllFrames = frameIndex == -1;
```
y-guyon, line 366: -1 could be a constant `ALL_FRAMES`

apps/avifdec.c
```diff
@@ -231,87 +273,38 @@ int main(int argc, char * argv[])
...
+    if (!inputFilename) {
+        fprintf(stderr, "Missing input filename\n");
+        syntax();
+        return 1;
+    }
+
+    avifAppFileFormat outputFormat = AVIF_APP_FILE_FORMAT_UNKNOWN;
     if (infoOnly) {
-        if (!inputFilename || outputFilename) {
+        if (outputFilename) {
```
y-guyon, line 284: Could be done for `iccOverrideFilename` but not worth it

apps/avifdec.c
```diff
@@ -337,78 +333,139 @@ int main(int argc, char * argv[])
...
+                // Should happen only if AVIF_STRICT_CLAP_VALID is disabled.
+                fprintf(stderr, "Warning: Invalid Clean Aperture values\n");
+            }
         }
-    } else if (outputFormat == AVIF_APP_FILE_FORMAT_JPEG) {
-        // Bypass alpha multiply step during conversion
-        if (rawColor) {
-            decoder->image->alphaPremultiplied = AVIF_TRUE;
+
+        if (ignoreICC && (decoder->image->icc.size > 0)) {
```
y-guyon, line 403: Note: I am surprised there is `ignoreExif` but no `ignoreICC` in `avifDecoder`.

apps/avifdec.c
```diff
@@ -337,78 +333,139 @@ int main(int argc, char * argv[])
...
+                fprintf(stderr, "ERROR: Failed to set ICC: %s\n", avifResultToString(result));
+                goto cleanup;
+            }
         }
-    } else if (outputFormat == AVIF_APP_FILE_FORMAT_PNG) {
-        if (!avifPNGWrite(outputFilename, decoder->image, requestedDepth, chromaUpsampling, pngCompressionLevel)) {
+
+        if (decodeAllFrames) {
+            // Create filename for individual frames, in the form path/to/output-000.ext
+            char * lastDot = strrchr(outputFilename, '.');
```
y-guyon, line 421: I think you need to look for the last `.` after the last path separator `/` (or `\`). Otherwise `path/to.some/output` may end as `path/to.some-000/output` and fail.

apps/avifdec.c
```diff
@@ -337,78 +333,139 @@ int main(int argc, char * argv[])
...
+            const size_t dotPos = (lastDot != NULL) ? (size_t)(lastDot - outputFilename) : strlen(outputFilename);
```
y-guyon, line 422: I assume this works with non-ASCII character sets.

apps/avifdec.c
```diff
@@ -337,78 +333,139 @@ int main(int argc, char * argv[])
...
+            if (result != AVIF_RESULT_OK) {
+                fprintf(stderr, "ERROR: Failed to set ICC: %s\n", avifResultToString(result));
+                goto cleanup;
+            }
         }
-    } else if (outputFormat == AVIF_APP_FILE_FORMAT_PNG) {
-        if (!avifPNGWrite(outputFilename, decoder->image, requestedDepth, chromaUpsampling, pngCompressionLevel)) {
+
+        if (decodeAllFrames) {
+            // Create filename for individual frames, in the form path/to/output-000.ext
```
y-guyon, line 420: `path/to/output-0000000000.ext`

apps/avifdec.c
```diff
@@ -337,78 +333,139 @@ int main(int argc, char * argv[])
...
+            break;
+        }
+        ++currIndex;
+    }
+
+    if (result == AVIF_RESULT_NO_IMAGES_REMAINING) {
+        if (decodeAllFrames) {
+            result = AVIF_RESULT_OK;
+        } else {
+            fprintf(stderr, "ERROR: Frame at index %d requested but the file does not contain enough frames\n", frameIndex);
```
y-guyon, line 460: `decoder->imageCount` could be mentioned in this error

apps/avifdec.c
```diff
@@ -337,78 +333,139 @@ int main(int argc, char * argv[])
...
+            }
         }
-    } else if (outputFormat == AVIF_APP_FILE_FORMAT_PNG) {
-        if (!avifPNGWrite(outputFilename, decoder->image, requestedDepth, chromaUpsampling, pngCompressionLevel)) {
+
+        if (decodeAllFrames) {
+            // Create filename for individual frames, in the form path/to/output-000.ext
+            char * lastDot = strrchr(outputFilename, '.');
+            const size_t dotPos = (lastDot != NULL) ? (size_t)(lastDot - outputFilename) : strlen(outputFilename);
+            const char * extension = (lastDot != NULL) ? lastDot + 1 : "";
```
y-guyon, line 423: Not sure if a feature or not but passing a folder will have different behaviors depending on whether it ends with a `/` or not:
- `some/output` will create files `some/output0000000000`
- `some/output/` will create files `some/output/0000000000`

This may be tricky but fine. It is a shame that this is not C++ where we could leverage `std::filesystem` for all this.

doc/avifdec.1.md
```diff
@@ -63,7 +63,7 @@ Output format can be either JPEG, PNG or YUV4MPEG2 (Y4M).
     (JPEG only; not applicable to y4m).
 
 **\--index** _I_
-:   When decoding an image sequence or progressive image, specify which frame index to decode (Default: 0).
+:   When decoding an image sequence or progressive image, specify which frame index to decode, where the first frame has index 0, or 'all' to decode all frames. (Default: all)
```
y-guyon, line 66: `(Default: 0)`

y-guyon (approved):
Thank you for adding the feature!

apps/avifdec.c
maryla-uc: Ack

apps/avifdec.c
maryla-uc: I just tried it and it seems to work fine yes.

apps/avifdec.c
maryla-uc: You can't pass in a directory because avifdec wouldn't be able to determine the output file type. In both of your examples it will fail with 'Cannot determine output file extension'

apps/avifdec.c
maryla-uc: The reason I didn't add it originally was because in my understanding, we don't know if that numer matches the actual number of frames. But you're right that it's useful information in most cases.

apps/avifdec.c
maryla-uc: Indeed, that could happen... but only if path/to.some/output is a file that already exists and is a jpeg or png or y4m, so that its type can be guessed from the first bytes (since it can't be guessed from the extension "some/output") which seems like a really far fetched corner case to me. Otherwise avifdec will fail with "Cannot determine output file extension" before we reach this point.

I think what we should do is only use the file extension, and not the first bytes, to guess the output filename in avifdec. This way we're guarantteed that the filename is sensible. Using the first bytes makes sense for avifenc's input files but for output files it seems strange to me. However this would be a small behavior change, that may break someone somewhere? (http://xkcd.com/1172 ...) I suspect it's ok to change though.

maryla-uc:
Thanks for the thorough review!

apps/avifdec.c
y-guyon: Looking at the code it seems to be equal to the number of track samples. I have no idea if that is equal to the number of frames in an AVIF image sequence.

apps/avifdec.c
y-guyon: I thought there was some `avifdec --output_format png` but I was mistaken.

y-guyon (approved):

apps/avifdec.c
maryla-uc: I think that in theory they should be the same, but the actual movie data might end up having a different number of frames if non compliant? In --info mode we print "(XXX expected frames)" before actually decoding the frames. But this is just a corner case.

apps/avifdec.c
```diff
@@ -337,78 +327,142 @@ int main(int argc, char * argv[])
...
+            // Create filename for individual frames, in the form path/to/output-0000000000.ext
+            char * lastDot = strrchr(outputFilename, '.');
+            const size_t dotPos = (lastDot != NULL) ? (size_t)(lastDot - outputFilename) : strlen(outputFilename);
+            const char * extension = (lastDot != NULL) ? lastDot + 1 : "";
+            const int maxFilenameWithoutExtensionLength = 1000;
+            const int maxExtensionLength = 10;
+            char frameFilename[1024];
+            int res = snprintf(frameFilename,
+                               sizeof(frameFilename),
+                               "%.*s-%010d.%.*s",
```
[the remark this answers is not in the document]
maryla-uc: > You could shorten this based on the max number of frames if that's available, but otherwise this seems OK.

I did that at first but we decided that having a predictible number of zeros (and therefore filename) would be easier for scripts. And that we could implement the "%03d" syntax if someone ever asked for it, but it really seems like a "nice to have".

apps/avifdec.c
```diff
@@ -38,7 +39,7 @@ static void syntax(void)
     printf("    -u,--upsampling U : Chroma upsampling (for 420/422). One of 'automatic' (default), 'fastest', 'best', 'nearest', or 'bilinear'\n");
     printf("    -r,--raw-color    : Output raw RGB values instead of multiplying by alpha when saving to opaque formats\n");
     printf("                        (JPEG only; not applicable to y4m)\n");
-    printf("    --index I         : When decoding an image sequence or progressive image, specify which frame index to decode (Default: 0)\n");
+    printf("    --index I         : When decoding an image sequence or progressive image, specify which frame index to decode, where the first frame has index 0, or 'all' to decode all frames. (Default: 0)\n");
```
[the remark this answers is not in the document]
maryla-uc: Indeed, I forgot (again)

apps/avifdec.c
```diff
@@ -54,6 +55,34 @@ static void syntax(void)
...
+                         int pngCompressionLevel,
+                         int requestedDepth,
+                         int chromaUpsampling)
+{
+    if (outputFormat == AVIF_APP_FILE_FORMAT_Y4M) {
+        if (image->icc.size || image->exif.size || image->xmp.size) {
+            fprintf(stderr, "Warning: metadata dropped when saving to y4m.\n");
+        }
+        return y4mWrite(outputFilename, image);
+    } else if (outputFormat == AVIF_APP_FILE_FORMAT_JPEG) {
```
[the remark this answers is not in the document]
maryla-uc: It can but I'm not sure it makes the code inhrently better? It's a code style preference, and there's nothing in the styleguide about this AFAIK. I prefer it this way, it's more compact and I find it clearer.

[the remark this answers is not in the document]
jzern: As long as things are consistent. You're right, it's not in the style guide, but a [clang-tidy warning](https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html).

PR #2672, opened 2025-03-07
Add missing changelog for `avifdec --index all`

maryla-uc:
Fix int => avifBool.


y-guyon (approved):

jzern (commented):
LGTM. Thanks for following up.

PR #2673, opened 2025-03-07
Fix monochrome ICC profile error message in apps

y-guyon:
Add CHANGELOG entry for https://github.com/AOMediaCodec/libavif/pull/2669.

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -911,7 +911,7 @@ static avifBool avifJPEGReadInternal(FILE * f,
             iccData = iccDataTmp;
             if (requestedFormat == AVIF_PIXEL_FORMAT_YUV400) {
                 fprintf(stderr,
-                        "The image contains an RGB ICC profile which is incompatible with the requested output "
+                        "The image contains an ICC profile which is incompatible with the requested output "
                         "format YUV400 (grayscale). Pass --ignore-icc to discard the ICC profile.\n");
                 goto cleanup;
             }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 
 ### Added since 1.2.0
 
-* Add support for outputting all frames of an image sequence in `avifdec`. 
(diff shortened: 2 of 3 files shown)


apps/shared/avifjpeg.c
```diff
@@ -911,7 +911,7 @@ static avifBool avifJPEGReadInternal(FILE * f,
             iccData = iccDataTmp;
             if (requestedFormat == AVIF_PIXEL_FORMAT_YUV400) {
                 fprintf(stderr,
-                        "The image contains an RGB ICC profile which is incompatible with the requested output "
+                        "The image contains an ICC profile which is incompatible with the requested output "
```
maryla-uc: Grayscale ICC profiles are also a thing (see iccmaker.h/c). It's true that here we don't actually check the type of profile, just that there is a profile. But I guess it's safe to assume that if the original image was not grayscale, then the icc is not grayscale. We handle grayscale jpegs, right? Is this code path triggered then as well? Because then it probably shouldn't be... I guess this is for another PR if something needs to be fixed.

maryla-uc (approved):

apps/shared/avifjpeg.c
maryla-uc: Actually if this is the main goal of this PR then it should be in this PR... (or dropped)

vrabaud (approved):

apps/shared/avifjpeg.c
y-guyon: > We handle grayscale jpegs, right? Is this code path triggered then as well?

No idea.

> I guess this is for another PR if something needs to be fixed.

Yes.

> Actually if this is the main goal of this PR then it should be in this PR... (or dropped)

No, just "RGB ICC" looked wrong, especially when reading a JPEG.

maryla-uc: IMO it's a bit misleading to say that ICC is incompatible with grayscale, because fundamentally it's not, but we can say this is meant to say that *we* don't support it...

y-guyon: Feel free to send a PR.

PR #2678, opened 2025-03-10
Use unspecified matrix coeffs in avifsvttest

y-guyon:
4:2:0 with identity matrix coefficients are not AV1 compliant and SVT-AV1 does not support 4:4:4.


jzern (approved):
This works. Another option would be to use YUV444 in the case of lossless.

jzern:
>> Another option would be to use YUV444 in the case of lossless.
> [...] and SVT-AV1 does not support 4:4:4.

Never mind then.

PR #2681, opened 2025-03-10
test_cmd_stdin: extract mdat offset automatically

jzern:


jzern:
```
++ grep -b -m 1 -o --text mdat avif_test_cmd_stdin_encoded.avif
++ cut -d: -f 1
+ MDAT_OFFSET=
+ dd if=avif_test_cmd_stdin_encoded.avif of=avif_test_cmd_stdin_encoded.avif.strip bs=1 skip=
dd: no value specified for skip
```

I'll need to have a closer look on macOS. A quick test outside of the script seemed to work.

jzern:
We can hold this until after the release to be safe.

tests/test_cmd_stdin.sh
```diff
@@ -35,11 +35,8 @@ strip_header_if() {
   STRIP_HEADER="$2"
 
   if ${STRIP_HEADER}; then
-    # The following does not work on all platforms:
-    #   grep --text mdat "${FILE}"
-    #   awk -b -v RS='mdat' '{print length($0); exit}' "${FILE}"
-    # Hence the hardcoded variable below.
-    MDAT_OFFSET=1061
+    MDAT_OFFSET=$(GREP_OPTIONS="" LC_ALL=C \
```
jzern, line 38: `LC_ALL` is key here, but since I was adding that I figured `GREP_OPTIONS` would make it more future proof. Another option would be to add `-h` explicitly to avoid any chance of the filename getting added to the output.

y-guyon (approved):
Thank you for the change.

> We can hold this until after the release to be safe.

Future-proofing a test is fine. It can be merged now to trigger postsubmit CI.

Issue #2685, opened 2025-03-13
Alpha encoded with AOM_CICP_MC_IDENTITY


y-guyon:
For color:

https://github.com/AOMediaCodec/libavif/blob/632b7eacdf428ec3a294808df08475fd49225b44/src/codec_aom.c#L1071

But for alpha, `mc` is left to its default value (0 because of `memset()` above), which means AOM_CICP_MC_IDENTITY, which is [forbidden](https://aomediacodec.github.io/av1-spec/av1-spec.pdf#page=131):

> subsampling_x |subsampling_y |mono_chrome |Description
> -|-|-|-
> 0 |0 |0 |YUV 4:4:4
> 1 |0 |0 |YUV 4:2:2
> 1 |1| 0 |YUV 4:2:0
> 1 |1 |1 |Monochrome 4:0:0
> 
> If matrix_coefficients is equal to MC_IDENTITY, it is a requirement of bitstream conformance that subsampling_x is equal to 0 and subsampling_y is equal to 0.

wantehchang:
It comes as a surprise that the libaom encoder ignores the `cp`, `tc`, and `mc` fields of the `aom_image_t` struct input passed to `aom_codec_encode()`. Although we set `aomImage.mc` to the wrong value (0 = `AOM_CICP_MC_IDENTITY`), it is ignored by the libaom encoder.

Have you verified that the libavif encoder writes MC=0 to the AV1 bitstream of an alpha auxiliary image?

Also, the libavif encoder should not write a `colr` box for an alpha auxiliary image.

y-guyon:
> It comes as a surprise that the libaom encoder ignores the `cp`, `tc`, and `mc` fields of the `aom_image_t` struct input passed to `aom_codec_encode()`. Although we set `aomImage.mc` to the wrong value (0 = `AOM_CICP_MC_IDENTITY`), it is ignored by the libaom encoder.

Good to know.

> Have you verified that the libavif encoder writes MC=0 to the AV1 bitstream of an alpha auxiliary image?

No. I checked that setting `aom_image_t::mc` to different values has no impact on the generated file though.  
The only change that impacts the generated file is

```c
aom_codec_control(&codec->internal->encoder, AV1E_SET_MATRIX_COEFFICIENTS, AOM_CICP_MC_UNSPECIFIED);
```

> Also, the libavif encoder should not write a `colr` box for an alpha auxiliary image.

[...]

PR #2687, opened 2025-03-14
Encode alpha with unspecified matrix coefficients

y-guyon:
Forward CICP to SVT-AV1.
Fix typo in aviflosslesstest.

Fixes https://github.com/AOMediaCodec/libavif/issues/2685.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ The changes are relative to the previous release, unless the baseline is specifi
   discarded.
 * Forbid encoding with AVIF_MATRIX_COEFFICIENTS_IDENTITY and
   AVIF_PIXEL_FORMAT_YUV400 to be AV1 spec compatible.
+* Forward the CICP color primaries, transfer characteristics, and matrix
+  coefficients to the SVT-AV1 encoder.
 * Update svt.cmd/svt.sh/LocalSvt.cmake to v3.0.1.
 
 ## [1.2.0] - 2025-02-25
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -1025,6 +1025,13 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     avifBool monochromeRequested = AVIF_FALSE;
 
     if (alpha) {
+        // AV1 specification section 6.4.2. Color config semantics:
+        //   subsampling_x | subsampling_y | mono_chrome | Description
+        //   1             | 1             | 1           | Monochrome 4:0:0
+        //   If matrix_coefficients is equal to MC_IDENTITY, it is a requirement of bitstream
+        //   conformance that subsampling_x is equal to 0 and subsampling_y is equal to 0.
+        // Use AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED instead.
+        aomImage.mc = (aom_matrix_coefficients_t)AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
         aomImage.range = AOM_CR_FULL_RANGE;
         monochromeRequested = AVIF_TRUE;
         if (aomImageAllocated) {
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -920,6 +920,13 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
     avifBool monochromeRequested = AVIF_FALSE;
 
     if (alpha) {
+        // AV1 specification section 6.4.2. Color config semantics:
+        //   subsampling_x | subsampling_y | mono_chrome | Description
+        //   1             | 1             | 1           | Monochrome 4:0:0
+        //   If matrix_coefficients is equal to MC_IDENTITY, it is a requirement of bitstream
+        //   conformance that subsampling_x is equal to 0 and subsampling_y is equal to 0.
+        // Use AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED instead.
+        aomImage.mc = (aom_matrix_coefficients_t)AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
         aomImage.range = AOM_CR_FULL_RANGE;
         monochromeRequested = AVIF_TRUE;
         if (aomImageAllocated) {
--- a/src/codec_rav1e.c
+++ b/src/codec_rav1e.c
@@ -89,12 +89,29 @@ static avifResult rav1eCodecEncodeImage(avifCodec * codec,
 
     if (!codec->internal->rav1eContext) {
         const avifBool supports400 = rav1eSupports400();
+        RaColorPrimaries rav1eCp;
+        RaTransferCharacteristics rav1eTf;
+        RaMatrixCoefficients rav1eMc;
         RaPixelRange rav1eRange;
         if (alpha) {
+            rav1eCp = (RaColorPrimaries)AVIF_COLOR_PRIMARIES_UNSPECIFIED;
+            rav1eTf = (RaTransferCharacteristics)AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
+            // AV1 specification section 6.4.2. Color config semantics:
+            //   subsampling_x | subsampling_y | mono_chrome | Description
+            //   1             | 1             | 1           | Monochrome 4:0:0
+            //   If matrix_coefficients is equal to MC_IDENTITY, it is a requirement of bitstream
+            //   conformance that subsampling_x is equal to 0 and subsampling_y is equal to 0.
+            // Use AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED instead.
+            rav1eMc = (RaMatrixCoefficients)AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
+
             rav1eRange = RA_PIXEL_RANGE_FULL;
             codec->internal->chromaSampling = supports400 ? RA_CHROMA_SAMPLING_CS400 : RA_CHROMA_SAMPLING_CS420;
             codec->internal->yShift = 1;
         } else {
+            rav1eCp = (RaColorPrimaries)image->colorPrimaries;
+            rav1eTf = (RaTransferCharacteristics)image->transferCharacteristics;
+            rav1eMc = (RaMatrixCoefficients)image->matrixCoefficients;
+
             rav1eRange = (image->yuvRange == AVIF_RANGE_FULL) ? RA_PIXEL_RANGE_FULL : RA_PIXEL_RANGE_LIMITED;
             codec->internal->yShift = 0;
             switch (image->yuvFormat) {
@@ -186,10 +203,7 @@ static avifResult rav1eCodecEncodeImage(avifCodec * codec,
             }
         }
 
-        rav1e_config_set_color_description(rav1eConfig,
-                                           (RaMatrixCoefficients)image->matrixCoefficients,
-                                           (RaColorPrimaries)image->colorPrimaries,
-                                           (RaTransferCharacteristics)image->transferCharacteristics);
+        rav1e_config_set_color_description(rav1eConfig, rav1eMc, rav1eCp, rav1eTf);
 
         codec->internal->rav1eContext = rav1e_context_new(rav1eConfig);
         if (!codec->internal->rav1eContext) {
--- a/src/codec_svt.c
+++ b/src/codec_svt.c
@@ -83,11 +83,28 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
     EbErrorType res = EB_ErrorNone;
 
     int y_shift = 0;
+    EbColorPrimaries color_primaries;
+    EbTransferCharacteristics transfer_characteristics;
+    EbMatrixCoefficients matrix_coefficients;
     EbColorRange svt_range;
     if (alpha) {
+        color_primaries = (EbColorPrimaries)AVIF_COLOR_PRIMARIES_UNSPECIFIED;
+        transfer_characteristics = (EbTransferCharacteristics)AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
+        // AV1 specification section 6.4.2. Color config semantics:
+        //   subsampling_x | subsampling_y | mono_chrome | Description
+        //   1             | 1             | 1           | Monochrome 4:0:0
+        //   If matrix_coefficients is equal to MC_IDENTITY, it is a requirement of bitstream
+        //   conformance that subsampling_x is equal to 0 and subsampling_y is equal to 0.
+        // Use AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED instead.
+        matrix_coefficients = (EbMatrixCoefficients)AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
+
         svt_range = EB_CR_FULL_RANGE;
         y_shift = 1;
     } else {
+        color_primaries = (EbColorPrimaries)image->colorPrimaries;
+        transfer_characteristics = (EbTransferCharacteristics)image->transferCharacteristics;
+        matrix_coefficients = (EbMatrixCoefficients)image->matrixCoefficients;
+
         svt_range = (image->yuvRange == AVIF_RANGE_FULL) ? EB_CR_FULL_RANGE : EB_CR_STUDIO_RANGE;
         switch (image->yuvFormat) {
             case AVIF_PIXEL_FORMAT_YUV444:
@@ -125,6 +142,9 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
         }
         svt_config->encoder_color_format = color_format;
         svt_config->encoder_bit_depth = (uint8_t)image->depth;
+        svt_config->color_primaries = color_primaries;
+        svt_config->transfer_characteristics = transfer_characteristics;
+        svt_config->matrix_coefficients = matrix_coefficients;
         svt_config->color_range = svt_range;
 #if !SVT_AV1_CHECK_VERSION(0, 9, 0)
         svt_config->is_16bit_pipeline = image->depth > 8;
--- a/tests/gtest/aviflosslesstest.cc
+++ b/tests/gtest/aviflosslesstest.cc
@@ -115,7 +115,7 @@ TEST_P(LosslessTest, EncodeDecode) {
   ASSERT_TRUE(testutil::AreImagesEqual(*image, *decoded));
 }
 
-// Reads an image losslessly, using identiy MC.
+// Reads an image losslessly, using identity MC.
 ImagePtr ReadImageLossless(const std::string& path,
                            avifPixelFormat requested_format,
                            avifBool ignore_icc) {


src/codec_aom.c
```diff
@@ -1025,6 +1025,13 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     avifBool monochromeRequested = AVIF_FALSE;
 
     if (alpha) {
+        // AV1 specification section 6.4.2. Color config semantics:
+        //   subsampling_x | subsampling_y | mono_chrome | Description
+        //   1             | 1             | 1           | Monochrome 4:0:0
+        //   If matrix_coefficients is equal to MC_IDENTITY, it is a requirement of bitstream
+        //   conformance that subsampling_x is equal to 0 and subsampling_y is equal to 0.
+        // Use AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED instead.
+        aomImage.mc = (aom_matrix_coefficients_t)AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
```
vrabaud, line 1034: I agree unspecified should be a good pick because according to h273, "MatrixCoefficients describes the matrix coefficients used in deriving luma and chroma signals from the green, blue and 
red or X, Y and Z primaries."

Now, we consider that grayscale (400), should have BT601 or BT$&)BG (cf https://github.com/AOMediaCodec/libavif/blob/086e7e127be93b7595fa2a2144bc6afb23ff3e22/src/reformat_libyuv.c#L278-L279  or https://github.com/AOMediaCodec/libavif/blob/086e7e127be93b7595fa2a2144bc6afb23ff3e22/src/reformat_libyuv.c#L762 ).

@wantehchang , any opinion on that?

jzern: In a change in another project I chose BT.601 (see cl/730991804).

vrabaud: Thx, is there a standard/good practice for that? MC do not make sense for grayscale ...

wantehchang: Yannis is right. The alpha auxiliary image is not converted to RGB, so MC should definitely be set to Unspecified (2) for alpha.

src/codec_aom.c
```diff
@@ -1025,6 +1025,13 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     avifBool monochromeRequested = AVIF_FALSE;
 
     if (alpha) {
+        // AV1 specification section 6.4.2. Color config semantics:
+        //   subsampling_x | subsampling_y | mono_chrome | Description
+        //   1             | 1             | 1           | Monochrome 4:0:0
+        //   If matrix_coefficients is equal to MC_IDENTITY, it is a requirement of bitstream
+        //   conformance that subsampling_x is equal to 0 and subsampling_y is equal to 0.
```
wantehchang, line 1032: This comment (lines 1028-1032) explains why we cannot use `AVIF_MATRIX_COEFFICIENTS__IDENTITY`, but it doesn't explain why we should use `AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED`. We should delete this comment. A better reason for using `AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED` is that the alpha image  is not converted to RGB, so `aomImage.mc` is ignored.

wantehchang: For the same reason, we should also set `aomImage.cp` and `aomImage.tc` to Unspecified (2) if alpha is true.

ronag:
Any chance this will land? I believe currently it's not possible to encode with alpha because of this?

y-guyon:
I am thinking of always setting CICP to 2/2/2 in Sequence Header OBU based on https://aomediacodec.github.io/av1-isobmff/#av1codecconfigurationbox-semantics allowing it. I will update this PR.

y-guyon:
> The sample entry SHOULD contain a colr box with a colour_type set to nclx. If a colr box with a colour_type set to nclx is present, the following applies:
> - If any of colour_primaries, transfer_characteristics, and matrix_coefficients is specified with a value other than 2 in the Sequence Header OBU (in the configOBUs field or in the associated samples), then the value of that field in the colr box SHALL match the value in the Sequence Header OBU.
> - If any of colour_primaries, transfer_characteristics, and matrix_coefficients is NOT specified or is specified with a value equal to 2 in the Sequence Header OBU, then the value of that field in the colr box MAY be different and the value in the colr box overrides the value in the bitstream, as specified in the ColourInformationBox definition in ISOBMFF.

wantehchang:
@ronag Robert: You asked:
> I believe currently it's not possible to encode with alpha because of this?

This is not true. libavif should be able to encode with alpha right now.

src/codec_aom.c
wantehchang: I am sorry I forgot to publish this comment earlier.

I believe the changes to codec_aom.c and codec_avm.c are just cosmetic because the libaom and AVM encoders ignore the `cp`, `tc`, `mc`, and `range` fields of `aom_image_t`. We need to call `aom_codec_control()` to instruct the libaom or AVM encoder to set these fields in the sequence header OBU, and the following code does that:

```
        // Set color_config() in the sequence header OBU.
        if (alpha) {
            aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_RANGE, AOM_CR_FULL_RANGE);
        } else {
            // libaom's defaults are AOM_CICP_CP_UNSPECIFIED, AOM_CICP_TC_UNSPECIFIED,
            // AOM_CICP_MC_UNSPECIFIED, AOM_CSP_UNKNOWN, and 0 (studio/limited range). Call
            // aom_codec_control() only if the values are not the defaults.
            if (image->colorPrimaries != AVIF_COLOR_PRIMARIES_UNSPECIFIED) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_PRIMARIES, (int)image->colorPrimaries);
            }
            if (image->transferCharacteristics != AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_TRANSFER_CHARACTERISTICS, (int)image->transferCharacteristics);
            }
            if (image->matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_MATRIX_COEFFICIENTS, (int)image->matrixCoefficients);
            }
            if (image->yuvChromaSamplePosition != AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_CHROMA_SAMPLE_POSITION, (int)image->yuvChromaSamplePosition);
            }
            if (image->yuvRange != AVIF_RANGE_LIMITED) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_RANGE, (int)image->yuvRange);
            }
        }
```

It shows that for `alpha` we never set the CP, TC, and MC fields in the sequence header OBU.

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ The changes are relative to the previous release, unless the baseline is specifi
   AVIF file has an ImageRotation property with angle set to 1 or 3, has no
   ImageMirror property, and carries an Exif chunk. Note that Exif orientation is
   usually ignored in PNG files, so this mainly impacts JPEG files.
+* Always forward Unspecified (2) CICP color primaries, transfer characteristics,
+  and matrix coefficients to the AV1 encoder. Rely on the 'colr' box instead.
```
wantehchang, line 29: The commit message still has the subject line "Encode alpha with unspecified matrix coefficients". Should it be updated to match this changelog entry?

wantehchang, line 29: Yannis: I will review this PR tomorrow. But a quick look shows that it doesn't do what it intends to do.

If we always forward CICP 2/2/2 to the AV1 encoder, then in src/codec_aom.c, the following if-else statement:

```
        // Set color_config() in the sequence header OBU.
        if (alpha) {
            aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_RANGE, AOM_CR_FULL_RANGE);
        } else {
            // libaom's defaults are AOM_CICP_CP_UNSPECIFIED, AOM_CICP_TC_UNSPECIFIED,
            // AOM_CICP_MC_UNSPECIFIED, AOM_CSP_UNKNOWN and 0 (studio/limited range). Call
            // aom_codec_control() only if the values are not the defaults.
            if (image->colorPrimaries != AVIF_COLOR_PRIMARIES_UNSPECIFIED) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_PRIMARIES, (int)image->colorPrimaries);
            }
            if (image->transferCharacteristics != AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_TRANSFER_CHARACTERISTICS, (int)image->transferCharacteristics);
            }
            if (image->matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_MATRIX_COEFFICIENTS, (int)image->matrixCoefficients);
            }
            if (image->yuvChromaSamplePosition != AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_CHROMA_SAMPLE_POSITION, (int)image->yuvChromaSamplePosition);
            }
            if (image->yuvRange != AVIF_RANGE_LIMITED) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_RANGE, (int)image->yuvRange);
            }
        }
```

should be simplified to the following:

```
        // Set color_config() in the sequence header OBU.
        if (alpha) {
            aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_RANGE, AOM_CR_FULL_RANGE);
        } else {
            // libaom's defaults are AOM_CSP_UNKNOWN, and 0 (studio/limited range). Call
            // aom_codec_control() only if the values are not the defaults.
            if (image->yuvChromaSamplePosition != AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_CHROMA_SAMPLE_POSITION, (int)image->yuvChromaSamplePosition);
            }
            if (image->yuvRange != AVIF_RANGE_LIMITED) {
                aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_RANGE, (int)image->yuvRange);
            }
        }
```

wantehchang: More details: In the AV1 spec, the color_config() structure has the following syntax elements:

```
    color_description_present_flag                       f(1)
    if ( color_description_present_flag ) {
        color_primaries                                  f(8)
        transfer_characteristics                         f(8)
        matrix_coefficients                              f(8)
    } else {
        color_primaries = CP_UNSPECIFIED
        transfer_characteristics = TC_UNSPECIFIED
        matrix_coefficients = MC_UNSPECIFIED
    }
```

If I understand it correctly, the goal of this PR is to cause the syntax element `color_description_present_flag` to have the value 0, so that the three syntax elements `color_primaries`, `transfer_characteristics`, and `matrix_coefficients` are all omitted. Is this correct?

y-guyon: "Encode alpha with unspecified matrix coefficients" still reflect the main intent of fixing #2685 in my opinion, but I do not mind merging this PR as "Set Unspecified CICP in Seq Header OBU".

wantehchang, line 29: I just realized an issue: right now we are relying on the AV1 encoders and decoders to check the following AV1 requirement:

After we start to ask the AV1 encoders to set `color_description_present_flag` to 0, we shoud check the above requirement in src/write.c and src/read.c. This can be done in a follow-up PR.

---

Another fact: After this PR, the AV1 encoders will not be able to use the following shortcut for lossless RGB compression:

```
        color_range = 1
        subsampling_x = 0
        subsampling_y = 0
    } else {
```

I think this is worthwhile because we will signal 1 or bits (`color_range` is always signaled, `subsampling_x` is signaled for 12 bits, `subsampling_y` isn't signaled).

wantehchang, line 29: Just FYI: I uploaded an AVIF file to the https://gpac.github.io/ComplianceWarden-wasm/ and it does not report any new errors. (It only reports a known error: https://github.com/AOMediaCodec/libavif/issues/2514.)

wantehchang (approved):
LGTM. Thanks!

CHANGELOG.md
y-guyon: > I just realized an issue: right now we are relying on the AV1 encoders and decoders to check the following AV1 requirement:
> 
> > If matrix_coefficients is equal to MC_IDENTITY, it is a requirement of bitstream conformance that subsampling_x is equal to 0 and subsampling_y is equal to 0.

This only applies to the AV1 bitstream and does not interact with AVIF, right? The mentioned `matrix_coefficients` is the one from the sequence header OBU, not the `nclx` `colr` box?

> After we start to ask the AV1 encoders to set `color_description_present_flag` to 0, we shoud check the above requirement in src/write.c and src/read.c. This can be done in a follow-up PR.

Assuming we still want to check that at the AVIF level, after this PR it should only be checked at reading, because at writing we will be always writing `MC_UNSPECIFIED`, thus making the first part `If matrix_coefficients is equal to MC_IDENTITY` never true. Unless I missed something and src/write.c should also be adapted?

---

> Another fact: After this PR, the AV1 encoders will not be able to use the following shortcut for lossless RGB compression:
> 
> ```
>     } else if ( color_primaries == CP_BT_709 &&
>                 transfer_characteristics == TC_SRGB &&
>                 matrix_coefficients == MC_IDENTITY ) {
>         color_range = 1
>         subsampling_x = 0
>         subsampling_y = 0
>     } else {
> ```
> 
> I think this is worthwhile because we will signal 1 or bits (`color_range` is always signaled, `subsampling_x` is signaled for 12 bits, `subsampling_y` isn't signaled).

Not sure where the 12 bits come from.  
From https://aomediacodec.github.io/av1-spec/av1-spec.pdf#page=44 I see:
- `color_description_present_flag=0` saves 8+8+8 bits
- `color_description_present_flag=1` and CICP 1/13/0 saves at most 1+1+1 bits

So keeping `color_description_present_flag=0` seems better.

src/codec_aom.c
```diff
@@ -1073,13 +1103,26 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
     }
 
+    // Section 2.3.4 of AV1-ISOBMFF says 'colr' with 'nclx' should be present and shall match CICP
+    // values in the Sequence Header OBU, unless the latter has 2/2/2 (Unspecified).
+    // So set CICP values to 2/2/2 (Unspecified) in the Sequence Header OBU for simplicity.
+    // It may also save 3 bytes since the AV1 encoder may set color_description_present_flag to 0
+    // (see Section 5.5.2 "Color config syntax" of the AV1 specification).
+    aomImage.cp = AOM_CICP_CP_UNSPECIFIED;
+    aomImage.tc = AOM_CICP_TC_UNSPECIFIED;
+    aomImage.mc = AOM_CICP_MC_UNSPECIFIED;
```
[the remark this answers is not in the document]
y-guyon: I removed this block and added a sentence to the above comment.

tests/gtest/avifcicptest.cc
```diff
@@ -0,0 +1,108 @@
...
+
+INSTANTIATE_TEST_SUITE_P(
+    UnknownIdentity, CicpTest,
+    // Identity MC require 4:4:4 and AVIF_CODEC_CHOICE_SVT only supports 4:2:0.
+    testing::Combine(testing::Values(AVIF_CODEC_CHOICE_AOM,
+                                     AVIF_CODEC_CHOICE_RAV1E),
+                     testing::Values(static_cast<avifColorPrimaries>(
+                         AVIF_COLOR_PRIMARIES_UNKNOWN)),
+                     testing::Values(static_cast<avifTransferCharacteristics>(
+                         AVIF_TRANSFER_CHARACTERISTICS_UNKNOWN)),
```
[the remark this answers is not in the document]
y-guyon: Testing 0/0/0 is intended in this suite. The suite below is for 2/2/2.

src/codec_avm.c
```diff
@@ -625,26 +625,41 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
 
         // Set color_config() in the sequence header OBU.
         if (alpha) {
+            // AV1-AVIF specification, Section 4 "Auxiliary Image Items and Sequences":
```
jzern, line 628: Should all of these be AVIF now for consistency?

It seems on github, the repository is av1-avif and the README has the same title. The document itself is 'AV1 Image File Format (AVIF)' so AVIF is fine for now. Hopefully future revisions for extra codecs aren't called 'AVIF'.

jzern (approved):

wantehchang (approved):
LGTM. Thanks!

CHANGELOG.md
y-guyon: > When we set `color_description_present_flag` to 0, the CP/TC/MC fields in the nclx colr box are effectively the `color_primaries`, `transfer_characteristics`, and `matrix_coefficients` syntax elements that would be in the AV1 bitstream if `color_description_present_flag` were 1. Do you agree?

I disagree. https://aomediacodec.github.io/av1-spec/av1-spec.pdf#page=129 just says

```
color_description_present_flag equal to 1 specifies that color_primaries, transfer_characteristics, and
matrix_coefficients are present. color_description_present_flag equal to 0 specifies that color_primaries,
transfer_characteristics and matrix_coefficients are not present.
```

Within the "AV1 realm", when `color_description_present_flag` is 0, we just have default values `color_primaries = CP_UNSPECIFIED`, `transfer_characteristics = TC_UNSPECIFIED`, and `matrix_coefficients = MC_UNSPECIFIED`.  
So when `color_description_present_flag` is 0, `matrix_coefficients` is NOT equal to `MC_IDENTITY`.

And "if `color_description_present_flag` were 1" (I do not really see what can be deduced from that), AV1-ISBOMFF mandates that CP/TC/MC in OBU shall match CP/TC/MC in `nclx` `colr`, or shall match 2/2/2. So it "may still not be `MC_IDENTITY`".

> If you agree with the above, then it follows that the AV1 bitstream requirement should be checked with `matrix_coefficients` replaced by the MC field in the nclx color box. Since libaom checks this AV1 bitstream requirement on both the encoder and decoder sides, libavif should check this requirement in both src/write.c and src/read.c.

Why not let libaom check that then? To have a better libavif error message? To catch the issue earlier?

y-guyon:
Note that setting OBU to 2/2/2 works because the `colr` `nclx` box is always written:

https://github.com/AOMediaCodec/libavif/blob/fa089a0c3f8ed6e202034589cf98394e0226d66c/src/write.c#L686

CHANGELOG.md
wantehchang: libaom still checks that requirement, but libaom cannot check that requirement when libavif passes 2/2/2 instead of the actual CP/TC/MC to libaom (to save three bytes in the sequence header OBU).

While trying to create an example to explain this better, I found that libavif already has a check for this requiremennt. It is the following code in `avifValidateGrid()`:

```
        // AV1 (Version 1.0.0 with Errata 1), Section 6.4.2. Color config semantics
        // If matrix coefficients is equal to MC_IDENTITY, it is a requirement of bitstream conformance that subsampling_x is equal to 0 and subsampling_y is equal to 0.
        if (cellImage->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY && cellImage->yuvFormat != AVIF_PIXEL_FORMAT_YUV444) {
            avifDiagnosticsPrintf(diag, "subsampling must be 0 (4:4:4) with identity matrix coefficients");
        }
```

y-guyon: Suggestion: https://github.com/AOMediaCodec/libavif/pull/2766

wantehchang: I understand your point. We should keep this check. I clarified this check in https://github.com/AOMediaCodec/libavif/pull/2766#discussion_r2045622979.

PR #2691, opened 2025-03-14
Bump libavif to 1.2.1

y-guyon:


jzern (approved):

y-guyon:
Binaries tested in https://github.com/AOMediaCodec/libavif/pull/2695

PR #2700, opened 2025-03-19
avifpng.c: support ImageMagick app1 exif text data

jzern:
ImageMagick 6.9.13-12 uses lowercase app1 in the raw profile string.

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -173,7 +173,7 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool 
             }
             avifRemoveHeader(&xmpApp1Header, &avif->xmp); // Ignore the return value because the header is optional.
             *ignoreXMP = AVIF_TRUE;                       // Ignore any other XMP chunk.
-        } else if (!strcmp(text->key, "Raw profile type APP1")) {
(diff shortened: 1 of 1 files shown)


y-guyon (approved):

apps/shared/avifpng.c
```diff
@@ -173,7 +173,7 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
             }
             avifRemoveHeader(&xmpApp1Header, &avif->xmp); // Ignore the return value because the header is optional.
             *ignoreXMP = AVIF_TRUE;                       // Ignore any other XMP chunk.
-        } else if (!strcmp(text->key, "Raw profile type APP1")) {
+        } else if (!strcmp(text->key, "Raw profile type APP1") || !strcmp(text->key, "Raw profile type app1")) {
```
wantehchang, line 176: Optional: It would be good to copy the comments from the libwebp source about which tool (Exiftool or ImageMagick) uses which key:
https://chromium-review.googlesource.com/c/webm/libwebp/+/6330329

wantehchang (approved):

apps/shared/avifpng.c
jzern: Done.

PR #2701, opened 2025-03-21
codec_aom: disable loop restoration w/12-bit input

jzern:
The libaom encoder may produce integer overflows with 12-bit input when
loop restoration is enabled. This is only done with animation for now as
that is the only time the issue has surfaced.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -877,9 +877,11 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             }
         }
 
-        if (image->depth == 12) {
-            // The encoder may produce integer overflows with 12-bit input when loop restoration is enabled.
-            aom_codec_control(&codec->internal->encoder, AV1E_SET_ENABLE_RESTORATION, 0);
+        if (image->depth == 12 && cfg->g_limit != 1) {
+            // The encoder may produce integer overflows with 12-bit input when loop restoration is enabled. See crbug.com/aomedia/42302587.
+            if (aom_codec_control(&codec->internal->encoder, AV1E_SET_ENABLE_RESTORATION, 0) != AOM_CODEC_OK) {
+                return AVIF_RESULT_UNKNOWN_ERROR;
+            }
         }
     } else {
         avifBool dimensionsChanged = AVIF_FALSE;
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -876,6 +876,11 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 return AVIF_RESULT_UNKNOWN_ERROR;
             }
         }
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -876,6 +876,11 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 return AVIF_RESULT_UNKNOWN_ERROR;
             }
         }
+
+        if (image->depth == 12) {
+            // The encoder may produce integer overflows with 12-bit input when loop restoration is enabled.
```
wantehchang, line 881: Please reference the libaom bug crbug.com/aomedia/42302587.

src/codec_aom.c
```diff
@@ -876,6 +876,11 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+            aom_codec_control(&codec->internal->encoder, AV1E_SET_ENABLE_RESTORATION, 0);
```
wantehchang, line 882: Nit: Please check the return value. See line 875 for an example.

src/codec_aom.c
```diff
@@ -876,6 +876,11 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 return AVIF_RESULT_UNKNOWN_ERROR;
             }
         }
+
+        if (image->depth == 12) {
```
wantehchang, line 880: I wonder if we should only turn this off when we are encoding animations (`image->depth == 12 && cfg->g_limit != 1`).

wantehchang (approved):
LGTM. Thanks!

src/codec_aom.c
jzern: I agree. I wanted to put this up first since I copied from another area in this function. Overall it's a bit inconsistent.

src/codec_aom.c
jzern: We can start with that.

wantehchang: Plese ignore this suggestion. This needs more investigation.

src/codec_aom.c
jzern: Done.

src/codec_aom.c
jzern: Done. See line 859 for an example of an unchecked return.

src/codec_aom.c
jzern: Sorry, I think there was a race. If you're OK with it as is we can let fuzzing continue, otherwise I'll undo the check in a follow up.

jzern: This check satisfied the fuzzing cases I tested, but I posted https://github.com/AOMediaCodec/libavif/pull/2702 if you'd like it.

PR #2702, opened 2025-03-21
codec_aom: rm g_limit check when disabling loop resto

jzern:
This may not be the best way to describe animation and may not cover all
fuzzing cases.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -877,7 +877,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             }
         }
 
-        if (image->depth == 12 && cfg->g_limit != 1) {
+        if (image->depth == 12) {
             // The encoder may produce integer overflows with 12-bit input when loop restoration is enabled. See crbug.com/aomedia/42302587.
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -877,7 +877,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             }
         }
 
-        if (image->depth == 12 && cfg->g_limit != 1) {
+        if (image->depth == 12) {
```
wantehchang, line 880: The reason for removing the animation check is that it is not robust to future changes.

libaom's speed_features.c disables wiener filter for speed >= 5 in all intra mode. So if we encode still images with speed < 5, we will still call the search_wiener function in libaom's av1/encoder/pickrst.c. Apparently our fuzztests only use speed >=5. In avif_fuzztest_helpers.h I see:

```
  // Fuzz only a small range of 'speed' values to avoid slowing down the fuzzer
  // too much. The main goal is to fuzz libavif, not the underlying AV1 encoder.
  const auto speed = fuzztest::InRange(6, AVIF_SPEED_FASTEST);
```

If we change that to include speed < 5, then our fuzztests may start to report integer overflows when encoding still images.

wantehchang (approved):

PR #2705, opened 2025-03-24
Move a changelog item from 1.2.1 to Unreleased

wantehchang:
The changelog item was added to the wrong section in commit 564ee1f.


maryla-uc (approved):

maryla-uc:
Thanks!

PR #2707, opened 2025-03-26
avifrgbtoyuvtest: increase max_abs_average_diff

wantehchang:
libyuv commit c060118 "ARGBToJ444 use 256 for fixed point scale UV" causes the test to fail. Temporarily relax the max_abs_average_diff threshold to allow the test to pass while we investigate why the libyuv conversion results seem to have become less accurate.


wantehchang:
@fbarchard Frank: FYI.

jzern (approved):
This should be temporary assuming we make changes to the test after discussing how the error is calculated.

PR #2708, opened 2025-03-26
avifrgbtoyuvtest: Do not compute diff_sum

wantehchang:
The sum of pixel sample differences is not a good measure of distortion
because it does not satisfy all intuitive properties of the length of a
vector (the mathematical term is the norm of a vector). Do not compute
diff_sum.
    
Rename max_abs_average_diff to max_average_abs_diff and use it as the
maximum of the average absolute value of pixel sample difference.

### DIFF
--- a/tests/gtest/avifrgbtoyuvtest.cc
+++ b/tests/gtest/avifrgbtoyuvtest.cc
@@ -281,9 +281,6 @@ void ConvertWholeBuffer(int rgb_depth, int yuv_depth, avifRGBFormat rgb_format,
       num_diffs += src_rgb.width * src_rgb.height * 3;
     }
   }
-  // max_average_abs_diff is not tested here because it is not meaningful for
-  // only 3*3 conversions as it takes the maximum difference per conversion.
-  // PSNR is averaged on all pixels so it can be tested here.
   EXPECT_GE(GetPsnr(static_cast<double>(sq_diff_sum),
                     static_cast<double>(num_diffs), rgb_max),
             min_psnr);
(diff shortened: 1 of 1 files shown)


y-guyon (approved):

wantehchang (commented):
Yannis: Thanks for the review. Unfortunately I forgot to publish my questions for you yesterday, so I will have to ask you to take a look again. Sorry!

Note: I rebased the pull request to resolve merge conflicts. The code is the exactly the same as the version you reviewed yesterday.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -690,24 +686,24 @@ INSTANTIATE_TEST_SUITE_P(
...
+        /*max_average_abs_diff=*/Values(6.57),  // Sharp YUV introduces some
+                                                // color shift.
         /*min_psnr=*/Values(34.)  // SharpYuv distortion is acceptable.
         ));
 INSTANTIATE_TEST_SUITE_P(
     SharpYuv16Bit, RGBToYUVTest,
     Combine(
         /*rgb_depth=*/Values(16),
-        // TODO(yguyon): Why max_abs_average_diff>28 if RGB16 to YUV10 full rng?
+        // TODO(yguyon): Why max_average_abs_diff>28 if RGB16 to YUV10 full rng?
```
y-guyon, line 698: This is probably obsolete.

y-guyon (approved):
Note that `max_abs_average_diff` was vaguely testing how "centered" is the color drift of a round trip conversion (meaning it can be close to 0 if the error is huge but evenly distributed between negative and positive shifts over the whole range).

`max_average_abs_diff` is more testing how far the color drift goes for all samples (meaning it is close to 0 if the error is low for each value of the whole range).

wantehchang (commented):
Yannis: Thanks for the replies, especially your original comment (before you edited it) that clearly explains the comment about the "only 3*3 conversions".

PR #2711, opened 2025-03-27
read: Fix some return values

vigneshvg:
`AVIF_FALSE` is being returned from some functions which return
`avifResult`. `AVIF_FALSE` is `0` which translates to
`AVIF_RESULT_OK` when (accidentally) converted to `avifResult`
which is incorrect.

I think this was a copy-paste error from functions which were
previously returning `avifBool`.


wantehchang (approved):
LGTM. Thanks for the fx! This is a common bug in C libraries.

maryla-uc (approved):
Thanks. We unfortunately do get this sort of mistake from time to time, they're hard to eliminate :/

wantehchang:
Yes, this is a common bug in C libraries that define their own boolean types. (I guess the `true` and `false` macros defined in <stdbool.h> can also cause this bug.) Vincent just wrote a pull request so that this kind of bug can be detected.

PR #2712, opened 2025-03-28
Have AVIF_FALSE and AVIF_TRUE be enums

vrabaud:
This offers a solution to see errors like fixed in https://github.com/AOMediaCodec/libavif/pull/2711

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -85,8 +85,11 @@ extern "C" {
     ((AVIF_VERSION_MAJOR * 1000000) + (AVIF_VERSION_MINOR * 10000) + (AVIF_VERSION_PATCH * 100) + AVIF_VERSION_DEVEL)
 
 typedef int avifBool;
-#define AVIF_TRUE 1
-#define AVIF_FALSE 0
+enum avifTrueFalse
+{
+    AVIF_FALSE = 0,
+    AVIF_TRUE = 1
+};
 
 #define AVIF_DIAGNOSTICS_ERROR_BUFFER_SIZE 256
 
--- a/src/read.c
+++ b/src/read.c
@@ -3390,7 +3390,7 @@ static avifResult avifParseMetaBox(avifMeta * meta, uint64_t rawOffset, const ui
                 //   The handler type for the MetaBox shall be 'pict'.
                 if (memcmp(handlerType, "pict", 4) != 0) {
                     avifDiagnosticsPrintf(diag, "Box[hdlr] handler_type is not 'pict'");
-                    return AVIF_FALSE;
+                    return AVIF_RESULT_BMFF_PARSE_FAILED;
                 }
                 firstBox = AVIF_FALSE;
             } else {
@@ -3958,7 +3958,7 @@ static avifResult avifParseMovieBox(avifDecoderData * data,
                 !memcmp(track->handlerType, "auxv", 4)) {
                 if ((track->width == 0) || (track->height == 0)) {
                     avifDiagnosticsPrintf(data->diag, "Track ID [%u] has an invalid size [%ux%u]", track->id, track->width, track->height);
-                    return AVIF_FALSE;
+                    return AVIF_RESULT_BMFF_PARSE_FAILED;
                 }
                 if (avifDimensionsTooLarge(track->width, track->height, imageSizeLimit, imageDimensionLimit)) {
                     avifDiagnosticsPrintf(data->diag,
@@ -3966,7 +3966,7 @@ static avifResult avifParseMovieBox(avifDecoderData * data,
                                           track->id,
                                           track->width,
                                           track->height);
-                    return AVIF_FALSE;
+                    return AVIF_RESULT_BMFF_PARSE_FAILED;
                 }
             }
         }
@@ -3984,7 +3984,7 @@ static avifResult avifParseMovieBox(avifDecoderData * data,
 static avifProperty * avifMetaCreateProperty(avifMeta * meta, const char * propertyType)
 {
     avifProperty * metaProperty = avifArrayPush(&meta->properties);
-    AVIF_CHECK(metaProperty);
+    AVIF_CHECKERR(metaProperty, NULL);
     memcpy(metaProperty->type, propertyType, 4);
     return metaProperty;
 }
@@ -3992,7 +3992,7 @@ static avifProperty * avifMetaCreateProperty(avifMeta * meta, const char * prope
 static avifProperty * avifDecoderItemAddProperty(avifDecoderItem * item, const avifProperty * metaProperty)
 {
     avifProperty * itemProperty = avifArrayPush(&item->properties);
-    AVIF_CHECK(itemProperty);
+    AVIF_CHECKERR(itemProperty, NULL);
     *itemProperty = *metaProperty;
     return itemProperty;
 }


include/avif/avif.h
```diff
@@ -85,8 +85,11 @@ extern "C" {
     ((AVIF_VERSION_MAJOR * 1000000) + (AVIF_VERSION_MINOR * 10000) + (AVIF_VERSION_PATCH * 100) + AVIF_VERSION_DEVEL)
 
 typedef int avifBool;
-#define AVIF_TRUE 1
-#define AVIF_FALSE 0
+enum avifTrueFalse
```
y-guyon, line 88: Is it necessary to name the enum? Could it be the following instead?

```c
typedef enum avifBool {
    AVIF_FALSE = 0,
    AVIF_TRUE = 1
} avifBool;
```

vrabaud: avifBool is used a lot all over and we have casts from bools to avifBool in tests all over so that creates new failures. We can do 
```
typedef int avifBool;
typedef enum avifTrueFalse {
    AVIF_FALSE = 0,
    AVIF_TRUE = 1
} avifTrueFalse;
```
I could not find any warning when using an int instead of an enum (which makes sense) but I could find a warning when casting an enum to a different enum: https://clang.llvm.org/docs/DiagnosticsReference.html#wenum-conversion

include/avif/avif.h
```diff
@@ -85,8 +85,11 @@ extern "C" {
...
+{
+    AVIF_FALSE = 0,
```
wantehchang, line 90: MSVC emits a warning:

D:\a\libavif\libavif\src\read.c(5482): warning C4127: conditional expression is constant

src/read.c:5482 is:

```
            AVIF_ASSERT_OR_RETURN(AVIF_FALSE);
```

It is strange that MSVC does not warn if the conditional expression is the integer literal 0, which is also a constant.

We can try changing that line to `AVIF_ASSERT_OR_RETURN(0);` or adding a new macro `AVIF_ASSERT_NOT_REACHED_OR_RETURN` (or `AVIF_NOT_REACHED_OR_RETURN`).

include/avif/avif.h
wantehchang, line 88: Do we still get the benefit of the `-Wenum-conversion` warning if this enum is unnamed?`

```
enum
{
    AVIF_FALSE = 0,
    AVIF_TRUE = 1
};
```

vrabaud: actually yes. We can name it though

apps/avifdec.c
```diff
@@ -359,7 +359,7 @@ int main(int argc, char * argv[])
 
     const avifBool decodeAllFrames = frameIndex == DECODE_ALL_FRAMES;
     int currIndex = decodeAllFrames ? 0 : frameIndex;
-    while (AVIF_TRUE) {
+    for (;;) {
```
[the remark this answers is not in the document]
wantehchang: I see. This is a potential source incompatibility, i.e., code that compiles successfully with older versions of libavif will get a MSVC compiler warning when they upgrade to a new version of libavif.

We probably should still go ahead with this pull request, because code like `if (!AVIF_FALSE)` or `while (AVIF_TRUE)` is unlikely to show up outside libavif. Another option is to make this change a compile-time option for libavif developers that we enable only in a new CI workflow.

I wonder if this is a MSVC compiler bug. `while (1)` and `while (true)` are commonly used idioms that the compiler should not warn about them.

wantehchang (approved):

vrabaud:
I thought about having a CI specific guard, but that's the only thing that would be guarded for now. We could also just use stdbool.h no? (it would not solve this problem though but it would make the code more standard)

vrabaud:
Using stdbool actually finds other issues ... https://github.com/AOMediaCodec/libavif/pull/2734
But not the ones here.

include/avif/avif.h
```diff
@@ -85,8 +85,11 @@ extern "C" {
     ((AVIF_VERSION_MAJOR * 1000000) + (AVIF_VERSION_MINOR * 10000) + (AVIF_VERSION_PATCH * 100) + AVIF_VERSION_DEVEL)
 
 typedef int avifBool;
-#define AVIF_TRUE 1
-#define AVIF_FALSE 0
+typedef enum avifTrueFalse
+{
+    AVIF_FALSE = 0,
+    AVIF_TRUE = 1
+} avifTrueFalse;
```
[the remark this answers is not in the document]
vrabaud: Done in CI

apps/avifdec.c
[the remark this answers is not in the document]
vrabaud: Done in CI only

.github/workflows/ci-c-warnings.yml
```diff
@@ -53,6 +53,18 @@ jobs:
...
+        run: |
+          sed -i 's/#include <stdint.h>/#include <stdint.h>\n#include <stdbool.h>/' include/avif/avif.h
+          sed -i 's/typedef int avifBool;/typedef bool avifBool;/' include/avif/avif.h
+          sed -i 's/#define AVIF_TRUE 1/#define AVIF_TRUE true/' include/avif/avif.h
+          sed -i 's/#define AVIF_FALSE 0/#define AVIF_FALSE false/' include/avif/avif.h
+      - name: Use enums for AVIF_TRUE/AVIF_FALSE
+        if: ${{ matrix.case == 2}}
+        run: |
+          sed -i 's/#define AVIF_TRUE 1/enum avifTrueFalse_{ AVIF_TRUE = 1,/' include/avif/avif.h
+          sed -i 's/#define AVIF_FALSE 0/AVIF_FALSE = 0};/' include/avif/avif.h
```
wantehchang, line 67: I suggest we add these two alternative definitions of `avifBool` to avif.h and avoid patching avif.h like this. The alternative definitions would be guarded by macros that we define in this CI workflow.

If we want to patch avif.h, we may want to add comments around the definition of `avifBool` to prevent people form changing or even reformatting it.

.github/workflows/ci-c-warnings.yml
```diff
@@ -53,6 +53,18 @@ jobs:
...
         if: ${{ matrix.case == 0}}
         run: echo "CMAKE_AVIF_FLAGS=\"-DAVIF_ENABLE_NODISCARD=ON\""  >> $GITHUB_ENV
+      - name: Use stdbool for AVIF_TRUE/AVIF_FALSE
+        if: ${{ matrix.case == 1}}
+        run: |
+          sed -i 's/#include <stdint.h>/#include <stdint.h>\n#include <stdbool.h>/' include/avif/avif.h
+          sed -i 's/typedef int avifBool;/typedef bool avifBool;/' include/avif/avif.h
+          sed -i 's/#define AVIF_TRUE 1/#define AVIF_TRUE true/' include/avif/avif.h
+          sed -i 's/#define AVIF_FALSE 0/#define AVIF_FALSE false/' include/avif/avif.h
+      - name: Use enums for AVIF_TRUE/AVIF_FALSE
```
wantehchang, line 63: Do we also need this alternative definition? Does stdbool not catch all the bugs?

vrabaud: Nope, cf the other PR: both found different bugs .....

.github/workflows/ci-c-warnings.yml
vrabaud: I think adding two guards (for stdbool and enum) is a lot of noise for a public header no? Especially just for a CI usage (not even a test).

PR #2713, opened 2025-03-28
avifjpeg.c: check for uint32_t overflow before add

wantehchang:
Bug: b:406974988

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -310,7 +310,7 @@ static uint16_t avifJPEGReadUint16LittleEndian(const uint8_t * src)
 // Reads 'numBytes' at 'offset', stores them in 'bytes' and increases 'offset'.
 static avifBool avifJPEGReadBytes(const avifROData * data, uint8_t * bytes, uint32_t * offset, uint32_t numBytes)
 {
-    if (data->size < (*offset + numBytes)) {
+    if ((UINT32_MAX - *offset) < numBytes || data->size < (*offset + numBytes)) {
         return AVIF_FALSE;
     }
     memcpy(bytes, &data->data[*offset], numBytes);


apps/shared/avifjpeg.c
```diff
@@ -310,7 +310,7 @@ static uint16_t avifJPEGReadUint16LittleEndian(const uint8_t * src)
 // Reads 'numBytes' at 'offset', stores them in 'bytes' and increases 'offset'.
 static avifBool avifJPEGReadBytes(const avifROData * data, uint8_t * bytes, uint32_t * offset, uint32_t numBytes)
 {
-    if (data->size < (*offset + numBytes)) {
+    if ((UINT32_MAX - *offset) < numBytes || data->size < (*offset + numBytes)) {
```
wantehchang, line 313: In the reproducer testcase for the bug, `*offset` is 4294967295 (0xffffffff), `numBytes` is 2, and `data->size` is 82.

We are adding `*offset` and `numBytes` (both `uint32_t` variables) here and later in line 317. So we need to make sure the sum does not overflow `uint32_t`.

vigneshvg (approved):

apps/shared/avifjpeg.c
wantehchang: @maryla-uc Maryla: I did not look into how `*offset` became 4294967295 (0xffffffff). That seems suspicious, so there may be a better place to defend against this at a higher level.

maryla-uc: This looks reasonable to me, thanks for the fix!
The offset 0xffffffff is read from the file itself. The MPF (multi picture format) contains a bunch of offsets which let you seek into the file. So indeed, they can be invalid.
Right now we can still get overflows in places where we add numbers of these offsets. I think it's not a big deal because we'll go to the wrong part of the file and will probably conclude that it's invalid, but it would be cleaner to use a larger type to avoid overflowing.

wantehchang: Thanks for taking a look. I wrote a pull request to check for overflows when we add numbers to `offset`: https://github.com/AOMediaCodec/libavif/pull/2736

Using a larger type to avoid overflowing also sounds good. I didn't look into that.

Also, it would make it easier to see the code is secure if we maintain the invariant that `offset` be <= the size of the data buffer (`data->size` in this function). Right now we allow `offset` to exceed the size of the data buffer, but we take care to check the `offset` before we read from the data buffer (which only occurs in this function), so the code is secure.

PR #2714, opened 2025-03-28
Enable quantization matrices for avm

FrankGalligan:


src/codec_avm.c
```diff
@@ -406,6 +406,12 @@ static const struct aomOptionDef aomOptionDefs[] = {
     { "film-grain-test", AV1E_SET_FILM_GRAIN_TEST_VECTOR, AVIF_AOM_OPTION_INT, NULL },
     // Film grain table file
     { "film-grain-table", AV1E_SET_FILM_GRAIN_TABLE, AVIF_AOM_OPTION_STR, NULL },
+    // Enable quantization matrices
+    { "enable-qm", AV1E_SET_ENABLE_QM, AVIF_AOM_OPTION_UINT, NULL },
+    // Quantization matrices range minimum
+    { "qm-min", AV1E_SET_QM_MIN, AVIF_AOM_OPTION_UINT, NULL },
+    // Quantization matrices range maximum
+    { "qm-max", AV1E_SET_QM_MAX, AVIF_AOM_OPTION_UINT, NULL },
```
wantehchang, line 414: If the `aom_codec_set_option()` function is present, we don't need to add these options to this array. Does AVM not have the `aom_codec_set_option()` function?

wantehchang: I verified that AVM has the `aom_codec_set_option()` function, and it supports the "enable-qm", "qm-min", and "qm-max" options. So we should not need this pull request.

I also verified that src/codec_avm.c (this file) calls the `aom_codec_set_option()` function.

Frank, could you double check if you really need this pull request?

FrankGalligan: If I don't include those it wail fail on line 489

wantehchang: Thanks. I found the bug in the `avifProcessAOMOptionsPostInit()` function. Please test my fix: https://github.com/AOMediaCodec/libavif/pull/2716

PR #2715, opened 2025-03-29
fix spelling

FrankGalligan:


wantehchang (approved):

apps/avifgainmaputil/swapbase_command.cc
```diff
@@ -116,7 +116,7 @@ SwapBaseCommand::SwapBaseCommand()
           "swapbase",
           "Swaps the base and alternate images (e.g. if the base image is SDR "
           "and the alternate is HDR, makes the base HDR). The alternate image "
-          "is the result ot fully applying the gain map.") {
```
jzern, line 119: Maybe this should have been 'of':
*The alternate image is the result to fully applying the gain map.*
vs.
*The alternate image is the result of fully applying the gain map.*

PR #2716, opened 2025-03-29
codec_avm.c: Removed unneeded code for options

wantehchang:
All versions of AVM that we need to support have the aom_codec_set_option() function. Remove the code that assumes aom_codec_set_option() is not present.

This also fixes a bug in avifProcessAOMOptionsPostInit() that incorrectly concatenates the two conditionally-compiled pieces of code when the file was copied from codec_aom.c and the
HAVE_AOM_CODEC_SET_OPTION macro was removed.

### DIFF
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -271,32 +271,6 @@ static aom_img_fmt_t avifImageCalcAOMFmt(const avifImage * image, avifBool alpha
     return fmt;
 }
 
-static avifBool aomOptionParseInt(const char * str, int * val)
-{
-    char * endptr;
-    const long rawval = strtol(str, &endptr, 10);
-
-    if (str[0] != '\0' && endptr[0] == '\0' && rawval >= INT_MIN && rawval <= INT_MAX) {
-        *val = (int)rawval;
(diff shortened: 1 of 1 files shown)


src/codec_avm.c
```diff
@@ -451,46 +381,6 @@ static avifBool avifProcessAOMOptionsPostInit(avifCodec * codec, avifBool alpha)
...
-                }
-                if (aomOptionDefs[j].controlId == AOME_SET_TUNING) {
-                    codec->internal->tuningSet = AVIF_TRUE;
-                }
-                break;
-            }
-        }
-        if (!match) {
-            return AVIF_FALSE;
-        }
```
wantehchang, line 493: Removing this block of code is a bug fix.

vrabaud (approved):

y-guyon (approved):

src/codec_avm.c
FrankGalligan: I just tried this and it works. So we can land this and I can delete #2714

wantehchang: Frank: Thanks for verifying this works. I have landed this.

PR #2721, opened 2025-03-31
Bump SVT-AV1 dependency to v3.1.0

vrabaud:


y-guyon (approved):

vrabaud:
Let's wait for a new release to fix https://gitlab.com/AOMediaCodec/SVT-AV1/-/merge_requests/2417

wantehchang:
Vincent: We can upgrade to the newly-released SVT-AV1 v3.1.0 now.

vrabaud:
For SVT-AV1, with https://gitlab.com/AOMediaCodec/SVT-AV1/-/commit/31fc5ca61f3a0459eade6c675ee0116800c9ab05, min_qp_allowed is not necessarily <= max_qp_allowed. We used to have:
```
/* Minimum QP value allowed for rate control use, only applicable when rate
     * control mode is set to 1 or 2. It has to be smaller or equal to maxQpAllowed.
     *
     * Default is 4. */
```
With https://gitlab.com/AOMediaCodec/SVT-AV1/-/commit/a95810b4eb6c1f909ea3b0050972dc2260bcab47, it has to be in [1 - 62]. This was actually specified in the doc since 2022 (cf https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/Docs/Parameters.md), just not enforced in the code.

y-guyon (approved):

src/codec_svt.c
```diff
@@ -173,10 +173,18 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
...
+#else
             svt_config->min_qp_allowed = AVIF_CLAMP(encoder->minQuantizerAlpha, 0, 63);
+#endif
             svt_config->max_qp_allowed = AVIF_CLAMP(encoder->maxQuantizerAlpha, 0, 63);
         } else {
+#if SVT_AV1_CHECK_VERSION(3, 1, 0)
+            svt_config->min_qp_allowed = AVIF_CLAMP(encoder->minQuantizer, 0, 62);
+#else
             svt_config->min_qp_allowed = AVIF_CLAMP(encoder->minQuantizer, 0, 63);
+#endif
```
wantehchang: Vincent: Thank you for noticing this newly-enforced range for min_qp_allowed. I suggest we use the [1-62] range for all versions of SVT-AV1, not just v3.1.0 or later.

vrabaud: Done

vrabaud:
We were wondering whether the range should be bumped to 1-62, but 0 is indeed allowed for lossless

wantehchang:
Vincent: Thank you for investitating this!

In Source/Lib/Globals/enc_settings.c of SVT-AV1 v3.1.0, I see the following code:

```
    if (config->max_qp_allowed > MAX_QP_VALUE) {
        SVT_ERROR("Instance %u: MaxQpAllowed must be [1 - %d]\n", channel_number + 1, MAX_QP_VALUE);
        return_error = EB_ErrorBadParameter;
    } else if (config->min_qp_allowed >= MAX_QP_VALUE) {
        SVT_ERROR("Instance %u: MinQpAllowed must be [1 - %d]\n", channel_number + 1, MAX_QP_VALUE - 1);
        return_error = EB_ErrorBadParameter;
    } else if ((config->min_qp_allowed) > (config->max_qp_allowed)) {
        SVT_ERROR("Instance %u:  MinQpAllowed must be smaller than MaxQpAllowed\n", channel_number + 1);
        return_error = EB_ErrorBadParameter;
    }
```

The last check shows `min_qp_allowed` can be equal to `max_qp_allowed`.

PR #2729, opened 2025-04-01
Fixes wrong EXIF Orientation from Irot Imir

### LINKED ISSUE #2727: Wrong EXIF Orientation converting from AVIF to JPG
Converting a JPG image to AVIF using avifenc and then converting the AVIF back to JPG using avifdec will write the wrong Orientation in the EXIF data.

# Version
```sh
avifenc -V
Version: 1.0.4 (dav1d [dec]:1.4.1, libgav1 [dec]:0.18.0, aom [enc/dec]:v3.8.2, rav1e [enc]:0.7.1 (UNKNOWN), svt [enc]:v1.7.0)
libyuv : available (1883)
```
# JPG to AVIF
```sh
avifenc -q 70 input.jpg output.avif
Directly copied JPEG pixel data (no YUV conversion): input.jpg
Successfully loaded: input.jpg
AVIF to be written: (Lossy)
 * Resolution     : 4000x2252
 * Bit Depth      : 8
 * Format         : YUV420
 * Chroma Sam. Pos: 0
 * Alpha          : Absent
 * Range          : Full
 * Color Primaries: 2
 * Transfer Char. : 2
 * Matrix Coeffs. : 6
 * ICC Profile    : Present (632 bytes)
 * XMP Metadata   : Absent
 * Exif Metadata  : Present (33975 bytes)
 * Transformations:
    * irot (Rotation)      : 3
 * Progressive    : Unavailable
Encoding with AV1 codec 'aom' speed [6], color quality [70 (Medium)], alpha quality [100 (Lossless)], tileRowsLog2 [0], tileColsLog2 [0], 1 worker thread(s), please wait...
Encoded successfully.
 * Color AV1 total size: 1367241 bytes
 * Alpha AV1 total size: 0 bytes
Wrote AVIF: output.avif
```
`* irot (Rotation)      : 3`  is the angle and means (90 * ANGLE) degree rotation anti-clockwise, which is correct, 90*3=270° anti-clockwise which equals the original JPG Orientation of 90° clockwise (CW) :
```sh
$ exiftool -a -u -g1 input.jpg | grep Orientation
Orientation                     : Rotate 90 CW
$ exiftool -Orientation# input.jpg 
Orientation                     : 6
```
# AVIF to JPG
```sh
avifdec output.avif upsidedown.jpg
Image decoded: output.avif
Image details:
 * Resolution     : 4000x2252
 * Bit Depth      : 8
 * Format         : YUV420
 * Chroma Sam. Pos: 0
 * Alpha          : Absent
 * Range          : Full
 * Color Primaries: 2
 * Transfer Char. : 2
 * Matrix Coeffs. : 6
 * ICC Profile    : Present (632 bytes)
 * XMP Metadata   : Absent
 * Exif Metadata  : Present (33975 bytes)
 * Transformations:
    * irot (Rotation)      : 3
 * Progressive    : Unavailable
Wrote JPEG: upsidedown.jpg
```
# Issue
6 and 8 EXIF Orientations are swapped in [avifImageGetExifOrientationFromIrotImir()](https://github.com/AOMediaCodec/libavif/blob/7d2c765bb0d142a571ae532c363900e390580185/apps/shared/avifexif.c#L7)
```sh
$ exiftool -a -u -g1 upsidedown.jpg | grep Orientation
Orientation                     : Rotate 270 CW
$ exiftool -Orientation# upsidedown.jpg 
Orientation                     : 8
```
Resulting in an upsidedown JPG

A temporary "solution" is manually changing Rotation before using avifdec:
```sh
$ exiftool -QuickTime:Rotation=90 output.avif
```
This will create an avif file with a wrong rotation but will make avifdec wrongly output a correct orientation of 6 (Rotate 90 CW) JPG

joojoooo:
fixes #2727

### DIFF
--- a/apps/shared/avifexif.c
+++ b/apps/shared/avifexif.c
@@ -13,7 +13,7 @@ uint8_t avifImageGetExifOrientationFromIrotImir(const avifImage * image)
             }
             return 5; // 90 degrees anti-clockwise then swap top and bottom.
         }
-        return 6; // 90 degrees anti-clockwise.
+        return 8; // 90 degrees anti-clockwise.
     }
     if ((image->transformFlags & AVIF_TRANSFORM_IROT) && (image->irot.angle == 2)) {
         if (image->transformFlags & AVIF_TRANSFORM_IMIR) {
@@ -31,7 +31,7 @@ uint8_t avifImageGetExifOrientationFromIrotImir(const avifImage * image)
             }
             return 7; // 270 degrees anti-clockwise then swap top and bottom.
         }
-        return 8; // 270 degrees anti-clockwise.
+        return 6; // 270 degrees anti-clockwise.
     }
     if (image->transformFlags & AVIF_TRANSFORM_IMIR) {
         if (image->imir.axis) {
--- a/tests/gtest/avifmetadatatest.cc
+++ b/tests/gtest/avifmetadatatest.cc
@@ -9,6 +9,8 @@
 
 #include "avif/avif.h"
 #include "avif/avif_cxx.h"
+#include "avif/internal.h"
+#include "avifexif.h"
 #include "avifjpeg.h"
 #include "avifpng.h"
 #include "aviftest_helpers.h"
@@ -282,6 +284,22 @@ TEST(MetadataTest, ExifOrientation) {
       avifTransformFlags{0});
   // TODO(yguyon): Fix orientation not being applied to PNG samples.
   EXPECT_EQ(image->width, temp_image->width /* should be height here */);
+}
+
+TEST(MetadataTest, AllExifOrientations) {
+  const ImagePtr image =
+      testutil::ReadImage(data_path, "paris_exif_orientation_5.jpg");
+  ASSERT_NE(image, nullptr);
+  image->transformFlags = AVIF_TRANSFORM_NONE;
+  for (uint8_t orientation = 1; orientation <= 8; ++orientation) {
+    // Check roundtrip.
+    ASSERT_EQ(avifSetExifOrientation(&image->exif, orientation),
+              AVIF_RESULT_OK);
+    ASSERT_EQ(avifImageExtractExifOrientationToIrotImir(image.get()),
+              AVIF_RESULT_OK);
+    ASSERT_EQ(avifImageGetExifOrientationFromIrotImir(image.get()),
+              orientation);
+  }
 }
 
 TEST(MetadataTest, ExifOrientationAndForcedImir) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ The changes are relative to the previous release, unless the baseline is specifi
   AVIF_PIXEL_FORMAT_YUV400 to be AV1 spec compatible.
 * Ignore tmap items not present in `grpl` box
 * Update libyuv.cmd: dc47c71b3 (1907)
+* Fix wrong Exif orientation set in JPEG or PNG output by avifdec when the input
+  AVIF file has an ImageRotation property with angle set to 1 or 3, has no
+  ImageMirror property, and carries an Exif chunk. Note that Exif orientation is
+  usually ignored in PNG files, so this mainly impacts JPEG files.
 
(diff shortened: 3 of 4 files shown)


y-guyon (approved):

vrabaud (approved):

apps/shared/avifexif.c
```diff
@@ -31,7 +31,7 @@ uint8_t avifImageGetExifOrientationFromIrotImir(const avifImage * image)
             }
             return 7; // 270 degrees anti-clockwise then swap top and bottom.
         }
-        return 8; // 270 degrees anti-clockwise.
+        return 6; // 270 degrees anti-clockwise.
```
wantehchang, line 34: I verified the bug fix by comparing this function with the corresponding code in Chrome. You can skip the following details.

Details:

In Chrome (third_party/blink/renderer/platform/image-decoders/avif/crabbyavif_image_decoder.cc), this operation is implemented as a table lookup (I replaced the symbolic constants with numeric values in the lookup table):

```
  // |angle| * 90 specifies the angle of anti-clockwise rotation in degrees.
  // Legal values: [0-3].
  int angle = 0;
  if (container->transformFlags & crabbyavif::AVIF_TRANSFORM_IROT) {
    angle = container->irot.angle;
    CHECK_LT(angle, 4);
  }
  // |axis| specifies how the mirroring is performed.
  //   -1: No mirroring.
  //    0: The top and bottom parts of the image are exchanged.
  //    1: The left and right parts of the image are exchanged.
  int axis = -1;
  if (container->transformFlags & crabbyavif::AVIF_TRANSFORM_IMIR) {
    axis = container->imir.axis;
    CHECK_LT(axis, 2);
  }
  // MIAF Section 7.3.6.7 (Clean aperture, rotation and mirror) says:
  //   These properties, if used, shall be indicated to be applied in the
  //   following order: clean aperture first, then rotation, then mirror.
  //
  // In the kAxisAngleToOrientation array, the first dimension is axis (with an
  // offset of 1). The second dimension is angle.
  constexpr std::array<std::array<ImageOrientationEnum, 4>, 3>
      kAxisAngleToOrientation = {{
          // No mirroring.
          {1, 8, 3, 6},
          // Top-to-bottom mirroring. Change Top<->Bottom in the first row.
          {4, 5, 2, 7},
          // Left-to-right mirroring. Change Left<->Right in the first row.
          {2, 7, 4, 5},
      }};
  orientation_ = kAxisAngleToOrientation[axis + 1][angle];
```

To match the code here, we need to read each column in Chrome's lookup table from the bottom up. For example, the second column read from the bottom up is 7, 5, 8. This should match the first outermost if block in this function. This confirms the last entry should be 8, not 6.

Similarly, the third column read from the bottom up is 4, 2, 3. This matches the second outermost if block in this function.

The fourth column read from the bottom up is 5, 7, 6. This matches the new code in the third outermost if block in this function.

Finally, the first column read from the bottom up is 2, 4, 1. This matches the last outermost if block and the final return statement in this function.

tests/gtest/avifmetadatatest.cc
```diff
@@ -284,6 +286,22 @@ TEST(MetadataTest, ExifOrientation) {
   EXPECT_EQ(image->width, temp_image->width /* should be height here */);
 }
 
+TEST(MetadataTest, AllExifOrientations) {
+  const ImagePtr image =
+      testutil::ReadImage(data_path, "paris_exif_orientation_5.jpg");
+  ASSERT_NE(image, nullptr);
+  image->transformFlags = AVIF_TRANSFORM_NONE;
```
wantehchang, line 293: Optional: It seems better to move these four lines (290-293) into the for loop so that each iteration of the for loop starts from a clean image. Right now each iteration starts from the results of the previous iteration.

wantehchang (commented):
LGTM. Thanks for the fix!

tests/gtest/avifmetadatatest.cc
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2735

PR #2726, opened 2025-04-01
zlibpng.cmd: zlib 5a82f71 and libpng 1.6.47

wantehchang:
ZLIB_BUILD_EXAMPLES has apparently been renamed ZLIB_BUILD_TESTING.

PNG_BUILD_ZLIB is deprecated. Use ZLIB_ROOT instead.

Revert the -DCMAKE_POLICY_VERSION_MINIMUM=3.5 workaround in .github/workflows/ci-windows.yml because the CMakeLists.txt in libpng v1.6.47 passes version 3.14 to cmake_minimum_required().

PR #2733, opened 2025-04-01
Update libpng to 1.6.47

vrabaud:
Using a newer zlib is not possible as the CMake is very different and is not compatible with FindZLib.cmake and libpng. It is easier to patch the old zlib


cmake/Modules/LocalZlibpng.cmake
```diff
@@ -18,44 +16,29 @@ if(ANDROID_ABI)
     set(ZLIB_BINARY_DIR "${ZLIB_BINARY_DIR}/${ANDROID_ABI}")
 endif()
 
+# The patch is taken from Googles's or-tools
```
wantehchang, line 19: Nit: Googles's => Google's

cmake/Modules/LocalZlibpng.cmake
```diff
@@ -18,44 +16,29 @@ if(ANDROID_ABI)
...
 
 # Put the value of ZLIB_INCLUDE_DIR in the cache. This works around cmake behavior that has been updated by
 # cmake policy CMP0102 in cmake 3.17. Remove the CACHE workaround when we require cmake 3.17 or later. See
 # https://gitlab.kitware.com/cmake/cmake/-/issues/21343.
 set(ZLIB_INCLUDE_DIR "${ZLIB_SOURCE_DIR}" CACHE PATH "zlib include dir")
-set(ZLIB_BUILD_EXAMPLES OFF CACHE BOOL "")
-# This include_directories() call must be before add_subdirectory(ext/zlib) to work around the
-# zlib/CMakeLists.txt bug fixed by https://github.com/madler/zlib/pull/818.
-include_directories(SYSTEM $<BUILD_INTERFACE:${ZLIB_INCLUDE_DIR}>)
+set(ZLIB_BUILD_TESTING OFF CACHE BOOL "")
```
wantehchang, line 34: IMPORTANT: This change of `ZLIB_BUILD_EXAMPLES` to `ZLIB_BUILD_TESTING` comes from my pull request (https://github.com/AOMediaCodec/libavif/pull/2726). It seems incorrect in this pull request, because cmake/Modules/ZLIB-v1.3.1.patch does NOT replace `ZLIB_BUILD_EXAMPLES` with `ZLIB_BUILD_TESTING`.

cmake/Modules/ZLIB-v1.3.1.patch
```diff
@@ -0,0 +1,150 @@
...
+ endif(MINGW)
+ 
+-add_library(zlib SHARED ${ZLIB_SRCS} ${ZLIB_DLL_SRCS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS})
+-target_include_directories(zlib PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
+-add_library(zlibstatic STATIC ${ZLIB_SRCS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS})
+-target_include_directories(zlibstatic PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
+-set_target_properties(zlib PROPERTIES DEFINE_SYMBOL ZLIB_DLL)
+-set_target_properties(zlib PROPERTIES SOVERSION 1)
++if(BUILD_SHARED_LIBS)
++  add_library(ZLIB SHARED ${ZLIB_SRCS} ${ZLIB_DLL_SRCS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS})
```
wantehchang, line 49: Do you know why this patch changes the name of the zlib library to all uppercase? It seems like an unnecessary change. The patch will be smaller without this library name change.

wantehchang (approved):
LGTM. Thanks!

cmake/Modules/ZLIB-v1.3.1.patch
```diff
@@ -0,0 +1,150 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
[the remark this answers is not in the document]
vrabaud: Let's move it to ext for now as it is cleaner to apply the patch in the zlibpng.cmd script. The ext folder should be removed at some point though as it is redundant with the CMake code and offers less flexibility (it just offers clarity as to where the sources are but we could add a comment in the README for developers)

ext/zlibpng.cmd
```diff
@@ -5,4 +5,4 @@
 : # The odd choice of comment style in this file is to try to share this script between *nix and win32.
 
 git clone -b v1.3.1 --depth 1 https://github.com/madler/zlib.git
-git clone -b v1.6.40 --depth 1 https://github.com/glennrp/libpng.git
+git clone -b v1.6.47 --depth 1 https://github.com/glennrp/libpng.git
```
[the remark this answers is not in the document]
vrabaud: Done

cmake/Modules/LocalZlibpng.cmake
vrabaud: Done

cmake/Modules/ZLIB-v1.3.1.patch
vrabaud: I believe this is to distinguish betwene the target (ZLIB), and the library name (zlib or zlibstatic).

ext/zlibpng.cmd
```diff
@@ -5,4 +5,5 @@
 : # The odd choice of comment style in this file is to try to share this script between *nix and win32.
 
 git clone -b v1.3.1 --depth 1 https://github.com/madler/zlib.git
-git clone -b v1.6.40 --depth 1 https://github.com/glennrp/libpng.git
+git apply --ignore-whitespace
```
wantehchang, line 8: This is missing the patch file name, so the command hangs (presumably waiting for the patch file from stdin).

Fixed in https://github.com/AOMediaCodec/libavif/pull/2743.

cmake/Modules/ZLIB-v1.3.1.patch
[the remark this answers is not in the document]
wantehchang: I found that libargparse.patch is also in the ext directory. So it's good to move the zlib patch there.

PR #2734, opened 2025-04-01
POC: use stdbool

vrabaud:
This breaks the ABI

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -4,6 +4,7 @@
 #ifndef AVIF_AVIF_H
 #define AVIF_AVIF_H
 
+#include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
 
@@ -84,9 +85,9 @@ extern "C" {
 #define AVIF_VERSION \
     ((AVIF_VERSION_MAJOR * 1000000) + (AVIF_VERSION_MINOR * 10000) + (AVIF_VERSION_PATCH * 100) + AVIF_VERSION_DEVEL)
 
-typedef int avifBool;
-#define AVIF_TRUE 1
-#define AVIF_FALSE 0
+typedef bool avifBool;
+#define AVIF_TRUE true
+#define AVIF_FALSE false
 
 #define AVIF_DIAGNOSTICS_ERROR_BUFFER_SIZE 256
 


src/read.c
```diff
@@ -6699,7 +6699,7 @@ static avifResult avifDecoderApplySampleTransform(const avifDecoder * decoder, a
         return result;
     }
 
-    for (avifBool alpha = AVIF_FALSE; alpha <= decoder->alphaPresent ? AVIF_TRUE : AVIF_FALSE; ++alpha) {
+    for (avifBool alpha = AVIF_FALSE; !alpha || decoder->alphaPresent; alpha = AVIF_TRUE) {
```
wantehchang, line 6702: The proposed new code is hard to understand. I suggest either:

```
    // `alpha` is used as a boolean inside the for loop.
    for (int alpha = 0; alpha <= decoder->alphaPresent; ++alpha) {
```

or

```
    const avifBool alpha[2] = { AVIF_FALSE, AVIF_TRUE };
    for (int i = 0; i <= decoder->alphaPresent; ++i) {
        // Replace `alpha` with `alpha[i]` in the loop's body
    }
```

Note: The conditional expression in the original code, `alpha <= decoder->alphaPresent ? AVIF_TRUE : AVIF_FALSE`, can be simplified to ` alpha <= decoder->alphaPresent`. This fact is used above.

apps/avifgainmaputil/extractgainmap_command.cc
```diff
@@ -18,7 +18,7 @@ ExtractGainMapCommand::ExtractGainMapCommand()
 
 avifResult ExtractGainMapCommand::Run() {
   DecoderPtr decoder(avifDecoderCreate());
-  if (decoder == NULL) {
+  if (decoder == nullptr) {
```
wantehchang, line 21: Please move the NULL-to-nullptr changes to a separate pull request.

src/write.c
```diff
@@ -975,10 +975,10 @@ static avifBool avifGainMapIdenticalChannels(const avifGainMap * gainMap)
 }
 
 // Returns the number of bytes written by avifWriteGainmapMetadata().
-static avifBool avifGainMapMetadataSize(const avifGainMap * gainMap)
+static uint32_t avifGainMapMetadataSize(const avifGainMap * gainMap)
 {
     const uint8_t channelCount = avifGainMapIdenticalChannels(gainMap) ? 1u : 3u;
-    return sizeof(uint16_t) * 2 + sizeof(uint8_t) + sizeof(uint32_t) * 4 + channelCount * sizeof(uint32_t) * 10;
+    return (uint32_t)(sizeof(uint16_t) * 2 + sizeof(uint8_t) + sizeof(uint32_t) * 4 + channelCount * sizeof(uint32_t) * 10);
```
wantehchang, line 981: Optional: Can we omit the `uint32_t` cast?

src/write.c
```diff
@@ -975,10 +975,10 @@ static avifBool avifGainMapIdenticalChannels(const avifGainMap * gainMap)
 }
 
 // Returns the number of bytes written by avifWriteGainmapMetadata().
-static avifBool avifGainMapMetadataSize(const avifGainMap * gainMap)
+static uint32_t avifGainMapMetadataSize(const avifGainMap * gainMap)
```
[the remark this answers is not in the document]
vrabaud: Done in https://github.com/AOMediaCodec/libavif/pull/2737

src/read.c
vrabaud: Both solutions still compare an int to a bool (alphaPresent). The solution in https://github.com/AOMediaCodec/libavif/pull/2737 is the cleanest IMHO

src/write.c
vrabaud: sizeof returns a size_t. In case somebody has a warning that warns about narrowing.

apps/avifgainmaputil/extractgainmap_command.cc
vrabaud: Done in https://github.com/AOMediaCodec/libavif/pull/2742

include/avif/avif.h
```diff
@@ -84,9 +85,9 @@ extern "C" {
 #define AVIF_VERSION \
     ((AVIF_VERSION_MAJOR * 1000000) + (AVIF_VERSION_MINOR * 10000) + (AVIF_VERSION_PATCH * 100) + AVIF_VERSION_DEVEL)
 
-typedef int avifBool;
-#define AVIF_TRUE 1
-#define AVIF_FALSE 0
+typedef bool avifBool;
```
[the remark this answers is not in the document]
vrabaud: Indeed, there is an ABI breakage, that was more a POC. Do we wait for 1.3.0 or we will never allow that?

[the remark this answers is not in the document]
wantehchang: v1.3.0 is a minor release, which allows new features but does not allow backward incompatible changes. v2.0.0 is a major release, which allows any changes.

We can either abandon this pull request and make this a conditionally-compiled change that we enable in a CI workflow.

vrabaud:
This PR is now part of the CI in https://github.com/AOMediaCodec/libavif/pull/2712

PR #2735, opened 2025-04-01
Move test body to inner loop in MetadataTest

y-guyon:


tests/gtest/avifmetadatatest.cc
```diff
@@ -287,11 +287,11 @@ TEST(MetadataTest, ExifOrientation) {
 }
 
 TEST(MetadataTest, AllExifOrientations) {
-  const ImagePtr image =
-      testutil::ReadImage(data_path, "paris_exif_orientation_5.jpg");
-  ASSERT_NE(image, nullptr);
-  image->transformFlags = AVIF_TRANSFORM_NONE;
   for (uint8_t orientation = 1; orientation <= 8; ++orientation) {
+    const ImagePtr image =
+        testutil::ReadImage(data_path, "paris_exif_orientation_5.jpg");
```
wantehchang, line 292: Optional: We could also decode the image before the for loop and call `avifImageCopy()` here. But this is fine.

wantehchang (approved):
LGTM. Thanks!

PR #2736, opened 2025-04-01
Detect overflows when adding numbers to `offset`

wantehchang:
Detect overflows when adding numbers to `offset` in avifJPEGExtractGainMapImageFromMpf().

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -715,6 +715,9 @@ static avifBool avifJPEGExtractGainMapImageFromMpf(FILE * f,
     for (int mpTagIdx = 0; mpTagIdx < mpTagCount; ++mpTagIdx) {
         uint16_t tagId;
         AVIF_CHECK(avifJPEGReadU16(segmentData, &tagId, &offset, isBigEndian));
+        if (UINT32_MAX - offset < 2 + 4) {
+            return AVIF_FALSE;
+        }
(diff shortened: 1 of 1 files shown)


apps/shared/avifjpeg.c
```diff
@@ -749,12 +752,18 @@ static avifBool avifJPEGExtractGainMapImageFromMpf(FILE * f,
...
+        }
         offset += 4; // Skip "Individual Image Attribute"
         uint32_t imageSize;
         AVIF_CHECK(avifJPEGReadU32(segmentData, &imageSize, &offset, isBigEndian));
         uint32_t imageDataOffset;
         AVIF_CHECK(avifJPEGReadU32(segmentData, &imageDataOffset, &offset, isBigEndian));
 
+        if (UINT32_MAX - offset < 4) {
+            return AVIF_FALSE;
+        }
```
wantehchang, line 766: Maryla: I found these by searching for "offset +" in this file.

We also add numbers to `offset` in `avifJPEGFindMpfSegmentOffset()`. I did not add similar checks to that function because I think `offset` cannot possibly overflow in that function. It would be good if you could also audit `avifJPEGFindMpfSegmentOffset()` to verify my claim.

maryla-uc: Yes I agree with you. Thanks for the fix!

maryla-uc (approved):

PR #2737, opened 2025-04-02
Fix some boolean logic

vrabaud:
This is just for clarity and in case avifBool is a proper boolean type.
Some of those fixes are from https://github.com/AOMediaCodec/libavif/pull/2734

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -6699,17 +6699,8 @@ static avifResult avifDecoderApplySampleTransform(const avifDecoder * decoder, a
         return result;
     }
 
-    for (int i_alpha = 0; i_alpha < 2; ++i_alpha) {
-        avifBool alpha;
-        if (i_alpha == 0) {
-            alpha = AVIF_FALSE;
-        } else {
-            if (decoder->alphaPresent) {
-                alpha = AVIF_TRUE;
-            } else {
-                break;
-            }
-        }
+    for (int pass = 0; pass < (decoder->alphaPresent ? 2 : 1); ++pass) {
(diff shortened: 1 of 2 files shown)


src/read.c
```diff
@@ -6699,7 +6699,17 @@ static avifResult avifDecoderApplySampleTransform(const avifDecoder * decoder, a
         return result;
     }
 
-    for (avifBool alpha = AVIF_FALSE; alpha <= decoder->alphaPresent ? AVIF_TRUE : AVIF_FALSE; ++alpha) {
+    for (int i_alpha = 0; i_alpha < 2; ++i_alpha) {
```
maryla-uc, line 6702: Could also do something like this:
```c
const int numPasses = decoder->alphaPresent ? 2 : 1;
for (int pass = 0; pass < numPasses; ++pass) {
  avifBool alpha = (pass == 1);
  ...
}
```
Not sure which one's clearer. Feel free to keep yours.

src/write.c
```diff
@@ -975,10 +975,10 @@ static avifBool avifGainMapIdenticalChannels(const avifGainMap * gainMap)
 }
 
 // Returns the number of bytes written by avifWriteGainmapMetadata().
-static avifBool avifGainMapMetadataSize(const avifGainMap * gainMap)
+static uint32_t avifGainMapMetadataSize(const avifGainMap * gainMap)
```
maryla-uc, line 978: my bad! thanks for the fix :)

maryla-uc (approved):

src/read.c
vrabaud: Done

PR #2739, opened 2025-04-02
Add "set -eu" at the top of every script.

maryla-uc:
Although it's set in cmd_test_common.sh, if cmd_test_common.sh fails to load for any reason, then the script will not fail.


y-guyon (approved):
Alternatively, could `source $(dirname "$0")/cmd_test_common.sh || die` be enough instead?

PR #2744, opened 2025-04-02
Set CONFIG_AV1_DECODER or CONFIG_AV1_ENCODER to 0

wantehchang:
If AVIF_CODEC_AOM_DECODE is OFF, set CONFIG_AV1_DECODER to 0. Similarly for CONFIG_AV1_ENCODER.


cmake/Modules/LocalAom.cmake
```diff
@@ -109,6 +109,14 @@ else()
...
     endif()
 
+    if(NOT AVIF_CODEC_AVM)
+        if(NOT AVIF_CODEC_AOM_DECODE)
+            set(CONFIG_AV1_DECODER 0 CACHE INTERNAL "")
+        endif()
+        if(NOT AVIF_CODEC_AOM_ENCODE)
+            set(CONFIG_AV1_ENCODER 0 CACHE INTERNAL "")
+        endif()
+    endif()
```
wantehchang, line 119: @fdintino Frankie: FYI. In https://github.com/python-pillow/Pillow/pull/8858/files/84485e1718869fecf1e6156e84f0e1e89a73ee96..ab55e543c8ea8bdbd3621f542da761610b38e403#r2025229855, you suggested using `set(CONFIG_AV1_DECODER 0)`. I assume that was a shorthand for `set(CONFIG_AV1_DECODER 0 CACHE INTERNAL "")` (line 114), right?

fdintino: you assume correctly.

fdintino (approved):

vrabaud (approved):

PR #2748, opened 2025-04-03
Change avifenc to start in automatic tiling mode

wantehchang:
Bug: b:328995727

### DIFF
--- a/doc/avifenc.1.md
+++ b/doc/avifenc.1.md
@@ -1,6 +1,6 @@
 % AVIFENC(1) | General Commands Manual
 %
-% 2022-04-30
+% 2025-04-11
 
 <!--
 This man page is written in pandoc's Markdown.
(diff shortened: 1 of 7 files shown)


maryla-uc (approved):
Please update the golden files by following the intrusctions in the the "How to fix failing tests" section of the failing test.
Basically run these two commands:
```
gh run download --dir /tmp/golden-tests-output-14401574812 --repo AOMediaCodec/libavif --name golden-tests-output 14401574812
cp /tmp/golden-tests-output-14401574812/*/*.xml ~/git/libavif/tests/data/goldens
```
(of course you might have to change the git path to the appropriate location on your machine in the second command)

It looks like the golden files contain some "start" field that is likely to change and should probably be redacted, but that should be fixed separately.

wantehchang (commented):
Thanks for the review and the tips on updating the golden files. I also realized that I forgot to update the avifenc man page. I took care of both in the second commit.

doc/avifenc.1.md
```diff
@@ -189,7 +189,7 @@ The following options can optionally have a **:u** (or **:update**) suffix like
 **\--autotiling**
 :   Set \--tilerowslog2 and \--tilecolslog2 automatically.
     If specified, switch to automatic tiling.
-    avifenc starts in manual tiling mode.
+    avifenc starts in automatic tiling mode.
```
jzern, line 192: The date of the man page could be updated too.

wantehchang: Done.

PR #2749, opened 2025-04-04
Document how the encoder's tiling mode changes

wantehchang:
Document how the --tilerowslog2, --tilecolslog2, and --autotiling options change the encoder's tiling mode between manual and automatic. Document that the encoder starts in manual tiling.

Bug: b:328995727

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -279,12 +279,12 @@ static void syntaxLong(void)
            AVIF_QUALITY_LOSSLESS);
 #endif
     printf("    --tilerowslog2 R                  : log2 of number of tile rows in 0..6. (Default: 0)\n");
-    printf("                                        If specified, switch to manual tiling\n");
+    printf("                                        If specified, switch to manual tiling.\n");
     printf("    --tilecolslog2 C                  : log2 of number of tile columns 0..6. (Default: 0)\n");
-    printf("                                        If specified, switch to manual tiling\n");
+    printf("                                        If specified, switch to manual tiling.\n");
     printf("    --autotiling                      : Set --tilerowslog2 and --tilecolslog2 automatically\n");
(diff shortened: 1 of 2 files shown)


apps/avifenc.c
```diff
@@ -2121,9 +2125,9 @@ int main(int argc, char * argv[])
         avifInputFileSettings * fileSettings = &file->settings;
 
         // Check tiling parameters.
-        // Auto tiling (autoTiling) and manual tiling (tileRowsLog2, tileColsLog2) are mutually exclusive, which means:
-        // - At each input, only one of the two shall be set.
-        // - At some input, specify one disables the other.
+        // Automatic tiling (autoTiling) and manual tiling (tileRowsLog2, tileColsLog2) are mutually exclusive, which means:
+        // - At each input, only one of the two may be set.
```
wantehchang, line 2129: The original word "shall" seems to imply that either auto tiling or manual tiling must be set at each input. I changed "shall" to "may" to suggest that it is optional to set one of the two tiling modes.

apps/avifenc.c
```diff
@@ -279,8 +279,12 @@ static void syntaxLong(void)
            AVIF_QUALITY_LOSSLESS);
 #endif
     printf("    --tilerowslog2 R                  : log2 of number of tile rows in 0..6. (Default: 0)\n");
+    printf("                                        If specified, switch to manual tiling\n");
```
maryla-uc, line 282: Please update avifenc.1.md as well

maryla-uc (approved):

apps/avifenc.c
wantehchang: Done. Thanks for the reminder.

wantehchang (commented):
Maryla: Please review the second commit. Thanks!

PR #2750, opened 2025-04-04
Fix cargo-c with Rust 1.86

vrabaud:
There is apparently no need to apply the patch in ext because `--locked` is used

### DIFF
--- a/cmake/Modules/LocalRav1e.cmake
+++ b/cmake/Modules/LocalRav1e.cmake
@@ -51,6 +51,9 @@ else()
             GIT_REPOSITORY https://github.com/lu-zero/cargo-c.git
             GIT_TAG "${AVIF_CARGOC_GIT_TAG}"
             GIT_SHALLOW ON
+            # Patch from https://github.com/lu-zero/cargo-c/pull/453
+            PATCH_COMMAND git apply --ignore-whitespace "${AVIF_SOURCE_DIR}/ext/cargo.patch"
+            UPDATE_COMMAND ""
         )
         FetchContent_MakeAvailable(cargoc)
 
--- a/ext/cargo.patch
+++ b/ext/cargo.patch
@@ -1,0 +1,28 @@
+--- Cargo.toml	2025-04-04 09:52:07.160571748 +0200
++++ Cargo.toml_new	2025-04-04 09:52:47.321096631 +0200
+@@ -1,6 +1,6 @@
+ [package]
+ name = "cargo-c"
+-version = "0.10.11+cargo-0.86.0"
++version = "0.10.12+cargo-0.87.0"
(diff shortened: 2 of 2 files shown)


cmake/Modules/LocalRav1e.cmake
```diff
@@ -51,6 +51,9 @@ else()
             GIT_REPOSITORY https://github.com/lu-zero/cargo-c.git
             GIT_TAG "${AVIF_CARGOC_GIT_TAG}"
             GIT_SHALLOW ON
+            # Patch from https://github.com/lu-zero/cargo-c/pull/453
+            PATCH_COMMAND sed -i.bak -e "s/0.10.11+cargo-0.86.0/0.10.12+cargo-0.87.0/g" Cargo.toml && sed -i.bak -e
```
maryla-uc, line 55: can you do a git patch file instead? sed can be annoying cross paltform as we've seen in the past

vrabaud: Done

cmake/Modules/LocalRav1e.cmake
```diff
@@ -51,6 +51,8 @@ else()
             GIT_REPOSITORY https://github.com/lu-zero/cargo-c.git
             GIT_TAG "${AVIF_CARGOC_GIT_TAG}"
             GIT_SHALLOW ON
+            # Patch from https://github.com/lu-zero/cargo-c/pull/453
+            PATCH_COMMAND git apply --ignore-whitespace "${AVIF_SOURCE_DIR}/ext/cargo.patch"
```
maryla-uc, line 55: You need `UPDATE_COMMAND ""` as well

maryla-uc (approved):
Thanks!

cmake/Modules/LocalRav1e.cmake
vrabaud: Done

wantehchang:
Vincent: Thanks a lot for the fix! Should we also use `--locked` in cmake/Modules/LocalRav1e.cmake?

cmake/Modules/LocalRav1e.cmake
maryla-uc: Yes, otherwise on update cmake will try to reapply the patch on the existing checkout of the depencency, which will fail the second time since the patch was already applied once. `UPDATE_COMMAND ""` tells it to not update but fetch a fresh copy every time.
That's my understanding at least.

vrabaud:
I tried as there is a LOCKED option in the corrosion_import_crate macro to no avail ... This is a remporary fix anyway so let's just wait. It seems there will be a longer term solution in cargo so that it does not happen again.

Issue #2752, opened 2025-04-07
x86 build compiled with mingw32 does not open


K4sum1:
I'm using mingw32 with clang to compile libavif. My setup works to compile x86 libjxl, so I don't think it's an issue with my setup.

I'm wanting to make a x86 build of libavif to package in my fork of [xl-converter](https://github.com/K4sum1/xl-converter). I'm able to get it to compile, but when I open it it errors asking for libgcc_s_dw2-1.dll. If I give it the dll, it will then error with 0xc000007b.

I've tried the Build everything from scratch debug command in readme.md and this custom release command I made below trying to make it statically link and not require the libgcc dll to no avail.

```
cmake -S libavif -B libavif/build -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DCMAKE_EXE_LINKER_FLAGS=-s -DCMAKE_C_FLAGS_RELEASE="-static -static-libgcc -static-libstdc++ -O3 -DNDEBUG" -DAVIF_CODEC_AOM=LOCAL -DAVIF_LIBYUV=LOCAL -DAVIF_LIBSHARPYUV=LOCAL -DAVIF_JPEG=LOCAL -DAVIF_ZLIBPNG=LOCAL -DAVIF_BUILD_APPS=ON -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
```

vrabaud:
We have a CI for mingw32 that works (the tests run). Maybe you can have a look at it? https://github.com/AOMediaCodec/libavif/blob/main/.github/workflows/ci-mingw.yml

K4sum1:
Well it builds with that, it's considerably smaller, but it depends on more dlls. I can satisfy it by running it from mingw32 prompt, but I want a static single binary like the GitHub releases.

wantehchang:
@K4sum1 Thanks for reporting the compiler warning. I wrote https://github.com/AOMediaCodec/libavif/pull/2753 to fix the compiler warning.

That compiler warning is benign because we have verified it is safe to shorten the `uint64_t` value to the 32-bit `size_t` type.

You can remove the `-DAVIF_ENABLE_WERROR=ON` cmake flag to allow the compiler warning.

K4sum1:
I was able to find a command that works, gives me a fully static libavif build that works on other systems. So I'm going to close the issue.

```
[...]

PR #2753, opened 2025-04-07
Cast uint64_t to size_t before assigning to size_t

wantehchang:
Reported in https://github.com/AOMediaCodec/libavif/issues/2752#issuecomment-2784725617

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -4766,7 +4766,7 @@ static avifResult avifParse(avifDecoder * decoder)
             if (header.isSizeZeroBox) {
                 // The box body goes till the end of the file.
                 if (decoder->io->sizeHint != 0 && decoder->io->sizeHint - parseOffset < SIZE_MAX) {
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -4766,7 +4766,7 @@ static avifResult avifParse(avifDecoder * decoder)
             if (header.isSizeZeroBox) {
                 // The box body goes till the end of the file.
                 if (decoder->io->sizeHint != 0 && decoder->io->sizeHint - parseOffset < SIZE_MAX) {
-                    sizeToRead = decoder->io->sizeHint - parseOffset;
+                    sizeToRead = (size_t)(decoder->io->sizeHint - parseOffset);
```
wantehchang, line 4769: Both `decoder->io->sizeHint` and `parseOffset` are of the `uint64_t` type. `sizeToRead` is of the `size_t` type.

jzern (approved):

PR #2754, opened 2025-04-08
Test manual and automatic tiling options

wantehchang:
Bug: b:328995727


maryla-uc (approved):
Do you want to have a test that uses ":u" as well?

wantehchang:
Yes, I should also add tests that use ":u".  I am not familiar with the ":u" suffix. That's why I didn't add tests that use ":u".

I found nontrivial uses of ":u" in test_cmd_animation.sh and test_cmd_progressive.sh. I will look into adding tests to test_cmd_animation.sh before I change avifenc's initial tiling mode from manual to automatic.

Issue #2741, opened 2025-04-02
avifgainmaptest fails on ppc64le


kmilos:
From a recent [Fedora scratch build](https://koji.fedoraproject.org/koji/taskinfo?taskID=130908363):
```
[ RUN      ] GainMapTest.EncodeDecodeGrid
WARNING: computing PSNR of images with different CICP: 2/2/2f vs 2/16/2f
/builddir/build/BUILD/libavif-1.2.1-build/libavif-1.2.1/tests/gtest/avifgainmaptest.cc:404: Failure
Expected: (testutil::GetPsnr(*merged_gain_map, *decoded->gainMap->image)) > (40.0), actual: 16.783084124813779 vs 40
```
P.S. Just in case libyuv is involved in this test, note that Fedora is currently shipping a relatively old one...

wantehchang:
Assigned the issue to Maryla. Miloš, do you have any suggestion on how we can debug this if we don't have a ppc64le machine?

jzern:
qemu might work.

jzern:
This reproduces on Fedora Workstation 41 under qemu:

```
[ RUN      ] GainMapTest.EncodeDecodeGrid
WARNING: computing PSNR of images with different CICP: 2/2/2f vs 2/16/2f
/home/user/external/libavif/tests/gtest/avifgainmaptest.cc:404: Failure
Expected: (testutil::GetPsnr(*merged_gain_map, *decoded->gainMap->image)) > (40.0), actual: 16.783084124813779 vs 40
```

Something like:

```
$ curl -O -L \
  https://download.fedoraproject.org/pub/fedora-secondary/releases/41/Workstation/ppc64le/iso/Fedora-Workstation-Live-ppc64le-41-1.4.iso
```

```
$ qemu-img create -f qcow2 fedora-workstation41-ppc64el.qcow2 16G
$ qemu-system-ppc64el -boot d \
  -cdrom Fedora-Workstation-Live-ppc64le-41-1.4.iso \
  -m 8G -smp 8 -hda fedora-workstation41-ppc64el.qcow2
```

(Install from the live disk, create a user/user account for simplicity; the UI will be slow...)

```
$ qemu-system-ppc64el -hda fedora-workstation41-ppc64el.qcow2 -m 8G -smp 8 -nic user,hostfwd=tcp::1234-:22
```

(Enable ssh in the settings and then you can ssh to avoid the UI)

```
[...]

PR #2759, opened 2025-04-10
update libaom to v3.12.0-4-g65ddc22823

jzern:
This fixes an encode & decode issue on PowerPC (ppc64le).

Fixes: #2741


jzern:
I went with the stable hash even though it means aom.cmd needs to do a full clone, just to keep things consistent. That way we know what was checked out by a user at any point in time.

wantehchang (approved):

PR #2762, opened 2025-04-11
Fix comment issues related to "auto"

wantehchang:
In particular, the comment on settings->autoProgressive is incorrect or out of date. The field is named "progressive", not "autoProgressive", and the action of the main function is to return an error, not set settings->progressive to false.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -230,7 +230,7 @@ static void syntaxLong(void)
     printf("                                        Use 2 for any you wish to leave unspecified\n");
     printf("    -r,--range RANGE                  : YUV range, one of 'limited' or 'l', 'full' or 'f'. (JPEG/PNG only, default: full; For y4m or stdin, range is retained)\n");
     printf("    --target-size S                   : Set target file size in bytes (up to 7 times slower)\n");
-    printf("    --progressive                     : EXPERIMENTAL: Auto set parameters to encode a simple layered image supporting progressive rendering from a single input frame.\n");
+    printf("    --progressive                     : EXPERIMENTAL: Automatically set parameters to encode a simple layered image supporting progressive rendering from a single input frame.\n");
     printf("    --layered                         : EXPERIMENTAL: Encode a layered AVIF. Each input is encoded as one layer and at most %d layers can be encoded.\n",
            AVIF_MAX_AV1_LAYER_COUNT);
     printf("    -g,--grid MxN                     : Encode a single-image grid AVIF with M cols & N rows. Either supply MxN identical W/H/D images, or a single\n");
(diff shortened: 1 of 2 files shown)


apps/avifenc.c
```diff
@@ -1213,8 +1213,8 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
            encoder->autoTiling ? "automatic tiling" : manualTilingStr,
            settings->jobs);
     if (settings->progressive) {
-        // If the color quality is less than 10, the main() function overrides
-        // --progressive and sets settings->autoProgressive to false.
+        // If --progressive is specified and the color quality is less than 10,
+        // the main() function returns an error. So we should not reach here.
```
wantehchang, line 1217: Here is the code in the main() function that this comment refers to:

```
            if (settings.progressive) {
                assert(DEFAULT_QUALITY >= PROGRESSIVE_WORST_QUALITY);
                if (fileSettings->quality.set && fileSettings->quality.value < PROGRESSIVE_WORST_QUALITY) {
                    fprintf(stderr, "ERROR: --qcolor must be at least %d when using --progressive.\n", PROGRESSIVE_WORST_QUALITY);
                    goto cleanup;
                }
                // --progressive only adjust color quality
            }
```

apps/avifenc.c
```diff
@@ -1213,8 +1213,8 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
            encoder->autoTiling ? "automatic tiling" : manualTilingStr,
            settings->jobs);
     if (settings->progressive) {
-        // If the color quality is less than 10, the main() function overrides
-        // --progressive and sets settings->autoProgressive to false.
+        // If --progressive is specified and the color quality is less than 10,
```
y-guyon, line 1216:
```diff
-// If --progressive is specified and the color quality is less than 10,
+// If --progressive is specified and the color quality is less than PROGRESSIVE_WORST_QUALITY,
```

to avoid the comment going out of sync in the future

y-guyon (approved):
LGTM.

I would even remove mentions of "automatic":  
"Automatically set parameters" -> "Set parameters"  
"Not doing automatic layered encoding" -> "Not doing layered encoding"

apps/avifenc.c
```diff
@@ -230,7 +230,7 @@ static void syntaxLong(void)
     printf("                                        Use 2 for any you wish to leave unspecified\n");
     printf("    -r,--range RANGE                  : YUV range, one of 'limited' or 'l', 'full' or 'f'. (JPEG/PNG only, default: full; For y4m or stdin, range is retained)\n");
     printf("    --target-size S                   : Set target file size in bytes (up to 7 times slower)\n");
-    printf("    --progressive                     : EXPERIMENTAL: Auto set parameters to encode a simple layered image supporting progressive rendering from a single input frame.\n");
+    printf("    --progressive                     : EXPERIMENTAL: Automatically set parameters to encode a simple layered image supporting progressive rendering from a single input frame.\n");
```
jzern, line 233: 'Automatically' does sound a bit redundant here.

jzern (approved):

apps/avifenc.c
wantehchang: Done.

apps/avifenc.c
wantehchang: I looked into this. The reason "automatically" and "automatic" are used to describe the `--progressive` option is that the `--layered` option below (line 234) can also be used to encode a layered image but the parameters for each layer must be set manually.

wantehchang (commented):
Thanks for the review. The avifenc man page needs to be kept in sync with the avifenc help message. I took care of that in the second commit.

apps/avifenc.c
y-guyon: Could the `--progressive` doc describe what `--layered` arg it is equivalent to?

PR #2763, opened 2025-04-14
Small improvements to avif_example_encode.c

wantehchang:
Use U,V plane height instead of image->height to allow the user to use a YUV format other than 4:4:4.

Print the diagnostic error message if avifEncoderAddImage() fails.

### DIFF
--- a/examples/avif_example_encode.c
+++ b/examples/avif_example_encode.c
@@ -52,9 +52,10 @@ int main(int argc, char * argv[])
         }
 
         // Fill your YUV(A) data here
+        const uint32_t uvHeight = avifImagePlaneHeight(image, AVIF_CHAN_U);
         memset(image->yuvPlanes[AVIF_CHAN_Y], 255, image->yuvRowBytes[AVIF_CHAN_Y] * image->height);
-        memset(image->yuvPlanes[AVIF_CHAN_U], 128, image->yuvRowBytes[AVIF_CHAN_U] * image->height);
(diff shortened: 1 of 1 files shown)


examples/avif_example_encode.c
```diff
@@ -52,9 +52,10 @@ int main(int argc, char * argv[])
         }
 
         // Fill your YUV(A) data here
+        const uint32_t uvHeight = avifImagePlaneHeight(image, AVIF_CHAN_U);
         memset(image->yuvPlanes[AVIF_CHAN_Y], 255, image->yuvRowBytes[AVIF_CHAN_Y] * image->height);
-        memset(image->yuvPlanes[AVIF_CHAN_U], 128, image->yuvRowBytes[AVIF_CHAN_U] * image->height);
-        memset(image->yuvPlanes[AVIF_CHAN_V], 128, image->yuvRowBytes[AVIF_CHAN_V] * image->height);
+        memset(image->yuvPlanes[AVIF_CHAN_U], 128, image->yuvRowBytes[AVIF_CHAN_U] * uvHeight);
+        memset(image->yuvPlanes[AVIF_CHAN_V], 128, image->yuvRowBytes[AVIF_CHAN_V] * uvHeight);
```
wantehchang, line 58: This file uses `AVIF_PIXEL_FORMAT_YUV444` (line 28), which is why the original code works. If we modify this file to use `AVIF_PIXEL_FORMAT_YUV420` instead, we will need the change here.

y-guyon (approved):

PR #2766, opened 2025-04-15
Add context to MC 0 with 4:4:4 requirement

y-guyon:


src/write.c
```diff
@@ -1660,7 +1660,11 @@ static avifResult avifValidateGrid(uint32_t gridCols,
         }
 
         // AV1 (Version 1.0.0 with Errata 1), Section 6.4.2. Color config semantics
-        // If matrix coefficients is equal to MC_IDENTITY, it is a requirement of bitstream conformance that subsampling_x is equal to 0 and subsampling_y is equal to 0.
+        //   If matrix_coefficients is equal to MC_IDENTITY, it is a requirement of bitstream conformance that
+        //   subsampling_x is equal to 0 and subsampling_y is equal to 0.
+        // Although matrix_coefficients in the Sequence Header OBU is set to Undefined (2), the requirement
+        // is still enforced here between what is written in the ColourInformationProperty of colour_type 'nclx'
+        // and the subsampling information in the Sequence Header OBU.
```
wantehchang, line 1667: You are right that we are not checking the exact AV1 requirement here but rather a generalization of that requirement.

The requirement we are checking here is:

> If MC=0 (the identity matrix) is used such that the Y plane contains G samples, the U plane contains B samples, and the V plane contains R samples, then the U and V planes shall not be subsampled.

The fact that MC=0 can be signaled in the AV1 sequence header OBU ("matrix_coefficients is equal to MC_IDENTITY") or in the nclx colr box.

wantehchang (approved):

PR #2768, opened 2025-04-18
Add integer overflow checks to makeRoom.

DanisJiang:
In the function `makeRoom`, when `stream->offset` or `size` is very large, the calculation `stream->offset + size` may cause an integer overflow, resulting in an abnormally small `neededSize`.

This can lead to an incorrect `newSize` and insufficient buffer allocation, which may subsequently cause a buffer overflow during the memcpy operation.

This commit add a integer overflow check to prevent such issues.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -335,6 +335,9 @@ avifBool avifROStreamReadAndEnforceVersion(avifROStream * stream, uint8_t enforc
 static avifResult makeRoom(avifRWStream * stream, size_t size)
 {
     size_t neededSize = stream->offset + size;
+    if (neededSize < stream->offset) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
     size_t newSize = stream->raw->size;
     while (newSize < neededSize) {
         newSize += AVIF_STREAM_BUFFER_INCREMENT;


y-guyon:

Have you encountered such an integer overflow using the libavif public API?  
If so, it would be good to add a test exercising the fix.

src/stream.c
```diff
@@ -335,6 +335,9 @@ avifBool avifROStreamReadAndEnforceVersion(avifROStream * stream, uint8_t enforc
 static avifResult makeRoom(avifRWStream * stream, size_t size)
 {
     size_t neededSize = stream->offset + size;
+    if (neededSize < stream->offset) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
```
wantehchang, line 340: Thanks for the patch. This detects unsigned integer overflow after it has happened. We can detect the overflow before it happens as follows:

```
    if (size > SIZE_MAX - stream->offset) {
        return  AVIF_RESULT_OUT_OF_MEMORY;
    }
    size_t neededSize = stream->offset + size;
```

Note that I also changed the error code to `AVIF_RESULT_OUT_OF_MEMORY`, which is more informative.

DanisJiang: Thanks for the suggestion, I've made the changes.

DanisJiang:
Sorry, I mistakenly included a commit from another pull request here. I've already removed that commit with a force push.

src/stream.c
```diff
@@ -334,6 +334,9 @@ avifBool avifROStreamReadAndEnforceVersion(avifROStream * stream, uint8_t enforc
 #define AVIF_STREAM_BUFFER_INCREMENT (1024 * 1024)
 static avifResult makeRoom(avifRWStream * stream, size_t size)
 {
+    if (size > SIZE_MAX - stream->offset) {
+        return  AVIF_RESULT_OUT_OF_MEMORY;
```
wantehchang, line 338: Nit: Use only one space character between "return" and "AVIF_RESULT_OUT_OF_MEMORY".

src/stream.c
```diff
@@ -334,6 +334,9 @@ avifBool avifROStreamReadAndEnforceVersion(avifROStream * stream, uint8_t enforc
...
+    }
```
wantehchang, line 339: Yannis: You asked:

> Have you encountered such an integer overflow using the libavif public API?
> If so, it would be good to add a test exercising the fix.

`makeRoom` is called with either a small `size` or the size of a valid memory buffer. We can assume that the size of a valid memory buffer is < `SIZE_MAX / 2`. Since `stream->offset` is <= the size of a valid memory buffer, the sum of `size` and `stream->offset` cannot possibly exceed `SIZE_MAX` under this assumption. So any test that tries to trigger this overflow will first fail in memory allocation.

Having said that, I think it is fine to add this kind of overflow check so that we don't need to perform this kind of analysis.

y-guyon: > So any test that tries to trigger this overflow will first fail in memory allocation.

I was thinking of not really allocating the full buffer, but just passing some non-NULL buffer. This should add coverage for the overflow check and fail before triggering an access past the end of the actually allocated buffer.

wantehchang: That's a clever idea. Yes, we should be able to make the overflow check fail before accessing the buffer with an invalid size.

wantehchang (approved):
Sorry about the delay. I was trying to avoid approving my own suggestion.

wantehchang:
Oops, I neglected to squash the three commits when I merged this pull request. Sorry!

Princemachiavelli:
This PR fixes CVE-2025-48174.

PR #2769, opened 2025-04-18
Declare *RowBytes as size_t in avifImageRGBToYUV()

wantehchang:
Declare rgbRowBytes, yRowBytes, uRowBytes, and vRowBytes as size_t in avifImageRGBToYUV(). This causes multiplications with these variables to be performed in size_t (which may be 64 bits) instead of uint32_t. For very large image width and height, these multiplications may overflow uint32_t.

Acknowledgements: DanisJiang
https://github.com/AOMediaCodec/libavif/security/advisories/GHSA-762c-2538-h844

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -273,14 +273,14 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
         const uint32_t offsetBytesG = state.rgb.offsetBytesG;
         const uint32_t offsetBytesB = state.rgb.offsetBytesB;
         const uint32_t offsetBytesA = state.rgb.offsetBytesA;
-        const uint32_t rgbRowBytes = rgb->rowBytes;
+        const size_t rgbRowBytes = rgb->rowBytes;
         const float rgbMaxChannelF = state.rgb.maxChannelF;
         uint8_t * yPlane = image->yuvPlanes[AVIF_CHAN_Y];
         uint8_t * uPlane = image->yuvPlanes[AVIF_CHAN_U];
         uint8_t * vPlane = image->yuvPlanes[AVIF_CHAN_V];
-        const uint32_t yRowBytes = image->yuvRowBytes[AVIF_CHAN_Y];
-        const uint32_t uRowBytes = image->yuvRowBytes[AVIF_CHAN_U];
-        const uint32_t vRowBytes = image->yuvRowBytes[AVIF_CHAN_V];
+        const size_t yRowBytes = image->yuvRowBytes[AVIF_CHAN_Y];
+        const size_t uRowBytes = image->yuvRowBytes[AVIF_CHAN_U];
+        const size_t vRowBytes = image->yuvRowBytes[AVIF_CHAN_V];
         for (uint32_t outerJ = 0; outerJ < image->height; outerJ += 2) {
             for (uint32_t outerI = 0; outerI < image->width; outerI += 2) {
                 int blockW = 2, blockH = 2;


y-guyon (approved):

jzern (approved):

wantehchang:
I just remembered that I previously took a different approach to fix this kind of bug: incrementing the buffer pointers in the for loops. But I don't have time to try that approach soon, so I will first merge this pull request as is.

charles2910:
Hi @wantehchang.

I'm working to fix CVE-2025-48175 in Debian stable versions, but the code is a bit different before 1.2.0 (more precisely before 1b4ce5ca24a) so I'm trying to get as much info as possible to correctly fix the vulnerability and I noticed [GHSA-762c-2538-h844](https://github.com/AOMediaCodec/libavif/security/advisories/GHSA-762c-2538-h844) is giving a 404 error, so I'm assuming is not public. Do you plan to make it public? If so, do you have a date in mind?

Thanks!

wantehchang:
Hi Carlos,

libavif v1.0.0 and later have a stable API and ABI. For libavif v1.0.0 and later, I recommend updating to the latest release (v1.3.0) rather than cherry-picking bug fixes. Cherry-picking a bug fix to an older version that you are not familiar with may have its own risks.

The PoC in [GHSA-762c-2538-h844](https://github.com/AOMediaCodec/libavif/security/advisories/GHSA-762c-2538-h844) is the following:

```
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include "avif/avif.h"

int main(void) {
    const uint32_t width = 65536;
    const uint32_t height = 65536;
    const uint32_t depth = 16;
    avifImage *image = avifImageCreate(width, height, depth,
                                       AVIF_PIXEL_FORMAT_YUV444);
    if (!image) {
        printf("Failed to create image\n");
        return 1;
    }

    avifRGBImage rgb;

    rgb.width = width;
    rgb.height = height;
    rgb.depth = depth;
    rgb.rowBytes = 3 * rgb.width * (depth > 8 ? 2 : 1);

    if (rgb.rowBytes > SIZE_MAX / rgb.height) {
        printf("Failed to calculate RGB buffer size\n");
        return 1;
    }
    // NOTE: Must cast operand to size_t.
    const size_t rgbBufSize = (size_t)rgb.rowBytes * rgb.height;
    rgb.pixels = malloc(rgbBufSize);
    if (!rgb.pixels) {
        printf("Failed to allocate RGB buffer\n");
        avifImageDestroy(image);
        return 1;
    }
    memset(rgb.pixels, 0, rgbBufSize);

    avifResult result = avifImageAllocatePlanes(image, AVIF_PLANES_YUV);
        printf("Plane allocation failed\n");
        free(rgb.pixels);
        avifImageDestroy(image);
        return 1;
    }

    result = avifImageRGBToYUV(image, &rgb);
    printf("Result: %s\n", avifResultToString(result));

    free(rgb.pixels);
    avifImageDestroy(image);

    return 0;
}
```

Running the PoC under ASan doesn't report errors. Run the PoC under UBSan with unsigned integer overflow checks enabled, i.e., compile the code with -fsanitize=undefined -fsanitize=unsigned-integer-overflow.

Another approach you can consider to avoid this class of bug is to impose a maximum image area (width * height) on the `avifImageRGBToYUV()` function. If you have a specific older version you need to fix, I can try preparing a patch for you.

charles2910:
Hi again,

thanks for the PoC and instructions on how to trigger the checks!

> libavif v1.0.0 and later have a stable API and ABI. For libavif v1.0.0 and later, I recommend updating to the latest release (v1.3.0) rather than cherry-picking bug fixes. Cherry-picking a bug fix to an older version that you are not familiar with may have its own risks.

Unfortunately we are talking about 0.11.1, 0.8.4 and older. I understand what you are saying and in Debian we occasionally do it for very specific cases (very very stable API/ABI or very nasty code to backport), but the preferred way is to only fix the vulnerability.

> If you have a specific older version you need to fix, I can try preparing a patch for you.

I'm very grateful, but I don't think it's fair since this is part of a sponsored work I'm doing for Debian ELTS initiative maintained by Freexian. Though I would appreciate if you could just take a quick look in the patches to see if I didn't do a terrible job and introduced regressions or vulnerabilities :-)

wantehchang:
I recommend the following patch for `avifImageRGBToYUV()` in libavif v0.11.1 and older:

```
index 951c46b5..9f760396 100644
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -196,6 +196,11 @@ static int
avifReformatStateUVToUNorm(avifReformatState * state, float v)

 avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
 {
+    // Avoid integer overflow in (32-bit) int or unsigned int arithmetic operations.
+    if ((uint64_t)rgb->width * rgb->height > AVIF_DEFAULT_IMAGE_SIZE_LIMIT) {
+        return AVIF_RESULT_REFORMAT_FAILED;
+    }
+
     if (!rgb->pixels || rgb->format == AVIF_RGB_FORMAT_RGB_565) {
     }
```

The idea of this patch is to assume the existence of integer overflow in the code in `avifImageRGBToYUV()` and only enter the function when the image width and height are not too big. We have a similar protection in `avifDecoder`. Since `avifImageRGBToYUV()` is typically used to prepare the input to `avifEncoder`, I didn't add this protection to `avifImageRGBToYUV()`.

https://github.com/AOMediaCodec/libavif/commit/2ded15b09 has some context for the image size (area) and dimension limits. For this `avifImageRGBToYUV()` issue, the image size (area) limit is sufficient. The image dimension limit is intended to avoid spending a very long time decoding an image.

wantehchang:
Yes. In commit https://github.com/AOMediaCodec/libavif/commit/980d585b6, the `AVIF_MAX_IMAGE_SIZE` macro was renamed `AVIF_DEFAULT_IMAGE_SIZE_LIMIT` and moved from the internal header "avif/internal.h" to the public header "avif/avif.h". The value of the macro did not change.

PR #2776, opened 2025-04-24
Fix the generator expression for libjpeg on Windows.

vrabaud:


cmake/Modules/LocalJpeg.cmake
```diff
@@ -35,7 +35,7 @@ else()
         GIT_REPOSITORY "https://github.com/libjpeg-turbo/libjpeg-turbo.git"
         GIT_TAG "${AVIF_JPEG_TAG}"
         LIST_SEPARATOR |
-        BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --config $<CONFIG> --target jpeg-static
+        BUILD_COMMAND ${CMAKE_COMMAND} --build $<BINARY_DIR> --config $<CONFIG> --target jpeg-static
```
wantehchang, line 38: The original code is correct according to CMake documentation:
https://cmake.org/cmake/help/latest/module/ExternalProject.html

Search for "BINARY_DIR" in that page.

So something else is wrong.

vrabaud: Reverting

cmake/Modules/LocalJpeg.cmake
```diff
@@ -1,4 +1,4 @@
-set(AVIF_JPEG_TAG "3.0.4")
+set(AVIF_JPEG_TAG "3.1.0")
```
wantehchang, line 1: Vincent: In libjpeg-turbo 3.1.0, a `src` subdirectory was added and the source files, including jpeglib.h, were moved into the `src` subdirectory. You can reproduce the compilation failures on any machine that doesn't have jpeglib.h installed as a system header. I bet that GitHub's os runner images for Linux and macOS already have jpeglib.h installed as a system header.

To fix the compilation errors, add the following to this pull request:

```
diff --git a/cmake/Modules/LocalJpeg.cmake b/cmake/Modules/LocalJpeg.cmake
index 302d9d9e..5b5bcd88 100644
--- a/cmake/Modules/LocalJpeg.cmake
+++ b/cmake/Modules/LocalJpeg.cmake
@@ -13,7 +13,7 @@ set(LIB_FILENAME "${LIB_DIR}/${LIB_BASENAME}")
 if(EXISTS "${LIB_FILENAME}")
     message(STATUS "libavif(AVIF_JPEG=LOCAL): ${LIB_FILENAME} found, using for local JPEG")
     set_target_properties(JPEG::JPEG PROPERTIES IMPORTED_LOCATION "${LIB_FILENAME}")
-    set(JPEG_INCLUDE_DIR "${AVIF_SOURCE_DIR}/ext/libjpeg-turbo")
+    set(JPEG_INCLUDE_DIR "${AVIF_SOURCE_DIR}/ext/libjpeg-turbo/src")
 else()
     message(STATUS "libavif(AVIF_JPEG=LOCAL): ${LIB_FILENAME} not found, fetching")
     set(LIB_DIR "${CMAKE_CURRENT_BINARY_DIR}/libjpeg/src/libjpeg-build")
@@ -70,11 +70,11 @@ else()
         INSTALL_COMMAND ""
     )
     add_dependencies(JPEG::JPEG libjpeg)
-    set(JPEG_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/libjpeg/src/libjpeg)
+    set(JPEG_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/libjpeg/src/libjpeg/src)
 endif()
 
 set_target_properties(JPEG::JPEG PROPERTIES AVIF_LOCAL ON)
-target_include_directories(JPEG::JPEG INTERFACE "${JPEG_INCLUDE_DIR}")
+target_include_directories(JPEG::JPEG INTERFACE $<INSTALL_INTERFACE:"${JPEG_INCLUDE_DIR}">)
 
 # Also add the build directory path because it contains jconfig.h,
 # which is included by jpeglib.h.
```

I had to add `$<INSTALL_INTERFACE:` to suppress CMake's check that the include directory exists. If CMake performs that check, we have not fetched the libjpeg-turbo source tree, so new value of `JPEG_INCLUDE_DIR` doesn't exists yet. If you have spare time, you may want to remove the `$<INSTALL_INTERFACE:` to reproduce the nonexistent include directory error, and see if you can come up with a better solution.

vrabaud: Thx for the details: it all works now!

Issue #2775, opened 2025-04-24
Build on Windows fails


LexaGV:
Win7 x64 SP1, cmake 4.0.1, python 3.8.10, nasm 2.16.03

When I do
    cmake -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_DAV1D=LOCAL -DAVIF_LIBYUV=LOCAL -DBUILD_SHARED_LIBS=OFF -DAVIF_BUILD_APPS=ON ../

I get this error:

    CMake Error at C:/Program Files/CMake/share/cmake-4.0/Modules/FindPackageHandleStandardArgs.cmake:227 (message):
    Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR)

Why cmake doesn't download it? I never had separate zlib, since all app support their dependencies.

vrabaud:
The apps need zlib indeed. To download it, just add `-DAVIF_ZLIBPNG=LOCAL`

vrabaud:
The README should contain everything you need, please let us know if there is something missing:  https://github.com/AOMediaCodec/libavif/blob/main/README.md#build-everything-from-scratch

LexaGV:
@vrabaud  -DAVIF_JPEG=LOCAL - that was missing. It could be nice to put both missing vars into [Building command line encoder avifenc](https://codelabs.developers.google.com/codelabs/avif#2)

One more fail:
    LINK : fatal error LNK1181: cannot open input file 'libjpeg\src\libjpeg-build\jpeg-static.lib' [r:\libavif\build\avifenc.vcxproj]

Mentioned jpeg-static.lib is one level deeper:

    'libjpeg\src\libjpeg-build\Release\jpeg-static.lib'
or
    'libjpeg\src\libjpeg-build\Debug\jpeg-static.lib'

LexaGV:
@vrabaud I have VS 2022 17.6.4; Linker is from the same toolchain. And using Release/Debug subfolders is standard for VS. Obviously something was custom configured to have lib in inappropriate folder.

wantehchang:
@LexaGV I can reproduce the build error if I generate a Visual Studio solution file. If I pass `-G Ninja` to cmake, then the build succeeds. Our CI workflows all use Ninja to build. That's why we don't see this build error in our CI workflows.

[...]

PR #2777, opened 2025-04-24
LocalJpeg.cmake: Support multi-config generators

wantehchang:
Add the locations of the libjpeg-turbo static library when a multi-configuration generator such as Visual Studio is used with cmake. Support the four typical configuration types: Debug, Release, RelWithDebInfo and MinSizeRel.

Fixes https://github.com/AOMediaCodec/libavif/issues/2775.


cmake/Modules/LocalJpeg.cmake
```diff
@@ -53,6 +62,14 @@ else()
...
     add_dependencies(JPEG::JPEG libjpeg)
     set(JPEG_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/libjpeg/src/libjpeg)
+    if(IS_MULTI_CONFIG)
+        set_target_properties(JPEG::JPEG PROPERTIES IMPORTED_LOCATION_DEBUG "${LIB_FILENAME_DEBUG}" AVIF_LOCAL ON)
+        set_target_properties(JPEG::JPEG PROPERTIES IMPORTED_LOCATION_MINSIZEREL "${LIB_FILENAME_MINSIZEREL}" AVIF_LOCAL ON)
+        set_target_properties(JPEG::JPEG PROPERTIES IMPORTED_LOCATION_RELEASE "${LIB_FILENAME_RELEASE}" AVIF_LOCAL ON)
+        set_target_properties(
+            JPEG::JPEG PROPERTIES IMPORTED_LOCATION_RELWITHDEBINFO "${LIB_FILENAME_RELWITHDEBINFO}" AVIF_LOCAL ON
+        )
+    endif()
```
wantehchang, line 72: Vincent: Ideally we would change line 75 to use a generator expression that contains `$<CONFIG>` as the value of the [`IMPORTED_LOCATION`](https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED_LOCATION.html) property. Unfortunately the value of `IMPORTED_LOCATION` cannot be a generator expression. So using `IMPORTED_LOCATION_<CONFIG>` is the best solution I came up with.

cmake/Modules/LocalJpeg.cmake
```diff
@@ -2,22 +2,27 @@ set(AVIF_JPEG_TAG "3.0.4")
...
-        set(LIB_FILENAME "${LIB_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}jpeg-static${CMAKE_STATIC_LIBRARY_SUFFIX}")
-    else()
-        set(LIB_FILENAME "${LIB_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}jpeg${CMAKE_STATIC_LIBRARY_SUFFIX}")
+    set(LIB_FILENAME "${LIB_DIR}/${LIB_BASENAME}")
+    get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+    if(IS_MULTI_CONFIG)
+        set(LIB_FILENAME_DEBUG "${LIB_DIR}/Debug/${LIB_BASENAME}")
+        set(LIB_FILENAME_MINSIZEREL "${LIB_DIR}/MinSizeRel/${LIB_BASENAME}")
+        set(LIB_FILENAME_RELEASE "${LIB_DIR}/Release/${LIB_BASENAME}")
+        set(LIB_FILENAME_RELWITHDEBINFO "${LIB_DIR}/RelWithDebInfo/${LIB_BASENAME}")
```
wantehchang, line 25: These four lines (22-25) and lines 66-71 below probably can be handled by for loops.

cmake/Modules/LocalJpeg.cmake
```diff
@@ -29,6 +34,10 @@ else()
     #   Manually-specified variables were not used by the project:
     #
     #     WITH_CRT_DLL
+    #
+    # TODO(wtc): Delete one of -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} and
+    # -DCMAKE_BUILD_TYPE=Release.
```
wantehchang, line 39: See line 49 and line 57.

cmake/Modules/LocalJpeg.cmake
```diff
@@ -29,6 +34,10 @@ else()
...
+    # TODO(wtc): Should BUILD_BYPRODUCTS include ${LIB_FILENAME_DEBUG}, etc.?
```
wantehchang, line 40: See line 60.

cmake/Modules/LocalJpeg.cmake
```diff
@@ -2,22 +2,27 @@ set(AVIF_JPEG_TAG "3.0.4")
...
     set(JPEG_INCLUDE_DIR "${AVIF_SOURCE_DIR}/ext/libjpeg-turbo")
 else()
     message(STATUS "libavif(AVIF_JPEG=LOCAL): ${LIB_FILENAME} not found, fetching")
     set(LIB_DIR "${CMAKE_CURRENT_BINARY_DIR}/libjpeg/src/libjpeg-build")
-    if(MSVC)
-        set(LIB_FILENAME "${LIB_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}jpeg-static${CMAKE_STATIC_LIBRARY_SUFFIX}")
-    else()
-        set(LIB_FILENAME "${LIB_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}jpeg${CMAKE_STATIC_LIBRARY_SUFFIX}")
+    set(LIB_FILENAME "${LIB_DIR}/${LIB_BASENAME}")
+    get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
```
wantehchang, line 20: Vincent: While investigating this Visual Studio build error, I noticed that we are using `_deps\dav1d-install\lib\libdav1d.a` for all configuration types. This means we are installing the Debug, MinSizeRel, Release, RelWithDebInfo versions of libdav1d.a to the same location and use the installed libdav1d.a in all configuration types. This is not ideal.

cmake/Modules/LocalJpeg.cmake
vrabaud: Then again you'd need to set the string to upper. Probably roughly the same number of lines

cmake/Modules/LocalJpeg.cmake
vrabaud: go with release, that is what we do for dav1d, rav1e, svt

cmake/Modules/LocalJpeg.cmake
vrabaud: I agree. That can be fixed in a different CL. ExternalProject_Add is only used for dav1d (because it uses meson) and libjpeg-turbo (it officially states in does not support add_subdirectory, cf https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/BUILDING.md#sub-project-builds)

cmake/Modules/LocalJpeg.cmake
vrabaud: I could not find a better solution either. I am not sure a for loop would be cleaner.

vrabaud (approved):

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Change avifenc to start in automatic tiling mode.
 * Always forward Unspecified (2) CICP color primaries, transfer characteristics,
   and matrix coefficients to the AV1 encoder. Rely on the 'colr' box instead.
+* LocalJpeg.cmake: Support multi-config generators.
```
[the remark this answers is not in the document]
vrabaud: we can add it to the matrix in ci-windows-installed which is the fastest windows CI.

Just add to https://github.com/AOMediaCodec/libavif/blob/2d0204485a30446d82770c115e0a4d61e2819f23/.github/workflows/ci-windows-installed.yml#L24:
```
generator: [ Ninja, "Visual Studio 17 2022"]
```
And then just modify the `-G Ninja` below with `-G ${{ matrix.generator }}`

.github/workflows/ci-windows.yml
```diff
@@ -75,7 +76,7 @@ jobs:
         run: cmake --build build --config Release --parallel 4
       - name: Run AVIF Tests
         working-directory: ./build
-        run: ctest -j $Env:NUMBER_OF_PROCESSORS --output-on-failure
+        run: ctest -C Release -j $Env:NUMBER_OF_PROCESSORS --output-on-failure
```
wantehchang, line 79: I am having a hard time getting the test_cmd*.sh tests to work with multi-config generators. The next step, "Check static link bundling", also needs to be updated to support multi-config generators. So I will remove the 6th commit from this pull request so that the pull request can be merged. Here are the (correct but incomplete) changes to .github/workflows/ci-windows.yml for future reference:

```
diff --git a/.github/workflows/ci-windows.yml b/.github/workflows/ci-windows.yml
index c40a9804..f6ea0ae2 100644
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -36,6 +36,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest]
+        generator: [Ninja, "Visual Studio 17 2022"]
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -58,7 +59,7 @@ jobs:
 
         run: >
-          cmake -G Ninja -S . -B build
+          cmake -G "${{ matrix.generator }}" -S . -B build
           -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF
           -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_DAV1D=LOCAL
           -DAVIF_CODEC_RAV1E=LOCAL -DAVIF_CODEC_SVT=LOCAL
@@ -75,7 +76,7 @@ jobs:
         run: cmake --build build --config Release --parallel 4
       - name: Run AVIF Tests
         working-directory: ./build
-        run: ctest -j $Env:NUMBER_OF_PROCESSORS --output-on-failure
+        run: ctest -C Release -j $Env:NUMBER_OF_PROCESSORS --output-on-failure
         run: >
           cl .\apps\avifenc.c .\apps\shared\*.c .\third_party\iccjpeg\iccjpeg.c /nologo
```

PR #2778, opened 2025-04-25
Add another integer overflow check to makeRoom

wantehchang:
Test the previous integer overflow check in makeRoom.

See https://github.com/AOMediaCodec/libavif/pull/2768.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -334,17 +334,16 @@ avifBool avifROStreamReadAndEnforceVersion(avifROStream * stream, uint8_t enforc
 #define AVIF_STREAM_BUFFER_INCREMENT (1024 * 1024)
 static avifResult makeRoom(avifRWStream * stream, size_t size)
 {
-    if (size > SIZE_MAX - stream->offset) {
-        return AVIF_RESULT_OUT_OF_MEMORY;
+    AVIF_CHECKERR(size <= SIZE_MAX - stream->offset, AVIF_RESULT_OUT_OF_MEMORY);
+    size_t newSize = stream->offset + size;
+    if (newSize <= stream->raw->size) {
+        return AVIF_RESULT_OK;
     }
-    size_t neededSize = stream->offset + size;
-    size_t newSize = stream->raw->size;
-    while (newSize < neededSize) {
-        if (newSize > SIZE_MAX - AVIF_STREAM_BUFFER_INCREMENT) {
-            return AVIF_RESULT_OUT_OF_MEMORY;
-        }
-        newSize += AVIF_STREAM_BUFFER_INCREMENT;
-    }
+    // Make newSize a multiple of AVIF_STREAM_BUFFER_INCREMENT.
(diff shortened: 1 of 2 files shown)


src/stream.c
```diff
@@ -340,6 +340,9 @@ static avifResult makeRoom(avifRWStream * stream, size_t size)
     size_t neededSize = stream->offset + size;
     size_t newSize = stream->raw->size;
     while (newSize < neededSize) {
+        if (newSize > SIZE_MAX - AVIF_STREAM_BUFFER_INCREMENT) {
+            return AVIF_RESULT_OUT_OF_MEMORY;
+        }
```
wantehchang, line 345: If `size_t` is 64 bits, the test for this new overflow check takes a very long time because `AVIF_STREAM_BUFFER_INCREMENT` is `(1024 * 1024)`.

tests/gtest/avifstreamtest.cc
```diff
@@ -202,6 +202,22 @@ TEST(StreamTest, WriteBitsLimit) {
...
+// Test the overflow checks in the makeRoom() function in src/stream.c.
+TEST(StreamTest, OverflowChecksInMakeRoom) {
+  testutil::AvifRwData rw_data;
+  avifRWStream rw_stream;
+  avifRWStreamStart(&rw_stream, &rw_data);
+  const char ten_bytes[10] = {0};
+  EXPECT_EQ(avifRWStreamWrite(&rw_stream, ten_bytes, 10), AVIF_RESULT_OK);
+  EXPECT_EQ(avifRWStreamWrite(&rw_stream, ten_bytes, SIZE_MAX - 9),
+            AVIF_RESULT_OUT_OF_MEMORY);
+  // The next test takes a very long time if size_t is 64 bits.
```
y-guyon, line 214: Could you clarify why `SIZE_MAX - 10` takes a long time but a larger size `SIZE_MAX - 9` does not?

src/stream.c
y-guyon: We could replace this loop with a formula:
- It would be slower for the general case but that should be negligible compared to calling the codec anyway,
- It would speed the test up.

Formula could be something like

```c
    AVIF_CHECKERR(size <= SIZE_MAX - stream->offset, AVIF_RESULT_OUT_OF_MEMORY);
    if (stream->raw->size >= stream->offset + size) return AVIF_RESULT_OK; // big enough already
    size_t neededExtra = size == AVIF_STREAM_BUFFER_INCREMENT ? 0 : AVIF_STREAM_BUFFER_INCREMENT - size % AVIF_STREAM_BUFFER_INCREMENT;
    AVIF_CHECKERR(stream->offset + size <= SIZE_MAX - neededExtra, AVIF_RESULT_OUT_OF_MEMORY);
    return avifRWDataRealloc(stream->raw, stream->offset + size + neededExtra);
```

y-guyon (approved):

y-guyon (approved):

PR #2784, opened 2025-05-01
Add an extra cache key for ci-windows

vrabaud:
Otherwise the two generators use the same cache

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -46,6 +46,7 @@ jobs:
           codec-aom: "LOCAL"
           codec-dav1d: "LOCAL"
           codec-rav1e: "LOCAL"
+          extra-cache-key: ${{ matrix.generator }}
 
       - name: Build libyuv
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-windows.yml
```diff
@@ -46,6 +46,7 @@ jobs:
           codec-aom: "LOCAL"
           codec-dav1d: "LOCAL"
           codec-rav1e: "LOCAL"
+          extra-cache-key: ${{ matrix.generator }}
```
wantehchang, line 49: The generator name "Visual Studio 17 2022" contains spaces. Is this okay?

vrabaud: No, it is safe:
```
Cache hit for: ext-Windows-X64-build-windows-Visual Studio 17 2022-cargo-1.86.0-3b65f88c2840ea41485837909a3894f4d2fec53d331e744c2ff8f78309be7abc-meson-1.8.0-13cdca5ece13259fe13b3545adcaf89e1c272c9d52977c36bd20987c52c7a50d
```

PR #2684, opened 2025-03-13
Do not go through RGB when reading a gray PNG/JPG

vrabaud:
This is a follow-up on https://github.com/AOMediaCodec/libavif/issues/2639 where I realized we are not treating grayscale images properly

PR #2790, opened 2025-05-08
Fix avifImageRGBToYUV for high bit depth gray

wantehchang:
If the output YUV format has the U and V planes, fill them with the half value for the bit depth as uint16_t samples.

See
https://github.com/AOMediaCodec/libavif/pull/2684#pullrequestreview-2816181370.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -16,9 +16,8 @@
 
 static void * avifMemset16(void * dest, int val, size_t count)
 {
-    size_t i;
     uint16_t * dest16 = (uint16_t *)dest;
-    for (i = 0; i < count; i++)
+    for (size_t i = 0; i < count; i++)
         *dest16++ = (uint16_t)val;
     return dest;
 }
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -14,6 +14,15 @@
 #include <pthread.h>
 #endif
 
+static void * avifMemset16(void * dest, int val, size_t count)
+{
+    size_t i;
+    uint16_t * dest16 = (uint16_t *)dest;
+    for (i = 0; i < count; i++)
+        *dest16++ = (uint16_t)val;
+    return dest;
+}
+
 struct YUVBlock
 {
     float y;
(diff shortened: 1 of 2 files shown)


src/reformat.c
```diff
@@ -506,17 +515,28 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
...
         const uint32_t shiftedH = (uint32_t)(((uint64_t)image->height + info.chromaShiftY) >> info.chromaShiftY);
+        const int half = 1 << (image->depth - 1);
         if (image->yuvPlanes[AVIF_CHAN_U]) {
+            uint8_t * uPlane = image->yuvPlanes[AVIF_CHAN_U];
             const uint32_t uRowBytes = image->yuvRowBytes[AVIF_CHAN_U];
-            memset(image->yuvPlanes[AVIF_CHAN_U], 128, shiftedH * uRowBytes);
+            if (state.yuv.channelBytes > 1) {
+                avifMemset16(uPlane, half, shiftedH * uRowBytes / 2);
+            } else {
+                memset(uPlane, half, shiftedH * uRowBytes);
```
wantehchang, line 529: If the row bytes (stride) is greater than the width, this memset() call is only safe if we have padding after the last row. This may not be true in general. So it would be better to replace the memset() call with a for loop that writes width pixels at a time.

src/reformat.c
```diff
@@ -14,6 +14,15 @@
 #include <pthread.h>
 #endif
 
+static void * avifMemset16(void * dest, int val, size_t count)
+{
+    size_t i;
```
jzern, line 19: C99 is OK here, you can scope this to the loop.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -411,6 +411,59 @@ TEST(RGBToYUVTest, AllMatrixCoefficients) {
   }
 }
 
+TEST(RGBToYUVTest, 8BitGrayToYUV420) {
+  // 2x2 8-bit image
+  const uint8_t gray[4] = {4, 3, 2, 1};
```
jzern, line 416: All of the const arrays could be `static`

jzern (approved):

src/reformat.c
wantehchang: Done.

tests/gtest/avifrgbtoyuvtest.cc
wantehchang: Done.

PR #2794, opened 2025-05-09
Add avifgainmaputil to linux,mac release artifacts

wantehchang:


.github/workflows/ci-linux-artifacts.yml
```diff
@@ -46,6 +46,7 @@ jobs:
           path: |
             avifenc
             avifdec
+            avifgainmaputil
```
wantehchang, line 49: In ci-windows-artifacts.yml, the `path` field is specifed as:

```
          path: "*.exe"
```

So avifgainmaputil.exe is included in the artifacts.

jzern (approved):

Issue #2787, opened 2025-05-05
Sample transforms: depth reported after avifDecoderParse() differs from final depth


maryla-uc:
avifDecoderParse() reports the base image's depth, which differs from the final decoded image. This is misleading and I think the final depth should be reported from the beginning.

This also affects avifdec which prints the wrong depth.

See example file: [weld_sato_8plus8bit.avif.zip](https://github.com/user-attachments/files/20033651/weld_sato_8plus8bit.avif.zip)

PR #2797, opened 2025-05-12
Fix avifDecoder image depth in avifDecoderParse()

y-guyon:
Fixes https://github.com/AOMediaCodec/libavif/issues/2787.


maryla-uc (approved):
thanks!

PR #2806, opened 2025-05-19
Fix Sample Transform pow operator

y-guyon:
Add test.

### DIFF
--- a/src/sampletransform.c
+++ b/src/sampletransform.c
@@ -244,24 +244,24 @@ static int32_t avifSampleTransformOperation32bTwoOperands(int32_t leftOperand, i
             if (leftOperand == 0 || leftOperand == 1) {
                 return leftOperand;
             }
-            const uint32_t exponent = rightOperand > 0 ? (uint32_t)rightOperand : (uint32_t) - (int64_t)rightOperand;
-            if (exponent == 0) {
+            if (rightOperand == 0) {
                 return 1;
             }
-            if (exponent == 1) {
+            if (rightOperand == 1) {
                 return leftOperand;
             }
             if (leftOperand == -1) {
-                return (exponent % 2 == 0) ? 1 : -1;
+                return (rightOperand % 2 == 0) ? 1 : -1;
             }
-
+            if (rightOperand < 0) {
+                // L^R is in ]0:1[ here, so truncating it always gives 0.
+                return 0;
+            }
             int64_t result = leftOperand;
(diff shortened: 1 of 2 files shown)


y-guyon:
See https://aomediacodec.github.io/av1-avif/#sample-transform-semantics

vrabaud (approved):
No need to optimize the for loop as we get out of bounds in 32 iters at most.

src/sampletransform.c
```diff
@@ -244,24 +244,24 @@ static int32_t avifSampleTransformOperation32bTwoOperands(int32_t leftOperand, i
...
+                return (rightOperand % 2 == 0) ? 1 : -1;
+            }
+            if (rightOperand < 0) {
+                // L^R is in ]0:1[ here, so truncating it always gives 0.
+                return 0;
             }
-
             int64_t result = leftOperand;
-            for (uint32_t i = 1; i < exponent; ++i) {
+            for (int32_t i = 1; i < rightOperand; ++i) {
```
wantehchang, line 261: Optional: A more efficient algorithm for the exponentiation is repeated squaring. However, since we will overflow `int32_t` after raising to the 31st or 32nd power (2^31 > INT32_MAX, (-2)^31 == INT32_MIN exactly but (-2)^32 > INT32_MAX), the repeated squaring operation can only reduce the number of multiplications from 32 to around 2*log2(32)=10. So it may not be worth it because the code will be more complicated.

wantehchang (commented):
LGTM. Thanks!

src/sampletransform.c
```diff
@@ -244,24 +244,24 @@ static int32_t avifSampleTransformOperation32bTwoOperands(int32_t leftOperand, i
...
                 return 1;
             }
-            if (exponent == 1) {
+            if (rightOperand == 1) {
                 return leftOperand;
             }
             if (leftOperand == -1) {
-                return (exponent % 2 == 0) ? 1 : -1;
+                return (rightOperand % 2 == 0) ? 1 : -1;
+            }
```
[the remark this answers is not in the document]
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2808

src/sampletransform.c
y-guyon: I agree, let's wait for performance issues before looking into that.

wantehchang: I just saw Vincent's comment about the same issue.

PR #2811, opened 2025-05-29
Update LocalAom.cmake to AVM research-v10.0.0

wantehchang:


tests/gtest/avifavmtest.cc
```diff
@@ -44,7 +44,7 @@ TEST_P(AvmTest, EncodeDecode) {
             AVIF_RESULT_OK);
 
   // Verify that the input and decoded images are close.
-  EXPECT_GT(testutil::GetPsnr(*image, *decoded), 39.0);
+  EXPECT_GT(testutil::GetPsnr(*image, *decoded), 38.1);
```
wantehchang, line 47: Fixes these two test failures:

```
[ RUN      ] HighBitDepthAndEvenDimensions/AvmTest.EncodeDecode/1
/usr/local/google/home/wtc/libavif.3/libavif/tests/gtest/avifavmtest.cc:47: Failure
Expected: (testutil::GetPsnr(*image, *decoded)) > (39.0), actual: 38.143876637567573 vs 39

[  FAILED  ] HighBitDepthAndEvenDimensions/AvmTest.EncodeDecode/1, where GetParam() = (2, 34, 10, 3, true) (702 ms)
[ RUN      ] HighBitDepthAndEvenDimensions/AvmTest.EncodeDecode/2
/usr/local/google/home/wtc/libavif.3/libavif/tests/gtest/avifavmtest.cc:47: Failure
Expected: (testutil::GetPsnr(*image, *decoded)) > (39.0), actual: 38.687015453905971 vs 39

[  FAILED  ] HighBitDepthAndEvenDimensions/AvmTest.EncodeDecode/2, where GetParam() = (2, 34, 10, 1, true) (737 ms)
```

cmake/Modules/LocalAom.cmake
```diff
@@ -1,5 +1,5 @@
 set(AVIF_AOM_GIT_TAG v3.12.1)
-set(AVIF_AVM_GIT_TAG research-v9.0.0)
+set(AVIF_AVM_GIT_TAG a1aa30d1911b0815b97bb544f63bec317e903be0)
```
wantehchang, line 2: research-v10.0.0-rc1 has a build system bug (https://gitlab.com/AOMediaCodec/avm/-/issues/712), so we need to check out a newer git tag with the bug fix.

tests/gtest/avifavmtest.cc
wantehchang: I moved this change to https://github.com/AOMediaCodec/libavif/pull/2813.

cmake/Modules/LocalAom.cmake
```diff
@@ -1,5 +1,5 @@
 set(AVIF_AOM_GIT_TAG v3.12.1)
-set(AVIF_AVM_GIT_TAG research-v9.0.0)
+set(AVIF_AVM_GIT_TAG 95163cd4a3c3a4714985e7f71ff021ffe096afd9)
```
wantehchang, line 2: There are test failures that prevent us from going further than this commit: https://gitlab.com/AOMediaCodec/avm/-/issues/717

wantehchang: I worked around https://gitlab.com/AOMediaCodec/avm/-/issues/717 by increasing the image width from 2 to 5 (the smallest width that allows the tests to pass).

tests/gtest/avifavmtest.cc
```diff
@@ -75,8 +75,10 @@ INSTANTIATE_TEST_SUITE_P(Tiny, AvmTest,
                                  Values(AVIF_PIXEL_FORMAT_YUV444),
                                  /*alpha=*/Values(false)));
 
+// width was originally 2. Increased it to 5 to work around
+// https://gitlab.com/AOMediaCodec/avm/-/issues/717.
```
y-guyon: I do not think necessary to keep that comment. A few pixels are not representative of the performance of a codec. The test image is very small to make the test fast, but the content does not matter much, and any PSNR above 20 or 30 is fine in my opinion. Thanks for investigating.

y-guyon (approved):

tests/gtest/avifavmtest.cc
wantehchang: It is the original width of 2 that detected the [AVM issue 717] (https://gitlab.com/AOMediaCodec/avm/-/issues/717). Until we create a minimal test case for that AVM bug, the original width in this test is the only way to reproduce the bug. This comment tells people interested in fixing that bug (or verifying its fix) how to reproduce the bug.

I can edit the comment to make that purpose clear.

y-guyon: My take is that how do we know it is an actual AVM bug? I would assume the PSNR drop on 68 pixels is only the byproduct of a tool change, such as the mentioned minimum block size increase or similar, and not the proof of incorrect or suboptimal AVM behavior. Unless CTC results on "reasonably-sized" images present the same quality drop, I wonder if spending more time on these 68 pixels is worth it.

With that said, it makes sense to keep this comment until the cited GitLab issue is investigated or resolved.

wantehchang: If we decrease `width` from 5 to 4 in the `HighBitDepthAndEvenDimensions` tests, the PSNR drops from 40.0 to 28.4 ~ 30.5 between the ["last known good commit"](https://gitlab.com/AOMediaCodec/avm/-/commit/c2f7e2f23f14b56c6a91dba364911603a645940d) and `research-v10.0.0`. I cannot dismiss a PSNR drop of this magnitude, even though the frame size is small (4x34). That's why I reported this issue to the AVM team.

tests/gtest/avifavmtest.cc
```diff
@@ -75,8 +75,12 @@ INSTANTIATE_TEST_SUITE_P(Tiny, AvmTest,
                                  Values(AVIF_PIXEL_FORMAT_YUV444),
                                  /*alpha=*/Values(false)));
 
+// The width in this test was originally 2 and that detected
+// https://gitlab.com/AOMediaCodec/avm/-/issues/717. 5 is the smallest width
+// that allows this test to pass. To reproduce the bug, decrease width to 4, 3,
+// or 2.
 INSTANTIATE_TEST_SUITE_P(HighBitDepthAndEvenDimensions, AvmTest,
-                         Combine(/*width=*/Values(2), /*height=*/Values(34),
+                         Combine(/*width=*/Values(5), /*height=*/Values(34),
```
wantehchang: I moved this change to https://github.com/AOMediaCodec/libavif/pull/2822.

wantehchang:
I will merge this pull request after AVM research-v10.0.0 has been imported into Google's internal repository.

tests/gtest/avifavmminitest.cc
```diff
@@ -45,7 +45,7 @@ TEST_P(AvmMiniTest, EncodeDecode) {
             AVIF_RESULT_OK);
 
   // Verify that the input and decoded images are close.
-  EXPECT_GT(testutil::GetPsnr(*image, *decoded), 40.0);
+  EXPECT_GT(testutil::GetPsnr(*image, *decoded), 37.0);
```
wantehchang, line 48: I found that the PSNR threshold 40.0 is too high for Arm64 and for x86_64 without AVX2. So I lowered it to 37.0.

This indicates a bug in AVM because SIMD optimizations should not change the encoded output.

Issue #2812, opened 2025-05-29
ext/avm.cmd is broken


wantehchang:
Although ext/avm.cmd builds AVM successfully, the libavif build fails when linking the libavif.so shared library:

[1/51] Linking CXX shared library libavif.so.16.2.1
FAILED: libavif.so.16.2.1 
: && /usr/bin/c++ -fPIC -g  -Wl,--dependency-file=CMakeFiles/avif.dir/link.d -shared -Wl,-soname,libavif.so.16 -o libavif.so.16.2.1 CMakeFiles/avif_obj.dir/src/alpha.c.o CMakeFiles/avif_obj.dir/src/avif.c.o CMakeFiles/avif_obj.dir/src/colr.c.o CMakeFiles/avif_obj.dir/src/colrconvert.c.o CMakeFiles/avif_obj.dir/src/diag.c.o CMakeFiles/avif_obj.dir/src/exif.c.o CMakeFiles/avif_obj.dir/src/gainmap.c.o CMakeFiles/avif_obj.dir/src/io.c.o CMakeFiles/avif_obj.dir/src/mem.c.o CMakeFiles/avif_obj.dir/src/obu.c.o CMakeFiles/avif_obj.dir/src/properties.c.o CMakeFiles/avif_obj.dir/src/rawdata.c.o CMakeFiles/avif_obj.dir/src/read.c.o CMakeFiles/avif_obj.dir/src/reformat.c.o CMakeFiles/avif_obj.dir/src/reformat_libsharpyuv.c.o CMakeFiles/avif_obj.dir/src/reformat_libyuv.c.o CMakeFiles/avif_obj.dir/src/scale.c.o CMakeFiles/avif_obj.dir/src/stream.c.o CMakeFiles/avif_obj.dir/src/utils.c.o CMakeFiles/avif_obj.dir/src/write.c.o CMakeFiles/avif_obj.dir/src/codec_avm.c.o  _deps/libyuv-build/libyuv.a  -lm  /usr/local/google/home/wtc/libavif.3/libavif/ext/avm/build.libavif/libaom.a  -l_aom_dep_lib_avm-NOTFOUND  -lm  /usr/lib/x86_64-linux-gnu/libpthread.a && :
/usr/bin/ld: cannot find -l_aom_dep_lib_avm-NOTFOUND: No such file or directory
collect2: error: ld returned 1 exit status

wantehchang:
Steps to reproduce:

1. cd ext
[...]

PR #2820, opened 2025-06-02
Remove ext/avm.cmd

y-guyon:
Fixes https://github.com/AOMediaCodec/libavif/issues/2812


CHANGELOG.md
```diff
@@ -13,7 +13,13 @@ The changes are relative to the previous release, unless the baseline is specifi
...
   after avifDecoderNextImage() when AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM is
   enabled and when the file to decode contains a 'sato' derived image item.
-* Update rav1e.cmd/LocalRav1e.cmake: cargo-c v0.10.13, corrosion v0.5.2, rav1e v0.8.0
+* Update rav1e.cmd/LocalRav1e.cmake: cargo-c v0.10.13, corrosion v0.5.2,
+  rav1e v0.8.0
+
+### Removed since 1.3.0
+
+* ext/avm.cmd was removed. Use the -DAVIF_CODEC_AVM=LOCAL CMake build option
+  instead.
```
wantehchang: The second sentence, "Use the -DAVIF_CODEC_AVM=LOCAL CMake build option instead.", is not accurate. `-DAVIF_CODEC_AVM=LOCAL` means "use a local libavm build", and ext/avm.cmd (if it works) is one of two ways to use a local libavm build.

If `-DAVIF_CODEC_AVM=LOCAL` is specified, we first see if there is a libavm build in the ext/avm directory (which we assume was created by running ext/avm.cmd). If Yes, we are done. If No, then we run cmake/Modules/LocalAvm.cmake to fetch and build libavm.

PR #2822, opened 2025-06-05
avifavmtest.cc: Increase width to 5 in HBD test

wantehchang:
In the HighBitDepthAndEvenDimensions test, increase the width to 5 to avoid https://gitlab.com/AOMediaCodec/avm/-/issues/717.

### DIFF
--- a/tests/gtest/avifavmtest.cc
+++ b/tests/gtest/avifavmtest.cc
@@ -75,8 +75,12 @@ INSTANTIATE_TEST_SUITE_P(Tiny, AvmTest,
                                  Values(AVIF_PIXEL_FORMAT_YUV444),
                                  /*alpha=*/Values(false)));
 
+// The width in this test was originally 2 and that detected
+// https://gitlab.com/AOMediaCodec/avm/-/issues/717. 5 is the smallest width
+// that allows this test to pass. To reproduce the bug, decrease width to 4, 3,
+// or 2.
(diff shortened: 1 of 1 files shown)


tests/gtest/avifavmtest.cc
```diff
@@ -75,8 +75,12 @@ INSTANTIATE_TEST_SUITE_P(Tiny, AvmTest,
                                  Values(AVIF_PIXEL_FORMAT_YUV444),
                                  /*alpha=*/Values(false)));
 
+// The width in this test was originally 2 and that detected
+// https://gitlab.com/AOMediaCodec/avm/-/issues/717. 5 is the smallest width
+// that allows this test to pass. To reproduce the bug, decrease width to 4, 3,
+// or 2.
```
wantehchang, line 81: We can delete this comment after https://gitlab.com/AOMediaCodec/avm/-/issues/717 is resolved. At that time we can leave the width at 5 or change it to a more natural number such as 4.

jzern: Do you want to make this a `TODO()` then?

jzern (approved):

tests/gtest/avifavmtest.cc
wantehchang: No. I wanted to avoid adding a TODO comment. I wrote this comment in a way such that it doesn't need to be deleted.

PR #2823, opened 2025-06-12
avifenc.c Use better formula in quantizerToQuality

wantehchang:
The original formula was apparently derived by inverting the quality-to-quantizer conversion formula in src/write.c. That derivation is not ideal because it included the `+ 50` rounding term. It is better to omit the `+ 50` rounding term, invert the formula, and then add a `+ 31` rounding term.

The new formula is already used in avifenc.c.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -1424,7 +1424,8 @@ static void avifInputAdd(avifInput * input, const char * filePath, uint64_t dura
 static int quantizerToQuality(int minQuantizer, int maxQuantizer)
 {
     const int quantizer = (minQuantizer + maxQuantizer) / 2;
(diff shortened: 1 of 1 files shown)


apps/avifenc.c
```diff
@@ -1424,7 +1424,8 @@ static void avifInputAdd(avifInput * input, const char * filePath, uint64_t dura
 static int quantizerToQuality(int minQuantizer, int maxQuantizer)
 {
     const int quantizer = (minQuantizer + maxQuantizer) / 2;
-    return (int)(100 - (100 * quantizer - 50) / 63.0);
```
wantehchang, line 1427: Here is the formula in the `avifQualityToQuantizer()` function in src/write.c:

```
```

maryla-uc:
LGTM thanks!

PR #2824, opened 2025-06-12
Set CMAKE_CXX_STANDARD to 17

wantehchang:
The Google Foundational C++ Support table requires C++17 as of 2024-12-17.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Set avifDecoder::image->depth to the same value after avifDecoderParse() as
   after avifDecoderNextImage() when AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM is
   enabled and when the file to decode contains a 'sato' derived image item.
+* Update googletest.cmd/LocalGTest.cmake: v1.17.0
 * Update rav1e.cmd/LocalRav1e.cmake: cargo-c v0.10.13, corrosion v0.5.2,
   rav1e v0.8.0
 
--- a/cmake/Modules/LocalGTest.cmake
+++ b/cmake/Modules/LocalGTest.cmake
@@ -1,4 +1,4 @@
-set(AVIF_GTEST_GIT_TAG v1.16.0)
+set(AVIF_GTEST_GIT_TAG v1.17.0)
 
 set(GTest_FOUND ON CACHE BOOL "")
 set(GTEST_INCLUDE_DIRS ${AVIF_SOURCE_DIR}/ext/googletest/googletest/include)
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -570,13 +570,7 @@ endif()
...
 if(AVIF_LIB_USE_CXX OR AVIF_BUILD_APPS OR (AVIF_BUILD_TESTS AND (AVIF_FUZZTEST OR AVIF_GTEST)))
     enable_language(CXX)
-    if(AVIF_ENABLE_NODISCARD OR AVIF_CODEC_AVM_ENABLED)
-        # [[nodiscard]] and tensorflow-lite require C++17.
-        set(CMAKE_CXX_STANDARD 17)
-        set(CMAKE_CXX_STANDARD_REQUIRED ON)
-    else()
-        set(CMAKE_CXX_STANDARD 14)
-    endif()
+    set(CMAKE_CXX_STANDARD 17)
```
wantehchang, line 573: It doesn't seem necessary to keep `set(CMAKE_CXX_STANDARD_REQUIRED ON)`. Should I keep it?

jzern: It's probably better to fail early than have the compile fail with an obscure error if it has fallen back to an earlier standard.

jzern (approved):

CMakeLists.txt
wantehchang: Done.

vrabaud (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Set avifDecoder::image->depth to the same value after avifDecoderParse() as
   after avifDecoderNextImage() when AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM is
   enabled and when the file to decode contains a 'sato' derived image item.
+* Update googletest.cmd/LocalGTest.cmake: v1.17.0
```
wantehchang, line 16: James, Vincent: This is needed because googletest v1.16.0 has the following GCC 15.1.0 warning:

```
In file included from D:/a/libavif/libavif/build/_deps/googletest-src/googletest/include/gtest/internal/gtest-port.h:295,
                 from D:/a/libavif/libavif/build/_deps/googletest-src/googletest/include/gtest/gtest-message.h:57,
                 from D:/a/libavif/libavif/build/_deps/googletest-src/googletest/include/gtest/gtest-assertion-result.h:46,
                 from D:/a/libavif/libavif/build/_deps/googletest-src/googletest/include/gtest/gtest.h:63,
                 from D:/a/libavif/libavif/build/_deps/googletest-src/googletest/src/gtest_main.cc:32:
D:/a/_temp/msys64/ucrt64/include/c++/15.1.0/ciso646:46:4: warning: #warning "<ciso646> is deprecated in C++17, use <version> to detect implementation-specific macros" [-Wcpp]
   46 | #  warning "<ciso646> is deprecated in C++17, use <version> to detect implementation-specific macros"
      |    ^~~~~~~
```

The googletest in libaom doesn't include `<ciso646>`, so we don't need to update it to v1.17.0 even though we now compile libaom with `--std=c++17`.

jzern (approved):

Issue #2827, opened 2025-06-15
1.3.0: Possible wrong results (grayscale PNG -> 12-bit AVIF)


WerctFourth:
Hello.
Avifenc 1.3.0 possibly produces wrong results while converting a 8- or 16-bit PNG image to a 12-bit AVIF.
1.2.1 produces correct results.
Commandline used: `-d 12 -q 54`
Platform: Windows 10 64-bit
Samples (Converted to PNG using avifdec for presentation):

Source image sample:

![Image](https://github.com/user-attachments/assets/d30d1939-9728-46d6-a288-60da4e638e3d)

1.3.0: 8-bit grayscale -> 12-bit AVIF:

![Image](https://github.com/user-attachments/assets/adc8dc8a-71d0-479f-bd54-7691d2381a6e)

1.3.0: 16-bit grayscale -> 12-bit AVIF:

![Image](https://github.com/user-attachments/assets/d7e37852-c09f-4d74-8b05-5d5a5fecb693)

1.2.1: 8-bit grayscale -> 12-bit AVIF:

![Image](https://github.com/user-attachments/assets/a3f7b536-2558-4cce-bfc4-97c97999def2)

All sample files are available in this archive:
https://1drv.ms/u/s!AkgUEzQmipA9gc0PYpqS7q7J7X6Epw?e=emLQOU

wantehchang:
Vincent: This may be caused by https://github.com/AOMediaCodec/libavif/pull/2684. Could you take a look? Thanks.

wantehchang:
WerctFourth: Thanks for the bug report. I can reproduce the bug with sample16bitGray.png and sample8bitGray.png. sample16bitRGB.png and sample8bitRGB.png can be converted to AVIF correctly.

I did a bisection. It says:

```
5450a8cddeeb631fd525751adc14333b5b20e6d6 is the first bad commit
commit 5450a8cddeeb631fd525751adc14333b5b20e6d6 (HEAD)
Author: Vincent Rabaud <vrabaud@google.com>
Date:   Tue Mar 18 11:01:24 2025 +0100

    Do not go through RGB when reading/writing a gray PNG/JPG (#2684)

 CHANGELOG.md                     |   4 +
 apps/shared/avifjpeg.c           |   5 +-
 apps/shared/avifpng.c            |  31 +++--
 include/avif/avif.h              |   4 +
 include/avif/internal.h          |  13 +-
 src/avif.c                       |  14 +-
[...]

PR #2828, opened 2025-06-16
Fix grayscale conversion when doing bit lifting.

vrabaud:
Fixes https://github.com/AOMediaCodec/libavif/issues/2827


src/reformat.c
```diff
@@ -505,7 +505,7 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                         }
                     }
                 }
-                g = avifRoundf(AVIF_CLAMP(g, 0.0f, state.yuv.maxChannel));
+                g = avifRoundf(AVIF_CLAMP(g * state.yuv.rangeY, 0.0f, state.yuv.maxChannel));
```
wantehchang: We need to add `state.yuv.biasY`:

```
                g = avifRoundf(AVIF_CLAMP(g * state.yuv.rangeY + state.yuv.biasY, 0.0f, state.yuv.maxChannel));
```

I also suggest we first round and then clamp. (Search for "Round(" in H.273.) We can just call `avifYUVColorSpaceInfoYToUNorm(&state.yuv, g)` here. Note that `avifYUVColorSpaceInfoYToUNorm()` returns `int`, so we can't store its return value in `g`. I think we can declare an `int` variable `y`.

Here is a test that demonstrates the need to add `state.yuv.biasY`:

```
TEST(RGBToYUVTest, 8BitGrayRoundTripWithLiftLimitedRange) {
  // 2x2 12-bit temporary image.
  ImagePtr image(avifImageCreate(2, 2, 12, AVIF_PIXEL_FORMAT_YUV400));
  ASSERT_NE(image, nullptr);
  image->yuvRange = AVIF_RANGE_LIMITED;
  // 2x2 8-bit original image.
  static constexpr uint8_t gray[4] = {5, 3, 2, 1};
  avifRGBImage rgb;
  avifRGBImageSetDefaults(&rgb, image.get());
  rgb.format = AVIF_RGB_FORMAT_GRAY;
  rgb.avoidLibYUV = AVIF_TRUE;
  rgb.depth = 8;
  rgb.pixels = const_cast<uint8_t*>(gray);
  rgb.rowBytes = 2 * sizeof(uint8_t);
  // Convert to 12 bits.
  ASSERT_EQ(avifImageRGBToYUV(image.get(), &rgb), AVIF_RESULT_OK);
  // Convert back to 8 bits.
  avifRGBImage rgb_final;
  avifRGBImageSetDefaults(&rgb_final, image.get());
  rgb_final.format = AVIF_RGB_FORMAT_GRAY;
  rgb_final.avoidLibYUV = AVIF_TRUE;
  rgb_final.depth = 8;
  ASSERT_EQ(avifRGBImageAllocatePixels(&rgb_final), AVIF_RESULT_OK);
  ASSERT_EQ(avifImageYUVToRGB(image.get(), &rgb_final), AVIF_RESULT_OK);
  // Compare to the original 8-bit image.
  const uint8_t* gray_plane = rgb_final.pixels;
  EXPECT_EQ(gray_plane[0], gray[0]);
  EXPECT_EQ(gray_plane[1], gray[1]);
  EXPECT_EQ(gray_plane[2], gray[2]);
  EXPECT_EQ(gray_plane[3], gray[3]);
  avifRGBImageFreePixels(&rgb_final);
}
```

In the `gray` array I replaced the value 4 with 5, because 4 does not roundtrip when going through 8-bit limited range. (4 is converted back to 3.) Although it does roundtrip when going through 12-bit limited range, it is better to avoid it.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -464,6 +464,38 @@ TEST(RGBToYUVTest, HighBitDepthGrayToYUV420) {
   }
 }
 
+TEST(RGBToYUVTest, 8BitGrayRoundTripWithLift) {
+  // 2x2 12-bit temporary image.
+  static constexpr uint8_t gray[4] = {4, 3, 2, 1};
```
wantehchang: Nit: Move this line after the comment at line 472, `2x2 8-bit original image.` Alternatively move this line before the comment at line 468, `2x2 12-bit temporary image.`

At the current location, this is confusing because the comment at line 468 says `12-bit` but this array is 8 bits.

src/reformat.c
vrabaud: Thx for the example, I changed the gray values accordingly.

tests/gtest/avifrgbtoyuvtest.cc
vrabaud: Done

src/reformat.c
```diff
@@ -505,7 +505,7 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                         }
                     }
                 }
-                g = avifRoundf(AVIF_CLAMP(g * state.yuv.rangeY, 0.0f, state.yuv.maxChannel));
+                g = avifYUVColorSpaceInfoYToUNorm(&state.yuv, g);
```
wantehchang: Nit: `avifYUVColorSpaceInfoYToUNorm()` returns `int`. I suggest storing the return value in an `int` variable `y` to avoid the unnecessary conversion to `float` and back to an integer.

```
                int y = avifYUVColorSpaceInfoYToUNorm(&state.yuv, g);
```

And then use `y` instead of `g` in line 511 and line 513.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -466,10 +466,10 @@ TEST(RGBToYUVTest, HighBitDepthGrayToYUV420) {
 
 TEST(RGBToYUVTest, 8BitGrayRoundTripWithLift) {
   // 2x2 12-bit temporary image.
-  static constexpr uint8_t gray[4] = {4, 3, 2, 1};
   ImagePtr image(avifImageCreate(2, 2, 12, AVIF_PIXEL_FORMAT_YUV400));
   ASSERT_NE(image, nullptr);
   // 2x2 8-bit original image.
+  static constexpr uint8_t gray[4] = {5, 3, 2, 1};
```
wantehchang: Could you also add a test for limited range? The easiest way is to add the test I provided.

wantehchang (approved):

tests/gtest/avifrgbtoyuvtest.cc
vrabaud: done

src/reformat.c
vrabaud: done

src/reformat.c
```diff
@@ -505,12 +505,12 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                         }
                     }
                 }
-                g = avifYUVColorSpaceInfoYToUNorm(&state.yuv, g);
+                int g_int = avifYUVColorSpaceInfoYToUNorm(&state.yuv, g);
```
wantehchang: Nit: g_int => gInt

wantehchang (approved):

src/reformat.c
vrabaud: ah right, thx, I'll never get used to that convention ....

wantehchang (approved):

src/reformat.c
```diff
@@ -505,12 +505,12 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
...
                 }
-                g = avifRoundf(AVIF_CLAMP(g, 0.0f, state.yuv.maxChannel));
+                int gInt = avifYUVColorSpaceInfoYToUNorm(&state.yuv, g);
                 if (state.yuv.channelBytes > 1) {
                     uint16_t * pY = (uint16_t *)&yPlane[(i * 2) + j * yRowBytes];
-                    *pY = (uint16_t)g;
+                    *pY = (uint16_t)gInt;
                 } else {
-                    yPlane[i + (j * yRowBytes)] = (uint8_t)g;
+                    yPlane[i + (j * yRowBytes)] = (uint8_t)gInt;
```
wantehchang, line 513: Vincent: Please add a changelog entry. Thanks!

vrabaud:
done in https://github.com/AOMediaCodec/libavif/pull/2829

PR #2830, opened 2025-06-18
Use AOM_TUNE_IQ by default

y-guyon:
Based on https://github.com/AOMediaCodec/libavif/pull/2599.

Note that "(Default: psnr)" was incorrect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Update rav1e.cmd/LocalRav1e.cmake: cargo-c v0.10.13, corrosion v0.5.2,
   rav1e v0.8.0
 * Fix grayscale conversion when changing the bit depth.
+* Use AOM_TUNE_IQ by default when possible.
 
 ### Removed since 1.3.0
 
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -312,7 +312,7 @@ static void syntaxLong(void)
         printf("    enable-chroma-deltaq=B            : Enable delta quantization in chroma planes. 0=disable (default), 1=enable\n");
         printf("    end-usage=MODE                    : Rate control mode, one of 'vbr', 'cbr', 'cq', or 'q'\n");
         printf("    sharpness=S                       : Bias towards block sharpness in rate-distortion optimization of transform coefficients in 0..7. (Default: 0)\n");
-        printf("    tune=METRIC                       : Tune the encoder for distortion metric, one of 'psnr' or 'ssim'. (Default: psnr)\n");
+        printf("    tune=METRIC                       : Tune the encoder for distortion metric, one of 'psnr', 'ssim' or 'iq'. (Default for aom 3.12+ still images: iq, otherwise ssim)\n");
         printf("    film-grain-test=TEST              : Film grain test vectors in 0..16. 0=none (default), 1=test1, 2=test2, ... 16=test16\n");
         printf("    film-grain-table=FILENAME         : Path to file containing film grain parameters\n");
         printf("\n");
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -886,8 +886,15 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         if (!avifProcessAOMOptionsPostInit(codec, alpha)) {
             return AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION;
         }
-        if (!codec->internal->tuningSet) {
-            if (aom_codec_control(&codec->internal->encoder, AOME_SET_TUNING, AOM_TUNE_SSIM) != AOM_CODEC_OK) {
+        if (!lossless && !codec->internal->tuningSet) {
+            aom_tune_metric tuneMetric = AOM_TUNE_SSIM;
+#if defined(AOM_HAVE_TUNE_IQ)
+            // AOM_TUNE_IQ sets --deltaq-mode=6 which can only be used in all intra mode.
+            if (aomUsage == AOM_USAGE_ALL_INTRA) {
+                tuneMetric = AOM_TUNE_IQ;
+            }
+#endif
+            if (aom_codec_control(&codec->internal->encoder, AOME_SET_TUNING, tuneMetric) != AOM_CODEC_OK) {
                 return AVIF_RESULT_UNKNOWN_ERROR;
             }
         }


src/codec_aom.c
```diff
@@ -886,8 +886,15 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
-            if (aom_codec_control(&codec->internal->encoder, AOME_SET_TUNING, AOM_TUNE_SSIM) != AOM_CODEC_OK) {
+        if (!lossless && !codec->internal->tuningSet) {
+            aom_tune_metric tuneMetric = AOM_TUNE_SSIM;
+#if defined(AOM_HAVE_TUNE_IQ)
+            // AOM_TUNE_IQ sets --deltaq-mode=6 which can only be used in all intra mode.
+            if (aomUsage == AOM_USAGE_ALL_INTRA) {
+                tuneMetric = AOM_TUNE_IQ;
+            }
+#endif
+            if (aom_codec_control(&codec->internal->encoder, AOME_SET_TUNING, tuneMetric) != AOM_CODEC_OK) {
```
juliobbv-p: Thanks for opening this PR, Yannis!

Can tune iq be applied to just the YUV planes, and not alpha? The perceptual optimizations are likely harmful for the alpha channel.

y-guyon: I understood "only YUV" as "not RGB", and not "not alpha". But I did not even test for identity MC. Do you know if `AOM_TUNE_IQ` for all matrix coefficients and color spaces (besides identity I guess)?

Good point for alpha.

juliobbv-p: > Do you know if AOM_TUNE_IQ for all matrix coefficients and color spaces (besides identity I guess)?

Good question! So, AOM_TUNE_IQ has been tuned for the YCbCr families of matrix coefficients and color spaces. That said, the tune tweaks do generalize (to various degrees) to all matrix coefficients, color spaces, and transfer characteristics that have one luma and two chroma channels.

I think we should be good turning it on globally on still images, except for the identity MC.

y-guyon: Should `AOM_TUNE_PSNR` or `AOM_TUNE_SSIM` be used for alpha?

juliobbv-p: Probably both PSNR or SSIM are fine for alpha, but anecdotally I've seen PSNR ring a bit less on grayscale images, so PSNR might be the better choice of the two for alpha (if the decoder interprets ringing as incorrect alpha values that make the underlying image too opaque or transparent).

maryla-uc:
WIth this PR, using `-a sharpness=0` does nothing because when you set tune=iq it overrides it.
This also happens if you manually set `-a sharpness=0 -a tune=iq`. It only works as expected with `-a tune=iq -a sharpness=0`.

juliobbv-p:
@maryla-uc good catch! tune iq sets up many optimized defaults (including `--sharpness`), so the codec control call to set tune iq should happen before the `-a` parameters.

y-guyon:
> @maryla-uc good catch! tune iq sets up many optimized defaults (including `--sharpness`), so the codec control call to set tune iq should happen before the `-a` parameters.

Done.

castano:
Any idea what's holding up this PR?

y-guyon:
> Any idea what's holding up this PR?

There are a few consequences of TUNE=IQ to take into consideration when using it by default:
- The quality encoding setting does not lead to the same file size and visual distortion, so the default encoding settings may end in noticeably different output after this change.
- TUNE=IQ is slower to encode. One option may be to increase the default speed to 7 instead of 6, but that could also have side effects on output file size and decoding time.

juliobbv-p:
@castano another thing worth mentioning: tune IQ is already considered production quality (on libaom 3.13.1 and newer). If a project doesn't have to worry about specific tune ssim behavior, then said project can just switch to use tune IQ right away.

For example, [The Guardian](https://www.theguardian.com/us) has been using tune IQ for a few months, and we've received positive feedback from the website admin in the form of overall better image quality and consistency. Tune IQ's improvement in bit allocation by itself can be seen as a "bug fix".

I wouldn't wait for this PR to be merged as a sign of feature "readiness". On the contrary, we're just trying to be super thorough in understanding the inherent differences in bitrate allocation and encode time with tune IQ across images, compared to tune SSIM.

castano:
I think it makes sense to evaluate the consequences of enabling tune IQ before making it the default choice, but it would be helpful to first add support for tune IQ as a user option, and only later make it the default after more users have had the chance to evaluate it and provide feedback.

I have clients interested in using tune IQ, but they won't maintain a custom build of libavif with modifications that they have to merge on every update.

Similarly, tools that use libavif won't offer this option until it's officially supported by libavif, so not having this exposed also delays its upstream adoption.

juliobbv-p:
@castano you can enable tune IQ in avifenc with the following parameter: `-a c:tune=iq`. This is the official way to switch between different tunings, so it's guaranteed to never change.

avifenc in turn passes in the parameter to libavif as a key ("tune")/ value ("iq") pair in the following manner:
```c
static avifBool avifCodecSpecificOptionsAdd(avifCodecSpecificOptions * options, const char * keyValue)
{
    const char * value = strchr(keyValue, '=');
    if (value) {
        // Keep the parts on the left and on the right of the equal sign,
        // but not the equal sign itself.
        const size_t keyLength = value - keyValue;
        return avifCodecSpecificOptionsAddKeyValue(options, keyValue, keyLength, value + 1);
    }
    // Pass in a non-NULL, empty string. Codecs can use the mere existence of a key as a boolean value.
    return avifCodecSpecificOptionsAddKeyValue(options, keyValue, strlen(keyValue), "");
}
```

[`avifenc.c`](https://github.com/AOMediaCodec/libavif/blob/40cb2c1fbff1e14b6c81816b73a364490b12ffb6/apps/avifenc.c) has the necessary code to talk to libavif, including passing in codec-specific options.

castano:
Hmm... I haven't tried avifenc, but I didn't see the iq option in the aomOptionDefs struct, so I assumed that would not work without additional changes:

https://github.com/AOMediaCodec/libavif/blob/40cb2c1fbff1e14b6c81816b73a364490b12ffb6/src/codec_aom.c#L397C39-L397C49

Ah, I see it gets routed to `aom_codec_set_option`, never mind!

wantehchang:
Ignacio: The aomOptionDefs struct is only used with older versions of libaom that don't have the `aom_codec_set_option()` function. In that case, the feature detection macro `HAVE_AOM_CODEC_SET_OPTION` is not defined.

For testing this pull request, please use the latest version of libaom, v3.13.1 if you can. Thanks!

src/codec_aom.c
```diff
@@ -991,14 +993,26 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
-        if (!codec->internal->tuningSet) {
-            if (aom_codec_control(&codec->internal->encoder, AOME_SET_TUNING, AOM_TUNE_SSIM) != AOM_CODEC_OK) {
+        if (!lossless && !avifAOMOptionsSetTuning(codec, alpha)) {
+            aom_tune_metric tuneMetric = AOM_TUNE_SSIM;
+            if (alpha) {
+                // Minimize ringing for alpha.
+                tuneMetric = AOM_TUNE_PSNR;
+            }
+#if defined(AOM_HAVE_TUNE_IQ)
+            // AOM_TUNE_IQ is favored for its low perceptual distortion on luma and chroma samples.
```
juliobbv-p: Because of the new quality-to-quantizer formula for tune IQ, we'll need to resolve the effective tuning mode before computing the quantizer. I think this means this piece of code needs to be moved much closer to the beginning of the encode function, around line 657.

y-guyon: I started doing that in https://github.com/AOMediaCodec/libavif/pull/2936 but I have a question:

What should the behavior be when doing the following?
1. set tune=psnr (whatever as long as different from libavif default)
2. encode the first frame
3. unset tune (`avifEncoderSetCodecSpecificOption("tune", NULL)`)
4. encode the second frame

juliobbv-p: In that case, my mind is telling me that the second frame should be encoded with the tune that libavif determines it's the best. If I'm reading the code, this is the current behavior, right?

y-guyon: > In that case, my mind is telling me that the second frame should be encoded with the tune that libavif determines it's the best.

I think it applies beyond `tune`. For example if I set `sharpness=7` for the first frame, then unset `sharpness` before the second frame, the current behavior of libaom-in-libavif at head is to keep `sharpness=7` because libavif does not "carry the option unsetting over to libaom".  
The `avifEncoderSetCodecSpecificOption()` doc says "passing a NULL value will delete the pending key". This is unclear or misleading. Maybe the word "pending" is key here, meaning it only applies if the option was not yet forwarded to libaom.

I think for now I will keep the current behavior of persistent codec-specific options until overwritten with a new value. Deleting a key does nothing if a frame already took it into account. We just need to make the comment more explicit.

EDIT: I misunderstood `avifEncoderSetCodecSpecificOption()`. I thought the `csOptions` were applied to all next frames, but `csOptions` is actually cleared when `avifEncoderAddImage()` is called. So current behavior is working as intended I guess.

y-guyon: See https://github.com/AOMediaCodec/libavif/pull/2937

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Set tuning before applying the user-provided specific aom codec options.
 * Use AOM_TUNE_PSNR by default when encoding alpha with libaom because
   AOM_TUNE_SSIM causes ringing for alpha.
+* Use AOM_TUNE_IQ by default when encoding still luma and chroma with libaom.
```
wantehchang: See my comment about "luma and chroma" in src/codec_aom.c.

apps/avifenc.c
```diff
@@ -321,7 +321,7 @@ static void syntaxLong(void)
         printf("    end-usage=MODE                    : Rate control mode, one of 'vbr', 'cbr', 'cq', or 'q'\n");
         printf("    sharpness=S                       : Bias towards block sharpness in rate-distortion optimization of transform coefficients in 0..7. (Default: 0)\n");
         printf("    tune=METRIC                       : Tune the encoder for distortion metric, one of 'psnr', 'ssim' or 'iq'.\n");
-        printf("                                        (Default for color: ssim, default for alpha: psnr)\n");
+        printf("                                        (Default for color: still images (aom v3.12+): iq, otherwise: ssim; default for alpha: psnr)\n");
```
wantehchang: I verified that tune=iq first appeared in the [libaom v3.12.0 release](https://groups.google.com/a/aomedia.org/g/av1-discuss/c/QJ-iK2CvmXw).

Optional: Change "aom v3.12+" to "libaom v3.12.0+" or "libaom v3.12.0 or later"

src/codec_aom.c
```diff
@@ -666,15 +666,22 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             libavifDefaultTuneMetric = AOM_TUNE_PSNR;
         } else {
             libavifDefaultTuneMetric = AOM_TUNE_SSIM;
+#if defined(AOM_HAVE_TUNE_IQ) && defined(AOM_USAGE_ALL_INTRA) && defined(ALL_INTRA_HAS_SPEEDS_7_TO_9)
```
wantehchang: Nit: I suggest we simplify this conditional expression based on what we know about libaom releases.

tune=iq was added much later than speeds 7-9 were added to all intra mode. So we can omit the `defined(ALL_INTRA_HAS_SPEEDS_7_TO_9)` test.

Optional: For the same reason we can omit the `defined(AOM_USAGE_ALL_INTRA)` test. But I am fine with keeping this test for clarity.

src/codec_aom.c
```diff
@@ -666,15 +666,22 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+            // AOM_TUNE_IQ is favored for its low perceptual distortion on luma and chroma samples.
```
wantehchang: Nit: "luma and chroma" is not clear. Did you mean YUV as opposed to RGB? Does "luma and chroma" include "luma only", i.e., monochrome?

src/codec_aom.c
```diff
@@ -666,15 +666,22 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+            // AOM_TUNE_IQ sets --deltaq-mode=6 which can only be used in all intra mode.
+            if (image->matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_IDENTITY && (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE)) {
```
wantehchang: IMPORTANT: If the `addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE` test is intended to detect whether all intra mode is used, ideally we should wait until we have set the `aomUsage` variable and then test `aomUsage == AOM_USAGE_ALL_INTRA`.

This requires moving some code around and possibly save some variables such as `aomUsage` and `useTuneIq` in the `codec->internal` struct.

Would you like me to take a stab at this and attach a patch to this pull request?

wantehchang:
Yannis: Sorry about the late review. I missed your review request.

src/codec_aom.c
juliobbv-p: I agree with Wan-Teh, `defined(AOM_HAVE_TUNE_IQ)` is enough here.

I'd actually argue that leaving `defined(AOM_USAGE_ALL_INTRA)` could be confusing, because that condition could be interpreted as detecting the ability for libaom to use tune IQ in all intra mode (as opposed to other modes), which it's not the case.

src/codec_aom.c
juliobbv-p: Depending on how detailed we want to be, this comment could go like this:

```c
// AOM_TUNE_IQ has been tuned for the YCbCr family of color spaces, and is favored for its low perceptual distortion
// AOM_TUNE_IQ partially generalizes to, and benefits from other "YUV-like" spaces (e.g. YCgCo and ICtCp) including monochrome (luma only)
// 
```

Or if we want to be more literal, we could simply phrase it in the negative:

```c
// AOM_TUNE_IQ is favored for its low perceptual distortion on non-RGB color samples
```

src/codec_aom.c
y-guyon: > Would you like me to take a stab at this and attach a patch to this pull request?

Thank you for the proposal. Could you do that in a separate pull request, so that this PR only switches t=IQ on by default?

src/codec_aom.c
y-guyon: Transformed into a kind of C static assert:

```c
#if !defined(AOM_USAGE_ALL_INTRA) || !defined(ALL_INTRA_HAS_SPEEDS_7_TO_9)
#error "AOM_HAVE_TUNE_IQ was introduced after AOM_USAGE_ALL_INTRA and ALL_INTRA_HAS_SPEEDS_7_TO_9"
#endif
```

src/codec_aom.c
```diff
@@ -666,15 +666,26 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             libavifDefaultTuneMetric = AOM_TUNE_PSNR;
         } else {
             libavifDefaultTuneMetric = AOM_TUNE_SSIM;
+#if defined(AOM_HAVE_TUNE_IQ)
+#if !defined(AOM_USAGE_ALL_INTRA) || !defined(ALL_INTRA_HAS_SPEEDS_7_TO_9)
```
wantehchang: This check looks strange here, especially the check for `ALL_INTRA_HAS_SPEEDS_7_TO_9`. The reason for this check is not obvious until we get to line 721 below:

```
            if (aomCpuUsed >= 7) {
#if defined(AOM_USAGE_ALL_INTRA) && defined(ALL_INTRA_HAS_SPEEDS_7_TO_9)
                if (!(addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE)) {
                }
#else
#endif
            }
```

If you want to avoid reordering the code now, I suggest we delete this check. This check is always false in libaom releases or on the libaom main branch.

src/codec_aom.c
wantehchang: Yes, we can reorder the code in a separate pull request.

Optional: If the `addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE` check is intended to detect whether all intra mode is used rather than whether we are encoding a still image, it would be good to use a local variable whose name clarifies the purpose of the check:

```
            avifBool useAllIntra = (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) != 0;
            if (image->matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_IDENTITY && useAllIntra) {
```

apps/avifenc.c
```diff
@@ -321,7 +321,7 @@ static void syntaxLong(void)
         printf("    end-usage=MODE                    : Rate control mode, one of 'vbr', 'cbr', 'cq', or 'q'\n");
         printf("    sharpness=S                       : Bias towards block sharpness in rate-distortion optimization of transform coefficients in 0..7. (Default: 0)\n");
         printf("    tune=METRIC                       : Tune the encoder for distortion metric, one of 'psnr', 'ssim' or 'iq'.\n");
-        printf("                                        (Default for color: ssim, default for alpha: psnr)\n");
+        printf("                                        (Default for color: still images (libaom v3.12.0+): iq, otherwise: ssim; default for alpha: psnr)\n");
```
wantehchang: Should we change "still images" to "still non-RGB images"?

src/codec_aom.c
y-guyon: Done and in https://github.com/AOMediaCodec/libavif/pull/2973.

src/codec_aom.c
y-guyon: > This check is always false in libaom releases or on the libaom main branch.

That is the point of a static assertion.

> If you want to avoid reordering the code now, I suggest we delete this check.

Done.

apps/avifenc.c

src/codec_aom.c
```diff
@@ -657,6 +657,40 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+    // Speed  2: GoodQuality CpuUsed 2
+    // Speed  3: GoodQuality CpuUsed 3
+    // Speed  4: GoodQuality CpuUsed 4
+    // Speed  5: GoodQuality CpuUsed 5
+    // Speed  6: GoodQuality CpuUsed 6
+    // Speed  7: RealTime    CpuUsed 7
+    // Speed  8: RealTime    CpuUsed 8
+    // Speed  9: RealTime    CpuUsed 9
+    // Speed 10: RealTime    CpuUsed 9
+    unsigned int libavifDefaultAomUsage = AOM_USAGE_GOOD_QUALITY;
```
wantehchang: Nit: Remove "Default" from the variable name.

This AOM usage won't be overridden by something else, so it isn't a default value.

I also think the "libavif" prefix is unnecessary. You didn't add the "libavif" prefix to "aomCpuUsed".

src/codec_aom.c
```diff
@@ -657,6 +657,40 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+        if (aomCpuUsed >= 7) {
+#if defined(AOM_USAGE_ALL_INTRA) && defined(ALL_INTRA_HAS_SPEEDS_7_TO_9)
+            if (!(addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE)) {
+                libavifDefaultAomUsage = AOM_USAGE_REALTIME;
+            }
+#else
+            libavifDefaultAomUsage = AOM_USAGE_REALTIME;
+#endif
+        }
+    }
```
wantehchang, line 734: I suggest also moving the following code that sets up `aomCpuUsed`:

```
        // aom_codec.h says: aom_codec_version() == (major<<16 | minor<<8 | patch)
        static const int aomVersion_2_0_0 = (2 << 16);
        const int aomVersion = aom_codec_version();
        if ((aomVersion < aomVersion_2_0_0) && (image->depth > 8)) {
            // Due to a known issue with libaom v1.0.0-errata1-avif, 10bpc and
            // 12bpc image encodes will call the wrong variant of
            // aom_subtract_block when cpu-used is 7 or 8, and crash. Until we get
            // a new tagged release from libaom with the fix and can verify we're
            // running with that version of libaom, we must avoid using
            // cpu-used=7/8 on any >8bpc image encodes.
            //
            // Context:
            //   * https://github.com/AOMediaCodec/libavif/issues/49
            //   * https://bugs.chromium.org/p/aomedia/issues/detail?id=2587
            //
            // Continued bug tracking here:
            //   * https://github.com/AOMediaCodec/libavif/issues/56

            if (aomCpuUsed > 6) {
                aomCpuUsed = 6;
            }
        }
```

src/codec_aom.c
wantehchang: This PR also includes the changes in https://github.com/AOMediaCodec/libavif/pull/2973. Is that intended? Or is https://github.com/AOMediaCodec/libavif/pull/2973 is prerequisite for this PR?

y-guyon: #2973 is a prerequisite for this PR #2830.

src/codec_aom.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2973.

Is this check still necessary by the way? libaom is at version 3+ now. v1.0.0-errata1-avif sounds quite old.

src/codec_aom.c
y-guyon: Done in #2973.

src/codec_aom.c
wantehchang: You are right. We can drop support for libaom version < 2.0.0 and remove this check.

src/codec_aom.c
```diff
@@ -666,15 +700,23 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             libavifDefaultTuneMetric = AOM_TUNE_PSNR;
         } else {
             libavifDefaultTuneMetric = AOM_TUNE_SSIM;
+#if defined(AOM_HAVE_TUNE_IQ)
+            // AOM_TUNE_IQ has been tuned for the YCbCr family of color spaces, and is favored for its low perceptual distortion.
+            // AOM_TUNE_IQ partially generalizes to, and benefits from other "YUV-like" spaces (e.g. YCgCo and ICtCp) including monochrome (luma only).
+            // AOM_TUNE_IQ sets --deltaq-mode=6 which can only be used in all intra mode.
+            if (image->matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_IDENTITY && libavifDefaultAomUsage == AOM_USAGE_ALL_INTRA) {
```
wantehchang: @juliobbv-p Julio: It just occurred to me that we may want to use tune=iq by default only for libaom v3.13.0+ because of the improvements you made in libaom v3.13.0.

As a reminder, here are the announcements of the libaom v3.12.0 and v3.13.0 releases:
https://groups.google.com/a/aomedia.org/g/av1-discuss/c/QJ-iK2CvmXw
https://groups.google.com/a/aomedia.org/g/av1-discuss/c/DWc3VKcao3o

Do you think we should even require the next libaom release (tentatively v3.14.0), which will have the latest improvements you made in November and December last year?

juliobbv-p: @wantehchang this is a great question!

IMO, the v3.13.0 implementation of tune=iq is mature enough for us to use in libavif by default -- improvements since then have been relatively minor. v3.12.0's implementation (used by The Guardian) is decent too, but it does have significantly different bit allocation characteristics that new quality-to-QP formula doesn't account for, so for simplicity reasons it'd be best to keep using tune SSIM there.

I don't think we should have a v3.14.0 requirement.

apps/avifenc.c
```diff
@@ -321,7 +321,7 @@ static void syntaxLong(void)
         printf("    end-usage=MODE                    : Rate control mode, one of 'vbr', 'cbr', 'cq', or 'q'\n");
         printf("    sharpness=S                       : Bias towards block sharpness in rate-distortion optimization of transform coefficients in 0..7. (Default: 0)\n");
         printf("    tune=METRIC                       : Tune the encoder for distortion metric, one of 'psnr', 'ssim' or 'iq'.\n");
-        printf("                                        (Default for color: ssim, default for alpha: psnr)\n");
+        printf("                                        (Default for color: still non-RGB images (libaom v3.12.0+): iq, otherwise: ssim; default for alpha: psnr)\n");
```
wantehchang: Please change `v3.12.0+` to `v3.13.0+`.

src/codec_aom.c
wantehchang: Julio: Thanks for the reply.

Yannis: Let's add a libaom version check here:

```
            static const int aomVersion_3_13_0 = (3 << 16) | (13 << 8);
            if (image->matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_IDENTITY &&
                libavifDefaultAomUsage == AOM_USAGE_ALL_INTRA &&
                aomVersion >= aomVersion_3_13_0) {
```

apps/avifenc.c
y-guyon: Done. In CHANGELOG too.

src/codec_aom.c

src/codec_aom.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2976.

src/codec_aom.c
```diff
@@ -722,15 +722,29 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
         }
     }
 
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
     // True if libavif knows that tune=iq is used, either by default by libavif, or explicitly set by the user.
     // False otherwise (including if libaom uses tune=iq by default, which is not the case as of v1.13.1 and earlier versions).
-    // This is only accurate for the first frame but tune=iq is only supported for still images in libavif and
-    // for all-intra coding in libaom (at least up to v1.13.1) anyway.
+    // This is only accurate for the first frame but tune=iq is only supported for all-intra coding in libaom (at least up to v1.13.1),
```
FooIbar: I assume you meant v3.13.1? Line 745 as well.

src/codec_aom.c
```diff
@@ -722,15 +722,29 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
 
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
     // True if libavif knows that tune=iq is used, either by default by libavif, or explicitly set by the user.
-    // False otherwise (including if libaom uses tune=iq by default, which is not the case as of v1.13.1 and earlier versions).
-    // This is only accurate for the first frame but tune=iq is only supported for still images in libavif and
-    // for all-intra coding in libaom (at least up to v1.13.1) anyway.
+    // False otherwise (including if libaom uses tune=iq by default, which is not the case as of v3.13.1 and earlier versions).
+    // This is only accurate for the first frame but tune=iq is only supported for all-intra coding in libaom (at least up to v3.13.1),
+    // and thus only supported for still images in libavif.
```
wantehchang: @juliobbv-p

Julio, Yannis: Please see issue https://github.com/AOMediaCodec/libavif/issues/2997. We also need to update this comment now that pull request https://github.com/AOMediaCodec/libavif/pull/2970 allows `tune=iq` to be used to encode some two-layer image items.

y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2999.

src/codec_aom.c
y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2999.

src/codec_aom.c
juliobbv-p: Thank you @y-guyon for taking care of this!

wantehchang:
Yannis: I am going to review the latest commit now.

src/codec_aom.c
```diff
@@ -758,6 +758,20 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+            // AOM_TUNE_IQ sets --deltaq-mode=6 which can only be used in all intra mode.
+            // AOM_TUNE_IQ was introduced in libaom v3.12.0 but it has significantly different bit
+            // allocation characteristics compared to v3.13.0. AOM_TUNE_IQ is used by default
+            // starting with v3.13.0 for fewer behavior changes in libavif.
+            static const int aomVersion_3_13_0 = (3 << 16) | (13 << 8);
+            if (image->matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_IDENTITY && aomUsage == AOM_USAGE_ALL_INTRA &&
+                aomVersion >= aomVersion_3_13_0) {
+                libavifDefaultTuneMetric = AOM_TUNE_IQ;
+            }
+#endif
```
wantehchang, line 773: IMPORTANT: I think this needs to be moved into the `else` block for `if (alpha)`, right?

y-guyon: Yes. I was wondering if we should merge https://github.com/AOMediaCodec/libavif/pull/3060 before this PR. What do you think?

wantehchang: I didn't review PR https://github.com/AOMediaCodec/libavif/pull/3060 because it was marked as a draft and had a compilation error. I suggest we merge this PR first (after moving the new code into the `else` block).

Are you worried that the code is complicated and prone to bugs if we merge this PR first?

y-guyon: > I suggest we merge this PR first (after moving the new code into the else block).

Done.

> Are you worried that the code is complicated and prone to bugs if we merge this PR first?

No. I just find the logic in #3060 easier to follow, and I was wondering if others do too.

wantehchang: I will review your PR https://github.com/AOMediaCodec/libavif/pull/3060 today or tomorrow.

Issue #2831, opened 2025-06-19
libavif ignores transformative properties of alpha auxiliary items


y-guyon:
Encoding the following PNG image (black opaque "abc" on translucent white background) with `irot` angle set to 1:

![Image](https://github.com/user-attachments/assets/b67108bb-1272-4312-8bfb-3543f872a782)

Gives the following AVIF image with no `irot` associated with the alpha auxiliary image item:

![Image](https://github.com/user-attachments/assets/67ed96bd-433e-40e0-b1b8-7334e954d8c1)

Which is decoded the same way by libavif (and CrabbyAvif and Chrome) as the following AVIF image which has that alpha `irot` association:

![Image](https://github.com/user-attachments/assets/c25d0f1a-48a9-41f6-bbd6-2730bb9c3fd1)

GIMP (which uses libheif I think) loads the AVIF above as (correct behavior according to some interpretation of the HEIF spec):

![Image](https://github.com/user-attachments/assets/c536dd36-d7f1-49d6-9ff8-3423653183c6)

y-guyon:
At encoding, libavif should write alpha `irot` associations.

At decoding, libavif should read and apply alpha `irot` associations. But that would break existing files encoded with libavif.

The same applies for `clap` and `imir`.

maryla-uc:
Tested in Safari (Version 18.5 (20621.2.5.11.8) on macOS 15.5 (24F74)) and the two middle avif images (which misleadingly have a .png extension but are actually avif) render identically (and same as Chrome).

y-guyon:
libavif no longer generates files without alpha `irot`/`clap`/`imir` associations.

libavif is permissive at decoding as described below:

https://github.com/AOMediaCodec/libavif/blob/841110fd8723abd0d5439cbb2cfb8e9cfa11f808/src/read.c#L5809-L5823

PR #2832, opened 2025-06-19
Associate clap,irot,imir with alpha aux items

y-guyon:
See https://github.com/AOMediaCodec/libavif/issues/2831

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -4477,7 +4477,7 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
     }
     if (orientation == 2 || orientation == 4 || orientation == 5 || orientation == 7) {
         imirPropIndex = meta->properties.count;
-        // Property with fixed index 10.
+        // Property with fixed 1-based index 10.
         assert(imirPropIndex + 1 == 10);
         avifProperty * imirProp = avifMetaCreateProperty(meta, "imir");
         AVIF_CHECKERR(imirProp, AVIF_RESULT_OUT_OF_MEMORY);


wantehchang:
Yannis: Thanks for the fix. I am reviewing this pull request now.

src/read.c
```diff
@@ -4469,6 +4469,9 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
         AVIF_CHECKERR(irotProp, AVIF_RESULT_OUT_OF_MEMORY);
         irotProp->u.irot.angle = orientation == 3 ? 2 : (orientation == 5 || orientation == 8) ? 1 : 3;
         AVIF_CHECKERR(avifDecoderItemAddProperty(colorItem, irotProp), AVIF_RESULT_OUT_OF_MEMORY);
+        if (hasAlpha) {
+            AVIF_CHECKERR(avifDecoderItemAddProperty(alphaItem, irotProp), AVIF_RESULT_OUT_OF_MEMORY);
+        }
```
wantehchang: Should we also add these transformative properties to the gain map item?

src/read.c
```diff
@@ -5773,7 +5779,47 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
...
 
-static avifResult aviDecoderCheckGainMapProperties(avifDecoder * decoder, const avifPropertyArray * gainMapProperties)
+static avifResult avifDecoderCheckAlphaProperties(avifDecoder * decoder, const avifPropertyArray * alphaProperties)
+{
+    const avifImage * image = decoder->image;
+    // The 'clap', 'irot' and 'imir' transformative properties should be applied to the alpha
+    // auxiliary image item before considering it a plane of the color image item.
+    // Otherwise, inequality with the transformative properties attached to the color image item
+    // should be treated as AVIF_RESULT_NOT_IMPLEMENTED.
+    // The latter is easier and is the behavior of libavif.
```
wantehchang, line 5815: This comment block (lines 5785-5789) is a little hard to understand. After reading this comment block several times, I think it will be slightly clearer if we change "Otherwise" to "Alternatively", to indicate that the first sentence and the second sentence are two options.

src/read.c
```diff
@@ -5773,7 +5779,47 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
...
+
+    // However, libavif up to version 1.3.0 generated images lacking transformative property
+    // associations with alpha auxiliary image items, so be lenient on their absence for
+    // backward compatibility with previously generated images.
```
wantehchang: It is not clear where the code corresponding to this comment is.

I guess this comment is referring to the if statement at lines 5797-5799. If so, I suggest moving this comment closer. For example,

```
    // However, libavif up to version 1.3.0 generated images lacking transformative property
    // associations with alpha auxiliary image items, so be lenient on their absence for
    // backward compatibility with previously generated images.
    if (clapProp == NULL && irotProp == NULL && imirProp == NULL) {
        return AVIF_RESULT_OK;
    }
```

or

```
    if (clapProp == NULL && irotProp == NULL && imirProp == NULL) {
        // However, libavif up to version 1.3.0 generated images lacking transformative property
        // associations with alpha auxiliary image items, so be lenient on their absence for
        // backward compatibility with previously generated images.
        return AVIF_RESULT_OK;
    }
```

src/read.c
```diff
@@ -5773,7 +5779,47 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
...
+
+    // TODO(yguyon): Check for 'ispe' values too.
+
+    if (!clapProp != !(image->transformFlags & AVIF_TRANSFORM_CLAP) ||
+        (clapProp && (clapProp->u.clap.widthN != image->clap.widthN || clapProp->u.clap.widthD != image->clap.widthD ||
+                      clapProp->u.clap.heightN != image->clap.heightN || clapProp->u.clap.heightD != image->clap.heightD ||
+                      clapProp->u.clap.horizOffN != image->clap.horizOffN || clapProp->u.clap.horizOffD != image->clap.horizOffD ||
+                      clapProp->u.clap.vertOffN != image->clap.vertOffN || clapProp->u.clap.vertOffD != image->clap.vertOffD))) {
+        avifDiagnosticsPrintf(&decoder->diag, "Clean aperture property mismatch between alpha auxiliary image item and color item");
+        return AVIF_RESULT_NOT_IMPLEMENTED;
```
wantehchang, line 5835: Similar code in the `avifDecoderCheckGainMapProperties()` function returns the `AVIF_RESULT_DECODE_GAIN_MAP_FAILED` error code under this condition. The `AVIF_RESULT_NOT_IMPLEMENTED` error code here seems to imply that this condition isn't really an error; it's just that we haven't written the code to handle this condition properly. Is that what you meant?

src/read.c
```diff
@@ -5773,7 +5779,47 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
...
+    }
+    if (!irotProp != !(image->transformFlags & AVIF_TRANSFORM_IROT) || (irotProp && irotProp->u.irot.angle != image->irot.angle)) {
+        avifDiagnosticsPrintf(&decoder->diag, "Rotation property mismatch between alpha auxiliary image item and color itemm");
+        return AVIF_RESULT_NOT_IMPLEMENTED;
+    }
+    if (!imirProp != !(image->transformFlags & AVIF_TRANSFORM_IMIR) || (imirProp && imirProp->u.imir.axis != image->imir.axis)) {
+        avifDiagnosticsPrintf(&decoder->diag, "Mirroring property mismatch between input items of tone-mapping derived image item");
```
wantehchang: The error message should read "Mirroring property mismatch between alpha auxiliary image item and color itemm".

src/read.c
```diff
@@ -6380,8 +6431,11 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         decoder->image->transformFlags |= AVIF_TRANSFORM_IMIR;
         decoder->image->imir = imirProp->u.imir;
     }
+    if (alphaProperties) {
```
wantehchang, line 6460: Nit: Change `gainMapProperties != NULL` (line 6437) to `gainMapProperties` for consistency.

src/write.c
```diff
@@ -3111,6 +3111,9 @@ static avifResult avifRWStreamWriteProperties(avifItemPropertyDedup * const dedu
             // Color specific properties
             // Note the 'tmap' (tone mapped image) item when a gain map is present also has itemCategory AVIF_ITEM_COLOR.
             AVIF_CHECKRES(avifEncoderWriteTransformativeProperties(&dedup->s, s, itemMetadata, &item->associations, dedup));
+        } else if (item->itemCategory == AVIF_ITEM_ALPHA) {
+            // Cropping, rotation and mirroring must also be applied to alpha auxiliary items.
+            AVIF_CHECKRES(avifEncoderWriteTransformativeProperties(&dedup->s, s, itemMetadata, &item->associations, dedup));
```
wantehchang, line 3116: Optional: This line is exactly the same as line 3113 above. Should we combine the two?

```
        if (item->itemCategory == AVIF_ITEM_COLOR || item->itemCategory == AVIF_ITEM_ALPHA) {
            // ...
            AVIF_CHECKRES(avifEncoderWriteTransformativeProperties(&dedup->s, s, itemMetadata, &item->associations, dedup));
```

tests/gtest/aviftransformtest.cc
```diff
@@ -96,6 +96,80 @@ TEST(TransformTest, ClopIrotImor) {
...
+  DecoderPtr ref_decoder(avifDecoderCreate());
+  ASSERT_NE(ref_decoder, nullptr);
+  ASSERT_EQ(avifDecoderReadFile(ref_decoder.get(), ref_decoded.get(),
+                                ref_path.c_str()),
+            AVIF_RESULT_OK);
+
+  // Read a file missing an 'irot' property associated with the alpha auxiliary
+  // image item. While this should not lead to the same output, libavif still
+  // decodes it the same way because libavif used to encode it without that
+  // association.
```
wantehchang, line 161: Optional/future: We can consider adding a strict flag for this issue. If that strict flag is set, libavif will ignore backward compatibility with libavif <= 1.3.0 and do the right thing.

wantehchang (approved):
LGTM. Thanks!

src/read.c

Also moved transformative property associations last.

src/read.c
y-guyon: Done.

src/read.c

src/read.c
y-guyon: Yes. This PR focuses on alpha, excepted for https://github.com/AOMediaCodec/libavif/pull/2832#discussion_r2159678461. What should `avifDecoderCheckGainMapProperties()` return is out of scope. Maybe it should be changed to `AVIF_RESULT_NOT_IMPLEMENTED` in another PR. @maryla-uc thoughts?

src/read.c

src/read.c
y-guyon: Done.

src/write.c
y-guyon: It also is the same line for `AVIF_ITEM_GAIN_MAP`. For consistency let's merge none or all (none it is).

tests/gtest/aviftransformtest.cc
y-guyon: I thought about it. It can be added later on if necessary.

src/read.c
```diff
@@ -4462,22 +4462,26 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
...
         AVIF_CHECKERR(irotProp, AVIF_RESULT_OUT_OF_MEMORY);
         irotProp->u.irot.angle = orientation == 3 ? 2 : (orientation == 5 || orientation == 8) ? 1 : 3;
-        AVIF_CHECKERR(avifDecoderItemAddProperty(colorItem, irotProp), AVIF_RESULT_OUT_OF_MEMORY);
     } else {
         AVIF_CHECKERR(avifMetaCreateProperty(meta, "skip"), AVIF_RESULT_OUT_OF_MEMORY); // Placeholder.
     }
     if (orientation == 2 || orientation == 4 || orientation == 5 || orientation == 7) {
+        imirPropIndex = meta->properties.count;
         // Property with fixed index 10.
+        assert(imirPropIndex + 1 == 10);
```
wantehchang, line 4481: Nit: Add "1-based" before "index 10" to match the comment at line 4470?

wantehchang (approved):

src/read.c
y-guyon: Done.

src/read.c
```diff
@@ -4593,6 +4597,34 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
     }
     AVIF_ASSERT_OR_RETURN(meta->properties.count == 29);
 
+    // ISO/IEC 23008-12 Section 6.5.1:
+    //   Readers shall allow and ignore descriptive properties following the first transformative or unrecognized
+    //   property, whichever is earlier, in the sequence associating properties with an item.
+    //   Writers should arrange the descriptive properties specified in 6.5 prior to any other properties in the
+    //   sequence associating properties with an item.
+    //
+    // irot and imir are transformative properties, so associate them last.
```
wantehchang, line 4606: Optional: When I read this comment, I was wondering why we didn't seem to be following the requirement on readers (this code is in src/read.c). Later I realized that we are simulating writing a meta box here. It would be good to clarify this. One possibility might be to remove the requirement on readers (lines 4601-4602).

y-guyon: Done in https://github.com/AOMediaCodec/libavif/pull/2839

PR #2834, opened 2025-06-19
Remove ext/libargparse.patch

wantehchang:


wantehchang:
This is just an experiment to verify that cmake 3.13 needs ext/libargparse.patch.

PR #2836, opened 2025-06-24
Check PSNR in transformtest because of rav1e

y-guyon:


y-guyon:
Fixes CI failure at head introduced in https://github.com/AOMediaCodec/libavif/pull/2832

PR #2837, opened 2025-06-24
MinGW: fix error of undefined strcmp()

carlo-bramini:
I cross-compiled latest sources of libavif with` x86_64-w64-mingw32` but I got an error:
```
src/libavif-1.3.0/contrib/gdk-pixbuf/loader.c:366:17: error: implicit declaration of function ‘strcmp’ [-Wimplicit-function-declaration]
  366 |             if (strcmp (*kiter, "quality") == 0) {
      |                 ^~~~~~
src/libavif-1.3.0/contrib/gdk-pixbuf/loader.c:11:1: note: include ‘<string.h>’ or provide a declaration of ‘strcmp’
   10 | #include <gdk-pixbuf/gdk-pixbuf.h>
  +++ |+#include <string.h>
   11 |
ninja: build stopped: subcommand failed.
*** ERROR: cygninja: command failed: all
```
Actually, the error is correct and it happens because `strcmp()` is used instead of `g_strcmp0()`, attached patch fixes this issue.

### DIFF
--- a/contrib/gdk-pixbuf/loader.c
+++ b/contrib/gdk-pixbuf/loader.c
@@ -4,6 +4,7 @@
 
 #include <avif/avif.h>
 #include <stdlib.h>
+#include <string.h>
 
 #define GDK_PIXBUF_ENABLE_BACKEND
 #include <gdk-pixbuf/gdk-pixbuf-io.h>


contrib/gdk-pixbuf/loader.c
```diff
@@ -363,7 +363,7 @@ avif_image_saver (FILE *f,
         gchar **viter = values;
 
         while (*kiter) {
-            if (strcmp (*kiter, "quality") == 0) {
+            if (g_strcmp0 (*kiter, "quality") == 0) {
```
wantehchang: Carlo: Thanks for the patch.

@novomesk Daniel: Could you please review this change? I don't know why `strcmp()` can't be used in this file (after we add `#include <string.h>`). Thanks!

wantehchang: Daniel: Note that `g_strcmp0()` is already used at line 333 above.

```
static gboolean
avif_is_save_option_supported (const gchar *option_key)
{
    if (g_strcmp0 (option_key, "quality") == 0) {
        return TRUE;
    }
    return FALSE;
}
```

novomesk: This is how the plug-ins in gdk-pixbuf repo looks like:

https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/blob/2.43.2/gdk-pixbuf/io-jpeg.c#L1385

They use `strcmp` at similar portion of code, while using `g_strcmp0` in the same file too.

I would prefer to try  adding `#include <string.h>` after `#include <stdlib.h>` (like it is in GNOME/gdk-pixbuf repo) first.

carlo-bramini: I tried both solutions and:
- adding `#include <string.h>` or
- replacing `strcmp()` with `g_strcmp0()`

and they both resolved the issue.
I think you just need to decide which one should be applied.

wantehchang: Carlo: Please add `#include <string.h>`. Thanks!

carlo-bramini: > Carlo: Please add `#include <string.h>`. Thanks!

Done!

wantehchang: I checked a few other plug-ins in the gdk-pixbuf repo. All of them use `g_strcmp0` only in their `is_save_option_supported` callback functions, and use `strcmp` elsewhere. So this is apparently the convention. Perhaps the `option_key` input parameter of the `is_save_option_supported` callback may be a null pointer? (`g_strcmp0` handles null pointer inputs gracefully.)

PR #2848, opened 2025-07-19
Don't hardcode 'include' as includedir

marcin-serwin:
GNUInstallDirs defines CMAKE_INSTALL_INCLUDEDIR as a location to install headers. This is important for distributions that want to install development files under a different prefix, see https://github.com/NixOS/nixpkgs/pull/424658.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,7 @@ include(ExternalProject)
 include(FetchContent)
 include(FindPkgConfig)
 include(AvifExternalProjectUtils)
+include(GNUInstallDirs)
 
 option(AVIF_ENABLE_NODISCARD "Add [[nodiscard]] to some functions. CMake must be at least 3.21 to force C23." OFF)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -591,7 +592,9 @@ endif()
 # Main avif library.
 set_target_properties(avif PROPERTIES VERSION ${LIBRARY_VERSION} SOVERSION ${LIBRARY_SOVERSION})
 target_link_libraries(avif PRIVATE avif_obj)
-target_include_directories(avif PUBLIC $<BUILD_INTERFACE:${libavif_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include>)
+target_include_directories(
+    avif PUBLIC $<BUILD_INTERFACE:${libavif_SOURCE_DIR}/include> $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+)
```
wantehchang, line 597: Thanks for the pull request. The original code comes from a very old commit: https://github.com/AOMediaCodec/libavif/commit/8737c2595

PR #2851, opened 2025-07-24
Bump clang-format to 19

y-guyon:

PR #2853, opened 2025-07-24
Remove empty comments

y-guyon:
They were introduced by the migration to clang-format-19 in #2851.
It may be a bug but it is not worth spending time on it.


y-guyon:
I did not find much help at https://clang.llvm.org/docs/ClangFormatStyleOptions.html#aligntrailingcomments.

PR #2855, opened 2025-07-30
Use proper generator expression for aom

vrabaud:
To prevent errors with the newer versions like:
CMake Error in build/_deps/libaom-src/CMakeLists.txt:
  Target "aom" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "/usr/local/google/home/vrabaud/software/avif/build/_deps/libaom-src"

  which is prefixed in the build directory.


wantehchang (approved):
LGTM. Thanks!

vrabaud:
It seems FetchContent_Populate sets libaom_SOURCE_DIR to something interpreted as relative by the generator expression, thus leading to
```
-I/home/runner/work/libavif/libavif/home/runner/work/libavif/libavif/build/_deps/libaom-src
```
instead of 
```
-Ihome/runner/work/libavif/libavif/build/_deps/libaom-src
```
It is not consistent though. Investigating (it works locally)

PR #2856, opened 2025-07-30
Add config/aom_config.h include dir only for AVM

wantehchang:
config/aom_config.h is a private header, so public headers should not include it. During AVM development, its public headers need to include config/aom_config.h for the definitions of experiment macros. So add the include directory for config/aom_config.h only for AVM.


cmake/Modules/LocalAom.cmake
```diff
@@ -185,7 +185,11 @@ else()
     endif()
 
     set_property(TARGET aom PROPERTY AVIF_LOCAL ON)
-    target_include_directories(aom INTERFACE $<BUILD_INTERFACE:$<PATH:ABSOLUTE_PATH,${libaom_SOURCE_DIR},/>;${AOM_BINARY_DIR}>)
+    target_include_directories(aom INTERFACE $<BUILD_INTERFACE:$<PATH:ABSOLUTE_PATH,${libaom_SOURCE_DIR},/>>)
+    if(AVIF_CODEC_AVM)
+        # avm/aom/aom_encoder.h includes config/aom_config.h which is generated by the local build of avm.
+        target_include_directories(aom INTERFACE $<BUILD_INTERFACE:${AOM_BINARY_DIR}>)
+    endif()
```
wantehchang, line 192: This is modeled after lines 21-25 above. I edited the comment at line 23 so that the same comment can be used in both places.

PR #2859, opened 2025-08-11
Revert "Set Unspecified CICP in Seq Header OBU (#2687)"

### LINKED ISSUE #2850: --cicp / --nclx X/X/X not working anymore since 1.3.0
The update from version 1.2.1 to 1.3.0 broke the --cicp command line parameter of avifenc (or the logic behind it).

I have a HDR BT.2020 PQ PNG test file which I encoded as an avif file:

avifenc -q 80 -d 10 --cicp 9/16/9 --jobs all INPUT_HDR_BT2020_PQ_16bit.png OUTPUT_HDR_BT2020_PQ_10bit.avif

With avifenc 1.2.1 ffprobe outputs this for the resulting file:
Stream #0:0[0x1]: Video: av1 (libdav1d) (High) (av01 / 0x31307661), yuv444p10le(pc, bt2020nc/bt2020/smpte2084), 8256x5504 [SAR 1:1 DAR 3:2], 1 fps, 1 tbr, 1 tbn (default)

With avifenc 1.3.0 ffprobe outputs this for the resulting file:
Stream #0:0[0x1]: Video: av1 (libdav1d) (High) (av01 / 0x31307661), yuv444p10le(pc), 8256x5504 [SAR 1:1 DAR 3:2], 1 fps, 1 tbr, 1 tbn (default)

As it can be seen the section "bt2020nc/bt2020/smpte2084" is missing from the file created with avifenc 1.3.0. Therefore some image viewers display it using the normal SDR BT.709 BT.1886 format which leads to incorrect display of the HDR data.

maryla-uc:
This partially reverts commit fa089a0c3f8ed6e202034589cf98394e0226d66c. The added test avifcicptest.cc is not reverted.

Fixes #2850

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -848,12 +848,15 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             // Keep default Unspecified (2) colour primaries, transfer characteristics,
             // and matrix coefficients.
         } else {
-            // libaom's defaults are AOM_CICP_CP_UNSPECIFIED, AOM_CICP_TC_UNSPECIFIED,
-            // AOM_CICP_MC_UNSPECIFIED, AOM_CSP_UNKNOWN, and 0 (studio/limited range). Call
-            // aom_codec_control() only if the values are not the defaults.
             // AVIF specification, Section 2.2.1. "AV1 Item Configuration Property":
             //   The values of the fields in the AV1CodecConfigurationBox shall match those
             //   of the Sequence Header OBU in the AV1 Image Item Data.
+            // CICP values could be set to 2/2/2 (Unspecified) in the Sequence Header OBU for
+            // simplicity and to save 3 bytes, but some decoders ignore the colr box and rely
+            // on the OBU contents instead. See #2850.
+            // libaom's defaults are AOM_CICP_CP_UNSPECIFIED, AOM_CICP_TC_UNSPECIFIED,
+            // AOM_CICP_MC_UNSPECIFIED, AOM_CSP_UNKNOWN, and 0 (studio/limited range). Call
+            // aom_codec_control() only if the values are not the defaults.
             if (image->colorPrimaries != AVIF_COLOR_PRIMARIES_UNSPECIFIED) {
                 aom_codec_control(&codec->internal->encoder, AV1E_SET_COLOR_PRIMARIES, (int)image->colorPrimaries);
             }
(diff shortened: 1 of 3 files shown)


src/codec_aom.c
```diff
@@ -1097,6 +1106,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         // AVIF specification, Section 2.2.1. "AV1 Item Configuration Property":
         //   The values of the fields in the AV1CodecConfigurationBox shall match those
         //   of the Sequence Header OBU in the AV1 Image Item Data.
+        aomImage.cp = (aom_color_primaries_t)image->colorPrimaries;
+        aomImage.tc = (aom_transfer_characteristics_t)image->transferCharacteristics;
+        aomImage.mc = (aom_matrix_coefficients_t)image->matrixCoefficients;
```
y-guyon, line 1111: These are not part of the `AV1CodecConfigurationBox`. I suggest moving these lines away from the above comment.

src/codec_rav1e.c
```diff
@@ -208,9 +208,9 @@ static avifResult rav1eCodecEncodeImage(avifCodec * codec,
         // It may also save 3 bytes since the AV1 encoder may set color_description_present_flag to 0
         // (see Section 5.5.2 "Color config syntax" of the AV1 specification).
         rav1e_config_set_color_description(rav1eConfig,
-                                           RA_MATRIX_COEFFICIENTS_UNSPECIFIED,
-                                           RA_COLOR_PRIMARIES_UNSPECIFIED,
-                                           RA_TRANSFER_CHARACTERISTICS_UNSPECIFIED);
+                                           (RaMatrixCoefficients)image->matrixCoefficients,
+                                           (RaColorPrimaries)image->colorPrimaries,
+                                           (RaTransferCharacteristics)image->transferCharacteristics);
```
y-guyon, line 214: The comment above is no longer accurate. I suggest replacing it with a link to the issue and/or a short explanation:

> CICP values could be set to 2/2/2 (Unspecified) in the Sequence Header OBU for simplicity and to save 3 bytes, but some decoders ignore the `colr` box and rely on the OBU contents instead. See https://github.com/AOMediaCodec/libavif/issues/2850.

To be done in other AV1 codec wrappers too.

src/codec_aom.c
```diff
@@ -848,12 +848,24 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
         } else {
-            // libaom's defaults are AOM_CSP_UNKNOWN and 0 (studio/limited range).
-            // Call aom_codec_control() only if the values are not the defaults.
-
             // AVIF specification, Section 2.2.1. "AV1 Item Configuration Property":
             //   The values of the fields in the AV1CodecConfigurationBox shall match those
             //   of the Sequence Header OBU in the AV1 Image Item Data.
+            // CICP values could be set to 2/2/2 (Unspecified) in the Sequence Header OBU for
+            // simplicity and to save 3 bytes, but some decoders ignore the colr box and rely
+            // on the OBU contents instead. See #2850.
```
wantehchang, line 856: Maryla, Yannis: It's good that we left AVM as is. We can also leave CP/TC/MC as 2/2/2 (Unspecified) for SlimHEIF because SlimHEIF is a new container format, i.e., add a check for `(encoder->headerFormat & AVIF_HEADER_MINI) != 0` here, guarded by `#if defined(AVIF_ENABLE_EXPERIMENTAL_MINI)`.

PR #2867, opened 2025-09-02
Remove comments that the gain map API is experimental.

maryla-uc:
The experimental build flag was removed so this can be considered stable.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -614,11 +614,6 @@ typedef struct avifContentLightLevelInformationBox
 // ---------------------------------------------------------------------------
 // avifGainMap
 // Gain Maps are a solution for a consistent and adaptive display of HDR images.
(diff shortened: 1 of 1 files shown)


vrabaud (approved):

include/avif/avif.h
```diff
@@ -614,11 +614,6 @@ typedef struct avifContentLightLevelInformationBox
 // ---------------------------------------------------------------------------
 // avifGainMap
 // Gain Maps are a solution for a consistent and adaptive display of HDR images.
-// Gain Maps are a HIGHLY EXPERIMENTAL FEATURE. The format might still change and
-// images containing a gain map encoded with the current version of libavif might
-// not decode with a future version of libavif. The API is not guaranteed
-// to be stable, and might even be removed in the future. Use at your own risk.
-// This is based on ISO/IEC JTC 1/SC 29/WG 3 m64379
```
wantehchang, line 621: Should we replace this line with the ISO standard number?

PR #2871, opened 2025-09-03
Update jpeg gain map doc links

maryla-uc:


apps/shared/avifjpeg.c
```diff
@@ -594,7 +594,7 @@ static inline void SwapDoubles(double * x, double * y)
 }
 
 // Parses gain map metadata from XMP.
-// See https://helpx.adobe.com/camera-raw/using/gain-map.html
+// See https://developer.android.com/media/platform/hdr-image-format
```
wantehchang, line 597: Should we also update the comment at line 606?

```
    // Set default values from Adobe's spec.
```

and the comment at line 654?

```
    // Not in Adobe's spec but both color spaces should be the same so this value doesn't matter.
```

wantehchang (approved):

apps/shared/avifjpeg.c
maryla-uc: Good catc! Done.

PR #2879, opened 2025-10-01
Add creationTime & modificationTime to avifEncoder

wantehchang:
Allow specifying the creation_time and modification_time fields in boxes such as mvhd for generating deterministic outputs when encoding images sequences. Set to 0 for the current behavior of using the current time.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -246,8 +246,8 @@ static void syntaxLong(void)
     printf("    --icc FILENAME                    : Provide an ICC profile payload to be associated with the primary item (implies --ignore-icc)\n");
     printf("    --timescale,--fps V               : Timescale for image sequences. If all frames are 1 timescale in length, this is equivalent to frames per second. (Default: 30)\n");
     printf("                                        If neither duration nor timescale are set, avifenc will attempt to use the framerate stored in a y4m header, if present.\n");
-    printf("    --creation-time                   : Creation time for image sequences, in seconds since 1970-01-01 00:00:00 UTC (the Unix epoch). (Default: 0, use current time)\n");
-    printf("    --modification-time               : Modification time for image sequences, in seconds since 1970-01-01 00:00:00 UTC (the Unix epoch). (Default: 0, use current time)\n");
+    printf("    --creation-time                   : Creation time for image sequences, in seconds since 1970-01-01 00:00:00 UTC (the Unix epoch). (Default: 0, use the modification time)\n");
+    printf("    --modification-time               : Modification time for image sequences, in seconds since 1970-01-01 00:00:00 UTC (the Unix epoch). (Default: 0, use the current time)\n");
     printf("    -k,--keyframe INTERVAL            : Maximum keyframe interval for image sequences (any set of INTERVAL consecutive frames will have at least one keyframe). Set to 0 to disable (default).\n");
     printf("    --ignore-exif                     : If the input file contains embedded Exif metadata, ignore it (no-op if absent)\n");
     printf("    --ignore-xmp                      : If the input file contains embedded XMP metadata, ignore it (no-op if absent)\n");
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -41,6 +41,8 @@ typedef struct
     int layers;
     int speed;
     avifHeaderFormatFlags headerFormat;
+    uint64_t creationTime;
+    uint64_t modificationTime;
 
     avifBool paspPresent;
     uint32_t paspValues[2];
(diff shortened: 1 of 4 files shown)


tests/test_cmd_stdin.sh
```diff
@@ -102,14 +79,14 @@ test_stdin() {
...
-  # only the encoded samples.
-  test_stdin "${INPUT_Y4M_ANIMATED}" true
+  test_stdin "${INPUT_Y4M_STILL}" y4m
+  test_stdin "${INPUT_PNG}" png
+  test_stdin "${INPUT_JPEG}" jpeg
+
+  # Make the output of avifenc for animations deterministic by specifying the
+  # creation_time and modification_time fields in boxes such as mvhd.
+  NOW=$(date +%s)
+  test_stdin "${INPUT_Y4M_ANIMATED}" y4m --creation-time ${NOW} --modification-time ${NOW}
```
wantehchang, line 89: Yannis: Please review my changes to this file. Thanks!

include/avif/avif.h
```diff
@@ -1603,6 +1603,14 @@ typedef struct avifEncoder
     // Version 1.2.0 ends here. Add any new members after this line.
     // --------------------------------------------------------------------------------------------
 
+    // creationTime and modificationTime are only used when encoding an image sequence (animated
+    // image).
+    //
+    // creationTime and modificationTime are in seconds since midnight, Jan. 1, 1970 UTC (the Unix
+    // epoch). If set to 0 (the default), libavif uses the current time.
+    uint64_t creationTime;
+    uint64_t modificationTime;
```
wantehchang, line 1613: Please review two API design decisions.

1. Use the Unix epoch (because it is more commonly used) instead of the 1904 epoch used by ISO BMFF.
2. Use the value 0 to indicate "use the current time". This is fine because no AVIF files were created or modified in 1970.

include/avif/avif.h
```diff
@@ -1603,6 +1603,14 @@ typedef struct avifEncoder
     // Version 1.2.0 ends here. Add any new members after this line.
     // --------------------------------------------------------------------------------------------
 
+    // creationTime and modificationTime are only used when encoding an image sequence (animated
+    // image).
+    //
+    // creationTime and modificationTime are in seconds since midnight, Jan. 1, 1970 UTC (the Unix
+    // epoch). If set to 0 (the default), libavif uses the current time.
```
wantehchang: It seems that only `modificationTime` should default to the current time. A better default for `creationTime` seems to be the modification time.

This way libavif clients that want deterministic encoder output will only need to set `modificationTime` in most cases.

What do you think?

apps/avifenc.c
```diff
@@ -1899,6 +1907,22 @@ int main(int argc, char * argv[])
                 goto cleanup;
             }
             settings.outputTiming.timescale = (uint64_t)timescaleInt;
+        } else if (!strcmp(arg, "--creation-time")) {
+            NEXTARG();
+            long long creationTime = strtoll(arg, NULL, 0);
+            if (creationTime < 0 || (unsigned long long)creationTime > UINT64_MAX) {
```
y-guyon, line 1913: That should be `INT64_MAX`

include/avif/avif.h
y-guyon: That sounds good to me.

include/avif/avif.h
y-guyon: 1. No opinion, Unix epoch looks good.
2. Sure.

tests/test_cmd_stdin.sh
y-guyon: LGTM

include/avif/avif.h
vigneshvg: sgtm.

apps/avifenc.c
wantehchang: This check makes sure that we can safely cast `creationTime` to `uint64_t` at line 1917. So it first checks for values outside the range of `0`..`UINT64_MAX`.

include/avif/avif.h
wantehchang: Thanks. I will go ahead and make this change.

Note that it is a little counter intuitive to make `modificationTime` the main timestamp because `creationTime` seems more natural when we encode an image sequence. But we may be writing the encoded output to an existing file (i.e., overwriting the file), so we are only sure that the current time is the modification time.

@vigneshvg Vignesh: This is opposite to what CrabbyAvif does. Is this change okay?

src/write.c
```diff
@@ -3219,10 +3221,14 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
 #endif // AVIF_ENABLE_EXPERIMENTAL_MINI
 
     const avifImage * imageMetadata = encoder->data->imageMetadata;
+    uint64_t now = (uint64_t)time(NULL);
+    uint64_t modificationTime = (encoder->modificationTime != 0) ? encoder->modificationTime : now;
+    uint64_t creationTime = (encoder->creationTime != 0) ? encoder->creationTime : modificationTime;
```
wantehchang, line 3226: This is the important change. `creationTime` defaults to `modificationTime`, rather than the other way around.

vigneshvg: yes i think this is fine.

Issue #2886, opened 2025-10-29
[bug] Overflow (read) in `avifRGBImagePremultiplyAlpha` on a `AVIF_RGB_FORMAT_GRAYA` image


hgarrereyn:
Hi, there is a potential bug in `avifRGBImagePremultiplyAlpha` reachable by an `avifRGBImage` with format `AVIF_RGB_FORMAT_GRAYA`.

This bug was reproduced on https://github.com/AOMediaCodec/libavif/commit/37ff3889b7b0b4ca34d708dc5bb5d0ef240d821d.

### Description

The original reproducer allocates two avifRGBImage buffers and eventually calls avifRGBImagePremultiplyAlpha on one of them. In the failing path the image is configured as GRAYA (two channels: gray + alpha) with depth 10 bits. libavif allocates pixels sized for GRAYA at >8-bit depth (2 channels x 16 bits = 4 bytes per pixel). However, in alpha.c at the highlighted lines, the 16-bit premultiply path treats the pixel as if it had four 16-bit channels (RGBA) by indexing pixel[0..3] and advancing across the row accordingly. For a 6x1 GRAYA image, the allocated buffer is 24 bytes, but the loop reads and writes beyond this, causing a heap buffer overflow on the very next iteration.

### POC

The following testcase demonstrates the bug:

**testcase.cpp**
```cpp
#include <cstdio>
#include <cstdint>
#include <cstring>
extern "C" {
#include "/fuzz/install/include/avif/avif.h"
}
int main(){
    avifRGBImage rgb; memset(&rgb, 0, sizeof(rgb));
    rgb.width = 6;
    rgb.height = 1;
    rgb.depth = 10;   // triggers 16-bit-per-channel path
    rgb.format = AVIF_RGB_FORMAT_GRAYA; // 2 channels with alpha
    if (avifRGBImageAllocatePixels(&rgb) != AVIF_RESULT_OK) return 0;
    // Premultiply assumes 4 channels for 16bpc and overruns the 2-channel buffer
    avifRGBImagePremultiplyAlpha(&rgb);
    avifRGBImageFreePixels(&rgb);
    return 0;
}

```

**stdout**
```text
/fuzz/testcase.cpp:15:5: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
   15 |     avifRGBImagePremultiplyAlpha(&rgb);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~ ~~~~
[...]

PR #2887, opened 2025-10-30
Avoid issue AOMediaCodec/libavif#2886

wantehchang:
avifRGBImagePremultiplyAlpha() and avifRGBImageUnpremultiplyAlpha() assume four-channel pixel buffers, so they would overrun the two-channel pixel buffer in an avifRGBImage of the AVIF_RGB_FORMAT_GRAYA or AVIF_RGB_FORMAT_AGRAY format. Return AVIF_RESULT_NOT_IMPLEMENTED before this bug is fixed.

The bug was reported by Harrison Green.

Avoids https://github.com/AOMediaCodec/libavif/issues/2886.

### DIFF
--- a/src/alpha.c
+++ b/src/alpha.c
@@ -165,6 +165,10 @@ avifResult avifRGBImagePremultiplyAlpha(avifRGBImage * rgb)
         return libyuvResult;
     }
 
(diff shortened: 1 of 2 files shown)


vrabaud (approved):
Thx

PR #2888, opened 2025-10-30
Use ninja to build dav1d

vrabaud:
This a revert of #2665 to get fuzz tests to build again.


vrabaud:
For some reason we did not get any email from the failures at https://oss-fuzz-build-logs.storage.googleapis.com/index.html#libavif

PR #2893, opened 2025-11-04
Call aom_codec_peek_stream_info() for size limit

y-guyon:
Fixes https://g-issues.oss-fuzz.com/issues/457325510.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -97,18 +97,13 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
                                      avifImage * image)
 {
     if (!codec->internal->decoderInitialized) {
-        if (!sample) {
-            // Nothing to decode: no need to initialize anything.
-            return AVIF_FALSE;
-        }
         aom_codec_iface_t * const decoderInterface = aom_codec_av1_dx();
         struct aom_codec_stream_info streamInfo = { 0 };
         if (aom_codec_peek_stream_info(decoderInterface, sample->data.data, sample->data.size, &streamInfo) != AOM_CODEC_OK) {
-            // The sample is incorrect or there are not enough bytes yet: no need to initialize anything.
             return AVIF_FALSE;
         }
         if (streamInfo.w == 0 || streamInfo.h == 0) {
-            // The sequence header was not found: no need to initialize anything.
+            // The sequence header was not found: treat it as an error.
             return AVIF_FALSE;
         }
         if (avifDimensionsTooLarge(streamInfo.w, streamInfo.h, codec->imageSizeLimit, codec->imageDimensionLimit)) {


y-guyon:
That may not be the best approach. Feel free to suggest an alternative.

My main concern is about the guarantees of finding a Sequence Header OBU with non-zero dimensions in an AV1 payload before outputting any frame in libavif.

The opaque returned type `avifBool` is also inconvenient for differentiating between not enough input bytes and an irrecoverable error.

`aom_codec_peek_stream_info()` seems of tiny complexity so I am not anxious about the performance implications of this change.

src/codec_aom.c
```diff
@@ -97,13 +97,30 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
...
+            // Nothing to decode: no need to initialize anything.
+            return AVIF_FALSE;
+        }
+        aom_codec_iface_t * const decoderInterface = aom_codec_av1_dx();
+        struct aom_codec_stream_info streamInfo = { 0 };
+        if (aom_codec_peek_stream_info(decoderInterface, sample->data.data, sample->data.size, &streamInfo) != AOM_CODEC_OK) {
+            // The sample is incorrect or there are not enough bytes yet: no need to initialize anything.
+            return AVIF_FALSE;
+        }
+        if (streamInfo.w == 0 || streamInfo.h == 0) {
```
wantehchang, line 105: I was surprised that `aom_codec_peek_stream_info()` did not return an error like "not enough data" when the sequence header was not found. I inspected the source code. Indeed, `aom_codec_peek_stream_info()` may return `AOM_CODEC_OK` with the local variable `got_sequence_header` equal to 0.

include/avif/internal.h
```diff
@@ -606,10 +606,11 @@ typedef struct avifCodec
     avifDiagnostics * diag;               // Shallow copy; owned by avifEncoder or avifDecoder
 
     // Decoder options (for getNextImage):
-    int maxThreads;          // See avifDecoder::maxThreads.
-    uint32_t imageSizeLimit; // See avifDecoder::imageSizeLimit.
-    uint8_t operatingPoint;  // Operating point, defaults to 0.
-    avifBool allLayers;      // if true, the underlying codec must decode all layers, not just the best layer
+    int maxThreads;               // See avifDecoder::maxThreads.
+    uint32_t imageSizeLimit;      // See avifDecoder::imageSizeLimit.
+    uint32_t imageDimensionLimit; // See avifDecoder::imageDimensionLimit.
```
wantehchang, line 611: I know `codec->imageDimensionLimit` is added so that we can call `avifDimensionsTooLarge()`. However, src/codec_dav1d.c only checks `codec->imageSizeLimit`. Perhaps we should aim for consistent behavior across AV1 decoders and only check `codec->imageSizeLimit` in src/codec_aom.c.

src/codec_aom.c
```diff
@@ -97,13 +97,30 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
...
+            // The sequence header was not found: no need to initialize anything.
+            return AVIF_FALSE;
+        }
+        if (avifDimensionsTooLarge(streamInfo.w, streamInfo.h, codec->imageSizeLimit, codec->imageDimensionLimit)) {
```
wantehchang, line 109: I assume the issue we are trying to fix is that the bitstream's frame size may be larger than the container's image size (from the 'ispe' box)?

It seems that libavif allows this. When it detects this condition, it will call `avifImageScaleWithLimit()` to downscale the decoded frame to the container's image size. Is this correct?

src/codec_aom.c
```diff
@@ -97,13 +97,30 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
                                      avifImage * image)
 {
     if (!codec->internal->decoderInitialized) {
+        if (!sample) {
```
wantehchang: Just curious (I didn't analyze the code): why do we call this function with a null `sample`? (I see we also check for a null `sample` at line 154 below.)

wantehchang:
Yannis wrote: 
> My main concern is about the guarantees of finding a Sequence Header OBU with non-zero dimensions in an AV1 payload before outputting any frame in libavif.

If I understand your question correctly, this is guaranteed. In AV1, a bitstream starts with a temporal delimiter OBU and a sequence header OBU, so when we decode a frame, a sequence header OBU must have been processed.

The dimensions in a sequence header OBU are nonzero by construction.

> `aom_codec_peek_stream_info()` seems of tiny complexity so I am not anxious about the performance implications of this change.

Yes. We should aim for software security first. A better long-term solution is to add to libaom a decoder option on the maximum frame size, similar to dav1d's `Dav1dSettings::frame_size_limit`.

include/avif/internal.h
y-guyon: If so, we should not have checked for `imageSizeLimit` in codec_dav1d. I do not think consistency here is important.

src/codec_aom.c
y-guyon: I understand it is the case, yes.

src/codec_aom.c
y-guyon: I do not know, I did not analyze the code yet either.

y-guyon:
> If I understand your question correctly, this is guaranteed. In AV1, a bitstream starts with a temporal delimiter OBU and a sequence header OBU, so when we decode a frame, a sequence header OBU must have been processed.
>
> The dimensions in a sequence header OBU are nonzero by construction.

This change should be fine then.

src/codec_aom.c
wantehchang: I found that the caller of  this function never passes a null `sample` parameter, so please delete this `if (!sample)` statement. I will write a pull request to remove the unnecessary null checks for `sample` in src/codec_aom.c and src/codec_avm.c. (No such checks are in src/codec_dav1d.c and src/codec_libgav1.c.)

src/codec_aom.c
```diff
@@ -97,13 +97,30 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
...
+            // Nothing to decode: no need to initialize anything.
+            return AVIF_FALSE;
+        }
+        aom_codec_iface_t * const decoderInterface = aom_codec_av1_dx();
+        struct aom_codec_stream_info streamInfo = { 0 };
+        if (aom_codec_peek_stream_info(decoderInterface, sample->data.data, sample->data.size, &streamInfo) != AOM_CODEC_OK) {
+            // The sample is incorrect or there are not enough bytes yet: no need to initialize anything.
```
wantehchang: Nit: I think this comment can be deleted because it is clear that we don't need to initialize `codec->internal->decoder` on error.

src/codec_aom.c
```diff
@@ -97,13 +97,30 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
...
+            return AVIF_FALSE;
+        }
+        if (streamInfo.w == 0 || streamInfo.h == 0) {
+            // The sequence header was not found: no need to initialize anything.
```
wantehchang: Nit: Remove the ": no need to initialize anything" part of the comment, or change "anything" to "codec->internal->decoder" or "the libaom decoder".

We can also add "treat it as an error" after "The sequence header was not found".

wantehchang (approved):

src/codec_aom.c
wantehchang: I know why we have a null check for `sample` below. But we should still delete this `if (!sample)` statement.

The reason we are checking `sample` at line 154 is that we set `sample` to NULL at line 160:

```
        } else if (sample) {
            codec->internal->iter = NULL;
            if (aom_codec_decode(&codec->internal->decoder, sample->data.data, sample->data.size, NULL)) {
                return AVIF_FALSE;
            }
            spatialID = sample->spatialID;
            sample = NULL;
        } else {
```

So the best we can do is to assert `sample` is not null on entry to this function.

src/codec_aom.c

src/codec_aom.c

src/codec_aom.c

PR #2894, opened 2025-11-05
Assert `sample` not null on entry to getNextImage

wantehchang:
The aomCodecGetNextImage() and avmCodecGetNextImage() functions may set `sample` to NULL internally.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -96,6 +96,8 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
                                      avifBool * isLimitedRangeAlpha,
                                      avifImage * image)
 {
+    assert(sample);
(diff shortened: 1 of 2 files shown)


src/codec_aom.c
```diff
@@ -96,6 +96,8 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
                                      avifBool * isLimitedRangeAlpha,
                                      avifImage * image)
 {
+    assert(sample);
```
wantehchang, line 99: Thanks for the review. Another approach is to use a boolean flag to indicate that the `sample` parameter has been consumed. Then we don't need to set `sample` to NULL after consuming it. In any case we can start with this assertion.

PR #2895, opened 2025-11-05
Implement avifRGBImagePremultiplyAlpha GRAYA,AGRAY

wantehchang:
Implement avifRGBImagePremultiplyAlpha and
avifRGBImageUnpremultiplyAlpha for AVIF_RGB_FORMAT_GRAYA and AVIF_RGB_FORMAT_AGRAY.

Fixes https://github.com/AOMediaCodec/libavif/issues/2886.

### DIFF
--- a/src/alpha.c
+++ b/src/alpha.c
@@ -165,10 +165,6 @@ avifResult avifRGBImagePremultiplyAlpha(avifRGBImage * rgb)
         return libyuvResult;
     }
 
-    if (rgb->format == AVIF_RGB_FORMAT_GRAYA || rgb->format == AVIF_RGB_FORMAT_AGRAY) {
-        return AVIF_RESULT_NOT_IMPLEMENTED;
-    }
-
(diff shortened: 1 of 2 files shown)


src/alpha.c
```diff
@@ -198,7 +194,7 @@ avifResult avifRGBImagePremultiplyAlpha(avifRGBImage * rgb)
                 }
                 row += rgb->rowBytes;
             }
-        } else {
+        } else if (rgb->format == AVIF_RGB_FORMAT_ARGB || rgb->format == AVIF_RGB_FORMAT_ABGR) {
```
wantehchang, line 197: I listed the supported RGB formats explicitly and avoided using `else` for "the rest of the formats with alpha". This makes the code more robust if we add a new RGB format with alpha in the future.

wantehchang:
Yes, the repetitive code is unfortunate. I am not good at writing code-generating macros, so I used brute force.

PR #2897, opened 2025-11-14
Initialize cicpSet in avifDecoderFindGainMapItem()

wantehchang:
avifDecoderFindGainMapItem() should initialize the local variable cicpSet to AVIF_FALSE, because avifReadColorNclxProperty() leaves its output parameter cicpSet unchanged if no colr nclx property was read.

Bug: https://issues.oss-fuzz.com/460535483


jzern:
> Bug: https://g-issues.oss-fuzz.com/issues/460535483

https://issues.oss-fuzz.com/460535483 will be a more accessible link when it's made public. The issue url is available in the drop down next to the bug number, or use the 'l-u' keyboard shortcut.

wantehchang:
James: Done. Thanks!

PR #2898, opened 2025-11-14
Set the gain map image's CICP directly

wantehchang:
In avifDecoderFindGainMapItem(), set the gain map images's CICP directly rather than through intermediate local variables. Then we don't need to pass a cicpSet output parameter to avifReadColorNclxProperty().

Bug: https://issues.oss-fuzz.com/460535483

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -5771,15 +5771,6 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
         return AVIF_RESULT_INVALID_TONE_MAPPED_IMAGE;
     }
 
-    avifColorPrimaries colorPrimaries = AVIF_COLOR_PRIMARIES_UNSPECIFIED;
-    avifTransferCharacteristics transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
-    avifMatrixCoefficients matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
-    avifRange yuvRange = AVIF_RANGE_FULL;
-    avifBool cicpSet;
-    // Look for a colr nclx box. Other colr box types (e.g. ICC) are not supported.
-    AVIF_CHECKRES(
-        avifReadColorNclxProperty(&gainMapItemTmp->properties, &colorPrimaries, &transferCharacteristics, &matrixCoefficients, &yuvRange, &cicpSet));
-
     // -- Everything is valid, do memory allocations and fill in output data. --
 
     decoder->image->gainMap = avifGainMapCreate();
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -5790,12 +5781,13 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
...
-            image->matrixCoefficients = matrixCoefficients;
-            image->yuvRange = yuvRange;
-        }
+        // Look for a colr nclx box. Other colr box types (e.g. ICC) are not supported.
+        AVIF_CHECKRES(avifReadColorNclxProperty(&gainMapItemTmp->properties,
+                                                &image->colorPrimaries,
+                                                &image->transferCharacteristics,
+                                                &image->matrixCoefficients,
+                                                &image->yuvRange,
+                                                NULL));
```
wantehchang, line 5790: Maryla: Please review this PR carefully. I believe it is equivalent to the original code except for how much of `decoder->image->gainMap` may be partially constructed on failure.

I wonder if we should destroy `decoder->image->gainMap` and reset it to NULL on failure (before returning from line 5783 or 5785).

y-guyon: > I believe it is equivalent to the original code except for how much of `decoder->image->gainMap` may be partially constructed on failure.

Another effect is that a file with a faulty `colr` associated with the gain map item will NOT fail to decode unless `decoder->imageContentToDecode&AVIF_IMAGE_CONTENT_GAIN_MAP!=0`, after this PR. That is fine in my opinion.

wantehchang: Yannis: I missed that difference. Thank you for catching that.

I will let you and Maryla decide whether this PR is a good idea as I am not familiar with the original PR https://github.com/AOMediaCodec/libavif/pull/2800.

maryla-uc: I think it was done this way for error management, to avoid having failure cases after the allocation of the gain map image (as shown by the comment above "Everything is valid, do memory allocations and fill in output data."). Maybe it's simpler to keep the previous code but initialize cicpSet to false?
Otherwise I think we should indeed destroy the gain map and it set back to NULL, but I'm not sure how much it matters.

wantehchang: Yes, we can use the other pull request https://github.com/AOMediaCodec/libavif/pull/2897.

I also thought error management was the intention. But that goal is not met if the second memory allocation (shown below) fails:

```
        decoder->image->gainMap->image = avifImageCreateEmpty();
        avifImage * image = decoder->image->gainMap->image;
        AVIF_CHECKERR(image, AVIF_RESULT_OUT_OF_MEMORY);
```

wantehchang:
I abandoned this pull request in favor of the simpler https://github.com/AOMediaCodec/libavif/pull/2897.

PR #2899, opened 2025-11-14
Document API contract of avifReadColorNclxProperty

wantehchang:
Document that if cicpSet is not NULL, *cicpSet must be AVIF_FALSE on entry.

Note: This API contract isn't that obvious, and may have been the source of https://issues.oss-fuzz.com/460535483.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -5521,6 +5521,8 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
     return AVIF_RESULT_OK;
 }
 
+// If cicpSet is not NULL, the caller must set |*cicpSet| to AVIF_FALSE before
+// calling this function.
 // On success, this function returns AVIF_RESULT_OK and does the following:
 // * If a nclx property was found in |properties|:
 //   - Set |*colorPrimaries|, |*transferCharacteristics|, |*matrixCoefficients|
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -5521,6 +5521,8 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
     return AVIF_RESULT_OK;
 }
 
+// If cicpSet is not NULL, the caller must set |*cicpSet| to AVIF_FALSE before
+// calling this function.
```
wantehchang, line 5525: This API contract isn't that obvious. In this PR I merely document it. It would be more constructive to change the API contract to be more intuitive.

src/read.c
```diff
@@ -5775,7 +5778,7 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
     avifTransferCharacteristics transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
     avifMatrixCoefficients matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
     avifRange yuvRange = AVIF_RANGE_FULL;
-    avifBool cicpSet;
+    avifBool cicpSet = AVIF_FALSE;
```
wantehchang, line 5781: This change is temporary. We need to merge either https://github.com/AOMediaCodec/libavif/pull/2897 or https://github.com/AOMediaCodec/libavif/pull/2898 as a prerequisite for this PR.

wantehchang: https://github.com/AOMediaCodec/libavif/pull/2898 has been merged.

PR #2900, opened 2025-11-14
Update LocalAom.cmake to AVM research-v12.0.0

wantehchang:


wantehchang:
@y-guyon FYI.

In Google's internal repository, we also need to apply https://gitlab.com/AOMediaCodec/avm/-/merge_requests/2692 to AVM research-v12.0.0 to fix memory leaks in avifavmtest and avifavmminitest. I don't know why our ASan build doesn't detect the memory leaks. I guess it doesn't build AVM?

jzern (approved):

PR #2904, opened 2025-11-18
Support sample transforms officially

y-guyon:
Sample transform support is always enabled now.

https://aomediacodec.github.io/av1-avif/v1.2.0.html is officially released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ The changes are relative to the previous release, unless the baseline is specifi
 ### Added since 1.3.0
 
 * Allow avifenc to read png or jpeg files through stdin using --stdin-format.
-* Support some Sample Transform schemes.
+* Support some Sample Transform schemes as defined in the version 1.2 of the
+  AVIF specification.
 
 ### Changed since 1.3.0
 


src/read.c
```diff
@@ -832,16 +832,14 @@ typedef struct avifMeta
...
-#endif
-
-#if defined(AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM)
     // Parsed from Sample Transform metadata if present, otherwise empty.
     avifSampleTransformExpression sampleTransformExpression;
     // Bit depth extracted from the pixi property of the Sample Transform derived image item, if any.
     uint32_t sampleTransformDepth;
+
+#if defined(AVIF_ENABLE_EXPERIMENTAL_MINI)
+    // If true, the fields above were extracted from a MinimizedImageBox.
```
wantehchang, line 841: IMPORTANT:  Please check if this comment is still correct after it is moved after the two sample transform fields. If not, we need to move the `fromMiniBox` field back to its original location.

tests/CMakeLists.txt
```diff
@@ -146,12 +143,21 @@ if(AVIF_GTEST)
...
+            avifallocationtest
+            avifgridapitest
+            avifincrtest
+            aviflosslesstest
+            avifmetadatatest
+            avifpixitest
+            # rav1e and svt do not support encoder changes (potentially different bit depth between
+            # primary item and bit depth extension).
+            avif16bittest
+            avifaltrtest
```
wantehchang, line 159: I am wondering why we didn't need to disable avif16bittest and avifaltrtest before and need to disable them now.

Is it because we didn't test rav1e and svt with `AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM` enabled?

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 ### Removed since 1.3.0
 
 * Remove ext/avm.cmd.
+* Remove the AVIF_ENABLE_EXPERIMENTAL_SAMPLE_TRANSFORM CMake flag.
```
wantehchang, line 42: Should we add in the "Added since 1.3.0" section that sample transforms of AVIF v1.2 are offically supported? Then we may not need this item in the "Removed since 1.3.0" section.

src/read.c
y-guyon: It is correct in the sense that the `MinimizedImageBox` cannot store any Sample Transform and these fields will be set to "none".

tests/CMakeLists.txt
y-guyon: Yes.

CHANGELOG.md
y-guyon: I added an entry and kept the other one. I think it is clearer.

PR #2905, opened 2025-11-18
Add --depthext flag to avifenc

y-guyon:
--depth and --depthext map to a Sample Transform recipe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Allow avifenc to read png or jpeg files through stdin using --stdin-format.
 * Support some Sample Transform schemes as defined in the version 1.2 of the
   AVIF specification.
+* Add an optional argument to the --depth flag of avifenc used to enable a bit
+  depth extension scheme in the encoded file.
 
 ### Changed since 1.3.0
 


y-guyon:
Requires https://github.com/AOMediaCodec/libavif/pull/2904

apps/avifenc.c
```diff
@@ -221,7 +223,8 @@ static void syntaxLong(void)
     printf("    --mini                            : EXPERIMENTAL: Use reduced header if possible (backward-incompatible)\n");
 #endif
     printf("    -l,--lossless                     : Set all defaults to encode losslessly, and emit warnings when settings/input don't allow for it\n");
-    printf("    -d,--depth D                      : Output depth, one of 8, 10 or 12. (JPEG/PNG only; For y4m or stdin, depth is retained)\n");
+    printf("    -d,--depth D                      : Primary item output bit depth per channel. One of: 8/10/12. (JPEG/PNG only; For y4m or stdin, depth is retained)\n");
+    printf("    --depthext Dext Doverlap          : Bit depth extension using Sample Transforms. Allowed [D Dext Doverlap] combinations: [D 0 0]/[8 8 0]/[12 4 0]/[12 8 4]\n");
```
wantehchang: I suggest using a comma (,) instead of space between Dexp and Doverlap. See the --crop, --clap, and --clli options below.

wantehchang: Optional: We can consider using a dash `-` to separate `depth` and `ext`, i.e., `--depth-ext`.

We are not consistent with the use of `-` to separate words in option names.

apps/avifenc.c
```diff
@@ -221,7 +223,8 @@ static void syntaxLong(void)
     printf("    --mini                            : EXPERIMENTAL: Use reduced header if possible (backward-incompatible)\n");
 #endif
     printf("    -l,--lossless                     : Set all defaults to encode losslessly, and emit warnings when settings/input don't allow for it\n");
-    printf("    -d,--depth D                      : Output depth, one of 8, 10 or 12. (JPEG/PNG only; For y4m or stdin, depth is retained)\n");
+    printf("    -d,--depth D                      : Primary item output bit depth per channel. One of: 8/10/12. (JPEG/PNG only; For y4m or stdin, depth is retained)\n");
```
wantehchang: Nit: The term "primary item" is jargon that I am afraid most avifenc users don't understand. Please see if it is possible to rephrase this.

One possibility is to only refer to "primary item" when the `--depthext` option is also used. I.e., we keep the original help message for the `-d/--depth` option, and then add a second sentence that says "If `--depthext` is also specified, this option means the bit depth of the primary item, which can be decoded by older AVIF decoders that don't support bit depth extension."

y-guyon: Rephrased. PTAL

apps/avifenc.c

wantehchang: Thanks. It also occurred to me that we could also "overload" the `-d,--depth` option to take two forms of argument:

```
    -d,--depth D
    -d,--depth D,Dext,Doverlap
```

Then we don't need to add the `--depth-ext` option. Please consider it and see if you like it. Let me take a quick look at the new code now.

tests/CMakeLists.txt
```diff
@@ -256,6 +256,7 @@ if(AVIF_BUILD_APPS)
     target_link_libraries(are_images_equal aviftest_helpers avif_enable_warnings)
     add_cmd_test(test_cmd ${CMAKE_CURRENT_SOURCE_DIR}/data)
     add_cmd_test(test_cmd_animation ${CMAKE_CURRENT_SOURCE_DIR}/data)
+    add_cmd_test(test_cmd_depthext ${CMAKE_CURRENT_SOURCE_DIR}/data)
```
wantehchang: Optional: rename this file `test_cmd_depth_ext` to match the new `--depth-ext` option name.

tests/CMakeLists.txt
```diff
@@ -292,6 +293,7 @@ if(AVIF_BUILD_APPS)
     if(NOT AVIF_CODEC_AOM_ENABLED OR NOT AVIF_CODEC_AOM_ENCODE)
         # Only aom encoder supports AV1 lossless encoding.
         set_property(TEST test_cmd_animation PROPERTY DISABLED True)
+        set_property(TEST test_cmd_depthext PROPERTY DISABLED True)
```
wantehchang: Future,unrelated: The latest SVT-AV1 releases also support lossless encoding now. So we can update the consition and the comment at lines 293-294.

apps/avifenc.c
```diff
@@ -221,7 +223,8 @@ static void syntaxLong(void)
     printf("    --mini                            : EXPERIMENTAL: Use reduced header if possible (backward-incompatible)\n");
 #endif
     printf("    -l,--lossless                     : Set all defaults to encode losslessly, and emit warnings when settings/input don't allow for it\n");
-    printf("    -d,--depth D                      : Output depth, one of 8, 10 or 12. (JPEG/PNG only; For y4m or stdin, depth is retained)\n");
+    printf("    -d,--depth D                      : Output bit depth per channel. One of: 8/10/12. (JPEG/PNG only; For y4m or stdin, depth is retained)\n");
```
wantehchang: Nit: Also add "bit" before "depth is retained".

apps/avifenc.c
```diff
@@ -221,7 +223,8 @@ static void syntaxLong(void)
...
+    printf("    --depth-ext Dext,Doverlap         : Bit depth extension added to the primary item's --depth D. Allowed [D Dext Doverlap] combinations: [D 0 0]/[8 8 0]/[12 4 0]/[12 8 4]\n");
```
wantehchang: The new help messages look good. Thanks.

Optional: In the examples of [8 8 0]/[12 4 0]/[12 8 4], end users may be more interested in knowing that the final bit depth is 16 bits. So it seems good to replace Doverlap with Dfinal (tentative name).

apps/avifenc.c
```diff
@@ -1198,6 +1225,22 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
...
+        encoder->sampleTransformRecipe = AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_8B_8B;
+    } else if (firstImage->depth == 16 && settings->depthExtension == 4 && settings->depthOverlap == 0) {
+        encoder->sampleTransformRecipe = AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_4B;
+    } else if (firstImage->depth == 16 && settings->depthExtension == 8 && settings->depthOverlap == 4) {
+        encoder->sampleTransformRecipe = AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_8B_OVERLAP_4B;
+    } else if (settings->depthExtension != 0 || settings->depthOverlap != 0) {
+        fprintf(stderr,
+                "ERROR: Unsupported bit depth extension scheme: %u = %d + %u - %u\n",
+                firstImage->depth,
+                (int)firstImage->depth - (int)settings->depthExtension,
```
wantehchang: IMPORTANT: I think this should be (`int` casts omitted for clarity):

```
firstImage->depth - (settings->depthExtension - settings->depthOverlap)
```

tests/CMakeLists.txt
y-guyon: Done.

apps/avifenc.c

apps/avifenc.c
y-guyon: This is done through the new help message "If specified, Dextension [...] reach 16-bit depth at decoding"

apps/avifenc.c

apps/avifenc.c
```diff
@@ -221,7 +222,9 @@ static void syntaxLong(void)
     printf("    --mini                            : EXPERIMENTAL: Use reduced header if possible (backward-incompatible)\n");
 #endif
     printf("    -l,--lossless                     : Set all defaults to encode losslessly, and emit warnings when settings/input don't allow for it\n");
-    printf("    -d,--depth D                      : Output depth, one of 8, 10 or 12. (JPEG/PNG only; For y4m or stdin, depth is retained)\n");
+    printf("    -d,--depth D[,Dextension]         : D is the output bit depth per channel. D must be 8, 10 or 12. (JPEG/PNG only; y4m or stdin: bit depth is retained)\n");
+    printf("                                        If specified, Dextension adds a hidden encoded image of Dextension bit depth in the same file as the primary image to reach 16-bit depth at decoding.\n");
+    printf("                                        See avifSampleTransformRecipe for the supported combinations (8,8 and 12,4 and 12,8).\n");
```
leo-barnes, line 227: Where is `avifSampleTransformRecipe`? Can't see it in this change.

leo-barnes:
> The new code LGTM. Thanks!
> 
> @leo-barnes Leo: If you have time, please take a look at the enhanced `-d,--depth` option and its help message in apps/avifenc.c and think about how we should expose bit depth extension to end users in the `avifenc` command-line interface. Thanks!

Sorry, completely missed this. Added a comment inline, but otherwise I think the command-line interface makes sense.

apps/avifenc.c
y-guyon: Here:

https://github.com/AOMediaCodec/libavif/blob/c3c6f7d7d05f927dfa7de69bae185285c3321440/include/avif/avif.h#L728-L757

PR #2909, opened 2025-11-20
Fix Extended XMP parsing

vrabaud:


apps/shared/avifjpeg.c
```diff
@@ -883,7 +877,7 @@ static avifBool avifJPEGReadInternal(FILE * f,
     memset(&rgb, 0, sizeof(avifRGBImage));
 
     // Standard XMP segment followed by all extended XMP segments.
```
maryla-uc: update comment?

apps/shared/avifjpeg.c
```diff
@@ -1163,7 +1152,10 @@ static avifBool avifJPEGReadInternal(FILE * f,
             uint8_t xmpNote[AVIF_JPEG_XMP_NOTE_TAG_LENGTH + AVIF_JPEG_EXTENDED_XMP_GUID_LENGTH];
             memcpy(xmpNote, AVIF_JPEG_XMP_NOTE_TAG, AVIF_JPEG_XMP_NOTE_TAG_LENGTH);
             memcpy(xmpNote + AVIF_JPEG_XMP_NOTE_TAG_LENGTH, extendedXMPGUID, AVIF_JPEG_EXTENDED_XMP_GUID_LENGTH);
-            if (!avifJPEGFindSubstr(standardXMPData, standardXMPSize, xmpNote, sizeof(xmpNote))) {
+            avifBool foundAlternativeXmpNote;
```
maryla-uc: nit: local style would dictate foundAlternativeXMPNote

apps/shared/avifjpeg.c
```diff
@@ -1172,17 +1164,119 @@ static avifBool avifJPEGReadInternal(FILE * f,
                     fprintf(stderr, "XMP extraction failed: standard and extended XMP GUID mismatch\n");
                     goto cleanup;
                 }
+                foundAlternativeXmpNote = AVIF_TRUE;
             }
+            (void)foundAlternativeXmpNote;
 
+#if defined(AVIF_ENABLE_JPEG_GAIN_MAP_CONVERSION)
```
maryla-uc, line 1282: AVIF_ENABLE_JPEG_GAIN_MAP_CONVERSION should be renamed maybe as this has nothing to do with gainmaps and everything to do with libxml support.
it's also weird to have checks for a GAIN_MAP macro in tests/gtest/avifmetadatatest.cc

apps/shared/avifjpeg.c
```diff
@@ -1172,17 +1164,119 @@ static avifBool avifJPEGReadInternal(FILE * f,
...
+            // Initialize the XMP RDF.
```
maryla-uc: is it possible to move some of this code to a helper function? (if too complicated you can leave it as is, this function is already super long and beyond saving I think)

apps/shared/avifjpeg.c
```diff
@@ -1172,17 +1164,119 @@ static avifBool avifJPEGReadInternal(FILE * f,
...
+        cleanup_xml:
+            xmlFreeDoc(xmpDoc);
+            xmlFreeDoc(extendedXMPDoc);
+            xmlFree(xmlBuff);
+            if (hasError) {
+                goto cleanup;
+            }
+#else
+            fprintf(stderr, "Must be compiled with libxml2 to handle HasExtendedXMP property\n");
+            goto cleanup;
```
maryla-uc: mhh, this actually sounds like a breaking change: previously, jpeg files with extended xmp could be converted to avif, although they would end up with weird concatenaed xmp. now they will fail? (if avifenc was built without libxml2). maybe we should fall back to the old behavior (concatenate), or ignore the extended xmp and print a warning? or ignore the xmp entirely? or make libxml2 mandatory?

apps/shared/avifjpeg.c
vrabaud: Indeed.

apps/shared/avifjpeg.c
vrabaud: Yes, they will fail: do we want to keep producing jpegs with broken xmp that we cannot parse later on ? I'd say no. But we can decide to. Maybe with a warning.

maryla-uc: This is not for producing jpeg. It's the jpeg reading code, so it would be used when converting jpeg to avif. I think producing an avif file with no xmp, partial xmp (discard the extended part), or even broken xmp (concatenated), is better than refusing to convert a file that would previously succeed.

apps/shared/avifjpeg.c
vrabaud: Actually, the changes here do not care about the gain map, just about libxml2. We could imagine keeping AVIF_ENABLE_JPEG_GAIN_MAP_CONVERSION but also have a AVIF_HAVE_LIBXML2

apps/shared/avifjpeg.c
vrabaud: Ok, I am displaying a warning and keeping the old behavior.

PR #2911, opened 2025-11-22
Update AVM to commit e3b7bd42

wantehchang:
AVM commit https://gitlab.com/AOMediaCodec/avm/-/commit/e3b7bd42 includes https://gitlab.com/AOMediaCodec/avm/-/merge_requests/2692. This allows us to re-enable Tiny tests in avifavmminitest.cc and avifavmtest.cc.

### DIFF
--- a/cmake/Modules/LocalAom.cmake
+++ b/cmake/Modules/LocalAom.cmake
@@ -1,5 +1,5 @@
 set(AVIF_AOM_GIT_TAG v3.13.1)
-set(AVIF_AVM_GIT_TAG research-v12.0.0)
+set(AVIF_AVM_GIT_TAG e3b7bd426c282494c07eaa3e559e461a3e96f823)
 
 if(AVIF_CODEC_AVM)
     # Building the avm repository generates files such as "libaom.a" because it is a fork of aom,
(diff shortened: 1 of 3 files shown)


cmake/Modules/LocalAom.cmake
```diff
@@ -1,5 +1,5 @@
 set(AVIF_AOM_GIT_TAG v3.13.1)
-set(AVIF_AVM_GIT_TAG research-v12.0.0)
+set(AVIF_AVM_GIT_TAG e3b7bd426c282494c07eaa3e559e461a3e96f823)
```
wantehchang, line 2: This AVM commit has just been imported into Google's internal repository.

cmake/Modules/LocalAom.cmake
```diff
@@ -95,7 +95,7 @@ else()
             BINARY_DIR "${AOM_BINARY_DIR}"
             GIT_TAG ${AVIF_AVM_GIT_TAG}
             GIT_PROGRESS ON
-            GIT_SHALLOW ON
+            GIT_SHALLOW OFF
```
wantehchang, line 98: `GIT_SHALLOW` only works for a tag. To check out a commit, we have to turn off `GIT_SHALLOW`.

wantehchang:
Vincent: I rebased the pull request. Thank you for fixing the CI hashFiles path.

PR #2912, opened 2025-11-24
Fix hashFiles path

vrabaud:
This made the CI fail

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -41,7 +41,7 @@ runs:
     - name: Generate cargo cache key
       if: ${{ inputs.use-rust == 'true' }}
       run: |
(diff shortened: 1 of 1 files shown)


wantehchang:
Vincent: Thanks a lot for fixing the CI workflow failures! Do you know what changed that caused the CI failures?

vrabaud:
I thought it was the '**' to '**/*'  fix but it might actually be due to a bug upstream on some VMs, cf https://github.com/orgs/community/discussions/180160

wantehchang:
Vincent: You are right. That does look like the issue we ran into. Do we need to revert this pull request?

vrabaud:
Well, i also made the change action -> actions . And the doc at https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#hashfiles does not mention '\*\*'. But it does '**/*.something' so I thought it got stricter. Let's stay closer to the doc.

PR #2914, opened 2025-11-24
Removed unused assignments

y-guyon:
Avoid static analysis complaining about unused values.


src/write.c
```diff
@@ -2073,6 +2073,7 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
                 }
                 AVIF_CHECKRES(
                     avifEncoderCreateBitDepthExtensionImage(encoder, item, itemWillBeEncodedLosslessly, cellImage, &sampleTransformedImage));
+                assert(cellImagePlaceholder == NULL);
```
wantehchang, line 2076: Strictly speaking, Coverity is right. But it's good to keep the assignment at line 2072 as defensive programming in case this function is modified in the future.

I think Coverity analyzes our Debug build, so it should see the assertion you added. But we can also just mark the Coverity defect as "Ignore".

PR #2915, opened 2025-11-26
scale: fix a potential memory leak

x2018:
In `src/scale.c`, `avifImageScaleWithLimit()` did not properly do the error handling when `avifImageAllocatePlanes()` fails, which can result in memory leak. This PR fixes it.


vrabaud:
Thx ! How did you find this one ?

x2018:
I found it with a tool that is still under development for research.

vrabaud:
Our fuzzers did not find it because OOM are treated differently. I'll fix that soonish.

PR #2920, opened 2025-12-06
Quality to QP refactor

juliobbv-p:
- Pass down `quality` to `xxxCodecEncodeImage()` instead of the `quantizer` (QP) value
- Delegate `quality` -> `quantizer` conversions to each encoder's `xxxCodecEncodeImage()` instead
- Remove double-conversions (`quality` -> `AOM quantizer` -> `quantizer`) in the rav1e and AVM encoders

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -544,8 +544,8 @@ typedef enum avifEncoderChange
     AVIF_ENCODER_CHANGE_MAX_QUANTIZER_ALPHA = (1 << 3),
     AVIF_ENCODER_CHANGE_TILE_ROWS_LOG2 = (1 << 4),
     AVIF_ENCODER_CHANGE_TILE_COLS_LOG2 = (1 << 5),
-    AVIF_ENCODER_CHANGE_QUANTIZER = (1 << 6),
-    AVIF_ENCODER_CHANGE_QUANTIZER_ALPHA = (1 << 7),
+    AVIF_ENCODER_CHANGE_QUALITY = (1 << 6),
+    AVIF_ENCODER_CHANGE_QUALITY_ALPHA = (1 << 7),
     AVIF_ENCODER_CHANGE_SCALING_MODE = (1 << 8),
 
     AVIF_ENCODER_CHANGE_CODEC_SPECIFIC = (1 << 30)
@@ -565,10 +565,10 @@ typedef avifBool (*avifCodecGetNextImageFunc)(struct avifCodec * codec,
 // encoder->tileRowsLog2, encoder->tileColsLog2, and encoder->autoTiling. The caller of
 // avifCodecEncodeImageFunc is responsible for automatic tiling if encoder->autoTiling is set to
 // AVIF_TRUE. The actual tiling values are passed to avifCodecEncodeImageFunc as parameters.
-// Similarly, avifCodecEncodeImageFunc should use the quantizer parameter instead of
-// encoder->quality and encoder->qualityAlpha. If disableLaggedOutput is AVIF_TRUE, then the encoder will emit the output frame
-// without any lag (if supported). Note that disableLaggedOutput is only used by the first call to this function (which
-// initializes the encoder) and is ignored by the subsequent calls.
+// Similarly, avifCodecEncodeImageFunc should use the quality parameter instead of
+// encoder->quality, encoder->qualityAlpha, and encoder->qualityGainMap. If disableLaggedOutput is AVIF_TRUE, then
+// the encoder will emit the output frame without any lag (if supported). Note that disableLaggedOutput is only
+// used by the first call to this function (which initializes the encoder) and is ignored by the subsequent calls.
 //
 // Note: The caller of avifCodecEncodeImageFunc always passes encoder->data->tileRowsLog2 and
 // encoder->data->tileColsLog2 as the tileRowsLog2 and tileColsLog2 arguments. Because
@@ -581,7 +581,7 @@ typedef avifResult (*avifCodecEncodeImageFunc)(struct avifCodec * codec,
                                                avifBool alpha,
                                                int tileRowsLog2,
                                                int tileColsLog2,
-                                               int quantizer,
+                                               int quality,
                                                avifEncoderChanges encoderChanges,
                                                avifBool disableLaggedOutput,
                                                avifAddImageFlags addImageFlags,
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -557,13 +557,20 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
 
 static avifBool aomCodecEncodeFinish(avifCodec * codec, avifCodecEncodeOutput * output);
 
+static int aomQualityToQuantizer(int quality)
+{
+    const int quantizer = ((100 - quality) * 63 + 50) / 100;
+
+    return quantizer;
+}
+
 static avifResult aomCodecEncodeImage(avifCodec * codec,
                                       avifEncoder * encoder,
                                       const avifImage * image,
                                       avifBool alpha,
                                       int tileRowsLog2,
                                       int tileColsLog2,
-                                      int quantizer,
+                                      int quality,
                                       avifEncoderChanges encoderChanges,
                                       avifBool disableLaggedOutput,
                                       avifAddImageFlags addImageFlags,
@@ -571,6 +578,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
 {
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
+    const int quantizer = aomQualityToQuantizer(quality);
 
     // For encoder->scalingMode.horizontal and encoder->scalingMode.vertical to take effect in AOM
     // encoder, config should be applied for each frame, so we don't care about changes on these
@@ -943,8 +951,8 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 quantizerUpdated = AVIF_TRUE;
             }
         }
-        const int quantizerChangedBit = alpha ? AVIF_ENCODER_CHANGE_QUANTIZER_ALPHA : AVIF_ENCODER_CHANGE_QUANTIZER;
-        if (encoderChanges & quantizerChangedBit) {
+        const int qualityChangedBit = alpha ? AVIF_ENCODER_CHANGE_QUALITY_ALPHA : AVIF_ENCODER_CHANGE_QUALITY;
+        if (encoderChanges & qualityChangedBit) {
             if ((cfg->rc_end_usage == AOM_VBR) || (cfg->rc_end_usage == AOM_CBR)) {
                 // cq-level is ignored in these two end-usage modes, so adjust minQuantizer and
                 // maxQuantizer to the target quantizer.
@@ -985,7 +993,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         if (encoderChanges & AVIF_ENCODER_CHANGE_TILE_COLS_LOG2) {
             aom_codec_control(&codec->internal->encoder, AV1E_SET_TILE_COLUMNS, tileColsLog2);
         }
-        if (encoderChanges & quantizerChangedBit) {
+        if (encoderChanges & qualityChangedBit) {
             if ((cfg->rc_end_usage == AOM_CQ) || (cfg->rc_end_usage == AOM_Q)) {
                 aom_codec_control(&codec->internal->encoder, AOME_SET_CQ_LEVEL, quantizer);
             }
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -438,6 +438,12 @@ static int avmScaleQuantizer(int quantizer, uint32_t depth)
     return AVIF_CLAMP((quantizer * 255 + 31) / 63, 0, 255);
 }
 
+static int avmQualityToQuantizer(int quality, uint32_t depth)
+{
+    int quantizer = ((100 - quality) * 63 + 50) / 100;
+    return quantizer;
+}
+
 static avifBool avmCodecEncodeFinish(avifCodec * codec, avifCodecEncodeOutput * output);
 
 static avifResult avmCodecEncodeImage(avifCodec * codec,
@@ -446,7 +452,7 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
                                       avifBool alpha,
                                       int tileRowsLog2,
                                       int tileColsLog2,
-                                      int quantizer,
+                                      int quality,
                                       avifEncoderChanges encoderChanges,
                                       avifBool disableLaggedOutput,
                                       avifAddImageFlags addImageFlags,
@@ -454,6 +460,7 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
 {
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
+    int quantizer = avmQualityToQuantizer(quality, image->depth);
 
     // For encoder->scalingMode.horizontal and encoder->scalingMode.vertical to take effect in AV2
     // encoder, config should be applied for each frame, so we don't care about changes on these
@@ -706,8 +713,8 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
                 quantizerUpdated = AVIF_TRUE;
             }
         }
-        const int quantizerChangedBit = alpha ? AVIF_ENCODER_CHANGE_QUANTIZER_ALPHA : AVIF_ENCODER_CHANGE_QUANTIZER;
-        if (encoderChanges & quantizerChangedBit) {
+        const int qualityChangedBit = alpha ? AVIF_ENCODER_CHANGE_QUALITY_ALPHA : AVIF_ENCODER_CHANGE_QUALITY;
+        if (encoderChanges & qualityChangedBit) {
             if ((cfg->rc_end_usage == AOM_VBR) || (cfg->rc_end_usage == AOM_CBR)) {
                 // cq-level is ignored in these two end-usage modes, so adjust minQuantizer and
                 // maxQuantizer to the target quantizer.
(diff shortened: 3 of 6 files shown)


include/avif/avif.h
```diff
@@ -1551,10 +1551,6 @@ typedef struct avifEncoder
     // 1. Each quantizer in 0..63 can be converted to quality and back to itself.
     // 2. The lossless quality 100 is the only quality that is converted to the
     //    lossless quantizer 0, and vice versa.
-    int minQuantizer;      // Deprecated, use `quality` instead.
-    int maxQuantizer;      // Deprecated, use `quality` instead.
-    int minQuantizerAlpha; // Deprecated, use `qualityAlpha` instead.
-    int maxQuantizerAlpha; // Deprecated, use `qualityAlpha` instead.
```
y-guyon, line 1557: This is an API/ABI breaking change. We should try our best to avoid that.

What about keeping them functional in the API and in `avifenc`, marking them as deprecated in the comments, and converting them internally the same way `avifenc`'s `quantizerToQuality()` does it?

src/write.c
```diff
@@ -221,18 +221,18 @@ typedef struct avifEncoderData
...
     int tileRowsLog2;
     int tileColsLog2;
     avifEncoder lastEncoder;
-    // lastQuantizer and lastQuantizerAlpha are the quantizer and quantizerAlpha values used last
+    // lastQuality and lastQualityAlpha are the quality and qualityAlpha values used last
     // time
-    int lastQuantizer;
-    int lastQuantizerAlpha;
+    int lastQuality;
+    int lastQualityAlpha;
```
wantehchang, line 236: By analogy we should add `lastQualityGainMap`. We should look into this.

include/avif/avif.h
wantehchang: Yes. include/avif/avif.h is a public header. The types defined in this header either cannot be modified or can only be extended carefully. The `avifEncoder` struct falls in the latter category. We can extend it (by adding new fields to the end of the struct) This is documented in the comment at line 1493:

```
// AVIF encoder struct. It may be extended in a future release. Code outside the libavif library
// must allocate avifEncoder by calling the avifEncoderCreate() function, and destroy it with
// avifEncoderDestroy().
```

src/codec_aom.c
```diff
@@ -373,6 +373,33 @@ static avifBool avifKeyEqualsName(const char * key, const char * name, avifBool
...
 }
 
+static const struct aomOptionEnumList tuningEnum[] = { //
+    { "psnr", AOM_TUNE_PSNR },                         //
+    { "ssim", AOM_TUNE_SSIM },                         //
+#if defined(AOM_HAVE_TUNE_IQ)
+    { "iq", AOM_TUNE_IQ },
+#endif
+    { NULL, 0 }
+};
```
wantehchang: Please revert the change to the `tuningEnum` array. Not that it is only used with older versions of libaom that don't have the `aom_codec_set_option()` function (the `HAVE_AOM_CODEC_SET_OPTION macro is not defined).

src/codec_aom.c
```diff
@@ -557,20 +578,63 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
+}
+
 static avifResult aomCodecEncodeImage(avifCodec * codec,
                                       avifEncoder * encoder,
                                       const avifImage * image,
                                       avifBool alpha,
                                       int tileRowsLog2,
                                       int tileColsLog2,
-                                      int quantizer,
+                                      int quality,
```
wantehchang, line 573: We may not need to add the `quality` parameter, because it is readily available in `encoder->quality` or `encoder->qualityAlpha` (selected by the `alpha` parameter`). Please look into this.

include/avif/avif.h
juliobbv-p: Thanks @y-guyon @wantehchang. I wasn't sure if we were planning for a major version bump, so I preemptively made my edits optimistically.

I'll go ahead and preserve the `min/maxQuantizer` types.

src/codec_aom.c
juliobbv-p: Will do. Note that I'm using this array in `avifAOMOptionsIsTuneIq()` as a mapping to detect whether `tune=iq` was set by these methods:
- Passing in `-a tune=AOM_TUNE_IQ` (i.e. the enum value)
- Passing in `-a tune=iq` (i.e. the string)
- Yannis's PR: libavif sets `tune=iq` for the "color" channel

I can modify `avifAOMOptionsIsTuneIq()` to explicitly check for all `tune=iq` permutations if that's clearer.

juliobbv-p: Actually, if there's a pattern for me to use to tell the tuning mode with `HAVE_AOM_CODEC_SET_OPTION`, I'll go ahead with that instead.

src/codec_aom.c
juliobbv-p: I think it's useful to keep `quality` as a parameter, as the caller (in `write.c`) has already determined the correct quality value for the layer that's being encoded.

Otherwise we'd need to push down this piece of logic to each of the four possible encoders:

```c
    int quality = isAlpha ? encoder->data->qualityAlpha
        : (item->itemCategory == AVIF_ITEM_GAIN_MAP) ? encoder->data->qualityGainMap
        : encoder->data->quality;
```

include/avif/avif.h
juliobbv-p:
```
// AVIF encoder struct. It may be extended in a future release. Code outside the libavif library
// must allocate avifEncoder by calling the avifEncoderCreate() function, and destroy it with
// avifEncoderDestroy().
```

@wantehchang just for my education: do we want to guarantee preservation of deprecated fields in public headers even across major version boundaries (i.e. 2.0)?

src/write.c
juliobbv-p: It does seem unusual that there wasn't a `lastQuantizerGainMap` in the first place. @y-guyon do you know if  the omission was intentional?

juliobbv-p:
@wantehchang @y-guyon PR is updated to remove the tune=iq stuff and restore the min/maxQuantizer fields. The tune=iq formula PR will follow up tomorrow.

src/write.c
```diff
@@ -1800,21 +1797,21 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
     }
 
     // -----------------------------------------------------------------------
-    // Map quality settings to quantizer values.
-    encoder->data->quantizer = avifQualityToQuantizer(encoder->quality, encoder->minQuantizer, encoder->maxQuantizer);
+    // Populate encoder data (color) quality, qualityAlpha and .
```
y-guyon: ... qualityGainMap?

src/write.c
y-guyon: No. Maybe gain maps are only supported for single images, so there is no need to store `lastQuantizerGainMap`? @maryla-uc do you know?

maryla-uc: Gain maps are indeed only for still images.

src/codec_aom.c
juliobbv-p: For the final implementation, I left the `tuningEnum` array in its original place, and just created an extra one just for the tune=iq string -> enum mapping.

src/write.c
```diff
@@ -221,18 +221,19 @@ typedef struct avifEncoderData
...
     int tileColsLog2;
     avifEncoder lastEncoder;
-    // lastQuantizer and lastQuantizerAlpha are the quantizer and quantizerAlpha values used last
+    // lastQuality and lastQualityAlpha are the quality and qualityAlpha values used last
     // time
-    int lastQuantizer;
-    int lastQuantizerAlpha;
+    int lastQuality;
+    int lastQualityAlpha;
+    // Note: Gain maps are supported for still images only, so we don't need a "lastQualityGainMap"
```
wantehchang: Nit: I was going to suggest adding this comment. Thank you! Please move this comment to line 233:

```
    // lastQuality and lastQualityAlpha are the quality and qualityAlpha values used last
    // time
    // Note: Gain maps are supported for still images only, so we don't need a "lastQualityGainMap"
```

wantehchang:
Yannis: Thank you for the review!

Julio: I am taking a close look now.

src/codec_aom.c
```diff
@@ -557,20 +557,31 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
 
 static avifBool aomCodecEncodeFinish(avifCodec * codec, avifCodecEncodeOutput * output);
 
+static int aomQualityToQuantizer(int quality)
+{
+    int quantizer;
+
+    quality = AVIF_CLAMP(quality, 0, 100);
```
wantehchang: Delete the clamping. The caller is responsible for clamping `quality`.

Note: Make the same change to the other src/codec_*.c files.

src/codec_aom.c
wantehchang: I understand why the `quality` parameter is needed now. Thanks for the explanation. The caller also converts the default value to an actual value, especially for alpha and gain map.

src/codec_aom.c
```diff
@@ -557,20 +557,31 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
-                                      int quantizer,
+                                      int quality,
                                       avifEncoderChanges encoderChanges,
                                       avifBool disableLaggedOutput,
                                       avifAddImageFlags addImageFlags,
                                       avifCodecEncodeOutput * output)
 {
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
+    int quantizer = aomQualityToQuantizer(quality);
```
wantehchang: Nit: Add `const`.

Note: Make the same change to the other src/codec_*.c files, if applicable.

src/codec_avm.c
```diff
@@ -438,6 +438,24 @@ static int avmScaleQuantizer(int quantizer, uint32_t depth)
...
+
+    if (depth == 10) {
+        return (((100 - quality) * (255 + 48) + 50) / 100) - 48;
+    }
+    if (depth == 12) {
+        return (((100 - quality) * (255 + 96) + 50) / 100) - 96;
+    }
+    assert(depth == 8);
+    return ((100 - quality) * 255 + 50) / 100;
+}
```
wantehchang, line 445: IMPORTANT: To preserve the current behavior, this function should do a standard conversion:

```
static int avmQualityToQuantizer(int quality, uint32_t depth)
{
    int quantizer = ((100 - quality) * 63 + 50) / 100;
    return quantizer;
}
```

The reason is that the caller will scale the return value of this function:

```
        // Scale from aom's [0:63] to avm's [0:255]. TODO(yguyon): Accept [0:255] directly in avifEncoder.
        quantizer = avmScaleQuantizer(quantizer, image->depth);
```

src/codec_avm.c
```diff
@@ -438,6 +438,24 @@ static int avmScaleQuantizer(int quantizer, uint32_t depth)
     return AVIF_CLAMP((quantizer * 255 + 31) / 63, 0, 255);
 }
 
+// Scales from quality [0:100] to avm's quantizer [M:255], where M=0/-48/-96 for 8/10/12 bit.
+// See --min-qp help in
+// https://gitlab.com/AOMediaCodec/avm/-/blob/main/apps/aomenc.c
+// TODO(yguyon): Accept [M:255] directly in avifEncoder.
```
wantehchang: Delete the entire comment block. See the comment at line 457.

src/codec_rav1e.c
```diff
@@ -49,13 +49,23 @@ static avifBool rav1eSupports400(void)
     return minorVersion >= 4;
 }
 
+static int rav1eQualityToQuantizer(int quality)
```
wantehchang, line 53: Add a comment:

```
// rav1e's QP range is [0,255]
```

src/codec_rav1e.c
```diff
@@ -49,13 +49,23 @@ static avifBool rav1eSupports400(void)
...
+{
+    int quantizer;
+
+    quality = AVIF_CLAMP(quality, 0, 100);
+    quantizer = ((100 - quality) * 255 + 50) / 100;
```
wantehchang: The new conversion is more accurate than the original conversion, so this will change the encoded output for some quality values. I think this is okay.

src/write.c
```diff
@@ -221,18 +221,19 @@ typedef struct avifEncoderData
 {
     avifEncoderItemArray items;
     avifEncoderFrameArray frames;
-    // Map the encoder settings quality and qualityAlpha to quantizer and quantizerAlpha
-    int quantizer;
-    int quantizerAlpha;
-    int quantizerGainMap;
+    // quality values for the image, alpha, and gain map
```
wantehchang, line 224: Since the `avifEncoder` struct also has these three fields, a natural question is why the `avifEncoderData` struct needs these three fields. I think the answer is similar to the comment below for `tileRowsLog2` and `tileColsLog2` -- these three fields are the actual quality values after the default values are resolved. Please clarify this point in the comment.

src/write.c
```diff
@@ -1204,19 +1205,16 @@ static avifResult avifImageCopyAndPad(avifImage * const dstImage, const avifImag
...
     if (quality == AVIF_QUALITY_DEFAULT) {
         // In older libavif releases, avifEncoder didn't have the quality and qualityAlpha fields.
-        // Supply a default value for quantizer.
+        // Supply a default value for quality.
         quantizer = (minQuantizer + maxQuantizer) / 2;
-        quantizer = AVIF_CLAMP(quantizer, 0, 63);
-    } else {
-        quality = AVIF_CLAMP(quality, 0, 100);
-        quantizer = ((100 - quality) * 63 + 50) / 100;
+        quality = ((63 - quantizer) * 100 + 31) / 63;
```
wantehchang, line 1217: To avoid behavior change, let's preserve the clamping in the original code. I think the if statement should be changed to the following (comments omitted for brevity):

```
    if (quality == AVIF_QUALITY_DEFAULT) {
        quantizer = (minQuantizer + maxQuantizer) / 2;
        quantizer = AVIF_CLAMP(quantizer, 0, 63);
        quality = ((63 - quantizer) * 100 + 31) / 63;
    } else {
        quality = AVIF_CLAMP(quality, 0, 100);
    }
```

wantehchang (approved):
LGTM. I review the code carefully. I suggest some simple changes.

src/codec_aom.c
```diff
@@ -557,20 +557,31 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
 
 static avifBool aomCodecEncodeFinish(avifCodec * codec, avifCodecEncodeOutput * output);
 
+static int aomQualityToQuantizer(int quality)
```
wantehchang, line 560: Optional: Rename this function `avifQualityToQuantizer()` and move it to include/avif/internal.h as a static inline function.

include/avif/internal.h
```diff
@@ -544,8 +544,8 @@ typedef enum avifEncoderChange
     AVIF_ENCODER_CHANGE_MAX_QUANTIZER_ALPHA = (1 << 3),
     AVIF_ENCODER_CHANGE_TILE_ROWS_LOG2 = (1 << 4),
     AVIF_ENCODER_CHANGE_TILE_COLS_LOG2 = (1 << 5),
-    AVIF_ENCODER_CHANGE_QUANTIZER = (1 << 6),
-    AVIF_ENCODER_CHANGE_QUANTIZER_ALPHA = (1 << 7),
+    AVIF_ENCODER_CHANGE_QUALITY = (1 << 6),
+    AVIF_ENCODER_CHANGE_QUALITY_ALPHA = (1 << 7),
```
wantehchang, line 548: Please apply the following patch to this file:

```
diff --git a/include/avif/internal.h b/include/avif/internal.h
index fd04bcfc..58a1fcf3 100644
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -565,8 +565,8 @@ typedef avifBool (*avifCodecGetNextImageFunc)(struct avifCodec * codec,
 // encoder->tileRowsLog2, encoder->tileColsLog2, and encoder->autoTiling. The caller of
 // avifCodecEncodeImageFunc is responsible for automatic tiling if encoder->autoTiling is set to
 // AVIF_TRUE. The actual tiling values are passed to avifCodecEncodeImageFunc as parameters.
-// Similarly, avifCodecEncodeImageFunc should use the quantizer parameter instead of
-// encoder->quality and encoder->qualityAlpha. If disableLaggedOutput is AVIF_TRUE, then the encoder will emit the output frame
+// Similarly, avifCodecEncodeImageFunc should use the quality parameter instead of
+// encoder->quality, encoder->qualityAlpha, and encoder->qualityGainMap. If disableLaggedOutput is AVIF_TRUE, then the encoder will emit the output frame
 // without any lag (if supported). Note that disableLaggedOutput is only used by the first call to this function (which
 // initializes the encoder) and is ignored by the subsequent calls.
 //
@@ -581,7 +581,7 @@ typedef avifResult (*avifCodecEncodeImageFunc)(struct avifCodec * codec,
                                                avifBool alpha,
                                                int tileRowsLog2,
                                                int tileColsLog2,
-                                               int quantizer,
+                                               int quality,
                                                avifEncoderChanges encoderChanges,
                                                avifBool disableLaggedOutput,
```

src/write.c
juliobbv-p: Yes, good call. Early on I had introduced a bug I didn't know `avifEncoderData` had resolved quality values.

src/codec_aom.c
juliobbv-p: I could add it to `include/avif/internal.h`, however I think it'd be actually be better if the name stays as `aomQualityToQuantizer()`, because AVM, rav1e, and now SVT (post 3.1.2) have different QP scaling and granularity.

wantehchang: You are right. If this function is not shared by multiple codecs, we should not move it to include/avif/internal.h.

src/codec_avm.c
```diff
@@ -438,6 +438,12 @@ static int avmScaleQuantizer(int quantizer, uint32_t depth)
     return AVIF_CLAMP((quantizer * 255 + 31) / 63, 0, 255);
 }
 
+static int avmQualityToQuantizer(int quality, uint32_t depth)
+{
+    int quantizer = ((100 - quality) * 63 + 50) / 100;
```
wantehchang, line 443: Optional: Add `const` in a follow-up CL.

src/write.c
wantehchang: I also had to study the code to refresh my memory. (I created a copy of your pull request at https://aomedia-review.googlesource.com/c/libavif/+/205142 so that I could use Gerrit's superior code review tool. It was worth the extra work.)

PR #2921, opened 2025-12-09
Add tune=iq quality to QP formula for libaom

juliobbv-p:
Add a "quality to quantizer (QP)" mapping formula designed for libaom's `tune=iq` AKA `AOM_TUNE_IQ`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Always forward the CICP color primaries, transfer characteristics,
   and matrix coefficients to the AV1 encoder, which writes them in the Sequence
   Header OBU, for compatibility with libraries that wrongly ignore the colr box.
+* Use a "quality to quantizer (QP)" mapping formula designed for AOM_TUNE_IQ.
 
 ### Removed since 1.3.0
 
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -389,6 +389,35 @@ static avifBool avifProcessAOMOptionsPreInit(avifCodec * codec, avifBool alpha, 
     return AVIF_TRUE;
 }
 
+static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
+{
+    avifBool ret = AVIF_FALSE;
+
+#if !defined(AOM_HAVE_TUNE_IQ)
+// Define the tune IQ value here if libaom doesn't define it. The enum value is guaranteed to never change
+// in libaom, so this definition won't ever get out of sync.
+#define AOM_TUNE_IQ 10
+#endif
+
+    // Tune IQ string -> enum mapping
+    static const struct aomOptionEnumList tuneIqEnum[] = {
+        { "iq", AOM_TUNE_IQ },
+    };
+
+    for (uint32_t i = 0; i < codec->csOptions->count; ++i) {
+        const avifCodecSpecificOption * entry = &codec->csOptions->entries[i];
+        int val;
+        // If there are multiple "tune" options specified, honor the last one.
+        // For consistent behavior, handle both cases where tune IQ was either specified as a string (tune=iq),
+        // or as an enum value (tune=10).
+        if (avifKeyEqualsName(entry->key, "tune", alpha) && aomOptionParseEnum(entry->value, tuneIqEnum, &val)) {
+            ret = (val == AOM_TUNE_IQ);
+        }
+    }
+
+    return ret;
+}
+
 #if !defined(HAVE_AOM_CODEC_SET_OPTION)
 typedef enum
 {
@@ -555,11 +584,47 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
     return lumaPels <= levelLumaPels && width <= levelWidth * levelDimMult && height <= levelHeight * levelDimMult;
 }
 
+// Quality (q) to quantizer (qp) formula for tune=iq (Image Quality), expressed as a look-up table for more clarity.
+// The x axis of the table represents the ones digit, while the y axis represents the tens digit
+// of the quality value [0-100], which is then mapped to a quantizer value [0-63].
+// The formula below is a piecewise linear function. Each segment was empirically selected to correct for the
+// non-linear bitrate increase from encoding with tune=iq (relative to tune=ssim), with a goal to make tune=iq behave
+// like tune=ssim at matching quality levels, with an overall smaller (but still predictable) file size and a similar
+// to better quality.
+// The qp of tune=ssim <= qp of tune=iq for all quality values because tune=iq needs a larger qp to reduce file size.
+// - [ 0 -  6]: quantizer = 63 - round(quality / 3)
+// - [ 7 - 28]: quantizer = 61 - round((quality - 6) / 2)
+// - [29 - 54]: quantizer = 50 - round((quality - 29) * 6 / 10)
+// - [55 - 99]: quantizer = 34 - round((quality - 54) * 34 / 45) + 1
+// - [    100]: quantizer = 0
+// clang-format off
+static const int tuneIqQualityToQuantizer[101] = {
+// 1s digit: *0  *1  *2  *3  *4  *5  *6  *7  *8  *9     10s digit:
+             63, 63, 63, 62, 62, 62, 61, 61, 60, 60, // 0*
+             59, 59, 58, 58, 57, 57, 56, 56, 55, 55, // 1*
+             54, 54, 53, 53, 52, 52, 51, 51, 50, 50, // 2*
+             49, 49, 48, 48, 47, 46, 46, 45, 44, 44, // 3*
+             43, 43, 42, 41, 41, 40, 40, 39, 39, 38, // 4*
+             38, 37, 37, 36, 35, 34, 34, 33, 32, 31, // 5*
+             30, 30, 29, 28, 28, 27, 26, 25, 25, 24, // 6*
+             23, 22, 22, 21, 20, 19, 19, 18, 17, 16, // 7*
+             15, 15, 14, 13, 12, 12, 11, 10,  9,  9, // 8*
+              8,  7,  6,  6,  5,  4,  3,  3,  2,  1, // 9*
+              0  // quality 100
+};
+// clang-format on
+
 static avifBool aomCodecEncodeFinish(avifCodec * codec, avifCodecEncodeOutput * output);
 
-static int aomQualityToQuantizer(int quality)
+static int aomQualityToQuantizer(int quality, avifBool isTuneIq)
 {
-    const int quantizer = ((100 - quality) * 63 + 50) / 100;
+    int quantizer;
+
+    if (isTuneIq) {
+        quantizer = tuneIqQualityToQuantizer[quality];
+    } else {
+        quantizer = ((100 - quality) * 63 + 50) / 100;
+    }
 
     return quantizer;
 }
@@ -578,7 +643,8 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
 {
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
-    const int quantizer = aomQualityToQuantizer(quality);
+    const avifBool tuneIq = avifImageUsesTuneIq(codec, alpha);
+    const int quantizer = aomQualityToQuantizer(quality, tuneIq);
 
     // For encoder->scalingMode.horizontal and encoder->scalingMode.vertical to take effect in AOM
     // encoder, config should be applied for each frame, so we don't care about changes on these
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -440,7 +440,8 @@ static int avmScaleQuantizer(int quantizer, uint32_t depth)
 
 static int avmQualityToQuantizer(int quality, uint32_t depth)
 {
-    int quantizer = ((100 - quality) * 63 + 50) / 100;
+    const int quantizer = ((100 - quality) * 63 + 50) / 100;
+
     return quantizer;
 }
 


juliobbv-p:
@wantehchang @y-guyon FYI here's the PR for the new quality to QP tune=iq libaom formula. The PR depends on !2920, so review just the last commit.

src/codec_aom.c
```diff
@@ -389,6 +389,29 @@ static avifBool avifProcessAOMOptionsPreInit(avifCodec * codec, avifBool alpha,
...
+    // Tune IQ string -> enum mapping
+    static const struct aomOptionEnumList tuneIqEnum[] = {
+        { "iq", AOM_TUNE_IQ },
+    };
+
+    for (uint32_t i = 0; i < codec->csOptions->count; ++i) {
+        const avifCodecSpecificOption * entry = &codec->csOptions->entries[i];
+        int val;
+        if (avifKeyEqualsName(entry->key, "tune", alpha)) {
+            if (aomOptionParseEnum(entry->value, tuneIqEnum, &val)) {
```
wantehchang: The numeric value of `AOM_TUNE_IQ` is 10. We only need to call `aomOptionParseEnum()` if we need to support both "tune=iq" and "tune=10". Otherwise, we can simplify this function as follows:

```
static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
{
#if defined(AOM_HAVE_TUNE_IQ)
    for (uint32_t i = 0; i < codec->csOptions->count; ++i) {
        const avifCodecSpecificOption * entry = &codec->csOptions->entries[i];
        if (avifKeyEqualsName(entry->key, "tune", alpha) && !strcmp(entry->value, "iq")) {
            return AVIF_TRUE;
        }
    }
#endif
    return AVIF_FALSE;
}
```
I checked libaom. It supports both "tune=iq" and "tune=10". But libavif can choose to make it simple and only support "tune=iq". Or libavif can play safe and support both.

If we want to support both, I think the new `tuneIqEnum` array seems better (than extending the existing `tuningEnum` array) because we don't need to waste time checking for "tune=psnr" and "tune=ssim".

src/codec_aom.c
```diff
@@ -389,6 +389,29 @@ static avifBool avifProcessAOMOptionsPreInit(avifCodec * codec, avifBool alpha,
...
+                if (val == AOM_TUNE_IQ) {
+                    return AVIF_TRUE;
```
wantehchang: Optional: There is a corner case that we should try to handle: there may be multiple `tune` options in the `codec->csOptions` array. A common convention is that the last `tune` option is honored, which means the early return here would be inappropriate.

To implement the common convention, we would need to do something like the following (assuming we don't need to support "tune=10"):

```
static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
{
    avifBool ret = AVIF_FALSE;
#if defined(AOM_HAVE_TUNE_IQ)
    for (uint32_t i = 0; i < codec->csOptions->count; ++i) {
        const avifCodecSpecificOption * entry = &codec->csOptions->entries[i];
        // If there are multiple "tune" options, honor the last one.
        if (avifKeyEqualsName(entry->key, "tune", alpha)) {
            ret = !strcmp(entry->value, "iq");
        }
    }
#endif
    return ret;
}
```

src/codec_aom.c
juliobbv-p: Yeah, I had in mind supporting both "tune=iq" and "tune=10" for the least-surprising UX behavior. Especially if you're coming from the SVT-AV1 world where tuning modes are always enum values, it might be a confusing experience to not treat the enum value the same as the string in libavif.

src/codec_aom.c
juliobbv-p: Applied your suggestion, while adapting it to keep "tune=10" support.

src/codec_aom.c
```diff
@@ -389,6 +389,26 @@ static avifBool avifProcessAOMOptionsPreInit(avifCodec * codec, avifBool alpha,
...
+#if defined(AOM_HAVE_TUNE_IQ)
+    // Tune IQ string -> enum mapping
+    static const struct aomOptionEnumList tuneIqEnum[] = {
+        { "iq", AOM_TUNE_IQ },
+    };
+
+    for (uint32_t i = 0; i < codec->csOptions->count; ++i) {
+        const avifCodecSpecificOption * entry = &codec->csOptions->entries[i];
+        int val;
+        if (avifKeyEqualsName(entry->key, "tune", alpha) && aomOptionParseEnum(entry->value, tuneIqEnum, &val)) {
```
wantehchang, line 413: Please add a comment to explain why we are doing this because this is not obvious. You can start with this suggestion:

```
// If there are multiple "tune" options, honor the last one.
```

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The changes are relative to the previous release, unless the baseline is specifi
   AVIF specification.
 * Add an optional argument to the --depth flag of avifenc used to enable a bit
   depth extension scheme in the encoded file.
+* Add a "quality to quantizer (QP)" mapping formula designed for AOM_TUNE_IQ.
```
wantehchang: Optional: This probably should go into the "Changed since 1.3.0" section instead because this isn't added to the public interface.

src/codec_aom.c
```diff
@@ -581,7 +637,8 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
 {
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
-    int quantizer = aomQualityToQuantizer(quality);
+    avifBool tuneIq = avifImageUsesTuneIq(codec, alpha);
+    int quantizer = aomQualityToQuantizer(quality, tuneIq);
```
wantehchang: Nit: Add `const` to both variables (`tuneIq` and `quantizer`).

src/codec_aom.c
```diff
@@ -555,14 +578,47 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
+// like tune=ssim at matching quality levels, with an overall smaller (but still predictable) file size and a similar
+// to better quality.
+// - [ 0 -  6]: quantizer = 63 - round(quality / 3)
+// - [ 7 - 28]: quantizer = 61 - round((quality - 6) / 2)
+// - [29 - 54]: quantizer = 50 - round((quality - 29) * 6 / 10)
+// - [55 - 99]: quantizer = 34 - round((quality - 54) * 34 / 45) + 1
+// - [    100]: quantizer = 0
+// clang-format off
+static const int tuneIqQualityToQuantizer[101] = {
+//1s digit: *0  *1  *2  *3  *4  *5  *6  *7  *8  *9.    10s digit:
```
wantehchang: Nit: The dot `.` after `*9` doesn't seem necessary

src/codec_aom.c
```diff
@@ -555,14 +578,47 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
+// of the quality value [0-100], which is then mapped to a quantizer value [0-63].
+// The formula below is a piecewise linear function. Each segment was empirically selected to correct for the
+// non-linear bitrate increase from encoding with tune=iq (relative to tune=ssim), with a goal to make tune=iq behave
+// like tune=ssim at matching quality levels, with an overall smaller (but still predictable) file size and a similar
+// to better quality.
+// - [ 0 -  6]: quantizer = 63 - round(quality / 3)
+// - [ 7 - 28]: quantizer = 61 - round((quality - 6) / 2)
+// - [29 - 54]: quantizer = 50 - round((quality - 29) * 6 / 10)
+// - [55 - 99]: quantizer = 34 - round((quality - 54) * 34 / 45) + 1
+// - [    100]: quantizer = 0
```
wantehchang, line 599: Please document that `qp of tune=ssim <= qp of tune=iq` for all quality values because tune=iq needs the larger qp to reduce file size (or, at the same qp, tune=iq produces larger file size).

I verified the `qp of tune=ssim <= qp of tune=iq` property with the following program:

```
#include <stdio.h>

static const int tuneIqQualityToQuantizer[101] = {
            63, 63, 63, 62, 62, 62, 61, 61, 60, 60, // 0*
            59, 59, 58, 58, 57, 57, 56, 56, 55, 55, // 1*
            54, 54, 53, 53, 52, 52, 51, 51, 50, 50, // 2*
            49, 49, 48, 48, 47, 46, 46, 45, 44, 44, // 3*
            43, 43, 42, 41, 41, 40, 40, 39, 39, 38, // 4*
            38, 37, 37, 36, 35, 34, 34, 33, 32, 31, // 5*
            30, 30, 29, 28, 28, 27, 26, 25, 25, 24, // 6*
            23, 22, 22, 21, 20, 19, 19, 18, 17, 16, // 7*
            15, 15, 14, 13, 12, 12, 11, 10,  9,  9, // 8*
             8,  7,  6,  6,  5,  4,  3,  3,  2,  1, // 9*
             0  // quality 100
};

static int aomQualityToQuantizer(int quality)
{
    int quantizer;

    return quantizer;
}

int main(void) {
  for (int quality = 0; quality < 101; quality++) {
    int qp_ssim = aomQualityToQuantizer(quality);
    int qp_iq = tuneIqQualityToQuantizer[quality];
    if (qp_ssim > qp_iq) {
      printf("%d %d %d\n", quality, qp_ssim, qp_iq);
    }
  }
  return 0;
}
```

wantehchang (approved):

src/codec_aom.c
```diff
@@ -555,22 +578,67 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
+            54, 54, 53, 53, 52, 52, 51, 51, 50, 50, // 2*
+            49, 49, 48, 48, 47, 46, 46, 45, 44, 44, // 3*
+            43, 43, 42, 41, 41, 40, 40, 39, 39, 38, // 4*
+            38, 37, 37, 36, 35, 34, 34, 33, 32, 31, // 5*
+            30, 30, 29, 28, 28, 27, 26, 25, 25, 24, // 6*
+            23, 22, 22, 21, 20, 19, 19, 18, 17, 16, // 7*
+            15, 15, 14, 13, 12, 12, 11, 10,  9,  9, // 8*
+             8,  7,  6,  6,  5,  4,  3,  3,  2,  1, // 9*
+             0  // quality 100
+};
```
wantehchang, line 614: Future: Once we are happy with this lookup table, we can consider publishing it as a source code comment in some libaom public header, so that it is distributed with libaom. (It probably should become a qp rescaling lookup table when it is distributed with libaom because libaom doesn't use a 0..100 quality level.)

src/codec_aom.c
juliobbv-p: Thanks for the independent verification!

src/codec_aom.c
juliobbv-p: MacOS moment 🤣

wantehchang:
Julio: Please rebase this pull request and I will take another look. Thanks!

src/codec_aom.c
wantehchang: The more you can share about how this lookup table was generated, the better. You can do that in a follow-up pull request.

Also, in addition to pointing out that qp of tune=ssim <= qp of tune=iq for all quality levels, it would be good to say roughly by how much in each of the quality intervals, and why the difference is bigger or smaller in a particular interval. Such insights of yours are extremely valuable and should be recorded.

src/codec_aom.c
wantehchang: Although I am also a macOS user, I don't get this joke. Please enlighten me...

juliobbv-p:
> Julio: Please rebase this pull request and I will take another look. Thanks!

@wantehchang done, thanks!

src/codec_aom.c
juliobbv-p: Sounds good, I'll add more information in a follow-up PR.

The tricky part is that I need to convey that the "qp of tune=ssim <= qp of tune=iq" differences came from a pragmatic trade-off of multiple effects in play, although it's just a case where I just need to be careful with my wording.

src/codec_aom.c
juliobbv-p: This is an interesting idea. The challenge is that for re-scaling to work well, we need at least one of these conditions to be fulfilled:
- Input level range should have at least ~50% more possible values so two or three input levels can map to one single qp, while not omitting any qp levels e.g. in the lookup table above, quality levels 0, 1 and 2 all map to qp 63
- Output level range should be a qindex instead of a qp, to make quantization differences more well even-spaced between adjacent input levels

src/codec_avm.c
```diff
@@ -460,7 +460,7 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
 {
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
-    int quantizer = avmQualityToQuantizer(quality, image->depth);
+    const int quantizer = avmQualityToQuantizer(quality, image->depth);
```
wantehchang: IMPORTANT: Remove this `const`. Instead, add it to the `quantizer` local variable inside `avmQualityToQuantizer()` (line 443).

This variable will be modified at 589 below:

```
        // Scale from aom's [0:63] to avm's [0:255]. TODO(yguyon): Accept [0:255] directly in avifEncoder.
        quantizer = avmScaleQuantizer(quantizer, image->depth);
```

src/codec_aom.c
```diff
@@ -389,6 +389,35 @@ static avifBool avifProcessAOMOptionsPreInit(avifCodec * codec, avifBool alpha,
     return AVIF_TRUE;
 }
 
+static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
+{
+    avifBool ret = AVIF_FALSE;
+#if !defined(AOM_HAVE_TUNE_IQ)
+    (void)codec;
+    (void)alpha;
+#endif
```
wantehchang: I just realize that we should do it like this:

```
#if !defined(AOM_HAVE_TUNE_IQ)
#define AOM_TUNE_IQ 10
#endif
```

And the delete the `#if defined(AOM_HAVE_TUNE_IQ)` below. The reason is that this allows libavif to be compiled against an older version of libaom shared library but works automatically when the libaom shared library is upgraded to a new version that supports tune=iq.

src/codec_aom.c
```diff
@@ -555,11 +584,48 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
+
 static avifBool aomCodecEncodeFinish(avifCodec * codec, avifCodecEncodeOutput * output);
 
-static int aomQualityToQuantizer(int quality)
+static int aomQualityToQuantizer(int quality, avifBool isTuneIq)
 {
-    const int quantizer = ((100 - quality) * 63 + 50) / 100;
+    int quantizer;
+
+    quality = AVIF_CLAMP(quality, 0, 100);
```
wantehchang: Delete the clamping. It's handled by the caller.

juliobbv-p: My bad, my rebase brought it back.

src/codec_aom.c
```diff
@@ -555,11 +584,48 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
     return lumaPels <= levelLumaPels && width <= levelWidth * levelDimMult && height <= levelHeight * levelDimMult;
 }
 
+// Quality (q) to quantizer (qp) formula for tune=iq (Image Quality), expressed as a look-up table for more clarity.
+// The x axis of the table represents the ones digit, while the y axis represents the tens digit
+// of the quality value [0-100], which is then mapped to a quantizer value [0-63].
+// The formula below is a piecewise linear function. Each segment was empirically selected to correct for the
+// non-linear bitrate increase from encoding with tune=iq (relative to tune=ssim), with a goal to make tune=iq behave
+// like tune=ssim at matching quality levels, with an overall smaller (but still predictable) file size and a similar
```
wantehchang, line 592: Nit/optional: "behave like tune=ssim" could be misinterpreted to mean tune=iq operates like tune=ssim, but I think you are referring to the encoding time? Would be good to clarify. I will also think about how to rephrase this.

juliobbv-p: Yeah, I see what you mean. In this case, it's both encoding and decoding time. Images tend to encode and decode faster as the qp increases, and tune=iq images are more complex than tune=ssim at the same frame qp due to CDEF usage, so the qp rescaling counteracts this effect.

PR #2924, opened 2025-12-11
Bump gha-setup-ninja and vcpkg-action in CI

vrabaud:
The CI is failing because of these two actions not being updated


vrabaud:
Actually, GitHub might have some hiccups.

PR #2926, opened 2025-12-13
avmQualityToQuantizer: remove unused `depth` param

wantehchang:


src/codec_avm.c
```diff
@@ -438,7 +438,7 @@ static int avmScaleQuantizer(int quantizer, uint32_t depth)
     return AVIF_CLAMP((quantizer * 255 + 31) / 63, 0, 255);
 }
 
-static int avmQualityToQuantizer(int quality, uint32_t depth)
+static int avmQualityToQuantizer(int quality)
```
y-guyon, line 441: This should be merged with `avmScaleQuantizer()` in another PR.

wantehchang: Done in https://github.com/AOMediaCodec/libavif/pull/2928. Thanks for the suggestion.

PR #2927, opened 2025-12-13
Tweak tune=iq quality to QP formula

juliobbv-p:
Tweak the formula to include another useful property: "The step size (cadence) of the computed quantizers monotonically decreases as quality increases (from 3 to 1)".

Additionally, simplify formulas as much as possible while preserving every existing useful property.

Thanks @FrankGalligan for the suggestion!

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -585,29 +585,37 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
 }
 
 // Quality (q) to quantizer (qp) formula for tune=iq (Image Quality), expressed as a look-up table for more clarity.
-// The x axis of the table represents the ones digit, while the y axis represents the tens digit
-// of the quality value [0-100], which is then mapped to a quantizer value [0-63].
 // The formula below is a piecewise linear function. Each segment was empirically selected to correct for the
-// non-linear bitrate increase from encoding with tune=iq (relative to tune=ssim), with a goal to make tune=iq behave
-// like tune=ssim at matching quality levels, with an overall smaller (but still predictable) file size and a similar
-// to better quality.
-// The qp of tune=ssim <= qp of tune=iq for all quality values because tune=iq needs a larger qp to reduce file size.
-// - [ 0 -  6]: quantizer = 63 - round(quality / 3)
-// - [ 7 - 28]: quantizer = 61 - round((quality - 6) / 2)
-// - [29 - 54]: quantizer = 50 - round((quality - 29) * 6 / 10)
-// - [55 - 99]: quantizer = 34 - round((quality - 54) * 34 / 45) + 1
-// - [    100]: quantizer = 0
+// non-linear bitrate increase from encoding content with tune=iq relative to tune=ssim with the same qp.
+//
+// | Quality | Quantizer                          | Step size |
+// |---------|------------------------------------|-----------|
+// |  0 -  6 | 63 - floor(quality / 3)            |         3 |
+// |  7 - 28 | 61 - round((quality - 7) / 2)      |         2 |
+// | 29 - 53 | 50 - round((quality - 29) * 3 / 5) |      1.66 |
+// | 54 - 99 | 35 - round((quality - 54) * 3 / 4) |      1.33 |
+// |     100 | 0 (lossless)                       |         1 |
+//
+// The formula has these properties, in addition to the general conversion formula properties described in avif.h:
+// - Encoding and decoding time with tune=iq are closer to tune=ssim's at a given quality level, with an
+//   overall smaller (but still predictable) file size and a similar to better quality
+// - The qp of tune=ssim <= qp of tune=iq for all quality values
+// - Quality 60 (the default in avifenc) = qp 30
+// - The step size of the quantizers monotonically decreases as quality increases (from 3 to 1)
+//
+// The x axis of the table represents the ones digit, while the y axis represents the tens digit
+// of the q value [0-100], which is then mapped to a qp value [0-63].
 // clang-format off
 static const int tuneIqQualityToQuantizer[101] = {
 // 1s digit: *0  *1  *2  *3  *4  *5  *6  *7  *8  *9     10s digit:
              63, 63, 63, 62, 62, 62, 61, 61, 60, 60, // 0*
              59, 59, 58, 58, 57, 57, 56, 56, 55, 55, // 1*
              54, 54, 53, 53, 52, 52, 51, 51, 50, 50, // 2*
-             49, 49, 48, 48, 47, 46, 46, 45, 44, 44, // 3*
-             43, 43, 42, 41, 41, 40, 40, 39, 39, 38, // 4*
-             38, 37, 37, 36, 35, 34, 34, 33, 32, 31, // 5*
-             30, 30, 29, 28, 28, 27, 26, 25, 25, 24, // 6*
-             23, 22, 22, 21, 20, 19, 19, 18, 17, 16, // 7*
+             49, 49, 48, 48, 47, 46, 46, 45, 45, 44, // 3*
+             43, 43, 42, 42, 41, 40, 40, 39, 39, 38, // 4*
+             37, 37, 36, 36, 35, 34, 33, 33, 32, 31, // 5*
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -585,29 +585,37 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
              54, 54, 53, 53, 52, 52, 51, 51, 50, 50, // 2*
-             49, 49, 48, 48, 47, 46, 46, 45, 44, 44, // 3*
-             43, 43, 42, 41, 41, 40, 40, 39, 39, 38, // 4*
-             38, 37, 37, 36, 35, 34, 34, 33, 32, 31, // 5*
-             30, 30, 29, 28, 28, 27, 26, 25, 25, 24, // 6*
-             23, 22, 22, 21, 20, 19, 19, 18, 17, 16, // 7*
+             49, 49, 48, 48, 47, 46, 46, 45, 45, 44, // 3*
+             43, 43, 42, 42, 41, 40, 40, 39, 39, 38, // 4*
+             37, 37, 36, 36, 35, 34, 33, 33, 32, 31, // 5*
+             30, 30, 29, 28, 27, 27, 26, 25, 24, 24, // 6*
```
y-guyon, line 617: At a glance, default quality seems to be [51](https://github.com/AOMediaCodec/libavif/blob/5dedb0b20c577bfbfac08da80d246f864c6bb3df/src/write.c#L1215-L1217) in libavif API and [60](https://github.com/AOMediaCodec/libavif/blob/5dedb0b20c577bfbfac08da80d246f864c6bb3df/apps/avifenc.c#L868) in avifenc. This PR does not seem to optimize for significantly steeper changes near these two values. On the other hand this is a conservative change so it should be safe.

src/codec_aom.c
```diff
@@ -585,29 +585,37 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
-// - [    100]: quantizer = 0
+// non-linear bitrate increase from encoding content with tune=iq relative to tune=ssim with the same qp.
+//
+// | Quality | Quantizer                          | Cadence |
+// |---------|------------------------------------|---------|
+// |  0 -  6 | 63 - floor(quality / 3)            |       3 |
+// |  7 - 28 | 61 - round((quality - 7) / 2)      |       2 |
+// | 29 - 53 | 50 - round((quality - 29) * 3 / 5) |     1.5 |
+// | 54 - 99 | 35 - round((quality - 54) * 3 / 4) |    1.33 |
+// |     100 |                                    |       1 |
```
wantehchang: The quantizer column should say "0" in the row for quality=100.

src/codec_aom.c
```diff
@@ -585,29 +585,37 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
+//
+// The formula has the following properties:
+// - Encoding and decoding time with tune=iq are closer to tune=ssim's at a given quality level, with an
+//   overall smaller (but still predictable) file size and a similar to better quality
+// - The qp of tune=ssim <= qp of tune=iq for all quality values
+// - Quality 60 (the default in libavif) = qp 30
+// - The cadence of the quantizers monotonically decreases as quality increases (from 3 to 1)
```
wantehchang: Add: Only quality 100 is mapped to quantizer 0 (lossless).

But this is a requirement on any quality-to-quantizer conversion formula. So it probably should not be mentioned in this list of properties.

src/codec_aom.c
```diff
@@ -585,29 +585,37 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
-// to better quality.
-// The qp of tune=ssim <= qp of tune=iq for all quality values because tune=iq needs a larger qp to reduce file size.
-// - [ 0 -  6]: quantizer = 63 - round(quality / 3)
-// - [ 7 - 28]: quantizer = 61 - round((quality - 6) / 2)
-// - [29 - 54]: quantizer = 50 - round((quality - 29) * 6 / 10)
-// - [55 - 99]: quantizer = 34 - round((quality - 54) * 34 / 45) + 1
-// - [    100]: quantizer = 0
+// non-linear bitrate increase from encoding content with tune=iq relative to tune=ssim with the same qp.
+//
+// | Quality | Quantizer                          | Cadence |
```
wantehchang: Optional: Change "cadence" to "step size"

src/codec_aom.c
```diff
@@ -585,29 +585,37 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
+// |---------|------------------------------------|---------|
+// |  0 -  6 | 63 - floor(quality / 3)            |       3 |
+// |  7 - 28 | 61 - round((quality - 7) / 2)      |       2 |
+// | 29 - 53 | 50 - round((quality - 29) * 3 / 5) |     1.5 |
+// | 54 - 99 | 35 - round((quality - 54) * 3 / 4) |    1.33 |
```
wantehchang: The new formula fixed a nit in the old formula that I complained about. The offset in the old formula is either the right endpoint of the previous segment or the left endpoint of the current segment. The offset in the new formua is always the left endpoint of the current segment, so it looks nicer.

src/codec_aom.c
juliobbv-p: Yeah, most of the steeper changes happen at the lower quality levels due to the non-linear nature of the bitrate increase with tune=iq compared to tune=ssim. At q levels around the default, step size happens to be very similar to the old formula.

src/codec_aom.c
juliobbv-p: Yep, I took some time to add Frank's property, while also making formulas look nicer. Multipliers and divisors are simpler too!

src/codec_aom.c
juliobbv-p: Yeah, I tried to refrain to not add the shared quality-to-quantizer properties to not make the comment too long. Quality 100 -> quantizer 0 will be implied by the formula definition anyway.

wantehchang: Quality 100 -> quantizer 0 is just half of the requirement. The other half is that no other qualities are mapped to quantizer 0. That is not obvious from the formula 35 - round((quality - 54) * 3 / 4).

I think the best solution is to expose the declaration of the lookup table in include/avif/internal.h and add a unit test that validates the requirements and desired properties.

wantehchang: Julio: The unit test suggestion is for future consideration. No need to add that test in this pull request.

src/codec_aom.c
```diff
@@ -585,29 +585,37 @@ static avifBool doesLevelMatch(int width, int height, int levelWidth, int levelH
...
+// |  7 - 28 | 61 - round((quality - 7) / 2)      |         2 |
+// | 29 - 53 | 50 - round((quality - 29) * 3 / 5) |       1.5 |
+// | 54 - 99 | 35 - round((quality - 54) * 3 / 4) |      1.33 |
+// |     100 | 0 (lossless)                       |         1 |
+//
+// The formula has the following properties:
+// - Encoding and decoding time with tune=iq are closer to tune=ssim's at a given quality level, with an
+//   overall smaller (but still predictable) file size and a similar to better quality
+// - The qp of tune=ssim <= qp of tune=iq for all quality values
+// - Quality 60 (the default in libavif) = qp 30
```
wantehchang: Nit: libavif => avifenc

wantehchang (approved):

src/codec_aom.c
juliobbv-p: Right, for that we do have the explanation in `avif.h` that applies to all conversion formulas:

```c
    // The conversion formulas have two nice properties:
    // 1. Each quantizer in 0..63 can be converted to quality and back to itself.
    // 2. The lossless quality 100 is the only quality that is converted to the
    //    lossless quantizer 0, and vice versa.
```

I'll add a reference of the baseline properties in the `tuneIqQualityToQuantizer[]` comment, just to make sure those additional properties aren't mistaken as a replacement.

We can add tests for the lookup table, but I propose we should do it for all quality-to-quantizer formulas, as it's as hard (if not harder) to catch property violations due to changes when the conversion formula is expressed as-is vs. as a lookup table.

PR #2928, opened 2025-12-13
Convert quality to avm's quantizer range directly

wantehchang:
We still need avmScaleQuantizer() as a separate function to scale
encoder->min/maxQuantizer to avm's [M:255] range.

Delete a TODO comment. Delete a duplicate comment that refers to the
[0:255] range.

### DIFF
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -425,7 +425,6 @@ static avifBool avifFindAOMScalingMode(const avifFraction * avifMode, AOM_SCALIN
 // Scales from aom's [0:63] to avm's [M:255], where M=0/-48/-96 for 8/10/12 bit.
 // See --min-qp help in
 // https://gitlab.com/AOMediaCodec/avm/-/blob/main/apps/aomenc.c
-// TODO(yguyon): Accept [M:255] directly in avifEncoder.
 static int avmScaleQuantizer(int quantizer, uint32_t depth)
 {
     if (depth == 10) {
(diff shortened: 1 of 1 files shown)


src/codec_avm.c
wantehchang: Yannis: Should we remove this TODO comment?

y-guyon: Probably. I guess the libavif API will only expose quality in the [0:100] range.

src/codec_avm.c
```diff
@@ -438,11 +438,11 @@ static int avmScaleQuantizer(int quantizer, uint32_t depth)
     return AVIF_CLAMP((quantizer * 255 + 31) / 63, 0, 255);
 }
 
-static int avmQualityToQuantizer(int quality)
+static int avmQualityToQuantizer(int quality, uint32_t depth)
 {
     const int quantizer = ((100 - quality) * 63 + 50) / 100;
```
y-guyon: My point was to avoid losing granularity here. We start with 101 quality levels, scale that down to 64 distinct "aom" quantizer levels in [0:63], and end with 64 distinct "avm" quantizer levels in the [M:255] range. Skipping the middle step would map the quality to 101 distinct quantizer levels in the [M:255], which offer more granularity to libavif users. That was unclear in the other PR, sorry.

juliobbv-p: So, in my very first attempt, my approach went straight from converting quality -> avm quantizer to avoid the granularity issue that comes from having an intermediate aom quantizer + rescaling step.

There's a caveat though: because with the current design, min/max quantizers still need to be rescaled from aom -> avm, you'll end up in a situation where the quantizer is converted with a finer granularity, but min/max quantizers with a coarser granularity, which feels unintuitive.

My question is: if min/max quantizers have already been deprecated in libavif, should the library just completely ignore those fields for avm? That'd make the design and code of avm quant derivation significantly simpler.

wantehchang: Julio: Hmm... You are right. libavif should completely ignore min/max quantizers for AVM. In addition, src/codec_avm.c should return an error if min/max quantizers are different from their default values.

src/codec_avm.c
wantehchang: Done.

src/codec_avm.c
wantehchang: Yannis: I changed avmQualityToQuantizer() as you suggested. Please take a look. Thanks!

juliobbv-p:
@wantehchang thanks for working on this change!

PR #2935, opened 2025-12-16
Set tuning before aom codec options

y-guyon:
Add test.


y-guyon:
Fixes https://github.com/AOMediaCodec/libavif/pull/2830#issuecomment-3056403684.

maryla-uc (approved):
Thanks!

PR #2936, opened 2025-12-16
Move default tune setting earlier in codec_aom.c

y-guyon:
Also make sure tune is NOT reset to libavif default when explicitly unset by the user.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -420,6 +420,10 @@ static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
         }
     }
 
+    // In practice this function should also return true if avifEncoderSetCodecSpecificOption("tune", "iq")
+    // was called for a previous frame and not called (or called with NULL) for this frame, because the tune
+    // option persists across frames in libaom. However AOM_TUNE_IQ is only supported with still images in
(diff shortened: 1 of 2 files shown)


src/codec_aom.c
```diff
@@ -420,6 +420,10 @@ static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
         }
     }
 
+    // In practice this function should also return true if avifEncoderSetCodecSpecificOption("tune", "iq")
+    // was called for a previous frame and not called (or called with NULL) for this frame, because the tune
+    // option persists across frames in libaom. However AOM_TUNE_IQ is only supported with still images in
+    // libavif and libaom as of today, so there is no need to remember this option across frames.
```
vrabaud, line 426: Do we have a test that says we cannot set tune=iq for two frames ? In case libaom changes behavior.

vrabaud (approved):

src/codec_aom.c
juliobbv-p: Small clarification: libaom **can** use tune IQ for sequences, but only in all intra mode. But libavif uses good mode for videos, so from the perspective of libavif, the assumption still holds true.

PR #2949, opened 2026-01-05
Use AOM_TUNE_PSNR by default for alpha

y-guyon:


apps/avifenc.c
```diff
@@ -320,7 +320,8 @@ static void syntaxLong(void)
         printf("    enable-chroma-deltaq=B            : Enable delta quantization in chroma planes. 0=disable (default), 1=enable\n");
         printf("    end-usage=MODE                    : Rate control mode, one of 'vbr', 'cbr', 'cq', or 'q'\n");
         printf("    sharpness=S                       : Bias towards block sharpness in rate-distortion optimization of transform coefficients in 0..7. (Default: 0)\n");
-        printf("    tune=METRIC                       : Tune the encoder for distortion metric, one of 'psnr', 'ssim' (default for aom) or 'iq'.\n");
+        printf("    tune=METRIC                       : Tune the encoder for distortion metric, one of 'psnr', 'ssim' or 'iq'.\n");
+        printf("                                        (Default for aom codec: YUV: ssim, alpha: psnr)\n");
```
wantehchang: Nit: Change "YUV" to "color". Although "color" is not ideal, it is the term we use to refer to the main image item.

wantehchang: Nit: Omit "for aom codec" because this entire section is aom-specific. See line 307:

```
        printf("aom-specific advanced options:\n");
```

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ The changes are relative to the previous release, unless the baseline is specifi
   Header OBU, for compatibility with libraries that wrongly ignore the colr box.
 * Use a "quality to quantizer (QP)" mapping formula designed for AOM_TUNE_IQ.
 * Set tuning before applying the user-provided specific aom codec options.
+* Use AOM_TUNE_PSNR by default when encoding alpha with libaom.
```
wantehchang: Optional: I think it would be good to mention the reason (AOM_TUNE_SSIM causes ringing for alpha).

PR #2951, opened 2026-01-07
Set diagnostic message on aom_codec_decode failure

wantehchang:
Improve the error reporting for bug b:471949427.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -148,6 +148,10 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
         } else if (sample) {
             codec->internal->iter = NULL;
             if (aom_codec_decode(&codec->internal->decoder, sample->data.data, sample->data.size, NULL)) {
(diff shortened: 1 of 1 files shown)


wantehchang:
Before this change, the `avifdec` error message is:

```
ERROR: Failed to decode image: Decoding of color planes failed
Diagnostics:
```

After this change, the `avifdec` error message is:

```
ERROR: Failed to decode image: Decoding of color planes failed
Diagnostics:
 * aom_codec_decode() failed: Bitstream not supported by this decoder: (null)
```

If we additionally apply the CL https://aomedia-review.googlesource.com/c/aom/+/205601 to libaom, the  `avifdec` error message becomes:

```
ERROR: Failed to decode image: Decoding of color planes failed
Diagnostics:
 * aom_codec_decode() failed: Bitstream not supported by this decoder: Value 20 of seq_level_idx[0] is not yet defined
```

Issue #2950, opened 2026-01-06
avifenc fails to generate correct HDR metadata when encoding HDR PNG files


realnc:
When encoding an HDR png, the resulting AVIF file lacks the correct color primaries and transfer function fields. They are not being copied from the PNG.

I'm attaching an example HDR PNG. It uses:
(These are shown by exiftool.)

```none
Color Primaries                 : BT.2020, BT.2100
Transfer Characteristics        : SMPTE ST 2084, ITU BT.2100 PQ
Matrix Coefficients             : Identity matrix
```

Encoding it with `avifenc Test.png Out.avif` results in an AVIF file that uses:

```none
Color Profiles                  : nclx
Color Primaries                 : Unspecified
Transfer Characteristics        : Unspecified
Matrix Coefficients             : BT.601
```

The ICC profile of the PNG is copied, but the NCLX stuff is not. Encoding with `--ignore-icc` does not help, since then the AVIF file shows these:

```none
Color Profiles                  : nclx
Color Primaries                 : BT.709
Transfer Characteristics        : sRGB or sYCC
Matrix Coefficients             : BT.601
```

Which is even worse since the AVIF is now an SDR file, not HDR.

Stripping the PNG from it's ICC profile (with `exiftool -icc_profile= Test.png`) and then encoding that with avifenc gives another result, different from the two previous ones:

```none
Color Profiles                  : nclx
Color Primaries                 : BT.2020, BT.2100
Transfer Characteristics        : Unspecified
Matrix Coefficients             : BT.601
```

Which of course is close, but no cigar since `Transfer Characteristics` is missing. (I don't know what the importance of the `Matrix Coefficients` field is.)

Example PNG to reproduce the issue:

https://github.com/user-attachments/assets/0a2f5cf5-b7e6-4452-acf8-037f748023a4

The correct behavior would be to respect the NCLX metadata of the PNG and use it in the generated AVIF. As it stands right now, the AVIF can't be viewed correctly in some image viewers, or it shows but generates error messages about incorrect/unsupported color primaries and transfer function.

Kaldaien:
I would add to this that all of those characteristics are defined in the PNG's CICP chunk. They are being ignored, even though AVIF has a 1:1 exact mapping between those things. The CICP data should be preserved and used verbatim, it would completely eliminate this problem.

wantehchang:
Maryla: Thank you for working on this issue.

@veluca93 Luca: FYI.

[...]

PR #2952, opened 2026-01-07
Add support for png cICP chunk

maryla-uc:
Simplify the image reading code by removing the allowChangingCicp parameter.
This is handled at the call site in avifenc.c instead.

#2950

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -210,14 +210,24 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool 
     return AVIF_TRUE;
 }
 
-// Callback for unknown chunks. Used to detect chunks that we are missing support for.
-// Could also be used to handle those chunks (not done right now for simplicity).
+typedef struct avifUnknownCicpChunkData
+{
+    avifImage * avif;
+    avifBool cicpChunkRead;
+} avifUnknownCicpChunkData;
+
+// Callback for unknown chunks. Used to detect/handle chunks that we are missing support for.
 static int avifPNGReadUnknownChunk(png_structp png_ptr, png_unknown_chunkp chunk)
 {
-    // If needed, the argument can be retrieved with: (MyType*)png_get_user_chunk_ptr(png_ptr);
-    (void)png_ptr;
+    avifUnknownCicpChunkData * data = (avifUnknownCicpChunkData *)png_get_user_chunk_ptr(png_ptr);
+    avifImage * avif = data->avif;
     if (!memcmp(chunk->name, "cICP\0", 5)) {
-        fprintf(stderr, "Warning: Unsupported cICP chunk. libpng version >= 1.6.45 needed.\n");
+        if (chunk->size > 2) {
+            data->cicpChunkRead = AVIF_TRUE;
+            avif->colorPrimaries = chunk->data[0];
+            avif->transferCharacteristics = chunk->data[1];
+            // The matrix coefficient and full range flag are ignored.
+        }
     }
     return 1;
 }
@@ -283,7 +293,8 @@ static avifBool avifPNGReadImpl(FILE * f,
         goto cleanup;
     }
 
-    png_set_read_user_chunk_fn(png, NULL, &avifPNGReadUnknownChunk);
+    avifUnknownCicpChunkData unknownChunkData = { avif, AVIF_FALSE };
+    png_set_read_user_chunk_fn(png, &unknownChunkData, &avifPNGReadUnknownChunk);
 
     png_init_io(png, f);
     png_set_sig_bytes(png, 8);
@@ -383,7 +394,9 @@ static avifBool avifPNGReadImpl(FILE * f,
             // The matrix coefficient and full range flag are ignored: they are about YUV encoding and the PNG values are irrelevant.
         } else
 #endif
-            if (png_get_iCCP(png, info, &iccpProfileName, &iccpCompression, &iccpData, &iccpDataLen) == PNG_INFO_iCCP) {
+            if (unknownChunkData.cicpChunkRead) {
+            // Already handled in avifPNGReadUnknownChunk()
+        } else if (png_get_iCCP(png, info, &iccpProfileName, &iccpCompression, &iccpData, &iccpDataLen) == PNG_INFO_iCCP) {
             if (!rawColorTypeIsGray && avif->yuvFormat == AVIF_PIXEL_FORMAT_YUV400) {
                 fprintf(stderr,
                         "The image contains a color ICC profile which is incompatible with the requested output "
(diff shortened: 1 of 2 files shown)


apps/shared/avifpng.c
```diff
@@ -210,6 +210,18 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
     return AVIF_TRUE;
 }
 
+// Callback for unknown chunks. Used to detect chunks that we are missing support for.
+// Could also be used to handle those chunks (not done right now for simplicity).
+static int avifPNGReadUnknownChunk(png_structp png_ptr, png_unknown_chunkp chunk)
+{
+    // If needed, the argument can be retreived with: (MyType*)png_get_user_chunk_ptr(png_ptr);
```
y-guyon: retrieved

apps/shared/avifpng.c
```diff
@@ -210,6 +210,18 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
+    (void)png_ptr;
+    if (!memcmp(chunk->name, "cICP\0", 5)) {
+        fprintf(stderr, "Warning: Unsupported cICP chunk. libpng version >= 1.6.45 needed.\n");
```
y-guyon: Could libavif parse that `cICP` chunk itself?

apps/shared/avifpng.c
```diff
@@ -351,13 +364,25 @@ static avifBool avifPNGReadImpl(FILE * f,
...
+        //   If a single image contains more than one of these chunk types, the chunk with the lowest
+        //   Priority number should take precedence and any higher-numbered chunk types should be ignored.
+#if defined(PNG_cICP_SUPPORTED)
+        png_byte cicpColorPrimaries;
+        png_byte cicpTransferFunction;
+        png_byte cicpMatrixCoefficients;
+        png_byte cicpVideoFullRangeFlag;
+        if (png_get_cICP(png, info, &cicpColorPrimaries, &cicpTransferFunction, &cicpMatrixCoefficients, &cicpVideoFullRangeFlag)) {
+            avif->colorPrimaries = cicpColorPrimaries;
+            avif->transferCharacteristics = cicpTransferFunction;
```
y-guyon, line 393: Matrix and full range are ignored?

maryla-uc: Yes. Added a comment.

apps/shared/avifpng.c
maryla-uc: Yes we could, as commented above ("Could also be used to handle those chunks (not done right now for simplicity).")
But I thought it was unnecessary and I didn't want to open the door for handling various chunks ourselves. Although admittedly it should be pretty easy for this chunk. Let me know if you think I should handle it here.

y-guyon: My bad for not reading the comment.

It sounds like this is fairly easy for this chunk and would simplify other parts of libavif (testing etc.).

apps/shared/avifpng.c
y-guyon: I wonder why PNG carries these values then, but I guess this is a deeper issue.

maryla-uc: From the spec: "RGB is currently the only supported color model in PNG, and as such Matrix Coefficients shall be set to 0." Maybe one day png could support YUV (unlikely?) but even then, how the PNG stored its YUV is I think somewhat irrelevant to how we choose to store our YUV, except if we wanted to directly copy the samples directly like we do for some jpegs.

The "Video Full Range Flag" could in theory be zero but I think this is also unlikely to be used in the near future.

apps/shared/avifpng.c
maryla-uc: Done

apps/shared/avifpng.c
```diff
@@ -210,6 +210,28 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
+    avifImage * avif;
+    avifBool cicpChunkRead;
+} avifUnknownCicpChunkData;
+
+// Callback for unknown chunks. Used to detect/handle chunks that we are missing support for.
+static int avifPNGReadUnknownChunk(png_structp png_ptr, png_unknown_chunkp chunk)
+{
+    avifUnknownCicpChunkData * data = (avifUnknownCicpChunkData *)png_get_user_chunk_ptr(png_ptr);
+    avifImage * avif = data->avif;
+    if (!memcmp(chunk->name, "cICP\0", 5)) {
```
wantehchang, line 224: The title of this pull request is "Add support for png iCCP chunk". Should we change "iCCP" to "cICP"?

wantehchang, line 224: Nit: The explicit `\0` in the string literal "cICP\0" is not necessary because there is an implicit `\0` at the end of the string literal. See this experiment:

```
$ cat nulchar.c 
#include <stdio.h>

int main(void) {
  printf("%zu\n", sizeof("cICP"));
  printf("%zu\n", sizeof("cICP\0"));
  return 0;
}
$ cc -Wall -Wextra nulchar.c
$ ./a.out
5
6
```

It is okay if you wanted the explicit `\0` for clarity.

apps/shared/avifpng.c
```diff
@@ -210,6 +210,28 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
...
+        if (chunk->size > 2) {
+            data->cicpChunkRead = AVIF_TRUE;
+            avif->colorPrimaries = chunk->data[0];
+            avif->transferCharacteristics = chunk->data[1];
+            // The matrix coefficient and full range flag are ignored.
```
wantehchang, line 229: Please add the reason for ignoring the matrix coefficients and full range flag. Perhaps we can copy the comment at line 394?

```
            // The matrix coefficient and full range flag are ignored: they are about YUV encoding and the PNG values are irrelevant.
```

apps/shared/avifpng.c
wantehchang: I remember there was a discussion about the support for a zero "Video Full Range Flag" on some W3C mailing list related to PNG. I will try to look that up.

wantehchang: I am not 100% sure, but I think this PNG issue is what I had in mind: https://github.com/w3c/png/issues/312#issuecomment-2325281113

(The reason I am not sure is that I seem to remember there was a discussion in the past few months.)

I also found a blog post relevant to this issue: https://svgees.us/blog/cICP.html#the-other-two-numbers

apps/shared/avifpng.c
maryla-uc: Ha indeed I made a mistake in the title. But I think it's hard to change now?

y-guyon: The title of the PR can be changed, but not the commit.

PR #2955, opened 2026-01-08
Refactor/clarify cICP support.

maryla-uc:
Add changelog entry.

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -228,9 +228,9 @@ static void avifSetPNGCicp(avifImage * avif,
     //  RGB is currently the only supported color model in PNG, and as such Matrix Coefficients shall be set to 0.
     // Note that we don't set avif->matrixCoefficients to this value as the Avif's YUV matrix is independent from the PNG's.
     if (cicpMatrixCoefficients != 0) {
-        fprintf(stderr, "Warning: Unsupported PNG CICP matrix coefficient value %d. Expected to be 0.\n", cicpMatrixCoefficients);
+        fprintf(stderr, "Warning: Unsupported PNG CICP matrix coefficients value %d. Expected to be 0.\n", cicpMatrixCoefficients);
     }
-    // Narrow range PNG files are uncommon and would require conversion to full range which we don't support for now for simplicity.
+    // Limited range PNG files are uncommon and would require conversion to full range which we don't support for now for simplicity.
     // See also https://github.com/w3c/png/issues/312#issuecomment-2325281113 and https://svgees.us/blog/cICP.html#the-other-two-numbers
     // Similarly, we don't set avif->yuvRange to this value as the Avif's YUV range is independent from the PNG's.
     if (cicpVideoFullRangeFlag != 1) {


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Update libyuv.cmd/LocalLibyuv.cmake: deeb764bb (1922)
 * Update LocalRav1e.cmake: cargo-c v0.10.14, corrosion v0.5.2,
 * Update zlibpng.cmd/LocalZlibpng.cmake: libpng 1.6.51, zlib 1.3.1.2
+* Add support for PNG cICP chunk when decoding PNG files.
```
y-guyon: Could be useful to mention that cICC may be ignored now.

y-guyon (approved):
Nice!

apps/shared/avifpng.c
```diff
@@ -216,17 +216,37 @@ typedef struct avifUnknownCicpChunkData
...
+                           png_byte cicpMatrixCoefficients,
+                           png_byte cicpVideoFullRangeFlag)
+{
+    avif->colorPrimaries = cicpColorPrimaries;
+    avif->transferCharacteristics = cicpTransferCharacteristics;
+    // PNG specification Third Edition Section 4.3:
+    //  RGB is currently the only supported color model in PNG, and as such Matrix Coefficients shall be set to 0.
+    // Note that we don't set avif->matrixCoefficients to this value as the Avif's YUV matrix is independent from the PNG's.
+    if (cicpMatrixCoefficients != 0) {
+        fprintf(stderr, "Warning: Unsupported PNG CICP matrix coefficient value %d. Expected to be 0.\n", cicpMatrixCoefficients);
```
wantehchang: Nit: coefficient => coefficients (plural)

apps/shared/avifpng.c
```diff
@@ -216,17 +216,37 @@ typedef struct avifUnknownCicpChunkData
...
+    }
+    // Narrow range PNG files are uncommon and would require conversion to full range which we don't support for now for simplicity.
```
wantehchang: Nit: Narrow => Limited

"Limited range" is the term we use in libavif.

wantehchang (approved):
LGTM. Thanks!

apps/shared/avifpng.c
```diff
@@ -216,17 +216,37 @@ typedef struct avifUnknownCicpChunkData
...
+}
+
 // Callback for unknown chunks. Used to detect/handle chunks that we are missing support for.
 static int avifPNGReadUnknownChunk(png_structp png_ptr, png_unknown_chunkp chunk)
 {
     avifUnknownCicpChunkData * data = (avifUnknownCicpChunkData *)png_get_user_chunk_ptr(png_ptr);
     avifImage * avif = data->avif;
-    if (!memcmp(chunk->name, "cICP\0", 5)) {
-        if (chunk->size > 2) {
+    if (!memcmp(chunk->name, "cICP", 4)) {
```
wantehchang, line 246: IMPORTANT: The last argument should still be 5. Sorry I missed this earlier.

The string literal "cICP" contains 5 bytes, with an implicit NUL character at the end.

We can still use "cICP\0" (which contains 6 bytes) for clarity if you consider this an esoteric detail of the C programming language.

maryla-uc: I know you didn't ask for the change from 5 to 4 but the documentation of the chunk->name field says "Textual chunk name with '\0' terminator" so the 5th char is always \0 and therefore it's not necessary to check it. Moreover I thought it looked confusing to have a memcmp call with 5 but 4 char string.
Otherwise I can just use strcmp but it needs a type cast from png_byte to char otherwise you get "warning: pointer targets in passing argument 1 of ‘strcmp’ differ in signedness".
In any case I don't think it's worth spending too much time on.

wantehchang: Thanks for the reply. I agree it is correct to pass either 4 or 5 to the `memcmp()` call.

PR #2959, opened 2026-01-09
Bump avm dependency to tag research-v13.0.0

y-guyon:


src/codec_avm.c
```diff
@@ -395,36 +396,36 @@ static avifBool avifProcessAOMOptionsPostInit(avifCodec * codec, avifBool alpha)
...
+            *avmMode = scalingModeMap[i].avmMode;
             return AVIF_TRUE;
         }
     }
 
     return AVIF_FALSE;
 }
 
-// Scales from aom's [0:63] to avm's [M:255], where M=0/-48/-96 for 8/10/12 bit.
+// Scales from avm's [0:63] to avm's [M:255], where M=0/-48/-96 for 8/10/12 bit.
```
FooIbar:
```suggestion
// Scales from aom's [0:63] to avm's [M:255], where M=0/-48/-96 for 8/10/12 bit.
```

src/codec_avm.c
```diff
@@ -395,36 +396,36 @@ static avifBool avifProcessAOMOptionsPostInit(avifCodec * codec, avifBool alpha)
...
 // See --min-qp help in
-// https://gitlab.com/AOMediaCodec/avm/-/blob/main/apps/aomenc.c
+// https://gitlab.com/AVMediaCodec/avm/-/blob/main/apps/avmenc.c
```
FooIbar:
```suggestion
// https://gitlab.com/AOMediaCodec/avm/-/blob/main/apps/avmenc.c
```

src/codec_avm.c
```diff
@@ -439,7 +440,7 @@ static int avmScaleQuantizer(int quantizer, uint32_t depth)
 
 // Converts quality to avm's quantizer in the range of [M:255], where M=0/-48/-96 for 8/10/12 bit.
 // See --min-qp help in
-// https://gitlab.com/AOMediaCodec/avm/-/blob/main/apps/aomenc.c
+// https://gitlab.com/AVMediaCodec/avm/-/blob/main/apps/avmenc.c
```
FooIbar:
```suggestion
// https://gitlab.com/AOMediaCodec/avm/-/blob/main/apps/avmenc.c
```

src/codec_avm.c
```diff
@@ -544,18 +545,18 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
             // Set the maximum number of frames to encode to 1. This instructs
             // libavm to set still_picture and reduced_still_picture_header to
-            // 1 in AV1 sequence headers.
+            // 1 in AV2 sequence headers.
             cfg->g_limit = 1;
 
-            // Use the default settings of the new AOM_USAGE_ALL_INTRA (added in
-            // https://crbug.com/aomedia/2959).
+            // Use the default settings of the new AVM_USAGE_ALL_INTRA (added in
+            // https://crbug.com/avmedia/2959).
```
FooIbar:
```suggestion
            // https://crbug.com/aomedia/2959).
```

src/codec_avm.c
```diff
@@ -782,100 +783,100 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
-    // We prefer to simply set the aomImage.planes[] pointers to the plane buffers in 'image'. When
-    // doing this, we set aomImage.w equal to aomImage.d_w and aomImage.h equal to aomImage.d_h and
-    // do not "align" aomImage.w and aomImage.h. Unfortunately this exposes a libaom bug in libavm
-    // (https://crbug.com/aomedia/3113) if chroma is subsampled and image->width or image->height is
-    // equal to 1. To work around this libavm bug, we allocate the aomImage.planes[] buffers and
+    avm_image_t avmImage;
+    // We prefer to simply set the avmImage.planes[] pointers to the plane buffers in 'image'. When
+    // doing this, we set avmImage.w equal to avmImage.d_w and avmImage.h equal to avmImage.d_h and
+    // do not "align" avmImage.w and avmImage.h. Unfortunately this exposes a libavm bug in libavm
+    // (https://crbug.com/avmedia/3113) if chroma is subsampled and image->width or image->height is
```
FooIbar:
```suggestion
    // (https://crbug.com/aomedia/3113) if chroma is subsampled and image->width or image->height is
```

src/codec_avm.c
```diff
@@ -782,100 +783,100 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
+    // equal to 1. To work around this libavm bug, we allocate the avmImage.planes[] buffers and
     // copy the image YUV data if image->width or image->height is equal to 1. This bug has been
-    // fixed in libaom v3.1.3 but not in libavm.
+    // fixed in libavm v3.1.3 but not in libavm.
```
FooIbar:
```suggestion
    // fixed in libaom v3.1.3 but not in libavm.
```

CHANGELOG.md
```diff
@@ -32,7 +32,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Update libjpeg.cmd/LocalJpeg.cmake: 3.1.1
 * Update libxml2.cmd/LocalLibXml2.cmake: v2.14.4
 * Update aom.cmd/LocalAom.cmake: v3.13.1
-* Update LocalAom.cmake: AVM research-v12.0.0
+* Update LocalAvm.cmake: research-v13.0.0
```
wantehchang, line 35: Yannis: Thank you for writing this pull request. I will try to review it today.

CMakeLists.txt
```diff
@@ -503,6 +503,14 @@ if(AVIF_CODEC_SVT_ENABLED)
     set(AVIF_PKG_CONFIG_EXTRA_REQUIRES_PRIVATE "${AVIF_PKG_CONFIG_EXTRA_REQUIRES_PRIVATE} SvtAv1Enc")
 endif()
 
+if(AVIF_CODEC_AVM)
+    # Avoid the following error:
+    #   CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
+    #   Missing variable is: CMAKE_ASM_COMPILE_OBJECT
+    # For some reason this must be done before check_avif_option(AVIF_CODEC_AOM TARGET aom PKG_NAME aom)
+    enable_language(ASM)
```
wantehchang, line 512: We should try to track down why this is necessary.

@jzern James: Do you have any speculations?

cmake/Modules/LocalAvm.diff
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index b70f5a9f31..81b801d509 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1101,7 +1101,7 @@ if(NOT AVM_DIST_DIR)
+ endif()
+ 
+ add_custom_target(
+-  dist
++  avm_dist
```
wantehchang, line 10: @jzern @vrabaud 

I wonder if there is a way to import another project into a namespace in CMake. Otherwise we will need to audit all the `add_custom_target()` calls in the libaom and AVM CMake build systems and make sure the custom target names have the "aom_" or "avm_" prefix.

src/obu.c
```diff
@@ -554,7 +484,7 @@ static avifBool parseAV1SequenceHeader(avifBits * bits, avifSequenceHeader * hea
 }
 
 #if defined(AVIF_CODEC_AVM)
-// See https://gitlab.com/AOMediaCodec/avm/-/blob/main/av1/decoder/decodeframe.c
+// See https://gitlab.com/AOMediaCodec/avm/-/blob/research-v13.0.0/common/av2_config.c?ref_type=tags#L290
```
wantehchang: Unless we are sure that the code in common/av2_config.c is correct, it is better to follow the `read_sequence_header_obu()` function in av2/decoder/obu.c as an example.

src/obu.c
```diff
@@ -563,29 +493,72 @@ static avifBool parseAV2SequenceHeader(avifBits * bits, avifSequenceHeader * hea
...
 
-#if CONFIG_CROP_WIN_CWG_F220
-    AVIF_CHECK(parseAV2SequenceHeaderCroppingWindow(bits));
-#endif // CONFIG_CROP_WIN_CWG_F220
+    if (avifBitsRead(bits, 1)) { // conf_window_flag
+        avifBitsReadVLC(bits);   // conf_win_left_offset
+        avifBitsReadVLC(bits);   // conf_win_right_offset
+        avifBitsReadVLC(bits);   // conf_win_top_offset
+        avifBitsReadVLC(bits);   // conf_win_bottom_offset
+    }
```
wantehchang, line 487: This if statement (lines 515-520) corresponds to the following code in `read_sequence_header_obu()`:

```
  av2_read_conformance_window(rb, seq_params);
  av2_validate_seq_conformance_window(seq_params, &cm->error);
```

Afterwards, `read_sequence_header_obu()` does the following:

```
#if CONFIG_CWG_F270_CI_OBU
  av2_read_chroma_format_bitdepth(rb, seq_params, &cm->error);
#else
  av2_read_color_config(rb, seq_params, &cm->error);
#endif  // CONFIG_CWG_F270_CI_OBU
```

So we should have similar code here. If `CONFIG_CWG_F270_CI_OBU` is equal to 1, we should only parse the chroma format and the bit depth. We can use a structure similar to the following code in av2/decoder/decodeframe.c:

```
#if CONFIG_CWG_F270_CI_OBU
void av2_read_chroma_format_bitdepth(
    struct avm_read_bit_buffer *rb,
#else
void av2_read_color_config(
    struct avm_read_bit_buffer *rb,
#endif  // CONFIG_CWG_F270_CI_OBU
    SequenceHeader *seq_params, struct avm_internal_error_info *error_info) {
  const uint32_t seq_chroma_format_idc = avm_rb_read_uvlc(rb);
  set_seq_chroma_format(seq_chroma_format_idc, seq_params, error_info);

  read_bitdepth(rb, seq_params, error_info);

  const int is_monochrome = (seq_chroma_format_idc == CHROMA_FORMAT_400);
  seq_params->monochrome = is_monochrome;
#if !CONFIG_CWG_F270_CI_OBU
  int color_description_present_flag = avm_rb_read_bit(rb);
  if (color_description_present_flag) {
    seq_params->color_primaries = avm_rb_read_literal(rb, 8);
    seq_params->transfer_characteristics = avm_rb_read_literal(rb, 8);
    seq_params->matrix_coefficients = avm_rb_read_literal(rb, 8);
  } else {
    seq_params->color_primaries = AVM_CICP_CP_UNSPECIFIED;
    seq_params->transfer_characteristics = AVM_CICP_TC_UNSPECIFIED;
    seq_params->matrix_coefficients = AVM_CICP_MC_UNSPECIFIED;
  }
  if (is_monochrome) {
    // [16,235] (including xvycc) vs [0,255] range
    seq_params->color_range = avm_rb_read_bit(rb);
    seq_params->subsampling_y = seq_params->subsampling_x = 1;
    seq_params->chroma_sample_position = AVM_CSP_UNSPECIFIED;
  } else {
    if (seq_params->color_primaries == AVM_CICP_CP_BT_709 &&
        seq_params->transfer_characteristics == AVM_CICP_TC_SRGB &&
        seq_params->matrix_coefficients == AVM_CICP_MC_IDENTITY) {
      seq_params->subsampling_y = seq_params->subsampling_x = 0;
      seq_params->color_range = 1;  // assume full color-range
      if (!(seq_params->profile == PROFILE_1 ||
            (seq_params->profile == PROFILE_2 &&
             seq_params->bit_depth == AVM_BITS_12))) {
        avm_internal_error(
            error_info, AVM_CODEC_UNSUP_BITSTREAM,
            "sRGB colorspace not compatible with specified profile");
      }
    } else {
      // [16,235] (including xvycc) vs [0,255] range
      seq_params->color_range = avm_rb_read_bit(rb);
      if (seq_params->matrix_coefficients == AVM_CICP_MC_IDENTITY &&
          (seq_params->subsampling_x || seq_params->subsampling_y)) {
        avm_internal_error(
            error_info, AVM_CODEC_UNSUP_BITSTREAM,
            "Identity CICP Matrix incompatible with non 4:4:4 color sampling");
      }
      if (seq_params->subsampling_x && !seq_params->subsampling_y) {
        // YUV 4:2:2
        seq_params->chroma_sample_position = AVM_CSP_UNSPECIFIED;
        const int csp_present_flag = avm_rb_read_bit(rb);
        if (csp_present_flag) {
          seq_params->chroma_sample_position = avm_rb_read_bit(rb);
        }
      } else if (seq_params->subsampling_x && seq_params->subsampling_y) {
        // YUV 4:2:0
        seq_params->chroma_sample_position = AVM_CSP_UNSPECIFIED;
        const int csp_present_flag = avm_rb_read_bit(rb);
        if (csp_present_flag) {
          const int chroma_sample_position = avm_rb_read_literal(rb, 3);
          if (chroma_sample_position > AVM_CSP_BOTTOM) {
            avm_internal_error(error_info, AVM_CODEC_UNSUP_BITSTREAM,
                               "Invalid chroma_sample_position");
          }
          seq_params->chroma_sample_position = chroma_sample_position;
        }
      }
    }
  }
#endif  // !CONFIG_CWG_F270_CI_OBU
}
```

CHANGELOG.md
wantehchang: Yannis: I only did a partial review today. Sorry1 I will try to do a complete review tomorrow.

CMakeLists.txt
jzern: For x86/x86_64 `enable_language(ASM_NASM)` is set if available. `enable_language(ASM)` is only used with Arm. This would take some digging, since libaom has largely the same structure (with updates), and I'm assuming it doesn't have the same issue. Any particular version of CMake?

cmake/Modules/LocalAvm.cmake
```diff
@@ -1 +1,172 @@
...
+
+if(EXISTS "${LIB_FILENAME}")
+    message(STATUS "libavif(AVIF_CODEC_AVM=LOCAL): compiled library found at ${LIB_FILENAME}")
+    add_library(avm STATIC IMPORTED GLOBAL)
+    set_target_properties(avm PROPERTIES IMPORTED_LOCATION "${LIB_FILENAME}" AVIF_LOCAL ON)
+    target_include_directories(avm INTERFACE $<BUILD_INTERFACE:${AVM_EXT_SOURCE_DIR}>)
+    # libavm-src/avm/avm_encoder.h includes config/avm_config.h which is generated by the local build of avm.
+    target_include_directories(avm INTERFACE $<BUILD_INTERFACE:${AVM_EXT_SOURCE_DIR}/build.libavif>)
+
+    # Add link dependency flags
```
wantehchang: Nit: In LocalAom.cmake, the corresponding comment is more verbose:

```
    # Add link dependency flags from the aom.pc file in ext/aom
    # by prepending the build directory to PKG_CONFIG_PATH and then calling
    # pkg_check_modules
```

Should we change this comment to match that?

src/codec_avm.c
```diff
@@ -610,92 +611,92 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
+        if (avmCpuUsed != -1) {
+            if (avm_codec_control(&codec->internal->encoder, AVME_SET_CPUUSED, avmCpuUsed) != AVM_CODEC_OK) {
                 return AVIF_RESULT_UNKNOWN_ERROR;
             }
         }
 
         // Set color_config() in the sequence header OBU.
         if (alpha) {
-            // AV1-AVIF specification, Section 4 "Auxiliary Image Items and Sequences":
+            // AV2-AVIF specification, Section 4 "Auxiliary Image Items and Sequences":
```
wantehchang: Change "AV2-AVIF" back to "AV1-AVIF".

Note: I have several comments like this. For each of these comments, please revert all other occurrences of the change.

src/codec_avm.c
```diff
@@ -610,92 +611,92 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
         } else {
-            // libaom's defaults are AOM_CSP_UNKNOWN and 0 (studio/limited range).
-            // Call aom_codec_control() only if the values are not the defaults.
+            // libavm's defaults are AVM_CSP_UNKNOWN and 0 (studio/limited range).
+            // Call avm_codec_control() only if the values are not the defaults.
 
-            // AV1-AVIF specification, Section 2.2.1. "AV1 Item Configuration Property":
-            //   The values of the fields in the AV1CodecConfigurationBox shall match those
-            //   of the Sequence Header OBU in the AV1 Image Item Data.
+            // AV2-AVIF specification, Section 2.2.1. "AV2 Item Configuration Property":
```
wantehchang: Change "AV2" back to "AV1" in "AV2 Item Configuration Property".

src/codec_avm.c
```diff
@@ -610,92 +611,92 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
+            //   The values of the fields in the AV2CodecConfigurationBox shall match those
```
wantehchang: Change "AV2CodecConfigurationBox" back to "AV1CodecConfigurationBox".

src/codec_avm.c
```diff
@@ -610,92 +611,92 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
+            //   of the Sequence Header OBU in the AV2 Image Item Data.
             if (image->yuvChromaSamplePosition != AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN) {
-                aom_codec_control(&codec->internal->encoder, AV1E_SET_CHROMA_SAMPLE_POSITION, (int)image->yuvChromaSamplePosition);
+                avm_codec_control(&codec->internal->encoder, AV2E_SET_CHROMA_SAMPLE_POSITION, (int)image->yuvChromaSamplePosition);
             }
 
-            // AV1-ISOBMFF specification, Section 2.3.4:
+            // AV2-ISOBMFF specification, Section 2.3.4:
```
wantehchang: Change "AV2-ISOBMFF" back to "AV1-ISOBMFF".

src/codec_avm.c
```diff
@@ -610,92 +611,92 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
+            // Section 2.3.4 of AV2-ISOBMFF says 'colr' with 'nclx' should be present and shall match CICP
             // values in the Sequence Header OBU, unless the latter has 2/2/2 (Unspecified).
             // So set CICP values to 2/2/2 (Unspecified) in the Sequence Header OBU for simplicity.
-            // It may also save 3 bytes since the AV1 encoder can set color_description_present_flag to 0
-            // (see Section 5.5.2 "Color config syntax" of the AV1 specification).
-            // libaom's defaults are AOM_CICP_CP_UNSPECIFIED, AOM_CICP_TC_UNSPECIFIED, and
-            // AOM_CICP_MC_UNSPECIFIED. No need to call aom_codec_control().
-            // aom_image_t::cp, aom_image_t::tc and aom_image_t::mc are ignored by aom_codec_encode().
+            // It may also save 3 bytes since the AV2 encoder can set color_description_present_flag to 0
+            // (see Section 5.5.2 "Color config syntax" of the AV2 specification).
```
wantehchang: This sentence, "It may also save 3 bytes since the AV2 encoder can set color_description_present_flag ..."
 is no longer correct. We can update it for AV2 or simply delete it.

src/obu.c
```diff
@@ -349,93 +350,51 @@ static avifBool parseAV1SequenceHeaderColorConfig(avifBits * bits, avifSequenceH
...
     AV2_CHROMA_FORMAT_420 = 0,
     AV2_CHROMA_FORMAT_400 = 1,
     AV2_CHROMA_FORMAT_444 = 2,
     AV2_CHROMA_FORMAT_422 = 3,
 };
-#endif // CONFIG_CWG_E242_CHROMA_FORMAT_IDC
 
-// Note: Does not parse separate_uv_delta_q.
-static avifBool parseAV2SequenceHeaderColorConfig(avifBits * bits, avifSequenceHeader * header)
+static avifBool av2ChromaSamplePositionToAv1ChromaSamplePosition(uint32_t av2ChromaSamplePosition, avifSequenceHeader * header)
```
wantehchang: IMPORTANT: This function is missing a return statement. I don't know why the compilation succeeds without any warnings.

I suggest changing the function prototype to:

```
static avifChromaSamplePosition av2ChromaSamplePositionToAv1ChromaSamplePosition(uint32_t av2ChromaSamplePosition)
```

and change lines such as

```
        header->chromaSamplePosition = AVIF_CHROMA_SAMPLE_POSITION_VERTICAL;
```

to

```
        return AVIF_CHROMA_SAMPLE_POSITION_VERTICAL;
```

And change the caller

```
        header->chromaSamplePosition = av2ChromaSamplePositionToAv1ChromaSamplePosition(chromaSamplePosition, header);
```

to

```
        header->chromaSamplePosition = av2ChromaSamplePositionToAv1ChromaSamplePosition(chromaSamplePosition);
```

src/obu.c
```diff
@@ -678,17 +653,36 @@ static avifBool av2SequenceHeaderParse(avifSequenceHeader * header, const avifRO
...
+            contentInterpretationFound = AVIF_TRUE;
+            break;
         }
-
-        // Skip this OBU
         obus.data += (size_t)obu_payload_size + init_byte_pos;
         obus.size -= (size_t)obu_payload_size + init_byte_pos;
     }
-    return AVIF_FALSE;
+    // Note: check for contentInterpretationFound instead if CI OBU is mandatory.
```
wantehchang: IMPORTANT: Since the CI OBU is optional, we have to parse the input until we have found the CI OBU or we are sure the CI OBU is absent. This requires knowing what OBUs may follow a CI OBU in the correct order of OBUs. If we have seen any of those OBUs but we haven't seen a CI OBU, then we can conclude the CI OBU is missing.

A simple solution is to parse until we have seen a frame. In AV2, a frame may be of one of several OBU types. So we need to check for all of the OBU types that may be a frame.

src/obu.c
```diff
@@ -446,65 +405,46 @@ static avifBool parseAV2SequenceHeaderColorConfig(avifBits * bits, avifSequenceH
...
+            header->av1C.chromaSubsamplingY = 0;
+            header->yuvFormat = AVIF_PIXEL_FORMAT_YUV422;
+        } else {
+            return AVIF_FALSE;
         }
-        header->av1C.chromaSubsamplingX = (uint8_t)subsampling_x;
-        header->av1C.chromaSubsamplingY = (uint8_t)subsampling_y;
-#endif // !CONFIG_CWG_E242_CHROMA_FORMAT_IDC
 
+#if !CONFIG_CWG_F270_CI_OBU
```
wantehchang: Please assume `CONFIG_CWG_F270_CI_OBU` is equal to 1 and simplify the code accordingly.

I verified that the version of AVM imported into Google's internal repository has the `CONFIG_CWG_F270_CI_OBU` macro. Since `CONFIG_CWG_F270_CI_OBU` defaults to 1, we can safely assume `CONFIG_CWG_F270_CI_OBU` is equal to 1.

CHANGELOG.md
wantehchang: Today I reviewed everything carefully except src/obu.c because I requested some changes to src/obu.c. I will review src/obu.c carefully in the next round.

src/obu.c
wantehchang: UPDATE: Now that we can safely assume `CONFIG_CWG_F270_CI_OBU` is equal to 1, please simplify my comment above accordingly.

CMakeLists.txt
y-guyon:
```
$ cmake --version
cmake version 3.31.6
CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

cmake/Modules/LocalAvm.diff
y-guyon: > I wonder if there is a way to import another project into a namespace in CMake

We could not find a simpler way than a local patch.

src/obu.c
y-guyon: For now it is parsing all OBUs. This is unnecessary as you said but should be correct for the current use of AVM in libavif (no CI change in the middle of a stream generated with libavif).

I vaguely understood from looking at avm code that a CI OBU, if present, must immediately follow a SH OBU. Is that the case?

src/obu.c
y-guyon: Done.

src/obu.c
y-guyon: Added a TODO.

src/obu.c
y-guyon: Done.

src/codec_avm.c
y-guyon: Deleted.

src/codec_avm.c
y-guyon: Done.

src/codec_avm.c
y-guyon: Done.

src/codec_avm.c
y-guyon: Done.

src/codec_avm.c
y-guyon: Done.

cmake/Modules/LocalAvm.cmake
y-guyon: There is no ext/avm so I just simplified the comment and kept the code. Are these instructions still necessary for AVM? Added a TODO.

src/obu.c
y-guyon: This is part of the TODO about following `read_sequence_header_obu()` instead of `parse_sequence_header()`.

src/obu.c
y-guyon: Added a TODO. Why is that better? I assume `parse_sequence_header()` is used to find/extract the codec config from an existing AV2 stream. This is exactly our use case in libavif. `read_sequence_header_obu()` is used to decode the whole AV2 stream, so it may be an example more complex than necessary.

CHANGELOG.md
y-guyon: Thanks!

CMakeLists.txt
wantehchang: On my Linux desktop at work, I can build libavif successfully with this pull request applied but without this `enable_language(ASM)` line. I am using the same version of CMake, and my cmake command line is:

```
cmake .. -G Ninja -DBUILD_SHARED_LIBS=OFF -DAVIF_ENABLE_WERROR=ON -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_AOM_DECODE=ON -DAVIF_CODEC_AOM_ENCODE=ON -DAVIF_CODEC_AVM=LOCAL -DAVIF_CODEC_DAV1D=LOCAL -DAVIF_CODEC_SVT=LOCAL -DAVIF_ZLIBPNG=SYSTEM -DAVIF_JPEG=SYSTEM -DAVIF_LIBYUV=LOCAL -DAVIF_LIBSHARPYUV=LOCAL -DAVIF_LIBXML2=LOCAL -DAVIF_BUILD_EXAMPLES=ON -DAVIF_BUILD_APPS=ON -DAVIF_BUILD_TESTS=ON -DAVIF_GTEST=LOCAL
```

On what platforms did you see the CMake error that requires the `enable_language(ASM)` workaround?

cmake/Modules/LocalAvm.diff
wantehchang: Thanks. By searching for the CMake error message, I found that a similar bug was also reported in libjpeg-turbo: https://github.com/libjpeg-turbo/libjpeg-turbo/issues/504. Their solution is to not support `add_subdirectory()` and instead only support `ExternalProject_Add()`. 

@jzern James: Do you know what the `dist` target is for? I have never used the `dist` target. If nobody is using the `dist` target, we can at least remove it from AVM's build system.

src/obu.c
wantehchang: Yannis: I checked the latest version of the AV2 spec. You are right. Section 7.3 (Ordering of OBUs) specifies that a CI OBU, if present, of the first embedded layer must immediate follow a SH OBU.

But I agree it is correct and simpler to parse all OBUs. So let's remove the TODO comment.

cmake/Modules/LocalAvm.cmake
wantehchang: I found the answer to your question. (The answer is No.) I will add a new comment about this.

src/obu.c
wantehchang: In general you are right. In the specific case of the research-v13.0.0 tag, the code in common/av2_config.c is incorrect. So we must use the more complex code in `read_sequence_header_obu()`.

Note that `read_sequence_header_obu()` is correct and authoritative by definition because it is the reference decoder. Its only problem is that it is more complex than necessary as an example.

src/obu.c
wantehchang: I will try to write a patch. Based on the code snippet I quoted, this should only require removing code.

src/obu.c
```diff
@@ -554,7 +462,9 @@ static avifBool parseAV1SequenceHeader(avifBits * bits, avifSequenceHeader * hea
 }
 
 #if defined(AVIF_CODEC_AVM)
-// See https://gitlab.com/AOMediaCodec/avm/-/blob/main/av1/decoder/decodeframe.c
+// See https://gitlab.com/AOMediaCodec/avm/-/blob/research-v13.0.0/common/av2_config.c?ref_type=tags#L290
+// TODO: b/398931194 - Follow read_sequence_header_obu() instead, see
+//                     https://gitlab.com/AOMediaCodec/avm/-/blob/research-v13.0.0/common/av2_config.c?ref_type=tags#L292
```
wantehchang: The following patch fixes this TODO comment. Please apply the patch. Then review and edit my changes as appropriate.

```
diff --git a/src/obu.c b/src/obu.c
index 00e1ef22..ce07f2b9 100644
--- a/src/obu.c
+++ b/src/obu.c
@@ -128,6 +128,19 @@ static uint32_t avifBitsReadVLC(avifBits * const bits)
     return numBits ? ((1U << numBits) - 1) + avifBitsRead(bits, numBits) : 0;
 }
 
+// Rice-Golomb coding with parameter n.
+static uint32_t avifBitsReadRG(avifBits * const bits, const uint32_t n)
+{
+    for (uint32_t q = 0; q < 32; q++) {
+        const uint32_t rg_bit = avifBitsRead(bits, 1);
+        if (rg_bit == 0) {
+            const uint32_t remainder = avifBitsRead(bits, n);
+            return (q << n) + remainder;
+        }
+    }
+    return 0xFFFFFFFFU;
+}
+
 // Variables in here use snake_case to self-document from the AV1 spec and the draft AV2 spec:
 //
@@ -374,13 +387,10 @@ static avifChromaSamplePosition av2ChromaSamplePositionToAv1ChromaSamplePosition
     }
 }
 
-static avifBool parseAV2SequenceHeaderColorConfig(avifBits * bits, avifSequenceHeader * header)
+static avifBool parseAV2ChromaFormatBitdepth(avifBits * bits, avifSequenceHeader * header)
 {
     const uint32_t chromaFormatIdc = avifBitsReadVLC(bits);
 
-    header->chromaSamplePosition = AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN;
-    header->av1C.chromaSamplePosition = (uint8_t)header->chromaSamplePosition;
-
     const uint32_t bitdepthIdx = avifBitsReadVLC(bits);
     if (bitdepthIdx == 0) {
         header->bitDepth = 10;
@@ -395,31 +405,11 @@ static avifBool parseAV2SequenceHeaderColorConfig(avifBits * bits, avifSequenceH
     header->av1C.twelveBit = header->bitDepth == 12;
     header->av1C.monochrome = chromaFormatIdc == AV2_CHROMA_FORMAT_400;
 
-    uint32_t color_description_present_flag = avifBitsRead(bits, 1);
-    if (color_description_present_flag) {
-        header->colorPrimaries = (avifColorPrimaries)avifBitsRead(bits, 8);                   // color_primaries
-        header->transferCharacteristics = (avifTransferCharacteristics)avifBitsRead(bits, 8); // transfer_characteristics
-        header->matrixCoefficients = (avifMatrixCoefficients)avifBitsRead(bits, 8);           // matrix_coefficients
-    } else {
-        header->colorPrimaries = AVIF_COLOR_PRIMARIES_UNSPECIFIED;
-        header->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
-        header->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
-    }
-
     if (header->av1C.monochrome) {
-        header->range = avifBitsRead(bits, 1) ? AVIF_RANGE_FULL : AVIF_RANGE_LIMITED; // color_range
         header->av1C.chromaSubsamplingX = 1;
         header->av1C.chromaSubsamplingY = 1;
         header->yuvFormat = AVIF_PIXEL_FORMAT_YUV400;
-    } else if (header->colorPrimaries == AVIF_COLOR_PRIMARIES_BT709 &&
-               header->transferCharacteristics == AVIF_TRANSFER_CHARACTERISTICS_SRGB &&
-               header->matrixCoefficients == AVIF_MATRIX_COEFFICIENTS_IDENTITY) {
-        header->range = AVIF_RANGE_FULL; // Assumed.
-        header->av1C.chromaSubsamplingX = 0;
-        header->av1C.chromaSubsamplingY = 0;
-        header->yuvFormat = AVIF_PIXEL_FORMAT_YUV444; // chroma_format_idc is ignored in this path apparently.
     } else {
-        header->range = avifBitsRead(bits, 1) ? AVIF_RANGE_FULL : AVIF_RANGE_LIMITED; // color_range
         if (chromaFormatIdc == AV2_CHROMA_FORMAT_420) {
             header->av1C.chromaSubsamplingX = 1;
             header->av1C.chromaSubsamplingY = 1;
@@ -467,12 +457,10 @@ static avifBool parseAV1SequenceHeader(avifBits * bits, avifSequenceHeader * hea
 //                     https://gitlab.com/AOMediaCodec/avm/-/blob/research-v13.0.0/common/av2_config.c?ref_type=tags#L292
 static avifBool parseAV2SequenceHeader(avifBits * bits, avifSequenceHeader * header)
 {
-#if CONFIG_CWG_E242_SEQ_HDR_ID
     uint32_t seqHeaderId = avifBitsReadVLC(bits);
     if (seqHeaderId >= 16) {
         return AVIF_FALSE;
     }
-#endif // CONFIG_CWG_E242_SEQ_HDR_ID
 
     AVIF_CHECK(parseSequenceHeaderProfile(bits, header));
     header->reduced_still_picture_header = (uint8_t)avifBitsRead(bits, 1); // single_picture_header_flag
@@ -499,13 +487,21 @@ static avifBool parseAV2SequenceHeader(avifBits * bits, avifSequenceHeader * hea
         avifBitsReadVLC(bits);   // conf_win_bottom_offset
     }
 
-    AVIF_CHECK(parseAV2SequenceHeaderColorConfig(bits, header));
+    AVIF_CHECK(parseAV2ChromaFormatBitdepth(bits, header));
+
+    header->colorPrimaries = AVIF_COLOR_PRIMARIES_UNSPECIFIED;
+    header->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
+    header->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED;
+    header->range = AVIF_RANGE_LIMITED;
+
+    header->chromaSamplePosition = AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN;
+    header->av1C.chromaSamplePosition = (uint8_t)header->chromaSamplePosition;
 
     // Other ignored fields.
     return !bits->error;
 }
 
-// See https://gitlab.com/AOMediaCodec/avm/-/blob/research-v13.0.0/common/av2_config.c?ref_type=tags#L457
+// See av2_read_content_interpretation_obu() in av2/decoder/obu_ci.c.
 static avifBool parseAV2ContentInterpretation(avifBits * bits, avifSequenceHeader * header)
 {
     const uint32_t scanTypeIdc = avifBitsRead(bits, 2);                 // ci_scan_type_idc
@@ -518,19 +514,42 @@ static avifBool parseAV2ContentInterpretation(avifBits * bits, avifSequenceHeade
 
     if (colorDescriptionPresent) {
         // Override the Sequence Header's CICP values.
-        const uint32_t colorDescriptionIdc = avifBitsRead(bits, 2); // color_description_idc
-        if (colorDescriptionIdc == 0) {
+        const uint32_t colorDescriptionIdc = avifBitsReadRG(bits, 2); // color_description_idc
+        if (colorDescriptionIdc == 0xFFFFFFFFU) {
+        }
+        if (colorDescriptionIdc == 0) {  // Explicitly signaled
             // parse_content_intrepretation_obu() uses AV2C_READ_UVLC_BITS_OR_RETURN_ERROR()
             // but av2_write_content_interpretation_obu() uses avm_wb_write_literal().
             header->colorPrimaries = (avifColorPrimaries)avifBitsRead(bits, 8);                   // color_primaries
             header->transferCharacteristics = (avifTransferCharacteristics)avifBitsRead(bits, 8); // transfer_characteristics
             header->matrixCoefficients = (avifMatrixCoefficients)avifBitsRead(bits, 8);           // matrix_coefficients
-        } else {
-            // Keep the Sequence Header's CICP values.
+        } else if (colorDescriptionIdc == 1) {  // BT.709 SDR
+            header->colorPrimaries = (avifColorPrimaries)1;
+            header->transferCharacteristics = (avifTransferCharacteristics)1;
+            header->matrixCoefficients = (avifMatrixCoefficients)5;
+        } else if (colorDescriptionIdc == 2) {  // BT.2100 PQ
+            header->colorPrimaries = (avifColorPrimaries)9;
+            header->transferCharacteristics = (avifTransferCharacteristics)16;
+            header->matrixCoefficients = (avifMatrixCoefficients)9;
+        } else if (colorDescriptionIdc == 3) {  // BT.2100 HLG
+            header->colorPrimaries = (avifColorPrimaries)9;
+            header->transferCharacteristics = (avifTransferCharacteristics)14;
+            header->matrixCoefficients = (avifMatrixCoefficients)9;
+        } else if (colorDescriptionIdc == 4) {  // sRGB
+            header->colorPrimaries = (avifColorPrimaries)1;
+            header->transferCharacteristics = (avifTransferCharacteristics)13;
+            header->matrixCoefficients = (avifMatrixCoefficients)0;
+        } else if (colorDescriptionIdc == 5) {  // sYCC
+            header->colorPrimaries = (avifColorPrimaries)1;
+            header->transferCharacteristics = (avifTransferCharacteristics)13;
+            header->matrixCoefficients = (avifMatrixCoefficients)5;
+        } else {  // Reserved
+            // Ignore a reserved value. Keep the default CICP values.
         }
         header->range = avifBitsRead(bits, 1) ? AVIF_RANGE_FULL : AVIF_RANGE_LIMITED; // color_range
     } else {
-        // Keep the Sequence Header's CICP values.
+        // Keep the default CICP values.
     }
 
     if (chromaSamplePositionPresent) {
```

src/obu.c
```diff
@@ -563,29 +473,72 @@ static avifBool parseAV2SequenceHeader(avifBits * bits, avifSequenceHeader * hea
...
+    const uint32_t colorDescriptionPresent = avifBitsRead(bits, 1);     // ci_color_description_present_flag
+    const uint32_t chromaSamplePositionPresent = avifBitsRead(bits, 1); // ci_chroma_sample_position_present_flag
+    avifBitsRead(bits, 1);                                              // ci_aspect_ratio_info_present_flag
+    avifBitsRead(bits, 1);                                              // ci_timing_info_present_flag
+    avifBitsRead(bits, 1);                                              // ci_extension_present_flag
+    avifBitsRead(bits, 1);                                              // reserved_bit
+
+    if (colorDescriptionPresent) {
+        // Override the Sequence Header's CICP values.
+        const uint32_t colorDescriptionIdc = avifBitsRead(bits, 2); // color_description_idc
```
wantehchang: The type of the `color_description_idc` syntax element is rg(2), not f(2). My patch took care of this.

cmake/Modules/LocalAvm.cmake
```diff
@@ -1 +1,173 @@
-include(LocalAom)
+set(AVIF_AVM_GIT_TAG research-v13.0.0)
+
+set(AVM_EXT_SOURCE_DIR "${AVIF_SOURCE_DIR}/ext/avm")
```
wantehchang: In commit https://github.com/AOMediaCodec/libavif/commit/963c4bb918bee37cf1ccbd81efa5d8cf530647cf, we removed ext/avm.cmd. So we can remove anything that contains "ext/avm" or "build.libavif" from this file.

cmake/Modules/LocalAvm.cmake
```diff
@@ -1 +1,173 @@
...
+if(EXISTS "${LIB_FILENAME}")
+    message(STATUS "libavif(AVIF_CODEC_AVM=LOCAL): compiled library found at ${LIB_FILENAME}")
+    add_library(avm STATIC IMPORTED GLOBAL)
+    set_target_properties(avm PROPERTIES IMPORTED_LOCATION "${LIB_FILENAME}" AVIF_LOCAL ON)
+    target_include_directories(avm INTERFACE $<BUILD_INTERFACE:${AVM_EXT_SOURCE_DIR}>)
+    # libavm-src/avm/avm_encoder.h includes config/avm_config.h which is generated by the local build of avm.
+    target_include_directories(avm INTERFACE $<BUILD_INTERFACE:${AVM_EXT_SOURCE_DIR}/build.libavif>)
+
+    # Add link dependency flags
+    # TODO: b/398931194 - Check if this is necessary
```
wantehchang: This is not necessary because ext/avm.cmd has been removed. See my comment at line 3 above.

src/obu.c
```diff
@@ -645,9 +598,12 @@ static avifBool av1SequenceHeaderParse(avifSequenceHeader * header, const avifRO
 #if defined(AVIF_CODEC_AVM)
 static avifBool av2SequenceHeaderParse(avifSequenceHeader * header, const avifROData * sample)
 {
+    avifBool sequenceHeaderFound = AVIF_FALSE;
+    avifBool contentInterpretationFound = AVIF_FALSE;
     avifROData obus = *sample;
 
-    // Find the sequence header OBU
+    // Find the Sequence Header OBU, and the Content Interpretation OBU if any.
+    // TODO: b/398931194 - Find when to stop looking for a CI OBU.
```
wantehchang: Let's remnove this TODO comment. On second thought I think it is simpler to just parse all the OBUs in `sample`.

CHANGELOG.md
wantehchang: Today I reviewed src/obu.c carefully.

cmake/Modules/LocalAvm.diff
jzern: > Thanks. By searching for the CMake error message, I found that a similar bug was also reported in libjpeg-turbo: [libjpeg-turbo/libjpeg-turbo#504](https://github.com/libjpeg-turbo/libjpeg-turbo/issues/504). Their solution is to not support `add_subdirectory()` and instead only support `ExternalProject_Add()`.
> 
> @jzern James: Do you know what the `dist` target is for? I have never used the `dist` target. If nobody is using the `dist` target, we can at least remove it from AVM's build system.

It comes from libvpx. It's meant to create a distribution directory of binaries and possibly sources. The need for this with the combination of the `install` target and `git archive` is questionable. This target is used in the libaom presubmit to make sure it stays functional, but if you wanted to propose removing it, that can be removed.

CMakeLists.txt
y-guyon: > I can build libavif successfully

Note that `-DAVIF_CODEC_SVT=LOCAL` enables `LocalSvt.cmake` which may set `enable_language(ASM)` too.

> On what platforms did you see the CMake error that requires the `enable_language(ASM)` workaround?

Linux laptop, compiled through VSCode:

```
[cmake] CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
[cmake] Missing variable is:
[cmake] CMAKE_ASM_COMPILE_OBJECT
[proc] The command: /usr/bin/cmake -DCMAKE_BUILD_TYPE:STRING=Debug -DBUILD_TESTING:STRING=ON -DAVIF_CODEC_AOM:STRING=LOCAL -DAVIF_CODEC_AVM:STRING=LOCAL -DAVIF_LIBYUV:STRING=LOCAL -DAVIF_BUILD_APPS:STRING=ON -DAVIF_BUILD_EXAMPLES:STRING=ON -DAVIF_BUILD_TESTS:STRING=ON -DAVIF_ENABLE_EXPERIMENTAL_EXTENDED_PIXI:STRING=ON -DAVIF_ENABLE_EXPERIMENTAL_MINI:STRING=ON -DAVIF_ENABLE_NODISCARD:STRING=ON -DAVIF_FUZZTEST:STRING=OFF -DAVIF_GTEST:STRING=LOCAL -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -DCMAKE_C_COMPILER:FILEPATH=/usr/bin/clang -DCMAKE_CXX_COMPILER:FILEPATH=/usr/bin/clang++ --no-warn-unused-cli -S /home/yguyon/git/libavif -B /home/yguyon/git/libavif/build -G Ninja exited with code: 1
```

`settings.json`

```json
  "cmake.configureSettings": {
    "AVIF_CODEC_AOM": "LOCAL",
    "AVIF_CODEC_AVM": "LOCAL",
    "AVIF_LIBYUV": "LOCAL",
    "AVIF_BUILD_APPS": "ON",
    "AVIF_BUILD_EXAMPLES": "ON",
    "AVIF_BUILD_TESTS": "ON",
    "AVIF_ENABLE_EXPERIMENTAL_EXTENDED_PIXI": "ON",
    "AVIF_ENABLE_EXPERIMENTAL_MINI": "ON",
    "AVIF_ENABLE_NODISCARD": "ON",
    "AVIF_FUZZTEST": "OFF",
    "AVIF_GTEST": "LOCAL"
  }
```

src/obu.c
y-guyon: Done.

cmake/Modules/LocalAvm.cmake
y-guyon: Done.

cmake/Modules/LocalAvm.cmake
y-guyon: Done.

src/obu.c
y-guyon: It was based on https://gitlab.com/AOMediaCodec/avm/-/blob/research-v13.0.0/common/av2_config.c?ref_type=tags#L492.

Done.

src/obu.c
y-guyon: Done.

cmake/Modules/LocalAvm.diff
wantehchang: James: Thank you for the history. I can try removing the `dist` target from AVM first.

wantehchang: I wrote [AVM merge request !2985](https://gitlab.com/AOMediaCodec/avm/-/merge_requests/2985) to remove the `dist` target from AVM.

src/obu.c
wantehchang: This is an example that common/av2_config.c in research-v13.0.0 is incorrect.

Details:

The `rg(n)` type descriptor is specified as follows:

```
rg( n ) {
    for ( q = 0; q < 32; q++ ) {
        f(1) rg_bit;
        if ( rg_bit == 0 ) {
            f(n) remainder;
            return (q << n) + remainder
        }
    }
    return -1
}
```

It follows that `rg(2)` is at least three bits (one-bit `rg_bit` and two-bit `remainder`). Even if we only support the shortest `rg(2)` codes, we need to read `f(3)`, not `f(2)`.

CMakeLists.txt
wantehchang: Thanks. I can reproduce the cmake error after removing `-DAVIF_CODEC_SVT=LOCAL`. Here is perhaps the shortest cmake command line to reproduce the cmake error:

```
cmake .. -G Ninja -DBUILD_SHARED_LIBS=OFF -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_AVM=LOCAL
```

I suggest we reference https://gitlab.kitware.com/cmake/cmake/-/issues/25042 in the comment for your workaround. I found that cmake bug report when I searched for the error message, and that cmake bug report is also cited in LocalSvt.cmake.

src/obu.c
```diff
@@ -40,7 +40,8 @@
 #include <string.h>
 
 #if defined(AVIF_CODEC_AVM)
-#include "config/aom_config.h"
+#include "avm/avm_codec.h"
```
wantehchang, line 44: This file should not depend on a particular AV2 codec such as AVM.

Please add a TODO comment that when AV2 is released, we should remove "avm/avm_codec.h" and replace OBU_SEQUENCE_HEADER and OBU_CONTENT_INTERPRETATION with their numeric values.

src/obu.c
```diff
@@ -127,6 +128,21 @@ static uint32_t avifBitsReadVLC(avifBits * const bits)
     return numBits ? ((1U << numBits) - 1) + avifBitsRead(bits, numBits) : 0;
 }
 
+#if defined(AVIF_CODEC_AVM)
+// Rice-Golomb coding with parameter n.
+static uint32_t avifBitsReadRG(avifBits * const bits, const uint32_t n)
+{
+    for (uint32_t q = 0; q < 32; q++) {
+        const uint32_t rg_bit = avifBitsRead(bits, 1);
```
wantehchang: Please change "rg_bit" to "rgBit".

src/obu.c
```diff
@@ -349,183 +365,67 @@ static avifBool parseAV1SequenceHeaderColorConfig(avifBits * bits, avifSequenceH
...
-                // CSP_CENTER: Horizontal offset 0.5, vertical offset 0.5
-                header->chromaSamplePosition = AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN;
-            } else if (chromaSamplePosition == 2) {
-                // CSP_TOPLEFT: Horizontal offset 0, vertical offset 0
-                header->chromaSamplePosition = AVIF_CHROMA_SAMPLE_POSITION_COLOCATED;
-            } else {
-                header->chromaSamplePosition = AVIF_CHROMA_SAMPLE_POSITION_UNKNOWN;
-            }
-            header->av1C.chromaSamplePosition = (uint8_t)header->chromaSamplePosition;
+        if (chromaFormatIdc == AV2_CHROMA_FORMAT_420) {
```
wantehchang: Nit: Please remove one level of indentation by applying the following patch:

```
diff --git a/src/obu.c b/src/obu.c
index bff8f912..6e733d28 100644
--- a/src/obu.c
+++ b/src/obu.c
@@ -411,22 +411,20 @@ static avifBool parseAV2ChromaFormatBitdepth(avifBits * bits, avifSequenceHeader
         header->av1C.chromaSubsamplingX = 1;
         header->av1C.chromaSubsamplingY = 1;
         header->yuvFormat = AVIF_PIXEL_FORMAT_YUV400;
+    } else if (chromaFormatIdc == AV2_CHROMA_FORMAT_420) {
+        header->av1C.chromaSubsamplingX = 1;
+        header->av1C.chromaSubsamplingY = 1;
+        header->yuvFormat = AVIF_PIXEL_FORMAT_YUV420;
+    } else if (chromaFormatIdc == AV2_CHROMA_FORMAT_444) {
+        header->av1C.chromaSubsamplingX = 0;
+        header->av1C.chromaSubsamplingY = 0;
+        header->yuvFormat = AVIF_PIXEL_FORMAT_YUV444;
+    } else if (chromaFormatIdc == AV2_CHROMA_FORMAT_422) {
+        header->av1C.chromaSubsamplingX = 1;
+        header->av1C.chromaSubsamplingY = 0;
+        header->yuvFormat = AVIF_PIXEL_FORMAT_YUV422;
     } else {
-        if (chromaFormatIdc == AV2_CHROMA_FORMAT_420) {
-            header->av1C.chromaSubsamplingX = 1;
-            header->av1C.chromaSubsamplingY = 1;
-            header->yuvFormat = AVIF_PIXEL_FORMAT_YUV420;
-        } else if (chromaFormatIdc == AV2_CHROMA_FORMAT_444) {
-            header->av1C.chromaSubsamplingX = 0;
-            header->av1C.chromaSubsamplingY = 0;
-            header->yuvFormat = AVIF_PIXEL_FORMAT_YUV444;
-        } else if (chromaFormatIdc == AV2_CHROMA_FORMAT_422) {
-            header->av1C.chromaSubsamplingX = 1;
-            header->av1C.chromaSubsamplingY = 0;
-            header->yuvFormat = AVIF_PIXEL_FORMAT_YUV422;
-        } else {
-        }
+        return AVIF_FALSE;
     }
 
     return !bits->error;
```

src/obu.c
```diff
@@ -554,42 +454,123 @@ static avifBool parseAV1SequenceHeader(avifBits * bits, avifSequenceHeader * hea
...
+    if (colorDescriptionPresent) {
+        // Override the default CICP values.
+        const uint32_t colorDescriptionIdc = avifBitsReadRG(bits, 2); // color_description_idc
+        if (colorDescriptionIdc == 0xFFFFFFFFU) {
+            return AVIF_FALSE;
+        }
+        if (colorDescriptionIdc == 0) {
+            // Explicitly signaled
+            // parse_content_intrepretation_obu() uses AV2C_READ_UVLC_BITS_OR_RETURN_ERROR()
+            // but av2_write_content_interpretation_obu() uses avm_wb_write_literal().
```
wantehchang: Delete this sentence (lines 523-524).

src/obu.c
```diff
@@ -554,42 +454,123 @@ static avifBool parseAV1SequenceHeader(avifBits * bits, avifSequenceHeader * hea
...
+            header->transferCharacteristics = (avifTransferCharacteristics)avifBitsRead(bits, 8); // transfer_characteristics
+            header->matrixCoefficients = (avifMatrixCoefficients)avifBitsRead(bits, 8);           // matrix_coefficients
+        } else if (colorDescriptionIdc == 1) {
+            // BT.709 SDR
+            header->colorPrimaries = AVIF_COLOR_PRIMARIES_BT709;                   // 1
+            header->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_BT709; // 1
+            header->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT470BG;         // 5
+        } else if (colorDescriptionIdc == 2) {
+            // BT.2100 PQ
+            header->colorPrimaries = AVIF_TRANSFER_CHARACTERISTICS_LOG100;      // 9
```
wantehchang: AVIF_COLOR_PRIMARIES_BT2020

src/obu.c
```diff
@@ -554,42 +454,123 @@ static avifBool parseAV1SequenceHeader(avifBits * bits, avifSequenceHeader * hea
...
+            header->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_PQ; // 16
+            header->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT2020_NCL;   // 9
+        } else if (colorDescriptionIdc == 3) {
+            // BT.2100 HLG
+            header->colorPrimaries = AVIF_TRANSFER_CHARACTERISTICS_LOG100;                // 9
```
wantehchang: AVIF_COLOR_PRIMARIES_BT2020

src/obu.c
```diff
@@ -645,9 +626,11 @@ static avifBool av1SequenceHeaderParse(avifSequenceHeader * header, const avifRO
 #if defined(AVIF_CODEC_AVM)
 static avifBool av2SequenceHeaderParse(avifSequenceHeader * header, const avifROData * sample)
 {
+    avifBool sequenceHeaderFound = AVIF_FALSE;
+    avifBool contentInterpretationFound = AVIF_FALSE;
```
wantehchang: Delete `contentInterpretationFound`. It is set but not used.

src/obu.c
```diff
@@ -678,17 +661,32 @@ static avifBool av2SequenceHeaderParse(avifSequenceHeader * header, const avifRO
...
-            return parseAV2SequenceHeader(&seqHdrBits, header);
+            if (!parseAV2SequenceHeader(&seqHdrBits, header)) {
+                return AVIF_FALSE;
+            }
+            sequenceHeaderFound = AVIF_TRUE;
+        } else if (obu_type == OBU_CONTENT_INTERPRETATION) { // optional Content Interpretation
+            if (!sequenceHeaderFound) {
+                return AVIF_FALSE;
+            }
+            avifBits seqHdrBits;
```
wantehchang: Nit: seqHdrBits => ciBits

CMakeLists.txt
y-guyon: Done.

cmake/Modules/LocalAvm.diff
y-guyon: Added a todo.

src/obu.c
y-guyon: Done.

src/obu.c
y-guyon: Done. Also changed variables in `av2SequenceHeaderParse()`.

src/obu.c
y-guyon: Done.

src/obu.c
y-guyon: Done.

src/obu.c
y-guyon: Went with AVIF_COLOR_PRIMARIES_BT2100.

src/obu.c
y-guyon: Went with AVIF_COLOR_PRIMARIES_BT2100.

src/obu.c
y-guyon: Done.

src/obu.c
y-guyon: Done.

CHANGELOG.md
wantehchang: I reviewed the merged code. LGTM!

Issue #2944, opened 2025-12-25
Apple iPhone HDR support (gain maps in jpeg)


VityaSchel:
Hello, I tried compiling libavif from source with libxml2 on macos and got it to enable --qgain-map option but avifenc still does not recognize iphone's gain maps and prints "Gain map       : Absent". I tried tinkering with avifgainmaputil for a few hours but it does not recognize apple's gain maps at all. I found this library https://github.com/grapeot/AppleJPEGGainMap that perfectly extracts grayscale gain map as a separate file. Since iPhone stores gain map at 0.5x resolution I also had to resize it to size of the hdr source image. Still, avifgainmaputil couldn't produce a proper image that would display both in unsupported browsers (firefox) and hdr supported browsers (safari, chrome).

Please add native support for apple's gain maps to avifenc! it would make my life so much easier since I wouldn't have to do some dark magic with tens of libraries to just transfer hdr from jpeg to avif :)

maryla-uc:
Hi, thanks for your interest in libavif and avifgainmaputil!
Could you provide example files that I could work with? Bonus points if it has a permissive license so that I could add it to the libavif repository as a test file.

I did find HDR.jpg in the repo you provided https://github.com/grapeot/AppleJPEGGainMap but I'd like another example, and also that repo doesn't have a license.

VityaSchel:
IT WORKS 🥳🥳🥳🥳🥳🥳 `* Gain Map AV1 total size: 188296 bytes` thank you sooooo much!!!!! i wish i could donate but i'm broke asylum seeker who will go on the street on next week but thank you so much now I can use iphone's hdr photos in avif and it just works!!!

for everyone coming here in the future until the next libavif release and push to packagers registries:

follow the instructions in README to pull the "main" branch, you'll have to compile it. I personally was able to do it like this on macos:

```
mkdir build && cd $_
cmake .. \
  -DCMAKE_BUILD_TYPE=Release \
  -DAVIF_BUILD_APPS=ON \
  -DAVIF_LIBYUV=LOCAL \
  -DAVIF_LIBXML2=SYSTEM \
  -DCMAKE_PREFIX_PATH="/opt/homebrew;/opt/homebrew/opt/libxml2" \
-DAVIF_CODEC_AOM=LOCAL
cmake --build . --config Release --parallel
```

then just use avifenc binary in the libavif/build directory

PR #2960, opened 2026-01-09
Add support for Apple gain maps (iPhone jpegs)

maryla-uc:
These use a different XMP syntax on the gain map image, and do not carry XMP on the main image.
Moreover, the alternate headroom may not be present in the XMP, in which case it's read from the Apple Maker Notes in Exif.

#2944

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -265,6 +265,9 @@ static const uint8_t * avifJPEGFindSubstr(const uint8_t * str, size_t strLength,
 #define AVIF_JPEG_EXTENDED_XMP_TAG "http://ns.adobe.com/xmp/extension/\0"
 #define AVIF_JPEG_EXTENDED_XMP_TAG_LENGTH 35
 
+#define AVIF_EXIF_APPLE_MAKER_NOTES_HEADER "Apple iOS\0\0\1MM"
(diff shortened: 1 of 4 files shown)


apps/shared/avifjpeg.c
```diff
@@ -401,15 +402,20 @@ avifBool avifGetExifAppleHeadroom(const avifROData * exif, double * altHeadroom)
...
             if (tagId == 0x8769) { // Exif Offset (offset to a sub IFD)
-                // Move back to just before the tagData which contians the offset of the Exif IFD.
+                // Move back to just before the tagData which contains the offset of the Exif IFD.
                 offset -= 4;
                 break;
             } else if (tagId == 0x927c) { // Maker Notes
-                uint32_t makerNotesOffset = tagData;
+                size_t makerNotesOffset = tagData;
+                uint8_t makerTag[AVIF_EXIF_APPLE_MAKER_NOTES_HEADER_LENGTH];
+                AVIF_CHECK(avifJPEGReadBytes(exif, makerTag, &makerNotesOffset, AVIF_EXIF_APPLE_MAKER_NOTES_HEADER_LENGTH));
```
vrabaud, line 411: You can probably add AVIF_NODISCARD to those functions to prevent you from forgetting a check

Issue #2956, opened 2026-01-08
[Feature Request] Support encoding 16-bit lossless tiled image


macdavis:
Please add support for encoding parameter `--lossless -d 12,8 --grid AxB`. This would be particular useful for large dimension images.

PR #2964, opened 2026-01-14
Support Sample Transform grids

y-guyon:
Fixes https://github.com/AOMediaCodec/libavif/issues/2956

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The changes are relative to the previous release, unless the baseline is specifi
   a cICP chunk and other color information chunks, such as iCCP (ICC profile),
   the other chunks are ignored as per the PNG Specification Third Edition
   Section 4.3.
+* Support Sample Transform derived image items with grid input image items.
 
 ### Changed since 1.3.0
 


src/read.c
```diff
@@ -2320,24 +2320,35 @@ static avifResult avifDecoderSampleTransformItemValidateProperties(const avifDec
...
+    // If item is a grid, check that its input image items share the same properties.
     for (uint32_t i = 0; i < item->meta->items.count; ++i) {
         avifDecoderItem * inputImageItem = item->meta->items.item[i];
         if (inputImageItem->dimgForID != item->id) {
             continue;
         }
-        // Even if inputImageItem is a grid, the ispe property from its first tile should have been copied to the grid item.
         const avifProperty * inputImageItemIspeProp = avifPropertyArrayFind(&inputImageItem->properties, "ispe");
         AVIF_ASSERT_OR_RETURN(inputImageItemIspeProp != NULL);
-        if (inputImageItemIspeProp->u.ispe.width != ispeProp->u.ispe.width ||
```
maryla-uc, line 2331: It seems that you are no longer comparing inputImageItemIspeProp to ispeProp, is that ok?
I.e. I see that now you have itemLevel1 -> itemLevel2 -> itemLevel3 and checking that level 2 and level 3 ispe match, but not level 1 and level 2 ispe match? (if I understand the change correctly)
I may be missing something, but checking just in case.

src/read.c
```diff
@@ -6745,6 +6759,76 @@ static avifBool avifDecoderDataFrameFullyDecoded(const avifDecoderData * data)
...
+    AVIF_CHECKRES(avifImageApplyExpression(dstImage,
+                                           AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_32,
+                                           &decoder->data->meta->sampleTransformExpression,
+                                           decoder->data->sampleTransformNumInputImageItems,
+                                           inputImages,
+                                           planes));
+    return AVIF_RESULT_OK;
+}
+
+// Intermediate function used to safely destroy temporary buffers even in case of error.
```
maryla-uc, line 6819: C is so much fun

maryla-uc:
Ha I forgot to say, while you're at it maybe you could update the description of the -d flag in avifenc.1.md, it's missing the part about bit depth extension

maryla-uc:
And update the changelog if you want

src/read.c
y-guyon: > It seems that you are no longer comparing inputImageItemIspeProp to ispeProp, is that ok?

Yes.

> I.e. I see that now you have itemLevel1 -> itemLevel2 -> itemLevel3 and checking that level 2 and level 3 ispe match, but not level 1 and level 2 ispe match? (if I understand the change correctly) I may be missing something, but checking just in case.

No. The code was checking `grid.ispe == cell1.ispe`. I could not remember or find why, but `sato` grids were not supported anyway. The correct check is `cellX.ispe == cellY.ispe`. That is implemented in this PR.

y-guyon:
> Ha I forgot to say, while you're at it maybe you could update the description of the -d flag in avifenc.1.md, it's missing the part about bit depth extension

Done in https://github.com/AOMediaCodec/libavif/pull/2967

PR #2968, opened 2026-01-20
Always set seqTier0 in parseAV2SequenceHeader()

y-guyon:
Also remove warning about scanTypeIdc being unused.


maryla-uc:
Could you explain the motivation behind this change? I have no idea what this does.

y-guyon:
> Could you explain the motivation behind this change? I have no idea what this does.

`header->av1C` is the struct used for AV1 codec configurations and is reused for now for AV2, but some of the fields do not match, such as `seqTier0`, which is hardcoded to 0. However, `header->av1C.seqTier0` was left unset unless `header->av1C.seqLevelIdx0 > 7 && !header->reduced_still_picture_header`, so move the assignment out of the condition's body.

`scanTypeIdc` is unused. Remove the variable to avoid compilation warnings treated as errors.

PR #2970, opened 2026-01-22
Switch layered images to all-intra encoding under certain conditions

juliobbv-p:
Switch layered images to all-intra encoding under certain conditions:
- The total number of layers is 2
- Quality of the first layer is very low (q <= 10)

Using all-intra encoding comes with the following benefits:
1. First layer will be smaller than the second layer (which can oftentimes not be the case with good mode at quality < 40)
2. Outputs have predictable file sizes: the sum of the first layer (quality 2) plus the second layer (quality set by the caller)
3. Because the first layer is very small, progressive encoding overhead is also smaller and more stable (about 5-8% for quality 40 and 2-4% for quality 60)
4. Availability of choosing tune IQ in libaom AV1 (which requires all-intra usage)

From the point of view of `avifenc`, encoding an image with `--progressive` will result in all-intra encoding being used, which is the main use case this change targets.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -49,6 +49,8 @@
 #endif
 #endif
 
+#define TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD 10
+
 struct avifCodecInternal
 {
 #if defined(AVIF_CODEC_AOM_DECODE)
@@ -65,6 +67,7 @@ struct avifCodecInternal
     avifPixelFormatInfo formatInfo;
     aom_img_fmt_t aomFormat;
     uint32_t currentLayer;
+    int qualityFirstLayer;
 #endif
 };
 
@@ -656,6 +659,21 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                                       avifAddImageFlags addImageFlags,
                                       avifCodecEncodeOutput * output)
 {
+    // Save the quality of the first layer because it's useful for further decision making
+    if (encoder->extraLayerCount > 0 && codec->internal->currentLayer == 0) {
+        codec->internal->qualityFirstLayer = quality;
+    }
+
+    // All-intra encoding is beneficial when encoding a two-layer image item and the quality of the first layer is very low.
+    // Switching to all-intra encoding comes with the following benefits:
+    // - The first layer will be smaller than the second layer (which is often not the case with inter encoding)
+    // - Outputs have predictable file sizes: the sum of the first layer (quality <= 10) plus the second layer (quality set by the caller)
+    // - Because the first layer is very small, layered encoding overhead is also smaller and more stable (about 5-8% for quality 40 and 2-4% for quality 60)
+    // - Option of choosing tune IQ (which requires AOM_USAGE_ALL_INTRA)
+    avifBool useAllIntraForLayered = encoder->extraLayerCount == 1 &&
+                                     codec->internal->qualityFirstLayer <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD;
+    avifBool useAllIntra = (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) || useAllIntraForLayered;
+
     // Map encoder speed to AOM usage + CpuUsed:
     // Speed  0: GoodQuality CpuUsed 0
     // Speed  1: GoodQuality CpuUsed 1
@@ -671,7 +689,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     unsigned int aomUsage = AOM_USAGE_GOOD_QUALITY;
     // Use AOM_USAGE_ALL_INTRA (added in https://crbug.com/aomedia/2959) for still image encoding if available.
 #if defined(AOM_USAGE_ALL_INTRA)
-    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) {
+    if (useAllIntra) {
         aomUsage = AOM_USAGE_ALL_INTRA;
     }
 #endif
@@ -680,7 +698,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         aomCpuUsed = AVIF_CLAMP(encoder->speed, 0, 9);
         if (aomCpuUsed >= 7) {
 #if defined(AOM_USAGE_ALL_INTRA) && defined(ALL_INTRA_HAS_SPEEDS_7_TO_9)
-            if (!(addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE)) {
+            if (!useAllIntra) {
                 aomUsage = AOM_USAGE_REALTIME;
             }
 #else
@@ -826,7 +844,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             // libaom to set still_picture and reduced_still_picture_header to
             // 1 in AV1 sequence headers.
             cfg->g_limit = 1;
-
+        }
+        if (useAllIntra) {
+#if !defined(AOM_USAGE_ALL_INTRA)
             // Use the default settings of the new AOM_USAGE_ALL_INTRA (added in
             // https://crbug.com/aomedia/2959).
             //
@@ -838,6 +858,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             cfg->kf_mode = AOM_KF_DISABLED;
             // Tell libaom that all frames will be key frames.
             cfg->kf_max_dist = 0;
+#endif
         } else {
             if (encoder->keyframeInterval > 0) {
                 cfg->kf_max_dist = encoder->keyframeInterval;
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -14,6 +14,8 @@
 #include <stdlib.h>
 #include <string.h>
 
+#define TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD 10
+
 struct avifCodecInternal
 {
     avifBool decoderInitialized;
@@ -547,7 +549,17 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
             // libavm to set still_picture and reduced_still_picture_header to
             // 1 in AV2 sequence headers.
             cfg->g_limit = 1;
+        }
 
+        // All-intra encoding is beneficial when encoding a two-layer image item and the quality of the first layer is very low.
+        // Switching to all-intra encoding comes with the following benefits:
+        // - The first layer will be smaller than the second layer (which is often not the case with inter encoding)
+        // - Outputs have predictable file sizes: the sum of the first layer (quality <= 10) plus the second layer (quality set by the caller)
+        // - Because the first layer is very small, layered encoding overhead is also smaller and more stable (about 5-8% for quality 40 and 2-4% for quality 60)
+        avifBool useAllIntraForLayered = encoder->extraLayerCount == 1 && quality <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD;
+        avifBool useAllIntra = (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) || useAllIntraForLayered;
+
+        if (useAllIntra) {
             // Use the default settings of the new AVM_USAGE_ALL_INTRA (added in
             // https://crbug.com/aomedia/2959).
             //


src/codec_aom.c
```diff
@@ -840,7 +840,8 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             // libaom to set still_picture and reduced_still_picture_header to
             // 1 in AV1 sequence headers.
             cfg->g_limit = 1;
-
+        }
+        if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE || addImageFlags & AVIF_ADD_IMAGE_FLAG_PROGRESSIVE) {
```
juliobbv-p: Technically I could also do:

```c
if (addImageFlags & (AVIF_ADD_IMAGE_FLAG_SINGLE | AVIF_ADD_IMAGE_FLAG_PROGRESSIVE))
```

But I felt the current code to be a bit more readable. Let me know if you have strong feelings either way on style.

include/avif/avif.h
```diff
@@ -1631,7 +1631,11 @@ typedef enum avifAddImageFlag
     // Use this flag when encoding a single frame, single layer image.
     // Signals "still_picture" to AV1 encoders, which tweaks various compression rules.
     // This is enabled automatically when using the avifEncoderWrite() single-image encode path.
-    AVIF_ADD_IMAGE_FLAG_SINGLE = (1 << 1)
+    AVIF_ADD_IMAGE_FLAG_SINGLE = (1 << 1),
+
+    // Use this flag when encoding a progressive frame
+    // Configures encoders with settings optimized for progressive encoding.
+    AVIF_ADD_IMAGE_FLAG_PROGRESSIVE = (1 << 2),
```
y-guyon: I wonder if this new `AVIF_ADD_IMAGE_FLAG_PROGRESSIVE` is necessary.

Without this PR, setting `AVIF_ADD_IMAGE_FLAG_SINGLE` does not work with progressive because it also sets `cfg->g_limit = 1`. That seems to be the only blocker. Is that accurate?
- Could `cfg->g_limit = 1` be undone through `avifCodecSpecificOption`s?
- Could we guess in `codec_aom.c` that progressive is wanted despite `AVIF_ADD_IMAGE_FLAG_SINGLE`, so that `cfg->g_limit = 1` is skipped in this case, without changing the libavif API?

juliobbv-p: So, there are two things at play here:

1. In libavif: we need a way to tell apart progressive from regular layered encoding to set the correct usage (all-intra vs. good mode)
2. In libaom/libavm: the encoders need to know when to set the "reduced still picture header" for single images (this is what `cfg->g_limit = 1` is doing here). Progressive images can use multiple operating points, so they need a full header

From what I can tell from libavif's architecture, signaling intent (single, progressive, layered) is the cleanest way to configure downstream encoders. Removing `cfg->g_limit = 1` is an option, but doing so would make single images a few bytes larger due to a lack of the reduced header.

Let me know what you think.

y-guyon: At head `AVIF_ADD_IMAGE_FLAG_SINGLE` requires `extraLayerCount=0`. The combination `AVIF_ADD_IMAGE_FLAG_SINGLE && extraLayerCount > 0` could be used to signal "progressive". I am not saying it is better than adding a dedicated flag. I am wondering if it is.

Taking a step back, the question revolves around whether we want to:
- Migrate the "auto-progressive" code from the `avifenc` tool to the libavif API (so adding API elements is even more justified), or
- Keep the "auto-progressive" feature in `avifenc` only (here the libavif API just has to allow a behavior but the API does not have to be as user-friendly as with the first bullet)

And then we have to answer:
- Is the libavif API already complete enough to have `AOM_USAGE_ALL_INTRA` and `g_limit>1`, maybe thanks to `avifCodecSpecificOptions`?
- Otherwise, do we prefer to add explicit libavif API elements such as `AVIF_ADD_IMAGE_FLAG_PROGRESSIVE` or `avifEncoder::progressive` (same behavior as `avifenc --progressive`), or is "hacking" the current surfaces enough (such as `AVIF_ADD_IMAGE_FLAG_SINGLE && extraLayerCount=1`)?

juliobbv-p: > The combination `AVIF_ADD_IMAGE_FLAG_SINGLE` && `extraLayerCount > 0` could be used to signal "progressive". 

This is a good observation. AFAICT from reading the libavif API, this should be good enough, as long as the AV2 high-level syntax on encoding a still picture or a layered sequence didn't change from AV1 (which I believe it's the case). I currently don't have access to the most recent AV2 draft spec -- can you or @wantehchang confirm?

By doing this, we will need to change the meaning of `AVIF_ADD_IMAGE_FLAG_SINGLE` from "we're encoding a single layer, still picture element" to "we're encoding a single source image (that either produces a still picture or a sequence if it's a progressive image)". I think we'll be fine if we document this behavior appropriately in code.

wantehchang: Julio: libavif doesn't support encoding a layered image sequence.

My first reaction is that we should keep the current meaning of `AVIF_ADD_IMAGE_FLAG_SINGLE` (which means `avifEncoderAddImageInternal()` will be called only once), and we should use `extraLayerCount > 0` to detect a progressive image. The `extraLayerCount` field was added to `avifEncoder` specifically for encoding a progressive image.

src/codec_aom.c
wantehchang: Please switch to the `addImageFlags & (AVIF_ADD_IMAGE_FLAG_SINGLE | AVIF_ADD_IMAGE_FLAG_PROGRESSIVE)` form. This kind of expression is commonly used in code that deals with bit flags.

include/avif/avif.h
```diff
@@ -1631,7 +1631,11 @@ typedef enum avifAddImageFlag
     // Use this flag when encoding a single frame, single layer image.
     // Signals "still_picture" to AV1 encoders, which tweaks various compression rules.
     // This is enabled automatically when using the avifEncoderWrite() single-image encode path.
-    AVIF_ADD_IMAGE_FLAG_SINGLE = (1 << 1)
+    AVIF_ADD_IMAGE_FLAG_SINGLE = (1 << 1),
+
+    // Use this flag when encoding a progressive frame
```
wantehchang: Nit: Add a period at the end of this sentence.

But I believe we don't need to add the `AVIF_ADD_IMAGE_FLAG_PROGRESSIVE` flag. Please see my comment at avif.h:1638.

apps/avifenc.c
```diff
@@ -1187,6 +1306,9 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
         if (!avifInputHasRemainingData(input, imageIndex) && (settings->layers == 1)) {
             addImageFlags |= AVIF_ADD_IMAGE_FLAG_SINGLE;
         }
+        if (settings->progressive) {
```
wantehchang: The conditiona should be `settings->progressive || settings->layered`.

But I believe we don't need to add the `AVIF_ADD_IMAGE_FLAG_PROGRESSIVE` flag. Please see my comment at avif.h:1638.

include/avif/avif.h
juliobbv-p: Thanks for the info! I read the AVIF spec and the aom layer logic, and I was wondering how libavif's implementation of `settings->progressive` and `settings->layered` differed once they went through the encoder layer level 😅.

apps/avifenc.c
juliobbv-p: I think there's still value in telling `settings->progressive` and `settings->layered` apart. This is because the libavif UX syntax for layered allows for updating the quality level for each presented frame, and layered use cases don't have the same total overhead/file size ratio constraints requirements that progressive has, so in theory there's no reason for layered to switch away from good mode.

Let's talk about this during our meeting tomorrow.

include/avif/avif.h
juliobbv-p: Added period. Flag pending (will resolve with other comment).

include/avif/avif.h
wantehchang: Julio and I met and discussed this issue today. Here are the notes from our meeting.

---

Background:

AVIF spec:

1. The AVIF spec specifies layered image items but not layered image sequences.
2. The AV1 bitstream for a layered image item may have up to 4 spatial layers. Temporal layers are not used.
3. A layered image item may have a layer selector (`lsel`) property that selects one of the spatial layers in the AV1 bitstream. If the `lsel` property has the special value 0xFFFF, it means the application can either decode and render all the spatial layers progressively, or render only the highest spatial layer.
4. If a layered image item does not have a `lsel` property, it should be treated as if it has a `lsel` property with the special value 0xFFFF.

The libavif encoder:

1. The libavif encoder only encodes layered image items for progressive encoding.
2. For backward compatibility, the libavif encoder does not add a `lsel` property to layered image items. (You can verify this by searching for "lsel" in src/write.c.)

The avifenc program:

1. Both the `--progressive` and `--layered` options encode a layered image item.
2. The `--progressive` option is automatic layered encoding, a simple, turnkey solution.
3. The `--layered` option is manual layered encoding. It is intended for encoding a layered image item suitable for progressive decoding, even though it is more flexible than the intended purpose.

---

Proposed approach:

The libavif encoder can detect whether it is encoding a layered image item by checking if `encoder->extraLayerCount > 0`. But we cannot blindly use all-intra mode when encoding a layered image item.

All-intra mode performs better than good-quality mode only when a layered image item is encoded in certain ways. So either the libavif encoder needs to detect when all-intra mode should be used, or the libavif encoder needs to be told to use all-intra mode. The latter requires a change to the `avifEncoder` interface.

My preference is to have the libavif encoder detect when all-intra mode should be used, for two reasons:
- we should not burden the user with this decision, and
- no change to the `avifEncoder` interface is needed.

Julio and I tentatively agreed on the following criteria:
- There are two spatial layers, i.e., `encoder->extraLayerCount == 1`; and
- The quality of the base layer is very low.

In addition, when the `--layered` option is used, avifenc should constrain the quality levels and resolutions for the frames in a layered image item so that the encoded layered image item is suitable for progressive decoding. For example, the quality levels and resolutions for the frames should not decrease.

src/codec_aom.c
```diff
@@ -671,7 +671,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     unsigned int aomUsage = AOM_USAGE_GOOD_QUALITY;
     // Use AOM_USAGE_ALL_INTRA (added in https://crbug.com/aomedia/2959) for still image encoding if available.
 #if defined(AOM_USAGE_ALL_INTRA)
-    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) {
+    if (addImageFlags & (AVIF_ADD_IMAGE_FLAG_SINGLE | AVIF_ADD_IMAGE_FLAG_PROGRESSIVE)) {
```
wantehchang: Julio: At the place where we decide whether a layered image item should be encoded in all-intra mode, please add a comment to explain why. You can use your explanation in the commit message as a basis for the comment.

apps/avifenc.c
juliobbv-p: I discussed this item with Wan-Teh. We agreed on having libavif switch between all-intra and good-quality mode with a simple heuristic, explained in detail by Wan-Teh's other PR comment.

I'll resolve this comment and have that work tracked by the other thread.

include/avif/avif.h
juliobbv-p: I've pushed changes to my PR to implement the all-intra mode layered image encoding criteria that @wantehchang and I discussed.

src/codec_aom.c
```diff
@@ -671,7 +673,14 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     unsigned int aomUsage = AOM_USAGE_GOOD_QUALITY;
     // Use AOM_USAGE_ALL_INTRA (added in https://crbug.com/aomedia/2959) for still image encoding if available.
 #if defined(AOM_USAGE_ALL_INTRA)
-    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) {
+    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE ||
+        (encoder->extraLayerCount == 1 && quality <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD)) {
```
wantehchang: Here `quality` is the quality of the current frame, not the quality of the first frame.

The code here was originally inside an if block for the first frame. But it was recently moved out of that if block. So the best solution is to move this code (and the code that depends on it) back into that if block. In this CL, I suggest a simpler solution: save the quality of the first layer in the `codec->internal` structure. Please see the following patch:

```
diff --git a/src/codec_aom.c b/src/codec_aom.c
index 657e8d33..b54d327b 100644
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -67,6 +67,7 @@ struct avifCodecInternal
     aom_img_fmt_t aomFormat;
     uint32_t currentLayer;
+    int qualityFirstLayer;
 #endif
 };
 
@@ -673,8 +674,11 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     unsigned int aomUsage = AOM_USAGE_GOOD_QUALITY;
     // Use AOM_USAGE_ALL_INTRA (added in https://crbug.com/aomedia/2959) for still image encoding if available.
 #if defined(AOM_USAGE_ALL_INTRA)
+    if (encoder->extraLayerCount > 0 && codec->internal->currentLayer == 0) {
+        codec->internal->qualityFirstLayer = quality;
+    }
     if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE ||
-        (encoder->extraLayerCount == 1 && quality <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD)) {
+        (encoder->extraLayerCount == 1 && codec->internal->qualityFirstLayer <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD)) {
         // All-intra mode is beneficial when encoding a two-layer image item, and the quality of the first layer is very low.
         // Switching to all-intra mode comes with the following benefits:
         // - The first layer will be smaller than the second layer (which is often not the case with good mode)
@@ -690,7 +694,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         if (aomCpuUsed >= 7) {
 #if defined(AOM_USAGE_ALL_INTRA) && defined(ALL_INTRA_HAS_SPEEDS_7_TO_9)
             if (!(addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE ||
-                  (encoder->extraLayerCount == 1 && quality <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD))) {
+                  (encoder->extraLayerCount == 1 && codec->internal->qualityFirstLayer <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD))) {
             }
 #else
```

src/codec_aom.c
```diff
@@ -680,7 +689,8 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         aomCpuUsed = AVIF_CLAMP(encoder->speed, 0, 9);
         if (aomCpuUsed >= 7) {
 #if defined(AOM_USAGE_ALL_INTRA) && defined(ALL_INTRA_HAS_SPEEDS_7_TO_9)
-            if (!(addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE)) {
+            if (!(addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE ||
+                  (encoder->extraLayerCount == 1 && quality <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD))) {
```
wantehchang: It would be good to avoid duplicating this complicated expression, first evaluated at line 676. We can save the result of the expression in a local variable.

src/codec_aom.c
```diff
@@ -671,7 +673,14 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     unsigned int aomUsage = AOM_USAGE_GOOD_QUALITY;
     // Use AOM_USAGE_ALL_INTRA (added in https://crbug.com/aomedia/2959) for still image encoding if available.
 #if defined(AOM_USAGE_ALL_INTRA)
-    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) {
+    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE ||
```
wantehchang: Nit: Add parentheses around `addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE`

src/codec_aom.c
wantehchang, line 865: This probably should still be the `else` block for `if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE)`:

```
        if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) {
            // Set the maximum number of frames to encode to 1. This instructs
            // libaom to set still_picture and reduced_still_picture_header to
            // 1 in AV1 sequence headers.
            cfg->g_limit = 1;
        } else {
            if (encoder->keyframeInterval > 0) {
                cfg->kf_max_dist = encoder->keyframeInterval;
            }
        }
```

src/codec_aom.c
```diff
@@ -826,7 +836,9 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             // libaom to set still_picture and reduced_still_picture_header to
             // 1 in AV1 sequence headers.
             cfg->g_limit = 1;
-
+        }
+        if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE ||
+            (encoder->extraLayerCount == 1 && quality <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD)) {
```
wantehchang: Use the local variable I suggest at line 693 here.

src/codec_aom.c
```diff
@@ -671,7 +673,14 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
 #if defined(AOM_USAGE_ALL_INTRA)
-    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) {
+    if (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE ||
+        (encoder->extraLayerCount == 1 && quality <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD)) {
+        // All-intra mode is beneficial when encoding a two-layer image item, and the quality of the first layer is very low.
+        // Switching to all-intra mode comes with the following benefits:
+        // - The first layer will be smaller than the second layer (which is often not the case with good mode)
+        // - Outputs have predictable file sizes: the sum of the first layer (quality <= 10) plus the second layer (quality set by the caller)
+        // - Because the first layer is very small, progressive encoding overhead is also smaller and more stable (about 5-8% for quality 40 and 2-4% for quality 60)
+        // - Availability of choosing tune IQ (which requires all-intra mode)
```
wantehchang: Nit: Availability => Option ?

src/codec_aom.c
juliobbv-p: Thanks @wantehchang! Patch applied.

My proposal assumed that setting `AOM_USAGE_*` can only happen during encoder initialization, so only quality of the first layer is implicitly taken into consideration. Making it explicit is better though.

src/codec_aom.c
juliobbv-p: I don't believe so (correct me if I'm wrong), because setting `kf_max_dist > 0` would only be applicable when we configure the encoder to allow for inter frame encoding per our criteria (i.e. not for single images or 2 layer images with a first layer q<=10).

include/avif/avif.h
juliobbv-p: I've pushed more changes to address Wan-Teh's feedback. Note that I'm now referring to all-intra "encoding" instead of "mode" or "usage" because all-intra in AVM (and old AOM) happens by configuring the encoder to use intra frames only.

src/codec_aom.c
wantehchang: You are right. I later also realized that my suggestion was wrong because this `else` block sets `cfg->kf_max_dist`, which must be set to 0 if we are doing all-intra encoding. So the condition for this `else` block should be "we are not doing all-intra encoding". Sorry about my confusion.

src/codec_aom.c
wantehchang: Glad to hear you and I are on the same page. The change I proposed is to avoid confusing people who don't know this. Eventually we should move this code back into the if block that initializes the encoder.

src/codec_aom.c
```diff
@@ -656,6 +659,21 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                                       avifAddImageFlags addImageFlags,
                                       avifCodecEncodeOutput * output)
 {
+    // Save the quality of the first layer because it's useful for further decision making
+    if (encoder->extraLayerCount > 0 && codec->internal->currentLayer == 0) {
+        codec->internal->qualityFirstLayer = quality;
+    }
+
+    // All-intra encoding is beneficial when encoding a two-layer image item, and the quality of the first layer is very low.
```
wantehchang: Nit: Delete the comma before "and the quality of the first layer is very low."

src/codec_aom.c
```diff
@@ -656,6 +659,21 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+    // Switching to all-intra encoding comes with the following benefits:
+    // - The first layer will be smaller than the second layer (which is often not the case with good mode)
```
wantehchang: Nit: good => good-quality

src/codec_aom.c
```diff
@@ -656,6 +659,21 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+    // - Outputs have predictable file sizes: the sum of the first layer (quality <= 10) plus the second layer (quality set by the caller)
+    // - Because the first layer is very small, progressive encoding overhead is also smaller and more stable (about 5-8% for quality 40 and 2-4% for quality 60)
```
wantehchang: Optional: Change "progressive" to "layered" or "multi-layer"

src/codec_aom.c
```diff
@@ -656,6 +659,21 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+    // - Option of choosing tune IQ (which requires all-intra encoding)
```
wantehchang: Nit: all-intra encoding => "all-intra mode" or "AOM_USAGE_ALL_INTRA"

src/codec_aom.c
```diff
@@ -826,7 +844,8 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             // libaom to set still_picture and reduced_still_picture_header to
             // 1 in AV1 sequence headers.
             cfg->g_limit = 1;
-
+        }
+        if (useAllIntra) {
```
wantehchang, line 848: Optional: The code inside the `if` block is intended to emulate `AOM_USAGE_ALL_INTRA`, so we don't need it if we have `AOM_USAGE_ALL_INTRA`.

```
        if (useAllIntra) {
#if !defined(AOM_USAGE_ALL_INTRA)
            // Use the default settings of the new AOM_USAGE_ALL_INTRA (added in
            // https://crbug.com/aomedia/2959).
            //
            // Set g_lag_in_frames to 0 to reduce the number of frame buffers
            // (from 20 to 2) in libaom's lookahead structure. This reduces
            // memory consumption when encoding a single image.
            cfg->g_lag_in_frames = 0;
            // Disable automatic placement of key frames by the encoder.
            cfg->kf_mode = AOM_KF_DISABLED;
            // Tell libaom that all frames will be key frames.
            cfg->kf_max_dist = 0;
#endif
        } else {
            if (encoder->keyframeInterval > 0) {
                cfg->kf_max_dist = encoder->keyframeInterval;
            }
        }
```

src/codec_avm.c
```diff
@@ -31,6 +33,7 @@ struct avifCodecInternal
     // avifEncoderSetCodecSpecificOption(encoder, "tune", value) call.
     avifBool tuningSet;
     uint32_t currentLayer;
+    int qualityFirstLayer;
```
wantehchang: Undo this change.

src/codec_avm.c does not need `qualityFirstLayer` because we only need to change the code inside the if block that initializes the encoder.

src/codec_avm.c
```diff
@@ -467,6 +470,20 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
                                       avifAddImageFlags addImageFlags,
                                       avifCodecEncodeOutput * output)
 {
+    // Save the quality of the first layer because it's useful for further decision making
+    if (encoder->extraLayerCount > 0 && codec->internal->currentLayer == 0) {
+        codec->internal->qualityFirstLayer = quality;
+    }
+
+    // All-intra encoding is beneficial when encoding a two-layer image item, and the quality of the first layer is very low.
```
wantehchang: Nit: Delete the comma before "and the quality of the first layer is very low."

src/codec_avm.c
```diff
@@ -467,6 +470,20 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
+    // Switching to all-intra encoding comes with the following benefits:
+    // - The first layer will be smaller than the second layer (which is often not the case with good mode)
```
wantehchang: Nit: good => good-quality

src/codec_avm.c
```diff
@@ -467,6 +470,20 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
+    // - Outputs have predictable file sizes: the sum of the first layer (quality <= 10) plus the second layer (quality set by the caller)
+    // - Because the first layer is very small, progressive encoding overhead is also smaller and more stable (about 5-8% for quality 40 and 2-4% for quality 60)
```
wantehchang: Optional: Change "progressive" to "layered" or "multi-layer"

src/codec_avm.c
```diff
@@ -467,6 +470,20 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
+    avifBool useAllIntraForLayered = encoder->extraLayerCount == 1 &&
+                                     codec->internal->qualityFirstLayer <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD;
+    avifBool useAllIntra = (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) || useAllIntraForLayered;
```
wantehchang: Move this block of code (lines 473-485) right before where `useAllIntra` is used (at line 568) or some other suitable place inside the `if (!codec->internal->encoderInitialized)` block.

Then change `codec->internal->qualityFirstLayer` back to `quality`.

src/codec_aom.c
```diff
@@ -49,6 +49,8 @@
 #endif
 #endif
 
+#define TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD 10
```
wantehchang, line 52: Optional: Some projects recommend that the subject line of a commit message be <= 50 characters. Where possible please aim for that. This often requires using abbreviations or being creative.

src/codec_aom.c
juliobbv-p: Thanks for the catch! I over-corrected this part by accident.

FooIbar:
Should this be mentioned in the changelog?

PR #2975, opened 2026-01-27
Do not STRING REGEX when the string is empty

vrabaud:


cmake/Modules/merge_static_libs.cmake
```diff
@@ -16,7 +16,10 @@ function(avif_collect_deps target out_deps)
             set(in_special_block FALSE) # Latch off
         elseif(NOT in_special_block)
             string(REGEX REPLACE "\\\$<BUILD_INTERFACE:([^>]*)>" "\\1" dep ${dep})
-            string(REGEX REPLACE "\\\$<LINK_ONLY:([^>]*)>" "\\1" dep ${dep})
+            # dep could now be empty if dep was <BUILD_INTERFACE:> before the replace.
+            if(dep)
+                string(REGEX REPLACE "\\\$<LINK_ONLY:([^>]*)>" "\\1" dep ${dep})
+            endif()
             if(TARGET ${dep})
```
wantehchang, line 23: Vincent: Should we check for an empty `dep` in this line? Or does `TARGET ${dep}` handle an empty `dep` gracefully?

cmake/Modules/merge_static_libs.cmake
```diff
@@ -16,7 +16,10 @@ function(avif_collect_deps target out_deps)
             set(in_special_block FALSE) # Latch off
         elseif(NOT in_special_block)
             string(REGEX REPLACE "\\\$<BUILD_INTERFACE:([^>]*)>" "\\1" dep ${dep})
```
wantehchang, line 18: I guess by definition (line 12: `foreach(dep IN LISTS current_deps)`), `dep` cannot be empty in this line?

cmake/Modules/merge_static_libs.cmake
```diff
@@ -16,7 +16,10 @@ function(avif_collect_deps target out_deps)
...
-            string(REGEX REPLACE "\\\$<LINK_ONLY:([^>]*)>" "\\1" dep ${dep})
+            # dep could now be empty if dep was <BUILD_INTERFACE:> before the replace.
```
wantehchang, line 19: Did you mean `dep` was exactly `<BUILD_INTERFACE:>`, with nothing between `:` and `>`?

Do you know why `dep` was equal to `<BUILD_INTERFACE:>`?

cmake/Modules/merge_static_libs.cmake
vrabaud: TARGET ${dep} does, I tried

cmake/Modules/merge_static_libs.cmake
vrabaud: exactly

cmake/Modules/merge_static_libs.cmake
vrabaud: exactly. Usually, there is nothing in there but Threads. On some platforms, it is not even there and is therefore empty.

wantehchang: Do you know why we only ran into this issue when we upgraded SVT-AV1 to v4.0.0? I assume it's something in the `CMakeLists.txt` file in SVT-AV1 v4.0.0.

vrabaud: From what I understand, it is due to https://gitlab.com/AOMediaCodec/SVT-AV1/-/commit/82418552281a55f3708caa5beb0bfa08a026fb03 , which removed Threads and m on certain platforms, which lead to the empty list we found.

Issue #2979, opened 2026-01-30
Potential null pointer dereference in `avifImageApplyExpression32b` during Sample Transform handling


X-ANOY:
Hi, I found a potential null pointer dereference while fuzzing libavif. The issue is triggered when decoding a malformed AVIF file using the default decoder configuration created via `avifDecoderCreate()`.

The issue is related to the handling of the `Sample Transform` (stpe) property. While the metadata is parsed successfully, the decoder proceeds to pixel calculations without ensuring the required image planes have been allocated, leading to a SEGV in the transformation logic.

The segmentation fault occurs in `src/sampletransform.c` within `avifImageApplyExpression32b`.

The implementation assumes that if a transform is defined, the necessary image planes for the target channels are present. It calls `avifImagePlane(image, c)` and immediately performs pointer arithmetic:
`const uint8_t * row = avifImagePlane(image, c) + avifImagePlaneRowBytes(image, c) * y;`

In the provided sample, `avifImagePlane` returns `NULL`. The lack of a null-check before dereferencing or performing arithmetic on this pointer causes the crash.

### **Reproduction**

**1.`poc.c`**:

```c
#include <avif/avif.h>
#include <stdio.h>
#include <stdlib.h>

int main(int argc, char **argv) {
    if (argc < 2) return 1;

    FILE *f = fopen(argv[1], "rb");
    if (!f) return 1;
    fseek(f, 0, SEEK_END);
    size_t size = ftell(f);
    fseek(f, 0, SEEK_SET);
    uint8_t *data = malloc(size);
    fread(data, 1, size, f);
    fclose(f);

    if (size < 12) return 1;

    avifDecoder *decoder = avifDecoderCreate();
    avifImage *image = avifImageCreateEmpty();

    decoder->ignoreExif      = (data[0] & 0x01) ? AVIF_TRUE : AVIF_FALSE;
    decoder->ignoreXMP       = (data[0] & 0x02) ? AVIF_TRUE : AVIF_FALSE;
    decoder->allowProgressive = (data[0] & 0x04) ? AVIF_TRUE : AVIF_FALSE;
    decoder->allowIncremental = (data[0] & 0x08) ? AVIF_TRUE : AVIF_FALSE;

    decoder->imageContentToDecode = 0;
    if (data[1] & 0x01) decoder->imageContentToDecode |= AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA;
[...]

PR #2980, opened 2026-01-30
Check planes in avifImageApplyExpression32b()

y-guyon:
See https://github.com/AOMediaCodec/libavif/issues/2979.

### DIFF
--- a/src/sampletransform.c
+++ b/src/sampletransform.c
@@ -314,7 +314,9 @@ static avifResult avifImageApplyExpression32b(avifImage * dstImage,
                         stack[stackSize++] = token->constant;
                     } else if (token->type == AVIF_SAMPLE_TRANSFORM_INPUT_IMAGE_ITEM_INDEX) {
                         const avifImage * image = inputImageItems[token->inputImageItemIndex - 1]; // 1-based
-                        const uint8_t * row = avifImagePlane(image, c) + avifImagePlaneRowBytes(image, c) * y;
+                        const uint8_t * row = avifImagePlane(image, c);
+                        AVIF_ASSERT_OR_RETURN(row != NULL);
+                        row += avifImagePlaneRowBytes(image, c) * y;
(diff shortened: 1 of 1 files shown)


src/sampletransform.c
```diff
@@ -333,7 +335,9 @@ static avifResult avifImageApplyExpression32b(avifImage * dstImage,
                 // Fit to 'pixi'-defined range. TODO(yguyon): Take avifRange into account.
                 stack[0] = AVIF_CLAMP(stack[0], minValue, maxValue);
 
-                uint8_t * row = avifImagePlane(dstImage, c) + avifImagePlaneRowBytes(dstImage, c) * y;
+                uint8_t * row = avifImagePlane(dstImage, c);
+                AVIF_ASSERT_OR_RETURN(row != NULL);
```
wantehchang, line 339: Note: This comment is just a question, not a request for change.

`row` cannot possibly be NULL here, right?

If `avifImagePlane(dstImage, c)` returns NULL, then both `avifImagePlaneWidth(dstImage, c)` and `avifImagePlaneHeight(dstImage, c)` should return 0 and we won't reach here.

To fix issue https://github.com/AOMediaCodec/libavif/issues/2979, the change at lines 317-319 should suffice, right?

Since `row` cannot possibly be NULL here, the use of `AVIF_ASSERT_OR_RETURN` is appropriate.

y-guyon: > If `avifImagePlane(dstImage, c)` returns NULL, then both `avifImagePlaneWidth(dstImage, c)` and `avifImagePlaneHeight(dstImage, c)` should return 0 and we won't reach here.

`avifImagePlaneWidth()` does not check for plane allocation, so it can return anything even if `avifImagePlane()` returns NULL.

> To fix issue https://github.com/AOMediaCodec/libavif/issues/2979, the change at lines 317-319 should suffice, right?

Yes, in the sense that it would segfault or return an error earlier. But the change at lines 317-319 should not have been necessary either, so I went for extra certainty with the change at lines 338-340.

PR #2981, opened 2026-01-30
Skip sato tiles depending on imageContentToDecode

y-guyon:
Skip sato tiles depending on imageContentToDecode

Do not generate tiles for Sample Transform input image items unless `imageContentToDecode&AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA!=0`.
Explicitly test any expected `avifItemCategory` in `avifDecoderReset()` and return an error for unlisted `avifItemCategory` values.
Also guard `avifDecoderApplySampleTransform()` with `imageContentToDecode` in case there are other tiles, such as gain map items.
Add test for the non-gain-map case because libavif does not support encoding gain maps in the same file as the 'sato' items.

With this change, https://github.com/AOMediaCodec/libavif/issues/2979 no longer leads to a segmentation fault, because there is no generated tile (`decoder->data->tiles.count == 0`) so `avifDecoderNextImage` returns `AVIF_RESULT_NO_CONTENT`. However, in the case where there would still be other generated tiles (e.g. from an `AVIF_ITEM_GAIN_MAP`), this single edit is not sufficient to avoid calling `avifDecoderApplySampleTransform()`. This is why `avifDecoderApplySampleTransform()` is also guarded by `imageContentToDecode`.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -6357,6 +6357,7 @@ avifResult avifDecoderReset(avifDecoder * decoder)
             if (!mainItems[c]) {
                 continue;
             }
+            AVIF_ASSERT_OR_RETURN(c != AVIF_ITEM_SAMPLE_TRANSFORM); // See sampleTransformItem.
 
             if (avifIsAlpha((avifItemCategory)c) && !mainItems[c]->width && !mainItems[c]->height) {
                 // NON-STANDARD: Alpha subimage does not have an ispe property; adopt width/height from color item
@@ -6938,7 +6939,7 @@ avifResult avifDecoderNextImage(avifDecoder * decoder)
         AVIF_ASSERT_OR_RETURN(prepareTileResult[c] == AVIF_RESULT_OK);
     }
 
-    if (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA && decoder->data->meta->sampleTransformExpression.count > 0) {
+    if ((decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) && decoder->data->meta->sampleTransformExpression.count > 0) {
         // TODO(yguyon): Add a field in avifDecoder and only perform sample transformations upon request.
         AVIF_CHECKRES(avifDecoderApplySampleTransform(decoder, decoder->image));
     }
--- a/src/write.c
+++ b/src/write.c
@@ -1955,7 +1955,7 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
             encoder->sampleTransformRecipe == AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_4B ||
             encoder->sampleTransformRecipe == AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_8B_OVERLAP_4B) {
             // For now, only 16-bit depth is supported.
-            AVIF_ASSERT_OR_RETURN(firstCell->depth == 16);
+            AVIF_CHECKERR(firstCell->depth == 16, AVIF_RESULT_NOT_IMPLEMENTED);
             // TODO: b/480081865 - Support gain maps in same file as Sample Transforms
             AVIF_CHECKERR(!firstCell->gainMap, AVIF_RESULT_NOT_IMPLEMENTED);
             AVIF_CHECKRES(avifEncoderCreateBitDepthExtensionItems(encoder, gridCols, gridRows, gridWidth, gridHeight, colorItemID));


src/write.c
```diff
@@ -1956,7 +1956,8 @@ static avifResult avifEncoderAddImageInternal(avifEncoder * encoder,
             encoder->sampleTransformRecipe == AVIF_SAMPLE_TRANSFORM_BIT_DEPTH_EXTENSION_12B_8B_OVERLAP_4B) {
             // For now, only 16-bit depth is supported.
             AVIF_ASSERT_OR_RETURN(firstCell->depth == 16);
-            AVIF_CHECKERR(!firstCell->gainMap, AVIF_RESULT_NOT_IMPLEMENTED); // TODO(yguyon): Implement 16-bit HDR
+            // TODO: b/480081865 - Support gain maps in same file as Sample Transforms
+            AVIF_CHECKERR(!firstCell->gainMap, AVIF_RESULT_NOT_IMPLEMENTED);
```
wantehchang, line 1960: Should we also change line 1958 to return `AVIF_RESULT_NOT_IMPLEMENTED`?

```
            // For now, only 16-bit depth is supported.
            AVIF_CHECKERR(firstCell->depth == 16, AVIF_RESULT_NOT_IMPLEMENTED);
```

Or have we already checked that bit depth is 16 earlier, so this check should not fail?

src/read.c
```diff
@@ -6367,8 +6367,17 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
-            if (!(decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) && (c == AVIF_ITEM_COLOR || c == AVIF_ITEM_ALPHA)) {
-                continue;
+            if (c == AVIF_ITEM_COLOR || c == AVIF_ITEM_SAMPLE_TRANSFORM_INPUT_0_COLOR ||
+                c == AVIF_ITEM_SAMPLE_TRANSFORM_INPUT_1_COLOR || c == AVIF_ITEM_ALPHA ||
+                c == AVIF_ITEM_SAMPLE_TRANSFORM_INPUT_0_ALPHA || c == AVIF_ITEM_SAMPLE_TRANSFORM_INPUT_1_ALPHA) {
+                if (!(decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA)) {
+                    continue;
+                }
+            } else {
+                AVIF_ASSERT_OR_RETURN(c == AVIF_ITEM_GAIN_MAP);
```
wantehchang, line 6378: This assertion assumes that `mainItems[AVIF_ITEM_SAMPLE_TRANSFORM]` is null, so if `c == AVIF_ITEM_SAMPLE_TRANSFORM`, we have continued from line 6358:

```
            if (!mainItems[c]) {
                continue;
            }
```

This assumption is not obvious, so it would be good to add a comment, or add an assertion after line 6359:

```
            if (!mainItems[c]) {
                continue;
            }
            AVIF_ASSERT_OR_RETURN(c != AVIF_ITEM_SAMPLE_TRANSFORM);
```

Note: My analysis shows that `AVIF_ITEM_SAMPLE_TRANSFORM` is only used in src/write.c. So this is a simple way to prove that `mainItems[AVIF_ITEM_SAMPLE_TRANSFORM]` must be null here.

src/read.c
```diff
@@ -6929,7 +6938,7 @@ avifResult avifDecoderNextImage(avifDecoder * decoder)
         AVIF_ASSERT_OR_RETURN(prepareTileResult[c] == AVIF_RESULT_OK);
     }
 
-    if (decoder->data->meta->sampleTransformExpression.count > 0) {
+    if (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA && decoder->data->meta->sampleTransformExpression.count > 0) {
```
wantehchang: Nit: Add parentheses around `decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA`:

```
    if ((decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) && decoder->data->meta->sampleTransformExpression.count > 0) {
```

src/read.c
y-guyon: Done. See also:

https://github.com/AOMediaCodec/libavif/blob/d55eb9f414642f0b93c84026d31b6e2fa535400c/src/read.c#L6243

src/write.c
y-guyon: Done.

src/read.c
wantehchang: Thank you for explaining that `AVIF_ITEM_SAMPLE_TRANSFORM` is used through `sampleTransformItem` rather than `mainItems` in src/read.c.

PR #2986, opened 2026-01-31
numInputImageItems to avifImageApplyExpression32b

wantehchang:
Pass numInputImageItems (the size of the inputImageItems array) to avifImageApplyExpression32b(). Validate the index for the inputImageItems array.

### DIFF
--- a/src/sampletransform.c
+++ b/src/sampletransform.c
@@ -283,6 +283,7 @@ AVIF_ARRAY_DECLARE(avifSampleTransformStack32b, int32_t, elements);
 
 static avifResult avifImageApplyExpression32b(avifImage * dstImage,
                                               const avifSampleTransformExpression * expression,
+                                              uint8_t numInputImageItems,
                                               const avifImage * inputImageItems[],
                                               avifPlanesFlags planes,
                                               int32_t * stack,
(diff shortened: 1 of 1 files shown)


src/sampletransform.c
```diff
@@ -313,7 +314,10 @@ static avifResult avifImageApplyExpression32b(avifImage * dstImage,
                         AVIF_ASSERT_OR_RETURN(stackSize < stackCapacity);
                         stack[stackSize++] = token->constant;
                     } else if (token->type == AVIF_SAMPLE_TRANSFORM_INPUT_IMAGE_ITEM_INDEX) {
-                        const avifImage * image = inputImageItems[token->inputImageItemIndex - 1]; // 1-based
+                        // inputImageItemIndex is 1-based.
+                        const uint8_t zeroBasedIndex = token->inputImageItemIndex - 1;
```
vrabaud, line 318: There is absolutely no way inputImageItemIndex is 0 I guess?

y-guyon: Indeed. It is checked in `avifSampleTransformExpressionIsValid()`, but so is `inputImageItemIndex <= numInputImageItems`.

wantehchang: Yannis: The reason I wrote this PR is that it looks bad to pass an array but not the array size to a function. But then the function should use the array size, and the only meaningful way to use the array size in this function is to validate the array index `token->inputImageItemIndex`.

When I wrote this PR, I didn't analyze the code well enough to see if the caller has already validated `token->inputImageItemIndex`. Since that is true, this pull request will validate `token->inputImageItemIndex` redundantly. So I will abandon this pull request.

Issue #2983, opened 2026-01-30
`avifDecoder` should disallow setting only one of the (1 << 0) and (1 << 1) bits in imageContentToDecode


wantehchang:
The `AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA` enum constant is the bitwise-OR of the `(1 << 0)` and `(1 << 1)` bits. It has the following comment:

```
    // Color only or alpha only is not currently supported.
    AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA = (1 << 0) | (1 << 1),
```

`avifDecoder` should disallow color only or alpha only. I.e., if one of the `(1 << 0)` and `(1 << 1)` bits is set in `imageContentToDecode`, the other one must also be set.

PR #2988, opened 2026-02-02
Fail if imageContentToDecode is color xor alpha

y-guyon:
Fixes https://github.com/AOMediaCodec/libavif/issues/2983

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -5229,6 +5229,12 @@ avifResult avifDecoderParse(avifDecoder * decoder)
     if ((decoder->imageSizeLimit > AVIF_DEFAULT_IMAGE_SIZE_LIMIT) || (decoder->imageSizeLimit == 0)) {
         return AVIF_RESULT_NOT_IMPLEMENTED;
     }
+    // Color only or alpha only is not currently supported.
+    if ((decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) != 0 &&
+        (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) != AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) {
+        avifDiagnosticsPrintf(&decoder->diag, "imageContentToDecode set to only color or only alpha is not supported");
+        return AVIF_RESULT_NOT_IMPLEMENTED;
+    }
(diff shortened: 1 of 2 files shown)


src/read.c
```diff
@@ -6012,6 +6012,13 @@ avifResult avifDecoderReset(avifDecoder * decoder)
 
     memset(&decoder->ioStats, 0, sizeof(decoder->ioStats));
 
+    // Color only or alpha only is not currently supported.
+    if ((decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) != 0 &&
+        (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) != AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) {
+        avifDiagnosticsPrintf(&decoder->diag, "imageContentToDecode set to only color or only alpha is not supported");
+        return AVIF_RESULT_NOT_IMPLEMENTED;
+    }
```
wantehchang, line 6026: My first reaction is that we should perform this check in `avifDecoderParse()`. Did you decide to perform this check in `avifDecoderReset()` because `decoder->imageContentToDecode` is read in `avifDecoderReset()` and `avifDecoderDecodedRowCount()`?

wantehchang (approved):
LGTM. Thanks!

src/read.c
```diff
@@ -6012,6 +6012,13 @@ avifResult avifDecoderReset(avifDecoder * decoder)
 
     memset(&decoder->ioStats, 0, sizeof(decoder->ioStats));
 
+    // Color only or alpha only is not currently supported.
```
wantehchang, line 6021: It would be good to add a test to tests/gtest/avifdecodetest.cc.

src/read.c
y-guyon: Yes. I also added it in `avifDecoderParse()` because it looks like the proper location too.

src/read.c
y-guyon: Done.

src/read.c
wantehchang: Sounds good, thanks. Note that it suffices to just check in `avifDecoderReset()` because `avifDecoderParse()` calls `avifDecoderReset()`.

wantehchang (commented):
The new commit LGTM. Thanks!

PR #2991, opened 2026-02-03
Add a unit test for issue #2979

wantehchang:
Based on the PoC in https://github.com/AOMediaCodec/libavif/issues/2979.

### DIFF
--- a/tests/gtest/avifaltrtest.cc
+++ b/tests/gtest/avifaltrtest.cc
@@ -81,7 +81,7 @@ TEST(AltrTest, ZeroImageContentToDecode) {
   ImagePtr image(avifImageCreateEmpty());
   ASSERT_NE(image, nullptr);
   ASSERT_EQ(avifDecoderReadFile(decoder.get(), image.get(), file_path.c_str()),
-            AVIF_RESULT_INTERNAL_ERROR);
+            AVIF_RESULT_NO_CONTENT);
 }
 
 //------------------------------------------------------------------------------


tests/gtest/avifaltrtest.cc
```diff
@@ -70,6 +70,20 @@ TEST(AltrTest, SampleTransformDepthParseNextEqual) {
...
+  const std::string file_path =
+      std::string(data_path) + "weld_sato_12B_8B_q0.avif";
+
+  DecoderPtr decoder(avifDecoderCreate());
+  ASSERT_NE(decoder, nullptr);
+  decoder->imageContentToDecode = AVIF_IMAGE_CONTENT_NONE;
+  ImagePtr image(avifImageCreateEmpty());
+  ASSERT_NE(image, nullptr);
+  ASSERT_EQ(avifDecoderReadFile(decoder.get(), image.get(), file_path.c_str()),
+            AVIF_RESULT_INTERNAL_ERROR);
```
y-guyon: This should be `AVIF_RESULT_NO_CONTENT` once https://github.com/AOMediaCodec/libavif/pull/2981 is merged.

wantehchang: Done.

To verify the fix for issue https://github.com/AOMediaCodec/libavif/issues/2979, we just need to verify the test doesn't crash with a segmentation fault; the error code isn't important. But it's good to expect the `AVIF_RESULT_NO_CONTENT` error because it looks similar to the `AVIF_IMAGE_CONTENT_NONE` flag.

PR #2998, opened 2026-02-04
Move a comment about still image encoding

wantehchang:
That comment describes the (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) check, which was moved in
https://github.com/AOMediaCodec/libavif/pull/2970.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -672,6 +672,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     // - Option of choosing tune IQ (which requires AOM_USAGE_ALL_INTRA)
     avifBool useAllIntraForLayered = encoder->extraLayerCount == 1 &&
                                      codec->internal->qualityFirstLayer <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD;
+    // Also use all-intra encoding when encoding still images.
     avifBool useAllIntra = (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) || useAllIntraForLayered;
 
     // Map encoder speed to AOM usage + CpuUsed:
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -672,6 +672,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     // - Option of choosing tune IQ (which requires AOM_USAGE_ALL_INTRA)
     avifBool useAllIntraForLayered = encoder->extraLayerCount == 1 &&
                                      codec->internal->qualityFirstLayer <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD;
+    // Also use all-intra encoding when encoding still images.
```
wantehchang, line 675: @juliobbv-p Julio: Please review. Thanks.

src/codec_aom.c
```diff
@@ -672,6 +672,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
     avifBool useAllIntra = (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) || useAllIntraForLayered;
```
y-guyon, line 676: Should `useAllIntra` be true only `#if defined(AOM_USAGE_ALL_INTRA)`?

When was `ALL_INTRA_HAS_SPEEDS_7_TO_9` introduced?

y-guyon (approved):

src/codec_aom.c
juliobbv-p: Comment LGTM 👍

src/codec_aom.c
juliobbv-p: `useAllIntra` is a more general concept: it means the encoder will be configured to only use intra frames only, either by setting `AOM_USAGE_ALL_INTRA`, or by manually configuring the encoder so all frames will be keyframes (if `AOM_USAGE_ALL_INTRA` isn't available).

wantehchang: Yannis: Speeds 7, 8, and 9 for all intra mode were introduced in [libaom v3.2.0](https://groups.google.com/a/aomedia.org/g/av1-discuss/c/xrpGhv1nr8w/m/C3FNW-UdAwAJ) (announced on 2021-10-14). I will add this version to the comment in src/codec_aom.c.

Issue #2997, opened 2026-02-04
Update avifImageUsesTuneIq() after pull request #2970


wantehchang:
@juliobbv-p Julio: I can't assign this issue to you, so I assigned it to myself as a substitute.

After pull request https://github.com/AOMediaCodec/libavif/pull/2970, `AOM_USAGE_ALL_INTRA` may be used to encode a two-layer image item. So `tune=iq` may be used to encode two AV1 frames.

This means the comment "AOM_TUNE_IQ is only supported with still images in libavif and libaom as of today" at the end of the `avifImageUsesTuneIq()` function is no longer true:

```
static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
{
    ...

    // In practice this function should also return true if avifEncoderSetCodecSpecificOption("tune", "iq")
    // was called for a previous frame and not called (or called with NULL) for this frame, because the tune
    // option persists across frames in libaom. However AOM_TUNE_IQ is only supported with still images in
    // libavif and libaom as of today, so there is no need to remember this option across frames.
    return ret;
}
```

Now we need to remember the `tune=iq` option across frames.

juliobbv-p:
Thanks @wantehchang for letting me know -- I'll take a look at this as my next item (and manually keep an eye on the issue). We can discuss this during our meeting tomorrow.

juliobbv-p:
Opened #3029 to address this issue.

wantehchang:
Fixed by Julio's PR https://github.com/AOMediaCodec/libavif/pull/3029.

PR #2999, opened 2026-02-05
Fix useTuneIq comment in aomCodecEncodeImage()

y-guyon:
The comment at the end of `avifImageUsesTuneIq()` already describes that it is only valid for the first frame, so this comment is not necessary, and `avifImageUsesTuneIq()` will be updated anyway (see https://github.com/AOMediaCodec/libavif/issues/2997).

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -733,9 +733,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
     // True if libavif knows that tune=iq is used, either by default by libavif, or explicitly set by the user.
-    // False otherwise (including if libaom uses tune=iq by default, which is not the case as of v1.13.1 and earlier versions).
-    // This is only accurate for the first frame but tune=iq is only supported for still images in libavif and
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -733,9 +733,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
     // True if libavif knows that tune=iq is used, either by default by libavif, or explicitly set by the user.
-    // False otherwise (including if libaom uses tune=iq by default, which is not the case as of v1.13.1 and earlier versions).
-    // This is only accurate for the first frame but tune=iq is only supported for still images in libavif and
-    // for all-intra coding in libaom (at least up to v1.13.1) anyway.
+    // False otherwise (including if libaom uses tune=iq by default, which is not the case as of v3.13.1 and earlier versions).
```
wantehchang, line 736: @juliobbv-p

I think we need to impose a constraint that the `tune` option can only be set before encoding starts. Then we can store the result of `avifImageUsesTuneIq()` in the `codec->internal` structure.

wantehchang (approved):

src/codec_aom.c
juliobbv-p: 100% agreed, I'll take a look at implementing this.

PR #3000, opened 2026-02-05
Return 0 dec rows when contentToDecode is gainmap

y-guyon:


src/read.c
```diff
@@ -7125,6 +7125,14 @@ static uint32_t avifGetDecodedRowCount(const avifDecoder * decoder, const avifTi
 
 uint32_t avifDecoderDecodedRowCount(const avifDecoder * decoder)
 {
+    if (decoder->data->tileInfos[AVIF_ITEM_COLOR].tileCount == 0) {
+        // decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA
+        // was likely 0 when avifDecoderNextImage() was called.
+        // avifDecoderDecodedRowCount() only describes decoder->image->yuvPlanes[0].
+        // There is no available luma plane, so return 0 decoded rows.
```
wantehchang, line 7132: Yannis: Should we describe this case (`imageContentToDecode & AVIF_IMAGE_CONTENT_GAIN_MAP` is zero) in the comment for `avifDecoderDecodedRowCount()`?

The relevant part of that comment is quoted below:

```
// ...
// If a gain map is present and  (imageContentToDecode & AVIF_IMAGE_CONTENT_GAIN_MAP) is nonzero,
// the gain map's planes can also be accessed in the same way.
// ...
// WARNING: Experimental feature.
AVIF_API uint32_t avifDecoderDecodedRowCount(const avifDecoder * decoder);
```

By the way, should we remove the experimental feature warning now?

y-guyon: > Yannis: Should we describe this case (`imageContentToDecode & AVIF_IMAGE_CONTENT_GAIN_MAP` is zero) in the comment for `avifDecoderDecodedRowCount()`?

I thought about it but to me the current description is already enough ("number of top rows that can be immediately accessed from the luma plane"):

https://github.com/AOMediaCodec/libavif/blob/2b1478469b61ade38ed21edf8ecc1a59608cb864/include/avif/avif.h#L1438-L1441

> By the way, should we remove the experimental feature warning now?

I am somewhat glad it was kept till this fix. I guess it can be removed now.

wantehchang: Yannis: My previous comment has a copy-and-paste error. But it seems that you understood what I meant to ask.

What I meant to ask is:

> Should we describe this case (`imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA` is zero) in the comment for `avifDecoderDecodedRowCount()`?

But I agree that the comment you quoted implies that in this case, since nothing is decoded into the luma plane of `decoder->image`, `avifDecoderDecodedRowCount()` should return 0.

wantehchang, line 7132: Yannis: One more question about this comment:

```
        // There is no available luma plane, so return 0 decoded rows.
```

Does this mean `decoder->image->yuvPlanes[0]` is a null pointer? Or does it mean `decoder->image->yuvPlanes[0]` is allocated but nothing is decoded into `decoder->image->yuvPlanes[0]`?

y-guyon: The intent should be a null pointer, but there is no way to signal that through the existing `avifDecoderDecodedRowCount()` API. I think it does not matter. `yuvPlanes[0]` being `NULL` or containing 0 elements cannot be accessed anyway by the user, right?

wantehchang: I was just curious what the current implementation does. I hope `decoder->image->yuvPlanes[0]` is a null pointer because we don't need to allocate `decoder->image->yuvPlanes[0]` in this case.

y-guyon: It is a null pointer.

PR #3001, opened 2026-02-05
Rely on contentToDecode only in avifDecoderReset()

y-guyon:
Avoid unexpected effects when the value of `avifDecoder::contentToDecode` is changed between `avifDecoderReset()` and `avifDecoderNextImage()`.

Follow-up of https://github.com/AOMediaCodec/libavif/pull/2981.

No test added because libavif does not support encoding files with both Sample Transforms and gain maps.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -6953,7 +6953,7 @@ avifResult avifDecoderNextImage(avifDecoder * decoder)
     }
 
     // If decoder->data->tileInfos[AVIF_ITEM_COLOR].tileCount == 0, it means
-    // decoder->imageContentToDecode was not set to AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA.
+    // decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA was equal to 0.
     // Only apply Sample Transforms if there is a color item to apply it onto.
     if (decoder->data->tileInfos[AVIF_ITEM_COLOR].tileCount != 0 && decoder->data->meta->sampleTransformExpression.count > 0) {
         // TODO(yguyon): Add a field in avifDecoder and only perform sample transformations upon request.


src/read.c
```diff
@@ -6952,7 +6952,10 @@ avifResult avifDecoderNextImage(avifDecoder * decoder)
         AVIF_ASSERT_OR_RETURN(prepareTileResult[c] == AVIF_RESULT_OK);
     }
 
-    if ((decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) && decoder->data->meta->sampleTransformExpression.count > 0) {
+    // If decoder->data->tileInfos[AVIF_ITEM_COLOR].tileCount == 0, it means
+    // decoder->imageContentToDecode was not set to AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA.
```
wantehchang: This should say "decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA was equal to 0", right?

Or we can say "decoder->imageContentToDecode was not set to a value that contains AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA"?

include/avif/avif.h
```diff
@@ -1373,6 +1373,8 @@ typedef struct avifDecoder
     // --------------------------------------------------------------------------------------------
 
     // Image content to decode (if present). Defaults to AVIF_IMAGE_CONTENT_DECODE_DEFAULT.
+    // The value is only taken into account when calling avifDecoderRead*(), avifDecoderSetSource(),
+    // avifDecoderParse(), or avifDecoderReset().
```
wantehchang, line 1377: `imageContentToDecode` is also checked in `avifDecoderDecodedRowCount()`.

wantehchang (approved):

src/read.c

include/avif/avif.h
y-guyon: It is mentioned in the comment but not directly checked:

https://github.com/AOMediaCodec/libavif/blob/835d66a5664edfdf2f4d4a3ecc122ad50797ccb3/src/read.c#L7128-L7134

wantehchang: `avifDecoderDecodedRowCount()` checks whether `imageContentToDecode` contains `AVIF_IMAGE_CONTENT_GAIN_MAP`:

```
uint32_t avifDecoderDecodedRowCount(const avifDecoder * decoder)
{
...
            if ((decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_GAIN_MAP) && gainMap != NULL && gainMap->height != 0) {
...
}
```

y-guyon: Right, sorry I missed that one.

I think https://github.com/AOMediaCodec/libavif/pull/3012 is appropriate but let me know if you believe the `imageContentToDecode` check should be kept in `avifDecoderDecodedRowCount()`.

wantehchang: You know the `avifDecoderDecodedRowCount()` code much better than I do. I trust your judgment. Thanks!

PR #3002, opened 2026-02-05
Prevent integer overflow during buffer reallocation

rootvector2:
Add pre- and post-multiplication checks when doubling allocation sizes to prevent size_t overflow leading to undersized allocations and potential heap corruption during memcpy.

Includes defensive overflow detection and early failure on unsafe growth.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -5,6 +5,7 @@
 
 #include <assert.h>
 #include <math.h>
+#include <stdint.h>
 #include <string.h>
 
 float avifRoundf(float v)
@@ -89,6 +90,11 @@ avifBool avifArrayCreate(void * arrayStruct, uint32_t elementSize, uint32_t init
     arr->elementSize = elementSize ? elementSize : 1;
     arr->count = 0;
     arr->capacity = initialCapacity;
+    if (arr->capacity > SIZE_MAX / arr->elementSize) {
+        arr->ptr = NULL;
+        arr->capacity = 0;
+        return AVIF_FALSE;
+    }
     size_t byteCount = (size_t)arr->elementSize * arr->capacity;
     arr->ptr = (uint8_t *)avifAlloc(byteCount);
     if (!arr->ptr) {
@@ -111,10 +117,8 @@ void * avifArrayPush(void * arrayStruct)
         }
 
         size_t newByteCount = oldByteCount * 2;
-
         uint8_t * newPtr = (uint8_t *)avifAlloc(newByteCount);
         if (newPtr == NULL) {
-            avifFree(oldPtr);
             return NULL;
         }
 


src/utils.c
```diff
@@ -105,7 +105,23 @@ void * avifArrayPush(void * arrayStruct)
...
+        // Check for overflow before doubling the allocation size
+        // If oldByteCount > SIZE_MAX/2, then oldByteCount * 2 would overflow
+        if (oldByteCount > SIZE_MAX / 2) {
+            // Cannot safely double the allocation size
+            return NULL;
+        }
+        
+        size_t newByteCount = oldByteCount * 2;
+        
+        // Additional safety check: verify the multiplication didn't overflow
```
wantehchang: Delete this redundant check. This is an alternative way to detect unsigned integer overflow.

I understand your comment about defense in depth, but that defense in depth should use different protections.

src/utils.c
```diff
@@ -105,7 +105,23 @@ void * avifArrayPush(void * arrayStruct)
     if (arr->count == arr->capacity) {
         uint8_t * oldPtr = arr->ptr;
         size_t oldByteCount = (size_t)arr->elementSize * arr->capacity;
-        arr->ptr = (uint8_t *)avifAlloc(oldByteCount * 2);
+        
+        // Check for overflow before doubling the allocation size
```
wantehchang: We can also check for overflow before doubling `arr->capacity` (of type `uint32_t`). See line 130:

```
        arr->capacity *= 2;
```

src/utils.c
```diff
@@ -105,7 +105,23 @@ void * avifArrayPush(void * arrayStruct)
     if (arr->count == arr->capacity) {
         uint8_t * oldPtr = arr->ptr;
         size_t oldByteCount = (size_t)arr->elementSize * arr->capacity;
-        arr->ptr = (uint8_t *)avifAlloc(oldByteCount * 2);
+
```
wantehchang: The multiplication `(size_t)arr->elementSize * arr->capacity` in line 107 cannot overflow because it is the size of the current `arr->ptr` buffer.

On the other hand, the `avifArrayCreate()` function above performs the same multiplication (line 92), but that multiplication may overflow because it is before the `avifAlloc()` call (line 93). I just wrote pull request https://github.com/AOMediaCodec/libavif/pull/3005 to prevent overflow in that multiplication.

src/utils.c
```diff
@@ -105,7 +105,23 @@ void * avifArrayPush(void * arrayStruct)
...
+        
+        size_t newByteCount = oldByteCount * 2;
+        
+        // Additional safety check: verify the multiplication didn't overflow
+        if (newByteCount < oldByteCount) {
+            // Overflow occurred despite the check (shouldn't happen, but defense in depth)
+            return NULL;
+        }
+        
+        arr->ptr = (uint8_t *)avifAlloc(newByteCount);
```
wantehchang: I also found that `oldPtr` is leaked if this `avifAlloc()` call fails. I wrote pull request https://github.com/AOMediaCodec/libavif/pull/3006 to fix that leak.

wantehchang (commented):
Naveed8951: Thanks for the pull request. I reviewed your pull request and the related code carefully.

rootvector2:
I removed the redundant post-multiplication overflow check and updated the allocation logic to avoid leaking oldPtr when avifAlloc() fails.

src/utils.c
```diff
@@ -105,10 +105,22 @@ void * avifArrayPush(void * arrayStruct)
     if (arr->count == arr->capacity) {
         uint8_t * oldPtr = arr->ptr;
         size_t oldByteCount = (size_t)arr->elementSize * arr->capacity;
-        arr->ptr = (uint8_t *)avifAlloc(oldByteCount * 2);
-        if (arr->ptr == NULL) {
+
+        // Check for overflow before doubling the allocation size
+        // If oldByteCount > SIZE_MAX/2, then oldByteCount * 2 would overflow
```
wantehchang: Nit: We can omit this comment, because this kind of overflow check is quite common in libavif.

src/utils.c
```diff
@@ -105,10 +105,22 @@ void * avifArrayPush(void * arrayStruct)
...
+        if (oldByteCount > SIZE_MAX / 2) {
+            return NULL;
+        }
+
+        size_t newByteCount = oldByteCount * 2;
+
+        uint8_t * newPtr = (uint8_t *)avifAlloc(newByteCount);
+        if (newPtr == NULL) {
+            avifFree(oldPtr);
```
wantehchang: Your fix of the leak of `oldPtr` is incorrect. `arr->ptr` is still equal to `oldPtr`, a dangling pointer. So `avifArrayDestroy()` will free `arr->ptr`, resulting in double free.

I fixed the leak of `oldPtr` using a different approach in commit https://github.com/AOMediaCodec/libavif/commit/2b1478469b61ade38ed21edf8ecc1a59608cb864. Please rebase this pull request to the tip of the main branch.

The following patch is the net change (with the comment and blank lines omitted):

```
diff --git a/src/utils.c b/src/utils.c
index 9a121934..7de2d0af 100644
--- a/src/utils.c
+++ b/src/utils.c
@@ -111,11 +111,15 @@ void * avifArrayPush(void * arrayStruct)
     if (arr->count == arr->capacity) {
         uint8_t * oldPtr = arr->ptr;
         size_t oldByteCount = (size_t)arr->elementSize * arr->capacity;
-        arr->ptr = (uint8_t *)avifAlloc(oldByteCount * 2);
-        if (arr->ptr == NULL) {
-            arr->ptr = oldPtr;
+        if (oldByteCount > SIZE_MAX / 2) {
             return NULL;
         }
+        size_t newByteCount = oldByteCount * 2;
+        uint8_t * newPtr = (uint8_t *)avifAlloc(newByteCount);
+        if (newPtr == NULL) {
+            return NULL;
+        }
+        arr->ptr = newPtr;
         memset(arr->ptr + oldByteCount, 0, oldByteCount);
         memcpy(arr->ptr, oldPtr, oldByteCount);
         arr->capacity *= 2;
```

rootvector2:
I’ve removed the overflow check comment and pushed the update.

wantehchang:
When you rebased this pull request, the merge conflicts were not resolved correctly. For example, my recent change to the `avifArrayCreate()` function was reverted.

The net change should be exactly the same as the following patch:

```
diff --git a/src/utils.c b/src/utils.c
index 9a121934..7de2d0af 100644
--- a/src/utils.c
+++ b/src/utils.c
@@ -111,11 +111,15 @@ void * avifArrayPush(void * arrayStruct)
     if (arr->count == arr->capacity) {
         uint8_t * oldPtr = arr->ptr;
         size_t oldByteCount = (size_t)arr->elementSize * arr->capacity;
-        arr->ptr = (uint8_t *)avifAlloc(oldByteCount * 2);
-        if (arr->ptr == NULL) {
-            arr->ptr = oldPtr;
+        if (oldByteCount > SIZE_MAX / 2) {
             return NULL;
         }
+        size_t newByteCount = oldByteCount * 2;
+        uint8_t * newPtr = (uint8_t *)avifAlloc(newByteCount);
+        if (newPtr == NULL) {
+            return NULL;
+        }
+        arr->ptr = newPtr;
         memset(arr->ptr + oldByteCount, 0, oldByteCount);
         memcpy(arr->ptr, oldPtr, oldByteCount);
         arr->capacity *= 2;
```

PR #3003, opened 2026-02-05
read: prevent integer overflow in image grid validation

rootvector2:
Add explicit overflow checks before multiplying tile dimensions by grid rows/columns to avoid wraparound that could bypass validation and lead to malformed image handling.

Reject unsafe grids early with AVIF_RESULT_INVALID_IMAGE_GRID.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1738,7 +1738,6 @@ static avifResult avifDecoderDataAllocateImagePlanes(const avifDecoderData * dat
         // HEIF (ISO/IEC 23008-12:2017), Section 6.6.2.3.1:
         //   The tiled input images shall completely "cover" the reconstructed image grid canvas, ...
         
-        // Check for integer overflow before performing multiplications
         if (((uint64_t)tile->image->width * grid->columns < grid->outputWidth) ||
             ((uint64_t)tile->image->height * grid->rows < grid->outputHeight)) {
 


src/read.c
```diff
@@ -1737,12 +1737,29 @@ static avifResult avifDecoderDataAllocateImagePlanes(const avifDecoderData * dat
         //
         // HEIF (ISO/IEC 23008-12:2017), Section 6.6.2.3.1:
         //   The tiled input images shall completely "cover" the reconstructed image grid canvas, ...
+        
+        // Check for integer overflow before performing multiplications
```
wantehchang: Since we don't need to store these products in `uint32_t` variables, it would be simpler and better to just cast `tile->image->width` and `tile->image->height` to `uint64_` in the four multiplications.

rootvector2:
Updated the checks to cast the four multiplications to uint64_t instead of using explicit UINT32_MAX overflow guards. Thanks for the suggestion.

src/read.c
wantehchang: You need to first delete the integer overflow checks you added, and then add `uint64_t` casts to the four multiplications in the original code.

The following patch is the net change:

```
index 8f22250d..11f8af07 100644
--- a/src/read.c
+++ b/src/read.c
@@ -1737,14 +1737,14 @@ static avifResult avifDecoderDataAllocateImagePlanes(const avifDecoderData * dat
         //
         // HEIF (ISO/IEC 23008-12:2017), Section 6.6.2.3.1:
         //   The tiled input images shall completely "cover" the reconstructed image grid canvas, ...
-        if (((tile->image->width * grid->columns) < grid->outputWidth) || ((tile->image->height * grid->rows) < grid->outputHeight)) {
+        if ((((uint64_t)tile->image->width * grid->columns) < grid->outputWidth) || (((uint64_t)tile->image->height * grid->rows) < grid->outputHeight)) {
             avifDiagnosticsPrintf(data->diag,
                                   "Grid image tiles do not completely cover the image (HEIF (ISO/IEC 23008-12:2017), Section 6.6.2.3.1)");
         }
         // Tiles in the rightmost column and bottommost row must overlap the reconstructed image grid canvas. See MIAF (ISO/IEC 23000-22:2019), Section 7.3.11.4.2, Figure 2.
-        if (((tile->image->width * (grid->columns - 1)) >= grid->outputWidth) ||
-            ((tile->image->height * (grid->rows - 1)) >= grid->outputHeight)) {
+        if ((((uint64_t)tile->image->width * (grid->columns - 1)) >= grid->outputWidth) ||
+            (((uint64_t)tile->image->height * (grid->rows - 1)) >= grid->outputHeight)) {
             avifDiagnosticsPrintf(data->diag,
                                   "Grid image tiles in the rightmost column and bottommost row do not overlap the reconstructed image grid canvas. See MIAF (ISO/IEC 23000-22:2019), Section 7.3.11.4.2, Figure 2");
```

rootvector2:
Removed the added comment and left only the uint64_t casts as suggested. Thanks.

src/read.c
wantehchang: [Naveed8951](https://github.com/Naveed8951): The new code is still incorrect. You did not delete the integer overflow checks you added, and you added `uint64_t` casts to only two of the four multiplications in the original code.

To save time, I will create a new pull request based on my patch (which doesn't contain your code) and acknowledge you for reporting this issue in the commit message.

wantehchang: Done in pull request https://github.com/AOMediaCodec/libavif/pull/3010.

PR #3004, opened 2026-02-05
read: add bounds checks during extent merge to prevent buffer overflow

rootvector2:
Validate write offsets and per-extent copy sizes before memcpy when merging item extents to avoid out-of-bounds writes triggered by malformed or malicious AVIF metadata.

Reject invalid extents with AVIF_RESULT_BMFF_PARSE_FAILED.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1534,24 +1534,15 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item,
             item->mergedExtents.size = bytesToRead;
         } else {
             AVIF_ASSERT_OR_RETURN(item->ownsMergedExtents);
-            AVIF_ASSERT_OR_RETURN(front);
+
             size_t writeOffset = (size_t)(front - item->mergedExtents.data);
-            // Validate that the write will not exceed the allocated buffer
-            if (writeOffset > item->mergedExtents.size ||
-                bytesToRead > item->mergedExtents.size - writeOffset) {
 
-                avifDiagnosticsPrintf(
-                    diag,
-                    "Item ID %u extent would overflow merge buffer (buffer size: %zu, current offset: %zu, bytes to write: %zu)",
-                    item->id,
-                    item->mergedExtents.size,
-                    writeOffset,
-                    bytesToRead);
+            AVIF_ASSERT_OR_RETURN(writeOffset < item->mergedExtents.size);
+            AVIF_ASSERT_OR_RETURN(bytesToRead <= item->mergedExtents.size - writeOffset);
 
-                return AVIF_RESULT_INTERNAL_ERROR;
-            }
-            memcpy(front, offsetBuffer.data, bytesToRead);
-            front += bytesToRead;
+            uint8_t *dst = item->mergedExtents.data + writeOffset;
+            memcpy(dst, offsetBuffer.data, bytesToRead);
+            writeOffset += bytesToRead;
         }
 
         remainingBytes -= bytesToRead;


src/read.c
```diff
@@ -1535,6 +1535,17 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item,
         } else {
             AVIF_ASSERT_OR_RETURN(item->ownsMergedExtents);
             AVIF_ASSERT_OR_RETURN(front);
+            // Validate that the write will not exceed the allocated buffer
+            if ((size_t)(front - item->mergedExtents.data) > item->mergedExtents.size ||
+                bytesToRead > item->mergedExtents.size - (size_t)(front - item->mergedExtents.data)) {
```
wantehchang: At this point we have already validated all the sizes and offsets. These two conditions cannot be true if our code is correct. So this check is an assertion. We can consider adding this assertion because the code is quite complicated, but the error code should be `AVIF_RESULT_INTERNAL_ERROR`, not `AVIF_RESULT_BMFF_PARSE_FAILED`.

The expression `(size_t)(front - item->mergedExtents.data)` occurs three times. I suggest saving the result of this expression in a local variable to simplify the code.

rootvector2:
Updated to return AVIF_RESULT_INTERNAL_ERROR and stored the write offset in a local variable as suggested.

src/read.c
```diff
@@ -1535,6 +1535,21 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item,
         } else {
             AVIF_ASSERT_OR_RETURN(item->ownsMergedExtents);
             AVIF_ASSERT_OR_RETURN(front);
+            size_t writeOffset = (size_t)(front - item->mergedExtents.data);
+            // Validate that the write will not exceed the allocated buffer
+            if (writeOffset > item->mergedExtents.size ||
```
wantehchang: Nit: I suggest also treating equality as an error:

```
            if (writeOffset >= item->mergedExtents.size ||
```

src/read.c
```diff
@@ -1535,6 +1535,21 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item,
         } else {
             AVIF_ASSERT_OR_RETURN(item->ownsMergedExtents);
             AVIF_ASSERT_OR_RETURN(front);
+            size_t writeOffset = (size_t)(front - item->mergedExtents.data);
```
wantehchang, line 1538: With the checks / assertions you are adding, the `writeOffset` variable is more convenient than the `front` variable.

I suggest replacing the `front` variable with the `writeOffset` variable. Then in the `memcpy()` call below, recover `front` by using `item->mergedExtents.data + writeOffset`.

src/read.c
```diff
@@ -1535,6 +1535,21 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item,
...
+            // Validate that the write will not exceed the allocated buffer
+            if (writeOffset > item->mergedExtents.size ||
+                bytesToRead > item->mergedExtents.size - writeOffset) {
```
wantehchang: I suggest using the `AVIF_ASSERT_OR_RETURN` macro to check these two conditions are false.

The following patch is the net change:

```
index 8f22250d..a33bd1bb 100644
--- a/src/read.c
+++ b/src/read.c
@@ -1483,7 +1483,7 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item,
     // Set this until we manage to fill the entire mergedExtents buffer
     item->partialMergedExtents = AVIF_TRUE;
 
-    uint8_t * front = item->mergedExtents.data;
+    size_t writeOffset = 0; // Write offset for item->mergedExtents.data
     size_t remainingBytes = totalBytesToRead;
@@ -1534,9 +1534,10 @@ static avifResult avifDecoderItemRead(avifDecoderItem * item,
             item->mergedExtents.size = bytesToRead;
         } else {
             AVIF_ASSERT_OR_RETURN(item->ownsMergedExtents);
-            AVIF_ASSERT_OR_RETURN(front);
-            memcpy(front, offsetBuffer.data, bytesToRead);
-            front += bytesToRead;
+            AVIF_ASSERT_OR_RETURN(writeOffset < item->mergedExtents.size);
+            AVIF_ASSERT_OR_RETURN(bytesToRead <= item->mergedExtents.size - writeOffset);
+            memcpy(item->mergedExtents.data + writeOffset, offsetBuffer.data, bytesToRead);
+            writeOffset += bytesToRead;
         }
 
         remainingBytes -= bytesToRead;
```

wantehchang (commented):
I took the opportunity to audit the code in this function. Although the code is very complicated, I did not spot any errors. (I audited this function carefully for buffer overflow before.) So the error conditions you are checking should not happen and it is appropriate to check them with assertions. (`AVIF_ASSERT_OR_RETURN` is also enabled in release builds.)

rootvector2:
Switched the overflow guards to AVIF_ASSERT_OR_RETURN and updated the memcpy logic to consistently use writeOffset as suggested.

src/read.c
wantehchang: [Naveed8951](https://github.com/Naveed8951): The new code is incorrect, as the failing CI checks showed. The net change needs to be exactly the same as the patch I provided in my previous comment.

To save time, I can try creating a new pull request and setting the author to you on Monday.

wantehchang: Done in pull request https://github.com/AOMediaCodec/libavif/pull/3011.

PR #3005, opened 2026-02-05
Detect multiplication overflow in avifArrayCreate

wantehchang:
Detect whether (size_t)arr->elementSize * arr->capacity would overflow.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -5,6 +5,7 @@
 
 #include <assert.h>
 #include <math.h>
+#include <stdint.h>
 #include <string.h>
 
 float avifRoundf(float v)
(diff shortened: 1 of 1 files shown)


src/utils.c
```diff
@@ -89,6 +90,11 @@ avifBool avifArrayCreate(void * arrayStruct, uint32_t elementSize, uint32_t init
     arr->elementSize = elementSize ? elementSize : 1;
     arr->count = 0;
     arr->capacity = initialCapacity;
+    if (arr->capacity > SIZE_MAX / arr->elementSize) {
```
wantehchang, line 93: This division is safe because line 90 ensures `arr->elementSize` is nonzero.

vigneshvg (approved):

PR #3009, opened 2026-02-08
Prevent integer overflow when doubling capacity

wantehchang:
A follow-up to https://github.com/AOMediaCodec/libavif/pull/3002. Prevent integer overflow when doubling arr->capacity in avifArrayPush().

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -111,17 +111,14 @@ void * avifArrayPush(void * arrayStruct)
     if (arr->count == arr->capacity) {
         uint8_t * oldPtr = arr->ptr;
         size_t oldByteCount = (size_t)arr->elementSize * arr->capacity;
-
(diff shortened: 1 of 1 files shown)


src/utils.c
```diff
@@ -111,17 +111,14 @@ void * avifArrayPush(void * arrayStruct)
     if (arr->count == arr->capacity) {
         uint8_t * oldPtr = arr->ptr;
         size_t oldByteCount = (size_t)arr->elementSize * arr->capacity;
-
-        if (oldByteCount > SIZE_MAX / 2) {
+        if (oldByteCount > SIZE_MAX / 2 || arr->capacity > UINT32_MAX / 2) {
```
wantehchang, line 114: `arr->capacity` (of the `uint32_t` type) will be doubled at line 125 below:

```
        arr->capacity *= 2;
```

PR #3018, opened 2026-02-09
Disable Sample Transform decoding by default

y-guyon:
Add AVIF_IMAGE_CONTENT_SAMPLE_TRANSFORMS.
16-bit avifImage may not be gracefully handled by all current avifDecoder users, so it is safer to disable the feature by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The changes are relative to the previous release, unless the baseline is specifi
   avifDecoder::imageContentToDecode & AVIF_IMAGE_CONTENT_SAMPLE_TRANSFORMS is
   not zero.
 * Support Sample Transform derived image items with grid input image items.
+* Add --sato flag to avifdec to enable Sample Transforms support at decoding.
 * Add --grid option to avifgainmaputil.
 
 ### Changed since 1.3.0
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -34,7 +34,7 @@ static void syntax(void)
     printf("    -j,--jobs J       : Number of jobs (worker threads), or 'all' to potentially use as many cores as possible. (Default: all)\n");
     printf("    -c,--codec C      : Codec to use (choose from versions list below)\n");
     printf("    -d,--depth D      : Output depth, either 8 or 16. (PNG only; For y4m, depth is retained, and JPEG is always 8bpc)\n");
-    printf("                        Output depth set to 16 enables Sample Transform decoding no matter the output format.\n");
+    printf("    --sato            : Enable Sample Transforms decoding (e.g. 16-bit AVIF)\n");
     printf("    -q,--quality Q    : Output quality in 0..100. (JPEG only, default: %d)\n", DEFAULT_JPEG_QUALITY);
     printf("    --png-compress L  : PNG compression level in 0..9 (PNG only; 0=none, 9=max). Defaults to libpng's builtin default\n");
     printf("    -u,--upsampling U : Chroma upsampling (for 420/422). One of 'automatic' (default), 'fastest', 'best', 'nearest', or 'bilinear'\n");
@@ -89,6 +89,7 @@ int main(int argc, char * argv[])
     const char * inputFilename = NULL;
     const char * outputFilename = NULL;
     int requestedDepth = 0;
+    avifBool enableSampleTransforms = AVIF_FALSE;
     int jobs = -1;
     int jpegQuality = DEFAULT_JPEG_QUALITY;
     int pngCompressionLevel = -1; // -1 is a sentinel to avifPNGWrite() to skip calling png_set_compression_level()
@@ -169,6 +170,8 @@ int main(int argc, char * argv[])
                 fprintf(stderr, "ERROR: invalid depth: %s\n", arg);
                 return 1;
             }
+        } else if (!strcmp(arg, "--sato")) {
+            enableSampleTransforms = AVIF_TRUE;
         } else if (!strcmp(arg, "-q") || !strcmp(arg, "--quality")) {
             NEXTARG();
             jpegQuality = atoi(arg);
@@ -314,7 +317,7 @@ int main(int argc, char * argv[])
     decoder->allowProgressive = allowProgressive;
     if (infoOnly) {
         decoder->imageContentToDecode |= AVIF_IMAGE_CONTENT_GAIN_MAP | AVIF_IMAGE_CONTENT_SAMPLE_TRANSFORMS;
-    } else if (requestedDepth == 16) {
+    } else if (enableSampleTransforms) {
         decoder->imageContentToDecode |= AVIF_IMAGE_CONTENT_SAMPLE_TRANSFORMS;
     }
 


include/avif/avif.h
```diff
@@ -1234,6 +1234,11 @@ typedef enum avifImageContentTypeFlag
     AVIF_IMAGE_CONTENT_GAIN_MAP = (1 << 2),
     AVIF_IMAGE_CONTENT_ALL = AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA | AVIF_IMAGE_CONTENT_GAIN_MAP,
 
+    // Mostly used for bit depth extensions to go beyond the underlying codec capability
+    // (e.g. 16-bit AVIF). Not part of AVIF_IMAGE_CONTENT_ALL as this is a rare use case.
+    // Has no effect without AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA.
+    AVIF_IMAGE_CONTENT_SAMPLE_TRANSFORMS = (1 << 3),
```
wantehchang, line 1240: I have two comments on this.

1. It may be better to control this with a new boolean field in the `avifDecoder` struct. Since sample transforms also produce color and alpha, sample transforms are not a new content type.
2. It may be better to restrict this control to bit depth extensions only rather than all sample transforms, if we can determine the output bit depth without applying the sample transforms. (Maryla also said this in [her review comment](https://github.com/AOMediaCodec/libavif/pull/3018#pullrequestreview-3784732820).) I think we can get the bit depth from the 'pixi' box of the 'sato' derived image item, right?

In summary, please consider exposing this control as a new `avifBool enableExtendedBitDepths` field in the `avifDecoder` struct

wantehchang (commented):
I agree that returning a 16-bit output image is an incompatible change, so users need to opt in to this new feature. Thank you for thinking of this issue!

include/avif/avif.h
y-guyon: Thank you for the review.

1. I disagree. Gain maps complement the primary color image but adding color information. Bit depth extensions, and thus Sample Transforms, also extend the color information in a similar way. The dimension is just bit depth instead of brightness, if I may say so.  
Also `avifImageContentTypeFlags` is a bit mask rather than an enum, meaning the different `avifImageContentTypeFlag` values are more orthogonal concepts that can each be toggled on or off than distinct types.  
Each `avifImageContentTypeFlag`  maps to a separate image item in AVIFv1 (primary color item, aux alpha, derived gain map, derived sample transform).
2. As of today, libavif applies any Sample Transform formula with at most 2 input items. Also, there could be plenty of different formulas for extending the bit depth, not only the few that are available at encoding right now. I do not think it is easy to precisely detect whether the decoded formula is a bit depth extension, and only that.  
Maryla's comment "Tying sample transforms decoding to bit depth" was specifically about avifdec's `--depth` flag. I am not sure this is what you meant.

wantehchang: I am commenting on the change to the public header include/avif/avif.h.

If a 'sato' derived image item has one 8-bit AV1 image item and adds the constant 0 to the input image item, the  'sato' derived image item should have a 'pixi' box that indicates bit depth 8. Current clients of libavif can handle this output image. This is why we only need to disable by default the 'sato' derived image items whose 'pixi' box indicates a bit depth other than 8, 10, 12.

Since the outut image is still represented with the `avifImage` struct, it seems that the only field in `avifImage` that may catch existing clients by surprise is the `depth` field. Are there other kinds of 'sato' derived image items that produce an output `avifImage` that current clients of libavif cannot handle?

y-guyon: > Current clients of libavif can handle this output image.

Yes. The question is: should it still be applied?  
With this specific example (noop 0 addition), it is obviously better NOT to enable Sample Transforms, as it would just make decoding slower for the same result.

> Are there other kinds of 'sato' derived image items that produce an output `avifImage` that current clients of libavif cannot handle?

I do not think so. But it does not mean it is beneficial to do so.  
libavif only supports Sample Transform derived image items that are NOT the primary item, and in the same `altr` group as the primary item. So there is always an alternative, and the primary item should be the default output of libavif in my opinion.  
Advanced users can enable Sample Transform decoding to extract more information when needed.

wantehchang: I thought the only concern was breaking the clients unprepared to handle new bit depths. I did not realize the cost incurred by applying sample transforms was also a concern.

We need to provide guidance on whether libavif clients should enable sample transforms. Besides the two concerns mentioned above, are there other concerns?

I think we should constrain the new bit depths that libavif supports and document what they are. My code analysis shows that libavif malfunctions if `decoder->data->meta->sampleTransformDepth > 16`, so we need to at least disallow bit depths > 16.

---

Note: You can ignore the rest of this comment.

Re: `AVIF_IMAGE_CONTENT_SAMPLE_TRANSFORMS` vs. `enableSampleTransforms`: In your interpretation of `avifImageContentTypeFlags` based on the image item type, `AVIF_IMAGE_CONTENT_SAMPLE_TRANSFORMS` makes sense. I have a different interpretation of `avifImageContentTypeFlags` based on the output destination:

- `AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA`: the output goes to `decoder->image`
- `AVIF_IMAGE_CONTENT_GAIN_MAP`: the output goes to `decoder->image->gainMap->image`

When sample transforms are enabled, the output still goes to `decoder->image`. Therefore sample transforms, unlike gain maps, do not behave like a new image content type.

y-guyon: > I thought the only concern was breaking the clients unprepared to handle new bit depths. I did not realize the cost incurred by applying sample transforms was also a concern.

That was not explicit, sorry about that.

> We need to provide guidance on whether libavif clients should enable sample transforms. Besides the two concerns mentioned above, are there other concerns?

Security? More features expose more attack surfaces in general.

> we need to at least disallow bit depths > 16.

I think 16-bit is already the maximum libavif can output:

https://github.com/AOMediaCodec/libavif/blob/ec2ee83ddc27ec100cab07e7b81ef18a6cd045a1/src/avif.c#L143-L146

This is verified in https://github.com/AOMediaCodec/libavif/pull/3038 (although the error code was wrong).

> I think we should constrain the new bit depths that libavif supports and document what they are.

I agree they should be documented.

> My code analysis shows that libavif malfunctions if `decoder->data->meta->sampleTransformDepth > 16`, so we need to at least disallow bit depths > 16.

libavif "malfunctions" in the sense that 1. if >16-bit were to be allowed, some parts of libavif would need further changes for proper support, or 2. you can craft AVIF files that make libavif crash or incorrectly decode as of today?

> When sample transforms are enabled, the output still goes to `decoder->image`. Therefore sample transforms, unlike gain maps, do not behave like a new image content type.

For me the name `imageContentToDecode` refers more to what is inside the file (to decode) than where to output it into.

wantehchang: Re: libavif malfunctions if `decoder->data->meta->sampleTransformDepth > 16`

Yannis: I didn't do a complete code analysis, but what I found concerns me. (You can do a similar analysis by inspecting the code related to `sampleTransformDepth`.)

In `avifDecoderReset()`, we assign a value to `sampleTransformDepth`:

```
            data->meta->sampleTransformDepth = pixiProp->u.pixi.planeDepths[0];
```

Code Search shows this is the only place where we assign a value to `sampleTransformDepth`. We do not restrict `pixiProp->u.pixi.planeDepths[0]` in `avifDecoderSampleTransformItemValidateProperties()`.

**Possible solution:** We can change `avifDecoderSampleTransformItemValidateProperties()` to disallow `pixiProp->u.pixi.planeDepths[0] > 16`.

Later in `avifDecoderReset()`, we copy `data->meta->sampleTransformDepth` to `decoder->image->depth`:

```
    if (decoder->data->meta->sampleTransformExpression.count > 0) {
        AVIF_ASSERT_OR_RETURN(decoder->data->meta->sampleTransformDepth != 0);
        decoder->image->depth = decoder->data->meta->sampleTransformDepth;
    }
```

This bypasses the bit depth check in `avifImageCreate()` that you mentioned.

`decoder->image` is passed as the `dstImage` argument to `avifDecoderApplySampleTransform()`.

```
    if (decoder->data->tileInfos[AVIF_ITEM_COLOR].tileCount != 0 && decoder->data->meta->sampleTransformExpression.count > 0) {
        AVIF_CHECKRES(avifDecoderApplySampleTransform(decoder, decoder->image));
    }
```

Eventually we reach `avifImageApplyExpression32b()`, which uses `dstImage` as follows:

```
                if (avifImageUsesU16(dstImage)) {
                    ((uint16_t *)row)[x] = (uint16_t)stack[0];
                } else {
                    row[x] = (uint8_t)stack[0];
                }
```

The cast to `uint16_t *` for any bit depth > 8 is incorrect for bit depths > 16. That's what I meant by "libavif malfunctions".

wantehchang: One more thing: `avifDecoderSampleTransformItemValidateProperties()` performs the following check:

```
    for (uint8_t i = 0; i < pixiProp->u.pixi.planeCount; ++i) {
        if (pixiProp->u.pixi.planeDepths[i] != pixiProp->u.pixi.planeDepths[0]) {
            avifDiagnosticsPrintf(diag,
                                  "Item ID %u of type '%.4s' has different depths specified by pixi property [%u, %u], this is not supported",
                                  item->id,
                                  (const char *)item->type,
                                  pixiProp->u.pixi.planeDepths[0],
                                  pixiProp->u.pixi.planeDepths[i]);
        }
    }
```

This check is redundant because ` avifParsePixelInformationProperty()` already performs the same check.

wantehchang: I wrote:

> **Possible solution:** We can change `avifDecoderSampleTransformItemValidateProperties()` to disallow `pixiProp->u.pixi.planeDepths[0] > 16`.

Your PR https://github.com/AOMediaCodec/libavif/pull/3038 disallows bit depths > 16 in `avifParsePixelInformationProperty()`. That is stricter and also works.

y-guyon: > This check is redundant because ` avifParsePixelInformationProperty()` already performs the same check.

Done in https://github.com/AOMediaCodec/libavif/pull/3040

y-guyon: > This bypasses the bit depth check in `avifImageCreate()` that you mentioned.

Not really, because it goes through:

https://github.com/AOMediaCodec/libavif/blob/0812f444de0f5b4cd43b41facfb1f0fce2bca5ec/src/read.c#L6846-L6855

`dstImage->depth` is likely 8 or 12 because it contains the primary coded image item. `decoder->data->meta->sampleTransformDepth` is 16 (or higher). So the condition is true, and `avifImageCreate()` is called with `sampleTransformDepth` bits. For `sampleTransformDepth`>16, `avifImageCreate()` returns null and `avifDecoderApplySampleTransform()` returns the wrong `AVIF_RESULT_OUT_OF_MEMORY` error code.

If I understand correctly, the only way to bypass this would be for the user to explicitly set `avifDecoder::image::depth` to exactly `sampleTransformDepth` after parsing/decoding the primary item but before `avifDecoderApplySampleTransform()` is called.

y-guyon: > Your PR #3038 disallows bit depths > 16 in `avifParsePixelInformationProperty()`. That is stricter and also works.

I consider this thread resolved.

PR #3022, opened 2026-02-10
Adapt avm seq_profile to PROFILE_BITS

y-guyon:


wantehchang (commented):
LGTM. Thanks for the fix!

PR #3024, opened 2026-02-10
Remove out-of-date or unlikely to be fixed TODOs

y-guyon:


apps/avifenc.c
```diff
@@ -2590,7 +2590,7 @@ int main(int argc, char * argv[])
             fprintf(stderr, "ERROR: Not enough input files for grid image! (expecting %u, or a single image to be split)\n", gridCellCount);
             goto cleanup;
         }
-        // TODO(yguyon): Check if it is possible to use frames from a single input file as grid cells. Maybe forbid it.
+        // Frames from a single input file could be used as grid cells. This is allowed for simplicity.
```
wantehchang: "This is allowed for simplicity" doesn't seem correct. Did you mean "This is unsupported for simplicity"? Please double check.

y-guyon: "This is allowed for simplicity" is correct.

It sounds weird to use the frames from an animated input files as the grid cells of the still output image. But it is not worth taking the time to check whether this is possible and if/how to forbid it.

wantehchang: Thanks for the confirmation. Let me explain why this comment confuses me.

The word "could" in "Frames from a single input file could be used as grid cells." implies that we do not use frames from a single input file as grid cells, but we could potentially do that. Then the next sentence says "This is allowed", which implies we are doing that. So the two sentences seem to contradict each other.

wantehchang: I agree that it is a weird idea to use the frames from an animated input file as the grid cells of the still output image. So we can consider just deleting the TODO comment, without mentioning this idea.

PR #3026, opened 2026-02-10
Rename avifSequenceHeader::av1C to codecConfig

y-guyon:
So that it can also be used to store av2C.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -783,7 +783,7 @@ typedef struct avifSequenceHeader
     avifTransferCharacteristics transferCharacteristics;
     avifMatrixCoefficients matrixCoefficients;
     avifRange range;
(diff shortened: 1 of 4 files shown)


include/avif/internal.h
```diff
@@ -783,7 +783,7 @@ typedef struct avifSequenceHeader
     avifTransferCharacteristics transferCharacteristics;
     avifMatrixCoefficients matrixCoefficients;
     avifRange range;
-    avifCodecConfigurationBox av1C; // TODO(yguyon): Rename or add av2C
+    avifCodecConfigurationBox codecConfig; // "av1C" for AV1 ("av2C" for AV2 if AVIF_CODEC_AVM)
```
wantehchang, line 786: It is better to wait until `AV2CodecConfigurationBox` ('av2C') is defined, because we may want to have both an `av1C` field and an `av2C` field, or even different struct types for AV1 and AV2 sequence headers.

PR #3029, opened 2026-02-11
Remember tune=iq option across frames

juliobbv-p:
Also add a ProgressiveTest to test two layer encoding with tune=iq.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -69,6 +69,7 @@ struct avifCodecInternal
     aom_img_fmt_t aomFormat;
     uint32_t currentLayer;
     int qualityFirstLayer;
+    avifBool previousFrameUsedTuneIq;
 #endif
 };
 
@@ -403,15 +404,20 @@ static avifBool avifProcessAOMOptionsPreInit(avifCodec * codec, avifBool alpha, 
     return AVIF_TRUE;
 }
 
-static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
+static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha, avifBool useLibavifDefaultTuneMetric, aom_tune_metric libavifDefaultTuneMetric)
 {
-    avifBool ret = AVIF_FALSE;
-
 #if !defined(AOM_HAVE_TUNE_IQ)
 // Define the tune IQ value here if libaom doesn't define it. The enum value is guaranteed to never change
 // in libaom, so this definition won't ever get out of sync.
 #define AOM_TUNE_IQ 10
 #endif
+
+    if (useLibavifDefaultTuneMetric) {
+        return libavifDefaultTuneMetric == AOM_TUNE_IQ;
+    }
+
+    avifBool useTuneIq = AVIF_FALSE;
+    avifBool isAnyTuneDefined = AVIF_FALSE;
 
     // Tune IQ string -> enum mapping
     static const struct aomOptionEnumList tuneIqEnum[] = { { "iq", AOM_TUNE_IQ }, // Image Quality (IQ) mode
@@ -423,16 +429,24 @@ static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
         // If there are multiple "tune" options specified, honor the last one.
         // For consistent behavior, handle both cases where tune IQ was either specified as a string (tune=iq),
         // or as an enum value (tune=10).
-        if (avifKeyEqualsName(entry->key, "tune", alpha) && aomOptionParseEnum(entry->value, tuneIqEnum, &val)) {
-            ret = (val == AOM_TUNE_IQ);
+        if (avifKeyEqualsName(entry->key, "tune", alpha)) {
+            isAnyTuneDefined = AVIF_TRUE;
+
+            if (aomOptionParseEnum(entry->value, tuneIqEnum, &val)) {
+                useTuneIq = (val == AOM_TUNE_IQ);
+            }
         }
     }
 
-    // In practice this function should also return true if avifEncoderSetCodecSpecificOption("tune", "iq")
-    // was called for a previous frame and not called (or called with NULL) for this frame, because the tune
-    // option persists across frames in libaom. However AOM_TUNE_IQ is only supported with still images in
-    // libavif and libaom as of today, so there is no need to remember this option across frames.
-    return ret;
+    if (!isAnyTuneDefined && codec->internal->previousFrameUsedTuneIq) {
+        // Handle the case where the encoder was called with avifEncoderSetCodecSpecificOption("tune", "iq")
+        // for a previous frame and not called (or called with NULL) for this frame, because the tune
+        // option persists across frames in libaom.
+        // In this case, we know libaom will also use tune=iq for this frame.
+        return AVIF_TRUE;
+    }
+
+    return useTuneIq;
 }
 
 #if !defined(HAVE_AOM_CODEC_SET_OPTION)
@@ -725,22 +739,32 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
 
     avifBool useLibavifDefaultTuneMetric = AVIF_FALSE;        // If true, override libaom's default tune option.
     aom_tune_metric libavifDefaultTuneMetric = AOM_TUNE_PSNR; // Meaningless unless useLibavifDefaultTuneMetric.
-    if (quality != AVIF_QUALITY_LOSSLESS && !avifAOMOptionsContainExplicitTuning(codec, alpha)) {
-        useLibavifDefaultTuneMetric = AVIF_TRUE;
-        if (alpha) {
-            // Minimize ringing for alpha.
-            libavifDefaultTuneMetric = AOM_TUNE_PSNR;
-        } else {
-            libavifDefaultTuneMetric = AOM_TUNE_SSIM;
+
+    // libavif only needs to set the default tune metric for the first frame,
+    // because libaom will persist that setting until explicitly changed.
+    if (!codec->internal->encoderInitialized) {
+        if (quality != AVIF_QUALITY_LOSSLESS && !avifAOMOptionsContainExplicitTuning(codec, alpha)) {
+            useLibavifDefaultTuneMetric = AVIF_TRUE;
+            if (alpha) {
+                // Minimize ringing for alpha.
+                libavifDefaultTuneMetric = AOM_TUNE_PSNR;
+            } else {
+                libavifDefaultTuneMetric = AOM_TUNE_SSIM;
+            }
         }
+        AVIF_ASSERT_OR_RETURN(!codec->internal->previousFrameUsedTuneIq);
     }
 
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
     // True if libavif knows that tune=iq is used, either by default by libavif, or explicitly set by the user.
     // False otherwise (including if libaom uses tune=iq by default, which is not the case as of v3.13.1 and earlier versions).
-    const avifBool useTuneIq = useLibavifDefaultTuneMetric ? libavifDefaultTuneMetric == AOM_TUNE_IQ : avifImageUsesTuneIq(codec, alpha);
+    const avifBool useTuneIq = avifImageUsesTuneIq(codec, alpha, useLibavifDefaultTuneMetric, libavifDefaultTuneMetric);
     const int quantizer = aomQualityToQuantizer(quality, useTuneIq);
+
+    // libavif needs to know whether the current frame uses tune=iq for the next frame, as libaom persists
+    // tuning modes across frames
+    codec->internal->previousFrameUsedTuneIq = useTuneIq;
 
     // For encoder->scalingMode.horizontal and encoder->scalingMode.vertical to take effect in AOM
     // encoder, config should be applied for each frame, so we don't care about changes on these
--- a/tests/gtest/avifprogressivetest.cc
+++ b/tests/gtest/avifprogressivetest.cc
@@ -87,6 +87,37 @@ TEST_F(ProgressiveTest, QualityChange) {
   TestDecode(kImageSize, kImageSize);
 }
 
+// NOTE: This test requires libaom v3.12.0 or later, as this was the first
+// version where tune IQ was available
+TEST_F(ProgressiveTest, TuneIq) {
+  encoder_->extraLayerCount = 1;
+  // Tune IQ requires all-intra mode, which libavif determines when the first
+  // layer is encoded at a very low quality (e.g. quality 10)
+  encoder_->quality = 10;
+  encoder_->codecChoice = AVIF_CODEC_CHOICE_AOM;
+
+  ASSERT_EQ(avifEncoderSetCodecSpecificOption(encoder_.get(), "tune", "iq"),
+            AVIF_RESULT_OK);
+  avifResult result = avifEncoderAddImage(encoder_.get(), image_.get(), 1,
+                                          AVIF_ADD_IMAGE_FLAG_NONE);
+
+  if (result == AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION) {
+    // The aom version that libavif was built with likely does not support
+    // AOM_TUNE_IQ.
+    return;
+  }
+
+  ASSERT_EQ(result, AVIF_RESULT_OK);
+  encoder_->quality = 50;
+  ASSERT_EQ(avifEncoderAddImage(encoder_.get(), image_.get(), 1,
+                                AVIF_ADD_IMAGE_FLAG_NONE),
+            AVIF_RESULT_OK);
+
+  ASSERT_EQ(avifEncoderFinish(encoder_.get(), &encoded_avif_), AVIF_RESULT_OK);
+
+  TestDecode(kImageSize, kImageSize);
+}
+
 // NOTE: This test requires libaom v3.6.0 or later, otherwise the following
 // assertion in libaom fails:
 //   av1/encoder/mcomp.c:1717: av1_full_pixel_search: Assertion


src/codec_aom.c
```diff
@@ -739,9 +752,16 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
     avifBool quantizerUpdated = AVIF_FALSE;
     // True if libavif knows that tune=iq is used, either by default by libavif, or explicitly set by the user.
     // False otherwise (including if libaom uses tune=iq by default, which is not the case as of v3.13.1 and earlier versions).
-    const avifBool useTuneIq = useLibavifDefaultTuneMetric ? libavifDefaultTuneMetric == AOM_TUNE_IQ : avifImageUsesTuneIq(codec, alpha);
+    const avifBool useTuneIq = useLibavifDefaultTuneMetric ? libavifDefaultTuneMetric == AOM_TUNE_IQ
+                                                           : avifImageUsesTuneIq(codec, alpha, codec->internal->currentLayer == 0);
     const int quantizer = aomQualityToQuantizer(quality, useTuneIq);
 
+    if (!alpha && encoder->extraLayerCount > 0) {
```
y-guyon: I understand it matches the name of the variable `previousLayerUsedTuneIq`, but beyond that, why is the purpose of only setting it if `encoder->extraLayerCount > 0`? Why not the same behavior for consecutive frames of an animation as for layers?

src/codec_aom.c
```diff
@@ -739,9 +752,16 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+        // We need to know whether the current layer uses tune=iq for the next layer, as libaom persists
+        // tuning modes across frames
+        codec->internal->previousLayerUsedTuneIq = useTuneIq;
```
y-guyon: I tried to deduce what would happen in the following two-layer scenarios, assuming `libavifDefaultTuneMetric`=`AOM_TUNE_IQ`:

##### no `avifEncoderSetCodecSpecificOption()` call (equivalent to `avifEncoderSetCodecSpecificOption("tune", NULL)`)

1. `ContainExplicitTuning()`=0, `useLibavifDefaultTune`=1, `useTuneIq`=1, `previousLayerUsedIq`=1
1. `ContainExplicitTuning()`=0, `useLibavifDefaultTune`=1, `useTuneIq`=1

✅

##### `avifEncoderSetCodecSpecificOption("tune", "ssim")` before first frame, `avifEncoderSetCodecSpecificOption("tune", "iq")` after first frame

1. `ContainExplicitTuning()`=1, `useLibavifDefaultTune`=0, `useTuneIq`=0, `previousLayerUsedIq`=0
1. `ContainExplicitTuning()`=1, `useLibavifDefaultTune`=0, `useTuneIq`=1

✅

##### `avifEncoderSetCodecSpecificOption("tune", "iq")` before first frame, `avifEncoderSetCodecSpecificOption("tune", "ssim")` after first frame

1. `ContainExplicitTuning()`=1, `useLibavifDefaultTune`=0, `useTuneIq`=1, `previousLayerUsedIq`=1
1. `ContainExplicitTuning()`=1, `useLibavifDefaultTune`=0, `useTuneIq`=0

✅

##### `avifEncoderSetCodecSpecificOption("tune", "iq")` before first frame

1. `ContainExplicitTuning()`=1, `useLibavifDefaultTune`=0, `useTuneIq`=1, `previousLayerUsedIq`=1
1. `ContainExplicitTuning()`=0, `useLibavifDefaultTune`=1, `useTuneIq`=1

✅

##### `avifEncoderSetCodecSpecificOption("tune", "ssim")` before first frame

1. `ContainExplicitTuning()`=1, `useLibavifDefaultTune`=0, `useTuneIq`=0, `previousLayerUsedIq`=0
1. `ContainExplicitTuning()`=0, `useLibavifDefaultTune`=1, `useTuneIq`=1
**is this the expected behavior? ------------------------------------------^**

src/codec_aom.c
```diff
@@ -423,16 +425,27 @@ static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
         // If there are multiple "tune" options specified, honor the last one.
         // For consistent behavior, handle both cases where tune IQ was either specified as a string (tune=iq),
         // or as an enum value (tune=10).
-        if (avifKeyEqualsName(entry->key, "tune", alpha) && aomOptionParseEnum(entry->value, tuneIqEnum, &val)) {
-            ret = (val == AOM_TUNE_IQ);
+        if (avifKeyEqualsName(entry->key, "tune", alpha)) {
+            // Check whether any tuning mode has been defined, excluding the { "tune", NULL } case
+            isAnyTuneDefined = entry->value != NULL;
```
y-guyon: `entry->value` cannot be NULL, see `avifCodecSpecificOptionsSet()`.

src/codec_aom.c
```diff
@@ -423,16 +425,27 @@ static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
...
-    // was called for a previous frame and not called (or called with NULL) for this frame, because the tune
-    // option persists across frames in libaom. However AOM_TUNE_IQ is only supported with still images in
-    // libavif and libaom as of today, so there is no need to remember this option across frames.
-    return ret;
+    if (!isAnyTuneDefined && !isFirstLayer && codec->internal->previousLayerUsedTuneIq) {
+        // Handle the case where the encoder was called with avifEncoderSetCodecSpecificOption("tune", "iq")
+        // for a previous layer (frame) and not called (or called with NULL) for this frame, because the tune
+        // option persists across frames in libaom.
+        // In this case, we know libaom will also use tune=iq for this frame.
+        useTuneIq = AVIF_TRUE;
```
y-guyon: I think this is never reached. If `isAnyTuneDefined`=0, `useLibavifDefaultTuneMetric` is 1 and `avifImageUsesTuneIq()` is not called.

src/codec_aom.c
juliobbv-p: I see now. The NULL case is when the **key** is null, not the **value**. I'll fix this.

src/codec_aom.c
juliobbv-p: > avifEncoderSetCodecSpecificOption("tune", "ssim") before first frame
ContainExplicitTuning()=1, useLibavifDefaultTune=0, useTuneIq=0, previousLayerUsedIq=0
ContainExplicitTuning()=0, useLibavifDefaultTune=1, useTuneIq=1
is this the expected behavior? ------------------------------------------^

This is a good question. In this case, `useTuneIq` has to match the logic that determines `libavifDefaultTuneMetric`, and because for the second frame: `useLibavifDefaultTune=1` and `libavifDefaultTuneMetric=AOM_TUNE_IQ`, then `useTuneIq=1` is expected behavior here.

This does bring us another question: do we want to *persist* the tune IQ decision for future frames when the tuning mode is unset for a frame (and have that influence libavif's tune picking decision), not just *remember* whether the previous frame used tune IQ or not? That'd require a more complex implementation I think, because libavif would need to keep track of whether the caller has ever set the tuning mode for past frames (independently for color and alpha), and modify the tune picking decision to leverage that information. Doable, but I'm concerned the logic would be difficult to read and reason about.

@wantehchang and I also had the idea to mark the tuning mode as not-updatable -- libavif will enforce the same tuning mode from the first frame onto the rest. This makes tune-picking logic much more simple, but does come with a breaking API change through. IMO, I find it odd that the tuning mode can be updatable between frames in the first place 😄.

What do your thoughts on the situation?

src/codec_aom.c
juliobbv-p: No specific reason, my PR was originally targeted toward layered encoding. I can remove the `encoder->extraLayerCount > 0` part if that's better.

src/codec_aom.c
y-guyon: Remembering if any tune was set in the past should be simple enough.

There is no way to "reset" the tune setting in aom through libavif right now, so if any tune setting was forwarded till aom once, there is no need for libavif to set anything by default ever after.  
In other words, libavif should only set a tune by default on the first frame, if there is no explicit user tune. And libavif should remember the last default or explicit tune for any frame, so that it can adapt the quality in case of tune=iq.  
That should be enough, unless I missed something.

juliobbv-p: Got you, I believe you're right on resetting the tune -- I don't see a way to do it either.

Just to confirm: there's no way in libavif to encode multi-frame layers with alpha, right?

juliobbv-p: From an email convo with @wantehchang:

> Yes, libavif supports multi-frame encoding with alpha. Alpha is
commonly used in animated images (image sequences).

> libavif creates separate AV1 encoders to encode the color and alpha
bitstreams. The AV1 encoders receive the "avifBool alpha" parameter so
that they know whether they are encoding alpha or color.

> The AV1 encoders are created by the avifCodecCreate() call in
avifEncoderAddImageItems().

juliobbv-p: All right, I've reacted to PR feedback, implementing this behavior:

> There is no way to "reset" the tune setting in aom through libavif right now, so if any tune setting was forwarded till aom once, there is no need for libavif to set anything by default ever after.
In other words, libavif should only set a tune by default on the first frame, if there is no explicit user tune. And libavif should remember the last default or explicit tune for any frame, so that it can adapt the quality in case of tune=iq.

I kept the `previousFrameUsedTuneIq` boolean variable from my first revision, but I can easily convert it to `aom_tune_metric` if that's more preferable. Both work well here.

src/codec_aom.c
```diff
@@ -424,15 +430,20 @@ static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
         // For consistent behavior, handle both cases where tune IQ was either specified as a string (tune=iq),
         // or as an enum value (tune=10).
         if (avifKeyEqualsName(entry->key, "tune", alpha) && aomOptionParseEnum(entry->value, tuneIqEnum, &val)) {
-            ret = (val == AOM_TUNE_IQ);
+            isAnyTuneDefined = AVIF_TRUE;
```
y-guyon: With the current PR, `isAnyTuneDefined` is only true if `tuneIqEnum` is matched, right? If so, the name `isAnyTuneDefined` is inaccurate.

juliobbv-p: Yep, you're right. The enum parsing condition is only for determining `useTuneIq` anyway, so I moved that condition into its own clause.

src/codec_aom.c
```diff
@@ -725,23 +739,32 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+        if (quality != AVIF_QUALITY_LOSSLESS && !avifAOMOptionsContainExplicitTuning(codec, alpha)) {
+            useLibavifDefaultTuneMetric = AVIF_TRUE;
+            if (alpha) {
+                // Minimize ringing for alpha.
+                libavifDefaultTuneMetric = AOM_TUNE_PSNR;
+            } else {
+                libavifDefaultTuneMetric = AOM_TUNE_SSIM;
+            }
         }
+        codec->internal->previousFrameUsedTuneIq = AVIF_FALSE;
```
y-guyon: Note that this should be memset to 0 anyway. This line could probably be replaced with `AVIF_ASSERT_OR_RETURN(!codec->internal->previousFrameUsedTuneIq)`. This is fine as is too.

src/codec_aom.c
```diff
@@ -725,23 +739,32 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
     aom_tune_metric libavifDefaultTuneMetric = AOM_TUNE_PSNR; // Meaningless unless useLibavifDefaultTuneMetric.
-    if (quality != AVIF_QUALITY_LOSSLESS && !avifAOMOptionsContainExplicitTuning(codec, alpha)) {
-        useLibavifDefaultTuneMetric = AVIF_TRUE;
-        if (alpha) {
-            // Minimize ringing for alpha.
-            libavifDefaultTuneMetric = AOM_TUNE_PSNR;
-        } else {
-            libavifDefaultTuneMetric = AOM_TUNE_SSIM;
+
+    // libavif only needs to set the default tune metric for the first frame
```
y-guyon: Suggestion:
```c
// libavif only needs to set the default tune metric for the first frame,
// because libaom will persist that setting until explicitly changed.
```

src/codec_aom.c
```diff
@@ -403,16 +404,21 @@ static avifBool avifProcessAOMOptionsPreInit(avifCodec * codec, avifBool alpha,
...
-
 #if !defined(AOM_HAVE_TUNE_IQ)
 // Define the tune IQ value here if libaom doesn't define it. The enum value is guaranteed to never change
 // in libaom, so this definition won't ever get out of sync.
 #define AOM_TUNE_IQ 10
 #endif
 
+    if (useLibavifDefaultTuneMetric && libavifDefaultTuneMetric == AOM_TUNE_IQ) {
+        return AVIF_TRUE;
+    }
```
y-guyon: Suggestion:
```c
    if (useLibavifDefaultTuneMetric) {
        return libavifDefaultTuneMetric == AOM_TUNE_IQ;
    }
```

src/codec_aom.c
juliobbv-p: Sounds good, I'll convert this to an assert. I see that's the pattern used in the project.

src/codec_aom.c
```diff
@@ -423,16 +429,24 @@ static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
...
         // For consistent behavior, handle both cases where tune IQ was either specified as a string (tune=iq),
         // or as an enum value (tune=10).
-        if (avifKeyEqualsName(entry->key, "tune", alpha) && aomOptionParseEnum(entry->value, tuneIqEnum, &val)) {
-            ret = (val == AOM_TUNE_IQ);
+        if (avifKeyEqualsName(entry->key, "tune", alpha)) {
+            isAnyTuneDefined = AVIF_TRUE;
+
+            if (aomOptionParseEnum(entry->value, tuneIqEnum, &val)) {
+                useTuneIq = (val == AOM_TUNE_IQ);
+            }
```
wantehchang, line 437: IMPORTANT: This if statement should become an if-else statement, otherwise it cannot set `useTuneIq` back to false. Also, since the `tuneIqEnum` array has only one element (`AOM_TUNE_IQ`), if `aomOptionParseEnum(entry->value, tuneIqEnum, &val)` returns true, we know `val` must be equal to `AOM_TUNE_IQ`.  

```
            if (aomOptionParseEnum(entry->value, tuneIqEnum, &val)) {
                assert(val == AOM_TUNE_IQ);
                useTuneIq = AVIF_TRUE;
            } else {
                useTuneIq = AVIF_FALSE;
            }
```

src/codec_aom.c
```diff
@@ -423,16 +429,24 @@ static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha)
...
-    // option persists across frames in libaom. However AOM_TUNE_IQ is only supported with still images in
-    // libavif and libaom as of today, so there is no need to remember this option across frames.
-    return ret;
+    if (!isAnyTuneDefined && codec->internal->previousFrameUsedTuneIq) {
+        // Handle the case where the encoder was called with avifEncoderSetCodecSpecificOption("tune", "iq")
+        // for a previous frame and not called (or called with NULL) for this frame, because the tune
+        // option persists across frames in libaom.
+        // In this case, we know libaom will also use tune=iq for this frame.
+        return AVIF_TRUE;
+    }
```
wantehchang, line 447: IMPORTANT: I believe this if statement (and the return statement at the end of the function) should read:

```
    if (!isAnyTuneDefined) {
        // Some appropriate comment...
        return codec->internal->previousFrameUsedTuneIq;
    }
    
    return useTuneIq;
```

or equivalently (which I think is clearer):


```
    if (isAnyTuneDefined) {
        return useTuneIq;
    }

    // Some appropriate comment...
    return codec->internal->previousFrameUsedTuneIq;
```

src/codec_aom.c
```diff
@@ -725,23 +739,33 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+        if (quality != AVIF_QUALITY_LOSSLESS && !avifAOMOptionsContainExplicitTuning(codec, alpha)) {
+            useLibavifDefaultTuneMetric = AVIF_TRUE;
+            if (alpha) {
+                // Minimize ringing for alpha.
+                libavifDefaultTuneMetric = AOM_TUNE_PSNR;
+            } else {
+                libavifDefaultTuneMetric = AOM_TUNE_SSIM;
+            }
         }
+        AVIF_ASSERT_OR_RETURN(!codec->internal->previousFrameUsedTuneIq);
```
wantehchang, line 755: This assertion doesn't seem necessary. You want to assert `codec->internal->previousFrameUsedTuneIq` is initialized to false, right?

`codec->internal->previousFrameUsedTuneIq` should be initialized to false because libaom does not use tune=iq by default, right?

tests/gtest/avifprogressivetest.cc
```diff
@@ -87,6 +87,37 @@ TEST_F(ProgressiveTest, QualityChange) {
...
+// version where tune IQ was available
+TEST_F(ProgressiveTest, TuneIq) {
+  encoder_->extraLayerCount = 1;
+  // Tune IQ requires all-intra mode, which libavif determines when the first
+  // layer is encoded at a very low quality (e.g. quality 10)
+  encoder_->quality = 10;
+  encoder_->codecChoice = AVIF_CODEC_CHOICE_AOM;
+
+  ASSERT_EQ(avifEncoderSetCodecSpecificOption(encoder_.get(), "tune", "iq"),
+            AVIF_RESULT_OK);
```
wantehchang, line 100: [DO NOT COMMIT. FOR TESTING ONLY.]

I found that `avifEncoderSetCodecSpecificOption()` deduplicates the keys, but it does not consider the "color:" or "c:" prefix when it deduplicates the keys. So to test how to handle multiple "tune" options, we need to use the "color:" and "c:" prefixes.

Apply the following patch and then tweak it as you see fit to test the changes I suggested:

```
diff --git a/tests/gtest/avifprogressivetest.cc b/tests/gtest/avifprogressivetest.cc
index d9e211af..b8a75635 100644
--- a/tests/gtest/avifprogressivetest.cc
+++ b/tests/gtest/avifprogressivetest.cc
@@ -96,7 +96,11 @@ TEST_F(ProgressiveTest, TuneIq) {
   encoder_->quality = 10;
   encoder_->codecChoice = AVIF_CODEC_CHOICE_AOM;
 
-  ASSERT_EQ(avifEncoderSetCodecSpecificOption(encoder_.get(), "tune", "iq"),
+  ASSERT_EQ(avifEncoderSetCodecSpecificOption(encoder_.get(), "tune", "ssim"),
+            AVIF_RESULT_OK);
+  ASSERT_EQ(avifEncoderSetCodecSpecificOption(encoder_.get(), "c:tune", "psnr"),
+            AVIF_RESULT_OK);
+  ASSERT_EQ(avifEncoderSetCodecSpecificOption(encoder_.get(), "color:tune", "iq"),
             AVIF_RESULT_OK);
   avifResult result = avifEncoderAddImage(encoder_.get(), image_.get(), 1,
                                           AVIF_ADD_IMAGE_FLAG_NONE);
@@ -109,6 +113,10 @@ TEST_F(ProgressiveTest, TuneIq) {
 
   ASSERT_EQ(result, AVIF_RESULT_OK);
   encoder_->quality = 50;
+  ASSERT_EQ(avifEncoderSetCodecSpecificOption(encoder_.get(), "c:tune", "iq"),
+            AVIF_RESULT_OK);
+  ASSERT_EQ(avifEncoderSetCodecSpecificOption(encoder_.get(), "color:tune", "psnr"),
+            AVIF_RESULT_OK);
   ASSERT_EQ(avifEncoderAddImage(encoder_.get(), image_.get(), 1,
                                 AVIF_ADD_IMAGE_FLAG_NONE),
             AVIF_RESULT_OK);
```

src/codec_aom.c
```diff
@@ -725,23 +739,33 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
     }
 
     struct aom_codec_enc_cfg * cfg = &codec->internal->cfg;
     avifBool quantizerUpdated = AVIF_FALSE;
     // True if libavif knows that tune=iq is used, either by default by libavif, or explicitly set by the user.
     // False otherwise (including if libaom uses tune=iq by default, which is not the case as of v3.13.1 and earlier versions).
-    const avifBool useTuneIq = useLibavifDefaultTuneMetric ? libavifDefaultTuneMetric == AOM_TUNE_IQ : avifImageUsesTuneIq(codec, alpha);
+    const avifBool useTuneIq = avifImageUsesTuneIq(codec, alpha, useLibavifDefaultTuneMetric, libavifDefaultTuneMetric);
```
wantehchang, line 762: I think the original form is easier to understand:

```
    const avifBool useTuneIq = useLibavifDefaultTuneMetric ? libavifDefaultTuneMetric == AOM_TUNE_IQ : avifImageUsesTuneIq(codec, alpha);
```

wantehchang (commented):
Julio, Yannis: Sorry about the late review. I reviewed and tested the code carefully. (Thank you for adding the test!)

I didn't read your prior discussions, so I am sorry if I comment on an issue that you already discussed.

tests/gtest/avifprogressivetest.cc
wantehchang: Julio: We were planning to disallow setting the "tune" option after we have encoded a frame, but this PR allows that (and handles it correctly). I just wanted to confirm with you that this is intended. (This is fine by me.)

src/codec_aom.c
y-guyon: Previously this line was `codec->internal->previousFrameUsedTuneIq = AVIF_FALSE`. I requested it to be an assertion because "`previousFrameUsedTuneIq` should be initialized to false", as you said. We can also remove the assertion, this is fine with me.

tests/gtest/avifprogressivetest.cc
juliobbv-p: Correct, this is intended. After a conversation with Yannis, I was convinced to keep the current libavif behavior (being able to update the "tune" option after encoding a frame).

juliobbv-p:
> Julio: None of the issues I noted are serious, so I am going to merge this PR as is. Please address my comments in a follow-up PR.

Thanks @wantehchang for the review! I'll address your comments 👍.

src/codec_aom.c
juliobbv-p: After chatting with Wan-Teh, I'll just go ahead and remove the assert.

PR #3030, opened 2026-02-11
Fix avifCodec::csOptions comment

y-guyon:
I could not find any codec specific option value "marked as used".


include/avif/internal.h
```diff
@@ -598,10 +598,8 @@ typedef void (*avifCodecDestroyInternalFunc)(struct avifCodec * codec);
 typedef struct avifCodec
 {
     avifCodecSpecificOptions * csOptions; // Contains codec-specific key/value pairs for advanced tuning.
-                                          // If a codec uses a value, it must mark it as used.
```
wantehchang, line 601: I verified that the codec doesn't mark a value in this array as used. The following is from my code analysis (no need to add it as a comment):

The decoder doesn't use this array. The encoder clears this array before `avifEncoderAddImageInternal()` returns `AVIF_RESULT_OK`.

PR #3032, opened 2026-02-11
Fix the CICP for BT.709 SDR and BT.2100 HLG

wantehchang:
The CICP for BT.709 SDR should be 1/1/1, not 1/1/5.

The CICP for BT.2100 HLG should be 9/18/9, not 9/14/9.

Corresponds to the following two pull requests for the AV2 spec: https://github.com/AOMediaCodec/av2-spec-internal/pull/617 https://github.com/AOMediaCodec/av2-spec-internal/pull/615


wantehchang:
This is similar to the [AVM merge request !3098](https://gitlab.com/AOMediaCodec/avm/-/merge_requests/3098).

PR #3038, opened 2026-02-13
Reject invalid depth in avifImageAllocatePlanes()

y-guyon:
Also reject >16-bit in avifParsePixelInformationProperty() to avoid returning the wrong AVIF_RESULT_OUT_OF_MEMORY error code later. Add test.


wantehchang (approved):
LGTM. Thanks! I didn't review the test.

Issue #2427, opened 2024-09-11
Apply clap to the output of avifdec


y-guyon:
Clean Aperture must be applied to the decoded image item unless marked as non-essential.  
avifdec does not use the values of the `avifImage::clap` field.

Related: https://github.com/AOMediaCodec/libavif/pull/2426

PR #3041, opened 2026-02-17
Apply clean aperture crop when decoding to PNG or JPEG.

maryla-uc:
Apply clean aperture crop for PNG and JPEG. Because this is done in RGB, it does not cover the similar TODO for when saving to Y4M.
I chose to do the cropping in RGB and not in YUV because it bypasses the need to upsample the YUV to 444 in some cases.

#2427

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -376,7 +376,8 @@ if(AVIF_CODEC_AVM_ENABLED)
         if(AVIF_BUILD_APPS)
             # Disable all tests that use avifenc without explicitly setting --codec=avm.
             set_tests_properties(
-                test_cmd test_cmd_animation test_cmd_grid test_cmd_stdin test_cmd_targetsize PROPERTIES DISABLED True
+                test_cmd test_cmd_animation test_cmd_grid test_cmd_stdin test_cmd_targetsize test_cmd_transform
+                PROPERTIES DISABLED True
             )
         endif()
     endif()


apps/shared/avifjpeg.c
```diff
@@ -1664,29 +1665,39 @@ avifBool avifJPEGWrite(const char * outputFilename, const avifImage * avif, int
...
+    if (avifImageYUVToRGB(avif, &rgbData) != AVIF_RESULT_OK) {
         fprintf(stderr, "Conversion to RGB failed: %s\n", outputFilename);
         goto cleanup;
     }
 
+    avifRGBImage rgbView = rgbData;
+    if (avif->transformFlags & AVIF_TRANSFORM_CLAP) {
+        avifCropRect cropRect;
+        avifDiagnostics diag;
+        if (avifCropRectFromCleanApertureBox(&cropRect, &avif->clap, avif->width, avif->height, &diag) &&
```
y-guyon, line 1686: Is an incorrect `clap` silently ignored on purpose? Should it return an error instead? (I know, it was the case already)

y-guyon (approved):

apps/shared/avifjpeg.c
maryla-uc: Currently avifdec prints a warning ("Warning: Invalid Clean Aperture values").
I think it's enough...?
Returning an error for something that used to work seems like a risky change.

maryla-uc: Although to be fair this PR completely changes the result (by cropping the image), so I guess it's technically just as risky.

PR #3042, opened 2026-02-17
Add avifRGBImage to the C++ API

vrabaud:


maryla-uc:
The issue is that you are supposed to be able to also use RGBImage as a view, in which case it does not own the pixel, but it doesn't know it. RGBImage does not have any boolean saying whether it owns the pixels, it's the caller's responsibility to decide whether to call avifRGBImageFreePixels.
Maybe we can just call it RGBImageOwnedPixelsPtr ...? With a comment explaining usage?

vrabaud:
Let's go with RGBImagePtr + comment

include/avif/avif_cxx.h
```diff
@@ -18,17 +18,20 @@ namespace avif
 // Struct to call the destroy functions in a unique_ptr.
 struct UniquePtrDeleter
 {
-    void operator()(avifEncoder * encoder) const { avifEncoderDestroy(encoder); }
     void operator()(avifDecoder * decoder) const { avifDecoderDestroy(decoder); }
-    void operator()(avifImage * image) const { avifImageDestroy(image); }
+    void operator()(avifEncoder * encoder) const { avifEncoderDestroy(encoder); }
     void operator()(avifGainMap * gainMap) const { avifGainMapDestroy(gainMap); }
+    void operator()(avifImage * image) const { avifImageDestroy(image); }
+    void operator()(avifRGBImage * image) const { avifRGBImageFreePixels(image); }
```
wantehchang, line 25: Vincent: This doesn't seem correct. `avifRGBImageFreePixels()` is implemented as follows:

```
void avifRGBImageFreePixels(avifRGBImage * rgb)
{
    if (rgb->pixels) {
        avifFree(rgb->pixels);
    }

    rgb->pixels = NULL;
    rgb->rowBytes = 0;
}
```

Note that it doesn't free the `avifRGBImage` struct. Could you modify an existing test in tests/gtest/ to show how `avif::RGBImagePtr` should be used?

vrabaud: Fix at https://github.com/AOMediaCodec/libavif/pull/3048

wantehchang: Vincent: I thought about this class last night. I would like to suggest two alternatives.

The first alternative is to preserve the main functionality of `RGBImagePtr` but rename it `RGBImageCleanup` and define the class directly rather than using `std::unique_ptr`:

```
class RGBImageCleanup {
 public:
  RGBImageCleanup(avifRGBImage* rgb) : rgb_(rgb) {}
  ~RGBImageCleanup() { avifRGBImageFreePixels(rgb_); }
 private:
  avifRGBImage* rgb_ = nullptr;
};
```

By not using `std::unique_ptr`, we avoid the confusion caused by expecting it to free the `avifRGBImage` struct.

The second alternative is to derive a C++ `RGBImage` class from the C `avifRGBImage` struct. The destructor calls `avifRGBImageFreePixels(this)`. The class may optionally provide constructors and methods such as `AllocatePixels()` for convenience.

My preference is the first alternative because it is the minimal solution to the cleanup problem. It is similar to `absl::Cleanup` (or its predecessor in Google's internal repository).

PR #3043, opened 2026-02-18
Restrict sato image item's depth to 8,10,12,16

wantehchang:
Restrict sample transform derived image item's bit depth to 8, 10, 12, and 16, which are supported by commonly-used libraries such as libyuv.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2308,6 +2308,12 @@ static avifResult avifDecoderSampleTransformItemValidateProperties(const avifDec
         // This is enforced in avifParsePixelInformationProperty().
         AVIF_ASSERT_OR_RETURN(pixiProp->u.pixi.planeDepths[i] == pixiProp->u.pixi.planeDepths[0]);
     }
+    AVIF_ASSERT_OR_RETURN(pixiProp->u.pixi.planeCount >= 1);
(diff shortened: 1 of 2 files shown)


src/read.c
```diff
@@ -2308,6 +2308,12 @@ static avifResult avifDecoderSampleTransformItemValidateProperties(const avifDec
         // This is enforced in avifParsePixelInformationProperty().
         AVIF_ASSERT_OR_RETURN(pixiProp->u.pixi.planeDepths[i] == pixiProp->u.pixi.planeDepths[0]);
     }
+    AVIF_ASSERT_OR_RETURN(pixiProp->u.pixi.planeCount >= 1);
+    const uint8_t depth = pixiProp->u.pixi.planeDepths[0];
+    if (depth != 8 && depth != 10 && depth != 12 && depth != 16) {
```
wantehchang, line 2313: Yannis: This is a strawman proposal. The rationale is that the output image's bit depth should be one that clients of libavif are likely to be able to handle.

The criterion I used is whether one can use libyuv to convert the output YUV image to RGB.

Alternatively, we should describe how to convert an uncommon bit depth, such as 1 or 14, to a common bit depth, or provide such a conversion function.

PR #3044, opened 2026-02-18
AVM: be compatible with profiles when -DCONFIG_AV2_PROFILES=1

vrabaud:


src/codec_avm.c
```diff
@@ -509,6 +509,38 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
         // Profile 2.  8-bit and 10-bit 4:2:2
         //            12-bit 4:0:0, 4:2:0, 4:2:2 and 4:4:4
         uint8_t seqProfile = 0;
+#if defined(CONFIG_AV2_PROFILES) && CONFIG_AV2_PROFILES
+        if (image->depth == 12) {
```
y-guyon: If anything but 8 and 10 are unsupported, I would suggest to test that instead of 12 specifically, or to replace `// 8-bit or 10-bit` with `AVIF_ASSERT_OR_RETURN(image->depth == 8 || image->depth == 10)`

y-guyon: Sorry, that was not clear. I meant keeping the condition and testing for `!=8 && !=10`, or adding `AVIF_ASSERT` in the `else` branch.

Implemented in https://github.com/AOMediaCodec/libavif/pull/3046

PR #3046, opened 2026-02-18
Replace AVIF_ASSERT by avm image->depth condition

y-guyon:
Disable 12-bit AV2 tests.

### DIFF
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -512,7 +512,7 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
 #if defined(CONFIG_AV2_PROFILES) && CONFIG_AV2_PROFILES
         if (image->depth != 8 && image->depth != 10) {
             avifDiagnosticsPrintf(codec->diag, "%d-bit is not supported in AV2 encoder.", image->depth);
-            return AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION;
+            return AVIF_RESULT_INVALID_ARGUMENT;
         }
         // Based on https://gitlab.com/AOMediaCodec/avm/-/blob/main/av2/common/enums.h?ref_type=fcab0163f471b38fe593672fcbd24a6beb0be82e#L272
         if (alpha) {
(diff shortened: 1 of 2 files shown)


vrabaud (approved):

src/codec_avm.c
```diff
@@ -510,9 +510,10 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
         //            12-bit 4:0:0, 4:2:0, 4:2:2 and 4:4:4
         uint8_t seqProfile = 0;
 #if defined(CONFIG_AV2_PROFILES) && CONFIG_AV2_PROFILES
-        // Only 8-bit and 10-bit are supported.
-        AVIF_ASSERT_OR_RETURN(image->depth == 8 || image->depth == 10);
-
+        if (image->depth != 8 && image->depth != 10) {
+            avifDiagnosticsPrintf(codec->diag, "%d-bit is not supported in AV2 encoder.", image->depth);
+            return AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION;
```
wantehchang: The error code `AVIF_RESULT_INVALID_CODEC_SPECIFIC_OPTION` is for the `avifEncoderSetCodecSpecificOption()` function, so it is not right for this error condition. Perhaps `AVIF_RESULT_NOT_IMPLEMENTED` or `AVIF_RESULT_INVALID_ARGUMENT`?

tests/gtest/avifavmtest.cc
```diff
@@ -71,13 +71,29 @@ INSTANTIATE_TEST_SUITE_P(Basic, AvmTest,
 
 INSTANTIATE_TEST_SUITE_P(Tiny, AvmTest,
                          Combine(/*width=*/Values(1), /*height=*/Values(1),
-                                 /*depth=*/Values(8, 10, 12),
+                                 /*depth=*/Values(8, 10),
+                                 Values(AVIF_PIXEL_FORMAT_YUV444),
+                                 /*alpha=*/Values(false)));
+
+// TODO(yguyon): Implement or fix in avm then test the following combinations.
```
wantehchang: The initial set of profiles defined for AV2 only support bit depths 8 and 10. Bit depth 12 will need to be added in new profiles in a future update of the AV2 spec. So this test requires a future AV2 spec update.

wantehchang (approved):
LGTM. I suggest two changes.

src/codec_avm.c
y-guyon: Done.

tests/gtest/avifavmtest.cc
y-guyon: Replaced the TODO by a comment.

PR #3047, opened 2026-02-19
Make tests/test_cmd_transform.sh executable

y-guyon:


wantehchang:
There are three other shell scripts that aren't executable:

```
-rw-r----- 1 wtc primarygroup 2161 Nov 22 01:48 ./tests/test_cmd_depth_extension.sh
-rw-r----- 1 wtc primarygroup 1596 Jul 19  2025 ./tests/test_cmd_enc_boxes_golden.sh
-rw-r----- 1 wtc primarygroup 1598 Jan 27 10:50 ./tests/test_cmd_enc_gainmap_boxes_golden.sh
```

Should we also make them executable?

PR #3048, opened 2026-02-19
Replace RGBImagePtr by RGBImageCleanup

vrabaud:
Although RGBImagePtr, defined using std::unique_ptr, is correct, it is
confusing because it does not free the avifRGBImage struct. Since the
avifRGBImage struct is intended to be allocated on the stack, we should
not address this by forcing the users of RGBImagePtr to allocate
avifRGBImage from the heap.

Replace RGBImagePtr by RGBImageCleanup, which is modeled after
absl::Cleanup and is a lightweight cleanup object whose destructor
calls avifRGBImageFreePixels() on the associated avifRGBImage struct.

### DIFF
--- a/include/avif/avif_cxx.h
+++ b/include/avif/avif_cxx.h
@@ -31,6 +31,7 @@ using GainMapPtr = std::unique_ptr<avifGainMap, UniquePtrDeleter>;
 using ImagePtr = std::unique_ptr<avifImage, UniquePtrDeleter>;
 
 // Automatically cleans the ressources of the avifRGBImage.
+// To use when RGBImage actually owns the pixels. RGBImage can also be used as a view, in which case it does not own the pixels.
 class RGBImageCleanup
 {
 public:


include/avif/avif_cxx.h
```diff
@@ -22,16 +22,24 @@ struct UniquePtrDeleter
...
 
 // Use these unique_ptr to ensure the structs are automatically destroyed.
 using DecoderPtr = std::unique_ptr<avifDecoder, UniquePtrDeleter>;
 using EncoderPtr = std::unique_ptr<avifEncoder, UniquePtrDeleter>;
 using GainMapPtr = std::unique_ptr<avifGainMap, UniquePtrDeleter>;
 using ImagePtr = std::unique_ptr<avifImage, UniquePtrDeleter>;
-// To use when RGBImage actually owns the pixels. RGBImage can also be used as a view, in which case it does not own the pixels.
-using RGBImagePtr = std::unique_ptr<avifRGBImage, UniquePtrDeleter>;
+
+// Automatically cleans the ressources of the avifRGBImage if it owns it.
```
wantehchang: Nit: "if it owns it" is a conditional, but the destructor calls `avifRGBImageFreePixels()` unconditionally. I think the comment should instead say that this class assumes the avifRGBImage owns its `pixels` buffer.

tests/gtest/avifbasictest.cc
```diff
@@ -41,5 +41,17 @@ TEST(BasicTest, EncodeDecode) {
   //  testutil::WriteImage(decoded.get(), "/tmp/avifbasictest.png");
 }
 
+TEST(BasicTest, SharedPtr) {
+  // Make sure the following does not create sanitizer bugs.
+  ImagePtr image =
+      testutil::CreateImage(/*width=*/12, /*height=*/34, /*depth=*/8,
+                            AVIF_PIXEL_FORMAT_YUV420, AVIF_PLANES_ALL);
+
+  std::unique_ptr<avifRGBImage> rgb(new avifRGBImage());
```
wantehchang: Change this to a plain `avifRGBImage` local variable allocated on the stack. That is how `avifRGBImage` is intended to be used. `avifRGBImage` is really a struct that collects the parameters for the RGB-to-YUV and YUV-to-RGB conversion functions.

We should also create a local variable rather than a temporary object for `RGBImageCleanup`. The local variable `cleanup` is created as soon as the avifRGBImage variable is properly initialized:

```
  avifRGBImage rgb;
  avifRGBImageSetDefaults(&rgb, image.get());
  RGBImageCleanup cleanup(&rgb);
  ASSERT_EQ(avifRGBImageAllocatePixels(&rgb), AVIF_RESULT_OK);
```

tests/gtest/avifbasictest.cc
```diff
@@ -41,5 +41,17 @@ TEST(BasicTest, EncodeDecode) {
   //  testutil::WriteImage(decoded.get(), "/tmp/avifbasictest.png");
 }
 
+TEST(BasicTest, SharedPtr) {
```
wantehchang: Nit: "SharedPtr" doesn't seem to match what the test does...

include/avif/avif_cxx.h
```diff
@@ -22,16 +22,24 @@ struct UniquePtrDeleter
...
+class RGBImageCleanup
```
wantehchang, line 35: Vincent: I consider this PR as optional for the libavif v1.4.0 release, so I didn't label it with the v1.4.0 milestone.

But this PR is quite close. Would you like to try to address the review comments and merge it tomorrow (Friday)?

wantehchang: Vincent: I didn't realize that `RGBImagePtr` has already been merged. Then this PR is mandatory for v1.4.0.

I will use the "Edit file" feature to make the changes I suggested and then merge this PR.

include/avif/avif_cxx.h
wantehchang: I deleted "if it owns it".

tests/gtest/avifbasictest.cc
wantehchang: Done.

tests/gtest/avifbasictest.cc
wantehchang: I renamed the test to "RGBImageCleanup".

wantehchang:
Maryla: Since I edited this PR, could you please review it? Thanks.

include/avif/avif_cxx.h
```diff
@@ -22,16 +22,25 @@ struct UniquePtrDeleter
...
 // Use these unique_ptr to ensure the structs are automatically destroyed.
 using DecoderPtr = std::unique_ptr<avifDecoder, UniquePtrDeleter>;
 using EncoderPtr = std::unique_ptr<avifEncoder, UniquePtrDeleter>;
 using GainMapPtr = std::unique_ptr<avifGainMap, UniquePtrDeleter>;
 using ImagePtr = std::unique_ptr<avifImage, UniquePtrDeleter>;
+
+// Automatically cleans the ressources of the avifRGBImage.
 // To use when RGBImage actually owns the pixels. RGBImage can also be used as a view, in which case it does not own the pixels.
-using RGBImagePtr = std::unique_ptr<avifRGBImage, UniquePtrDeleter>;
+class RGBImageCleanup
```
wantehchang, line 35: Vignesh: Please review. This is modeled after `absl::MakeCleanup`. Thanks!

vigneshvg:
Since the new class is only to be used when the buffers are owned by the underlying pointer, can you please clarify a bit more in the commit message as to how the new class is different from the old unique_ptr wrapper?

Could one not use the existing unique_ptr wrapper the same way?

wantehchang:
Vignesh: I added a commit message in https://github.com/AOMediaCodec/libavif/pull/3048#issue-3963004983.

The existing unique_ptr wrapper (`RGBImagePtr`) is equivalent in functionality to the new `RGBImageCleanup`, so one can view this PR as only making the code less confusing. The existing unique_ptr wrapper is confusing because users expect unique_ptr to delete the struct that the pointer points to, but `RGBImagePtr` only frees the pixels buffer and does not free the `avifRGBImage` struct.

include/avif/avif_cxx.h
vrabaud: LGTM, thx !

PR #3049, opened 2026-02-20
gainmap: prevent integer overflow in plane allocation

rootvector2:


src/gainmap.c
```diff
@@ -558,11 +558,24 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
 
     avifResult res = AVIF_RESULT_OK;
     // --- After this point, the function should exit with 'goto cleanup' to free allocated resources.
+    // Overflow protection: 'width * height * sizeof(float)' uses signed int
+    // multiplication which is undefined behavior on overflow in C. Compute the
+    // allocation size in size_t with explicit overflow checks instead.
```
wantehchang, line 563: [All of my review comments are for future reference.]

Omit this kind of comment that explains an overflow check in detail. There are a lot of such overflow checks in libavif. We cannot afford to explain all of them in detail. libavif maintainers are expected to be familiar with them.

src/gainmap.c
```diff
@@ -558,11 +558,24 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
+    if (baseRgbImage->width > SIZE_MAX / sizeof(float)) {
+        res = AVIF_RESULT_INVALID_ARGUMENT;
+        goto cleanup;
+    }
+    const size_t gainMapPlaneRowBytes = (size_t)baseRgbImage->width * sizeof(float);
+    if (gainMapPlaneRowBytes != 0 && baseRgbImage->height > SIZE_MAX / gainMapPlaneRowBytes) {
+        res = AVIF_RESULT_INVALID_ARGUMENT;
+        goto cleanup;
+    }
```
wantehchang, line 572: If the width and height are for a buffer that have been allocated successfully, it is safe to multiply width and height in the size_t type. We can take advantage of this fact here. So for the problematic expression `width * height * sizeof(float)`, we only need one overflow check:

```
    const size_t numPixels = (size_t)width * height;
    if (numPixels > SIZE_MAX / sizeof(float)) {
        res = AVIF_RESULT_INVALID_ARGUMENT;
        goto cleanup;
    }
    const size_t gainMapPlaneSize = numPixels * sizeof(float);
```

rootvector2:
Thank you for the detailed review and for auditing the file more thoroughly. I appreciate the improvements in #3051 and am glad the report was useful.

wantehchang:
Dexter.k: I noticed you have focused your audit of libavif on calculations of allocation sizes. Since width, height, and rowBytes are all stored in `uint32_t` struct members, when we multiply width or rowBytes by height (or a loop index that may be as large as height - 1), we need to cast the operands to `size_t`. If you are interested in auditing more libavif code, I would encourage you to search for "rowBytes" (case-insensitive) and check any multiplication involving "rowBytes". After that is done, repeat it for "width".

PR #3050, opened 2026-02-20
Address #3029 PR feedback

juliobbv-p:


juliobbv-p:
@wantehchang as promised, here's the follow-up PR to address your comments.

src/codec_aom.c
```diff
@@ -433,20 +429,23 @@ static avifBool avifImageUsesTuneIq(const avifCodec * codec, avifBool alpha, avi
...
+    if (isAnyTuneDefined) {
+        return useTuneIq;
     }
 
-    return useTuneIq;
+    // Handle the case where the encoder was called with avifEncoderSetCodecSpecificOption("tune", "iq")
+    // for a previous frame and not called (or called with NULL) for this frame, because the tune
+    // option persists across frames in libaom.
+    // In this case, return what the previous frame used.
+    return codec->internal->previousFrameUsedTuneIq;
```
wantehchang, line 448: Re: the assertion at line 755 that I asked you to delete:

Ideally the property we want to assert is that `codec->internal->previousFrameUsedTuneIq` is not used when we encode the first frame because there isn't a previous frame. So we could add the following assertion before this return statement:

```
    assert(!codec->internal->encoderInitialized);
    return codec->internal->previousFrameUsedTuneIq;
```

We can almost prove this is true except for a small extra condition: we also need `quality != AVIF_QUALITY_LOSSLESS`. See lines 739-754.

We should look into why `quality != AVIF_QUALITY_LOSSLESS` was added to line 745.

juliobbv-p: > We should look into why quality != AVIF_QUALITY_LOSSLESS was added to line 745.

@y-guyon can correct me if I'm wrong, but my understanding is that because lossless doesn't have to care about distortion, there's nothing in the rate-distortion optimization process to "tune" for. As a result, setting a particular tuning mode is meaningless in libaom (with lossless), so libavif doesn't bother setting a default tune.

y-guyon: That is my understanding as well.

PR #3051, opened 2026-02-21
Prevent integer overflows in src/gainmap.c

wantehchang:
Prevent integer overflows in multiplications involving width, height, and rowBytes in src/gainmap.c by performing the multiplications in the size_t type. The size_t type is large enough because pixel buffers for the width, height, and rowBytes have been allocated successfully.

"Dexter.k" <164054284+rootvector2@users.noreply.github.com> reported an integer overflow in the allocation of the gainMapF buffers in avifRGBImageComputeGainMap() and suggested a fix in https://github.com/AOMediaCodec/libavif/pull/3049.


wantehchang:
@rootvector2 Dexter.k: FYI. I wrote this pull request by starting from your pull request https://github.com/AOMediaCodec/libavif/pull/3049. After I discovered that much more changes should be made, I decided to create my own pull request rather than suggest changes to your pull request, and treat your pull request as a bug report.

PR #3054, opened 2026-02-23
fix: resolve alignment violations

uwezkhan:


include/avif/internal.h
```diff
@@ -19,6 +21,22 @@ extern "C" {
 #define AVIF_MIN(a, b) (((a) < (b)) ? (a) : (b))
 #define AVIF_MAX(a, b) (((a) > (b)) ? (a) : (b))
 
+// Alignment-safe load/store for 16-bit values from potentially unaligned byte pointers.
```
wantehchang, line 24: Thank you for reporting this issue. The libavi API assumes that the `uint8_t *` buffers, if they contain high bit depth (> 8) pixels, point to `uint16_t` arrays, so that it is safe to cast the `uint8_t *` buffers for high bit depth pixels to `uint16_t *`. This is an API contract.

I did not check this pull request in its entirety (because it is very long). Under this API contract, I believe none of the changes in this pull request are needed. Could you please go over your pull request and see if there are any changes that should be kept? Thanks.

include/avif/internal.h
```diff
@@ -19,6 +21,22 @@ extern "C" {
...
+// Using memcpy avoids undefined behavior from casting uint8_t* to uint16_t* without
+// alignment guarantees (C standard §6.3.2.3). Modern compilers optimize these to a single
+// load/store instruction on platforms that support unaligned access (e.g., x86/x64).
```
wantehchang, line 27: This is correct. libavif takes care to use `memcpy` for unaligned integers. You can search for "memcpy" in src/stream.c to find the `memcpy` calls used to read or write unaligned 16-bit, 32-bit, or 64-bit integers.

include/avif/internal.h
wantehchang: We can consider checking this API contract on entry to the relevant functions.

`avifImage` structs are managed by libavif, so we can safely assume their pixels buffers are aligned.

The `pixels` buffers in `avifRGBImage` structs may be allocated by libavif, but they can also point to buffers allocated by the caller. So I think it suffices to validate that the `pixels` buffers in `avifRGBImage` structs are aligned.

But this is an extremely unlikely issue in practice because it would be inconvenient to the callers themselves if they allocated unaligned buffers for high bit depth pixels.

wantehchang:
uwezkhan: I am going to close this PR.

PR #3055, opened 2026-02-23
Set a callback function for dav1d_log()

wantehchang:
The callback function directs dav1d_log() output to the error message buffer in the avifDiagnostics struct.

### DIFF
--- a/src/codec_dav1d.c
+++ b/src/codec_dav1d.c
@@ -41,9 +41,7 @@ static void avifDav1dFreeCallback(const uint8_t * buf, void * cookie)
 static void avifDav1dLogCallback(void * cookie, const char * format, va_list ap)
 {
     avifCodec * codec = (avifCodec *)cookie;
-    size_t size = AVIF_DIAGNOSTICS_ERROR_BUFFER_SIZE;
-    vsnprintf(codec->diag->error, size - 1, format, ap);
-    codec->diag->error[size - 1] = '\0';
+    vsnprintf(codec->diag->error, AVIF_DIAGNOSTICS_ERROR_BUFFER_SIZE, format, ap);
 }
 
 static void dav1dCodecDestroyInternal(avifCodec * codec)


src/codec_dav1d.c
```diff
@@ -70,6 +79,8 @@ static avifBool dav1dCodecGetNextImage(struct avifCodec * codec,
         // a message, so we set frame_size_limit to at most 8192 * 8192 to avoid the dav1d_log
         // message.
         dav1dSettings.frame_size_limit = (sizeof(size_t) < 8) ? AVIF_MIN(codec->imageSizeLimit, 8192 * 8192) : codec->imageSizeLimit;
+        dav1dSettings.logger.cookie = codec;
+        dav1dSettings.logger.callback = avifDav1dLogCallback;
```
wantehchang, line 81: Here is an example of avifdec output.

Before:

```
Frame size 49337x8192 exceeds limit 268435456
Diagnostics:
```

Note that dav1d_log() writes its error message "Frame size 49337x8192 exceeds limit 268435456" to stderr directly. libavif's error diagnostics message is the less informative "tile->codec->getNextImage() failed".

After:

```
Diagnostics:
 * Frame size 49337x8192 exceeds limit 268435456
```

Nothing from dav1d is written to stderr. libavif's error diagnostics message is "Frame size 49337x8192 exceeds limit 268435456", written by dav1d_log().

src/codec_dav1d.c
```diff
@@ -37,6 +38,14 @@ static void avifDav1dFreeCallback(const uint8_t * buf, void * cookie)
     (void)cookie;
 }
 
+static void avifDav1dLogCallback(void * cookie, const char * format, va_list ap)
+{
+    avifCodec * codec = (avifCodec *)cookie;
+    size_t size = AVIF_DIAGNOSTICS_ERROR_BUFFER_SIZE;
+    vsnprintf(codec->diag->error, size - 1, format, ap);
+    codec->diag->error[size - 1] = '\0';
```
wantehchang: The reason we are passing `size - 1` to `vsnprintf()` and ensuring the `codec->diag->error` buffer is null-terminated is for compatibility with the `_vsnprintf()` function in older versions of Microsoft Visual C++. This is described in the Note in (Microsoft's documentation for vsnprintf, _vsnprintf, etc](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/vsnprintf-vsnprintf-vsnprintf-l-vsnwprintf-vsnwprintf-l?view=msvc-170):

> To ensure that there's room for the terminating null when calling **`_vsnprintf`**, **`_vsnprintf_l`**, **`_vsnwprintf`** and **`_vsnwprintf_l`**, be sure that `count` is strictly less than the buffer length and initialize the buffer to null prior to calling the function.

Since we don't have a macro like:

```
#define vsnprintf _vsnprintf
```

I wonder if we still need this accomodation for the `_vsnprintf()` function. What do you think?

jzern: This is safer, but since there's a C99 requirement I think it's OK to assume a working implementation. Otherwise a comment would help keeping this from being updated in the future.

I don't see anything under https://learn.microsoft.com/en-us/cpp/c-runtime-library/compatibility?view=msvc-170 that would suggest `_vsnprintf` could be used directly.

jzern (approved):

src/codec_dav1d.c
vigneshvg: +1 to adding the explanation as a comment in the code.

vigneshvg (approved):

src/codec_dav1d.c
wantehchang: Thank you both for your comments. I will assume we have a recent MSVC with `vsnprintf()` and remove the workaround for `_vsnprintf()`.

PR #3056, opened 2026-02-23
Handle null detail after aom_codec_decode failure

wantehchang:
After aom_codec_decode() fails, aom_codec_error_detail() may return a null pointer. Handle it and replace a null pointer with the "no error detail" string.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -153,10 +153,11 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
         } else if (sample) {
             codec->internal->iter = NULL;
             if (aom_codec_decode(&codec->internal->decoder, sample->data.data, sample->data.size, NULL)) {
+                const char * error_detail = aom_codec_error_detail(&codec->internal->decoder);
                 avifDiagnosticsPrintf(codec->diag,
                                       "aom_codec_decode() failed: %s: %s",
                                       aom_codec_error(&codec->internal->decoder),
-                                      aom_codec_error_detail(&codec->internal->decoder));
+                                      error_detail ? error_detail : "no error detail");
                 return AVIF_FALSE;
             }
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -153,10 +153,11 @@ static avifBool aomCodecGetNextImage(struct avifCodec * codec,
         } else if (sample) {
             codec->internal->iter = NULL;
             if (aom_codec_decode(&codec->internal->decoder, sample->data.data, sample->data.size, NULL)) {
+                const char * error_detail = aom_codec_error_detail(&codec->internal->decoder);
                 avifDiagnosticsPrintf(codec->diag,
                                       "aom_codec_decode() failed: %s: %s",
                                       aom_codec_error(&codec->internal->decoder),
-                                      aom_codec_error_detail(&codec->internal->decoder));
+                                      error_detail ? error_detail : "no error detail");
```
wantehchang, line 160: There are many other `aom_codec_error_detail()` calls in this file. To save time, I only handle a null pointer return value here.

Should we change `aom_codec_error_detail()` to return the "no error detail" or "error detail not available" string instead of a null pointer?

wantehchang, line 160: Here is an example of the avifdec output.

Before:

```
Diagnostics:
 * aom_codec_decode() failed: Corrupt frame detected: (null)
```

Note that "(null)" is printed in the error diagnostics message at the end.

After:

```
Diagnostics:
 * aom_codec_decode() failed: Corrupt frame detected: no error detail
```

vigneshvg: updating libaom seems like the right thing to do here.

wantehchang: On second thought, I think it is more flexible to leave `aom_codec_error_detail()` unchanged to make it easy to detect that no detailed error information is available.

PR #3059, opened 2026-02-25
Apply rotation and mirroring when decoding to PNG or JPEG.

maryla-uc:
#2427

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -534,7 +534,7 @@ static avifBool avifPNGReadImpl(FILE * f,
         fprintf(stderr, "avifPNGRead internal error: rowBytes mismatch libavif %u vs libpng %" AVIF_FMT_ZU "\n", rgb.rowBytes, rowBytes);
         goto cleanup;
     }
-    rowPointers = (png_bytep *)malloc(sizeof(png_bytep) * (size_t)rgb.height);
+    rowPointers = (png_bytep *)malloc(sizeof(png_bytep) * rgb.height);
     if (rowPointers == NULL) {
         fprintf(stderr, "avifPNGRead internal error: memory allocation failure");
         goto cleanup;
@@ -827,7 +827,7 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
         png_write_chunk(png, cicp, cicpData, 4);
     }
 
-    rowPointers = (png_bytep *)malloc(sizeof(png_bytep) * (size_t)height);
+    rowPointers = (png_bytep *)malloc(sizeof(png_bytep) * height);
     if (rowPointers == NULL) {
         fprintf(stderr, "Error writing PNG: memory allocation failure");
         goto cleanup;


apps/shared/avifjpeg.c
```diff
@@ -1724,19 +1725,10 @@ avifBool avifJPEGWrite(const char * outputFilename, const avifImage * avif, int
...
+        result = avifSetExifOrientation(&exif, 1);
         if (result != AVIF_RESULT_OK) {
-            // Ignore errors if the orientation is the default one because not being able to set Exif orientation now
-            // means a reader will not be able to parse it later either.
-            if (orientation != 1) {
-                fprintf(stderr, "Error writing JPEG metadata: %s\n", avifResultToString(result));
-                avifRWDataFree(&exif);
-                goto cleanup;
-            }
+            // Ignore errors: if the exif is invalid, we can consider it as equivalent to not having an orientation.
```
y-guyon: Maybe `assert(result == AVIF_RESULT_INVALID_EXIF_PAYLOAD || result == AVIF_RESULT_NOT_IMPLEMENTED);` to be safe

apps/shared/avifutil.c
```diff
@@ -659,11 +659,150 @@ void avifRGBImageSetViewRect(avifRGBImage * dstImage, const avifRGBImage * srcIm
...
+    dstImage->height = newHeight;
+    dstImage->pixels = NULL;
+    avifResult result = avifRGBImageAllocatePixels(dstImage);
+    if (result != AVIF_RESULT_OK) {
+        return result;
+    }
+
+    if (rotation->angle == 0) {
+        const size_t bytesPerRow = (size_t)bytesPerPixel * srcImage->width;
+        // 0 degrees. Just copy the rows as is.
```
y-guyon, line 689: Nit: In this case the function could return right after `*dstImage = *srcImage;` (note that this branch is never reached when called from `avifApplyTransforms()` anyway)

apps/shared/avifutil.c
```diff
@@ -659,11 +659,150 @@ void avifRGBImageSetViewRect(avifRGBImage * dstImage, const avifRGBImage * srcIm
...
+        for (uint32_t j = 0; j < srcImage->height; ++j) {
+            memcpy(dstImage->pixels + (j * dstImage->rowBytes), srcImage->pixels + (j * srcImage->rowBytes), bytesPerRow);
```
y-guyon: `(size_t)j * dstImage->rowBytes`

Or store and increment pointers instead

apps/shared/avifutil.c
```diff
@@ -659,11 +659,150 @@ void avifRGBImageSetViewRect(avifRGBImage * dstImage, const avifRGBImage * srcIm
...
+}
+
+avifResult avifRGBImageMirror(avifRGBImage * image, const avifImageMirror * mirror)
+{
+    if (mirror->axis == 0) { // Horizontal axis.
+        const uint32_t bytesPerPixel = avifRGBImagePixelSize(image);
+        // May be less than image->rowBytes e.g. if image is a cropped view.
+        const size_t bytesPerRowToMove = (size_t)bytesPerPixel * image->width;
+        // Top-to-bottom
+        uint8_t * tempRow = (uint8_t *)avifAlloc(bytesPerRowToMove);
```
y-guyon, line 737: Nit: Not strictly necessary, is it faster and/or more convenient to use a temporary buffer? `mirror=1` do it pixel by pixel anyway.

apps/shared/avifutil.c
```diff
@@ -659,11 +659,150 @@ void avifRGBImageSetViewRect(avifRGBImage * dstImage, const avifRGBImage * srcIm
...
+        for (uint32_t y = 0; y < image->height / 2; ++y) {
+            uint8_t * row1 = &image->pixels[(size_t)y * image->rowBytes];
+            uint8_t * row2 = &image->pixels[(size_t)(image->height - 1 - y) * image->rowBytes];
+            memcpy(tempRow, row1, bytesPerRowToMove);
+            memcpy(row1, row2, bytesPerRowToMove);
+            memcpy(row2, tempRow, bytesPerRowToMove);
+        }
+        avifFree(tempRow);
+    } else if (mirror->axis == 1) { // Vertical axis.
+        const uint32_t pixelSize = avifRGBImagePixelSize(image);
```
y-guyon: `bytesPerPixel`?

apps/shared/avifutil.c
```diff
@@ -659,11 +659,150 @@ void avifRGBImageSetViewRect(avifRGBImage * dstImage, const avifRGBImage * srcIm
...
+            avifRGBImageSetViewRect(dstView, srcImage, &cropRect);
+        } else {
+            fprintf(stderr, "Invalid clean aperture box\n");
+            return AVIF_RESULT_INVALID_ARGUMENT;
+        }
+    }
+    if (avif->transformFlags & AVIF_TRANSFORM_IROT && avif->irot.angle != 0) {
+        avifRGBImage tmpRgbImage;
+        avifResult result = avifRGBImageRotate(&tmpRgbImage, dstView, &avif->irot);
+        if (result != AVIF_RESULT_OK) {
```
y-guyon, line 792: Not necessary with the current PR, but to be safe:

```c
avifRGBImageFreePixels(tmpRgbImage);
```

apps/shared/avifpng.c
```diff
@@ -685,18 +685,14 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
     avifRGBImage rgbView = rgbData;
-    if (hasClap) {
-        avifCropRect cropRect;
-        avifDiagnostics diag;
-        if (avifCropRectFromCleanApertureBox(&cropRect, &avif->clap, avif->width, avif->height, &diag) &&
-            (cropRect.x != 0 || cropRect.y != 0 || cropRect.width != avif->width || cropRect.height != avif->height)) {
-            avifRGBImageSetViewRect(&rgbView, &rgbData, &cropRect);
-            width = cropRect.width;
-            height = cropRect.height;
+    avifResult transformResult = avifApplyTransforms(&rgbView, &rgbData, avif);
```
y-guyon, line 690: Suggestion:

```c
// rgbView is a view on rgbData. avifApplyTransforms() may modify rgbData.
```

apps/shared/avifjpeg.c
maryla-uc: Done, I added a check for those errors, and I return in other cases.

apps/shared/avifutil.c
maryla-uc: This would make the contract more complicated. The current contract says that dstImage owns the returned pixels.  The RGBImage struct itself doesn't know whether it owns its pixels, so this would have to be communicated in some way. And in any case this path is indeed not used but just there for completeness.

apps/shared/avifutil.c
maryla-uc: Well it should be faster although I haven't benchmarked it.

apps/shared/avifpng.c
```diff
@@ -534,7 +534,7 @@ static avifBool avifPNGReadImpl(FILE * f,
         fprintf(stderr, "avifPNGRead internal error: rowBytes mismatch libavif %u vs libpng %" AVIF_FMT_ZU "\n", rgb.rowBytes, rowBytes);
         goto cleanup;
     }
-    rowPointers = (png_bytep *)malloc(sizeof(png_bytep) * rgb.height);
+    rowPointers = (png_bytep *)malloc(sizeof(png_bytep) * (size_t)rgb.height);
```
y-guyon: `sizeof()` returns `size_t`. This should not be necessary.

apps/shared/avifjpeg.c
```diff
@@ -1724,15 +1726,14 @@ avifBool avifJPEGWrite(const char * outputFilename, const avifImage * avif, int
         }
         memcpy(exif.data, AVIF_JPEG_EXIF_HEADER, AVIF_JPEG_EXIF_HEADER_LENGTH);
         memcpy(exif.data + AVIF_JPEG_EXIF_HEADER_LENGTH, avif->exif.data + exifTiffHeaderOffset, avif->exif.size - exifTiffHeaderOffset);
-        // Make sure the Exif orientation matches the irot/imir values.
-        // libheif does not have the same behavior. The orientation is applied to samples and orientation data is discarded there,
-        // see https://github.com/strukturag/libheif/blob/ea78603d8e47096606813d221725621306789ff2/examples/encoder_jpeg.cc#L187
-        const uint8_t orientation = avifImageGetExifOrientationFromIrotImir(avif);
-        result = avifSetExifOrientation(&exif, orientation);
+        // We already rotated the pixels if necessary in avifApplyTransforms(), so we set the orientation to 1 (no rotation, no mirror).
+        result = avifSetExifOrientation(&exif, 1);
```
wantehchang, line 1730: Maryla: This `avifSetExifOrientation(&exif, 1)` call was added to avifjpeg.c but was not added to avifpng.c. I can observe a difference when decoding the attached happy_dog.avif file to a JPEG output and a PNG output. The following patch (to illustrate the bug, not for check-in) fixes the difference:

```
diff --git a/apps/shared/avifpng.c b/apps/shared/avifpng.c
index 3e6b034f..826b4f2b 100644
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -783,6 +783,7 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
             fprintf(stderr, "Error writing PNG: Exif metadata is too big\n");
             goto cleanup;
         }
+        avifResult result = avifSetExifOrientation((avifRWData *)&avif->exif, 1);
         png_set_eXIf_1(png, info, (png_uint_32)avif->exif.size, avif->exif.data);
     }
     if (avif->xmp.data && (avif->xmp.size > 0)) {
```

[happy_dog.avif.zip](https://github.com/user-attachments/files/25618197/happy_dog.avif.zip)

wantehchang: Maryla: I filed issue https://github.com/AOMediaCodec/libavif/issues/3070. Let's continue the discussion there.

PR #3060, opened 2026-02-25
Refactor avifAOMOptionsContainExplicitTuning()

y-guyon:
Hopefully the logic around useTuneIq is easier to follow.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -390,29 +390,34 @@ static avifBool avifKeyEqualsName(const char * key, const char * name, avifBool 
 // in libaom, so this definition won't ever get out of sync.
 #define AOM_TUNE_IQ 10
 #endif
-static const struct aomOptionEnumList tuningEnum[] = { { "psnr", AOM_TUNE_PSNR },
-                                                       { "ssim", AOM_TUNE_SSIM },
-                                                       { "iq", AOM_TUNE_IQ },
-                                                       { NULL, 0 } };
+// Tune IQ string -> enum mapping
+static const struct aomOptionEnumList tuneIqEnum[] = { { "iq", AOM_TUNE_IQ }, { NULL, 0 } };
 
 // Returns true if codec-specific options for AOM contain a tune metric setting. Returns false otherwise.
-// Sets tuneMetric if codec-specific options for AOM contain a known tune metric setting. Otherwise sets tuneMetric to -1.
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -385,11 +385,30 @@ static avifBool avifKeyEqualsName(const char * key, const char * name, avifBool
...
-static avifBool avifAOMOptionsContainExplicitTuning(const avifCodec * codec, avifBool alpha)
+#if !defined(AOM_HAVE_TUNE_IQ)
+// Define the tune IQ value here if libaom doesn't define it. The enum value is guaranteed to never change
+// in libaom, so this definition won't ever get out of sync.
+#define AOM_TUNE_IQ 10
+#endif
+static const struct aomOptionEnumList tuningEnum[] = { { "psnr", AOM_TUNE_PSNR },
+                                                       { "ssim", AOM_TUNE_SSIM },
+                                                       { "iq", AOM_TUNE_IQ },
+                                                       { NULL, 0 } };
```
wantehchang: I changed this array back to the `tuneIqEnum` array in the original code. I also restored the `tuningEnum` array at its original location.

src/codec_aom.c
```diff
@@ -385,11 +385,30 @@ static avifBool avifKeyEqualsName(const char * key, const char * name, avifBool
...
+
+// Returns true if codec-specific options for AOM contain a tune metric setting. Returns false otherwise.
+// Sets tuneMetric if codec-specific options for AOM contain a known tune metric setting. Otherwise sets tuneMetric to -1.
+static avifBool avifAOMOptionsContainExplicitTuning(const avifCodec * codec, avifBool alpha, int * tuneMetric)
```
wantehchang: I changed the output parameter `int * tuneMetric` to `avifBool * useTuneIq` because that's what the caller cares about.

src/codec_aom.c
```diff
@@ -385,11 +385,30 @@ static avifBool avifKeyEqualsName(const char * key, const char * name, avifBool
...
 {
+    *tuneMetric = -1;
+    // If there are multiple "tune" options specified, honor the last one.
```
wantehchang, line 403: To honor the last "tune" option, the for loop must run to completion as the `avifImageUsesTuneIq()` function in the original code does.

wantehchang (approved):

PR #3063, opened 2026-02-27
Audit avifImageYUVAnyToRGBAnySlow(): use signed arithmetic for UV sam…

rootvector2:
…pling offsets

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -683,10 +683,10 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
     for (uint32_t j = 0; j < image->height; ++j) {
         // uvJ is used only when yuvHasColor is true.
         const uint32_t uvJ = yuvHasColor ? (j >> state->yuv.formatInfo.chromaShiftY) : 0;
-        const uint8_t * ptrY8 = &yPlane[j * yRowBytes];
-        const uint8_t * ptrU8 = uPlane ? &uPlane[(uvJ * uRowBytes)] : NULL;
-        const uint8_t * ptrV8 = vPlane ? &vPlane[(uvJ * vRowBytes)] : NULL;
-        const uint8_t * ptrA8 = aPlane ? &aPlane[j * aRowBytes] : NULL;
+        const uint8_t * ptrY8 = &yPlane[(size_t)j * yRowBytes];
+        const uint8_t * ptrU8 = uPlane ? &uPlane[(size_t)uvJ * uRowBytes] : NULL;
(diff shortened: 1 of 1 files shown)


wantehchang:
Dexter.k: Thank you for the pull request. We are in the middle of creating the libavif v1.4.0 release. I will review this PR next week.

src/reformat.c
```diff
@@ -683,10 +683,10 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
     for (uint32_t j = 0; j < image->height; ++j) {
         // uvJ is used only when yuvHasColor is true.
         const uint32_t uvJ = yuvHasColor ? (j >> state->yuv.formatInfo.chromaShiftY) : 0;
-        const uint8_t * ptrY8 = &yPlane[j * yRowBytes];
-        const uint8_t * ptrU8 = uPlane ? &uPlane[(uvJ * uRowBytes)] : NULL;
-        const uint8_t * ptrV8 = vPlane ? &vPlane[(uvJ * vRowBytes)] : NULL;
-        const uint8_t * ptrA8 = aPlane ? &aPlane[j * aRowBytes] : NULL;
+        const uint8_t * ptrY8 = &yPlane[(size_t)j * yRowBytes];
```
vrabaud: Please use uint64_t as size_t can be 32 bit on 32 bit platforms.

rootvector2:
@vrabaud done

src/reformat.c
wantehchang: Vincent: Please see my comment at https://github.com/AOMediaCodec/libavif/pull/3097#pullrequestreview-3936435562, which explains why we can multiply rowBytes by height or an index that is less than height using the `size_t` type after the pixel buffer has been allocated successfully.

This is why libavif uses the `size_t` type in such multiplications. Before we allocate a pixel buffer, it is not safe to do these multiplications using the `size_t` type, so we need to either detect overflow or use the `uin64_t` type.

In the `avifImageYUVAnyToRGBAnySlow()` function (the subject of this pull request), we may multiply rowBytes by -1 to go back the the previous row, so we need to use the signed `ptrdiff_t` type in those places. We need to review the changes to this function carefully.

rootvector2:
Reverted `uint64_t` back to `size_t` for the row pointer offset calculations. After going through your explanation on #3097, it makes sense — once the pixel buffer has been successfully allocated, `size_t` is guaranteed to be safe for `rowBytes * index` multiplications. The `ptrdiff_t` changes for the signed UV adjustments remain as-is. Let me know if anything else needs to be adjusted.

rootvector2:
@wantehchang any update

wantehchang:

PR #3064, opened 2026-02-27
This PR replaces the use of ftell() with platform-correct 64-bit file…

nidu-ninja:
… APIs when determining file size in avifIOCreateFileReader().
The previous implementation relied on long, which may be 32-bit on some platforms, causing incorrect file size detection for files larger than 2GB.

The change ensures consistent and correct large-file handling across Windows and POSIX systems without altering public APIs or parsing behavior.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -157,7 +157,6 @@ avifIO * avifIOCreateFileReader(const char * filename)
         fclose(f);
         return NULL;
     }
-
 #else
     // POSIX large file support
     if (fseeko(f, 0, SEEK_END) != 0) {


src/io.c
```diff
@@ -142,13 +142,40 @@ avifIO * avifIOCreateFileReader(const char * filename)
...
+    if (fileSizeSigned < 0) {
+        fclose(f);
+        return NULL;
+    }
+    if (fseeko(f, 0, SEEK_SET) != 0) {
+        fclose(f);
+        return NULL;
+    }
+    uint64_t fileSize = (uint64_t)fileSizeSigned;
+#endif
```
wantehchang, line 175: NiDU-NINJA: Thank you for the pull request. We are in the middle of creating the libavif v1.4.0 release. I will review this PR next week.

Some preliminary comments.

1. It is disappointing that C23 did not address the use of the `long` type in `fseek()` and `ftell()`.
2. I think your PR is correct. Since the `fseek()` call at line 113 in `avifIOFileReaderRead()` needs the same treatment, it would be better to create libavif's own wrapper functions for the `fseek()` and `ftell()` variants that support large files.
    - We can be POSIX-centric and provide `fseeko()` and `ftello()` wrappers for Windows, or
    - we can be neutral and provide `avifFseek64()` and `avifFtell64()` for POSIX and Windows.
    - I can take care of the wrappers in a follow-up PR.

src/io.c
```diff
@@ -142,13 +142,40 @@ avifIO * avifIOCreateFileReader(const char * filename)
...
+    if (fileSizeSigned < 0) {
+        fclose(f);
+        return NULL;
+    }
+    if (_fseeki64(f, 0, SEEK_SET) != 0) {
+        fclose(f);
+        return NULL;
+    }
+    uint64_t fileSize = (uint64_t)fileSizeSigned;
+
```
wantehchang: Nit: Delete the blank line.

nidu-ninja:
Thank you for the feedback and for reviewing the PR.

I agree that the `fseek()` call in `avifIOFileReaderRead()` would benefit from the same large-file handling, and that introducing wrapper functions for the large-file variants would provide a cleaner and more consistent solution across the codebase.

For this PR, I intentionally kept the change minimal and focused on correcting the file size detection in `avifIOCreateFileReader()`.

I appreciate your suggestion and your offer to handle the wrappers in a follow-up PR. I look forward to your review.

nidu-ninja:
I have applied the change and updated the code to keep fileSize as a signed variable, casting it to uint64_t only when assigning it to sizeHint, as recommended. The updated patch has been edited in file.

PR #3068, opened 2026-02-27
Use Tune IQ for layered image inter-frame encoding

juliobbv-p:
Additionally, for avifenc's `--progressive`: scale the first layer by a half for further file size savings.

Note: Tune IQ for inter-frame encoding is expected to be available in libaom v3.14.0.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -993,6 +993,10 @@ static avifBool avifEncodeRestOfLayeredImage(avifEncoder * encoder,
             // reversed lerp, so that last layer reaches exact targetQuality
             encoder->quality = targetQuality - (targetQuality - PROGRESSIVE_START_QUALITY) *
                                                    (encoder->extraLayerCount - layerIndex) / encoder->extraLayerCount;
+            // We scaled the first layer by a half (numerator: 1, denominator: 2).
+            // Don't perform any scaling for the second layer (numerator: 1, denominator: 1).
+            const avifScalingMode scalingMode = { { 1, 1 }, { 1, 1 } };
+            encoder->scalingMode = scalingMode;
         } else {
             const avifInputFile * nextFile = avifInputGetFile(input, layerIndex);
             // main() function should set number of layers to number of input,
@@ -1160,8 +1164,11 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
         // we should not reach here.
         assert(encoder->quality >= PROGRESSIVE_WORST_QUALITY);
         // Encode the base layer with a very low quality to ensure a small encoded size.
-        encoder->quality = 2;
+        encoder->quality = 10;
         // Low alpha quality resulted in weird artifact, so we don't do it.
+        // For further size savings, scale the first layer by a half (numerator: 1, denominator: 2).
+        const avifScalingMode scalingMode = { { 1, 2 }, { 1, 2 } };
+        encoder->scalingMode = scalingMode;
     }
 
     if (settings->layers > 1) {
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -666,16 +666,25 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         codec->internal->qualityFirstLayer = quality;
     }
 
-    // Determine whether the encoder should be configured to use intra frames only, either by setting aomUsage to AOM_USAGE_ALL_INTRA,
-    // or by manually configuring the encoder so all frames will be key frames (if AOM_USAGE_ALL_INTRA isn't available).
+    // Determine whether the encoder should be configured to use intra frames only, either by setting aomUsage to
+    // AOM_USAGE_ALL_INTRA, or by manually configuring the encoder so all frames will be key frames (if AOM_USAGE_ALL_INTRA isn't
+    // available).
 
-    // All-intra encoding is beneficial when encoding a two-layer image item and the quality of the first layer is very low.
-    // Switching to all-intra encoding comes with the following benefits:
+    // For libaom versions older than 3.14.0, all-intra encoding is beneficial when encoding a two-layer image item and the
+    // quality of the first layer is very low. Switching to all-intra encoding comes with the following benefits:
     // - The first layer will be smaller than the second layer (which is often not the case with inter encoding)
-    // - Outputs have predictable file sizes: the sum of the first layer (quality <= 10) plus the second layer (quality set by the caller)
-    // - Because the first layer is very small, layered encoding overhead is also smaller and more stable (about 5-8% for quality 40 and 2-4% for quality 60)
-    // - Option of choosing tune IQ (which requires AOM_USAGE_ALL_INTRA)
-    avifBool useAllIntraForLayered = encoder->extraLayerCount == 1 &&
+    // - Outputs have predictable file sizes: the sum of the first layer (quality <= 10) plus the second layer (quality set by
+    //   the caller)
+    // - Because the first layer is very small, layered encoding overhead is also smaller and more stable (about 5-8% for quality
+    //   40 and 2-4% for quality 60)
+    // Note: libaom 3.14.0 introduces a mechanism to completely control each layer's QP, and extends tune IQ to inter-frame
+    // encoding modes (AOM_USAGE_GOOD_QUALITY and AOM_USAGE_REALTIME), so there's no need to use all-intra encoding for layered.
+
+    // aom_codec.h says: aom_codec_version() == (major<<16 | minor<<8 | patch)
+    // TODO(wtc): Update constant with the libaom version that contains the layered-encoding improvements
+    static const int aomVersion_3_14_0 = INT_MAX;
+    const int aomVersion = aom_codec_version();
+    avifBool useAllIntraForLayered = aomVersion < aomVersion_3_14_0 && encoder->extraLayerCount == 1 &&
                                      codec->internal->qualityFirstLayer <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD;
     // Also use all-intra encoding when encoding still images.
     avifBool useAllIntra = (addImageFlags & AVIF_ADD_IMAGE_FLAG_SINGLE) || useAllIntraForLayered;
@@ -713,9 +722,7 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         }
     }
 
-    // aom_codec.h says: aom_codec_version() == (major<<16 | minor<<8 | patch)
     static const int aomVersion_2_0_0 = (2 << 16);
-    const int aomVersion = aom_codec_version();
     if (aomVersion <= aomVersion_2_0_0) {
         // Issue with v1.0.0-errata1-avif: https://github.com/AOMediaCodec/libavif/issues/56
         // Issue with v2.0.0: https://aomedia-review.googlesource.com/q/I26a39791f820b4d4e1d63ff7141f594c3c7181f5
@@ -752,13 +759,18 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
                 // AOM_TUNE_IQ has been tuned for the YCbCr family of color spaces, and is favored for
                 // its low perceptual distortion. AOM_TUNE_IQ partially generalizes to, and benefits
                 // from other "YUV-like" spaces (e.g. YCgCo and ICtCp) including monochrome (luma only).
-                // AOM_TUNE_IQ sets --deltaq-mode=6 which can only be used in all intra mode.
+                //
                 // AOM_TUNE_IQ was introduced in libaom v3.12.0 but it has significantly different bit
                 // allocation characteristics compared to v3.13.0. AOM_TUNE_IQ is used by default
                 // starting with v3.13.0 for fewer behavior changes in libavif.
+                //
+                // Starting with libaom v3.14.0, AOM_TUNE_IQ supports all-intra, good-quality and
+                // realtime modes (for single and layered images). Prior to v3.14.0, AOM_TUNE_IQ is only
+                // supported in all-intra mode.
                 static const int aomVersion_3_13_0 = (3 << 16) | (13 << 8);
-                if (image->matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_IDENTITY && aomUsage == AOM_USAGE_ALL_INTRA &&
-                    aomVersion >= aomVersion_3_13_0) {
+                if (image->matrixCoefficients != AVIF_MATRIX_COEFFICIENTS_IDENTITY &&
+                    ((aomUsage == AOM_USAGE_ALL_INTRA && aomVersion >= aomVersion_3_13_0) ||
+                     (encoder->extraLayerCount > 0 && aomVersion >= aomVersion_3_14_0))) {
                     libavifDefaultTuneMetric = AOM_TUNE_IQ;
                 }
 #endif
@@ -906,6 +918,14 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
             // For layered image, disable lagged encoding to always get output
             // frame for each input frame.
             cfg->g_lag_in_frames = 0;
+
+            if (aomVersion >= aomVersion_3_14_0) {
+                // Disable QP offsets, so CQ level = frame QP for every frame.
+                // This feature requires libaom 3.14.0 or later.
+                if (cfg->rc_end_usage == AOM_Q) {
+                    cfg->use_fixed_qp_offsets = 2;
+                }
+            }
         }
         if (disableLaggedOutput) {
             cfg->g_lag_in_frames = 0;


src/codec_aom.c
juliobbv-p, line 917: FYI @wantehchang: the encoder already sets `g_limit` to the right value for layered images.

apps/avifenc.c
```diff
@@ -993,6 +993,8 @@ static avifBool avifEncodeRestOfLayeredImage(avifEncoder * encoder,
             // reversed lerp, so that last layer reaches exact targetQuality
             encoder->quality = targetQuality - (targetQuality - PROGRESSIVE_START_QUALITY) *
                                                    (encoder->extraLayerCount - layerIndex) / encoder->extraLayerCount;
+            // Don't perform any scaling for the second layer (numerator: 1, denominator: 1).
```
wantehchang, line 997: Add the following sentence before this comment:

```
            // We scaled the first layer by a half (numerator: 1, denominator: 2).
            // Don't perform any scaling for the second layer (numerator: 1, denominator: 1).
```

The reason this is useful is that this points out `encoder->scalingMode` isn't 1:1 at this point.

apps/avifenc.c
```diff
@@ -993,6 +993,8 @@ static avifBool avifEncodeRestOfLayeredImage(avifEncoder * encoder,
...
+            encoder->scalingMode = scalingModeSettingsEntryOf(1, 1).value;
```
wantehchang: Nit: Avoid calling the `scalingModeSettingsEntryOf()` function. (See patch.)

apps/avifenc.c
```diff
@@ -1160,8 +1162,10 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
         // we should not reach here.
         assert(encoder->quality >= PROGRESSIVE_WORST_QUALITY);
         // Encode the base layer with a very low quality to ensure a small encoded size.
-        encoder->quality = 2;
+        encoder->quality = 10;
```
wantehchang, line 1167: Please confirm that you intend to increase the quality of the base layer from 2 to 10. Is this change necessary because we changed from all-intra to good-quality mode?

Since libavif may use either all-intra or good-quality depending on the libaom version, please make sure quality 10 works for both modes.

Should we make this change in a separate PR now?

apps/avifenc.c
```diff
@@ -1160,8 +1162,10 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
...
         // Low alpha quality resulted in weird artifact, so we don't do it.
+        // For further size savings, scale the first layer by a half (numerator: 1, denominator: 2).
+        encoder->scalingMode = scalingModeSettingsEntryOf(1, 2).value;
```
wantehchang: Nit: Avoid calling the `scalingModeSettingsEntryOf()` function. (See patch.)

src/codec_aom.c
```diff
@@ -664,16 +666,21 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         codec->internal->qualityFirstLayer = quality;
     }
 
+    // aom_codec.h says: aom_codec_version() == (major<<16 | minor<<8 | patch)
+    static const int aomVersion_3_14_0 = (3 << 16) | (14 << 8);
+    const int aomVersion = aom_codec_version();
```
wantehchang, line 686: Nit: I moved these three lines closer to where they are used. (See patch.)

src/codec_aom.c
```diff
@@ -664,16 +666,21 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
-    // Switching to all-intra encoding comes with the following benefits:
+    // For libaom versions older than 3.14.0, all-intra encoding is beneficial when encoding a two-layer image item and the quality of the first layer is
+    // very low. Switching to all-intra encoding comes with the following benefits:
     // - The first layer will be smaller than the second layer (which is often not the case with inter encoding)
     // - Outputs have predictable file sizes: the sum of the first layer (quality <= 10) plus the second layer (quality set by the caller)
     // - Because the first layer is very small, layered encoding overhead is also smaller and more stable (about 5-8% for quality 40 and 2-4% for quality 60)
-    // - Option of choosing tune IQ (which requires AOM_USAGE_ALL_INTRA)
-    avifBool useAllIntraForLayered = encoder->extraLayerCount == 1 &&
+    // Note: libaom 3.14.0 introduces a mechanism to completely control each layer's QP, and extends tune IQ to inter-frame encoding modes
+    // (AOM_USAGE_GOOD_QUALITY and AOM_USAGE_REALTIME), so there's no need to use all-intra encoding for layered.
```
wantehchang: Nit: The .clang-format file has the following:

```
ColumnLimit:     130
```

but it doesn't seem to have an effect. I manually reformatted the comments here so that each line is shorter than 130 chars. (See patch.)

src/codec_aom.c
```diff
@@ -664,16 +666,21 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
         codec->internal->qualityFirstLayer = quality;
     }
 
+    // aom_codec.h says: aom_codec_version() == (major<<16 | minor<<8 | patch)
+    static const int aomVersion_3_14_0 = (3 << 16) | (14 << 8);
```
wantehchang: I have an idea that can allow us to merge this PR before libaom v3.14.0 is released. We can set `aomVersion_3_14_0` to `INT_MAX`. When libaom v3.14.0 is released, we change it back to the correct value.

juliobbv-p: Sounds good. Made change and added a TODO comment as discussed during our meeting.

apps/avifenc.c
juliobbv-p: The increase in quality from 2 to 10 is to make the first layer look a bit nicer, after factoring in that this layer is now scaled by a half (independently of the encoding mode). Given this info, it's best to keep this change here.

juliobbv-p:
@wantehchang thanks for reviewing this PR! I pushed new changes with your patch applied, and reacted to your feedback.

I'm planning to add the layer scaling bug guard check in a separate PR.

src/codec_aom.c
wantehchang: Thanks for checking. I also verified that yesterday.

src/codec_aom.c
```diff
@@ -664,16 +666,22 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
+    // TODO(wtc): Update constant with the libaom version that contains the layered-encoding improvements
+    static const int aomVersion_3_14_0 = INT_MAX;
+    const int aomVersion = aom_codec_version();
+
     // Determine whether the encoder should be configured to use intra frames only, either by setting aomUsage to AOM_USAGE_ALL_INTRA,
     // or by manually configuring the encoder so all frames will be key frames (if AOM_USAGE_ALL_INTRA isn't available).
 
-    // All-intra encoding is beneficial when encoding a two-layer image item and the quality of the first layer is very low.
-    // Switching to all-intra encoding comes with the following benefits:
+    // For libaom versions older than 3.14.0, all-intra encoding is beneficial when encoding a two-layer image item and the quality of the first layer is
```
wantehchang: Julio: Did you apply my patch? This comment block still have several lines longer than 130 characters.

wantehchang: To help me review this PR, I created a copy in Gerrit. The following diffs show the changes in this update:
https://aomedia-review.git.corp.google.com/c/libavif/+/209521/1..2

wantehchang: Actually GitHub also shows the changes:
https://github.com/AOMediaCodec/libavif/compare/f63f6b3cea8f64506241873525aa8ba962a0ec10..73679bb6f0c2f3c66ac8b9b54337bb148f940900

juliobbv-p: Looks like I pushed a version without your patch applied by mistake 🙁.

I'll take care of this and repush.

juliobbv-p: OK, pushed with your patch applied for real this time. Sorry about the confusion.

wantehchang: LGTM! Thanks.

src/codec_aom.c
```diff
@@ -904,6 +918,14 @@ static avifResult aomCodecEncodeImage(avifCodec * codec,
...
             // frame for each input frame.
             cfg->g_lag_in_frames = 0;
+
+            if (aomVersion >= aomVersion_3_14_0) {
+                // Disable QP offsets, so CQ level = frame QP for every frame.
+                // This feature requires libaom 3.14.0 or later.
+                if (cfg->rc_end_usage == AOM_Q) {
+                    cfg->use_fixed_qp_offsets = 2;
+                }
+            }
```
juliobbv-p, line 928: I had a meeting with @wantehchang to come up with a plan to best handle the two frame scaling bugs we recently found in libaom (targeting layered encoding scenarios):
1. In `src/codec_aom.c`: if libaom version is <= 3.13.2 and frame scaling is enabled, disable both comp mask prediction and restoration filtering with their respective codec controls.
2. (if 1 doesn't work) In `src/codec_aom.c`: if libaom version is <= 3.13.2 and frame scaling is enabled, disable the frame scaling feature. If the quality of the first layer is 10, re-adjust the quality to 2 (to compensate for the increased frame dimensions).

wantehchang: Julio: I found that the crash in loop restoration filtering is unrelated to frame scaling. It is, however, related to multithreading. We can avoid this crash by either setting the `AV1E_SET_ENABLE_RESTORATION` codec control to 0 or setting `g_threads` in `aom_codec_enc_cfg_t` to 1.

Setting the `AV1E_SET_ENABLE_MASKED_COMP` codec control to 0 unfortunately does not avoid the crash in `aom_comp_mask_upsampled_pred()`. By trial and error I found that setting the `AV1E_SET_ENABLE_INTERINTRA_WEDGE` codec control to 0 avoids the crash in your ResizeCrashTest.TestCompoundMaskPredictionCrash test, but I cannot show by code analysis that this works in general. So the best way to avoid this crash is to disable the frame scaling feature.

juliobbv-p: Thanks Wan-Teh for the investigation. What you said about loop restoration filter makes sense, given that the it was two parts of the code that disagreed on the number of actual LR workers during multithreading.

Now that we don't know for certain we can't avoid the crash setting in `aom_comp_mask_upsampled_pred()`, I agree that disabling the frame scaling feature in libavif is the way to go.

tongyuantongyu: I'd like to mention that without `cfg->use_fixed_qp_offsets = 2`, the outcome of setting quality is often unpredictable or at least confusing, so maybe we should force all intra usage for libaom < 3.14.

Not sure if the loop restoration filtering crash still happens with all intra usage, but this should avoid the `aom_comp_mask_upsampled_pred()` crash at least.

wantehchang: The loop restoration filtering crash also occurs with all intra usage. I just wrote down more info on the loop restoration filtering crash in the [libaom CL that fixed the crash](https://aomedia-review.git.corp.google.com/c/aom/+/208901/6#message-dd0b573bae2307c9627bead3eb2fb94b8ce62588). I found that this crash affects libaom versions as old as v3.2.0, but for some reason we have not received reports of this crash. Since the crash is unrelated to the encoding of layered images, I don't think libavif needs to avoid it by turnning off loop restoration or multithreading.

---

Yuan: I believe we already do what you suggested:

```
    avifBool useAllIntraForLayered = aomVersion < aomVersion_3_14_0 && encoder->extraLayerCount == 1 &&
                                     codec->internal->qualityFirstLayer <= TWO_LAYER_ALL_INTRA_QUALITY_THRESHOLD;
```

This already checks libaom < 3.14. Or are you suggesting we broaden the condition to the following?

```
    avifBool useAllIntraForLayered = aomVersion < aomVersion_3_14_0 && encoder->extraLayerCount > 0;
```

wantehchang: Julio: I just realized that we are already protected from the [frame scaling crash in `aom_comp_mask_upsampled_pred()`](https://aomedia-review.git.corp.google.com/c/aom/+/208801).

When we use `avifenc --progressive` to encode a two-layered image with the first layer downscaled by 1/2:
- For libaom < 3.14.0, we will use all-intra mode. Since `aom_comp_mask_upsampled_pred()` is not called in all-intra mode, we are protected from this crash.
- For libaom >= 3.14.0, we will use good-quality or real-time mode (depending on the encoder speed). This crash is fixed in libaom >= 3.14.0.

Is my anaysis correct?

Note: I just released [libaom v3.13.3](https://groups.google.com/a/aomedia.org/g/av1-discuss/c/HMmF0gxEqmQ) yesterday with your two crash fixes. But it doesn't affect my analysis above.

tongyuantongyu: > Or are you suggesting we broaden the condition to the following?

Exactly. I was confused by the unpredictable output size when I was implementing layered encoding, and [it turns out I'm not alone](https://jakearchibald.com/2025/present-and-future-of-progressive-image-rendering/#:~:text=Supposedly%2C%20additional%20layers%20are%20possible%2C%20allowing%20for%20extra%20steps%20in%20the%20progressive%20render%2C%20but%20whenever%20I%20tried%20this%2C%20the%20encoder%20would%20error%20out%2C%20or%20explode%20the%20file%20size%20to%20~400%20kB%2C%20even%20at%20lowest%20quality.%20I%20guess%20that%27s%20why%20it%27s%20marked%20%27experimental%27.). Enforcing all intra at least provides predictability.

But anyway, based on the current situation, I would strongly recommend any user to encode layered image with aom >= 3.14. I think we should also add a warning in avifenc when encoding layered image with aom < 3.14.

wantehchang: Julio: I can now characterize the condition under which the [multithreading loop restoration crash](https://aomedia-review.git.corp.google.com/c/aom/+/208901) occurs: the number of worker threads for loop restoration needs to decrease. For a fixed frame size, this can happen if we
- decrease the number of tiles, or
- switch from AV1E_SET_ROW_MT=1 to AV1E_SET_ROW_MT=0.

Note: The number of worker threads for loop restoration is calculated by the libaom function [`compute_num_lr_workers()`](https://aomedia-review.git.corp.google.com/c/aom/+/208901/6#message-db07386e34706233456a9db7f970600b69b7f1a9), which does not consider the scaling factor. This explains why this crash is unrelated to frame scaling.

When we use `avifenc --progressive` to encode a two-layered image, whether we downscale the first layer by 1/2 or not, the number of worker threads for loop restoration is the same for the two layers because the number of tiles and the row-MT mode won't change.

Therefore we don't need to defend against this crash for `avifenc --progressive`.

juliobbv-p: Hi @wantehchang,

First of all, thanks for the analysis!

> When we use avifenc --progressive to encode a two-layered image, whether we downscale the first layer by 1/2 or not, the number of worker threads for loop restoration is the same for the two layers because the number of tiles and the row-MT mode won't change.

BTW, this isn't what I saw in practice. When I configured `--progressive` to downscale the first layer by a 1/2, I could see the number of **reported** workers change on my machine (as reported by `av1_get_num_mod_workers_for_alloc()`) in between layers, which triggered the crash that was fixed by https://aomedia-review.googlesource.com/c/aom/+/209681.

> For libaom < 3.14.0, we will use all-intra mode. Since aom_comp_mask_upsampled_pred() is not called in all-intra mode, we are protected from this crash.
> For libaom >= 3.14.0, we will use good-quality or real-time mode (depending on the encoder speed). This crash is fixed in libaom >= 3.14.0.
> Is my anaysis correct?

If we're using all-intra mode for libaom < 3.14.0, (which protects us from the crash in `[aom_comp_mask_upsampled_pred()](https://aomedia-review.git.corp.google.com/c/aom/+/208801)`, this means that we can disable loop restoration with the corresponding codec control and be fully protected from both crashes. This particular version check could also be tightened up to libaom < 3.13.3, given that it contains the LR multithreading fix.

juliobbv-p: BTW, here's an image that crashes avifenc if you encode with a version of libaom that doesn't have the LR fix:

US_Navy_091029-M-5645B-091_A_landing_craft,_air_cushion_(LCAC)_and_the_U S _Coast_Guard_patrol_boat_Albacore_are_underway_near_Naval_Station_Norfolk

`./avifenc -s 0 --progressive -c aom -q 10 -y 444 "original/US_Navy_091029-M-5645B-091_A_landing_craft,_air_cushion_(LCAC)_and_the_U.S._Coast_Guard_patrol_boat_Albacore_are_underway_near_Naval_Station_Norfolk.png" -o usnavy.avif"
`

wantehchang: Julio: Thank you very much for the steps to reproduce the bug. I can reproduce the crash now. I was mistaken. The libaom function `compute_num_lr_workers()` does consider the scaling factor when the `avifenc --progressive` command executes. I suspect that libaom ignores the scaling factors in libaom's ResizeCrashTest.TestRestorationFilterCrash test because that test does not use spatial layers. (I haven't verified this.) Another libaom bug I found is that `compute_num_lr_workers()` is called before the current frame size is set, so the number of worker threads for loop restoration is calculated based on the previous frame size. This is why the number of worker threads decreases when the `avifenc --progressive` command executes. Without this bug, the number of worker threads should increase because the frame size increases.

To avoid this crash, we can disable either loop restoration or multithreading. It is also possible to avoid this crash by setting `cfg.g_threads` to a small enough number that depends on the frame size, but that requires duplicating the logic of the libaom function `compute_num_lr_workers()` and is not worth the effort.

wantehchang: Julio: Please see pull request https://github.com/AOMediaCodec/libavif/pull/3170.

PR #3072, opened 2026-03-01
Use size_t arithmetic for pixel offset calculations in reformat.c

uwezkhan:
This change promotes several pixel offset multiplications in avifImageRGBToYUV and avifImageYUVAnyToRGBAnySlow to size_t before pointer arithmetic.

Currently, expressions such as:

(i * rgbPixelBytes)

are evaluated using 32-bit arithmetic before being promoted to size_t. While current AV1 frame dimension limits prevent overflow under normal conditions, performing these calculations directly in size_t ensures correctness and avoids reliance on implicit width constraints.

PR #3073, opened 2026-03-01
Promote pixel offset multiplications to size_t in reformat.c

uwezkhan:
This patch improves the safety of pixel offset calculations in src/reformat.c by ensuring that all byte offset multiplications are performed using size_t arithmetic.

In several places inside avifImageRGBToYUV and avifImageYUVAnyToRGBAnySlow, expressions such as:

i * rgbPixelBytes
uvI * yuvChannelBytes
i * 2

were evaluated using 32-bit integer arithmetic before being implicitly promoted to size_t for pointer arithmetic. Although current AV1 frame dimension limits prevent these expressions from overflowing in practice, relying on 32-bit intermediate arithmetic introduces unnecessary fragility.

This change explicitly casts loop indices to size_t before multiplication, for example:

(size_t)i * rgbPixelBytes
(size_t)uvI * yuvChannelBytes

This guarantees that offset computations are performed in the correct width type prior to pointer arithmetic. The update does not modify any control flow, logic, or behavior. It simply ensures arithmetic correctness and strengthens robustness in code that processes untrusted image data.

A total of 36 offset calculations were updated. No functional changes are expected.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -307,24 +307,24 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
 
                         // Unpack RGB into normalized float
                         if (state.rgb.channelBytes > 1) {
-                            rgbPixel[0] = *((uint16_t *)(&rgb->pixels[offsetBytesR + (i * rgbPixelBytes) + (j * rgbRowBytes)])) /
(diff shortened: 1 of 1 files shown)


vrabaud:
This is a duplicate of https://github.com/AOMediaCodec/libavif/pull/3072 ? Please close if so

PR #3074, opened 2026-03-02
Bump libpng to 1.6.55

vrabaud:
Bump libpng to 1.6.55

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Update rav1e.cmd/LocalRav1e.cmake: cargo-c v0.10.20, corrosion v0.6.1,
   rav1e v0.8.1
 * Update svt.cmd/svt.sh/LocalSvt.cmake: v4.0.1
-* Update zlibpng.cmd/LocalZlibpng.cmake: libpng 1.6.51, zlib 1.3.2
+* Update zlibpng.cmd/LocalZlibpng.cmake: libpng 1.6.55, zlib 1.3.2
 * Fix grayscale conversion when changing the bit depth.
 * Bump cmake_minimum_required from 3.13 to 3.22
 * Associate transformative properties with alpha auxiliary image items.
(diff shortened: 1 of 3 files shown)


cmake/Modules/Findrav1e.cmake
```diff
@@ -41,7 +41,11 @@ if(NOT RAV1E_LIBRARY)
     find_library(RAV1E_LIBRARY NAMES rav1e PATHS ${_RAV1E_LIBDIR})
 endif()
 
-set(RAV1E_LIBRARIES ${RAV1E_LIBRARIES} ${RAV1E_LIBRARY} ${_RAV1E_LDFLAGS})
+if(MSVC)
```
maryla-uc: If this related to the rest of the change?

vrabaud: Yes it is: I am trying to fix the fact that bumping libpng (which has no CMake changes !) actually messes up the linker flags by adding some gcc flags (e.g.  -lkernel32 -lntdll -luserenv -lws2_32 -ldbghelp -lntdll -luserenv -lws2_32 -lbcrypt) while using the MSVC linker ....

wantehchang:
Vincent: Please see my test PR https://github.com/AOMediaCodec/libavif/pull/3081.

The second commit in PR https://github.com/AOMediaCodec/libavif/pull/3081 patches v1.6.55 to behave like v1.6.52. libavif will build successfully and the tests pass, but the combined static library test will fail because libpng static library is built with the .a file extension.

The third commit in PR https://github.com/AOMediaCodec/libavif/pull/3081 works.

Obviously PR https://github.com/AOMediaCodec/libavif/pull/3081 is not ready for check-in. But it shows the assembler program is related to the build failures.

vrabaud:
@wantehchang , the bug was actually due to https://github.com/pnggroup/libpng/commit/ff615fdcaf393ec2402bfafebadaa30b32fe5f3a

Enabling ASM on MSVC while it does not support it, creates some weird mix of Unix linked flags. What is actually missing is NASM (as recommended by SVT-AV1) when SVT-AV1 is built locally. I am trying one last hack and we can then split this PR.

vrabaud:
Just updated the commit description

Issue #3070, opened 2026-02-28
Should avifpng.c also call avifSetExifOrientation(&exif, 1) after calling avifApplyTransforms()?


wantehchang:
Note: I originally reported this issue in https://github.com/AOMediaCodec/libavif/pull/3059#pullrequestreview-3869959393

PR https://github.com/AOMediaCodec/libavif/pull/3059 added an `avifSetExifOrientation(&exif, 1)` call to avifjpeg.c (at line 1730) but but did not add it to avifpng.c. I can observe a difference when decoding the attached happy_dog.avif file to a JPEG output versus a PNG output. The following patch (to illustrate the bug, not for check-in) fixes the difference:

```
diff --git a/apps/shared/avifpng.c b/apps/shared/avifpng.c
index 3e6b034f..826b4f2b 100644
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -783,6 +783,7 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
             fprintf(stderr, "Error writing PNG: Exif metadata is too big\n");
             goto cleanup;
         }
+        avifResult result = avifSetExifOrientation((avifRWData *)&avif->exif, 1);
         png_set_eXIf_1(png, info, (png_uint_32)avif->exif.size, avif->exif.data);
     }
     if (avif->xmp.data && (avif->xmp.size > 0)) {
```

[happy_dog.avif.zip](https://github.com/user-attachments/files/25618197/happy_dog.avif.zip)

maryla-uc:
The difference exists because the PNG specification third edition states that "It is recommended that unless a decoder has independent knowledge of the validity of the Exif data, the data should be considered to be of historical value only."
So PNG decoders are supposed to ignore the exif data, including orientation (at least that's our interpretation), therefore we do not set it, and just keep the old "historical value". 

Buuuut you are right, it looks like Chrome and a bunch of other places are actually honoring the Exif orientation, leading to a rotated image. Apparently Firefox used to ignore it but now supports it according to https://bugzilla.mozilla.org/show_bug.cgi?id=1627423 (have not tested). Also Preview on Mac uses it but NOT the Finder thumbnail.
Some software does ignore it. e.g. https://www.photopea.com/ or a bunch of viewers on my Linux machine (eye of gnome, feh, Gwenview). It might be intentional, or it might just be because the eXIf chunk is new if version 3 (published in June 2025).

[...]

PR #3075, opened 2026-03-02
Set PNG Exif orientation to 1 (no rotation, no mirror) when saving.

maryla-uc:
#3070

### DIFF
--- a/apps/shared/avifpng.c
+++ b/apps/shared/avifpng.c
@@ -785,7 +785,7 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
             fprintf(stderr, "Error writing PNG: Exif metadata is too big\n");
             goto cleanup;
         }
-        // Make a copy of the Exif data to avoid modifying the input image when setting rotation.
+        // Make a copy of the Exif data to avoid modifying the input image when setting orientation.
         avifRWData exif = { NULL, 0 };
         if (avifRWDataRealloc(&exif, avif->exif.size) != AVIF_RESULT_OK) {
             fprintf(stderr, "Error writing PNG metadata: out of memory\n");


apps/shared/avifpng.c
```diff
@@ -138,6 +138,8 @@ static avifBool avifExtractExifAndXMP(png_structp png, png_infop info, avifBool
             // It is easier to set it to 1 (the default top-left) than actually removing the tag.
             // libheif has the same behavior, see
             // https://github.com/strukturag/libheif/blob/18291ddebc23c924440a8a3c9a7267fe3beb5901/examples/heif_enc.cc#L703
+            // However note that some software does use Exif orientation, e.g. Chrome, see also
```
wantehchang, line 141: The Chrome example is incorrect because Chrome uses Exif orientation in PNG files.

apps/shared/avifpng.c
```diff
@@ -783,7 +785,28 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
             fprintf(stderr, "Error writing PNG: Exif metadata is too big\n");
             goto cleanup;
         }
-        png_set_eXIf_1(png, info, (png_uint_32)avif->exif.size, avif->exif.data);
+        // Make a copy of the Exif data to avoid modifying the input image when setting rotation.
```
wantehchang: Nit: rotation => orientation

The orientation could indicate not only rotation but also mirroring.

apps/shared/avifpng.c
```diff
@@ -783,7 +785,28 @@ avifBool avifPNGWrite(const char * outputFilename, const avifImage * avif, uint3
...
+            goto cleanup;
+        }
+        memcpy(exif.data, avif->exif.data, avif->exif.size);
+        // We already rotated the pixels if necessary in avifApplyTransforms(), so we set the orientation to 1 (no rotation, no mirror).
+        avifResult result = avifSetExifOrientation(&exif, 1);
+        if (result != AVIF_RESULT_OK) {
+            if (result == AVIF_RESULT_INVALID_EXIF_PAYLOAD || result == AVIF_RESULT_NOT_IMPLEMENTED) {
+                // Either the Exif is invalid, or it doesn't have an orientation field.
+                // If it's invalid, we can consider it as equivalent to not having an orientation.
+                // In both cases, we can ignore the error.
```
wantehchang, line 801: Please confirm that `avifSetExifOrientation()` leaves the `exif` buffer unchanged when it fails.

apps/shared/avifpng.c
vrabaud: It seems it sticks to the spec and does not take exit into account: https://issues.chromium.org/issues/40125224

apps/shared/avifpng.c
maryla-uc: Yes it's unchanged.

apps/shared/avifpng.c
maryla-uc: Wan-Teh I don't understand your comment, the code comment says that Chrome does use orientation.
Vincent, I think this is for the case where Exif appears after the image data.
In any case, if you try it, you can see that otherwise Chrome does use the orientation.

wantehchang: Maryla: I am so sorry. Somehow I thought the comment said "some software does not use Exif orientation".

wantehchang (approved):
LGTM. Thanks!

Issue #3078, opened 2026-03-02
Should avifJPEGRead() set the orientation field to 1 in the Exif metadata copied from the JPEG input file?


wantehchang:
@vrabaud @y-guyon 

In apps/shared/avifjpeg.c, the `avifJPEGReadInternal()` function has the following code:

```
                // Exif orientation, if any, is imported to avif->irot/imir and kept in avif->exif.
                // libheif has the same behavior, see
                // https://github.com/strukturag/libheif/blob/ea78603d8e47096606813d221725621306789ff2/examples/heif_enc.cc#L403
                if (avifImageSetMetadataExif(avif,
                                             marker->data + AVIF_JPEG_EXIF_HEADER_LENGTH,
                                             marker->data_length - AVIF_JPEG_EXIF_HEADER_LENGTH) != AVIF_RESULT_OK) {
                    fprintf(stderr, "Setting Exif metadata failed: %s (out of memory)\n", inputFilename);
                    goto cleanup;
                }
```

This means orientation will be present in both `avif->irot/imir` and `avif->exif`. The rationale is not clear; the comment "libheif has the same behavior" suggests that we wanted to match libheif's behavior. Yannis, is that why?

It seems better to set the orientation to 1 in `avif->exif`. If we have to preserve the Exif orientation in `avif->exif`, I suggest adding the rationale to the comment.

joojoooo:
Some image viewers (like immich web) still use exif orientation in avif images to display it, resulting in weird stretched images when using the new encoder to convert from jpg

maryla-uc:
I think this should fundamentally be fixed in the viewer, and I see you did file https://github.com/immich-app/immich/issues/27771

We could revert the change but that would be encouraging non compliant viewers.

maryla-uc:
By the way, does immich display the image correctly if the Orientation tag (or the whole Exif metadata) is removed?
We could remove Orientation but it's a bit more tricky to implement, and if immich does not correctly take into account the irot/imir properties anyway then it won't change anything.

joojoooo:
> does immich display the image correctly if the Orientation tag (or the whole Exif metadata) is removed?

Unfortunately no, it's completely ignoring irot/imir

> I think this should fundamentally be fixed in the viewer

I totally agree. Immich also has the same issue for HEIF photos took by cameras, so they must fix the viewer

joojoooo:
> We could remove Orientation

[...]

PR #3079, opened 2026-03-02
Remove Exif orientation after saving it to irot/imir when reading JPEGs.

maryla-uc:
Currently we read the Exif orientation tag and translate it to irot/imir, but also leave the Exif as is.
However, MIAF says that AVIF files should not contain Exif transformations.

#3078

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Add --sato flag to avifdec to enable Sample Transforms support at decoding.
 * Add --grid option to avifgainmaputil.
 * Apply clean aperture crop, rotation and mirror when decoding to PNG or JPEG.
+  Remove orientation information from Exif if present.
 
 ### Changed since 1.3.0
 
@@ -62,7 +63,6 @@ The changes are relative to the previous release, unless the baseline is specifi
   now also splits the gain map into a grid.
 * In avifenc, set Exif orientation to 1 (no transformation) when converting
   JPEGs to Avif.
-* In avifedc, set Exif orientation to 1 when writing PNGs.
 
 ### Removed since 1.3.0
 


apps/shared/avifjpeg.c
```diff
@@ -1394,7 +1394,7 @@ static avifBool avifJPEGReadInternal(FILE * f,
                     goto cleanup;
                 }
 
-                // Exif orientation, if any, is imported to avif->irot/imir and kept in avif->exif.
+                // Exif orientation, if any, is imported to avif->irot/imir, and the Exif data is saved to avif->exif.
```
wantehchang, line 1397: Should we delete the next two lines?

```
                // libheif has the same behavior, see
                // https://github.com/strukturag/libheif/blob/ea78603d8e47096606813d221725621306789ff2/examples/heif_enc.cc#L403
```

apps/shared/avifutil.c
```diff
@@ -622,7 +622,7 @@ avifBool avifImageSplitGrid(const avifImage * gridSplitImage, uint32_t gridCols,
         }
     }
     if (gridSplitImage->exif.size > 0) {
-        const avifResult result = avifImageSetMetadataExif(firstCell, gridSplitImage->exif.data, gridSplitImage->exif.size);
+        const avifResult result = avifRWDataSet(&firstCell->exif, gridSplitImage->exif.data, gridSplitImage->exif.size);
```
wantehchang, line 625: IMPORTANT: Is this a bug fix? If so, this should be mentioned in the commit message.

apps/shared/avifjpeg.c
```diff
@@ -1403,6 +1403,15 @@ static avifBool avifJPEGReadInternal(FILE * f,
...
                 }
+                // Set the Exif orientation to 1 (no transformation).
+                // ISO/IEC 23000-22:2024 (MIAF), Section 7.3.10.1:
+                //   There should be no image transformations expressed by Exif (rotation,
+                //   mirroring, etc.) indicated in the Exif metadata, in files encoded according
+                //   to this document.
+                // Do not check for errors, it's a "should" so ok to do on a best-effort basis.
+                // Moreover it should only fail if the Exif is marlformed or there is no orientation
+                // tag to begin with.
+                (void)avifSetExifOrientation(&avif->exif, 1);
```
wantehchang, line 1412: Do you think this is worth mentioning in CHANGELOG.md?

apps/shared/avifutil.c
maryla-uc: It's not a bugfix. It was working before, but it's a necessary change with this PR. Since we remove the orientation from exif, when later calling avifImageSetMetadataExif() it will use the modified exif and lose the actual orientation.
The call to avifImageSetMetadataExif was not really a bug but was not necessary either, calling avifRWDataSet is enough.
This was caught by one of the golden tests.

apps/shared/avifjpeg.c
maryla-uc: Done. I also added a line for my previous PR #3075

CHANGELOG.md
```diff
@@ -60,6 +60,9 @@ The changes are relative to the previous release, unless the baseline is specifi
   libaom v3.13.0 or later.
 * Converting an image containing a gain map using avifenc with the --grid flag
   now also splits the gain map into a grid.
+* In avifenc, set Exif orientation to 1 (no transformation) when converting
+  JPEGs to Avif.
+* In avifedc, set Exif orientation to 1 when writing PNGs.
```
wantehchang: I suggest we merge this item with the existing item at line 29:

> * Apply clean aperture crop, rotation and mirror when decoding to PNG or JPEG.

apps/shared/avifutil.c
wantehchang: Thanks for the clarification. I thought this was an independent bug fix and was going to suggest you move it to a separate PR. I will need to study this change.

apps/shared/avifjpeg.c
```diff
@@ -1394,15 +1394,22 @@ static avifBool avifJPEGReadInternal(FILE * f,
                     goto cleanup;
                 }
 
-                // Exif orientation, if any, is imported to avif->irot/imir and kept in avif->exif.
-                // libheif has the same behavior, see
-                // https://github.com/strukturag/libheif/blob/ea78603d8e47096606813d221725621306789ff2/examples/heif_enc.cc#L403
```
wantehchang, line 1399: I checked the current libheif code. This code has been moved to another file (heifio/decoder_jpeg.cc) but is still the same.

I also found the following TODO comment in examples/heif_enc.cc:

```
    // write EXIF to HEIC
    if (!input_image.exif.empty()) {
      // Note: we do not modify the EXIF Orientation here because we want it to match the HEIF transforms.
      // TODO: is this a good choice? Or should we set it to 1 (normal) so that other, faulty software will not transform it once more?

      error = heif_context_add_exif_metadata(context, handle,
                                             input_image.exif.data(), (int) input_image.exif.size());
      if (error.code != 0) {
        heif_nclx_color_profile_free(nclx);
        std::cerr << "Could not write EXIF metadata: " << error.message << "\n";
        return 1;
      }
    }
```

tests/gtest/avifmetadatatest.cc
```diff
@@ -246,11 +246,21 @@ TEST(MetadataTest, ExifOrientation) {
...
   EXPECT_EQ(image->irot.angle, 1u);
   EXPECT_EQ(image->imir.axis, 0u);
+  testutil::AvifRwData originalReadExif;
+  ASSERT_EQ(
+      avifRWDataSet(&originalReadExif, image->exif.data, image->exif.size),
+      AVIF_RESULT_OK);
+  // For testing purposes, set back the orientation.
+  ASSERT_EQ(avifSetExifOrientation(&image->exif, 5), AVIF_RESULT_OK);
+  // The two Exifs are different, showing that the Orientation was indeed
+  // modifiedwhen reading.
```
y-guyon: Missing space

tests/gtest/avifmetadatatest.cc
```diff
@@ -313,6 +323,15 @@ TEST(MetadataTest, ExifOrientationAndForcedImir) {
...
   image->transformFlags = AVIF_TRANSFORM_IMIR;
   image->imir.axis = 1;
+  testutil::AvifRwData originalReadExif;
+  ASSERT_EQ(
+      avifRWDataSet(&originalReadExif, image->exif.data, image->exif.size),
+      AVIF_RESULT_OK);
+  // For testing purposes, set back the orientation.
+  ASSERT_EQ(avifSetExifOrientation(&image->exif, 5), AVIF_RESULT_OK);
+  // The two Exifs are different, showing that the Orientation was indeed
+  // modifiedwhen reading.
```
y-guyon: Missing space

apps/shared/avifutil.c
wantehchang: Maryla: I understand this change now. So we could make this change as an independent change, before the rest of the PR, because the original `avifImageSetMetadataExif()` call extracts the Exif orientation to irot/imir redundantly. Did I understand it correctly?

maryla-uc: Yes this is correct.

PR #3083, opened 2026-03-03
SVT-AV1: Enable NASM on x86 only, ASM on ARM only.

vrabaud:
Also force CMP0194 to make sure MSVC is not chosen for NASM or ASM.

This is what upstream is doing: https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/CMakeLists.txt?ref_type=heads#L68


CMakeLists.txt
```diff
@@ -493,6 +499,16 @@ if(AVIF_CODEC_RAV1E_ENABLED)
     set(AVIF_PKG_CONFIG_EXTRA_REQUIRES_PRIVATE "${AVIF_PKG_CONFIG_EXTRA_REQUIRES_PRIVATE} rav1e")
 endif()
 
+if(AVIF_CODEC_AVM)
+    # Avoid the following error:
+    #   CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
+    #   Missing variable is: CMAKE_ASM_COMPILE_OBJECT
+    # For some reason this must be done before check_avif_option(AVIF_CODEC_AOM TARGET aom PKG_NAME aom).
+    # See also https://gitlab.kitware.com/cmake/cmake/-/issues/25042.
+    # Enable ASM before NASM in SVT-AV1.
```
wantehchang, line 508: Just wanted to confirm that this needs to be moved here because of the change to cmake/Modules/LocalSvt.cmake? This is not related to the CMP0194 policy change, right?

vrabaud: Exactly: as ASM needs to be enabled before NASM, and now that SVT can enable NASM only, if AVM enables ASM, it has to be done before SVT enables NASM

PR #3086, opened 2026-03-05
Unify installing packages in the CI

vrabaud:
Also:
- test more OSes with installed rav1e and svt-av1
- clean rav1e needed libraries in CMake


cmake/Modules/Findrav1e.cmake
```diff
@@ -61,9 +61,8 @@ if(RAV1E_LIBRARY)
             rav1e::rav1e PROPERTIES IMPORTED_LOCATION "${RAV1E_LIBRARY}" IMPORTED_IMPLIB "${RAV1E_LIBRARY}" IMPORTED_SONAME rav1e
         )
         target_include_directories(rav1e::rav1e INTERFACE ${RAV1E_INCLUDE_DIR})
-        # The following is copied from the main CMakeLists.txt.
         if(WIN32)
-            target_link_libraries(rav1e::rav1e INTERFACE ntdll.lib userenv.lib ws2_32.lib bcrypt.lib)
+            target_link_libraries(rav1e::rav1e INTERFACE ntdll userenv ws2_32 bcrypt)
```
wantehchang, line 65: The removal of the .lib filename suffix is a good change. CMake documentation recommends this.

cmake/Modules/Findrav1e.cmake
```diff
@@ -61,9 +61,8 @@ if(RAV1E_LIBRARY)
...
         elseif(UNIX AND NOT APPLE)
             target_link_libraries(rav1e::rav1e INTERFACE ${CMAKE_DL_LIBS}) # for backtrace
```
wantehchang, line 67: Does anyone know what the comment "for backtrace" mean?

cmake/Modules/LocalRav1e.cmake
```diff
@@ -105,6 +105,9 @@ else()
     add_library(rav1e::rav1e STATIC IMPORTED)
     add_dependencies(rav1e::rav1e rav1e)
     target_link_libraries(rav1e::rav1e INTERFACE "${Rust_CARGO_TARGET_LINK_NATIVE_LIBS}")
+    if(UNIX AND NOT APPLE)
+        target_link_libraries(rav1e::rav1e INTERFACE ${CMAKE_DL_LIBS}) # for backtrace
+    endif()
```
wantehchang, line 110: Just curious: Why is this needed in this file, and why don't we need the WIN32 "ntdll userenv ws2_32 bcrypt" libs in this file?

vrabaud: Those are actually included in Rust_CARGO_TARGET_LINK_NATIVE_LIBS

cmake/Modules/Findrav1e.cmake
vrabaud: Nope, I just copy-pasted. Maybe that's also included in Rust_CARGO_TARGET_LINK_NATIVE_LIBS. Let's try.

wantehchang: @joedrago 

I tracked this down. Joe added this line in commit https://github.com/AOMediaCodec/libavif/commit/0b6e7a7b4:

```
        set(AVIF_CODEC_LIBARIES ${AVIF_CODEC_LIBARIES} dl) # for backtrace
```

and then `dl` (for the `-ldl` linker flag) was replaced with `${CMAKE_DL_LIBS}` in commit https://github.com/AOMediaCodec/libavif/commit/d5643554f:

```
        set(AVIF_PLATFORM_LIBRARIES ${AVIF_PLATFORM_LIBRARIES} ${CMAKE_DL_LIBS}) # for backtrace
```

Joe, do you still remember what the comment "for backtrace" means? Is "backtrace" a feature of rav1e? I found something like that here:

https://www.rpmfind.net/linux/RPM/epel/9/aarch64/Packages/r/rust-rav1e+backtrace-devel-0.8.1-2.el9.noarch.html

and a Rust crate called "backtrace":

https://docs.rs/backtrace/latest/backtrace/

cmake/Modules/LocalRav1e.cmake
joedrago: Heh, it's been a minute, but based on that commit, I'm sure that version of rav1e failed to link on Linux without libdl. If it no longer depends on it and you can link against rav1e on Linux without it, feel free to remove it.

PR #3089, opened 2026-03-06
Set CMP0169 to NEW

vrabaud:
FetchContent_Populate(${name}) cannot be used anymore. Now use FetchContent_MakeAvailable and unify how _SOURCE_DIR and _BINARY_DIR are set.

### DIFF
--- a/cmake/Modules/AvifExternalProjectUtils.cmake
+++ b/cmake/Modules/AvifExternalProjectUtils.cmake
@@ -1,4 +1,4 @@
-function(avif_fetchcontent_populate_cmake name)
+function(avif_fetchcontent_makeavailable_cmake name)
     FetchContent_GetProperties(${name})
     if(NOT ${name}_POPULATED)
         # Force static build
--- a/cmake/Modules/LocalAom.cmake
+++ b/cmake/Modules/LocalAom.cmake
@@ -68,7 +68,10 @@ else()
     endfunction()
 
     FetchContent_Declare(
-        libaom URL "https://aomedia.googlesource.com/aom/+archive/${AVIF_AOM_GIT_TAG}.tar.gz" UPDATE_COMMAND "" EXCLUDE_FROM_ALL
+        libaom
+        EXCLUDE_FROM_ALL
+        URL "https://aomedia.googlesource.com/aom/+archive/${AVIF_AOM_GIT_TAG}.tar.gz"
+        UPDATE_COMMAND ""
     )
 
     if(NOT AVIF_CODEC_AOM_DECODE)
@@ -102,7 +105,7 @@ else()
             endforeach()
         endforeach()
 
-        avif_fetchcontent_populate_cmake(libaom)
+        avif_fetchcontent_makeavailable_cmake(libaom)
 
         set(_aom_config RELEASE)
         if(CMAKE_BUILD_TYPE)
--- a/cmake/Modules/LocalAvm.cmake
+++ b/cmake/Modules/LocalAvm.cmake
@@ -27,6 +27,7 @@ endfunction()
 
 FetchContent_Declare(
     libavm
+    EXCLUDE_FROM_ALL
     GIT_REPOSITORY "https://gitlab.com/AOMediaCodec/avm.git"
     GIT_TAG ${AVIF_AVM_GIT_TAG}
     GIT_PROGRESS ON
@@ -44,7 +45,7 @@ FetchContent_Declare(
     #   git diff > LocalAvm.diff
     # TODO: b/398931194 - Remove the patch when using a libavm version past
     #                     https://gitlab.com/AOMediaCodec/avm/-/merge_requests/2985
-    PATCH_COMMAND git apply ${AVIF_SOURCE_DIR}/cmake/Modules/LocalAvm.diff EXCLUDE_FROM_ALL
+    PATCH_COMMAND git apply ${AVIF_SOURCE_DIR}/cmake/Modules/LocalAvm.diff
 )
 # There can be a duplicate cpuinfo in SVT so find_package has to be used.
 set(RUY_FIND_CPUINFO ON CACHE INTERNAL "")
@@ -76,7 +77,7 @@ if(NOT libavm_POPULATED)
         endforeach()
     endforeach()
 
-    avif_fetchcontent_populate_cmake(libavm)
+    avif_fetchcontent_makeavailable_cmake(libavm)
 
     set(_avm_config RELEASE)
     if(CMAKE_BUILD_TYPE)
--- a/cmake/Modules/LocalFuzztest.cmake
+++ b/cmake/Modules/LocalFuzztest.cmake
@@ -13,6 +13,7 @@ else()
 
     FetchContent_Declare(
         fuzztest
+        EXCLUDE_FROM_ALL
         GIT_REPOSITORY "https://github.com/google/fuzztest.git"
         GIT_TAG "${AVIF_FUZZTEST_TAG}"
         # Fixes for https://github.com/google/fuzztest/issues/1124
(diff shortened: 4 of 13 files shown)


CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ endif()
 # New in CMake version 3.30. FetchContent_Populate(<name>) is deprecated, call
 # FetchContent_MakeAvailable(<name>) instead.
 if(POLICY CMP0169)
-    cmake_policy(SET CMP0169 OLD)
+    cmake_policy(SET CMP0169 NEW)
```
wantehchang, line 25: If NEW is the default for CMP0169, we should delete this block of code (lines 22-26).  I guess NEW is still not the default?

vrabaud: Not the default, it's like the other policies. cf https://cmake.org/cmake/help/latest/policy/CMP0169.html  "If it is not set, CMake warns, and uses OLD behavior."
But: "The OLD behavior of a policy is deprecated by definition and may be removed in a future version of CMake."

vrabaud:
The goal here really is to have less code (56 lines less)

cmake/Modules/AvifExternalProjectUtils.cmake
```diff
@@ -1,16 +1,20 @@
...
-
         # Force static build
-        set(BUILD_SHARED_LIBS_ORIG ${BUILD_SHARED_LIBS})
-        set(BUILD_SHARED_LIBS OFF CACHE INTERNAL "")
-        set(BUILD_TESTING_ORIG ${BUILD_TESTING})
-        set(BUILD_TESTING OFF CACHE INTERNAL "")
+        set(BUILD_SHARED_LIBS OFF)
+        set(BUILD_TESTING OFF)
+
+        FetchContent_MakeAvailable(${name})
```
wantehchang, line 8: @fdintino Hi Frankie,

The `FetchContent_MakeAvailable()` function was added in CMake 3.14:
https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable

When this file was added in commit f1b936d7394da74cddb505f4dcd61e492b5c2233, our CMake minimum required version was 3.13. Is that why you emulated `FetchContent_MakeAvailable()` with `FetchContent_Populate()` and `add_subdirectory()`? Or did you want to see `BUILD_SHARED_LIBS` and `BUILD_TESTING` to `OFF` for the `add_subdirectory()` call?

Our CMake minimum required version is 3.22 now.

cmake/Modules/AvifExternalProjectUtils.cmake
```diff
@@ -1,16 +1,20 @@
-macro(avif_fetchcontent_populate_cmake name)
+function(avif_fetchcontent_populate_cmake name)
```
wantehchang: Rename this function to "avif_fetchcontent_makeavailable_cmake"?

The "avif_fetchcontent_populate_cmake" name is clearly suggesting it wraps the FetchContent_Populate cmake function. Now it wraps the FetchContent_MakeAvailable cmake function.

cmake/Modules/AvifExternalProjectUtils.cmake
```diff
@@ -1,16 +1,20 @@
...
     if(NOT ${name}_POPULATED)
-        FetchContent_Populate(${name})
-
         # Force static build
-        set(BUILD_SHARED_LIBS_ORIG ${BUILD_SHARED_LIBS})
-        set(BUILD_SHARED_LIBS OFF CACHE INTERNAL "")
-        set(BUILD_TESTING_ORIG ${BUILD_TESTING})
-        set(BUILD_TESTING OFF CACHE INTERNAL "")
+        set(BUILD_SHARED_LIBS OFF)
+        set(BUILD_TESTING OFF)
```
wantehchang, line 6: Please confirm that we don't need to save and restore the values of these two variables.

cmake/Modules/LocalAom.cmake
```diff
@@ -67,15 +67,8 @@ else()
...
-    set(AOM_BINARY_DIR "${FETCHCONTENT_BASE_DIR}/aom-build")
-
-    if(ANDROID_ABI)
-        set(AOM_BINARY_DIR "${AOM_BINARY_DIR}/${ANDROID_ABI}")
-    endif()
-
     FetchContent_Declare(
-        libaom URL "https://aomedia.googlesource.com/aom/+archive/${AVIF_AOM_GIT_TAG}.tar.gz" BINARY_DIR "${AOM_BINARY_DIR}"
-        UPDATE_COMMAND ""
+        libaom URL "https://aomedia.googlesource.com/aom/+archive/${AVIF_AOM_GIT_TAG}.tar.gz" UPDATE_COMMAND "" EXCLUDE_FROM_ALL
```
wantehchang: [This comment applies to all the Local*.cmake files.]

Optional: It would be good if `EXCLUDE_FROM_ALL` could be moved to a separate line.

It's not too bad in this file, but it's quite confusing in LocalAvm.cmake and LocalFuzztest.cmake.

cmake/Modules/AvifExternalProjectUtils.cmake
fdintino: This was quite some time ago, but my recollection is that I used `FetchContent_Populate` and `add_subdirectory` in place of `FetchContent_MakeAvailable` since it wasn't an option. But `BUILD_SHARED_LIBS` and `BUILD_TESTING` do also need to be set for `add_subdirectory`. I'm pretty sure that `FetchContent_MakeAvailable` calls `add_subdirectory`, so this should still do that.

cmake/Modules/AvifExternalProjectUtils.cmake
fdintino: I think this hinges on whether the `CACHE` argument is necessary or not. I'd imagine that if it was the CI failures would be obvious;

cmake/Modules/LocalAom.cmake
vrabaud: Done

cmake/Modules/AvifExternalProjectUtils.cmake
vrabaud: Right, it does not need to be put in the cache, it can stay local and it will be used by add_subdirectory properly

cmake/Modules/AvifExternalProjectUtils.cmake
vrabaud: Done

cmake/Modules/LocalAom.cmake
```diff
@@ -67,14 +67,10 @@ else()
...
-
-    if(ANDROID_ABI)
-        set(AOM_BINARY_DIR "${AOM_BINARY_DIR}/${ANDROID_ABI}")
-    endif()
-
     FetchContent_Declare(
-        libaom URL "https://aomedia.googlesource.com/aom/+archive/${AVIF_AOM_GIT_TAG}.tar.gz" BINARY_DIR "${AOM_BINARY_DIR}"
+        libaom
+        EXCLUDE_FROM_ALL
+        URL "https://aomedia.googlesource.com/aom/+archive/${AVIF_AOM_GIT_TAG}.tar.gz"
```
wantehchang, line 73: Do you know why we use the URL method rather than the GIT_REPOSITORY + GIT_TAG method with libaom?

vrabaud: I saw that too while tweaking that code. I vaguely remember the git repo failing once in a while. Other sources use gitlab and github. Downloading archives all the time would probably be a bit faster but we would lose some git tooling once in the repo (like "git grep").

wantehchang: I traced it to https://github.com/AOMediaCodec/libavif/pull/1744.

@fdintino Frankie: Do you remember why you used the URL method rather than the GIT_REPOSITORY + GIT_TAG method with libaom?

PR #3091, opened 2026-03-09
auto-generate output filename when not specified

### LINKED ISSUE #3090: [Feature Request] Make specifying output filename optional, and make avifenc.exe accept wildcard (*.jpg)
This is a feature request for improving the ease of use of the avifenc command line encoder.

1. It would be great if avifenc.exe could accept an input filename without specifying an output filename, where the output filename would then simply be copied from the input filename, but with the new .avif extensions instead of .jpg/.png.
For example: `avifenc.exe input1.jpg` which would create a file called input1.avif, instead of needing `avifenc.exe input1.jpg output1.avif`

2. When the above is made possible, this would also open the door to accepting multiple input filenames in one go.
For example: `avifenc.exe input1.jpg input2.jpg input3.jpg`

3. When that is possible, we arrive at the destination where I would love avifenc to end up, which is being able to use it with a wildcard.
For example: `avifenc.exe *.jpg`

A very common (possibly the most common?) scenario is where a user wants to compress a folder of images to AVIF.
Currently, to do this on Windows, you need a cumbersome powershell script like so:
`ls *.jpg|foreach{.\avifenc.exe $_ $_.name.replace("jpg", "avif")}`
This is not very user/beginner friendly.

Compare this to cavif.exe for example. It accepts an input filename only, and will then automatically copy this to the output filename, but with .avif extension. This allows dragging-and-dropping a file onto the cavif.exe executable in Windows Explorer, making it possible to compress to AVIF without even needing the command line. Cavif also allows you to drag-and-drop _multiple_ files onto the executable. Unfortunately it does not seem to accept a wildcard (*.jpg) input yet.

Other command line encoders, for example for FLAC or WavPack audio, can be used with a wildcard. You drop the executable in a folder, and by simply doing `flac.exe *.wav` or `wavpack.exe *.wav` you automatically compress all files in the folder. Super easy. Or you select all input files in the folder and drag-and-drop onto the executable. Even easier!

I sincerely hope this proposal can be considered for implementation.

rootvector2:
Fixes #3090

Allow `avifenc` to auto-generate the output filename when only a single input file is provided and no output is specified. The output name is derived by replacing the input extension with `.avif`, or appending `.avif` if the input has no extension.

This behavior only applies when exactly one input file is given and no `-o` or positional output argument is present, so existing CLI behavior remains unchanged.

Also adds a safeguard to prevent accidentally overwriting the input file when the input already has a `.avif` extension.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -188,15 +188,17 @@ typedef struct avifEncodedByteSizes
 
 static void syntaxShort(void)
 {
-    printf("Syntax: avifenc [options] -q quality input.[jpg|jpeg|png|y4m] output.avif\n");
+    printf("Syntax: avifenc [options] -q quality input.[jpg|jpeg|png|y4m] [output.avif]\n");
     printf("where quality is between %d (worst quality) and %d (lossless).\n", AVIF_QUALITY_WORST, AVIF_QUALITY_LOSSLESS);
-    printf("Typical value is 60-80.\n\n");
+    printf("Typical value is 60-80.\n");
+    printf("If output.avif is not specified, it is derived from the input filename with a .avif extension.\n\n");
     printf("Try -h for an exhaustive list of options.\n");
 }
 
 static void syntaxLong(void)
 {
-    printf("Syntax: avifenc [options] input.[jpg|jpeg|png|y4m] output.avif\n");
+    printf("Syntax: avifenc [options] input.[jpg|jpeg|png|y4m] [output.avif]\n");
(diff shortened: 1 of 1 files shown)


rootvector2:
This PR focuses only on the optional output filename behavior. The other suggestions in the issue (multi-input batch conversion and wildcard handling) may require maintainer guidance since multiple inputs currently have existing semantics (e.g., animations).

pinballmachine:
Thank you very much for this improvement! I hope this will make it possible to use drag-and-drop of single files onto the executable.

PR #3094, opened 2026-03-10
Add avif_off_t and avif_fseeko() and avif_ftello()

wantehchang:
Add simple wrappers around POSIX's fseeko() and ftello() and Windows' _fseeki64() and _ftelli64().

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -9,7 +9,9 @@
 
 #include "avif/internal.h"
 
+#include <assert.h>
 #include <limits.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <string.h>
 
@@ -28,6 +30,7 @@ static avif_off_t avif_ftello(FILE * stream)
 }
 #else
 // POSIX large file support
+static_assert(sizeof(off_t) == sizeof(int64_t), "");
 typedef off_t avif_off_t;
 
 static int avif_fseeko(FILE * stream, avif_off_t offset, int whence)


src/io.c
```diff
@@ -104,13 +138,13 @@ static avifResult avifIOFileReaderRead(struct avifIO * io, uint32_t readFlags, u
     }
 
     if (size > 0) {
-        if (offset > LONG_MAX) {
+        if (offset > INT64_MAX) {
```
wantehchang, line 144: James: This is the only weakness of this pull request. There is no symbolic constant such as `OFF_MAX` for the maximum value of the `off_t` type. So here we assume without checking that `off_t` is 64 bits. (`off_t` is a signed integer by definition.)

This check ensures that it is safe to cast `offset` to `avif_off_t` at line 147.

src/io.c
```diff
@@ -1,12 +1,46 @@
 // Copyright 2020 Joe Drago. All rights reserved.
 // SPDX-License-Identifier: BSD-2-Clause
 
+#if !defined(_WIN32)
```
jzern, line 4: I was thinking cygwin/mingw might support this, but they should also support the windows specifics you added. We can test this if the toolchains are covered by the presubmit.

src/io.c
jzern: Nothing common anyway. I see gcc/clang report `#define __OFF_T_MATCHES_OFF64_T 1`. I wonder if it's worth adding a compile time assert in the posix path for `sizeof(off_t) == 8`?

wantehchang: Done. Now that libavif uses C11, we can include <assert.h> and use `static_assert()`.

I use `sizeof(int64_t)` instead of 8 in the static assertion. I hope that is okay.

src/io.c
wantehchang: Cygwin doesn't define `_WIN32`, so Cygwin will use `fseeko` and `ftello`.

I want only MSVC and MinGW to use `_fseeki64` and `_ftelli64`.

PR #3096, opened 2026-03-11
Add more CI for the oldest supported CMake

vrabaud:
And fix the resulting bugs.


cmake/Modules/LocalAom.cmake
```diff
@@ -144,7 +144,7 @@ else()
     endif()
 
     set_property(TARGET aom PROPERTY AVIF_LOCAL ON)
-    target_include_directories(aom INTERFACE $<BUILD_INTERFACE:$<PATH:ABSOLUTE_PATH,${libaom_SOURCE_DIR},/>>)
+    target_include_directories(aom INTERFACE $<BUILD_INTERFACE:${libaom_SOURCE_DIR}> $<BUILD_INTERFACE:${libaom_BINARY_DIR}>)
```
wantehchang: IMPORTANT: Why is ${libaom_BINARY_DIR} added?

`${libaom_BINARY_DIR}` should not be added to the `INTERFACE` or `PUBLIC` include directories.

cmake/Modules/LocalAvm.cmake
```diff
@@ -118,7 +118,7 @@ endif()
 target_link_libraries(avm PRIVATE tensorflow-lite)
 
 set_property(TARGET avm PROPERTY AVIF_LOCAL ON)
-target_include_directories(avm INTERFACE $<BUILD_INTERFACE:$<PATH:ABSOLUTE_PATH,${libavm_SOURCE_DIR},/>>)
+target_include_directories(avm INTERFACE $<BUILD_INTERFACE:${libavm_SOURCE_DIR}>)
```
wantehchang, line 121: Good catch! [CMake documentation](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#path-transformations) says `$<PATH:ABSOLUTE_PATH>` was added in CMake 3.24.

.github/workflows/ci-linux-oldest-cmake.yml
```diff
@@ -0,0 +1,55 @@
...
+      - uses: ./.github/actions/setup-linux
+        with:
+          codec-aom: "LOCAL"
+          codec-dav1d: "LOCAL"
+          codec-rav1e: "LOCAL"
+          codec-svt: "LOCAL"
+          libjpeg-turbo: "LOCAL"
+          libxml2: "LOCAL"
+          libyuv: "LOCAL"
+          oldest-cmake: true
```
wantehchang, line 33: I verified this uses CMake 3.22.x, and the [Foundational C++ Support Table](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md#compilers-tools-build-systems) that we follow still requires CMake 3.22. Good.

.github/workflows/ci-linux-oldest-cmake.yml
```diff
@@ -0,0 +1,55 @@
+# This is a copy of ci-unix-shared-local.yml for Linux only and using the oldest supported CMake
```
wantehchang, line 1: Please add the following to the commit message:

```
Fixes https://github.com/AOMediaCodec/libavif/issues/3095.
```

wantehchang, line 1: Please add a changelog entry for this bug fix. Thanks!

cmake/Modules/LocalAom.cmake
vrabaud: It's part of BUILD_INTERFACE so it's fine (we do it for AVM, that's why I put it here). It is not necessary for libaom indeed, I removed it.

wantehchang: Before AVM is final, its public headers need to include the private header "config/avm_config.h" and test the experiment macros defined in "config/avm_config.h". This is why we allow `${libavm_SOURCE_DIR}`.

For libaom, we need to ensure that its public headers don't include any private headers by mistake.

.github/workflows/ci-linux-oldest-cmake.yml
```diff
@@ -0,0 +1,55 @@
...
+      fail-fast: false
+
+    name: build-linux-oldest-cmake
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: ./.github/actions/setup-linux
+        with:
+          codec-aom: "LOCAL"
+          codec-dav1d: "LOCAL"
```
wantehchang, line 27: Just wanted to confirm that there is no `codec-libgav1` variable, so we don't need to add `codec-libgav1: "LOCAL"`.

.github/workflows/ci-linux-oldest-cmake.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Prepare libavif (cmake)
+        run: >
+          cmake -G Ninja -S . -B build
+          -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON
+          -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_DAV1D=LOCAL
+          -DAVIF_CODEC_LIBGAV1=LOCAL -DAVIF_CODEC_RAV1E=LOCAL
+          -DAVIF_CODEC_SVT=LOCAL
+          -DAVIF_JPEG=LOCAL -DAVIF_LIBSHARPYUV=LOCAL
+          -DAVIF_LIBXML2=LOCAL -DAVIF_LIBYUV=LOCAL
+          -DAVIF_ZLIBPNG=LOCAL
```
wantehchang, line 45: I guess the goal of this CI workflow is to use `LOCAL` for all dependencies?

wantehchang:
Nit: I suggest using "Fix build with CMake 3.22" as the title of the commit message.

.github/workflows/ci-linux-oldest-cmake.yml
vrabaud: Indeed: it is because libgav1 does not require any deps (dav1d requires meson, rav1e requires Rust ...)

.github/workflows/ci-linux-oldest-cmake.yml
vrabaud: Exactly, just to test the CMake with all deps.

PR #3097, opened 2026-03-12
guard against integer overflow in width * height pixel count

sayedkaif27:
## Description

Guard against integer overflow when computing the pixel count in `avifRGBImageComputeGainMap`.

On 32-bit platforms (`size_t` = 32 bits), the expression `(size_t)width * height` can wrap
silently when both `uint32_t` operands are large, causing the subsequent
`numPixels > SIZE_MAX / sizeof(float)` check to operate on an already-corrupted value and
pass erroneously — leading to a heap buffer overflow during gain map computation.

**Fix:** add the standard unsigned overflow pre-check before the multiplication:

```c
if (width != 0 && height > SIZE_MAX / width) {
    res = AVIF_RESULT_INVALID_ARGUMENT;
    goto cleanup;
}

### DIFF
--- a/src/gainmap.c
+++ b/src/gainmap.c
@@ -285,7 +285,7 @@ avifResult avifRGBImageApplyGainMap(const avifRGBImage * baseImage,
 
         // Convert extended SDR (where 1.0 is SDR white) to nits.
         clli->maxCLL = (uint16_t)AVIF_CLAMP(avifRoundf(rgbMaxLinear * SDR_WHITE_NITS), 0.0f, (float)UINT16_MAX);
-        const float rgbAverageLinear = rgbSumLinear / ((uint64_t)width * height);
+        const float rgbAverageLinear = rgbSumLinear / ((size_t)width * height);
         clli->maxPALL = (uint16_t)AVIF_CLAMP(avifRoundf(rgbAverageLinear * SDR_WHITE_NITS), 0.0f, (float)UINT16_MAX);
     }
 
@@ -559,12 +559,11 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
     avifResult res = AVIF_RESULT_OK;
     // --- After this point, the function should exit with 'goto cleanup' to free allocated resources.
 
-    const uint64_t numPixels64 = (uint64_t)width * (uint64_t)height;
-    if (numPixels64 > SIZE_MAX / sizeof(float)) {
+    const size_t numPixels = (size_t)width * height;
+    if (numPixels > SIZE_MAX / sizeof(float)) {
         res = AVIF_RESULT_INVALID_ARGUMENT;
         goto cleanup;
     }
-    const size_t numPixels = (size_t)numPixels64;
     const size_t gainMapChannelSize = numPixels * sizeof(float);
     const avifBool singleChannel = (gainMap->image->yuvFormat == AVIF_PIXEL_FORMAT_YUV400);
     const int numGainMapChannels = singleChannel ? 1 : 3;

### CODE REVIEW

**Comment by vrabaud** [comment]
Replacing the size_t arithmetic with uint64_t is easier no? There is another (size_t)width in that file (the only other one in the code base).

**Comment by sayedkaif27** [comment]
I updated the patch to perform the multiplication in uint64_t and added the same change to the other (size_t)width * height

File: src/gainmap.c [hunk 1]
```diff
@@ -559,11 +559,12 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
     // --- After this point, the function should exit with 'goto cleanup' to free allocated resources.
 
-    const size_t numPixels = (size_t)width * height;
-    if (numPixels > SIZE_MAX / sizeof(float)) {
+    const uint64_t numPixels64 = (uint64_t)width * (uint64_t)height;
+    if (numPixels64 > SIZE_MAX / sizeof(float)) {
         res = AVIF_RESULT_INVALID_ARGUMENT;
         goto cleanup;
     }
+    const size_t numPixels = (size_t)numPixels64;
```
wantehchang: The original code is correct. Please verify the following analysis.

`width` and `height` are defined as follows:

```
    const uint32_t width = baseRgbImage->width;
    const uint32_t height = baseRgbImage->height;
```

The original code relies on the API contract that `baseRgbImage->width` and `baseRgbImage->height` are the width and height of the `baseRgbImage->pixels` buffer. It is safe to multiply `baseRgbImage->rowBytes` by `baseRgbImage->height` using the `size_t` type. Since `baseRgbImage->width <= baseRgbImage->rowBytes`, it is also safe to multiply `baseRgbImage->width` by `baseRgbImage->height` using the `size_t` type.

src/gainmap.c
```diff
@@ -285,7 +285,7 @@ avifResult avifRGBImageApplyGainMap(const avifRGBImage * baseImage,
 
         // Convert extended SDR (where 1.0 is SDR white) to nits.
         clli->maxCLL = (uint16_t)AVIF_CLAMP(avifRoundf(rgbMaxLinear * SDR_WHITE_NITS), 0.0f, (float)UINT16_MAX);
-        const float rgbAverageLinear = rgbSumLinear / ((size_t)width * height);
+        const float rgbAverageLinear = rgbSumLinear / ((uint64_t)width * height);
```
wantehchang: The original code is correct. Please check the following analysis.

`width` and `height` are defined as follows:

```
    const uint32_t width = baseImage->width;
    const uint32_t height = baseImage->height;
```

The original code relies on the API contract that `baseImage->width` and `baseImage->height` are the width and height of the `baseImage->pixels` buffer. It is safe to multiply `baseImage->rowBytes` by `baseImage->height` using the `size_t` type (done at line 119 and inside the `avifGetRGBAPixel()` call at line 147). Since `baseImage->width <= baseImage->rowBytes`, it is also safe to multiply `baseImage->width` by `baseImage->height` using the `size_t` type.

wantehchang (commented):
Thanks for analyzing the libavif code. The code in question relies on the fact that if a pixel buffer has been allocated successfully, we can multiply the width or the rowBytes (also known as the stride) of the buffer by the height of the buffer using the `size_t` type.

In the code modified in this pull request, the pixel buffer is an input parameter of the functions. Since the functions did not allocate the pixel buffer, this becomes an **API contract** that the pixel buffer and its width, height, and rowBytes are all valid.

sayedkaif27:
Thanks for the clarification. I’ve updated the code back to size_t.

sayedkaif27:
@wantehchang why this pr was closed?

wantehchang:
sayedkaif27: Your third commit ["updated unit64_t to size_t"](https://github.com/AOMediaCodec/libavif/pull/3097/changes/dbc1e3c9fd68cccefb59fc7a51b8314323248a00) restored the current code, so there are no [net changes in this PR](https://github.com/AOMediaCodec/libavif/pull/3097/changes). That's why I closed the PR. (I waited four days after your third commit to close the PR.)

PR #3098, opened 2026-03-13
Bump libyuv to 30809ff64

vrabaud:


vrabaud:
Maybe let's wait for the chain of commits to be submitted and the libyuv version actually bumped at the end of https://chromium-review.git.corp.google.com/c/libyuv/libyuv/+/7512030

PR #3101, opened 2026-03-15
Support long path names in Windows

Delphox:
Since Windows 10 1607, it's been possible to have paths longer than 260 characters, however apps need an entry in their manifest to be aware of it.  See [Microsoft Docs](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#application-manifest-updates-to-declare-long-path-capability).

This PR adds the `longPathAware` entry to the manifest. I've tested it and so far had no issues.

### DIFF
--- a/.github/actions/setup-common/action.yml
+++ b/.github/actions/setup-common/action.yml
@@ -37,9 +37,6 @@ runs:
     - name: Print CMake version
       run: cmake --version
       shell: bash
-    - name: Set up nasm
-      if: ${{ inputs.codec-aom == 'LOCAL' || inputs.codec-dav1d == 'LOCAL' || inputs.codec-svt == 'LOCAL' }}
-      uses: ilammy/setup-nasm@72793074d3c8cdda771dba85f6deafe00623038b # v1.5.2
     - name: Set up rust
       if: ${{ inputs.codec-rav1e == 'LOCAL' }}
       uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
--- a/.github/actions/setup-linux/action.yml
+++ b/.github/actions/setup-linux/action.yml
@@ -4,6 +4,9 @@ inputs:
   codec-aom:
     description: "Can take the values: OFF, LOCAL, SYSTEM"
     default: "OFF"
+  codec-avm:
+    description: "Can take the values: OFF, LOCAL, SYSTEM"
+    default: "OFF"
   codec-dav1d:
     description: "Can take the values: OFF, LOCAL, SYSTEM"
     default: "OFF"
@@ -82,6 +85,10 @@ runs:
     - name: Install libyuv library
       if: ${{ inputs.libyuv == 'SYSTEM' }}
       run: echo "AVIF_UNIX_LIBRARIES=${{ env.AVIF_UNIX_LIBRARIES }} libyuv-dev" >> "$GITHUB_ENV"
+      shell: bash
+    - name: Install nasm
+      if: ${{ inputs.codec-aom == 'LOCAL' || inputs.codec-avm == 'LOCAL' || inputs.codec-dav1d == 'LOCAL' || inputs.codec-svt == 'LOCAL' }}
+      run: echo "AVIF_UNIX_LIBRARIES=${{ env.AVIF_UNIX_LIBRARIES }} nasm" >> "$GITHUB_ENV"
       shell: bash
     - name: Install zlibpng library
       if: ${{ inputs.zlibpng == 'SYSTEM' }}
--- a/.github/actions/setup-macos/action.yml
+++ b/.github/actions/setup-macos/action.yml
@@ -4,6 +4,9 @@ inputs:
   codec-aom:
     description: "Can take the values: OFF, LOCAL, SYSTEM"
     default: "OFF"
+  codec-avm:
+    description: "Can take the values: OFF, LOCAL, SYSTEM"
+    default: "OFF"
   codec-dav1d:
     description: "Can take the values: OFF, LOCAL, SYSTEM"
     default: "OFF"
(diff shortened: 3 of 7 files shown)


y-guyon:

The change looks good to me. \
Could you please add a small test in `.github/workflows/ci-windows.yml`, for example similar to:

```yaml
      - name: Verify longPathAware
        run: .\avifenc.exe tests\data\abc.png .\verylongpath_000000010_000000020_000000030_000000040_000000050_000000060_000000070_000000080_000000090_000000100_000000110_000000120_000000130_000000140_000000150_000000160_000000170_000000180_000000190_000000200_000000210_000000220_000000230_000000240_000000250_000000260.avif
```

apps/utf8.manifest
```diff
@@ -3,6 +3,7 @@
   <application>
     <windowsSettings>
       <activeCodePage xmlns="http://schemas.microsoft.com/SMI/2019/WindowsSettings">UTF-8</activeCodePage>
+      <longPathAware xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">true</longPathAware>
```
wantehchang, line 6: I am not familiar with the manifest file format. The [Microsoft documentation page](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#application-manifest-updates-to-declare-long-path-capability) you linked to gives the following example, which looks quite different in format from the new line you added:

```
<application xmlns="urn:schemas-microsoft-com:asm.v3">
    <windowsSettings xmlns:ws2="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
        <ws2:longPathAware>true</ws2:longPathAware>
    </windowsSettings>
</application>
```

I found [another example](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#longPathAware):

```
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
 ...
  <asmv3:application>
    <asmv3:windowsSettings xmlns:ws2="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
      <ws2:longPathAware>true</ws2:longPathAware>
    </asmv3:windowsSettings>
  </asmv3:application>
 ...
</assembly>
```

The main difference seems to be `ws2:longPathAware` vs. `longPathAware`.

Delphox: In the example in the MS page, they aliased the "http://schemas.microsoft.com/SMI/2016/WindowsSettings" namespace to `ws2`, while in my line I just declared the namespace directly.

Delphox:
Seems like the test if failing because in order for the longpathAware to take effect it also requires `HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled` key in the registry to be set to 1, which requires a restart to take effect (which is not recommended in a CI...). Presumably for legacy reasons.

.github/workflows/ci-windows.yml
```diff
@@ -99,3 +99,7 @@ jobs:
           /out:avifenc.exe
 
           .\avifenc.exe --help
+      - name: Verify longPathAware
+        run: |
```
wantehchang: Delphox: Note that I added a "pipe" character `|` at the end of the "run:" line.

Delphox: Thanks! I had missed it.

Delphox:
> Seems like the test if failing because in order for the longpathAware to take effect it also requires `HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled` key in the registry to be set to 1, which requires a restart to take effect (which is not recommended in a CI...). Presumably for legacy reasons.

So turns out a restart isn't necessary, it was just missing a linking step. I fixed it in my testing branch and the test passed, I'll just clean it up and push here soon.

Delphox:
Should be good now.

Note: the reason I went with a chain of folders rather than a big name is that individual file/folder names are still limited to 255 characters, only the whole path has its 260 character limit lifted.

Delphox:
So it turns out the registry key isn't necessary in modern Windows, which is what I originally thought. I was just running in circles trying to troubleshoot why the test was failing and thought it might be necessary. But it was just the missing linking step. Should be all good now hopefully.

.github/workflows/ci-windows.yml
y-guyon: Note that it is `>` at line 82. I did not look up the difference.

Delphox: `>` squashes new lines into spaces unless it's a double new line, so its good for long commands with a lot of arguments. While `|` just runs commands from each line normally.

I could change it into `>` and replace the underlines with spaces so I could split the long path into smaller blocks that are easier to follow and doesn't make the document so horizontally large. Should I do that?

y-guyon: If you would like to bother and it works, sure! Up to you.

Delphox: Done. I tested in another branch earlier and it works.

.github/workflows/ci-windows.yml
```diff
@@ -98,4 +98,16 @@ jobs:
...
+      - name: Verify longPathAware
+        run: >
+          New-Item -ItemType "Directory" -Name "verylongpath 000000010 000000020 000000030 000000040
+          000000050\000000060 000000070 000000080 000000090 000000100\000000110 000000120 000000130
+          000000140 000000150\000000160 000000170 000000180 000000190\000000200 000000210 000000220"
+          
+          .\avifenc.exe tests\data\abc.png ".\verylongpath 000000010 000000020 000000030 000000040
+          000000050\000000060 000000070 000000080 000000090 000000100\000000110 000000120 000000130
+          000000140 000000150\000000160 000000170 000000180 000000190\000000200 000000210
+          000000220\000000230 000000240 000000250 000000260.avif"
```
y-guyon: Add a trailing new line at end of file

.github/workflows/ci-windows.yml
```diff
@@ -98,4 +98,16 @@ jobs:
...
+          mt.exe -manifest apps\utf8.manifest -outputresource:avifenc.exe;1
+
           .\avifenc.exe --help
+      - name: Verify longPathAware
+        run: >
+          New-Item -ItemType "Directory" -Name "verylongpath 000000010 000000020 000000030 000000040
+          000000050\000000060 000000070 000000080 000000090 000000100\000000110 000000120 000000130
+          000000140 000000150\000000160 000000170 000000180 000000190\000000200 000000210 000000220"
+          
+          .\avifenc.exe tests\data\abc.png ".\verylongpath 000000010 000000020 000000030 000000040
```
wantehchang: It would be better to test the avifenc.exe binary from the "Build libavif" step.

wantehchang (approved):
LGTM. Thanks. I tested the patch locally. I had to first set a registry value to enable long paths on my computer. I tested both avifenc and avifdec.

wantehchang:
Delphox wrote:
> So it turns out the registry key isn't necessary in modern Windows, which is what I originally thought.

My computer is running Windows 11 Enterprise, Version 25H2, OS build 26200.8037. I had to set the registry value to enable long paths on my computer.

I guess GitHub's Windows runner image already has the registry value set to 1.

.github/workflows/ci-windows.yml
Delphox: Done!

wantehchang:
Delphox: Thank you very much for updating the test. I think the long pathname you constructed is very clever.

I am going to ignore the three MinGW CI failures. They are all caused by ImageMagick errors in test_cmd_icc_profile and test_cmd_transform. Here is a representative error:

```
+ magick avif_test_cmd_icc_profile_decoded.png -profile D:/a/libavif/libavif/tests/data/sRGB2014.icc avif_test_cmd_icc_profile_corrected.png
magick.exe: unable to open image 'avif_test_cmd_icc_profile_corrected.png': Invalid argument @ error/blob.c/OpenBlob/3710.
magick.exe: WriteBlob Failed `avif_test_cmd_icc_profile_corrected.png' @ error/png.c/MagickPNGError/1305.
```

wantehchang:
Update: The ImageMagick error is a segmentation fault in two of the three MinGW CI worflows, for example,

```
+ magick avif_test_cmd_icc_profile_decoded.png -profile D:/a/libavif/libavif/tests/data/sRGB2014.icc avif_test_cmd_icc_profile_corrected.png
D:/a/libavif/libavif/tests/test_cmd_icc_profile.sh: line 84:   721 Segmentation fault         "${IMAGEMAGICK}" "${DECODED_FILE}" -profile "${SRGB_ICC}" "${CORRECTED_FILE}"
```

.github/workflows/ci-windows.yml
```diff
@@ -98,4 +98,16 @@ jobs:
...
+          mt.exe -manifest apps\utf8.manifest -outputresource:avifenc.exe;1
+
           .\avifenc.exe --help
+      - name: Verify longPathAware
+        run: >
+          New-Item -ItemType "Directory" -Name "verylongpath 000000010 000000020 000000030 000000040
+          000000050\000000060 000000070 000000080 000000090 000000100\000000110 000000120 000000130
+          000000140 000000150\000000160 000000170 000000180 000000190\000000200 000000210 000000220"
+          
+          & ".\build\${env:CONFIG_DIR}avifenc.exe" tests\data\abc.png ".\verylongpath 000000010 000000020 000000030 000000040
```
wantehchang, line 110: Delphox: Do we need the `&` before the second command?

Delphox: Yes, it's necessary to expand ${env:CONFIG_DIR} variable in Powershell.

Issue #3103, opened 2026-03-16
Unsupported nasm: multipass optimization not supported


y-guyon:
```sh
git clone https://github.com/AOMediaCodec/libavif.git
cmake -S libavif -B libavif/build -DAVIF_CODEC_AOM=LOCAL -DAVIF_LIBYUV=LOCAL
```
```
-- The C compiler identification is GNU 15.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.8.1")
-- libavif(AVIF_LIBYUV=LOCAL): compiled library not found at /home/yguyon/tmp/libavif/ext/libyuv/build/libyuv.a; using FetchContent
-- libavif(AVIF_LIBYUV=LOCAL): fetching and configuring libyuv
-- The CXX compiler identification is GNU 15.2.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- libavif(AVIF_LIBYUV=LOCAL): fetching and configuring libyuv - complete
-- libavif: libyuv (1922) found; libyuv-based fast paths enabled.
-- libavif: Enabling warnings for GCC
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- libavif(AVIF_CODEC_AOM=LOCAL): compiled library not found at /home/yguyon/tmp/libavif/ext/aom/build.libavif/libaom.a, using FetchContent
-- libavif(AVIF_CODEC_AOM=LOCAL): fetching and configuring aom
--- aom_configure: Detected CPU: x86_64
-- The ASM_NASM compiler identification is NASM
-- Found assembler: /usr/bin/nasm
CMake Error at build/_deps/libaom-src/build/cmake/aom_optimization.cmake:219 (message):
  Unsupported nasm: multipass optimization not supported.
Call Stack (most recent call first):
  build/_deps/libaom-src/build/cmake/aom_configure.cmake:172 (test_nasm)
  build/_deps/libaom-src/CMakeLists.txt:73 (include)
```

[...]

PR #3104, opened 2026-03-16
Cherrypick libaom nasm fix

y-guyon:
Fixes https://github.com/AOMediaCodec/libavif/issues/3103


cmake/Modules/LocalAom.cmake
```diff
@@ -72,6 +72,15 @@ else()
...
         UPDATE_COMMAND ""
+        # Avoid the following error:
+        #   CMake Error at build/_deps/libaom-src/build/cmake/aom_optimization.cmake:219 (message):
+        #     Unsupported nasm: multipass optimization not supported.
+        #   Call Stack (most recent call first):
+        #     build/_deps/libaom-src/build/cmake/aom_configure.cmake:172 (test_nasm)
+        #     build/_deps/libaom-src/CMakeLists.txt:73 (include)
+        # TODO: Remove the patch when using a libaom version past
+        #       https://aomedia.googlesource.com/aom/+/6d2b7f71b98bfa28e372b1f2d85f137280bdb3de%5E%21/
+        PATCH_COMMAND git apply ${AVIF_SOURCE_DIR}/cmake/Modules/LocalAom.diff
```
wantehchang, line 83: Do you think we should add a changelog entry?

y-guyon: Done.

PR #3108, opened 2026-03-18
Update to libaom v3.13.2

wantehchang:
Remove the workaround for issue
https://github.com/AOMediaCodec/libavif/issues/3103.


CHANGELOG.md
```diff
@@ -11,9 +11,10 @@ The changes are relative to the previous release, unless the baseline is specifi
 ### Changed since 1.4.0
 
 * Fix build with CMake 3.22
+* Update aom.cmd/LocalAom.cmake: v3.13.2
 * Update libxml2.cmd/LocalLibXml2.cmake: v2.15.2
 * Update libyuv.cmd/LocalLibyuv.cmake: 6067afde5 (1922)
-* Fix libaom v3.13.1 build with nasm
+* Fix libaom build with nasm
```
wantehchang: Yannis: Do you think we should remove "v3.13.1" from this entry?

cmake/Modules/LocalAom.cmake
```diff
@@ -75,18 +75,6 @@ else()
...
-        #   CMake Error at build/_deps/libaom-src/build/cmake/aom_optimization.cmake:219 (message):
-        #     Unsupported nasm: multipass optimization not supported.
-        #   Call Stack (most recent call first):
-        #     build/_deps/libaom-src/build/cmake/aom_configure.cmake:172 (test_nasm)
-        #     build/_deps/libaom-src/CMakeLists.txt:73 (include)
-        # TODO: Remove the patch when using a libaom version past
-        #       https://aomedia.googlesource.com/aom/+/6d2b7f71b98bfa28e372b1f2d85f137280bdb3de%5E%21/
-        # TODO: Switch back to URL "https://aomedia.googlesource.com/aom/+archive/${AVIF_AOM_GIT_TAG}.tar.gz"
-        #       instead of GIT_* above when using a libaom version past
-        #       https://aomedia.googlesource.com/aom/+/6d2b7f71b98bfa28e372b1f2d85f137280bdb3de%5E%21/
```
wantehchang, line 88: Vincent, Yannis: It seems fine to use GIT_*. Do you want me to switch back to the URL method?

vrabaud: Please keep git for now: let's see if it is stable enough.

CHANGELOG.md
```diff
@@ -11,9 +11,10 @@ The changes are relative to the previous release, unless the baseline is specifi
 ### Changed since 1.4.0
 
 * Fix build with CMake 3.22
+* Update aom.cmd/LocalAom.cmake: v3.13.2
```
y-guyon, line 14: Should it be `v3.13.2-rc1`? Or is it due to this PR being a Draft?

CHANGELOG.md
y-guyon: I would drop the whole line. The issue was in aom which was bumped.

CHANGELOG.md
wantehchang: This is due to this PR being a Draft. When the v3.13.2 release tag is available, I will change aom.cmd and LocalAom.cmake to use v3.13.2 instead.

CHANGELOG.md
wantehchang: Done.

Issue #3111, opened 2026-03-19
ImageMagick fails or crashes in MinGW CI workflows


wantehchang:
The three MinGW CI workflows are failing. They are all caused by ImageMagick errors in test_cmd_icc_profile and test_cmd_transform.

In one MinGW CI workflow, the ImageMagick error is as follows:

```
+ magick avif_test_cmd_icc_profile_decoded.png -profile D:/a/libavif/libavif/tests/data/sRGB2014.icc avif_test_cmd_icc_profile_corrected.png
magick.exe: unable to open image 'avif_test_cmd_icc_profile_corrected.png': Invalid argument @ error/blob.c/OpenBlob/3710.
magick.exe: WriteBlob Failed `avif_test_cmd_icc_profile_corrected.png' @ error/png.c/MagickPNGError/1305.
```

In the other two MinGW CI workflows, the ImageMagick error is a segmentation fault, for example,

```
+ magick avif_test_cmd_icc_profile_decoded.png -profile D:/a/libavif/libavif/tests/data/sRGB2014.icc avif_test_cmd_icc_profile_corrected.png
D:/a/libavif/libavif/tests/test_cmd_icc_profile.sh: line 84:   721 Segmentation fault         "${IMAGEMAGICK}" "${DECODED_FILE}" -profile "${SRGB_ICC}" "${CORRECTED_FILE}"
```

kmilos:
FWIW, upstream report: https://github.com/ImageMagick/ImageMagick/issues/8630

PR #3112, opened 2026-03-19
Skip ImageMagick test in MinGW

wantehchang:
ImageMagick 7.1.2-17 fails or crashes in MinGW. Skip the ImageMagick test to avoid test failures in test_cmd_icc_profile.sh and test_cmd_transform.sh.

Fixes https://github.com/AOMediaCodec/libavif/issues/3111.


vrabaud:
I tried 7.1.2-17 on Linux and the tests pass so this is a MinGW issue.

maryla-uc:
Thanks for the fix!

wantehchang:
Vincent: Thanks for testing ImageMagick 7.1.2-17 on Linux. Before I wrote this PR, I found that ImageMagick 7.1.2-17 is the latest version and suspected that the MinGW ImageMagick 7.1.2-17 package was likely built incorrectly.

PR #3114, opened 2026-03-19
DO NOT SUBMIT: Generate v1.4.1 release artifacts

wantehchang:


wantehchang:
I tested the release artifacts on macOS and Windows. I verified that the executable programs are statically linked with libavif. I  also encoded an image with avifenc and decoded the image with avifdec.

On Windows I tested avifenc and avifdec with a long path name :-)

wantehchang:
I tested the release artifacts on Linux.

PR #3115, opened 2026-03-20
Add a --jobs flag to avifgainmaputil for multithreading support.

maryla-uc:
Default to using all available cores like avifenc/avifdec. Also enable automatic tiling (avifenc default, not configurable for now in avifgainmaputil).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 
 ### Added since 1.4.1
 
-* Add --jobs flag to avifgainmaputil to use use multiple worker threads when
+* Add --jobs flag to avifgainmaputil to use multiple worker threads when
   reading/writing AVIF files.
 
 ### Changed since 1.4.1
--- a/apps/avifgainmaputil/convert_command.cc
+++ b/apps/avifgainmaputil/convert_command.cc
@@ -62,7 +62,7 @@ avifResult ConvertCommand::Run() {
   avifResult result =
       ReadImage(image.get(), arg_input_filename_.value(), pixel_format,
                 arg_image_read_.depth, arg_image_read_.ignore_profile,
-                /*ignore_gain_map*/ false, arg_jobs_.jobs.value());
+                /*ignore_gain_map=*/false, arg_jobs_.jobs.value());
   if (result != AVIF_RESULT_OK) {
     std::cout << "Failed to decode image: " << arg_input_filename_;
     return result;
--- a/apps/avifgainmaputil/imageio.cc
+++ b/apps/avifgainmaputil/imageio.cc
@@ -67,7 +67,7 @@ avifResult WriteImage(const avifImage* image, int grid_cols, int grid_rows,
 }
 
 namespace {
-std::string QualityString(int quality) {
+std::string QualityLevelString(int quality) {
   if (quality == AVIF_QUALITY_LOSSLESS) {
     return "Lossless";
   }
@@ -84,12 +84,12 @@ std::string QualityString(int quality) {
 }
 
 // Based on avifenc.c, changes here may be mirrored there if relevant.
-void PrintEncodingSettings(avifEncoder* encoder, bool has_gain_map) {
+void PrintEncodingSettings(const avifEncoder* encoder, bool has_gain_map) {
   std::string gain_map_str;
   if (has_gain_map) {
     gain_map_str = ", gain map quality [" +
                    std::to_string(encoder->qualityGainMap) + " (" +
-                   QualityString(encoder->qualityGainMap) + ")]";
+                   QualityLevelString(encoder->qualityGainMap) + ")]";
   }
   std::string manual_tiling_str =
       "tileRowsLog2 [" + std::to_string(encoder->tileRowsLog2) +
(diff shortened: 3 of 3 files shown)


apps/avifenc.c
```diff
@@ -1143,7 +1143,7 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
 
     char manualTilingStr[128];
     snprintf(manualTilingStr, sizeof(manualTilingStr), "tileRowsLog2 [%d], tileColsLog2 [%d]", encoder->tileRowsLog2, encoder->tileColsLog2);
-
+    // Note: this is mirrored in apps/avifgainmaputil/imageio.cc
```
wantehchang: Is this comment intended to remind people to keep this printf statement and the `PrintEncodingSettings()` function in apps/avifgainmaputil/imageio.cc in sync? If so, please explicitly state that.

apps/avifgainmaputil/program_command.h
```diff
@@ -143,6 +150,19 @@ struct ImageReadArgs {
...
+// Flags for setting the number of worker threads.
+struct JobsArgs {
+  argparse::ArgValue<int> jobs;
+
+  void Init(argparse::ArgumentParser& argparse) {
+    argparse.add_argument<int, JobsConverter>(jobs, "--jobs", "-j")
+        .help(
+            "Number of jobs (worker threads), or 'all' to potentially use as "
+            "many cores as possible.")
+        .default_value("all");
```
wantehchang, line 162: Please add a changelog entry for the --jobs option.

maryla-uc: Thanks, I still forget half of the time -_-

apps/avifenc.c
maryla-uc: Done. Indeed the point was to suggest keeping them in sync, even though it's not stricly necessary.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ The changes are relative to the previous release, unless the baseline is specifi
 
 ## [Unreleased]
 
+### Changed since 1.4.1
+
+* Add --jobs flag to avifgainmaputil and enable auto tiling.
```
wantehchang: Add the `Added since 1.4.1` section and move "Add --jobs flag to avifgainmaputil" to that section.

I think we can keep "Enable auto tiling in avifgainmaputil." in the `Changed since 1.4.1` section.

apps/avifgainmaputil/program_command.h
wantehchang: Me too. I only became sensitive to this after working on the v1.4.0 and v1.4.1 releases recently :-)

apps/avifgainmaputil/convert_command.cc
```diff
@@ -61,7 +62,7 @@ avifResult ConvertCommand::Run() {
   avifResult result =
       ReadImage(image.get(), arg_input_filename_.value(), pixel_format,
                 arg_image_read_.depth, arg_image_read_.ignore_profile,
-                /*ignore_gain_map*/ false);
+                /*ignore_gain_map*/ false, arg_jobs_.jobs.value());
```
wantehchang: Nit: `/*ignore_gain_map=*/false`

apps/avifgainmaputil/imageio.h
```diff
@@ -15,11 +15,11 @@ namespace avif {
 // other formats.
 avifResult WriteImage(const avifImage* image, int grid_cols, int grid_rows,
                       const std::string& output_filename, int quality,
-                      int speed);
+                      int speed, int jobs);
 // Reads an image in any of the supported formats. Ignores any gain map.
 avifResult ReadImage(avifImage* image, const std::string& input_filename,
                      avifPixelFormat requested_format, uint32_t requested_depth,
-                     bool ignore_profile, bool ignore_gain_map = true);
+                     bool ignore_profile, bool ignore_gain_map, int jobs);
```
wantehchang, line 22: Optional: If you want to keep the default argument (of `true`) for `ignore_gain_map`, we can declare `int jobs` before `bool ignore_profile`.

apps/avifgainmaputil/imageio.cc
```diff
@@ -65,15 +66,55 @@ avifResult WriteImage(const avifImage* image, int grid_cols, int grid_rows,
   return AVIF_RESULT_OK;
 }
 
+namespace {
+std::string QualityString(int quality) {
```
wantehchang: Optional: Perhaps name this function "QualityLevelString" or "QualityRangeString"?

Otherwise `std::to_string(encoder->qualityGainMap)` looks quite similar to `QualityString(encoder->qualityGainMap)`.

apps/avifgainmaputil/imageio.cc
```diff
@@ -65,15 +66,55 @@ avifResult WriteImage(const avifImage* image, int grid_cols, int grid_rows,
...
+    return "Medium";
+  }
+  if (quality == AVIF_QUALITY_WORST) {
+    return "Worst";
+  }
+  return "Low";
+}
+
+// Based on avifenc.c, changes here may be mirrored there if relevant.
+void PrintEncodingSettings(avifEncoder* encoder, bool has_gain_map) {
```
wantehchang: Nit: Add `const` to `avifEncoder* encoder`

CHANGELOG.md
```diff
@@ -8,6 +8,15 @@ The changes are relative to the previous release, unless the baseline is specifi
 
 ## [Unreleased]
 
+### Added since 1.4.1
+
+* Add --jobs flag to avifgainmaputil to use use multiple worker threads when
```
wantehchang: Nit: duplicate word "use"

apps/avifgainmaputil/imageio.h
maryla-uc: Forgot to reply to this. Yes I thought that was an option too but felt like it was better to keep this order and not use default values. This function is not called in that many places anyway.

PR #3119, opened 2026-03-20
Remove unused libaom and AVM image formats

wantehchang:
AOM_IMG_FMT_AOMYV12 and AOM_IMG_FMT_AOMI420 are unused.

AVM_IMG_FMT_AVMYV12 and AVM_IMG_FMT_AVMI420 are unused.


wantehchang:
James: I found that AVM also has these two unused image formats.

PR #3126, opened 2026-03-23
Improve are_images_equal for gain maps.

maryla-uc:
Compare the alternate image/gainmap by tone mapping to the alternate. Explicitly compare gainmap metadata, although if the file is jpeg (and are_images_equel doesn't actually support avif so it's gonna be jpeg), then the XMP will also differ which is already checked for.
Add logging to tell where images differ.
Add negative tests for are_images_equal with gainmaps.

### DIFF
--- a/tests/gtest/are_images_equal.cc
+++ b/tests/gtest/are_images_equal.cc
@@ -1,6 +1,17 @@
 // Copyright 2022 Google LLC
 // SPDX-License-Identifier: BSD-2-Clause
+//
 // Compares two files and returns whether they are the same once decoded.
+//
+// Usage:
+//   are_images_equal file1 file2 ignore_alpha_flag [psnr_threshold]
+//   [ignore_gain_map_flag]
+// If psnr_threshold is not provided or is 0, the images are compared for strict
+// equality, including all metadata. If psnr_threshold is greater than 0, the
+// images are compared for similarity using PSNR. Most metadata is NOT checked.
+// If a gain map is present and ignore_gain_map_flag is absent or 0, the
+// tone-mapped images are also compared using PSNR. ignore_alpha_flag and
+// ignore_gain_map_flag are booleans (0 or 1).
 
 #include <cstdint>
 #include <cstring>
@@ -56,7 +67,7 @@ ImagePtr ToneMapToAlternate(const ImagePtr& image) {
 }  // namespace
 
 int main(int argc, char** argv) {
-  if (argc != 4 && argc != 5 && argc != 6) {
+  if (argc < 4 || argc > 6) {
     std::cerr << "Wrong argument: " << argv[0]
               << " file1 file2 ignore_alpha_flag [psnr_threshold] "
                  "[ignore_gain_map_flag]"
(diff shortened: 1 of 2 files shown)


tests/gtest/are_images_equal.cc
```diff
@@ -21,6 +64,7 @@ int main(int argc, char** argv) {
     return 2;
   }
   const bool ignore_gain_map = argc > 5 && std::stoi(argv[5]) != 0;
+  const double psnr_threshold = argc > 4 ? std::stod(argv[4]) : 0;
```
y-guyon, line 78: Nit: Could be 99dB instead of 0, but if I understand correctly, 99 has a different behavior than 0.

tests/gtest/are_images_equal.cc
```diff
@@ -70,14 +121,61 @@ int main(int argc, char** argv) {
...
       std::cerr << "PSNR: " << psnr << ", images " << argv[1] << " and "
                 << argv[2] << " are not similar enough (threshold: " << argv[4]
                 << ")." << std::endl;
       return 1;
     }
     std::cout << "PSNR: " << psnr << ", images " << argv[1] << " and "
               << argv[2] << " are similar." << std::endl;
+
+    // If there is a gain map, tone map both images to their alternate headroom
+    // and check the PSNR of the tone mapped images.
```
y-guyon, line 147: Nit: What if the tone mapped image is equal to the the other image which has no gain map? It is currently considered as a failure "have different gainmap presence" but I am not sure why. Convenience?

tests/gtest/are_images_equal.cc
y-guyon, line 126: This will not be equal to `avif::testutil::GetPsnr(*toneMapped0, *toneMapped1)`, right? Maybe mention the PSNR only describes the base image in this case?

tests/gtest/aviftest_helpers.cc
y-guyon, line 266: Is this always used in positive tests? Otherwise std::err output when the test passes may be misleading.

tests/test_cmd_avifgainmaputil.sh
```diff
@@ -81,22 +83,32 @@ pushd ${TMP_DIR}
...
+
   "${AVIFGAINMAPUTIL}" convert "${INPUT_JPEG_GAINMAP_SDR}" "${AVIF_OUTPUT}"
    # should fail because icc profiles are not supported
   "${AVIFGAINMAPUTIL}" convert "${INPUT_JPEG_GAINMAP_SDR}" "${AVIF_OUTPUT}" --swap-base && exit 1
   "${AVIFGAINMAPUTIL}" convert "${INPUT_JPEG_GAINMAP_SDR}" "${AVIF_OUTPUT}" --swap-base --ignore-profile \
       --cicp 2/3/4
+
+  # Negative tests for are_images_equal: different images should not be equal.
+  "${ARE_IMAGES_EQUAL}" "${INPUT_JPEG_GAINMAP_SDR}" "${INPUT_JPEG_DIFFERENT_GAINMAP_SDR}" 0 0 0 && exit 1
+  "${ARE_IMAGES_EQUAL}" "${INPUT_JPEG_GAINMAP_SDR}" "${INPUT_JPEG_DIFFERENT_GAINMAP_SDR}" 0 40 0 && exit 1
```
y-guyon, line 107: To be sure to understand: would the following pass?

```sh
"${ARE_IMAGES_EQUAL}" "${INPUT_JPEG_GAINMAP_SDR}" "${INPUT_JPEG_DIFFERENT_GAINMAP_SDR}" 0 1 0
```

y-guyon (approved):
Thank you for following up.

tests/gtest/are_images_equal.cc
maryla-uc: Indeed the behavior is different. It's confusing but it was already this way. It could be changed but I don't think it's worth doing right now.

tests/gtest/are_images_equal.cc
maryla-uc: Well an image with a gainmap and without a gainmap are different things IMO. And image with a gainmap embeds two images. If you wanted to compare only the tone mapped image you could use avifgainmap util to tonemap and compare the result I guess.
What you could potentially consider equal is two images with gain maps where one is sdr base and the other one is hdr base but I don't think we need to handle this case.

tests/gtest/aviftest_helpers.cc
maryla-uc: There may be some tests where we expect it to fail but I think that's ok. I could change it all to stdout if you prefer.

tests/test_cmd_avifgainmaputil.sh
maryla-uc: Correct

PR #3127, opened 2026-03-23
Fix memory leak of altICC if avifDecoderFindGainMapItem returns early.

maryla-uc:
Make sure none of the inputs variables are modified until the very end.

Fixes b/495087335

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -5893,8 +5893,7 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
     }
 
     if (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_GAIN_MAP) {
-        gainMapTmp.image = avifImageCreateEmpty();
-        avifImage * image = gainMapTmp.image;
+        avifImage * image = avifImageCreateEmpty();
         if (!image) {
             avifRWDataFree(&gainMapTmp.altICC);
             return AVIF_RESULT_OUT_OF_MEMORY;
@@ -5905,6 +5904,7 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
             image->matrixCoefficients = matrixCoefficients;
             image->yuvRange = yuvRange;
         }
+        gainMapTmp.image = image;
     }
 
     // Only set the output pointers after everything has been validated.


src/read.c
```diff
@@ -5792,37 +5792,43 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
...
-                                          &toneMappedImageItemTmp->properties,
-                                          &gainMapTmp.altICC,
-                                          &gainMapTmp.altColorPrimaries,
-                                          &gainMapTmp.altTransferCharacteristics,
-                                          &gainMapTmp.altMatrixCoefficients,
-                                          &gainMapTmp.altYUVRange,
-                                          /*cicpSet=*/NULL));
+    AVIF_CHECKRES(result);
+
+    // This may allocate gainMapTmp.altICC which must be freed in case of error.
```
wantehchang, line 5819: I verified this is documented:

```
// ... The output parameters may be populated even in case of failure
// and must be ignored (and the |icc| object may need to be freed).

```

src/read.c
```diff
@@ -5868,19 +5877,28 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
...
+    if (!decoder->image->gainMap) {
+        avifRWDataFree(&gainMapTmp.altICC);
+        return AVIF_RESULT_OUT_OF_MEMORY;
+    }
 
     if (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_GAIN_MAP) {
-        decoder->image->gainMap->image = avifImageCreateEmpty();
-        avifImage * image = decoder->image->gainMap->image;
-        AVIF_CHECKERR(image, AVIF_RESULT_OUT_OF_MEMORY);
+        gainMapTmp.image = avifImageCreateEmpty();
```
wantehchang: Optional: Inside this if block, I suggest assigning to `gainMapTmp.image` after everything is done:

```
    if (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_GAIN_MAP) {
        avifImage * image = avifImageCreateEmpty();
        if (!image) {
            avifRWDataFree(&gainMapTmp.altICC);
        }
        if (cicpSet) {
            image->colorPrimaries = colorPrimaries;
            image->transferCharacteristics = transferCharacteristics;
            image->matrixCoefficients = matrixCoefficients;
            image->yuvRange = yuvRange;
        }
        gainMapTmp.image = image;
    }
```

src/read.c
```diff
@@ -5792,37 +5792,43 @@ static avifResult avifDecoderFindGainMapItem(const avifDecoder * decoder,
...
+                                     &toneMappedImageItemTmp->properties,
+                                     &gainMapTmp.altICC,
+                                     &gainMapTmp.altColorPrimaries,
+                                     &gainMapTmp.altTransferCharacteristics,
+                                     &gainMapTmp.altMatrixCoefficients,
+                                     &gainMapTmp.altYUVRange,
+                                     /*cicpSet=*/NULL);
+    if (result != AVIF_RESULT_OK) {
+        avifRWDataFree(&gainMapTmp.altICC);
+        return result;
```
wantehchang, line 5830: Optional: The `AVIF_CHECK*` macros call `avifBreakOnError()` before returning. For example,

```
#define AVIF_CHECKERR(A, ERR)   \
    do {                        \
        if (!(A)) {             \
            avifBreakOnError(); \
            return ERR;         \
        }                       \
    } while (0)
```

I personally don't use that, but you may want to add the `avifBreakOnError()` calls back manually.

maryla-uc: That's a fair point, although the function calling this function will break on error.
I think adding avifBreakOnError() here is too noisy and not consistent with the rest of the code.
We could maybe add a macro:

```
#define AVIF_RETURNERR(ERR) \
    do {                    \
        avifBreakOnError(); \
        return ERR;         \
    } while (0)
```

What do you think? My only fear is that it makes it a less obvious that this is essentially a return statement.
In any case I would do this in another PR (if at all).

wantehchang: You are right! The only call to this function is enclosed in the `AVIF_CHECKRES` macro. Let's not add `avifBreakOnError()` then.

Issue #1665, opened 2023-10-06
avifenc: Add an option to ignore the alpha channel


vigneshvg:
Add an option to avifenc that will make it ignore the alpha channel even if it is found in the input. The option can be named `--ignore-alpha` or `--no-alpha`.

Some tools add alpha channels to PNG files inadvertently. So it would help `avifenc` encode such images without the alpha channel (see https://github.com/AOMediaCodec/libavif/issues/1658#issuecomment-1751388161 for an example).

vrabaud:
We could also have an option to pre-multiply RGB with A.
FYI, WebP also has an option to blend the alpha against a given color.
```
  -blend_alpha <hex> ..... blend colors against background color
                           expressed as RGB values written in
                           hexadecimal, e.g. 0xc0e0d0 for red=0xc0
                           green=0xe0 and blue=0xd0
  -noalpha ............... discard any transparency information
```

vigneshvg:
> We could also have an option to pre-multiply RGB with A. FYI, WebP also has an option to blend the alpha against a given color.
> 
> ```
>   -blend_alpha <hex> ..... blend colors against background color
>                            expressed as RGB values written in
>                            hexadecimal, e.g. 0xc0e0d0 for red=0xc0
>                            green=0xe0 and blue=0xd0
>   -noalpha ............... discard any transparency information
> ```

Yes, we could certainly add that option as well. This issue was just to ignore the alpha channel (similar to `-noalpha`) in webp to deal with files which inadvertently have an alpha channel in the input and have no use for them.

wantehchang:
Fixed by PR https://github.com/AOMediaCodec/libavif/pull/3216. Added the --ignore-alpha option to avifenc.

PR #3130, opened 2026-03-24
avifenc: add --noalpha flag to discard alpha channel on encode

enyetsuen:
Adds an option to avifenc to ignore and discard any alpha channel present in the input image and encode the output as fully opaque. This is useful for cases where input assets contain unintended or accidental alpha channels (for example, PNGs exported by certain tools that always include an alpha plane), but the final AVIF output is intended to be RGB-only. Also helps with avoiding unnecessary alpha processing overhead during encoding.

The feature aligns with existing behavior in encoders such as WebP's [cwebp](https://developers.google.com/speed/webp/docs/cwebp), which already provides a -noalpha option to discard transparency when it is not required.

It was also requested by users at [#1665 ](https://github.com/AOMediaCodec/libavif/issues/1665)

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -528,6 +528,7 @@ static avifBool avifInputReadImage(avifInput * input,
                                    avifBool ignoreColorProfile,
                                    avifBool ignoreExif,
                                    avifBool ignoreXMP,
+                                   avifBool ignoreAlpha,
                                    avifBool allowChangingCicp,
                                    avifBool ignoreGainMap,
                                    avifImage * image,
@@ -635,6 +636,11 @@ static avifBool avifInputReadImage(avifInput * input,
         }
         return AVIF_FALSE;
     }
+    
+    if (ignoreAlpha) {
+        dstImage->alphaPlane = NULL;
+        dstImage->alphaRowBytes = 0;
+    }
     if (!allowChangingCicp) {
         // Restore the previous primaries/transfer in case avifReadImage changed them.
         dstImage->colorPrimaries = colorPrimariesBefore;
@@ -660,6 +666,7 @@ static avifBool avifInputReadImage(avifInput * input,
                                   ignoreColorProfile,
                                   ignoreExif,
                                   ignoreXMP,
+                                  ignoreAlpha,
                                   allowChangingCicp,
                                   ignoreGainMap,
                                   image,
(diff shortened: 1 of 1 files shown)


apps/avifenc.c
```diff
@@ -2445,6 +2450,11 @@ int main(int argc, char * argv[])
         image->clli.maxPALL = (uint16_t)settings.clliValues[1];
     }
 
+    if (settings.ignoreAlpha) {
+        image->alphaPlane    = NULL;
+        image->alphaRowBytes = 0;
```
wantehchang: This is safe only if `image->imageOwnsAlphaPlane` is false. I think that is the case, but we should analyze the code and verify that.

Rather than setting `image->alphaPlane` to NULL and `image->alphaRowBytes` to 0 here, I would consider passing `ignoreAlpha` as a new parameter to `avifInputReadImage()` and doing the work in that function istead.

enyetsuen: I believe image->imageOwnsAlphaPlane is always false at the point alpha is nulled out. I moved the work to avifInputReadImage() and passed the ignoreAlpha parameter there.

I tested encoding with aom and --noalpha and checked the output with avifdec --info and alpha is indeed absent when the parameter is passed. Let me know if you'd like me to adjust anything further.

wantehchang: enyetsuen: I am sorry for the late review. I took a close look at this pull request last night. Here are my comments.

My code analysis showed that `image->imageOwnsAlphaPlane` should be true for PNG files that have an alpha channel, and I verified that in both the first commit and the second commit of this PR with the following commands:

```
./avifenc --ignore-alpha ../tests/data/circle-trns-before-plte.png out.avif
./avifenc --ignore-alpha ../tests/data/circle-trns-after-plte.png out.avif
```

You said you believe `image->imageOwnsAlphaPlane` is always false at the point alpha is nulled out. Did you verify that?

Because of this, both the first commit and the second commit will leak the alpha plane after it is nulled out. (In addition, the first commit only handles the single image case and misses the multiple-image case such as a grid image or an image sequence.) To fix the leak of the alpha plane, we need to propagate `ignoreAlpha` further down the call chain. Unfortunately this requires changing many files because of several functions have a new `ignoreAlpha` parameter. I have made the needed changes. Since I am not good at pushing commits to your branch, I will create a copy of this pull request and then push my changes as the third commit.

wantehchang:
This PR was replaced by PR https://github.com/AOMediaCodec/libavif/pull/3216.

PR #3132, opened 2026-03-25
Get fseeko to compile with Android < 24

vrabaud:


src/io.c
wantehchang: The use of `INT64_MAX` here assumes `avif_off_t` is 64 bits. In 32-bit Android, `long` is 32 bits, so this is incorrect.

We will need to define an `AVIF_OFF_MAX` macro for the maximum value of `avif_off_t` and use `AVIF_OFF_MAX` here.

src/io.c
```diff
@@ -29,6 +30,23 @@ static avif_off_t avif_ftello(FILE * stream)
     return _ftelli64(stream);
 }
 #else
+
+#if defined(__ANDROID__)
+#include <android/api-level.h>
+#if __ANDROID_API__ >= 24
+#define USE_FSEEKO
+#else
+#define USE_FSEEK_FALLBACK
```
wantehchang, line 42: Google's [Foundational C++ Support table](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md#android-ndk) says Android NDK version >= 21 is supported. So this is necessary.

src/io.c
vrabaud: Done. Some computers apparently have long defined as __int128 though .... My patch only assumes 32 and 64 bits for avif_off_t.

src/io.c
vrabaud: Right, that's why we have 21 here: https://github.com/AOMediaCodec/libavif/blob/f962648242c69bf10e9a0525ea378ba0d03823f4/android_jni/avifandroidjni/build.gradle#L11 (and why adding -Werror fails in the CI)

android_jni/avifandroidjni/build.gradle
```diff
@@ -13,6 +13,12 @@ android {
         versionCode 1
         versionName "1.0"
         testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
+        externalNativeBuild {
+            cmake {
+                // Add -Werror to the C and C++ flags
+                arguments "-DCMAKE_C_FLAGS=-Werror", "-DCMAKE_CXX_FLAGS=-Werror"
```
wantehchang: @jzern @vigneshvg James, Vignesh: Please take a look.

Vincent: I will allow this for now. I am worried that this may override the default values of `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS`. We want to append, not override.

If you are also unsure, I suggest removing this change from this PR.

src/io.c
```diff
@@ -5,6 +5,7 @@
 // Ensure off_t is 64 bits.
 #undef _FILE_OFFSET_BITS
 #define _FILE_OFFSET_BITS 64
+#define _POSIX_C_SOURCE 200112L
```
wantehchang, line 10: (This can be done in a follow-up PR.)

1. Undefine this macro first. This file may be compiled by an alternative build system, which may define `_POSIX_C_SOURCE` on the compiler command line.
2. Please add a comment to explain the purpose of this macro definition.

src/io.c
```diff
@@ -42,7 +60,23 @@ static avif_off_t avif_ftello(FILE * stream)
...
+
+static avif_off_t avif_ftello(FILE * stream)
+{
+    return ftell(stream);
+}
+#endif // defined(USE_FSEEKO)
+
+#endif // defined(_WIN32)
+
+#define AVIF_OFF_MAX (sizeof(avif_off_t) == 8 ? INT64_MAX : INT32_MAX)
```
wantehchang: (This can be done in a follow-up PR.) It is better to define the `AVIF_OFF_MAX` macro immediately after each definition of the `avif_off_t` type alias.

android_jni/avifandroidjni/build.gradle
vigneshvg: you can append these instead by using:
```
cmake {
  cFlags "-Werror"
  cppFlags "-Werror"
}
```

vrabaud: Right, I saw those too. Actually, what would be the right spot to put those for the CI only ? We probably do not want to add those by default right ?

src/io.c
```diff
@@ -5,6 +5,9 @@
 // Ensure off_t is 64 bits.
 #undef _FILE_OFFSET_BITS
 #define _FILE_OFFSET_BITS 64
+// Ensure we have some POSIC compatible with fseeko/ftello.
```
wantehchang: Typo: POSIC

android_jni/avifandroidjni/build.gradle
vigneshvg: if it doesn't break any builds, we can simply put it in the `build.gradle` itself.

wantehchang: Now I remember that some libavif package maintainers asked us to **not** enable -Werror by default. That's why we added the `AVIF_ENABLE_WERROR` build option:

```
option(AVIF_ENABLE_WERROR "Treat all compiler warnings as errors" OFF)
```

Let's leave the build system unchanged and set `AVIF_ENABLE_WERROR` to `ON` in the CI workflow file.

PR #3133, opened 2026-03-25
Pass AVIF_ENABLE_WERROR to libavif in Android tests

vrabaud:
This would have been useful to catch the bug fixed in https://github.com/AOMediaCodec/libavif/pull/3132

### DIFF
--- a/android_jni/avifandroidjni/src/main/jni/CMakeLists.txt
+++ b/android_jni/avifandroidjni/src/main/jni/CMakeLists.txt
@@ -5,6 +5,10 @@ cmake_minimum_required(VERSION 3.22)
 
 project(avif_android_jni)
 
+# Setting the same option as libavif here ensures the AVIF_ENABLE_WERROR
+# variable is set in the CACHE for add_subdirectory to consume below.
+option(AVIF_ENABLE_WERROR "Treat all compiler warnings as errors" OFF)
+
 # Perform a static build of libavif and link it with the jni shared library so
 # that we end up with a single libavif_android.so file in the end.
 #
--- a/.github/workflows/ci-android-jni.yml
+++ b/.github/workflows/ci-android-jni.yml
@@ -42,6 +42,8 @@ jobs:
         run: sdkmanager "cmake;3.22.1"
       - name: Build the libavif JNI Wrapper
         working-directory: android_jni
-        run: ./gradlew --no-daemon assembleRelease
+        run: |
+          ./gradlew --no-daemon assembleRelease \
+          -Pandroid.extraCMakeFlags="-DAVIF_ENABLE_WERROR=ON"
         env:
           ANDROID_NDK_HOME: ${{ steps.setup-ndk.outputs.ndk-path }}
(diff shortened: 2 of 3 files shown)


vrabaud:
@wantehchang , the PR right now reverts the two commits that fixed the fseeko bug to show that my PR works. I will obviously remove them before merging.

android_jni/avifandroidjni/src/main/jni/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ cmake_minimum_required(VERSION 3.22)
 
 project(avif_android_jni)
 
+# Setting the same option as libavif here ensures the AVIF_ENABLE_WERROR
+# variable is set in the CACHE for add_subdirectory to consume below.
+option(AVIF_ENABLE_WERROR "Treat all compiler warnings as errors" OFF)
```
wantehchang, line 10: In libavif/CMakeLists.txt, we also have the following:

```
if(AVIF_ENABLE_WERROR)
    # Warnings as errors
    if(MSVC)
        target_compile_options(avif_enable_warnings INTERFACE /WX)
    elseif(CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID MATCHES "GNU")
        target_compile_options(avif_enable_warnings INTERFACE -Werror)
    else()
        message(FATAL_ERROR "libavif: Unknown compiler, bailing out")
    endif()
endif()
```

Does this file not need that?

vigneshvg:
> @wantehchang , the PR right now reverts the two commits that fixed the fseeko bug to show that my PR works. I will obviously remove them before merging.

The Android CIs are failing with this PR. So is this PR actually fixing them?

android_jni/avifandroidjni/src/main/jni/CMakeLists.txt
vrabaud: No, we just need to put AVIF_ENABLE_WERROR in the cache and it is then passed to libavif that interprets it by going through those conditions.

vrabaud:
@vigneshvg : the CI was failing to show that with my changes, the CI would have caught the fseeko error we fixed before. I reverted those 2 commits to make the PR ready to merge.

android_jni/avifandroidjni/src/main/jni/CMakeLists.txt
wantehchang: Thanks for the explanation. So it is not possible to do this using the `set(AVIF_ENABLE_WERROR ... CACHE ...)` command?

vrabaud:
You could use set(AVIF_ENABLE_WERROR ... CACHE ...) but that would be definitive. The current approach with an option restricts it to the CI.

vigneshvg:
> @vigneshvg : the CI was failing to show that with my changes, the CI would have caught the fseeko error we fixed before. I reverted those 2 commits to make the PR ready to merge.

ah ok makes sense. sorry i misunderstood your earlier comment.

vigneshvg (approved):

PR #3134, opened 2026-03-25
src/io.c: Some cleanup related to fseeko/ftello

wantehchang:
Include <unistd.h> for the definition of _POSIX_VERSION. See https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/unistd.h.html

Add the AVIF_ prefix to USE_FSEEKO.

Remove the definition of the unused macro USE_FSEEK_FALLBACK.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -39,14 +39,14 @@ static avif_off_t avif_ftello(FILE * stream)
 #if defined(__ANDROID__)
 #include <android/api-level.h>
 #if __ANDROID_API__ >= 24
-#define USE_FSEEKO
+#define AVIF_USE_FSEEKO
 #endif
 #elif defined(_POSIX_VERSION) && _POSIX_VERSION >= 200112L
 /* Standard Modern POSIX */
-#define USE_FSEEKO
+#define AVIF_USE_FSEEKO
 #endif
 
-#if defined(USE_FSEEKO)
+#if defined(AVIF_USE_FSEEKO)
 // POSIX large file support
 static_assert(sizeof(off_t) == sizeof(int64_t), "");
 typedef off_t avif_off_t;
@@ -75,7 +75,7 @@ static avif_off_t avif_ftello(FILE * stream)
 {
     return ftell(stream);
 }
-#endif // defined(USE_FSEEKO)
+#endif // defined(AVIF_USE_FSEEKO)
 
 #endif // defined(_WIN32)
 


src/io.c
```diff
@@ -34,19 +34,16 @@ static avif_off_t avif_ftello(FILE * stream)
 }
 #else
 
+#include <unistd.h>
```
wantehchang, line 37: The need to include <unistd.h> is documented in
https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/unistd.h.html

> **Version Test Macros**
> 
> The <unistd.h> header shall define the following symbolic constants. The values shall be suitable for use in **#if** preprocessing directives.
> 
> _POSIX_VERSION
> Integer value indicating version of this standard (C-language binding) to which the implementation conforms. For implementations conforming to POSIX.1-2024, the value shall be 202405L.

That webpage also gives an example of using the `#if _POSIX_VERSION >= 200112L` test for `ftello`, which matches our test at line 44:

> **APPLICATION USAGE**
> 
> POSIX.1-2024 only describes the behavior of systems that claim conformance to it. However, application developers who want to write applications that adapt to other versions of this standard (or to systems that do not conform to any POSIX standard) may find it useful to code them so as to conditionally compile different code depending on the value of _POSIX_VERSION, for example:
> 
> ```
> #if _POSIX_VERSION >= 200112L
> /* Use the newer function that copes with large files. */
> off_t pos=ftello(fp);
> #else
> /* Either this is an old version of POSIX, or _POSIX_VERSION is
>    not even defined, so use the traditional function. */
> long pos=ftell(fp);
> #endif
> ```

Note: `fseeko` and `ftello` were added in [The Open Group Base Specifications Issue 6](https://pubs.opengroup.org/onlinepubs/009695399/), in which `_POSIX_VERSION == 200112L`:
https://pubs.opengroup.org/onlinepubs/009695399/basedefs/unistd.h.html

src/io.c
```diff
@@ -5,7 +5,7 @@
 // Ensure off_t is 64 bits.
 #undef _FILE_OFFSET_BITS
 #define _FILE_OFFSET_BITS 64
-// Ensure we have some POSIX compatible with fseeko/ftello.
+// Ensure we have some POSIX compatibility with fseeko/ftello.
 #undef _POSIX_C_SOURCE
 #define _POSIX_C_SOURCE 200112L
```
wantehchang, line 10: I haven't figured out what value we should define `_POSIX_C_SOURCE` to, and whether we should not define it if the application defines it to a higher value. For example, whether we should take this kind of care:

```
// Ensure we have some POSIX compatibility with fseeko/ftello.
#if !defined(_POSIX_C_SOURCE) || _POSIX_C_SOURCE < 200112L
#undef _POSIX_C_SOURCE
#define _POSIX_C_SOURCE 200112L
#endif
```

https://pubs.opengroup.org/onlinepubs/9799919799/functions/V2_chap02.html#tag_16_02 says:

> **2.2.1.1 The _POSIX_C_SOURCE Feature Test Macro**
> 
> A POSIX-conforming application shall ensure that the feature test macro _POSIX_C_SOURCE is defined before inclusion of any header.
> 
> When an application includes a header described by POSIX.1-2024, and when this feature test macro is defined to have the value 202405L:

PR #3139, opened 2026-03-26
Set offsetBytesA for AVIF_RGB_FORMAT_GRAY

radarhere:
Within `avifGetRGBColorSpaceInfo()`, `AVIF_RGB_FORMAT_GRAY` is the only `rgb->format` that doesn't set `info->offsetBytesA`. This adds it.

`avifImageRGBToYUV()` calls `avifPrepareReformatState()`, which calls `avifGetRGBColorSpaceInfo()`.
https://github.com/AOMediaCodec/libavif/blob/24eff5475e6275b7655c4bbc68386b5fb89dba02/src/reformat.c#L218-L225
https://github.com/AOMediaCodec/libavif/blob/24eff5475e6275b7655c4bbc68386b5fb89dba02/src/reformat.c#L158-L169

I expected `offsetBytesA` to be set so that `avifImageRGBToYUV()` could access it later at
https://github.com/AOMediaCodec/libavif/blob/24eff5475e6275b7655c4bbc68386b5fb89dba02/src/reformat.c#L471

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -93,6 +93,7 @@ avifBool avifGetRGBColorSpaceInfo(const avifRGBImage * rgb, avifRGBColorSpaceInf
             break;
         case AVIF_RGB_FORMAT_GRAY:
             info->offsetBytesGray = info->channelBytes * 0;
+            info->offsetBytesA = 0;
             break;
         case AVIF_RGB_FORMAT_GRAYA:
             info->offsetBytesGray = info->channelBytes * 0;


wantehchang:
Andrew: Thank you for the pull request. Did you observe a bug? I analyzed the code. Although the uninitialized `state.rgb.offsetBytesA` value is copied to the local variable `offsetBytesA`, I verified that `offsetBytesA` is only used when the RGB format has alpha.

I like this change; I am just wondering if I should treat this PR as a bug fix and mention it in the changelog.

I took the opportunity to audit the code. There is a similar issue that `offsetBytesR/G/B` are not initialized for the three gray formats. I will write a pull request to initialize all members of the `avifRGBColorSpaceInfo` struct.

radarhere:
When calling the code with Valgrind, I found an error
```
==417== Use of uninitialised value of size 8
==417==    at 0x8EFC3B8: avifImageRGBToYUV (in /usr/lib/x86_64-linux-gnu/libavif.so.16.4.1)
```

When I added a patch with this change, the error disappeared.

So what you are saying about local variable `offsetBytesA` makes sense, and my error would have just been an automated detection.

wantehchang:
Andrew: If you have spare time, could you undo your bug fix and apply the following patch as an experiment?

```
index 7b53addf..c581a2dd 100644
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -468,7 +468,6 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
     } else if (!converted && isGray) {
         const uint32_t grayPixelBytes = state.rgb.pixelBytes;
         const uint32_t offsetBytesGray = state.rgb.offsetBytesGray;
-        const uint32_t offsetBytesA = state.rgb.offsetBytesA;
         const size_t grayRowBytes = rgb->rowBytes;
         const float grayMaxChannelF = state.rgb.maxChannelF;
         uint8_t * yPlane = image->yuvPlanes[AVIF_CHAN_Y];
@@ -484,9 +483,9 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                 if (alphaMode != AVIF_ALPHA_MULTIPLY_MODE_NO_OP) {
                     float a;
                     if (state.rgb.channelBytes > 1) {
-                        a = *((uint16_t *)(&rgb->pixels[offsetBytesA + (i * grayPixelBytes) + (j * grayRowBytes)])) / grayMaxChannelF;
+                        a = *((uint16_t *)(&rgb->pixels[state.rgb.offsetBytesA + (i * grayPixelBytes) + (j * grayRowBytes)])) / grayMaxChannelF;
                     } else {
-                        a = rgb->pixels[offsetBytesA + (i * grayPixelBytes) + (j * grayRowBytes)] / grayMaxChannelF;
+                        a = rgb->pixels[state.rgb.offsetBytesA + (i * grayPixelBytes) + (j * grayRowBytes)] / grayMaxChannelF;
                     }
 
                     if (alphaMode == AVIF_ALPHA_MULTIPLY_MODE_MULTIPLY) {
```

If this patch also fixes the valgrind error, that verifies my analysis is correct (the uninitialized value isn't used for important purposes). Thanks!

radarhere:
Yes, that patch by itself also fixes the valgrind error.

PR #3140, opened 2026-03-26
Initialize all offsets in avifGetRGBColorSpaceInfo

wantehchang:
In avifGetRGBColorSpaceInfo(), first initialize all offsets to 0. Then set the offsets used by the RGB format to the appropriate values. This ensures that avifGetRGBColorSpaceInfo() doesn't leave any members of the avifRGBColorSpaceInfo struct uninitialized.


wantehchang:
@radarhere Andrew: FYI.

PR #3145, opened 2026-03-28
Simplify casts for gridCells

wantehchang:


apps/avifgainmaputil/imageio.cc
```diff
@@ -159,10 +159,9 @@ avifResult WriteAvifGrid(const avifImage* image, int grid_cols, int grid_rows,
   avifImageDump(image, grid_cols, grid_rows,
                 AVIF_PROGRESSIVE_STATE_UNAVAILABLE);
   PrintEncodingSettings(encoder, image->gainMap != nullptr);
-  avifResult result = avifEncoderAddImageGrid(
-      encoder, grid_cols, grid_rows,
-      const_cast<const avifImage* const*>(grid_cells_ptrs.data()),
```
wantehchang, line 164: Maryla: If I understand it correctly, the return type of `grid_cells_ptrs.data()` is either `avifImage**` or `avifImage* const*`. Our target is `const avifImage* const*`, which has more `const`.

Since `const_cast` is used to remove `const`, it does not seem like the right cast to use.

apps/avifgainmaputil/imageio.cc
```diff
@@ -159,10 +159,9 @@ avifResult WriteAvifGrid(const avifImage* image, int grid_cols, int grid_rows,
...
-      AVIF_ADD_IMAGE_FLAG_SINGLE);
+  avifResult result = avifEncoderAddImageGrid(encoder, grid_cols, grid_rows,
+                                              grid_cells_ptrs.data(),
```
wantehchang, line 163: I am surprised that we don't need any cast in C++.

apps/avifenc.c
```diff
@@ -2611,7 +2611,7 @@ int main(int argc, char * argv[])
                   settings.layers > 1 ? AVIF_PROGRESSIVE_STATE_AVAILABLE : AVIF_PROGRESSIVE_STATE_UNAVAILABLE);
 
     avifEncodedByteSizes byteSizes = { 0, 0, 0 };
-    if (!avifEncodeImages(&settings, &input, firstFile, image, (const avifImage * const *)gridCells, &raw, &byteSizes)) {
+    if (!avifEncodeImages(&settings, &input, firstFile, image, (const avifImage **)gridCells, &raw, &byteSizes)) {
```
wantehchang, line 2614: `gridCells` is of the `avifImage **` type. We need a cast in C. We just need to cast it to `const avifImage **`, which is automatically converted to `const avifImage * const *`.

PR #3146, opened 2026-03-29
validate rowBytes and width overflow in avifImageRGBToYUV

uwezkhan:
Added validation in avifImageRGBToYUV to make sure rowBytes and width are safe before processing.

What was the issue

The function was using rowBytes directly without checking if it was actually large enough for the given width and pixel size.
If rowBytes is smaller than expected, it can read past the buffer.
Also, width * pixelSize could overflow in some cases.

What I changed

* Added a check to prevent overflow in width * pixelSize
* Added a check to ensure rowBytes is not smaller than required

Impact

This prevents out-of-bounds reads when invalid RGB input is passed.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -224,6 +224,29 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
         return AVIF_RESULT_REFORMAT_FAILED;
     }
 
+    // Validate that rgb->rowBytes is consistent with the image dimensions and
+    // pixel format. The slow-path conversion loop indexes pixels as:
+    //   rgb->pixels[offsetBytes + (i * pixelBytes) + (j * rowBytes)]
+    // A rowBytes that is smaller than (width * pixelBytes) would cause the
+    // per-row offset to under-stride, producing out-of-bounds reads on the
+    // first row that is not the last. Reject such inputs early.
+    {
+        const uint32_t pixelSize = avifRGBImagePixelSize(rgb);
+        // Guard width * pixelSize against uint32_t overflow before comparing.
+        if (rgb->width > 0 && pixelSize > 0) {
+            if (rgb->width > UINT32_MAX / pixelSize) {
+                // Dimensions imply a row size that cannot be represented in
+                // uint32_t; the allocation could never have been large enough.
+                return AVIF_RESULT_INVALID_ARGUMENT;
+            }
+            const uint32_t minRowBytes = rgb->width * pixelSize;
+            if (rgb->rowBytes < minRowBytes) {
+                // Caller-supplied rowBytes is too small for the declared width.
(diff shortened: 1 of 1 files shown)


b4l4-sec:
I've opened a fix PR for this in #3291.

**Root cause confirmed under ASan:** `avifImageRGBToYUV()` (and `avifImageYUVToRGB()`) never validate that `rgb->rowBytes >= rgb->width * avifRGBImagePixelSize(rgb)` before entering the pixel loop. The offset calculation at `reformat.c:320`:

```c
rgb->pixels[ offsetBytesR + (i * rgbPixelBytes) + (j * rgbRowBytes) ]
```

reads 1 byte past the end of the heap allocation at column `i=1` when `rowBytes = pixelSize` (1 pixel wide) and `width = 100`.

```
AddressSanitizer: heap-buffer-overflow READ of size 1
    #0 avifImageRGBToYUV src/reformat.c:320
    0x… is located 0 bytes after 4-byte region
```

**Fix:** 4-line early-return guard in each entry point + a regression test. See #3291.

src/reformat.c
```diff
@@ -224,6 +224,29 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
         return AVIF_RESULT_REFORMAT_FAILED;
     }
 
+    // Validate that rgb->rowBytes is consistent with the image dimensions and
+    // pixel format. The slow-path conversion loop indexes pixels as:
+    //   rgb->pixels[offsetBytes + (i * pixelBytes) + (j * rowBytes)]
+    // A rowBytes that is smaller than (width * pixelBytes) would cause the
+    // per-row offset to under-stride, producing out-of-bounds reads on the
```
y-guyon, line 231: Please add a test showing out-of-bounds reads with reasonable API use.

uwezkhan: Went to write this and couldn't get an out-of-bounds read with reasonable use. The slow path reads rgb->pixels at the caller-supplied rgb->rowBytes, and avif.h already documents that rowBytes must be at least width * avifRGBImagePixelSize(). Every in-tree caller goes through avifRGBImageAllocatePixels(), which sets rowBytes to exactly that, so none of them can trip it.

The only way to reach the over-read is to hand-set rowBytes below that documented minimum. The ASan repro in #3291 does the same thing (rowBytes = pixelSize with width = 100), so it's that contract violation rather than reasonable use.

So this is really defense-in-depth against documented misuse, not a reachable bug. I'm fine closing it, or keeping the guard as a cheap INVALID_ARGUMENT return instead of an OOB read. Your call.

wantehchang: Let's close this PR. This kind of broken input is not part of our threat model. Although we defend against an attacker that controls the width, height and pixel values, we assume that the pixel buffer is allocated properly with the width and height.

PR #3148, opened 2026-03-29
prevent underflow in stream remaining bytes check

uwezkhan:
Summary-
Added a safety check in stream functions to avoid underflow when offset goes past the buffer size.

What was the issue-
Both functions were doing size - offset using unsigned values.
If offset becomes greater than size, it wraps around to a very large number instead of failing.
This can make the parser think there is still data left when there isn’t.

What I changed-
Added a simple check to return false or 0 if offset is already beyond the buffer size.

Impact
Prevents incorrect parsing behavior and avoids using invalid buffer sizes.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -30,11 +30,22 @@ void avifROStreamStart(avifROStream * stream, avifROData * raw, avifDiagnostics 
 
 avifBool avifROStreamHasBytesLeft(const avifROStream * stream, size_t byteCount)
 {
+    // Guard against unsigned underflow: if offset has somehow advanced past the
+    // end of the buffer (e.g. via a mishandled early-return leaving stale state),
+    // the subtraction below would wrap to SIZE_MAX and incorrectly return TRUE.
+    if (stream->offset > stream->raw->size) {
+        return AVIF_FALSE;
+    }
     return byteCount <= (stream->raw->size - stream->offset);
 }
 
 size_t avifROStreamRemainingBytes(const avifROStream * stream)
 {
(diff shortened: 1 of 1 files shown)


src/stream.c
```diff
@@ -30,11 +30,22 @@ void avifROStreamStart(avifROStream * stream, avifROData * raw, avifDiagnostics
 
 avifBool avifROStreamHasBytesLeft(const avifROStream * stream, size_t byteCount)
 {
+    // Guard against unsigned underflow: if offset has somehow advanced past the
+    // end of the buffer (e.g. via a mishandled early-return leaving stale state),
+    // the subtraction below would wrap to SIZE_MAX and incorrectly return TRUE.
+    if (stream->offset > stream->raw->size) {
```
wantehchang, line 36: Thank you for the pull request.

The `avifROStream*` functions maintain the invariant that `stream->offset <= stream->raw->size`. Therefore these two functions can safely assume `stream->offset <= stream->raw->size`. 

If you are interested in analyzing the code in src/stream.c, please try to find the code that does not maintain this invariant.

wantehchang: I wrote two pull requests to document the invariant and to make it easier to verify the invariant is maintained:
- https://github.com/AOMediaCodec/libavif/pull/3156
- https://github.com/AOMediaCodec/libavif/pull/3159

After these pull requests are merged, we only need to check the code in src/stream.c to verify the invariant `stream->offset <= stream->raw->size` is maintained.

PR #3149, opened 2026-03-30
Fix integer arithmetic inconsistency in YUV/RGB conversion slow paths

YooLCD:
In `avifImageYUVAnyToRGBAnySlow()` and the YUV420/422 write path of `avifImageRGBToYUV()`, chroma row offsets are computed as `uvJ * uRowBytes` where both operands are `uint32_t`. This means the multiplication is done in 32-bit arithmetic before being used as an array index into the UV planes.

The fast paths (`avifImageYUV16ToRGB16Color`, `avifImageYUV8ToRGB8Color`, etc.) already guard against this by casting to `size_t` first:

```
// fast path — already correct
(uint16_t *)&image->yuvPlanes[AVIF_CHAN_U][((size_t)uvJ * image->yuvRowBytes[AVIF_CHAN_U])]

// slow path — before this patch
&uPlane[(uvJ * uRowBytes) + (uvI * yuvChannelBytes)]
```

If `yuvRowBytes` is provided externally (it is a public field on `avifImage`), the 32-bit product can wrap on large images. This patch adds `(size_t)` casts to all 22 affected sites in the slow path and the encode write path to match the fast paths.

Also replaces `strcpy` with `memcpy` in `read.c` when copying `AVIF_URN_ALPHA0` into `alphaAuxProp->u.auxC.auxType`. No overflow today (`AVIF_URN_ALPHA0` is 44 bytes, `AUXTYPE_SIZE` is 64), but every other string copy in the file uses `memcpy` or `avifROStreamReadString`, so this makes it consistent.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -4524,8 +4524,6 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
         colorItem->premByID = alphaIsPremultiplied;
         avifProperty * alphaAuxProp = avifMetaCreateProperty(meta, "auxC");
         AVIF_CHECKERR(alphaAuxProp, AVIF_RESULT_OUT_OF_MEMORY);
-        // Use memcpy instead of strcpy to make the buffer boundary explicit.
-        // AVIF_URN_ALPHA0 is 43 bytes + null terminator, well within AUXTYPE_SIZE (64).
         memcpy(alphaAuxProp->u.auxC.auxType, AVIF_URN_ALPHA0, sizeof(AVIF_URN_ALPHA0));
         AVIF_CHECKERR(avifDecoderItemAddProperty(alphaItem, alphaAuxProp), AVIF_RESULT_OUT_OF_MEMORY);
 
--- a/src/read.c
+++ b/src/read.c
@@ -4524,7 +4524,7 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
         colorItem->premByID = alphaIsPremultiplied;
         avifProperty * alphaAuxProp = avifMetaCreateProperty(meta, "auxC");
         AVIF_CHECKERR(alphaAuxProp, AVIF_RESULT_OUT_OF_MEMORY);
-        memcpy(alphaAuxProp->u.auxC.auxType, AVIF_URN_ALPHA0, sizeof(AVIF_URN_ALPHA0));
+        strcpy(alphaAuxProp->u.auxC.auxType, AVIF_URN_ALPHA0);
         AVIF_CHECKERR(avifDecoderItemAddProperty(alphaItem, alphaAuxProp), AVIF_RESULT_OUT_OF_MEMORY);
 
         // Property with fixed index 2 (reused).
--- a/src/read.c
+++ b/src/read.c
@@ -4524,7 +4524,9 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
         colorItem->premByID = alphaIsPremultiplied;
         avifProperty * alphaAuxProp = avifMetaCreateProperty(meta, "auxC");
         AVIF_CHECKERR(alphaAuxProp, AVIF_RESULT_OUT_OF_MEMORY);
-        strcpy(alphaAuxProp->u.auxC.auxType, AVIF_URN_ALPHA0);
+        // Use memcpy instead of strcpy to make the buffer boundary explicit.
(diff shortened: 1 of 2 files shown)


src/read.c
```diff
@@ -4524,7 +4524,9 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
         colorItem->premByID = alphaIsPremultiplied;
         avifProperty * alphaAuxProp = avifMetaCreateProperty(meta, "auxC");
         AVIF_CHECKERR(alphaAuxProp, AVIF_RESULT_OUT_OF_MEMORY);
-        strcpy(alphaAuxProp->u.auxC.auxType, AVIF_URN_ALPHA0);
+        // Use memcpy instead of strcpy to make the buffer boundary explicit.
```
wantehchang: Nit: Delete this comment.

One problem with this comment is that it describes the change rather than the resulting code.

In general, the change should be described in the commit message of a pull request.

The resulting code, if subtle, can be described in a comment.

src/read.c
```diff
@@ -4524,7 +4524,9 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
...
+        // AVIF_URN_ALPHA0 is 43 bytes + null terminator, well within AUXTYPE_SIZE (64).
```
wantehchang: Nit: Delete this comment

libavif recently switched to the ISO C11 standard. This is best valdiated with a static assertion. See pull request https://github.com/AOMediaCodec/libavif/pull/3154.

src/read.c
```diff
@@ -4524,7 +4524,9 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
...
+        memcpy(alphaAuxProp->u.auxC.auxType, AVIF_URN_ALPHA0, sizeof(AVIF_URN_ALPHA0));
```
wantehchang: Yannis: Since this is your code, I will let you decide if you want to accept this change. I am okay either way.

src/reformat.c
```diff
@@ -686,10 +686,10 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
     for (uint32_t j = 0; j < image->height; ++j) {
         // uvJ is used only when yuvHasColor is true.
         const uint32_t uvJ = yuvHasColor ? (j >> state->yuv.formatInfo.chromaShiftY) : 0;
-        const uint8_t * ptrY8 = &yPlane[j * yRowBytes];
-        const uint8_t * ptrU8 = uPlane ? &uPlane[(uvJ * uRowBytes)] : NULL;
-        const uint8_t * ptrV8 = vPlane ? &vPlane[(uvJ * vRowBytes)] : NULL;
-        const uint8_t * ptrA8 = aPlane ? &aPlane[j * aRowBytes] : NULL;
+        const uint8_t * ptrY8 = &yPlane[(size_t)j * yRowBytes];
```
wantehchang: Please undo the changes to the `avifImageYUVAnyToRGBAnySlow()` function. I have previously committed to reviewing pull request https://github.com/AOMediaCodec/libavif/pull/3063 for this function.

YooLCD: Thank you for the review! I've reverted the changes to `avifImageYUVAnyToRGBAnySlow()` and removed the comments.

src/read.c
wantehchang: Yoo_LCD: Could you move this strcpy-to-memcpy change to a separate pull request so that it can be reviewed and merged quickly? Thanks.

YooLCD: Done! I've opened a separate pull request for this change: #3161

src/reformat.c
```diff
@@ -430,13 +430,13 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                     int uvI = outerI >> chromaShiftX;
                     int uvJ = outerJ >> chromaShiftY;
                     if (state.yuv.channelBytes > 1) {
-                        uint16_t * pU = (uint16_t *)&uPlane[(uvI * 2) + (uvJ * uRowBytes)];
+                        uint16_t * pU = (uint16_t *)&uPlane[(uvI * 2) + ((size_t)uvJ * uRowBytes)];
```
wantehchang, line 433: Sorry about the late review. In commit https://github.com/AOMediaCodec/libavif/commit/f17110dac426a7dd046a34b99276e64d9fb73c28 we decided to declare `uvJ` as `size_t` rather than add `size_t` casts to `uvJ`. So this PR is no longer needed. Thanks again!

PR #3150, opened 2026-03-30
Fix minor coverity finding

maryla-uc:
CID 559069:         Performance inefficiencies  (COPY_INSTEAD_OF_MOVE)
"manual_tiling_str" is passed-by-value as parameter to "operator <<", when it could be moved instead.

### DIFF
--- a/apps/avifgainmaputil/imageio.cc
+++ b/apps/avifgainmaputil/imageio.cc
@@ -101,7 +101,8 @@ void PrintEncodingSettings(const avifEncoder* encoder, bool has_gain_map) {
             << ")], alpha quality ['" << encoder->qualityAlpha << "' ("
             << QualityLevelString(encoder->qualityAlpha) << ")]" << gain_map_str
             << ", "
-            << (encoder->autoTiling ? "automatic tiling" : manual_tiling_str)
+            << (encoder->autoTiling ? "automatic tiling"
+                                    : std::move(manual_tiling_str))
             << ", " << encoder->maxThreads
             << " worker thread(s), please wait...\n";
 }


apps/avifgainmaputil/imageio.cc
```diff
@@ -101,7 +101,8 @@ void PrintEncodingSettings(const avifEncoder* encoder, bool has_gain_map) {
             << ")], alpha quality ['" << encoder->qualityAlpha << "' ("
             << QualityLevelString(encoder->qualityAlpha) << ")]" << gain_map_str
             << ", "
-            << (encoder->autoTiling ? "automatic tiling" : manual_tiling_str)
+            << (encoder->autoTiling ? "automatic tiling"
+                                    : std::move(manual_tiling_str))
```
wantehchang, line 105: Maryla: Thank you for fixing the Coverity defect quickly. I think it is fine to ignore this Coverity defect. If we want to fix this Coverity defect, I suggest also adding `std::move` to `gain_map_str` because it has the same issue.

apps/avifgainmaputil/imageio.cc
wantehchang, line 96: Optional: We can emulate `gain_map_str` and assign a value to `manual_tiling_str` only when it will be used:

```
  std::string manual_tiling_str;
  if (!encoder->autoTiling) {
    manual_tiling_str =
        "tileRowsLog2 [" + std::to_string(encoder->tileRowsLog2) +
        "], tileColsLog2 [" + std::to_string(encoder->tileColsLog2) + "]";
  }
```

apps/avifgainmaputil/imageio.cc
wantehchang: According to https://en.cppreference.com/w/cpp/string/basic_string/operator_ltltgtgt.html, `std::string` is passed by const reference to the `std::cout <<` operator:

```
template< class CharT, class Traits, class Allocator >
std::basic_ostream<CharT, Traits>&
    operator<<( std::basic_ostream<CharT, Traits>& os,
                const std::basic_string<CharT, Traits, Allocator>& str );
```

So this Coverity defect report seems incorrect.

wantehchang: I misunderstood the Coverity defect report. `manual_tiling_str` is passed to the `? :` expression rather than the `std::out <<` operator. So I believe Coverity is correct but what happens in that `? :` expression is beyond my C++ knowledge. In any case, your rewrite in PR https://github.com/AOMediaCodec/libavif/pull/3152 eliminated the issue completely.

PR #3153, opened 2026-03-30
gainmap: guard pixel-count arithmetic overflow

rootvector2:
This PR adds a small safety check in gainmap code to avoid overflow when width and height are very large.
I kept the change minimal and only touched src/gainmap.c.
It should not change normal behavior and it fails early for invalid huge inputs.

### DIFF
--- a/src/gainmap.c
+++ b/src/gainmap.c
@@ -4,6 +4,7 @@
 #include "avif/internal.h"
 #include <assert.h>
 #include <float.h>
+#include <limits.h>
 #include <math.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


src/gainmap.c
```diff
@@ -559,11 +560,11 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
     avifResult res = AVIF_RESULT_OK;
     // --- After this point, the function should exit with 'goto cleanup' to free allocated resources.
 
+    AVIF_CHECKERR(width == 0 || height <= (SIZE_MAX / width), AVIF_RESULT_INVALID_ARGUMENT);
     const size_t numPixels = (size_t)width * height;
-    if (numPixels > SIZE_MAX / sizeof(float)) {
-        res = AVIF_RESULT_INVALID_ARGUMENT;
-        goto cleanup;
-    }
+    AVIF_CHECKERR(numPixels <= (SIZE_MAX / sizeof(float)), AVIF_RESULT_INVALID_ARGUMENT);
```
wantehchang, line 565: As the comment at line 561 noted, after line 561, every error require going to the `cleanup` label for resource cleanup. So we can't use the `AVIF_CHECKERR` macro here.

src/gainmap.c
```diff
@@ -559,11 +560,11 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
...
+    // avifFindMinMaxWithoutOutliers() takes numPixels as int.
```
wantehchang, line 566: This is stale information. In pull requerst https://github.com/AOMediaCodec/libavif/pull/3051 you and I also changed the `numPixels` parameter of `avifFindMinMaxWithoutOutliers()` to `size_t`.

If you analyzed the source code using a tool, please check the configuration of that tool.

src/gainmap.c
```diff
@@ -559,11 +560,11 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
     avifResult res = AVIF_RESULT_OK;
     // --- After this point, the function should exit with 'goto cleanup' to free allocated resources.
 
+    AVIF_CHECKERR(width == 0 || height <= (SIZE_MAX / width), AVIF_RESULT_INVALID_ARGUMENT);
```
wantehchang, line 563: Dexter.k: You and I worked on this function (avifRGBImageComputeGainMap) in pul request https://github.com/AOMediaCodec/libavif/pull/3051. At that time I audited this function thoroughly, and this file has not changed since then.

This check is not necessary here. `width` and `height` are the width and height of `baseRgbImage`. I audited the code and proved that the pixel buffers of `baseRgbImage` have been allocated. Therefore we can safely multiply `width` by `height` using the `size_t` type.

We take advantage of this property to omit unnecessary overflow checks like this one.

Note: I can certainly make mistakes when analyzing code and prove properties. Since you didn't point out a flaw in my analysis, I am going to assume that my analysis was correct.

rootvector2:
Sorry about that I missed that we had already worked on this in the earlier PR. My mistake.

rootvector2:
@wantehchang I’d really appreciate your support. If possible would you consider sponsoring my work?

PR #3154, opened 2026-03-30
Statically assert a strcpy is safe

wantehchang:


src/read.c
```diff
@@ -4524,6 +4524,7 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
         colorItem->premByID = alphaIsPremultiplied;
         avifProperty * alphaAuxProp = avifMetaCreateProperty(meta, "auxC");
         AVIF_CHECKERR(alphaAuxProp, AVIF_RESULT_OUT_OF_MEMORY);
+        static_assert(sizeof(alphaAuxProp->u.auxC.auxType) >= sizeof(AVIF_URN_ALPHA0), "");
```
wantehchang, line 4527: This static assertion is correct whether we copy using `strcpy` or `memcpy`.

The macro `AVIF_URN_ALPHA0` is defined as a string literal:

```
#define AVIF_URN_ALPHA0 "urn:mpeg:mpegB:cicp:systems:auxiliary:alpha"
```

In C, `sizeof` a string literal includes the terminating NUL character.

PR #3156, opened 2026-03-31
Document the invariant on avifROStream

wantehchang:
The invariant is stream->offset <= stream->raw->size.


wantehchang:
@uwezkhan FYI. This is related to your pull request https://github.com/AOMediaCodec/libavif/pull/3148.

PR #3157, opened 2026-03-31
Follow up on aomDiagPrintf()

wantehchang:
Call aom_codec_error() and aom_codec_error_detail() inside aomDiagPrintf().

Call aomDiagPrintf() on encoder function failures.

### DIFF
--- a/src/codec_aom.c
+++ b/src/codec_aom.c
@@ -90,13 +90,15 @@ static void aomCodecDestroyInternal(avifCodec * codec)
     avifFree(codec->internal);
 }
 
-#if defined(AVIF_CODEC_AOM_DECODE)
-
 // Writes a libaom error code and error detail into diagnostics.
-static void aomDiagPrintf(avifDiagnostics * diag, const char * func, const char * error, const char * detail)
+static void aomDiagPrintf(avifDiagnostics * diag, const char * func, const aom_codec_ctx_t * ctx)
 {
-    avifDiagnosticsPrintf(diag, "%s failed: %s: %s", func, error, detail ? detail : "no error detail");
+    const char * error = aom_codec_error(ctx);
+    const char * error_detail = aom_codec_error_detail(ctx);
+    avifDiagnosticsPrintf(diag, "%s failed: %s: %s", func, error, error_detail ? error_detail : "no error detail");
 }
+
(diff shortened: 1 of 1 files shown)


src/codec_aom.c
```diff
@@ -90,14 +90,16 @@ static void aomCodecDestroyInternal(avifCodec * codec)
     avifFree(codec->internal);
 }
 
-#if defined(AVIF_CODEC_AOM_DECODE)
-
 // Writes a libaom error code and error detail into diagnostics.
-static void aomDiagPrintf(avifDiagnostics * diag, const char * func, const char * error, const char * detail)
+static void aomDiagPrintf(avifDiagnostics * diag, const char * func, const aom_codec_ctx_t * ctx)
```
wantehchang, line 94: @rootvector2 Dexter.k: This is the function prototype I had in mind. Also, this function can be used for both libaom decoder and encoder.

src/codec_aom.c
```diff
@@ -1367,11 +1345,7 @@ static avifBool aomCodecEncodeFinish(avifCodec * codec, avifCodecEncodeOutput *
     for (;;) {
         // flush encoder
         if (aom_codec_encode(&codec->internal->encoder, NULL, 0, 1, 0) != AOM_CODEC_OK) {
-            const char * error_detail = aom_codec_error_detail(&codec->internal->encoder);
-            avifDiagnosticsPrintf(codec->diag,
-                                  "aom_codec_encode() with img=NULL failed: %s: %s",
-                                  aom_codec_error(&codec->internal->encoder),
-                                  error_detail ? error_detail : "no error detail");
+            aomDiagPrintf(codec->diag, "aom_codec_encode() with img=NULL", &codec->internal->encoder);
```
wantehchang, line 1352: Note that the function name string is "aom_codec_encode() with img=NULL".

This restores the original diagnostic message:

```
                                  "aom_codec_encode() with img=NULL failed: %s: %s",
                                  aom_codec_error(&codec->internal->encoder),
```

src/codec_aom.c
wantehchang, line 526: This is the only libaom function call whose diagnostic message does not fit the function name pattern of `aomDiagPrintf()`. So we cannot call `aomDiagPrintf()` here.

PR #3159, opened 2026-03-31
Access stream->offset through avifROStreamOffset()

wantehchang:
In src/read.c, change three direct accesses to stream->offset to avifROStreamOffset() calls. This will make it easier to verify the invariant on avifROStream's data size and offset is maintained.


uwezkhan:
Hi @wantehchang, thanks for merging this!

I noticed this change is related to my earlier PR (#3148 – “prevent underflow in stream remaining bytes check”). Could you please tag or mention me in the merged PR for attribution?

Appreciate your work on this!

PR #3160, opened 2026-03-31
add overflow check for width * height in gainmap

uwezkhan:
Added checks to avoid overflow when calculating total pixels in gainmap functions.

The code was doing width * height without checking for overflow.
On 32-bit systems this can wrap around and lead to incorrect memory allocation.

Added checks before multiplication to make sure the values stay within safe limits.
Also added a small guard to avoid division by zero just in case.

Prevents possible memory issues when handling very large or invalid images.

### DIFF
--- a/src/gainmap.c
+++ b/src/gainmap.c
@@ -285,7 +285,11 @@ avifResult avifRGBImageApplyGainMap(const avifRGBImage * baseImage,
 
         // Convert extended SDR (where 1.0 is SDR white) to nits.
         clli->maxCLL = (uint16_t)AVIF_CLAMP(avifRoundf(rgbMaxLinear * SDR_WHITE_NITS), 0.0f, (float)UINT16_MAX);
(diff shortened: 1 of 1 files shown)


src/gainmap.c
```diff
@@ -559,6 +563,9 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
     avifResult res = AVIF_RESULT_OK;
     // --- After this point, the function should exit with 'goto cleanup' to free allocated resources.
 
+    if (width == 0 || height == 0 || height > (SIZE_MAX / width)) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
```
wantehchang, line 568: These two `width * height` multiplications were audited recently.

The `width` and `height` variables in these two functions are the width and height of an `avifRGBImage` struct whose `pixels` buffer has been allocated successfully. So we know `rowBytes * height` fits in `size_t`. Since `width <= rowBytes`, it follows that `width * height` also fits in `size_t`.

PR #3163, opened 2026-04-01
Explicitly forbid reusing encoder instance

### LINKED ISSUE #3129: Decoder can be reused, Encoder can't
Currently, I have a small decoding program that decodes a sequence of given images individually as fast as possible.
The program is designed like this:
```
1. Decoder creation
2. WHILE new_image_in_input_buffer <─┐
3.   |  Set up decoder for new image │
4.   |  Decode image to RGB          │
5.   |  Copy RGB output somewhere    │
6.   |  Cleanup RGB output ──────────┘
7. Decoder destruction
```
This works because the decoder automatically resets when calling it again for the next image.

My problem is that, when wanting to write an equivalent program for encoding, things change. Even when encoding in single image mode with `AVIF_ADD_IMAGE_FLAG_SINGLE` and cleaning up after each iteration with `avifImageDestroy()` and  `avifRWDataFree()`, the encoder still panics after the first iteration when I try to reuse it. 

Now, I think encoding/decoding images in a loop of some kind is a frequent use case for applications wanting to use this library. Either this lack of an automatic reset is intended behavior, in which case I think it should be documented somewhere, or I am doing something wrong and you actually _can_ reuse the encoder without destroying it and constructing it every single time, in which case, again, I think it should be documented how to do it.

y-guyon:
Fixes #3129.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1416,7 +1416,7 @@ AVIF_API avifResult avifDecoderReadFile(avifDecoder * decoder, avifImage * image
 // avifDecoderSetSource() or avifDecoderParse() will automatically Reset the decoder.
 //
 // The decoder must be destroyed once there is no need for further parsing or decoding.
-// The decoder instance cannot be reused for another file. Call avifDecoderCreate() instead.
+// Reusing the decoder instance for another file is not recommended. Call avifDecoderCreate() instead.
 //
 // avifDecoderSetSource() allows you not only to choose whether to parse tracks or
 // items in a file containing both, but switch between sources without having to


include/avif/avif.h
```diff
@@ -1415,6 +1415,9 @@ AVIF_API avifResult avifDecoderReadFile(avifDecoder * decoder, avifImage * image
 // to reset the internal decoder back to before the first frame. Calling either
 // avifDecoderSetSource() or avifDecoderParse() will automatically Reset the decoder.
 //
+// The decoder must be destroyed once there is no need for further parsing or decoding.
+// The decoder instance cannot be reused for another asset. Call avifDecoderCreate() instead.
+//
```
wantehchang, line 1420: We should study exactly what `avifDecoderReset()` does. The name of the function suggests it may allow the decoder instance to be reused.

include/avif/avif.h
```diff
@@ -1415,6 +1415,9 @@ AVIF_API avifResult avifDecoderReadFile(avifDecoder * decoder, avifImage * image
 // to reset the internal decoder back to before the first frame. Calling either
 // avifDecoderSetSource() or avifDecoderParse() will automatically Reset the decoder.
 //
+// The decoder must be destroyed once there is no need for further parsing or decoding.
+// The decoder instance cannot be reused for another asset. Call avifDecoderCreate() instead.
```
wantehchang: Nit: Change "asset" to "image" or some other term that's used in this header or the AVIF spec, or simply omit "for another asset".

include/avif/avif.h
```diff
@@ -1667,6 +1670,9 @@ typedef uint32_t avifAddImageFlags;
 // The image passed to avifEncoderAddImage() or avifEncoderAddImageGrid() is encoded during the
 // call (which may be slow) and can be freed after the function returns.
 //
+// The encoder must be destroyed after avifEncoderFinish() is called.
+// The encoder instance cannot be reused for another asset. Call avifEncoderCreate() instead.
+//
```
wantehchang, line 1677: I think the encoder-side comment is correct.

include/avif/avif.h
wantehchang, line 1670: I think the "Usage / function call order" comment is quite clear. We can consider addinf a blank line after `avifEncoderCreate()` and another blank line before `avifEncoderFinish()`:

```
// * avifEncoderCreate()
//
// - Still image:
//   * avifEncoderAddImage() [exactly once]
// - Still image grid:
//   * avifEncoderAddImageGrid() [exactly once, AVIF_ADD_IMAGE_FLAG_SINGLE is assumed]
// - Image sequence (animation):
//   * Set encoder->timescale (Hz) correctly
//   * avifEncoderAddImage() ... [repeatedly; at least once]
// - Still layered image:
//   * Set encoder->extraLayerCount correctly
//   * avifEncoderAddImage() ... [exactly encoder->extraLayerCount+1 times]
// - Still layered grid:
//   * Set encoder->extraLayerCount correctly
//   * avifEncoderAddImageGrid() ... [exactly encoder->extraLayerCount+1 times]
//
// * avifEncoderFinish()
// * avifEncoderDestroy()
```

include/avif/avif.h
```diff
@@ -1667,6 +1670,9 @@ typedef uint32_t avifAddImageFlags;
 // The image passed to avifEncoderAddImage() or avifEncoderAddImageGrid() is encoded during the
 // call (which may be slow) and can be freed after the function returns.
 //
+// The encoder must be destroyed after avifEncoderFinish() is called.
+// The encoder instance cannot be reused for another asset. Call avifEncoderCreate() instead.
```
wantehchang: Nit: Change "asset" to "image" or some other term that's used in this header or the AVIF spec, or simply omit "for another asset".

y-guyon: Omitted "for another asset".

include/avif/avif.h
y-guyon: Done.

include/avif/avif.h
y-guyon: I cannot omit "for another asset" because the decoder instance can be "reused" to decode the same image or sequence multiple times, by calling `avifDecoderReset()`.

Used "file" instead.

include/avif/avif.h
y-guyon: Hints toward NOT reusing the decoder instance:

```c
// Data passed into avifDecoderParse() is NOT
// copied, so it must continue to exist until the decoder is destroyed.
```

This sentence does not mention `avifDecoderReset()`, so data would need to continue to exist across multiple assets. It makes little sense.

```c
// You can use avifDecoderReset() any time after a successful call to avifDecoderParse()
// to reset the internal decoder back to before the first frame.
```

To me, "before the first frame" is implicitly followed with "of the same asset".

wantehchang: Please see my findings at issue https://github.com/AOMediaCodec/libavif/issues/3129#issuecomment-4181079818. I suggest that we
- say nothing about decoder reuse to avoid encouraging it, and
- change the title of the commit to "Explicitly forbid reusing encoder instance".

Alternatively, we can change the last sentence to a recommendation:

```
// We recommend not reusing the decoder instance for another file. Call avifDecoderCreate() instead.
```

wantehchang (approved):

include/avif/avif.h
y-guyon: Done.

PR #3164, opened 2026-04-01
fix overflow in libyuv allocation

uwezkhan:
fix overflow when calculating allocation size in libyuv path by adding proper checks before multiplication

prevents wrong memory allocation for large or invalid inputs,

### DIFF
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -178,7 +178,17 @@ static int avifReorderARGBThenConvertToYUV(int (*ReorderARGB)(const uint8_t *, i
         // allocating more than soft_allocation_limit, unless min_num_rows rows need more than that.
         num_allocated_rows = AVIF_MAX(1, soft_allocation_limit / (src_stride_argb * min_num_rows)) * min_num_rows;
     }
-    src_argb = (uint8_t *)avifAlloc(num_allocated_rows * src_stride_argb);
+    if (src_stride_argb < 0) {
+        return -1;
+    }
+
+    // Check for overflow before multiplication
+    if (num_allocated_rows > 0 && (size_t)num_allocated_rows > SIZE_MAX / (size_t)src_stride_argb) {
+        return -1;
+    }
+
+    const size_t allocationSize = (size_t)num_allocated_rows * (size_t)src_stride_argb;
+    src_argb = (uint8_t *)avifAlloc(allocationSize);
     if (!src_argb) {
         return -1;
     }


uwezkhan:
Any update on this PR

src/reformat_libyuv.c
```diff
@@ -178,7 +178,17 @@ static int avifReorderARGBThenConvertToYUV(int (*ReorderARGB)(const uint8_t *, i
         // allocating more than soft_allocation_limit, unless min_num_rows rows need more than that.
         num_allocated_rows = AVIF_MAX(1, soft_allocation_limit / (src_stride_argb * min_num_rows)) * min_num_rows;
     }
-    src_argb = (uint8_t *)avifAlloc(num_allocated_rows * src_stride_argb);
+    if (src_stride_argb < 0) {
+        return -1;
+    }
+
+    // Check for overflow before multiplication
+    if (num_allocated_rows > 0 && (size_t)num_allocated_rows > SIZE_MAX / (size_t)src_stride_argb) {
```
vrabaud, line 186: if (src_stride_argb > 0

src/reformat_libyuv.c
```diff
@@ -178,7 +178,17 @@ static int avifReorderARGBThenConvertToYUV(int (*ReorderARGB)(const uint8_t *, i
         // allocating more than soft_allocation_limit, unless min_num_rows rows need more than that.
         num_allocated_rows = AVIF_MAX(1, soft_allocation_limit / (src_stride_argb * min_num_rows)) * min_num_rows;
     }
-    src_argb = (uint8_t *)avifAlloc(num_allocated_rows * src_stride_argb);
+    if (src_stride_argb < 0) {
```
wantehchang, line 181: Code analysis shows that the caller of this function passes the `width` field of `avifImage` as the `width` parameter to this function, and the caller checks if `image->width > INT_MAX`. So the `width` parameter cannot be negative, which implies `src_stride_argb` cannot be negative.

Note: The caller is `avifImageRGBToYUVLibYUV()` -> `avifImageRGBToYUVLibYUV8bpc()`.

src/reformat_libyuv.c
wantehchang: The code above (lines 168-180) ensures that `num_allocated_rows * src_stride_argb` is either equal to `src_stride_argb` or is less than or equal to `soft_allocation_limit`. Therefore `num_allocated_rows * src_stride_argb` fits in the `size_t` type.

PR #3178, opened 2026-04-19
Android JNI: reject invalid length at direct buffer entry points

### LINKED ISSUE #3177: [Bug] High: Unchecked signed Android JNI length causes parser OOB read
Priority: High

Affected component: Android JNI decoder wrapper (`android_jni/avifandroidjni`), specifically the public `AvifDecoder.getInfo(ByteBuffer, int, Info)` and `AvifDecoder.decode(ByteBuffer, int, Bitmap, int)` entry points.

Root cause:
The Java API accepts a caller-supplied signed `int length`, and the JNI layer forwards it unchecked into `CreateDecoderAndParse()` and then `avifDecoderSetIOMemory()`, whose `size` parameter is `size_t`.

That signed-to-unsigned boundary lets negative or oversized lengths advertise more bytes than the direct `ByteBuffer` actually holds. The parser then trusts the inflated size and reads past the heap allocation.

Relevant file paths:
- `android_jni/avifandroidjni/src/main/java/org/aomedia/avif/android/AvifDecoder.java`
- `android_jni/avifandroidjni/src/main/jni/libavif_jni.cc`
- `src/read.c`
- `src/stream.c`

Exact reproduction steps:

Android/JNI surface:
1. Build current `main` (`b357f08`) with AddressSanitizer enabled for the Android JNI target.
2. Use an 8-byte direct buffer containing a truncated `ftyp` box header:

```java
ByteBuffer buf = ByteBuffer.allocateDirect(8);
buf.putInt(16);
buf.put((byte) 'f');
buf.put((byte) 't');
buf.put((byte) 'y');
buf.put((byte) 'p');
buf.flip();

org.aomedia.avif.android.AvifDecoder.Info info =
    new org.aomedia.avif.android.AvifDecoder.Info();
org.aomedia.avif.android.AvifDecoder.getInfo(buf, -1, info);
```

3. Observe an ASan `heap-buffer-overflow` in the parser path.
4. As a control, pass `length == 8`; that returns truncated-data behavior instead of crashing.

Verified native-boundary replay used during confirmation:
1. Run a small ASan-enabled harness that mirrors `CreateDecoderAndParse()` after `GetDirectBufferAddress()` with the same 8-byte payload and an advertised length of `-1`.
2. The verification used this exact command:

```sh
timeout --preserve-status 180s \
  env ASAN_OPTIONS=detect_leaks=0:halt_on_error=1 \
      UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1 \
      ASAN_SYMBOLIZER_PATH="$(command -v llvm-symbolizer || true)" \
  build-asan/tr1_neg_length_parse \
  fuzz/corpus/tr1_setiomemory/tiny_ftyp.avif -1
```

3. ASan reports:

```text
ERROR: AddressSanitizer: heap-buffer-overflow
#1 avifROStreamRead src/stream.c:74
#2 avifParseFileTypeBox src/read.c:4780
#3 avifParse src/read.c:4909
#4 avifDecoderParse src/read.c:5314
```

4. The same crash reproduces with an advertised length of `0x80000000`.
5. The same payload with an honest advertised size of `8` does not crash; it returns `Truncated data`.

Suggested fix direction:
- In `libavif_jni.cc`, reject `length < 0` before any cast.
- Use `GetDirectBufferCapacity(encoded)` and require `capacity >= 0` and `length <= capacity` before calling `CreateDecoderAndParse()` / `avifDecoderSetIOMemory()`.
- Cast to `size_t` only after the bounds checks pass.
- Apply the same validation consistently to every JNI entry point that consumes a direct buffer and explicit length.
- Add a regression test that passes a direct buffer with `length = -1` (and another oversized positive value such as `0x80000000`) and verifies a clean failure instead of a parser crash.

parasol-aser:
## Summary

Fixes #3177.

The public Android JNI decoder wrapper (`android_jni/avifandroidjni/src/main/jni/libavif_jni.cc`) accepts a caller-supplied signed `int length` at every direct-`ByteBuffer` entry point and forwards it, unchecked, into the native `size_t size` boundary of `avifDecoderSetIOMemory()`. A negative length (e.g. `-1`) or a negative-as-int oversize (e.g. `0x80000000`) sign-extends to a near-`SIZE_MAX` value; the parser then trusts the inflated advertised size and reads past the real direct-buffer allocation. ASan reports a `heap-buffer-overflow` in `avifROStreamRead` (see the stack in the issue).

This PR enforces the contract at the JNI boundary — where the signed `int` enters — so that every downstream caller of `avifDecoderSetIOMemory()` stays free to pass any `size_t` it likes.

### Changes

- **New `ValidateDirectBuffer()` helper** in the `libavif_jni.cc` anonymous namespace that:
  - Rejects `length < 0` before any cast.
  - Reads `GetDirectBufferCapacity(encoded)` once and rejects non-direct buffers (`capacity < 0`) and `length > capacity` (strict `>`).
  - Rejects a `nullptr` address from `GetDirectBufferAddress()`.
  - Casts to `size_t` only after the checks pass; returns the validated `(buffer, size_t size)` pair.
- **Hardened all four public JNI entry points** that consume a direct `ByteBuffer` and an explicit length: `isAvifImage`, `getInfo`, `decode`, and `createDecoder`. Each now early-exits with its clean-failure value (`false` / `0`) when validation fails.
- **`CreateDecoderAndParse()` signature** changed from `int length` to `size_t length` so the signed-to-unsigned sign-extension is gone from the internal helper too. All four call sites updated.
- **New Android instrumentation regression test** `AvifDecoderLengthValidationTest` covering:
  - The two issue-named boundary values (`length = -1`, `length = Integer.MIN_VALUE`).
  - `length = capacity + 1` and `length = Integer.MAX_VALUE` (positive-but-oversized).
  - Non-direct (heap-backed) buffers.
  - Boundary cases that must still succeed: `length == capacity`, `length == 0`, and a valid image through `AvifDecoder.create()` / `AvifDecoder.getInfo()`.
  - Every case verifies the clean-failure contract (`false` / `null`) instead of a crash.

### Scope and non-goals

- `src/read.c`, `src/stream.c` — **unchanged**. The crash manifests there, but the parser is trusting an already-bad size handed to it. Fixing it lower down would silently paper over other callers of `avifDecoderSetIOMemory()` that are free to pass a large `size_t` deliberately.
- `AvifDecoder.java` public API surface — **unchanged**. The issue's secondary suggestion of deprecating the `int length` overloads in favor of `encoded.remaining()`-driven variants is deliberately not taken here; it would be a source-incompatible change, and the issue frames it as secondary hardening. Happy to do that as a follow-up if maintainers want it.
- No new linker deps, no new JNI reflection lookups, no C++ standard bump.

## Test plan

- [x] **Repro confirmed pre-fix.** Against `b357f08` (current `main`), the 8-byte truncated-`ftyp` payload with `length = -1` or `length = 0x80000000` produces the ASan `heap-buffer-overflow` at `avifROStreamRead` → `avifParseFileTypeBox` → `avifParse` → `avifDecoderParse` (matches the issue).
- [x] **Repro cleared post-fix at the JNI boundary.** With `ValidateDirectBuffer()` in place, the same inputs through `AvifDecoder.getInfo` / `decode` / `isAvifImage` / `create` return the clean-failure value (`false` / `null`) without entering the parser. Verified with a simulated-JNI-validation harness that mirrors the new helper before calling `avifDecoderSetIOMemory()`.
- [x] **Parser contract unchanged.** Calling `avifDecoderSetIOMemory()` directly with `(size_t)-1` still crashes as before — expected; the fix is at the JNI boundary, not in the parser. This is intentionally not a behavior change for non-JNI callers.
- [ ] **Android instrumentation suite.** Run `AvifDecoderLengthValidationTest` on an Android emulator/device, e.g.:
    ```sh
    cd android_jni && ./gradlew :avifandroidjni:connectedDebugAndroidTest \
      --tests org.aomedia.avif.android.AvifDecoderLengthValidationTest
    ```
    Every test case must pass. I don't have an Android device/emulator in this environment to run the instrumentation suite myself; maintainers with CI access should confirm. The tests reuse the existing `avif/fox.profile0.8bpc.yuv420.avif` asset already shipped under `androidTest/assets` for the happy-path cases.
- [ ] **Build hygiene.** `./android_jni/gradlew :avifandroidjni:assembleDebug` must continue to link cleanly after the `CreateDecoderAndParse` signature change. `-Wsign-compare` / `-Wconversion` remain clean across the edited block (the helper uses an explicit `static_cast<jlong>(length)` for the capacity comparison).

### Boundary behavior verified by the test matrix

| Entry point | `length` | Expected |
|---|---|---|
| `getInfo` | `-1` | `false`, no crash |
| `getInfo` | `Integer.MIN_VALUE` | `false`, no crash |
| `getInfo` | `capacity + 1` (9) | `false`, no crash |
| `getInfo` | `Integer.MAX_VALUE` | `false`, no crash |
| `getInfo` | `capacity` (8, truncated ftyp) | `false` (parser — truncated data control) |
| `getInfo` | `0` (empty direct buffer) | `false` (parser — not short-circuited at JNI) |
| `getInfo` | valid image, honest length | `true` (happy path guard) |
| `decode` | `-1` / `MIN_VALUE` / `cap+1` / `MAX_VALUE` | `false`, no crash |
| `isAvifImage` | truncated ftyp / heap-backed / empty | `false`, no crash |
| `create` | truncated / heap-backed / empty | `null` |
| `create` | valid image | non-`null` (happy path guard) |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
+++ b/android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
@@ -50,8 +50,39 @@ struct AvifDecoderWrapper {
   avifCropRect crop;
 };
 
+// Returns true when `encoded` is a direct ByteBuffer of at least `length`
+// bytes and `length` is non-negative. On success, `*out_buffer` is set to the
+// direct buffer address and `*out_size` is set to `(size_t)length`. On
+// failure, logs via LOGE() and returns false; callers should propagate a
+// clean failure (false / 0 / nullptr) to the Java layer.
+bool ValidateDirectBuffer(JNIEnv* env, jobject encoded, jint length,
+                          const uint8_t** out_buffer, size_t* out_size) {
+  if (length < 0) {
+    LOGE("AVIF JNI: negative length (%d) rejected.", length);
+    return false;
+  }
+  const jlong capacity = env->GetDirectBufferCapacity(encoded);
+  if (capacity < 0) {
+    LOGE("AVIF JNI: encoded is not a direct ByteBuffer.");
+    return false;
+  }
+  if (static_cast<jlong>(length) > capacity) {
+    LOGE("AVIF JNI: length (%d) exceeds direct buffer capacity (%lld).",
+         length, static_cast<long long>(capacity));
+    return false;
+  }
+  const void* const address = env->GetDirectBufferAddress(encoded);
+  if (address == nullptr) {
+    LOGE("AVIF JNI: GetDirectBufferAddress returned null.");
+    return false;
+  }
+  *out_buffer = static_cast<const uint8_t*>(address);
+  *out_size = static_cast<size_t>(length);
+  return true;
+}
+
 bool CreateDecoderAndParse(AvifDecoderWrapper* const decoder,
-                           const uint8_t* const buffer, int length,
+                           const uint8_t* const buffer, size_t length,
                            int threads) {
   decoder->decoder = avifDecoderCreate();
   if (decoder->decoder == nullptr) {
@@ -255,9 +286,12 @@ jint JNI_OnLoad(JavaVM* vm, void* /*reserved*/) {
 
 FUNC(jboolean, isAvifImage, jobject encoded, int length) {
   IGNORE_UNUSED_JNI_PARAMETERS;
-  const uint8_t* const buffer =
-      static_cast<const uint8_t*>(env->GetDirectBufferAddress(encoded));
-  const avifROData avif = {buffer, static_cast<size_t>(length)};
+  const uint8_t* buffer = nullptr;
+  size_t size = 0;
+  if (!ValidateDirectBuffer(env, encoded, length, &buffer, &size)) {
+    return false;
+  }
+  const avifROData avif = {buffer, size};
   return avifPeekCompatibleFileType(&avif);
 }
 
@@ -278,10 +312,13 @@ FUNC(jboolean, isAvifImage, jobject encoded, int length) {
 
 FUNC(jboolean, getInfo, jobject encoded, int length, jobject info) {
   IGNORE_UNUSED_JNI_PARAMETERS;
-  const uint8_t* const buffer =
-      static_cast<const uint8_t*>(env->GetDirectBufferAddress(encoded));
+  const uint8_t* buffer = nullptr;
+  size_t size = 0;
+  if (!ValidateDirectBuffer(env, encoded, length, &buffer, &size)) {
+    return false;
+  }
   AvifDecoderWrapper decoder;
-  if (!CreateDecoderAndParse(&decoder, buffer, length, /*threads=*/1)) {
+  if (!CreateDecoderAndParse(&decoder, buffer, size, /*threads=*/1)) {
     return false;
   }
   FIND_CLASS(info_class, "org/aomedia/avif/android/AvifDecoder$Info", false);
@@ -303,10 +340,13 @@ FUNC(jboolean, getInfo, jobject encoded, int length, jobject info) {
 FUNC(jboolean, decode, jobject encoded, int length, jobject bitmap,
      jint threads) {
   IGNORE_UNUSED_JNI_PARAMETERS;
-  const uint8_t* const buffer =
-      static_cast<const uint8_t*>(env->GetDirectBufferAddress(encoded));
+  const uint8_t* buffer = nullptr;
+  size_t size = 0;
+  if (!ValidateDirectBuffer(env, encoded, length, &buffer, &size)) {
+    return false;
+  }
   AvifDecoderWrapper decoder;
-  if (!CreateDecoderAndParse(&decoder, buffer, length,
+  if (!CreateDecoderAndParse(&decoder, buffer, size,
                              getThreadCount(threads))) {
     return false;
   }
@@ -314,14 +354,17 @@ FUNC(jboolean, decode, jobject encoded, int length, jobject bitmap,
 }
 
 FUNC(jlong, createDecoder, jobject encoded, jint length, jint threads) {
-  const uint8_t* const buffer =
-      static_cast<const uint8_t*>(env->GetDirectBufferAddress(encoded));
+  const uint8_t* buffer = nullptr;
+  size_t size = 0;
+  if (!ValidateDirectBuffer(env, encoded, length, &buffer, &size)) {
+    return 0;
+  }
   std::unique_ptr<AvifDecoderWrapper> decoder(new (std::nothrow)
                                                   AvifDecoderWrapper());
   if (decoder == nullptr) {
     return 0;
   }
-  if (!CreateDecoderAndParse(decoder.get(), buffer, length,
+  if (!CreateDecoderAndParse(decoder.get(), buffer, size,
                              getThreadCount(threads))) {
     return 0;
   }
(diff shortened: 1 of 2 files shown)


vigneshvg (approved):

wantehchang:
parasol-aser: Thank you for the bug report and the pull request. A weakness of the pull request is that it compares the `length` parameter with the `capacity()` of the buffer. We should compare `length` with the `limit()` of the buffer, but it is more complicated to call the `limit()` method in JNI.

Vignesh: I found that all the callers of the native methods that take `ByteBuffer encoded` pass `encoded.remaining()` as the `int length` parameter. In addition, these native methods apparently assume that `encoded.position()` is equal to 0 (this assumption is documented for some but not all of these native methods).

Ideally we should remove the `int length` parameter from these native methods and just use `env->GetDirectBufferCapacity(encoded)`. Since some of these native methods are public, we want to avoid changing their function prototypes. It would be better to first check the following two assumptions in Java before calling the native methods, and make all the native methods private:
- `encoded.position() == 0`. Note that this implies `encoded.remaining() == encoded.limit()`.
- `encoded.limit() == encoded.capacity()`

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -50,8 +50,39 @@ struct AvifDecoderWrapper {
...
+// direct buffer address and `*out_size` is set to `(size_t)length`. On
+// failure, logs via LOGE() and returns false; callers should propagate a
+// clean failure (false / 0 / nullptr) to the Java layer.
+bool ValidateDirectBuffer(JNIEnv* env, jobject encoded, jint length,
+                          const uint8_t** out_buffer, size_t* out_size) {
+  if (length < 0) {
+    LOGE("AVIF JNI: negative length (%d) rejected.", length);
+    return false;
+  }
+  const jlong capacity = env->GetDirectBufferCapacity(encoded);
```
wantehchang, line 64: According to [JNI documentation](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#GetDirectBufferCapacity), `GetDirectBufferCapacity()` returns -1 if if the given object is not a direct `java.nio.Buffer`. So the new `capacity < 0` check below is correct unless we know `encoded` must be a direct buffer.

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
```diff
@@ -50,8 +50,39 @@ struct AvifDecoderWrapper {
...
+  if (capacity < 0) {
+    LOGE("AVIF JNI: encoded is not a direct ByteBuffer.");
+    return false;
+  }
+  if (static_cast<jlong>(length) > capacity) {
+    LOGE("AVIF JNI: length (%d) exceeds direct buffer capacity (%lld).",
+         length, static_cast<long long>(capacity));
+    return false;
+  }
+  const void* const address = env->GetDirectBufferAddress(encoded);
```
wantehchang, line 74: From the [JNI documentation](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#GetDirectBufferAddress), it is less clear whether the `address == nullptr` check below is needed, given that `GetDirectBufferCapacity()` succeeded. But we can keep this check.

Issue #2849, opened 2025-07-20
`--depth` parameter silently ignored when using Y4M input files


rachelplusplus:
Currently, when `avifenc` is run with an input file in .y4m format, the colour depth used in the resulting AVIF file is always the same as in the original Y4M file. This currently overrides any explicit bit depth specified on the command line.

While this behaviour is documented, it is non-intuitive. In addition, the `--depth` parameter is silently ignored, rather than printing an appropriate warning.

Ideally, if the input bit depth and `--depth` parameter differ, the input should be converted to the requested precision before compression. This can help squeeze a bit more out of the format: usually, even if the input file uses an 8-bit depth, compressing at a 10-bit internal precision gains a few % better compression at a negligible time cost, due to a reduction in rounding error within the codec.

If this is done, I would also like to request that avifenc normalize "unknown" colour space components to definite values. Otherwise, the colour space inferred by downstream tools can differ between 8-bit and 10-bit files, leading to the wrong colours being displayed.

wantehchang:
Rachel: Thank you for the bug report. I will first make `avifenc` fail with an error if the .y4m input bit depth and the `--depth` parameter differ.

Could you clarify why the colour space inferred by downstream tools can differ between 8-bit and 10-bit AVIF files with "unknown" colour space components?

juliobbv-p:
> I will first make avifenc fail with an error if the .y4m input bit depth and the --depth parameter differ.

Another option to consider: we could treat y4ms like they were pngs, and treat `--depth` as if it were the "output" depth, so an 8-bit y4m can be up-converted to a 10-bit AVIF with `--depth 10`. y4ms are closer to pngs in usability than to raw yuvs, as all the width/height/bit depth info is included in them. Aligning y4m/png behavior feels like the least astonishing behavior.

Up-converting content from 8 to 10 bits can be noticeably beneficial for efficiency, as shown by the following comparison by Rachel herself: https://www.rachelplusplus.me.uk/blog/2025/07/a-better-image-compression-comparison/.

PR #3179, opened 2026-04-27
avifenc: reject mismatched --depth for Y4M input

kjg0724:
Reject mismatched `--depth` values for Y4M input instead of silently ignoring them.

Y4M inputs carry their bit depth in the stream header, and avifenc currently retains that depth. If a different `--depth` is specified, fail with a clear error because Y4M bit-depth conversion is not supported. Bit-depth extension is also rejected for Y4M input.

This covers both file input and `--stdin --input-format y4m`.

Refs #2849

Tested:
- cmake --build build/local -j 8
- ctest --test-dir build/local --output-on-failure
- bash tests/test_cmd_stdin.sh "" /Users/solario/Projects/Solido/libavif/build/local /Users/solario/Projects/Solido/libavif/tests/data /tmp/libavif-test-stdin

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -222,7 +222,7 @@ static void syntaxLong(void)
     printf("    --mini                            : EXPERIMENTAL: Use reduced header if possible (backward-incompatible)\n");
 #endif
     printf("    -l,--lossless                     : Set all defaults to encode losslessly, and emit warnings when settings/input don't allow for it\n");
-    printf("    -d,--depth D[,Dextension]         : D is the output bit depth per channel. D must be 8, 10 or 12. (JPEG/PNG only; y4m or stdin: bit depth is retained)\n");
+    printf("    -d,--depth D[,Dextension]         : D is the output bit depth per channel. D must be 8, 10 or 12. (JPEG/PNG only; y4m or stdin: bit depth is retained and must match if specified; Dextension is unsupported)\n");
     printf("                                        If specified, Dextension adds a hidden encoded image of Dextension bit depth in the same file as the primary image to reach 16-bit depth at decoding.\n");
     printf("                                        See avifSampleTransformRecipe for the supported combinations (8,8 and 12,4 and 12,8).\n");
     printf("    -y,--yuv FORMAT                   : Output format, one of 'auto' (default), 444, 422, 420 or 400. Ignored for y4m or stdin (y4m format is retained)\n");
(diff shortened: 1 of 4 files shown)


vigneshvg (changes requested):
Thank you for the PR. It seems like there are some presubmit failures (clang format), could you please fix them?

vigneshvg (approved):
thank you.

apps/avifenc.c
```diff
@@ -633,6 +634,19 @@ static avifBool avifInputReadImage(avifInput * input,
...
+    if (actualInputFormat == AVIF_APP_FILE_FORMAT_Y4M) {
+        if (input->requestedDepthExtension) {
+            fprintf(stderr, "ERROR: --depth %d,%d is not supported for Y4M input\n", input->requestedDepth, input->requestedDepthExtension);
+            return AVIF_FALSE;
+        }
+        if (input->requestedDepth && (input->requestedDepth != (int)dstImage->depth)) {
+            fprintf(stderr,
+                    "ERROR: --depth %d does not match Y4M bit depth %u; Y4M bit-depth conversion is not supported\n",
+                    input->requestedDepth,
+                    dstImage->depth);
```
wantehchang, line 646: cc: @maryla-uc

@kjg0724: The condition for this error message does not seem to map your change to the help message at line 225, reproduced below:

```
    printf("    -d,--depth D[,Dextension]         : D is the output bit depth per channel. D must be 8, 10 or 12. (JPEG/PNG only; y4m or stdin: bit depth is retained and must match if specified; Dextension is unsupported)\n");
```

The help message says "y4m or stdin", but the corresponding condition at line 637 is `actualInputFormat == AVIF_APP_FILE_FORMAT_Y4M`, which does not seem to consider stdin.

Should we change "y4m or stdin" to "y4m" in the help message at line 225 as follows?

```
    printf("    -d,--depth D[,Dextension]         : D is the output bit depth per channel. D must be 8, 10 or 12. (JPEG/PNG only; y4m: bit depth is retained and must match if specified; Dextension is unsupported)\n");
```

apps/avifenc.c
wantehchang, line 683: @kjg0724: Note that `inputFormat` is used here, which may be `AVIF_APP_FILE_FORMAT_UNKNOWN`. Should we also change `inputFormat` to `actualInputFormat` here, as you did at line 660?

PR #3182, opened 2026-04-30
Update comment for avifDecoderItem::mergedExtents

wantehchang:
The mergedExtents field of the avifDecoderItem struct is now always used. However, it may or may not own the buffer it points to, depending on the value of the ownsMergedExtents field.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -222,8 +222,10 @@ typedef struct avifDecoderItem
     avifContentType contentType;
     avifPropertyArray properties;
     avifExtentArray extents;       // All extent offsets/sizes
-    avifRWData mergedExtents;      // if set, is a single contiguous block of this item's extents (unused when extents.count == 1)
-    avifBool ownsMergedExtents;    // if true, mergedExtents must be freed when this item is destroyed
+    avifRWData mergedExtents;      // A single contiguous block of this item's extents
+    avifBool ownsMergedExtents;    // If true, mergedExtents must be freed when this item is destroyed.
+                                   // If false, mergedExtents is used as an avifROData and points to a
+                                   // buffer it doesn't own.
     avifBool partialMergedExtents; // If true, mergedExtents doesn't have all of the item data yet
     uint32_t thumbnailForID;       // if non-zero, this item is a thumbnail for Item #{thumbnailForID}
     uint32_t auxForID;             // if non-zero, this item is an auxC plane for Item #{auxForID}


src/read.c
```diff
@@ -222,8 +222,10 @@ typedef struct avifDecoderItem
     avifContentType contentType;
     avifPropertyArray properties;
     avifExtentArray extents;       // All extent offsets/sizes
-    avifRWData mergedExtents;      // if set, is a single contiguous block of this item's extents (unused when extents.count == 1)
-    avifBool ownsMergedExtents;    // if true, mergedExtents must be freed when this item is destroyed
+    avifRWData mergedExtents;      // A single contiguous block of this item's extents
```
wantehchang, line 225: I noticed the comment for `mergedExtents` is out of date while researching whether https://b.corp.google.com/issues/506387278 affects libavif.

src/read.c
```diff
@@ -222,8 +222,10 @@ typedef struct avifDecoderItem
...
+    avifBool ownsMergedExtents;    // If true, mergedExtents must be freed when this item is destroyed.
+                                   // If false, mergedExtents is used as an avifROData and points to a
+                                   // buffer it doesn't own.
```
wantehchang, line 228: The following code is related to the ownsMergedExtents=true case:

```
    // If there is a single extent for this item and the source of the read buffer is going to be
    // persistent for the lifetime of the avifDecoder (whether it comes from its own internal
    // idatBuffer or from a known-persistent IO), we can avoid buffer duplication and just use the
    // preexisting buffer.
    avifBool singlePersistentBuffer = ((item->extents.count == 1) && (idatBuffer || io->persistent));
    if (!singlePersistentBuffer) {
        // Always allocate the item's full size here, as progressive image decodes will do partial
        // reads into this buffer and begin feeding the buffer to the underlying AV1 decoder, but
        // will then write more into this buffer without flushing the AV1 decoder (which is still
        // holding the address of the previous allocation of this buffer). This strategy avoids
        // use-after-free issues in the AV1 decoder and unnecessary reallocs as a typical
        // progressive decode use case will eventually decode the final layer anyway.
        AVIF_CHECKRES(avifRWDataRealloc(&item->mergedExtents, item->size));
        item->ownsMergedExtents = AVIF_TRUE;
    }
```

The following code is related to the ownsMergedExtents=false case. Note that offsetBuffer is an avifROData:

```
        avifROData offsetBuffer;
        ...

        if (singlePersistentBuffer) {
            memcpy(&item->mergedExtents, &offsetBuffer, sizeof(avifRWData));
            item->mergedExtents.size = bytesToRead;
        } else {
```

wantehchang: The use of `memcpy` in this code is a little confusing. I will write a pull request to clean that up.

vigneshvg (approved):

PR #3184, opened 2026-05-01
Declare rgbPixelBytes and grayPixelBytes as size_t

wantehchang:
In avifImageRGBToYUV(), declare the local variables rgbPixelBytes and grayPixelBytes as size_t and replace the constant 2 with sizeof(uint16_t) so that multiplications with them are performed in the size_t type and avoid integer overflows.

Based on the changes to avifImageRGBToYUV() in the first two commits in the pull request https://github.com/AOMediaCodec/libavif/pull/3072.

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -279,7 +279,7 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
 
         struct YUVBlock yuvBlock[2][2];
         float rgbPixel[3];
-        const uint32_t rgbPixelBytes = state.rgb.pixelBytes;
+        const size_t rgbPixelBytes = state.rgb.pixelBytes;
         const uint32_t offsetBytesR = state.rgb.offsetBytesR;
         const uint32_t offsetBytesG = state.rgb.offsetBytesG;
         const uint32_t offsetBytesB = state.rgb.offsetBytesB;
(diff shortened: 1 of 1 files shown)


src/reformat.c
```diff
@@ -279,7 +279,7 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
 
         struct YUVBlock yuvBlock[2][2];
         float rgbPixel[3];
-        const uint32_t rgbPixelBytes = state.rgb.pixelBytes;
+        const size_t rgbPixelBytes = state.rgb.pixelBytes;
```
wantehchang, line 282: Vincent: This PR is no longer necessary now that pull request https://github.com/AOMediaCodec/libavif/pull/3072 has been updated to declare the index variables `i`, `j`, `uvI`, `uvJ` as `size_t`. But I think it is still good to declare `rgbPixelBytes` and `grayPixelBytes` as `size_t` because the related `*RowBytes` variables are declared as `size_t`.

src/reformat.c
```diff
@@ -387,13 +387,13 @@ avifResult avifImageRGBToYUV(avifImage * image, const avifRGBImage * rgb)
                         }
 
                         if (state.yuv.channelBytes > 1) {
-                            uint16_t * pY = (uint16_t *)&yPlane[(i * 2) + (j * yRowBytes)];
+                            uint16_t * pY = (uint16_t *)&yPlane[(i * sizeof(uint16_t)) + (j * yRowBytes)];
```
wantehchang, line 390: However, if you think changing `2` to `sizeof(uint16_t)` is too verbose, I can revert these changes.

PR #3187, opened 2026-05-02
Add defensive limits for BMFF metadata counts during parsing

uwezkhan:
This patch adds internal defensive limits for attacker-controlled BMFF metadata counts parsed from AVIF/HEIF files.

Several parser paths allocate memory or perform repeated processing based on counts read directly from the bitstream (items, properties, extents, groups, etc.). Malformed files can use excessive values here to trigger disproportionate memory usage and parser overhead.

To mitigate this, this change introduces conservative internal upper bounds for key metadata structures and rejects files that exceed those limits during parsing.

The limits are intentionally generous for legitimate AVIF content while preventing pathological resource-exhaustion cases from malformed inputs.


src/read.c
```diff
@@ -38,6 +38,12 @@ static const size_t VISUALSAMPLEENTRY_SIZE = 78;
 // can't be more than 4 unique tuples right now.
 #define MAX_IPMA_VERSION_AND_FLAGS_SEEN 4
 
+// Security limits for BMFF structures to prevent resource exhaustion (DoS).
+#define AVIF_MAX_ITEM_COUNT 4096
+#define AVIF_MAX_PROPERTY_COUNT 4096
+#define AVIF_MAX_EXTENT_COUNT 64
+#define AVIF_MAX_GROUP_COUNT 1024
```
y-guyon: I would have expected dynamic limits, to be set at runtime like `avifDecoder::imageDimensionLimit`.

src/read.c
```diff
@@ -4945,7 +5006,8 @@ static avifResult avifParse(avifDecoder * decoder)
 #if defined(AVIF_ENABLE_EXPERIMENTAL_MINI)
             AVIF_CHECKERR(!miniSeen, AVIF_RESULT_BMFF_PARSE_FAILED);
 #endif
-            AVIF_CHECKRES(avifParseMetaBox(data->meta, boxOffset, boxContents.data, boxContents.size, data->diag));
+            avifMeta * meta = data->meta;
+            AVIF_CHECKRES(avifParseMetaBox(meta, boxOffset, boxContents.data, boxContents.size, data->diag));
```
y-guyon: This change looks unrelated.

src/read.c
uwezkhan: Thanks for the feedback — that makes sense.

I started with fixed constants to keep the initial patch minimal, but I agree runtime-configurable limits are a better fit and more consistent with existing decoder limits like imageDimensionLimit.

I’ll rework this accordingly.

src/read.c
uwezkhan: that change was unrelated cleanup and should not have been included in this PR.

I’ll remove it in the next update.

include/avif/avif.h
```diff
@@ -1383,6 +1395,23 @@ typedef struct avifDecoder
...
+    // The default is AVIF_DEFAULT_PROPERTY_COUNT_LIMIT, and setting this to 0 disables the limit.
+    uint32_t propertyCountLimit;
+
+    // This provides an upper bound on how many extents an item can have.
+    // The default is AVIF_DEFAULT_EXTENT_COUNT_LIMIT, and setting this to 0 disables the limit.
+    uint32_t extentCountLimit;
+
+    // This provides an upper bound on how many groups can be created in a grpl box.
+    // The default is AVIF_DEFAULT_GROUP_COUNT_LIMIT, and setting this to 0 disables the limit.
+    uint32_t groupCountLimit;
```
wantehchang: It is a little excessive to add four decoder options that most users of libavif are unlikely to understand.

Also, the libavif code related to these four limits grows an array one element at a time. So the input file must have as many objects to cause the array size to grow. It is more important to defend against a small malicious file that causes libavif to allocate a lot of memory, i.e., what your description refers to as "disproportionate memory usage". Could you generate a PoC to demonstrate that? Alternatively, it would be good if you could justify why the default limits are guaranteed to be large enough in practice.

src/read.c
```diff
@@ -3605,6 +3646,10 @@ static avifResult avifParseChunkOffsetBox(avifSampleTable * sampleTable, avifBoo
 
     uint32_t entryCount;
     AVIF_CHECKERR(avifROStreamReadU32(&s, &entryCount), AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(32) entry_count;
+    if (entryCount > AVIF_DEFAULT_IMAGE_COUNT_LIMIT) {
```
wantehchang: This and similar checks should use `decoder->imageCountLimit` instead.

Note: It is not obvious why `decoder->imageCountLimit` is an appropriate maximum in these checks. So we need to review these checks carefully.

src/read.c
```diff
@@ -3772,9 +3835,10 @@ static avifResult avifParseSampleTableBox(avifTrack * track, uint64_t rawOffset,
         avifDiagnosticsPrintf(diag, "Duplicate Box[stbl] for a single track detected");
         return AVIF_RESULT_BMFF_PARSE_FAILED;
     }
-    track->sampleTable = avifSampleTableCreate();
-    AVIF_CHECKERR(track->sampleTable != NULL, AVIF_RESULT_OUT_OF_MEMORY);
-
+    if (!track->sampleTable) {
```
wantehchang: Undo this change. This change is incorrect. `track->sampleTable` must be a null pointer here.

uwezkhan:
@wantehchang review on this PR?

uwezkhan:
hey can i get a review on this? its been a month

android_jni/avifandroidjni/src/main/jni/libavif_jni.cc
y-guyon, line 1: Please undo formatting changes in this PR. It makes it harder to compare the behavior changes.

include/avif/avif.h
y-guyon: This was resolved without addressing the comments. Unresolving.

uwezkhan: Thanks for pointing that out. I've updated the patch to remove the four public decoder options entirely and replaced them with internal implementation limits, so there is no longer any API change.

Regarding the resource-exhaustion concern, I agree that the key question is whether the affected metadata structures can cause disproportionate parser cost in practice. I'm investigating that further and will follow up with either concrete measurements/PoCs or a narrower justification for the remaining limits.

uwezkhan:
any update ?

include/avif/avif.h
y-guyon: In my opinion this was resolved without addressing the comments. Unresolving.

> It is a little excessive to add four decoder options that most users of libavif are unlikely to understand.

Would introducing a single boolean suit you? I am worried about adding hardcoded limits that cannot be disabled by any compile- or run-time flag.

wantehchang: Yannis: You are much more familiar with ISOBMFF than I am, so I defer the risk assessment and API design to you. I think you can make a better judgment.

y-guyon:
I think a file/PoC exhibiting "disproportionate memory usage and parser overhead" would prove the need for this PR. Could you first add a test triggering a memory allocation failure, and/or an image that takes too long to decode, that cannot be handled by a reasonable `imageSizeLimit` value in the existing API please?

uwezkhan:
Went back through the array code to answer this properly. The item, property, extent and group counts don't pre-size anything: the arrays start small (8/16/1/2 elements) and grow by doubling in `avifArrayPush`, one push per object actually read from the stream. So an array only reaches N if the file physically contains N objects, and the allocation stays bounded by the input size. Wantehchang's read was correct, there's no small-file amplification here.

Because of that I can't produce a PoC where one of these counts causes an allocation failure, or a decode that runs long, that a reasonable `imageSizeLimit` or the file size itself wouldn't already bound. The justification for the count limits doesn't hold up.

So I'm fine closing this. If there's a specific box you'd still want a hard cap on, tell me which one and I'll cut the PR down to just that with a real repro.

include/avif/avif.h
uwezkhan: On the boolean: it's moot if the limits go away, since there'd be nothing to gate. I left the longer answer in a PR comment, but the short version is these counts are already bounded by file size (arrays grow one push at a time), so the caps aren't buying anything over the existing limits.

PR #3189, opened 2026-05-04
Fix NaN bypass of AVIF_CLAMP in gain map pixel clamping

### LINKED ISSUE #3188: NaN bypass of AVIF_CLAMP in gain map computation causes assertion failure
## Summary

`powf()` in `avifRGBImageComputeGainMap()` can return NaN when processing crafted gain map gamma fractions. The NaN value bypasses `AVIF_CLAMP(v, 0.0f, 1.0f)` because IEEE 754 defines all comparisons with NaN as false, so the macro's ternary expressions all evaluate to the NaN input. The unclamped NaN propagates to `avifSetRGBAPixel()`, which asserts that pixel values are in `[0.0, 1.0]`.

- **Debug builds**: assertion failure → abort (DoS)
- **Release builds** (`-DNDEBUG`): NaN reaches `(uint8_t)(0.5f + NaN * maxChannelF)` — undefined behavior per C11 §6.3.1.4. On ARM64 and x86_64 hardware this produces 0 (data corruption, not memory corruption).

## Affected versions

- v1.4.1 (latest stable, confirmed)
- main branch at commit 2fdc999 (confirmed — the integer overflow fix does not address NaN)

## Affected component

- `src/gainmap.c`, lines 764–765 (`avifRGBImageComputeGainMap`)
- `include/avif/internal.h`, line 18 (`AVIF_CLAMP` macro)
- `src/reformat.c`, lines 1896–1898 (`avifSetRGBAPixel` — assertion site)

## Preconditions

- Build with `-DAVIF_ENABLE_EXPERIMENTAL_GAIN_MAP=ON`
- Attacker controls gain map metadata (e.g., via a crafted AVIF file with an embedded gain map)

## Root cause

In `avifRGBImageComputeGainMap()` at `src/gainmap.c:764–765`:

```c
const float gainMapGamma = avifUnsignedFractionToFloat(gainMap->gainMapGamma[c]);
// ...
v = powf((v - gainMapMinLog2[c]) / range, gainMapGamma);          // line 764
gainMapF[c][(size_t)j * width + i] = AVIF_CLAMP(v, 0.0f, 1.0f);  // line 765
```

When `gainMapGamma` is a degenerate value from a crafted fraction, `powf()` returns NaN. The `AVIF_CLAMP` macro:

```c
#define AVIF_CLAMP(x, low, high) (((x) < (low)) ? (low) : (((high) < (x)) ? (high) : (x)))
```

evaluates as:
- `NaN < 0.0f` → **false** (IEEE 754)
- `1.0f < NaN` → **false** (IEEE 754)
- Result: **NaN passes through unclamped**

The NaN then reaches `avifSetRGBAPixel()` in `src/reformat.c`:

```c
assert(rgbaPixel[0] >= 0.0f && rgbaPixel[0] <= 1.0f);  // NaN >= 0.0f → false → abort
```

Two additional AVIF_CLAMP sites in `avifRGBImageApplyGainMap()` (lines 156, 273) have the same pattern and are also vulnerable if upstream gain map math produces NaN.

### Minimal PoC (81 bytes)

```c
unsigned char poc[] = {
  0x30, 0x01, 0x01, 0xfd, 0x30, 0x30, 0x30, 0x01, 0x30, 0x30, 0x30, 0xfe,
  0x30, 0x30, 0x30, 0x30, 0x01, 0x01, 0x1b, 0x30, 0x30, 0x30, 0x30, 0x30,
  0x30, 0xfb, 0x30, 0x30, 0x30, 0x30, 0x7c, 0x01, 0x30, 0x01, 0x30, 0x01,
  0x30, 0x01, 0x30, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x30, 0x01,
  0x12, 0x01, 0x1e, 0x30, 0x30, 0x30, 0x30, 0x01, 0x30, 0x01, 0x30, 0x01,
  0x30, 0x01, 0x7f, 0x30, 0x30, 0x30, 0x30, 0x01, 0x30, 0x01, 0x30, 0x30,
  0x30, 0x30, 0x30, 0x30, 0x30, 0xe8, 0x30, 0x30, 0xff
};
```

This PoC is consumed by an API-level fuzzer that constructs `avifImage` objects and gain map metadata from the raw bytes, then calls `avifRGBImageComputeGainMap()`. A standalone reproducer is included in the linked PR.

### Build steps

```bash
git clone --depth 1 --branch v1.4.1 https://github.com/AOMediaCodec/libavif.git
cd libavif && mkdir build && cd build
cmake .. -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_AOM_ENCODE=ON \
         -DAVIF_ENABLE_EXPERIMENTAL_GAIN_MAP=ON \
         -DCMAKE_C_FLAGS="-g -O1 -fsanitize=address" \
         -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=OFF
make -j$(nproc)
```

### Expected output

```
Assertion failed: (rgbaPixel[1] >= 0.0f && rgbaPixel[1] <= 1.0f),
function avifSetRGBAPixel, file reformat.c, line 1897.
Abort trap: 6
```

## Crash distribution

Fuzzing with AFL++ 4.40c (12 instances, ~250M executions) produced 168 unique crash inputs (by content hash), all hitting the same root cause across three assertion sites:

| Assertion site | Approx. share | Channel |
|---------------|--------------|---------|
| `reformat.c:1896` | ~60% | R (`rgbaPixel[0]`) |
| `reformat.c:1897` | ~33% | G (`rgbaPixel[1]`) |
| `reformat.c:1898` | ~7% | B (`rgbaPixel[2]`) |

## Suggested fix

Replace the comparison-based `AVIF_CLAMP` with `fminf`/`fmaxf` at the three call sites feeding pixel data into `avifSetRGBAPixel()`. Per C99 §7.12.12, `fmaxf`/`fminf` return the non-NaN argument when one argument is NaN:

```c
// Before (NaN-unsafe):
gainMapF[c][...] = AVIF_CLAMP(v, 0.0f, 1.0f);

// After (NaN-safe):
gainMapF[c][...] = fminf(1.0f, fmaxf(0.0f, v));
```

A PR with this fix and verification against all 168 crash inputs is forthcoming.

## Notes

- This experimental feature is **not fuzzed by OSS-Fuzz** (the OSS-Fuzz build does not enable `-DAVIF_ENABLE_EXPERIMENTAL_GAIN_MAP=ON`).
- Found via custom AFL++ harness targeting the gain map API surface.

jortles:
Fixes #3188

## Summary

`powf()` in `avifRGBImageComputeGainMap()` can return NaN from degenerate gain map gamma fractions. `AVIF_CLAMP(v, 0.0f, 1.0f)` passes NaN through because IEEE 754 comparisons with NaN always return false. The NaN then triggers the assertion in `avifSetRGBAPixel()` (debug builds) or causes undefined behavior in the float-to-integer conversion (release builds).

## Changes

Replace `AVIF_CLAMP(x, 0.0f, 1.0f)` with `fminf(1.0f, fmaxf(0.0f, x))` at the three call sites in `src/gainmap.c` that feed pixel data into `avifSetRGBAPixel()`:

1. **Line 156** — `avifRGBImageApplyGainMap()`, identity copy path after color space conversion
2. **Line 277** — `avifRGBImageApplyGainMap()`, tone-mapped pixel output
3. **Line 770** — `avifRGBImageComputeGainMap()`, gain map value after `powf()` gamma remapping

Per C99 §7.12.12.2/§7.12.12.3, `fmaxf`/`fminf` return the non-NaN argument when one argument is NaN, which correctly clamps NaN to 0.0f.

## Verification

- 81-byte minimized PoC (afl-tmin) that triggers the assertion on v1.4.1 and main — passes clean with this fix
- All 168 unique crash inputs from AFL++ fuzzing (12 instances, ~250M executions) pass clean with this fix
- No regressions: the fix is a strict behavioral superset of the original AVIF_CLAMP for non-NaN inputs

## Discovery

Found via custom AFL++ fuzzer targeting the experimental gain map API (`avifRGBImageComputeGainMap` / `avifRGBImageApplyGainMap`). This code path is not covered by OSS-Fuzz as the build does not enable `-DAVIF_ENABLE_EXPERIMENTAL_GAIN_MAP=ON`.


src/gainmap.c
```diff
@@ -153,7 +153,11 @@ avifResult avifRGBImageApplyGainMap(const avifRGBImage * baseImage,
                         avifLinearRGBConvertColorSpace(basePixelRGBA, conversionCoeffs);
                     }
                     for (int c = 0; c < 3; ++c) {
-                        basePixelRGBA[c] = AVIF_CLAMP(linearToGamma(basePixelRGBA[c]), 0.0f, 1.0f);
+                        // Use fminf/fmaxf instead of AVIF_CLAMP for NaN safety:
+                        // AVIF_CLAMP passes NaN through because IEEE 754 comparisons
+                        // with NaN always return false. fmaxf/fminf return the non-NaN
+                        // argument per C99 §7.12.12.
+                        basePixelRGBA[c] = fminf(1.0f, fmaxf(0.0f, linearToGamma(basePixelRGBA[c])));
```
wantehchang: Anthony: Thank you for the pull request. Please add an entry for this fix to the "Changed since 1.4.1" section in CHANGELOG.md.

maryla-uc:
Hi, thank you for the issue and pull request.

I'm mostly curious/confused what kind of "degenerate values" cause the powf to return NaN.

In the issue, you say that "A standalone reproducer is included in the linked PR." but I don't see it? And you say that your 81 byte poc is for a fuzzer that is not provided so I can't use it to reproduce.

I'm wondering if the root cause might be earlier in the function, making us pass bad values to powf.

jortles:
You're right that the root cause is upstream of `powf` — it's receiving NaN, not producing it.

**Apply path** (the two sites at lines 156 and 277): when `baseLinear + baseOffset` is zero (e.g., a black pixel with `baseOffset = {0, 1}`), and the gain map log2 value is large enough for `exp2f()` to overflow to +Inf, the multiplication `0.0f * +Inf` evaluates to NaN per IEEE 754. The NaN then passes through `linearToGamma()` and `AVIF_CLAMP()` unmodified, since the macro's ternary comparisons all return false for NaN operands.

A similar NaN path exists in the **compute path** (line 764): if `base + baseOffset[c]` is near-zero after color space conversion, the ratio at line 695 overflows to +Inf. `AVIF_MAX(ratio, kEpsilon)` only lower-bounds, so +Inf passes through into `gainMapF`. After `avifFindMinMaxWithoutOutliers` picks up +Inf as the channel max, the normalization `(+Inf - min) / +Inf` evaluates to NaN (indeterminate form).

The `fminf`/`fmaxf` fix is intentionally a safety net at the output boundary — it ensures NaN can't reach `avifSetRGBAPixel()` regardless of which upstream path produced it. If you'd prefer a root-cause fix as well (e.g., clamping the ratio or guarding `exp2f` overflow), I'm happy to add that.

Regarding the reproducer: apologies, that was incorrectly stated in the issue. The 81-byte PoC is consumed by a custom fuzzer harness that constructs `avifImage` objects and gain map metadata from raw bytes, then calls `avifImageComputeGainMap()` / `avifImageApplyGainMap()`. I'll push a standalone reproducer that triggers the apply-path NaN directly — black base pixel, `baseOffset = 0`, `gainMapMax = 1000` so `exp2f(1000) = +Inf`.

jortles:
Updated the reproducer in 608ca50. The original used sRGB output transfer, which masked the crash: `avifToGammaSRGB()` absorbs NaN because all branch conditions (`< 0`, `< 0.003`, `< 1.0`) are false for NaN, so it falls through to `return 1.0f` — turning NaN into silent data corruption (black pixels render as white) rather than triggering the assertion.

Switched to `AVIF_TRANSFER_CHARACTERISTICS_LINEAR`, whose transfer function (`AVIF_CLAMP`) also passes NaN through, allowing it to reach the assertion in `avifSetRGBAPixel`.

Verified:
- **Without fix**: `Assertion 'rgbaPixel[0] >= 0.0f && rgbaPixel[0] <= 1.0f' failed` (SIGABRT)
- **With fix**: `PASS: no crash`

This also highlights that the NaN issue affects correctness even with sRGB — the tone-mapped output is silently wrong (1.0 instead of the correct value).

src/gainmap.c
maryla-uc: can you create a helper function e.g. avifNanSafeClamp()?

tests/reproduce_gainmap_nan.c
```diff
@@ -0,0 +1,152 @@
+/*
+ * reproduce_gainmap_nan.c — Demonstrates NaN crash in gain map tone mapping.
```
maryla-uc: This should be added as a test in avifgainmaptest.cc
The avifCreateGainMap bug can also be added as a test.

For example:

```
TEST(GainMapTest, ApplyGainMapNaN) {
  /* 2x2 black base image (sRGB, BT.709) */
  ImagePtr base(avifImageCreate(2, 2, 8, AVIF_PIXEL_FORMAT_YUV444));
  ASSERT_NE(base, nullptr);
  base->colorPrimaries = AVIF_COLOR_PRIMARIES_SRGB;
  base->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_SRGB;
  base->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT709;
  base->yuvRange = AVIF_RANGE_FULL;
  ASSERT_EQ(avifImageAllocatePlanes(base.get(), AVIF_PLANES_YUV),
            AVIF_RESULT_OK)
      << "Failed to allocate base planes";
  /* Y=0 (black), U=128/V=128 (neutral chroma) */
  memset(base->yuvPlanes[0], 0, (size_t)base->yuvRowBytes[0] * 2);
  memset(base->yuvPlanes[1], 128, (size_t)base->yuvRowBytes[1] * 2);
  memset(base->yuvPlanes[2], 128, (size_t)base->yuvRowBytes[2] * 2);

  /* 2x2 gain map image — all pixels at maximum (255 -> 1.0 normalized) */
  GainMapPtr gainMap(avifGainMapCreate());
  ASSERT_NE(gainMap, nullptr);
  gainMap->image = avifImageCreate(2, 2, 8, AVIF_PIXEL_FORMAT_YUV444);
  ASSERT_NE(gainMap->image, nullptr) << "Failed to create gain map image";
  gainMap->image->yuvRange = AVIF_RANGE_FULL;
  gainMap->image->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_IDENTITY;
  ASSERT_EQ(avifImageAllocatePlanes(gainMap->image, AVIF_PLANES_YUV),
            AVIF_RESULT_OK)
      << "Failed to allocate gain map planes";
  memset(gainMap->image->yuvPlanes[0], 255,
         (size_t)gainMap->image->yuvRowBytes[0] * 2);
  memset(gainMap->image->yuvPlanes[1], 255,
         (size_t)gainMap->image->yuvRowBytes[1] * 2);
  memset(gainMap->image->yuvPlanes[2], 255,
         (size_t)gainMap->image->yuvRowBytes[2] * 2);

  /*
   * Gain map metadata crafted to trigger NaN:
   *   gainMapMin  = 0     -> lerp lower bound
   *   gainMapMax  = 1000  -> lerp upper bound
   *   gamma       = 1     -> no gamma distortion
   *   baseOffset  = 0     -> (baseLinear + 0) = 0 for black pixels
   *   altOffset   = 0
   *
   * The math: lerp(0, 1000, powf(1.0, 1.0)) = 1000
   *           exp2f(1000 * weight) = +Inf
   *           (0.0 + 0.0) * +Inf = NaN  (IEEE 754)
   */
  for (int c = 0; c < 3; ++c) {
    gainMap->gainMapMin[c] = (avifSignedFraction){0, 1};
    gainMap->gainMapMax[c] = (avifSignedFraction){1000, 1};
    gainMap->gainMapGamma[c] = (avifUnsignedFraction){1, 1};
    gainMap->baseOffset[c] = (avifSignedFraction){0, 1};
    gainMap->alternateOffset[c] = (avifSignedFraction){0, 1};
  }
  gainMap->baseHdrHeadroom = (avifUnsignedFraction){0, 1};
  gainMap->alternateHdrHeadroom = (avifUnsignedFraction){6, 1};
  gainMap->useBaseColorSpace = 1;
  gainMap->altColorPrimaries = AVIF_COLOR_PRIMARIES_SRGB;
  gainMap->altTransferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_SRGB;
  gainMap->altMatrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT709;
  gainMap->altYUVRange = AVIF_RANGE_FULL;
  gainMap->altDepth = 8;
  gainMap->altPlaneCount = 3;

  /* Output tone-mapped image — set format/depth only.
   * avifRGBImageApplyGainMap sets width/height and allocates pixels internally. */
  avifRGBImage toneMap;
  memset(&toneMap, 0, sizeof(toneMap));
  toneMap.depth = 8;
  toneMap.format = AVIF_RGB_FORMAT_RGBA;

  avifContentLightLevelInformationBox clli;
  memset(&clli, 0, sizeof(clli));
  avifDiagnostics diag;
  avifDiagnosticsClearError(&diag);

  /* Apply with full HDR headroom (weight = 1.0).
   * Use LINEAR transfer so NaN propagates through to avifSetRGBAPixel.
   * (sRGB's gamma function absorbs NaN to 1.0f, hiding the crash.) */
  avifResult result = avifImageApplyGainMap(base.get(), gainMap.get(), 6.0f,
                                            AVIF_COLOR_PRIMARIES_SRGB,
                                            AVIF_TRANSFER_CHARACTERISTICS_LINEAR,
                                            &toneMap, &clli, &diag);

  // Expecting an error because the crafted metadata leads to NaN.
  EXPECT_EQ(result, AVIF_RESULT_INVALID_TONE_MAPPED_IMAGE)
      << avifResultToString(result) << " (" << diag.error << ")";

  avifRGBImageFreePixels(&toneMap);
}
```

maryla-uc (commented):
Thank you for the follow up analysis!

maryla-uc (approved):
Sorry for the delay. Could you fix the formatting? (which is on the code I provided so it's my fault I guess)
It would be nice to also have a test for the gain map creation case but that's optional.

src/gainmap.c
```diff
@@ -762,7 +775,7 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
                     float v = gainMapF[c][(size_t)j * width + i];
                     v = AVIF_CLAMP(v, gainMapMinLog2[c], gainMapMaxLog2[c]);
```
wantehchang, line 776: Just wondering: Does this `AVIF_CLAMP` call need to handle `v` equal to NaN?

PR #3195, opened 2026-05-15
Print diag msg if AOMD_SET_FRAME_SIZE_LIMIT fails

wantehchang:


src/codec_aom.c
wantehchang, line 156: cc: @juliobbv-p 

Vignesh: Here is an example of printing a diagnostic message after an `aom_codec_control()` call fails.

vigneshvg (approved):

PR #2710, opened 2025-03-27
avifenc: read jpeg/png through stdin with --stdin-format flag

maryla-uc:
When transcoding jpegs with a gain map, the gain map is ignored and a warning is printed.

#2676

PR #3197, opened 2026-05-15
Change "y4m or stdin" to "y4m" in avifenc help msg

wantehchang:
A follow-up to PR #2710.


wantehchang:
cc: @kjg0724 

Maryla: Please review. I hope I understand your PR https://github.com/AOMediaCodec/libavif/pull/2710 (avifenc: read jpeg/png through stdin with --stdin-format flag) correctly. Thanks!

maryla-uc (approved):
Good catch, thanks for the fix!

PR #3199, opened 2026-05-16
Reject oversized BMFF box sizes before serialization

jmestwa-coder:
## Summary

Reject box sizes that cannot be represented in the 32-bit ISO BMFF size field before writing the box header.

This prevents silent truncation when `headerSize + contentSize` exceeds `UINT32_MAX`, which could otherwise produce invalid serialized BMFF output.

## Changes

* Add an overflow-safe size validation in `avifRWStreamWriteFullBox()`
* Reject unrepresentable box sizes with `AVIF_RESULT_INVALID_ARGUMENT`
* Move marker assignment until after validation/allocation succeeds
* Add regression tests covering:

  * maximum valid regular box size
  * overflow regular box size
  * maximum valid full box size
  * overflow full box size
  * stream offset stability on failure

## Result

Oversized boxes are now rejected before serialization instead of emitting truncated 32-bit BMFF size values.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -397,13 +397,11 @@ avifResult avifRWStreamWriteChars(avifRWStream * stream, const char * chars, siz
 avifResult avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, size_t contentSize, int version, uint32_t flags, avifBoxMarker * marker)
 {
     assert(stream->numUsedBitsInPartialByte == 0); // Byte alignment is required.
-    if (marker) {
-        *marker = stream->offset;
-    }
     size_t headerSize = sizeof(uint32_t) + 4 /* size of type */;
     if (version != -1) {
         headerSize += 4;
     }
+    AVIF_CHECKERR(contentSize <= UINT32_MAX - headerSize, AVIF_RESULT_INVALID_ARGUMENT);
 
     AVIF_CHECKRES(makeRoom(stream, headerSize));
     memset(stream->raw->data + stream->offset, 0, headerSize);
@@ -415,6 +413,9 @@ avifResult avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, si
         stream->raw->data[stream->offset + 9] = (uint8_t)((flags >> 16) & 0xff);
         stream->raw->data[stream->offset + 10] = (uint8_t)((flags >> 8) & 0xff);
         stream->raw->data[stream->offset + 11] = (uint8_t)((flags >> 0) & 0xff);
+    }
+    if (marker) {
+        *marker = stream->offset;
     }
     stream->offset += headerSize;
 
--- a/src/stream.c
+++ b/src/stream.c
@@ -397,7 +397,6 @@ avifResult avifRWStreamWriteChars(avifRWStream * stream, const char * chars, siz
 avifResult avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, size_t contentSize, int version, uint32_t flags, avifBoxMarker * marker)
 {
     assert(stream->numUsedBitsInPartialByte == 0); // Byte alignment is required.
-    avifBoxMarker oldOffset = stream->offset;
     size_t headerSize = sizeof(uint32_t) + 4 /* size of type */;
     if (version != -1) {
         headerSize += 4;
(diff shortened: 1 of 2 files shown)


y-guyon (approved):

y-guyon:
Thank you for this change. Please run clang-format.

jmestwa-coder:
@y-guyon I fixed the clang-format issue and added coverage for the valid full-box upper-bound case as well.

src/stream.c
```diff
@@ -416,6 +414,9 @@ avifResult avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, si
         stream->raw->data[stream->offset + 10] = (uint8_t)((flags >> 8) & 0xff);
         stream->raw->data[stream->offset + 11] = (uint8_t)((flags >> 0) & 0xff);
     }
+    if (marker) {
+        *marker = stream->offset;
+    }
```
wantehchang: jmestwa-coder: The original location of this if statement is also correct; it only has the minor downside that the output parameter `*marker` is set even when the function fails. If we want to set the output parameter `*marker` only when the function succeeds, I would prefer we do this as soon as we know the function cannot fail, before we use `stream->offset`, because it will be clearer that we save the old value of `stream->offset` before it is modified.

tests/gtest/avifstreamtest.cc
```diff
@@ -202,6 +202,43 @@ TEST(StreamTest, WriteBitsLimit) {
...
+          std::numeric_limits<uint32_t>::max() - sizeof(uint32_t) - 4, &marker),
+      AVIF_RESULT_OK);
+  EXPECT_EQ(marker, size_t{0});
+  EXPECT_EQ(avifRWStreamOffset(&rw_stream), sizeof(uint32_t) + 4);
+
+  EXPECT_EQ(
+      avifRWStreamWriteBox(
+          &rw_stream, box_type,
+          std::numeric_limits<uint32_t>::max() - sizeof(uint32_t) - 3, &marker),
+      AVIF_RESULT_INVALID_ARGUMENT);
```
wantehchang, line 224: Nit: The code below assumes that `avifRWStreamWriteBox()` does not modify the stream's offset when it fails. This is too strong a requirement because we stop using a stream as soon as there is an error. Tests should avoid "over-specifying" the expected behavior.

We can allow this, but let's add the following `EXPECT_EQ`, similar to the `EXPECT_EQ` at line 239:

```
  EXPECT_EQ(avifRWStreamOffset(&rw_stream), sizeof(uint32_t) + 4);
```

src/stream.c
```diff
@@ -397,13 +397,11 @@ avifResult avifRWStreamWriteChars(avifRWStream * stream, const char * chars, siz
...
 {
     assert(stream->numUsedBitsInPartialByte == 0); // Byte alignment is required.
-    if (marker) {
-        *marker = stream->offset;
-    }
     size_t headerSize = sizeof(uint32_t) + 4 /* size of type */;
     if (version != -1) {
         headerSize += 4;
     }
+    AVIF_CHECKERR(contentSize <= UINT32_MAX - headerSize, AVIF_RESULT_INVALID_ARGUMENT);
```
wantehchang, line 404: `avifRWStreamWriteFullBox()` and the related `avifRWStreamWriteBox()` function are internal functions, so they don't need to check for invalid input arguments as thoroughly as a public API function. But it is fine to hold them to a higher standard.

I audited the callers of `avifRWStreamWriteFullBox()` and `avifRWStreamWriteBox()`. They all pass `AVIF_BOX_SIZE_TBD` as the `contentSize` argument, with only two exceptions: `sizeof(uint16_t)` and 0.

jmestwa-coder:
@wantehchang Thanks for the detailed review. I updated the patch to keep the overflow guard while narrowing the tests to the boundary validation itself, and adjusted the marker handling to use the pre-modification offset directly.

src/stream.c
jmestwa-coder: done, moved the marker assignment to right after makeRoom() so it's set once the function can't fail, before stream->offset is modified. dropped the extra local this time.

tests/gtest/avifstreamtest.cc
jmestwa-coder: added the offset check after the failing avifRWStreamWriteBox() call.

PR #3204, opened 2026-05-18
Check null return value from avifImageCreateEmpty

wantehchang:
Audit all calls to avifImageCreateEmpty() and avifImageCreate() in the libavif source tree (except the tests) and make sure that the callers check for a null return value.

Fix nits in two error messages.

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -1062,7 +1062,7 @@ static avifBool avifJPEGExtractGainMapImage(FILE * f,
             !memcmp(marker->data, tagMpf.data, tagMpf.size)) {
             avifImage * image = avifImageCreateEmpty();
             if (!image) {
-                fprintf(stderr, "Warning: out of memory\n");
+                fprintf(stderr, "Warning: out of memory when reading gain map\n");
                 return AVIF_FALSE;
             }
             // Set jpeg native matrix coefficients to allow copying YUV values directly.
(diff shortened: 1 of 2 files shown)


wantehchang:
@rootvector2 Dexter.k: FYI. If you are doing the same audit (I saw your PR https://github.com/AOMediaCodec/libavif/pull/3205), you can just review my results in this PR. To save our time, please do NOT audit the tests. Thanks!

apps/shared/avifjpeg.c
```diff
@@ -1061,6 +1061,10 @@ static avifBool avifJPEGExtractGainMapImage(FILE * f,
         if ((marker->marker == (JPEG_APP0 + 2)) && (marker->data_length > tagMpf.size) &&
             !memcmp(marker->data, tagMpf.data, tagMpf.size)) {
             avifImage * image = avifImageCreateEmpty();
+            if (!image) {
+                fprintf(stderr, "Warning: out of memory\n");
```
y-guyon: This is an error and not a warning, right?

y-guyon (approved):

apps/shared/avifjpeg.c
wantehchang: I imitated the current error messages in this function, which all start with "Warning". (There is only one excepption, at line 1078, which starts with "Note".) I think the error messages say "Warning" because the caller of this function ignores the failure:

```
        // Ignore the return value: continue even if we fail to find/parse/decode the gain map.
        if (avifJPEGExtractGainMapImage(f, sizeLimit, &cinfo, avif, gainMap, chromaDownsampling, expectIsoGainMap)) {
            ...
            avif->gainMap = gainMap;
        } else {
            avifGainMapDestroy(gainMap);
        }
```

wantehchang: I discussed this error message with Maryla. I've changed the error message to what Maryla suggested: "Warning: out of memory when reading gain map"

PR #3206, opened 2026-05-18
Reject unsupported external iloc data references

jmestwa-coder:
libavif currently supports same-file item extents and `idat` extents only.

`iloc.data_reference_index` values were previously parsed but ignored, causing extents declared as external references to be interpreted as same-file offsets.

This change rejects unsupported nonzero `data_reference_index` values during `iloc` parsing so downstream extent handling operates under validated source-boundary assumptions.

### Changes

* Reject nonzero `iloc.data_reference_index` values in `avifParseItemLocationBox()`
* Add a regression test covering unsupported external data references

### Test

* Mutate `white_1x1.avif` to set `data_reference_index = 1`
* Verify parsing now fails with `AVIF_RESULT_BMFF_PARSE_FAILED`
* Existing `IlocTest.TwoExtents` continues to pass

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2054,6 +2054,10 @@ static avifResult avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw,
 
         uint16_t dataReferenceIndex;
         AVIF_CHECKERR(avifROStreamReadU16(&s, &dataReferenceIndex), AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(16) data_reference_index;
+        if (dataReferenceIndex != 0) {
+            avifDiagnosticsPrintf(diag, "Item ID [%u] contains an unsupported data reference index [%u]", itemID, dataReferenceIndex);
+            return AVIF_RESULT_BMFF_PARSE_FAILED;
+        }
         uint64_t baseOffset;
         AVIF_CHECKERR(avifROStreamReadUX8(&s, &baseOffset, baseOffsetSize), AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(base_offset_size*8) base_offset;
         uint16_t extentCount;
--- a/src/read.c
+++ b/src/read.c
@@ -2033,13 +2033,15 @@ static avifResult avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw,
             return AVIF_RESULT_BMFF_PARSE_FAILED;
         }
 
+        // Version 0 has no construction_method field; the data is located by file offsets,
+        // matching construction method 0.
+        uint8_t constructionMethod = 0;
         if (version == 1 || version == 2) {
             AVIF_CHECKERR(avifROStreamReadBitsU32(&s, &reserved, /*bitCount=*/12), AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(12) reserved = 0;
             if (reserved) {
                 avifDiagnosticsPrintf(diag, "Box[iloc] has a non null reserved field [%u]", reserved);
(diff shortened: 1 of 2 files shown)


src/read.c
```diff
@@ -2054,6 +2054,10 @@ static avifResult avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw,
 
         uint16_t dataReferenceIndex;
         AVIF_CHECKERR(avifROStreamReadU16(&s, &dataReferenceIndex), AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(16) data_reference_index;
+        if (dataReferenceIndex != 0) {
+            avifDiagnosticsPrintf(diag, "Item ID [%u] contains an unsupported data reference index [%u]", itemID, dataReferenceIndex);
+            return AVIF_RESULT_BMFF_PARSE_FAILED;
+        }
```
wantehchang, line 2062: Thank you for the pull request. ISO/IEC 14496-12:2022 Clause 8.11.3 (Item location box) does not require or recommend that `data_reference_index` be set to 0 when this field is not used. Therefore I think it is better to simply ignore this field.

Note that the value 0 is a valid value for `data_reference_index`, so one can argue that the value 0 can't really be used to indicate that the field is not used.

wantehchang: I took a closer look. I think the condition at line 2057 should also check that `construction_method` is equal to 0:

```
        if (constructionMethod == 0 && dataReferenceIndex != 0) {
```

But it is not clear what value `constructionMethod` should be set to for version 0. Do you know?

tests/gtest/avifilocextenttest.cc
```diff
@@ -30,6 +33,31 @@ TEST(IlocTest, TwoExtents) {
...
+      std::search(avif.data, avif.data + avif.size, kIloc, kIloc + 4);
+  ASSERT_NE(iloc_position, avif.data + avif.size);
+  ASSERT_GE(static_cast<size_t>(avif.data + avif.size - iloc_position),
+            size_t{16});
+
+  // white_1x1.avif uses iloc version 0 with a single item. The
+  // data_reference_index field follows the item_ID field.
+  ASSERT_EQ(iloc_position[4], 0);
+  iloc_position[14] = 0;
+  iloc_position[15] = 1;
```
wantehchang, line 56: I suggest also verifying `data_reference_index` is equal to 0 before changing it to 1:

```
   ASSERT_EQ(iloc_position[4], 0);
+  ASSERT_EQ(iloc_position[14], 0);
+  ASSERT_EQ(iloc_position[15], 0);
   iloc_position[14] = 0;
   iloc_position[15] = 1;
```

src/read.c
jmestwa-coder: Thanks for taking a closer look.

From what I understand, `construction_method` is only explicitly present starting from `iloc` version 1/2, so for version 0 I’m not fully sure what the intended interpretation should be either.

I’ve updated the test to make the fixture assumptions explicit before mutating `data_reference_index`.

tests/gtest/avifilocextenttest.cc
jmestwa-coder: I verified the fixture layout and added assertions checking the original data_reference_index value before mutating it.

src/read.c
jmestwa-coder: done, pushed. the check is now gated on construction method 0.

on version 0: the construction_method field isn't present, and the data is located by file offsets, which is what construction method 0 codifies. so constructionMethod now defaults to 0 for version 0 and the check still applies there, while idat items are left alone.

PR #3207, opened 2026-05-18
Detect invalid box size in avifRWStreamFinishBox()

wantehchang:
Unfortunately avifRWStreamFinishBox() returns void and there are a lot of avifRWStreamFinishBox() calls, so have avifRWStreamFinishBox() call abort() on an invalid box size for now.

PR #3209, opened 2026-05-19
Change avifRWStreamFinishBox to return avifResult

wantehchang:
Follow up on PR #3207. Allow error handling.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -742,7 +742,7 @@ avifResult avifRWStreamWriteBox(avifRWStream * stream, const char * type, size_t
 avifResult avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, size_t contentSize, int version, uint32_t flags, avifBoxMarker * marker);
 // marker is the offset of the size field in stream, returned by a previous
 // avifRWStreamWriteBox() or avifRWStreamWriteFullBox() call.
-void avifRWStreamFinishBox(avifRWStream * stream, avifBoxMarker marker);
+AVIF_NODISCARD avifResult avifRWStreamFinishBox(avifRWStream * stream, avifBoxMarker marker);
 avifResult avifRWStreamWriteU8(avifRWStream * stream, uint8_t v);
 avifResult avifRWStreamWriteU16(avifRWStream * stream, uint16_t v);
 avifResult avifRWStreamWriteU32(avifRWStream * stream, uint32_t v);
--- a/src/stream.c
+++ b/src/stream.c
@@ -426,7 +426,7 @@ avifResult avifRWStreamWriteBox(avifRWStream * stream, const char * type, size_t
     return avifRWStreamWriteFullBox(stream, type, contentSize, -1, 0, marker);
 }
 
-void avifRWStreamFinishBox(avifRWStream * stream, avifBoxMarker marker)
+avifResult avifRWStreamFinishBox(avifRWStream * stream, avifBoxMarker marker)
 {
     assert(stream->numUsedBitsInPartialByte == 0); // Byte alignment is required.
     size_t boxSize = stream->offset - marker;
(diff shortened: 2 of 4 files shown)


include/avif/internal.h
```diff
@@ -742,7 +742,7 @@ avifResult avifRWStreamWriteBox(avifRWStream * stream, const char * type, size_t
 avifResult avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, size_t contentSize, int version, uint32_t flags, avifBoxMarker * marker);
 // marker is the offset of the size field in stream, returned by a previous
 // avifRWStreamWriteBox() or avifRWStreamWriteFullBox() call.
-void avifRWStreamFinishBox(avifRWStream * stream, avifBoxMarker marker);
+AVIF_NODISCARD avifResult avifRWStreamFinishBox(avifRWStream * stream, avifBoxMarker marker);
```
wantehchang, line 745: I can remove this `AVIF_NODISCARD` after this PR passes the CI workflows.

src/write.c
```diff
@@ -3475,9 +3475,9 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
     altImageMetadata = encoder->data->altImageMetadata;
     avifResult result = avifRWStreamWriteProperties(dedup, &s, encoder, imageMetadata, altImageMetadata);
     avifItemPropertyDedupDestroy(dedup);
-    AVIF_CHECKRES(result);
-    avifRWStreamFinishBox(&s, ipco);
     dedup = NULL;
+    AVIF_CHECKRES(result);
+    AVIF_CHECKRES(avifRWStreamFinishBox(&s, ipco));
```
wantehchang, line 3480: Note that I moved the statement `dedup = NULL;` immediately after the `avifItemPropertyDedupDestroy(dedup)` call here.

src/stream.c
```diff
@@ -426,20 +426,19 @@ avifResult avifRWStreamWriteBox(avifRWStream * stream, const char * type, size_t
...
     // special values 0 and 1.
-    assert(boxSize >= sizeof(uint32_t) + 4);
-    if (boxSize > UINT32_MAX) {
-        abort();
-    }
+    AVIF_ASSERT_OR_RETURN(boxSize >= sizeof(uint32_t) + 4);
+    AVIF_CHECKERR(boxSize <= UINT32_MAX, AVIF_RESULT_INVALID_ARGUMENT);
     uint32_t noSize = avifHTONL((uint32_t)boxSize);
     memcpy(stream->raw->data + marker, &noSize, sizeof(uint32_t));
+    return AVIF_RESULT_OK;
```
wantehchang, line 441: Please review. Only this file and the change at src/write.c:3480 (annotated with a comment) need to be checked carefully. All the other changes are mechanical changes. Thanks!

tests/gtest/avifstreamtest.cc
```diff
@@ -52,9 +52,10 @@ TEST(StreamTest, Roundtrip) {
   const uint16_t rw_someu16 = 0xAABB;
   EXPECT_EQ(avifRWStreamWriteU16(&rw_stream, rw_someu16), AVIF_RESULT_OK);
 
-  avifRWStreamFinishBox(&rw_stream, rw_full_box_marker);
+  EXPECT_EQ(avifRWStreamFinishBox(&rw_stream, rw_full_box_marker),
```
vigneshvg: these could be ASSERT_EQs but this is consistent with the rest of the file.

vigneshvg (approved):

tests/gtest/avifstreamtest.cc
wantehchang: Done. You are right. The test should not keep going if any of these `avifRWStream` functions fails.

vigneshvg (approved):

PR #3212, opened 2026-05-20
Make avifAlloc(0) deterministic and fix RWData shrink-to-zero handling

metsw24-max:
This PR updates the `avifAlloc()` contract to handle zero-size allocations deterministically and fixes a latent bug in `avifRWDataRealloc()` when shrinking buffers to size `0`.

Previously, `avifAlloc(0)` asserted in debug builds and relied on implementation-defined `malloc(0)` behavior in release builds. This patch standardizes the behavior by returning `NULL` for zero-size allocations.

During the allocator audit, a bug was identified in `avifRWDataRealloc()`: shrinking an existing buffer to zero incorrectly routed through `avifAlloc(0)`, which could previously assert/crash or incorrectly surface as an out-of-memory failure. The shrink-to-zero path is now handled explicitly.

### Memory allocation contract

* Updated `include/avif/avif.h`

  * Documented that `avifAlloc()` returns `NULL` when `size == 0`.

* Updated `src/mem.c`

  * Removed the debug-only assertion for `size != 0`.
  * Added deterministic handling for zero-size allocations:

    * `avifAlloc(0)` now returns `NULL`.
  * Avoids implementation-defined `malloc(0)` behavior.

### RWData realloc fix

* Updated `src/rawdata.c`

  * Added explicit handling for:

    ```c
    avifRWDataRealloc(raw, 0)
    ```
  * Shrinking to zero now:

    * releases the existing buffer,
    * resets the allocation state correctly,
    * returns `AVIF_RESULT_OK`.

* Simplified the copy condition after introducing the early-return path:

  ```c
  if (raw->size)
  ```

## Tests Added

Added regression tests covering both the new allocator contract and the realloc fix:

* `AvifAllocTest.ZeroReturnsNull`

  * Verifies `avifAlloc(0)` consistently returns `NULL`.

* `AvifAllocTest.RWDataReallocShrinkToZeroSucceeds`

  * Verifies shrinking RWData buffers to zero:

    * succeeds,
    * releases memory,
    * clears internal state correctly.

### DIFF
--- a/src/rawdata.c
+++ b/src/rawdata.c
@@ -8,9 +8,15 @@
 avifResult avifRWDataRealloc(avifRWData * raw, size_t newSize)
 {
     if (raw->size != newSize) {
+        if (newSize == 0) {
+            // Shrink-to-zero: release the buffer instead of calling avifAlloc(0),
+            // which would return NULL under the new contract and falsely look like OOM.
+            avifRWDataFree(raw);
+            return AVIF_RESULT_OK;
+        }
         uint8_t * newData = (uint8_t *)avifAlloc(newSize);
         AVIF_CHECKERR(newData, AVIF_RESULT_OUT_OF_MEMORY);
-        if (raw->size && newSize) {
+        if (raw->size) {
             memcpy(newData, raw->data, AVIF_MIN(raw->size, newSize));
         }
         avifFree(raw->data);
--- a/src/rawdata.c
+++ b/src/rawdata.c
@@ -9,8 +9,7 @@ avifResult avifRWDataRealloc(avifRWData * raw, size_t newSize)
 {
     if (raw->size != newSize) {
         if (newSize == 0) {
-            // Shrink-to-zero: release the buffer instead of calling avifAlloc(0),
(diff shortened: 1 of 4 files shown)


src/rawdata.c
```diff
@@ -8,9 +8,15 @@
 avifResult avifRWDataRealloc(avifRWData * raw, size_t newSize)
 {
     if (raw->size != newSize) {
+        if (newSize == 0) {
+            // Shrink-to-zero: release the buffer instead of calling avifAlloc(0),
+            // which would return NULL under the new contract and falsely look like OOM.
```
maryla-uc: nit: don't mention "the new contract", the comment should only document the current state, no need to hint at a past behavior unless there is a specific reason to do so

maryla-uc (approved):

src/rawdata.c
metsw24-max: reworded to document only the current behavior dropped the "new contract" framing.

maryla-uc:
Thank you for the PR!

PR #3213, opened 2026-05-20
reformat: complete remaining size_t index conversions

metsw24-max:
Follow up f17110da by converting the remaining loop indices in `avifImageYUVAnyToRGBAnySlow()` from `uint32_t` to `size_t`.

This ensures pointer-offset multiplications involving row strides and pixel offsets are performed in `size_t` arithmetic rather than
potentially overflowing in 32-bit intermediates.

## Changes

- Convert `j` and `i` loop counters to `size_t`
- Convert derived chroma indices `uvJ` and `uvI` to `size_t`
- Remove redundant cast-at-use stride multiplications

## Rationale

The earlier sweep in f17110da standardized this approach across other YUV/RGB conversion paths. This patch applies the same fix
pattern to the remaining slow conversion path that was missed.

## Testing

- Build succeeds without warnings
- Existing conversion paths remain unchanged functionally
- Large-image stride arithmetic now consistently uses `size_t`

### DIFF
--- a/src/reformat.c
+++ b/src/reformat.c
@@ -683,9 +683,9 @@ static avifResult avifImageYUVAnyToRGBAnySlow(const avifImage * image,
     // to clang's analyzer.
     assert((alphaMultiplyMode == AVIF_ALPHA_MULTIPLY_MODE_NO_OP) || aPlane);
 
-    for (uint32_t j = 0; j < image->height; ++j) {
+    for (size_t j = 0; j < image->height; ++j) {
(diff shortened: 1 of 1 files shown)


wantehchang:

metsw24-max:
No worries, #3269 handles it more thoroughly than mine did. The ptrdiff_t treatment for the adjustable column/row offsets is the right call, I'd only touched the loop indices. Glad it made it in, thanks for consolidating.

PR #3214, opened 2026-05-20
Replace strcpy/strcat with snprintf in apps and tests

metsw24-max:
Replace remaining `strcpy`/`strcat` usages in `apps/` and `tests/` with bounded `snprintf()` calls.

This change is a small cleanup/hardening improvement that aligns with recent repository changes replacing unsafe string APIs with bounded alternatives.

## Changes

* Replace:

  * `strcpy()` + `strcat()` sequences in `tests/aviftest.c`
  * `strcpy()` in `apps/avifenc.c`

* Use `snprintf()` with explicit destination sizes instead.

* Remove now-unnecessary intermediate length variables and manual concatenation logic.

## Notes

The existing code paths were already bounded and safe by construction. This patch does not change behavior; it simplifies the logic and removes usage of traditionally unsafe libc string APIs.

## Verification

Built successfully with MSVC 19.44:

* `aviftest.exe`
* `avifenc.exe`
* `avif.lib`

No warnings or errors observed.

### DIFF
--- a/apps/avifenc.c
+++ b/apps/avifenc.c
@@ -1152,7 +1152,7 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
     const char * const codecName = avifCodecName(settings->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
     char speedStr[16];
     if (settings->speed == AVIF_SPEED_DEFAULT) {
-        strcpy(speedStr, "default");
+        snprintf(speedStr, sizeof(speedStr), "default");
     } else {
         snprintf(speedStr, sizeof(speedStr), "%d", settings->speed);
     }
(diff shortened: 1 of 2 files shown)


apps/avifenc.c
```diff
@@ -1152,7 +1152,7 @@ static avifBool avifEncodeImagesFixedQuality(const avifSettings * settings,
     const char * const codecName = avifCodecName(settings->codecChoice, AVIF_CODEC_FLAG_CAN_ENCODE);
     char speedStr[16];
     if (settings->speed == AVIF_SPEED_DEFAULT) {
-        strcpy(speedStr, "default");
+        snprintf(speedStr, sizeof(speedStr), "default");
```
wantehchang: Thanks for the pull request. Since the original code is correct, and modern tools have no difficulty verifying its correctness, these changes are really a matter of preference. I am afraid that I am going to decline this pull request.

We can, however, consider changing the `strcpy()` and `strcat()` calls to `memcpy()` because we have the string lengths available.

Note: This change especially stands out. It is obvious that the code is correct even though it uses `strcpy()`:

```
    char speedStr[16];
    strcpy(speedStr, "default");
```

metsw24-max:
I have reverted `avifenc.c` and changed `aviftest.c` to use memcpy with the existing length variables. Please take a look when you have a moment.

wantehchang (approved):
LGTM. Thanks. I made a small edit (moved the declaration of `ioDirLen` earlier).

PR #3216, opened 2026-05-20
avifenc: add --ignore-alpha flag to discard alpha channel on encode

wantehchang:
The first and second commits are copies of the first and second commits in pull request https://github.com/AOMediaCodec/libavif/pull/3130 by @enyetsuen.

The third commit contains my edits and further changes.


wantehchang (commented):
Vincent: I requested a review from you because I remember you worked on the `avifReadImage()` function. This PR is quite large, but there are a lot of mechanical changes. I suggest reviewing the code in the following order:
1. CHANGELOG.md
2. apps/avifenc.c
3. doc/avifenc.1.md
4. apps/shared/avifutil.{h,c}
5. apps/shared/avifpng.{h,c}
6. apps/shared/y4m.{h,c}
7. tests/test_cmd.sh
8. Other files

Thank you!

@vigneshvg Vignesh: FYI. You requested the `--ignore-alpha` option in issue https://github.com/AOMediaCodec/libavif/issues/1665.

apps/avifenc.c
```diff
@@ -622,6 +625,7 @@ static avifBool avifInputReadImage(avifInput * input,
                                                               ignoreColorProfile,
                                                               ignoreExif,
                                                               ignoreXMP,
+                                                              ignoreAlpha,
```
wantehchang, line 628: Vincent: When I wrote this PR, I didn't realize that `avifImageFreePlanes()` is a public function. Since adding the `ignoreAlpha` parameter to `avifReadImage()` makes it necessary to update its callers, an alternative solution is to leave `avifReadImage()` unchanged and call `avifImageFreePlanes(dstImage, AVIF_PLANES_A)` if `ignoreAlpha` is true. This will make the PR much smaller at the cost of allocating and copying the alpha plane from the input image unnecessarily. Would you prefer this alternative solution?

vrabaud: It's ok, it's all internal. We should probably merge those ignores flags into a struct anyway.

vrabaud (approved):

PR #3218, opened 2026-05-21
Fix bug in avifToLinearIEC61966

maryla-uc:
The negative portion of the curve did not match the spec, and was not the inverse of the fromLinear function.

### DIFF
--- a/src/colr.c
+++ b/src/colr.c
@@ -317,7 +317,7 @@ static float avifToGammaLog100Sqrt10(float linear)
 static float avifToLinearIEC61966(float gamma)
 {
     if (gamma < -4.5f * 0.018053968510807f) {
-        return powf((-gamma + 0.09929682680944f) / -1.09929682680944f, 1.0f / 0.45f);
+        return -powf((gamma - 0.09929682680944f) / -1.09929682680944f, 1.0f / 0.45f);
     } else if (gamma < 4.5f * 0.018053968510807f) {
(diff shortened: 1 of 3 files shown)


src/colr.c
```diff
@@ -317,7 +317,7 @@ static float avifToGammaLog100Sqrt10(float linear)
 static float avifToLinearIEC61966(float gamma)
 {
     if (gamma < -4.5f * 0.018053968510807f) {
-        return powf((-gamma + 0.09929682680944f) / -1.09929682680944f, 1.0f / 0.45f);
+        return -powf((gamma - 0.09929682680944f) / -1.09929682680944f, 1.0f / 0.45f);
```
wantehchang, line 320: Is this bug fix worth documenting in the changelog?

Note: I didn't compare the new code with the spec, but the original code is incorrect mathematically because the base of the power, `(-gamma + 0.09929682680944f) / -1.09929682680944f`, is negative.

wantehchang (approved):

src/colr.c
maryla-uc: I added a changelog entry, I think it's good hygiene, thanks for reminding me.
Fortunately I think this bug is unlikely to have ever affected anyone. It only affects negative values, and only for transfer function 11 which probably nobody uses. And the transfer functions are only used by the gain map code (creation/tone mapping).

wantehchang (approved):

PR #3220, opened 2026-05-22
Harden gain-map pixel count arithmetic and dimension validation

metsw24-max:
## Summary

This PR hardens gain-map dimension handling by introducing centralized checked pixel-count arithmetic and replacing unchecked `width * height` calculations in gain-map processing paths.

## Changes

- Added `avifDimensionsToPixelCount()` helper for overflow-safe pixel-count computation
- Rejected zero-sized and overflow-triggering dimensions
- Hardened `avifDimensionsTooLarge()` with explicit zero-dimension validation
- Replaced unchecked gain-map arithmetic with validated pixel-count usage
- Added safer allocation-size validation in gain-map computation paths
- Removed unused return value from `avifRWStreamFinishBox()`
- Minor comment/formatting cleanup

## Testing

Added regression tests covering:
- valid pixel-count computation
- zero dimensions
- overflow-triggering dimensions
- invalid dimension handling in `avifDimensionsTooLarge()`

Validated against existing build/tests and normal gain-map decoding paths.

### DIFF
--- a/include/avif/internal.h
+++ b/include/avif/internal.h
@@ -380,6 +380,7 @@ avifResult avifRGBImagePremultiplyAlphaLibYUV(avifRGBImage * rgb);
 avifResult avifRGBImageUnpremultiplyAlphaLibYUV(avifRGBImage * rgb);
 
 AVIF_NODISCARD avifBool avifDimensionsTooLarge(uint32_t width, uint32_t height, uint32_t imageSizeLimit, uint32_t imageDimensionLimit);
+AVIF_NODISCARD avifBool avifDimensionsToPixelCount(uint32_t width, uint32_t height, size_t * pixelCount);
 
 // Given the number of encoding threads or decoding threads available and the image dimensions,
 // chooses suitable values of *tileRowsLog2 and *tileColsLog2.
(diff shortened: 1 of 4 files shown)


include/avif/internal.h
```diff
@@ -742,7 +743,7 @@ avifResult avifRWStreamWriteBox(avifRWStream * stream, const char * type, size_t
 avifResult avifRWStreamWriteFullBox(avifRWStream * stream, const char * type, size_t contentSize, int version, uint32_t flags, avifBoxMarker * marker);
 // marker is the offset of the size field in stream, returned by a previous
 // avifRWStreamWriteBox() or avifRWStreamWriteFullBox() call.
-AVIF_NODISCARD avifResult avifRWStreamFinishBox(avifRWStream * stream, avifBoxMarker marker);
+void avifRWStreamFinishBox(avifRWStream * stream, avifBoxMarker marker);
```
wantehchang, line 746: This is a merge error.

src/gainmap.c
```diff
@@ -109,6 +109,10 @@ avifResult avifRGBImageApplyGainMap(const avifRGBImage * baseImage,
     memset(&rgbGainMap, 0, sizeof(rgbGainMap));
 
     avifResult res = AVIF_RESULT_OK;
+    size_t numPixels;
+    if (!avifDimensionsToPixelCount(width, height, &numPixels)) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
```
wantehchang, line 115: cc: @maryla-uc 

metsw24-max: Thank you for your interest in reviewing libavif source code. The code in this file modified by this PR was reviewed at least twice recently. The reason we can multiply width and height (after casting to `size_t`) with no risk of integer overflow is that a buffer of dimensions width x height has been allocated successfully. Therefore we know width x height must fit in `size_t`. I am going to decline this PR.

PR #3233, opened 2026-05-29
Fix unchecked integer overflow in encoder UV plane allocation

jortles:
## Summary

- Add pre-multiplication overflow checks to `codec_avm.c` and `codec_svt.c` for UV plane size computations, consistent with the existing safe pattern in `avifImageAllocatePlanes()` (`avif.c:435-441`)
- Add overflow check in `avifImageCopyProperties()` before `numProperties` allocation, consistent with `avifImagePushProperty()` (`avif.c:387`)

## Details

Three sites compute allocation sizes via unchecked integer multiplication, unlike their parallel functions which validate before multiplying:

**codec_avm.c:919** — `channelSize * monoUVWidth` (uint32_t) and `monoUVHeight * monoUVRowBytes` (size_t) used for `avifAlloc()` without overflow checks. The parallel code in `avifImageAllocatePlanes()` checks `width > UINT32_MAX / channelSize` and `height > PTRDIFF_MAX / fullRowBytes` before the same operations.

**codec_svt.c:286** — `uvWidth * bytesPerPixel` (uint32_t) and `uvHeight * uvRowBytes` (size_t) computed without pre-multiplication validation. The existing post-hoc check (`uvSize > UINT32_MAX / 2`) cannot detect a wrap that already occurred.

**avif.c:237** — `numProperties * sizeof(...)` passed to `avifAlloc()` without an overflow guard. `avifImagePushProperty()` in the same file already checks `numProperties < SIZE_MAX / sizeof(avifImageItemProperty)` before an identical allocation.

## Test plan

- [x] Builds cleanly with `cmake --build . --target avif_obj` (zero warnings)
- [ ] Existing test suite passes (CI)

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -234,6 +234,7 @@ static avifResult avifImageCopyProperties(avifImage * dstImage, const avifImage 
     dstImage->numProperties = 0;
 
     if (srcImage->numProperties != 0) {
+        AVIF_CHECKERR(srcImage->numProperties < SIZE_MAX / sizeof(srcImage->properties[0]), AVIF_RESULT_INVALID_ARGUMENT);
         dstImage->properties = (avifImageItemProperty *)avifAlloc(srcImage->numProperties * sizeof(srcImage->properties[0]));
         AVIF_CHECKERR(dstImage->properties != NULL, AVIF_RESULT_OUT_OF_MEMORY);
         memset(dstImage->properties, 0, srcImage->numProperties * sizeof(srcImage->properties[0]));
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -911,17 +911,17 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
             // monochrome. Manually set UV planes to 0.5.
 
             // avmImage is always 420 when we're monochrome
+            if (image->width == UINT32_MAX || image->height == UINT32_MAX) {
+                return AVIF_RESULT_INVALID_ARGUMENT;
+            }
             uint32_t monoUVWidth = (image->width + 1) >> 1;
             uint32_t monoUVHeight = (image->height + 1) >> 1;
 
             // Allocate the U plane if necessary.
             if (!avmImageAllocated) {
                 uint32_t channelSize = avifImageUsesU16(image) ? 2 : 1;
-                if (monoUVWidth > UINT32_MAX / channelSize) {
-                    return AVIF_RESULT_INVALID_ARGUMENT;
-                }
                 uint32_t monoUVRowBytes = channelSize * monoUVWidth;
-                if (monoUVHeight > PTRDIFF_MAX / monoUVRowBytes) {
+                if (monoUVHeight > SIZE_MAX / monoUVRowBytes) {
                     return AVIF_RESULT_INVALID_ARGUMENT;
                 }
                 size_t monoUVSize = (size_t)monoUVHeight * monoUVRowBytes;
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -917,7 +917,13 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
             // Allocate the U plane if necessary.
             if (!avmImageAllocated) {
                 uint32_t channelSize = avifImageUsesU16(image) ? 2 : 1;
+                if (monoUVWidth > UINT32_MAX / channelSize) {
+                    return AVIF_RESULT_INVALID_ARGUMENT;
+                }
                 uint32_t monoUVRowBytes = channelSize * monoUVWidth;
+                if (monoUVHeight > PTRDIFF_MAX / monoUVRowBytes) {
+                    return AVIF_RESULT_INVALID_ARGUMENT;
+                }
                 size_t monoUVSize = (size_t)monoUVHeight * monoUVRowBytes;
 
                 monoUVPlane = avifAlloc(monoUVSize);
--- a/src/codec_svt.c
+++ b/src/codec_svt.c
@@ -283,7 +283,13 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
 #if SVT_AV1_CHECK_VERSION(1, 8, 0)
         // Simulate 4:2:0 UV planes. SVT-AV1 does not support 4:0:0 samples.
         const uint32_t uvWidth = (image->width + y_shift) >> y_shift;
+        if (uvWidth > UINT32_MAX / bytesPerPixel) {
+            goto cleanup;
+        }
         const uint32_t uvRowBytes = uvWidth * bytesPerPixel;
+        if (uvHeight > PTRDIFF_MAX / uvRowBytes) {
+            goto cleanup;
+        }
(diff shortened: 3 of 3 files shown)


src/codec_avm.c
```diff
@@ -917,7 +917,13 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
             // Allocate the U plane if necessary.
             if (!avmImageAllocated) {
                 uint32_t channelSize = avifImageUsesU16(image) ? 2 : 1;
+                if (monoUVWidth > UINT32_MAX / channelSize) {
```
y-guyon: `monoUVWidth` is always below `UINT32_MAX / channelSize` because `channelSize` is at most 2 and `monoUVWidth = (image->width + 1) >> 1`.

However I wonder if there should be a check for the `image->width + 1` part. Maybe `avm_codec_enc_init()` fails earlier in case of a huge dimension, but that could change in the future.

src/codec_avm.c
```diff
@@ -917,7 +917,13 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
+                    return AVIF_RESULT_INVALID_ARGUMENT;
+                }
                 uint32_t monoUVRowBytes = channelSize * monoUVWidth;
+                if (monoUVHeight > PTRDIFF_MAX / monoUVRowBytes) {
```
y-guyon: I see these checks were inspired from:

https://github.com/AOMediaCodec/libavif/blob/d8b4e042cde136995909a7c5753d5f2910772248/src/avif.c#L433-L440

But `PTRDIFF_MAX` makes little sense in both places in my opinion. It should be `SIZE_MAX` right?

y-guyon (commented):

src/codec_avm.c
jortles: Good catch — you're right that `monoUVWidth` can never exceed `UINT32_MAX / channelSize` given the current math. I've removed that unreachable check and added a guard for the `image->width + 1` overflow instead, which is the actual risk. Same change applied to `codec_svt.c`.

src/codec_avm.c
jortles: Agreed — `SIZE_MAX` is the correct bound here. Changed in both `codec_avm.c` and `codec_svt.c`.

src/codec_avm.c
```diff
@@ -911,13 +911,19 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
             // monochrome. Manually set UV planes to 0.5.
 
             // avmImage is always 420 when we're monochrome
+            if (image->width == UINT32_MAX || image->height == UINT32_MAX) {
```
y-guyon, line 914: Maybe `image->width` should be cast to `size_t` instead of compared against `UINT32_MAX`? It does not matter much for such huge dimensions I guess.

wantehchang: Yes, we should allow `UINT32_MAX`.

As Yannis suggested, we can cast `image->width` to a larger type such as `uint64_t`. (`size_t` is 32 bits in 32-bit systems.) We use this approach elsewhere.

Another approach is to replace `(image->width + 1) >> 1` with `image->width / 2 + image->width % 2`, or (since `image->width >= 1`) `1 + (image->width - 1) / 2`, or some other similar trick.

src/avif.c
```diff
@@ -234,6 +234,7 @@ static avifResult avifImageCopyProperties(avifImage * dstImage, const avifImage
     dstImage->numProperties = 0;
 
     if (srcImage->numProperties != 0) {
+        AVIF_CHECKERR(srcImage->numProperties < SIZE_MAX / sizeof(srcImage->properties[0]), AVIF_RESULT_INVALID_ARGUMENT);
```
wantehchang, line 237: Please revert the change to this file. This has been addressed by PR https://github.com/AOMediaCodec/libavif/pull/3210.

src/codec_avm.c
wantehchang: I just realized that by the time we reach here, `image->width` and `image->height` have been validated by `avifValidateGrid()` to be valid AV1 frame width and height. So we know `image->width <= 65536` and `image->height <= 65536`. Therefore we should simply delete this if statement.

src/codec_avm.c
```diff
@@ -911,13 +911,19 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
...
             uint32_t monoUVWidth = (image->width + 1) >> 1;
             uint32_t monoUVHeight = (image->height + 1) >> 1;
 
             // Allocate the U plane if necessary.
             if (!avmImageAllocated) {
                 uint32_t channelSize = avifImageUsesU16(image) ? 2 : 1;
                 uint32_t monoUVRowBytes = channelSize * monoUVWidth;
+                if (monoUVHeight > SIZE_MAX / monoUVRowBytes) {
+                    return AVIF_RESULT_INVALID_ARGUMENT;
+                }
```
wantehchang, line 926: We should also delete this if statement because:

```
    monoUVWidth <= 32768
    monoUVHeight <= 32768
    channelSize <= 2
    monoUVHeight * monoUVRowBytes <= 32768 * 2 * 32768 < UINT32_MAX <= SIZE_MAX
```

src/codec_svt.c
```diff
@@ -282,8 +282,14 @@ static avifResult svtCodecEncodeImage(avifCodec * codec,
...
 #if SVT_AV1_CHECK_VERSION(1, 8, 0)
         // Simulate 4:2:0 UV planes. SVT-AV1 does not support 4:0:0 samples.
+        if (image->width == UINT32_MAX || image->height == UINT32_MAX) {
+            goto cleanup;
+        }
         const uint32_t uvWidth = (image->width + y_shift) >> y_shift;
         const uint32_t uvRowBytes = uvWidth * bytesPerPixel;
+        if (uvHeight > SIZE_MAX / uvRowBytes) {
+            goto cleanup;
+        }
```
wantehchang, line 292: `y_shift` is equal to 1 for 4:2:0. So my analysis in codec_avm.c also applies here. We should delete these two if statements.

wantehchang (commented):
Anthony: After merging PR https://github.com/AOMediaCodec/libavif/pull/3210, and after analyzing the changes to codec_avm.c and codec_svt.c carefully, I concluded that we don't need any of the changes in this PR. So I am going to close this PR.

Please check if my analysis is correct. Thank you!

src/codec_avm.c
y-guyon: Are there the same frame dimension limits in AV2?

In general it could be `AVIF_ASSERT_OR_RETURN()` instead of checks, but leaving the code as is is fine with me too.

wantehchang: Yannis: Thank you for checking my analysis. This was a flaw in my analysis. I just checked the [AV2 specification (version 1.0.0)](https://av2.aomedia.org/v1.0.0/index.html#general_sequence_header_obu_syntax). AV2 has the same frame dimension limits as AV1. (We can verify this by searching for "max_frame_width_minus_1" and "max_frame_height_minus_1" in the AV2 spec.) I just wrote PR https://github.com/AOMediaCodec/libavif/pull/3238 to note this fact.

PR #3234, opened 2026-05-31
Avoid self-copy data loss in avifImageCopy()

jmestwa-coder:
## Summary

`avifImageCopy()` does not handle self-copy correctly.

When the same image is passed as both source and destination:

```c
avifImageCopy(image, image, AVIF_PLANES_ALL);
```

the function frees the image planes before copying, causing owned image data to be discarded while still returning `AVIF_RESULT_OK`.

## Fix

Add an early return for self-copy operations:

```c
if (dstImage == srcImage) {
    return AVIF_RESULT_OK;
}
```

This preserves existing image data and avoids unnecessary work when the source and destination are the same object.

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -251,6 +251,9 @@ static avifResult avifImageCopyProperties(avifImage * dstImage, const avifImage 
 
 avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
 {
+    if (dstImage == srcImage) {
+        return AVIF_RESULT_OK;
+    }
     avifImageFreePlanes(dstImage, AVIF_PLANES_ALL);
     avifImageCopyNoAlloc(dstImage, srcImage);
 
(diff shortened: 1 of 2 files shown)


y-guyon:

I would consider this as incorrect use of the API and in my opinion there is no need to handle that case nor explain it in the API comment.

Why would a user call `avifImageCopy(image, image, planes)`?

jmestwa-coder:
@y-guyon Thank you for the clarification.

I viewed self-copy as a valid operation because the API does not currently document that the source and destination must be different.

My concern was that the function currently returns `AVIF_RESULT_OK` while modifying the image when `dstImage == srcImage`.

Would you prefer this to remain undocumented caller responsibility, or would documenting `dstImage != srcImage` as a precondition be worthwhile?

wantehchang:
In API design it seems more elegant to allow self copy and handle it properly, similar to the assignment operator `=` in C++.

But I agree with Yannis that calling `avifImageCopy(image, image, planes)` is extremely unlikely and would almost always indicate a programming error. We probably should strive to detect such bugs rather than ignore them. I suggest we change your check for self copy to return an error.

There are two other libavif public functions that take two `avifImage` parameters:
- avifImageSetViewRect: This should require `dstImage != srcImage`. See PR https://github.com/AOMediaCodec/libavif/pull/3236.
- avifImageComputeGainMap: I think it is fine if `baseImage == altImage`.

wantehchang:
This PR has been replaced by PR https://github.com/AOMediaCodec/libavif/pull/3236.

PR #3236, opened 2026-06-01
Disallow dstImage == srcImage

wantehchang:
Disallow dstImage == srcImage in avifImageCopy() and avifImageSetViewRect().

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -251,6 +251,9 @@ static avifResult avifImageCopyProperties(avifImage * dstImage, const avifImage 
 
 avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
 {
+    if (dstImage == srcImage) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
     avifImageFreePlanes(dstImage, AVIF_PLANES_ALL);
     avifImageCopyNoAlloc(dstImage, srcImage);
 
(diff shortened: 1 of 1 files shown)


src/avif.c
```diff
@@ -251,6 +251,9 @@ static avifResult avifImageCopyProperties(avifImage * dstImage, const avifImage
 
 avifResult avifImageCopy(avifImage * dstImage, const avifImage * srcImage, avifPlanesFlags planes)
 {
+    if (dstImage == srcImage) {
+        return AVIF_RESULT_INVALID_ARGUMENT;
+    }
```
y-guyon: Suggestion:

```c
AVIF_CHECKERR(dstImage != srcImage, AVIF_RESULT_INVALID_ARGUMENT);
```

y-guyon (approved):

src/avif.c
wantehchang: Done.

y-guyon (approved):

vrabaud:
@y-guyon , wouldn't it be better to actually check for a partial overlap of the buffers?

y-guyon:
> wouldn't it be better to actually check for a partial overlap of the buffers?

I assume you mean for `avifImageSetViewRect()` and not `avifImageCopy()`.  
The issue could be ownership. If `image->imageOwnsYUVPlanes` and `avifImageSetViewRect(image, image, {1,0,1,1})`, `avifImageDestroy(image)` cannot free the originally allocated memory region at its first byte because `avifImage` only stores one pointer per plane:

https://github.com/AOMediaCodec/libavif/blob/c64ea9468ddf5f0b3cdf14d07da3e200c8c208b5/src/avif.c#L488-L495

In the case of `!image->imageOwnsYUVPlanes` that could be useful, but then the API gets tricky. I think it is simpler to forbid it for now.

PR #3245, opened 2026-06-08
cast to unsigned char before ctype calls in image readers

dxbjavid:
avifHexStringToBytes() in avifpng.c and the gain map property parser in avifjpeg.c hand a plain char straight to isxdigit()/isspace(), so a byte above 0x7f coming from a png raw profile or an xmp property value promotes to a negative int and falls outside the [-1, 255] domain those functions are defined for, which is undefined behaviour and can index before the ctype table on some libcs. cast to unsigned char first, the same way read.c and avifutil.c already do for their isprint()/tolower() calls.

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -766,7 +766,7 @@ static avifBool avifJPEGFindGainMapPropertyDoubles(const xmlNode * descriptionNo
             // Make sure that remaining characters (if any) are only whitespace.
             const int len = (int)strlen(textValues[i]);
             while (charsRead < len) {
-                if (!isspace(textValues[i][charsRead])) {
+                if (!isspace((unsigned char)textValues[i][charsRead])) {
                     return AVIF_FALSE; // Invalid character.
                 }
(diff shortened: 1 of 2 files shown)


apps/shared/avifjpeg.c
```diff
@@ -766,7 +766,7 @@ static avifBool avifJPEGFindGainMapPropertyDoubles(const xmlNode * descriptionNo
             // Make sure that remaining characters (if any) are only whitespace.
             const int len = (int)strlen(textValues[i]);
             while (charsRead < len) {
-                if (!isspace(textValues[i][charsRead])) {
+                if (!isspace((unsigned char)textValues[i][charsRead])) {
```
wantehchang, line 769: This is a good bug fix. Not casting the input parameter to `unsigned char` when calling <ctype.h> functions is a common programming error in C.

I verified that this PR is complete.

dxbjavid:
@wantehchang @y-guyon 
Thanks for taking the time to review, verify, and merge the fix. Much appreciated.

PR #3248, opened 2026-06-11
Replace archived release asset upload action

Alb3e3:
This removes the dependency on the archived `actions/upload-release-asset` action from the release artifact workflows.

The workflows now use `gh release upload` with the existing `contents: write` permission and `GITHUB_TOKEN`-backed authentication. This preserves the current release asset upload behavior while avoiding an archived action in the release path.

I verified the workflow YAML still parses, `git diff --check` passes, and `zizmor` no longer reports archived action usage for these workflows.

### DIFF
--- a/.github/workflows/ci-macos-artifacts.yml
+++ b/.github/workflows/ci-macos-artifacts.yml
@@ -59,15 +59,10 @@ jobs:
             avifgainmaputil
             README.txt
       - name: Upload artifacts
-        uses: actions/upload-release-asset@e8f9f06c4b078e705bd2ea027f0926603fc9b4d5 # v1.0.2
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          # See https://docs.github.com/en/webhooks/webhook-events-and-payloads#release.
-          upload_url: ${{ github.event.release.upload_url }}
-          asset_path: build/${{ runner.os }}-artifacts.zip
-          asset_name: ${{ runner.os }}-artifacts.zip
-          asset_content_type: application/zip
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        run: gh release upload "$RELEASE_TAG" "build/${{ runner.os }}-artifacts.zip"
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-macos-artifacts.yml
```diff
@@ -59,15 +59,11 @@ jobs:
...
-        uses: actions/upload-release-asset@e8f9f06c4b078e705bd2ea027f0926603fc9b4d5 # v1.0.2
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          # See https://docs.github.com/en/webhooks/webhook-events-and-payloads#release.
-          upload_url: ${{ github.event.release.upload_url }}
-          asset_path: build/${{ runner.os }}-artifacts.zip
-          asset_name: ${{ runner.os }}-artifacts.zip
-          asset_content_type: application/zip
+          ARTIFACT_PATH: build/${{ runner.os }}-artifacts.zip
```
y-guyon: For simplicity and consistency, I suggest removing `ARTIFACT_PATH` and using `build/macos-artifacts.zip` directly in the `run:` line. Same for windows.

.github/workflows/ci-macos-artifacts.yml
```diff
@@ -59,15 +59,11 @@ jobs:
             avifgainmaputil
             README.txt
```
y-guyon, line 60: I wonder if this should also be replaced with a command line rather than a third-party action.

y-guyon (approved):
Thank you for the proposed change.

.github/workflows/ci-macos-artifacts.yml
Alb3e3: Updated in 574e56b: removed ARTIFACT_PATH from both upload steps and pass the archive paths directly to gh release upload. I kept the existing macOS archive name expression so the generated asset path/name stays unchanged.

y-guyon (approved):

wantehchang:
alb3e3:  Thank you for the pull request. Just curious: what is an archived action?

wantehchang:
alb3e3: I am answering my own question. I found the answer at https://github.com/actions/upload-release-asset#github-action---releases-api. An archived action means an unmaintained action.

PR #3249, opened 2026-06-11
Guard JPEG XMP parsing sizes

Alb3e3:
## Summary
- route JPEG gain-map XMP parsing through one helper that rejects buffers larger than libxml's int-sized API can represent
- apply the guard to gain-map detection, gain-map metadata parsing, and extended-XMP merging
- add a regression test that passes an oversized XMP length without allocating it

## Testing
- git diff --check
- /tmp/cmake-3.29.6-linux-x86_64/bin/cmake -S . -B build-xmp-size -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=RelWithDebInfo -DAVIF_BUILD_TESTS=ON -DAVIF_GTEST=LOCAL -DAVIF_JPEG=LOCAL -DAVIF_LIBXML2=LOCAL -DAVIF_ZLIBPNG=LOCAL -DAVIF_CODEC_AOM=OFF -DAVIF_CODEC_DAV1D=OFF -DAVIF_CODEC_LIBGAV1=OFF -DAVIF_CODEC_RAV1E=OFF -DAVIF_CODEC_SVT=OFF -DAVIF_LIBYUV=OFF -DAVIF_LIBSHARPYUV=OFF -DAVIF_ENABLE_WERROR=ON -DAVIF_BUILD_APPS=OFF -DAVIF_ENABLE_GOLDEN_TESTS=OFF
- /tmp/cmake-3.29.6-linux-x86_64/bin/cmake --build build-xmp-size --target avifjpeggainmaptest -- -j2
- build-xmp-size/tests/avifjpeggainmaptest tests/data/

### DIFF
--- a/tests/gtest/avifjpeggainmaptest.cc
+++ b/tests/gtest/avifjpeggainmaptest.cc
@@ -3,6 +3,8 @@
 
 #include <math.h>
 
+#include <limits>
+
 #include "avif/avif.h"
 #include "avifjpeg.h"
 #include "aviftest_helpers.h"
@@ -331,6 +333,15 @@ TEST(JpegTest, EmptyXMP) {
   avifBool is_avif_gain_map;
   EXPECT_FALSE(avifJPEGParseGainMapXMP((const uint8_t*)xmp.data(), xmp.size(),
                                        gain_map.get(), &is_avif_gain_map));
+}
+
+TEST(JpegTest, TooLargeXMP) {
+  const uint8_t xmp = 0;
+  GainMapPtr gain_map(avifGainMapCreate());
+  avifBool is_avif_gain_map;
+  EXPECT_FALSE(avifJPEGParseGainMapXMP(
+      &xmp, static_cast<size_t>(std::numeric_limits<int>::max()) + 1,
+      gain_map.get(), &is_avif_gain_map));
 }
 
 //------------------------------------------------------------------------------
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -7,6 +7,7 @@
 
 #include <assert.h>
 #include <ctype.h>
+#include <limits.h>
 #include <math.h>
 #include <setjmp.h>
 #include <stdint.h>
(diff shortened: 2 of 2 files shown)


maryla-uc (approved):
Looks good to me. Thanks for the fix!

tests/gtest/avifjpeggainmaptest.cc
```diff
@@ -333,6 +335,15 @@ TEST(JpegTest, EmptyXMP) {
                                        gain_map.get(), &is_avif_gain_map));
 }
 
+TEST(JpegTest, TooLargeXMP) {
+  const uint8_t xmp = 0;
+  GainMapPtr gain_map(avifGainMapCreate());
+  avifBool is_avif_gain_map;
+  EXPECT_FALSE(avifJPEGParseGainMapXMP(
+      &xmp, static_cast<size_t>(std::numeric_limits<int>::max()) + 1,
```
wantehchang, line 343: This test passes an incorrect xmp buffer size  to the `avifJPEGParseGainMapXMP()` function. (The correct xmp buffer size is 1.) We should avoid this.

This test assumes the only thing `avifJPEGParseGainMapXMP()` does with its `xmpSize` parameter is to pass it to `avifJPEGReadXMLMemory()`. We should not assume that.

Another problem with this test is that the `xmp` buffer does not contain valid data, so if I remove the `size > INT_MAX` check in `avifJPEGReadXMLMemory()`, the test still passes (because `avifJPEGReadXMLMemory()` still fails, for the wrong reason).

I fixed both of these problems in PR https://github.com/AOMediaCodec/libavif/pull/3250.

wantehchang:
In the summary of this PR, we have:

> - add a regression test that passes an oversized XMP length without allocating it

I have seen this technique in several regression tests recently. The problem with this technique is that the XMP length is not only oversized but also incorrect.

Did you write this regression test manually, or by using an AI tool? If you used an AI tool, please instruct the tool to NOT use an incorrect buffer size, i.e., the test should pass the actual allocated buffer size to the function under test.

Alb3e3:
Thanks for catching this. Yes, the regression test was AI-assisted, and I missed that it passed a size that did not match the actual allocated buffer. That made the test weaker than intended and could let it pass for the wrong reason. I'll make sure future tests pass the actual allocated buffer size and validate the intended failure path. Thanks for fixing it in #3250.

PR #3250, opened 2026-06-11
Use valid xmp_data buffer in JpegTest.TooLargeXMP

wantehchang:
Actually allocate a too large xmp_data buffer. Make sure the avifJPEGParseGainMapXMP() call would succeed if the size > INT_MAX check were removed.

A follow-up to PR https://github.com/AOMediaCodec/libavif/pull/3249.

### DIFF
--- a/tests/gtest/avifjpeggainmaptest.cc
+++ b/tests/gtest/avifjpeggainmaptest.cc
@@ -6,6 +6,8 @@
 #include <stdint.h>
 #include <string.h>
 
+#include <limits>
+#include <memory>
 #include <new>
 
 #include "avif/avif.h"
@@ -335,6 +337,15 @@ TEST(JpegTest, EmptyXMP) {
   avifBool is_avif_gain_map;
   EXPECT_FALSE(avifJPEGParseGainMapXMP((const uint8_t*)xmp.data(), xmp.size(),
                                        gain_map.get(), &is_avif_gain_map));
+}
+
+TEST(JpegTest, TooLargeXMPFake) {
+  const uint8_t xmp = 0;
+  GainMapPtr gain_map(avifGainMapCreate());
+  avifBool is_avif_gain_map;
+  EXPECT_FALSE(avifJPEGParseGainMapXMP(
+      &xmp, static_cast<size_t>(std::numeric_limits<int>::max()) + 1,
+      gain_map.get(), &is_avif_gain_map));
 }
 
 // This test requires a 64-bit size_t type.
--- a/tests/gtest/avifjpeggainmaptest.cc
+++ b/tests/gtest/avifjpeggainmaptest.cc
@@ -1,9 +1,12 @@
 // Copyright 2023 Google LLC
 // SPDX-License-Identifier: BSD-2-Clause
 
+#include <limits.h>
 #include <math.h>
+#include <stdint.h>
+#include <string.h>
 
-#include <limits>
+#include <new>
 
 #include "avif/avif.h"
 #include "avifjpeg.h"
@@ -215,8 +218,7 @@ TEST(JpegTest, ParseXMPAllDefaultValues) {
     </rdf:Description>
   </rdf:RDF>
 </x:xmpmeta>
-<?xpacket end="w"?>
-  )";
+<?xpacket end="w"?>)";
   GainMapPtr gain_map(avifGainMapCreate());
   avifBool is_avif_gain_map;
   ASSERT_TRUE(avifJPEGParseGainMapXMP((const uint8_t*)xmp.data(), xmp.size(),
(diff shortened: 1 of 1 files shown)


tests/gtest/avifjpeggainmaptest.cc
```diff
@@ -215,8 +218,7 @@ TEST(JpegTest, ParseXMPAllDefaultValues) {
     </rdf:Description>
   </rdf:RDF>
 </x:xmpmeta>
-<?xpacket end="w"?>
-  )";
+<?xpacket end="w"?>)";
```
wantehchang, line 223: I removed the unnecessary newline and spaces at the end of the xmp string. I then copied this xmp string to the new test.

wantehchang (commented):
cc: @Alb3e3

tests/gtest/avifjpeggainmaptest.cc
```diff
@@ -335,14 +337,33 @@ TEST(JpegTest, EmptyXMP) {
...
+      xmlns:hdrgm="http://ns.adobe.com/hdr-gain-map/1.0/" hdrgm:Version="1.0">
+    </rdf:Description>
+  </rdf:RDF>
+</x:xmpmeta>
+<?xpacket end="w"?>)";
+  // When xmp_size is cast to int, it is equal to xmp.size().
+  const size_t xmp_size = size_t{UINT_MAX} + 1 + xmp.size();
+  std::unique_ptr<uint8_t[]> xmp_data(new (std::nothrow) uint8_t[xmp_size]);
+  if (!xmp_data) {
+    GTEST_SKIP() << "Allocation of " << xmp_size << " bytes failed";
```
maryla-uc, line 367: Should we run the test without allocating in that case? I feel like this more correct test might end up testing less if it gets skipped?
I originally thought the version that didn't allocate was good enough...

maryla-uc (approved):

tests/gtest/avifjpeggainmaptest.cc
wantehchang: Indeed, this more correct test hurts test coverage. Note that it is also skipped on 32-bit platforms. See the `#if SIZE_MAX > UINT_MAX` at line 341.

I understand that not everyone objects to a test that deliberately passes an incorrect buffer size to the function under test. (The reason I object to this practice is that the function under test cannot detect an incorrect buffer size except in trivial cases such as a 0 or negative buffer size.)

In this particular case, the original test has another, perhaps more serious problem: the test still passes if we remove the bug fix. This experiment shows that:

```
$ git diff
diff --git a/apps/shared/avifjpeg.c b/apps/shared/avifjpeg.c
index 426c5d0a..c1eda541 100644
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -582,9 +582,7 @@ static xmlNode * avifJPEGFindXMLNodeByName(const xmlNode * parentNode, const cha
 
 static xmlDoc * avifJPEGReadXMLMemory(const uint8_t * data, size_t size, const char * url)
 {
-    if (size > INT_MAX) {
-        return NULL;
-    }
+    fprintf(stderr, "Passing (int)size=%d to xmlReadMemory()\n", (int)size);
     return xmlReadMemory((const char *)data, (int)size, url, NULL, /*options=*/0);
 }
 
$ tests/avifjpeggainmaptest --gtest_filter=JpegTest.TooLargeXMP ../tests/data/
Note: Google Test filter = JpegTest.TooLargeXMP
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from JpegTest
[ RUN      ] JpegTest.TooLargeXMP
Passing (int)size=-2147483648 to xmlReadMemory()
[       OK ] JpegTest.TooLargeXMP (0 ms)
[----------] 1 test from JpegTest (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 1 test.
```

The reason the test still passes is that the `avifJPEGReadXMLMemory()` function still fails, for the wrong reason. The test is a regression test but it fails to detect the bug.

It's a little tricky to make the test detect the bug. The method I came up with is to make `avifJPEGReadXMLMemory()` succeed if the bug fix is removed. Unfortunately this method requires a 64-bit `size_t` type.

maryla-uc: Thanks for the extra info. Indeed if it still passed without the bugfix, it was not a very useful test. Your version is better then.

wantehchang: I decided to be less pedantic and keep the original JpegTest.TooLargeXMP test. I renamed it JpegTest.TooLargeXMPFake.

The new policy is that a test can pass to the function under test a buffer size larger than the actual allocation size under the following conditions:
- it is possible to allocate that buffer size if we want to (i.e., we are just taking a short cut), and
- the function under test checks the buffer size before accessing the buffer.

PR #3251, opened 2026-06-11
Avoid const cast of avifJPEGFindXMLNodeByName rv

wantehchang:
Change avifJPEGFindXMLNodeByName() to return a non-const xmlNode * so that we don't need apply a const cast to its return value.

Also fix an incorrect error message (we are parsing an extended XMP segment, not a standard XMP segment).

### DIFF
--- a/apps/shared/avifjpeg.c
+++ b/apps/shared/avifjpeg.c
@@ -556,17 +556,17 @@ static avifBool avifJPEGFindMpfSegmentOffset(FILE * f, uint32_t * mpfOffset)
 
 // Searches for a node called 'nameSpace:nodeName' in the children (or descendants if 'recursive' is set) of 'parentNode'.
 // Returns the first such node found (in depth first search). Returns NULL if no such node is found.
-static const xmlNode * avifJPEGFindXMLNodeByName(const xmlNode * parentNode, const char * nameSpace, const char * nodeName, avifBool recursive)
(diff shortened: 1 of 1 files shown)


apps/shared/avifjpeg.c
```diff
@@ -1137,10 +1137,10 @@ static avifBool avifJPEGMergeXMP(const uint8_t * standardXMPData,
         isValid = AVIF_FALSE;
         goto cleanup_xml;
     }
-    xmlNode * xmpRdf = (xmlNode *)avifJPEGFindXMLNodeByName(xmlDocGetRootElement(xmpDoc),
```
wantehchang, line 1140: This is the const cast that I want to avoid. This return value will be passed to `xmlAddChild(xmpRdf, childCopy)`, which requires a non-const `xmlNode` pointer.

apps/shared/avifjpeg.c
```diff
@@ -1199,7 +1199,7 @@ static avifBool avifJPEGMergeXMP(const uint8_t * standardXMPData,
                                                                "RDF",
                                                                /*recursive=*/AVIF_TRUE);
     if (!extendedXMPRdf) {
-        fprintf(stderr, "XMP extraction failed: invalid standard XMP segment\n");
+        fprintf(stderr, "XMP extraction failed: cannot find RDF node\n");
```
wantehchang, line 1202: I copied this error message from line 1145.

maryla-uc (approved):

PR #3254, opened 2026-06-15
Update VS version in Windows CI

y-guyon:
This change avoids errors such as:

```
CMake Error at CMakeLists.txt:34 (project):
  Generator
    Visual Studio 17 2022
  could not find any instance of Visual Studio.
```

([log](https://github.com/AOMediaCodec/libavif/actions/runs/27531891859/job/81371768152))

Likely related: https://github.com/actions/runner-images/issues/14017

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -33,7 +33,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest]
-        generator: [Ninja, "Visual Studio 17 2022"]
+        generator: [Ninja, "Visual Studio 18 2026"]
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -33,7 +33,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest]
-        generator: [Ninja, "Visual Studio 18 2026"]
+        generator: ["-G Ninja", ""]
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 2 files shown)


y-guyon:
This fixes the `could not find any instance of Visual Studio` compile error but the distortion failures remain. See https://github.com/AOMediaCodec/libavif/issues/3255.

vrabaud:
Sorry, I forgot to port that fix: I ended up doing that https://github.com/webmproject/libsmpte2094-50/blob/023ec1619aad8e983627d5f50ad570850224198c/.github/workflows/ci-windows.yml#L24    It is more forward compatible

y-guyon:
> Sorry, I forgot to port that fix: I ended up doing that https://github.com/webmproject/libsmpte2094-50/blob/023ec1619aad8e983627d5f50ad570850224198c/.github/workflows/ci-windows.yml#L24    It is more forward compatible

Done. Are we certain the default generator is and will stay Visual Studio?

vrabaud:
There is no guarantee no ... It's just the current CMake behavior

vrabaud (approved):

.github/workflows/ci-windows.yml
wantehchang, line 47: Since the old generator value "Visual Studio 17 2022" also contained spaces, I assume that it is also safe to use "-G Ninja" as the extra cache key.

wantehchang, line 74: We need to update the comparison here and at line 78 to compare with "-G Ninja" instead of 'Ninja'.

wantehchang (commented):
Thanks for the fix. Please see my suggested change.

wantehchang:
Re: the default generator on Windows: Historically the build tools provided by the OS vendor have a higher status. So the default generator on Windows is likely to stay as Visual Studio, even though it is not guaranteed.

.github/workflows/ci-windows.yml
wantehchang: I took care of this in PR https://github.com/AOMediaCodec/libavif/pull/3258.

Issue #3255, opened 2026-06-15
High distortion with SVT-AV1 in CI Windows


y-guyon:
`tests/gtest/avifcodectest.cc` fails at head. It was passing 3 days ago.

<details>

<summary><code>Expected: (testutil::GetPsnr(*image, *decoded)) > (32.0), actual: 12 vs 32</code></summary>

```
[ RUN      ] All/CodecTest.EncodeDecode/6
Svt[info]: -------------------------------------------
Svt[info]: SVT [version]:	SVT-AV1 Encoder Lib v3.1.2
Svt[info]: SVT [build]  :	Visual Studio 2022 64 bit
Svt[info]: -------------------------------------------
Svt[info]: Level of Parallelism: 1
Svt[info]: Number of PPCS 74
Svt[info]: [asm level on system : up to avx2]
Svt[info]: [asm level selected : up to avx2]
Svt[info]: -------------------------------------------
Svt[info]: SVT [config]: main profile	tier (auto)	level (auto)
Svt[info]: SVT [config]: width / height / fps numerator / fps denominator 		: 64 / 64 / 60000 / 1000
Svt[info]: SVT [config]: bit-depth / color format 					: 8 / YUV420
Svt[info]: SVT [config]: preset / tune / pred struct 					: 8 / PSNR / random access
Svt[info]: SVT [config]: gop size / mini-gop size / key-frame type 			: 321 / 32 / key frame
Svt[info]: SVT [config]: BRC mode / rate factor 					: CRF / 6 
Svt[info]: SVT [config]: AQ mode / Variance Boost 					: 2 / 0
Svt[info]: SVT [config]: sharpness / luminance-based QP bias 				: 0 / 0
Svt[info]: -------------------------------------------
Svt[info]: -------------------------------------------
Svt[info]: SVT [version]:	SVT-AV1 Encoder Lib v3.1.2
Svt[info]: SVT [build]  :	Visual Studio 2022 64 bit
Svt[info]: -------------------------------------------
Svt[info]: Level of Parallelism: 1
Svt[info]: Number of PPCS 74
Svt[info]: [asm level on system : up to avx2]
Svt[info]: [asm level selected : up to avx2]
Svt[info]: -------------------------------------------
Svt[info]: SVT [config]: main profile	tier (auto)	level (auto)
Svt[info]: SVT [config]: width / height / fps numerator / fps denominator 		: 64 / 64 / 60000 / 1000
[...]

PR #3257, opened 2026-06-16
Patch SVT-AV1 to avoid corrupted files with MSVC

y-guyon:


y-guyon:
This should fix https://github.com/AOMediaCodec/libavif/issues/3255 for `CI Windows` but not `CI Windows Installed`.

cmake/Modules/LocalSvt.cmake
```diff
@@ -42,6 +42,7 @@ else()
         GIT_TAG "${AVIF_SVT_GIT_TAG}"
         UPDATE_COMMAND ""
         GIT_SHALLOW ON
+        PATCH_COMMAND git apply ${AVIF_SOURCE_DIR}/cmake/Modules/LocalSvt.diff
```
wantehchang, line 46: Please add a comment citing https://github.com/AOMediaCodec/libavif/issues/3255 or the upstream SVT-AV1 issue (https://gitlab.com/AOMediaCodec/SVT-AV1/-/work_items/2322 ?).

wantehchang (approved):
LGTM. Thanks for the investigation and workaround.

wantehchang:
Yannis: PR https://github.com/AOMediaCodec/libavif/pull/3259 allows `CI Windows Installed` to succeed. We can use PR https://github.com/AOMediaCodec/libavif/pull/3259 instead, or use a half-and-half solution.

cmake/Modules/LocalSvt.cmake
y-guyon: Done.

y-guyon:
> Yannis: PR #3259 allows `CI Windows Installed` to succeed. We can use PR #3259 instead, or use a half-and-half solution.

While there is little to be done for installed SVT-AV1 binaries on the libavif side (maybe displaying a warning when compiling libavif?), this PR helps for a locally built SVT-AV1 dependency. Fixing the CI itself avoids noise but does not help users of libavif. I would suggest the half-and-half approach.

PR #3260, opened 2026-06-17
Fix integer overflows in avifImageYUVAnyToRGBAnySlow

MXguru1:
Widen loop counters and rowBytes variables to size_t to prevent 32-bit unsigned integer overflow when calculating row offsets for large images in the slow format conversion path.


src/reformat.c
y-guyon, line 814: Should these be adapted too? I see `size_t + int`.

y-guyon (commented):

wantehchang:

src/reformat.c
wantehchang: Yes. In PR https://github.com/AOMediaCodec/libavif/pull/3269 I use `ptrdiff_t` instead of `int` and add the `ptrdiff_t` variables to a pointer instead of a `size_t` variable.

Issue #3263, opened 2026-06-17
cross-compilation: failure to cross compile on Ubunto for Windows


vtorri:
commands to compile:

```
cmake \
    -DCMAKE_TOOLCHAIN_FILE=../cross_toolchain.txt \
    -DCMAKE_INSTALL_PREFIX=$prefix_unix \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=TRUE \
    -DAVIF_CODEC_AOM=SYSTEM \
    -DAVIF_CODEC_DAV1D=SYSTEM \
    -DAVIF_BUILD_APPS:BOOL=OFF \
    -G Ninja \
    .. 
```

cross compilation file:

```set(CMAKE_SYSTEM_NAME Windows)
set(CMAKE_SYSTEM_PROCESSOR @proc@)

# binaries
set(CMAKE_C_COMPILER @host@-gcc)
set(CMAKE_CXX_COMPILER @host@-g++)
set(CMAKE_LINKER @host@-ld)
set(CMAKE_STRIP @host@-strip)
set(CMAKE_RC_COMPILER @host@-windres)

set(CMAKE_FIND_ROOT_PATH @prefix@)

# flags
set(CMAKE_C_FLAGS "-I@prefix@/include -D__USE_MINGW_ANSI_STDIO=0 -D_WIN32_WINNT=@winver@")
set(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS}")

set(CMAKE_C_FLAGS_DEBUG "-Og -g3 -ggdb3")
set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG}")

set(CMAKE_C_FLAGS_RELEASE "-O2 -pipe -march=native")
set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")

set(CMAKE_SHARED_LINKER_FLAGS "-L@prefix@/lib")
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_SHARED_LINKER}")

```

output:

```[1/26] Building C object CMakeFiles/avif_obj.dir/src/diag.c.obj
[2/26] Building C object CMakeFiles/avif_obj.dir/src/mem.c.obj
[3/26] Building C object CMakeFiles/avif_obj.dir/src/properties.c.obj
[...]

PR #3266, opened 2026-06-17
Link with -lm when NOT WIN32 AND NOT APPLE

wantehchang:
Change the condition to link with the math library (-lm) to if(NOT WIN32 AND NOT APPLE). This condition has worked well in libaom.

Address the -l_aom_dep_lib_m-NOTFOUND part of
issue https://github.com/AOMediaCodec/libavif/issues/3263.


wantehchang:
cc: @vtorri

Hi vtorri,

Could you please test this pull request and see if it fixed the `-l_aom_dep_lib_m-NOTFOUND` part of issue https://github.com/AOMediaCodec/libavif/issues/3263? Thanks.

jzern (approved):

Issue #3275, opened 2026-06-19
SEGV / out-of-bounds read in `avifROStreamReadBoxHeaderPartial()` on a crafted ISOBMFF box header (src/stream.c)


OwenSanzas:
## Summary

While parsing the header of an ISOBMFF box, `avifROStreamReadBoxHeaderPartial()` reads a 32-bit box size via `avifROStreamReadU32()` → `avifROStreamRead()` without sufficiently validating the stream cursor against the remaining input length. On a crafted, truncated 8-byte box header the read dereferences a wild / out-of-range pointer, causing a segmentation fault. This is reachable through the public libavif decode/parse path on an untrusted file.

Confirmed reproducing on current `main` (HEAD `3f4212f`, 2026-06-19).

## Root Cause

`avifROStreamReadBoxHeaderPartial()` immediately reads the box `size` and `type` fields from the stream. The underlying `avifROStreamReadU32()` copies `sizeof(uint32_t)` bytes from the stream cursor, but the cursor/remaining-length validation is insufficient for a crafted header, so the read goes out of bounds.

`src/stream.c`:

```c
avifBool avifROStreamReadU32(avifROStream * stream, uint32_t * v)
{
    assert(stream->numUsedBitsInPartialByte == 0); // Byte alignment is required.
    AVIF_CHECK(avifROStreamRead(stream, (uint8_t *)v, sizeof(uint32_t)));   // stream.c:128 — OOB read
    *v = avifNTOHL(*v);
    return AVIF_TRUE;
}

avifBool avifROStreamReadBoxHeaderPartial(avifROStream * stream, avifBoxHeader * header, avifBool topLevel)
{
    // Section 4.2.2 of ISO/IEC 14496-12.
    size_t startOffset = stream->offset;

    uint32_t smallSize;
    AVIF_CHECK(avifROStreamReadU32(stream, &smallSize));   // stream.c:254 — unsigned int(32) size;
    AVIF_CHECK(avifROStreamRead(stream, header->type, 4)); //              unsigned int(32) type = boxtype;
    ...
}
```

The crash chain is `avifROStreamReadBoxHeaderPartial` (stream.c:254) → `avifROStreamReadU32` (stream.c:128) → `avifROStreamRead`, which dereferences an out-of-range pointer.

## PoC

An 8-byte crafted ISOBMFF box header (`segv_avifROStreamRead.bin`) fed to the libavif decode/parse entry point.

## Reproduction

[...]

PR #3276, opened 2026-06-19
Add the BasicTest.DecodeIssue3275 test

wantehchang:
Add a test to reproduce issue #3275.


wantehchang:
Abandoned this PR because it does not reproduce the bug alleged in issue https://github.com/AOMediaCodec/libavif/issues/3275.

PR #3278, opened 2026-06-24
No need to destroy gainMap if ignore_gain_map

wantehchang:
In avif::ReadImage(), if ignore_gain_map is true,
decoder->imageContentToDecode has the default value AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA, decoder->image->gainMap must be null and doesn't need to be destroyed. Add an assertion to assert that.

Also explicitly unset AVIF_IMAGE_CONTENT_GAIN_MAP in decoder->imageContentToDecode if ignore_gain_map is true in case the default value of decoder->imageContentToDecode changes in the future.

### DIFF
--- a/apps/avifgainmaputil/imageio.cc
+++ b/apps/avifgainmaputil/imageio.cc
@@ -3,6 +3,7 @@
 
 #include "imageio.h"
 
+#include <cassert>
 #include <cstring>
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


maryla-uc (commented):
I think the gainMap metadata (decoder->image->gainMap) is always present if a gainmap is detected. It's the gainmap's image (decoder->image->gainMap->image) that is decoded or not based on imageContentToDecode.

wantehchang:
Thanks for the explanation.

Issue #3277, opened 2026-06-22
avif::ReadAvif() in apps/avifgainmaputil/imageio.cc should not modify decoder->image


wantehchang:
In apps/avifgainmaputil/imageio.cc, the `avif::ReadAvif()` function has the following code at the end of the function:

```
avifResult ReadAvif(avifDecoder* decoder, const std::string& input_filename,
                    bool ignore_profile, bool ignore_alpha) {
  ...
  if (ignore_profile) {
    avifRWDataFree(&decoder->image->icc);
    if (decoder->image->gainMap) {
      avifRWDataFree(&decoder->image->gainMap->altICC);
    }
  }
  if (ignore_alpha && decoder->image->alphaPlane) {
    avifImageFreePlanes(decoder->image, AVIF_PLANES_A);
  }

  return AVIF_RESULT_OK;
}
```

Since `decoder->image` is owned by `decoder`, clients of `decoder` should treat `decoder->image` as read-only.

Although avifgainmaputil is part of the libavif project, it should act as a client of the libavif library. So `avif::ReadAvif()` should not modify `decoder->image`.

We can consider calling `avifImageCopy()` to copy `decoder->image` to an output `avifImage` object that we own.

maryla-uc:
Indeed `image` is marked as an output parameter "which users of the API may read". So we're not supposed to modify them.
I guess the risk is that the decoder would have some kind of invariant based on the image, and that modifying the image would break this invariant.
In practice I don't think it causes any issues, especially if we are done with the decoder at this point anway.
It seems a bit wasteful to do full image copy (which can be quite large) just for this.
There's actually a bunch more places in avifgainmaputil where we are modifying decoder->image.
E.g. in `ReadImage` we do `std::swap(*image, *decoder->image);` which avoids a copy, but is actually also modifying `decoder->image` so I guess it's wrong as well...

Actually in `apps/avifdec.c` we also modify the image's icc https://github.com/AOMediaCodec/libavif/blob/32393ae1eefc568feed206d126434cd0c28893fd/apps/avifdec.c#L403-L417
Should avifdec follow the same rule?

wantehchang:
Thank you for looking into this. Yes, avifdec should follow the same policy.

The `std::swap(*image, *decoder->image);` operation in `avif::ReadImage()` is the worst among these. I suggest we deal with that first.

[...]

PR #3280, opened 2026-06-25
Add the ignoreICC option to avifDecoder

wantehchang:
This allows us to remove the ignoreColorProfile parameter from the
avifImageCreateView() function.

There is a change of behavior to avifdec:
The message "[--ignore-icc] Discarding ICC profile.\n" won't be printed
if the --ignore-icc or --icc option is specified and the image has an
ICC profile.

Bug: https://github.com/AOMediaCodec/libavif/issues/3277

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 
 ### Added since 1.4.2
 
+* Add the ignoreColorProfile option to avifDecoder
 * avifenc: add --ignore-alpha flag to discard alpha channel on encode
 * avifgainmaputil: add --ignore-alpha flag to discard alpha channel
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 
 ### Added since 1.4.2
 
-* Add the ignoreColorProfile option to avifDecoder
+* Add the ignoreICC option to avifDecoder
 * avifenc: add --ignore-alpha flag to discard alpha channel on encode
 * avifgainmaputil: add --ignore-alpha flag to discard alpha channel
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ The changes are relative to the previous release, unless the baseline is specifi
 * Fix printed gain map size in avifenc when converting a jpeg with a gain map
   and using --grid.
 * Allow setting avifDecoder::imageSizeLimit to any value.
+* avifdec: The message "[--ignore-icc] Discarding ICC profile.\n" won't be
+  printed if the --ignore-icc or --icc option is specified and the image has an
+  ICC profile.
 
 ## [1.4.2] - 2026-05-26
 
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The changes are relative to the previous release, unless the baseline is specifi
 
 ### Added since 1.4.2
 
+* Add the ignoreColorProfile option to avifDecoder
```
maryla-uc: Another minor user facing change is that the decoder no longer prints "[--ignore-icc] Discarding ICC profile." when ignoring ICC

maryla-uc (approved):
Are you planning on porting this change to crabbyavif as well?

CHANGELOG.md
wantehchang: I mentioned this in the commit message. Were you suggesting that this should also be mentioned in the changlog?

wantehchang (commented):
Thanks for the review. I've ported the `avifDecoder::ignoreColorProfile` option to CrabbyAvif as you suggested. (See CrabbyAvif feature request https://github.com/webmproject/CrabbyAvif/issues/829 and Google internal CL cl/942193170.)

CHANGELOG.md
maryla-uc: Yes that's what I meant, adding it to the changelog, since you could imagine someone having a script relying on this message for example. But this seems fairly unlikely so up to you.

wantehchang: Done. Thanks for the confirmation.

include/avif/avif.h
```diff
@@ -1384,6 +1384,10 @@ typedef struct avifDecoder
 
     // Version 1.2.0 ends here. Add any new members after this line.
     // --------------------------------------------------------------------------------------------
+
+    // Enable this to avoid reading and surfacing ICC profile to the decoded avifImage and gain map
+    // metadata.
+    avifBool ignoreICC;
```
wantehchang, line 1390: I also renamed the `ignoreColorProfile` field to `ignoreICC`. The original reason is that the name `ignore_icc` blends in better with the CrabbyAvif source code. A new reason is that the `ignoreColorProfile` parameter of the `avifReadImage()` function in apps/shared/avifutil.h also controls the setting of the colorPrimaries and transferCharacteristics fields when decoding a PNG input file:

```
// Reads an image from a file with the requested format and depth.
// ...
// May set the image's colorPrimaries, transferCharacteristics and ICC fields
// based on color information found in the image (unless ignoreColorProfile is true).
avifAppFileFormat avifReadImage(const char * filename,
                                avifAppFileFormat inputFormat,
                                avifPixelFormat requestedFormat,
                                int requestedDepth,
                                avifChromaDownsampling chromaDownsampling,
                                avifBool ignoreColorProfile,
                                avifBool ignoreExif,
                                avifBool ignoreXMP,
                                avifBool ignoreAlpha,
                                avifBool ignoreGainMap,
                                uint32_t imageSizeLimit,
                                avifImage * image,
                                uint32_t * outDepth,
                                avifAppSourceTiming * sourceTiming,
                                struct y4mFrameIterator ** frameIter);
```

So the `ignoreICC` name is more precise.

PR #3281, opened 2026-06-26
Fix direct modifications of decoder->image.

maryla-uc:
Directly modifying it goes against the API contract. Create views instead.

#3277

### DIFF
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -367,6 +367,10 @@ int main(int argc, char * argv[])
     const avifBool decodeAllFrames = frameIndex == DECODE_ALL_FRAMES;
     int currIndex = decodeAllFrames ? 0 : frameIndex;
     for (;;) {
+        if (imageView != NULL && imageView != decoder->image) {
+            avifImageDestroy(imageView);
+        }
+        imageView = NULL;
         result = decodeAllFrames ? avifDecoderNextImage(decoder) : avifDecoderNthImage(decoder, frameIndex);
         if (result != AVIF_RESULT_OK) {
             break;
@@ -401,21 +405,18 @@ int main(int argc, char * argv[])
             }
         }
 
-        if (ignoreICC || iccOverrideFilename) {
-            if (imageView != NULL && imageView != decoder->image) {
-                avifImageDestroy(imageView);
-            }
+        if (ignoreICC) {
             imageView = avifImageCreateEmpty();
             if (!imageView) {
                 fprintf(stderr, "ERROR: Out of memory\n");
                 goto cleanup;
             }
-            if (ignoreICC && (decoder->image->icc.size > 0)) {
+            if (decoder->image->icc.size > 0) {
                 printf("[--ignore-icc] Discarding ICC profile.\n");
             }
             result = avifImageCreateView(imageView,
                                          decoder->image,
-                                         /*ignoreColorProfile=*/ignoreICC,
+                                         /*ignoreColorProfile=*/AVIF_TRUE,
                                          /*ignoreAlpha=*/AVIF_FALSE,
                                          /*ignoreGainMap=*/AVIF_FALSE);
             if (result != AVIF_RESULT_OK) {
--- a/apps/avifdec.c
+++ b/apps/avifdec.c
@@ -304,6 +304,7 @@ int main(int argc, char * argv[])
 
     // ------ After this point, use 'goto cleanup;' in case of failure ------
     int returnCode = 1;
+    avifImage * imageView = NULL;
     avifDecoder * decoder = avifDecoderCreate();
     if (!decoder) {
         fprintf(stderr, "Memory allocation failure\n");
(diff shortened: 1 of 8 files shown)


maryla-uc:
Rebased on top of #3283

apps/avifdec.c
wantehchang, line 226: Note that `iccOverrideFilename` implies `ignoreICC` here. This fact will be used at line 404 below.

apps/avifdec.c
```diff
@@ -400,20 +401,37 @@ int main(int argc, char * argv[])
...
-            printf("[--ignore-icc] Discarding ICC profile.\n");
-            // This cannot fail.
-            result = avifImageSetProfileICC(decoder->image, NULL, 0);
-            assert(result == AVIF_RESULT_OK);
-        }
-
-        if (iccOverrideFilename) {
-            printf("[--icc] Setting ICC profile: %s\n", iccOverrideFilename);
-            result = avifImageSetProfileICC(decoder->image, iccOverride.data, iccOverride.size);
+        if (ignoreICC || iccOverrideFilename) {
```
wantehchang: `iccOverrideFilename` implies `ignoreICC` (see line 226), so don't need `|| iccOverrideFilename`.

apps/avifdec.c
```diff
@@ -400,20 +401,37 @@ int main(int argc, char * argv[])
...
+            if (imageView != NULL && imageView != decoder->image) {
+                avifImageDestroy(imageView);
+            }
+            imageView = avifImageCreateEmpty();
+            if (!imageView) {
+                fprintf(stderr, "ERROR: Out of memory\n");
+                goto cleanup;
+            }
+            if (ignoreICC && (decoder->image->icc.size > 0)) {
```
wantehchang: Omit `ignoreICC &&`

apps/avifdec.c
```diff
@@ -400,20 +401,37 @@ int main(int argc, char * argv[])
...
+                printf("[--ignore-icc] Discarding ICC profile.\n");
+            }
+            result = avifImageCreateView(imageView,
+                                         decoder->image,
+                                         /*ignoreColorProfile=*/ignoreICC,
```
wantehchang: Optional: Use `AVIF_TRUE` instead of `ignoreICC`

apps/avifdec.c
wantehchang, line 374: IMPORTANT: Since this statement invalidates certain fields in `decoder->image`, I suggest we destroy `imageView` before this statement to avoid dangling pointers:

```
        if (imageView != NULL && imageView != decoder->image) {
            avifImageDestroy(imageView);
        }
        imageView = NULL;
        result = decodeAllFrames ? avifDecoderNextImage(decoder) : avifDecoderNthImage(decoder, frameIndex);
```

apps/avifdec.c
```diff
@@ -400,20 +401,37 @@ int main(int argc, char * argv[])
...
-            assert(result == AVIF_RESULT_OK);
-        }
-
-        if (iccOverrideFilename) {
-            printf("[--icc] Setting ICC profile: %s\n", iccOverrideFilename);
-            result = avifImageSetProfileICC(decoder->image, iccOverride.data, iccOverride.size);
+        if (ignoreICC || iccOverrideFilename) {
+            if (imageView != NULL && imageView != decoder->image) {
+                avifImageDestroy(imageView);
+            }
```
wantehchang: Delete this if statement. See my comment at line 370.

apps/shared/avifutil.h
```diff
@@ -128,6 +128,12 @@ void avifImageFixXMP(avifImage * image);
 // This must match the cited fallback for "--yuv auto" in avifenc.c's syntax() function.
 #define AVIF_APP_DEFAULT_PIXEL_FORMAT AVIF_PIXEL_FORMAT_YUV444
 
+// Creates a shallow copy of an image (a view of the pixels, but deep copy of metadata).
+// If a gain map is present in 'srcImage', the gain map of 'dstImage' is also set to
+// a view of the original gain map.
+// 'dstImage' should be an empty image. It will not own the pixel data.
+avifResult avifImageCreateView(avifImage * dstImage, const avifImage * srcImage, avifBool ignoreColorProfile, avifBool ignoreAlpha, avifBool ignoreGainMap);
```
wantehchang, line 135: Future: This function probably should be moved to the libavif library.

Also, some of the `ignore*` parameters could be handled by new `avifDecoder` options or by the caller, for example,
- PR [#3280](https://github.com/AOMediaCodec/libavif/pull/3280)
- PR [#3284](https://github.com/AOMediaCodec/libavif/pull/3284)

We can deal with this iteratively.

apps/shared/avifutil.c
```diff
@@ -823,3 +823,94 @@ avifResult avifApplyTransforms(avifRGBImage * dstView, avifRGBImage * srcImage,
...
+
+        if (srcImage->gainMap->image) {
+            dstImage->gainMap->image = avifImageCreateEmpty();
+            if (!dstImage->gainMap->image) {
+                return AVIF_RESULT_OUT_OF_MEMORY;
+            }
+            res = avifImageCreateView(dstImage->gainMap->image,
+                                      srcImage->gainMap->image,
+                                      ignoreColorProfile,
+                                      /*ignoreAlpha=*/AVIF_TRUE,
```
wantehchang, line 907: Is a gain map image not allowed to have an alpha channel?

wantehchang (approved):

apps/shared/avifutil.h
maryla-uc: I think we wanted to stop adding new features to libavif...

apps/shared/avifutil.c
maryla-uc: No, gain maps don't have alpha. I'm not sure if it's explicitly forbidden in the spec but an alpha channel would not be used for anything anyway

wantehchang (approved):
LGTM. Thanks!

PR #3283, opened 2026-06-26
TonemapCommand::Run: Use `image` consistently

wantehchang:
Once the local variable `image` is declared and initialized to decoder->image, use `image` intead of `decoder->image` consistently.

### DIFF
--- a/apps/avifgainmaputil/tonemap_command.cc
+++ b/apps/avifgainmaputil/tonemap_command.cc
@@ -80,7 +80,7 @@ avifResult TonemapCommand::Run() {
         arg_input_cicp_.value().matrix_coefficients;
   }
 
-  avifImage* image = decoder->image;
+  avifImage* const image = decoder->image;
   if (image->gainMap == nullptr || image->gainMap->image == nullptr) {
     std::cerr << "Input image " << arg_input_filename_
               << " does not contain a gain map\n";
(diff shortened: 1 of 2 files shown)


apps/avifgainmaputil/tonemap_command.cc
```diff
@@ -80,7 +80,7 @@ avifResult TonemapCommand::Run() {
         arg_input_cicp_.value().matrix_coefficients;
   }
 
-  avifImage* image = decoder->image;
+  avifImage* const image = decoder->image;
```
wantehchang, line 83: Adding this `const` makes it easier to prove that `image` is equal to `decoder->image` after this point.

apps/avifgainmaputil/tonemap_command.cc
```diff
@@ -203,7 +203,7 @@ avifResult TonemapCommand::Run() {
   }
   avifDiagnostics diag;
   result = avifImageApplyGainMap(
-      decoder->image, image->gainMap, arg_headroom_, cicp.color_primaries,
+      image, image->gainMap, arg_headroom_, cicp.color_primaries,
```
wantehchang, line 206: I noticed this issue while reviewing your PR https://github.com/AOMediaCodec/libavif/pull/3281. Please feel free to incorporate these changes into your PR, and I can abandon this PR.

maryla-uc (approved):

PR #3284, opened 2026-06-27
Don't delete gain map metadata if ignore_gain_map

wantehchang:
In avif::ReadImage(), if ignore_gain_map is true, don't ignore gain map metadata. Let the caller do that.


wantehchang:
I abandoned this PR. Replaced by PR https://github.com/AOMediaCodec/libavif/pull/3290.

PR #3288, opened 2026-07-01
Support ignoring alpha in imageContentToDecode

wantehchang:
Add AVIF_IMAGE_CONTENT_COLOR and AVIF_IMAGE_CONTENT_ALPHA and allow
decoding color only. Decoding alpha only is still not supported.
    
Remove the ignoreAlpha parameter from the avifImageCreateView()
function. If alpha should be ignored, the caller of
avifImageCreateView() is responsible for clearing the
AVIF_IMAGE_CONTENT_ALPHA bit in decoder->imageContentToDecode before
decoding (assuming srcImage is decoder->image).

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1229,8 +1229,10 @@ AVIF_API const char * avifProgressiveStateToString(avifProgressiveState progress
 typedef enum avifImageContentTypeFlag
 {
     AVIF_IMAGE_CONTENT_NONE = 0,
-    // Color only or alpha only is not currently supported.
-    AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA = (1 << 0) | (1 << 1),
+    AVIF_IMAGE_CONTENT_COLOR = (1 << 0),
+    // Alpha only is not currently supported.
+    AVIF_IMAGE_CONTENT_ALPHA = (1 << 1),
+    AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA = AVIF_IMAGE_CONTENT_COLOR | AVIF_IMAGE_CONTENT_ALPHA,
     AVIF_IMAGE_CONTENT_GAIN_MAP = (1 << 2),
     AVIF_IMAGE_CONTENT_ALL = AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA | AVIF_IMAGE_CONTENT_GAIN_MAP,
 
--- a/src/read.c
+++ b/src/read.c
@@ -6178,59 +6178,60 @@ avifResult avifDecoderReset(avifDecoder * decoder)
             }
         }
 
+        uint32_t alphaTrackIndex = 0;
         avifCodecType alphaCodecType = AVIF_CODEC_TYPE_UNKNOWN;
-        if (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_ALPHA) {
-            uint32_t alphaTrackIndex = 0;
-            for (; alphaTrackIndex < data->tracks.count; ++alphaTrackIndex) {
-                avifTrack * track = &data->tracks.track[alphaTrackIndex];
-                if (!track->sampleTable) {
-                    continue;
-                }
-                if (!track->id) {
-                    continue;
-                }
-                if (!track->sampleTable->chunks.count) {
-                    continue;
-                }
-                alphaCodecType = avifSampleTableGetCodecType(track->sampleTable);
-                if (alphaCodecType == AVIF_CODEC_TYPE_UNKNOWN) {
-                    continue;
-                }
-                const avifPropertyArray * properties = avifSampleTableGetProperties(track->sampleTable, alphaCodecType);
-                const avifProperty * auxiProp = properties ? avifPropertyArrayFind(properties, "auxi") : NULL;
-                // If auxi is present, check that it contains the alpha URN.
-                // If auxi is not present, assume that the track is alpha. This is for backward compatibility with
-                // old versions of libavif that did not write this property, see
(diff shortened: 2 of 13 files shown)


src/read.c
```diff
@@ -6180,44 +6178,46 @@ avifResult avifDecoderReset(avifDecoder * decoder)
...
-            // If auxi is not present, assume that the track is alpha. This is for backward compatibility with
-            // old versions of libavif that did not write this property, see
-            // https://github.com/AOMediaCodec/libavif/commit/98faa17
-            if (auxiProp && !isAlphaURN(auxiProp->u.auxC.auxType)) {
-                continue;
-            }
-            // Do not check the track's handlerType. It should be "auxv" according to
-            // HEIF (ISO/IEC 23008-12:2022), Section 7.5.3.1, but old versions of libavif used to write
-            // "pict" instead. See https://github.com/AOMediaCodec/libavif/commit/65d0af9
+        if (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_ALPHA) {
```
wantehchang: Here I added `if (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_ALPHA) {` around this block of code. The only change is to the block of code is that I moved the declaration of `alphaCodecType` outside the if block, because it is used after the if block.

src/read.c
```diff
@@ -5289,10 +5289,9 @@ avifResult avifDecoderParse(avifDecoder * decoder)
 {
     avifDiagnosticsClearError(&decoder->diag);
 
-    // Color only or alpha only is not currently supported.
-    if ((decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) != 0 &&
-        (decoder->imageContentToDecode & AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) != AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA) {
-        avifDiagnosticsPrintf(&decoder->diag, "imageContentToDecode set to only color or only alpha is not supported");
+    // Alpha only is not currently supported.
```
wantehchang, line 5292: All the changes in this PR are straightforward except the changes to this file. Please review the changes to src/read.c carefully.

Maryla: If you are not familiar with the code in src/read.c, we can wait for Yannis's review.

maryla-uc (approved):
Currently the documentation of alphaPresent is "This is true when avifDecoderParse() detects an alpha plane."
It's unclear from the documentation what the value should be when imageContentToDecode does not include alpha. To me it seems that the intention of the API is that imageContentToDecode is about the pixels actually decoded. E.g. for gain maps, we still read the gain map metadata even when we don't decode the pixels. So it would make sense to me that alphaPresent would still be set even when not decoding alpha plane pixels.
Feel free to disagree and keep the current behavior but in any case the behavior should be clarified in avif.h

include/avif/avif.h
```diff
@@ -1229,8 +1229,10 @@ AVIF_API const char * avifProgressiveStateToString(avifProgressiveState progress
 typedef enum avifImageContentTypeFlag
 {
     AVIF_IMAGE_CONTENT_NONE = 0,
-    // Color only or alpha only is not currently supported.
-    AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA = (1 << 0) | (1 << 1),
+    AVIF_IMAGE_CONTENT_COLOR = (1 << 0),
+    // Alpha only is not currently supported.
+    AVIF_IMAGE_CONTENT_ALPHA = (1 << 1),
+    AVIF_IMAGE_CONTENT_COLOR_AND_ALPHA = AVIF_IMAGE_CONTENT_COLOR | AVIF_IMAGE_CONTENT_ALPHA,
```
wantehchang, line 1235: Maryla: Thank you for pointing out that the documentation of `decoder->alphaPresent` suggests it should be independent of whether the `AVIF_IMAGE_CONTENT_ALPHA` bit is set in `decoder->imageContentToDecode`.

When `decoder->alphaPresent` was added (for Chrome), it was intended to indicate whether the decoder image would have an alpha plane. Since `decoder->imageContentToDecode` didn't exist then, we didn't need to deal with this issue.

Before this PR, `decoder->imageContentToDecode` could already be set to `AVIF_IMAGE_CONTENT_GAIN_MAP` without the `AVIF_IMAGE_CONTENT_ALPHA` bit. I think it is better to preserve the behavior of `decoder->alphaPresent` when `decoder->imageContentToDecode` is set to `AVIF_IMAGE_CONTENT_GAIN_MAP`. I will find out what the behavior is, but I suspect it is what the comment says, "This is true when avifDecoderParse() detects an alpha plane."

wantehchang: Maryla: PR https://github.com/AOMediaCodec/libavif/pull/3301 shows you are right. I reverted `decoder->alphaPresent` to the original meaning. Thanks again for catching this bug.

Since the new changes to src/read.c are quite different, it would be good to review this PR again. I suggest first reviewing the second commit except src/read.c, and then reviewing the net changes to src/read.c.

maryla-uc (approved):

src/read.c
wantehchang: Yannis: In the interest of time, I have merged this PR. It would be great if you could still do a post-commit review of the changes to include/avif/avif.h and src/read.c in this PR. Thanks!

y-guyon: LGTM.

Nit: `decoder->imageContentToDecode` could have been set to `AVIF_IMAGE_CONTENT_NONE` instead of `0` in avifanimationtest.

wantehchang: Yannis: Thank you for the review! I made the change you suggested in https://github.com/AOMediaCodec/libavif/pull/3305.

PR #3290, opened 2026-07-01
Don't delete gain map metadata if ignore_gain_map

wantehchang:
Don't delete gain map metadata if ignore_gain_map
    
In avif::ReadImage(), if ignore_gain_map is true, don't delete gain map
metadata. Let the caller do that.
    
Also, since avif::ReadImage() calls avifImageCopy(), it doesn't need to
call avifImageCreateView().

### DIFF
--- a/apps/avifgainmaputil/combine_command.cc
+++ b/apps/avifgainmaputil/combine_command.cc
@@ -135,9 +135,11 @@ avifResult CombineCommand::Run() {
             << gain_map_height << "\n";
 
   // Because base_image is read with ignore_gain_map=true, there is no
-  // preexisting gain map. Otherwise, overwriting the pointer would cause a
-  // memory leak.
-  assert(base_image->gainMap == nullptr);
+  // preexisting gain map but there may be preexisting gain map metadata.
+  if (base_image->gainMap) {
+    assert(base_image->gainMap->image == nullptr);
+    avifGainMapDestroy(base_image->gainMap);
+  }
   base_image->gainMap = avifGainMapCreate();
   base_image->gainMap->image =
       avifImageCreate(gain_map_width, gain_map_height, arg_gain_map_depth_,
--- a/apps/avifgainmaputil/combine_command.cc
+++ b/apps/avifgainmaputil/combine_command.cc
@@ -139,6 +139,7 @@ avifResult CombineCommand::Run() {
   if (base_image->gainMap) {
     assert(base_image->gainMap->image == nullptr);
     avifGainMapDestroy(base_image->gainMap);
+    base_image->gainMap = nullptr;
   }
   base_image->gainMap = avifGainMapCreate();
   base_image->gainMap->image =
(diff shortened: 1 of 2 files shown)


apps/avifgainmaputil/combine_command.cc
```diff
@@ -135,9 +135,11 @@ avifResult CombineCommand::Run() {
...
 
   // Because base_image is read with ignore_gain_map=true, there is no
-  // preexisting gain map. Otherwise, overwriting the pointer would cause a
-  // memory leak.
-  assert(base_image->gainMap == nullptr);
+  // preexisting gain map but there may be preexisting gain map metadata.
+  if (base_image->gainMap) {
+    assert(base_image->gainMap->image == nullptr);
+    avifGainMapDestroy(base_image->gainMap);
+  }
```
wantehchang, line 143: This is the only caller of `avif::ReadImage()` that passes `ignore_gain_map=true`.

apps/avifgainmaputil/combine_command.cc
```diff
@@ -135,9 +135,11 @@ avifResult CombineCommand::Run() {
             << gain_map_height << "\n";
 
   // Because base_image is read with ignore_gain_map=true, there is no
-  // preexisting gain map. Otherwise, overwriting the pointer would cause a
-  // memory leak.
-  assert(base_image->gainMap == nullptr);
+  // preexisting gain map but there may be preexisting gain map metadata.
+  if (base_image->gainMap) {
+    assert(base_image->gainMap->image == nullptr);
+    avifGainMapDestroy(base_image->gainMap);
```
maryla-uc, line 141: base_image->gainMap = NULL;
?

apps/avifgainmaputil/imageio.cc
```diff
@@ -210,25 +210,24 @@ avifResult ReadImage(avifImage* image, const std::string& input_filename,
...
     const avifTransferCharacteristics in_transfer =
         image->transferCharacteristics;
     const avifMatrixCoefficients in_matrix = image->matrixCoefficients;
 
-    result = avifImageCopy(image, view.get(), AVIF_PLANES_ALL);
+    result = avifImageCopy(image, decoder->image, AVIF_PLANES_ALL);
     if (result != AVIF_RESULT_OK) {
       return result;
     }
+    if (ignore_profile) {
```
maryla-uc: The idea was to use avifImageCreateView()'s logic for the ignore_* parameters. This works too although I'm not sure it makes much difference to be honest.

maryla-uc (approved):

apps/avifgainmaputil/combine_command.cc
wantehchang: Done. (I omitted this because `base_image->gainMap` is immediately assigned a new value in the next line. But it's good to write the code in the way the original author would.)

apps/avifgainmaputil/imageio.cc
wantehchang: I suspected that's why the `avifImageCreateView()` call was there. I added it back but changed the `ignoreGainMap` argument to `false` with a comment.

Note: These two extra if statements would only be transitional. They would be deleted in PR https://github.com/AOMediaCodec/libavif/pull/3280 and PR https://github.com/AOMediaCodec/libavif/pull/3288.

wantehchang (commented):
Thanks for the review.

PR #3291, opened 2026-07-02
reformat: validate rowBytes before entering pixel loop (fixes #3146)

b4l4-sec:
## Summary

Fixes #3146.

`avifImageRGBToYUV()` and `avifImageYUVToRGB()` accepted an `avifRGBImage`
whose `rowBytes` was smaller than `width * avifRGBImagePixelSize(rgb)`.
The documentation at `include/avif/avif.h:929-932` and the allocation helper
`avifRGBImageAllocatePixels()` both require:

```
rowBytes >= width * avifRGBImagePixelSize(rgb)
```

but neither conversion entry-point enforced this invariant.

### Root cause

The inner pixel loop in the manual RGB→YUV path computes offsets as:

```c
// src/reformat.c:320
rgb->pixels[ offsetBytesR + (i * rgbPixelBytes) + (j * rgbRowBytes) ]
```

When `rgbRowBytes < rgb->width * rgbPixelBytes`, column `i = 1` already
reads 1 byte past the end of the allocation.

The same issue exists in the YUV→RGB direction (`avifImageYUVToRGB`).

### Proof of Concept (ASan confirmed on HEAD)

```c
avifImage *image = avifImageCreate(100, 1, 8, AVIF_PIXEL_FORMAT_YUV444);
avifRGBImage rgb;
avifRGBImageSetDefaults(&rgb, image);
rgb.rowBytes    = avifRGBImagePixelSize(&rgb); // 4, but width = 100
rgb.pixels      = malloc(rgb.rowBytes * rgb.height); // only 4 bytes
rgb.avoidLibYUV = AVIF_TRUE;
avifImageRGBToYUV(image, &rgb); // ← CRASH
```

```
AddressSanitizer: heap-buffer-overflow READ of size 1
    #0 avifImageRGBToYUV src/reformat.c:320
0x… is located 0 bytes after 4-byte region [0x…, 0x…)
```

### Fix

Two early-return guards — one in each conversion entry point — that return
`AVIF_RESULT_REFORMAT_FAILED` when `rowBytes < width * pixelSize`.

### Test

`TEST(RGBToYUVTest, UndersizedRowBytesFails)` in
`tests/gtest/avifrgbtoyuvtest.cc` covers both directions.

After the patch the PoC returns `AVIF_RESULT_REFORMAT_FAILED` with zero ASan
events.

### DIFF
--- a/tests/gtest/avifrgbtoyuvtest.cc
+++ b/tests/gtest/avifrgbtoyuvtest.cc
@@ -903,5 +903,27 @@ TEST(RGBToYUVTest, UndersizedRowBytesFails) {
     EXPECT_EQ(avifImageYUVToRGB(image.get(), &rgb), AVIF_RESULT_REFORMAT_FAILED);
 }
 
+// Same class of bug in alpha premultiply / unpremultiply (issue #3146 follow-up).
+TEST(AlphaTest, UndersizedRowBytesFailsPremultiply) {
+    for (uint32_t depth : {8u, 10u, 16u}) {
+        ImagePtr image(avifImageCreate(100, 1, depth, AVIF_PIXEL_FORMAT_YUV444));
+        ASSERT_NE(image, nullptr);
+
+        avifRGBImage rgb;
+        avifRGBImageSetDefaults(&rgb, image.get());
+        rgb.depth = depth;
+
+        const uint32_t pixelSize = avifRGBImagePixelSize(&rgb);
+        rgb.rowBytes = pixelSize;  // one pixel wide, but width = 100
+        std::vector<uint8_t> buf(static_cast<size_t>(rgb.rowBytes) * rgb.height, 0x80);
+        rgb.pixels = buf.data();
+
+        EXPECT_EQ(avifRGBImagePremultiplyAlpha(&rgb), AVIF_RESULT_REFORMAT_FAILED)
+            << "depth=" << depth;
+        EXPECT_EQ(avifRGBImageUnpremultiplyAlpha(&rgb), AVIF_RESULT_REFORMAT_FAILED)
+            << "depth=" << depth;
+    }
+}
+
 }  // namespace
 }  // namespace avif
--- a/tests/gtest/avifrgbtoyuvtest.cc
+++ b/tests/gtest/avifrgbtoyuvtest.cc
@@ -880,5 +880,28 @@ INSTANTIATE_TEST_SUITE_P(
             /*max_average_abs_diff=*/Values(10.),
             /*min_psnr=*/Values(10.)));
 
+// Regression test for GitHub issue #3146:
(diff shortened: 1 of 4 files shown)


b4l4-sec:
## Update: 4 more instances of the same class found and patched

The second commit extends the fix to all public API functions with the same missing `rowBytes` validation:

### `avifRGBImagePremultiplyAlpha` / `avifRGBImageUnpremultiplyAlpha` — `src/alpha.c`

Both functions step through pixels with `pixel += pixelSize` per column but never check `rowBytes >= width * pixelSize`. **ASan-confirmed heap-buffer-overflow READ at `alpha.c:261`:**

```
AddressSanitizer: heap-buffer-overflow READ of size 1
    #0 avifRGBImagePremultiplyAlpha src/alpha.c:261
    0x… is located 3 bytes after 4-byte region
```

Same PoC pattern — `rowBytes = pixelSize`, `width = 100`, 4-byte allocation, crash at column `i=1`.

### `avifRGBImageApplyGainMap` / `avifRGBImageComputeGainMap` — `src/gainmap.c`

Both call `avifGetRGBAPixel(baseImage/baseRgbImage/altRgbImage, i, j, ...)` which computes `y * rowBytes + x * pixelBytes` with no prior validation of `rowBytes`. Added early-return guards on the caller-supplied images.

### All four guards added, test suite extended:

- `TEST(AlphaTest, UndersizedRowBytesFailsPremultiply)` covers both alpha functions at 8-, 10-, and 16-bit depth.

tests/gtest/avifrgbtoyuvtest.cc
```diff
@@ -880,5 +880,50 @@ INSTANTIATE_TEST_SUITE_P(
...
+    ImagePtr image(avifImageCreate(100, 1, 8, AVIF_PIXEL_FORMAT_YUV444));
+    ASSERT_NE(image, nullptr);
+
+    avifRGBImage rgb;
+    avifRGBImageSetDefaults(&rgb, image.get());
+
+    // rowBytes = 4 (one pixel wide) but rgb.width = 100 — deliberate mismatch.
+    const uint32_t pixelSize = avifRGBImagePixelSize(&rgb);
+    rgb.rowBytes = pixelSize;
+    std::vector<uint8_t> buf(static_cast<size_t>(rgb.rowBytes) * rgb.height, 0x41);
```
y-guyon, line 896: This line is incorrect. It allocates only 4 * 1 values and the call below expects 4 * 100 * 1 values to be converted. Same in the test below. This PR is about wrong `rowBytes` values, which is breaking the API contract anyway, and the tests are about accessing buffers too small to also respect the API contract.

wantehchang: I am going to close this PR. This kind of broken input is not part of our threat model. Although we defend against an attacker that controls the width, height and pixel values, we assume that the pixel buffer is allocated properly with the width and height.

PR #3292, opened 2026-07-02
Remove ignoreGainMap param of avifImageCreateView

wantehchang:
Remove the ignoreGainMap parameter of the avifImageCreateView() function.

All the callers of avifImageCreateView() pass ignoreGainMap=false to the function. (The only exception is the recursive call on the gain map image, but the gain map image cannot have its own gain map image, so it doesn't really matter.)

### DIFF
--- a/apps/shared/avifutil.h
+++ b/apps/shared/avifutil.h
@@ -132,7 +132,7 @@ void avifImageFixXMP(avifImage * image);
 // If a gain map is present in 'srcImage', the gain map of 'dstImage' is also set to
 // a view of the original gain map.
 // 'dstImage' should be an empty image. It will not own the pixel data.
-avifResult avifImageCreateView(avifImage * dstImage, const avifImage * srcImage, avifBool ignoreColorProfile, avifBool ignoreAlpha, avifBool ignoreGainMap);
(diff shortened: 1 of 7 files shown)


maryla-uc (approved):

apps/shared/avifutil.h
```diff
@@ -132,7 +132,7 @@ void avifImageFixXMP(avifImage * image);
 // If a gain map is present in 'srcImage', the gain map of 'dstImage' is also set to
 // a view of the original gain map.
 // 'dstImage' should be an empty image. It will not own the pixel data.
-avifResult avifImageCreateView(avifImage * dstImage, const avifImage * srcImage, avifBool ignoreColorProfile, avifBool ignoreAlpha, avifBool ignoreGainMap);
+avifResult avifImageCreateView(avifImage * dstImage, const avifImage * srcImage, avifBool ignoreColorProfile, avifBool ignoreAlpha);
```
wantehchang, line 135: This PR is the first of a series of PRs.

The three `ignore*` parameters can be removed from `avifImageCreateView(). I am removing them in three steps.
- `ignoreGainMap` is the easiest to remove.
- `ignoreColorProfile` will be removed in #3280. That PR is also simple.
- `ignoreAlpha` will be removed in #3288. That PR is more complicated, so I save it for the last.

After all this is done, we will be able to remove some of the `avifImageCreateView()` calls.

PR #3293, opened 2026-07-06
Bump libyuv to 078d7f6ff

wantehchang:
Used on the Chrome M151 branch, plus the fix for three Coverity Control flow issues (DEADCODE): CID 561616, CID 561617, CID 561618.


wantehchang:
Vincent: I am abandoning this PR because the two CI workflows that use CMake 3.22 have compilation errors when compiling libyuv. The compliation error is only in the `cpuid` executable program in the libyuv project. So we can work around this by building only the `yuv` library target. But it seems that all targets in the libyuv project are built when CMake 3.22 is used.

PR #3296, opened 2026-07-13
LocalLibXml2.cmake: Avoid duplicates in PREFIXES

wantehchang:
If ${AVIF_LIBRARY_PREFIX} is equal to lib, PREFIXES will check for the prefix "lib" twice. If a compiled libxml2 library exists, we will call add_library() twice. The second add_library() call will fail.

As a second fix, add an early return after finding a compiled libxml2 library.

### DIFF
--- a/cmake/Modules/LocalLibXml2.cmake
+++ b/cmake/Modules/LocalLibXml2.cmake
@@ -1,7 +1,10 @@
 set(AVIF_LIBXML_GIT_TAG "v2.15.3")
 
 # First, whether the library exists.
-set(PREFIXES lib ${AVIF_LIBRARY_PREFIX})
+set(PREFIXES "lib")
+if(NOT AVIF_LIBRARY_PREFIX STREQUAL "lib")
+    list(APPEND PREFIXES ${AVIF_LIBRARY_PREFIX})
+endif()
 set(SUFFIXES "s" "sd" "")
 foreach(PREFIX IN LISTS PREFIXES)
     foreach(SUFFIX IN LISTS SUFFIXES)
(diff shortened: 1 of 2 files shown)


cmake/Modules/LocalLibXml2.cmake
```diff
@@ -19,14 +22,11 @@ foreach(PREFIX IN LISTS PREFIXES)
...
             target_include_directories(LibXml2 INTERFACE "${AVIF_SOURCE_DIR}/ext/libxml2/install.libavif/include/libxml2")
             add_library(LibXml2::LibXml2 ALIAS LibXml2)
+            return()
         endif()
     endforeach()
 endforeach()
 
-if(TARGET LibXml2::LibXml2)
-    return()
-endif()
```
wantehchang, line 28: Vincent: You added the original code in PR https://github.com/AOMediaCodec/libavif/pull/2923.

cmake/Modules/LocalLibXml2.cmake
```diff
@@ -1,7 +1,10 @@
 set(AVIF_LIBXML_GIT_TAG "v2.15.3")
 
 # First, whether the library exists.
-set(PREFIXES lib ${AVIF_LIBRARY_PREFIX})
+set(PREFIXES "lib")
+if(NOT AVIF_LIBRARY_PREFIX STREQUAL "lib")
+    list(APPEND PREFIXES ${AVIF_LIBRARY_PREFIX})
```
wantehchang, line 6: Vincent: If `${AVIF_LIBRARY_PREFIX}` may be an empty string, should we quote it?

vrabaud: No, that should be fine

cmake/Modules/LocalLibXml2.cmake
vrabaud: Ack

vrabaud (approved):

PR #3301, opened 2026-07-15
Test alphaPresent when decoder ignores alpha

wantehchang:
Modify GainMapTest.IgnoreColorAndAlpha and GainMapTest.IgnoreAll to check whether alphaPresent and other main image metadata are set when the decoder ignores color and alpha.

Since this requires examining the decoder's fields after avifDecoderParse() is called, the avifDecoderReadMemory() call in GainMapTest.IgnoreColorAndAlpha is replaced with calls to the equivalent avifDecoder functions (except avifImageCopy()).

### DIFF
--- a/tests/gtest/avifgainmaptest.cc
+++ b/tests/gtest/avifgainmaptest.cc
@@ -581,8 +581,6 @@ TEST(GainMapTest, IgnoreColorAndAlpha) {
   ASSERT_EQ(result, AVIF_RESULT_OK)
       << avifResultToString(result) << ": " << encoder->diag.error;
 
-  ImagePtr decoded(avifImageCreateEmpty());
-  ASSERT_NE(decoded, nullptr);
   DecoderPtr decoder(avifDecoderCreate());
   ASSERT_NE(decoder, nullptr);
   // Decode just the gain map.
@@ -604,9 +602,6 @@ TEST(GainMapTest, IgnoreColorAndAlpha) {
   result = avifDecoderNextImage(decoder.get());
   ASSERT_EQ(result, AVIF_RESULT_OK)
       << avifResultToString(result) << ": " << decoder->diag.error;
-  result = avifImageCopy(decoded.get(), decoder->image, AVIF_PLANES_ALL);
-  ASSERT_EQ(result, AVIF_RESULT_OK)
-      << avifResultToString(result) << ": " << decoder->diag.error;
 
   // Main image metadata is available.
   EXPECT_EQ(decoder->image->width, 12u);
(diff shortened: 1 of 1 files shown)


tests/gtest/avifgainmaptest.cc
```diff
@@ -587,18 +587,41 @@ TEST(GainMapTest, IgnoreColorAndAlpha) {
...
+  EXPECT_EQ(decoder->image->width, 12u);
+  EXPECT_EQ(decoder->image->height, 34u);
+  EXPECT_EQ(decoder->image->depth, 10);
+  EXPECT_EQ(decoder->image->yuvFormat, AVIF_PIXEL_FORMAT_YUV420);
+  EXPECT_EQ(decoder->alphaPresent, AVIF_TRUE);
+
+  result = avifDecoderNextImage(decoder.get());
+  ASSERT_EQ(result, AVIF_RESULT_OK)
+      << avifResultToString(result) << ": " << decoder->diag.error;
+  result = avifImageCopy(decoded.get(), decoder->image, AVIF_PLANES_ALL);
```
wantehchang: This `avifImageCopy()` call is here to emulate the original `avifDecoderReadMemory()` call. We could omit this `avifImageCopy()` call and use `decoder->image` instead of `decoded`. Would you like me to do that?

tests/gtest/avifgainmaptest.cc
```diff
@@ -587,18 +587,41 @@ TEST(GainMapTest, IgnoreColorAndAlpha) {
...
+  result = avifDecoderParse(decoder.get());
+  ASSERT_EQ(result, AVIF_RESULT_OK)
+      << avifResultToString(result) << ": " << decoder->diag.error;
+
+  // Main image metadata is available.
+  EXPECT_EQ(decoder->image->width, 12u);
+  EXPECT_EQ(decoder->image->height, 34u);
+  EXPECT_EQ(decoder->image->depth, 10);
+  EXPECT_EQ(decoder->image->yuvFormat, AVIF_PIXEL_FORMAT_YUV420);
+  EXPECT_EQ(decoder->alphaPresent, AVIF_TRUE);
```
wantehchang, line 600: This shows that `decoder->alphaPresent` is assigned a value exactly according to the comment "This is true when avifDecoderParse() detects an alpha plane."

tests/gtest/avifgainmaptest.cc
maryla-uc: Indeed I think there's no reason to do a copy here but it doesn't matter much to be honest, I'm ok with it either way.

maryla-uc (approved):
Thanks!

tests/gtest/avifgainmaptest.cc
wantehchang: Done. I removed the `avifImageCopy() ` call. Several tests in this file call `avifDecoderReadMemory()` or `avifDecoderReadFile()`, so we have sufficient test coverage for `avifImageCopy() ` when the decoded image has a gain map.

PR #3307, opened 2026-07-21
Add --ignore-exif and --ignore-xmp to avifdec

### LINKED ISSUE #3306: Invalid Exif payload error
Hello,

[testfile.zip](https://github.com/user-attachments/files/30221409/testfile.zip)

The APZGTUXT.Avif file I downloaded is not decodable via libavif, but libheif, Google Chrome, Firefox are all able to open it. I don't know what software created the file.

Image

```
avifdec -i --no-strict APZGTUXT.Avif
Decoding with codec 'dav1d' (12 worker threads), please wait...
ERROR: Failed to parse image: Invalid Exif payload
```

```
avifdec --version
Version: 1.4.1 (dav1d [dec]:1.5.1, aom [enc/dec]:v3.12.1, rav1e [enc]:0.8.1 (), svt [enc]:v4.0.1)
libyuv : unavailable
```

```
heif-info APZGTUXT.Avif
MIME type: image/avif
main brand: avif
compatible brands: avif, mif1, miaf

image: 1440x1080 (id=1), primary
  colorspace: YCbCr, 4:2:0
  bit depth: 8
  color profile: nclx
    colour primaries: 2 (unspecified)
    transfer characteristics: 2 (unspecified)
    matrix coefficients: 6 (BT.601)
    full range flag: 1 (full)
  alpha channel: no
  depth channel: no
metadata:
  Exif: 10 bytes
transformations:
  none
region annotations:
  none
text items:
  none
properties:
none
MIME items:
```

I believe that the file is probably not perfect, but would it be possible to configure libavif to allow decoding it? Especially when the browsers don't reject it and the problem is only metadata.

y-guyon:
Fixes https://github.com/AOMediaCodec/libavif/issues/3306

### DIFF
--- a/tests/test_cmd_metadata.sh
+++ b/tests/test_cmd_metadata.sh
@@ -28,7 +28,7 @@ ENCODED_FILE="avif_test_cmd_metadata_encoded.avif"
 ENCODED_FILE_NO_METADATA="avif_test_cmd_metadata_encoded_no_metadata.avif"
 ENCODED_FILE_MORE_METADATA="avif_test_cmd_metadata_encoded_more_metadata.avif"
 DECODED_FILE="avif_test_cmd_metadata_decoded.png"
-DECODED_FILE_CHANGED_METADATA="avif_test_cmd_metadata_decoded_changed_icc.png"
+DECODED_FILE_CHANGED_METADATA="avif_test_cmd_metadata_decoded_changed_metadata.png"
 
 # Cleanup
 cleanup() {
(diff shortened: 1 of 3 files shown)


tests/test_cmd_metadata.sh
```diff
@@ -28,14 +28,14 @@ ENCODED_FILE="avif_test_cmd_metadata_encoded.avif"
 ENCODED_FILE_NO_METADATA="avif_test_cmd_metadata_encoded_no_metadata.avif"
 ENCODED_FILE_MORE_METADATA="avif_test_cmd_metadata_encoded_more_metadata.avif"
 DECODED_FILE="avif_test_cmd_metadata_decoded.png"
-DECODED_FILE_CHANGED_ICC="avif_test_cmd_metadata_decoded_changed_icc.png"
+DECODED_FILE_CHANGED_METADATA="avif_test_cmd_metadata_decoded_changed_icc.png"
```
maryla-uc: nit: change the file name as well

maryla-uc (commented):
Thanks! It's hard to believe we didn't have this already.

tests/test_cmd_metadata.sh
y-guyon: Done

Issue #3308, opened 2026-07-21
AvifDecodeTest.NonPersistentIOBug506387278 fails with libaom


y-guyon:
To reproduce:

```sh
git clone https://github.com/AOMediaCodec/libavif.git  # at 3e6bfe8
cmake -S libavif -B libavif/build \
  -DAVIF_CODEC_AOM=LOCAL -DAVIF_CODEC_DAV1D=OFF -DAVIF_LIBYUV=LOCAL \
  -DAVIF_BUILD_TESTS=ON -DAVIF_GTEST=LOCAL
cmake --build libavif/build -j
libavif/build/tests/avifdecodetest libavif/tests/data/
```

```
[ RUN      ] AvifDecodeTest.NonPersistentIOBug506387278
/home/yguyon/tmp/libavif/tests/gtest/avifdecodetest.cc:178: Failure
Expected equality of these values:
  avifDecoderNextImage(decoder.get())
    Which is: 11
  AVIF_RESULT_OK
    Which is: 0

/home/yguyon/tmp/libavif/tests/gtest/avifdecodetest.cc:179: Failure
Expected equality of these values:
  avifDecoderNextImage(decoder.get())
    Which is: 11
  AVIF_RESULT_OK
    Which is: 0

[  FAILED  ] AvifDecodeTest.NonPersistentIOBug506387278
```

I found the test to fail at https://aomedia.googlesource.com/aom/+/refs/tags/v3.14.1/av1/decoder/obu.c#857.  
The test passes with `-DAVIF_CODEC_DAV1D=LOCAL`.

---

Looking at the AV1 spec, I do not understand the following Padding OBU syntax (https://aomediacodec.github.io/av1-spec/#general-obu-syntax):
```c
open_bitstream_unit( sz ) {
    obu_header()
    obu_size  // leb128()
    startPosition = get_position( )
    // Inlined padding_obu( ):
    for ( i = 0; i < obu_padding_length; i++ )	 
        obu_padding_byte  // f(8)
    currentPosition = get_position( )
    payloadBits = currentPosition - startPosition
    if ( obu_size > 0 )
        trailing_bits( obu_size * 8 - payloadBits )
}
```

The spec says:
> Note: `obu_padding_length` is not coded in the bitstream but can be computed based on `obu_size` minus the number of trailing bytes.

I assume "the number of trailing bytes" to mean `nbBits/8`, `nbBits` being the number of trailing bits passed to `trailing_bits()`.

[...]

PR #3309, opened 2026-07-21
Fix AvifDecodeTest.NonPersistentIOBug506387278

wantehchang:
AvifDecodeTest.NonPersistentIOBug506387278 fails when libaom or libgav1 is used as the AV1 decoder, because the test input file poc_b_506387278.avif contains errors.

The first error is easier to fix. The two padding OBUs do not contain trailing bits. This is fixed with the following Python script:

```
with open("tests/data/poc_b_506387278.avif", "r+b") as f:
  f.seek(0x1702)
  f.write(b'\x80')
  f.seek(0x559a)
  f.write(b'\x80')
```

Note: The two padding OBUs can be identified by searching for the byte 0x7a, which is the OBU header for a padding OBU. The OBU sizes for the two padding OBUs are 0xbc 0x26 and 0xf0 0x7c, respectively, when encoded in leb128().

The second error is harder to fix. The data for the first image contains partial frame data for the second frame. The extra incomplete frame data makes avifDecoderNextImage() fail when libaom or libgav1 is used as the AV1 decoder. Since this test is originally intended for CrabbyAvif and dav1d, in libavif it is used as a fuzz test, so we can relax the expected results of the avifDecoderNextImage() calls in the test.


wantehchang:
cc: @jzern @vigneshvg FYI. It's not clear which AV1 decoder's behavior is the most appropriate in this case.

y-guyon (approved):

wantehchang:
This PR fixes https://github.com/AOMediaCodec/libavif/issues/3308.

PR #3310, opened 2026-07-21
Make decoder reset preserve generated alpha grids

carrerasdarren-cell:
`avifMetaFindAlphaItem()` synthesizes an alpha grid when a color grid's
individual tiles each have alpha auxiliaries. The first reset rewrites those
alpha tile `dimgForID` associations to the generated grid. A later
`avifDecoderReset()` then mistakes the rewritten associations for malformed
input and returns `AVIF_RESULT_INVALID_IMAGE_GRID`.

Record and reuse the generated alpha grid on later resets, and avoid appending
duplicate adopted codec configuration properties to grid items. Extend the
existing `color_grid_alpha_nogrid.avif` regression test to decode successfully
across three passes with two resets.

Tests:
- Full CTest suite with system AOM and dav1d: 50/50 passed
- `avifdecodetest` with UndefinedBehaviorSanitizer: 8/8 passed
- 10,000 alternating full/color-only resets on the affected fixture

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -232,8 +232,9 @@ typedef struct avifDecoderItem
     uint32_t descForID;            // if non-zero, this item is a content description for Item #{descForID}
     uint32_t dimgForID;            // if non-zero, this item is an input of derived Item #{dimgForID}
     uint32_t dimgIdx; // If dimgForId is non-zero, this is the zero-based index of this item in the list of Item #{dimgForID}'s dimg.
-    avifBool hasDimgFrom; // whether there is a 'dimg' box with this item's id as 'fromID'
-    uint32_t premByID;    // if non-zero, this item is premultiplied by Item #{premByID}
+    uint32_t generatedAlphaForID; // if non-zero, this is a generated alpha grid for color Item #{generatedAlphaForID}
+    avifBool hasDimgFrom;         // whether there is a 'dimg' box with this item's id as 'fromID'
+    uint32_t premByID;            // if non-zero, this item is premultiplied by Item #{premByID}
     avifBool hasUnsupportedEssentialProperty; // If true, this item cites a property flagged as 'essential' that libavif doesn't support (yet). Ignore the item, if so.
     avifBool ipmaSeen;    // if true, this item already received a property association
     avifBool progressive; // if true, this item has progressive layers (a1lx), but does not select a specific layer (the layer_id value in lsel is set to 0xFFFF)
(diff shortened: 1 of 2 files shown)


y-guyon:

This behavior may no longer be compliant (assuming it was so far): https://github.com/AOMediaCodec/libavif/issues/1203#issuecomment-5046044638

carrerasdarren-cell:
Thanks for flagging this. This PR only makes repeated decoder resets consistent with libavif's existing synthesized-alpha-grid behavior; the generated item is internal and the patch does not introduce a new association in the input file.

The cited draft does make the premise of that compatibility path questionable. If libavif removes support for carrying tile alpha into a derived color grid, this patch and its reset regression should not land as-is. I suggest holding #3310 until the direction in #1203 is decided. If the behavior remains as an optional compatibility path, the reset failure still needs this kind of fix and I can adjust the patch to the preferred design.

src/read.c
```diff
@@ -1654,9 +1655,11 @@ static avifResult avifDecoderAdoptGridTileCodecType(avifDecoder * decoder,
...
                 return AVIF_RESULT_INVALID_IMAGE_GRID;
             }
-            avifProperty * dstProp = (avifProperty *)avifArrayPush(&gridItem->properties);
-            AVIF_CHECKERR(dstProp != NULL, AVIF_RESULT_OUT_OF_MEMORY);
-            *dstProp = *srcProp;
+            if (!avifPropertyArrayFind(&gridItem->properties, configPropName)) {
+                avifProperty * dstProp = (avifProperty *)avifArrayPush(&gridItem->properties);
+                AVIF_CHECKERR(dstProp != NULL, AVIF_RESULT_OUT_OF_MEMORY);
+                *dstProp = *srcProp;
+            }
```
y-guyon, line 1662: This change is unrelated to alpha.

If I understand correctly, at head every consecutive call to `avifDecoderReset()` will allocate a new property if the input file contains a grid. This is why a check is added here to avoid adding the property if it already exists. Is this correct?

This is a bit late in the libavif development process, but maybe `avifDecoderReset()` should not create "fake" properties. That should be done in `avifDecoderParse()` only in my opinion.

I sent https://github.com/AOMediaCodec/libavif/pull/3311 to avoid having to create that fake property at all. That does not fix the other issue about alpha.

carrerasdarren-cell: Yes, that is correct: the added check prevents each avifDecoderReset() call from appending another copied codec property. I agree that #3311 removes the need for that part of this patch entirely. I will keep #3310 on hold pending the direction in #1203; if the compatibility behavior remains, I will rebase on #3311 and retain only the generated-alpha identity/reset handling and its regression test.

PR #3315, opened 2026-08-03
Bump libyuv to 7138e0a2a (1953)

wantehchang:
Used on the Chrome M152 branch, plus three more commits.


wantehchang:
I am abandoning this PR because the three extra commits (beyond what Chrome M152 uses) are huge.

PR #3316, opened 2026-08-06
Honour the base image stride in the tone map early exit

evilgensec:
The "nothing to tone map" early exit in `avifRGBImageApplyGainMap()` copies the base image with a single `memcpy` sized from `baseImage->rowBytes`, which assumes the source and destination strides are equal. That assumption was only enforced by an `assert()`, so it is gone in a release build.

## Detail

`src/gainmap.c`, lines 120 to 129:

```c
if (weight == 0.0f && outputTransferCharacteristics == baseTransferCharacteristics &&
    outputColorPrimaries == baseColorPrimaries && baseImage->format == toneMappedImage->format &&
    baseImage->depth == toneMappedImage->depth && baseImage->isFloat == toneMappedImage->isFloat) {
    assert(baseImage->rowBytes == toneMappedImage->rowBytes);
    assert(baseImage->height == toneMappedImage->height);
    // Copy the base image.
    memcpy(toneMappedImage->pixels, baseImage->pixels, (size_t)baseImage->rowBytes * baseImage->height);
    goto cleanup;
}
```

The destination is usually allocated a few lines earlier by `avifRGBImageAllocatePixels()`, which sets a tight stride at `src/avif.c:733`:

```c
const uint32_t rowBytes = rgb->width * pixelSize;
```

Meanwhile `include/avif/avif.h:929` documents the input contract as:

> which should be at least (`->rowBytes` * `->height`) bytes, where `->rowBytes` is at least (`->width` * `avifRGBImagePixelSize()`)

So a base image with row padding is a documented legal input, but the fast path sizes the copy from the padded source stride and writes it into a tightly strided destination. The copy then runs `(baseImage->rowBytes - width * pixelSize) * height` bytes past the end of the destination allocation.

Every other RGB path in the function, including the general path just below through `avifGetRGBAPixel` and `avifSetRGBAPixel`, already handles strides per row. This branch is the only one that does not.

The early exit is easy to reach. An SDR target display gives `hdrHeadroom` of 0, which equals a typical `baseHdrHeadroom` of 0, so `avifGetGainMapWeight()` returns exactly `0.0f`. Matching base and output color spaces, for example sRGB to sRGB, completes the condition. Padded strides are normal for buffers coming from Android Bitmaps, GPU readbacks, cropped views and any 4, 16 or 64 byte aligned image pipeline.

## The change

Promote the `assert` into the branch condition. When the strides differ the early exit is skipped and control falls through to the existing per pixel loop, which honours both strides. There is no behaviour change when the strides are equal, so the fast path still applies in the common case.

The `rowBytes` assert is removed because it is now redundant with the condition. The `height` assert is left alone.

## Test

`ToneMapTest.ToneMapRGBPaddedBaseStride` builds a base `avifRGBImage` with 16 bytes of row padding, sets `baseHdrHeadroom` to 0 so the weight is 0 and the early exit is taken, calls `avifRGBImageApplyGainMap`, and then compares the output to the base image row by row over each buffer own stride. It constructs its images in memory and needs no test data files.

Without the source change, in a Release build with ASAN:

```
==31748==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x611000000640
WRITE of size 384 at 0x611000000640 thread T0
    #1 avifRGBImageApplyGainMap gainmap.c:127
0x611000000640 is located 0 bytes after 256-byte region
allocated by thread T0 here:
    #1 avifRGBImageAllocatePixels avif.c:733
    #2 avifRGBImageApplyGainMap gainmap.c:114
SUMMARY: AddressSanitizer: heap-buffer-overflow gainmap.c:127 in avifRGBImageApplyGainMap
```

With the change applied:

```
[       OK ] ToneMapTest.ToneMapRGBPaddedBaseStride (0 ms)
[  PASSED  ] 1 test.
```

## Suite result

Built with `-DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="-DNDEBUG -fsanitize=address -g -O1"`, dav1d as the only codec:

```
clean tree      : 37 passed, 15 failed
with this change: 38 passed, 15 failed
```

The 15 failures are identical in both runs and are environmental. They report "No codec available" because I built with a decoder only and those cases encode. The change adds one passing test and alters nothing else.

## Note

`avifRGBImageApplyGainMap` and `avifImageApplyGainMap` have no fuzz target under `tests/gtest/`, and the in tree callers in `apps/avifgainmaputil` and the existing tests all pass tight strides, which is why this has not surfaced before.

### DIFF
--- a/src/gainmap.c
+++ b/src/gainmap.c
@@ -118,10 +118,9 @@ avifResult avifRGBImageApplyGainMap(const avifRGBImage * baseImage,
     const float weight = avifGetGainMapWeight(hdrHeadroom, gainMap);
 
     // Early exit if the gain map does not need to be applied and the pixel format is the same.
-    if (weight == 0.0f && outputTransferCharacteristics == baseTransferCharacteristics &&
-        outputColorPrimaries == baseColorPrimaries && baseImage->format == toneMappedImage->format &&
-        baseImage->depth == toneMappedImage->depth && baseImage->isFloat == toneMappedImage->isFloat) {
-        assert(baseImage->rowBytes == toneMappedImage->rowBytes);
+    if (weight == 0.0f && outputTransferCharacteristics == baseTransferCharacteristics && outputColorPrimaries == baseColorPrimaries &&
+        baseImage->format == toneMappedImage->format && baseImage->depth == toneMappedImage->depth &&
+        baseImage->isFloat == toneMappedImage->isFloat && baseImage->rowBytes == toneMappedImage->rowBytes) {
         assert(baseImage->height == toneMappedImage->height);
         // Copy the base image.
         memcpy(toneMappedImage->pixels, baseImage->pixels, (size_t)baseImage->rowBytes * baseImage->height);
--- a/tests/gtest/avifgainmaptest.cc
+++ b/tests/gtest/avifgainmaptest.cc
@@ -1251,6 +1251,75 @@ TEST(ToneMapTest, ToneMapImageSameHeadroom) {
   }
 }
 
+// avifRGBImage::rowBytes only has to be at least width * pixelSize, so a base
+// image with row padding is a legal input. Check that the "nothing to tone map"
+// early exit honours the source stride instead of assuming it is tight.
+TEST(ToneMapTest, ToneMapRGBPaddedBaseStride) {
+  constexpr uint32_t kWidth = 8;
+  constexpr uint32_t kHeight = 8;
+
+  GainMapPtr gain_map(avifGainMapCreate());
+  ASSERT_NE(gain_map, nullptr);
+  for (int c = 0; c < 3; ++c) {
+    gain_map->gainMapMin[c] = {0, 1};
+    gain_map->gainMapMax[c] = {1, 1};
+    gain_map->gainMapGamma[c] = {1, 1};
+    gain_map->baseOffset[c] = {0, 64};
+    gain_map->alternateOffset[c] = {0, 64};
+  }
+  // baseHdrHeadroom 0 and a target headroom of 0 give a weight of 0, which is
+  // what makes the memcpy fast path eligible. The padded base stride below is
+  // what must keep it from being taken.
+  gain_map->baseHdrHeadroom = {0, 1};
+  gain_map->alternateHdrHeadroom = {1, 1};
+  gain_map->useBaseColorSpace = AVIF_TRUE;
+  gain_map->image =
+      avifImageCreate(kWidth, kHeight, 8, AVIF_PIXEL_FORMAT_YUV400);
+  ASSERT_NE(gain_map->image, nullptr);
+  ASSERT_EQ(avifImageAllocatePlanes(gain_map->image, AVIF_PLANES_YUV),
+            AVIF_RESULT_OK);
+
+  avifRGBImage base = {};
+  base.width = kWidth;
+  base.height = kHeight;
+  base.depth = 8;
+  base.format = AVIF_RGB_FORMAT_RGBA;
+  const uint32_t pixel_size = avifRGBImagePixelSize(&base);
+  // Padded stride, as produced by an Android Bitmap or a cropped GPU readback.
+  base.rowBytes = kWidth * pixel_size + 16;
+  std::vector<uint8_t> base_pixels((size_t)base.rowBytes * base.height, 0x41);
+  base.pixels = base_pixels.data();
+
+  // avifRGBImageApplyGainMap() allocates the output itself, so only the fields
+  // it does not set are needed here.
+  avifRGBImage tone_mapped = {};
+  tone_mapped.depth = 8;
+  tone_mapped.format = AVIF_RGB_FORMAT_RGBA;
+
+  avifDiagnostics diag;
+  avifDiagnosticsClearError(&diag);
+  ASSERT_EQ(avifRGBImageApplyGainMap(
+                &base, AVIF_COLOR_PRIMARIES_BT709,
+                AVIF_TRANSFER_CHARACTERISTICS_SRGB, gain_map.get(),
+                /*hdrHeadroom=*/0.0f, AVIF_COLOR_PRIMARIES_BT709,
+                AVIF_TRANSFER_CHARACTERISTICS_SRGB, &tone_mapped,
+                /*clli=*/nullptr, &diag),
+            AVIF_RESULT_OK)
+      << diag.error;
+
+  // The gain map is a no-op here, so the output should equal the base image
+  // row by row, each row compared over its own stride.
+  for (uint32_t y = 0; y < kHeight; ++y) {
+    EXPECT_EQ(memcmp(tone_mapped.pixels + (size_t)y * tone_mapped.rowBytes,
+                     base.pixels + (size_t)y * base.rowBytes,
+                     (size_t)kWidth * pixel_size),
+              0)
+        << "row " << y << " differs";
+  }
+
+  avifRGBImageFreePixels(&tone_mapped);
+}
+
 TEST(GainMapTest, OpaqueProperties) {
   ImagePtr image = CreateTestImageWithGainMap(/*base_rendition_is_hdr=*/false);
   ASSERT_NE(image, nullptr);


evilgensec:
Both review comments were correct. Updated the test:

**Redundant pre-allocation removed.** `avifRGBImageApplyGainMap()` calls `avifRGBImageAllocatePixels()` at `gainmap.c:114`, and that function begins with `avifRGBImageFreePixels()`, so allocating the output in the test was wasted work and implied the caller owned a buffer it does not. The test now sets only `depth` and `format`, the two fields the function does not set itself, and still frees at the end.

**Comment reworded.** The old wording said the early exit is taken, which is backwards for this test. A weight of 0 is what makes the memcpy path eligible, and the padded base stride is what must stop it being taken. The comment now says that.

Also wrapped one line that ran past the column limit so clang-format is clean.

Re-verified after the change. With the patch the test passes and the full `avifgainmaptest` suite goes from 37 to 38 passing. With `src/gainmap.c` reverted to its prior state the test still catches the overflow:

```
==47491==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x611000000500
WRITE of size 384 at 0x611000000500 thread T0
    #1 avifRGBImageApplyGainMap gainmap.c:127
```

The 15 failures in that suite are pre-existing in my environment and identical before and after, all reporting "No codec available" because I built with a decoder only.

tests/gtest/avifgainmaptest.cc
```diff
@@ -1251,6 +1251,75 @@ TEST(ToneMapTest, ToneMapImageSameHeadroom) {
...
+  // row by row, each row compared over its own stride.
+  for (uint32_t y = 0; y < kHeight; ++y) {
+    EXPECT_EQ(memcmp(tone_mapped.pixels + (size_t)y * tone_mapped.rowBytes,
+                     base.pixels + (size_t)y * base.rowBytes,
+                     (size_t)kWidth * pixel_size),
+              0)
+        << "row " << y << " differs";
+  }
+
+  avifRGBImageFreePixels(&tone_mapped);
```
maryla-uc, line 1320: You could also use RGBImageCleanup from include/avif/avif_cxx.h after the avifRGBImageApplyGainMap call

maryla-uc (approved):
Thanks for the fix!

evilgensec:
Thanks for the review.

Fixed the `clang-format-check` failure. The condition needed re-wrapping against `src/`'s 130 column limit once the fourth clause was added. I had also wrapped a line in the test at 80 columns by mistake, not realising `tests/gtest/` has its own Google-style config, so both files are now formatted against their own `.clang-format` rather than one assumption applied to both.

On `build-windows-installed (clang-cl)`, I do not think that one is mine, and I would rather say so than quietly force a re-run. `avifgainmaptest.cc.obj` compiled and `avifgainmaptest.exe` linked, both fine, and the step that actually stopped ninja was `FAILED: tests/avifpropertytest.exe`, a target this PR does not touch. The same job also failed to fetch `libiconv-1.19.tar.gz` earlier with a 502 from the GNU mirror during vcpkg setup, and `build-windows-installed (msvc)` plus all three `build-shared-installed (windows-latest)` jobs passed on identical code. The push above should give a clean run. If clang-cl fails again on the same target I will dig further rather than assume.

Local verification is unchanged after the reformat. The test passes with the patch, and with `src/gainmap.c` reverted it still catches the overflow:

```
==47491==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x611000000500
WRITE of size 384 at 0x611000000500 thread T0
    #1 avifRGBImageApplyGainMap gainmap.c:127
```

evilgensec:
Small correction to my last message: the workflows on the new commit are sitting in `action_required`, so they need a maintainer to approve the run before anything executes. Nothing is failing, it just cannot start on its own from a fork. Whenever it is convenient.

src/gainmap.c
```diff
@@ -118,10 +118,9 @@ avifResult avifRGBImageApplyGainMap(const avifRGBImage * baseImage,
     const float weight = avifGetGainMapWeight(hdrHeadroom, gainMap);
 
     // Early exit if the gain map does not need to be applied and the pixel format is the same.
-    if (weight == 0.0f && outputTransferCharacteristics == baseTransferCharacteristics &&
-        outputColorPrimaries == baseColorPrimaries && baseImage->format == toneMappedImage->format &&
-        baseImage->depth == toneMappedImage->depth && baseImage->isFloat == toneMappedImage->isFloat) {
-        assert(baseImage->rowBytes == toneMappedImage->rowBytes);
+    if (weight == 0.0f && outputTransferCharacteristics == baseTransferCharacteristics && outputColorPrimaries == baseColorPrimaries &&
+        baseImage->format == toneMappedImage->format && baseImage->depth == toneMappedImage->depth &&
+        baseImage->isFloat == toneMappedImage->isFloat && baseImage->rowBytes == toneMappedImage->rowBytes) {
```
wantehchang, line 123: Thanks for the patch. This fix is correct. The following is an alternative fix that copies row by row and doesn't require `rowBytes` to be the same:

```
diff --git a/src/gainmap.c b/src/gainmap.c
index d87f96a8..e076953d 100644
--- a/src/gainmap.c
+++ b/src/gainmap.c
@@ -121,10 +121,15 @@ avifResult avifRGBImageApplyGainMap(const avifRGBImage * baseImage,
     if (weight == 0.0f && outputTransferCharacteristics == baseTransferCharacteristics &&
         outputColorPrimaries == baseColorPrimaries && baseImage->format == toneMappedImage->format &&
         baseImage->depth == toneMappedImage->depth && baseImage->isFloat == toneMappedImage->isFloat) {
-        assert(baseImage->rowBytes == toneMappedImage->rowBytes);
+        assert(baseImage->width == toneMappedImage->width);
         assert(baseImage->height == toneMappedImage->height);
         // Copy the base image.
-        memcpy(toneMappedImage->pixels, baseImage->pixels, (size_t)baseImage->rowBytes * baseImage->height);
+        const uint32_t pixelSize = avifRGBImagePixelSize(baseImage);
+        for (uint32_t j = 0; j < baseImage->height; ++j) {
+            memcpy(toneMappedImage->pixels + (size_t)j * toneMappedImage->rowBytes,
+                   baseImage->pixels + (size_t)j * baseImage->rowBytes,
+                   baseImage->width * pixelSize);
+        }
         goto cleanup;
     }
 
```

tests/gtest/avifgainmaptest.cc
```diff
@@ -1251,6 +1251,75 @@ TEST(ToneMapTest, ToneMapImageSameHeadroom) {
...
+
+  avifRGBImage base = {};
+  base.width = kWidth;
+  base.height = kHeight;
+  base.depth = 8;
+  base.format = AVIF_RGB_FORMAT_RGBA;
+  const uint32_t pixel_size = avifRGBImagePixelSize(&base);
+  // Padded stride, as produced by an Android Bitmap or a cropped GPU readback.
+  base.rowBytes = kWidth * pixel_size + 16;
+  std::vector<uint8_t> base_pixels((size_t)base.rowBytes * base.height, 0x41);
```
wantehchang, line 1290: When I run this test in Release build without the bug fix, the `memcmp()` check at the end of this test still passes. (The buffer overwrite does not cause a crash.) The reason is that the whole `base` image is set to the same value (0x41). I will write a PR to set the `base` image to different sample values to make the `memcmp()` check fail.

src/gainmap.c
evilgensec: Thanks — your alternative is better than what landed, and I would like to send it as a follow-up if you have not already started one.

Mine makes stride equality a precondition, so a padded base image now falls through to the full conversion path. Yours keeps the fast path and copies `width * pixelSize` per row, which produces the same output for strictly less work.

Worth noting why the asserts are sound in your version where the original one was not. `avifRGBImageApplyGainMap()` assigns `toneMappedImage->width` and `->height` from the base image at `gainmap.c:111-112`, immediately before the early exit, so `assert(baseImage->width == toneMappedImage->width)` restates an invariant the function has just established. `rowBytes` was different in kind: the base stride comes from the caller and padding is explicitly legal, while the output stride comes from `avifRGBImageAllocatePixels()`. The old assert was checking something the function did not control, which is why it held in debug builds and vanished in release.

If you would like me to open it, I will also extend `ToneMapRGBPaddedBaseStride` so the base and output images have *different* padded strides, exercising the row-by-row path rather than only the equal-stride case. Happy to leave it with you if you would rather take it yourself.

@maryla-uc — good call on `RGBImageCleanup`, I will fold that into the same PR.

PR #3317, opened 2026-08-06
Improve ToneMapTest.ToneMapRGBPaddedBaseStride

wantehchang:
This is a follow-up to PR #3316.

In the ToneMapTest.ToneMapRGBPaddedBaseStride test, set the `base` image to different sample values to enable the memcmp() check at the end of the test to detect the bug.

Add a comment to describe how avifImageApplyGainMap() sets the fields of the 'toneMappedImage' in/out parameter.

### DIFF
--- a/include/avif/avif.h
+++ b/include/avif/avif.h
@@ -1721,6 +1721,8 @@ AVIF_NODISCARD AVIF_API avifBool avifPeekCompatibleFileType(const avifROData * i
 // Performs tone mapping on a base image using the provided gain map.
 // The HDR headroom is log2 of the ratio of HDR to SDR white brightness of the display to tone map for.
 // 'toneMappedImage' should have the 'format', 'depth', and 'isFloat' fields set to the desired values.
+// This function sets the 'width' and 'height' fields of 'toneMappedImage' to the width and height of
+// `baseImage`, and allocates the 'pixels' buffer and sets the 'rowBytes' field of 'toneMappedImage'.
 // If non NULL, 'clli' will be filled with the light level information of the tone mapped image.
 AVIF_API avifResult avifImageApplyGainMap(const avifImage * baseImage,
                                           const avifGainMap * gainMap,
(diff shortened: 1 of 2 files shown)


include/avif/avif.h
```diff
@@ -1721,6 +1721,8 @@ AVIF_NODISCARD AVIF_API avifBool avifPeekCompatibleFileType(const avifROData * i
 // Performs tone mapping on a base image using the provided gain map.
 // The HDR headroom is log2 of the ratio of HDR to SDR white brightness of the display to tone map for.
 // 'toneMappedImage' should have the 'format', 'depth', and 'isFloat' fields set to the desired values.
+// This function sets the 'width' and 'height' fields of 'toneMappedImage' to the width and height of
+// `baseImage`, and allocates the 'pixels' buffer and sets the 'rowBytes' field of 'toneMappedImage'.
```
wantehchang, line 1725: This function uses `toneMappedImage` as an in/out parameter. The new sentence documents the fields of 'toneMappedImage' set by this function, which implies the values of those fields are ignored by this function on input. (Any existing `toneMappedImage->pixels` buffer is freed first.)

tests/gtest/avifgainmaptest.cc
```diff
@@ -1287,7 +1287,10 @@ TEST(ToneMapTest, ToneMapRGBPaddedBaseStride) {
   const uint32_t pixel_size = avifRGBImagePixelSize(&base);
   // Padded stride, as produced by an Android Bitmap or a cropped GPU readback.
   base.rowBytes = kWidth * pixel_size + 16;
-  std::vector<uint8_t> base_pixels((size_t)base.rowBytes * base.height, 0x41);
+  std::vector<uint8_t> base_pixels((size_t)base.rowBytes * base.height);
+  for (size_t i = 0; i < base_pixels.size(); ++i) {
+    base_pixels[i] = i & 0xff;
+  }
```
wantehchang, line 1293: cc: @evilgensec FYI.

evilgensec:
Thanks for catching this. You are right that the constant `0x41` fill defeated the `memcmp`: with every byte identical the row comparison passes whether or not the copy honours the destination stride, so the only thing the test actually caught was the overflow itself under a sanitizer build. Filling with `i & 0xff` makes each row distinguishable, so the comparison now fails on the unfixed code for the right reason.

The `avifImageApplyGainMap()` doc note is a good addition too. LGTM.

y-guyon (approved):

Issue #3319, opened 2026-08-07
1.4.2 fails on static_assert with older versions of clang


Dave-Allured:
libavif 1.4.2 fails build on older versions of clang.  The context is older versions of macOS on MacPorts.  Build on macOS 10.10 (Mojave) with Apple Clang 7.0.2 and CMake 3.31.12 fails:
```
[...]

PR #3320, opened 2026-08-10
CMakeLists.txt: Guard for static_assert

Dave-Allured:
* Add CMake guard for C11 static_assert, which was added in libavif 1.4.2.
* This will not fix missing support, but will catch insufficient compiler at configure time.
* https://github.com/AOMediaCodec/libavif/issues/3319

I am not experienced with CMake.  Please adjust or reposition this patch, as appropriate.  Thanks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,11 @@ endif()
 add_library(avif_obj OBJECT)
 add_library(avif)
 
+# Require static_assert support in function context, as added in 1.4.2.
+# Was missing in some older C11 compilers: Apple Clang <= 7.0.2, etc.
+# https://github.com/AOMediaCodec/libavif/issues/3319
+target_compile_features(avif PRIVATE c_static_assert)
+
 # Adds <target> to avif_obj's public link libraries for build, and adds
 # the <target> library as an install link library for export in case a consumer
 # needs to include that library alongside libavif when statically linking.


CMakeLists.txt
```diff
@@ -155,6 +155,11 @@ endif()
 add_library(avif_obj OBJECT)
 add_library(avif)
 
+# Require static_assert support in function context, as added in 1.4.2.
+# Was missing in some older C11 compilers: Apple Clang <= 7.0.2, etc.
+# https://github.com/AOMediaCodec/libavif/issues/3319
+target_compile_features(avif PRIVATE c_static_assert)
```
wantehchang, line 161: cc: @vrabaud 

Dave: Thank you for the pull request. We can try the following simple change:

```
index c9a7f82d..f29d4b39 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,10 +51,10 @@ option(AVIF_ENABLE_NODISCARD "Add [[nodiscard]] to some functions." OFF)
 if(AVIF_ENABLE_NODISCARD)
     # [[nodiscard]] requires C23.
     set(CMAKE_C_STANDARD 23)
-    set(CMAKE_C_STANDARD_REQUIRED ON)
 else()
     set(CMAKE_C_STANDARD 11)
 endif()
+set(CMAKE_C_STANDARD_REQUIRED ON)
 
 # SOVERSION scheme: MAJOR.MINOR.PATCH
 #   If there was an incompatible interface change:
```

If we want to avoid setting `CMAKE_C_STANDARD` globally, I would prefer to use `c_std_11` instead of `c_static_assert` because we may use other C11 features such as `alignas` in the future. Also, we would need to call `target_compile_features(... PRIVATE c_std_11)` on all the C targets, not just `avif`.

Dave-Allured: @wantehchang I understand your concerns.  I think you should avoid setting `CMAKE_C_STANDARD` globally, only for this purpose.  I am concerned about possible wider disruption, which seems unnecessary right now.  My intent was a very minor improvement in diagnostic messaging for legacy compilers.  I was looking for a way to set `c_static_assert` globally, but I did not find this in CMake documentation.

Let's cancel this PR and not disturb the code base.  I have a viable solution for the problem that I raised in #3319.

wantehchang: Dave: I will look into avoiding setting `CMAKE_C_STANDARD` globally. I believe `target_compile_features()` is the recommended method:

https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#requiring-language-standards

I need to figure out a way to use either `c_std_11` or `c_std_23` depending on whether `AVIF_ENABLE_NODISCARD` is enabled.

Dave-Allured: I have reconsidered this and checked (briefly) the CMake documentation.  `c_static_assert` is a fine-grained approach which targets only the code base in its current state at 1.4.2.  Meta-options such as `c_std_11` carry more requirements, with a slight risk of disturbing current builds in some contexts.  I see this as unnecessary disturbance for new language features which are not yet used or needed.

https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_C_KNOWN_FEATURES.html

Therefore I recommend only two conservative alternatives at this time.

* Either apply `c_static_assert` as I originally suggested,
* Or make no change at this time, wait for new code requirements before adding new compiler requirements.

This addition of CMake `c_static_assert` is a simple one-line update.  You can easily discard or change this at any time in the future, as code requirements change.

PR #3321, opened 2026-08-10
Parse 'iref' and 'grpl' children within their own box size

Nexory:
### Summary

`avifParseItemReferenceBox()` (`src/read.c:3347`) and `avifParseGroupsListBox()`
(`src/read.c:3423`) read a child box header and then read the child's fields directly from the
parent stream, driven by `reference_count` and `num_entities_in_group`. Neither uses the
`.size` it just read. Every other child box loop in `read.c` parses the child inside a stream
bounded by the child's size and then skips the parent stream by that size: fourteen functions
read child box headers, twelve of them bound the child at thirteen call sites, and these two do
not.

The consequence is an interoperability defect on `grpl` and a parsing inconsistency on `iref`.
This patch makes both loops behave like their thirteen siblings. No memory safety issue is
involved, see Notes.

### Why an EntityToGroupBox may have data after the entity_id array

The strongest case does not need HEIF at all. `prsl` (Preselection grouping) is registered in
MP4RA under `ISO`, that is in ISO/IEC 14496-12 itself, not in HEIF
(`mp4ra.github.io/main/data/entity-groups.csv`). It carries further fields after the entity_id
array, and child boxes after those. From `Dash-Industry-Forum/codem-isoboxer`,
`src/processors/prsl.js`:

```js
// ISO/IEC 14496-12:202x - 8.18.4.1 Preselection group box
this._procField('group_id', 'uint', 32);
this._procField('num_entities_in_group', 'uint', 32);
this._procEntries('entities', this.num_entities_in_group, ...);
if (this.flags & 0x1000) this._procField('preselection_tag', 'utf8string');
if (this.flags & 0x2000) this._procField('selection_priority', 'uint', 8);
if (this.flags & 0x4000) this._procField('interleaving_tag', 'utf8string');
/* rest of the boxes will be read by _parseContainerBox */
```

MPEG's own reference software says the same thing in a comment and implements it.
`MPEGGroup/isobmff`, `IsoLib/libisomediafile/src/MP4Atoms.h:2241`:

```c
/** ISOBMFF allows extensions to EntityToGroupBox. This buffer holds the data for it */
char *remainingData;
u32 remainingDataSize;
```

and `IsoLib/libisomediafile/src/EntityToGroupBox.c`, `createFromInputStream()`, reads the
entity IDs and then everything up to `self->size` into that buffer.

Two more independent implementations already do what this patch does:

- Nokia HEIF (`srcs/common/groupslistbox.cpp:54-63`) extracts a size bounded sub bitstream per
  child (`srcs/common/bitstream.cpp:512-514`), and the child
  (`srcs/common/entitytogroupbox.cpp:65-75`) reads only the generic fields. Leftover bytes are
  simply not read.
- GPAC (`src/isomedia/iff.c:776-782`) keeps the remainder in `data`/`data_len` and writes it
  back out (`iff.c:811,819`), and reads every `grpl` child generically
  (`src/isomedia/box_funcs.c:322-325`).

A concrete file exists today. `heif-enc --add-pyramid-group -A ...` writes a `pymd` group
(`pymd,Image pyramid,HEIF` in the same MP4RA list) with `tile_size_x/y` and per layer data after
the entity IDs:

```
00 00 00 3e 67 72 70 6c   grpl, size 62
00 00 00 36 70 79 6d 64   pymd, size 54
00 00 00 00               version 0, flags 0
00 00 00 01               group_id
00 00 00 03               num_entities_in_group
00 00 00 01 00 00 00 02 00 00 00 03   entity_id[0..2]
02 00 02 00               tile_size_x = 512, tile_size_y = 512   <- parsing resumes here
00 04 00 00 00 00         layer 0
00 02 00 00 00 00         layer 1
00 01 00 00 00 00         layer 2
```

libavif stops 22 bytes short of the child box end, reads `02 00 02 00` as the size of the next
EntityToGroupBox, and rejects the file:

```
$ avifdec pyramid.avif out.png
ERROR: Failed to parse image: BMFF parsing failed
Diagnostics:
 * Box[grpl]: Child box too large, possibly truncated data
```

libavif does not need to understand any of these grouping types. It keeps unknown types in
`meta->entityToGroups` and only acts on `altr`. It only needs to stop at the child box end,
which `avifROStreamReadBoxHeader()` read and validated one line earlier.

### The 'iref' half

Same shape, opposite direction: `irefHeader.size` is read and never used, so a
`reference_count` that asks for more `to_item_ID` values than the declared child box holds is
satisfied from the bytes following the box, and libavif accepts the file.

Two other parsers reject exactly that input:

- ffmpeg, `libavformat/mov.c:9733`: rejects when `entries * item_id_size > size - item_id_size - 2`,
  with the message `iref %s entry count %d exceeds the sub-box size`. Its parent loop computes
  `next = start + size` and seeks there at the child end (`mov.c:9775-9803`).
- GPAC, `src/isomedia/box_code_meta.c:875`: rejects when `ptr->size / 2 < ptr->reference_count`.

Inside libavif the nearest relative is `avifTrackReferenceBox()`, the track side of the same
reference family, which already does it:

```c
AVIF_CHECK(avifROStreamReadU32(&s, &toID));                       // unsigned int(32) track_IDs[];
AVIF_CHECK(avifROStreamSkip(&s, header.size - sizeof(uint32_t))); // just take the first one
```

### How common are such files today

Not common, and I would rather say so than let it be discovered in review. I scanned four
corpora with a structural scanner that compares each `iref` and `grpl` child box payload against
its field driven length:

| Corpus | Files | With iref/grpl | Child boxes | Size mismatches |
|---|---|---|---|---|
| Nokia HEIF conformance suite | 63 | 44 | 67 | 0 |
| AVIF conformance (`AOMediaCodec/av1-avif`, testFiles) | 172 | 19 | 24 | 0 |
| libavif `tests/data` | 54 | 45 | 125 | 0 |
| libheif `tests/data` | 93 | 3 | 3 | 0 |
| Total | 382 | 111 | 219 | 0 |

219 real child boxes, none of them with trailing data. The scanner was checked against the
pyramid file first, which it reports as a mismatch, so a zero here is a result and not a broken
tool. The libheif writing side is behind `ENABLE_EXPERIMENTAL_FEATURES` and is not enabled in
any distribution build I checked (Debian, Arch, Alpine, Fedora, vcpkg; Gentoo and Homebrew I
could not determine). Searching this repository's issues and pull requests for `pymd` or
`Child box too large` returns nothing.

So: files like this are practically absent today. The reason to do this is that `prsl` is in the
base specification, that the failure mode is losing the whole file rather than ignoring one
group, and that two of thirteen sibling loops in this file behave differently from the other
eleven.

### What the patch does

Both loops now read their fields from a stream bounded by the child box size, and skip the
parent stream by that size, mirroring `iprp`/`ipco`, `iprp`/`ipma`, `iinf`/`infe` and the generic
`meta` loop. Diff is +21/-9 in `src/read.c`: two `BEGIN_STREAM` lines, two `avifROStreamSkip`
calls, seven comment lines, and ten lines that only redirect existing reads from `&s` to the
bounded stream. No other logic changes.

There is prior art for the reasoning in this file. `2be0ec98` ("read.c: Use header size when
parsing VisualSampleEntry") says:

> The current code uses avifROStreamRemainingBytes which is not correct. We are inside a for loop
> where each loop is a box with a fixed header size. So within each loop, we should not parse
> more than that loop's header size.

To be precise about what that is: it is the rationale of one commit from 2024 that changed six
lines in `avifParseSampleDescriptionBox`, not a written project rule, and the situation there is
not identical. I cite it as the same reasoning applied to a neighbouring loop, not as an argument
that the current code violates a policy.

### Two review questions, asked before you have to

**Leftover bytes are now skipped silently.** That is a real behaviour change and it is not always
an improvement. Concrete case: an `iref` whose `dimg` child declares a size large enough to
swallow a following complete `dimg` box.

```
before:  BMFF parsing failed
         Box[iinf] contains duplicate boxes of type 'dimg' with the same from_item_ID value 1
after:   accepted, the swallowed box is skipped
```

Today that malformed file is caught by accident, as a duplicate. After the change it is not
caught at all. Skipping is what the specification asks for, since the bytes inside the declared
child box belong to that child box, and it is what the other bounded loops in this file already
do with leftover bytes. But the two boxes are not in the same position:

- `EntityToGroupBox` is extensible per grouping type. Rejecting leftover bytes there would
  reintroduce the failure above and break every grouping type that defines trailing data.
  Skipping is required.
- `SingleItemTypeReferenceBox` has no defined trailing data. Erroring out on leftover bytes there
  would be defensible.

I kept both consistent and silent, which is the smaller change. If you prefer an error for
leftover bytes in `iref` only, say so and I will add it, with a test.

**The new `avifROStreamSkip()` calls have an unreachable error branch.** After a successful
`avifROStreamReadBoxHeader()` the skip cannot fail, because the header read already rejects a
child size larger than the parent remainder and nothing between the two calls moves the stream.
The return value still has to be checked: `avifROStreamSkip()` is `AVIF_NODISCARD`
(`include/avif/internal.h:703`) and the oss-fuzz build uses `-DAVIF_ENABLE_WERROR=ON`. Ten
existing sites in `read.c` write it the same way (2978, 3210, 3244, 3330, 3520, 3816, 3833,
3856, 3987, 4067), and there is no `(void)` cast on a stream function anywhere in the file.

### Tests

Two cases, one per direction:

- `EntityToGroupTest.PyramidGroupWithGroupingTypeSpecificData` parses the pyramid file written by
  libheif and expects `AVIF_RESULT_OK`. On `main` it fails with
  `Box[grpl]: Child box too large`.
- `ItemReferenceTest.ReferenceCountBeyondChildBoxSizeIsRejected` uses a
  `SingleItemTypeReferenceBox` whose `reference_count` asks for one more `to_item_ID` than the
  declared box size holds, and expects `AVIF_RESULT_BMFF_PARSE_FAILED`. On `main` the file is
  accepted and the `to_item_ID` is taken from the bytes after the box.

Both fail before the change, for the right reason, and pass after it. The test lives in
`tests/gtest/avifpymdtest.cc`, is registered in `tests/CMakeLists.txt` as
`add_avif_gtest_with_data(avifpymdtest)`, and its two data files are `tests/data/pyramid_pymd.avif`
and `tests/data/iref_count_past_box_end.avif`, each with an entry in `tests/data/README.md`.

### Verification

- Full `ctest` suite with gtest, aom and dav1d as system codecs: 61 of 61 pass before and after the
  change, same verdicts per test, no new compiler warnings. With the test below added the suite is
  62 of 62 on this branch. Same with clang: 61 of 61 in both states, zero
  warnings in `src/`, `include/`, `apps/` and `tests/`.
- ASAN and UBSAN over the 54 `tests/data/*.avif` plus the crafted and real files: no reports in
  either state, verified against a deliberate positive control first. The full suite under ASAN
  and UBSAN produces the same three pre-existing failures with and without the change.
- A mutation driver linked against the ASAN and UBSAN builds, same fixed input stream for both
  states, 200 000 mutated inputs per state: zero sanitizer reports on either side.
- FuzzTest in fuzzing mode, four targets, same seed corpus and same time budget per state, about
  4.07 million executions in total: no crash, abort or failure in either state.
- The whole `AVIF_STRICT` matrix, 65 files times 8 flag combinations per state: the 54 real corpus
  files are byte identical between `main` and the patch in all eight combinations, and the flags
  do not change the patch effect on any file. The libheif pyramid file parses after the patch
  under `AVIF_STRICT_ENABLED`, that is the default, not only with strictness disabled.
- All 54 `tests/data/*.avif` through a public API parse harness: byte identical output before and
  after.
- 32 bit build (`-m32`), zero warnings in libavif sources, and the 32 bit parser shows character
  for character the same behaviour difference as the 64 bit one. Not covered: MSVC, macOS, big
  endian, Arm. Your CI runs more than I could.
- `clang-format-19` with the repository style: no diff. `git diff --check` clean, LF only.

### Notes

- No memory safety issue is involved. Every read in both loops already goes through
  `avifROStream*`, which bounds checks against the parent buffer, and
  `avifROStreamReadBoxHeader()` already rejects a child size larger than the parent remainder.
  This is about which bytes get parsed as what, and about not rejecting files that other
  implementations write.
- The pyramid file was produced with libheif built from source with
  `ENABLE_EXPERIMENTAL_FEATURES=ON`, since `--add-pyramid-group` is behind that flag. It is in this
  branch as `tests/data/pyramid_pymd.avif`, 2273 bytes. If you would rather have a smaller hand
  built equivalent, say so and I will replace it.

Nexory (https://github.com/Nexory)


tests/gtest/avifpymdtest.cc
```diff
@@ -0,0 +1,71 @@
...
+
+// The mirror direction: a SingleItemTypeReferenceBox whose reference_count
+// asks for more to_item_ID values than its declared box size holds must not be
+// satisfied from the bytes that follow the box.
+TEST(ItemReferenceTest, ReferenceCountBeyondChildBoxSizeIsRejected) {
+  const testutil::AvifRwData encoded = testutil::ReadFile(
+      std::string(data_path) + "iref_count_past_box_end.avif");
+  ASSERT_NE(encoded.size, size_t{0});
+  EXPECT_EQ(Parse(encoded), AVIF_RESULT_BMFF_PARSE_FAILED);
+}
```
y-guyon: This test already passes at head.

tests/data/pyramid_pymd.avif
y-guyon, line 1: Chrome fails to show this file. Could you check if CrabbyAvif has the same issue, and send a PR there once this one is merged in libavif please?

Gnome's Image Viewer successfully shows a gradient.

src/read.c
y-guyon, line 1: Please describe this change in CHANGELOG.

y-guyon (changes requested):
Thank you for this fix. It looks good.

I agree `irefHeader.size` and `groupHeader.size` are read and ignored, and could contain any wrong value without libavif rejecting the input file. They could also contain more bytes than the ones explicitly read by libavif, which is also an issue.

`EntityToGroupTest.PyramidGroupWithGroupingTypeSpecificData` covers the latter.

I am not sure about `ItemReferenceTest.ReferenceCountBeyondChildBoxSizeIsRejected`. I think it should start from a valid, complete AVIF file, and the only change would be a `SingleItemTypeReferenceBox` size off by a byte (a test for -1 and another test for 1). That file should wrongly decode at head and fail to decode once this patch is applied.

Nexory:
Both points are addressed in 44e0bfe.

### iref test

It now starts from `color_grid_alpha_nogrid.avif`, and the only change is the declared size of its
first SingleItemTypeReferenceBox, the 16 byte `dimg` box at offset 340: one case for -1, one for +1.
The size field is written by the test, so no new data file is needed.

Measured on the same input, one build per state:

| declared size | main | this branch |
|---|---|---|
| 15 (-1) | `AVIF_RESULT_OK` | `AVIF_RESULT_BMFF_PARSE_FAILED`, `Box[iref] SingleItemTypeReference: Failed to read 2 bytes, truncated data?` |
| 16 (unchanged) | `AVIF_RESULT_OK` | `AVIF_RESULT_OK` |
| 17 (+1) | `AVIF_RESULT_OK` | `AVIF_RESULT_BMFF_PARSE_FAILED`, `Box[iref]: Child box too large, possibly truncated data` |

On main both malformed files parse as if the size field still said 16, because the reference fields
come from the parent stream. With the patch the -1 case runs out of bytes on the last `to_item_ID`,
and the +1 case skips one byte into the following `auxl` box, whose header is then rejected.

One detail worth naming: the +1 case only behaves that way on a file whose `iref` has more than one
child. With a single child, `avifROStreamReadBoxHeader()` already rejects the oversized child
against the parent remainder, so that file fails on main as well and the test would pass for the
wrong reason. `color_grid_alpha_nogrid.avif` has three children, which is why I picked it.

I added `ItemReferenceTest.ValidFileParses` as a control, so a failure in the two cases above can
only come from the one byte change, and deleted `tests/data/iref_count_past_box_end.avif` since
nothing uses it now. The `grpl` case no longer sets `AVIF_STRICT_DISABLED`, it passes with the
default strictness.

### CHANGELOG

Entry added under Unreleased, `Changed since 1.4.2`.

Full suite is 62 of 62 with gtest, aom and dav1d as system codecs, no new warnings, and
`clang-format-19` reports no diff.

tests/data/pyramid_pymd.avif
Nexory: I looked at the file first. A full box dump shows nothing structurally off: every box
ends exactly where its parent ends, no slack and no overrun. The 22 bytes after the
entity_id array are a conforming PyramidEntityToGroupBox of ISO/IEC 23008-12:
tile_size_x and tile_size_y of 512, then layer_binning, tiles_in_layer_row_minus1 and
tiles_in_layer_column_minus1 per entity, so 4 plus 3 times 6 bytes, consumed exactly.
The binnings 4, 2 and 1 match the three ispe in ipco (128x128, 256x256, 512x512). To
make that testable I built a control variant of the same file: same total length, same
mdat offset, every byte of image data identical, and the only change is that those 22
bytes become a free box with the pymd and grpl sizes reduced to match. Chrome loads that
variant and reports 128x128. libavif at head and CrabbyAvif accept it as well. Only the
original fails.

CrabbyAvif has the same gap this PR fixes. In src/parser/mp4box.rs, parse_grpl (line
1553) reads the child box header and then never uses header.size, so group_id,
num_entities_in_group and the entity ids all come straight from the parent stream, while
parse_meta in the same file (line 1612) does it correctly with
stream.sub_stream(&header.size)?. parse_iref (line 1499) has the same gap. Running the
file through CrabbyAvif gives BmffParseFailed("possibly truncated box"), which is what
parse_header returns after reading the first bytes of the pymd trailing data as the size
of the next EntityToGroupBox; the control variant parses fine. I did not verify which
decoder Chrome uses, so I am not claiming the two failures share an implementation, only
that they fail and recover on exactly the same pair of files.

Happy to send the CrabbyAvif change once this one is merged, covering both parse_grpl
and parse_iref.

tests/gtest/avifchildboxsizetest.cc
y-guyon, line 1: The last tests in this file have nothing to do with pymd, right? Maybe rename the file to `avifireftest` or something?

y-guyon (approved):

tests/gtest/avifchildboxsizetest.cc
Nexory: True, the ItemReferenceTest cases have nothing to do with pymd anymore.

avifireftest would have the opposite problem though, since EntityToGroupTest would then sit in a file named after iref. Both groups check the same thing, that a child box is parsed within its own declared size, so how about avifchildboxsizetest.cc for the whole file? I am happy to use avifireftest and move the grpl case elsewhere instead if you prefer two files.

y-guyon: `avifchildboxsizetest.cc` sounds good, thanks.

PR #3327, opened 2026-08-19
Reset Sample Transform state in avifDecoderReset()

srkyn:
`avifDecoderReset()` leaves the parsed Sample Transform expression and input count in the decoder data. Calling reset again then adds the inputs a second time. In debug builds this reaches an assertion because the count is expected to start at zero. In release builds the reset returns `AVIF_RESULT_INTERNAL_ERROR`.

Clear both fields with the rest of the per-parse decoder state. The new test uses the existing `weld_sato_12B_8B_q0.avif` fixture and covers explicit reset and the reset performed by `avifDecoderSetSource()`.

Tested with the full ASan and UBSan configuration. All 51 tests passed.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -6093,6 +6093,8 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         memset(&data->tileInfos[c].grid, 0, sizeof(data->tileInfos[c].grid));
     }
     avifDecoderDataClearTiles(data);
+    data->sampleTransformNumInputImageItems = 0;
+    avifArrayDestroy(&data->meta->sampleTransformExpression);
 
     // Prepare / cleanup decoded image state
     if (decoder->image) {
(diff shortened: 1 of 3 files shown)


y-guyon (approved):
Thank you for the fix. Could you add it to the CHANGELOG please?

srkyn:
Added the CHANGELOG entry under Unreleased. Thanks!

y-guyon (approved):

src/read.c
```diff
@@ -6093,6 +6093,8 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         memset(&data->tileInfos[c].grid, 0, sizeof(data->tileInfos[c].grid));
     }
     avifDecoderDataClearTiles(data);
+    data->sampleTransformNumInputImageItems = 0;
+    avifArrayDestroy(&data->meta->sampleTransformExpression);
```
wantehchang, line 6097: I suggest we also set `data->meta->sampleTransformDepth` to an invalid value such as 0.

srkyn: Done in #3333. Thanks!

PR #3331, opened 2026-08-21
Keep every target of a 'thmb', 'auxl', 'cdsc' or 'prem' reference

Nexory:
A SingleItemTypeReferenceBox links one item to an array of to_item_IDs
(ISO/IEC 14496-12, section 8.11.12.1). libavif stored the 'thmb', 'auxl',
'cdsc' and 'prem' relations in a single-valued field, and the parsing loop
overwrote it per entry, so the target parsed last silently won.

Measured on files differing from a working one only in the order of two
to_item_ID values, the Exif item describing items [1, 2]:

```
cdsc 3 -> [1]      Exif Metadata : Present (14 bytes)   exit 0
cdsc 3 -> [1, 2]   Exif Metadata : Absent               exit 0
cdsc 3 -> [2, 1]   Exif Metadata : Present (14 bytes)   exit 0
```

The decoded pixels are identical in all three, only the Exif payload is gone. For
'auxl' the effect is heavier: with the alpha item declared auxiliary for
[1, 3], the alpha plane is dropped and the image decodes as opaque, again with
exit code 0 and no diagnostic. Two boxes of one type sharing a from_item_ID
behave the same.

This keeps every target instead. The references live in a per item array, the
parsing loop pushes rather than overwrites, and the read sites became
membership checks. 'dimg' is unchanged: it refers in the opposite direction and
already refuses what it cannot represent.

Without the change both tests fail with the Exif data and the alpha plane
missing, while the decode itself still returns AVIF_RESULT_OK.

A second commit fixes the MinimizedImageBox path, which stored the one bit
alpha_is_premultiplied value in the same field. Under the array introduced here
that is a reference to item 1, the primary item itself, so a premultiplied image
decoded as straight alpha. The parameterized test in avifminitest.cc already
compares a MinimizedImageBox against a regular MetaBox and compares
alphaPremultiplied, but no parameter set marks the alpha as premultiplied, so
the flag was never exercised. The added test does.

Two limits remain. I found no encoder that writes such files (libheif passes a
single ID at every call site), so both test files are hand-crafted. I built with
dav1d and aom from the system, rav1e, svt and libgav1 off, not the five codecs
the CI builds from source.


y-guyon:

I am not so sure that rejecting files is a better behavior. How about fixing the issue instead?

Nexory:
Fair point, and I think you are right.

I checked how the four relations are consumed. Across the tree they appear only
in src/read.c, and the eight read sites on avifDecoderItem never ask for "the"
target, only whether one specific item is among them; thumbnailForID is just
compared against 0. The avifTrack fields of the same name are separate.

So the fields can become id arrays, the loop pushes instead of overwriting, and
those comparisons become a membership check.

Happy to rework the PR that way, with the two tests asserting that the metadata
and the alpha plane survive instead of asserting a rejection. In this PR, or a
fresh one?

y-guyon:
> Happy to rework the PR that way, with the two tests asserting that the metadata and the alpha plane survive instead of asserting a rejection. In this PR, or a fresh one?

As you wish!

Nexory:
Reworked as suggested.

The four single-valued fields became one per item array of (type, toID)
entries. The parsing loop pushes instead of overwriting, and the read sites
became membership checks. 'dimg' is untouched: it refers in the opposite
direction and already refuses what it cannot represent.

Both tests now assert survival instead of rejection. The Exif metadata is
present for a 'cdsc' naming items [1, 2], and the alpha plane is there for an
'auxl' naming [1, 3].

One thing I did not change: in the MinimizedImageBox path, the line that used
to read

stores a one bit flag in a field the consumer compares against the alpha item
ID, which is fixed at 2 in that path. I translated it unchanged rather than
fixing it here. Happy to file that separately if you want it looked at.

Nexory:
I measured the MinimizedImageBox line I mentioned, and folded the fix into this
PR after all, because it turned out not to be a separate issue.

Under the array this PR introduces, toID is an item ID. The old line stored the
one bit alpha_is_premultiplied value, and the primary item ID in that path is 1,
so translating it faithfully wrote "item 1 is premultiplied by item 1" into the
new structure. Keeping that seemed worse than fixing it.

The behaviour on main, measured with avifenc on one input image: encoded with
-p it decodes as premultiplied through a regular MetaBox, and as straight alpha
through --mini. Changing only the reader and decoding the same files restores
it, while an image encoded without -p stays unpremultiplied either way.

The parameterized test in avifminitest.cc already compares a MinimizedImageBox
against a regular MetaBox for the same image, and AreImagesEqual compares
alphaPremultiplied. No parameter set marks the alpha as premultiplied, so the
flag was never exercised. The added test does, and it fails without the change.

Happy to split it back out if you would rather review the two separately.

src/read.c
```diff
@@ -3390,12 +3440,12 @@ static avifResult avifParseItemReferenceBox(avifMeta * meta, const uint8_t * raw
...
-                item->thumbnailForID = toID;
-            } else if (!memcmp(irefHeader.type, "auxl", 4)) {
-                item->auxForID = toID;
-            } else if (!memcmp(irefHeader.type, "cdsc", 4)) {
-                item->descForID = toID;
+            if (!memcmp(irefHeader.type, "thmb", 4) || !memcmp(irefHeader.type, "auxl", 4) ||
+                !memcmp(irefHeader.type, "cdsc", 4) || !memcmp(irefHeader.type, "prem", 4)) {
+                // Section 8.11.12.1 of ISO/IEC 14496-12 represents the items linked to as an array
+                // of to_item_IDs, and several boxes of the same type may share a from_item_ID, so
+                // keep every target instead of only the one parsed last.
```
y-guyon: The comments should describe the current implementation and not explain why it changed or how it was before.

y-guyon: The comments should describe the current implementation and not explain why it changed or how it was before.

src/read.c
```diff
@@ -6450,8 +6501,8 @@ avifResult avifDecoderReset(avifDecoder * decoder)
                 AVIF_CHECKERR(!mainItems[alphaCategory] == !mainItems[AVIF_ITEM_ALPHA], AVIF_RESULT_NOT_IMPLEMENTED);
                 if (mainItems[alphaCategory] != NULL) {
                     AVIF_CHECKERR(isAlphaInputImageItemInInput == isAlphaItemInInput, AVIF_RESULT_NOT_IMPLEMENTED);
-                    AVIF_CHECKERR((mainItems[*category]->premByID == mainItems[alphaCategory]->id) ==
-                                      (mainItems[AVIF_ITEM_COLOR]->premByID == mainItems[AVIF_ITEM_ALPHA]->id),
+                    AVIF_CHECKERR(avifDecoderItemReferences(mainItems[*category], "prem", mainItems[alphaCategory]->id) ==
+                                      avifDecoderItemReferences(mainItems[AVIF_ITEM_COLOR], "prem", mainItems[AVIF_ITEM_ALPHA]->id),
```
y-guyon: These should be reversed, right? It was `premByID`, not `premForID` like all other ref types.

src/read.c
```diff
@@ -6551,8 +6602,8 @@ avifResult avifDecoderReset(avifDecoder * decoder)
         decoder->image->width = mainItems[AVIF_ITEM_COLOR]->width;
         decoder->image->height = mainItems[AVIF_ITEM_COLOR]->height;
         decoder->alphaPresent = (mainItems[AVIF_ITEM_ALPHA] != NULL);
-        decoder->image->alphaPremultiplied = decoder->alphaPresent &&
-                                             (mainItems[AVIF_ITEM_COLOR]->premByID == mainItems[AVIF_ITEM_ALPHA]->id);
+        decoder->image->alphaPremultiplied =
+            decoder->alphaPresent && avifDecoderItemReferences(mainItems[AVIF_ITEM_COLOR], "prem", mainItems[AVIF_ITEM_ALPHA]->id);
```
y-guyon: Same. Please also add a test covering this, this should have been caught in the CI.

src/read.c
y-guyon, line 6356: Just a note: `avifTrack`s have a similar issue with `auxForID` and `premByID` not being arrays. This may be less of a problem though.

tests/gtest/avifmetadatatest.cc
```diff
@@ -467,6 +467,50 @@ TEST(MetadataTest, XMPWithTrailingNullCharacter) {
...
+TEST(MetadataTest, CdscWithMultipleTargets) {
+  testutil::AvifRwData avif = testutil::ReadFile(
+      std::string(data_path) + "circle_cdsc_two_targets.avif");
+  ASSERT_NE(avif.size, 0u);
+  ImagePtr decoded(avifImageCreateEmpty());
+  ASSERT_NE(decoded, nullptr);
+  DecoderPtr decoder(avifDecoderCreate());
+  ASSERT_NE(decoder, nullptr);
+  ASSERT_EQ(
+      avifDecoderReadMemory(decoder.get(), decoded.get(), avif.data, avif.size),
```
y-guyon: Why not directly `avifDecoderReadFile()`?

tests/data/circle_auxl_two_targets.avif
y-guyon, line 1: Note: this file is displayed with a green background in Chrome (uses CrabbyAvif) and with a transparent background in Gnome's image viewer Loupe (uses libheif I think).

tests/gtest/avifmetadatatest.cc
```diff
@@ -467,6 +467,50 @@ TEST(MetadataTest, XMPWithTrailingNullCharacter) {
...
 
+//------------------------------------------------------------------------------
+// Multiple targets in a single item type reference box
+
+// Section 8.11.12.1 of ISO/IEC 14496-12 represents the targets of a
+// SingleItemTypeReferenceBox as an array of to_item_IDs. libavif used to keep
+// one target per item, so the target parsed last silently won.
+// In circle_cdsc_two_targets.avif the Exif item describes items [1, 2]. The
+// Exif metadata used to disappear without any error, exit code or diagnostic;
+// swapping the two IDs made it reappear. It must survive either way.
```
y-guyon, line 478: Describing the old implementation is fine for a regression test.

tests/gtest/avifminitest.cc
```diff
@@ -129,6 +129,54 @@ TEST_P(AvifMinimizedImageBoxTest, All) {
   }
 }
 
+// The parameterized test above compares a MinimizedImageBox against a regular
+// MetaBox for the same image, and AreImagesEqual() compares alphaPremultiplied.
+// No parameter set marks the alpha as premultiplied though, so the flag was
+// never exercised: the MinimizedImageBox path stored the one bit
+// alpha_is_premultiplied value in a field the decoder compares against the
+// alpha item ID, and a premultiplied image silently decoded as straight alpha.
+TEST(MinimizedImageBoxPremultipliedTest, SurvivesTheRoundTrip) {
```
y-guyon: Please do not insert a `TEST` between a `TEST_P` and its `INSTANTIATE_TEST_SUITE_P`.

tests/gtest/avifminitest.cc
```diff
@@ -129,6 +129,54 @@ TEST_P(AvifMinimizedImageBoxTest, All) {
   }
 }
 
+// The parameterized test above compares a MinimizedImageBox against a regular
+// MetaBox for the same image, and AreImagesEqual() compares alphaPremultiplied.
```
y-guyon: I suggest adding a premul field in `AvifMinimizedImageBoxTest` instead of adding this whole new test.

src/read.c
```diff
@@ -4532,8 +4580,10 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
 
     if (hasAlpha) {
         // Property with fixed index 7.
-        alphaItem->auxForID = colorItem->id;
-        colorItem->premByID = alphaIsPremultiplied;
+        AVIF_CHECKRES(avifDecoderItemAddReference(alphaItem, "auxl", colorItem->id));
+        if (alphaIsPremultiplied) {
+            AVIF_CHECKRES(avifDecoderItemAddReference(colorItem, "prem", alphaItem->id));
```
y-guyon, line 4586: Note: That corresponds to the spec:

> An item reference of type 'prem' from the master image item to the auxiliary image item signals that the master image is pre-multiplied by the alpha value.

y-guyon (changes requested):
Thanks. Please also describe the added test images in `tests/data/README.md`.

Nexory:
Thanks, all five are done.

- `src/read.c`: the comment is in the present tense now, following the `dimg`
  branch a few lines below it.
- `tests/data/README.md`: both files described, with the `avifenc` command and
  the manual `iref` edit that produced them.
- `tests/gtest/avifmetadatatest.cc`: both tests go through `avifDecoderReadFile`.
- `tests/gtest/avifminitest.cc`: the standalone `TEST` is gone. The
  premultiplied case is a `create_premul` parameter on
  `AvifMinimizedImageBoxTest` instead, enabled only in `Dimensions`, which is
  the one instantiation with a single combination and `AVIF_PLANES_ALL`, so it
  costs one extra case.
- The regression-test comments describing the old behaviour are left as they
  are, per your note.

One thing worth flagging on the parameter: `AreImagesEqual()` compares
`alphaPremultiplied` only when an alpha plane is present and the image is not
opaque. Today the second guard happens to hold because `FillImageGradient()`
also writes a gradient into `AVIF_CHAN_A`, which is not obvious from the test.
I added an explicit assertion next to `AreImagesEqual()` so the check does not
depend on that.

On the `avifTrack` note: same shape, different box. `avifTrackReferenceBox()`
reads one `track_ID` and skips the rest, and the loop has no duplicate guard,
so a second `auxl` box overwrites `auxForID`. The consumer compares `auxForID`
against `colorTrack->id`, so a dropped ID loses the alpha track the same way it
did for items. It sits in `tref` (8.3.3) rather than `iref` (8.11.12) and it
needs an animated file carrying two `track_ID`s, which I would have to build by
hand, so I would rather send it separately than double the size of this PR.
Happy to file it.

On the Chrome and Loupe difference, thanks for checking it in two more
implementations. I went looking, and it is the same defect in a third decoder
rather than a rendering choice. CrabbyAvif keeps `auxl` in a single field
(`src/decoder/item.rs`, `"auxl" => item.aux_for_id = reference.to_item_id`),
while `cdsc` on the line below it already collects into a `Vec`. Running its
CLI over the two files in this PR: for the `auxl` one it reports
`Alpha: Absent` and still finds the Exif, and for the `cdsc` one it reports the
alpha normally. The lost alpha association is what makes the image opaque, so
the green background of the source shows through. I ran that CLI, not Chrome or
libheif, so the browser half stays your observation. I can report it there
separately; it does not belong in this PR.

tests/data/README.md
```diff
@@ -54,6 +54,29 @@ follow the PLTE chunk, if any", libpng 1.6.46 or older considers the tRNS chunk
...
+![](circle_auxl_two_targets.avif)
+
+License: [same as libavif](https://github.com/AOMediaCodec/libavif/blob/main/LICENSE)
+
+Source: `avifenc -q 60 -y 420 --exif exif.bin circle-trns-after-plte.png`, where
+exif.bin is a minimal 14-byte TIFF header, with the `auxl` reference from the
+alpha item 2 manually edited to list the two targets 1,3 instead of the single
+target 1. Item 1 is the primary color item and item 3 is the Exif item.
+Section 8.11.12.1 of ISO/IEC 14496-12 represents the items linked to as an
+array of to_item_IDs, so the file is valid.
```
y-guyon: I would omit this part. Alpha being an `auxl` of an Exif item makes no sense, so saying the file is valid is somewhat misleading.

tests/gtest/avifminitest.cc
```diff
@@ -119,6 +122,12 @@ TEST_P(AvifMinimizedImageBoxTest, All) {
   EXPECT_TRUE(
       testutil::AreImagesEqual(*decoded_meta.get(), *decoded_mini.get()));
   EXPECT_EQ(decoded_meta->gainMap != nullptr, decoded_mini->gainMap != nullptr);
+  // AreImagesEqual() only compares alphaPremultiplied when an alpha plane is
+  // present and the image is not opaque, so assert the flag on its own too.
+  EXPECT_EQ(decoded_mini->alphaPremultiplied, decoded_meta->alphaPremultiplied);
```
y-guyon, line 127: I doubt this is an API invariant worth enforcing and supporting in libavif, but this can stay in the test.

y-guyon (approved):
Thanks! Do you mind creating the same PR for https://github.com/webmproject/CrabbyAvif once this PR is merged?

> On the `avifTrack` note: same shape, different box. `avifTrackReferenceBox()`
> reads one `track_ID` and skips the rest, and the loop has no duplicate guard,
> so a second `auxl` box overwrites `auxForID`. The consumer compares `auxForID`
> against `colorTrack->id`, so a dropped ID loses the alpha track the same way it
> did for items. It sits in `tref` (8.3.3) rather than `iref` (8.11.12) and it
> needs an animated file carrying two `track_ID`s, which I would have to build by
> hand, so I would rather send it separately than double the size of this PR.
> Happy to file it.

Sure!

Nexory:
Sure, I will send it once this one is merged.

I looked at where it would go. In `src/decoder/item.rs` the parsed references are
folded into per-item fields, and three of the four are scalar assignments that a
later reference overwrites: `thmb` into `thumbnail_for_id`, `auxl` into
`aux_for_id` and `prem` into `prem_by_id`, all `u32`. `cdsc` already collects into
`desc_for_id: Vec<u32>`, so the shape to move the other three to is in the same
match. `parse_iref` keeps every reference with its index, so nothing is lost
before that point and the change stays inside that fold.

`circle_auxl_two_targets.avif` carries over directly.
`circle_cdsc_two_targets.avif` should already pass there, since `cdsc` is the one
of the four that already collects.

y-guyon (approved):

src/read.c
```diff
@@ -4532,8 +4581,10 @@ static avifResult avifParseMinimizedImageBox(avifDecoderData * data,
...
+        }
```
wantehchang: This pull request may have introduced [an assertion failure in fuzzing](https://g-issues.oss-fuzz.com/issues/553221196), which is why I am taking a look.

To preserve the original behavior faithfully, we should change this if statement (lines 4585-4587) to:

```
        AVIF_CHECKRES(avifDecoderItemAddReference(colorItem, "prem", alphaIsPremultiplied));
```

src/read.c
wantehchang: What about `colorTrack->premByID`?

src/read.c
```diff
@@ -870,6 +911,7 @@ static void avifMetaDestroy(avifMeta * meta)
         avifDecoderItem * item = meta->items.item[i];
         avifPropertyArrayDestroy(&item->properties);
         avifArrayDestroy(&item->extents);
+        avifArrayDestroy(&item->references);
```
wantehchang: Nit: Move after `avifRWDataFree(&item->mergedExtents)`

PR #3333, opened 2026-08-21
Reset Sample Transform depth in avifDecoderReset()

srkyn:
Follow-up to #3327. This resets sampleTransformDepth to 0 alongside the other Sample Transform state in avifDecoderReset(), as suggested in the review. A fresh Release library build passes.


wantehchang (approved):
LGTM. Thanks!

PR #3335, opened 2026-08-25
Keep every target of a track 'auxl' or 'prem' reference

Nexory:
Follow-up to your note on #3331, that `avifTrack`s have a similar issue with
`auxForID` and `premByID` not being arrays:
https://github.com/AOMediaCodec/libavif/pull/3331#discussion_r3844045380

`avifTrackReferenceBox` loses track references in two ways.

The first is the one the code comment admits to:

    AVIF_CHECK(avifROStreamReadU32(&s, &toID));                       // unsigned int(32) track_IDs[];
    AVIF_CHECK(avifROStreamSkip(&s, header.size - sizeof(uint32_t))); // just take the first one

Section 8.3.3.2 of ISO/IEC 14496-12 gives TrackReferenceTypeBox an array of
track_IDs with no count field, so the number of entries follows from the box
size and everything after the first is dropped.

The second is that `track->auxForID = toID` is unconditional, and nothing limits
a 'tref' to one box of a given type, so a second 'auxl' box overwrites what the
first one recorded. `avifParseTrackBox` keeps a seen flag for 'tkhd' and for
'edts' but not for 'tref', so several 'tref' boxes in one 'trak' do the same.

## What it costs

`auxForID` is compared against `colorTrack->id` when the alpha track is looked
up. If the color track is not the target that survived, the loop ends without a
match, `alphaTrack` stays null and `decoder->alphaPresent` is false. There is no
diagnostic and no error code: the file decodes as a plain color sequence.

`premByID` has the same shape, with `alphaPremultiplied` as its consumer. I did
not build a file for that path, so I am reporting it from the code rather than
from a run.

You wrote that this may be less of a problem, and that matches what it does: a
file that a conformant reader shows with alpha decodes here without it, quietly.
There is no attacker gain and no memory error in it.

## The change

The three reference helpers added in #3331 are renamed and now take the
`avifItemReferenceArray` instead of the item, so tracks can use them unchanged.
That is why the diff touches the item call sites.

`avifTrack` gets the same `references` array, created with the track and
destroyed with it. `avifTrackReferenceBox` reads track_IDs until the box runs
out, and the three places that compared `auxForID` and `premByID` now ask
`avifReferences` and `avifHasReference`.

Three deliberate decisions, so they do not have to be found in the diff:

1. The `header.size >= sizeof(uint32_t)` check from #3324 stays, so a box that
   holds no track_ID at all is still refused.
2. The box size is not required to be a multiple of four. Trailing bytes that do
   not make up a whole track_ID are skipped as before.
3. A track_ID of 0 is skipped rather than recorded. `auxForID` was an in-band
   sentinel where 0 meant "no reference", so recording it would turn a file that
   used to read as "no reference" into one that has one. The item side answers
   the same question differently: `avifParseItemReferenceBox` passes to_item_ID
   through `avifCheckItemID`, which refuses 0 outright. There is no equivalent
   check for track_IDs today, and adding one would refuse files that are read
   now, so this skips instead. Happy to make it refuse if you prefer the item
   side's answer here too.

That third one is a compatibility choice, not a measurement, and it is not
complete: a 'tref' whose only 'auxl' names 0 and then a real track used to leave
`auxForID` at 0, so the track stayed a color candidate. It now has a reference
and is skipped, and if it were the only usable track the decode would end in
`AVIF_RESULT_NO_CONTENT`. I think refusing to treat a track that declares itself
auxiliary as the color track is the better reading, but say the word if you would
rather keep the old one.

## Tests

The alpha track of `colors-animated-8bpc-alpha-exif-xmp.avif` holds a 20 byte
'tref' at offset 1586 with one 'auxl' box naming track 1, directly followed by a
44 byte 'edts'. The three tests rewrite those 64 bytes in place and pad with a
'free' box, so the file size and every absolute offset in it are untouched, which
matters because the file carries 'iloc' and 'stco' entries.

The first test is the control: it keeps the single reference and only turns the
'edts' box into a 'free' box, so the rewrite itself does not cost the alpha
channel. Dropping 'edts' leaves that track's repetition count unknown, which the
decoder takes from the color track.

The other two present the two cases above: two 'auxl' boxes, and one 'auxl' box
naming two tracks with the color track second. They fail without the change and
pass with it, and the control passes either way.

## One thing I left alone

`avifTrackReferenceBox` still returns `avifBool`, so an allocation failure inside
it surfaces as `AVIF_RESULT_BMFF_PARSE_FAILED` rather than
`AVIF_RESULT_OUT_OF_MEMORY`. Changing the signature would widen the diff; happy
to do it here if you would rather have it in one go.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -252,21 +252,21 @@ typedef struct avifDecoderItem
 } avifDecoderItem;
 AVIF_ARRAY_DECLARE(avifDecoderItemArray, avifDecoderItem *, item);
 
-// Records "this item is a {type} for item #{toID}".
-static avifResult avifDecoderItemAddReference(avifDecoderItem * item, const char * type, uint32_t toID)
+// Records "this item or track is a {type} for #{toID}".
+static avifResult avifAddReference(avifItemReferenceArray * references, const char * type, uint32_t toID)
 {
-    avifItemReference * reference = (avifItemReference *)avifArrayPush(&item->references);
+    avifItemReference * reference = (avifItemReference *)avifArrayPush(references);
     AVIF_CHECKERR(reference != NULL, AVIF_RESULT_OUT_OF_MEMORY);
     memcpy(reference->type, type, 4);
     reference->toID = toID;
     return AVIF_RESULT_OK;
 }
 
-// Returns AVIF_TRUE if this item is a {type} for item #{toID}.
-static avifBool avifDecoderItemReferences(const avifDecoderItem * item, const char * type, uint32_t toID)
+// Returns AVIF_TRUE if this item or track is a {type} for #{toID}.
+static avifBool avifReferences(const avifItemReferenceArray * references, const char * type, uint32_t toID)
 {
-    for (uint32_t i = 0; i < item->references.count; ++i) {
-        const avifItemReference * reference = &item->references.ref[i];
+    for (uint32_t i = 0; i < references->count; ++i) {
+        const avifItemReference * reference = &references->ref[i];
         if (!memcmp(reference->type, type, 4) && reference->toID == toID) {
             return AVIF_TRUE;
         }
@@ -274,11 +274,11 @@ static avifBool avifDecoderItemReferences(const avifDecoderItem * item, const ch
     return AVIF_FALSE;
 }
 
-// Returns AVIF_TRUE if this item is a {type} for at least one item.
-static avifBool avifDecoderItemHasReference(const avifDecoderItem * item, const char * type)
+// Returns AVIF_TRUE if this item or track is a {type} for at least one item or track.
+static avifBool avifHasReference(const avifItemReferenceArray * references, const char * type)
 {
-    for (uint32_t i = 0; i < item->references.count; ++i) {
-        if (!memcmp(item->references.ref[i].type, type, 4)) {
+    for (uint32_t i = 0; i < references->count; ++i) {
+        if (!memcmp(references->ref[i].type, type, 4)) {
             return AVIF_TRUE;
         }
     }
@@ -501,8 +501,10 @@ typedef struct avifTrack
 {
     uint32_t id;
     uint8_t handlerType[4];
-    uint32_t auxForID; // if non-zero, this track is an auxC plane for Track #{auxForID}
-    uint32_t premByID; // if non-zero, this track is premultiplied by Track #{premByID}
+    // The 'auxl' and 'prem' references from this track. Section 8.3.3.2 of ISO/IEC 14496-12
+    // allows an array of track_IDs per box, and nothing stops a 'tref' from holding several
+    // boxes of the same type, so one track can hold several of these.
+    avifItemReferenceArray references;
     uint32_t mediaTimescale;
     uint64_t mediaDuration;
     uint64_t trackDuration;
@@ -1128,6 +1130,11 @@ static avifTrack * avifDecoderDataCreateTrack(avifDecoderData * data)
         avifArrayPop(&data->tracks);
         return NULL;
     }
+    if (!avifArrayCreate(&track->references, sizeof(avifItemReference), 1)) {
+        avifMetaDestroy(track->meta);
+        avifArrayPop(&data->tracks);
+        return NULL;
+    }
     return track;
 }
 
@@ -1179,6 +1186,7 @@ static void avifDecoderDataDestroy(avifDecoderData * data)
         if (track->meta) {
             avifMetaDestroy(track->meta);
         }
+        avifArrayDestroy(&track->references);
     }
     avifArrayDestroy(&data->tracks);
     avifDecoderDataClearTiles(data);
@@ -1947,7 +1955,7 @@ static avifResult avifDecoderFindMetadata(avifDecoder * decoder, avifMeta * meta
             continue;
         }
 
-        if ((colorId > 0) && !avifDecoderItemReferences(item, "cdsc", colorId)) {
+        if ((colorId > 0) && !avifReferences(&item->references, "cdsc", colorId)) {
             // Not a content description (metadata) for the colorOBU, skip it
             continue;
         }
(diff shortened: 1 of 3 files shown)


src/read.c
```diff
@@ -501,8 +501,10 @@ typedef struct avifTrack
 {
     uint32_t id;
     uint8_t handlerType[4];
-    uint32_t auxForID; // if non-zero, this track is an auxC plane for Track #{auxForID}
-    uint32_t premByID; // if non-zero, this track is premultiplied by Track #{premByID}
+    // The 'auxl' and 'prem' references from this track. Section 8.3.3.2 of ISO/IEC 14496-12
+    // allows an array of track_IDs per box, and nothing stops a 'tref' from holding several
+    // boxes of the same type, so one track can hold several of these.
```
maryla-uc: 8.3.3.1 actually says "Each reference type shall occur at most once." but we have never enforced this it seems therefore it's slightly risky to start enforcing now (alhtough I doubt it would cause issues in practice...)
You can say that we allow multiple ones for backward compatibility.

src/read.c
```diff
@@ -3910,18 +3918,28 @@ static avifBool avifTrackReferenceBox(avifTrack * track, const uint8_t * raw, si
         avifBoxHeader header;
         AVIF_CHECK(avifROStreamReadBoxHeader(&s, &header));
 
-        if (!memcmp(header.type, "auxl", 4)) {
+        if (!memcmp(header.type, "auxl", 4) || !memcmp(header.type, "prem", 4)) {
+            // Keep refusing a box that holds no track_ID at all, as added in #3324.
```
maryla-uc: No need to document past behavior especially when you are not changing it.
// Refuse a box that holds no track_ID at all.

src/read.c
```diff
@@ -3910,18 +3918,28 @@ static avifBool avifTrackReferenceBox(avifTrack * track, const uint8_t * raw, si
...
-        } else if (!memcmp(header.type, "prem", 4)) {
-            AVIF_CHECK(header.size >= sizeof(uint32_t));
-            uint32_t byID;
-            AVIF_CHECK(avifROStreamReadU32(&s, &byID));                       // unsigned int(32) track_IDs[];
-            AVIF_CHECK(avifROStreamSkip(&s, header.size - sizeof(uint32_t))); // just take the first one
-            track->premByID = byID;
+            // Section 8.3.3.2 of ISO/IEC 14496-12:
+            //   unsigned int(32) track_IDs[];
+            // There is no count field: the number of track_IDs follows from the box size. Every
+            // target is kept. Bytes that do not make up a whole track_ID are skipped as before,
```
maryla-uc: again no need to document that you are not changing the behavior. remove everything after "skipped".

src/read.c
```diff
@@ -3910,18 +3918,28 @@ static avifBool avifTrackReferenceBox(avifTrack * track, const uint8_t * raw, si
...
+            // so this does not refuse a file that used to be read.
+            const size_t numTrackIDs = header.size / sizeof(uint32_t);
+            for (size_t i = 0; i < numTrackIDs; ++i) {
+                uint32_t toID;
+                AVIF_CHECK(avifROStreamReadU32(&s, &toID));
+                if (toID == 0) {
+                    // auxForID and premByID used to be plain IDs where 0 meant "no reference", so a
```
maryla-uc: Again I think there is a simpler way to document this without mentioning the previous code in detail e.g.
```
// Section 8.3.3.3 of ISO/IEC 14496-12:
//   The value 0 shall not be present.
// Skip it instead of rejecting for backward compatibility.
```

tests/gtest/avifanimationtest.cc
```diff
@@ -98,6 +101,122 @@ TEST(AvifDecodeTest, AnimatedImageWithAlphaAndMetadata) {
...
+constexpr size_t kTrefAndEdtsSize = 64;
+
+void WriteBE32(uint32_t value, uint8_t* bytes) {
+  bytes[0] = static_cast<uint8_t>(value >> 24);
+  bytes[1] = static_cast<uint8_t>(value >> 16);
+  bytes[2] = static_cast<uint8_t>(value >> 8);
+  bytes[3] = static_cast<uint8_t>(value);
+}
+
+void AppendBox(const char* type, const std::vector<uint32_t>& track_ids,
```
maryla-uc: AppendTrackReferenceTypeBox would be a more explicit name

tests/gtest/avifanimationtest.cc
```diff
@@ -98,6 +101,122 @@ TEST(AvifDecodeTest, AnimatedImageWithAlphaAndMetadata) {
...
+  AppendBox("auxl", {1}, &tref);
+  std::vector<uint8_t> boxes;
+  AppendBox("tref", {}, &boxes);
+  WriteBE32(static_cast<uint32_t>(8 + tref.size()), boxes.data());
+  boxes.insert(boxes.end(), tref.begin(), tref.end());
+  ASSERT_NO_FATAL_FAILURE(ReplaceTrefAndEdts(boxes, &encoded));
+  ExpectAlphaTrackIsFound(encoded);
+}
+
+// The 'tref' box holds two 'auxl' boxes. The second one used to overwrite the
```
maryla-uc: rather than saing "used to"... just say "check that both are read" or something. same for the other test.

maryla-uc (commented):
Thanks for the change. It's a good fix and the code looks good in general although I have some nits mostly on comments that describe behavior change (or lack thereof). I think y-guyon also made a similar comment on the previous PR, please add this to your prompt or something. Comments should describe the current behavior, and only mention past behavior when particularly relevant, for example explaning that we do not enforce a particular constraint for backward compatibility (because it used not to be enforced).

For the avifTrackReferenceBox return value, it's fine to do in another PR.

Can you also check if this issue also exists in CrabbyAvif?

Nexory:
Thanks, all of these are addressed:

- the struct comment cites 8.3.3.1 and says multiple boxes of a type are allowed
  for backward compatibility, instead of describing the old code
- the empty box check no longer points at the earlier PR
- the box size comment ends at "skipped"
- the track_ID 0 comment is your wording, with 8.3.3.3
- `AppendBox` is now `AppendTrackReferenceTypeBox`
- both test comments say what is checked

The return value of `avifTrackReferenceBox` I will send separately.

On CrabbyAvif: yes, on both sides.

The item side is the same defect as #3331. It is already in, as
webmproject/CrabbyAvif#878: `fold_item_references` kept one target per reference
type, while `cdsc` had already been moved to a `Vec` in 7ccb19e.

The track side has it too and is still open there. `parse_tref` in
`src/parser/mp4box.rs` reads one track_ID per box and says so in a comment, and
`track.aux_for_id` is assigned unconditionally, so a second `auxl` box wins.
`parse_elst` right below it keeps a seen flag, so the shape is already in the
file. I will send that once this one lands, unless you would rather have it now.

PR #3336, opened 2026-08-26
Fix memory leaks on error paths in avifRGBImageComputeGainMap()

krishna28238-arch:
Two error paths in `avifRGBImageComputeGainMap()` returned directly instead of going through the `cleanup` label, breaking the contract documented right above the allocations ("After this point, the function should exit with 'goto cleanup' to free allocated resources"):

1. the "Unsupported RGB color space" path, reachable with a gain map image depth that is only rejected after the intermediate buffers are allocated (e.g. 9),
2. the `avifImageScale()` failure path.

Both leaked:
- the `gainMapF[]` channel buffers (up to 3 × width × height × sizeof(float)),
- the `gainMapRGB` pixels,

and left the gain map image planes allocated.

The fix routes both paths through the existing `cleanup` label. A regression test computes a gain map with an unsupported depth and checks that the image planes are freed on error.

Validation (build with `-fsanitize=address`, `AVIF_GTEST=LOCAL`, no codec):

- Before the fix, LeakSanitizer reports 1200 bytes in 3 objects (`gainMapF[]`, gainmap.c:586) and 800 bytes in 1 object (`gainMapRGB` pixels, gainmap.c:799) leaked by the new test; after the fix, no leaks.
- Full `avifgainmaptest` suite: no regression, the only delta is the new test going from FAIL to PASS.

### DIFF
--- a/src/gainmap.c
+++ b/src/gainmap.c
@@ -804,7 +804,8 @@ avifResult avifRGBImageComputeGainMap(const avifRGBImage * baseRgbImage,
     avifRGBColorSpaceInfo gainMapRGBInfo;
     if (!avifGetRGBColorSpaceInfo(&gainMapRGB, &gainMapRGBInfo)) {
         avifDiagnosticsPrintf(diag, "Unsupported RGB color space");
-        return AVIF_RESULT_NOT_IMPLEMENTED;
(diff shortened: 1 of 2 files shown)


y-guyon (approved):
Thanks!

PR #3339, opened 2026-08-27
Allow unset elements in the dimgIdxToItemIdx array

wantehchang:
Allow the dimgIdxToItemIdx and dimgIdxToAlphaItemIdx arrays to have unset elements. When these arrays are consumed, skip the unset elements.

Unset elements in the dimgIdxToItemIdx and dimgIdxToAlphaItemIdx arrays need to be allowed after commit 2f8eb91 "Keep every target of a 'thmb', 'auxl', 'cdsc' or 'prem' reference".

Bug: b:553462837
Bug: b:553463408

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1652,7 +1652,7 @@ static avifResult avifFillDimgIdxToItemIdxArray(uint32_t * dimgIdxToItemIdx, uin
         }
     }
     // The number of tiles has been verified in avifDecoderItemReadAndParse().
-    AVIF_ASSERT_OR_RETURN(numTiles == numExpectedTiles);
+    AVIF_ASSERT_OR_RETURN(numTiles <= numExpectedTiles);
     return AVIF_RESULT_OK;
 }
 
@@ -1663,9 +1663,13 @@ static avifResult avifDecoderAdoptGridTileCodecType(avifDecoder * decoder,
                                                     const uint32_t * dimgIdxToItemIdx,
                                                     uint32_t numTiles)
 {
+    const uint32_t itemIndexNotSet = UINT32_MAX;
     avifDecoderItem * firstTileItem = NULL;
     for (uint32_t dimgIdx = 0; dimgIdx < numTiles; ++dimgIdx) {
         const uint32_t itemIdx = dimgIdxToItemIdx[dimgIdx];
+        if (itemIdx == itemIndexNotSet) {
+            continue;
+        }
         AVIF_ASSERT_OR_RETURN(itemIdx < gridItem->meta->items.count);
         avifDecoderItem * item = gridItem->meta->items.item[itemIdx];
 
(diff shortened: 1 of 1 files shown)


wantehchang:
cc: @Nexory

Please review this pull request carefully because I don't fully understand the reason the assertion may fail after commit 2f8eb917768c722b09fdcfacdee839dad3f7068a. However, if we want to relax that assertion, I believe this pull request is correct.

Nexory:
I built this branch and ran it against a file that reaches the assertion. It is
tests/data/color_grid_alpha_nogrid.avif with a single byte changed, 0x06 to 0x05
at offset 379. That byte is the low half of the from_item_ID of the second 'auxl'
box, so item 5 becomes the alpha auxiliary of both grid tiles and item 6 of
neither.

The branch does fix the abort: the file decodes with exit 0, where b6fb186 aborts
at read.c:1655. A trace at the two consumer skip sites shows the expected path,
once each:

    [GRID] cell 0 of 2 UNSET, skipped

What I would like to flag is what the skipped cell leaves behind. Under valgrind
the same binary is clean on the unmodified file, 0 errors from 0 contexts, and
reports 157770 errors from 214 contexts on the modified one:

    Conditional jump or move depends on uninitialised value(s)
       at png_write_row
       by png_write_image
       by avifPNGWrite

So the uninitialised bytes reach the written image, not only an internal buffer.

Comparing the two decoded outputs, the alpha channel differs in 813 of 6400
pixels, rows 64 to 74, and in that band the modified file gives one value, 0,
where the unmodified one has 55 distinct values. I first read that constant as
evidence against uninitialised memory, and two runs do give byte identical
output, but a freshly mapped page reads as 0 as well, so neither observation
rules it out.

I could not work out why that band rather than a whole cell, and the build I used
carries no symbols for an origin trace, so I cannot point at the allocation. You
know the compositing side.

I am happy to send the file and a regression test with it.

Nexory:
I found what the band in my earlier comment was, and it is not about
uninitialised memory. Skipping a cell also moves the tiles after it.

The destination of a tile comes from the loop counter:

    unsigned int rowIndex = tileIndex / info->grid.columns;
    unsigned int colIndex = tileIndex % info->grid.columns;

With a trace at avifDecoderDataCopyTileToImage on the same 2x1 grid file:

    valid file     [COPY] alpha tileIndex=0 grid=2x1 tileCount=2
                   [COPY] alpha tileIndex=1 grid=2x1 tileCount=2
    modified file  [COPY] alpha tileIndex=0 grid=2x1 tileCount=1

The surviving alpha item has dimgIdx 1, so it belongs in cell 1. Once cell 0 is
skipped it is copied as tileIndex 0 and lands in cell 0 instead, and cell 1 is
never written. That is why the upper half of my earlier comparison matched the
unmodified file: the tile is there, just not where the file puts it. The band I
could not place is in the lower half, the cell that stayed empty.

So the gap is not only undefined, the remaining content is silently relocated.
Clearing the plane after allocation makes the output defined, and I measured
that it takes valgrind from 157770 errors to 0, but the tile is still in the
wrong cell, so I do not think that is enough on its own.

Two ways I can see. Either refuse the file, which is what my branch does and it
is ready if you want it. Or carry the cell index through to the copy so a tile
keeps its place, and clear the cells that have no tile.

Happy to test whichever you prefer.

wantehchang:
Hi Nexory,

Thank you for testing and analyzing my patch. I am afraid that I am not qualified to work on this bug. Please upload your pull request with the test file for Vignesh and Yannis to review.

Since our internal fuzzing is reporting a lot of bugs, I may check in a change to supress the bug or roll back your two commits.

wantehchang:
I wrote and merged PR https://github.com/AOMediaCodec/libavif/pull/3340 as a temporary workaround to avoid the assertion failure. I am abandoning this PR.

PR #3340, opened 2026-08-27
Add avifDecoderItemIsAlphaAuxOld as temporary fix

wantehchang:
Avoid the failure of the assertion numTiles == numExpectedTiles in avifFillDimgIdxToItemIdxArray() and allow us to investigate the bug.

Add avifDecoderItemIsAlphaAuxOld() to emulate the old behavior of avifDecoderItemIsAlphaAux() before commit 2f8eb91 and call avifDecoderItemIsAlphaAuxOld() in the loop that builds the dimgIdxToAlphaItemIdx array in avifMetaFindAlphaItem().

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -274,6 +274,20 @@ static avifBool avifReferences(const avifItemReferenceArray * references, const 
     return AVIF_FALSE;
 }
 
+// Bug oss-fuzz:553221196: The same as avifReferences() except that it emulates
+// the old behavior before commit 2f8eb91 and considers only the last reference.
+static avifBool avifReferencesOld(const avifItemReferenceArray * references, const char * type, uint32_t toID)
+{
+    for (uint32_t i = references->count; i > 0;) {
+        --i;
+        const avifItemReference * reference = &references->ref[i];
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -274,6 +274,20 @@ static avifBool avifReferences(const avifItemReferenceArray * references, const
     return AVIF_FALSE;
 }
 
+// Bug oss-fuzz:553221196: The same as avifReferences() except that it emulates
+// the old behavior before commit 2f8eb91 and considers only the last reference.
+static avifBool avifReferencesOld(const avifItemReferenceArray * references, const char * type, uint32_t toID)
+{
+    for (uint32_t i = references->count; i > 0;) {
+        --i;
```
wantehchang, line 282: Note that the index `i` counts down, so we traverse the references array backwards.

At line 285 the last reference, if present, gives the answer.

vigneshvg (approved):

wantehchang:
cc: @Nexory @y-guyon

This is a temporary workaround to prevent the fuzzer from filing more bugs.

PR #3342, opened 2026-08-28
read: keep 'dimg' inputs as an ordered array on the derived item

Nexory:
Draft, because you are weighing whether to drop the synthetic alpha grid
generation altogether. @y-guyon asked to see the array change, so here it is with
the cost visible. If the generation goes away this becomes moot and I will close
it without any fuss.

The 'dimg' box of a derived image item is an ordered array of input item IDs.
libavif stores the reverse: each input item carries the ID of its derived item
and its position in that list, one value each. The list then has to be rebuilt by
scanning every item whenever it is needed, and the inversion cannot express a
value twice.

This keeps the list where the file puts it, on the derived item, and turns the
read sites into membership or position lookups. avifFillDimgIdxToItemIdxArray()
reads the list directly instead of rebuilding it, and its assertion on the tile
count becomes an ordinary check.

On the citation you asked for: section 8.11.12.1 cuts both ways. It says a given
value shall occur at most once within a given array, so repeating a tile inside
one grid stays refused, now checked against the list. It also says a value may
occur within several arrays, which the single valued field could not represent at
all. That is the gap this closes, rather than any permission to reuse a tile.

It also drops avifReferencesOld() and avifDecoderItemIsAlphaAuxOld() from eb67309.
With the list, the synthetic alpha grid can hold one auxiliary item in several
cells, which is what that workaround was avoiding.

Measured: without the change the new test fails, with it the four DimgTest cases
pass, the suite is 63 of 63 both with the change and on an untouched checkout of
main, and clang-format 22.1.8 is clean on both changed files. Not measured: fault
injection, a positive control, and a run in the CI image.

The test file differs from color_grid_alpha_nogrid.avif in one field, the
from_item_ID of the second auxl box, so item 5 is the alpha auxiliary of both
tiles. On current main it decodes with its alpha channel dropped; here it decodes
with the second cell repeating the first.

#3341 is the narrow alternative, a guard that refuses that file instead. Happy to
mirror this in CrabbyAvif if you want it, and happy to drop both if the generation
is removed.


src/read.c
```diff
@@ -219,6 +219,12 @@ typedef struct avifItemReference
 } avifItemReference;
 AVIF_ARRAY_DECLARE(avifItemReferenceArray, avifItemReference, ref);
 
+// The ordered list of input item IDs of a derived image item, in the order given by its
+// 'dimg' box. Section 8.11.12.1 of ISO/IEC 14496-12 allows a given item to occur in the
+// arrays of several derived items, so the list belongs to the derived item rather than
+// to its inputs.
```
wantehchang: Delete the second sentence. It is confusing and does not add useful information to the first sentence.

src/read.c
```diff
@@ -219,6 +219,12 @@ typedef struct avifItemReference
 } avifItemReference;
 AVIF_ARRAY_DECLARE(avifItemReferenceArray, avifItemReference, ref);
 
+// The ordered list of input item IDs of a derived image item, in the order given by its
```
wantehchang: Add "(to_item_ID)" after "input item IDs"

src/read.c
```diff
@@ -237,9 +243,9 @@ typedef struct avifDecoderItem
                                    // If false, mergedExtents is used as an avifROData and points to a
                                    // buffer it doesn't own.
     avifBool partialMergedExtents; // If true, mergedExtents doesn't have all of the item data yet
-    uint32_t dimgForID;            // if non-zero, this item is an input of derived Item #{dimgForID}
-    uint32_t dimgIdx; // If dimgForId is non-zero, this is the zero-based index of this item in the list of Item #{dimgForID}'s dimg.
-    avifBool hasDimgFrom; // whether there is a 'dimg' box with this item's id as 'fromID'
+    avifItemIDArray dimgInputs;    // if this item is derived, its input items in 'dimg' order
+    avifBool isDimgInput;          // whether this item is an input of some derived item
+    avifBool hasDimgFrom;          // whether there is a 'dimg' box with this item's id as 'fromID'
```
wantehchang: `hasDimgFrom` is redundant now because `hasDimgFrom` is equal to `dimgInputs.count != 0`.

src/read.c
```diff
@@ -252,6 +258,39 @@ typedef struct avifDecoderItem
...
+{
+    uint32_t * dimgInput = (uint32_t *)avifArrayPush(&derivedItem->dimgInputs);
+    AVIF_CHECKERR(dimgInput != NULL, AVIF_RESULT_OUT_OF_MEMORY);
+    *dimgInput = inputItem->id;
+    inputItem->isDimgInput = AVIF_TRUE;
+    return AVIF_RESULT_OK;
+}
+
+// Returns the position of inputItemID among derivedItem's 'dimg' inputs, or UINT32_MAX.
+// A given value occurs at most once in a 'dimg' array, so the first match is the only one.
```
wantehchang: IMPORTANT: Is this required by some standard?

Searching for "to_item_ID" in ISO/IEC 14496-12:2022 and ISO/IEC 23008-12:2022, I do not see such a requirement.

src/read.c
wantehchang: IMPORTANT: We should rename `dimg` as `dimgInput` or `inputItem`.

This variable is an input image item. It is not a derived image item.

Note: This is a pre-existing problem, so we can also ignore it in this PR. But I find it very confusing.

src/read.c
```diff
@@ -5700,8 +5722,16 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
+        alphaTileItem->isDimgInput = AVIF_TRUE;
+        // The alpha grid is synthesized here rather than read from a file, so one item
+        // may stand in for several cells: an auxiliary item can be the alpha of more
+        // than one color tile.
+        uint32_t * alphaInput = (uint32_t *)avifArrayPush(&(*alphaItem)->dimgInputs);
+        if (alphaInput == NULL) {
+            avifFree(dimgIdxToAlphaItemIdx);
+            return AVIF_RESULT_OUT_OF_MEMORY;
+        }
+        *alphaInput = alphaTileItem->id;
```
wantehchang: lines 5725 - 5734 should be equivalent to:

```
        result = avifAddDimgInput(*alphaItem, alphaTileItem);
            avifFree(dimgIdxToAlphaItemIdx);
            return result;
        }
```

src/read.c
wantehchang: Add ";  within a given array, a given value shall occur at most once" after "The items linked to are then represented by an array of to_item_IDs".

And then point out that we do not (but should) check this requirement.

src/read.c
```diff
@@ -3477,11 +3512,10 @@ static avifResult avifParseItemReferenceBox(avifMeta * meta, const uint8_t * raw
                 // Section 8.11.12.1 of ISO/IEC 14496-12:
                 //   The items linked to are then represented by an array of to_item_IDs;
                 //   within a given array, a given value shall occur at most once.
-                AVIF_CHECKERR(dimg->dimgForID != fromID, AVIF_RESULT_INVALID_IMAGE_GRID);
+                AVIF_CHECKERR(!avifIsDimgInput(item, toID), AVIF_RESULT_INVALID_IMAGE_GRID);
                 // A given value may occur within multiple arrays but this is not supported by libavif.
-                AVIF_CHECKERR(dimg->dimgForID == 0, AVIF_RESULT_NOT_IMPLEMENTED);
-                dimg->dimgForID = fromID;
-                dimg->dimgIdx = refIndex;
+                AVIF_CHECKERR(!dimg->isDimgInput, AVIF_RESULT_NOT_IMPLEMENTED);
```
wantehchang: IMPORTANT: I thought this PR was intended to address this very limitation. Should we delete this comment and this check (lines 3516-3517)?

src/read.c
Nexory: Done in 99aad10.

src/read.c
Nexory: Done in 99aad10. It said the list belongs to the derived item because a value may occur in several arrays, which belongs next to the code that relies on it, not on the declaration.

src/read.c
Nexory: Done in a1b6fac, and it says the requirement is not checked for these four types.

Nexory: Done in 9e417f7. Only the variable: the box type in the memcmp and the comments about the 'dimg' box keep their name, so the change is four lines.

src/read.c
Nexory: Done in 55513c3, exactly as you wrote it.

src/read.c
Nexory: I tried that and it is not equivalent. A 'dimg' box with reference_count zero sets hasDimgFrom while leaving the list empty, and nothing rejects such a box today. With an empty box before a filled one for the same from_item_ID:

    with hasDimgFrom       BMFF parsing failed, duplicate boxes of type 'dimg'
    with dimgInputs.count  the check does not fire, refused later for another reason

They do agree when the empty box comes second, which is the case I measured first and why I nearly agreed with you. Happy to remove the field if you also want reference_count zero rejected.

src/read.c
Nexory: You are right, and it does lift. I pointed both 'dimg' lists of the existing shared tile file at the same four color tiles, so items 2 to 5 are each an input of two derived items:

    with the check     Not implemented
    without the check  decodes, 80x80, YUV444, alpha present

The reason that file cannot show it as it stands is that its two grids are a color grid and an alpha grid. Their av1C differ by construction, 81 20 00 00 against 81 00 1c 00, so the shared tile stops at the tile consistency check instead.

I can remove the check and add a test file for it. The part I would rather not decide alone is DimgTest.ItemShared: its expectation moves from NOT_IMPLEMENTED to BMFF_PARSE_FAILED and then it no longer tests what its name says. In this PR, or separately?

src/read.c
```diff
@@ -252,6 +256,39 @@ typedef struct avifDecoderItem
...
+static avifResult avifAddDimgInput(avifDecoderItem * derivedItem, avifDecoderItem * inputItem)
+{
+    uint32_t * dimgInput = (uint32_t *)avifArrayPush(&derivedItem->dimgInputs);
+    AVIF_CHECKERR(dimgInput != NULL, AVIF_RESULT_OUT_OF_MEMORY);
+    *dimgInput = inputItem->id;
+    inputItem->isDimgInput = AVIF_TRUE;
+    return AVIF_RESULT_OK;
+}
+
+// Returns the position of inputItemID among derivedItem's 'dimg' inputs, or UINT32_MAX.
```
y-guyon: Suggestion:

```c
// Returns the 0-based position
```

y-guyon: Should it return a `bool` instead? Could `UINT32_MAX` be an index in a conform file?

src/read.c
```diff
@@ -1651,22 +1684,22 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
 // to its corresponding 0-based index in the avifMeta::items array.
 static avifResult avifFillDimgIdxToItemIdxArray(uint32_t * dimgIdxToItemIdx, uint32_t numExpectedTiles, const avifDecoderItem * gridItem)
 {
-    const uint32_t itemIndexNotSet = UINT32_MAX;
+    // The input items of a derived item are kept in 'dimg' order, so a cell index is a
+    // position in that list. Nothing has to be reconstructed and one input can stand in
+    // for several cells.
```
y-guyon: The second sentence can be removed.

src/read.c
```diff
@@ -1651,22 +1684,22 @@ static avifCodecType avifDecoderItemGetGridCodecType(const avifDecoderItem * gri
...
+        const uint32_t inputItemID = gridItem->dimgInputs.id[dimgIdx];
+        uint32_t itemIdx = UINT32_MAX;
+        for (uint32_t i = 0; i < gridItem->meta->items.count; ++i) {
+            if (gridItem->meta->items.item[i]->id == inputItemID) {
+                itemIdx = i;
+                break;
+            }
         }
+        AVIF_CHECKERR(itemIdx != UINT32_MAX, AVIF_RESULT_INVALID_IMAGE_GRID);
+        dimgIdxToItemIdx[dimgIdx] = itemIdx;
```
y-guyon, line 1692: Are item indices constant over a single decode call?  
If so, I wonder if the new `avifItemIDArray dimgInputs` field should also (or only?) store item indices. That way there would be no need for `avifFillDimgIdxToItemIdxArray()`.  
I did not verify my reasoning, this is just an open question.

tests/data/README.md
```diff
@@ -556,6 +556,17 @@ The color and alpha planes are arranged as 2x2 grid items. The color `dimg`
...
+### File [color_grid_alpha_item_shared_in_auxl.avif](color_grid_alpha_item_shared_in_auxl.avif)
+
+License: [same as libavif](https://github.com/AOMediaCodec/libavif/blob/main/LICENSE)
+
+Source: Same as `color_grid_alpha_nogrid.avif`
+
+The color plane is a grid of two tiles and each tile has its own alpha auxiliary
+item. The `from_item_ID` of the second `auxl` box is manually edited from 6 to 5, so
+item 5 is the alpha auxiliary of both tiles and item 6 of neither. Only that one
+field differs from the source file.
```
y-guyon, line 568: Could you also add a test file derived from `color_grid_alpha_grid.avif`, with the same edit (a cell used in two different grids, for example the same cell used for both color and alpha)? I think this PR also fixes this (unlikely) use case.

y-guyon (approved):
Thank you for this PR.

@wantehchang I assume that you are fine with this PR after all, because you reviewed it.

src/read.c
Nexory: Done in 6aeb4850. 6d827287 then folded the function into `avifIsDimgInput()`,
so that wording moved with it.

Nexory: Yes, and it turned out to be more than a signature change. `avifIsDimgInput()`
and `avifDimgInputIndex()` ran the same loop over the same list, and all three
callers that wanted an index had already called `avifIsDimgInput()` on the same
pair first. The sentinel was therefore never observed, and each of those three
sites walked the list twice.

6d827287 gives `avifIsDimgInput()` an optional out parameter for the position
and drops `avifDimgInputIndex()`. Callers that only ask the question pass NULL.
A position can no longer be read without first learning that there is one.

On the other half of your question: `UINT32_MAX` could not have been a valid
index anyway, since `reference_count` is `unsigned int(16)` and a 'dimg' box
holds at most 65535 entries. With the sentinel gone it no longer matters.

src/read.c
Nexory: Removed in bfd661f7.

src/read.c
Nexory: They are, but it is closer than it looks. `meta->items` only ever grows:
`avifArrayPush()` appends, and the five `avifArrayPop(&meta->items)` calls are
all inside `avifMetaFindOrCreateItem()` itself, each rolling back the item
pushed a few lines earlier before returning `AVIF_RESULT_OUT_OF_MEMORY`.
Nothing sorts or removes.

Two things weigh against storing indices. The synthetic alpha grid item is
created at read.c:5702, after the 'dimg' boxes are parsed, so an index list
would be filled at two different points in the decode. And `avifArrayPush()`
reallocates: indices survive that, pointers do not, and storing the ID keeps
the field independent of the array entirely. What would remain of
`avifFillDimgIdxToItemIdxArray()` is the count check.

I would rather not change the representation in this PR now that it is
approved. Happy to try it as a follow-up if you want to see it.

tests/data/README.md
Nexory: That file already exists: `color_grid_alpha_grid_tile_shared_in_dimg.avif`. Its
'dimg' boxes are

    from_item_ID 1 -> 2, 3, 4, 10
    from_item_ID 6 -> 7, 8, 9, 10

so item 10 is a cell of both the color grid and the alpha grid, which is the
case you describe. `DimgTest.ItemShared` covers it.

It still ends in `AVIF_RESULT_NOT_IMPLEMENTED` though, and this PR does not
change that. The expectation in that test is the same at eb673097 and on this
branch, and it passes in both. What refuses the file is read.c:3509,
`AVIF_CHECKERR(!dimgInput->isDimgInput, AVIF_RESULT_NOT_IMPLEMENTED)`: item 10
is already an input of item 1 when it turns up in item 6's array. The ordered
list can represent the case now, but that check is separate from the
representation, and I did not touch it here.

src/read.c
wantehchang: What I meant is that since it is simpler to store the list of to_item_IDs in the derived item, no extra explanation is needed.

Nexory: Sorry for the confusion. The sentence is gone in 99aad10, and nothing moved
either: the one next to the check has been there since d591bb29.

src/read.c
wantehchang: Thank you for the explanation. I didn't consider the reference_count=0 case. Let's keep `hasDimgFrom` then.

I suggest clarifying its comment as follows:
1. Change "fromID" to "from_item_ID"
2. Add "even if reference_count is zero" at the end of the comment.

I also suggest moving `hasDimgFrom` before `dimgInputs` so that the order implies that `dimgInputs` should be used only when `hasDimgFrom` is true. (The new order also matches the order of the corresponding fields in `SingleItemTypeReferenceBox`.)

src/read.c
wantehchang: Nexory: I was wondering what "it does lift" meant. Did it mean "it does work"?

Yannis: Could you take a look at this discussion? We need your expertise.

src/read.c
Nexory: Done in 051f9f5c, all three. isDimgInput stayed where it was, since it is not a
field of that box.

src/read.c
Nexory: Yes, sorry for the wording: I meant the restriction can be lifted. With that
check removed, the file I crafted decodes, 80x80, YUV444, alpha present.

src/read.c
```diff
@@ -5634,13 +5648,14 @@ static avifResult avifMetaFindAlphaItem(avifMeta * meta,
...
                 avifDecoderItem * auxlItem = meta->items.item[j];
-                if (avifDecoderItemIsAlphaAuxOld(auxlItem, item->id)) {
-                    if (seenAlphaForCurrentItem || auxlItem->dimgForID != 0 || item->dimgIdx >= tileCount ||
-                        dimgIdxToAlphaItemIdx[item->dimgIdx] != itemIndexNotSet) {
+                if (avifDecoderItemIsAlphaAux(auxlItem, item->id)) {
+                    if (seenAlphaForCurrentItem || auxlItem->isDimgInput || colorTileIdx >= tileCount ||
+                        dimgIdxToAlphaItemIdx[colorTileIdx] != itemIndexNotSet) {
                         // One of the following invalid cases:
                         // * Multiple items are claiming to be the alpha auxiliary of the current item.
                         // * Alpha auxiliary is dimg for another item.
```
y-guyon: Why is that invalid?

tests/data/README.md
y-guyon: I agree with Wan-Teh, I think we should remove that check in this PR. I also believe `isDimgInput` is no longer necessary afterwards.

src/read.c
Nexory: It was the same restriction seen from the other side: an alpha auxiliary that is
a 'dimg' input was already a cell of some derived item, which the old
single-valued mapping could not express. The condition and that bullet are gone
in 87918daf, together with isDimgInput.

tests/data/README.md
Nexory: Done in 87918daf. The check, that alpha auxiliary condition and isDimgInput are
all gone, and you were right that nothing else used it: it had four occurrences,
now none.

DimgTest.ItemShared moves from NOT_IMPLEMENTED to BMFF_PARSE_FAILED. Item 10 is
a cell of both grids, so the file now reaches the tile consistency check and
stops there, since a color av1C cannot match the alpha ones.

b9eec4c7 adds a file where the case actually decodes: both 'dimg' lists point at
the same four cells, five bytes edited from that same source file. Without the
removal that new test gets NOT_IMPLEMENTED instead of OK.

y-guyon (approved):

y-guyon:
Could you check if CrabbyAvif would benefit from the same change, if applicable, please?

Nexory:
@y-guyon Sure, I'll take a look

Nexory:
Yes, the same shape applies, and CrabbyAvif already has the structure to put it in.

`Item::source_item_ids` is documented as the source item ids "in the same order as
they appear in the `dimg` box", which is what `dimgInputs` became here. It is
derived rather than parsed, though: `populate_source_item_ids()` walks every item,
keeps the ones whose single-valued `dimg_for_id` matches, and sorts by `dimg_index`
to recover the box order.

That single-valued mapping carries the same restriction this PR removed. In
`item.rs` a second `dimg` reference to the same item is rejected, with
`BmffParseFailed` when it comes from the same derived item and `NotImplemented`
when it comes from a different one. The second branch is the one that is no longer
needed.

Filling `source_item_ids` while parsing the box would remove `dimg_for_id`,
`dimg_index`, the walk and the sort. The rule that a value occurs at most once
within one array still needs its own check, the way it does here.

One thing the single-valued field does better, for the record: it answers "which
derived item is this a cell of" in constant time, while walking the list does not.
That cost is real here: of the eleven call sites of `avifIsDimgInput()`, ten sit
inside a loop over the item table, two of them nested in a second one. In CrabbyAvif it would not be paid, because nothing reads
`dimg_for_id` or `dimg_index` outside the check, the assignment and
`populate_source_item_ids()` itself.

Happy to prepare that as a PR if you would like it.

y-guyon:
Please do, thanks!

PR #3343, opened 2026-08-29
Replace avifAlloc(n * sizeof(X)) with avifCalloc(n, sizeof(X)) at call sites lacking overflow checks

tinhien11:
# Replace avifAlloc(n * sizeof(X)) with avifCalloc(n, sizeof(X)) at call sites lacking overflow checks

## Motivation

libavif currently uses two inconsistent patterns for size-multiplication at
allocation sites:

1. Sites with an explicit `SIZE_MAX / sizeof(X)` check before calling `avifAlloc()`.
2. Sites that call `avifAlloc(n * sizeof(X))` directly with no check.

Pattern (2) silently wraps on 32-bit targets (and on unusual `sizeof()` values
in general), which can produce a small allocation followed by out-of-bounds
writes. This is the same class of bug fixed upstream in libjpeg-turbo
(c692fea and related), where integer overflow in a TurboJPEG helper allowed a
buffer overrun.

`calloc()` (C11 §7.22.3.2) is required to verify that its arguments' product
does not overflow before computing it, and to fail if it does. libavif already
exposes `avifCalloc()` internally, so we can migrate pattern (2) call sites to
`avifCalloc()` without adding any new API surface and without paying a
meaningful performance cost (these sites are all parse-time or one-time
allocations, not hot paths).

## Changes

Replaced the following call sites with `avifCalloc(count, elemSize)`:

- `src/avif.c:393` — `numProperties * sizeof(avifImageItemProperty)`.
  (Also dropped the now-redundant explicit `SIZE_MAX /` check, since
  `avifCalloc()` already performs it.)
- `src/read.c:1755` — `numTiles * sizeof(uint32_t)` (dimg-index → item-index).
- `src/read.c:5628` — `tileCount * sizeof(uint32_t)` (dimg-index → alpha item).
- `src/read.c:6122` — `numTiles * sizeof(uint32_t)` (grid alpha lookup).
- `src/reformat.c:580` and `src/reformat.c:591` — `cpCount * sizeof(float)`
  (YUV→RGB unorm→float LUTs).
- `src/reformat_libyuv.c:181` — `num_allocated_rows * src_stride_argb`.
- `src/sampletransform.c:381` — `stackCapacity * sizeof(int32_t)`.
- `src/codec_avm.c:923` — `monoUVSize = monoUVHeight * monoUVRowBytes`
  (dropped the now-unused `monoUVSize` temporary).

## Scope

Not touched:

- Call sites that already compute `n * sizeof(X)` with an explicit
  `SIZE_MAX /` guard (e.g. `src/utils.c:93`, `src/gainmap.c:578`). Those
  remain as-is; consolidating them is possible follow-up work.
- Call sites with a fixed `sizeof(X)` operand only (`avifAlloc(sizeof(T))`) —
  no arithmetic to overflow.
- `memcpy()` sites that propagate already-validated sizes (e.g.
  `src/gainmap.c:126`).

## Security impact

This is a defensive hardening change. It does not fix a specific known
vulnerability, but it removes a systematic source of future integer-overflow
bugs in the AVIF parser path — the category explicitly eligible for the
Google Patch Rewards Program as "secure-by-design memory safety improvements"
on Tier-1 "Core infrastructure data parsers".

## Verification

- `cmake -DCMAKE_BUILD_TYPE=Release -DAVIF_LIBYUV=OFF -DAVIF_LIBSHARPYUV=OFF
  -DAVIF_BUILD_APPS=OFF -DAVIF_BUILD_TESTS=ON ..` — configure OK.
- `make -j8` — build passes with no new warnings on any changed file.
- `ctest` — `avifyuv_limited` and `avifyuv_rgb` (which exercise the changed
  `reformat.c` and `sampletransform.c` allocation sites) pass.
  `aviftest` fails identically with and without this patch when the optional
  codec backends (`AVIF_LIBYUV=OFF`, no aom/dav1d/libgav1) are disabled —
  baseline failure is pre-existing, unrelated to this change.

### DIFF
--- a/src/avif.c
+++ b/src/avif.c
@@ -387,10 +387,11 @@ avifResult avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_
 
 avifResult avifImagePushProperty(avifImage * image, const uint8_t boxtype[4], const uint8_t usertype[16], const uint8_t * boxPayload, size_t boxPayloadSize)
 {
-    AVIF_CHECKERR(image->numProperties < SIZE_MAX / sizeof(avifImageItemProperty), AVIF_RESULT_INVALID_ARGUMENT);
     // Shallow copy the current properties.
     const size_t numProperties = image->numProperties + 1;
-    avifImageItemProperty * const properties = (avifImageItemProperty *)avifAlloc(numProperties * sizeof(properties[0]));
+    // calloc() verifies that numProperties * sizeof() does not overflow size_t
+    // before computing the product, so this is safe even on 32-bit targets.
+    avifImageItemProperty * const properties = (avifImageItemProperty *)avifCalloc(numProperties, sizeof(properties[0]));
     AVIF_CHECKERR(properties != NULL, AVIF_RESULT_OUT_OF_MEMORY);
     if (image->numProperties != 0) {
         memcpy(properties, image->properties, image->numProperties * sizeof(properties[0]));
--- a/src/codec_avm.c
+++ b/src/codec_avm.c
@@ -918,9 +918,10 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
             if (!avmImageAllocated) {
                 uint32_t channelSize = avifImageUsesU16(image) ? 2 : 1;
                 uint32_t monoUVRowBytes = channelSize * monoUVWidth;
-                size_t monoUVSize = (size_t)monoUVHeight * monoUVRowBytes;
 
-                monoUVPlane = avifAlloc(monoUVSize);
+                // calloc() verifies that the product does not overflow size_t
+                // before computing it, so this is safe even on 32-bit targets.
+                monoUVPlane = avifCalloc(monoUVHeight, monoUVRowBytes);
                 AVIF_CHECKERR(monoUVPlane != NULL, AVIF_RESULT_OUT_OF_MEMORY); // No need for avm_img_free() because !avmImageAllocated
                 avmImage.planes[1] = monoUVPlane;
                 avmImage.stride[1] = monoUVRowBytes;
--- a/src/reformat_libyuv.c
+++ b/src/reformat_libyuv.c
@@ -178,7 +178,10 @@ static int avifReorderARGBThenConvertToYUV(int (*ReorderARGB)(const uint8_t *, i
         // allocating more than soft_allocation_limit, unless min_num_rows rows need more than that.
         num_allocated_rows = AVIF_MAX(1, soft_allocation_limit / (src_stride_argb * min_num_rows)) * min_num_rows;
     }
-    src_argb = (uint8_t *)avifAlloc(num_allocated_rows * src_stride_argb);
+    // calloc() verifies that num_allocated_rows * src_stride_argb does not
+    // overflow size_t before computing the product, so this is safe even on
+    // 32-bit targets.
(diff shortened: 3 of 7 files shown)


src/avif.c
```diff
@@ -387,10 +387,11 @@ avifResult avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_
 
 avifResult avifImagePushProperty(avifImage * image, const uint8_t boxtype[4], const uint8_t usertype[16], const uint8_t * boxPayload, size_t boxPayloadSize)
 {
-    AVIF_CHECKERR(image->numProperties < SIZE_MAX / sizeof(avifImageItemProperty), AVIF_RESULT_INVALID_ARGUMENT);
     // Shallow copy the current properties.
     const size_t numProperties = image->numProperties + 1;
-    avifImageItemProperty * const properties = (avifImageItemProperty *)avifAlloc(numProperties * sizeof(properties[0]));
+    // calloc() verifies that numProperties * sizeof() does not overflow size_t
+    // before computing the product, so this is safe even on 32-bit targets.
+    avifImageItemProperty * const properties = (avifImageItemProperty *)avifCalloc(numProperties, sizeof(properties[0]));
```
wantehchang, line 394: In these places `avifAlloc` is used because we don't need to zero-initialize the allocated memory.

src/codec_avm.c
```diff
@@ -918,9 +918,10 @@ static avifResult avmCodecEncodeImage(avifCodec * codec,
             if (!avmImageAllocated) {
                 uint32_t channelSize = avifImageUsesU16(image) ? 2 : 1;
                 uint32_t monoUVRowBytes = channelSize * monoUVWidth;
-                size_t monoUVSize = (size_t)monoUVHeight * monoUVRowBytes;
 
-                monoUVPlane = avifAlloc(monoUVSize);
+                // calloc() verifies that the product does not overflow size_t
+                // before computing it, so this is safe even on 32-bit targets.
+                monoUVPlane = avifCalloc(monoUVHeight, monoUVRowBytes);
```
wantehchang, line 924: Overflow checks are omitted when we can prove overflow is impossible. I give two examples.

Here we can prove the product `monoUVHeight * monoUVRowBytes` cannot overflow as follows:

1. In 4:2:0, the size of the U plane is smaller than the size of the Y plane.
2. The Y plan has been allocated successfully by the caller. Therefore the size of the Y plane can be expressed in `size_t`.
3. It follows that the size of the U plane can be expressed in `size_t`.

src/reformat_libyuv.c
```diff
@@ -178,7 +178,10 @@ static int avifReorderARGBThenConvertToYUV(int (*ReorderARGB)(const uint8_t *, i
         // allocating more than soft_allocation_limit, unless min_num_rows rows need more than that.
         num_allocated_rows = AVIF_MAX(1, soft_allocation_limit / (src_stride_argb * min_num_rows)) * min_num_rows;
     }
-    src_argb = (uint8_t *)avifAlloc(num_allocated_rows * src_stride_argb);
+    // calloc() verifies that num_allocated_rows * src_stride_argb does not
+    // overflow size_t before computing the product, so this is safe even on
+    // 32-bit targets.
+    src_argb = (uint8_t *)avifCalloc(num_allocated_rows, src_stride_argb);
```
wantehchang, line 184: The code above ensures `num_allocated_rows * src_stride_argb` does not overflow `int`.

Issue #3338, opened 2026-08-27
`--yuv` being ignored with `avifgainmaputil combine`


BPplays:
im trying to use the `--yuv`/`-y` option with `avifgainmaputil combine` but it only seems to have an effect for the gain map not the base image.

---

here the base image says `Format : YUV444` but the gain map says: `YUV420`

```powershell
-> avifgainmaputil combine `
>   "C:\Users\censored\AppData\Local\Temp\avifgainmappwsh\a7210c6d-be58-4070-836f-fe94c5c62bcb\thm-tmp-sdr.avif" `
>   "C:\Users\censored\AppData\Local\Temp\avifgainmappwsh\a7210c6d-be58-4070-836f-fe94c5c62bcb\thm-tmp-hdr.avif" `
>   "C:\Users\censored\AppData\Local\Temp\avifgainmappwsh\a7210c6d-be58-4070-836f-fe94c5c62bcb\test-420.avif" `
>   -d 10 `
>   --depth-gain-map 10 `
>   -q 85 `
>   --qgain-map 85 `
>   --qalpha 99 `
>   -y 420 `
>   --yuv-gain-map 420 `
>   --cicp-base "1/13/6" `
>   --cicp-alternate "9/16/9" `
>   --ignore-profile
Creating a gain map of size 1025 x 768
AVIF to be written:
 * Resolution     : 1025x768
 * Bit Depth      : 10
 * Format         : YUV444
 * Alpha          : Absent
 * Range          : Full
 * Color Primaries: 1
 * Transfer Char. : 13
 * Matrix Coeffs. : 6
 * ICC Profile    : Absent
 * XMP Metadata   : Absent
 * Exif Metadata  : Present (180 bytes)
 * Transformations: None
 * Progressive    : Unavailable
 * Gain map       : 1025x768 pixels, 10 bit, YUV420, Full Range, Matrix Coeffs. 2, Base Headroom 0.00 (SDR), Alternate Headroom 1.92 (HDR)
 * Alternate image:
    * Color Primaries: 9
    * Transfer Char. : 16
    * Matrix Coeffs. : 9
    * ICC Profile    : Absent
    * Bit Depth      : 10
    * Planes         : 3

Encoding AVIF with settings: codec 'aom' speed ['6'], color quality ['85' (High)], alpha quality ['99' (High)], gain map quality [85 (High)], automatic tiling, 32 worker thread(s), please wait...
[...]

PR #3346, opened 2026-08-31
avifgainmaputil: Implement YUV 444/420 conversion for AVIF base images.

maryla-uc:
This allows avifgainmaputil to honor the --yuv 420 and --yuv 444 flags even when the input is an AVIF image, so already YUV.
Not all conversions are supported, e.g. to/from 422 or  400. If not supported, a warning is printed so at least it's no longer silent.

It includes both a libyuv integration and a C fallback path for when libavif is compiled without libyuv.

#3338

### DIFF
--- a/apps/avifgainmaputil/imageio.cc
+++ b/apps/avifgainmaputil/imageio.cc
@@ -16,7 +16,235 @@
 #include "avifutil.h"
 #include "y4m.h"
 
+#if defined(AVIF_LIBYUV_ENABLED)
+#include <libyuv.h>
+#endif
+
 namespace avif {
+
+static avifResult UpsampleYUV420To444(avifImage* image) {
+  if (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV420) {
+    return AVIF_RESULT_NOT_IMPLEMENTED;
+  }
+  if (!image->imageOwnsYUVPlanes) {
+    // We need to replace the U and V planes with new, owned planes. Y has to
+    // be owned as well since there is a single ownership boolean for all
+    // three planes. If Y is not already owned, we could copy it, but this is
+    // not needed in practice for now so just reject it.
+    return AVIF_RESULT_NOT_IMPLEMENTED;
+  }
+
+  const uint32_t new_uv_width = image->width;
+  const uint32_t new_uv_height = image->height;
+  const uint32_t old_uv_width = (image->width + 1) / 2;
+  const uint32_t old_uv_height = (image->height + 1) / 2;
+  const uint32_t bytes_per_pixel = (image->depth > 8) ? 2 : 1;
+
+  uint8_t* new_u =
+      (uint8_t*)avifAlloc(new_uv_width * new_uv_height * bytes_per_pixel);
+  uint8_t* new_v =
+      (uint8_t*)avifAlloc(new_uv_width * new_uv_height * bytes_per_pixel);
+  if (!new_u || !new_v) {
+    avifFree(new_u);
+    avifFree(new_v);
+    return AVIF_RESULT_OUT_OF_MEMORY;
+  }
+
+  const uint32_t new_stride_u = new_uv_width * bytes_per_pixel;
+  const uint32_t new_stride_v = new_uv_width * bytes_per_pixel;
+
+#if defined(AVIF_LIBYUV_ENABLED)
+  if (image->depth == 8) {
+    libyuv::ScalePlane(image->yuvPlanes[AVIF_CHAN_U],
+                       image->yuvRowBytes[AVIF_CHAN_U], old_uv_width,
+                       old_uv_height, new_u, new_stride_u, new_uv_width,
+                       new_uv_height, libyuv::kFilterBilinear);
+    libyuv::ScalePlane(image->yuvPlanes[AVIF_CHAN_V],
+                       image->yuvRowBytes[AVIF_CHAN_V], old_uv_width,
+                       old_uv_height, new_v, new_stride_v, new_uv_width,
+                       new_uv_height, libyuv::kFilterBilinear);
+  } else {
+#if LIBYUV_VERSION >= 1774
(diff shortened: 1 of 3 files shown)


apps/avifgainmaputil/imageio.cc
```diff
@@ -16,8 +16,236 @@
...
 
+#if defined(AVIF_LIBYUV_ENABLED)
+#include <libyuv.h>
+#endif
+
 namespace avif {
 
+static avifResult UpsampleYUV420To444(avifImage* image) {
+  if (image->yuvFormat != AVIF_PIXEL_FORMAT_YUV420) {
+    return AVIF_RESULT_NOT_IMPLEMENTED;
```
y-guyon: `INVALID_ARGUMENT`?

apps/avifgainmaputil/imageio.cc
```diff
@@ -16,8 +16,236 @@
...
+  const uint32_t new_uv_width = image->width;
+  const uint32_t new_uv_height = image->height;
+  const uint32_t old_uv_width = (image->width + 1) / 2;
+  const uint32_t old_uv_height = (image->height + 1) / 2;
+  const uint32_t bytes_per_pixel = (image->depth > 8) ? 2 : 1;
+
+  uint8_t* new_u =
+      (uint8_t*)avifAlloc(new_uv_width * new_uv_height * bytes_per_pixel);
+  uint8_t* new_v =
+      (uint8_t*)avifAlloc(new_uv_width * new_uv_height * bytes_per_pixel);
```
y-guyon: Maybe cast to size_t first or check for overflow

apps/avifgainmaputil/imageio.cc
```diff
@@ -234,6 +477,31 @@ avifResult ReadImage(avifImage* image, const std::string& input_filename,
...
+    // Attempt to honor the requested format if it was explicitly asked for and
+    // it differs from the source.
+    if (requested_format != AVIF_PIXEL_FORMAT_NONE &&
+        requested_format != image->yuvFormat) {
+      if (image->yuvFormat == AVIF_PIXEL_FORMAT_YUV444 &&
+          requested_format == AVIF_PIXEL_FORMAT_YUV420) {
+        avifResult scale_res = DownsampleYUV444To420(image);
+        if (scale_res != AVIF_RESULT_OK) {
+          return scale_res;
+        }
```
y-guyon, line 513: Nit: `AVIF_CHECKRES` if available here

y-guyon (approved):

apps/avifgainmaputil/imageio.cc
wantehchang: Yes, we need to calculate the new U and V buffer size in `size_t` and check for overflow. Since the new buffers are larger than the current buffers, we cannot rely on the caller's successful allocation of the current buffers (which implies the current buffer sizes can be represented in `size_t`) and omit the overflow check.

We also need to check for overflow when calculaitng `new_uv_width * bytes_per_pixel` at line 53. That multiplication should be performed in `uint32_t`.

apps/avifgainmaputil/imageio.cc
wantehchang: `AVIF_CHECKRES` is defined in `avif/internal.h`, so it is not available here. (The code in apps/ should only include public headers.)

apps/avifgainmaputil/imageio.cc
```diff
@@ -16,8 +16,236 @@
...
+    // not needed in practice for now so just reject it.
+    return AVIF_RESULT_NOT_IMPLEMENTED;
+  }
+
+  const uint32_t new_uv_width = (image->width + 1) / 2;
+  const uint32_t new_uv_height = (image->height + 1) / 2;
+  const uint32_t bytes_per_pixel = (image->depth > 8) ? 2 : 1;
+
+  uint8_t* new_u =
+      (uint8_t*)avifAlloc(new_uv_width * new_uv_height * bytes_per_pixel);
```
wantehchang: In this function, the new buffer sizes are smaller than the current buffer sizes. We should still calculate the new buffer sizes in `size_t`, but it is safe to omit the overflow check.

Similarly at line 156 it is safe to omit the overflow check; so no change is needed there.

apps/avifgainmaputil/imageio.cc
```diff
@@ -16,8 +16,240 @@
 #include "avifutil.h"
 #include "y4m.h"
 
+#if defined(AVIF_LIBYUV_ENABLED)
+#include <libyuv.h>
+#endif
+
 namespace avif {
 
+static avifResult UpsampleYUV420To444(avifImage* image) {
```
wantehchang: Optional: Since this is a C++ file, we should use an anonymous namespace instead of `static`, and use C++ casts instead of C-style casts.

wantehchang (approved):

apps/avifgainmaputil/imageio.cc
```diff
@@ -16,8 +16,240 @@
...
+                [src_y * (image->yuvRowBytes[AVIF_CHAN_U] / 2) + src_x];
+        ((uint16_t*)new_v)[(size_t)y * (new_stride_v / 2) + x] =
+            ((const uint16_t*)image->yuvPlanes[AVIF_CHAN_V])
+                [src_y * (image->yuvRowBytes[AVIF_CHAN_V] / 2) + src_x];
+      }
+    }
+  }
+#endif
+
+  if (image->imageOwnsYUVPlanes) {
```
wantehchang: Because of the check at line 29, `image->imageOwnsYUVPlanes` must be true here, so this check is not necessary.

Similarly, the assignment `image->imageOwnsYUVPlanes = AVIF_TRUE;` at line 130 is not necessary.

wantehchang (approved):

maryla-uc:
Thanks for the thorough reviews and apologies for the churn.

wantehchang:
Maryla: No worries. It's a shame that we could not use `avifImageAllocatePlanes`, which took care of calculating the buffer sizes securely.

PR #3353, opened 2026-09-10
Use uint32_t for encoder item IDs

krishna28238-arch:
avifEncoderItem, lastItemID, primaryItemID, the 'altr' alternative item ID array and all related identifiers are now uint32_t, matching the unsigned int(32) item IDs of the file format (ISO/IEC 14496-12 Section 8.11.6), so that item IDs no longer wrap around when a file requires more than 65535 items. The 32-bit item ID box variants ('pitm' version 1, 'iloc' version 2, 'iinf' version 1 with 'infe' version 3 entries, 'iref' version 1 and 'ipma' version 1) are not written yet, so avifEncoderFinish() refuses to write a file needing more than 65535 items with AVIF_RESULT_NOT_IMPLEMENTED instead of silently truncating the item IDs.

The 'trak' meta box still writes unsigned int(16) item IDs because it only contains metadata items of image sequences, which are too few to require 32-bit item IDs.

Part of #3337.


src/write.c
```diff
@@ -887,11 +887,11 @@ static avifResult avifEncoderWriteTrackMetaBox(avifEncoder * encoder, avifRWStre
             continue;
         }
 
-        AVIF_CHECKRES(avifRWStreamWriteU16(s, item->id));          // unsigned int(16) item_ID;
-        AVIF_CHECKRES(avifRWStreamWriteU16(s, 0));                 // unsigned int(16) data_reference_index;
-        AVIF_CHECKRES(avifRWStreamWriteU16(s, 1));                 // unsigned int(16) extent_count;
-        AVIF_CHECKRES(avifEncoderItemAddMdatFixup(item, s));       //
-        AVIF_CHECKRES(avifRWStreamWriteU32(s, 0 /* set later */)); // unsigned int(offset_size*8) extent_offset;
+        AVIF_CHECKRES(avifRWStreamWriteU16(s, (uint16_t)item->id)); // unsigned int(16) item_ID;
```
y-guyon:
```c
AVIF_ASSERT_OR_RETURN(item->id <= UINT16_MAX);
```

src/write.c
```diff
@@ -908,7 +908,9 @@ static avifResult avifEncoderWriteTrackMetaBox(avifEncoder * encoder, avifRWStre
         AVIF_ASSERT_OR_RETURN(!item->hiddenImage);
         avifBoxMarker infe;
         AVIF_CHECKRES(avifRWStreamWriteFullBox(s, "infe", AVIF_BOX_SIZE_TBD, 2, 0, &infe));
-        AVIF_CHECKRES(avifRWStreamWriteU16(s, item->id));                             // unsigned int(16) item_ID;
+        // The trak meta box only contains metadata items (Exif/XMP) of image sequences,
+        // which are too few to require 32-bit item IDs.
+        AVIF_CHECKRES(avifRWStreamWriteU16(s, (uint16_t)item->id));                   // unsigned int(16) item_ID;
```
y-guyon:
```c
AVIF_ASSERT_OR_RETURN(item->id <= UINT16_MAX);
```

tests/gtest/avifgridapitest.cc
```diff
@@ -277,5 +277,41 @@ TEST(GridApiTest, DifferentMatrixCoefficients) {
...
+  const std::vector<avifImage*> cell_image_ptrs(128 * 256, cell.get());
+
+  EncoderPtr encoder(avifEncoderCreate());
+  ASSERT_NE(encoder, nullptr);
+  // Worst quality to keep the encoding of the 65536 cells fast.
+  encoder->speed = AVIF_SPEED_FASTEST;
+  encoder->minQuantizer = 63;
+  encoder->maxQuantizer = 63;
+  encoder->minQuantizerAlpha = 63;
+  encoder->maxQuantizerAlpha = 63;
```
y-guyon: Please use quality instead

tests/gtest/avifgridapitest.cc
```diff
@@ -277,5 +277,41 @@ TEST(GridApiTest, DifferentMatrixCoefficients) {
...
+      /*width=*/64, /*height=*/64, /*depth=*/8, AVIF_PIXEL_FORMAT_YUV420,
+      AVIF_PLANES_ALL /* alpha channel needed */);
+  ASSERT_NE(cell, nullptr);
+  // The pixels do not matter but avoid use-of-uninitialized-value errors.
+  testutil::FillImageGradient(cell.get());
+  const std::vector<avifImage*> cell_image_ptrs(128 * 256, cell.get());
+
+  EncoderPtr encoder(avifEncoderCreate());
+  ASSERT_NE(encoder, nullptr);
+  // Worst quality to keep the encoding of the 65536 cells fast.
```
y-guyon, line 301: Side note: https://github.com/AOMediaCodec/libavif/pull/2381 could help in this case

src/write.c
y-guyon, line 890: For each remaining u32>u16 cast in this change after supporting 32-bit `iloc` etc., please add either:
- `AVIF_ASSERT_OR_RETURN(item->id <= UINT16_MAX);` if there is no way it could be `>UINT16_MAX`,
- `AVIF_CHECK(item->id <= UINT16_MAX, AVIF_RESULT_NOT_IMPLEMENTED);` otherwise.

src/write.c
```diff
@@ -3158,6 +3160,17 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
...
+    if (encoder->data->lastItemID > UINT16_MAX) {
+        // The 32-bit item ID box variants ('pitm' version 1, 'iloc' version 2, 'iinf' version 1 with
+        // 'infe' version 3 entries, 'iref' version 1 and 'ipma' version 1) are not written by
+        // avifEncoderFinish() yet, and item ID 0 is invalid (ISO/IEC 14496-12 Section 8.11.1.1), so
+        // refuse to write a file needing more than 65535 items rather than silently truncating IDs.
+        avifDiagnosticsPrintf(&encoder->diag,
+                              "cannot write a file with %u items because the 32-bit item ID box variants are not supported yet",
+                              encoder->data->lastItemID);
+        return AVIF_RESULT_NOT_IMPLEMENTED;
+    }
```
y-guyon, line 3172: I do not see the benefit of this change without supporting 32-bit item ID box variants. Please either:
- Close this PR and add a check in `avifEncoderDataCreateItem()` instead, to make sure `lastItemID` does not overflow (if that is the correct fix for the issue described in the added test)
- Support 32-bit item ID box variants

y-guyon (commented):
Thank you for this change.

src/write.c
krishna28238-arch: Thank you for the review! Closing this PR as suggested: switching to uint32_t indeed
brings no benefit without writing the 32-bit item ID box variants.

I followed your guidance in #3352 instead: avifEncoderDataCreateItem() now returns
AVIF_RESULT_NOT_IMPLEMENTED when lastItemID would overflow the 16-bit item ID space.
Its signature was changed to return an avifResult through an out-pointer so that the
error code is properly propagated (also fixing latent NULL-dereferences on allocation
failure at three call sites). The upfront check in avifEncoderAddImageItems() is kept
so that oversized grids are rejected before any cell is encoded, keeping the test fast
(nothing gets encoded, so the worst-quality settings are no longer needed either).

Noted on the side notes: thank you for the pointers to quality and to #2381, kept in
mind for future tests. Supporting the 32-bit item ID box variants would be a nice
follow-up but is out of scope for now.

PR #3354, opened 2026-09-10
Reject grids exceeding the 'dimg' reference count limit

krishna28238-arch:
avifEncoderAddImageGrid() now refuses grids with more than 65535 cells (e.g. 256x256 cells) upfront with AVIF_RESULT_INVALID_IMAGE_GRID and a diagnostic: Section 8.11.12.1 of ISO/IEC 14496-12 requires all the references from one item to be collected into a single item type reference box whose reference_count field is unsigned int(16) whatever the 'iref' version, and ISO/IEC 23008-12 (HEIF) Section 6.6.1 forbids more than one 'dimg' box with the same from_item_ID. Such grids therefore cannot be encoded in a conformant way, whatever the item ID width, and used to silently generate a broken file instead.

Part of #3337.


y-guyon:
Thank you for this change. Is there the same issue in CrabbyAvif? If so, would you mind sending a similar patch there?

src/write.c
```diff
@@ -2154,6 +2154,20 @@ avifResult avifEncoderAddImageGrid(avifEncoder * encoder,
     if ((gridCols == 0) || (gridCols > 256) || (gridRows == 0) || (gridRows > 256)) {
         return AVIF_RESULT_INVALID_IMAGE_GRID;
     }
+    if ((uint64_t)gridCols * gridRows > 65535) {
```
wantehchang, line 2157: Thanks for the fix.

Nit: Because of the check at line 2154, `gridCols * gridRows` is at most 256 * 256 = 65536, so the `uint64_t` cast is not necessary. I will take care of this.

PR #3355, opened 2026-09-10
Avoid freeing planes the image does not own

seonglae:
Three public entry points free planes they do not own.

`avifImageCopy(dst, src)` where `src` is a view into `dst`'s own planes frees the destination first and then reads the source, which is a use-after-free. `avifImageSetViewRect()` has the same shape. `avifImageAllocatePlanes()` on an image whose planes are a non-owning view calls `avifFree()` on borrowed memory.

The fix keeps ownership explicit rather than inferring it from the pointer: the copy detects the alias and copies through the source before releasing anything, `SetViewRect` rejects a view into destination-owned planes, and allocation no longer frees planes the image does not own.

Each commit carries its regression. `avifimagetest` covers the copy and view aliases and the unrelated-owning-destination case that must keep working, `avifallocationtest` covers allocating over a view, and a FuzzTest composes the operations. All eight `AvifImageTest` cases pass; the three `EncodingTest` failures in `avifallocationtest` are present on an unpatched tree too, since this build has no codec.

Every touched file is clean against the project's `.clang-format`. Reported privately first; happy to hold this until the maintainers prefer it public.


seonglae:
Could a maintainer approve the workflow runs? None have run on this PR, as it is my first here.

include/avif/avif.h
```diff
@@ -870,8 +871,10 @@ AVIF_API avifResult avifImageSetMetadataExif(avifImage * image, const uint8_t *
 AVIF_API avifResult avifImageSetMetadataXMP(avifImage * image, const uint8_t * xmp, size_t xmpSize);
 
 // Allocate/free/steal planes. These functions ignore the gainMap field.
-AVIF_API avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags planes); // Ignores any pre-existing planes
-AVIF_API void avifImageFreePlanes(avifImage * image, avifPlanesFlags planes);           // Ignores already-freed planes
+// Allocates missing planes. A complete set of pre-existing planes is left unchanged, including its ownership.
+// Incomplete non-owned YUV planes are rejected because ownership is tracked for the whole YUV plane set.
+AVIF_API avifResult avifImageAllocatePlanes(avifImage * image, avifPlanesFlags planes);
```
wantehchang, line 876: seonglae: Thank you for the pull request. This PR defends against programming errors when using these three functions. For avifImageCopy() and avifImageSetViewRect(), the problem is caused by the lifetime of the image that a view rect points to. View rects need to be used carefully. I can add a warning to the avifImageSetViewRect() function.

The issue with avifImageAllocatePlanes() is thought-provoking and I have analyzed its uses. I will think about how to best handle this function.

In the interest of time, I will treat this PR as a bug report and write my own PR for the avifImageAllocatePlanes() function.

PR #3357, opened 2026-09-10
Support the 32-bit item ID box variants for files with more than 6553…

krishna28238-arch:
…5 items

The encoder previously used 16-bit item IDs, so encoding a file containing more than 65535 items (e.g. a grid image with an alpha channel) silently wrapped the item IDs around and generated an invalid file.

Item IDs are now uint32_t internally. When any item ID exceeds UINT16_MAX, avifEncoderFinish() and avifEncoderWriteTrackMetaBox() write the 32-bit item ID variants of the boxes with item ID fields: 'pitm' version 1, 'iloc' version 2, 'iinf' version 1, 'infe' version 3, 'iref' version 1 and 'ipma' version 1. All these variants were already supported when reading.

Files with 65535 items or fewer are unchanged (byte-identical output). avifEncoderDataCreateItem() returns NULL instead of wrapping around if the 32-bit item ID space is exhausted.


src/write.c
y-guyon, line 878: largeItemIDs |= (metadataItemCount > UINT16_MAX);

src/write.c
```diff
@@ -288,8 +288,15 @@ static avifEncoderData * avifEncoderDataCreate(void)
...
 
+// Returns NULL if a memory allocation failed, or if no item ID is left (item IDs span a
+// 32-bit space, so the latter would require more than 4 billion items).
 static avifEncoderItem * avifEncoderDataCreateItem(avifEncoderData * data, const char * type, const char * infeName, size_t infeNameSize, uint32_t cellIndex)
 {
+    if (data->lastItemID == UINT32_MAX) {
+        // All item IDs are taken. This is unreachable in practice but avoid a wrap-around of
+        // item IDs, which would silently truncate or duplicate them.
+        return NULL;
+    }
```
y-guyon, line 299: Make that an `AVIF_ASSERT_OR_RETURN()` since it is an invariant. The comment can remain.

src/write.c
```diff
@@ -3347,26 +3361,50 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
     // -----------------------------------------------------------------------
     // Write pitm
 
+    // Use the 32-bit item ID variants of the 'pitm', 'iloc', 'iinf', 'infe', 'iref' and
+    // 'ipma' boxes below if any item ID exceeds the 16-bit item ID space, i.e. if there are
+    // more than 65535 items (item IDs are sequential and start at 1). Reading these box
+    // variants is already supported by libavif.
```
y-guyon, line 3367: Do not include commit-like sentences like "Reading these box variants is already supported by libavif" in comments.

Same with other sentences containing "now" in this PR.

src/write.c
```diff
@@ -3347,26 +3361,50 @@ avifResult avifEncoderFinish(avifEncoder * encoder, avifRWData * output)
...
+    const avifBool largeItemIDs = (encoder->data->lastItemID > UINT16_MAX);
```
y-guyon, line 3368:
```c
const avifBool largeItemIDs = encoder->data->lastItemID > UINT16_MAX || encoder->data->items.count > UINT16_MAX;
```

or the following below

```c
AVIF_ASSERT_OR_RETURN(largeItemIDs || encoder->data->items.count <= UINT16_MAX);
```

There should not be silent truncation like `(uint16_t)encoder->data->items.count` without an extra check in my opinion. I guess it is fine as is for `primaryItemID`, it is obvious enough that it cannot be greater than `lastItemID`.

tests/gtest/avifgridapitest.cc
```diff
@@ -221,6 +241,163 @@ TEST(GridApiTest, CellsOfDifferentDimensions) {
...
+  // cell items), which does not fit in the 16-bit item ID space (item ID 0 is
+  // also invalid, see ISO/IEC 14496-12 Section 8.11.1.1). The encoder now
+  // writes the 32-bit item ID variants of the boxes with item ID fields
+  // ('pitm' version 1, 'iloc' version 2, 'iinf' version 1, 'infe' version 3,
+  // 'iref' version 1 and 'ipma' version 1) so that such files are valid.
+  // The cells are 64x64 because it is the smallest size allowed by MIAF.
+  // Note: Most of the time spent in this test is the encoding of the 65536
+  // cells. The generated file is not parsed with avifDecoderParse() because
+  // that would cost minutes with the current quadratic lookups for this many
+  // items; walking the boxes below is enough to check what was written.
```
y-guyon, line 256: Was that quadratic lookup introduced recently?

y-guyon (commented):
Thank you for this change.

src/write.c
```diff
@@ -288,19 +288,22 @@ static avifEncoderData * avifEncoderDataCreate(void)
...
-{
-    if (data->lastItemID == UINT32_MAX) {
-        // All item IDs are taken. This is unreachable in practice but avoid a wrap-around of
-        // item IDs, which would silently truncate or duplicate them.
-        return NULL;
-    }
+// Returns AVIF_RESULT_OUT_OF_MEMORY if a memory allocation failed, or AVIF_RESULT_INTERNAL_ERROR
+// if no item ID is left (item IDs span a 32-bit space, so the latter would require more than
+// 4 billion items).
+static avifResult avifEncoderDataCreateItem(avifEncoderItem ** itemPtr,
```
y-guyon: Move `avifEncoderItem ** itemPtr` arg last (output arg)

y-guyon (approved):

y-guyon (approved):

PR #3358, opened 2026-09-10
Remove uint64_t cast for gridCols * gridRows

wantehchang:
Since gridCols <= 256 and gridRows <= 256, the gridCols * gridRows product can be represented in uint32_t, so the uint64_t cast is not necessary.

Note: We can alternatively test (gridCols == 256) && (gridRows == 256).


wantehchang:
cc: @krishna28238-arch FYI.

y-guyon (approved):

PR #3360, opened 2026-09-11
gainmap.c: Omit unneeded avifImageAllocatePlanes

wantehchang:
Omit the unneeded avifImageAllocatePlanes() call in avifRGBImageComputeGainMap(). The avifImage is passed to avifImageRGBToYUV(), which will call avifImageAllocatePlanes().


src/gainmap.c
wantehchang, line 820: This `avifImageRGBToYUV()` call will call `avifImageAllocatePlanes()`.

PR #3364, opened 2026-09-14
Remove tests/avifoptionstest.c

wantehchang:
tests/avifoptionstest.c was added in
https://github.com/AOMediaCodec/libavif/pull/3350, but I don't have time to make it conform to the style of the tests we wrote ourselves. The test also verifies properties of the current implementation, such as the initial capacity of 4 and the fact that we grow an avif array by doubling. Some of that may be necessary so that we can count the number of codec-specific options we need to add before the avif array is reallocated, but it could make the test a maintenance burden.


wantehchang:
cc: @agamal026 FYI.

y-guyon (approved):

PR #1069, opened 2022-08-26
Support encoding layered image with pre-scaled inputs

tongyuantongyu:
A layered image may use different frame for each layer, like using a small and blurred version as the first layer. It is more efficient and flexible to directly encode the small image, instead of scale it up only for the encoder to scale it back down; and it allows the scale ratio to be smaller than 1/8 as well.

Only single cell layered image is supported in this version. Layered grid may be revisited in the future.

PR #3367, opened 2026-09-17
Use output_width/height in avifJPEGCopyPixels()

wantehchang:
Use cinfo->output_width/height instead of cinfo->image_width/height in avifJPEGCopyPixels() because cinfo->output_width/height is the resolution of the pixels jpeg_read_raw_data() produces.

Suggested by Yuan Tong in
https://github.com/AOMediaCodec/libavif/pull/1069#discussion_r4034202339.


vigneshvg (approved):

apps/shared/avifjpeg.c
wantehchang, line 108: cc: @tongyuantongyu

Vignesh: Note that `cinfo->output_height` is used here to guard the `jpeg_read_raw_data()` call.

tongyuantongyu: ref: https://github.com/libjpeg-turbo/libjpeg-turbo/blob/0431f8047c912a0ea5d32580318686666d1cc704/src/jdapistd.c#L731-L734

Issue #3344, opened 2026-08-30
Inconsistent quality <-> quantizer mapping cause `tune=iq` + `end-usage=cbr/vbr` encoding to tail


tongyuantongyu:
This encoding command can fail with 

```
$ avifenc --min 32 --max 34 -a end-usage=cbr input.png out.avif
  ERROR: Failed to encode image: Encoding of color planes failed
  Diagnostics:
   * aom_codec_enc_init() failed: Invalid parameter: rc_min_quantizer out of range [..cfg->rc_max_quantizer]
```

---
                                                                                                                                                                                                                                                                                                                          
When no explicit `-q` is given, libavif synthesize one using a generic linear mapping (very early, before encoder has a chance to say anything):

https://github.com/AOMediaCodec/libavif/blob/02e61469422e51c9dbb37b44c4d1b12556aeafdb/src/write.c#L1254-L1269

then the quality is mapped back to quantizer in aom encoder, this time using a custom mapping when `tune=iq`:

https://github.com/AOMediaCodec/libavif/blob/02e61469422e51c9dbb37b44c4d1b12556aeafdb/src/codec_aom.c#L643-L654

which may give a `quantize` value outside the original `[minQuantize, maxQuantize]` range:

https://github.com/AOMediaCodec/libavif/blob/02e61469422e51c9dbb37b44c4d1b12556aeafdb/src/codec_aom.c#L981-L982

In the example `quantize` turns out to be `39`, so `minQuantizer = quantizer - 4 = 35` becomes larger than `maxQuantize=34`, resulting in the failure.

---

A simple fix would be to clamp `quantize` back to the `[minQuantize, maxQuantize]` range, but this reveals a fundamental inconsistency in how we handle the `quality` / `quantizer` signaling. It may worth a refactor to make a correct design.

wantehchang:
cc: @juliobbv-p 

Thanks for the bug report.

Julio, do you have time to look into this?

juliobbv-p:
Hi @wantehchang, thanks for the heads-up.

A question: `--min` and `--max` have been deprecated in libavif for a good while, right? The reason is: directly controlling quantizer limits goes against the ethos of tune=iq -- to let the encoder determine the best quantizers for the content to hit a given target quality. In this case, just "fixing" the quality<->quantizer comparison above wouldn't bring anything to the table.

[...]

PR #3368, opened 2026-09-17
Set quality instead of quantizer in fuzz test

tongyuantongyu:
Use `quality` instead of the now-deprecated  `quantizer` config.

This prevents the issue described in #3344 happen in fuzz tests. #1069 added fuzzing for encoding layered images, which uses CBR mode by default and triggers the problematic code path.


tests/gtest/avif_fuzztest_helpers.h
```diff
@@ -261,23 +261,16 @@ inline auto ArbitraryAvifEncoder() {
...
   const auto tile_cols_log2 = fuzztest::InRange(0, 6);
   // Fuzz only a small range of 'speed' values to avoid slowing down the fuzzer
   // too much. The main goal is to fuzz libavif, not the underlying AV1 encoder.
   const auto speed = fuzztest::InRange(6, AVIF_SPEED_FASTEST);
-  return fuzztest::Map(CreateAvifEncoder, codec_choice, max_threads,
-                       min_quantizer, max_quantizer, min_quantizer_alpha,
-                       max_quantizer_alpha, tile_rows_log2, tile_cols_log2,
-                       speed);
+  return fuzztest::Map(CreateAvifEncoder, codec_choice, max_threads, quality,
+                       quality_alpha, tile_rows_log2, tile_cols_log2, speed);
```
wantehchang, line 272: Yuan: There is a compilation error here. I am looking into this.

wantehchang: I found the problem. I will take care of it.

wantehchang (approved):
LGTM. Thanks!

PR #3371, opened 2026-09-18
ci: attest release artifacts

mcc0nnell:
## Summary

Add GitHub/SLSA build provenance attestations for the Linux, macOS, and Windows release artifact ZIPs.

Each release workflow already builds a deterministic local ZIP and uploads that exact file to the GitHub release. This change grants the workflows the OIDC/attestation permissions they need and attests each ZIP immediately before upload.

No artifact contents, filenames, build flags, or upload paths change.

## Current gap

The current v1.4.2 release publishes:

- `linux-artifacts.zip`
- `macOS-artifacts.zip`
- `windows-artifacts.zip`

GitHub records SHA-256 digests for all three, but the release artifacts have no GitHub provenance attestations. For example, `gh attestation verify` for the current Linux and macOS ZIPs returns 404.

## Change

Each artifact workflow now:

1. creates the release ZIP exactly as before;
2. generates build provenance for that exact ZIP with `actions/attest-build-provenance`;
3. uploads the same ZIP to the existing release.

The attestation action is pinned to the full commit for v3.

## Validation

- current v1.4.2 release ZIPs have recorded SHA-256 digests
- current Linux/macOS release ZIPs return no existing GitHub provenance attestation
- `actionlint` on all three changed workflows: PASS
- `git diff --check`: PASS
- no competing provenance PR found


y-guyon:

https://github.com/actions/attest-build-provenance#usage says:

> new implementations should use actions/attest instead

PR #3372, opened 2026-09-18
Reject item-location boxes that over-allocate memory

matejsmycka:
### Problem

`avifParseItemLocationBox()` reads a 16-bit `extent_count` per item and pushes that many `avifExtent` records without checking the count against the bytes remaining in the box. The `iloc` size fields may be zero, and `avifROStreamReadUX8()` consumes no input for a zero size factor, so each ~6-byte item can allocate ~1 MiB. This runs inside `avifDecoderParse()` with no AV1 payload and no codec linked, and none of the decoder limits gate it. The cost is paid while the file is ultimately rejected.

### Fix

Before the extent loop, compute the per-extent wire size, reject a non-zero `extent_count` whose extents would consume no input, and reject any `extent_count` whose total wire size exceeds the bytes left in the box.

### Testing

Built parse-only from a clean checkout, patched vs unpatched, with a harness calling `avifDecoderParse()` and reporting peak RSS:

| Input (12,373-byte iloc, 2048 items x extent_count 65535, zero size fields) | Result | Peak RSS |
|---|---|---|
| unpatched | rejected late after allocating | 2.11 GB |
| patched | rejected early with a parse error | 1.9 MB |

Valid AVIFs, including files with genuine multi-byte extents, still parse unchanged.

This is availability-only (no memory corruption, no information disclosure).


src/read.c
```diff
@@ -2154,6 +2154,9 @@ static avifResult avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw,
         AVIF_CHECKERR(avifROStreamReadUX8(&s, &baseOffset, baseOffsetSize), AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(base_offset_size*8) base_offset;
         uint16_t extentCount;
         AVIF_CHECKERR(avifROStreamReadU16(&s, &extentCount), AVIF_RESULT_BMFF_PARSE_FAILED); // unsigned int(16) extent_count;
+        const uint64_t bytesPerExtent = (uint64_t)(((version == 1 || version == 2) ? indexSize : 0)) + offsetSize + lengthSize;
```
wantehchang, line 2157: Nit: since `indexSize` is initialized to 0, we can omit the `(version == 1 || version == 2)` check.

wantehchang, line 2157: It is not necessary to use the `uint64_t` type for the sum of `indexSize`, `offsetSize`, and `lengthSize`. All three variables are <= 8, so their sum is at most 24.

src/read.c
wantehchang, line 2194: I found that we don't handle a zero `extentLength` correctly. ISO BMFF says that if the value of `exteng_length` is 0, then the length of the extent is the length of the entire referenced container.

src/read.c
```diff
@@ -2154,6 +2154,9 @@ static avifResult avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw,
...
+        AVIF_CHECKERR(extentCount == 0 || bytesPerExtent > 0, AVIF_RESULT_BMFF_PARSE_FAILED);
```
wantehchang, line 2158: It seems that we should also allow `bytesPerExtent` to be zero when `extentCount == 1`. Please see my comment at line 2194.

src/read.c
```diff
@@ -2154,6 +2154,9 @@ static avifResult avifParseItemLocationBox(avifMeta * meta, const uint8_t * raw,
...
+        AVIF_CHECKERR((uint64_t)extentCount * bytesPerExtent <= avifROStreamRemainingBytes(&s), AVIF_RESULT_BMFF_PARSE_FAILED);
```
wantehchang, line 2159: Since `bytesPerExtent` is at most 24, it is not necessary to cast `extentCount` to `uint64_t` because the usual integer promotion to `int` is large enough for the product. Or we can cast to `uint32_t` or `size_t` (the return type of `avifROStreamRemainingBytes()`.

Also, from a security point of view, this check is less important than the `bytesPerExtent > 0` check, because it is the zero `bytesPerExtent` case that causes us to allocate a disproportionate amount of memory relative to the input size. So we could omit this check.

matejsmycka:
Thanks for the quick review Wan-Teh; feedback added

